PR #591, opened 2016-06-04
Add NetStandard support

am11:
Addresses #568.

This delta adds a solution file, `sln/Microsoft.OData.NetStandard.sln`,
which references six `.xproj` files in the following directories:
- NetStandard.Projects/src/Microsoft.OData.Client
- NetStandard.Projects/src/Microsoft.OData.Core
- NetStandard.Projects/src/Microsoft.OData.Edm
- NetStandard.Projects/src/Microsoft.Spatial
- NetStandard.Projects/test/EndToEndTests/Framework
- NetStandard.Projects/test/EndToEndTests/Services/TestServices

(basically all the project covered by
`sln/Microsoft.OData.E2E.Portable.sln`)

To build this project with VS 2015 and dotnet CLI Preview 1, open the
`sln/Microsoft.OData.NetStandard.sln` and build (`Ctrl+Shift+B`).

To build via command line on Windows or Unix, installed dotnet CLI
Preview 1 (from https://dot.net/) and run the following commands:

``` sh

git clone https://github.com/am11/odata.net -b feature/netstd-port
cd odata.net

dotnet restore
dotnet build NetStandard.Projects/test/EndToEndTests/Services/TestServices
```

Building `test/EndToEndTests/Services/TestServices` will trigger the
build on other five projects as TestServices is the most dependent one
in the given set.

The third way is to use latest msbuild (v14.0 or xplat for Unix:
https://github.com/microsoft/msbuild/tree/xplat) and execute:

``` sh
git clone https://github.com/am11/odata.net -b feature/netstd-port
cd odata.net

dotnet restore
msbuild sln/Microsoft.OData.NetStandard.sln
```

PR #774, opened 2017-03-17
Adding .NET Standard 1.1 support for ODataV4-7.x

AlanWong-MS:
### Description
This PR adds separate .csproj files for the ODataV4-7.x libraries built with .NET Standard 1.1, and they are gathered with a newly added Microsoft.Test.OData.DotNetStandard.sln file. The intent is to have both PCL and .NET Standard libraries shipped in NuGet packages; this will help ease the transition.

- [x] *Test cases added.*
- [x] *Build and test with one-click build and test script passed (around six tests fail intermittently and are not due to the effects of this changelist)*

### Additional work necessary
N/A: the user should be able to run build.cmd and all solutions should be able to build out of the box.


AlanWong-MS:
I accounted for this existing issue with the [project.json/Visual Studio reference](https://github.com/aspnet/Tooling/issues/245) error by adding one project.lock.json file to the Microsoft.OData.E2E.DotNetCore.xproj directory. Hopefully others won't have issues with Visual Studio not being able to locate dependencies.

README.md
```diff
@@ -30,7 +28,7 @@ For these libraries, we accept bug reports and pull requests. The corresponding
 
 **ODataV4-7.x branch**
 
-The ODataV4-7.x branch includes the .NET libraries for OData V4 only that are in development now. It is evolved from ODataV4-6.x with kinds of improvements, new features and so on. It is in-compatible with ODataV4-6.x release and will become the master branch after it is released. It has same libraries as ODataV4-6.x branch.
+The ODataV4-7.x branch includes .NET Standard 1.1 libraries for OData V4 only that are in development now. This branch builds upon ODataV4-6.x with new features with the same libraries but is incompatible with ODataV4-6.x--it will become the master branch after it is released. Due to the new use of `project.json`, there is a change in the build process for those who are unfamiliar. Please either 1) run `nuget restore Microsoft.Test.OData.DotNetStandard.sln` in the commandline of the `odata.net/sln` directory, or 2) enable automatic package restore via Visual Studio to build them (in VS2015, go to `Tools->NuGet Package Manager->Package Manager Settings`. Enable "Allow NuGet to download missing packages".). More info regarding the command can be found [here](https://docs.microsoft.com/en-us/nuget/consume-packages/package-restore).
```
AlanWong-MS: Need to revise statement. It can build both PCL and .NET Standard. Also rewrite some sentences for easier understanding.

am11:
@AlanWong-MS, thank you so much for moving this forward! 🎉 
Azure Table Storage, Azure Media Service and many other downstream packages that rely on OData would able to provide support for pure .NET Core support on Unix platform by this effort. In the current release of OData packages, we need to take dependency on mono PCL to run on Unix platforms (which is definitely not a non-blocker, but bit inconvenient IMO).

Now that .NET Core SDK is supported by VS2017 and .NET CLI, would it make sense to do the direct transition to CSPROJ based project instead of taking the XPROJ route? I have done the port-to-NetStd work in some projects, I would be happy to take a stab at it and provide the CSPROJ based implementation. The solution would be able to build on VS Code, VS 2017 for Mac (preview) and VS2017 {any flavor} on Windows as well as the dotnet CLI v1.0 (that ships with VS2017), MSBuild15 and nuget4. However, that SLN wouldn't be able to build on VS2015 as (CPS based) CSPROJ schema has been changed, as it provides build for multiple frameworks with a simplified structure.

Please note, (if I remember it correctly) that dotnet-cli v1.0.0-preview3 is the last version that supports project.json / xproj based projects. Starting from preview4 and now v1.0.0 (RTM) of dotnet-cli would not support project.sjon/xproj based projects. Hence this new-csproj transition is somewhat rendered as mandatory at this point.

Some information can be found here: https://docs.microsoft.com/en-us/dotnet/articles/core/tools/project-json-to-csproj

Please let me know if I can contribute. :)

AlanWong-MS:
@am11, thank you for your contributions and support. :)

Thanks for your comment as well as your previous PR (#591) related to this. We did a thorough review of your PR and we ultimately decided to create a separate one for two main reasons: 1) we're going to target .NET Standard 1.1 to enable a wider range of support and 2) we’re trying to maintain capability with our existing test infrastructure so we went with .csproj/project.json files for .NET Standard (whereas yours went the .xproj route) to be consistent with our PCL versions.

> Now that .NET Core SDK is supported by VS2017 and .NET CLI, would it make sense to do the direct transition to CSPROJ based project instead of taking the XPROJ route?

We’re just as excited as everyone else with the release of VS2017 and would like to onboard to the new product as soon as we can. However, we’re still in the process of getting VS2017 set up on our machines and build infrastructure (some migrations from older systems left to take care of). We do hear you on the points you made regarding VS2017’s advantages—we’re trying to get to the .NET Standard checkpoint first so VS2015 is an intermediate step for us on the way to VS2017.

Having said that, you are one of our amazing pioneers of transitioning to .NET Standard, and we’d welcome your help again in assisting the community on moving to VS2017. :) We don’t have a solid timeframe at this point, but it’s definitely on our radar.

Thank you!

README.md
```diff
@@ -30,7 +28,9 @@ For these libraries, we accept bug reports and pull requests. The corresponding
 
 **ODataV4-7.x branch**
 
-The ODataV4-7.x branch includes the .NET libraries for OData V4 only that are in development now. It is evolved from ODataV4-6.x with kinds of improvements, new features and so on. It is in-compatible with ODataV4-6.x release and will become the master branch after it is released. It has same libraries as ODataV4-6.x branch.
+The ODataV4-7.x builds upon the [ODataV4-6.x branch](https://github.com/OData/odata.net/tree/ODataV4-6.x) and produces both [PCL (Portable Class Libraries) Profile259](https://msdn.microsoft.com/library/gg597391.aspx) and [.NET Standard 1.1](https://docs.microsoft.com/en-us/dotnet/articles/standard/library) libraries. Following ODataV4-6.x, this branch implements new features with the same libraries and will become the master branch after release.
```
robward-ms: The current projects are Profile111, not 259.

tools/CustomMSBuild/Build.props
```diff
@@ -38,6 +38,9 @@
 
     <!-- This ensures OutputPath always matches OutDir, especially when TFS overrides OutDir. -->
     <OutputPath>$(OutDir)</OutputPath>
+    
+    <!-- Setting this variable to false preserves the OutputPath set above. Without this variable, our OutputPath would eventually be overwritten with a different one in the Microsoft.Common.CurrentVersion.Target script. -->
```
AlanWong-MS: Done.

src/Microsoft.OData.Edm/Csdl/Semantics/IEdmReferentialConstraint.cs
```diff
@@ -1,26 +0,0 @@
-﻿//---------------------------------------------------------------------
```
AlanWong-MS: Will add consolidated descriptions of the commits into the overall description when we do a squash PR.

src/Microsoft.OData.Client/Build.NetStandard/project.json
```diff
@@ -0,0 +1,14 @@
+﻿{
```
AlanWong-MS: Done.

PR #779, opened 2017-03-22
Batch response in json format w/ Accept header application/json

biaol-odata:
### Issues
*Batch response in json format w/ Accept header application/json*  

### Description
*This change uses JsonOutputContext for the batch response writing, therefore first implementation using RawOutputContext has become mostly noise, which is the reason for using a new forked branch (still based on ODataV4-6.x).*

- [ x ] Test cases added

### Additional work necessary


biaol-odata:
@mikepizzo is added to the review. #Closed

biaol-odata:
@markdstafford is added to the review. #Closed

biaol-odata:
@robward-ms is added to the review. #Closed

biaol-odata:
@manivannanse is added to the review. #Closed

src/Microsoft.OData.Core/ODataBatchJsonWriter.cs
```diff
@@ -0,0 +1,387 @@
...
+                    foreach (KeyValuePair<string, string> headerPair in headers)
+                    {
+                        this.jsonWriter.WriteName(headerPair.Key);
+                        this.jsonWriter.WriteValue(headerPair.Value);
+                    }
+                }
+                this.jsonWriter.EndObjectScope();
+
+                
+                // body attribute
```
mikepizzo: Remove one of the two empty lines prior to this line

src/Microsoft.OData.Core/ODataBatchJsonWriter.cs
```diff
@@ -0,0 +1,387 @@
...
+                if (!ODataBatchWriterUtils.HasResponseBody(this.CurrentOperationResponseMessage))
+                {
+                    // Close the individual response object now since there are no content.
+                    this.jsonWriter.EndObjectScope();
```
mikepizzo: Where does this get closed if there is a body? Why doesn't it get closed the same place whether or not there is a body?

src/Microsoft.OData.Core/ODataBatchJsonReader.cs
```diff
@@ -0,0 +1,80 @@
...
+
+
+        /// <summary>
+        /// Returns the cached <see cref="ODataBatchOperationRequestMessage"/> for reading the content of an operation 
+        /// in a batch request.
+        /// </summary>
+        /// <returns>The message that can be used to read the content of the batch request operation from.</returns>
+        internal override ODataBatchOperationRequestMessage CreateOperationRequestMessageImplementation()
+        {
+            return null;
```
mikepizzo: Todo, I assume?

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -16,27 +18,26 @@ namespace Microsoft.OData.Core
 #if ODATALIB_ASYNC
     using System.Threading.Tasks;
 #endif
+
+    using Microsoft.OData.Core.JsonLight;
```
mikepizzo: The JsonLight namespace should not be used by this base class. It  looks like it's not actually used and this line should be able to safely be deleted.

mikepizzo:
>     internal sealed class ODataBatchReaderStream

Is this class only used by MimeReader? If so should probably rename ODataBatchMimeReaderStream (or ODataMimeBatchReaderStream if we rename ODataBatchMimeReader to ODataMimeBatchReader).

---
Refers to: src/Microsoft.OData.Core/ODataBatchReaderStream.cs:25 in 9296350. [](commit_id = 929635024faeed6f72b4204ffc585572b7a9c783, deletion_comment = False)

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -4,6 +4,8 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using Microsoft.OData.Core.JsonLight;
```
mikepizzo: This namespace is not (and should not) be used by this base class, so this line should be deleted.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -429,12 +348,9 @@ private void VerifyCanWriteStartBatch(bool synchronousCall)
...
 
         /// <summary>
-        /// Starts a new batch - implementation of the actual functionality.
+        /// Starts a new batch.
         /// </summary>
-        private void WriteStartBatchImplementation()
-        {
-            this.SetState(BatchWriterState.BatchStarted);
-        }
+        internal abstract void WriteStartBatchImplementation();
```
mikepizzo: Implementation methods are not generally in base classes, they are called from the derived class's override method. -- is there a reason to have this here?

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -447,26 +363,9 @@ private void VerifyCanWriteEndBatch(bool synchronousCall)
...
-            this.SetState(BatchWriterState.BatchCompleted);
-
-            // write the end boundary for the batch
-            ODataBatchWriterUtils.WriteEndBoundary(this.rawOutputContext.TextWriter, this.batchBoundary, !this.batchStartBoundaryWritten);
-
-            // For compatibility with WCF DS we write a newline after the end batch boundary.
-            // Technically it's not needed, but it doesn't violate anything either.
-            this.rawOutputContext.TextWriter.WriteLine();
-        }
+        internal abstract void WriteEndBatchImplementation();
```
mikepizzo: Implementation methods are not generally in base classes, they are called from the derived classe's override mehtod. -- is there a reason to have this here?

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -481,27 +380,7 @@ private void VerifyCanWriteStartChangeset(bool synchronousCall)
...
-
-            // write the boundary string
-            ODataBatchWriterUtils.WriteStartBoundary(this.rawOutputContext.TextWriter, this.batchBoundary, !this.batchStartBoundaryWritten);
-            this.batchStartBoundaryWritten = true;
-
-            // write the change set headers
-            ODataBatchWriterUtils.WriteChangeSetPreamble(this.rawOutputContext.TextWriter, this.changeSetBoundary);
-            this.changesetStartBoundaryWritten = false;
-        }
+        internal abstract void WriteStartChangesetImplementation();
```
mikepizzo: Implementation methods are not generally in base classes, they are called from the derived classe's override mehtod. -- is there a reason to have this here?

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -514,62 +393,31 @@ private void VerifyCanWriteEndChangeset(bool synchronousCall)
...
-
-            // In the case of an empty changeset the start changeset boundary has not been written yet
-            // we will leave it like that, since we want the empty changeset to be represented only as
-            // the end changeset boundary.
-            // Due to WCF DS V2 compatiblity we must not write the start boundary in this case
-            // otherwise WCF DS V2 won't be able to read it (it fails on the start-end boundary empty changeset).
-
-            // write the end boundary for the change set
-            ODataBatchWriterUtils.WriteEndBoundary(this.rawOutputContext.TextWriter, currentChangeSetBoundary, !this.changesetStartBoundaryWritten);
+        internal abstract void WriteEndChangesetImplementation();
```
mikepizzo: Implementation methods are not generally in base classes, they are called from the derived classe's override mehtod. -- is there a reason to have this here?

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -4,6 +4,8 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+
+
 namespace Microsoft.OData.Core
```
mikepizzo: remove extra added blank lines above

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -203,49 +203,24 @@ void IODataBatchOperationListener.BatchOperationContentStreamDisposed()
...
-                    // We should never get here when in Exception state.
-                    throw new ODataException(Strings.General_InternalError(InternalErrorCodes.ODataBatchReader_GetEndBoundary_Exception));
-
-                default:
-                    // Invalid enum value
-                    throw new ODataException(Strings.General_InternalError(InternalErrorCodes.ODataBatchReader_GetEndBoundary_UnknownValue));
-            }
-        }
+        /// <returns>true if more items were read; otherwise false.</returns>
+        internal abstract bool ReadImplementation();
```
mikepizzo: >ReadImplementation [](start = 31, length = 18)

Implementation methods should be protected abstract in base class.

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -203,49 +203,24 @@ void IODataBatchOperationListener.BatchOperationContentStreamDisposed()
...
+
+        /// <summary>
+        /// Returns the cached <see cref="ODataBatchOperationRequestMessage"/> for reading the content of an operation 
+        /// in a batch request.
+        /// </summary>
+        /// <returns>The message that can be used to read the content of the batch request operation from.</returns>
+        internal abstract ODataBatchOperationRequestMessage CreateOperationRequestMessageImplementation();
```
mikepizzo: >CreateOperationRequestMessageImplementation [](start = 60, length = 43)

Implementation methods should be protected abstract in base class.

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -203,49 +203,24 @@ void IODataBatchOperationListener.BatchOperationContentStreamDisposed()
...
+
+        /// <summary>
+        /// Returns the cached <see cref="ODataBatchOperationRequestMessage"/> for reading the content of an operation 
+        /// in a batch request.
+        /// </summary>
+        /// <returns>The message that can be used to read the content of the batch request operation from.</returns>
+        internal abstract ODataBatchOperationResponseMessage CreateOperationResponseMessageImplementation();
```
mikepizzo: Implementation methods should be protected abstract in base class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
mikepizzo: Implementation methods should be protected abstract in base class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
mikepizzo: Implementation methods should be protected abstract in base class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
mikepizzo: Implementation methods should be protected abstract in base class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
mikepizzo: Implementation methods should be protected abstract in base class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -581,55 +429,8 @@ private void VerifyCanCreateOperationRequestMessage(bool synchronousCall, string
...
-
-            // write the operation's start boundary string
-            this.WriteStartBoundaryForOperation();
-
-            // write the headers and request line
-            ODataBatchWriterUtils.WriteRequestPreamble(this.rawOutputContext.TextWriter, method, uri, changeSetBoundary != null, contentId);
-
-            return this.CurrentOperationRequestMessage;
-        }
+        internal abstract ODataBatchOperationRequestMessage CreateOperationRequestMessageImplementation(string method,
```
mikepizzo: Implementation methods should be protected abstract in base class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -640,127 +441,54 @@ private void VerifyCanCreateOperationResponseMessage(bool synchronousCall)
...
-
-            // write the operation's start boundary string
-            this.WriteStartBoundaryForOperation();
-
-            // write the headers and request separator line
-            ODataBatchWriterUtils.WriteResponsePreamble(this.rawOutputContext.TextWriter, changeSetBoundary != null, contentId);
-
-            return this.CurrentOperationResponseMessage;
-        }
+        internal abstract ODataBatchOperationResponseMessage CreateOperationResponseMessageImplementation(
```
mikepizzo: Implementation methods should be protected abstract in base class.

src/Microsoft.OData.Core/ODataBatchJsonWriter.cs
```diff
@@ -0,0 +1,387 @@
...
+            this.CurrentOperationResponseMessage = null;
+
+            // Close the response object. Do not dispose the json writer.
+            this.jsonWriter.EndObjectScope();
+        }
+
+        /// <summary>
+        /// Starts a new batch - implementation of the actual functionality.
+        /// </summary>
+        internal override void WriteStartBatchImplementation()
```
mikepizzo: implementation methods should be protected override in derived class

src/Microsoft.OData.Core/ODataBatchJsonWriter.cs
```diff
@@ -0,0 +1,387 @@
...
+
+            // Start the responses array
+            this.jsonWriter.WriteName("responses");
+            this.jsonWriter.StartArrayScope();
+        }
+
+        /// <summary>
+        /// Ends a batch - implementation of the actual functionality.
+        /// </summary>
+        internal override void WriteEndBatchImplementation()
```
mikepizzo: implementation methods should be protected override in derived class

src/Microsoft.OData.Core/ODataBatchJsonWriter.cs
```diff
@@ -0,0 +1,387 @@
...
+            jsonWriter.EndArrayScope();
+
+            // Close the top level scope
+            jsonWriter.EndObjectScope();
+        }
+
+        /// <summary>
+        /// Starts a new changeset - implementation of the actual functionality.
+        /// </summary>
+        internal override void WriteStartChangesetImplementation()
```
mikepizzo: implementation methods should be protected override in derived class

src/Microsoft.OData.Core/ODataBatchJsonWriter.cs
```diff
@@ -0,0 +1,387 @@
...
+
+            // reset the size of the current changeset and increase the size of the batch
+            this.ResetChangeSetSize();
+            this.InterceptException(this.IncreaseBatchSize);
+        }
+
+        /// <summary>
+        /// Ends an active changeset - implementation of the actual functionality.
+        /// </summary>
+        internal override void WriteEndChangesetImplementation()
```
mikepizzo: implementation methods should be protected override in derived class

src/Microsoft.OData.Core/ODataBatchJsonReader.cs
```diff
@@ -0,0 +1,80 @@
...
+    using System.Threading.Tasks;
+#endif
+
+    using Microsoft.OData.Core.JsonLight;
+    #endregion Namespaces
+
+    /// <summary>
+    /// Class for reading OData batch messages in json format; also verifies the proper sequence of read calls on the reader.
+    /// </summary>
+    internal sealed class ODataBatchJsonReader: ODataBatchReader
```
mikepizzo: Should probably be named "ODataJsonLightBatchReader

src/Microsoft.OData.Core/ODataBatchJsonWriter.cs
```diff
@@ -0,0 +1,387 @@
...
+#endif
+
+    using Microsoft.OData.Core.Json;
+    using Microsoft.OData.Core.JsonLight;
+    #endregion Namespaces
+
+    /// <summary>
+    /// Class for writing OData batch messages of JSON type.
+    /// </summary>
+    internal sealed class ODataBatchJsonWriter : ODataBatchWriter
```
mikepizzo: >ODataBatchJsonWriter [](start = 26, length = 20)

Should probably be named "ODataJsonLightBatchWriter", and put in the "JsonLight" folder.

src/Microsoft.OData.Core/ODataBatchJsonWriter.cs
```diff
@@ -0,0 +1,387 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ODataBatchJsonWriter.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Core
```
mikepizzo: > Microsoft.OData.Core [](start = 9, length = 21)

Classes specific to JSON (light) format should be in Microsoft.OData.Core.JsonLight namespace

src/Microsoft.OData.Core/ODataBatchJsonReader.cs
```diff
@@ -0,0 +1,80 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ODataBatchJsonReader.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Core
```
mikepizzo: Classes specific to JSON (light) format should be in Microsoft.OData.Core.JsonLight namespace

src/Microsoft.OData.Core/ODataBatchMimeReader.cs
```diff
@@ -0,0 +1,375 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ODataBatchMimeReader.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Core
```
mikepizzo: >Microsoft.OData. [](start = 10, length = 16)

Should probably create a "Mime" namespace (and corresponding folder) and put the Mime classes there.

src/Microsoft.OData.Core/ODataBatchMimeReader.cs
```diff
@@ -0,0 +1,375 @@
...
+    using System.Diagnostics;
+    using System.Diagnostics.CodeAnalysis;
+    using System.Linq;
+    using System.Text;
+#if ODATALIB_ASYNC
+    using System.Threading.Tasks;
+#endif
+    #endregion Namespaces
+
+    internal sealed class ODataBatchMimeReader: ODataBatchReader
```
mikepizzo: Should be ODataMimeBatchReader.

src/Microsoft.OData.Core/ODataBatchMimeWriter.cs
```diff
@@ -0,0 +1,564 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ODataBatchMimeWriter.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Core
```
mikepizzo: Should probably create a "Mime" namespace (and corresponding folder) and put the Mime classes there.

src/Microsoft.OData.Core/ODataBatchMimeWriter.cs
```diff
@@ -0,0 +1,564 @@
...
+    using System.Globalization;
+#if ODATALIB_ASYNC
+    using System.Threading.Tasks;
+#endif
+    #endregion Namespaces
+
+    /// <summary>
+    /// Class for writing OData batch messages of MIME type.
+    /// </summary>
+    internal sealed class ODataBatchMimeWriter : ODataBatchWriter
```
mikepizzo: Should be ODataMimeBatchWriter

src/Microsoft.OData.Core/Json/ODataJsonOutputContextBase.cs
```diff
@@ -120,6 +122,17 @@ internal abstract class ODataJsonOutputContextBase : ODataOutputContext
         }
 
         /// <summary>
+        /// The output stream to write the payload to.
+        /// </summary>
+        internal Stream OutputStream
```
mikepizzo: >OutputStream [](start = 24, length = 12)

Rather than add a new property for OutputStream which is a duplicate of either messageOutputStream or asynchronousOutputStream, consider a GetMessageStream method that returns asynchronousOutputStream if not null, otherwise messageOutputStream.

test/FunctionalTests/Tests/DataOData/Tests/OData.Common.Tests/ObjectModelTests/ODataMessageTests.cs
```diff
@@ -4,6 +4,8 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using System.Reflection;
```
mikepizzo: Move to appropriate location within namespace

mikepizzo:
:clock1:

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: I was thinking about using protected too, which would make this API accessible from outside of the assembly (e.g. possible for a public client to derive ODataMyBatchReader). It appears to me that internal is providing better restriction. Can you share the related guidance in OData for using protected/internal?

---
In reply to: [108274517](https://github.com/OData/odata.net/pull/779#discussion_r108274517) [](ancestors = 108274517)

src/Microsoft.OData.Core/ODataBatchJsonReader.cs
biaol-odata: yes.

---
In reply to: [108258104](https://github.com/OData/odata.net/pull/779#discussion_r108258104) [](ancestors = 108258104)

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: deleted.

---
In reply to: [108261884](https://github.com/OData/odata.net/pull/779#discussion_r108261884) [](ancestors = 108261884)

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: talked to Mike. Third-party derivation is allowed. There are some grey areas, but we will stick to the established pattern.

---
In reply to: [108537013](https://github.com/OData/odata.net/pull/779#discussion_r108537013) [](ancestors = 108537013,108274517)

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
```diff
@@ -92,6 +103,14 @@ public string Method
         }
 
         /// <summary>
+        /// Gets the requst prerequisites.
```
mikepizzo: >requst [](start = 21, length = 6)

spelling: "request"

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
```diff
@@ -92,6 +103,14 @@ public string Method
         }
 
         /// <summary>
+        /// Gets the requst prerequisites.
+        /// </summary>
+        public IList<string> DependsOnRequestIds
```
mikepizzo: >DependsOnRequestIds [](start = 29, length = 19)

Might be simpler to call this just "DependsOn"

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
```diff
@@ -184,13 +203,15 @@ Uri IODataUrlResolver.ResolveUrl(Uri baseUri, Uri payloadUri)
             ODataBatchOperationHeaders headers,
             IODataBatchOperationListener operationListener,
             string contentId,
-            IODataUrlResolver urlResolver)
+            IODataUrlResolver urlResolver,
+            IList<string> dependsOnReqIds = null)
```
mikepizzo: >dependsOnReqIds [](start = 26, length = 15)

can just call this dependsOn

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -16,24 +16,22 @@ namespace Microsoft.OData.Core
...
-        /// <summary>The batch stream used by the batch reader to devide a batch payload into parts.</summary>
-        private readonly ODataBatchReaderStream batchStream;
+        private ODataInputContext inputContext;
 
         /// <summary>True if the writer was created for synchronous operation; false for asynchronous.</summary>
         private readonly bool synchronous;
 
         /// <summary>The batch-specific URL resolver that stores the content IDs found in a changeset and supports resolving cross-referencing URLs.</summary>
-        private readonly ODataBatchUrlResolver urlResolver;
+        internal readonly ODataBatchUrlResolver urlResolver;
```
mikepizzo: >urlResolver [](start = 48, length = 11)

internal variables should be initial caps.

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -57,31 +55,51 @@ public sealed class ODataBatchReader : IODataBatchOperationListener
...
 
         /// <summary>
         /// Constructor.
         /// </summary>
         /// <param name="inputContext">The input context to read the content from.</param>
         /// <param name="batchBoundary">The boundary string for the batch structure itself.</param>
-        /// <param name="batchEncoding">The encoding to use to read from the batch stream.</param>
         /// <param name="synchronous">true if the reader is created for synchronous operation; false for asynchronous.</param>
-        internal ODataBatchReader(ODataRawInputContext inputContext, string batchBoundary, Encoding batchEncoding, bool synchronous)
+        internal ODataBatchReader(ODataInputContext inputContext, bool synchronous)
```
mikepizzo: >internal [](start = 8, length = 8)

If ODataBatchReader is abstract, shouldn't this constructor be protected?  Shouldn't code always create one of the derived types?

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -106,7 +124,7 @@ public ODataBatchReaderState State
                 return this.batchReaderState;
             }
 
-            private set
+            set
```
mikepizzo: >set [](start = 12, length = 3)

Should set be protected?  Is State set outside of a derived class?

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -596,7 +335,7 @@ private void ParseRequestLine(string requestLine, out string httpMethod, out Uri
         /// <param name="responseLine">The response line as a string.</param>
         /// <returns>The parsed status code from the response line.</returns>
         [SuppressMessage("Microsoft.Performance", "CA1822:MarkMembersAsStatic", Justification = "'this' is used when built in debug")]
-        private int ParseResponseLine(string responseLine)
+        internal int ParseResponseLine(string responseLine)
```
mikepizzo: >ParseResponseLine [](start = 21, length = 17)

Is this needed in the base class? It looks like it's only called by the ODataMultipartMixedBatchReader; can it be moved there?

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -270,274 +263,25 @@ private Task<bool> ReadAsynchronously()
...
-        /// Returns the cached <see cref="ODataBatchOperationRequestMessage"/> for reading the content of an operation 
-        /// in a batch request.
+        /// Parses the request line of a batch operation request, without HTTP method validation.
         /// </summary>
-        /// <returns>The message that can be used to read the content of the batch request operation from.</returns>
-        private ODataBatchOperationResponseMessage CreateOperationResponseMessageImplementation()
+        /// <param name="requestLine">The request line as a string.</param>
+        /// <param name="httpMethod">The parsed HTTP method of the request.</param>
+        /// <param name="requestUri">The parsed <see cref="Uri"/> of the request.</param>
+        internal void ParseRequestLine(string requestLine, out string httpMethod, out Uri requestUri)
```
mikepizzo: >ParseRequestLine [](start = 22, length = 16)

Is this needed? It doesn't appear to have any callers.

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -270,274 +263,25 @@ private Task<bool> ReadAsynchronously()
...
 
         /// <summary>
         /// Parses the request line of a batch operation request.
         /// </summary>
         /// <param name="requestLine">The request line as a string.</param>
+        /// <param name="httpMethodValidation">The validation for the HTTP method in the request line.</param>
         /// <param name="httpMethod">The parsed HTTP method of the request.</param>
         /// <param name="requestUri">The parsed <see cref="Uri"/> of the request.</param>
-        private void ParseRequestLine(string requestLine, out string httpMethod, out Uri requestUri)
+        internal void ParseRequestLine(string requestLine, Action<string> httpMethodValidation, out string httpMethod, out Uri requestUri)
```
mikepizzo: >ParseRequestLine [](start = 22, length = 16)

Is this needed in the base class? If it's only called by the ODataMultipartMixedBatchReader, can it be moved there (and kept private)?

mikepizzo: >httpMethodValidation [](start = 74, length = 20)

If this is only called by ODataMultipartMixedBatchReader, and you move it there, can you inline this logic rather than pass it in as an action?

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -745,7 +484,7 @@ private void VerifyCallAllowed(bool synchronousCall)
         /// <summary>
         /// Increases the size of the current batch message; throws if the allowed limit is exceeded.
         /// </summary>
-        private void IncreaseBatchSize()
+        internal void IncreaseBatchSize()
```
mikepizzo: >IncreaseBatchSize [](start = 22, length = 17)

It looks like this is only called for MultipartMixed; shouldn't it also be called from the JSON format? Should it be protected?

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -758,7 +497,7 @@ private void IncreaseBatchSize()
         /// <summary>
         /// Increases the size of the current change set; throws if the allowed limit is exceeded.
         /// </summary>
-        private void IncreaseChangeSetSize()
+        internal void IncreaseChangeSetSize()
```
mikepizzo: >IncreaseChangeSetSize [](start = 22, length = 21)

This also seems only to be called for the MultipartMixed format; should be called for JSON as well. Consider making protected.

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -771,7 +510,7 @@ private void IncreaseChangeSetSize()
         /// <summary>
         /// Resets the size of the current change set to 0.
         /// </summary>
-        private void ResetChangeSetSize()
+        internal void ResetChangeSetSize()
```
mikepizzo: >ResetChangeSetSize [](start = 22, length = 18)

Same as previous 2 comments; should be called in JSON format, consider making protected.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -313,81 +313,34 @@ public Task<ODataBatchOperationResponseMessage> CreateOperationResponseMessageAs
 #endif
 
         /// <summary>Flushes the write buffer to the underlying stream.</summary>
-        public void Flush()
-        {
-            this.VerifyCanFlush(true);
-
```
mikepizzo: Why is this made abstract? The implementation should be the same in both the JsonLight and MultipartMixed derived classes, you just have to change this.rawOutputContext.Flush() to this.outputContext.Flush(), then you can remove the overrides in the derived classes.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -313,81 +313,34 @@ public Task<ODataBatchOperationResponseMessage> CreateOperationResponseMessageAs
...
 #if ODATALIB_ASYNC
         /// <summary>Flushes the write buffer to the underlying stream asynchronously.</summary>
         /// <returns>A task instance that represents the asynchronous operation.</returns>
-        public Task FlushAsync()
-        {
-            this.VerifyCanFlush(false);
-
-            // make sure we switch to state FatalExceptionThrown if an exception is thrown during flushing.
-            return this.rawOutputContext.FlushAsync().FollowOnFaultWith(t => this.SetState(BatchWriterState.Error));
-        }
```
mikepizzo: As above; you should be able to keep this public implementation in the base class and remove from derived classes by referencing this.outputContext, rather than this.rawOutputContext.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -313,81 +313,34 @@ public Task<ODataBatchOperationResponseMessage> CreateOperationResponseMessageAs
...
+        public abstract Task FlushAsync();
 #endif
 
         /// <summary>
         /// This method is called to notify that the content stream for a batch operation has been requested.
         /// </summary>
-        void IODataBatchOperationListener.BatchOperationContentStreamRequested()
-        {
-            // Write any pending data and flush the batch writer to the async buffered stream
-            this.StartBatchOperationContent();
```
mikepizzo: As above; you should be able to keep the implementation in the base class, and remove the implementations in the derived classes, by referencing the outputContext rather than the rawOutputContext.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -313,81 +313,34 @@ public Task<ODataBatchOperationResponseMessage> CreateOperationResponseMessageAs
...
-        Task IODataBatchOperationListener.BatchOperationContentStreamRequestedAsync()
-        {
-            // Write any pending data and flush the batch writer to the async buffered stream
-            this.StartBatchOperationContent();
-
-            // Asynchronously flush the async buffered stream to the underlying message stream (if there's any);
-            // then dispose the batch writer (since we are now writing the operation content) and set the corresponding state.
-            return this.rawOutputContext.FlushBuffersAsync()
-                .FollowOnSuccessWith(task => this.DisposeBatchWriterAndSetContentStreamRequestedState());
-        }
```
mikepizzo: As above; keep this implementation in the base class and remove from derived classes.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -313,81 +313,34 @@ public Task<ODataBatchOperationResponseMessage> CreateOperationResponseMessageAs
...
         /// <summary>
         /// This method is called to notify that the content stream of a batch operation has been disposed.
         /// </summary>
-        void IODataBatchOperationListener.BatchOperationContentStreamDisposed()
-        {
-            Debug.Assert(this.CurrentOperationMessage != null, "Expected non-null operation message!");
-
-            this.SetState(BatchWriterState.OperationStreamDisposed);
-            this.CurrentOperationRequestMessage = null;
-            this.CurrentOperationResponseMessage = null;
```
mikepizzo: Should put common logic (setting state, setting request/response message to null) in the base class. In MultipartMixed derived class, override this method with a method that first calls the base implementation, then calls this.rawOutputContext.InitializeRawValueWriter (or figure out some other way to call InitializeRawValueWriter at the appropriate time in the derived class).

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -396,17 +349,7 @@ void IODataBatchOperationListener.BatchOperationContentStreamDisposed()
...
-        void IODataOutputInStreamErrorListener.OnInStreamError()
-        {
-            this.rawOutputContext.VerifyNotDisposed();
-            this.SetState(BatchWriterState.Error);
-            this.rawOutputContext.TextWriter.Flush();
-
-            // The OData protocol spec did not defined the behavior when an exception is encountered outside of a batch operation. The batch writer
-            // should not allow WriteError in this case. Note that WCF DS Server does serialize the error in XML format when it encounters one outside of a
-            // batch operation.
-            throw new ODataException(Strings.ODataBatchWriter_CannotWriteInStreamErrorForBatch);
```
mikepizzo: Common logic, especially state handling, should be done in base class implementation. Both ODataRawOutputContext and ODataJsonOutputContextBase have Flush methods that call flush on the underlying rawValueWriter/JsonWriter, so if you can add Flush to the ODataOutputContext base class you can just call outputContext.Flush() instead of this.rawOutputContext.TextWriter.Flush() in the MultipartMixed derived class, or this.JsonLightOutputContext.JsonWriter.Flush() in the JsonLight derived class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -429,12 +372,9 @@ private void VerifyCanWriteStartBatch(bool synchronousCall)
         }
 
         /// <summary>
-        /// Starts a new batch - implementation of the actual functionality.
+        /// Starts a new batch.
         /// </summary>
-        private void WriteStartBatchImplementation()
-        {
-            this.SetState(BatchWriterState.BatchStarted);
-        }
```
mikepizzo: This implementation deals with state management; its implementation (which doesn't differ between multipartmixed and json) should be left in the base class and, since it's only caller is the base class, it should be kept private..

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -447,26 +387,9 @@ private void VerifyCanWriteEndBatch(bool synchronousCall)
...
-
-            this.SetState(BatchWriterState.BatchCompleted);
-
-            // write the end boundary for the batch
-            ODataBatchWriterUtils.WriteEndBoundary(this.rawOutputContext.TextWriter, this.batchBoundary, !this.batchStartBoundaryWritten);
-
-            // For compatibility with WCF DS we write a newline after the end batch boundary.
-            // Technically it's not needed, but it doesn't violate anything either.
-            this.rawOutputContext.TextWriter.WriteLine();
-        }
```
mikepizzo: State management and other common logic in xyzImplementation() methods should be in the base class, which should call Start/End methods in the derived class for the actual writing.  For comparison, see how ODataWriterCore.WriteStartEntryImplementation implements all of the state/validation logic and then calls the protected abstract StartEntry(), overridden in ODataJsonLightWriter, to write the actual value.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -481,27 +404,7 @@ private void VerifyCanWriteStartChangeset(bool synchronousCall)
...
-            this.InterceptException(this.IncreaseBatchSize);
-
-            // write the boundary string
-            ODataBatchWriterUtils.WriteStartBoundary(this.rawOutputContext.TextWriter, this.batchBoundary, !this.batchStartBoundaryWritten);
-            this.batchStartBoundaryWritten = true;
-
-            // write the change set headers
-            ODataBatchWriterUtils.WriteChangeSetPreamble(this.rawOutputContext.TextWriter, this.changeSetBoundary);
-            this.changesetStartBoundaryWritten = false;
-        }
```
mikepizzo: as above; this *Implementation() method should be private void and contain the state management logic, then call into an abstract StartChangeset implementation, overriden in the derived classes, to contain the actual logic for writing the changeset.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -514,62 +417,31 @@ private void VerifyCanWriteEndChangeset(bool synchronousCall)
...
-            this.SetState(BatchWriterState.ChangeSetCompleted);
-
-            // In the case of an empty changeset the start changeset boundary has not been written yet
-            // we will leave it like that, since we want the empty changeset to be represented only as
-            // the end changeset boundary.
-            // Due to WCF DS V2 compatiblity we must not write the start boundary in this case
-            // otherwise WCF DS V2 won't be able to read it (it fails on the start-end boundary empty changeset).
-
-            // write the end boundary for the change set
-            ODataBatchWriterUtils.WriteEndBoundary(this.rawOutputContext.TextWriter, currentChangeSetBoundary, !this.changesetStartBoundaryWritten);
```
mikepizzo: As above, End*Implementation methods should be private in the base class, and contain shared logic, then call abstract End* methods that are implemented in the derived classes and do the actual format-specific writing.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -514,62 +417,31 @@ private void VerifyCanWriteEndChangeset(bool synchronousCall)
...
-
-            // Reset the cache of content IDs here. As per spec, content IDs are only unique inside a change set.
-            this.urlResolver.Reset();
-            this.currentOperationContentId = null;
-        }
```
mikepizzo: logic should remain private in base class, and call abstract EndChangeset, implemented in derived class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -581,55 +453,8 @@ private void VerifyCanCreateOperationRequestMessage(bool synchronousCall, string
...
-
-            // Add a potential Content-ID header to the URL resolver so that it will be available
-            // to subsequent operations.
-            // Note that what we add here is the Content-ID header of the previous operation (if any).
-            // This also means that the Content-ID of the last operation in a changeset will never get
-            // added to the cache which is fine since we cannot reference it anywhere.
-            if (this.currentOperationContentId != null)
-            {
-                this.urlResolver.AddContentId(this.currentOperationContentId);
-            }
```
mikepizzo: This method should be private and call into abstract method implemented in derived class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -581,55 +453,8 @@ private void VerifyCanCreateOperationRequestMessage(bool synchronousCall, string
...
-                uri,
-                /*operationListener*/ this,
-                this.urlResolver);
-
-            if (this.changeSetBoundary != null)
-            {
-                this.RememberContentIdHeader(contentId);
-            }
-
-            this.SetState(BatchWriterState.OperationCreated);
```
mikepizzo: > this.SetState(BatchWriterState.OperationCreated); [](start = 11, length = 50)

State management should be in base class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -640,127 +465,54 @@ private void VerifyCanCreateOperationResponseMessage(bool synchronousCall)
...
 
         /// <summary>
-        /// Creates an <see cref="ODataBatchOperationResponseMessage"/> for writing an operation of a batch response - implementation of the actual functionality.
+        /// Creates an <see cref="ODataBatchOperationResponseMessage"/> for writing an operation of a batch response.
         /// </summary>
         /// <param name="contentId">The Content-ID value to write in ChangeSet head.</param>
         /// <returns>The message that can be used to write the response operation.</returns>
-        private ODataBatchOperationResponseMessage CreateOperationResponseMessageImplementation(string contentId)
-        {
-            this.WritePendingMessageData(true);
```
mikepizzo: This method should remain private, contain common/state management logic, and call abstract method in derived class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -640,127 +465,54 @@ private void VerifyCanCreateOperationResponseMessage(bool synchronousCall)
...
 
         /// <summary>
         /// Disposes the batch writer and set the 'OperationStreamRequested' batch writer state;
         /// called after the flush operation(s) have completed.
         /// </summary>
-        private void DisposeBatchWriterAndSetContentStreamRequestedState()
-        {
-            this.rawOutputContext.CloseWriter();
-
-            this.SetState(BatchWriterState.OperationStreamRequested);
```
mikepizzo: State management should be done in base class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -640,127 +465,54 @@ private void VerifyCanCreateOperationResponseMessage(bool synchronousCall)
...
         /// <param name="synchronousCall">true if the call is to be synchronous; false otherwise.</param>
-        private void VerifyCanFlush(bool synchronousCall)
-        {
-            this.rawOutputContext.VerifyNotDisposed();
-            this.VerifyCallAllowed(synchronousCall);
-            if (this.state == BatchWriterState.OperationStreamRequested)
-            {
-                this.ThrowODataException(Strings.ODataBatchWriter_FlushOrFlushAsyncCalledInStreamRequestedState);
-            }
-        }
```
mikepizzo: You should be able to make this a private method in the base class; see ODataWriterCore's VerifyCanFlush method.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -640,127 +465,54 @@ private void VerifyCanCreateOperationResponseMessage(bool synchronousCall)
...
+        protected abstract void VerifyCanFlush(bool synchronousCall);
 
         /// <summary>
         /// Verifies that a call is allowed to the writer.
         /// </summary>
         /// <param name="synchronousCall">true if the call is to be synchronous; false otherwise.</param>
-        private void VerifyCallAllowed(bool synchronousCall)
+        internal void VerifyCallAllowed(bool synchronousCall)
```
mikepizzo: >internal [](start = 8, length = 8)

Should be able to make this private once you restore VerifyCanFlush to a private method in this class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -796,68 +548,36 @@ private void InterceptException(Action action)
         /// Sets a new writer state; verifies that the transition from the current state into new state is valid.
         /// </summary>
         /// <param name="newState">The writer state to transition into.</param>
-        private void SetState(BatchWriterState newState)
-        {
-            this.InterceptException(() => this.ValidateTransition(newState));
```
mikepizzo: SetState should remain private -- all state management should be done in the base class.  Logic that is manipulating state in the derived classes needs to be moved back to the base class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -796,68 +548,36 @@ private void InterceptException(Action action)
...
 
-            switch (newState)
+        /// <summary>
+        /// Wrapper method to validate state transition with optional customized validation.
+        /// </summary>
+        /// <param name="newState">Teh new writer state to transition into.</param>
+        /// <param name="customizedValidationAction">Optional validation action.</param>
+        internal void ValidateTransition(BatchWriterState newState, Action customizedValidationAction)
+        {
+            if (customizedValidationAction != null)
```
mikepizzo: >customizedValidationAction [](start = 16, length = 26)

ValidateTransition should only validate that the state transition is valid, which should be independent of any implementation so it shouldn't require passing in a customValdiationAction.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -796,68 +548,36 @@ private void InterceptException(Action action)
...
-
-            // make sure that we are not completing a batch while a changeset is still active
-            if (newState == BatchWriterState.BatchCompleted)
-            {
-                if (this.changeSetBoundary != null)
-                {
-                    throw new ODataException(Strings.ODataBatchWriter_CannotCompleteBatchWithActiveChangeSet);
-                }
-            }
-
```
mikepizzo: all of this logic should continue to be in the base class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -939,82 +659,26 @@ private void ValidateTransition(BatchWriterState newState)
...
-        private void ValidateWriterReady()
-        {
-            this.rawOutputContext.VerifyNotDisposed();
-
-            // If the operation stream was requested but not yet disposed, the writer can't be used to do anything.
-            if (this.state == BatchWriterState.OperationStreamRequested)
-            {
-                throw new ODataException(Strings.ODataBatchWriter_InvalidTransitionFromOperationContentStreamRequested);
-            }
-        }
```
mikepizzo: should remain in the base class.

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -270,274 +263,25 @@ private Task<bool> ReadAsynchronously()
             return TaskUtils.GetTaskForSynchronousOperation<bool>(this.ReadImplementation);
         }
 #endif
-
-        /// <summary>
-        /// Continues reading from the batch message payload.
-        /// </summary>
-        /// <returns>true if more items were read; otherwise false.</returns>
-        private bool ReadImplementation()
```
mikepizzo: >ReadImplementation [](start = 21, length = 18)

The *Implementation methods should be private methods in the base class that deal with common logic (primarily statemanagement/validation logic).  Those methods should call into abstract methods that are overridden in the derived classes to do the actual reading.

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
biaol-odata: fixed.

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
biaol-odata: I believe your thought is from the "DependsOn" request attribute in the Spec. The "DependsOn" in the Spec is a list of request Ids and group Ids. However, the member variable here only contains "RequestIds", considering groupIds are not the same between request and response in current protocol (for both multipart and json formats) so the only reliable way for the client to ensure that prerequisites of a request is satisfy is thru request ids.

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
biaol-odata: same as above.

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: Cannot be protected: class ODataBatchUrlResolver is internal, whose accessibility is more restricted than protected. Usually library should apply restrictive scope (such as "internal" here) as much as possible to prevent misuse from client, right?

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: make sense. fixed.

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: Yes. VS's Resharper suggests so as well.

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: Observation is correct. Moved.

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: wrapper method removed; it was some leftover from removed impl. Also moved the wrapped method to ODataMultipartMixedBatchReader, which becomes the only caller.

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: Yes. Moved.

biaol-odata: Fixed. I cannot recall the exact reason for passing in an no-op action at the time, maybe during some interim implementation, but seems like that is no longer the case any more.

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: Added to Json Batch Reader.

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: Added to Json Batch Reader.

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: Added to Json Batch Reader.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: I don't think it would work, because the common base ODataOutputContext context doesn't have a Flush() definition. The design has been letting the derived context (raw, json or text, etc) to handle underlying writer respectively. Original BatchWriter has only one format and it can provide implementation here. Now with the additional Json format, it needs to be abstract.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: same as above, no FlushAsync supported on base context.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: same as above. No Flush operations supported in base context.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: same as above. No Flush operations supported in base context.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: The downside of do this is that disposed will be spread in multipart reader and the base reader. Four lines of code is not big amount of duplicates but it makes the disposal logic all in one place for each class. Let me know how you think?

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: I don't think we can justify ripples to five other derived classes of ODataOutputContext due to adding Flush needed by these specific two context types? We can solve it within the limited scope of batch writer.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: Observation is correct. Original design of this work item started from overriding all methods related to state but this one is the same and should be cleaned up.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: The ODataWriterCore is for entry writer, not necessary suitable for batch writer, right? I would prefer keeping the original flow for batching to avoid potential surprise.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: Let's discuss more about this (likely when current QP milestone finishes) to understand the risk & benefits.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: same as above.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: same as above.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: not sure I understand your comment here, can you please elaborate? Thanks.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: I think this is related to discussions above.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: related to above.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: Actually, the customizedValidationAction passed into the ValidateTransition is required by the Multipart batch reader (related to changeset boundary) but not the Json batch reader. and doing so that common logic can be shared.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: base class's OutputContext doesn't support VerifyNotDisposed()

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: Let's discuss more after QP.

biaol-odata:
All CR comments here have been addressed, and this PR has been rebased to maintenance-V4:
https://github.com/OData/odata.net/pull/951

Close it now.

abelykh0:
What "in-progress" label indicates on this closed issue?

biaol-odata:
@abelykh0 This PR originally targeted OData:ODataV4-6.x, which is a maintenance branch. It has been rebased to the main dev branch and then released.

PR #796, opened 2017-04-13
Fix client test failures

robward-ms:
This change fixes test failures in both local machine and in the lab
because of the switch to VS2015. Switching to VS2015 changed the testing
database engine from SQL Express to LocalDb. Since VS2015 installs LocalDb,
this is a change that make sense to keep.

For local tests, some client tests were failing due to a failure to
clean up local database entries in LocalDB. The change to artdbclean.js
uses SQL Native Client for LocalDB to fix a connection issue, allowing the
script to properly clean up after the tests.

For lab tests, the machine configuration is a bit different and LocalDb is not
available; therefore, use SQL Express when building the lab. Since the current
VS2015 symbol was only used for test DB selection, I renamed the symbol to
reflect its usage.

- [   ] Test cases added

### Issues

### Description

- [ x ] Test cases added

### Additional work necessary

### DIFF
--- a/tools/Scripts/artdbclean.js
+++ b/tools/Scripts/artdbclean.js
@@ -17,9 +17,9 @@ function OpenResultSet(connection, queryText)
 
 // Opens an ADO SQL Server driver connection to the specified.
 function OpenSqlServerConnection(serverName, initialCatalog, useTrustedConnection,
-  username, password)
+  username, password, provider)
 {
-  var connectionString = "Provider='sqloledb';Data Source=" + serverName + ";";
+  var connectionString = "Provider='" + provider + "';Data Source=" + serverName + ";";
   if (initialCatalog != null && initialCatalog != "")
   {
     connectionString += "Initial Catalog='" + initialCatalog + "';";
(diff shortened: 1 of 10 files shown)


tools/Scripts/artdbclean.js
```diff
@@ -85,9 +85,19 @@ if (enlistmentRoot == null || enlistmentRoot == "")
...
-for (var i in serverNames) {
-  DropDatabasesForServer(serverNames[i], enlistmentRoot);
+var serverNames = ["(local)", ".\\SQLEXPRESS"];
+
+for (var i in serverNames)
+{
+  DropDatabasesForServer(serverNames[i], enlistmentRoot, "sqloledb");
+}
+
+var serverNames = ["(localdb)\\MSSQLLocalDB"];
```
AlanWong-MS: Nit: serverNames declared again.

tools/Scripts/artdbclean.js
```diff
@@ -85,9 +85,19 @@ if (enlistmentRoot == null || enlistmentRoot == "")
...
 }
 
-var serverNames = ["(local)", ".\\SQLEXPRESS", "(localdb)\\MSSQLLocalDB"];
-for (var i in serverNames) {
-  DropDatabasesForServer(serverNames[i], enlistmentRoot);
+var serverNames = ["(local)", ".\\SQLEXPRESS"];
+
+for (var i in serverNames)
+{
+  DropDatabasesForServer(serverNames[i], enlistmentRoot, "sqloledb");
```
AlanWong-MS: Nit: indentation might be slightly off... double-check?

Issue #782, opened 2017-03-27
Add support for optional parameters in function


jatin085:
The library already supports OData actions with optional parameters i.e. a client can execute an OData action with one or more optional parameters missing. For functions, we need a similar support. Presently, it seems the code flows through method `FilterFunctionsByParameterNames` that picks a function only if its parameter count matches with count of input parameters.

### Assemblies affected
OData .Net lib 7.0.0

### Reproduce steps
*Define a function with optional parameter.  i.e. one that does not have `Nullable="true"` field value set. Make call to the OData function without passing the optional parameter.*

### Expected result
*OData function call is successful*

### Actual result
*"Resource not found" error.*

xuzhg:
Similar requirement at: #403

mikepizzo:
The OData protocol doesn't support optional parameters for function invocation because it wants the service to be able to correctly route function overloads based on the URL (actions don't have overloads based on non-binding parameters for this same reason -- non-binding parameters are passed in the body).

The OData-compliant way to do this would be to define two versions of the function, one with the optional parameter and one without.

Does that not work for your scenario?

jatin085:
Thanks for the update. We have few functions with 3 or 4 optional parameters. Exposing 16 different overloads does not seem to be a right workaround.

- Jatin

-----Original Message-----

jatin085:
Hi Mike, if I get it correctly, a single routing pattern won't work even if the caller changes the order of input arguments to the function but the OData specification supports reordering of the function arguments. Let me know if I am missing something.

mikepizzo:
OData requires that *both* of the following are true:
-the unordered set of parameters names must uniquely identify an overload
-the ordered set of parameter types must uniquely identify an overload

The first rule is to support routing (since OData doesn't require the parameters to be passed in order) and the second rule was added largely for compatibility with programming languages which may want to generate corresponding strongly typed functions.

As long as we can live with these rules still being supported for all *required* parameters, I believe we could add optional parameters without breaking routing or language overload rules.

Specifically:
-Optional parameters must be defined after any required parameters (for language compatibility)
-Unordered list of *required* parameter names must uniquely identify the overload
-Ordered list of *required* parameter types must uniquely identify the overload

Would that address your requirements?

jatin085:
[...]

PR #804, opened 2017-05-02
Support Reading/Writing Optional Parameters

mikepizzo:
Support selecting best function overload based on optional parameters

### Issues
*This pull request fixes issue #782  

### Description
*Supports new Core annotation for marking a property as optional.*
*Adds IEdmOptionalParameter for marking a parameter as optional, getting/setting default value*
*Adds EdmOptionalParameter for creating an optional parameter within a model*
*Validates that optional parameters must come at the end*
*Adds logic to function binder and uri resolver to handle optional parameters*

- [ x ] Test cases added
- [   ] Build and test with one-click build and test script passed

### Additional work necessary
*Docs required*
*Public API needs to be updated*
*Support in WebAPI OData*
*Support for out-of-line optional annotation*


mikepizzo:
Closing in favor of PR #828, which targets the new master branch.

PR #825, opened 2017-06-07
Update expired certificate for WindowsStore test project.

robward-ms:
This change fixes a build issue with an expired certificate.

- [   ] Test cases added


robward-ms:
This change should be in the maintenance-6.x branch. We may also need it in master.

Issue #799, opened 2017-04-28
Support for $compute in $select and $filter


ysanghi:
*Short summary (3-5 sentences) describing the issue.*  
OData 4.01 adds a new "$compute" clause for computing expressions that can then appear in $select, and in results as dynamic columns. 

See http://docs.oasis-open.org/odata/odata/v4.01/csprd01/part2-url-conventions/odata-v4.01-csprd01-part2-url-conventions.html#_Toc470093702.

We need this functionality so that we can issue requests like:
http://host/api/Collections(‘id’)/Items?$select=Property1,Property2
       &$compute=Cast(Prop1,’Edm.Int32) as Property1, Cast(Prop2,’Edm.String’) as Property2

### Assemblies affected
OData.NET 7.X 

### Reproduce steps
Trying $compute with ODataUriParser doesn't parse the query parameter as a valid supported query parameter as detailed in the spec.

### Expected result
*What would happen if there wasn't a bug.*  
We can parse the URI segments and also write computed values in the response for both typed and untyped values.

### Actual result
*What is actually happening.*  
The query parameter is just ignored.

PR #827, opened 2017-06-12
Implementation of $compute

mtroth-microsoft:
### Issues
This pull request fixes issue #799 .

### Description
Adding support to ODL parser for the $compute query option.

- [ x ] Test cases added

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/ODataQueryOptionParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/ODataQueryOptionParser.cs
@@ -44,6 +44,11 @@ namespace Microsoft.OData.UriParser
         private ApplyClause applyClause;
 
         /// <summary>
+        /// Compute clause for computation queries
+        /// </summary>
+        private ComputeClause computeClause;
+
+        /// <summary>
         /// The path info about parsed segments and target edm type and navigation source.
         /// </summary>
         private ODataPathInfo odataPathInfo;
@@ -326,6 +331,31 @@ namespace Microsoft.OData.UriParser
             string deltaTokenQuery;
             return this.TryGetQueryOption(UriQueryConstants.DeltaTokenQueryOption, out deltaTokenQuery) ? deltaTokenQuery : null;
         }
+
+        /// <summary>
+        /// Parses a compute clause on the given full Uri, binding
+        /// the text into semantic nodes using the constructed mode.
+        /// </summary>
+        /// <returns>A <see cref="ComputeClause"/> representing the computed properties.</returns>
+        public ComputeClause ParseCompute()
+        {
+            if (this.computeClause != null)
+            {
+                return this.computeClause;
+            }
+
+            string computeQuery;
+
+            if (!this.TryGetQueryOption(UriQueryConstants.ComputeQueryOption, out computeQuery)
+                || string.IsNullOrEmpty(computeQuery)
+                || this.targetEdmType == null)
+            {
+                return null;
+            }
+
+            this.computeClause = ParseComputeImplementation(computeQuery, this.Configuration, this.odataPathInfo);
+            return this.computeClause;
+        }
         #endregion public methods
 
         #region private methods
@@ -541,6 +571,34 @@ namespace Microsoft.OData.UriParser
             SearchBinder searchBinder = new SearchBinder(binder.Bind);
 
             return searchBinder.BindSearch(queryToken);
+        }
+
+        /// <summary>
+        /// Parses the <paramref name="compute"/> clause, binding
+        /// the text into a metadata-bound list of compuations using the provided model.
+        /// </summary>
+        /// <param name="compute">String representation of the compute expression from the URI.</param>
+        /// <param name="configuration">The configuration used for binding.</param>
+        /// <param name="odataPathInfo">The path info from Uri path.</param>
+        /// <returns>A <see cref="ComputeClause"/> representing the metadata bound compute expression.</returns>
+        private static ComputeClause ParseComputeImplementation(string compute, ODataUriParserConfiguration configuration, ODataPathInfo odataPathInfo)
+        {
+            ExceptionUtils.CheckArgumentNotNull(configuration, "configuration");
+            ExceptionUtils.CheckArgumentNotNull(compute, "compute");
+
+            // Get the syntactic representation of the apply expression
+            UriQueryExpressionParser expressionParser = new UriQueryExpressionParser(configuration.Settings.FilterLimit, configuration.EnableCaseInsensitiveUriFunctionIdentifier);
+            ComputeToken computeToken = expressionParser.ParseCompute(compute);
+
+            // Bind it to metadata
+            BindingState state = new BindingState(configuration);
+            state.ImplicitRangeVariable = NodeFactory.CreateImplicitRangeVariable(odataPathInfo.TargetEdmType.ToTypeReference(), odataPathInfo.TargetNavigationSource);
+            state.RangeVariables.Push(state.ImplicitRangeVariable);
+            MetadataBinder binder = new MetadataBinder(state);
+            ComputeBinder computeBinder = new ComputeBinder(binder.Bind);
+            ComputeClause boundNode = computeBinder.BindCompute(computeToken);
+
+            return boundNode;
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Core/UriParser/Parsers/UriQueryExpressionParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/UriQueryExpressionParser.cs
@@ -205,6 +205,38 @@ namespace Microsoft.OData.UriParser
             }
         }
 
+        // parses $compute query option.
+        internal ComputeToken ParseCompute(string compute)
+        {
+            Debug.Assert(compute != null, "compute != null");
+
+            List<ComputeExpressionToken> transformationTokens = new List<ComputeExpressionToken>();
+
+            if (string.IsNullOrEmpty(compute))
+            {
+                return new ComputeToken(transformationTokens);
+            }
+
+            this.recursionDepth = 0;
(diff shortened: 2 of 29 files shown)


kosinsky:
Could you support inside $apply too? For example, GET ~/Sales?$apply=compute(Amount mul Product/TaxRate as Tax) I think it should be the same approach as $filter and filter transformation in $apply
http://docs.oasis-open.org/odata/odata-data-aggregation-ext/v4.0/cs02/odata-data-aggregation-ext-v4.0-cs02.html#_Toc435016588

src/Microsoft.OData.Core/UriParser/Parsers/UriQueryExpressionParser.cs
```diff
@@ -205,6 +205,38 @@ internal static string GetEdmConstantNames(IEdmTypeReference edmTypeReference)
...
+            {
+                return transformationTokens;
+            }
+
+            this.recursionDepth = 0;
+            this.lexer = CreateLexerForFilterOrOrderByOrApplyExpression(compute);
+
+            while (true)
+            {
+                ComputeToken computed = this.ParseComputeExpression();
```
robward-ms: >computed [](start = 29, length = 8)

could computed by null or invalid?

src/Microsoft.OData.Core/UriParser/SyntacticAst/ComputeToken.cs
```diff
@@ -0,0 +1,79 @@
...
+            this.expression = expression;
+            this.alias = alias;
+        }
+
+        /// <summary>
+        /// Gets the kind of this token.
+        /// </summary>
+        public override QueryTokenKind Kind
+        {
+            get { return QueryTokenKind.Compute; }
```
AlanWong-MS: format nit: the bracelets are on the line as the return, whereas they're on separate lines for the other properties below

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
```diff
@@ -677,7 +677,152 @@ public void ParsePathFunctionWithNullEntitySetPath()
             pathSegmentList[0].ShouldBeOperationImportSegment(getCurrentCustomerImport);
             pathSegmentList[1].ShouldBeNavigationPropertySegment(customerDetail);
         }
+        #endregion
 
+        #region Parse Compute Tests
```
robward-ms: >Tests [](start = 30, length = 5)

What about negative testing? Maybe a test for missing "as"?

mtroth-microsoft:
As discussed offline, this makes sense but will be handled through a separate PR.

---
In reply to: [309063099](https://github.com/OData/odata.net/pull/827#issuecomment-309063099) [](ancestors = 309063099)

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
mtroth-microsoft: I'll add something for 'as', but the remainder should be covered by the expression parsing reuse.

---
In reply to: [123090919](https://github.com/OData/odata.net/pull/827#discussion_r123090919) [](ancestors = 123090919)

src/Microsoft.OData.Core/UriParser/Parsers/UriQueryExpressionParser.cs
mtroth-microsoft: computed can't be null since we are testing for the a null compute string and returning empty token list if encountered. computed can be invalid in a variety of ways. your negative testing comment points in that direction. invalid expression formation is caught be the reuse of the expression binding code path, only the 'as' tests should be improved.  testing the alias that follows the 'as' is already covered by the apply code path that is being reused.

---
In reply to: [122828867](https://github.com/OData/odata.net/pull/827#discussion_r122828867) [](ancestors = 122828867)

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
```diff
@@ -677,7 +677,170 @@ public void ParsePathFunctionWithNullEntitySetPath()
...
+            subPropertyInfo.TargetMultiplicity = EdmMultiplicity.One;
+            EdmProperty subnavigation = EdmNavigationProperty.CreateNavigationProperty(targetType, subPropertyInfo);
+            targetType.AddProperty(subnavigation);
+
+            EdmEntityContainer container = model.AddEntityContainer("Default", "Container");
+            container.AddEntitySet("Entities", elementType);
+
+            // Define queries and new up parser.
+            Uri root = new Uri("http://host");
+            Uri url = new Uri("http://host/Entities?$compute=cast(Prop1, 'Edm.String') as Property1AsString, tolower(Prop1) as Property1Lower&$expand=Nav1($compute=cast(Prop1, 'Edm.String') as NavProperty1AsString;$expand=SubNav1($compute=cast(Prop1, 'Edm.String') as SubNavProperty1AsString))");
```
AlanWong-MS: Can we break this line into multiple lines?

src/Microsoft.OData.Core/UriParser/SemanticAst/ExpandedReferenceSelectItem.cs
```diff
@@ -45,6 +45,33 @@ public ExpandedReferenceSelectItem(ODataExpandPath pathToNavigationProperty, IEd
...
+        /// Create an expand item, using a navigationProperty, its entity set, and any expand options.
+        /// </summary>
+        /// <param name="pathToNavigationProperty">the path to the navigation property for this expand item, including any type segments</param>
+        /// <param name="navigationSource">the navigation source for this expand level.</param>
+        /// <param name="filterOption">A filter clause for this expand (can be null)</param>
+        /// <param name="orderByOption">An Orderby clause for this expand (can be null)</param>
+        /// <param name="topOption">A top clause for this expand (can be null)</param>
+        /// <param name="skipOption">A skip clause for this expand (can be null)</param>
+        /// <param name="countOption">An query count clause for this expand (can be null)</param>
+        /// <param name="searchOption">An levels clause for this expand (can be null)</param>
```
AlanWong-MS: An levels -> A search

src/Microsoft.OData.Core/UriParser/SyntacticAst/ComputeToken.cs
```diff
@@ -0,0 +1,82 @@
...
+
+        /// <summary>
+        /// Accept a <see cref="ISyntacticTreeVisitor{T}"/> to walk a tree of <see cref="QueryToken"/>s.
+        /// </summary>
+        /// <typeparam name="T">Type that the visitor will return after visiting this token.</typeparam>
+        /// <param name="visitor">An implementation of the visitor interface.</param>
+        /// <returns>An object whose type is determined by the type parameter of the visitor.</returns>
+        public override T Accept<T>(ISyntacticTreeVisitor<T> visitor)
+        {
+            return default(T);
```
mikepizzo: Why is this just returning default(T) and not visitor.Visit(this)?

src/Microsoft.OData.Core/UriParser/SyntacticAst/ExpandTermToken.cs
```diff
@@ -104,6 +109,36 @@ public ExpandTermToken(PathSegmentToken pathToNavigationProp, SelectToken select
...
+            PathSegmentToken pathToNavigationProp,
+            QueryToken filterOption,
+            IEnumerable<OrderByToken> orderByOptions,
+            long? topOption,
+            long? skipOption,
+            bool? countQueryOption,
+            long? levelsOption,
+            QueryToken searchOption,
+            SelectToken selectOption,
+            IEnumerable<ComputeToken> computeOptions,
```
mikepizzo: Why did you add computeOptions *before* expandOption? Generally better for overloads to build on previous parameter ordering.

mikepizzo:
:clock1:

src/Microsoft.OData.Core/UriParser/ODataQueryOptionParser.cs
```diff
@@ -326,6 +331,31 @@ public string ParseDeltaToken()
...
+            if (this.computeClause != null)
+            {
+                return this.computeClause;
+            }
+
+            string computeQuery;
+
+            if (!this.TryGetQueryOption(UriQueryConstants.ComputeQueryOption, out computeQuery)
+                || string.IsNullOrEmpty(computeQuery)
+                || this.targetEdmType == null)
```
mikepizzo: >|| this.targetEdmType == null [](start = 16, length = 29)

Under what situations will targetEdmType be null (a dynamic property?) and what are the ramifications of returning a null in this case? Is the compute clause silently ignored? Do we have a test case for this case?

src/Microsoft.OData.Core/UriParser/SyntacticAst/ComputeToken.cs
mtroth-microsoft: ISyntacticTreeVisitor<T> is a public interface. It does not have a Visit override that specifically handles ComputeTokens. I opted not to introduce a breaking change since visitation doesn't appear to be required to address the scenarios specified in this issue. I also opted not to introduce a larger non-breaking change for the same reason.

---
In reply to: [123611743](https://github.com/OData/odata.net/pull/827#discussion_r123611743) [](ancestors = 123611743)

src/Microsoft.OData.Core/UriParser/ODataQueryOptionParser.cs
mtroth-microsoft: there are two test cases that address returning null, i believe. both seem to be based on the assertion that all query options must be able to handle null scenarios.

---
In reply to: [123615935](https://github.com/OData/odata.net/pull/827#discussion_r123615935) [](ancestors = 123615935)

PR #835, opened 2017-06-14
Cleanup skipped tests in Microsoft.OData.Core.Tests.

robward-ms:
### Description
Skipped tests have been resolved by:
1.) Unskipping the test
2.) Modifying the test to match current product behavior
3.) Commenting out the test where matching product work is not complete, add the matching GitHiub issue
4.) Removing the test

- [ x ] Test cases added

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/Evaluation/ODataEntryMetadataContextTest.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/Evaluation/ODataEntryMetadataContextTest.cs
@@ -209,14 +209,6 @@ namespace Microsoft.OData.Tests.Evaluation
             this.entryMetadataContextWithoutModel.ETagProperties.Should().BeEmpty();
         }
 
-        [Fact(Skip = "we should remove this case because we need model to calc ETag")]
-        public void ETagPropertiesShouldReturnEmptyForEntryMetadataContextWithModelWithoutETagsAndThereAreSerializationInfoOnEntryProperties()
-        {
-            var odataEntry = new ODataResource { Properties = new[] { new ODataProperty { Name = "Name", Value = null, SerializationInfo = new ODataPropertySerializationInfo { PropertyKind = ODataPropertyKind.ETag } } } };
-            var entryMetadataContext = ODataResourceMetadataContext.Create(odataEntry, this.typeContext, /*serializationInfo*/null, new EdmEntityType("ns", "TypeName"), new TestMetadataContext(), SelectedPropertiesNode.EntireSubtree);
-            entryMetadataContext.ETagProperties.Should().BeEmpty();
-        }
-
         [Fact]
         public void ETagPropertiesShouldThrowWhenSerializationInfoIsSetAndPropertyValueIsNonPrimitive()
         {
@@ -232,28 +224,6 @@ namespace Microsoft.OData.Tests.Evaluation
             this.entryMetadataContextWithoutModel.ETagProperties.Should().HaveCount(1).And.Contain(p => p.Key == "Name" && p.Value == null);
         }
 
-        [Fact(Skip = "we should remove this case because we need model to calc ETag")]
-        public void ETagPropertiesShouldThrowWhenMetadataIsPresentAndPropertyValueIsNonPrimitive()
-        {
-            this.entry.Properties = new[] { new ODataProperty { Name = "Name2", Value = new ODataCollectionValue() } };
-            Action test = () => this.entryMetadataContextWithModel.ETagProperties.Should().NotBeNull();
-            test.ShouldThrow<ODataException>(Strings.ODataResourceMetadataContext_KeyOrETagValuesMustBePrimitiveValues("Name2", "ns.TypeName"));
-        }
-
-        [Fact(Skip = "we should remove this case because we need model to calc ETag")]
-        public void ETagPropertiesShouldNotThrowWhenMetadataIsPresentAndPropertyValueIsNull()
-        {
-            this.entry.Properties = new[] { new ODataProperty { Name = "Name2", Value = null }, new ODataProperty { Name = "Name3", Value = null }};
-            this.entryMetadataContextWithModel.ETagProperties.Should().HaveCount(2).And.Contain(p => p.Key == "Name2" && p.Value == null).And.Contain(p => p.Key == "Name3" && p.Value == null);
-        }
-
-        [Fact(Skip = "we should remove this case because we need model to calc ETag")]
-        public void ETagPropertiesShouldThrowWhenETagPropertyInMetadataIsNotInEntry()
-        {
-            Action test = () => this.entryMetadataContextWithModel.ETagProperties.Should().NotBeNull();
-            test.ShouldThrow<ODataException>(Strings.EdmValueUtils_PropertyDoesntExist("ns.TypeName", "Name2"));
-        }
-
         [Fact]
         public void ShouldGetETagPropertiesBasedOnSerializationInfo()
         {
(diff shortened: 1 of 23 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/Evaluation/ODataEntryMetadataContextTest.cs
```diff
@@ -209,14 +209,6 @@ public void ETagPropertiesShouldReturnEmptyForEntryMetadataContextWithoutModelAn
             this.entryMetadataContextWithoutModel.ETagProperties.Should().BeEmpty();
         }
 
-        [Fact(Skip = "we should remove this case because we need model to calc ETag")]
```
biaol-odata: In case as such where the test's dependency is not satisfied, should we comment out the tests (well pretty much same as using skip but w/o impacts to summary)instead of deleting it unless we are sure it is not needed anymore?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FullUriFunctionalTests.cs
```diff
@@ -71,52 +71,63 @@ public void SelectOrExpandCanOnlyBeCalledOnEntity()
...
+        [Fact]
         public void ParseSelectExpand()
         {
             ODataUriParser parser = new ODataUriParser(HardCodedTestModel.TestModel, new Uri("http://www.odata.com/OData"), new Uri("http://www.odata.com/OData/Dogs?$select=Color, MyPeople&$expand=MyPeople"));
             ODataUri parsedUri = parser.ParseUri();
-            parsedUri.SelectAndExpand.SelectedItems.First().ShouldBePathSelectionItem(new ODataPath(new PropertySegment(HardCodedTestModel.GetDogColorProp())));
-            var myPeopleExpand = parsedUri.SelectAndExpand.SelectedItems.Last().ShouldBeSelectedItemOfType<ExpandedNavigationSelectItem>().And;
+            parsedUri.SelectAndExpand.SelectedItems.Count().Should().Be(3);
+            parsedUri.SelectAndExpand.SelectedItems.ElementAt(1).ShouldBePathSelectionItem(new ODataPath(new PropertySegment(HardCodedTestModel.GetDogColorProp())));
+            var myPeopleExpand = parsedUri.SelectAndExpand.SelectedItems.First().ShouldBeSelectedItemOfType<ExpandedNavigationSelectItem>().And;
```
biaol-odata: q: don't understand why verification on Last() is changed to First(). Can we add verification related to Last() as  well? 
IE Browser cannot view the whole line of the code, limiting max length to 120-ish might be helpful?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FullUriFunctionalTests.cs
```diff
@@ -71,52 +71,63 @@ public void SelectOrExpandCanOnlyBeCalledOnEntity()
...
 
-        [Fact(Skip = "This test currently fails.")]
+        [Fact]
         public void TopOnlyValidOnEntities()
         {
             ODataUriParser parser = new ODataUriParser(HardCodedTestModel.TestModel, new Uri("http://www.odata.com/OData"), new Uri("http://www.odata.com/OData/People(1)/MyAddress?$top=1"));
-            Action parseWitnNonEntity = () => parser.ParseUri();
-            parseWitnNonEntity.ShouldThrow<ODataException>().WithMessage(ODataErrorStrings.MetadataBinder_QueryOptionNotApplicable("$top"));
+            ODataUri parsedUri = parser.ParseUri();
+            parsedUri.Top.Should().Be(1);
```
biaol-odata: Assuming the Uri is valid, should we update the method test name?
q: why it verification is on parsedUri.Top? Can it be something else more specific from the parsed result?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/SelectPathSegmentTokenBinderTests.cs
```diff
@@ -159,19 +159,14 @@ public void ValidateThrowInFindOperationsByBindingParameterTypeHierarchyExceptio
...
+        [Fact]
         public void ValidateNonCatchableExceptionsThrowInFindOperationsByBindingParameterTypeHierarchyExceptionAndSurfaces()
         {
-            var model = new FindOperationsByBindingParameterTypeHierarchyThrowingStackOverflowEdmModel();
+            var model = new FindOperationsByBindingParameterTypeHierarchyThrowingNonCatchableExceptionEdmModel();
             IEdmEntityType entityType = new EdmEntityType("n", "EntityType");
             ODataPathSegment foundPathSegment = null;
             Action test = () => SelectPathSegmentTokenBinder.TryBindAsOperation(new SystemToken("foo", null), model, entityType, out foundPathSegment);
-
-#if NETCOREAPP1_0
```
biaol-odata: should we preserve this if/else ?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/ExpandOptionParserTests.cs
```diff
@@ -139,11 +139,12 @@ public void DeepNestedExpandsWithSemisAreAllowed()
...
 
-        [Fact(Skip = "This test currently fails.")]
-        public void CannotHaveAnOptionMoreThanOnce()
+        [Fact]
+        public void CanHaveAnOptionMoreThanOnce()
         {
-            Action parse = () => this.ParseExpandOptions("($filter=true; $filter=false)");
-            parse.ShouldThrow<ODataException>().WithMessage("TODO");
+            // This test was once written to support only one instance of filter
+            // but oin 6.x and 7.x, it is supported.
```
biaol-odata: nit: typo "oin"

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/SelectPathSegmentTokenBinderTests.cs
AlanWong-MS: I think Rob's proposed change makes sense and is cleaner than what I had there so we can forego the #if/else.

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/SelectPathSegmentTokenBinderTests.cs
```diff
@@ -235,15 +230,11 @@ public override IEnumerable<IEdmOperation> FindDeclaredBoundOperations(IEdmType
...
+        private class FindOperationsByBindingParameterTypeHierarchyThrowingNonCatchableExceptionEdmModel : EdmModel
         {
             public override IEnumerable<IEdmOperation> FindDeclaredBoundOperations(IEdmType bindingType)
             {
-#if NETCOREAPP1_0
-                throw new Exception("Oh no!");
-#else
-                throw new StackOverflowException("Oh no!");
-#endif
+                throw new OutOfMemoryException("Oh no!");
```
AlanWong-MS: Side comment: the message irked me a little because it was uninformative. :P I probably should've changed it when I saw it to give more context in some way.

test/FunctionalTests/Microsoft.OData.Core.Tests/Evaluation/ODataEntryMetadataContextTest.cs
robward-ms: I looked at rewriting the test, it looked like it was impacted by the 7.x API change. I took the hint from the skip comment and just removed them. We don't seem to have test gaps for ETags should removing should be safe.

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/SelectPathSegmentTokenBinderTests.cs
robward-ms: This #if is here because StackOverflowException is not in NetCore but the product code ignore StackOverflowException anyway and only responds to OutOfMemoryException, which is in both platforms, hence no need for #if.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FullUriFunctionalTests.cs
robward-ms: Done. Same for Skip...

PR #842, opened 2017-06-16
Re-enable priority test cases in Microsoft.Data.Web.UnitTests suite

AlanWong-MS:
### Description
This change modifies and re-enables relevant tests that had been skipped in the Microsoft.Data.Web.UnitTests test suite. These tests help fill important missing test gaps for our code coverage. A separate pull request will be sent out to address the rest of the skipped tests.

- [ x ] Test cases added

### DIFF
--- a/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
+++ b/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
@@ -962,7 +962,7 @@ namespace Microsoft.OData.Edm
                 catch (NotImplementedException)
                 {
                     // model.EntityContainer can throw NotImplementedException
-                    return foundEntitySet;
+                    return null;
                 }
             }
 
@@ -992,7 +992,7 @@ namespace Microsoft.OData.Edm
                 catch (NotImplementedException)
                 {
                     // model.EntityContainer can throw NotImplementedException
-                    return foundSingleton;
+                    return null;
                 }
             }
 
--- a/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
+++ b/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
@@ -951,10 +951,18 @@ namespace Microsoft.OData.Edm
             if (!model.TryFindContainerQualifiedEntitySet(qualifiedName, out foundEntitySet))
             {
                 // try searching by entity set name in container and extended containers:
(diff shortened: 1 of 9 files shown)


src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -951,10 +951,18 @@ public static IEdmEntitySet FindDeclaredEntitySet(this IEdmModel model, string q
...
-                if (container != null)
+                try
+                {
+                    IEdmEntityContainer container = model.EntityContainer;
+                    if (container != null)
+                    {
+                        return container.FindEntitySetExtended(qualifiedName);
+                    }
+                }
+                catch (NotImplementedException)
```
AlanWong-MS: If we perform any logging, I think it'd be appropriate to track any exceptions here.

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -951,10 +951,18 @@ public static IEdmEntitySet FindDeclaredEntitySet(this IEdmModel model, string q
...
                 {
-                    return container.FindEntitySetExtended(qualifiedName);
+                    // model.EntityContainer can throw NotImplementedException
+                    return foundEntitySet;
```
robward-ms: >foundEntitySet [](start = 27, length = 14)

Is this null in this case? Would it make more sense to be explicit and return null?  Same below.

robward-ms:
>         public static XmlDocument VerifyXPaths(Stream resultStream, string responseFormat, string[] web3sXPaths, string[] jsonXPaths, string[] atomXPaths)

This variable is not used?

---
Refers to: test/FunctionalTests/Tests/DataServices/UnitTests/ServerUnitTests/Utils/UnitTestsUtil.cs:258 in a57c94d. [](commit_id = a57c94dcb47125db8f1ca4758eba154c4f0c19aa, deletion_comment = False)

robward-ms:
:clock1:

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
AlanWong-MS: Yes, that's a good idea. I'll change for the case below as well.

mtroth-microsoft:
:shipit:

PR #845, opened 2017-06-20
add test cases for the ODataAtomErrorDeserializer coverage

xuzhg:
### Issues

Related to MQ/QP.

### Description

* Originally, there are 113/100% codes missing. We should to add test cases to cover most of codes in it.
* After added, 

image

- [ x ] Test cases added

### Additional work necessary
* None


xuzhg:
Merged at: https://github.com/OData/odata.net/commit/3d94ab583edf32cc65ec74db0ccd331129e1bf8f

Issue #801, opened 2017-04-28
Supporting Primitive Type Casts


ysanghi:
*Short summary (3-5 sentences) describing the issue.*  
It will be nice to have the ability to do cast properties as primitive types in $select, $filter, $order by. Today you can only use non primitive types in the type segment.

### Assemblies affected
OData.Net 7.x

### Reproduce steps
CurrentCollections('id')/Items?$select=Prop1/Edm.String&$filter=Prop2/Edm.String%20eq%20'sdb'

### Expected result
Found a path with multiple navigation properties or a bad complex property path in a select clause. Please reword your query such that each level of select or expand only contains either TypeSegments or Properties.

### Actual result
It work as normal type segment

mikepizzo:
Addressed in PR #848.

Issue #747, opened 2017-01-10
Cannot select dynamic property of a dynamic property


ysanghi:
Select of dynamic properties of dynamic properties is not supported.

### Assemblies affected
OData .NET Lib 7.0

### Reproduce steps
Try to select dynamic property of a dynamic property on an open type. Let's assume there is an dynamic complex type called 'foo' which has a sub property called 'bar'. Trying to select $select = foo/bar would throw the following error:
Found a path with multiple navigation properties or a bad complex property path in a select clause. Please reword your query such that each level of select or expand only contains either TypeSegments or Properties.

### Expected result
No error with a list or stack of path segments.

### Actual result
Exceptions is being thrown: Found a path with multiple navigation properties or a bad complex property path in a select clause. Please reword your query such that each level of select or expand only contains either TypeSegments or Properties.

PR #848, opened 2017-06-21
Add Support for PrimitiveType casts

mtroth-microsoft:
### Issues

This pull request fixes issue #801 and #747 

### Description

Modify the binding operations so that type casts and dynamic properties on dynamic complex properties can be successfully bound to Nodes during the parsing process.

- [ x ] Test cases added

### Additional work necessary

No documentation changes needed.

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/DottedIdentifierBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/DottedIdentifierBinder.cs
@@ -82,7 +82,15 @@ namespace Microsoft.OData.UriParser
                     IEdmTypeReference edmTypeReference = UriEdmHelpers.FindTypeFromModel(state.Model, dottedIdentifierToken.Identifier, this.Resolver).ToTypeReference();
                     if (edmTypeReference is IEdmPrimitiveTypeReference || edmTypeReference is IEdmEnumTypeReference)
                     {
-                        return new ConstantNode(dottedIdentifierToken.Identifier, dottedIdentifierToken.Identifier);
+                        IEdmPrimitiveType childPrimitiveType = childType as IEdmPrimitiveType;
+                        if (childPrimitiveType != null && dottedIdentifierToken.NextToken != null)
+                        {
+                            return new SingleValueCastNode(parent as SingleValueNode, childPrimitiveType);
+                        }
+                        else
+                        {
+                            return new ConstantNode(dottedIdentifierToken.Identifier, dottedIdentifierToken.Identifier);
+                        }
                     }
                     else
                     {
--- a/src/Microsoft.OData.Core/UriParser/SemanticAst/SingleValueCastNode.cs
+++ b/src/Microsoft.OData.Core/UriParser/SemanticAst/SingleValueCastNode.cs
@@ -1,0 +1,76 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="SingleValueCastNode.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using Microsoft.OData.Edm;
+
+namespace Microsoft.OData.UriParser
+{
+    /// <summary>
+    /// Node representing a type segment that casts a single primitive value node.
+    /// </summary>
+    public sealed class SingleValueCastNode : SingleValueNode
+    {
+        /// <summary>
+        /// The resource that we're casting to a different type.
+        /// </summary>
+        private readonly SingleValueNode source;
+
+        /// <summary>
+        /// The target type that the source is cast to.
+        /// </summary>
+        private readonly IEdmPrimitiveTypeReference primitiveTypeReference;
+
+        /// <summary>
+        /// Created a SingleValueCastNode with the given source node and the given type to cast to.
+        /// </summary>
+        /// <param name="source"> Source <see cref="SingleValueNode"/> that is being cast.</param>
+        /// <param name="primitiveType">Type to cast to.</param>
+        /// <exception cref="System.ArgumentNullException">Throws if the input primitiveType is null.</exception>
+        public SingleValueCastNode(SingleValueNode source, IEdmPrimitiveType primitiveType)
+        {
+            ExceptionUtils.CheckArgumentNotNull(source, "source");
+            ExceptionUtils.CheckArgumentNotNull(primitiveType, "primitiveType");
+            this.source = source;
+            this.primitiveTypeReference = new EdmPrimitiveTypeReference(primitiveType, true);
+        }
+
+        /// <summary>
+        /// Gets the property that we're casting to a different type.
+        /// </summary>
+        public SingleValueNode Source
+        {
+            get { return this.source; }
+        }
+
+        /// <summary>
+        /// Gets the target type that the source is cast to.
+        /// </summary>
+        public override IEdmTypeReference TypeReference
+        {
+            get { return this.primitiveTypeReference; }
+        }
+
+        /// <summary>
+        /// Gets the kind of this query node.
+        /// </summary>
+        internal override InternalQueryNodeKind InternalKind
+        {
+            get { return InternalQueryNodeKind.SingleValueCast; }
+        }
+
+        /// <summary>
+        /// Accept a <see cref="QueryNodeVisitor{T}"/> that walks a tree of <see cref="QueryNode"/>s.
+        /// </summary>
+        /// <typeparam name="T">Type that the visitor will return after visiting this token.</typeparam>
+        /// <param name="visitor">An implementation of the visitor interface.</param>
+        /// <returns>An object whose type is determined by the type parameter of the visitor.</returns>
+        public override T Accept<T>(QueryNodeVisitor<T> visitor)
+        {
+            ExceptionUtils.CheckArgumentNotNull(visitor, "visitor");
+            return visitor.Visit(this);
+        }
+    }
+}
--- a/src/Microsoft.OData.Core/UriParser/TreeNodeKinds/QueryNodeKind.cs
+++ b/src/Microsoft.OData.Core/UriParser/TreeNodeKinds/QueryNodeKind.cs
@@ -158,6 +158,11 @@ namespace Microsoft.OData.UriParser
         /// Count of a collection contains primitive or enum or complex or entity type.
         /// </summary>
         Count = InternalQueryNodeKind.Count,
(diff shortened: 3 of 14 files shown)


src/Microsoft.OData.Core/UriParser/Binders/DottedIdentifierBinder.cs
```diff
@@ -55,6 +55,7 @@ internal QueryNode BindDottedIdentifier(DottedIdentifierToken dottedIdentifierTo
             SingleResourceNode parentAsSingleResource = parent as SingleResourceNode;
             IEdmSchemaType childType = UriEdmHelpers.FindTypeFromModel(state.Model, dottedIdentifierToken.Identifier, this.Resolver);
             IEdmStructuredType childStructuredType = childType as IEdmStructuredType;
+            IEdmPrimitiveType childPrimitiveType = childType as IEdmPrimitiveType;
```
mikepizzo: Consider moving this line to just before line 86 (if (childPrimitiveType != null...), since childPrimitiveType is only used within that block.

src/Microsoft.OData.Core/UriParser/SemanticAst/SinglePropertyCastNode.cs
```diff
@@ -0,0 +1,75 @@
...
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using Microsoft.OData.Edm;
+
+namespace Microsoft.OData.UriParser
+{
+    /// <summary>
+    /// Node representing a type segment that casts a single entity/complex parent node.
```
mikepizzo: >single entity/complex parent node [](start = 54, length = 33)

Update description (...casts a single primitive value node)

src/Microsoft.OData.Core/UriParser/SemanticAst/SinglePropertyCastNode.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// </summary>
+        private readonly IEdmPrimitiveTypeReference primitiveTypeReference;
+
+        /// <summary>
+        /// Created a SinglePropertyCastNode with the given source node and the given type to cast to.
+        /// </summary>
+        /// <param name="source"> Source <see cref="SingleValueNode"/> that is being cast.</param>
+        /// <param name="primitiveType">Type to cast to.</param>
+        /// <exception cref="System.ArgumentNullException">Throws if the input entityType is null.</exception>
+        public SinglePropertyCastNode(SingleValueNode source, IEdmPrimitiveType primitiveType)
```
mikepizzo: >source [](start = 54, length = 6)

Can source be null? If not, should we add an not null check here? I notice that we don't on SingleResourceCastNode, which also seems strange.

src/Microsoft.OData.Core/UriParser/SemanticAst/SinglePropertyCastNode.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// The target type that the source is cast to.
+        /// </summary>
+        private readonly IEdmPrimitiveTypeReference primitiveTypeReference;
+
+        /// <summary>
+        /// Created a SinglePropertyCastNode with the given source node and the given type to cast to.
+        /// </summary>
+        /// <param name="source"> Source <see cref="SingleValueNode"/> that is being cast.</param>
+        /// <param name="primitiveType">Type to cast to.</param>
+        /// <exception cref="System.ArgumentNullException">Throws if the input entityType is null.</exception>
```
mikepizzo: >entityType [](start = 79, length = 10)

primitiveType

src/Microsoft.OData.Core/UriParser/TreeNodeKinds/QueryNodeKind.cs
```diff
@@ -158,6 +158,11 @@ public enum QueryNodeKind
         /// Count of a collection contains primitive or enum or complex or entity type.
         /// </summary>
         Count = InternalQueryNodeKind.Count,
+
+        /// <summary>
+        /// Cast on a single property.
+        /// </summary>
+        SinglePropertyCast = InternalQueryNodeKind.SinglePropertyCast,
```
mikepizzo: >SinglePropertyCast [](start = 8, length = 18)

I wonder if we should call it SinglePrimitiveCast (here and elsewhere)

src/Microsoft.OData.Core/UriParser/Visitors/SelectPropertyVisitor.cs
```diff
@@ -151,17 +156,38 @@ private void ProcessTokenAsPath(NonSystemToken tokenIn)
...
+                        EdmPrimitiveTypeKind nextTypeKind = EdmCoreModel.Instance.GetPrimitiveTypeKind(nextToken.Identifier);
+                        IEdmPrimitiveType castType = EdmCoreModel.Instance.GetPrimitiveType(nextTypeKind);
+                        if (castType != null)
+                        {
+                            lastSegment = new TypeSegment(castType, castType, null);
+                        }
+                        else if (dynamicPath != null)
+                        {
+                            lastSegment = new DynamicPathSegment(nextToken.Identifier);
+                        }
```
mikepizzo: move this outside of this nested else, so logic is if (primitiveType == null && dynamicPath == null) {/*handle complex type*/} else if (dynamicPath != null) {/*handle dynamic path segment*/} else {/*handle primitive type cast segment*/ else{throw}}

test/FunctionalTests/Service/Microsoft/OData/Service/Parsing/NodeToExpressionTranslator.cs
```diff
@@ -348,6 +348,28 @@ public override Expression Visit(SingleResourceCastNode node)
...
+        /// <param name="node">The node to translate.</param>
+        /// <returns>The translated expression.</returns>
+        public override Expression Visit(SinglePropertyCastNode node)
+        {
+            WebUtil.CheckArgumentNull(node, "node");
+
+            // Whenever we encounter the type segment, we need to only verify that the MPV is set to 4.0 or higher.
+            // There is no need to check for request DSV, request MaxDSV since there are no protocol changes in
+            // the payload for uri's with type identifier.
+            this.verifyProtocolVersion(ODataProtocolVersion.V4);
```
mikepizzo: Do we still need these Protocol Version checks? Is the code used by the 3.x stack?

src/Microsoft.OData.Core/UriParser/TreeNodeKinds/QueryNodeKind.cs
mtroth-microsoft: Resource / Property makes more sense than Resource / Primitive.

---
In reply to: [124892629](https://github.com/OData/odata.net/pull/848#discussion_r124892629) [](ancestors = 124892629)

src/Microsoft.OData.Core/UriParser/Visitors/SelectPropertyVisitor.cs
mtroth-microsoft: the difference is just sugar and would force the type derivation to occur even in cases when it is not necessary.

---
In reply to: [124894287](https://github.com/OData/odata.net/pull/848#discussion_r124894287) [](ancestors = 124894287)

src/Microsoft.OData.Core/UriParser/TreeNodeKinds/QueryNodeKind.cs
mtroth-microsoft: the cast works on complex properties as well as primitive properties.

---
In reply to: [125081678](https://github.com/OData/odata.net/pull/848#discussion_r125081678) [](ancestors = 125081678,124892629)

test/FunctionalTests/Service/Microsoft/OData/Service/Parsing/NodeToExpressionTranslator.cs
mtroth-microsoft: I believe we still do need this. the test runs in the maintenance 6x branch and since this class won't exist there, we'll need to expect the failure.

---
In reply to: [124896810](https://github.com/OData/odata.net/pull/848#discussion_r124896810) [](ancestors = 124896810)

test/FunctionalTests/Tests/DataOData/Tests/OData.Common.Tests/PublicApi/PublicApi.bsl
```diff
@@ -5173,6 +5173,7 @@ public enum Microsoft.OData.UriParser.QueryNodeKind : int {
 	SearchTerm = 24
 	SingleComplexNode = 27
 	SingleNavigationNode = 11
+	SinglePropertyCast = 29
```
mikepizzo: >SinglePropertyCast [](start = 1, length = 18)

Should rename to SingleValueCast to match renamed SingleValueCastNode.

src/Microsoft.OData.Core/UriParser/SemanticAst/SingleValueCastNode.cs
```diff
@@ -0,0 +1,76 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="SinglePropertyCastNode.cs" company="Microsoft">
```
mikepizzo: Update to SingleValueCastNode.cs

src/Microsoft.OData.Core/UriParser/TreeNodeKinds/QueryNodeKind.cs
mikepizzo: Rename to SingleValueCast as per discussion.

---
In reply to: [125102394](https://github.com/OData/odata.net/pull/848#discussion_r125102394) [](ancestors = 125102394,125081678,124892629)

src/Microsoft.OData.Core/UriParser/TreeNodeKinds/QueryNodeKind.cs
```diff
@@ -309,5 +314,10 @@ internal enum InternalQueryNodeKind
         /// Node describing count of a collection contains primitive or enum or complex or entity type.
         /// </summary>
         Count = 28,
+
+        /// <summary>
+        /// Cast on a single property.
+        /// </summary>
+        SinglePropertyCast = 29,
```
mikepizzo: >SinglePropertyCast [](start = 8, length = 18)

Renamed to SingleValueCast

mikepizzo:
:shipit:

Issue #846, opened 2017-06-20
Raw value serializer for Spatial type not correct


xuzhg:
In [RawValueWriter.cs](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/RawValueWriter.cs),  There's a function named [`WriteRawValue(object value)`](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/RawValueWriter.cs#L108), in which it has the following codes to write the Spatial object:
```C#
else if (value is Geometry || value is Geography)
{
   PrimitiveConverter.Instance.TryWriteAtom(value, textWriter);
}
```
However, it should call `WriteJsonLight`, not `TryWriteAtom`.

### Assemblies affected

*OData .Net lib*

### Reproduce steps

For example:
```C#
RawValueWriter target = new RawValueWriter(this.settings, this.stream, new UTF32Encoding());
var value2 = GeometryPoint.Create(1.2, 3.16);
target.WriteRawValue(value2);
```

### Expected result

```json
{"type":"Point","coordinates":[1.2,3.16],"crs":{"type":"name","properties":{"name":"EPSG:4326"}}}
```
### Actual result

```txt
RID=4326;POINT (1.2 3.16)
```

xuzhg:
I have the following example to illustrate the problem:

```xml

<?xml version="1.0" encoding="utf-8"?>
<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
  <edmx:DataServices>
    <Schema Namespace="WaoTest" xmlns="http://docs.oasis-open.org/odata/ns/edm">
      <EntityType Name="Customer">
        <Key>
          <PropertyRef Name="Id" />
        </Key>
        <Property Name="Id" Type="Edm.Int32" Nullable="false" />
        <Property Name="Location" Type="Edm.GeographyPoint" SRID="4326" />
      </EntityType>
    </Schema>
    <Schema Namespace="Default" xmlns="http://docs.oasis-open.org/odata/ns/edm">
      <EntityContainer Name="Container">
        <EntitySet Name="Customers" EntityType="WaoTest.Customer" />
      </EntityContainer>
    </Schema>
  </edmx:DataServices>
</edmx:Edmx>
```
If I query: `~/Customers(1)/Location`

I can get:
```json
{
  "@odata.context":"http://localhost/odata/$metadata#Customers(1)/Location","value":{
    "type":"Point","coordinates":[
[...]

PR #849, opened 2017-06-21
Clean up codes related to Spatial type deseializer

xuzhg:
### Issues

Part of related to https://github.com/OData/odata.net/issues/846

### Description

The output of raw value of Spatial is JSON format. So, such codes are needless because they are assuming the xml format of Spatial date.

- [    ] Test cases added
- [   ] Build and test with one-click build and test script passed

### Additional work necessary

It's part of code coverage related to Metadata subfolder.


xuzhg:
Merged at: 02d336caf52312ea6a0af33d4f5f22ebcc9b0658

PR #850, opened 2017-06-21
Add odata.core.test project into full solution

xuzhg:
### Issues

Add odata.core.test project into full solution

### Description

Full solution doesn't include the OData.Core.Test project in which most of test cases related to the OData.Core lib includes.

- [   ] Test cases added

### Additional work necessary

It's related to code coverage.


xuzhg:
Merged at: 28e5fa7007ffe1eaed83bf3f29c9813eb6840b56

PR #852, opened 2017-06-23
Clean codes in EdmLibExtensions for coverage

xuzhg:
### Description

Related to test coverage.

So far, it has the following code coverage:

image

- [ x ] Test cases added

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/EndPathBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/EndPathBinder.cs
@@ -111,7 +111,7 @@ namespace Microsoft.OData.UriParser
         internal SingleValueOpenPropertyAccessNode GeneratePropertyAccessQueryForOpenType(EndPathToken endPathToken, SingleValueNode parentNode)
         {
             if (parentNode.TypeReference == null ||
-                parentNode.TypeReference.Definition.IsOpenType() ||
+                parentNode.TypeReference.Definition.IsOpen() ||
                 IsAggregatedProperty(endPathToken.Identifier))
             {
                 return new SingleValueOpenPropertyAccessNode(parentNode, endPathToken.Identifier);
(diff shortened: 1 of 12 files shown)


src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1641,6 +1641,35 @@ public static bool IsOpen(this IEdmStructuredTypeReference type)
...
+            if (structuredType != null)
+            {
+                return structuredType.IsOpen;
+            }
+
+            // If its a collection, return whether its element type is open.
+            // This is because when processing a navigation property, the target type
+            // may be a collection type even though a key expression has been applied.
+            // This will be cleaned up in a subsequent change.
+            // TODO: when SingleResult is removed from the semantic path parser, change this to return false.
```
AlanWong-MS: Do we have a way to track this TODO? Should we open a GitHub issue and reference the number here?

src/Microsoft.OData.Core/UriParser/Binders/EndPathBinder.cs
```diff
@@ -111,7 +111,7 @@ internal static SingleValueNode CreateParentFromImplicitRangeVariable(BindingSta
         internal SingleValueOpenPropertyAccessNode GeneratePropertyAccessQueryForOpenType(EndPathToken endPathToken, SingleValueNode parentNode)
         {
             if (parentNode.TypeReference == null ||
-                parentNode.TypeReference.Definition.IsOpenType() ||
```
AlanWong-MS: What's the reasoning for moving from .IsOpenType() to .IsOpen()?

xuzhg: EdmLib has `IsOpen()` extension method defined ,  `IsOpenType()` was similar function defined in ODL and I removed it.

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
xuzhg: Oh, I think we should remove this "TODO". I just copied from the ODL lib.

AlanWong-MS:
:shipit:

xuzhg:
Merged at: 5ece8b56164f44c4b09101362953e0a76b6591d0

PR #856, opened 2017-06-27
Raw Value serializer output json string for Spatial type

xuzhg:
### Issues

* #846 

### Description

In `RawValueWriter.cs`, There's a function named `WriteRawValue(object value)`, in which it has the following codes to write the Spatial object:

```C#
else if (value is Geometry || value is Geography)
{
   PrimitiveConverter.Instance.TryWriteAtom(value, textWriter);
}
```
However, it should call `WriteJsonLight`, not `TryWriteAtom`.

For example:
```C#
RawValueWriter target = new RawValueWriter(this.settings, this.stream, new UTF32Encoding());
var value2 = GeometryPoint.Create(1.2, 3.16);
target.WriteRawValue(value2);
```

Expected result

```json
{"type":"Point","coordinates":[1.2,3.16],"crs":{"type":"name","properties":{"name":"EPSG:4326"}}}
```
Actual result
```txt
RID=4326;POINT (1.2 3.16)
```

- [ x ] Test cases added

### DIFF
--- a/src/Microsoft.OData.Core/PrimitiveConverter.cs
+++ b/src/Microsoft.OData.Core/PrimitiveConverter.cs
@@ -78,34 +78,6 @@ namespace Microsoft.OData
         }
 
         /// <summary>
-        /// Try to write the XML representation of <paramref name="instance"/> to the specified <paramref name="writer"/>
-        /// </summary>
-        /// <param name="instance">Object to convert to XML representation.</param>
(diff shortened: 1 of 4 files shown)


xuzhg:
Merged at: https://github.com/OData/odata.net/pull/856/commits/f49b05ad4fcd7fa66ba0852708ede1848b033bca

PR #862, opened 2017-06-29
Increase coverage for JsonLight in Microsoft.OData.Core

biaol-odata:
### Issues

*This pull request is for QP #194197.*

### Description

*Added test coverage for major gaps identified thru VS code analysis report.*
*Buddy build passed: https://identitydivision.visualstudio.com/OData/_build/index?buildId=7728&_a=summary*
*Coverage data for Microsoft.OData.JsonLight: debug:91.02%; release:89.93% .*

- [ x ] Test cases added

### Additional work necessary


sln/.nuget/packages.config
```diff
@@ -23,4 +23,5 @@
   <package id="xunit.extensibility.core" version="2.1.0" />
   <package id="xunit.extensibility.execution" version="2.1.0" />
   <package id="xunit.runner.visualstudio" version="2.1.0" />
+  <package id="Microsoft.NETCore.Portable.Compatibility" version="1.0.1" />
```
AlanWong-MS: nit: re-arrange to alphabetical order for consistency

biaol-odata:
Alan, for your second comment: The lock.json files are required to be updated due to the additional package in the global package config, otherwise VS2015 have issue resolving the package references in the related projects.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightReaderUtilsTests.cs
```diff
@@ -0,0 +1,90 @@
...
+            }
+            catch (TargetInvocationException e)
+            {
+                e.InnerException.Should().BeOfType<ODataException>();
+                e.InnerException.Message.Should()
+                    .Be("An internal error 'ODataJsonLightReader_ReadResourceStart' occurred.");
+                return;
+            }
+
+            false.Should().BeTrue("Exception is expected to be thrown and test execution should not reach here.");
```
robward-ms: Look at the test CannotWriteExpandedNavigationPropertyOutsideDeltaEntry for an example of how to assert a particular exception, maybe make it like CannotWriteExpandedNavigationPropertyOutsideDeltaEntry to be consistent in catching expected exceptions?

AlanWong-MS: I missed this one. Interesting... this was compilable?

biaol-odata: @robward-ms The actions.ShouldThrow<ExceptionType>.WithMessage() is not entirely applicable here because it is a Reflection call, which always throws the TargetInvocationException w/ inner exception (here as ODataException). In other words, the ExcpetionType would be always TargetInvocationException but we want to assert the inner exception of ODataException. Therefor e the catch / innerException is used above.
I don't like the false.should().BeTrue() either. But, for FluenctAssertion, I couldn't anything such as Assert.Failed(msg) available in other assertion libraries. I am now updating it to "exceptionThrown.Should().BeTrue()" hopefully it is better.

@AlanWong-MS yes it was compilable.

biaol-odata:
@robward-ms The actions.ShouldThrow.WithMessage() is not entirely applicable here because it is a Reflection call, which always throws the TargetInvocationException w/ inner exception (here as ODataException). In other words, the ExcpetionType would be always TargetInvocationException but we want to assert the inner exception of ODataException. Therefor e the catch / innerException is used above.
I don't like the false.should().BeTrue() either. But, for FluenctAssertion, I couldn't anything such as Assert.Failed(msg) available in other assertion libraries. I am now updating it to "exceptionThrown.Should().BeTrue()" hopefully it is better.
@AlanWong-MS yes it was compilable.

PR #870, opened 2017-07-05
Standardize on EntityFramework 5.0 for tests to reduce race conditions

robward-ms:
### Description
Standardize on EntityFramework 5.0 for tests to reduce race conditions
for EntityFramework dependent tests.

- [   ] Test cases added


biaol-odata:
LGTM

PR #871, opened 2017-07-05
Modify database strategy to eliminate Attach.

robward-ms:
### Description
Modify database strategy to eliminate Attach. This approach takes longer to run
but eliminates one of the reasons tests must be run ad Administrator.

See this for details:
https://stackoverflow.com/questions/2330439/access-is-denied-when-attaching-a-database

- [   ] Test cases added


test/FunctionalTests/Framework/TestExecutionLayer/AstoriaDatabase.cs
```diff
@@ -17,6 +17,7 @@ namespace System.Data.Test.Astoria.TestExecutionLayer
     using System.Threading;
     using Microsoft.Test.ModuleCore;
     using System.Security;
+    using System.Text;
```
biaol-odata: nit: import order for "Microsoft.Test.ModuleCore", if you can clean it up?

test/FunctionalTests/Framework/TestExecutionLayer/AstoriaDatabase.cs
```diff
@@ -458,27 +374,71 @@ private void AttachDatabase()
...
+                using (SqlCommand sqlCmd = sqlConn.CreateCommand())
+                {
+                    // Read from the file until it is empty.
+                    StringBuilder sqlStatements = new StringBuilder();
+                    String fileLine = reader.ReadLine();
+                    while (fileLine != null)
+                    {
+                        // If we find a "GO", execute everything we received up
+                        // to this point and ignore the "GO".
+                        if (fileLine == "GO")
```
biaol-odata: Does the sql db file always end with last line of "GO"? otherwise we need to always flush the buffered sqlStatements as the last step?

robward-ms: The files comes from SSMS's scripting tool and includes GO as the last line. I agree that is a possibility but I'd rather change the SQL if needed then make this code handle that case.

PR #874, opened 2017-07-05
Modify database strategy to eliminate Attach.

robward-ms:
### Description
Modify database strategy to eliminate Attach. This approach takes longer to run
but eliminates one of the reasons tests must be run ad Administrator.

- [   ] Test cases added


test/FunctionalTests/Framework/TestExecutionLayer/AstoriaDatabase.cs
```diff
@@ -458,27 +374,71 @@ private void AttachDatabase()
...
+                using (SqlCommand sqlCmd = sqlConn.CreateCommand())
+                {
+                    // Read from the file until it is empty.
+                    StringBuilder sqlStatements = new StringBuilder();
+                    String fileLine = reader.ReadLine();
+                    while (fileLine != null)
+                    {
+                        // If we find a "GO", execute everything we received up
+                        // to this point and ignore the "GO".
+                        if (fileLine == "GO")
```
biaol-odata: same comment as the change on master branch.

PR #882, opened 2017-07-06
Clean the skipped test cases, part 2

xuzhg:
### Issues

QP

### Description

Clean the skipped test cases in full.sln.
It's part of the clean skipped test cases, most of which are enabled in 6.x but disabled in 7.x.
So, comment out them and use a Github issue to track.

- [ x ] Test cases added


test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/Atom/WriterMetadataDocumentAtomTests.cs
```diff
@@ -23,8 +23,9 @@ public class WriterMetadataDocumentAtomTests : ODataWriterTestCase
         [InjectDependency]
         public MetadataWriterTestDescriptor.Settings Settings { get; set; }
 
+        // For comment out test cases, see github: https://github.com/OData/odata.net/issues/883
         [Ignore] // Remove Atom
```
biaol-odata: If I understand correctly, this case was propagated from 6.x branch, not applicable to 7.x due to Atom, is ignored, so we are commenting out the TestMethod. Assuming Atom will never come back to 7.x, why we are not deleting this file (or even the Atom folder) completely, since we have it on 6.x already. Am I missing anything?

xuzhg: Some test cases are mixed with atom and json, Some are not. To distinguish them needs time. So, after QP, we have tasks for it.

xuzhg:
Merged at: 8cdb4a6c5b5ebb8f30007d852e3976da7ed0a5c3

Issue #898, opened 2017-07-14
OData query throws InvalidOperationException: Collection was modified; enumeration operation may not execute


onlyann:
Exception of type `InvalidOperationException: Collection was modified; enumeration operation may not execute`  occurs when running queries from a generated OData client. 

### Assemblies affected

6.17.0

### Reproduce steps

I will try to come back with a consistent way to reproduce the behaviour. However, I feel that the additional detail section should give enough information to fix it.

It happens while calling an ASP.NET application that uses `DataServiceQuery<T>.GetAllPagesAsync`. However, it doesn't happen every time.
No `DataServiceContext` instance is shared between requests.

### Additional detail

Here's a stack trace of the exception
```
System.InvalidOperationException: Collection was modified; enumeration operation may not execute.
   at System.ThrowHelper.ThrowInvalidOperationException(ExceptionResource resource)
   at System.Collections.Generic.List`1.Enumerator.MoveNextRare()
   at System.Linq.Enumerable.FirstOrDefault[TSource](IEnumerable`1 source, Func`2 predicate)
   at Microsoft.OData.Client.ClientEdmModel.GetOrCreateEdmTypeInternal(IEdmStructuredType edmBaseType, Type type, PropertyInfo[] keyProperties, Boolean isEntity, Nullable`1 hasProperties)
   at Microsoft.OData.Client.ClientEdmModel.GetOrCreateEdmType(Type type)
   at Microsoft.OData.Client.ClientEdmModel.CreateEdmProperty(IEdmStructuredType declaringType, PropertyInfo propertyInfo)
   at Microsoft.OData.Client.ClientEdmModel.<>c__DisplayClass19.<GetOrCreateEdmTypeInternal>b__e(EdmEntityTypeWithDelayLoadedProperties entityType)
   at Microsoft.OData.Client.Providers.EdmEntityTypeWithDelayLoadedProperties.EnsurePropertyLoaded()
   at Microsoft.OData.Client.Providers.EdmEntityTypeWithDelayLoadedProperties.get_DeclaredProperties()
   at Microsoft.OData.Client.Metadata.ClientTypeAnnotation.<DiscoverEdmProperties>d__8.MoveNext()
   at System.Linq.Buffer`1..ctor(IEnumerable`1 source)
   at System.Linq.Enumerable.ToArray[TSource](IEnumerable`1 source)
   at Microsoft.OData.Client.Metadata.ClientTypeAnnotation.EdmProperties()
[...]

PR #902, opened 2017-07-16
Make ClientEdmModel.EdmStructuredSchemaElements thread-safe

onlyann:
### Issues

This pull request fixes issue #898 

### Description

Change structure of `ClientEdmModel.EdmStructuredSchemaElements` from `List<IEdmSchemaElement>` to `ConcurrentDictionary<string,IEdmSchemaElement>` where the key is the `Name` property.

This should make the property thread-safe while at the same time improve performance for large number of schema elements.

The reason this property should be made thread-safe is because the `ClientEdmModel` instances are shared through the [ClientEdmModelCache](https://github.com/OData/odata.net/blob/c0389b23b383cf9f6b9009b23b9c2bff98fa660f/src/Microsoft.OData.Client/DataServiceContext.cs#L3462) while the property gets updated by `TypeResolver` when sending requests / processing responses.

- [x] Build and test with one-click build and test script passed

### DIFF
--- a/src/Microsoft.OData.Client/ClientEdmModel.cs
+++ b/src/Microsoft.OData.Client/ClientEdmModel.cs
@@ -24,6 +24,7 @@ namespace Microsoft.OData.Client
     using Microsoft.OData.Edm.Library.Annotations;
     using Microsoft.OData.Edm.Library.Values;
     using c = Microsoft.OData.Client;
+    using System.Collections.Concurrent;
 
     #endregion Namespaces.
 
@@ -109,13 +110,12 @@ namespace Microsoft.OData.Client
         }
 
         /// <summary>
-        /// Gets the state of whether the edm structured schema elements have been set.
+        /// Gets the state of edm structured schema elements keyed by their Name.
         /// </summary>
-        internal List<IEdmSchemaElement> EdmStructuredSchemaElements
+        internal ConcurrentDictionary<string, IEdmSchemaElement> EdmStructuredSchemaElements
         {
-            get;
-            set;
-        }
+            get; 
+        } = new ConcurrentDictionary<string, IEdmSchemaElement>();
 
         /// <summary>
         /// Gets the max protocol version of the model.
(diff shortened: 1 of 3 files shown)


onlyann:
Would it also be preferable to change EdmStructuredSchemaElements to a `HashSet<T>` considering its average size and the way `TypeResolver` uses it:
```
foreach (var element in serviceModel.SchemaElements.Where(se => se is IEdmStructuredType))
                        {
                            if (!clientEdmModel.EdmStructuredSchemaElements.Contains(element))
                            {
                                clientEdmModel.EdmStructuredSchemaElements.Add(element);
                            }
                        }
```

robward-ms:
@onlyann - Can you also create a PR for 7.x as well?

onlyann:
@robward-ms . Any timeline of when this fix can be packaged in a new minor version? Thanks

src/Microsoft.OData.Client/ClientEdmModel.cs
```diff
@@ -109,13 +110,12 @@ public IEdmEntityContainer EntityContainer
...
+        /// Gets the state of edm structured schema elements keyed by their Name.
         /// </summary>
-        internal List<IEdmSchemaElement> EdmStructuredSchemaElements
+        internal ConcurrentDictionary<string, IEdmSchemaElement> EdmStructuredSchemaElements
         {
-            get;
-            set;
-        }
+            get; 
+        } = new ConcurrentDictionary<string, IEdmSchemaElement>();
```
robward-ms: the property_initializer syntax does not work in the VS2013 compiler, which we use for official builds. You'll need to change this.

robward-ms:
Compile error:
2017-07-19T16:59:22.0876927Z   c:\BuildAgent\_work\19\s\src\Microsoft.OData.Client\ClientEdmModel.cs(116,18): error CS1070: The type name 'ConcurrentDictionary' could not be found. This type has been forwarded to assembly 'System.Collections.Concurrent, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a'. Consider adding a reference to that assembly. [C:\BuildAgent\_work\19\s\src\Microsoft.OData.Client\Build.Portable\Microsoft.OData.Client.Portable.csproj]

Are you using the build.cmd script to catch these errors?

onlyann:
@robward-ms - Apologies. I didn't have VS2015 to run the `build.cmd`.
I am now able to replicate the issue.
Profile 111 means Windows Phone 8.0 and it doesn't support `ConcurrentDictionary`.
I can create a shim the same way SQLite [does it](https://github.com/praeclarum/sqlite-net/blob/master/src/SQLite.cs#L33). 
What compiler flag should I use here to ideally only target Windows Phone 8.0 ?

robward-ms:
I'm not seeing an appropriate symbol, can you define one like SQLLITE and add it to the Profile111 projects?

onlyann:
@robward-ms - I have used `PORTABLELIB` that is used is other places. My understanding is that the PCL111 portable version is one library that targets several platforms and there is no build that will be specific to Windows Phone 8.0 . 
The shim can be removed when `Odata.net` drops the support for Windows Phone 8.0.

I have verified that the build succeeds.

robward-ms:
If you look in the root Build.props file, you will see PORTABLELIB is defined for any platform that targets .NETPortable. 6,x ships with 2 portable libs and 2 dnx libs so your change is effectively only in the dnx libs.

I think that is the right thing to so but it begs the question as to whether this will fix a problem you are having. Are you consuming one of the dnx versions of the lib or one of the portable versions?

onlyann:
@robward-ms I have a dependency on the net40 version of the assembly, not on the portable version. 
That being said, this should also fix the problem on the portable version given the `EdmStructuredSchemaElements`dictionary isn't enumerated on top of a performance boost.

PR #905, opened 2017-07-17
change EdmStructuredSchemaElements to a ConcurrentDictionary

onlyann:
### Issues

This pull request cherry picked from #902 fixes issue #898.

### Description

Change structure of `ClientEdmModel.EdmStructuredSchemaElements` from `List<IEdmSchemaElement>` to `ConcurrentDictionary<string,IEdmSchemaElement>` where the key is the `Name` property.

This should make the property thread-safe while at the same time improve performance for large number of schema elements.

The reason this property should be made thread-safe is because the `ClientEdmModel` instances are shared through the [ClientEdmModelCache](https://github.com/OData/odata.net/blob/c0389b23b383cf9f6b9009b23b9c2bff98fa660f/src/Microsoft.OData.Client/DataServiceContext.cs#L3462) while the property gets updated by `TypeResolver` when sending requests / processing responses.

- [x] Build and test with one-click build and test script passed

### DIFF
--- a/src/Microsoft.OData.Client/ClientEdmModel.cs
+++ b/src/Microsoft.OData.Client/ClientEdmModel.cs
@@ -21,6 +21,7 @@ namespace Microsoft.OData.Client
     using Microsoft.OData.Edm;
     using Microsoft.OData.Edm.Vocabularies;
     using c = Microsoft.OData.Client;
+    using System.Collections.Concurrent;
 
     #endregion Namespaces.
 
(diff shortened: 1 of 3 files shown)


robward-ms:
@onlyann - Thanks!

src/Microsoft.OData.Client/ClientEdmModel.cs
```diff
@@ -106,13 +107,12 @@ public IEdmEntityContainer EntityContainer
...
+        /// Gets the state of edm structured schema elements keyed by their Name.
         /// </summary>
-        internal List<IEdmSchemaElement> EdmStructuredSchemaElements
+        internal ConcurrentDictionary<string, IEdmSchemaElement> EdmStructuredSchemaElements
         {
-            get;
-            set;
-        }
+            get; 
+        } = new ConcurrentDictionary<string, IEdmSchemaElement>();
```
robward-ms: the property_initializer syntax does not work in the VS2013 compiler, which we use for official builds. You'll need to change this.

onlyann: Done

Issue #899, opened 2017-07-14
Special char in hostname interferes ODL unit test run 


biaol-odata:
*My machine happened to have a host name with hypen char "-", and unit test failed. After removing special chars in the host name, the issue was resolved.*

### Assemblies affected

*Latest development branch odata.net/tree/master.*

### Reproduce steps

*Stack trace when running test\functionaltests\tests\dataservices\unittests\serverunittests\tests\providertests.cs:

Test method AstoriaUnitTests.Tests.UnitTestModule+ProviderTests.BaseProviderGetDerivedTypesTest threw exception: 

System.Data.SqlClient.SqlException: Incorrect syntax near 'Rds2_5d71adfbbeeb417387b337275aa8f10e'.
    at System.Data.SqlClient.SqlConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.SqlInternalConnection.OnError(SqlException exception, Boolean breakConnection, Action`1 wrapCloseInAction)
   at System.Data.SqlClient.TdsParser.ThrowExceptionAndWarning(TdsParserStateObject stateObj, Boolean callerHasConnectionLock, Boolean asyncClose)
   at System.Data.SqlClient.TdsParser.TryRun(RunBehavior runBehavior, SqlCommand cmdHandler, SqlDataReader dataStream, BulkCopySimpleResultSet bulkCopyHandler, TdsParserStateObject stateObj, ref Boolean dataReady)
   at System.Data.SqlClient.SqlCommand.RunExecuteNonQueryTds(String methodName, Boolean async, Int32 timeout, Boolean asyncWrite)
   at System.Data.SqlClient.SqlCommand.InternalExecuteNonQuery(TaskCompletionSource`1 completion, String methodName, Boolean sendToPipe, Int32 timeout, ref Boolean usedCache, Boolean asyncWrite, Boolean inRetry)
   at System.Data.SqlClient.SqlCommand.ExecuteNonQuery()
   at System.Data.Test.Astoria.TestExecutionLayer.AstoriaDatabase.CreatePopulatedDatabase() in AstoriaDatabase.cs: line 428
   at System.Data.Test.Astoria.TestExecutionLayer.AstoriaDatabase.CreateDatabase() in AstoriaDatabase.cs: line 222
*

### Expected result

*Test should pass.*

### Actual result

*Error.*

### Additional detail

*database name was constructed by:
[...]

PR #907, opened 2017-07-18
Re-organizing solution file layout for ODL master branch

AlanWong-MS:
### Description
Re-organizing solution file layout for ODL master branch
- Renaming solutions from Microsoft.(Test.)OData.* to OData.Product/Tests to reduce name size
- Creating new light solutions for each supported framework (Net35, Net45, and NetStandard) and corresponding unit tests.
- Reducing number of solutions by moving most tests to Microsoft.OData.Full.sln, which has been renamed to OData.Tests.E2E.sln. This solution builds on Net45
- There are two .NetStandard solutions. The product one contains only the product, and the test one contains both the product and unit tests. The product-only solution will eventually be removed once ODL migrates to Visual Studio 2017; this is a current workaround to pre-released version of .NET Core tooling in VS2015.

- [ x ] Test cases added


test/FunctionalTests/Framework/TestExecutionLayer/AstoriaDatabase.cs
```diff
@@ -173,7 +173,10 @@ public string DatabaseName
             {
                 if (_databaseName == null)
                 {
-                    _databaseName = String.Format("{0}_{1}_{2}", this.DatabasePrefixName, System.Net.Dns.GetHostName(), Guid.NewGuid().ToString("N"));
+                    _databaseName = String.Format("{0}_{1}_{2}",
+                        this.DatabasePrefixName,
+                        System.Net.Dns.GetHostName().Replace('-', '_'),
```
biaol-odata: A general theme should be converting other special characters as well as '-' here. Alternatively we can clarify the requirement for DNS hostname of not allowing special character. Here is the github issue submitted last week: https://github.com/OData/odata.net/issues/899

AlanWong-MS: Looks like the only special character allowed in the hostname is the '-', so I'll leave my change as-is.

biaol-odata: OK. Confirmed.

PR #836, opened 2017-06-15
remove the auto generated project.lock.json files

xuzhg:
### Description
project.lock.json files are auto-generated files. Don't push it into Github repository.

It also looks like a cache file and make the build.cmd failed.

PR #921, opened 2017-07-31
Migrating .NET Core tests and .NET Standard source projects to VS2017

AlanWong-MS:
### Description
- OData.Tests.NetStandard.sln is now OData.Tests.NetStandard.VS2017.sln
to indicate that it is to be opened with VS2017. Because we have a
massive number of projects to migrate from VS2015, we will do them
incrementally and use the VS2017 suffix to indicate which solutions are
VS2017. Once we are done with the migration, we'll remove the suffix.
- Replaced .xproj with new VS2017 .csproj for .NET Core tests. You will
need to have VS2017 installed if you wish to run the unit tests in .NET
Core. As a result of the migration, also cleaning up the project.jsons
and related files.
- Added VS2017 .csproj versions of the .NET Standard product source. You
may still launch OData.NetStandard.sln to modify and build .NET Standard
with VS2015. However, once we are in a state where we can flip the
switch to be 100% VS2017, OData.NetStandard.sln will also be in VS2017.
- Deleted src\Microsoft.OData.Core\PortabilityExtensionMethods.cs as it
is not used
- Fixing a few stylecop errors
- Updating build.ps1 to build and run the .NET Core tests if VS2017 and dotnet CLI are installed
- Updating the logic in the test results calculation in build.ps1. It was previously parsing each line in the test logs to count number of passed/skipped tests. The dotnet CLI doesn't provide this output, so the script would missing the passed/skipped tests for .NET Core tests. Modifying the algorithm to parse the "Total tests..." line for total passed/skipped tests, which is generalized across both VSTest and dotnet.

- [ x ] Test cases added

### Additional work necessary
N/A


AlanWong-MS:
This pull request includes the changes presented by #836.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -740,8 +740,8 @@ private bool ReadAtResourceStartImplementationSynchronously()
 
             if (this.CurrentResource == null)
             {
-                Debug.Assert(this.IsExpandedLinkContent || this.CurrentResourceType.IsODataComplexTypeKind() || this.CurrentResourceType.TypeKind == EdmTypeKind.Untyped,
-                    "null or untyped resource can only be reported in an expanded link or in collection of complex instance.");
+                // Debug.Assert(this.IsExpandedLinkContent || this.CurrentResourceType.IsODataComplexTypeKind(),
```
AlanWong-MS: Need re-review this. Is there value in this assert?

build.ps1
```diff
@@ -472,30 +494,13 @@ Function NugetRestoreSolution
...
-    Write-Host '**********Copying NetCore Output Binaries*********'
-
-    $NETCOREUNITTESTROOT = $ENLISTMENT_ROOT + "\test\FunctionalTests\"
-    $UNITTESTFOLDERS = "Microsoft.OData.Core.Tests",
-        "Microsoft.OData.Edm.Tests",
-        "Microsoft.Spatial.Tests"
-
-    ForEach($TEST in $UNITTESTFOLDERS)
-    {
-        Copy-Item ($NETCOREUNITTESTROOT + $TEST + "\bin\$Configuration\*") ($ENLISTMENT_ROOT + "\bin\AnyCPU\$Configuration\Test\.NETPortable") -Recurse -Force
```
AlanWong-MS: VS2017 csproj's OutputPath works properly so we no longer need this.

build.ps1
```diff
@@ -557,6 +596,15 @@ Function TestProcess
         exit
     }
 
+    if ($DOTNETTEST)
+    {
+        RunTest -title 'NetCoreTests' -testdir $NetCoreXUnitTestProjs -framework 'dotnet'
+    }
+    else
+    {
+        Write-Host 'The dotnet CLI must be installed to run any .NET Core tests.'
```
robward-ms: Color yellow?

AlanWong-MS: Done.

PR #930, opened 2017-08-08
Add the GC Allocations measure for performance 

xuzhg:
It's required to add the GC allocations measure for the performance test.
So, modify the performance tests to accept the [MeasureGCAllocations].

The performance test runner outputs the GC Allocations as below with "bytes" unit:

image

Meantime, to run the analysis will output the following:

image

- [  ] Test cases added
- [  ] Build and test with one-click build and test script passed

### Additional work necessary


xuzhg:
Merged at 77150134332deafc1aca64ab97f8278a21198da2

PR #932, opened 2017-08-10
Removing tests for Windows 8.0 apps

AlanWong-MS:
### Issues
N/A

### Description
Removing Microsoft.Test.OData.Tests.Client.Portable.WindowsStore.csproj and OData.Tests.WindowsStore.VS2013.sln due to complexity of maintaining VS2013. With this removal, the VS2013 will no longer be an installation requirement for comprehensive testing.

Per .NET Standard support matrix, .NET Standard 1.1 supports Windows 8.0 apps but we are making a conscious decision to remove the tests for that platform. This allows us to simplify the number of installations needed to develop for OData and achieve our overall goal of migration to the latest version of Visual Studio (2017 as of writing). Further, Windows 8.0 has been long superseded by Windows 8.1 (and 10), so there is low risk in removing such tests. Should any developer wish to run these tests, he/she may review the commit history to retrieve the test project.

- [ x ] Test cases added

### Additional work necessary
N/A


AlanWong-MS:
The buddy build check here was utilizing an outdated parameter so it will indicate failure. I've done a separate buddy build that passed. Will merge according to that result.

PR #939, opened 2017-08-11
Adding files that are part of the VSIX package.

AlanWong-MS:
### Description
Adding files that are part of the VSIX package.

- [ x ] Test cases added

### Additional work necessary
N/A


AlanWong-MS:
These files are part of a separate branch and do not impact the build result so foregoing buddy build result. Merging PR.

PR #569, opened 2016-05-17
Ignore Null Values in output

kosinsky:
Ignoring null values during serialization to improve performance and decrease output size in case of sparse entities.

Behavior controlled by ODataWriterSettings and disabled by default

PR #944, opened 2017-08-14
Ignore Null Values in output

kosinsky:
Ignoring null values during serialization to improve performance and decrease output size in case of sparse entities.
Behavior controlled by ODataWriterSettings and disabled by default

Recreating PR https://github.com/OData/odata.net/pull/569 to start from different branch


xuzhg:
As @mikepizzo said, we should consider to use the request header, Right?

kosinsky:
Yes. Prefer: omit-values=nulls (http://docs.oasis-open.org/odata/odata/v4.01/csprd01/part1-protocol/odata-v4.01-csprd01-part1-protocol.html#_Toc470092999 ). Still working on implementation

mikepizzo:
Closed in favor of PR #1106.

PR #958, opened 2017-09-19
6.x to 7.x Porting : Batching refactor - Multipart mime only

manivannanse:
### Issues

*Ported multi part mime refactoring from 6.X.*

### Description

*Refactored code to support multiple batch formats. This supports only support multi part mime*

- [  ] Test cases added
- [  ] Build and test with one-click build and test script passed

### Additional work necessary
4 Test failures needs to be fixed. Details are as follows
1 - Public Api test. This can be fixed by updating baseline file. But not sure if this is correct
2-  of them are ODataMessageTests.cs refereing to ODatabatch Writter and Reader. With new model it is abstract class. There are test changes for this. But I haven't ported this yet. Need to make sure this is not breaking change before porting test code
1 - renaming ChangeSetStarted to ChangesetStarted broke the test.

### DIFF
--- a/src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchReader.cs
+++ b/src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchReader.cs
@@ -1,0 +1,472 @@
+//---------------------------------------------------------------------
+// <copyright file="ODataMultipartMixedBatchReader.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.MultipartMixed
+{
+    #region Namespaces
+    using System;
+    using System.Diagnostics;
+    using System.Diagnostics.CodeAnalysis;
+    using System.Text;
+
+#if PORTABLELIB
+    using System.Threading.Tasks;
+#endif
+    #endregion Namespaces
+
+    /// <summary>
+    /// Class for reading OData batch messages of MIME multipart/mixed type.
+    /// Also verifies the proper sequence of read calls on the reader.
+    /// </summary>
+    internal sealed class ODataMultipartMixedBatchReader : ODataBatchReader
+    {
+        /// <summary>The batch stream used by the batch reader to divide a batch payload into parts.</summary>
+        private readonly ODataMultipartMixedBatchReaderStream batchStream;
+
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="inputContext">The input context to read the content from.</param>
+        /// <param name="batchBoundary">The boundary string for the batch structure itself.</param>
+        /// <param name="batchEncoding">The encoding to use to read from the batch stream.</param>
+        /// <param name="synchronous">true if the reader is created for synchronous operation; false for asynchronous.</param>
+        internal ODataMultipartMixedBatchReader(ODataInputContext inputContext, string batchBoundary, Encoding batchEncoding, bool synchronous)
+            : base(inputContext, synchronous)
+        {
+            Debug.Assert(inputContext != null, "inputContext != null");
+            Debug.Assert(this.RawInputContext != null, "this.RawInputContext != null");
+            Debug.Assert(!string.IsNullOrEmpty(batchBoundary), "!string.IsNullOrEmpty(batchBoundary)");
+
+            this.batchStream = new ODataMultipartMixedBatchReaderStream(this.RawInputContext, batchBoundary, batchEncoding);
+        }
+
+        /// <summary>
+        /// Gets the reader's input context as the real runtime type.
+        /// </summary>
+        internal ODataRawInputContext RawInputContext
+        {
+            get
+            {
+                return this.InputContext as ODataRawInputContext;
+            }
+        }
+
+        /// <summary>
+        /// Parses the request line of a batch operation request.
+        /// </summary>
+        /// <param name="requestLine">The request line as a string.</param>
+        /// <param name="httpMethod">The parsed HTTP method of the request.</param>
+        /// <param name="requestUri">The parsed <see cref="Uri"/> of the request.</param>
+        internal void ParseRequestLine(string requestLine, out string httpMethod, out Uri requestUri)
+        {
+            Debug.Assert(!this.RawInputContext.ReadingResponse, "Must only be called for requests.");
+
+            // Batch Request: POST /Customers HTTP/1.1
+            // Since the uri can contain spaces, the only way to read the request url, is to
+            // check for first space character and last space character and anything between
+            // them.
+            int firstSpaceIndex = requestLine.IndexOf(' ');
+
+            // Check whether there are enough characters after the first space for the 2nd and 3rd segments
+            // (and a whitespace in between)
+            if (firstSpaceIndex <= 0 || requestLine.Length - 3 <= firstSpaceIndex)
+            {
+                // only 1 segment or empty first segment or not enough left for 2nd and 3rd segments
+                throw new ODataException(Strings.ODataBatchReaderStream_InvalidRequestLine(requestLine));
+            }
+
+            int lastSpaceIndex = requestLine.LastIndexOf(' ');
+            if (lastSpaceIndex < 0 || lastSpaceIndex - firstSpaceIndex - 1 <= 0 || requestLine.Length - 1 <= lastSpaceIndex)
+            {
+                // only 2 segments or empty 2nd or 3rd segments
+                // only 1 segment or empty first segment or not enough left for 2nd and 3rd segments
+                throw new ODataException(Strings.ODataBatchReaderStream_InvalidRequestLine(requestLine));
+            }
+
+            httpMethod = requestLine.Substring(0, firstSpaceIndex);               // Request - Http method
+            string uriSegment = requestLine.Substring(firstSpaceIndex + 1, lastSpaceIndex - firstSpaceIndex - 1);      // Request - Request uri
+            string httpVersionSegment = requestLine.Substring(lastSpaceIndex + 1);             // Request - Http version
+
+            // Validate HttpVersion
+            if (string.CompareOrdinal(ODataConstants.HttpVersionInBatching, httpVersionSegment) != 0)
+            {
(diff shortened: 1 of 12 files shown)


src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchReaderStream.cs
```diff
@@ -0,0 +1,684 @@
...
+        private void DetermineChangesetBoundaryAndEncoding(string contentType)
+        {
+            Debug.Assert(!string.IsNullOrEmpty(contentType), "Should have validated that non-null, non-empty content type header exists.");
+
+            ODataMediaType mediaType;
+            ODataPayloadKind readerPayloadKind;
+            MediaTypeUtils.GetFormatFromContentType(
+                contentType,
+                new ODataPayloadKind[] { ODataPayloadKind.Batch },
+                new ODataMediaTypeResolver(),
```
manivannanse: >ODataMediaTypeResolver() [](start = 20, length = 24)

In 6.x it was ODataMediaTypeResolver.DefaultMediaTypeResolver. Need to revisit this

manivannanse:
Test failures still exists are as follows. These can be fixed by porting appropriate test files. Would like to understand why refactoring needs test fixes before porting the test code

13 Test failures needs to be fixed. Details are as follows
9 of them are -  batch Buffer was converted to upper case. Instead of fixing the test i changed the code to fix this. Need to figure out which one is correct

1 - Public Api test. This can be fixed by updating baseline file. But not sure if this is correct
2-  of them are ODataMessageTests.cs refereing to ODatabatch Writter and Reader. With new model it is abstract class. There are test changes for this. But I haven't ported this yet. Need to make sure this is not breaking change before porting test code
1 - renaming ChangeSetStarted to ChangesetStarted broke the test.

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -17,24 +17,23 @@ namespace Microsoft.OData
     #endregion Namespaces
 
     /// <summary>
-    /// Class for reading OData batch messages; also verifies the proper sequence of read calls on the reader.
+    /// Abstract Class for reading OData batch messages; also verifies the proper sequence of read calls on the reader.
     /// </summary>
-    public sealed class ODataBatchReader : IODataBatchOperationListener
+    public abstract class ODataBatchReader : IODataBatchOperationListener
```
manivannanse: >abstract [](start = 11, length = 8)

Is this breaking change.  I see tests that were initializing the reader and writer. Need to analyzed more of usage patterns

src/Microsoft.OData.Core/ODataBatchReaderStream.cs
```diff
@@ -10,55 +10,34 @@ namespace Microsoft.OData
...
-
-        /// <summary>The boundary string for the batch structure itself.</summary>
-        private readonly string batchBoundary;
 
         /// <summary>The buffer used by the batch reader stream to scan for boundary strings.</summary>
-        private readonly ODataBatchReaderStreamBuffer batchBuffer;
-
-        /// <summary>The media type resolver to use when interpreting the incoming content type.</summary>
-        private readonly ODataMediaTypeResolver mediaTypeResolver;
+        protected readonly ODataBatchReaderStreamBuffer batchBuffer;
```
manivannanse: >batchBuffer [](start = 56, length = 11)

This is renamed to BatchBuffer in original pull request in 6.x branch. This casued 6 test failure. Theoretically refactoring should not break any existing tests.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -899,14 +688,14 @@ private void ValidateTransition(BatchWriterState newState)
...
                         throw new ODataException(Strings.ODataBatchWriter_InvalidTransitionFromBatchStarted);
                     }
 
                     break;
-                case BatchWriterState.ChangeSetStarted:
-                    if (newState != BatchWriterState.OperationCreated && newState != BatchWriterState.ChangeSetCompleted)
+                case BatchWriterState.ChangesetStarted:
+                    if (newState != BatchWriterState.OperationCreated && newState != BatchWriterState.ChangesetCompleted)
                     {
                         throw new ODataException(Strings.ODataBatchWriter_InvalidTransitionFromChangeSetStarted);
```
AlanWong-MS: It looks like we decided to change the letter casing for "Change**S**etStarted" to "Change**s**etStarted". We should probably do a search across the project for other variables that have "ChangeSet" and update them for consistency.

ODataBatchWriter_InvalidTransitionFromChange**S**etStarted

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataBatchReaderStreamTests.cs
```diff
@@ -10,7 +10,7 @@
 using FluentAssertions;
 using Xunit;
 using ErrorStrings = Microsoft.OData.Strings;
-
+using Microsoft.OData.MultipartMixed;
```
AlanWong-MS: nit: empty line after this line

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -17,24 +17,23 @@ namespace Microsoft.OData
...
     /// <summary>
-    /// Class for reading OData batch messages; also verifies the proper sequence of read calls on the reader.
+    /// Abstract Class for reading OData batch messages; also verifies the proper sequence of read calls on the reader.
     /// </summary>
-    public sealed class ODataBatchReader : IODataBatchOperationListener
+    public abstract class ODataBatchReader : IODataBatchOperationListener
     {
         /// <summary>The input context to read the content from.</summary>
-        private readonly ODataRawInputContext inputContext;
+        private  ODataInputContext inputContext;
```
AlanWong-MS: nit: there's an extra space between "private" and "ODataInputContext"

robward-ms:
If we want to accept this portion of batching with failing tests, I suggest we target a feature branch instead of submitting to master.

src/Microsoft.OData.Core/ODataBatchReaderStream.cs
AlanWong-MS: I saw four failures in the build. Did you end up fixing some?

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchReader.cs
```diff
@@ -0,0 +1,472 @@
+//---------------------------------------------------------------------
+// <copyright file="ODataMultipartMixedBatchReader.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.MultipartMixed
```
mtroth-microsoft: >MultipartMixed [](start = 26, length = 14)

not enough classes here to merit its own namespace.

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -204,440 +236,81 @@ void IODataBatchOperationListener.BatchOperationContentStreamDisposed()
         }
 
         /// <summary>
-        /// Returns the next state of the batch reader after an end boundary has been found.
+        /// Increases the size of the current batch message; throws if the allowed limit is exceeded.
         /// </summary>
-        /// <returns>The next state of the batch reader.</returns>
-        private ODataBatchReaderState GetEndBoundaryState()
```
mikepizzo: >GetEndBoundaryState [](start = 38, length = 19)

I would expect these states and transitions to be common between the multipart/mixed and JSON formats, in which case they should be in the base class.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -136,66 +110,92 @@ private enum BatchWriterState
...
+                {
+                    Debug.Assert(!this.outputContext.WritingResponse, "Request message can only be set when writing request.");
+                    return this.currentOperationRequestMessage.OperationMessage;
+                }
+                else if (this.currentOperationResponseMessage != null)
+                {
+                    Debug.Assert(this.outputContext.WritingResponse, "Response message can only be set when writing response.");
+                    return this.currentOperationResponseMessage.OperationMessage;
+                }
+                else
```
mikepizzo: >               else [](start = 1, length = 19)

Don't really need the "else" as both the previous cases end in returns.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -136,66 +110,92 @@ private enum BatchWriterState
...
+            set { this.currentOperationContentId = value; }
+        }
+
+        /// <summary>
+        /// Gets or Sets the batch writer's state.
+        /// </summary>
+        protected BatchWriterState State
+        {
+            get { return this.state; }
+            set { this.state = value; }
```
mikepizzo: > set { this.state = value; } [](start = 11, length = 28)

I would generally expect the state to be set within the ODataBatchWriter base class as the derived classed called methods for each transition.

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchReader.cs
```diff
@@ -0,0 +1,472 @@
...
+
+                case ODataBatchReaderState.Exception:    // fall through
+                case ODataBatchReaderState.Completed:
+                    Debug.Assert(false, "Should have checked in VerifyCanRead that we are not in one of these states.");
+                    throw new ODataException(Strings.General_InternalError(InternalErrorCodes.ODataBatchReader_ReadImplementation));
+
+                default:
+                    Debug.Assert(false, "Unsupported reader state " + this.State + " detected.");
+                    throw new ODataException(Strings.General_InternalError(InternalErrorCodes.ODataBatchReader_ReadImplementation));
+            }
```
mikepizzo: Ideally state transitions would be done in the base class; the derived class would never set the state directly, but would call the next "Startxyz" method in the base class, which would transition state and then call the implementation method in the derived class.  Take a look at how this is done in ODataReader.

AlanWong-MS:
Closing this PR as it is superseded by #960.

PR #962, opened 2017-09-27
fix the CustomUriLiteralPrefixedTests to handle interference to stati…

biaol-odata:
…c prefix registration from other test runs

### Issues

*This pull request fixes stability issue for test CustomUriLiteralPrefixedTests.*

### Description

*The setup stage of CustomUriLiteralPrefixedTests could be subjected to interference from other tests (in the same test execution) via the static registrations of CustomUriLiteralPrefixes, and resulted in false-positive failure. The fix here is the handled the exception gracefully if the exception was caused by such interference.*
*see failure details in https://identitydivision.visualstudio.com/DefaultCollection/OData/_build/index?buildId=29037&_a=summary*

- [  ] Test cases added

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/CustomUriLiteralPrefixesTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/CustomUriLiteralPrefixesTests.cs
@@ -327,10 +327,30 @@ namespace Microsoft.OData.Tests.UriParser.Parsers
         [Fact]
         public void CustomUriLiteralPrefix_CannotParseTypeWithWrongLiteralPrefix()
         {
+            // Ensure the prefix under test is registered. Handle the exception gracefully if
+            // the prefix is already registered.
+            IEdmTypeReference booleanTypeReference = EdmCoreModel.Instance.GetBoolean(false);
+
             try
             {
-                IEdmTypeReference booleanTypeReference = EdmCoreModel.Instance.GetBoolean(false);
-                CustomUriLiteralPrefixes.AddCustomLiteralPrefix(CustomUriLiteralParserUnitTests.BOOLEAN_LITERAL_PREFIX, booleanTypeReference);
+                CustomUriLiteralPrefixes.AddCustomLiteralPrefix(
+                    CustomUriLiteralParserUnitTests.BOOLEAN_LITERAL_PREFIX, booleanTypeReference);
+            }
+            catch (ODataException e)
+            {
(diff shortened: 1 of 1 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/CustomUriLiteralPrefixesTests.cs
```diff
@@ -327,10 +327,30 @@ public void CustomUriLiteralPrefix_CannotParseWithCustomLiteralPrefix_IfBuiltInP
...
+                    // unexpected exception, re-throw.
+                    throw;
+                }
+
+                // Swallow the exception since it is due to trying to register a prefix that is already added.
+            }
+
+            try
+            {
+
```
AlanWong-MS: nit: extra empty line

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/CustomUriLiteralPrefixesTests.cs
```diff
@@ -327,10 +327,30 @@ public void CustomUriLiteralPrefix_CannotParseWithCustomLiteralPrefix_IfBuiltInP
         [Fact]
         public void CustomUriLiteralPrefix_CannotParseTypeWithWrongLiteralPrefix()
         {
+            // Ensure the prefix under test is registered. Handle the exception gracefully if
```
AlanWong-MS: Is there a way to unregister the instance during test clean-up?

biaol-odata: This are no ways to unregister the static instance. At this point we are not sure where the interference is coming from, but ensure proper setup for test here is the right thing to do. I have check other places, and they are all using unique prefixes, therefore should be no other issues.

---
In reply to: [141485235](https://github.com/OData/odata.net/pull/962#discussion_r141485235) [](ancestors = 141485235)

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/CustomUriLiteralPrefixesTests.cs
biaol-odata: removed.

---
In reply to: [141484785](https://github.com/OData/odata.net/pull/962#discussion_r141484785) [](ancestors = 141484785)

biaol-odata:
merged.

PR #964, opened 2017-09-29
fix the CustomUriLiteralPrefixedTests to handle interference to stati…

biaol-odata:
…c prefix registration from other test runs

### Issues

*This pull request fixes stability issue for test CustomUriLiteralPrefixedTests.*

### Description

*The setup stage of CustomUriLiteralPrefixedTests could be subjected to interference from other tests (in the same test execution) via the static registrations of CustomUriLiteralPrefixes, and resulted in false-positive failure. The fix here is the handled the exception gracefully if the exception was caused by such interference.*
*see failure details in https://identitydivision.visualstudio.com/DefaultCollection/OData/_build/index?buildId=29037&_a=summary*

- [  ] Test cases added

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/CustomUriLiteralPrefixesTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/CustomUriLiteralPrefixesTests.cs
@@ -353,8 +353,6 @@ namespace Microsoft.OData.Core.Tests.UriParser.Parsers
 
             try
             {
-
-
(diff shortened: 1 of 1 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/CustomUriLiteralPrefixesTests.cs
```diff
@@ -330,10 +330,30 @@ public void CustomUriLiteralPrefix_CannotParseWithCustomLiteralPrefix_IfBuiltInP
...
+                    // unexpected exception, re-throw.
+                    throw;
+                }
+
+                // Swallow the exception since it is due to trying to register a prefix that is already added.
+            }
+
+            try
+            {
+
```
AlanWong-MS: nit: extra empty line

PR #966, opened 2017-10-05
Cache EDMX in memory.

bhaarathms:
### Description

*Added support for storing the EDMX in memory statically so we only have to xml write when returning instead of traversing the EDM model serveral times.*

- [ ] *Test cases added*

### Additional work necessary

*Will add test cases once it is known that the proposed changes looks good for uptaking*


bhaarathms:
Sent prematurely

PR #968, opened 2017-10-06
Support String to Enum conversion

bhaarathms:
### Description

*Support String to Enum conversion*

- [x] *Test cases added*


src/Microsoft.OData.Core/UriParser/TypePromotionUtils.cs
```diff
@@ -261,6 +261,13 @@ internal static void GetTypeFacets(IEdmTypeReference type, out int? precision, o
                     return true;
                 }
 
+                // Comparing an enum with a string is valid
+                if (left.IsEnum() && right.IsString())
+                {
+                    right = left;
+                    return true;
+                }
```
mikepizzo: Any reason not to support the same comparison with the string on the left and enum on the right?

src/Microsoft.OData.Core/UriParser/Binders/BinaryOperatorBinder.cs
```diff
@@ -72,6 +72,12 @@ internal static void PromoteOperandTypes(BinaryOperatorKind binaryOperatorKind,
                 throw new ODataException(ODataErrorStrings.MetadataBinder_IncompatibleOperandsError(leftTypeName, rightTypeName, binaryOperatorKind));
             }
 
+            // If comparing an enum with a string, we must convert the string into the enum equivalent
+            if (left.TypeReference.IsEnum() && right.TypeReference.IsString())
+            {
+                right = new ConstantNode(new ODataEnumValue(((ConstantNode)right).Value.ToString(), leftType.Definition.ToString()), ((ConstantNode)right).Value.ToString(), rightType);
+            }
+
```
mikepizzo: After PromoteOperandTypes sets the type of the string operand to the appropriate enum type, would it be better to put the logic to create the appropriate constant node within  MetadataBindingUtils.ConvertToTypeIfNeeded?

mikepizzo:
Thanks for the contribution; eager to get this one integrated!  OData 4.01 adds support for implicit conversions of string to both enums and duration values; any interest in expanding this PR to also support string to duration conversions?

src/Microsoft.OData.Core/UriParser/TypePromotionUtils.cs
bhaarathms: That looks like a valid case. Added logic for conversion when string is on left

src/Microsoft.OData.Core/UriParser/Binders/BinaryOperatorBinder.cs
bhaarathms: Sure. Moved the conversion logic to ConvertToTypeIfNeeded. Looks cleaner this way.

bhaarathms:
@mikepizzo Thank you for taking a look at the PR. 
I think I would have to do some research on duration conversions. Right now I have few other things on my schedule. I will check back in a week and try adding them if they are not added by then.

AlanWong-MS:
@bhaarathms, heads up: we recently made committed some changes to the master branch that you'll have to pull. The build check will fail without the new changes.

bhaarathms:
Looks like I messed something up during rebase. I will wait to see the build status. I will create a new pull request and close this one

bhaarathms:
@mikepizzo created a new PR as I messed this one up. https://github.com/OData/odata.net/pull/980 is this new PR. Please take a look.

Issue #965, opened 2017-10-01
Parsing encoded character with special meaning in  ODataJsonLightContextUriParser


OfirArviv:
When receiving an OData response with 'odata.context' uri that contain encoded special characters with special meaning in uri parsing, for example %23 which is the encoding for '#', the OData client throws the following exception The context URL '[context url ]' is invalid.' This occur due to parsing the URI with 'UriFormat.Unescaped' param.

### Assemblies affected

OData .Net lib 6.16.0 (and also 7.X by looking at the source code).

### Reproduce steps

An example for a request that fails:
1. Making the following OData GET request using a generated OData client :
	a. GET /api/v1.0/me/notes/classnotebooks/[guid]/students?$expand=members HTTP/1.1
2. Receiving a valid response with one of the inner complex fields having a metadata property as following:
	a. "members@odata.context":"https://www.onenote.com/api/v1.0/$metadata#me/notes/classnotebooks('[guid]')/students('i%3A0%23.f%7Cmembership%7Cexample%40example.org')/members"

### Expected result

The OData response is parsed normally.

### Actual result
The following exception is thrown 'The context URL 'https://www.onenote.com/api/v1.0/$metadata#me/notes/classnotebooks('[guid]')/students('i%3A0%23.f%7Cmembership%7Cexample%40example.org')/members' is invalid.'

### Additional detail

This is the main flow that cause this:

	1. Processing the response is taking place in the OData client.
	2. A new ODataJsonLightContextUriParser class is created with the context URI as it's input.
	3. The class method TokenizeContextUri() is called.
	4. The fragment of the context URI is parsed in the following line of codes:
		1. Uri contextUriFromPayload = this.parseResult.ContextUri; 
		2.  this.parseResult.Fragment = contextUriFromPayload.GetComponents(UriComponents.Fragment, UriFormat.Unescaped);
		* This is the root - using UriFormat.Unescaped instead of UriFormat.SafeUnescaped or UriFormat.UriEscaped*
	5. Which results in parser.parseResult.Fragment = "me/notes/classnotebooks('[guid]')/students('i:0#.f|membership|example@example.org')/members"
	6. A chain of methods continues until we get to the method ParseContextUriFragment() of the ODataJsonLightContextUriParser class is executed with the fragment variable as one of its parameters.
	7. Which results in the following lines of code being exectued:
		1. var rawSegments = pathParser.ParsePathIntoSegments(pathUri, this.ServiceRoot); in  ODataUriParser class.
[...]

PR #969, opened 2017-10-09
Escaped special meaning characters remain escaped when parsing the co…

OfirArviv:
…ntext uri +UT

### Issues

*This pull request fixes issue #965 *

### Description

Parsing the tokens of a context url in ODataJsonLightContextUriParser.TokenizeContextUri now use UriFormat.SafeUnescaped instead of UriFormat.Unescaped

- [X] *Test cases added*

### Additional work necessary
None


biaol-odata:
@OfirArviv The change looks good to me. We can merge it into master after @mikepizzo sign off on it. Thanks.

PR #980, opened 2017-10-26
Support enum to string comparision

bhaarathms:
### Description

Support string to Enum conversion in filters

- [ X] *Test cases added*

### Additional work necessary

No

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
@@ -55,14 +55,18 @@ namespace Microsoft.OData.UriParser
                     return new ConvertNode(source, targetTypeReference);
                 }
 
+                ConstantNode constantNode = source as ConstantNode;
+                if (source.TypeReference.IsString() && targetTypeReference.IsEnum())
+                {
+                    return new ConstantNode(new ODataEnumValue(constantNode.Value.ToString(), targetTypeReference.Definition.ToString()), constantNode.Value.ToString(), targetTypeReference);
+                }
+                
                 if (!TypePromotionUtils.CanConvertTo(source, source.TypeReference, targetTypeReference))
                 {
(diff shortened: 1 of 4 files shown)


bhaarathms:
This PR is same as https://github.com/OData/odata.net/pull/968 
I messed something up while rebasing with master on the branch corresponding old PR. So I created this one afresh. 
I will close the old PR.

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/ODataUriResolverTests.cs
```diff
@@ -139,33 +139,19 @@ public override void PromoteBinaryOperandTypes(BinaryOperatorKind binaryOperator
         [Fact]
         public void TestStringAsEnumInFunctionParameter()
         {
-            this.TestStringAsEnum(
```
robward-ms: Why is this test being removed?

bhaarathms: With the change in this PR String to Enum Conversion is supported. The two tests here check for error messages on converting a string to Enum. So they are no longer valid. So I am removing them and adding a test to verify the conversion.

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
```diff
@@ -55,14 +55,18 @@ internal static SingleValueNode ConvertToTypeIfNeeded(SingleValueNode source, IE
                     return new ConvertNode(source, targetTypeReference);
                 }
 
+                ConstantNode constantNode = source as ConstantNode;
+                if (source.TypeReference.IsString() && targetTypeReference.IsEnum())
+                {
+                    return new ConstantNode(new ODataEnumValue(constantNode.Value.ToString(), targetTypeReference.Definition.ToString()), constantNode.Value.ToString(), targetTypeReference);
```
mtroth-microsoft: >constantNode [](start = 63, length = 12)

how do you know this is not null?

bhaarathms: added null checks

Issue #731, opened 2016-12-05
Microsoft.OData.Client 7.0


meancrazy:
Please update Microsoft.OData.Client nuget package to version 7.0.

patrickmichalina:
Thanks - not sure why they haven't shipped it yet...

MAOliver:
I have an application using the 6.15 version of the client to contact a service which is using 7.0 with a ComplexType w/ a NavigationProperty.

The client currently fails to instantiate. I can make it work by creating an intermediate complextype and mapping my entity to the complextype and changing my complextype to only reference other complextypes. Is there another, more elegant way to work around this situation while I wait for the new client?

earloc:
any near plans for shipping this?

AlanWong-MS:
Hi everyone, Microsoft.OData.Client 7.x is on our radar (it has not been released with the recent ODL 7.2.0). We're currently sorting out any possible breaking changes with the library and will proceed to publish it when we're ready.

Thanks for your patience!

Slimicus:
@MAOliver do you have any more documentation on your workaround for this issue? I'm currently facing the same error, the edmx won't parse because of a NavigationProperty in the ComplexType. Thanks!

MAOliver:
@Slimicus My situation arose from trying to convert a pre-existing service to OData. I have an entry point that was originally a WebAPI entrypoint whose data contract was created as "XSD-first" and there are several container objects that don't actually map to an entity, but they contain entities themselves.

To solve the immediate issue, I just removed navigation properties from the complex types and replaced them with a complex type that I populate from the entity that the navigation property was pointing to. It means that anyone calling the endpoint won't be able to traverse to the entity endpoint that populated that complextype, but I didn't want to lose the ability to generate clients.

It's hard for me to be clearer about what I did without creating a code sample to show you. Is this clear enough?

wmalgadey:
any update on this? nearly 2 month later?

JamieG:
I'm researching tooling for a series of upcoming projects and I really wanted to use OData, I'm extremely disappointed with the state of the tooling and the lack of VS2017 support. I've attempted to workaround the issues by forking the repo but the solution is a cross referencing headache and the client template is a nightmare, is a 5000 line ttinclude really the best was to go? The whole thing needs a reboot and a rewrite for .NET Core. Is this being neglected because the core team are going in a different direction, should we be looking at an alternative solution?

MAOliver:
[...]

PR #981, opened 2017-10-27
Updating Nuget package for OData Client to remove NetFx45 binaries

AlanWong-MS:
### Issues
Issue #731

### Description
- Updating Nuget package for OData Client to remove NetFx45 binaries.
- Fixing bug in ServiceWrapperApp where exception is ignored.

- [x] *Test cases added (N/A)*

### Additional work necessary
N/A

### DIFF
--- a/test/EndToEndTests/Services/ServiceWrapperApp/ServiceWrapperApp.cs
+++ b/test/EndToEndTests/Services/ServiceWrapperApp/ServiceWrapperApp.cs
@@ -67,6 +67,8 @@ namespace ServiceWrapperApp
             {
                 Console.WriteLine(e.Message);
                 PrintHelp();
+
+                return 1;
             }
 
             if (!isAutomation)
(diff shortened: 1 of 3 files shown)


test/EndToEndTests/Services/ServiceWrapperApp/ServiceWrapperApp.cs
```diff
@@ -67,6 +67,8 @@ static int Main(string[] args)
             {
                 Console.WriteLine(e.Message);
                 PrintHelp();
+
+                return 1;
```
robward-ms: How about make this 2 so you have a unique return code here, might be useful later on the calling side?

robward-ms: Your change certainly covers the case I asked for :)

AlanWong-MS: Wanted to polish the code a little bit more while I was in there. :)

AlanWong-MS:
Looks like the buddy build failed for a separate reason. Will squash and merge this PR.

PR #960, opened 2017-09-21
porting to master branch: multipart/mixed refactoring

biaol-odata:
### Issues

*This pull request is for porting the JsonBatch feature work from the maintenance-6.x branch. As the first step, it includes the refactoring of Multipart/Mixed reader & write into base + derived classes. The base class ODataBatchReader and ODataBatchWriter will be the the base class for JsonBatch format specific reader & writer as well.*

### Description

*Refactoring of Multipart/Mixed reader & writer with tests.*

### Checklist (Uncheck if it is not completed)

- [ x ] Test cases added
- [ x ] Build and test with one-click build and test script passed

### Additional work necessary

*If documentation update is needed, please add "Docs Needed" label to the issue and provide details about the required document change in the issue.*

PR #992, opened 2017-11-10
porting to master branch: JsonBatch step2

biaol-odata:
### Issues

*This pull request is for porting the JsonBatch feature work from the maintenance-6.x branch with further re-factorization and modifications for master branch resolving breaking changes.
This PR is based on, and contains all changes from, [PR for step#1](https://github.com/OData/odata.net/pull/960), which is now therefore superseded.*

### Description

*The first 13 commits are inherited from [PR for step#1](https://github.com/OData/odata.net/pull/960).*

*New change here includes introducing the ODL-Core support for JsonBatch, e.g. batch reader / writer for JsonBatch format specified thru Content-Type header, along with auxiliary functionalities. The corresponding commits are those starting from 7a8cd77.*

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/ODataBatchUtils.cs
```diff
@@ -142,5 +146,121 @@ internal static void EnsureArraySize(ref byte[] buffer, int numberOfBytesInBuffe
...
+                // Validate that dependsOn contains the Id reference.
+                // Note that the dependsOnRequestIds here is already flattened and consists of individual request
+                // ids, therefore it should contain the request id referenced by the Uri.
+                if (dependsOnRequestIds == null || !dependsOnRequestIds.Contains(referenceId))
+                {
+                    throw new ODataException(Strings.ODataBatchReader_ReferenceIdNotIncludedInDependsOn(
+                        referenceId, UriUtils.UriToString(uri), dependsOnRequestIds));
+                }
+            }
+        }
```
mikepizzo: ODataBatchUtils should be format-independent.
Validating the reference uri is a format-independent task, it's just that the set of valid ids are different between json and multipart/mixed.
In Json, the valid ids have to be called out as dependencies. In multipart/mixed, they are previous content-ids (in v4, they must be in the same batch; in v4.01 they can be any other content ids in the batch).

ODataBatchUtils should take the Uri, the set of valid references, and BaseUri.  The JsonWriter should call with the specified dependency ids; the multipart/mixed should call with the contentids of previous statements in the changeset (v4) or batch (v4.01). ODataBatchUtils should have no references to JSON or MultipartMixed derived classes.

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
```diff
@@ -66,9 +84,13 @@ public sealed class ODataBatchOperationRequestMessage : IODataRequestMessage, IO
             this.Method = method;
             this.Url = requestUrl;
             this.ContentId = contentId;
+            this.groupId = groupId;
 
             this.message = new ODataBatchOperationMessage(contentStreamCreatorFunc, headers, operationListener, payloadUriConverter, writing);
             this.Container = container;
+            this.dependsOnIds = dependsOnIds != null
+                             ? new List<string>(dependsOnIds)
+                             : null;
```
mikepizzo: If the caller passes in an empty list, we will set dependsOnIds to an empty list.  If the caller passes in null we will create a null dependsOnIds.  That means that the public caller will have to check for both conditions. Why not just always create a List<string> containing any values passed in to the constructor?

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
```diff
@@ -99,6 +121,31 @@ public string Method
...
+        {
+            get
+            {
+                return this.dependsOnIds;
+            }
+
+            set
+            {
+                this.dependsOnIds = value;
+            }
```
mikepizzo: Rather than expose a setter for an IEnumerable, if we expose as a read-only property of type List<string> the user can add dependency Ids rather than replacing the entire enumerable..

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -204,48 +247,167 @@ void IODataBatchOperationListener.BatchOperationContentStreamDisposed()
...
+        /// <param name="method">The HTTP method used for this request message.</param>
+        /// <param name="requestUri">The request Url for this request message.</param>
+        /// <param name="headers">The headers for this request message.</param>
+        /// <param name="contentId">The contentId of this request message.</param>
+        /// <param name="dependsOnRequestIds">
+        /// The prerequisite request Ids of this request message. For batch in Json format,
+        /// some of these request Ids are resolved from prerequisite atomic groups that are
+        /// specified in the dependsOn attribute of the request.</param>
+        /// <returns>The <see cref="ODataBatchOperationRequestMessage"/> instance.</returns>
+        protected ODataBatchOperationRequestMessage BuildOperationRequestMessage(
```
mikepizzo: >BuildOperationRequestMessage [](start = 52, length = 28)

Why doesn't buildoperationrequestmessage in batchreader set the groupid on the request message?  Given that it can be set when creating a request message, why wouldn't we set it when reading the request message?

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -204,48 +247,167 @@ void IODataBatchOperationListener.BatchOperationContentStreamDisposed()
...
 
-                case ODataBatchReaderState.Operation:
-                    // If we find an end boundary in state 'Operation' we have finished
-                    // processing an operation and found the end boundary of either the
-                    // current changeset or the batch.
-                    return this.batchStream.ChangeSetBoundary == null
-                        ? ODataBatchReaderState.Completed
-                        : ODataBatchReaderState.ChangesetEnd;
+                ODataBatchUtils.ValidateReferenceUri(requestUri, dependsOnRequestIds,
+                    this.inputContext);
```
mikepizzo: >this.inputContext [](start = 20, length = 17)

See comment on ValidateReferenceUri in ODataBatchUtils.  This should just take the this.inputContext.BaseUri, and the logic in ODataBatchUtils should not depend on the format.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchReaderStream.cs
```diff
@@ -0,0 +1,120 @@
...
+        /// <param name="userBufferOffset">The offset in the buffer where to start reading bytes into.</param>
+        /// <param name="count">The number of bytes to read.</param>
+        /// <returns>The number of bytes actually read.</returns>
+        internal override int ReadWithLength(byte[] userBuffer, int userBufferOffset, int count)
+        {
+            Debug.Assert(userBuffer != null, "userBuffer != null");
+            Debug.Assert(userBufferOffset >= 0, "userBufferOffset >= 0");
+            Debug.Assert(count >= 0, "count >= 0");
+            Debug.Assert(this.batchEncoding != null, "Batch encoding should have been established.");
+
```
mikepizzo: this.batchEncoding is not used in JSON, so you can remove this assert.

src/Microsoft.OData.Core/ODataBatchReaderStream.cs
```diff
@@ -8,332 +8,68 @@ namespace Microsoft.OData
...
-        /// <summary>The media type resolver to use when interpreting the incoming content type.</summary>
-        private readonly ODataMediaTypeResolver mediaTypeResolver;
+        protected readonly ODataBatchReaderStreamBuffer BatchBuffer;
 
         /// <summary>The encoding to use to read from the batch stream.</summary>
-        private Encoding batchEncoding;
-
-        /// <summary>The boundary string for a changeset (or null if not in a changeset part).</summary>
-        private string changesetBoundary;
+        protected Encoding batchEncoding;
```
mikepizzo: batchEncoding is only used in multipartmixed, so you can move it to the derived class.

src/Microsoft.OData.Core/ODataBatchReaderStream.cs
```diff
@@ -8,332 +8,68 @@ namespace Microsoft.OData
...
 
         /// <summary>The encoding for a given changeset.</summary>
-        private Encoding changesetEncoding;
+        protected Encoding changesetEncoding;
```
mikepizzo: changesetEncoding is only used in multipartmixed, so you can move it there.

src/Microsoft.OData.Core/ODataBatchReaderStream.cs
```diff
@@ -8,332 +8,68 @@ namespace Microsoft.OData
...
         /// </summary>
         /// <remarks>This is the changeset encoding when reading a changeset or the batch encoding otherwise.</remarks>
-        private Encoding CurrentEncoding
+        protected Encoding CurrentEncoding
         {
             get
             {
                 return this.changesetEncoding ?? this.batchEncoding;
             }
         }
```
mikepizzo: Can move this to multipartmixed.

src/Microsoft.OData.Core/ODataBatchReaderStream.cs
```diff
@@ -343,308 +79,45 @@ internal int ReadWithDelimiter(byte[] userBuffer, int userBufferOffset, int coun
...
-            return this.CurrentEncoding.GetString(bytesForString, 0, lineBufferSize);
-        }
+        internal abstract int ReadWithLength(byte[] userBuffer, int userBufferOffset, int count);
 
         /// <summary>
         /// Ensure that a batch encoding exists; if not, detect it from the first couple of bytes of the stream.
         /// </summary>
-        private void EnsureBatchEncoding()
+        /// <param name="stream">The stream to read.</param>
+        protected void EnsureBatchEncoding(Stream stream)
```
mikepizzo: Can be moved to multipart/mixed

src/Microsoft.OData.Core/ODataBatchReaderStream.cs
```diff
@@ -343,308 +79,45 @@ internal int ReadWithDelimiter(byte[] userBuffer, int userBufferOffset, int coun
...
         /// <summary>Detect the encoding based data from the stream.</summary>
+        /// <param name="stream">The stream to read.</param>
         /// <returns>The encoding discovered from the bytes in the buffer or the fallback encoding.</returns>
         /// <remarks>
         /// We don't have to skip a potential preamble of the encoding since the batch reader
         /// will skip over everything (incl. the potential preamble) until it finds the first
         /// boundary.
         /// </remarks>
-        private Encoding DetectEncoding()
+        protected Encoding DetectEncoding(Stream stream)
```
mikepizzo: >DetectEncoding [](start = 27, length = 14)

Can be moved to multipart/mixed

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -236,20 +199,55 @@ public Task WriteEndBatchAsync()
...
 #endif
 
-        /// <summary>Starts a new changeset; can only be called after WriteStartBatch and if no other active operation or changeset exists.</summary>
+        /// <summary>
+        /// Starts a new changeset without specifying group id.
+        /// This can only be called after WriteStartBatch and if no other active operation or changeset exists.
+        /// </summary>
         public void WriteStartChangeset()
+        {
+            this.WriteStartChangeset(null);
```
mikepizzo: I would pass Guid.NetGuid().ToString to this.WriteStartChangeset.  Then, in the WriteStartChangesetImplementation for JsonLight you don't need to account for a null groupId, and in multipartmixed you can generate the changeset boundary using that groupid.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -236,20 +199,55 @@ public Task WriteEndBatchAsync()
...
+        /// <summary>
+        /// Starts a new changeset with the specified group id, which is null for multipart batch.
+        /// This can only be called after WriteStartBatch and if no other active operation or changeset exists.</summary>
+        /// <param name="groupId">
+        /// For Json batch, this is the atomic group id. Default value of null can be used for writing request and an GUID will be generated.
+        /// Non-null value from corresponding request atomic group should be used for writing response so that atomic groups
+        /// from request and response can be correlated.
+        /// For multipart batch, null value should be used and is ignored.
+        /// </param>
+        public void WriteStartChangeset(string groupId)
```
mikepizzo: >string groupId [](start = 40, length = 14)

disallow passing null, so that downstream logic can rely  on a non-null groupid (either passed by user or a guid generated in the parameterless WriteStartChangeset).

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchWriter.cs
```diff
@@ -0,0 +1,651 @@
...
+
+            // write pending message data (headers, response line) for a previously unclosed message/request
+            this.WritePendingMessageData(true);
+
+            // important to do this first since it will set up the change set boundary.
+            this.SetState(BatchWriterState.ChangesetStarted);
+
+            // When the changeset start is created for request, the <paramref name="groupId"/> could be null
+            // and we will create an GUID here.
+            this.atomicityGroupId = groupId ?? Guid.NewGuid().ToString();
```
mikepizzo: create the groupid in the parameterless WriteStartChangeset() of the base class; assert in this method that it is non-null.

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
```diff
@@ -0,0 +1,413 @@
...
+            return this.RawOutputContext.FlushAsync();
+        }
+#endif
+
+        /// <summary>
+        /// Starts a new changeset - implementation of the actual functionality.
+        /// </summary>
+        /// <param name="groupId">Ignored value for multipart batch start changeset implementation
+        /// since in this format the changeset boundary is always created randomly for both request and response.</param>
+        protected override void WriteStartChangesetImplementation(/*ignored*/string groupId)
```
mikepizzo: >/*ignored*/string groupId [](start = 66, length = 25)

don't ignore; pass in a non-null groupId from the base class and pass it on to CreateChangeSetBoundary, rather than generating a new guid there.

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriterUtils.cs
```diff
@@ -1,43 +1,45 @@
...
+            return string.Format(CultureInfo.InvariantCulture, template, Guid.NewGuid());
         }
 
         /// <summary>
         /// Creates a new changeset boundary string based on a randomly created GUID.
         /// </summary>
-        /// <param name="isResponse">A flag indicating whether the boundary should be created for a request or a resonse.</param>
+        /// <param name="isResponse">A flag indicating whether the boundary should be created for a request or a response.</param>
         /// <returns>The newly created changeset boundary as string.</returns>
         internal static string CreateChangeSetBoundary(bool isResponse)
```
mikepizzo: >CreateChangeSetBoundary [](start = 31, length = 23)

Pass in the groupid passed to startchangesetimplementation rather than create it here.

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriterUtils.cs
```diff
@@ -1,43 +1,45 @@
...
-    internal static class ODataBatchWriterUtils
+    internal static class ODataMultipartMixedBatchWriterUtils
     {
         /// <summary>
         /// Creates a new batch boundary string based on a randomly created GUID.
         /// </summary>
-        /// <param name="isResponse">A flag indicating whether the boundary should be created for a request or a resonse.</param>
+        /// <param name="isResponse">A flag indicating whether the boundary should be created for a request or a response.</param>
         /// <returns>The newly created batch boundary as string.</returns>
         internal static string CreateBatchBoundary(bool isResponse)
```
mikepizzo: >CreateBatchBoundary [](start = 31, length = 19)

For consistency with CreateChangeSetBoundary, consider passing in a boundaryid that is generated when calling this function.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -236,20 +199,55 @@ public Task WriteEndBatchAsync()
...
         }
 
 #if PORTABLELIB
-        /// <summary>Asynchronously starts a new change set; can only be called after WriteStartBatch and if no other active operation or change set exists.</summary>
+        /// <summary>Asynchronously starts a new change set without specifying group id;
+        /// This can only be called after WriteStartBatch and if no other active operation or change set exists.</summary>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
         public Task WriteStartChangesetAsync()
+        {
+            return WriteStartChangesetAsync(null);
```
mikepizzo: As above, call WriteStartChangesetAsync(Guid.NewGuid().ToString()) so that downstream logic always has a non-null groupId.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -632,104 +686,55 @@ private ODataBatchOperationRequestMessage CreateOperationRequestMessageImplement
...
+            this.isInChangset = true;
         }
 
         /// <summary>
-        /// Disposes the batch writer and set the 'OperationStreamRequested' batch writer state;
-        /// called after the flush operation(s) have completed.
+        /// Perform updates after changeset is ended.
         /// </summary>
-        private void DisposeBatchWriterAndSetContentStreamRequestedState()
+        private void UpdateAfterWriteEndChangeset()
```
mikepizzo: >UpdateAfterWriteEndChangeset [](start = 21, length = 28)

How about "FinishWriteEndChangeset"?

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -632,104 +686,55 @@ private ODataBatchOperationRequestMessage CreateOperationRequestMessageImplement
...
             return this.CurrentOperationRequestMessage;
         }
 
         /// <summary>
-        /// Verifies that calling CreateOperationResponseMessage is valid.
+        /// Perform updates after changeset is started.
         /// </summary>
-        /// <param name="synchronousCall">true if the call is to be synchronous; false otherwise.</param>
-        private void VerifyCanCreateOperationResponseMessage(bool synchronousCall)
+        private void UpdateAfterWriteStartChangeset()
```
mikepizzo: >UpdateAfterWriteStartChangeset [](start = 21, length = 30)

how about "FinishWriteStartChangeset"?

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
```diff
@@ -0,0 +1,413 @@
...
+        /// <param name="uri">The Uri to be used for this request operation.</param>
+        /// <param name="contentId">The Content-ID value to write in ChangeSet head.</param>
+        /// <param name="payloadUriOption">
+        /// The format of operation Request-URI, which could be AbsoluteUri, AbsoluteResourcePathAndHost, or RelativeResourcePath.</param>
+        /// <param name="dependsOnIds">The prerequisite request ids of this request.
+        /// For multipart format, it is less significant than it is for Json batch format and has a default value of null.</param>
+        /// <returns>The message that can be used to write the request operation.</returns>
+        protected override ODataBatchOperationRequestMessage CreateOperationRequestMessageImplementation(
+            string method, Uri uri, string contentId, BatchPayloadUriOption payloadUriOption,
+            /*ignored*/IEnumerable<string> dependsOnIds)
```
mikepizzo: >/*ignored*/IEnumerable<string> dependsOnIds [](start = 12, length = 43)

Rather than ignore, should validate that the dependsOnIds are all contentIds that have previously been read.

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
biaol-odata: updated to use List<string> and exposed as read-only getter.

---
In reply to: [155349744](https://github.com/OData/odata.net/pull/992#discussion_r155349744) [](ancestors = 155349744)

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -323,6 +366,8 @@ public Task<ODataBatchOperationRequestMessage> CreateOperationRequestMessageAsyn
         public ODataBatchOperationResponseMessage CreateOperationResponseMessage(string contentId)
         {
             this.VerifyCanCreateOperationResponseMessage(true);
+            Debug.Assert(this.currentOperationContentId == null, "The Content-ID header is only supported in request messages.");
+
             return this.CreateOperationResponseMessageImplementation(contentId);
```
mikepizzo: Really?  In json implementation we pass it in to BuildOperationResponseMessage (as we should), and in multipartmixed we *should* pass it in to BuildOperationResponseMessage and *do* use it when writing the response preamble.

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
```diff
@@ -0,0 +1,413 @@
...
+        /// <param name="contentId">The Content-ID value to write in ChangeSet head.</param>
+        /// <returns>The message that can be used to write the response operation.</returns>
+        protected override ODataBatchOperationResponseMessage CreateOperationResponseMessageImplementation(string contentId)
+        {
+            this.WritePendingMessageData(true);
+
+            // In responses we don't need to use our batch URL resolver, since there are no cross referencing URLs
+            // so use the URL resolver from the batch message instead.
+            this.CurrentOperationResponseMessage = BuildOperationResponseMessage(
+                this.RawOutputContext.OutputStream, /*contentId*/ null, /*groupId*/ null);
```
mikepizzo: >/*contentId*/ null [](start = 52, length = 18)

We should pass in the contentId that was passed in to the method.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -355,32 +401,24 @@ public void Flush()
             }
         }
 
+
+
 #if PORTABLELIB
```
mikepizzo: Extra blank lines

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: Moved the setting groupid here.

---
In reply to: [155353014](https://github.com/OData/odata.net/pull/992#discussion_r155353014) [](ancestors = 155353014)

src/Microsoft.OData.Core/ODataBatchReaderStream.cs
biaol-odata: moved to multipart derived class.

---
In reply to: [155379677](https://github.com/OData/odata.net/pull/992#discussion_r155379677) [](ancestors = 155379677)

src/Microsoft.OData.Core/ODataBatchReaderStream.cs
biaol-odata: moved to multipart derived class.

---
In reply to: [155379765](https://github.com/OData/odata.net/pull/992#discussion_r155379765) [](ancestors = 155379765)

src/Microsoft.OData.Core/ODataBatchReaderStream.cs
biaol-odata: moved.

---
In reply to: [155380026](https://github.com/OData/odata.net/pull/992#discussion_r155380026) [](ancestors = 155380026)

src/Microsoft.OData.Core/ODataBatchReaderStream.cs
biaol-odata: moved.

---
In reply to: [155380791](https://github.com/OData/odata.net/pull/992#discussion_r155380791) [](ancestors = 155380791)

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: OK. This requires other refactoring of the ODataMultipartMixedBatchWriterUtils.CreateChangeSetBoundary as well.

---
In reply to: [155383582](https://github.com/OData/odata.net/pull/992#discussion_r155383582) [](ancestors = 155383582)

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
```diff
@@ -57,7 +73,9 @@ public sealed class ODataBatchOperationRequestMessage : IODataRequestMessage, IO
             string contentId,
             IODataPayloadUriConverter payloadUriConverter,
             bool writing,
-            IServiceProvider container)
+            IServiceProvider container,
+            IEnumerable<string> dependsOnIds = null,
```
mikepizzo: This internal constructor is called in two places, and both provide dependsOnIds, so there is no reason to define a default null value.

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: updated to throw exception if it is null.

---
In reply to: [155383773](https://github.com/OData/odata.net/pull/992#discussion_r155383773) [](ancestors = 155383773)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriterUtils.cs
biaol-odata: I don't think the user/client can specify the boundary the same way as can be done for changeSetBoundary.
The batch boundary is created by ODataMultipartMixedBatchFormat.GetContentType().

---
In reply to: [155385370](https://github.com/OData/odata.net/pull/992#discussion_r155385370) [](ancestors = 155385370)

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: sure.

---
In reply to: [155387064](https://github.com/OData/odata.net/pull/992#discussion_r155387064) [](ancestors = 155387064)

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: yes, that is the current implementation.
[https://github.com/OData/odata.net/blob/75df8f44f2b81f984589790be4885b6ee8946ad0/src/Microsoft.OData.Core/ODataBatchOperationResponseMessage.cs#L164](https://github.com/OData/odata.net/blob/75df8f44f2b81f984589790be4885b6ee8946ad0/src/Microsoft.OData.Core/ODataBatchOperationResponseMessage.cs#L164)
, where the response object is created with /*contentId*/ null.
I think it is due to the fact that contentId in the response is not much useful comparing to that in the request?
I agree that it doesn't hurt to add it. The question is whether we have to do this now.
thoughts?


---
In reply to: [155392178](https://github.com/OData/odata.net/pull/992#discussion_r155392178) [](ancestors = 155392178)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
biaol-odata: see above. This is consistent with the pre-refactoring logic in current master branch.

---
In reply to: [155392475](https://github.com/OData/odata.net/pull/992#discussion_r155392475) [](ancestors = 155392475)

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
biaol-odata: cleaned up. Thanks.

---
In reply to: [155401617](https://github.com/OData/odata.net/pull/992#discussion_r155401617) [](ancestors = 155401617)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
biaol-odata: this will be related to introducing dependsOn to multipart batch side.

---
In reply to: [155388122](https://github.com/OData/odata.net/pull/992#discussion_r155388122) [](ancestors = 155388122)

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: Updated to use  this.inputContext.BaseUri directly. Also use the batchFormat value to indicate whether this is from MultipartMixed batch, in which case the reference uri validate should be skipped if the dependsOnIds is not specified (value is null) for backward compatibility with original Multipart/Mixed request. 

---
In reply to: [155353476](https://github.com/OData/odata.net/pull/992#discussion_r155353476) [](ancestors = 155353476)

src/Microsoft.OData.Core/ODataBatchUtils.cs
biaol-odata: Updates have been made related to using BaseUri directly and passing in batchFormat, so that the coupling to the format-specific namespaces are reduced.
However I think the batchFormat is still needed. This method is invoked from a ODataBatchReader/Writer which are base classes for both formats,  therefore we cannot tell either formats apart. On the other hand we need to consider the backward compatibility to the original mulipart/mixed format where the dependsOnIds are not specified and shouldn't be rejected.

---
In reply to: [155347419](https://github.com/OData/odata.net/pull/992#discussion_r155347419) [](ancestors = 155347419)

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
```diff
@@ -57,7 +76,9 @@ public sealed class ODataBatchOperationRequestMessage : IODataRequestMessage, IO
             string contentId,
             IODataPayloadUriConverter payloadUriConverter,
             bool writing,
-            IServiceProvider container)
+            IServiceProvider container,
+            IList<string> dependsOnIds,
```
mikepizzo: >IList<string> [](start = 12, length = 13)

The input parameter should be an IEnumerable<string>. It doesn't need to be an IList, we copy it to our own list anyway (as we should).

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
mikepizzo: We should still only require that the caller pass in an IEnumerable<string>, rather than an IList<string>.  We copy it to our list (as we should) so it doesn't have to be a list. The rest is of the code is fine.

---
In reply to: [155390374](https://github.com/OData/odata.net/pull/992#discussion_r155390374) [](ancestors = 155390374,155349744)

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -204,48 +249,172 @@ void IODataBatchOperationListener.BatchOperationContentStreamDisposed()
...
+        /// <param name="batchFormat">Format of the batch.</param>
+        /// <returns>The <see cref="ODataBatchOperationRequestMessage"/> instance.</returns>
+        protected ODataBatchOperationRequestMessage BuildOperationRequestMessage(
+            Func<Stream> streamCreatorFunc,
+            string method,
+            Uri requestUri,
+            ODataBatchOperationHeaders headers,
+            string contentId,
+            string groupId,
+            IList<string> dependsOnRequestIds,
```
mikepizzo: >IList<string> [](start = 12, length = 13)

input should be IEnumerable<string> instead of IList<string>

src/Microsoft.OData.Core/ODataBatchReader.cs
mikepizzo: Still don't like passing in format, but will cover that in a separate comment.

---
In reply to: [156197925](https://github.com/OData/odata.net/pull/992#discussion_r156197925) [](ancestors = 156197925,155353476)

src/Microsoft.OData.Core/ODataBatchWriter.cs
mikepizzo: You can just use:
             ExceptionUtils.CheckArgumentNotNull(groupOrChangesetId, "groupOrChangesetId");


---
In reply to: [155405601](https://github.com/OData/odata.net/pull/992#discussion_r155405601) [](ancestors = 155405601,155383773)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchWriter.cs
```diff
@@ -0,0 +1,650 @@
...
+            Debug.Assert(groupId != null, "groupId != null");
+
+            // write pending message data (headers, response line) for a previously unclosed message/request
+            this.WritePendingMessageData(true);
+
+            // important to do this first since it will set up the change set boundary.
+            this.SetState(BatchWriterState.ChangesetStarted);
+
+            // When the changeset start is created for request, the <paramref name="groupId"/> could be null
+            // and we will create an GUID here.
```
mikepizzo: This part of the comment is no longer true.

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
```diff
@@ -0,0 +1,413 @@
...
+        {
+            return this.RawOutputContext.FlushAsync();
+        }
+#endif
+
+        /// <summary>
+        /// Starts a new changeset - implementation of the actual functionality.
+        /// </summary>
+        /// <param name="changeSetGuid">The value for changeset boundary for multipart batch.</param>
+        protected override void WriteStartChangesetImplementation(string changeSetGuid)
```
mikepizzo: >changeSetGuid [](start = 73, length = 13)

Should not be called "changeSetGuid" -- it's not necessarily a Guid, it could be any string identifier. "changeSetId" would be more accurate.

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -236,20 +201,63 @@ public Task WriteEndBatchAsync()
...
+        /// Cannot be null.
+        /// </param>
+        /// <returns>A task instance that represents the asynchronous write operation.</returns>
+        /// <exception cref="ODataException">Thrown if the <paramref name="groupOrChangesetId"/> is null.</exception>
+        public Task WriteStartChangesetAsync(string groupOrChangesetId)
+        {
+            if (groupOrChangesetId == null)
+            {
+                throw new ODataException(Strings.ODataBatch_GroupIdOrChangeSetIdCannotBeNull);
+            }
```
mikepizzo: as above, use:
             ExceptionUtils.CheckArgumentNotNull(groupOrChangesetId, "groupOrChangesetId");

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriterUtils.cs
mikepizzo: Yep; I understand. I was just suggesting that, for API consistency, the caller call Guid.NewGuid() rather than generating it here. But generating it here is fine too.

---
In reply to: [155407361](https://github.com/OData/odata.net/pull/992#discussion_r155407361) [](ancestors = 155407361,155385370)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
```diff
@@ -0,0 +1,413 @@
...
+        /// <param name="method">The Http method to be used for this request operation.</param>
+        /// <param name="uri">The Uri to be used for this request operation.</param>
+        /// <param name="contentId">The Content-ID value to write in ChangeSet head.</param>
+        /// <param name="payloadUriOption">
+        /// The format of operation Request-URI, which could be AbsoluteUri, AbsoluteResourcePathAndHost, or RelativeResourcePath.</param>
+        /// <param name="dependsOnIds">The prerequisite request ids of this request.</param>
+        /// <returns>The message that can be used to write the request operation.</returns>
+        protected override ODataBatchOperationRequestMessage CreateOperationRequestMessageImplementation(
+            string method, Uri uri, string contentId, BatchPayloadUriOption payloadUriOption,
+            IList<string> dependsOnIds)
```
mikepizzo: >IList<string> [](start = 12, length = 13)

Should be IEnumerable<string> rather than IList<string> (here and in any other input parameters, unless we are explicitly using the list (storing, adding/removing, etc.) which we generally shouldn't)

src/Microsoft.OData.Core/ODataBatchWriter.cs
mikepizzo: Yes, we should remove the assert and, in multiparmixedbatchwriter.CreationOperationResponseMessageImplementation, we should pass the contentId in to BuildOoperationResponseMessage.

---
In reply to: [155410202](https://github.com/OData/odata.net/pull/992#discussion_r155410202) [](ancestors = 155410202,155392178)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
mikepizzo: See above; if we have a content-id, we should pass it.

---
In reply to: [155410508](https://github.com/OData/odata.net/pull/992#discussion_r155410508) [](ancestors = 155410508,155392475)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
```diff
@@ -0,0 +1,413 @@
...
+        /// <param name="contentId">The Content-ID value to write in ChangeSet head.</param>
+        /// <returns>The message that can be used to write the response operation.</returns>
+        protected override ODataBatchOperationResponseMessage CreateOperationResponseMessageImplementation(string contentId)
+        {
+            this.WritePendingMessageData(true);
+
+            // In responses we don't need to use our batch URL resolver, since there are no cross referencing URLs
+            // so use the URL resolver from the batch message instead.
+            this.CurrentOperationResponseMessage = BuildOperationResponseMessage(
+                this.RawOutputContext.OutputStream, /*contentId*/ null, /*groupId*/ null);
```
mikepizzo: >/*groupId*/ null [](start = 72, length = 16)

for multipart/mixed, we should pass in the id used in generating the batchboundary.

src/Microsoft.OData.Core/ODataBatchOperationResponseMessage.cs
```diff
@@ -39,6 +38,8 @@ public sealed class ODataBatchOperationResponseMessage : IODataResponseMessage,
         /// <summary>The result status code of the response message.</summary>
         private int statusCode;
 
+
+
         /// <summary>
```
mikepizzo: remove extra blank lines

src/Microsoft.OData.Core/ODataBatchOperationResponseMessage.cs
```diff
@@ -94,6 +98,12 @@ public int StatusCode
         /// </summary>
         public IServiceProvider Container { get; private set; }
 
+        /// <summary>
+        /// For Json batch, this is the atomic group id of the corresponding request (null for top level request).
+        /// For multipart batch, this is not used and the value should be null.
```
mikepizzo: Technically, base class shouldn't comment on how one if its properties may/may not be used by derived class.  More importantly, as per previous comment, this should be set in multipart/mixed to the id used in the changeset boundary.

src/Microsoft.OData.Core/ODataBatchUtils.cs
mikepizzo: How about this:
1) Remove batchFormat from ODataBatchUtils.ValidateReferenceUri.
2) Replace the third parameter in ODataBatchUtils.ValidateReferenceUri with MessageReaderSettings (we may need additional information from MessageReaderSettings besides baseUri to properly validate in the future).
3) In BuildOperationRequestMessage, replace the "batchFormat" parameter with a Function that takes a requestUri, IEnumerable<string> dependency ids, and MessageReaderSettings (for extensibility this should probably actually be a dependency-injected ValidateUriReferenceService, but for now we can just make it a Func).
4) When called from Json, pass in ODataBatchUtils.ValidateReferenceUri.
5) When called from Multipart/Mixed, pass in an in-line defined Func that checks to see if dependencyIds are null and, if not, calls ODataBatchUtils.ValidateReferenceUri.
6) Same for BuildOperationRequestMessage in ODataBatchWriter

That way:
1) format-specific logic continues to be in the derived classes
2) multipart-mixed can retain existing non-validating behavior
3) we have extensibility for plugging in other validation logic

A cheaper version, if we don't think we'll want the ability to plug in custom validation logic, is to simply pass in a boolean ShouldValidateReferencedIds to BuildOperationRequestMessage.


---
In reply to: [156200555](https://github.com/OData/odata.net/pull/992#discussion_r156200555) [](ancestors = 156200555,155347419)

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
biaol-odata: Restored to IEnumerable for the argument type. Thanks.

---
In reply to: [156760101](https://github.com/OData/odata.net/pull/992#discussion_r156760101) [](ancestors = 156760101,155390374,155349744)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriterUtils.cs
```diff
@@ -172,14 +212,20 @@ internal static void WriteChangeSetPreamble(TextWriter writer, string changeSetB
...
             writer.WriteLine("{0}: {1}", ODataConstants.ContentTransferEncoding, ODataConstants.BatchContentTransferEncoding);
             if (inChangeSetBound && contentId != null)
             {
                 writer.WriteLine("{0}: {1}", ODataConstants.ContentIdHeader, contentId);
             }
+
+            if (!string.IsNullOrEmpty(dependsOnIds))
+            {
+                writer.WriteLine("{0}: {1}", ODataConstants.DependsOnIdsHeader, dependsOnIds);
+            }
```
mikepizzo: Sorry; I think I wasn't clear on my comment on DependencyIds in multipart/mixed.  I wasn't suggesting that we find a way to write out/read in dependencyIds in multipart/mixed -- that would be interpreting the existing format beyond what the protocol currently defines. The protocol currently says that, outside of a changeset, every statement is implicitly dependent upon the previous one.  So we shouldn't be writing out, or reading in, anything explicit for dependencyIds. Instead, when given a set of dependencyIds we can validate that all specified Ids are valid Ids of previous statements, and in populating dependencyids in multipartmixed we can include a list of all previously read statement ids (which would then make validation work).

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchReaderStream.cs
```diff
@@ -0,0 +1,811 @@
...
+                }
+
+                // Verify that we only allow single byte encodings and UTF-8 for now.
+                ReaderValidationUtils.ValidateEncodingSupportedInBatch(this.changesetEncoding);
+            }
+            else if (this.ChangeSetBoundary != null)
+            {
+                headers.TryGetValue(ODataConstants.ContentIdHeader, out contentId);
+                headers.TryGetValue(ODataConstants.DependsOnIdsHeader, out dependsOnIds);
+            }
```
mikepizzo: See other comments. I wasn't suggesting that we introduce a specific way to extend multipart/mixed with dependencyids. The protocol already defines that, for the multipartmixed format, each statement is implicitly dependent upon the previous one, so we can just populate the dependencyIds list with the list of all previous statements in the same changeset (or at the top level, for requests outside of a changeset).

src/Microsoft.OData.Core/ODataBatchOperationResponseMessage.cs
biaol-odata: done.

---
In reply to: [156778833](https://github.com/OData/odata.net/pull/992#discussion_r156778833) [](ancestors = 156778833)

src/Microsoft.OData.Core/ODataBatchOperationRequestMessage.cs
mikepizzo: >IList<string> [](start = 12, length = 13)

Should be IEnumerable<string>

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: updated. Also updated test as well.

---
In reply to: [156767675](https://github.com/OData/odata.net/pull/992#discussion_r156767675) [](ancestors = 156767675,155405601,155383773)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchWriter.cs
biaol-odata: removed. Thanks.

---
In reply to: [156768909](https://github.com/OData/odata.net/pull/992#discussion_r156768909) [](ancestors = 156768909)

src/Microsoft.OData.Core/ODataUtils.cs
```diff
@@ -301,5 +301,31 @@ public static string AppendDefaultHeaderValue(string headerName, string headerVa
...
+                if (string.CompareOrdinal(ODataConstants.HttpVersionInBatching, httpVersionSegment) != 0)
+                {
+                    throw new ODataException(Strings.ODataBatchReaderStream_InvalidHttpVersionSpecified(
+                        httpVersionSegment,
+                        ODataConstants.HttpVersionInBatching));
+                }
+            }
+
+            return new Uri(urlSegment, UriKind.RelativeOrAbsolute);
+        }
```
mikepizzo: why are we doing this validation? what if the service uses https?  or http 1.0?
We should just be validating the odata batching semantics and not get in the way of whatever other protocol they may be trying to use.

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
biaol-odata: Renamed here and in ODataMultipartMixedBatchWriterUtils. Solution-wide search doesn't indicate other changes needed.

---
In reply to: [156769543](https://github.com/OData/odata.net/pull/992#discussion_r156769543) [](ancestors = 156769543)

src/Microsoft.OData.Core/ODataUtils.cs
biaol-odata: OK, removed the validation part.

BTW, It was done so to follow the same pattern on multipart batch side:
[https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/ODataBatchReader.cs#L573](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/ODataBatchReader.cs#L573)
but I think we don't have to do this on JSON batch side.

---
In reply to: [156807126](https://github.com/OData/odata.net/pull/992#discussion_r156807126) [](ancestors = 156807126)

src/Microsoft.OData.Core/ODataBatchOperationResponseMessage.cs
biaol-odata: Updated with general description.

---
In reply to: [156779179](https://github.com/OData/odata.net/pull/992#discussion_r156779179) [](ancestors = 156779179)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
biaol-odata: OK, done. Shouldn't hurt to add, as long as we don't remove it.

---
In reply to: [156777000](https://github.com/OData/odata.net/pull/992#discussion_r156777000) [](ancestors = 156777000,155410508,155392475)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
biaol-odata: OK, also pass in the changeset id extracted from *changesetboundary* for groupId; contentId is directly using the value passed in. Please let me know any different thoughts.

---
In reply to: [156778151](https://github.com/OData/odata.net/pull/992#discussion_r156778151) [](ancestors = 156778151)

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: OK, done. 
Shouldn't hurt to add, as long as we don't remove the contentId later.

---
In reply to: [156776160](https://github.com/OData/odata.net/pull/992#discussion_r156776160) [](ancestors = 156776160,155410202,155392178)

src/Microsoft.OData.Core/ODataBatchUtils.cs
biaol-odata: As discussed offline with Mike, dependsOnIds for multipartMixed will include changes to enable tracking top-level statement for both V4 and V401 (and beyond). And this will be worked on after the 7.4.0-beta2 release. 

---
In reply to: [156797673](https://github.com/OData/odata.net/pull/992#discussion_r156797673) [](ancestors = 156797673,156200555,155347419)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriterUtils.cs
biaol-odata: rolled back the previous implementation. Please see the other comment of this similar issue, about the dependsOn Ids for top-level requests.

---
In reply to: [156802615](https://github.com/OData/odata.net/pull/992#discussion_r156802615) [](ancestors = 156802615)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchReaderStream.cs
biaol-odata: rolled back the previous implementation. 

---
In reply to: [156803180](https://github.com/OData/odata.net/pull/992#discussion_r156803180) [](ancestors = 156803180)

src/Microsoft.OData.Core/ODataBatchUtils.cs
mikepizzo: Do we have an issue to track this work if we close the comment?

---
In reply to: [157276865](https://github.com/OData/odata.net/pull/992#discussion_r157276865) [](ancestors = 157276865,156797673,156200555,155347419)

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -236,20 +201,57 @@ public Task WriteEndBatchAsync()
...
+        /// <summary>
+        /// Asynchronously starts a new change set; can only be called after WriteStartBatch and if no other active operation or change set exists.
+        /// </summary>
+        /// <param name="groupOrChangesetId">
+        /// The atomic group id (for Json batch) / changeset GUID (for Multipart/Mixed batch) of the batch request.
+        /// Cannot be null.
+        /// </param>
+        /// <returns>A task instance that represents the asynchronous write operation.</returns>
+        /// <exception cref="ODataException">Thrown if the <paramref name="groupOrChangesetId"/> is null.</exception>
+        public Task WriteStartChangesetAsync(string groupOrChangesetId)
```
mikepizzo: >groupOrChangesetId [](start = 52, length = 18)

From an API perspective, which should be format-agnostic, this is the id associated with the changeset.  Can we just call this "changesetId" and remove the format-specific comments in the <param> description?

src/Microsoft.OData.Core/ODataBatchUtils.cs
biaol-odata: I just created issue #1102  to track this.

---
In reply to: [157300821](https://github.com/OData/odata.net/pull/992#discussion_r157300821) [](ancestors = 157300821,157276865,156797673,156200555,155347419)

mikepizzo: Excellent; thanks!

---
In reply to: [157302502](https://github.com/OData/odata.net/pull/992#discussion_r157302502) [](ancestors = 157302502,157300821,157276865,156797673,156200555,155347419)

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: updated. Will push the commit.

---
In reply to: [157302410](https://github.com/OData/odata.net/pull/992#discussion_r157302410) [](ancestors = 157302410)

Issue #990, opened 2017-11-06
OData Client 7.4.0-Beta performance and bug


tchuan:
Updated my application from 6.18.0 to 7.4.0 beta, a performance test case showed about 20% decrease. 
Average time run from 10S down to 12S. Sometimes even failed. 

### Assemblies affected

OData Client 7.4.0-Beta

### Reproduce steps

```
    var tasks = new Task[1000];
    for (var i = 0; i < tasks.Length; i++)
    {
        tasks[i] = Task.Run(async () =>
        {
            var entity = await oDataRepository.FirstOrDefaultAsync<SampleEntity>();
            Assert.AreNotEqual(null, entity);
        });
    }

    Task.WaitAll(tasks);
```

### StackTrace:
Test Name:	PerformanceTest
Test Outcome:	Failed
Test Duration:	0:00:28.2052726

Result StackTrace:	
at System.Collections.Generic.Dictionary`2.Insert(TKey key, TValue value, Boolean add)
   at System.Collections.Generic.Dictionary`2.Add(TKey key, TValue value)
   at Microsoft.OData.Client.DictionaryExtensions.TryAdd[TKey,TValue](IDictionary`2 dict, TKey key, TValue value)
   at Microsoft.OData.Client.TypeResolver..ctor(ClientEdmModel model, Func`2 resolveTypeFromName, Func`2 resolveNameFromType, IEdmModel serviceModel)
   at Microsoft.OData.Client.RequestInfo..ctor(DataServiceContext context)
   at Microsoft.OData.Client.DataServiceRequest.CreateExecuteResult(Object source, DataServiceContext context, AsyncCallback callback, Object state, String method)
   at Microsoft.OData.Client.DataServiceQuery`1.BeginExecute(AsyncCallback callback, Object state)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncImpl(Func`3 beginMethod, Func`2 endFunction, Action`1 endAction, Object state, TaskCreationOptions creationOptions)
   at System.Threading.Tasks.TaskFactory`1.FromAsync(Func`3 beginMethod, Func`2 endMethod, Object state)
   at Microsoft.OData.Client.DataServiceQuery`1.ExecuteAsync()
   at ERM.CRM.OData.Repository.Extentions.ODataExtention.<ExecuteQueryAsync>d__1`1.MoveNext() in xxx.cs:line 41
Result Message:	
Test method PerformanceTest threw exception: 
[...]

PR #995, opened 2017-11-15
Use ConcurrentDictionary in all platforms to make client edm model thread safe

tchuan:
### Issues

*This pull request fixes issue #990 .*

### Description

* Generic Dictionary does not support multi thread read and write. .Net Standard 1.1 does support ConcurrentDictionary, which makes this a easy fix.

### DIFF
--- a/src/Microsoft.OData.Client/ClientEdmModel.cs
+++ b/src/Microsoft.OData.Client/ClientEdmModel.cs
@@ -21,13 +21,7 @@ namespace Microsoft.OData.Client
     using Microsoft.OData.Edm;
     using Microsoft.OData.Edm.Vocabularies;
     using c = Microsoft.OData.Client;
-
-#if PORTABLELIB
-    // Windows Phone 8.0 doesn't support ConcurrentDictionary
-    using ConcurrentEdmSchemaDictionary = System.Collections.Generic.Dictionary<string, Edm.IEdmSchemaElement>;
-#else
     using ConcurrentEdmSchemaDictionary = System.Collections.Concurrent.ConcurrentDictionary<string, Edm.IEdmSchemaElement>;
-#endif
 
     #endregion Namespaces.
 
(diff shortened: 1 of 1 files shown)


biaol-odata:
@AlanWong-MS This is related to the issue #990 that you might have some context. Can you take a look? I am not sure whether the latest ODL completely has switched to .Net Standard 1.1.

src/Microsoft.OData.Client/ClientEdmModel.cs
```diff
@@ -21,13 +21,7 @@ namespace Microsoft.OData.Client
     using Microsoft.OData.Edm;
     using Microsoft.OData.Edm.Vocabularies;
     using c = Microsoft.OData.Client;
-
-#if PORTABLELIB
```
AlanWong-MS: @tchuan, just had a discussion with the team and we think the appropriate change is actually to make the preprocessor target specifically Windows Phone 8.0 instead of removing the preprocessor entirely. This is due to the fact that we still build a test project against Windows Phone 8.0 (using OData Client portable lib, non-Net Standard version) and even if we don't plan to ship the portable lib, we don't want to publish a broken test until we've removed the test project and related projects (which we currently don't have the resources to review at this time).

Therefore, I recommend changing the preprocessor directive from
`#if PORTABLELIB`
to
`#if PORTABLELIB && WINDOWSPHONE`

The `WINDOWSPHONE` constant is defined in the [Windows Phone test project](https://github.com/OData/odata.net/blob/master/test/EndToEndTests/Tests/Client/Build.PortableLibrary.Phone/Microsoft.Test.OData.Tests.Client.Phone.csproj#L18) that I referred to.

tchuan:
Updated based on @AlanWong-MS 's suggestion.

AlanWong-MS:
Tests pass on our buddy build. Thanks for the fix!

Issue #1009, opened 2017-12-05
Writing undeclared untyped properties to open types fails.


mikepizzo:
*Writing an undeclared untyped property to an open entity fails*

### Assemblies affected

*ODL V7*

### Reproduce steps
1. Define an open entity type
2. Attempt to write an undefined property:
   new ODataProperty { Name = "UndeclaredProperty", Value = new ODataUntypedValue { RawValue = "\"rawValue\"" }) };
 
### Expected result

*The undeclared property should be written since the type is open*

### Actual result

Error Message "The property 'OpenCollectionNoType' does not exist on type 'microsoft.graph.S1E2'. Make sure to only use property names that are defined by the type or mark the type as open type."

### Additional detail

This works for undefined *typed* properties in an open entity type, it's only an issue for untyped properties.
This worked in ODL V6 and is blocking folks from moving to ODL V7.

mikepizzo:
Fixed by change #1010.

Issue #1008, opened 2017-12-05
Validations.None still validates property names in ODL V7


mikepizzo:
*In ODL V6, setting ODataMessageWriterSettings.EnableFullValidation = false enables writing special characters (@, .) in property names.  Setting Validations.None in ODL V7 does not support writing these special characters.*

### Assemblies affected

*OData .Net V7.x*

### Reproduce steps

Set WriterSettings.Validations to none and attempt to write a property whose name contains "@" or ".".

### Expected result

The property should be written.

### Actual result

Error Message=The property name 'dynamicExpandedNavigation@odata.nextLink' is invalid; property names must not contain any of the reserved characters ':', '.', '@'.

### Additional Information

This is blocking customers attempting to move from ODL V6 to ODL V7.

mikepizzo:
Fixed by change #1010.

srinivasupadhya:
@mikepizzo can you please repose to this [post](https://stackoverflow.com/questions/55085301/odata-disable-property-name-validation) if you know the answer?

PR #1010, opened 2017-12-05
Fix property validation issues

mikepizzo:
### Issues

*This pull request fixes issue #1008*
*This pull request fixes issue #1009*

### Description

*Supports V6 capability of:
1) Writing special characters in property names if Validations.None is specified, and
2) Writing undefined untyped properties to open types*

- [x] *Test cases added*

### Additional work necessary

*None.*

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
@@ -183,9 +183,13 @@ namespace Microsoft.OData.JsonLight
 
             string propertyName = property.Name;
 
-            if (!this.JsonLightOutputContext.PropertyCacheHandler.InResourceSetScope())
+            if (this.JsonLightOutputContext.MessageWriterSettings.Validations != ValidationKinds.None)
             {
                 WriterValidationUtils.ValidatePropertyName(propertyName);
+            }
+
+            if (!this.JsonLightOutputContext.PropertyCacheHandler.InResourceSetScope())
+            {
                 this.currentPropertyInfo = new PropertySerializationInfo(propertyName, owningType) { IsTopLevel = isTopLevel };
             }
             else
@@ -261,17 +265,9 @@ namespace Microsoft.OData.JsonLight
 
         private void WriteUntypedValue(ODataUntypedValue untypedValue)
         {
-            if (!this.MessageWriterSettings.ThrowOnUndeclaredPropertyForNonOpenType)
-            {
-                this.JsonWriter.WriteName(this.currentPropertyInfo.WireName);
-                this.jsonLightValueSerializer.WriteUntypedValue(untypedValue);
-                return;
-            }
-
-            Debug.Assert(
-                this.MessageWriterSettings.ThrowOnUndeclaredPropertyForNonOpenType,
-                "this.MessageWriterSettings.ThrowOnUndeclaredPropertyForNonOpenType");
-            throw new ODataException(ODataErrorStrings.ValidationUtils_PropertyDoesNotExistOnType(this.currentPropertyInfo.PropertyName, this.currentPropertyInfo.MetadataType.OwningType.FullTypeName()));
+            this.JsonWriter.WriteName(this.currentPropertyInfo.WireName);
+            this.jsonLightValueSerializer.WriteUntypedValue(untypedValue);
+            return;
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Core/PropertyCache.cs
+++ b/src/Microsoft.OData.Core/PropertyCache.cs
@@ -21,8 +21,6 @@ namespace Microsoft.OData
             PropertySerializationInfo propertyInfo;
             if (!propertyInfoDictionary.TryGetValue(uniqueName, out propertyInfo))
             {
-                WriterValidationUtils.ValidatePropertyName(name);
-
                 propertyInfo = new PropertySerializationInfo(name, owningType);
                 propertyInfoDictionary[uniqueName] = propertyInfo;
             }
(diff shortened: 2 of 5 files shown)


src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -261,17 +265,9 @@ private bool IsOpenProperty(ODataProperty property)
...
-            {
-                this.JsonWriter.WriteName(this.currentPropertyInfo.WireName);
-                this.jsonLightValueSerializer.WriteUntypedValue(untypedValue);
-                return;
-            }
-
-            Debug.Assert(
-                this.MessageWriterSettings.ThrowOnUndeclaredPropertyForNonOpenType,
-                "this.MessageWriterSettings.ThrowOnUndeclaredPropertyForNonOpenType");
-            throw new ODataException(ODataErrorStrings.ValidationUtils_PropertyDoesNotExistOnType(this.currentPropertyInfo.PropertyName, this.currentPropertyInfo.MetadataType.OwningType.FullTypeName()));
```
xuzhg: Remove this is a breaking change?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -183,9 +183,13 @@ private bool IsOpenProperty(ODataProperty property)
 
             string propertyName = property.Name;
 
-            if (!this.JsonLightOutputContext.PropertyCacheHandler.InResourceSetScope())
+            if (this.JsonLightOutputContext.MessageWriterSettings.Validations != ValidationKinds.None)
```
xuzhg: maybe:
```C#
if ( ... .InResourceSetScope)
{
          if ( .. != ValidationKinds.None)
          { 
                    ValidatePropertyName(...);
          }
          ...
}
else
{
      ...
}
```

src/Microsoft.OData.Core/PropertyCache.cs
```diff
@@ -21,8 +21,6 @@ public PropertySerializationInfo GetProperty(string name, string uniqueName, IEd
             PropertySerializationInfo propertyInfo;
             if (!propertyInfoDictionary.TryGetValue(uniqueName, out propertyInfo))
             {
-                WriterValidationUtils.ValidatePropertyName(name);
```
xuzhg: Just remove the validate?  or use the validate setting?

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Writer/JsonLight/ODataJsonLightWriterIntegrationTests.cs
```diff
@@ -748,7 +790,8 @@ private static string WriteJsonLightEntry(bool isRequest, Uri serviceDocumentUri
         private static string WriteJsonLightEntry(bool isRequest, Uri serviceDocumentUri, bool specifySet,
             ODataResource odataEntry, IEdmNavigationSource entitySet, IEdmStructuredType resourceType,
             ODataUri odataUri, Action<ODataWriter> writeAction = null, bool isResourceSet = false,
-            EdmModel model = null)
+            EdmModel model = null,
+            bool ValidatePropertyNames = true)
```
xuzhg: maybe it's better to accept "ValidationKinds" enum?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
mikepizzo: We do the check before this, in WriteProperty (which is where we should check the property name, not in WriteUntypedValue). There we correctly check to see if the type is no-open.  So this code is only reached after we've verified that that it's valid.

So the only behavioral change is, if the property is undeclared and the type is open we no longer throw the exception for untyped values.  This enables a code path that was previously an error (but shouldn't have been).   

---
In reply to: [154829236](https://github.com/OData/odata.net/pull/1010#discussion_r154829236) [](ancestors = 154829236)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
mikepizzo: In the else case, PropertyCacheHandler eventually also called ValidatePropertyName. So it was called in both cases, but in weird places. Moving it out of the if statement provided a common place to validate, regardless of whether or not we were in a resourcesetscope (which shouldn't affect property validation).

---
In reply to: [154829506](https://github.com/OData/odata.net/pull/1010#discussion_r154829506) [](ancestors = 154829506)

src/Microsoft.OData.Core/PropertyCache.cs
mikepizzo: This is only called from WriteProperty.  By moving the ValidatePropertyName first in WriteProperty, we do the validation (correctly checking whether or not the type is open) before this code ever gets called.  Which is good, since this code doesn't have the context of the defining type to know whether or not it was open.

---
In reply to: [154829612](https://github.com/OData/odata.net/pull/1010#discussion_r154829612) [](ancestors = 154829612)

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Writer/JsonLight/ODataJsonLightWriterIntegrationTests.cs
mikepizzo: Good suggestion, but this is internal test code; we can change easily change it later if/when we add tests that handle other ValidationKinds.

---
In reply to: [154829788](https://github.com/OData/odata.net/pull/1010#discussion_r154829788) [](ancestors = 154829788)

mikepizzo:
Thanks Sam!

I added replies to your comments.  The change is a bit subtle, since it looks like I removed validations from some cases and added it to others, but by moving the ValidatePropertyNames to the front of WriteProperty it makes sure that the property name is validated in the right situations and with the right context, rather than further down where the context is not available. The only change in functionality should be that undeclared untyped properties on open types should now succeed rather than fail.

The change in the test method to take ValidationKinds enum, rather than ValidatePropertyNames, probably makes sense, except that it implies we handle other types of validations (we don't)  If I had to make other changes I'd probably take this suggestion, but I don't think it's worth updating the PR for since we) we can easily change this (since it's test code) in the future if we want to handle other ValidationKinds.

xuzhg:
@mikepizzo Do I need to merge it or you will do it by yourself?

PR #1015, opened 2017-12-15
Align PublicApi.bsl with preceding API prolog changes.

biaol-odata:
### Issues

*This pull request fixes issue in Public API bsl.*

### Description

*Align the changed parameter name in PublicAPI.bsl since it checks param name as well.*

- [ ] *Test cases added*

### Additional work necessary


biaol-odata:
cancelled

PR #1017, opened 2017-12-16
FxCopFixes: use ToLowerInvariant

biaol-odata:
### Issues

*This pull request fixes FxCop error related to using ToLowerInvariant().*

### Description

*Replacing ToLower() with ToLowerInvariant() in ODataJsonLightBatchWriter.*

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchWriter.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchWriter.cs
@@ -586,6 +586,7 @@ namespace Microsoft.OData.JsonLight
         /// <summary>
         /// Writing pending data for the current request message.
         /// </summary>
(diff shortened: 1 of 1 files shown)


src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchWriter.cs
```diff
@@ -598,7 +598,7 @@ private void WritePendingRequestMessageData()
             {
                 foreach (KeyValuePair<string, string> headerPair in headers)
                 {
-                    this.jsonWriter.WriteName(headerPair.Key.ToLower());
+                    this.jsonWriter.WriteName(headerPair.Key.ToLowerInvariant());
```
xuzhg: Please use the VSTS to verify all such build/fxcop/stylecop issues.

Issue #1011, opened 2017-12-08
Microsoft.OData.Client 7.0+ needs to support GetValue like the older OData Client


ImGonaRot:
Current users that have the Microsoft.OData.Client below 7.0 could use GetValue in non-async methods on the clients.

There are tons of examples on the Internet on using GetValue with the OData Client.
Some developers (like me) are using Winforms and in some cases are not allow to use async methods in some of the code and as such will not be able to use the OData Client 7.0+ without creating wrapper methods to call BeginGetValue and EndGetValue.

This also breaks all current T4 template proxy generators like connected services and OData client generator.
The T4 template proxy generator (connected services) is also broken because "Microsoft.OData.Edm.Csdl.EdmxReader.Parse(reader)" change to "Microsoft.OData.Edm.Csdl.CsdlReader.Parse(reader)"

### Assemblies affected

Microsoft.OData.Client 7.0+

AlanWong-MS:
Thanks for the feedback @ImGonaRot. We're adding back the OData Client Net45 library so that the sync versions come back. This is exactly what we needed to hear to make informed decisions. Our OData 7.4 beta 2 release will include the Net45 lib, per PR #1018.

PR #1018, opened 2017-12-18
Adding OData Client Net45 binary

AlanWong-MS:
Adding back the OData Client binary for Net45 framework per feedback from community. For reference, the PR that removed the binaries was #981.

### Issues
#731, #1011

### Description
A few users depended on particular code patterns found only in the non-portable versions of OData Client (i.e. Net45 framework), so this PR is to add back that Client version based on the feedback.

- [x] *Test cases added* (N/A)

### Additional work necessary
None

### DIFF
--- a/src/Microsoft.OData.Client/Build.NuGet/Microsoft.OData.Client.Nightly.Release.nuspec
+++ b/src/Microsoft.OData.Client/Build.NuGet/Microsoft.OData.Client.Nightly.Release.nuspec
@@ -24,6 +24,12 @@ OData .NET library is open source at http://github.com/OData/odata.net</descript
     </frameworkAssemblies>
   </metadata>
   <files>
+    <file src="$ProductRoot$\Desktop\Microsoft.OData.Client.dll" target="lib\net45" />
+    <file src="$ProductRoot$\Desktop\Microsoft.OData.Client.xml" target="lib\net45" />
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Client/Build.NuGet/Microsoft.OData.Client.Nightly.Release.nuspec
```diff
@@ -24,6 +24,12 @@ OData .NET library is open source at http://github.com/OData/odata.net</descript
     </frameworkAssemblies>
   </metadata>
   <files>
+    <file src="$ProductRoot$\Desktop\Microsoft.OData.Client.dll" target="lib\net45" />
+    <file src="$ProductRoot$\Desktop\Microsoft.OData.Client.xml" target="lib\net45" />
+    <file src="$ProductRoot$\Desktop\Microsoft.OData.Client.pdb" target="lib\net45" />
+    <file src="$ProductRoot$\.NETPortable\v4.5\Microsoft.OData.Client.dll" target="lib\portable-net45+win8+wpa81" />
```
robward-ms: Let's just do the net45, Netstandard1.1 and portable-net45+win8+wpa81 should be equivalent.

AlanWong-MS: Woops, I jumped the gun there.

PR #1020, opened 2017-12-20
DependsOn Ids for Multipart/Mixed Batch

biaol-odata:
- Reader extracts contentId header value for both changeset and top-level requests;
- Uses DependsOnIdsTracker to track request ids in Multipart/Mixed batch;
- DependsOnIds validation is performed only for JsonBatch since it is client provided; not for Multipart/Mixed since it is not provided by client.

### Issues

*This pull request is to introduce DependsOn ids to Multipart/Mixed batch format. This makes DependsOn Ids available for query when service processes the batch request.*

### Description

*Introducing DependsOn IDs to Multipart/Mixed batch so that both formats are symmetrical in semantics. The dependsOn ids for multipart/mixed format is derived from protocol and is not explicitly provided by client.*

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -534,17 +533,18 @@ protected virtual IEnumerable<string> GetDependsOnRequestIds(IEnumerable<string>
...
+            string contentId, string groupId, IEnumerable<string> dependsOnIds, bool dependsOnIdsValidationRequired)
         {
             IEnumerable<string> flattenDependsOnIds = dependsOnIds == null
                 ? null
                 : GetDependsOnRequestIds(dependsOnIds);
 
-            if (flattenDependsOnIds != null)
+            if (flattenDependsOnIds != null && dependsOnIdsValidationRequired)
             {
+                // Validate dependsOn ids in Json batch.
```
mikepizzo: Base class should not refer to derived classes, even in comments.  We validate dependsOnIds if dependsOnIdsValidationRequired is true.  Why it's true is outside of the scope of this class.

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
```diff
@@ -207,25 +216,42 @@ protected override void WriteStartChangesetImplementation(string changeSetId)
...
             string method, Uri uri, string contentId, BatchPayloadUriOption payloadUriOption,
             IEnumerable<string> dependsOnIds)
         {
+            if (dependsOnIds != null)
+            {
+                throw new ODataException(
+                    "For Multipart/Mixed batch format, request dependency is implicit by the protocol and should be null.");
+            }
+
             // write pending message data (headers, response line) for a previously unclosed message/request
```
mikepizzo: Do not enforce null. We should validate that any ids specified are in the ids read by the tracker for this scope (changeset or toplevel).

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
```diff
@@ -207,25 +216,42 @@ protected override void WriteStartChangesetImplementation(string changeSetId)
...
             // create the new request operation
+            // For Multipart batch format, dependsOnIds is implicitly derived from batch message itself and is not
+            // user input, therefore no validation is needed.
             ODataBatchOperationRequestMessage operationRequestMessage = BuildOperationRequestMessage(
                 this.RawOutputContext.OutputStream,
-                method, uri, contentId, /*groupId*/null, dependsOnIds, ODataFormat.Batch);
+                method, uri, contentId,
+                ODataMultipartMixedBatchWriterUtils.GetChangeSetIdFromBoundary(this.changeSetBoundary),
+                this.dependsOnIdsTracker.GetDependsOnIds(),
+                /*dependsOnIdsValidationRequired*/ false);
```
mikepizzo: >/*dependsOnIdsValidationRequired*/ false [](start = 16, length = 40)

Not sure why we wouldn't pass true in order to validate against the ids already read? Is this just for backward compatibility? Seems like the request is destined to fail if it doesn't reference a valid id.

src/Microsoft.OData.Core/DependsOnIdsTracker.cs
```diff
@@ -0,0 +1,90 @@
...
+    /// <summary>
+    /// This class is to keep track of dependsOn ids and associated change set state,
+    /// and to return dependsOn ids pertaining to current state of the batch processing.
+    /// </summary>
+    internal sealed class DependsOnIdsTracker
+    {
+        /// <summary>
+        /// List of top-level dependsOn ids seen so far.
+        /// </summary>
+        private readonly IList<string> topLevelDepeondsOnIds;
```
mikepizzo: >topLevelDepeondsOnIds [](start = 39, length = 21)

Spelling; should be topLevelDependsOnIds

mikepizzo:
>                 if (HttpUtils.IsQueryMethod(method))

This restriction (inability to have query methods in a changeset) is no longer true.  Please add "&& this.outputContext.MessageWriterSettings.Version <= ODataVersion.V4"

---
Refers to: src/Microsoft.OData.Core/ODataBatchWriter.cs:811 in 2ce1695. [](commit_id = 2ce169536b0f827326ada572678efc66cc3762fa, deletion_comment = False)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
biaol-odata: removed null check / exception thrown

---
In reply to: [158350729](https://github.com/OData/odata.net/pull/1020#discussion_r158350729) [](ancestors = 158350729)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
biaol-odata: Updated to use only dependsOnIds which is only null for the implicit dependsOnIds Multipart/Mixed case; and the GetDependsOnIds overrides in derived classes getting the convertedDependsOnIds in the BuildOperationRequestMessage method.

---
In reply to: [158351447](https://github.com/OData/odata.net/pull/1020#discussion_r158351447) [](ancestors = 158351447)

src/Microsoft.OData.Core/DependsOnIdsTracker.cs
biaol-odata: oops! fixed.

---
In reply to: [159333403](https://github.com/OData/odata.net/pull/1020#discussion_r159333403) [](ancestors = 159333403)

biaol-odata:
>                 if (HttpUtils.IsQueryMethod(method))

I think this is aligned to the protocol and we keep this restriction at least for now per offline discussion with the group. Right?

---
In reply to: [355098382](https://github.com/OData/odata.net/pull/1020#issuecomment-355098382) [](ancestors = 355098382)

---
Refers to: src/Microsoft.OData.Core/ODataBatchWriter.cs:811 in 2ce1695. [](commit_id = 2ce169536b0f827326ada572678efc66cc3762fa, deletion_comment = False)

src/Microsoft.OData.Core/ODataBatchWriter.cs
mikepizzo: Please remove references to JSON and Multipart from comments in the base class.  Comment should read something like "Validate dependsOnIds if specified".

---
In reply to: [158349752](https://github.com/OData/odata.net/pull/1020#discussion_r158349752) [](ancestors = 158349752)

src/Microsoft.OData.Core/DependsOnIdsTracker.cs
```diff
@@ -0,0 +1,90 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="DependsOnIdsTracker.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData
+{
```
mikepizzo: If this is only used by multipart/mixed, should it go in the multipart/mixed namespace?

src/Microsoft.OData.Core/ODataBatchWriter.cs
```diff
@@ -554,8 +549,7 @@ protected virtual IEnumerable<string> GetDependsOnRequestIds(IEnumerable<string>
                 }
             }
 
-            ODataBatchUtils.ValidateReferenceUri(uri, flattenDependsOnIds,
-                this.outputContext.MessageWriterSettings.BaseUri, batchFormat);
+            ODataBatchUtils.ValidateReferenceUri(uri, convertedDependsOnIds, this.outputContext.MessageWriterSettings.BaseUri);
```
mikepizzo: >convertedDependsOnIds [](start = 54, length = 21)

I don't think this is the right logic.  I thought that, if the derived class passed in null for dependsOnIds, we were going to use the ids from the payloadUriConverter (which would be be correctly scoped for the odata version).  The ids from the DependsOnIdsTracker will be return only the ids in the changeset (if currently in changeset) otherwise top level ids, which is correct for 4.0, but wrong for 4.01.  I actually don't think we need the dependsOnIdsTracker at all for the ODataMultipartMixedBatchWriter, only for the ODataMultipartMixedBatchReader.

biaol-odata: There are pros and cons of using payloadUrlConverter. I thought about using it but ended up not doing it because it is undesirable to couple with legacy issue of difference between changeset scope in V4.0 implementation and batch scope in V4.0 protocol. In other words, current payloadUrlConverter is actually implemented incorrectly against the V4.0 protocol.
Furthermore, Json Batch Format 19.1: (for referencing url...)"The id of this request MUST be specified in the dependsOn name/value pair", it is natural extension to validate the referenced Id in url using the dependsOnIds? 

I understand doing this will allow, on the writer side, the referencing of preceding top level request, e.g. request_2's url uses "$request_1", which should not be allowed in V4.0 implementation due to scope issue above but should not be an problem in V4.01. 

All these said, I just don't want to let the scope issue above limit the validation on the writer side, and the referencing urls in both cases above will still be validated / processed correctly on the reader side.

---
In reply to: [160558905](https://github.com/OData/odata.net/pull/1020#discussion_r160558905) [](ancestors = 160558905)

mikepizzo: For the case of the non-null dependsOnIds (JSON batch and multipartmixed with explicit dependsOnIds) yes, absolutely we should use the (flattened) list of dependsOnIds explicitly passed by the user.  However, for the dependsOnIds==null (multipartmixed implicit) case, we should default to the set of Ids in the payloadUrlConverter.

---
In reply to: [160567393](https://github.com/OData/odata.net/pull/1020#discussion_r160567393) [](ancestors = 160567393,160558905)

src/Microsoft.OData.Core/DependsOnIdsTracker.cs
biaol-odata: yes, moved.

---
In reply to: [160557298](https://github.com/OData/odata.net/pull/1020#discussion_r160557298) [](ancestors = 160557298)

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: done for base classes reader and writer. 

---
In reply to: [160554679](https://github.com/OData/odata.net/pull/1020#discussion_r160554679) [](ancestors = 160554679,158349752)

src/Microsoft.OData.Core/ODataBatchWriter.cs
biaol-odata: updated with tests added.

---
In reply to: [160573714](https://github.com/OData/odata.net/pull/1020#discussion_r160573714) [](ancestors = 160573714,160567393,160558905)

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
```diff
@@ -217,15 +240,24 @@ protected override void WriteStartChangesetImplementation(string changeSetId)
             this.WritePendingMessageData(true);
 
             // create the new request operation
+            // For Multipart batch format, validate dependsOnIds if it is user explicit input, otherwise skip validation
+            // when it is implicitly derived per protocol.
             ODataBatchOperationRequestMessage operationRequestMessage = BuildOperationRequestMessage(
                 this.RawOutputContext.OutputStream,
-                method, uri, contentId, /*groupId*/null, dependsOnIds, ODataFormat.Batch);
```
mikepizzo: >/*groupId*/null [](start = 40, length = 15)

Why did we start passing null, rather than ODataMultipartMixedBatchWriterUtils.GetChangeSetIdFromBoundary(this.changeSetBoundary)?

mikepizzo:
>                 if (HttpUtils.IsQueryMethod(method))

Actually, I don't think this restriction is in the json specification. We should probably make it either 4.0 specific (here), move it to multipart/mixed, or remove it all together. In general, if a service can support GET requests in a changeset I'd rather not get in their way, so we should probably just make this a 4.0 restriction.

---
In reply to: [356447377](https://github.com/OData/odata.net/pull/1020#issuecomment-356447377) [](ancestors = 356447377,355098382)

---
Refers to: src/Microsoft.OData.Core/ODataBatchWriter.cs:811 in 2ce1695. [](commit_id = 2ce169536b0f827326ada572678efc66cc3762fa, deletion_comment = False)

src/Microsoft.OData.Core/ODataBatchReader.cs
```diff
@@ -333,9 +327,9 @@ protected virtual string GetCurrentGroupIdImplementation()
             string contentId,
             string groupId,
             IEnumerable<string> dependsOnRequestIds,
-            ODataFormat batchFormat)
+            bool dependsOnIdsValidationRequired)
```
mikepizzo: >bool dependsOnIdsValidationRequired [](start = 12, length = 35)

Can we do the same thing that we did in the writer, and pass null versus an empty list of requestIds instead of a boolean flag here, and validate the Url against the payloadUriConverter in the null dependsOnIds case?

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
biaol-odata: groupId is introduced with JSON Batch. For MultipartMixed batch, we have changeset boundary instead and we originally considered groupId value is always null. And then we decided to use the value extracted from changeset boundary as group Id, likely from one of your comments.

---
In reply to: [160844104](https://github.com/OData/odata.net/pull/1020#discussion_r160844104) [](ancestors = 160844104)

mikepizzo: Sorry; I was reacting to an apparent change to pass null here, but that was an old iteration.  Now we pass GetChangeSetIdFromBoundary, which is what I would expect.

---
In reply to: [160846771](https://github.com/OData/odata.net/pull/1020#discussion_r160846771) [](ancestors = 160846771,160844104)

src/Microsoft.OData.Core/ODataBatchReader.cs
biaol-odata: I thought about that but we are out of luck / by design here. The dependsOnRequestIds is always not null, and it is always derived (implicit) for Multipart/Mixed batch, and always explicit for Json batch.

---
In reply to: [160846029](https://github.com/OData/odata.net/pull/1020#discussion_r160846029) [](ancestors = 160846029)

mikepizzo: What would happen if we removed the dependsOnIdsValidationRequired flag and always validated the dependsOnRequestIds?  Shouldn't the implicit dependsOnRequestIds accumulated in the multipart/mixed case always be valid?

---
In reply to: [160847347](https://github.com/OData/odata.net/pull/1020#discussion_r160847347) [](ancestors = 160847347,160846029)

biaol-odata: In V4, since the payloadUriConverter is reset after changset, the validation (L336) will throw because implicit dependsOnIds could contain Content-IDs from preceding top-level requests --> error.

---
In reply to: [160848457](https://github.com/OData/odata.net/pull/1020#discussion_r160848457) [](ancestors = 160848457,160847347,160846029)

mikepizzo: Right; in 4.0, requests can only reference other requests within the same changeset, so payloadUriConverter only has the set of ids in the changeset, which is a subset of the requests that this request depends on coming before this request. Makes sense. 

---
In reply to: [160849419](https://github.com/OData/odata.net/pull/1020#discussion_r160849419) [](ancestors = 160849419,160848457,160847347,160846029)

PR #1025, opened 2018-01-06
Fix Issue 1193; add back extension function to derived class.

mikepizzo:
### Issues

*This pull request fixes WebAPI OData issue https://github.com/OData/WebApi/issues/1193.*

### Description

*In refactoring the code for Delta support we moved an extension method from a derived class to a new base class.  No code changes for newly compiled apps, but existing apps break because the resolver doesn't find the old signature (and doesn't resolve it to the new method on the base class).*

- [n/a ] *Test cases added*
- [ x ] *Build and test with one-click build and test script passed*

### Additional work necessary
None.

### DIFF
--- a/src/Microsoft.OData.Core/ODataObjectModelExtensions.cs
+++ b/src/Microsoft.OData.Core/ODataObjectModelExtensions.cs
@@ -35,6 +35,17 @@ namespace Microsoft.OData
         }
 
         /// <summary>
+        /// Provide additional serialization information to the <see cref="ODataWriter"/> for <paramref name="resource"/>.
+        /// </summary>
+        /// <param name="resource">The instance to set the serialization info.</param>
+        /// <param name="serializationInfo">The serialization info to set.</param>
+        public static void SetSerializationInfo(this ODataResource resource, ODataResourceSerializationInfo serializationInfo)
+        {
+            ExceptionUtils.CheckArgumentNotNull(resource, "resource");
+            resource.SerializationInfo = serializationInfo;
+        }
+
+        /// <summary>
         /// Provide additional serialization information to the <see cref="ODataWriter"/> for <paramref name="nestedResourceInfo"/>.
         /// </summary>
         /// <param name="nestedResourceInfo">The instance to set the serialization info.</param>
--- a/test/FunctionalTests/Tests/DataOData/Tests/OData.Common.Tests/PublicApi/PublicApi.bsl
+++ b/test/FunctionalTests/Tests/DataOData/Tests/OData.Common.Tests/PublicApi/PublicApi.bsl
@@ -4541,6 +4541,11 @@ public sealed class Microsoft.OData.ODataObjectModelExtensions {
 	[
 	ExtensionAttribute(),
 	]
+	public static void SetSerializationInfo (Microsoft.OData.ODataResource resource, Microsoft.OData.ODataResourceSerializationInfo serializationInfo)
+
+	[
+	ExtensionAttribute(),
+	]
 	public static void SetSerializationInfo (Microsoft.OData.ODataResourceBase resource, Microsoft.OData.ODataResourceSerializationInfo serializationInfo)
 
 	[


xuzhg:
@mikepizzo It seems the PublicApi test case should be modified also.

mikepizzo:
Right. Didn't get a chance to do that before I left on Friday, but will finish the PR for review today.

xuzhg:
>         public static void SetSerializationInfo(this ODataResourceBase resource, ODataResourceSerializationInfo serializationInfo)

So, how about the overload? Maybe:

1. Make it internal?
2. Or remove it?



---
Refers to: src/Microsoft.OData.Core/ODataObjectModelExtensions.cs:31 in 6f0c634. [](commit_id = 6f0c6343a8cef790e7159e0431b96be34ee971ef, deletion_comment = False)

src/Microsoft.OData.Core/ODataObjectModelExtensions.cs
```diff
@@ -34,6 +34,17 @@ public static void SetSerializationInfo(this ODataResourceBase resource, ODataRe
             resource.SerializationInfo = serializationInfo;
         }
 
+        /// <summary>
+        /// Provide additional serialization information to the <see cref="ODataWriter"/> for <paramref name="resource"/>.
+        /// </summary>
+        /// <param name="resource">The instance to set the serialization info.</param>
+        /// <param name="serializationInfo">The serialization info to set.</param>
+        public static void SetSerializationInfo(this ODataResource resource, ODataResourceSerializationInfo serializationInfo)
```
xuzhg: >public static void SetSerializationInfo [](start = 8, length = 39)

The whole method body is to assign the value to its property. 
So, what's the advantages to add such extension method?

Can we just make the internal property 'SerializationInfo'  as public?

mikepizzo: I honestly don't know why this extension method was created in the first place, but it's shipped so we need to retain it. Can revisit in 8.x.

mikepizzo:
The overload that specifies the base class is required for applying serializationInfo to other classes derived from ODataResourceBase (notably, ODataDeletedResource).

xuzhg:
So, Maybe it's better to add the similar extension method for the derived types for consistent? For example:
```C#
public static void SetSerializationInfo(this ODataDeletedResource resource...)
```

mikepizzo:
The serializationInfo is on the base class, so it makes sense to have the extension method on the base class. That way if we add a new derived class in the future it will also automatically have the extension method.  The only reason to have the extension method on the derived class is for backward compatibility with compiled code, otherwise we would just define it on the base class.

mikepizzo:
>         public static void SetSerializationInfo(this ODataResourceBase resource, ODataResourceSerializationInfo serializationInfo)

We need to keep the overload for the base class in order to support ODataDeletedResource and other derived types in the future. It makes sense for the extension to be on the base class, since that's where SerializationInfo is defined, but we need to keep it in the derived type in order not to break compatibility with compiled clients.

---
In reply to: [356116599](https://github.com/OData/odata.net/pull/1025#issuecomment-356116599) [](ancestors = 356116599)

---
Refers to: src/Microsoft.OData.Core/ODataObjectModelExtensions.cs:31 in 6f0c634. [](commit_id = 6f0c6343a8cef790e7159e0431b96be34ee971ef, deletion_comment = False)

src/Microsoft.OData.Core/ODataObjectModelExtensions.cs
mikepizzo: I honestly don't know why this extension method was created in the first place, but it's shipped so we need to retain it. Can revisit in 8.x.

---
In reply to: [160273548](https://github.com/OData/odata.net/pull/1025#discussion_r160273548) [](ancestors = 160273548)

mprigge:
@mikepizzo Hi Mike! Do you have an idea of when this change might find its way into a prerelease binary?

mikepizzo:
Hi Matt;

I'd like to do a very short beta 3 with this fix (and a few other features) in the next couple of days.  

In the meantime, our nightly builds are available at https://www.myget.org/F/odata/.

Please let us know right away if you run into any problems!

mprigge:
Fantastic. Thanks Mike!

PR #1030, opened 2018-01-09
modify the error text file for TT generator

xuzhg:
### Description

When we change the EdmxTarget to CsdlTarget, we forgot to rename the error message. That will be used to AutoTextGenerator. 

Without change, if we regenerate the error message file, we will have build error because the TT will generate "UnknownEnumVal_EdmxTarget".

- [] *Test cases added*
- [] *Build and test with one-click build and test script passed*


xuzhg:
Merged at: https://github.com/OData/odata.net/commit/684918dbf54ab62d9d37c8fdfae4495d6cf624dd

Thanks for review.

Issue #1028, opened 2018-01-08
EnumMember not supported in the Annotation


zmyvszk:
When I ran the validation library to validate my csdl. I got the following errors:
**ExpressionNotValidForTheAssertedType:The type of the expression is incompatible with the asserted type.**

When I looked into the code, I found that the code doesn't support the EmumMember in the Annotation.

Just want to verify this situation and hope it helps!

Thanks!

### Assemblies affected

6.x

xuzhg:
Merged at #1031. Thanks!

xuzhg:
Keep it open whether we should make sure to port it to maintenance-6.x branch.

AlanWong-MS:
Nope

PR #1031, opened 2018-01-09
Add the enum member into the expression check and don't return the Ex…

xuzhg:
### Issues

*This PR is related to https://github.com/OData/odata.net/issues/1028. But it's fixed in master branch. Once we have done it, we can consider to port to maintenance-6x branch.*

### Description

Originally, if the CSDL has annotation with 'EnumMember', it will output an error message:
```C#
ExpressionNotValidForTheAssertedType:The type of the expression is incompatible with the asserted type.
```
However, the `EnumMember` is valid expression type and should not return error message.

This PR is to fix it. 

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Validation/ExpressionTypeChecker.cs
+++ b/src/Microsoft.OData.Edm/Validation/ExpressionTypeChecker.cs
@@ -488,7 +488,7 @@ namespace Microsoft.OData.Edm.Validation
             return true;
         }
 
-        private static bool TryCastEnumConstantAsType(IEdmEnumMemberExpression expression, IEdmTypeReference type, bool matachExactly, out IEnumerable<EdmError> discoveredErrors)
+        private static bool TryCastEnumConstantAsType(IEdmEnumMemberExpression expression, IEdmTypeReference type, bool matchExactly, out IEnumerable<EdmError> discoveredErrors)
         {
             if (!type.IsEnum())
             {
@@ -503,7 +503,7 @@ namespace Microsoft.OData.Edm.Validation
 
             foreach (var member in expression.EnumMembers)
             {
-                if (!TestTypeMatch(member.DeclaringType, type.Definition, expression.Location(), matachExactly, out discoveredErrors))
+                if (!TestTypeMatch(member.DeclaringType, type.Definition, expression.Location(), matchExactly, out discoveredErrors))
                 {
                     return false;
                 }
--- a/src/Microsoft.OData.Edm/Validation/ExpressionTypeChecker.cs
+++ b/src/Microsoft.OData.Edm/Validation/ExpressionTypeChecker.cs
@@ -123,6 +123,8 @@ namespace Microsoft.OData.Edm.Validation
                     return TestTypeReferenceMatch(((IEdmCastExpression)expression).Type, type, expression.Location(), matchExactly, out discoveredErrors);
                 case EdmExpressionKind.LabeledExpressionReference:
                     return TryCast(((IEdmLabeledExpressionReferenceExpression)expression).ReferencedLabeledExpression, type, out discoveredErrors);
+                case EdmExpressionKind.EnumMember:
+                    return TryCastEnumConstantAsType((IEdmEnumMemberExpression)expression, type, matchExactly, out discoveredErrors);
                 default:
                     discoveredErrors = new EdmError[] { new EdmError(expression.Location(), EdmErrorCode.ExpressionNotValidForTheAssertedType, Edm.Strings.EdmModel_Validator_Semantic_ExpressionNotValidForTheAssertedType) };
                     return false;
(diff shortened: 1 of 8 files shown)


src/Microsoft.OData.Edm/Validation/ExpressionTypeChecker.cs
```diff
@@ -486,6 +488,31 @@ private static bool TryCastBinaryConstantAsType(IEdmBinaryConstantExpression exp
             return true;
         }
 
+        private static bool TryCastEnumConstantAsType(IEdmEnumMemberExpression expression, IEdmTypeReference type, bool matachExactly, out IEnumerable<EdmError> discoveredErrors)
```
AlanWong-MS: typo: bool matachExactly -> bool matchExactly

src/Microsoft.OData.Edm/Validation/ExpressionTypeChecker.cs
```diff
@@ -486,6 +488,31 @@ private static bool TryCastBinaryConstantAsType(IEdmBinaryConstantExpression exp
...
+                    new EdmError(expression.Location(),
+                    EdmErrorCode.ExpressionEnumKindNotValidForAssertedType,
+                    Edm.Strings.EdmModel_Validator_Semantic_ExpressionEnumKindNotValidForAssertedType)
+                };
+                return false;
+            }
+
+            foreach (var member in expression.EnumMembers)
+            {
+                if (!TestTypeMatch(member.DeclaringType, type.Definition, expression.Location(), matachExactly, out discoveredErrors))
```
AlanWong-MS: I see that the typo "matachExactly" came from somewhere else. I think it should be fixed there too.

test/FunctionalTests/Microsoft.OData.Edm.Tests/ScenarioTests/OasisRelationshipChangesAcceptanceTests.cs
```diff
@@ -463,6 +463,67 @@ public void ValidationShouldFailIfNavigationParterIsSpecifiedButCannotBeFound()
                 });
         }
 
+        [Fact]
+        public void ValidationShouldFailIfEnumMemberIsSpecifiedButCannotBeFound()
+        {
+            IEdmModel model = GetEdmModel(@"<EnumMember>TestNS2.UnknowColor/Blue</EnumMember>");
```
AlanWong-MS: typo: UnknowColor -> UnknownColor

test/FunctionalTests/Microsoft.OData.Edm.Tests/ScenarioTests/OasisRelationshipChangesAcceptanceTests.cs
```diff
@@ -463,6 +463,67 @@ public void ValidationShouldFailIfNavigationParterIsSpecifiedButCannotBeFound()
...
+            IEnumerable<EdmError> errors;
+            model.Validate(out errors).Should().BeFalse();
+            errors.Should().HaveCount(1);
+            errors.Should().Contain(e => e.ErrorCode == EdmErrorCode.BadUnresolvedEnumMember && e.ErrorMessage == ErrorStrings.Bad_UnresolvedEnumMember("Blue"));
+        }
+
+        [Fact]
+        public void ValidationShouldFailIfEnumMemberIsSpecifiedButCannotBeFoundTheMember()
+        {
+            IEdmModel model = GetEdmModel(@"<EnumMember>TestNS2.Color/Yellow</EnumMember>");
```
AlanWong-MS: Maybe update the color from Yellow to "UnknownMember" to be consistent with the previous test case?

src/Microsoft.OData.Edm/Validation/ExpressionTypeChecker.cs
xuzhg: #Resolved.

src/Microsoft.OData.Edm/Validation/ExpressionTypeChecker.cs
xuzhg: #resolved.

test/FunctionalTests/Microsoft.OData.Edm.Tests/ScenarioTests/OasisRelationshipChangesAcceptanceTests.cs
xuzhg: #resolved.

test/FunctionalTests/Microsoft.OData.Edm.Tests/ScenarioTests/OasisRelationshipChangesAcceptanceTests.cs
xuzhg: #resolved.

src/Microsoft.OData.Edm/Validation/ExpressionTypeChecker.cs
```diff
@@ -486,6 +488,31 @@ private static bool TryCastBinaryConstantAsType(IEdmBinaryConstantExpression exp
...
+                discoveredErrors = new EdmError[]
+                {
+                    new EdmError(expression.Location(),
+                    EdmErrorCode.ExpressionEnumKindNotValidForAssertedType,
+                    Edm.Strings.EdmModel_Validator_Semantic_ExpressionEnumKindNotValidForAssertedType)
+                };
+                return false;
+            }
+
+            foreach (var member in expression.EnumMembers)
```
mikepizzo: >expression [](start = 35, length = 10)

Is there any chance that expression may be null? Should we at least assert, if not check for null?

src/Microsoft.OData.Edm/Validation/ExpressionTypeChecker.cs
```diff
@@ -123,6 +123,8 @@ public static bool TryCast(this IEdmExpression expression, IEdmTypeReference typ
                     return TestTypeReferenceMatch(((IEdmCastExpression)expression).Type, type, expression.Location(), matchExactly, out discoveredErrors);
                 case EdmExpressionKind.LabeledExpressionReference:
                     return TryCast(((IEdmLabeledExpressionReferenceExpression)expression).ReferencedLabeledExpression, type, out discoveredErrors);
+                case EdmExpressionKind.EnumMember:
+                    return TryCastEnumConstantAsType((IEdmEnumMemberExpression)expression, type, matchExactly, out discoveredErrors);
```
mikepizzo: >expression [](start = 79, length = 10)

Is there any chance that type is an enum, but expression is not an IEdmEnumMemberExpression? If so, TryCastEnumConstantAsType will throw a null ref exception.

src/Microsoft.OData.Edm/Validation/ExpressionTypeChecker.cs
```diff
@@ -486,6 +488,31 @@ private static bool TryCastBinaryConstantAsType(IEdmBinaryConstantExpression exp
             return true;
         }
 
+        private static bool TryCastEnumConstantAsType(IEdmEnumMemberExpression expression, IEdmTypeReference type, bool matchExactly, out IEnumerable<EdmError> discoveredErrors)
+        {
+            if (!type.IsEnum())
+            {
+                discoveredErrors = new EdmError[]
+                {
+                    new EdmError(expression.Location(),
```
mikepizzo: >Location [](start = 44, length = 8)

Is there any chance that expression may be null, leading to a null ref exception here (and below)? We cast to an IEdmEnumMemberExpression when calling this function; if the expression is not an IEdmEnumMemberExpression this will be null.  Is there any chance, when calling this function, that the expression will not be an IEdmEnumMemberExpression? Should we at least assert not-null, if not explicitly check for/handle the null case?

Issue #1037, opened 2018-01-17
Support reading/writing OData 4.01 compatible JSON payloads


mikepizzo:
OASIS has defined a [4.01 version](http://docs.oasis-open.org/odata/odata-json-format/v4.01/csprd04/odata-json-format-v4.01-csprd04.html) of the OData JSON format with a number of simplifications. ODL should support reading/writing JSON format according to OData 4.01.

### Assemblies affected

ODL 7.4.

### Additional detail

In PR #989 we added support for writing the new 4.01 delta format by specifying ODataVersion.401 in WriterSettings.  In addition to supporting the new Delta format, we need to make sure that the generated JSON complies with additional 4.01 requirements.  Specifically, when 4.01 is specified, the writer should:
1) Omit the odata. prefix from control information in the JSON payload
2) Omit the # prefix from @type primitive values
3) Recognize metadata and streaming type parameters with and without the "odata" prefix

In Addition, the ODataReader should support reading valid 4.01 payloads, including:
1) Interpret control information with or without the odata. supports control information without the odata. prefix
2) Consume primitive value type annotations with or without the # prefix
3) Recognize metadata and streaming type parameters with and without the "odata" prefix

robward-ms:
Fixed in 7.4.0.

PR #989, opened 2017-11-06
Add support for 4.01 Delta Format

mikepizzo:
### Issues

*This pull request fixes issue #988.*

### Description

OData 4.01 adds support for a new delta format: http://docs.oasis-open.org/odata/odata-json-format/v4.01/csprd03/odata-json-format-v4.01-csprd03.html#sec_DeletedEntity.

The new format marks deleted entries using an @odata.removed annotation that contains the deleted reason, and nested delta resource sets using an @Delta annotation. In the new format, a deleted entry can have properties, and is identified either by @odata.id or by key properties.

This PR adds support for an OData 4.01 version, and uses that to control writing of 4.0 versus 4.01 format. Delta functionality is integrated into ODataReader and ODataWriter; the previous ODataDeltaReader and ODataDeltaWriter are retained as shims for backward compatibility, but marked as obsolete.  Replace CreateODataDeltaWriter with CreateODataDeltaResourceSetWriter and CreateODataDeltaReader with CreateODataDeltaResourceSetReader

### Checklist (Uncheck if it is not completed)

- [x] *Test cases added*
- [x] *Build and test with one-click build and test script passed*

### Additional work necessary

*Docs needed.*
*4.01 also adds support for deltas in request payloads.  This PR sets up deltas to be able to use in a request payload, but does not remove the current restrictions nor ensure critical annotations are written in order to support using in request payloads.*

PR #1038, opened 2018-01-17
401 format fixes

mikepizzo:
### Issues

*This pull request fixes issue #1037.*

### Description

In PR #989 we added support for writing the new 4.01 delta format by specifying ODataVersion.401 in WriterSettings. This PR supports additional requirements when writing, or reading, a 4.01 payload.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/MediaTypeUtils.cs
+++ b/src/Microsoft.OData.Core/MediaTypeUtils.cs
@@ -176,6 +176,21 @@ namespace Microsoft.OData
                 format = selectedMediaTypeWithFormat.Format;
                 mediaType = selectedMediaTypeWithFormat.MediaType;
 
+                // If any accept types in request used non-prefixed odata metadata or streaming type parameter,
+                // use the non-prefixed version in the response
+                if (specifiedTypes != null && mediaType.Parameters != null)
+                {
+                    if (specifiedTypes.Any(t => t.Key.Parameters != null && t.Key.Parameters.Any(p => String.Compare(p.Key, MimeConstants.MimeShortMetadataParameterName, StringComparison.OrdinalIgnoreCase) == 0)))
+                    {
+                        mediaType = new ODataMediaType(mediaType.Type, mediaType.SubType, mediaType.Parameters.Select(p => new KeyValuePair<string, string>(String.Compare(p.Key, MimeConstants.MimeMetadataParameterName, StringComparison.OrdinalIgnoreCase) == 0 ? MimeConstants.MimeShortMetadataParameterName : p.Key, p.Value)));
+                    }
+
+                    if (specifiedTypes.Any(t => t.Key.Parameters != null && t.Key.Parameters.Any(p => String.Compare(p.Key, MimeConstants.MimeShortStreamingParameterName, StringComparison.OrdinalIgnoreCase) == 0)))
+                    {
+                        mediaType = new ODataMediaType(mediaType.Type, mediaType.SubType, mediaType.Parameters.Select(p => new KeyValuePair<string, string>(String.Compare(p.Key, MimeConstants.MimeStreamingParameterName, StringComparison.OrdinalIgnoreCase) == 0 ? MimeConstants.MimeShortStreamingParameterName : p.Key, p.Value)));
+                    }
+                }
+
                 // If a charset was specified with the accept header, consider it for the encoding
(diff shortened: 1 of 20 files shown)


src/Microsoft.OData.Core/MediaTypeUtils.cs
```diff
@@ -547,13 +549,15 @@ private static void ConvertApplicationJsonInAcceptableMediaTypes(IList<KeyValueP
...
+                        !mediaType.Parameters.Any(p => HttpUtils.CompareMediaTypeParameterNames(p.Key, MimeConstants.MimeMetadataParameterName) || HttpUtils.CompareMediaTypeParameterNames(p.Key, MimeConstants.MimeShortMetadataParameterName)))
                     {
                         // application/json detected; convert it to Json Light
                         IList<KeyValuePair<string, string>> existingParams = mediaType.Parameters != null ? mediaType.Parameters.ToList() : null;
                         int newCount = existingParams == null ? 1 : existingParams.Count + 1;
                         List<KeyValuePair<string, string>> newParams = new List<KeyValuePair<string, string>>(newCount);
-                        newParams.Add(new KeyValuePair<string, string>(MimeConstants.MimeMetadataParameterName, MimeConstants.MimeMetadataParameterValueMinimal));
+                        newParams.Add(new KeyValuePair<string, string>(
+                            version < ODataVersion.V401 ? MimeConstants.MimeMetadataParameterName : MimeConstants.MimeShortMetadataParameterName,
```
xuzhg: > < [](start = 35, length = 2)

Sorry, would you please help double check the syntax is what you are expected.

src/Microsoft.OData.Core/ODataObjectModelExtensions.cs
```diff
@@ -34,6 +34,17 @@ public static void SetSerializationInfo(this ODataResourceBase resource, ODataRe
...
+        /// Provide additional serialization information to the <see cref="ODataWriter"/> for <paramref name="resource"/>.
+        /// </summary>
+        /// <param name="resource">The instance to set the serialization info.</param>
+        /// <param name="serializationInfo">The serialization info to set.</param>
+        public static void SetSerializationInfo(this ODataResource resource, ODataResourceSerializationInfo serializationInfo)
+        {
+            ExceptionUtils.CheckArgumentNotNull(resource, "resource");
+            resource.SerializationInfo = serializationInfo;
+        }
+
```
xuzhg: That's merged by other PR, right?

mikepizzo: Yes; I was working off a branch that already had that commit, but it looks like the merge works fine.

---
In reply to: [162159319](https://github.com/OData/odata.net/pull/1038#discussion_r162159319) [](ancestors = 162159319)

src/Microsoft.OData.Core/MediaTypeUtils.cs
mikepizzo: For ODataVersions prior to 4.01, the application/json metadata type parameter should be called "odata.metadata".  For 4.01 and beyond, the header should be called "metadata".  The default value remains "minimal".

Adding some test cases and simplifying the code a bit for comparing prefixed with non-prefixed odata metadata and odata streaming type parameters.

---
In reply to: [162159043](https://github.com/OData/odata.net/pull/1038#discussion_r162159043) [](ancestors = 162159043)

src/Microsoft.OData.Core/ODataSimplifiedOptions.cs
```diff
@@ -28,16 +28,15 @@ public ODataSimplifiedOptions(ODataVersion? version)
         {
             this.EnableParsingKeyAsSegmentUrl = true;
             this.EnableWritingKeyAsSegment = false;
+            this.EnableReadingKeyAsSegment = false;
```
biaol-odata: Question: this will change reader's EnableReadingKeyAsSegment default to false for V401 ? Just want to make sure this is intentional.

mikepizzo: I had mistakenly changed it to true in a previous PR to this release, thinking it allowed the reader to recognize key-as-segment.  In fact, it means that generated keys use key-as-segment, which should be off by default, so I fixed it here.  Thanks for checking!

PR #1042, opened 2018-01-24
Remove the NavigationPropertyEntityMustNotIndirectlyContainItself rule

xuzhg:
### Issues

* Issue from the Graph beta validation. It shows: 

Problem 3: NavigationPropertyEntityMustNotIndirectlyContainItself:The navigation property 'events' is invalid because it indirectly contains itself. 

### Description

When we validate the Graph beta CSDL, it gives us the following error messages:

```C#
The navigation property 'drive' is invalid because it indirectly contains itself.
The navigation property 'driveItem' is invalid because it indirectly contains itself.
The navigation property 'site' is invalid because it indirectly contains itself.
The navigation property 'worksheet' is invalid because it indirectly contains itself.
The navigation property 'worksheet' is invalid because it indirectly contains itself.
The navigation property 'charts' is invalid because it indirectly contains itself.
```
As discussed with Mike, it's better to remove the in-correct validation rule.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Validation/ValidationRuleSet.cs
+++ b/src/Microsoft.OData.Edm/Validation/ValidationRuleSet.cs
@@ -103,7 +103,6 @@ namespace Microsoft.OData.Edm.Validation
                 ValidationRules.AnnotationInaccessibleTerm,
                 ValidationRules.ElementDirectValueAnnotationFullNameMustBeUnique,
                 ValidationRules.VocabularyAnnotationInaccessibleTarget,
-                ValidationRules.NavigationPropertyEntityMustNotIndirectlyContainItself,
                 ValidationRules.EntitySetRecursiveNavigationPropertyMappingsMustPointBackToSourceEntitySet,
                 ValidationRules.NavigationPropertyMappingMustPointToValidTargetForProperty,
                 ValidationRules.DirectValueAnnotationHasXmlSerializableName,
--- a/test/FunctionalTests/Microsoft.OData.Edm.Tests/Validation/ValidationRuleSetTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Edm.Tests/Validation/ValidationRuleSetTests.cs
@@ -22,7 +22,8 @@ namespace Microsoft.OData.Edm.Tests.Validation
         public void NewValidationRulesShouldBeInTheRuleSetExceptBaselinedExceptionRules()
         {
             var validationRules = new Dictionary<object, string>();
-            var items = typeof(ValidationRules).GetFields().Select(f=> new KeyValuePair<object, string>(f.GetValue(null), f.Name));
+            var items = typeof(ValidationRules).GetFields().Where(f => f.Name != "NavigationPropertyEntityMustNotIndirectlyContainItself")
+                .Select(f=> new KeyValuePair<object, string>(f.GetValue(null), f.Name));
             foreach (var item in items)
             {
                 validationRules.Add(item.Key, item.Value);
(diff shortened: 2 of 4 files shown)


xuzhg:
@mikepizzo

test/FunctionalTests/Tests/DataOData/Tests/OData.Common.Tests/PublicApi/PublicApi.bsl
```diff
@@ -3115,7 +3114,6 @@ public sealed class Microsoft.OData.Edm.Validation.ValidationRules {
 	public static readonly Microsoft.OData.Edm.Validation.ValidationRule`1[[Microsoft.OData.Edm.IEdmNavigationProperty]] NavigationPropertyDependentPropertiesMustBelongToDependentEntity = Microsoft.OData.Edm.Validation.ValidationRule`1[Microsoft.OData.Edm.IEdmNavigationProperty]
 	public static readonly Microsoft.OData.Edm.Validation.ValidationRule`1[[Microsoft.OData.Edm.IEdmNavigationProperty]] NavigationPropertyDuplicateDependentProperty = Microsoft.OData.Edm.Validation.ValidationRule`1[Microsoft.OData.Edm.IEdmNavigationProperty]
 	public static readonly Microsoft.OData.Edm.Validation.ValidationRule`1[[Microsoft.OData.Edm.IEdmNavigationProperty]] NavigationPropertyEndWithManyMultiplicityCannotHaveOperationsSpecified = Microsoft.OData.Edm.Validation.ValidationRule`1[Microsoft.OData.Edm.IEdmNavigationProperty]
-	public static readonly Microsoft.OData.Edm.Validation.ValidationRule`1[[Microsoft.OData.Edm.IEdmNavigationProperty]] NavigationPropertyEntityMustNotIndirectlyContainItself = Microsoft.OData.Edm.Validation.ValidationRule`1[Microsoft.OData.Edm.IEdmNavigationProperty]
```
mikepizzo: Removing this is a breaking change that would break apps that attempt to access this property, no?  Should we still support the property and just return a dummy rule that always validates to true?  Or retain the rule but not include it as part of the default set of validations? Or modify the rule to enforce the semantics introduced in 4.0, namely: "If the containment is recursive, the partner navigation property MUST be nullable and specify a single entity type (i.e. have a cardinality of 0..1). If the containment is not recursive, the partner navigation property MUST NOT be nullable (i.e. have a cardinality of 1)."

src/Microsoft.OData.Edm/Validation/EdmErrorCode.cs
```diff
@@ -579,10 +579,10 @@ public enum EdmErrorCode
         /// </summary>
         InvalidPolymorphicComplexType = 221,
 
-        /// <summary>
-        /// A navigation property without direct containment cannot contain its declaring entity indirectly.
-        /// </summary>
-        NavigationPropertyEntityMustNotIndirectlyContainItself = 222,
+        // /// <summary>
```
mikepizzo: Is this commented, rather than removed, in order to make sure we don't re-use the value 222? Should we add a comment to that effect?

xuzhg: Yes. removing it is a breaking change. I will keep it unchanged.

test/FunctionalTests/Tests/DataOData/Tests/OData.Common.Tests/PublicApi/PublicApi.bsl
xuzhg: Thanks for pointing out. I am prefer to remove the validate rule from the default set and keep the rule unchanged.

xuzhg:
@mikepizzo I changed the codes. Please take a look.

Issue #1043, opened 2018-01-24
Thread safety of context item collections


chrisbigart:
We have code that calls .First() on a collection in a Microsoft.Dynamics.CRM.System context that is deadlocked because of the lack of thread-safety of the Dictionary<> class.  A process dump shows that multiple threads are in the EntityTracker.InternalAttachEntityDescriptor() method, calling the .Add()/.Insert() method of the 'identityToDescriptor' Dictionary<>.

The calling code looks like:

```
public static string GetContactById(Microsoft.Dynamics.CRM.System _context, Guid id)
        {
            var contact = _context.Contacts.Expand(a => a.Parentcustomerid_account).Where(a => a.Contactid == id);
            var cContact = ConvertContact(_context, contact.First());
            return Util.Serialize<Business.Contact>(cContact);
        }
```

It appears that the fix would be to change each Dictionary<> to a ConcurrentDictionary<> for the EntityTracker internal members: 'entityDescriptors', 'identityToDescriptor', and 'bindings'.

If there is some other work-around, please let me know.  Thanks.

chrisbigart:
The call stack looks like:
```
 >mscorlib.dll!System.Collections.Generic.Dictionary<object, Microsoft.OData.Client.EntityDescriptor>.Insert(object key, Microsoft.OData.Client.EntityDescriptor value, bool add)
 
>Microsoft.OData.Client.dll!Microsoft.OData.Client.EntityTracker.InternalAttachEntityDescriptor(Microsoft.OData.Client.EntityDescriptor entityDescriptorFromMaterializer, bool failIfDuplicated) Line 286	C#	

 >Microsoft.OData.Client.dll!Microsoft.OData.Client.AtomMaterializerLog.ApplyToContext() Line 68	C#	

 >Microsoft.OData.Client.dll!Microsoft.OData.Client.MaterializeAtom.MoveNextInternal() Line 280	C#	

 >Microsoft.OData.Client.dll!Microsoft.OData.Client.MaterializeAtom.MoveNext() Line 217	C#	

 >System.Core.dll!System.Linq.Enumerable.<CastIterator>d__b1<2016WebApiService.Microsoft.Dynamics.CRM.Contact>.MoveNext()	
 	>System.Core.dll!System.Linq.Enumerable.First<2016WebApiService.Microsoft.Dynamics.CRM.Contact>(System.Collections.Generic.IEnumerable<2016WebApiService.Microsoft.Dynamics.CRM.Contact> source)	

[...]

PR #1054, opened 2018-01-29
Use thread-safe ConcurrentDictionary instead of Dictionary for internal EntityTracker objects.

chrisbigart:
### Issues

*This pull request fixes issue #1043.*

### Description

*Replaces EntityTracker's internal Dictionary<> objects with thread-safe ConcurrentDictionary<> ones.*

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Client/DictionaryExtensions.cs
+++ b/src/Microsoft.OData.Client/DictionaryExtensions.cs
@@ -8,11 +8,12 @@ namespace Microsoft.OData.Client
 {
     using System;
     using System.Collections.Generic;
-    using System.Diagnostics;
+    using System.Collections.Concurrent;
+	using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
 
     /// <summary>
-    /// Useful extension methods for IDictionary
+    /// Useful extension methods for IDictionary and ConcurrentDictionary
     /// </summary>
     internal static class DictionaryExtensions
     {
@@ -54,6 +55,22 @@ namespace Microsoft.OData.Client
             }
         }
 
+		/// <summary>
+		/// Convenience funcion that wraps ConcurrentDictionary.TryAdd() to allow same signature as IDictionary
+		/// </summary>
+		public static bool Add<TKey, TValue>(this ConcurrentDictionary<TKey, TValue> self, TKey key, TValue value)
+		{
+			return self.TryAdd(key, value);
+		}
+
+		/// <summary>
+		/// Convenience function for ConcurrentDictionary to allow same signature as IDictionary
+		/// </summary>
+		public static bool Remove<TKey, TValue>(this ConcurrentDictionary<TKey, TValue> self, TKey key)
+		{
+			return ((IDictionary<TKey, TValue>)self).Remove(key);
+		}
+
 #if PORTABLELIB
         /// <summary>
         /// Try to add a key/value pair to the dictionary and returns true if the key was added, false if it was already present.
@@ -86,5 +103,5 @@ namespace Microsoft.OData.Client
             return false;
         }
 #endif
-    }
+	}
 }
--- a/src/Microsoft.OData.Client/DictionaryExtensions.cs
+++ b/src/Microsoft.OData.Client/DictionaryExtensions.cs
@@ -11,7 +11,7 @@ namespace Microsoft.OData.Client
     using System.Collections.Concurrent;
     using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
-
+    
     /// <summary>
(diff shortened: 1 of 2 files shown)


biaol-odata:
@chrisbigart It looks harmless to change to use ConcurrentDictionary. But if you run the build.cmd you will see test error in StateManagementTests. When trying to add same object, an exception is expected but it is happening anymore. Can you please investigate more?

chrisbigart:
I think this is because of an inconsistency in the Add() method between IDictionary and ConcurrentDictionary.  IDictionary.Add() signature:

> 
> ```
> public void Add(
> 	TKey key,
> 	TValue value
> )
> ```
> 
> **ArgumentException** | _An element with the same key already exists in the Dictionary<TKey, TValue>._
> -- | --
> 

(Which should return a bool instead of throwing an exception in my opinion, but I digress...)

ConcurrentDictionary.TryAdd() is (in my opinion) more sensible:

> ```
> public bool TryAdd(
> 	TKey key,
> 	TValue value
> )
> ```
> **Return Value**
> Type: System.Boolean
> _true if the key/value pair was added to the System.Collections.Concurrent.ConcurrentDictionary<TKey, TValue> successfully; false if the key already exists._

I guess the best way to handle this is to explicitly throw System.ArgumentException if ConcurrentDictionary.TryAdd() returns false.

src/Microsoft.OData.Client/DictionaryExtensions.cs
```diff
@@ -8,11 +8,12 @@ namespace Microsoft.OData.Client
 {
     using System;
     using System.Collections.Generic;
+    using System.Collections.Concurrent;
     using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
-
+
```
biaol-odata: nit: remove empty line.

src/Microsoft.OData.Client/DictionaryExtensions.cs
```diff
@@ -54,6 +55,23 @@ internal static class DictionaryExtensions
             }
         }
 
+        /// <summary>
+        /// Convenience funcion that wraps ConcurrentDictionary.TryAdd() to allow same signature as IDictionary
```
biaol-odata: nit: typo "funcion"

src/Microsoft.OData.Client/DictionaryExtensions.cs
```diff
@@ -54,6 +55,23 @@ internal static class DictionaryExtensions
...
+        /// </summary>
+        public static void Add<TKey, TValue>(this ConcurrentDictionary<TKey, TValue> self, TKey key, TValue value)
+        {
+            if(!self.TryAdd(key, value))
+                throw new ArgumentException("Argument_AddingDuplicate");
+        }
+
+        /// <summary>
+        /// Convenience function for ConcurrentDictionary to allow same signature as IDictionary
```
biaol-odata: I am wondering why we need this wrapper, since IDictinoary<TKey, TValue>.Remove(TKey) already exists?

biaol-odata:
@chrisbigart LGTM, with some additional comments for small updates. Thanks!

biaol-odata:
@chrisbigart Ping for followups. Thanks.

AlanWong-MS:
@chrisbigart, thank you for the contribution and work thus far. We strive to provide quick support for PRs, such as yours, and merge them to the source code where appropriate. We noticed that there hasn't been activity on your PR in the last 30 days and because of this, we are considering moving it to a closed state until you are have the opportunity to continue adding to the PR. Doing so allows us to devote the same level of attentiveness to other contributors’ PRs. Within the next two weeks, we'd like to know where you are with the contribution before we decide to park it in "Closed". Thank you for your understanding!

src/Microsoft.OData.Client/DictionaryExtensions.cs
mikepizzo: Because Remove is an explicitly implemented member of IDictionary<TKey,TValue>; you have to cast to IDictionary<TKey,TValue> to call it.  This helper method avoids having to cast where Remove is called.

---
In reply to: [166143263](https://github.com/OData/odata.net/pull/1054#discussion_r166143263) [](ancestors = 166143263)

src/Microsoft.OData.Client/DictionaryExtensions.cs
mikepizzo: removed spaces

---
In reply to: [166142363](https://github.com/OData/odata.net/pull/1054#discussion_r166142363) [](ancestors = 166142363)

mikepizzo:
Replaced by PR #1199, which addresses comments in original PR.

PR #1056, opened 2018-01-29
Add KeyAsSegmentSupported annotation term to Capabiliites vocabulary

xuzhg:
### Issues

See issue at: https://issues.oasis-open.org/browse/ODATA-1134

### Description

In OData 4.01 we introduced semantics around supporting the popular key-as-segment URL syntax, but we have no generic way for clients to know whether or not the service supports this sytnax. 

This PR is to add a new boolean term to the capabilities vocabulary, `KeyAsSegmentSupported`

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Vocabularies/CapabilitiesVocabularies.xml
+++ b/src/Microsoft.OData.Edm/Vocabularies/CapabilitiesVocabularies.xml
@@ -242,17 +242,18 @@ Overview:
         <Annotation Term="Core.Description" String="Supports key values according to OData URL conventions" />
       </Term>
 
+      <Term Name="KeyAsSegmentSupported" Type="Core.Tag" DefaultValue="true" AppliesTo="EntityContainer">
+        <Annotation Term="Core.Description" String="Supports key as segment" />
+      </Term>
       <Term Name="TopSupported" Type="Core.Tag" DefaultValue="true" AppliesTo="EntitySet">
         <Annotation Term="Core.Description" String="Supports $top" />
       </Term>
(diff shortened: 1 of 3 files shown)


mikepizzo:
CapabilitiesVocabularyModel.cs only defines IEdmTerms for ChangeTrackingTerm (where-as CoreVocabularyModel.cs defines IEdmTerms for most/all of the terms defined in the vocabulary).  Should we define a term for KeyAsSegmentTerm (and the other capabilities terms) or do we not think there is value in defining these IEdmTerms?

xuzhg:
@mikepizzo

Actually, we doesn't need to do that . 

1. We designed `CapabilitiesVocabularyModel` as an internal classes. Define `KeyAsSegmentTerm` can't help customer.

2. We don't use it in our product codes. The customers should call "FindTerm("Org.OData...KeyAsSegmentSupported") to get the term. 

3.  We planned to remove all the term variables defined `CoreVocabularyModel.cs`, But, it's a breaking change because this class is public class.

Any thoughts?

mikepizzo:
That's fine with me; I just wanted to be sure we had a good reason for the difference.  

It appears that we do use ChangeTrackingTerm from the capabilities vocabulary, and presumably get a performance improvement for having it a static value and not having to look it up each time we use it. 

Maybe consider adding a code comment to CapabilitiesVocabularyModel.cs calling out why we define an IEdmTerm for Changetracking but intentionally don't define IEdmTerm values for the other terms.

---
In reply to: [361427705](https://github.com/OData/odata.net/pull/1056#issuecomment-361427705) [](ancestors = 361427705)

PR #1051, opened 2018-01-26
Add the build-in abstract type into Edm core model - Edm Type Part.

xuzhg:
### Issues

*See Issue: https://github.com/OData/odata.net/issues/1027

### Description

OData spec says at: http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part3-csdl/odata-v4.0-errata03-os-part3-csdl-complete.html#_Built-In_Abstract_Types

The following built-in abstract types can be used within a model:
·         Edm.PrimitiveType 
·         Edm.ComplexType
·         Edm.EntityType

Vocabulary terms can, in addition, use
·         Edm.AnnotationPath
·         Edm.PropertyPath
·         Edm.NavigationPropertyPath 

However, the current implementation doesn't include this part. It will make the model validation failed. Show as https://github.com/OData/odata.net/issues/1027

### Design & Implementation

1.  Move the core model related files into CoreModel subfolder.
2.  Add `Edm.ComplexType` using `IEdmComplexType` into Core model.
3.  Add `Edm.EntityType` using `IEdmEntityType` into Core model.
4.  Add `Edm.PrimitiveType` using `IEdmPrimitiveType` into Core model.
5. Add "Path" into EdmTypeKind.
6. Add `IEdmPathType` & `IEdmPathTypeReference`.
7. Add `Edm.AnnotationPath, Edm.PropertyPath, Edm.NavigationPropertyPath` using `IEdmPathType` into Core model.

### Usage 

For example, to use "Edm.EntityType" to define the navigation property.

```C#
EdmModel model = new EdmModel();
EdmEntityType customer = new EdmEntityType("NS", "Customer");
customer.AddKeys(customer.AddStructuralProperty("Id", EdmCoreModel.Instance.GetInt32(false)));
customer.AddStructuralProperty("Complex", EdmCoreModel.Instance.GetComplexType(true));
customer.AddUnidirectionalNavigation(new EdmNavigationPropertyInfo
{
       Target = EdmCoreModel.Instance.GetEntityType(true).EntityDefinition(),
       ContainsTarget = false,
       TargetMultiplicity = EdmMultiplicity.ZeroOrOne,
       Name = "Properties"
});
model.AddElement(customer);
```
The corresponding CSDL:
```XML
<?xml version="1.0" encoding="utf-16"?>
<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
  <edmx:DataServices>
[...]

Issue #1027, opened 2018-01-08
Edm.NavigationPropertyPath not supported


zmyvszk:
The type Edm.NavigationPropertyPath is not exist in the Capacibilities.xml, for example:

<Property Name="NonCountableNavigationProperties" Type="Collection(Edm.NavigationPropertyPath)">
   <Annotation Term="Core.Description" String="These navigation properties do not allow /$count segments" />
 </Property>
</ComplexType> 

In this case, when I validate my csdl, I will get the errors like:
BadUnresolvedType:The type 'Edm.NavigationPropertyPath' could not be found.
BadUnresolvedType:The type 'Edm.PropertyPath' could not be found.
BadUnresolvedType:The type 'Edm.PropertyPath' could not be found.
BadUnresolvedType:The type 'Edm.PropertyPath' could not be found.
BadUnresolvedType:The type 'Edm.PropertyPath' could not be found.
BadUnresolvedType:The type 'Edm.NavigationPropertyPath' could not be found.
BadUnresolvedType:The type 'Edm.NavigationPropertyPath' could not be found.
BadUnresolvedType:The type 'Edm.PropertyPath' could not be found.
BadUnresolvedType:The type 'Edm.NavigationPropertyPath' could not be found.
BadUnresolvedType:The type 'Edm.NavigationPropertyPath' could not be found.
BadUnresolvedType:The type 'Edm.NavigationPropertyPath' could not be found.
BadUnresolvedType:The type 'Edm.PropertyPath' could not be found.
BadUnresolvedType:The type 'Edm.PropertyPath' could not be found.
BadUnresolvedType:The type 'Edm.NavigationPropertyPath' could not be found.

This issue also happened to PropertyPath.

### Assemblies affected

6.x

PR #1058, opened 2018-01-31
Add the validation rules to the abstract types

xuzhg:
### Issues

*See Issue #1027 & PR #1051

### Description

Add the validation rules for the abstract types

1. Edm.EntityType 
* cannot be used as the type of a singleton in an entity container because it doesn’t define a structure, which defeats the purpose of a singleton.
* cannot be used as the type of an entity set because all entities in an entity set must have the same key fields to uniquely identify them within the set.
* cannot be the base type of an entity type or complex type. 

2. Edm.ComplexType 
* cannot be the base type of an entity type or complex type. 

3. Edm.PrimitiveType 
* cannot be used as the type of a key property of an entity type.
* cannot be used as the underlying type of a type definition or enumeration type.
4. Collection(Edm.PrimitiveType) and Collection(Edm.ComplexType)
* cannot be used as the type of a property.
* cannot be used as the return type of a function.

- [x] *Test cases added*

### Additional work necessary


test/FunctionalTests/Tests/DataOData/Tests/OData.Common.Tests/PublicApi/PublicApi.bsl
```diff
@@ -536,10 +536,10 @@ public enum Microsoft.OData.Edm.EdmOnDeleteAction : int {
 }
 
 public enum Microsoft.OData.Edm.EdmPathTypeKind : int {
-	Annotation = 1
```
robward-ms: This seems like a breaking change.

xuzhg: `EdmPathTypeKind` is added in my previous PR. https://github.com/OData/odata.net/pull/1051 . It's not in public official release. 

It's a breaking change for my previous PR, not a breaking change for the office release. :)

Thanks for catch this.

#Won't fix.

mikepizzo:
The only rule that I didn't see was that an entity type used in an entity set, singleton, or nav prop must not use any of the path types (directly, or through complex-typed properties).

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
```diff
@@ -39,6 +39,17 @@ public static bool IsEntity(this IEdmTypeReference type)
             return type.TypeKind() == EdmTypeKind.Entity;
         }
 
+        /// <summary>
+        /// Returns true if this reference refers to a path type.
+        /// </summary>
+        /// <param name="type">Type reference.</param>
+        /// <returns>This reference refers to a pah type.</returns>
```
biaol-odata: typo.

src/Microsoft.OData.Edm/Parameterized.Microsoft.OData.Edm.cs
```diff
@@ -642,6 +642,20 @@ internal static class Strings {
...
+        /// A string like "The 'Edm.PrimitiveType' cannot be used as the type of a key property '{0}' of an entity type '{1}'."
+        /// </summary>
+        internal static string EdmModel_Validator_Semantic_EdmPrimitiveTypeCannotBeUsedAsTypeOfKey(object p0, object p1) {
+            return Microsoft.OData.Edm.EntityRes.GetString(Microsoft.OData.Edm.EntityRes.EdmModel_Validator_Semantic_EdmPrimitiveTypeCannotBeUsedAsTypeOfKey, p0, p1);
+        }
+
+        /// <summary>
+        /// A string like "The 'Edm.PrimitiveType' cannot be used as the underlying type of '{0}' type '{1}'."
+        /// </summary>
+        internal static string EdmModel_Validator_Semantic_EdmPrimitiveTypeCannotBeUsedAsUnderlyingType(object p0, object p1) {
```
biaol-odata: looks like p0 is always "type definition" only one argument typeDefinition.FullName() is needed?
Please check the chained definitions as well.

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
xuzhg: #Resolved.

src/Microsoft.OData.Edm/Parameterized.Microsoft.OData.Edm.cs
xuzhg: It can be used as "enumeration type" as well. 
The rule is:
*The 'Edm.PrimitiveType' cannot be used as the underlying type of type definition or enumeration type. *
Owing that the

xuzhg:
@mikepizzo I added the new rules for the path type property. Please take a look, thanks.

mikepizzo:
Do we have a general test that a (nullable) property of a complex type whose type is the same complex type (i.e., <ComplexType Name= "foo"><Property Name="nested" Type="ns.foo" Nullable="true"/></ComplexType>), an entity containing such a recursive complex type, and a singleton/entity set/nav prop whose type is an entity containing such a recursive type,  doesn't cause infinite recursion in validation? I see you handle it in code, but want to make sure we have a general test for it.

xuzhg:
@mikepizzo I added the new rules for the recursive type with path type property. Please take a look, thanks

src/Microsoft.OData.Edm/Validation/EdmErrorCode.cs
```diff
@@ -1309,5 +1322,40 @@ public enum EdmErrorCode
...
+        /// </summary>
+        EntityTypeOfSingletonCannotBeEdmEntityType = 384,
+
+        /// <summary>
+        /// Edm.EntityType cannot be used as the type of an entity set in an entity container.
+        /// </summary>
+        EntityTypeOfEntitySetCannotBeEdmEntityType = 385,
+
+        /// <summary>
+        /// The dclaring type of navigation source cannot have path type property.
```
AlanWong-MS: typo: dclaring -> declaring

PR #1063, opened 2018-02-02
Support annotations in $Select

mikepizzo:
### Description

-Supports annotations in $select only; does not support annotations in expressions used by $orderby or $filter.
-Annotations must be prefixed with "@" and namespace-qualified.
-Selecting control information as an odata annotation is explicitly not allowed.

- [x] *Test cases added*


src/Microsoft.OData.Core/UriParser/SemanticAst/AnnotationSegment.cs
```diff
@@ -0,0 +1,95 @@
...
+
+    using Edm.Vocabularies;
+    using Microsoft.OData.Edm;
+
+    #endregion Namespaces
+
+    /// <summary>
+    /// A segment representing an Annotation
+    /// </summary>
+    public sealed class AnnotationSegment : ODataPathSegment
```
xuzhg: >AnnotationSegment [](start = 24, length = 17)

Maybe it's better to name it as "TermSegment" ?

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -127,6 +128,33 @@ public virtual IEdmProperty ResolveProperty(IEdmStructuredType type, string prop
...
+        /// Resolve term name from model.
+        /// </summary>
+        /// <param name="model">The model to be used.</param>
+        /// <param name="termName">The term name to be resolved.</param>
+        /// <returns>Resolved term.</returns>
+        public virtual IEdmTerm ResolveTerm(IEdmModel model, string termName)
+        {
+            if (EnableCaseInsensitive)
+            {
+                var result = model.SchemaElements.OfType<IEdmTerm>()
```
xuzhg: >model.SchemaElements [](start = 29, length = 20)

how about if the input term is in CoreModel?

I am wondering "SchemaElements" includes the Core annotation and capabilities annotation.

src/Microsoft.OData.Core/UriParser/Parsers/UriParserHelper.cs
```diff
@@ -261,6 +261,16 @@ internal static IEdmTypeReference GetLiteralEdmTypeReference(ExpressionTokenKind
...
         }
 
+        /// <summary>
+        /// Determines whether or not an identifier is an annotation term name
+        /// </summary>
+        /// <param name="identifier">The identifer that may be an annotation term name</param>
+        /// <returns>True if the identifier is an annotation term, otherwise false</returns>
+        internal static bool IsAnnotation(string identifier)
+        {
+            return !string.IsNullOrEmpty(identifier) && identifier[0] == '@' && identifier.Contains(".");
```
xuzhg: >'@' [](start = 73, length = 3)

Maybe use the existing constants defined in "JsonLightConstants"

src/Microsoft.OData.Core/ODataConstants.cs
```diff
@@ -144,6 +144,9 @@ internal static class ODataInternalConstants
         /// <summary>A segment name in a URI that indicates metadata is being requested.</summary>
         internal const string UriMetadataSegment = "$metadata";
 
+        /// <summary>The OData prefix</summary>
+        internal const string ODataPrefix = "odata";
```
xuzhg: >odata [](start = 45, length = 5)

JsonLightConstants defines "ODataAnnotationNamespacePrefix = "odata.";

Can we re-use it?

src/Microsoft.OData.Core/UriParser/Visitors/PathSegmentToContextUrlPathTranslator.cs
```diff
@@ -126,6 +126,17 @@ public override string Translate(PropertySegment segment)
...
 
+        /// <summary>
+        /// Translate an AnnotationSegment
+        /// </summary>
+        /// <param name="segment">the segment to Translate</param>
+        /// <returns>Defined by the implementer.</returns>
+        public override string Translate(AnnotationSegment segment)
+        {
+            Debug.Assert(segment != null, "segment != null");
+            return "/" + segment.Term.Name;
```
xuzhg: >Name [](start = 38, length = 4)

Maybe FullName ?

src/Microsoft.OData.Core/UriParser/Visitors/PathSegmentToResourcePathTranslator.cs
```diff
@@ -115,6 +115,17 @@ public override string Translate(PropertySegment segment)
...
 
+        /// <summary>
+        /// Translate an AnnotationSegment
+        /// </summary>
+        /// <param name="segment">the segment to Translate</param>
+        /// <returns>Defined by the implementer.</returns>
+        public override string Translate(AnnotationSegment segment)
+        {
+            Debug.Assert(segment != null, "segment != null");
+            return "/" + segment.Term.Name;
```
xuzhg: >Name [](start = 38, length = 4)

Maybe Full Name

src/Microsoft.OData.Core/UriParser/Visitors/PathSegmentToStringTranslator.cs
```diff
@@ -61,6 +61,17 @@ public override string Translate(PropertySegment segment)
...
 
+        /// <summary>
+        /// Translate an AnnotationSegment
+        /// </summary>
+        /// <param name="segment">the segment to Translate</param>
+        /// <returns>Defined by the implementer.</returns>
+        public override string Translate(AnnotationSegment segment)
+        {
+            Debug.Assert(segment != null, "segment != null");
+            return segment.Term.Name;
```
xuzhg: >Name [](start = 32, length = 4)

maybe full name

src/Microsoft.OData.Core/UriParser/Visitors/SelectPropertyVisitor.cs
```diff
@@ -75,7 +75,7 @@ public SelectExpandClause DecoratedExpandClause
         public override void Visit(SystemToken tokenIn)
         {
             ExceptionUtils.CheckArgumentNotNull(tokenIn, "tokenIn");
-            throw new ODataException(ODataErrorStrings.SelectPropertyVisitor_SystemTokenInSelect(tokenIn.Identifier));
+            throw new ODataException(ODataErrorStrings.UriSelectParser_TermIsNotValid(tokenIn.Identifier));
```
xuzhg: >UriSelectParser_TermIsNotValid [](start = 55, length = 30)

Is it a breaking change?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
```diff
@@ -397,6 +398,100 @@ public void ParseQueryOptionsShouldWork(string relativeUriString, bool enableNoD
...
+        }
+
+        [Fact]
+        public void ParseSelectUnknownODataControlInformationShouldFail()
+        {
+            string relativeUriString = "People?$select=@odata.unknown";
+            var parser = new ODataUriParser(HardCodedTestModel.TestModel, new Uri(relativeUriString, UriKind.Relative));
+            Action action = () => parser.ParseSelectAndExpand();
+            action.ShouldThrow<ODataException>().WithMessage(ODataErrorStrings.UriSelectParser_TermIsNotValid("@odata.unknown"));
+        }
```
xuzhg: Maybe merge them as one test using [Theory].

src/Microsoft.OData.Core/UriParser/SemanticAst/AnnotationSegment.cs
mikepizzo: Well, the term is what is defined in metadata, the annotation is the application of the term.  In this case, what is being projected is not the term definition, but the application of the term to the data (i.e., the value of the term, rather than the definition.

---
In reply to: [166081172](https://github.com/OData/odata.net/pull/1063#discussion_r166081172) [](ancestors = 166081172)

src/Microsoft.OData.Core/UriParser/Parsers/UriParserHelper.cs
mikepizzo: I'd rather not have the parser reference the jsonlight constants, but I can define '@' in someplace like ODataConstants (in fact, I had done that previously but shelved it as part of the orderby work).

---
In reply to: [166082602](https://github.com/OData/odata.net/pull/1063#discussion_r166082602) [](ancestors = 166082602)

src/Microsoft.OData.Core/ODataConstants.cs
mikepizzo: I would rather not have uriparser reference jsonlightconstants. Perhaps we should just move it.

---
In reply to: [166082858](https://github.com/OData/odata.net/pull/1063#discussion_r166082858) [](ancestors = 166082858)

src/Microsoft.OData.Core/UriParser/Visitors/SelectPropertyVisitor.cs
mikepizzo: Actually, I think I updated the exception in the wrong place.

---
In reply to: [166084054](https://github.com/OData/odata.net/pull/1063#discussion_r166084054) [](ancestors = 166084054)

src/Microsoft.OData.Core/UriParser/Visitors/PathSegmentToContextUrlPathTranslator.cs
mikepizzo: Good catch; thanks.

---
In reply to: [166083770](https://github.com/OData/odata.net/pull/1063#discussion_r166083770) [](ancestors = 166083770)

src/Microsoft.OData.Core/UriParser/Visitors/PathSegmentToResourcePathTranslator.cs
mikepizzo: Good catch; thanks.

---
In reply to: [166083905](https://github.com/OData/odata.net/pull/1063#discussion_r166083905) [](ancestors = 166083905)

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
mikepizzo: I don't think capabilities would ever be projected; will have to think about Core (maybe permissions?)

---
In reply to: [166082192](https://github.com/OData/odata.net/pull/1063#discussion_r166082192) [](ancestors = 166082192)

src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
```diff
@@ -680,7 +680,7 @@ protected virtual ExpressionToken NextTokenImplementation(out Exception error)
                         }
 
                         this.ParseIdentifier();
-                        t = ExpressionTokenKind.ParameterAlias;
+                        t = this.parsingFunctionParameters && !(this.ExpressionText[0] == '@' && this.ExpressionText.Contains(".")) ? ExpressionTokenKind.ParameterAlias : ExpressionTokenKind.Identifier;
```
robward-ms: nit: multiline would be easier to read.

src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
```diff
@@ -1195,7 +1195,7 @@ private void AdvanceThroughBalancedExpression(char startingCharacter, char endin
         /// <summary>Parses an identifier by advancing the current character.</summary>
         private void ParseIdentifier()
         {
-            Debug.Assert(this.IsValidStartingCharForIdentifier, "Expected valid starting char for identifier");
+            Debug.Assert(this.IsValidStartingCharForIdentifier || this.ch == '@', "Expected valid starting char for identifier");
```
robward-ms: Maybe provide a function to check for '@' instead of hard-coding '@' in a bunch of places? Here and ExpressionLexer.cs.. I say function because the intent of the check would be expressed in the function name, making the conditional more readable.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -127,6 +128,33 @@ public virtual IEdmProperty ResolveProperty(IEdmStructuredType type, string prop
...
+               .ToList();
+
+                if (result.Count == 1)
+                {
+                    return result.Single();
+                }
+                else if (result.Count > 1)
+                {
+                    throw new ODataException(Strings.UriParserMetadata_MultipleMatchingTypesFound(termName));
+                }
```
biaol-odata: should we add returning with null here if result.Count==0?

src/Microsoft.OData.Core/UriParser/Visitors/IsCollectionTranslator.cs
```diff
@@ -62,6 +62,18 @@ public override bool Translate(PropertySegment segment)
             return false;
         }
 
+        /// <summary>
+        /// Translate an AnnotationSegment
```
biaol-odata: <code>AnnotationSegement</code>?

src/Microsoft.OData.Core/UriParser/Parsers/UriParserHelper.cs
mikepizzo: Defined a constant in UriQueryConstants for the annotation prefix.

---
In reply to: [166082602](https://github.com/OData/odata.net/pull/1063#discussion_r166082602) [](ancestors = 166082602)

src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
mikepizzo: indeed.  fixed.

---
In reply to: [167023378](https://github.com/OData/odata.net/pull/1063#discussion_r167023378) [](ancestors = 167023378)

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
mikepizzo: simplified logic in latest pr

---
In reply to: [167038073](https://github.com/OData/odata.net/pull/1063#discussion_r167038073) [](ancestors = 167038073)

src/Microsoft.OData.Core/UriParser/Visitors/IsCollectionTranslator.cs
mikepizzo: Internal class, so not really worth applying change here and (for consistency) in all the other methods.

---
In reply to: [167040541](https://github.com/OData/odata.net/pull/1063#discussion_r167040541) [](ancestors = 167040541)

src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
mikepizzo: Note that, in some cases we are comparing to '@' to determine if it's an annotation; in other cases we're trying to determine if it's a parameter alias.  I have an "IsAnnotation" method which also verifies that it contains a ".", but that can't be used here because we only have the single char.  I have updated all cases where we are comparing a character to "@" to use the new AnnotationPrefix constant, which provides context as to what we are checking for.

---
In reply to: [167023686](https://github.com/OData/odata.net/pull/1063#discussion_r167023686) [](ancestors = 167023686)

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
mikepizzo: Added code to traverse referenced schemas, which would include core, etc.

---
In reply to: [166162587](https://github.com/OData/odata.net/pull/1063#discussion_r166162587) [](ancestors = 166162587,166082192)

PR #1080, opened 2018-02-15
Add nuget package testing.

robward-ms:
### Description

This intention is to provide a small test project that can ingest the latest NuPkg's
from the build, restore them, build the project which results in a testable Dll. These
tests are very limited in scope, i.e. they only test the Dlls ability to load and
perform basic functions; the functional tests will test the full functionality.

- [x] *Test cases added*

### Additional work necessary

An additional build task will need to be added to update/compile/test this code.


test/PackageTests/Microsoft.OData.Tests.Nuget.Client/Microsoft.OData.Tests.Nuget.Client.csproj
```diff
@@ -0,0 +1,110 @@
...
+    <PlatformTarget>AnyCPU</PlatformTarget>
+    <DebugType>pdbonly</DebugType>
+    <Optimize>true</Optimize>
+    <OutputPath>bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="Microsoft.OData.Client, Version=7.4.0.20111, Culture=neutral, PublicKeyToken=31bf3856ad364e35">
```
AlanWong-MS: CR note: the hardcode here will be updated during automation

robward-ms: Correct. However, it makes sense to checkin a lower version, say 7.0.0. Automation will update it but never checkin the update. I'll make a change here...

AlanWong-MS: Agreed, that would be cleaner.

robward-ms: Changed. Core, Edm and Spatial point to 7.0.0, Client points to 7.4.0 since that is the earliest 7.x RTM version of the client.

PR #1082, opened 2018-02-15
Merge branch 'master' into release

AlanWong-MS:
Merge branch 'master' into release


AlanWong-MS:
Will squash merge.

robward-ms:
@AlanWong-MS - we should not squash merge; the history in release should match what's in master as much as possible.

PR #1084, opened 2018-02-16
Updating package ID to match the existing VSIX ID in the marketplace

AlanWong-MS:
### Issues
The ID in the new VS2017 Client Code Gen package is different from the ID of the current VSIX package on the [VS marketplace](https://marketplace.visualstudio.com/items?itemName=bingl.ODatav4ClientCodeGenerator).

### Description
This PR changes the ID to the existing ID.

- [ ] *Test cases added* N/A

### Additional work necessary
N/A

### DIFF
--- a/src/CodeGen/ODataT4ItemTemplate/source.extension.vsixmanifest
+++ b/src/CodeGen/ODataT4ItemTemplate/source.extension.vsixmanifest
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <PackageManifest Version="2.0.0" xmlns="http://schemas.microsoft.com/developer/vsx-schema/2011" xmlns:d="http://schemas.microsoft.com/developer/vsx-schema-design/2011">
     <Metadata>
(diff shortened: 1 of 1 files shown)


Nicte:
Hi, when will this new version of the Client Code Generator will be available in the marketplace? I can't find it even though the package version 7.4.1 has already been published and I was able to update the packages in my solution. Probably there is something I'm missing. Thanks!

AlanWong-MS:
@Nicte, we were anticipating releasing a beta Client Code Gen along with our 7.4.1 release, but unfortunately, the new package didn't have the same ID as the existing one on the Marketplace (which cause the new package to be rejected) so I had to fix it with this PR. For the time being, I'm considering uploading the beta package at #731 for people to try out--I should have that figured out early this week. I recommend subscribing to that thread for further updates. Thanks!

Nicte:
@AlanWong-MS thanks for the info! I was not aware of #731 I will keep an eye on it!

Issue #645, opened 2016-07-22
Enable updating top-level properties to null


TomDu:
Re-enable updating top-level properties to null after "odata.null" is removed (#630).

maubin-kl:
Is there an ETA on this? This is a breaking change from version prior to 6.0.0.

We currently cannot update to 6.0.0 because we have custom actions that returns a null value. We do expect to receive a null value on the client side and instead it is now an exception.

maubin-kl:
@TomDu @robward-ms

lpperras:
@TomDu Any news on this?

robward-ms:
The scenario is not supported per the language of the [spec](http://docs.oasis-open.org/odata/odata/v4.0/odata-v4.0-part1-protocol.html):

11.2.3 Requesting Individual Properties
To retrieve an individual property, the client issues a GET request to the property URL. The property URL is the entity read URL with "/" and the property name appended.
For complex typed properties, the path can be further extended with the name of an individual property of the complex type.
See [OData‑URL] for details.
**_If the property is single-valued and has the null value, the service responds with 204 No Content._**
If the property is not available, for example due to permissions, the service responds with 404 Not Found.

**_So the right thing to do is return 204._** 

However, I've made a change to the library to enable back-compat behavior via a writer validation that is on by default. When turned off, the exception is no longer throws an "value":null is written. Not that thus behavior is not compliant with the latest protocol. See #1093 for the change.

robward-ms:
Add a change to allow the odata@null to be written under 6.x compatibility mode and read unconditionally, see #1093

maubin-kl:
@robward-ms Thanks for making the fix. I understand that this is not in the spec but in order to not break our existing clients integration we still need to support that on our end.

I've reviewed the PR and the changes that you've made. However it's not really apparent to me how can I change the behaviour in my application. 

Where from my WebApiConfig can I set the ODataLibraryCompatibility to version 6.xxxx?

robward-ms:
@maubin-kl - The writer settings would need to be injected into the service container. [This ](https://github.com/OData/WebApi/blob/88c5f01f2ee2d685d873b2ab075909b7da6651f1/src/System.Web.OData/Extensions/HttpRequestMessageExtensions.cs#L317-L330)shows where in WebApi the writer settings are read.

You can read more about ODL and dependency injection [here](http://odata.github.io/odata.net/v7/#01-05-di-support).

robward-ms:
[...]

PR #1093, opened 2018-02-21
Enable updating top-level properties to null.

robward-ms:
### Issues

This pull request fixes issue #645.

### Description

Support for writing top-level properties with null values at
the time @odata.null:true was removed. The 4.0 spec states in
1.2.3 Requesting Individual Properties that if the property is
single-valued and has the null value, the service responds with
204 No Content. The library is correct in throwing an exception
when asked to write a null top-level values.

This change adds support for the scenario in the following ways:

1.) Writing a null top-level value still throws but with a well-defined
    error message.

2.) The scenario can be enabled by turning off the validation in the
    writer settings, i.e. settings.Validations = ~ ValidationKinds.ThrowOnTopLevelNullProperty.
    In this case, the value is written as: "value":null

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
@@ -1093,116 +1093,136 @@ namespace Microsoft.OData.JsonLight
             object propertyValue = missingPropertyValue;
             var customInstanceAnnotations = new Collection<ODataInstanceAnnotation>();
 
-            string payloadTypeName = null;
-            if (this.ReadingComplexProperty(propertyAndAnnotationCollector, expectedPropertyTypeReference, out payloadTypeName))
+            // Check for the special top-level OData 3.0 null marker in order to accommodate
+            // the responses written by 6.x version of this library.
+            if (this.IsTopLevel6xNullValue())
             {
-                // Figure out whether we are reading a complex property or not; complex properties are not wrapped while all others are.
-                // Since we don't have metadata in all cases (open properties), we have to detect the type in some cases.
-                this.AssertJsonCondition(JsonNodeType.Property, JsonNodeType.EndObject);
-
-                // Now read the property value
-                propertyValue = this.ReadNonEntityValue(
-                    payloadTypeName,
+                // NOTE: when reading a null value we will never ask the type resolver (if present) to resolve the
+                //       type; we always fall back to the expected type.
+                this.ReaderValidator.ValidateNullValue(
                     expectedPropertyTypeReference,
-                    propertyAndAnnotationCollector,
-                    /*collectionValidator*/ null,
                     /*validateNullValue*/ true,
-                    /*isTopLevelPropertyValue*/ true,
-                    /*insideComplexValue*/ true,
-                    /*propertyName*/ null);
+                    /*propertyName*/ null,
+                    null);
+
+                // We don't allow properties or non-custom annotations in the null payload.
+                this.ValidateNoPropertyInNullPayload(propertyAndAnnotationCollector);
+
+                propertyValue = null;
             }
             else
             {
-                bool isReordering = this.JsonReader is ReorderingJsonReader;
-
-                Func<string, object> propertyAnnotationReaderForTopLevelProperty =
-                    annotationName => { throw new ODataException(ODataErrorStrings.ODataJsonLightPropertyAndValueDeserializer_UnexpectedODataPropertyAnnotation(annotationName)); };
-
-                // Read through all top-level properties, ignore the ones with reserved names (i.e., reserved
-                // characters in their name) and throw if we find none or more than one properties without reserved name.
-                while (this.JsonReader.NodeType == JsonNodeType.Property)
+                string payloadTypeName = null;
+                if (this.ReadingComplexProperty(propertyAndAnnotationCollector, expectedPropertyTypeReference, out payloadTypeName))
                 {
-                    this.ProcessProperty(
+                    // Figure out whether we are reading a complex property or not; complex properties are not wrapped while all others are.
+                    // Since we don't have metadata in all cases (open properties), we have to detect the type in some cases.
+                    this.AssertJsonCondition(JsonNodeType.Property, JsonNodeType.EndObject);
+
+                    // Now read the property value
+                    propertyValue = this.ReadNonEntityValue(
+                        payloadTypeName,
+                        expectedPropertyTypeReference,
                         propertyAndAnnotationCollector,
-                        propertyAnnotationReaderForTopLevelProperty,
-                        (propertyParsingResult, propertyName) =>
-                        {
-                            switch (propertyParsingResult)
+                        /*collectionValidator*/ null,
+                        /*validateNullValue*/ true,
+                        /*isTopLevelPropertyValue*/ true,
+                        /*insideComplexValue*/ true,
+                        /*propertyName*/ null);
+                }
+                else
+                {
+                    bool isReordering = this.JsonReader is ReorderingJsonReader;
+
+                    Func<string, object> propertyAnnotationReaderForTopLevelProperty =
+                        annotationName => { throw new ODataException(ODataErrorStrings.ODataJsonLightPropertyAndValueDeserializer_UnexpectedODataPropertyAnnotation(annotationName)); };
+
+                    // Read through all top-level properties, ignore the ones with reserved names (i.e., reserved
+                    // characters in their name) and throw if we find none or more than one properties without reserved name.
+                    while (this.JsonReader.NodeType == JsonNodeType.Property)
+                    {
+                        this.ProcessProperty(
+                            propertyAndAnnotationCollector,
+                            propertyAnnotationReaderForTopLevelProperty,
+                            (propertyParsingResult, propertyName) =>
                             {
-                                case PropertyParsingResult.ODataInstanceAnnotation:
-                                    if (string.CompareOrdinal(ODataAnnotationNames.ODataType, propertyName) == 0)
-                                    {
-                                        // When we are not using the reordering reader we have to ensure that the 'odata.type' property appears before
-                                        // the 'value' property; otherwise we already scanned ahead and read the type name and have to now
-                                        // ignore it (even if it is after the 'value' property).
-                                        if (isReordering)
+                                switch (propertyParsingResult)
+                                {
+                                    case PropertyParsingResult.ODataInstanceAnnotation:
+                                        if (string.CompareOrdinal(ODataAnnotationNames.ODataType, propertyName) == 0)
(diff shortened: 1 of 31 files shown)


src/Microsoft.OData.Core/ValidationKinds.cs
```diff
@@ -36,6 +36,16 @@ public enum ValidationKinds
...
 
+        /// <summary>
+        /// Validates that top level properties cannot be null. If a top level property is null,
+        /// the response should be 204 (NoContent) per the spec: 
+        /// 11.2.3 Requesting Individual Properties
+        /// ...
+        /// If the property is single-valued and has the null value, the service responds with 204 No Content.
+        /// ...
+        /// </summary>
+        ThrowOnTopLevelNullProperty = 5,
```
AlanWong-MS: Should this be 8?

robward-ms: Yep.

biaol-odata:
PublicApi.bsl needs to be updated.

Also, the issue described in #645 is for writing the null value, but the protocol 11.2.3 is for the response side. Your change here is only for the writing side, correct?

robward-ms:
@biaol-odata - Yep, I'll update the baseline. My read of 11.2.3 was how to request a single property, which is when the writer would be used write a top-level property. Since that section specifies what happens on the server side, it seemed appropriate to follow for this fix. I fixed the writer to support the scenario and verified the reader would work in this case. If you think there is something I missed, let me know.

biaol-odata:
Do you mean that service shouldn't throw an exception when asked to write a null top-level values, if protocol section for request should be applicable for the response? It seems that we are turning an error case to a success case, which match the protocol, so we shouldn't need to provide the option of the flag ValidationKinds.ThrowOnTopLevelNullProperty in the change.

src/Microsoft.OData.Core/ValidationKinds.cs
```diff
@@ -36,6 +36,16 @@ public enum ValidationKinds
...
 
+        /// <summary>
+        /// Validates that top level properties cannot be null. If a top level property is null,
+        /// the response should be 204 (NoContent) per the spec: 
+        /// 11.2.3 Requesting Individual Properties
+        /// ...
+        /// If the property is single-valued and has the null value, the service responds with 204 No Content.
+        /// ...
+        /// </summary>
+        ThrowOnTopLevelNullProperty = 8,
```
biaol-odata: It seems that we are turning an error case to a success case, which match the protocol, so we shouldn't need to provide the option of the flag ValidationKinds.ThrowOnTopLevelNullProperty?

robward-ms: My read of 11.2.3 is that is a top-level property is all that is to be written, and the value of that property is null, the service should return 204. My thought is the writer could either write nothing to the stream and let the caller determine 204 is appropriate because the stream is empty OR the writer could throw in this case since 204 is appropriate and it's the callers job to return 204. So the default is throw and it's controlled by the validation value ThrowOnTopLevelNullProperty.

For when this causes problems for folks, as in the case of this bug, the options allows ThrowOnTopLevelNullProperty to be turned off, which will allow the writer to write the null value to the stream. This is NOT compliant with protocol. 

So I would describe the fix as formalize the existing error case but allow this behavior to be turned off and get a similar behavior to the 6.x library.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightOutputContextTests.cs
```diff
@@ -345,6 +375,10 @@ private static ODataJsonLightOutputContext CreateJsonLightOutputContext(MemorySt
             var settings = new ODataMessageWriterSettings { Version = ODataVersion.V4 };
             settings.SetServiceDocumentUri(new Uri("http://odata.org/test"));
             settings.ShouldIncludeAnnotation = ODataUtils.CreateAnnotationFilter("*");
+            if (allowNull)
+            {
+                settings.Validations = ~ ValidationKinds.ThrowOnTopLevelNullProperty;
```
biaol-odata: should be "~="?

biaol-odata: should be "&= !ValidationKinds.ThrowOnTopLevelNullProperty"?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -366,12 +360,21 @@ private void WriteStreamReferenceProperty(string propertyName, ODataStreamRefere
...
-                throw new ODataException("A null top-level property is not allowed to be serialized.");
+                // From the spec:
+                // 11.2.3 Requesting Individual Properties
+                // ...
+                // If the property is single-valued and has the null value, the service responds with 204 No Content.
+                // ...
+                // ValidateNullPropertyValue will throw by default but might allow this code to be reached
+                // if validation for this scenario is off. In that case, write "value": null which is not
+                // per spec but the best we could do indexer this case.
+                this.JsonWriter.WriteName(JsonLightConstants.ODataValuePropertyName);
```
marych: ODL6.x would not write a value property at all in this case. It would include "@odata.null" : true instead.

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
```diff
@@ -182,6 +183,11 @@ public ODataUri ODataUri
         /// </summary>
         internal bool ThrowOnUndeclaredPropertyForNonOpenType { get; private set; }
 
+        /// <summary>
+        /// Returns whether ThrowOnTopLevelNullProperty validation setting is enabled.
+        /// </summary>
+        internal bool ThrowOnTopLevelNullProperty { get; private set; }
```
biaol-odata: This new field should also be included in the CopyFrom method of this class.

robward-ms: Done. Also added tests for it.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -366,12 +360,27 @@ private void WriteStreamReferenceProperty(string propertyName, ODataStreamRefere
...
-                this.currentPropertyInfo.MetadataType.TypeReference, property.Name, this.Model);
+                this.currentPropertyInfo.MetadataType.TypeReference, property.Name,
+                this.currentPropertyInfo.IsTopLevel, this.Model);
 
             if (this.currentPropertyInfo.IsTopLevel)
             {
-                // TODO: Enable updating top-level properties to null #645
-                throw new ODataException("A null top-level property is not allowed to be serialized.");
+                if (this.JsonLightOutputContext.MessageWriterSettings.LibraryCompatibility <=
+                    ODataLibraryCompatibility.Version6)
```
mikepizzo: Should we define an ODataLibraryCompatibility.Version7, and use < ODataLibraryCompatibilityVersion7 here, in case we end up defining other 6.x compatibility versions? (which I know we don't plan to do, but might be safer to do <7 just in case?)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1555,6 +1574,13 @@ private object ReadEnumValue(bool insideJsonObjectValue, IEdmEnumTypeReference e
             // Try to read a null value
             if (ODataJsonReaderCoreUtils.TryReadNullValue(this.JsonReader, this.JsonLightInputContext, targetTypeReference, validateNullValue, propertyName, isDynamicProperty))
             {
+                if (isTopLevelPropertyValue && this.MessageReaderSettings.LibraryCompatibility == ODataLibraryCompatibility.Version6)
```
mikepizzo: >this.MessageReaderSettings.LibraryCompatibility == ODataLibraryCompatibility.Version6 [](start = 47, length = 85)

What if isTopLevelPropertyValue == true and this.MessageReaderSettings.LibraryCompatibility != 6?  I assume we don't expect to get here in that case, but if we did, wouldn't we want to raise an error? I'm wondering if the LibraryCompatibility == 6 is required here...

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1093,116 +1093,135 @@ private ODataProperty ReadTopLevelPropertyImplementation(IEdmTypeReference expec
             object propertyValue = missingPropertyValue;
             var customInstanceAnnotations = new Collection<ODataInstanceAnnotation>();
 
-            string payloadTypeName = null;
-            if (this.ReadingComplexProperty(propertyAndAnnotationCollector, expectedPropertyTypeReference, out payloadTypeName))
+            // Check for the special top-level null marker 
+            if (this.IsTopLevelNullValue() && this.MessageReaderSettings.LibraryCompatibility == ODataLibraryCompatibility.Version6)
```
mikepizzo: >this.IsTopLevelNullValue() && this.MessageReaderSettings.LibraryCompatibility == ODataLibraryCompatibility.Version6 [](start = 16, length = 115)

What if this.IsTopLevelNullValue() && != ODataLibCompatibility6 here? Should we handle that case here, when we read the IsTopLevelNullValue?

Or, since the payload is unambiguous in what it's trying to convey, even if it's not 4.0 compatible, should we be more lax *on reading* and always support, rather than requiring the client to specify "I might be talking to a bonehead service that returned the OData 3.0 representation of a top-level null"?  In other words, are there any cases where we would tell the client that they shouldn't set LibraryCompatibility6 *on the reader*?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
robward-ms: I thought we discussed having a simply compatibility flag, one that would not support 6.4.2. Just 6.x, 7.x which is why the enum is a simple int matching the major version number. Did I remember wrong?

mikepizzo: No, you recall correctly. However, I got back to my office and started working on some of the contextUrl fixes and it occurred to me that this would be a place where I could use the library compatibility flag to make a spec-compliant change that was still a change in behavior, so I started reconsidering how we might use this in the future, and thought perhaps introducing a slightly more granular system might make sense.  I wouldn't make it a lot more complex by trying to specify support for future builds of earlier major version (maintenance) releases, but being able to distinguish, i.e., between 7.4.3 and 7.4.4. might be useful.

PR #1106, opened 2018-03-06
Pr omit values null

biaol-odata:
### Issues

*This pull request introduces omitNullValues controlled by omit-values=nulls Preference header per OData v4.01.*

### Description

* Current changes include the following areas: *
- serialize JSON object to response with omit null values option according to Preference-Applied header's parameter value.
- de-serialize JSON object from response with omitted null values restored per the option used according to the Preference-Applied header's parameter value.
- has no effects on request payload and delta payload.

Based on PR #944 from kosinsky.

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
```diff
@@ -196,6 +196,11 @@ public ODataMessageQuotas MessageQuotas
         /// </summary>
         internal bool ThrowOnUndeclaredPropertyForNonOpenType { get; private set; }
 
+        /// <summary>
+        /// True if null values are omitted, per the omit-values parameter in Preference header of the message.
+        /// </summary>
+        internal bool NullValuesOmitted { get; set; }
```
kosinsky: >internal [](start = 8, length = 8)

Should we make that public to allow tests in projects using ODL simple injection? Injecting via PreferenceAppliedHeader() can be a bit complicated (to much stuff to mock)

test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/JsonLight/JsonLightEntryWriterTests.cs
```diff
@@ -432,14 +432,104 @@ public void ActionAndFunctionPayloadOrderTest()
...
+                    PayloadEdmElementType = testCase.EntityType,
+                    SkipTestConfiguration = testCase.SkipTestConfiguration
+                });
+
+            this.CombinatorialEngineProvider.RunCombinations(
+                testDescriptors,
+                this.WriterTestConfigurationProvider.JsonLightFormatConfigurationsWithIndent,
+                (testDescriptor, testConfiguration) =>
+                {
+                    testConfiguration.MessageWriterSettings.OmitNullValues = true;
```
kosinsky: Could you add EndToEnd test that uses  responseMessageWithModel.PreferenceAppliedHeader().OmitValues= "nulls"; to check e2e? Something like tests for AnnotationFilter in WritePayloadTests or InstanceAnnotationWriterIntegrationTests

kosinsky:
>             if (this.JsonLightOutputContext.MessageWriterSettings.Validations != ValidationKinds.None)

For performance reasons consider adding following optimization:

            ODataValue value = property.ODataValue;

            // If no validation is required, we don't need property serialization info and could try to write null property right away
            // For very wide and sparse outputs it allows to avoid a lot of dictionary lookups
            if (!this.MessageWriterSettings.ThrowIfTypeConflictsWithMetadata && omitNullValues)
            {
                if (value is ODataNullValue || value == null)
                {
                    return;
                }
            }

---
Refers to: src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs:192 in ba5d506. [](commit_id = ba5d5065a929d89b052279fa2ad89e4182187d2e, deletion_comment = False)

src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
biaol-odata: I think the NullValuesOmitted is caused by Preference-Applied header parameter value in the specific request. ODataPreferenceHeader(responseMessage).OmitValus = null should do the trick. 
If it were the case for DependencyInjection, we would need to do it with bool flag for convenience, but it is not the case here, right?

test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/JsonLight/JsonLightEntryWriterTests.cs
biaol-odata: Will do.

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
```diff
@@ -353,6 +367,64 @@ public void ReadingTypeDefinitionPayloadJsonLight()
...
+            this.ReadEntryPayload(model, payload, entitySet, entityType,
+                reader =>
+                {
+                    switch (reader.State)
+                    {
+                        case ODataReaderState.ResourceStart:
+                            entries.Add(reader.Item as ODataResource);
+                            break;
+                        case ODataReaderState.NestedResourceInfoStart:
+                            navigationLink = (ODataNestedResourceInfo)reader.Item;
```
biaol-odata: rename navigationLink to "nestedResourceInfo".

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
```diff
@@ -773,6 +845,246 @@ public void ReadingNullValueForDeclaredCollectionPropertyInComplexTypeShouldFail
...
+            person.Should().NotBeNull();
+            person.Properties.FirstOrDefault(s => string.Equals(s.Name, "Id", StringComparison.Ordinal)).Value.Should().Be(0);
+            // omitted null-able property should be restored as null.
+            person.Properties.FirstOrDefault(s => string.Equals(s.Name, "Height", StringComparison.Ordinal)).Value.Should().BeNull();
+            // missing non-null-able property should not be restored.
+            person.Properties.Any(s => string.Equals(s.Name, "Weight", StringComparison.Ordinal)).Should().BeFalse();
+        }
+
+        [Fact]
+        public void ReadingPropertyInTopLevelOrInComplexTypeShouldRestoreOmittedNullValuesWithSelectedPropertiesEntireSubTree()
```
biaol-odata: Add a test case for dynamic property selected, and it is not in the payload, should be materialized as null.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2259,91 @@ private void EndEntry()
...
+                                // Skip declared properties that are not null-able types.
+                                continue;
+                            }
+
+                            // Response should be generated using case sensitive, matching EDM defined by metadata.
+                            // In order to find potential omitted properties, need to check
+                            // resource properties read and navigation properties read.
+                            if (!resource.Properties.Any(
+                                p => p.Name.Equals(currentProperty.Name, StringComparison.Ordinal))
+                                && !resourceState.NavigationPropertiesRead.Contains(currentProperty.Name))
```
biaol-odata: Make sure we have test case to cover complex type and navigation property.

test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/JsonLight/JsonLightEntryWriterTests.cs
biaol-odata: Added e2e test WriteJsonWittoutModelTests::ExpandedCustomerEntryOmitNullValuesTest with proper header and null values for various cases.

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
```diff
@@ -353,6 +367,64 @@ public void ReadingTypeDefinitionPayloadJsonLight()
             address.Properties.FirstOrDefault(s => string.Equals(s.Name, "CountryRegion", StringComparison.OrdinalIgnoreCase)).Value.Should().Be("China");
         }
 
+        [Fact]
+        public void ReadingTypeDefinitionPayloadJsonLightWithOmittedNullValues()
```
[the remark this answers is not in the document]
biaol-odata: Added false case for comparison for all new tests as well (omittedNullValue=false should be the default case for all old tests). 

---
In reply to: [173004227](https://github.com/OData/odata.net/pull/1106#discussion_r173004227) [](ancestors = 173004227)

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
biaol-odata: Add unknown properties to context URI in test cases to verify unknown properties are ignored during deserialization.
Dynamic properties don't show up in context URI in response, therefore omitted one are not materialized as nulls.

---
In reply to: [173008510](https://github.com/OData/odata.net/pull/1106#discussion_r173008510) [](ancestors = 173008510)

kosinsky (dismissed):
:shipit:

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
biaol-odata: Already have coverage for complex type.
Added coverage for navigation link in e2e test ExpandedCustomerEntryOmitNullValuesTest.
Added various test case for expanded navigation property in PropertyAndValueJsonLightReaderIntegrationTests: expanded navigation link should have no content when the link is null (see sample link: [http://services.odata.org/V4/TripPinService/People](http://services.odata.org/V4/TripPinService/People)('ronaldmundy')?$expand=Photo)

---
In reply to: [173012135](https://github.com/OData/odata.net/pull/1106#discussion_r173012135) [](ancestors = 173012135)

src/Microsoft.OData.Core/ODataMessageReader.cs
```diff
@@ -169,6 +169,17 @@ public ODataMessageReader(IODataResponseMessage responseMessage, ODataMessageRea
...
+            // Whether null values were omitted in the response.
+            this.settings.NullValuesOmitted = false;
+            string omitValuePreferenceApplied = responseMessage.PreferenceAppliedHeader().OmitValues;
+            if (omitValuePreferenceApplied != null)
+            {
+                // If the Preference-Applied header's omit-values parameter is present in the response, its value should
+                // be applied to the reader's setting.
+                this.settings.NullValuesOmitted =
+                    omitValuePreferenceApplied.Equals(ODataConstants.OmitValuesNulls, StringComparison.OrdinalIgnoreCase);
+            }
```
mikepizzo: The check for the omit-nulls preference in the ODataMessageWriter is more concise.  Any reason not to use the same code here?

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -336,11 +339,40 @@ internal IEnumerable<IEdmNavigationProperty> GetSelectedNavigationProperties(IEd
...
+
+            if (this.selectionType == SelectionType.EntireSubtree || this.hasWildcard)
+            {
+                return structuredType.DeclaredProperties;
+            }
+
+            Debug.Assert(this.selectedProperties != null, "selectedProperties != null");
+
+            // Get declared properties selected, and filter out unrecognized properties.
+            return this.selectedProperties.Select(structuredType.FindProperty).OfType<IEdmProperty>();
```
mikepizzo: Why are we omitting undefined properties? If a dynamic property is explicitly requested in the select list, I would expect to return a null value for it if it was a null value or omitted in the payload.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2259,109 @@ private void EndEntry()
...
+        /// <summary>
+        /// For response de-serialization, restore the null values for properties that are omitted
+        /// by the omit-values=nulls preference header.
+        /// </summary>
+        private void RestoreOmittedNullValues()
+        {
+            // Restore omitted null value only when processing response and when
+            // Preference header omit-values=nulls is specified.
+            if (this.jsonLightInputContext.ReadingResponse
+                && this.jsonLightInputContext.MessageReaderSettings.NullValuesOmitted)
```
mikepizzo: should this be encapsulated in a method on the input context, similar to the ShouldWriteNullValues() method on the output context?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -202,8 +223,6 @@ private bool IsOpenProperty(ODataProperty property)
 
             WriteInstanceAnnotation(property, isTopLevel, currentPropertyInfo.MetadataType.IsUndeclaredProperty);
```
mikepizzo: this logic should come before the check for null values.  Even if the property is null, it can have instance annotations (in fact, a common use case for instance annotations is annotating a null value with why it is null).  Please move this before the check for null values, and add a test to verify that instance annotations are correctly written/read for annotations applied to null values, with and without omit-nulls.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -227,9 +246,9 @@ private bool IsOpenProperty(ODataProperty property)
                 return;
             }
 
-            if (value is ODataNullValue || value == null)
+            if (isNullValue)
             {
-                this.WriteNullProperty(property);
+                this.WriteNullProperty(property, omitNullValues);
```
mikepizzo: do you need the new omitNullValues parameter in WriteNullProperty? Why would you get this far if the property was a null value and omitNullValues was true?  Is this only for validating the metadata?  Can you handle that when you check for isNullValue, and get rid of the new omitNullValues flag to WriteNullProperty?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -356,8 +375,10 @@ private void WriteStreamReferenceProperty(string propertyName, ODataStreamRefere
         /// Writes a Null property.
         /// </summary>
         /// <param name="property">The property to write out.</param>
+        /// <param name="omitNullValues">Whether to omit the writing of this null property value.</param>
         private void WriteNullProperty(
-            ODataProperty property)
+            ODataProperty property,
+            bool omitNullValues)
```
mikepizzo: See above.  I would think that you can handle all cases of null and omit=null in WriteProperty, and never get to WriteNullProperty for null properties in the omit nulls case.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2259,109 @@ private void EndEntry()
...
+                                as IEnumerable<IEdmProperty>;
+                            selectedProperties = selectedProperties.Concat(
+                                resourceState.SelectedProperties.GetSelectedProperties(edmStructuredType))
+                                .Distinct();
+                        }
+
+                        foreach (IEdmProperty currentProperty in selectedProperties)
+                        {
+                            Debug.Assert(currentProperty.Type != null, "currentProperty.Type != null");
+                            if (!currentProperty.Type.IsNullable || currentProperty.PropertyKind == EdmPropertyKind.Navigation)
```
mikepizzo: >|| currentProperty.PropertyKind == EdmPropertyKind.Navigation [](start = 65, length = 61)

this seems suspect. Doesn't that mean that we wouldn't return a null value for a nav prop that was explicitly selected?  why did we add the nav properties to the selectedProperties list if we just ignore them here?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
mikepizzo: I don't think TripPinService is handling this case correctly; if an expanded single-value navigation property is null, I would expect a null value be returned in the payload, and I would expect to get the null value as nested content when I ready the payload.  We should make sure that ODL does support reading and writing a single-value null navigation property, and fix this case to handle a missing explicitly selected null navigation property the same as an explicit null value for a nullable single-value navigation property in the payload.

---
In reply to: [174645486](https://github.com/OData/odata.net/pull/1106#discussion_r174645486) [](ancestors = 174645486,173012135)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
biaol-odata: good point. Will do.

---
In reply to: [186531887](https://github.com/OData/odata.net/pull/1106#discussion_r186531887) [](ancestors = 186531887)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
biaol-odata: Moved. Thanks.


---
In reply to: [186558582](https://github.com/OData/odata.net/pull/1106#discussion_r186558582) [](ancestors = 186558582)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
biaol-odata: For all cases of null and omit=null, code block around L392 is needed for <V7. Am I missing anything?

---
In reply to: [186562291](https://github.com/OData/odata.net/pull/1106#discussion_r186562291) [](ancestors = 186562291)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
biaol-odata: Yes, see comment inside the if block. I believe nav links (@odata.navigationlink annotation) are added by SelectedPropertiesNodes.GetSelectedNavigationProperties method, it looks like a logical method, additionally it has 10 references of src+tests. Filtering out here during the processing of null value restoration should be a reasonable approach?

I will work on ensuring that nested nav property which should be PropertyKind of EdmPropertyKind.Structual can be restored as null.

---
In reply to: [186563420](https://github.com/OData/odata.net/pull/1106#discussion_r186563420) [](ancestors = 186563420)

src/Microsoft.OData.Core/Microsoft.OData.Core.txt
```diff
@@ -242,6 +242,8 @@ HttpHeaderValueLexer_FailedToReadTokenOrQuotedString=An error occurred when pars
 HttpHeaderValueLexer_InvalidSeparatorAfterQuotedString=An error occurred when parsing the HTTP header '{0}'. The header value '{1}' is incorrect at position '{2}' because '{3}' is not a valid separator after a quoted-string.
 HttpHeaderValueLexer_EndOfFileAfterSeparator=An error occurred when parsing the HTTP header '{0}'. The header value '{1}' is incorrect at position '{2}' because the header value should not end with the separator '{3}'.
 
+HttpPreferenceHeader_InvalidValueForToken=The value '{0}' is not a valid value for OData token '{1}' in HTTP Prefer or Preference-Applied headers.
```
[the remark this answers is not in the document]
biaol-odata: removed because unknown preference header value should just be ignored.

---
In reply to: [186522770](https://github.com/OData/odata.net/pull/1106#discussion_r186522770) [](ancestors = 186522770)

src/Microsoft.OData.Core/ODataMessageReader.cs
biaol-odata: Updated to be the same as the writer; should be equivalent except that on reader side this flag should be false by default.

---
In reply to: [186526760](https://github.com/OData/odata.net/pull/1106#discussion_r186526760) [](ancestors = 186526760)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
biaol-odata: Moved to ODataReaderCore.ShouldRestoreNullValues().

---
In reply to: [186532720](https://github.com/OData/odata.net/pull/1106#discussion_r186532720) [](ancestors = 186532720)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
biaol-odata: The omitNullValues is needed in WriteNullProperty to differentiate delta and others; When we get this far, we knows property is a null but omitNullValues can still be true or false; I think validating metadata is done before this (L215,L217); I believe WriteNullProperty properly encapsulates the logic to write null property, without the omitNullValues flag it wouldn't be able to tell whether null values needs to be serialized (e.g. delata or non-delta).
I have added additional information about the omitNullValues param in pending changes.

---
In reply to: [186561856](https://github.com/OData/odata.net/pull/1106#discussion_r186561856) [](ancestors = 186561856)

src/Microsoft.OData.Core/ODataPreferenceHeader.cs
```diff
@@ -243,6 +250,34 @@ public string AnnotationFilter
...
+                ExceptionUtils.CheckArgumentStringNotEmpty(value, "OmitValues");
+
+                string normalizedValue = value.Trim().ToLowerInvariant();
+                if (string.Equals(normalizedValue, ODataConstants.OmitValuesNulls, StringComparison.Ordinal))
+                {
+                    this.Set(new HttpHeaderValueElement(OmitValuesPreferenceToken, ODataConstants.OmitValuesNulls, EmptyParameters));
+                }
+
+                // No-ops for other preference values, including "defaults" which is in the OData protocol but not implemented yet.
+            }
```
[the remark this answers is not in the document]
biaol-odata: We could, that would require figuring out the value and associated parameters of the omit-values header (see HttpHeaderValueElement c'tor) that we don't have details about or doesn't belong to this task yet. It is typical to silently ignore unrecognizable preference header, and we have inline-comment. Sounds reasonable?

---
In reply to: [186528628](https://github.com/OData/odata.net/pull/1106#discussion_r186528628) [](ancestors = 186528628)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
biaol-odata: Will look into this issue about expanded nav property and it should be nestedInfo with null restored.

---
In reply to: [186565783](https://github.com/OData/odata.net/pull/1106#discussion_r186565783) [](ancestors = 186565783,174645486,173012135)

biaol-odata:
@mikepizzo I just pushed latest commit for CR updates above. As we discussed offline, please check whether it is OK to not emitting the null for nested resource omitted on the wire; otherwise we need to make further change related to the overall json reader. Thanks.

biaol-odata:
@mikepizzo  I have updated with the latest commit to not omit null values for expanded navigation properties. Can you please take a look? Thanks.

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -336,11 +339,62 @@ internal IEnumerable<IEdmNavigationProperty> GetSelectedNavigationProperties(IEd
...
+
+            // We cannot determine the selected properties without the user model. This means we won't be computing the missing properties.
+            if (structuredType == null)
+            {
+                return EmptyEdmProperties;
+            }
+
+            if (this.selectionType == SelectionType.EntireSubtree || this.hasWildcard)
+            {
+                return structuredType.DeclaredProperties;
```
mikepizzo: >DeclaredProperties [](start = 38, length = 18)

This only returns the properties declared directly on the type; you should use structuredType.Properties to include inherited properties.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2259,117 @@ private void EndEntry()
...
+                    resource.Properties.Concat(new List<ODataProperty>() { property });
+            }
+        }
+
+        /// <summary>
+        /// Get all properties defined by the EDM structural type and its base types.
+        /// </summary>
+        /// <param name="edmStructuredType">The EDM structural type.</param>
+        /// <returns>All the properties of this type.</returns>
+        private static IEnumerable<IEdmProperty> GetSelfAndBaseTypeProperties(IEdmStructuredType edmStructuredType)
```
mikepizzo: >GetSelfAndBaseTypeProperties [](start = 49, length = 28)

Note that you can use the Properties() extension method to get all declared and inherited properties for a structured type.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2259,117 @@ private void EndEntry()
...
+                    {
+                        ODataResourceBase resource = resourceState.Resource;
+
+                        IEnumerable<IEdmProperty> selectedProperties;
+
+                        if (resourceState.SelectedProperties == SelectedPropertiesNode.EntireSubtree)
+                        {
+                            selectedProperties = GetSelfAndBaseTypeProperties(edmStructuredType);
+                        }
+                        else
```
mikepizzo: This case is handled already in GetSelectedProperties()

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2259,117 @@ private void EndEntry()
...
+                        {
+                           // Partial subtree.
+                           selectedProperties = resourceState.SelectedProperties.GetSelectedProperties(edmStructuredType);
+                        }
+
+                        // All dynamic properties: null value restoration.
+                        foreach (string name in resourceState.SelectedProperties.GetSelectedDynamicProperties(edmStructuredType))
+                        {
+                            RestoreNullODataProperty(name, resourceState);
+                        }
```
mikepizzo: What if the dynamic property already existed in the payload; this seems like it would add the property a second time with the null value.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2259,117 @@ private void EndEntry()
...
+                IODataJsonLightReaderResourceState resourceState = this.CurrentResourceState;
+                IEdmStructuredType edmStructuredType = resourceState.ResourceType;
+
+                if (resourceState.SelectedProperties == SelectedPropertiesNode.Empty)
+                {
+                    return;
+                }
+                else
+                {
+                    if (resourceState.Resource != null && edmStructuredType != null)
```
mikepizzo: >&& edmStructuredType != null [](start = 55, length = 28)

If edmStructuredType is null, why not just treat all properties as dynamic and restore anything that is not in the payload?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2259,117 @@ private void EndEntry()
...
+                        {
+                            RestoreNullODataProperty(name, resourceState);
+                        }
+
+                        foreach (IEdmProperty currentProperty in selectedProperties)
+                        {
+                            Debug.Assert(currentProperty.Type != null, "currentProperty.Type != null");
+
+                            // Skip declared properties that are not null-able types.
+                            // Skip navigation properties (no navigation link means link value is null)
```
mikepizzo: >(no navigation link means link value is null) [](start = 58, length = 45)

this is not true

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2259,117 @@ private void EndEntry()
...
+                            // Skip properties that have been read (primitive or structural).
+                            if (!currentProperty.Type.IsNullable
+                                || currentProperty.PropertyKind == EdmPropertyKind.Navigation
+                                || resource.Properties.Any(p => p.Name.Equals(currentProperty.Name, StringComparison.Ordinal))
```
mikepizzo: Does selectedProperties ever include navigation properties?  From the code above, it looks like it should only ever include structural properties.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2259,117 @@ private void EndEntry()
...
+                                || resourceState.NavigationPropertiesRead.Contains(currentProperty.Name))
```
mikepizzo: I don't think this check is required either (you already eliminated navigation properties twice...)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -336,11 +339,62 @@ internal IEnumerable<IEdmNavigationProperty> GetSelectedNavigationProperties(IEd
...
+        /// <summary>
+        /// Return names of dynamic properties if current structured node is open type.
+        /// </summary>
+        /// <param name="structuredType">The current structured type.</param>
+        /// <returns>The names of dynamic properties, empty if the current structured type is not open type.</returns>
+        internal IEnumerable<string> GetSelectedDynamicProperties(IEdmStructuredType structuredType)
+        {
+            IEnumerable<string> dynamicProperties = Enumerable.Empty<string>();
+
+            if (structuredType.IsOpen && this.selectedProperties != null && this.selectedProperties.Count > 0)
```
mikepizzo: >structuredType.IsOpen &&  [](start = 16, length = 25)

remove this check.  the service can always return properties not defined on the type, whether open or not.  if it is in the list of selectedProperties, and not in the payload, we should restore it.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
```diff
@@ -331,6 +331,7 @@ protected override void StartResource(ODataResource resource)
                 this.ResourceType,
                 resource.Properties,
                 false /* isComplexValue */,
+                this.ShouldOmitNullValues(),
```
mikepizzo: why do you pass this in, rather than have the jsonLightResourceSerializer get it from the Input Context?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
mikepizzo: On L219, if the value is the null value and omitNullValues is true, you would have returned, so you would never reach this line for a null value if omitNullValues is true.  Right?

---
In reply to: [187209557](https://github.com/OData/odata.net/pull/1106#discussion_r187209557) [](ancestors = 187209557,186561856)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
mikepizzo: You only call WriteNullProperty in one place, and you have an optimization before calling writeNullProperty that, if the property is null and omitNullValues is true, you return.  So this will only ever be called if it is a null value that you really want to write.

---
In reply to: [187208933](https://github.com/OData/odata.net/pull/1106#discussion_r187208933) [](ancestors = 187208933,186562291)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
biaol-odata: ok

---
In reply to: [198068206](https://github.com/OData/odata.net/pull/1106#discussion_r198068206) [](ancestors = 198068206)

src/Microsoft.OData.Core/ODataMessageReader.cs
```diff
@@ -169,6 +169,16 @@ public ODataMessageReader(IODataResponseMessage responseMessage, ODataMessageRea
             this.model = model ?? GetModel(this.container);
             this.edmTypeResolver = new EdmTypeReaderResolver(this.model, this.settings.ClientCustomTypeResolver);
 
+            // Whether null values were omitted in the response, false by default.
+            this.settings.NullValuesOmitted = false;
```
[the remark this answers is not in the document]
biaol-odata: Setting the false as the default value, see comment in the line above.

---
In reply to: [198055075](https://github.com/OData/odata.net/pull/1106#discussion_r198055075) [](ancestors = 198055075)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
biaol-odata: OK. Thanks.

---
In reply to: [198061988](https://github.com/OData/odata.net/pull/1106#discussion_r198061988) [](ancestors = 198061988)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
biaol-odata: Not sure what you meant. It is handled here by GetSelectedProperties().

---
In reply to: [198062679](https://github.com/OData/odata.net/pull/1106#discussion_r198062679) [](ancestors = 198062679)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
biaol-odata: Resolved. Thanks.

---
In reply to: [198063535](https://github.com/OData/odata.net/pull/1106#discussion_r198063535) [](ancestors = 198063535)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
biaol-odata: In what case can edmStructuredType be null?
For OpenType w/ dynamic props, it should be the edm type; for Untype it should be 'IEdmComplexType', right? 

---
In reply to: [198065705](https://github.com/OData/odata.net/pull/1106#discussion_r198065705) [](ancestors = 198065705)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
biaol-odata: There is a nested if-clause inside L219, code won't return when the if-clause @L221 is not satisfied.

---
In reply to: [198078564](https://github.com/OData/odata.net/pull/1106#discussion_r198078564) [](ancestors = 198078564,187209557,186561856)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
biaol-odata: omitNullVallues could be either true or false and needs to be passed in. See related comment in caller.

---
In reply to: [198079167](https://github.com/OData/odata.net/pull/1106#discussion_r198079167) [](ancestors = 198079167,187208933,186562291)

biaol-odata:
@mikepizzo I have updated the PR per your comments. Can you please take a look at the pending comments? Thanks!

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
mikepizzo: Presence or absence of navigation link doesn't have anything to do with whether or not the navigation property is null.

Navigation links are omitted by default in minimal metadata responses when they follow convention.  If you request full metadata, you will get back the link, even if the value of the navigation property is false.  For example:

  [https://services.odata.org/V4/TripPinService/People](https://services.odata.org/V4/TripPinService/People)('ronaldmundy')?$select=UserName,Photo&$format=application/json;odata.metadata=full

returns:
{
  "@odata.context": "[http://services.odata.org/V4/TripPinService/$metadata#People(UserName,Photo)/$entity](http://services.odata.org/V4/TripPinService/$metadata#People(UserName,Photo)/$entity)",
  "@odata.type": "#Microsoft.OData.SampleService.Models.TripPin.Person",
  "@odata.id": "[http://services.odata.org/V4/TripPinService/People(](http://services.odata.org/V4/TripPinService/People()'ronaldmundy')",
  "@odata.etag": "W/\"08D60D2CA6FBB30E\"",
  "@odata.editLink": "[http://services.odata.org/V4/TripPinService/People(](http://services.odata.org/V4/TripPinService/People()'ronaldmundy')",
  "UserName": "ronaldmundy",
  "Photo@odata.associationLink": "[http://services.odata.org/V4/TripPinService/People(](http://services.odata.org/V4/TripPinService/People()'ronaldmundy')/Photo/$ref",
  "Photo@odata.navigationLink": "[http://services.odata.org/V4/TripPinService/People(](http://services.odata.org/V4/TripPinService/People()'ronaldmundy')/Photo"
}

On the other hand, in minimal metadata, the links are omitted even if the value of the navigation property is non-null:

  [https://services.odata.org/V4/TripPinService/People](https://services.odata.org/V4/TripPinService/People)('russellwhyte')?$select=UserName,Photo

---
In reply to: [209789007](https://github.com/OData/odata.net/pull/1106#discussion_r209789007) [](ancestors = 209789007,198066211)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -336,11 +339,62 @@ internal IEnumerable<IEdmNavigationProperty> GetSelectedNavigationProperties(IEd
...
+
+            // Get declared properties selected, and filter out unrecognized properties.
+            return this.selectedProperties.Select(structuredType.FindProperty).OfType<IEdmProperty>();
+        }
+
+        /// <summary>
+        /// Return names of dynamic properties if current structured node is open type.
+        /// </summary>
+        /// <param name="structuredType">The current structured type.</param>
+        /// <returns>The names of dynamic properties, empty if the current structured type is not open type.</returns>
```
mikepizzo: >, empty if the current structured type is not open type [](start = 52, length = 55)

This part of the description should be deleted now that we removed the check for non-open type.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
mikepizzo: In the above conditional you explicitly check for resourceState.SelectedProperties==SelectedPropertiesNode.EntireSubtree.  If not SelectedPropertiesNode.EntireSubtree, you call GetSelectedProperties, which also checks for SelectionType.EntireSubtree (along with this.hasWildcard) and treats (almost) the same way (by returning all of the properties of the structured type).  You should be able to remove the check here for SelectedPropertiesNode.EntireSubtree, and just invoke GetSelectedProperties.

---
In reply to: [209784487](https://github.com/OData/odata.net/pull/1106#discussion_r209784487) [](ancestors = 209784487,198062679)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
mikepizzo: Why is this by design?  If this is a derived type, why are we only restoring properties declared directly on the derived type and not inherited properties?

---
In reply to: [209788492](https://github.com/OData/odata.net/pull/1106#discussion_r209788492) [](ancestors = 209788492,198059137)

src/Microsoft.OData.Core/ODataMessageReader.cs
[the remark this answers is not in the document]
mikepizzo: in this case, you are *always* setting the value to false, and then *possibly* setting it back to true.  Why not just set it to true if the preference apply matches, otherwise set it to false, as in:

            if (string.Equals(responseMessage.PreferenceAppliedHeader().OmitValues, ODataConstants.OmitValuesNulls,
                StringComparison.OrdinalIgnoreCase))
            {
                // If the Preference-Applied header's omit-values parameter is present in the response, its value should
                // be applied to the reader's setting.
                this.settings.NullValuesOmitted = true;
            }
            else
            {
                this.settings.NullValuesOmitted = false;
            }

---
In reply to: [209783953](https://github.com/OData/odata.net/pull/1106#discussion_r209783953) [](ancestors = 209783953,198055075)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2259,109 @@ private void EndEntry()
...
+                                .Distinct();
+                        }
+
+                        foreach (IEdmProperty currentProperty in selectedProperties)
+                        {
+                            Debug.Assert(currentProperty.Type != null, "currentProperty.Type != null");
+                            if (!currentProperty.Type.IsNullable || currentProperty.PropertyKind == EdmPropertyKind.Navigation)
+                            {
+                                // Skip declared properties that are not null-able types, and declared navigation properties
+                                // which are specified using odata.navigationlink annotations whose absence stand for null values.
```
mikepizzo: As per other comment; this is not true.  An omitted odata.navigationlink does not imply a null value.  the odata.navigationlink always has a (declared or implied) value; in minimal metadata it is omitted if it matches convention, in full metadata it is always written. this has nothing to do with the value of the navigation property.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
mikepizzo: The nested if clause checks two things:
1) If this is top level, in which case omitNullValues doesn't apply (it's already ignored in WriteNullProperty)
2) whether we should check metadata

For #2, I'm not convinced we should skip checking metadata for null values as this would introduce a change in what the writer was able to write whether omit values was true or false:
a) this would allow the writer to write a non-null value and one or more null values for the same property when omitNullValues was true, which should really not be valid.
b) this would allow the writer to write a null value for an undeclared property, even though throwOnUndeclaredProperty was true, when omitNullValues is true

So, we should simplify move the validations in L227 and L229 before the check for null values in L219, and in L219 return for null non-top level elements if omitNullValues is true.  Then we can remove the omitNullValues argument from WriteNullProperty, since it will only be called when we really want to write the null value.

---
In reply to: [209787942](https://github.com/OData/odata.net/pull/1106#discussion_r209787942) [](ancestors = 209787942,198078564,187209557,186561856)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
mikepizzo: see comment in caller. I don't see any reason why we can't check this case before calling the method and would have to pass it in.

---
In reply to: [209788338](https://github.com/OData/odata.net/pull/1106#discussion_r209788338) [](ancestors = 209788338,198079167,187208933,186562291)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -336,11 +339,62 @@ internal IEnumerable<IEdmNavigationProperty> GetSelectedNavigationProperties(IEd
...
+        internal IEnumerable<string> GetSelectedDynamicProperties(IEdmStructuredType structuredType)
```
mikepizzo: >GetSelectedDynamicProperties [](start = 37, length = 28)

Why are we breaking dynamic properties into a separate method from declared properties?  Doesn't that require us to do multiple loops through selected properties (and then, for each selected property, find out if it is declared)?

Why don't we just do a single loop through selectedProperties and process each as appropriate?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
biaol-odata: Updated the inline comment. Thanks for the clarification!

---
In reply to: [213502295](https://github.com/OData/odata.net/pull/1106#discussion_r213502295) [](ancestors = 213502295,209789007,198066211)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
biaol-odata: Updated to use method structuredType.Properties().

---
In reply to: [213509115](https://github.com/OData/odata.net/pull/1106#discussion_r213509115) [](ancestors = 213509115,209788492,198059137)

src/Microsoft.OData.Core/ODataMessageReader.cs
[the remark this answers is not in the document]
biaol-odata: These two ways are mathematically equivalent. I still slightly incline to initializing it to false first in order to emphasize the fact that preference header omit-values=nulls is required to be set explicitly.

---
In reply to: [213509847](https://github.com/OData/odata.net/pull/1106#discussion_r213509847) [](ancestors = 213509847,209783953,198055075)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
biaol-odata: I see, though that your comment was related to the else branch at L2291. Removed the if-branch above.

---
In reply to: [213508907](https://github.com/OData/odata.net/pull/1106#discussion_r213508907) [](ancestors = 213508907,209784487,198062679)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
biaol-odata: These two methods are invoked from code segment around L2287. Regular property restoration needs to check for nullable type (otherwise no restoration); while dynamic property doesn't require such check. Therefore two passes are used here (same linear complexity of O(N)).

---
In reply to: [213517170](https://github.com/OData/odata.net/pull/1106#discussion_r213517170) [](ancestors = 213517170)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
biaol-odata: For performance reason, kosinsky has requested (see his comment at L192) that skipping the validations at L227-229, using check for null values in L219 *before hand*. Switching the order as you suggest above would contradict that, right?

---
In reply to: [213515156](https://github.com/OData/odata.net/pull/1106#discussion_r213515156) [](ancestors = 213515156,209787942,198078564,187209557,186561856)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
biaol-odata: see response in the caller.

---
In reply to: [213515321](https://github.com/OData/odata.net/pull/1106#discussion_r213515321) [](ancestors = 213515321,209788338,198079167,187208933,186562291)

biaol-odata:
@Michael Pizzo Thanks for the comments. Can you check again to see whether all your comments have been addressed properly?

biaol-odata:
@mikepizzo The latest changes for expanded entity's omit-nulls are in last commit above. All other commits above are the original commits w/ re-base to the tip of master branch.
Can you please take a look? Thanks!

src/Microsoft.OData.Core/ODataMessageReader.cs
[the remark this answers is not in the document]
mikepizzo: I still prefer setting it once to twice as it feels cleaner, but I'm not going to force it.

---
In reply to: [213861831](https://github.com/OData/odata.net/pull/1106#discussion_r213861831) [](ancestors = 213861831,213509847,209783953,198055075)

src/Microsoft.OData.Core/ODataResource.cs
```diff
@@ -186,6 +186,11 @@ public IEnumerable<ODataProperty> Properties
             set { this.properties = value; }
         }
 
+        internal bool IsExpandedEntityProcessed(string name)
+        {
```
mikepizzo: should have code comments and be moved.

is this even needed? why doesn't the calling code just do resource.MetadataBuilder.IsExpandedEntityProcessed(name)?

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -204,6 +210,19 @@ internal enum SelectionType
...
+        /// <summary>
+        /// Gets names of the expanded entity nodes at current level.
+        /// </summary>
+        internal IEnumerable<string> NamesOfSelectedExpandedEntities
+        {
+            get
+            {
+                return this.children == null
+                    ? Enumerable.Empty<string>()
+                    : this.children.Where(_ => _.Value.isExpandedNavigationProperty).Select(c => c.Key);
```
mikepizzo: >_ [](start = 42, length = 1)

I prefer to use a somewhat descriptive name, such as "node" or "childNode"

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -204,6 +210,19 @@ internal enum SelectionType
             PartialSubtree = 2,
         }
 
+        /// <summary>
+        /// Gets names of the expanded entity nodes at current level.
+        /// </summary>
+        internal IEnumerable<string> NamesOfSelectedExpandedEntities
```
mikepizzo: >Selected [](start = 44, length = 8)

This name implies that the function returns the names of selected as well as expanded entities.  I think it just returns the expanded entities.  Also, I don't think it adds much to put "NamesOf" in the name, and doing so is not a pattern we follow other places, so I'd just call this ExpandedEntities.

src/Microsoft.OData.Core/Evaluation/ODataConventionalResourceMetadataBuilder.cs
```diff
@@ -365,6 +368,26 @@ internal override void MarkNestedResourceInfoProcessed(string navigationProperty
...
+
+        /// <summary>
+        /// Return whether the given expanded entity of specified name has been processed.
+        /// </summary>
+        /// <param name="name">The name of the expanded entity.</param>
+        internal override bool IsExpandedEntityProcessed(string name)
+        {
+            return this.ProcessedExpandedEntityNames.Contains(name);
+        }
+
```
mikepizzo: Why not move this implementation to the ODataResourceMetadataBuilder as a (more useful) virtual implementation than a no-op for MarkExpandedEntityProcessed and returning false for IsExpandedEntityProcessed. A derived type can still override if they have alternate behavior.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
mikepizzo: Quite honestly, throwOnUndeclaredProperty is a carry-over from OData V3.  In OData V4 it should always be false which means this check should never have to occur.  Unfortunately, the default value is currently false, but we should consider changing it.

I still dislike having the behavior change based on the omit-nulls flag; having the same code throw an exception if omit-nulls is false and succeed if omit-nulls is true is unintuitive; for example, avoiding the duplicateNamesChecker means that the app could succeed in writing the same property twice, once as null and once not as null, with omit-nulls equal to true, and fail with omit-nulls equal to false. Omit-nulls should we a wire optimization, not a change in behavior.

All that said, it's a nice optimization, even if throwOnUndeclaredProperty is false, to avoid the duplicateNameChecker....

---
In reply to: [213868365](https://github.com/OData/odata.net/pull/1106#discussion_r213868365) [](ancestors = 213868365,213515156,209787942,198078564,187209557,186561856)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
mikepizzo: then should edmStructuredType != null be an assert, rather than a condition, on line 2287?

---
In reply to: [209787888](https://github.com/OData/odata.net/pull/1106#discussion_r209787888) [](ancestors = 209787888,198065705)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2265,107 @@ private void EndEntry()
...
+                        {
+                            Debug.Assert(currentProperty.Type != null, "currentProperty.Type != null");
+
+                            // Skip declared properties that are not null-able types.
+                            // Skip navigation properties (no navigation links need to be restored as null.)
+                            // Skip properties that have been read (primitive or structural).
+                            if (!currentProperty.Type.IsNullable
+                                || currentProperty.PropertyKind == EdmPropertyKind.Navigation
+                                || resource.Properties.Any(p => p.Name.Equals(currentProperty.Name, StringComparison.Ordinal))
+                                || resourceState.NavigationPropertiesRead.Contains(currentProperty.Name))
```
mikepizzo: >resourceState.NavigationPropertiesRead.Contains(currentProperty.Name) [](start = 35, length = 69)

if you already excluded navigation properties, why would you have to look in NavigationPropertiesRead?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2265,107 @@ private void EndEntry()
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Restore the null property value in the resource.
+        /// </summary>
+        /// <param name="nullPropertyName">Name of the property to restore the null value.</param>
+        /// <param name="resourceState">The resource state.</param>
+        private static void RestoreNullODataProperty(string nullPropertyName, IODataJsonLightReaderResourceState resourceState)
```
mikepizzo: >RestoreNullODataProperty [](start = 28, length = 24)

Seems inefficient to attempt a cast and do a concat for each individual property being restored.  Could we instead pass a collection of properties and add them in a single concatenation?

mikepizzo:
What is the behavior if we have annotations for a property but no value? Do we write the annotations but no value? What happens on restore; do we restore a null in that case?  What do we do without omit-nulls when we ready a property with annotations but no value? is the value of the ODataProperty null, rather than ODataNullValue?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
```diff
@@ -270,15 +270,27 @@ protected override void StartResource(ODataResource resource)
...
+                    this.jsonWriter.WriteName(parentNavLink.Name);
+                }
             }
 
             if (resource == null)
             {
-                this.jsonWriter.WriteValue((string)null);
+                if (!this.ShouldOmitNullValues())
+                {
+                    this.jsonWriter.WriteValue((string) null);
```
mikepizzo: Why not put line the WriteValue immediately after the WriteNames on lines 277 and remove lines 287-292?

biaol-odata:
If we have annotations for a property but no value, we write annotations but no value; and restore null on read. Without omit-nulls, read would do nothing w.r.t. null restoration (write does nothing either).
The property value of null or ODataNullValue are equivalent.

---
In reply to: [435924987](https://github.com/OData/odata.net/pull/1106#issuecomment-435924987) [](ancestors = 435924987)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
biaol-odata: sure, updated to Assert(edmStructuredType != null)

---
In reply to: [229534051](https://github.com/OData/odata.net/pull/1106#discussion_r229534051) [](ancestors = 229534051,209787888,198065705)

src/Microsoft.OData.Core/ODataResource.cs
biaol-odata: I thought having this convenient method on ODataResouce would save the indirection underneath and therefore would be beneficial for the user. 
But you are right that it is not strictly needed. Removed.

---
In reply to: [229462320](https://github.com/OData/odata.net/pull/1106#discussion_r229462320) [](ancestors = 229462320)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
biaol-odata: renamed.

---
In reply to: [229463175](https://github.com/OData/odata.net/pull/1106#discussion_r229463175) [](ancestors = 229463175)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
biaol-odata: renamed to SelectedExpandedEntities. What we need is the expanded entities' names from the keys, instead from the value objects.

---
In reply to: [229464026](https://github.com/OData/odata.net/pull/1106#discussion_r229464026) [](ancestors = 229464026)

src/Microsoft.OData.Core/Evaluation/ODataConventionalResourceMetadataBuilder.cs
biaol-odata: sure, moved.

---
In reply to: [229530080](https://github.com/OData/odata.net/pull/1106#discussion_r229530080) [](ancestors = 229530080)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
biaol-odata: As discussed, we need to check NavigationPropertiesRead for structural property (which is not nav link) in order to skip its null restoration as required.

---
In reply to: [229536601](https://github.com/OData/odata.net/pull/1106#discussion_r229536601) [](ancestors = 229536601)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
biaol-odata: The original ReadOnlyEnumerable only supports adding an item to the underneath list. I just expand the API for adding a list of items, so that we can add list of items once for all from here.
Now we pass a collection and add all of the items at once.

---
In reply to: [230799899](https://github.com/OData/odata.net/pull/1106#discussion_r230799899) [](ancestors = 230799899)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
biaol-odata: As we discussed, line 287-292 is still needed for case of parentNavLink == null.
Per your suggestion for optimization, added the writeValue + return after Line 27, as tradeoff between optimization and non-duplicated code.

---
In reply to: [230806136](https://github.com/OData/odata.net/pull/1106#discussion_r230806136) [](ancestors = 230806136)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2256,6 +2265,107 @@ private void EndEntry()
...
+                    {
+                        ODataResourceBase resource = resourceState.Resource;
+
+                        IEnumerable<IEdmProperty> selectedProperties = resourceState.SelectedProperties.GetSelectedProperties(edmStructuredType);
+
+                        // All dynamic properties: null value restoration for properties that are not present.
+                        foreach (string name in resourceState.SelectedProperties.GetSelectedDynamicProperties(edmStructuredType))
+                        {
+                            if (resource.Properties.Any(p => p.Name.Equals(name, StringComparison.Ordinal))
+                                || resourceState.NavigationPropertiesRead.Contains(name))
```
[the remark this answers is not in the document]
biaol-odata: Dynamic property is represented by string; while regular declared property is represented by IEdmProperty.
For declared property, we need to check the IEdmProperty.Type.IsNullable; while for dynamic property, we don't.
Combining declared properties and dynamic properties together would require using strings to represent both; and then for each name we would need to try resolving it back to IEdmProperty to check Type.IsNullable; additionally, as we discussed, internal processing for entire-subtree and empty-subtree are for declared properties only, we would have to perform an additional step to build the list of selected dynamic properties to be combined with selected declared properties. With all these, it would not be efficient if we combining both kinds.

Therefore I think we should keep processing of above two kinds separate, and it is cleaner when we separate different processing logic.

---
In reply to: [229536394](https://github.com/OData/odata.net/pull/1106#discussion_r229536394) [](ancestors = 229536394)

biaol-odata:
@mikepizzo 
Thanks for the latest round of review (primarily for $expand case of omit-null)! 
I have addressed the new comments. Please also follow up with @kosinsky about throwOnUndeclaredProperty optimization as we discussed, and I will do update as needed.

biaol-odata:
@mikepizzo I have made one more update for the optimization we discussed in office yesterday. Thanks for the suggestion.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
biaol-odata: As we discussed with kosinsky, we can ignore the optimization for validation. I have updated the code accordingly including removal of the omitNullValues argument from WriteNullProperty. Thanks.

---
In reply to: [229533703](https://github.com/OData/odata.net/pull/1106#discussion_r229533703) [](ancestors = 229533703,213868365,213515156,209787942,198078564,187209557,186561856)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -202,7 +211,18 @@ private bool IsOpenProperty(ODataProperty property)
 
             WriteInstanceAnnotation(property, isTopLevel, currentPropertyInfo.MetadataType.IsUndeclaredProperty);
 
-            ODataValue value = property.ODataValue;
+            WriterValidationUtils.ValidatePropertyDefined(this.currentPropertyInfo, this.MessageWriterSettings.ThrowOnUndeclaredPropertyForNonOpenType);
+
+            duplicatePropertyNameChecker.ValidatePropertyUniqueness(property);
+
```
mikepizzo: Not sure why these were moved? Can we put the two checks back before line 206 (before we write the type annotation for undeclared properties)?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -164,17 +170,23 @@ private bool IsOpenProperty(ODataProperty property)
...
             bool isTopLevel,
             bool allowStreamProperty,
+            bool omitNullValues,
             IDuplicatePropertyNameChecker duplicatePropertyNameChecker)
         {
             WriterValidationUtils.ValidatePropertyNotNull(property);
 
+            ODataValue value = property.ODataValue;
+
             string propertyName = property.Name;
```
mikepizzo: To reduce unnecessary changes, can we move this back to after writing the instance annotation on line 213?

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -441,11 +461,131 @@ internal IEnumerable<IEdmNavigationProperty> GetSelectedNavigationProperties(IEd
...
+            {
+                dynamicProperties = this.selectedProperties.Select(
+                    prop => structuredType.FindProperty(prop) == null
+                    ? prop
+                    : null)
+                    .OfType<string>();
+            }
+
+            return dynamicProperties;
+        }
```
mikepizzo: These two methods should be deleted if they are no longer used.

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -441,11 +461,131 @@ internal IEnumerable<IEdmNavigationProperty> GetSelectedNavigationProperties(IEd
...
+
+        /// <summary>
+        /// Gets the list of names for selected declared or dynamic properties having values of null.
```
mikepizzo: > Gets the list of names for selected declared or dynamic properties having values of null. [](start = 11, length = 90)

Does this return the names of the properties having null values, or the names of (nullable) properties missing from the payload?

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -441,11 +461,131 @@ internal IEnumerable<IEdmNavigationProperty> GetSelectedNavigationProperties(IEd
...
+        /// </summary>
+        /// <param name="structuredType">The current structured type.</param>
+        /// <param name="resourceState">The current resource state.</param>
+        /// <returns>The list of property names with null values.</returns>
+        internal IList<string> GetSelectedNullValueProperties(IEdmStructuredType structuredType,
```
mikepizzo: >GetSelectedNullValueProperties [](start = 31, length = 30)

We are doing a lot of the same work (i.e., determining the set of nullable properties) for each resource in the result.  I wonder if it wouldn't be more efficient to determine the set of nullable property names up front, and then, at the end of writing each resource, restore nulls for the properties from that static list that have not been read.

mikepizzo:
I like the new code much better.  See my comments inline:
1) it looks like the old methods, that separately read declared and dynamic, are still there and should be removed?
2) rather than determine the list of nullable properties each time we read a resource, can we do it once for the result, and then compare what we've read with that static list of nullable properties?

---
In reply to: [436729775](https://github.com/OData/odata.net/pull/1106#issuecomment-436729775) [](ancestors = 436729775)

mikepizzo:
As per separate thread, based on clarification from OASIS, we should not omit annotated values that are null, and should not restore missing values for annotated properties.

---
In reply to: [436424268](https://github.com/OData/odata.net/pull/1106#issuecomment-436424268) [](ancestors = 436424268,435924987)

src/Microsoft.OData.Core/Evaluation/ODataConventionalResourceMetadataBuilder.cs
mikepizzo: The implementation in ODataResourceMetadataBuilder just always returns false, and MarkExpandedEntityProcessed is a no-op. Why not put this actual implementation as the virtual implementation in ODataResourceMetadataBuilder, and not override at all in ODataConventionalResourceMetadataBuilder?

---
In reply to: [231309971](https://github.com/OData/odata.net/pull/1106#discussion_r231309971) [](ancestors = 231309971,229530080)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
biaol-odata: Moved back to before L206. Thanks.

---
In reply to: [233183156](https://github.com/OData/odata.net/pull/1106#discussion_r233183156) [](ancestors = 233183156)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
biaol-odata: sure. done.

---
In reply to: [233183812](https://github.com/OData/odata.net/pull/1106#discussion_r233183812) [](ancestors = 233183812)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
biaol-odata: cleaned up GetSelectedDynamicProperties.
GetSelectedProperties is still used by the test w/o using resource state.

---
In reply to: [233187531](https://github.com/OData/odata.net/pull/1106#discussion_r233187531) [](ancestors = 233187531)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
biaol-odata: Updated. Should be the names of properties missing from payload, which will be restored to null. 

---
In reply to: [233188777](https://github.com/OData/odata.net/pull/1106#discussion_r233188777) [](ancestors = 233188777)

biaol-odata:
The changes related to excluding annotated values is in a forked branch I have shared with you. Will be integrated into this PR, pending on double-checking with detailed requirements as we discussed.

---
In reply to: [438406870](https://github.com/OData/odata.net/pull/1106#issuecomment-438406870) [](ancestors = 438406870,436424268,435924987)

biaol-odata:
@mikepizzo Please see the latest updates per your comments today. Currently there is one issue pending related to the annotated properties skipping omit-nulls, which will be merged into this PR when the detailed requirement is finalized. Thanks.

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
biaol-odata: The current logic follows the overall design of ODataJsonLightReader's EndEntry() method and has been the design from the beginning. It restores all missing nulls for the current entry all at once when the entry is deserialized. If we restored at the end of writing each resource, it would still need to perform the same restoration of null values, and with additional cost of caching information for null properties. Therefore doing it in flight is more efficient.

---
In reply to: [233190829](https://github.com/OData/odata.net/pull/1106#discussion_r233190829) [](ancestors = 233190829)

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
mikepizzo: Note that "navigationLink" is still used for nestedResourceInfo in a number of places

---
In reply to: [173003599](https://github.com/OData/odata.net/pull/1106#discussion_r173003599) [](ancestors = 173003599)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
mikepizzo: I understand that we are restoring null values at the end -- that's fine.  What I am questioning is whether we have to identify the set of nullable properties each time, or if we can do that once for the result.

---
In reply to: [233299115](https://github.com/OData/odata.net/pull/1106#discussion_r233299115) [](ancestors = 233299115,233190829)

biaol-odata: Since SelectedPropertiesNode is designed as part of context url parsing and is used to represent relatively static context url information, I am not sure it is proper for SelectedPropertiesNode to cache null properties, which is related to runtime payload and is not intrinsic part of SelectedPropertiesNode.
Additionally, there are no 'recalculating' per se; the calculations of set of omitted null properties are required for each structured property or entry on different level regardless whether restoration is done in-flight or via caching, and restoring it in-flight can save the cost of caching and fits with the design of the reader.

---
In reply to: [233553959](https://github.com/OData/odata.net/pull/1106#discussion_r233553959) [](ancestors = 233553959,233299115,233190829)

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
biaol-odata: I think you are looking at some original test cases that are not changed by this PR. Since you have spotted them, and it makes sense to renaming it, I have updated the names to 'nestedResourceInfo'.

---
In reply to: [233550345](https://github.com/OData/odata.net/pull/1106#discussion_r233550345) [](ancestors = 233550345,173003599)

biaol-odata:
@mikepizzo I have updated the names to 'nestedResourceInfo'. They are from some original test cases that are not changed by this PR but it makes sense to update them. I have also provided my thoughts on your question related to caching null properties during restoration. Thanks.

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
mikepizzo: This comment seems wrong. Dynamic properties do show up in ContextUrl if explicitly selected.

---
In reply to: [174266340](https://github.com/OData/odata.net/pull/1106#discussion_r174266340) [](ancestors = 174266340,173008510)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
mikepizzo: I don't like having methods (even internal) that are only used by tests. Is there another way the test can get the selected properties it is validating?

---
In reply to: [233297316](https://github.com/OData/odata.net/pull/1106#discussion_r233297316) [](ancestors = 233297316,233187531)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
mikepizzo: Again, I'm interested in whether we can avoid determining the set of nullable properties in the SelectedPropertiesNode each time, not evaluating what the set of actual null property values is for each resource.

---
In reply to: [233666008](https://github.com/OData/odata.net/pull/1106#discussion_r233666008) [](ancestors = 233666008,233553959,233299115,233190829)

mikepizzo:
I don't see fd97a4455c as part of this PR.

---
In reply to: [438516788](https://github.com/OData/odata.net/pull/1106#issuecomment-438516788) [](ancestors = 438516788,438406870,436424268,435924987)

agrabhi:
@mikepizzo @biaol-odata @xuzhg 
we need above feature for one of above api, because many properties tend to be null there. It seems like this PR has not been touched for a while. What can be done to get this PR checked in

agrabhi:
@mikepizzo @biaol-odata @xuzhg 
Is this PR on radar for checkin soon?

PR #1110, opened 2018-03-09
[Feature] Set-based operations

AlanWong-MS:
### Description
Implementation of the $filter and $each path segments as part of the [set-based operation](http://docs.oasis-open.org/odata/new-in-odata/v4.01/cn01/new-in-odata-v4.01-cn01.html#_Toc485385072) in the OData v4.01 specification ([ANBF](http://docs.oasis-open.org/odata/odata/v4.01/csprd04/abnf/)).

This PR includes the $filter, $ref, and $each path segment features, which include three new classes (FilterSegment, ReferenceSegment, and EachSegment), modifications to the URI parser to process the new path segments, unit tests, and functional tests. A new $ref is created to address a previous bug where the segment was missing scenarios such as TypeSegments, entity sets, and was not compatible with the new $filter path segment.

Other changes in this PR include cleaning up StyleCop issues.

- [x] *Test cases added*


AlanWong-MS:
Ready for review now

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -450,80 +454,222 @@ private bool TryCreateCountSegment(string segmentText)
...
+            return filterBinder.BindFilter(filterToken);
+        }
+
+        /// <summary>
+        /// Try to handle the segment as $filter.
+        /// </summary>
+        /// <param name="segmentText">The raw segment text.</param>
+        /// <returns>Whether the segment was $filter.</returns>
+        /// <remarks>$filter path segment is different from existing path segments in that it strictly
+        /// follows the format of "$filter(@a)", where @a represents a parameter alias. Thus, this function
```
mikepizzo: >@a [](start = 43, length = 2)

should now be $filter(expression) where expression could be an alias or inline expression.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -450,80 +454,222 @@ private bool TryCreateCountSegment(string segmentText)
...
+            int index = UriQueryConstants.FilterSegment.Length;
+            if (segmentText.Length <= index + 2 || segmentText[index] != '(' || segmentText[segmentText.Length - 1] != ')')
             {
-                throw ExceptionUtil.CreateSyntaxError();
+                throw new ODataException(ODataErrorStrings.RequestUriProcessor_FilterPathSegmentSyntaxError);
             }
 
-            // Create a stack to keep track of KeySegments
-            Stack<KeySegment> keySegmentsForPreviousPathSegment = new Stack<KeySegment>();
+            // 3) Extract the parameter alias and perform the rest of the validations on it.
```
mikepizzo: >parameter alias [](start = 30, length = 15)

expression

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -716,6 +860,18 @@ private void CreateFirstSegment(string segmentText)
...
+                // $filter on root: throw
+                throw ExceptionUtil.ResourceNotFoundError(ODataErrorStrings.RequestUriProcessor_FilterOnRoot);
+            }
+
+            if (this.IdentifierIs(UriQueryConstants.EachSegment, identifier))
+            {
+                // $each on root: throw
+                throw ExceptionUtil.ResourceNotFoundError(ODataErrorStrings.RequestUriProcessor_EachOnRoot);
+            }
+
```
mikepizzo: should you also check this.IdentifierIs(UriQueryConstants.RefSegment, identifier) here?

src/Microsoft.OData.Core/UriParser/SemanticAst/EachSegment.cs
```diff
@@ -0,0 +1,90 @@
...
+        /// </summary>
+        private readonly IEdmType edmType;
+
+        /// <summary>
+        /// Build a segment representing each.
+        /// </summary>
+        /// <param name="navigationSource">The entity collection that this set-based operation applies to.</param>
+        /// <param name="targetEdmType">Target type for the entity being referenced.</param>
+        /// <exception cref="System.ArgumentNullException">Throws if any input parameter is null.</exception>
+        /// <remarks>$filter cannot be applied on singletons.</remarks>
```
mikepizzo: >filter [](start = 22, length = 6)

$each

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
AlanWong-MS: Thanks for the catch!

src/Microsoft.OData.Core/UriParser/SemanticAst/EachSegment.cs
```diff
@@ -0,0 +1,90 @@
...
+        public EachSegment(IEdmNavigationSource navigationSource, IEdmType targetEdmType)
```
mikepizzo: >IEdmType targetEdmType [](start = 66, length = 22)

when would targetEdmType be different from navigationSource.Type?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
AlanWong-MS: :+1:

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
AlanWong-MS: Yep, thanks for catching that.

src/Microsoft.OData.Core/UriParser/SemanticAst/EachSegment.cs
AlanWong-MS: :+1:

src/Microsoft.OData.Core/UriParser/SemanticAst/EachSegment.cs
```diff
@@ -0,0 +1,90 @@
...
+        {
+            ExceptionUtils.CheckArgumentNotNull(navigationSource, "navigationSource");
+
+            this.Identifier = UriQueryConstants.EachSegment;
+            this.SingleResult = false;
+            this.TargetEdmNavigationSource = navigationSource;
+            this.TargetKind = RequestTargetKind.Resource;
+            this.TargetEdmType = targetEdmType ?? navigationSource.EntityType();
```
mikepizzo: >navigationSource.EntityType() [](start = 50, length = 29)

why navigationSource.EntityType() rather than navigationSource.Type?

src/Microsoft.OData.Core/UriParser/SemanticAst/ReferenceSegment.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// Build a segment representing $ref.
+        /// </summary>
+        /// <param name="navigationSource">The navigation source to which this segment references.</param>
+        /// <exception cref="System.ArgumentNullException">Throws if any input parameter is null.</exception>
+        public ReferenceSegment(IEdmNavigationSource navigationSource)
+        {
+            ExceptionUtils.CheckArgumentNotNull(navigationSource, "navigationSource");
+
+            this.Identifier = UriQueryConstants.RefSegment;
+            this.SingleResult = false;
```
mikepizzo: >this.SingleResult = false; [](start = 12, length = 26)

what does setting this to false mean?  Can I still do Employees(1)/Manager/$ref?

src/Microsoft.OData.Core/UriParser/Visitors/IsCollectionTranslator.cs
```diff
@@ -98,6 +98,42 @@ public override bool Translate(CountSegment segment)
...
+        /// <summary>
+        /// Translate a ReferenceSegment
+        /// </summary>
+        /// <param name="segment">the segment to Translate</param>
+        /// <returns>UserDefinedValue</returns>
+        /// <exception cref="System.ArgumentNullException">Throws if the input segment is null.</exception>
+        public override bool Translate(ReferenceSegment segment)
+        {
+            ExceptionUtils.CheckArgumentNotNull(segment, "segment");
+            return false;
```
mikepizzo: >return false; [](start = 12, length = 13)

Wouldn't this depend on whether the reference segment was applied to a collection (i.e., Employees(1)/ns.Manager/Employees/$ref) or a single value (i.e., Employees(1)/Manager/$ref)?

src/Microsoft.OData.Core/UriParser/Visitors/IsCollectionTranslator.cs
```diff
@@ -157,5 +193,17 @@ public override bool Translate(MetadataSegment segment)
...
+        /// <summary>
+        /// Translate a PathTemplateSegment
+        /// </summary>
+        /// <param name="segment">the segment to Translate</param>
+        /// <returns>UserDefinedValue</returns>
+        /// <exception cref="System.ArgumentNullException">Throws if the input segment is null.</exception>
+        public override bool Translate(PathTemplateSegment segment)
+        {
+            ExceptionUtils.CheckArgumentNotNull(segment, "segment");
+            return false;
```
mikepizzo: >return false [](start = 12, length = 12)

maybe we should return segment.SingleResult (in case we someday figure out how to determine whether the path template is single- or multi-valued.)

src/Microsoft.OData.Core/UriParser/Visitors/PathSegmentToContextUrlPathTranslator.cs
```diff
@@ -180,6 +180,36 @@ public override string Translate(CountSegment segment)
...
         }
 
+        /// <summary>
+        /// Translate a FilterSegment
+        /// </summary>
+        /// <param name="segment">the segment to Translate</param>
+        /// <returns>Defined by the implementer.</returns>
+        public override string Translate(FilterSegment segment)
+        {
+            return "/" + segment.FullSegment;
```
mikepizzo: >return "/" + segment.FullSegment; [](start = 12, length = 33)

Filter segment should not appear in contextUrl; just return string.Empty.

src/Microsoft.OData.Core/UriParser/Visitors/PathSegmentToContextUrlPathTranslator.cs
```diff
@@ -180,6 +180,36 @@ public override string Translate(CountSegment segment)
...
+        }
+
+        /// <summary>
+        /// Translate a ReferenceSegment
+        /// </summary>
+        /// <param name="segment">the segment to Translate</param>
+        /// <returns>Defined by the implementer.</returns>
+        public override string Translate(ReferenceSegment segment)
+        {
+            return string.Empty;
```
mikepizzo: This actually came up in our TC meeting today.  I'm unclear what we should do in a ContextUrl for a reference segment. Please open an issue to track, and reference it from the code here.

mikepizzo:
>     internal sealed class SplitEndingSegmentOfTypeHandler<T> : PathSegmentHandler where T : ODataPathSegment

should this handle EachSegment, ReferenceSegment, and PathTemplateSegment?

---
Refers to: src/Microsoft.OData.Core/UriParser/Visitors/SplitEndingSegmentOfTypeHandler.cs:18 in a9bc40c. [](commit_id = a9bc40c1c134f32a398ec5da28c12cb5ee8db877, deletion_comment = False)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/PathBuilderTests.cs
```diff
@@ -126,6 +126,30 @@ public void BuildPathWithEscapeMarkerWithTypeSegmentInKeyAsSegment()
...
+        [Fact]
+        public void BuildPathWithEachSegmentRelativeUri()
+        {
+            Uri queryUri = new Uri("People/$each/Fully.Qualified.Namespace.SummonPuppies", UriKind.Relative);
+
+            Uri actualUri = UriBuilder(queryUri, ODataUrlKeyDelimiter.Parentheses, settings);
+            Assert.Equal(new Uri("http://gobbledygook/People/$each/Fully.Qualified.Namespace.SummonPuppies"), actualUri);
+
+            actualUri = UriBuilder(queryUri, ODataUrlKeyDelimiter.Slash, settings);
+            Assert.Equal(new Uri("http://gobbledygook/People/$each/Fully.Qualified.Namespace.SummonPuppies"), actualUri);
```
mikepizzo: >new Uri("http://gobbledygook/People/$each/Fully.Qualified.Namespace.SummonPuppies" [](start = 25, length = 82)

just define this once and use it in both asserts

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/UrlBuilderWithParameterAliasTests.cs
```diff
@@ -394,6 +394,46 @@ public void BuildExpandOrderby_AliasInFunction()
...
+            ODataUri odataUri = odataUriParser.ParseUri();
+
+            odataUri.Filter.Should().BeNull();
+            odataUri.ParameterAliasNodes["@p1"].ShouldBeConstantQueryNode(true);
+
+            Uri actualUri = odataUri.BuildUri(ODataUrlKeyDelimiter.Parentheses);
+            Assert.Equal(new Uri("http://gobbledygook/People/$filter=@p1?@p1=true"), actualUri);
+
+            actualUri = odataUri.BuildUri(ODataUrlKeyDelimiter.Slash);
+            Assert.Equal(new Uri("http://gobbledygook/People/$filter=@p1?@p1=true"), actualUri);
```
mikepizzo: >new Uri("http://gobbledygook/People/$filter=@p1?@p1=true" [](start = 25, length = 57)

just define once and use in both asserts.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/UrlBuilderWithParameterAliasTests.cs
```diff
@@ -394,6 +394,46 @@ public void BuildExpandOrderby_AliasInFunction()
...
+        [Fact]
+        public void BuildFullUri_AliasInFilterPathSegment_AliasAsBoolean()
+        {
+            Uri fullUri = new Uri("http://gobbledygook/People/$filter=@p1?@p1=true");
+            ODataUriParser odataUriParser = new ODataUriParser(HardCodedTestModel.TestModel, serviceRoot, fullUri);
+            SetODataUriParserSettingsTo(this.settings, odataUriParser.Settings);
+            odataUriParser.UrlKeyDelimiter = ODataUrlKeyDelimiter.Parentheses;
+            ODataUri odataUri = odataUriParser.ParseUri();
+
+            odataUri.Filter.Should().BeNull();
```
mikepizzo: >odataUri [](start = 12, length = 8)

do you want to add a check that the expected filter segment is present in the odataUri?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/UrlBuilderWithParameterAliasTests.cs
```diff
@@ -394,6 +394,46 @@ public void BuildExpandOrderby_AliasInFunction()
...
+        [Fact]
+        public void BuildRelativeUri_AliasInFilterPathSegment_AliasAsBoolean()
+        {
+            Uri relativeUri = new Uri("People/$filter=@p1?@p1=true", UriKind.Relative);
+            ODataUriParser odataUriParser = new ODataUriParser(HardCodedTestModel.TestModel, serviceRoot, relativeUri);
+            SetODataUriParserSettingsTo(this.settings, odataUriParser.Settings);
+            odataUriParser.UrlKeyDelimiter = ODataUrlKeyDelimiter.Parentheses;
+            ODataUri odataUri = odataUriParser.ParseUri();
+
+            odataUri.Filter.Should().BeNull();
```
mikepizzo: >odataUri [](start = 12, length = 8)

do you want to add a check that the expected filter segment is present in the odataUri?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/UrlBuilderWithParameterAliasTests.cs
```diff
@@ -394,6 +394,46 @@ public void BuildExpandOrderby_AliasInFunction()
         }
         #endregion
 
+        #region alias in path segment
+        [Fact]
+        public void BuildFullUri_AliasInFilterPathSegment_AliasAsBoolean()
+        {
+            Uri fullUri = new Uri("http://gobbledygook/People/$filter=@p1?@p1=true");
```
mikepizzo: >$filter=@p1 [](start = 62, length = 11)

this should now be $filter(@p1)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/UrlBuilderWithParameterAliasTests.cs
```diff
@@ -394,6 +394,46 @@ public void BuildExpandOrderby_AliasInFunction()
...
+            Assert.Equal(fullUri, actualUri);
+
+            actualUri = odataUri.BuildUri(ODataUrlKeyDelimiter.Slash);
+            Assert.Equal(fullUri, actualUri);
+        }
+
+        [Fact]
+        public void BuildRelativeUri_AliasInFilterPathSegment_AliasAsBoolean()
+        {
+            Uri relativeUri = new Uri("People/$filter=@p1?@p1=true", UriKind.Relative);
```
mikepizzo: >$filter=@p1 [](start = 46, length = 11)

should now be $filter(@p1)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/UrlBuilderWithParameterAliasTests.cs
```diff
@@ -394,6 +394,46 @@ public void BuildExpandOrderby_AliasInFunction()
         }
         #endregion
 
+        #region alias in path segment
+        [Fact]
+        public void BuildFullUri_AliasInFilterPathSegment_AliasAsBoolean()
```
mikepizzo: >BuildFullUri_AliasInFilterPathSegment_AliasAsBoolean [](start = 20, length = 52)

Can you add UriBuilder tests for expression in $filter segment (rather than alias). i.e., [http://gobbledygook/People/$filter(age](http://gobbledygook/People/$filter(age) ge 18)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EntityReferenceFunctionalTests.cs
```diff
@@ -26,6 +26,25 @@ public void UseMultipleEscapeSequencesWithRefInKeyAsSegment()
...
+        public void EntityReferenceCannotAppearAfterMetadata()
+        {
+            PathFunctionalTestsUtil.RunParseErrorPath("$metadata/$ref", ODataErrorStrings.RequestUriProcessor_MustBeLeafSegment("$metadata"));
+        }
+
+        [Fact]
+        public void EntityReferenceCannotAppearAfterSingleton()
+        {
+            PathFunctionalTestsUtil.RunParseErrorPath("Boss/$ref", ODataErrorStrings.PathParser_EntityReferenceNotSupported("Boss"));
+        }
```
mikepizzo: This is wrong; you should be able to have /$ref after a singleton...

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EntityReferenceFunctionalTests.cs
```diff
@@ -38,17 +57,36 @@ public void KeyLookupCannotAppearAfterEntityReference()
...
+
         [Fact]
         public void CountCannotAppearAfterEntityReferenceCollectionProperties()
         {
             PathFunctionalTestsUtil.RunParseErrorPath("Dogs(1)/MyPeople/$ref/$count", ODataErrorStrings.RequestUriProcessor_MustBeLeafSegment(UriQueryConstants.RefSegment));
         }
 
         [Fact]
-        public void EntityReferenceCannotAppearAfterAnEntitySet()
+        public void EntityReferenceCanAppearAfterAnEntitySet()
```
mikepizzo: >EntityReferenceCanAppearAfterAnEntitySet [](start = 20, length = 40)

please add EntityReferenceCanAppearAfterACollectionValuedNavigationProperty and EntityReferenceCanAppearAfterASingleValuedNavigationProperty.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EntityReferenceFunctionalTests.cs
```diff
@@ -57,6 +95,32 @@ public void EntityReferenceCannotAppearAfterAComplexProperty()
...
+
+        [Fact]
+        public void EntityReferenceCanAppearAfterCastedType()
+        {
+            var path = PathFunctionalTestsUtil.RunParsePath("People/Fully.Qualified.Namespace.Employee/$ref");
+            path.LastSegment.ShouldBeReferenceSegment(HardCodedTestModel.GetPeopleSet());
+        }
+
+        [Fact]
+        public void EntityReferenceCanAppearAfterBoundFunctionReturningCollection()
```
mikepizzo: >EntityReferenceCanAppearAfterBoundFunctionReturningCollection [](start = 20, length = 61)

please add EntityReferenceCanAppearAfterBoundFunctionReturningSingleEntity()

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EntityReferenceFunctionalTests.cs
```diff
@@ -38,17 +57,36 @@ public void KeyLookupCannotAppearAfterEntityReference()
...
+        public void EntityReferenceCanAppearAfterAnEntitySet()
         {
-            // TODO: We can improve error message drastically when we refactor path parsing
-            PathFunctionalTestsUtil.RunParseErrorPath("People/$ref", ODataErrorStrings.PathParser_EntityReferenceNotSupported("People"));
+            var path = PathFunctionalTestsUtil.RunParsePath("People/$ref");
+            path.LastSegment.ShouldBeReferenceSegment(HardCodedTestModel.GetPeopleSet());
+        }
+
+        [Fact]
+        public void EntityReferenceCanAppearAfterAFilteredEntitySet()
```
mikepizzo: >EntityReferenceCanAppearAfterAFilteredEntitySet [](start = 20, length = 47)

please add EntityReferenceCannotAppearAfterAnEachSegment()

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/PathFunctionalTests.cs
```diff
@@ -796,6 +796,20 @@ public void KeyLookupCannotAppearAfterBatchReference()
             parse.ShouldThrow<ODataException>().WithMessage(ODataErrorStrings.RequestUriProcessor_SyntaxError);
         }
 
+        [Fact]
+        public void EntityReferenceCannotAppearAfterBatchReference()
+        {
+            ODataUriParser parser = new ODataUriParser(HardCodedTestModel.TestModel, new Uri("http://gobbldygook/"), new Uri("http://gobbldygook/$42/$ref"));
```
mikepizzo: >http://gobbldygook/$42/$ref [](start = 126, length = 27)

I think that should work...

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/SetBasedOperationTests.cs
```diff
@@ -0,0 +1,420 @@
...
+                    filterClause.Should().BeNull();
+                    aliasNodes["@p1"].ShouldBeBinaryOperatorNode(BinaryOperatorKind.Equal).And.Right.ShouldBeConstantQueryNode(42);
+
+                    EachSegment eachSegment = oDataPath.Last() as EachSegment;
+                    eachSegment.Should().NotBeNull();
+                    eachSegment.TargetEdmNavigationSource.Should().Be(HardCodedTestModel.GetPeopleSet());
+                });
+        }
+
+        [Fact]
```
mikepizzo: >[Fact] [](start = 8, length = 6)

maybe add a comment on what this means?  this would be used, i.e., in a PATCH to update all people with id eq 42, and return the updated people whose SSN was equal to 'num'

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/SetBasedOperationTests.cs
```diff
@@ -0,0 +1,420 @@
...
+        {
+            Action parse = () => ParseUriAndVerify(
+                new Uri("http://gobbledygook/People/$each/Fully.Qualified.Namespace.SummonPuppies/$each"),
+                (oDataPath, filterClause, aliasNodes) =>
+                {
+                });
+            parse.ShouldThrow<ODataException>().WithMessage(ODataErrorStrings.RequestUriProcessor_MustBeLeafSegment("Fully.Qualified.Namespace.SummonPuppies"));
+        }
+
+        // NOTE: Per OData 4.01 spec, GET operation and functions may proceed $each, but we are limiting the scope of that feature
```
mikepizzo: >proceed [](start = 70, length = 7)

follow?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/SetBasedOperationTests.cs
```diff
@@ -0,0 +1,420 @@
...
+        {
+            Action parse = () => ParseUriAndVerify(
+                new Uri("http://gobbledygook/People/$each/Fully.Qualified.Namespace.AllMyFriendsDogs"),
+                (oDataPath, filterClause, aliasNodes) =>
+                {
+                });
+            parse.ShouldThrow<ODataException>().WithMessage(ODataErrorStrings.RequestUriProcessor_OnlySingleActionCanProceedEachPathSegment);
+        }
+
+        // NOTE: Per OData 4.01 spec, GET operation and functions may proceed $each, but we are limiting the scope of that feature
```
mikepizzo: >proceed [](start = 70, length = 7)

follow?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/UriTemplateParserTests.cs
```diff
@@ -247,6 +247,18 @@ public void ParsePathTemplateSegmentWithTemplateParser()
             templateSegment.LiteralText.Should().Be("{some}");
         }
 
+        [Fact]
+        public void EntityReferenceCannotAppearAfterAPathTemplateSegment()
```
mikepizzo: >EntityReferenceCannotAppearAfterAPathTemplateSegment [](start = 20, length = 52)

why?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/SemanticAst/ReferenceSegmentTests.cs
```diff
@@ -0,0 +1,57 @@
...
+        [Fact]
+        public void ReferenceSegmentWithNavigationSourceConstructsSuccessfully()
+        {
+            ReferenceSegment referenceSegment = new ReferenceSegment(HardCodedTestModel.GetPet1Set());
+
+            referenceSegment.Identifier.Should().Be(UriQueryConstants.RefSegment);
+            referenceSegment.SingleResult.Should().BeFalse();
+            referenceSegment.TargetEdmNavigationSource.Should().Be(HardCodedTestModel.GetPet1Set());
+            referenceSegment.TargetKind.Should().Be(RequestTargetKind.Resource);
+            referenceSegment.EdmType.Should().BeNull();
```
mikepizzo: >referenceSegment [](start = 12, length = 16)

add validation that (in this case) referenceSegment.SingleResult.Should().BeTrue().  Add another test case with a single-valued nav source where referenceSegment.SingleResult.Should().BeFalse().

src/Microsoft.OData.Core/UriParser/SemanticAst/FilterSegment.cs
```diff
@@ -0,0 +1,150 @@
...
+        /// </summary>
+        public SingleValueNode Expression
+        {
+            get { return this.expression; }
+        }
+
+        /// <summary>
+        /// Gets the string representing "$filter(expression)".
+        /// </summary>
+        public string FullSegment
```
mikepizzo: >FullSegment [](start = 22, length = 11)

Should this be "LiteralText" (similar to PathTemplateSegment)?  Should it be public (versus used for internally, i.e. for serialization to a string)?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -450,80 +454,222 @@ private bool TryCreateCountSegment(string segmentText)
...
+            // KeySegments or before the $ref (i.e. NavPropSeg/KeySeg/KeySeg/.../$ref or NavPropSeg/$ref). Then the NavPropSeg is replaced
+            // with NavigationPropertyLinkSegment. In a previous implementation, if either a NavPropSeg didn't exist before KeySegments or
+            // $ref itself, then this function would throw. This was not correct as $ref can apply to entity sets and similar segments
+            // (e.g. TypeSegment and FilterSegment), and therefore 2) below is implemented to support those scenarios.
+            // 2) If the NavPropSeg does not exist, then this algorithm appends a ReferenceSegment to the existing list of parsed segments.
 
-            // Add this segment to parsedSegments
-            this.parsedSegments.Add(navPropLinkSegment);
+            // Determine the index of the NavigationPropertySegment in either of the following formats:
+            // NavPropSeg/KeySeg/KeySeg/.../$ref or NavPropSeg/$ref
```
mikepizzo: Can we remove the code to get the last navigation property segment, since above we added code to keep track of the lastNavigationSource?

src/Microsoft.OData.Core/UriParser/SemanticAst/EachSegment.cs
AlanWong-MS: I don't have the in-depth reason off the top of my head but from previous tests, I found that they're different when the previous segment is either a TypeSegment or FilterSegment.

```
EachSegment eachSegment = new EachSegment(
                lastNavigationSource,
                prevSegment is TypeSegment || prevSegment is FilterSegment ? prevSegment.TargetEdmType : null);
```

This is in the ODataPathParser.cs. Please let me know if we should dig deeper on whether we should change this design.

src/Microsoft.OData.Core/UriParser/SemanticAst/EachSegment.cs
AlanWong-MS: navigationSource.EntityType() helps resolve to the target EdmType while navigationSource.Type resolves to the collection of the type.

navigationSource.EntityType():
TargetEdmType	{Fully.Qualified.Namespace.Person}	Microsoft.OData.Edm.IEdmType {Microsoft.OData.Edm.Csdl.CsdlSemantics.CsdlSemanticsEntityTypeDefinition}

navigationSource.Type:
Type	{Collection([Fully.Qualified.Namespace.Person Nullable=False])}	Microsoft.OData.Edm.IEdmType {Microsoft.OData.Edm.EdmCollectionType}

src/Microsoft.OData.Core/UriParser/SemanticAst/ReferenceSegment.cs
AlanWong-MS: From looking at the ODataPathParser, it looks like we throw if the source is a single result. I don't recall why we do that. If we decide that this is should not be the case, then I can re-write to accept single results.

```
if (this.parsedSegments.Last().TargetKind != RequestTargetKind.Resource || this.parsedSegments.Last().SingleResult)
{
  throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.PathParser_EntityReferenceNotSupported(this.parsedSegments.Last().Identifier));
}

ReferenceSegment referenceSegment = new ReferenceSegment(lastNavigationSource);
```

src/Microsoft.OData.Core/UriParser/Visitors/PathSegmentToContextUrlPathTranslator.cs
AlanWong-MS: Gotcha.

src/Microsoft.OData.Core/UriParser/Visitors/PathSegmentToContextUrlPathTranslator.cs
AlanWong-MS: Adding https://github.com/OData/odata.net/issues/1292 to code comments there.

src/Microsoft.OData.Core/UriParser/Visitors/IsCollectionTranslator.cs
AlanWong-MS: Changed to return `!segment.SingleResult;`

src/Microsoft.OData.Core/UriParser/Visitors/IsCollectionTranslator.cs
AlanWong-MS: Changed to `return !segment.SingleResult;`

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
AlanWong-MS: Resolved offline. Will keep as-is due to maintaining backward compat and actual need for last nav prop segment.

src/Microsoft.OData.Core/UriParser/SemanticAst/FilterSegment.cs
AlanWong-MS: Good idea. Changing to LiteralText and making it internal.

src/Microsoft.OData.Core/UriParser/SemanticAst/ReferenceSegment.cs
AlanWong-MS: Resolved offline; will allow single results and base SingleResult off of whether navigationSource is a collection.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/PathBuilderTests.cs
AlanWong-MS: :+1:

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/UrlBuilderWithParameterAliasTests.cs
AlanWong-MS: :+1:

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/UrlBuilderWithParameterAliasTests.cs
AlanWong-MS: :+1:

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/UrlBuilderWithParameterAliasTests.cs
AlanWong-MS: :+1:

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/ParameterAliasFunctionalTests.cs
```diff
@@ -337,6 +336,563 @@ public void ParseFilter_AliasInsideInOperator_OperandsAsExpressions()
...
+            ParseUriAndVerify(
+                new Uri("http://gobbledygook/People/$filter(ID eq 1)"),
+                (oDataPath, filterClause, orderByClause, selectExpandClause, aliasNodes) =>
+                {
+                    oDataPath.Count.Should().Be(2);
+
+                    List<FilterSegment> filterSegments = oDataPath.OfType<FilterSegment>().ToList();
+                    filterSegments.Count.Should().Be(1);
+
+                    filterSegments[0].Expression.ShouldBeBinaryOperatorNode(BinaryOperatorKind.Equal).And.Right.ShouldBeConstantQueryNode(1);
```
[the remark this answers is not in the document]
AlanWong-MS: :+1:

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/SetBasedOperationTests.cs
AlanWong-MS: :+1:

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/SetBasedOperationTests.cs
AlanWong-MS: :+1:

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/SetBasedOperationTests.cs
AlanWong-MS: :+1:

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EntityReferenceFunctionalTests.cs
AlanWong-MS: Corrected.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EntityReferenceFunctionalTests.cs
AlanWong-MS: Done

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/PathFunctionalTests.cs
AlanWong-MS: Yep, fixed it now.

AlanWong-MS:
> > ```
> > internal sealed class SplitEndingSegmentOfTypeHandler<T> : PathSegmentHandler where T : ODataPathSegment
> > ```
> 
> should this handle EachSegment, ReferenceSegment, and PathTemplateSegment?
> 
> Refers to: src/Microsoft.OData.Core/UriParser/Visitors/SplitEndingSegmentOfTypeHandler.cs:18 in [a9bc40c](https://github.com/OData/odata.net/commit/a9bc40c1c134f32a398ec5da28c12cb5ee8db877). [](commit_id = [a9bc40c](https://github.com/OData/odata.net/commit/a9bc40c1c134f32a398ec5da28c12cb5ee8db877), deletion_comment = False)

Added

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/UriTemplateParserTests.cs
AlanWong-MS: I'm not too familiar with the PathTemplateSegment. This exception is thrown before we get to parsing $ref:

Microsoft.OData.UriParser.ODataUnrecognizedPathException: 'The request URI is not valid. The segment '{some}' must be the last segment in the URI because it is one of the following: $ref, $batch, $count, $value, $metadata, a named media resource, an action, a noncomposable function, an action import, a noncomposable function import, an operation with void return type, or an operation import with void return type.'

src/Microsoft.OData.Core/UriParser/SemanticAst/FilterSegment.cs
AlanWong-MS: Update: upon making it internal, a test service `test\FunctionalTests\Service\Microsoft.OData.Service.csproj` has caused a lot of issues. (Not sure if I really like the way that the test service is designed.)

1. The test service tried to access some of the internals of the source code (specifically the new internal LiteralText).
2. Upon either enabling internals to the test service or adding a link of FilterSegment.cs to the test service, the test service tries to access more internals.
3. Upon adding those internals that the test service tries to access, the chain of access to internals  continues to grow, so I abandoned this approach because it looked like it was going to take a while to add everything needed.
---
1. I also tried to move the LiteralText building into the Translator classes. However, due to the need of using the NodeToStringBuilder (which is also internal) to help us determine the text, we come across the same problem as the one mentioned above. The Microsoft.OData.Services project will not build due to the lack of access to the internals.

I spent several hours to try to work around the test service and I'd probably need more time to review whether we can remove the test service.

Therefore, the easiest solution that I see is to leave LiteralText as public so we come across issues with the Microsoft.OData.Service test project; in trying to solve this issue, I noticed that a few parts of the production code has `#if ODATA_SERVICE` defines just to cater to the test project which I don't think is a good thing. Perhaps the FilterSegment interface could be re-evaluated during the next major release.

src/Microsoft.OData.Core/UriParser/SemanticAst/EachSegment.cs
mikepizzo: Doesn't appear to be fixed?

---
In reply to: [224584557](https://github.com/OData/odata.net/pull/1110#discussion_r224584557) [](ancestors = 224584557)

src/Microsoft.OData.Core/UriParser/SemanticAst/EachSegment.cs
mikepizzo: Got it.

Rather than make targetEdmType parameter nullable, wouldn't it be simpler to always pass in the targetEdmType and simply set this.TargetEdmType = targetEdmType in the constructor (rather than check for null)?  

In the code above, couldn't you just pass in prevSegment.TargetEdmType for all cases (presumably, if the previous segment was the navigation source, then lastNavigationSource.EntityType() == prevSegment.TargetEdmType?)

---
In reply to: [224878387](https://github.com/OData/odata.net/pull/1110#discussion_r224878387) [](ancestors = 224878387)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -450,80 +454,225 @@ private bool TryCreateCountSegment(string segmentText)
...
+            {
+                throw new ODataException(ODataErrorStrings.RequestUriProcessor_CannotApplyEachOnSingleEntities(lastNavigationSource.Name));
+            }
 
-            // If this is a navigation property, find target navigation source
-            IEdmPathExpression bindingPath;
-            var targetNavigationSource = this.parsedSegments[parsedSegments.Count - 1].TargetEdmNavigationSource.FindNavigationTarget(navPropSegment.NavigationProperty, BindingPathHelper.MatchBindingPath, this.parsedSegments, out bindingPath);
+            EachSegment eachSegment = new EachSegment(
+                lastNavigationSource,
+                prevSegment is TypeSegment || prevSegment is FilterSegment ? prevSegment.TargetEdmType : null);
```
mikepizzo: Can you just do EachSegment eachSegment = new EachSegment(lastNavigationSource,prevSegment.TargetEdmType)?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/SetBasedOperationTests.cs
mikepizzo: Not fixed?  Also, wording is a bit awkward; a function is an operations, but we are checking for actions, which are also operations. I think what we are really enforcing is only one operation segment (function or action) can follow /$each? (is that right, that a function can follow $each?)

---
In reply to: [224938789](https://github.com/OData/odata.net/pull/1110#discussion_r224938789) [](ancestors = 224938789)

src/Microsoft.OData.Core/UriParser/SemanticAst/EachSegment.cs
AlanWong-MS: Odd...I distinctly remember fixing this as well as the other code comment ones. Actually doing it right now. Marking unresolved for now.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/SetBasedOperationTests.cs
AlanWong-MS: Distinctly remember updating the text. Marking unresolved.

AlanWong-MS: I think we decided that only one action can follow $each.

For convenience, here's the [spec](http://docs.oasis-open.org/odata/odata/v4.01/cs01/part1-protocol/odata-v4.01-cs01-part1-protocol.html#sec_BindinganOperationtoaResource).

src/Microsoft.OData.Core/UriParser/SemanticAst/EachSegment.cs
```diff
@@ -0,0 +1,90 @@
...
```
mikepizzo: Note that, according to OData Protocol, $each is valid following any collection (not just an entity collection).  If we change this to get this.TargetKind based on targetEdmType, and this.TargetEdmType to be the passed in targetEdmType, is there anything else that assumes the collection being iterated is a collection of entities?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
AlanWong-MS: Stepped through the debugger a several times. Just posting results here (follow-up response after):

1. If previous segment was TypeSegment (e.g. "People/Fully.Qualified.Namespace.Employee/$each")

prevSegment.TargetEdmType:
TargetEdmType	{Fully.Qualified.Namespace.Employee}	Microsoft.OData.Edm.IEdmType {Microsoft.OData.Edm.Csdl.CsdlSemantics.CsdlSemanticsEntityTypeDefinition}

lastNavigationSource.EntityType():
TargetEdmType	{Fully.Qualified.Namespace.Person}	Microsoft.OData.Edm.IEdmType {Microsoft.OData.Edm.Csdl.CsdlSemantics.CsdlSemanticsEntityTypeDefinition}

2. If previous segment was FilterSegment (e.g. "People/$filter(@p1)/$each")

prevSegment.TargetEdmType:
TargetEdmType	{Fully.Qualified.Namespace.Person}	Microsoft.OData.Edm.IEdmType {Microsoft.OData.Edm.Csdl.CsdlSemantics.CsdlSemanticsEntityTypeDefinition}

lastNavigationSource.EntityType():
TargetEdmType	{Fully.Qualified.Namespace.Person}	Microsoft.OData.Edm.IEdmType {Microsoft.OData.Edm.Csdl.CsdlSemantics.CsdlSemanticsEntityTypeDefinition}

3. If previous segment was FilterSegment whose previous segment was TypeSegment (e.g. "People/Fully.Qualified.Namespace.Employee/$filter(@p1)/$each")

prevSegment.TargetEdmType:
TargetEdmType	{Fully.Qualified.Namespace.Person}	Microsoft.OData.Edm.IEdmType {Microsoft.OData.Edm.Csdl.CsdlSemantics.CsdlSemanticsEntityTypeDefinition}

lastNavigationSource.EntityType():
TargetEdmType	{Fully.Qualified.Namespace.Person}	Microsoft.OData.Edm.IEdmType {Microsoft.OData.Edm.Csdl.CsdlSemantics.CsdlSemanticsEntityTypeDefinition}

4. If previous segment was EntitySetSegment (e.g. "People/$each")

prevSegment.TargetEdmType:
TargetEdmType	{Fully.Qualified.Namespace.Person}	Microsoft.OData.Edm.IEdmType {Microsoft.OData.Edm.Csdl.CsdlSemantics.CsdlSemanticsEntityTypeDefinition}

lastNavigationSource.EntityType():
TargetEdmType	{Fully.Qualified.Namespace.Person}	Microsoft.OData.Edm.IEdmType {Microsoft.OData.Edm.Csdl.CsdlSemantics.CsdlSemanticsEntityTypeDefinition}

5. If previous segment was NavigationPropertySegment (e.g. "People/1/MyFriendsDogs/$each")

prevSegment.TargetEdmType:
TargetEdmType	{Collection([Fully.Qualified.Namespace.Dog Nullable=False])}	Microsoft.OData.Edm.IEdmType {Microsoft.OData.Edm.EdmCollectionType}

lastNavigationSource.EntityType():
TargetEdmType	{Fully.Qualified.Namespace.Dog}	Microsoft.OData.Edm.IEdmType {Microsoft.OData.Edm.Csdl.CsdlSemantics.CsdlSemanticsEntityTypeDefinition}

AlanWong-MS: From case 3, we need to specifically use the prevSegment.TargetEdmType for both FilterSegment and TypeSegment scenarios to retain the derived type information. Therefore we can't just pass the lastNavigationSource.EntityType().

From case 5, we can't just pass the prevSegment.TargetEdmType as it refers to a collection.

Therefore, I think the code there might be the cleanest form to account for all the cases.

src/Microsoft.OData.Core/UriParser/SemanticAst/EachSegment.cs
AlanWong-MS: Great point, I'll move the null check out of the constructor and make it not nullable. Additionally the check is made prior to construction. There may be some other path segments that are following this pattern, where the check is done inside the constructor; I think your suggestion is cleaner.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1200,5 +1386,56 @@ private bool IdentifierIs(string expected, string identifier)
...
+
+        /// <summary>
+        /// Per OData 4.01 spec, GET operation and functions may proceed $each, but we are limiting the scope of that spec
+        /// by permitting only ONE action segment to follow $each. As such, at most one $each segment can exist in the URI.
+        /// This function enforces those restrictions and throws if any of them are violated.
+        /// </summary>
+        /// <param name="index">Index of path segment to examine in the list of parsed segments.</param>
+        /// <exception cref="ODataException">Throws if there's a violation of $each restrictions.</exception>
+        /// <remarks>Should the restrictions on the $each be removed, this function can be deleted.</remarks>
+        private void CheckDollarEachSegmentRestrictions(int index)
```
AlanWong-MS: @mikepizzo, regarding the single bound action/function following $each, I have this function here that's enforcing that constraint that we decided to impose at the time. It's going to be a little difficult for me to dig deeper into this or make new changes given some time constraints that I have. If there should be further adjustments to this logic, could that be done in a separate PR prior to shipping the next ODL?

src/Microsoft.OData.Core/UriParser/SemanticAst/EachSegment.cs
AlanWong-MS: Thanks for point this out. There aren't any other code paths that I wrote which would assume that the collection is a collection of entities. I've updated the line to:

`this.TargetKind = targetEdmType.GetTargetKindFromType();`

mikepizzo: Cool. Do we have any tests that validate the use against a collection(complex) or collection(primitive)?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
mikepizzo: In line 1434, can you just remove:
       || !(operationSegment.Operations.All(a => a is IEdmAction))
?

---
In reply to: [225763923](https://github.com/OData/odata.net/pull/1110#discussion_r225763923) [](ancestors = 225763923)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
mikepizzo: How about prevSegment.TargetEdmType.AsElementType()?

---
In reply to: [225761341](https://github.com/OData/odata.net/pull/1110#discussion_r225761341) [](ancestors = 225761341)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
AlanWong-MS: Just saw your edit from one of the other posts. Yes, removing that condition will allow any single operation to follow $each. Making that change and adjusting tests.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
AlanWong-MS: That works! That also makes sense that there's a property in there somewhere that refers to the singular type.

src/Microsoft.OData.Core/UriParser/SemanticAst/EachSegment.cs
AlanWong-MS: Adding two tests to explicitly test this:
```
        [Fact]
        public void SetBasedOperations_EachSegmentOnPrimitiveTypeCollection_ReturnsSuccess()
        {
            ParseUriAndVerify(
                new Uri("http://gobbledygook/People/1/RelatedIDs/$each"),
                (oDataPath, filterClause, aliasNodes) =>
                {
                    oDataPath.Count.Should().Be(4);

                    EachSegment eachSegment = oDataPath.Last() as EachSegment;
                    eachSegment.Should().NotBeNull();
                    eachSegment.TargetKind.Should().Be(RequestTargetKind.Primitive);
                    eachSegment.TargetEdmNavigationSource.Should().Be(HardCodedTestModel.GetPeopleSet());
                });
        }

        [Fact]
        public void SetBasedOperations_EachSegmentOnComplexTypeCollection_ReturnsSuccess()
        {
            ParseUriAndVerify(
                new Uri("http://gobbledygook/People/1/PreviousAddresses/$each"),
                (oDataPath, filterClause, aliasNodes) =>
                {
                    oDataPath.Count.Should().Be(4);

                    EachSegment eachSegment = oDataPath.Last() as EachSegment;
                    eachSegment.Should().NotBeNull();
                    eachSegment.TargetKind.Should().Be(RequestTargetKind.Resource);
                    eachSegment.TargetEdmNavigationSource.Should().Be(HardCodedTestModel.GetPeopleSet());
                });
        }
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/ParameterAliasFunctionalTests.cs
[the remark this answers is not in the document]
mikepizzo: still doesn't look like you validate filterSegments[0].Expression.Left

---
In reply to: [224609615](https://github.com/OData/odata.net/pull/1110#discussion_r224609615) [](ancestors = 224609615)

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/SemanticAst/ReferenceSegmentTests.cs
mikepizzo: Second one would be something like:
            ReferenceSegment referenceSegment = new ReferenceSegment(HardCodedTestModel.GetBossSingleton());
            referenceSegment.Identifier.Should().Be(UriQueryConstants.RefSegment);
            referenceSegment.SingleResult.Should().BeTrue();
            referenceSegment.TargetEdmNavigationSource.Should().Be(HardCodedTestModel.GetBossSingleton());
            referenceSegment.TargetKind.Should().Be(RequestTargetKind.Resource);

---
In reply to: [224945402](https://github.com/OData/odata.net/pull/1110#discussion_r224945402) [](ancestors = 224945402)

mikepizzo: >referenceSegment.EdmType.Should().BeNull(); [](start = 12, length = 43)

Why is EdmType null?  Shouldn't it be Pet?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/ParameterAliasFunctionalTests.cs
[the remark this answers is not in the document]
AlanWong-MS: GitHub's diff tool seems to have obstructed the view. I was scratching my head for a bit but here it is:

image

I can probably re-write that test differently if you prefer.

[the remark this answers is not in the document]
AlanWong-MS: Looks like this won't be the final review, so I'll write the test slightly differently.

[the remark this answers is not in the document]
AlanWong-MS: Actually, just caught the redundancy in checks, so re-write on the checking the right side is the right thing to do:
```
List<FilterSegment> filterSegments = oDataPath.OfType<FilterSegment>().ToList();
filterSegments.Count.Should().Be(1);
filterSegments[0].TargetEdmType.ToString().ShouldBeEquivalentTo(HardCodedTestModel.GetPersonType().ToString());
filterSegments[0].SingleResult.Should().BeFalse();

BinaryOperatorNode binaryOperatorNode = filterSegments[0].Expression as BinaryOperatorNode;
binaryOperatorNode.Should().NotBeNull();
binaryOperatorNode.OperatorKind.Should().Be(BinaryOperatorKind.Equal);
binaryOperatorNode.Left.ShouldBeSingleValuePropertyAccessQueryNode(HardCodedTestModel.GetPersonIdProp());
binaryOperatorNode.Right.ShouldBeConstantQueryNode(1);

filterClause.Should().BeNull();
aliasNodes.Should().BeEmpty();
```

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/SemanticAst/ReferenceSegmentTests.cs
AlanWong-MS: Ah okay, that's added now.

src/Microsoft.OData.Core/UriParser/Visitors/PathSegmentToContextUrlPathTranslator.cs
```diff
@@ -180,6 +180,38 @@ public override string Translate(CountSegment segment)
...
+        /// <summary>
+        /// Translate a ReferenceSegment
+        /// </summary>
+        /// <param name="segment">the segment to Translate</param>
+        /// <returns>Defined by the implementer.</returns>
+        public override string Translate(ReferenceSegment segment)
+        {
+            // How a ReferenceSegment should be addressed in the context URI is still a pending decision.
+            // https://github.com/OData/odata.net/issues/1292
+            return string.Empty;
```
mikepizzo: See comments in 1292.

If the /$ref is in the path, the contextUrl should simply be #$ref (or #Collection($ref).

See [http://docs.oasis-open.org/odata/odata/v4.01/cs01/part1-protocol/odata-v4.01-cs01-part1-protocol.html#sec_EntityReference](http://docs.oasis-open.org/odata/odata/v4.01/cs01/part1-protocol/odata-v4.01-cs01-part1-protocol.html#sec_EntityReference).

Note that the entity set/singleton/navprop is *not* in the Url.  So, for example, Customers(1)/Orders(1)/Items/$ref would just be $metadata#Collection($ref).


That's a bit problemmatic for this translator as encountering a ReferenceSegment should replace the entire ContextUrl fragment with #ref or Collection(#ref).

PR #1112, opened 2018-03-15
Fix for not throwing and exception for an invalid enum value

mikepizzo:
### Description

When we added support for casting strings to enum values in URLs, we didn't validate that the string was a valid member of the enum type.  This means that using a prefixed enum value would throw an exception if the value was not a member of the enum, but using the string form would not.

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
@@ -6,7 +6,9 @@
 
 namespace Microsoft.OData.UriParser
 {
+    using System;
     using System.Diagnostics;
+    using System.Linq;
     using Microsoft.OData.Edm;
     using Microsoft.OData;
     using Microsoft.OData.Metadata;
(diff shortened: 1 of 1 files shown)


src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
```diff
@@ -58,7 +60,14 @@ internal static SingleValueNode ConvertToTypeIfNeeded(SingleValueNode source, IE
                 ConstantNode constantNode = source as ConstantNode;
                 if (constantNode != null && constantNode.Value != null && source.TypeReference.IsString() && targetTypeReference.IsEnum())
                 {
-                    return new ConstantNode(new ODataEnumValue(constantNode.Value.ToString(), targetTypeReference.Definition.ToString()), constantNode.Value.ToString(), targetTypeReference);
+                    string memberName = constantNode.Value.ToString();
+                    IEdmEnumType enumType = targetTypeReference.Definition as IEdmEnumType;
+                    if (enumType.Members.Any(m => m.Name == memberName))
+                    {
```
biaol-odata: Should we use string.Equals here?

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
```diff
@@ -58,7 +60,14 @@ internal static SingleValueNode ConvertToTypeIfNeeded(SingleValueNode source, IE
                 ConstantNode constantNode = source as ConstantNode;
                 if (constantNode != null && constantNode.Value != null && source.TypeReference.IsString() && targetTypeReference.IsEnum())
                 {
-                    return new ConstantNode(new ODataEnumValue(constantNode.Value.ToString(), targetTypeReference.Definition.ToString()), constantNode.Value.ToString(), targetTypeReference);
+                    string memberName = constantNode.Value.ToString();
+                    IEdmEnumType enumType = targetTypeReference.Definition as IEdmEnumType;
+                    if (enumType.Members.Any(m => m.Name == memberName))
+                    {
+                        return new ConstantNode(new ODataEnumValue(constantNode.Value.ToString(), targetTypeReference.Definition.ToString()), constantNode.Value.ToString(), targetTypeReference);
```
biaol-odata: > return [](start = 23, length = 7)

nit: put throw inside an "else" branch?

PR #1113, opened 2018-03-15
Default build and solution configuration to VS2017

AlanWong-MS:
### Description
Migrating all solutions to VS2017. Removing the VS2015 versions of NetStandard library projects and solution. Renaming solutions and projects with .VS2017 suffix to remove the suffix. Updating build.ps1 to default to VS2017 and ReadMe to address changes.

With this change, the build.cmd script will default to VS2017 and if none of the versions are found (Enterprise/Professional/Community), then the script will fall back to VS2015. All of the solutions except the ones for NetStandard, NetCore, and ClientCodeGen can still be built with VS2015. Official support for NetStandard and NetCore starts with VS2017 so it is now the recommended IDE version to use for development; existing users may still use VS2015 to build most solutions and new users should install just VS2017 for development.


AlanWong-MS:
Note: build failures are expected due to solution file changes.

jamisonhyatt:
@AlanWong-MS - These changes seem to break the building of the net standard binaries in the location nuspec expects. In master, building the NetStandard solution publishes correctly to `bin\AnyCPU\Release\Product\.NETPortable\v5.0` (where the nuspec is looking).  As far as I can tell, with your changes, building the NetStandard sln file actually publishes those under `Desktop`, instead.   Am I missing something?

AlanWong-MS:
Yes, this does seem to be the case.

Here's the vs2015 version: https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Edm/Build.NetStandard/Microsoft.OData.Edm.NetStandard.csproj
 
Here's the vs2017 version:
https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Edm/Microsoft.OData.Edm.NetStandard.VS2017.csproj

Indeed it looks like the vs2015 version specifies the output location: https://github.com/OData/odata.net/blob/39f4b4c346a61822fb238e4914fb9345695e3cfa/src/Microsoft.OData.Edm/Build.NetStandard/Microsoft.OData.Edm.NetStandard.csproj#L28

The VS2017 csproj files should specify the same locations for output.

jamisonhyatt:
Thanks @AlanWong-MS  - I couldn't get msbuild or visual studio to respect those outputs after modifying them.  They don't appear to affect the location at all.  However, It appears the root [Build.props](https://github.com/OData/odata.net/blob/39f4b4c346a61822fb238e4914fb9345695e3cfa/Build.props#L92) is expecting a value "TargetFrameworkFolderName", and defaults to "Desktop" even for .NetStandard (after your changes).

This can be solved inside the csproj files themselves, either by adding `<TargetFrameworkFolderName>.NETPortable\v5.0</TargetFrameworkFolderName>` to each of the 4 *.NetStandard.csproj files, or when building the `.\sln\OData.NetStandard.sln` passing the `/p:TargetFrameworkFolderName=".NETPortable\v5.0"` property during msbuild.

Pinging @xuzhg to see which approach he'd prefer here.

xuzhg:
Close it since it's not applicable

PR #1118, opened 2018-03-21
Fix path for EdmContainedEntitySet

mikepizzo:
### Issues

EdmContainedEntitySet.Path should include the navProp. Currently, the path returned is the path of the parent.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Edm/Schema/EdmContainedEntitySet.cs
+++ b/src/Microsoft.OData.Edm/Schema/EdmContainedEntitySet.cs
@@ -179,7 +179,7 @@ namespace Microsoft.OData.Edm
                 newPath.Add(targetType.FullTypeName());
             }
 
-            newPath.Concat(this.NavigationPath.PathSegments);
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Edm/Schema/EdmContainedEntitySet.cs
```diff
@@ -179,7 +179,7 @@ private IEdmPathExpression ComputePath()
                 newPath.Add(targetType.FullTypeName());
             }
 
-            newPath.Concat(this.NavigationPath.PathSegments);
+            newPath.AddRange(this.NavigationPath.PathSegments);
```
xuzhg: What is the difference between the AddRange and Concat functions on a generic List

mikepizzo: Concat returns a new sequence, but we never assigned the result to a variable so it was effectively a no-op. So we could either do:
   newPath = newPath.Concat(...)
or simply:
   newPath.AddRange(...)

PR #1123, opened 2018-03-21
Include dots in parseIdentifier for annotation parsing

biaol-odata:
### Issues

*This pull request fixes an issue for annotation parsing in ODL 7.4.2.*

### Description

*Annotation such as "@NS.myAnnotation" are not parsed correctly as Identifier token because lexer stops at the dot char during the traversing. This can be fixed by allowing to include dot chars for paring string token starting with "@", which could be either an parameter alias (w/o dots) or an annotation.*

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
+++ b/src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
@@ -680,7 +680,9 @@ namespace Microsoft.OData.UriParser
                         }
 
                         int start = this.textPos;
-                        this.ParseIdentifier();
+
+                        // Include dots for the case of annotation.
+                        this.ParseIdentifier(true /*includingDots*/);
 
                         // Extract the identifier from expression.
                         string leftToken = ExpressionText.Substring(start, this.textPos - start);
@@ -1201,14 +1203,15 @@ namespace Microsoft.OData.UriParser
 
 
         /// <summary>Parses an identifier by advancing the current character.</summary>
-        private void ParseIdentifier()
+        /// <param name="includingDots">Optional flag for whether to include dots as part of the identifier.</param>
+        private void ParseIdentifier(bool includingDots = false)
         {
             Debug.Assert(this.IsValidStartingCharForIdentifier || this.ch == UriQueryConstants.AnnotationPrefix, "Expected valid starting char for identifier");
             do
             {
                 this.NextChar();
             }
-            while (this.IsValidNonStartingCharForIdentifier);
+            while (this.IsValidNonStartingCharForIdentifier || (includingDots && this.ch == '.'));
         }
 
         /// <summary>Sets the text position.</summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
```diff
@@ -1201,14 +1203,15 @@ private void AdvanceThroughBalancedExpression(char startingCharacter, char endin
 
 
         /// <summary>Parses an identifier by advancing the current character.</summary>
-        private void ParseIdentifier()
+        /// <param name="includingDots">Optional flag for whether to include dots as part of the identifier.</param>
+        private void ParseIdentifier(bool includingDots = false)
```
mikepizzo: >bool includingDots = false [](start = 37, length = 26)

We pass this as true if the leading character is an "@", otherwise we pass this as false.  Rather than making this a parameter, should we just set includingDots = this.ch==UriQueryConstants.AnnotationPrefix within the function?

src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
```diff
@@ -680,7 +680,9 @@ protected virtual ExpressionToken NextTokenImplementation(out Exception error)
                         }
 
                         int start = this.textPos;
-                        this.ParseIdentifier();
+
+                        // Include dots for the case of annotation.
+                        this.ParseIdentifier(true /*includingDots*/);
```
mikepizzo: >true /*includingDots*/ [](start = 45, length = 22)

So we allow dots for either the parameter case or the annotation case.  Dots aren't really allowed in parameter names, but for the purpose of parsing the expression it is probably fine to include them as part of the identifier (i.e., there's not a valid case where a dot would be a valid separator between a parameter and something else, so including it in the identifier is reasonable)

mikepizzo:
Can we add some additional tests for the UriParser, to parse different expressions containing combinations of annotations and parameter markers?  i.e.; with this change, the lexer will support the parser parsing @my.annotation eq 5 in a filter expression; that should either throw a reasonable exception or parse to an annotation node as the left operand; it should *not* successfully parse as a parameter as the left operand.

biaol-odata:
Sure. Please see two test cases added in ODataUriParserTests in the commit above.

src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
biaol-odata: It would be equivalent to the current logic, but explicitly passing in as bool from caller is causing less ripple, cleaner therefore less risky and more flexible in case this method is needed later in other place. Therefore I prefer explicitly passing in a bool value unless you have strong opinion on this.

---
In reply to: [176544495](https://github.com/OData/odata.net/pull/1123#discussion_r176544495) [](ancestors = 176544495)

src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
biaol-odata: Yes, should be OK to include dots, which is further used to differentiate parameter alias and annotation per OData protocol. Please let me know if I miss anything in your comment.

---
In reply to: [176545588](https://github.com/OData/odata.net/pull/1123#discussion_r176545588) [](ancestors = 176545588)

PR #1130, opened 2018-03-23
Add StringComparison in String.Equal() method call

xuzhg:
### Issues

* similar issue at: https://github.com/OData/WebApi/issues/1215

### Description

* Add StringComparison in String.Equal() method call.*

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/BatchSaveResult.cs
+++ b/src/Microsoft.OData.Client/BatchSaveResult.cs
@@ -471,7 +471,7 @@ namespace Microsoft.OData.Client
                     Encoding encoding;
                     Exception inner = contentTypeException;
                     ContentTypeUtil.ReadContentType(this.batchResponseMessage.GetHeader(XmlConstants.HttpContentType), out mime, out encoding);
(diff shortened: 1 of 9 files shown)


biaol-odata:
@xuzhg We should make sure the suppresions for CA1307 such as the following are cleaned up as well:
test\EndToEndTests\Framework\Reliability\GlobalSuppressions.cs Line8:
[assembly: System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Globalization", "CA1307:SpecifyStringComparison"......

xuzhg:
@biaol-odata Thanks your information. Do you know what's the usage of "Reliability"? And I saw all the 'CA1307' are not related to `String.Equals`.

biaol-odata:
@xuzhg I am not sure what the usage of "Reliability" is. I was just thinking it would be logical to remove the CA1307, which is either no further changes required or leading us to more finding. Since this has been merged already, no further action is required now (good item to work on for next MQ?).

PR #1135, opened 2018-03-30
Make reading/writing navigationPropertyMappings threadsafe

mikepizzo:
### Description

Fixes two issues:
-Locks navigationPropertyMappings dictionary when reading/writing entries
-Makes sure NavigationPropertyBindings are not written for containment navigation properties

- [x] *Test cases added*


src/Microsoft.OData.Edm/Schema/EdmContainedEntitySet.cs
```diff
@@ -179,7 +179,7 @@ private IEdmPathExpression ComputePath()
                 newPath.Add(targetType.FullTypeName());
             }
 
-            newPath.Concat(this.NavigationPath.PathSegments);
+            newPath.AddRange(this.NavigationPath.PathSegments);
```
xuzhg: >AddRange [](start = 20, length = 8)

I remember this change has been out in 7.4.3, Don't we?

src/Microsoft.OData.Edm/Schema/EdmNavigationSource.cs
```diff
@@ -110,8 +114,10 @@ public virtual IEnumerable<IEdmNavigationPropertyBinding> FindNavigationProperty
...
             EdmUtil.CheckArgumentNull(navigationProperty, "navigationProperty");
 
-            Dictionary<string, IEdmNavigationPropertyBinding> result;
-            if (this.navigationPropertyMappings.TryGetValue(navigationProperty, out result))
+            Dictionary<string, IEdmNavigationPropertyBinding> result =
+                EdmUtil.DictionarySafeGet(this.navigationPropertyMappings, navigationProperty);
+
+            if (result != null)
             {
                 return result.Select(item => item.Value);
```
xuzhg: >result.Select(item => item.Value); [](start = 23, length = 34)

why can't return result.Values?

src/Microsoft.OData.Edm/Schema/EdmContainedEntitySet.cs
mikepizzo: Yes; the PR ended up pulling in that commit, which has already been merged, but since it's the same commit it doesn't result in a conflict.

test/FunctionalTests/Tests/DataEdmLib/FunctionalUtilities/NavigationTestModelBuilder.cs
```diff
@@ -1412,9 +1396,9 @@ public static IEdmModel RecursiveThreeContainmentNavigationsWithEntitySet()
 
             personSet.AddNavigationTarget(personToPet, petSet);
             petSet.AddNavigationTarget(personToPet.Partner, personSet);
-            petSet.AddNavigationTarget(petToHome, homeSet);
+//            petSet.AddNavigationTarget(petToHome, homeSet);
```
xuzhg: if we don't want to remove these, it's better to have some comments?

src/Microsoft.OData.Edm/Schema/EdmNavigationSource.cs
mikepizzo: I didn't change anything here, but it does seem like we could just return result.Values.

test/FunctionalTests/Tests/DataEdmLib/FunctionalUtilities/NavigationTestModelBuilder.cs
mikepizzo: I will create a separate pull request to remove these.

PR #1142, opened 2018-04-02
fix write document metadata properties for apply ComputeTransformatio…

voronov-maxim:
### Description

When write response for query "OrderItems?$apply=compute(Count mul Price as Total,Id add OrderId as SumId)" compute properties not append metadata "@odata.context": "http://dummy/$metadata#OrderItems".

Response must be:
"@odata.context": "http://dummy/$metadata#OrderItems(Total,SumId)"

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Aggregation/ApplyClause.cs
+++ b/src/Microsoft.OData.Core/UriParser/Aggregation/ApplyClause.cs
@@ -106,7 +106,7 @@ namespace Microsoft.OData.UriParser.Aggregation
 
             if (computeExpressions != null)
             {
-                string computeProperties = string.Join(",", computeExpressions.Select(e => e.Alias));
+                string computeProperties = string.Join(",", computeExpressions.Select(e => e.Alias).ToArray());
                 if (!string.IsNullOrEmpty(computeProperties))
                     result = string.IsNullOrEmpty(result)
                         ? computeProperties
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/UriParser/Aggregation/ApplyClause.cs
```diff
@@ -102,6 +104,15 @@ internal List<string> GetLastAggregatedPropertyNames()
                     : CreateAggregatePropertiesUriSegment(aggregateExpressions);
             }
 
+            if (computeExpressions != null)
+            {
+                string computeProperties = string.Join(",", computeExpressions.Select(e => e.Alias));
+                if (!string.IsNullOrEmpty(computeProperties))
+                    result = string.IsNullOrEmpty(result)
```
AlanWong-MS: Even though this is a single line within the if-statement, we prefer to still have the { } based on our style guide.

src/Microsoft.OData.Core/UriParser/Aggregation/ApplyClause.cs
```diff
@@ -102,6 +104,15 @@ internal List<string> GetLastAggregatedPropertyNames()
                     : CreateAggregatePropertiesUriSegment(aggregateExpressions);
             }
 
+            if (computeExpressions != null)
+            {
+                string computeProperties = string.Join(",", computeExpressions.Select(e => e.Alias));
```
AlanWong-MS: There's a build failure at this line:
`src\Microsoft.OData.Core\UriParser\Aggregation\ApplyClause.cs(109,61): error CS1503: Argument 2: cannot convert from 'System.Collections.Generic.IEnumerable<string>' to 'string[]' [src\Microsoft.OData.Core\Build.Net35\Microsoft.OData.Core.NetFX35.csproj]`

You may want to apply a `.ToArray()` at the end of the LINQ expression, similar to lines 93-94.

AlanWong-MS:
Thanks for addressing my comments. It would be great if you're able to add a test case to cover your change; I don't think there were any existing tests. I think you may be able to add it to `test\FunctionalTests\Microsoft.OData.Core.Tests\ScenarioTests\Writer\JsonLight\FullPayloadValidateTests.cs` and use the existing ones as an example.

AlanWong-MS:
Thanks!

Issue #1151, opened 2018-04-13
Port client-side DateTime support fix to 7.x


AlanWong-MS:
DateTime Support on the client side is missing. See #715 for PR details.

### Assemblies affected
OData Client/Core 7.4.4

### Reproduce steps
See #473. 

### Expected result
DateTime should be supported in the client library.

### Actual result
DateTime is not supported in the client library.

wmalgadey:
Is there a release plan, when this could be pushed to nuget? after all tests are green of course.

wmalgadey:
So! Will you merge it? Could it happen in the next release of odata client 7.x?

AlanWong-MS:
Pending CR on the PR, I've pinged some members to help with that. :) We should be able to fit it into the next release of 7.x.

AlanWong-MS:
Merged, it'll be in next release of 7.x.

PR #1152, opened 2018-04-13
Client-side DateTime support

AlanWong-MS:
### Issues
#1151

### Description
Adding client-side DateTime implementation and corresponding test cases.

- [x] *Test cases added*


xuzhg:
DateTime was the part of Edm.Primititive. In OData V4, it's removed from the spec to meet the requirement for OASIS standard. I think it's better to sync with @mikepizzo about the supporting for DateTime.

AlanWong-MS:
Got it, thanks for the heads up.

seanrodda:
Is this change included in the recently release version 7.5.0 of the client library? I notice this mentions being merged into master on May 11, however it is still "in progress" with some recent changes 5 days ago. Just wondering what, if anything, was included in 7.5.0. 
Thanks

AlanWong-MS:
@seanrodda, yes it should be included in 7.5.0. Thanks for bringing up the note. You can refer to #1152 to see what changes were included.

Issue #1153, opened 2018-04-21
UriParser should throw exception for mismatched keys count when UnqualifiedODataUriResolver is configured.


biaol-odata:
*UriParser should throw exception for mismatched keys count (extraneous) when UnqualifiedODataUriResolver is configured.*
Sample case:
With the following EDM model:
```
            var pet = new EdmEntityType("TestNS", "Pet");
            PetType = pet;
            model.AddElement(pet);
            var key1 = pet.AddStructuralProperty("key1", EdmCoreModel.Instance.GetInt32(false));
            var key2 = pet.AddStructuralProperty("key2", EdmCoreModel.Instance.GetString(false));
            pet.AddKeys(key1, key2);
            var petSet = container.AddEntitySet("PetSet", pet);
```

and the following relative uri: 
`"PetSet(key1=1, key2='aStr', nonExistingKey='bStr')"`

ODataUriParser instance configured with UnqualifiedODataUriResolver should throw exception for mismatched key count during uri parsing.
Note the default ODataUriParser instance configured with default ODataUriResolver does throw exception as expected. Therefore, the difference is caused by the different types of Uri resolver configured.

### Assemblies affected

*OData .Net lib < 7.4.4.*

### Reproduce steps

*See description above.*

### Expected result

*ODataException thrown for mismatched keys count.*

### Actual result

*No exception thrown. Extraneous keys are ignored. *

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

biaol-odata:
Closed as merged.

PR #1154, opened 2018-04-21
Fix SegmentKeyHandler to include UnqualifiedODataUriResolver proccessing

biaol-odata:
### Issues

*This pull request fixes issue #1153.*

### Description

*Include UnqualifiedODataUriResolve type when checking for mismatched keys count.*

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Parsers/SegmentKeyHandler.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/SegmentKeyHandler.cs
@@ -153,7 +153,7 @@ namespace Microsoft.OData.UriParser
 
             Debug.Assert(targetEntityType != null, "targetEntityType != null");
 
-            // Make sure the keys specified in the uri matches with the number of keys in the metadata
+            // Adjust the keys for navigation segment.
             var keyProperties = targetEntityType.Key().ToList();
             if (keyProperties.Count != key.ValueCount)
             {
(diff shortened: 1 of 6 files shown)


biaol-odata:
This PR fixes an dependency issue for OData/WebApi#1409.

src/Microsoft.OData.Core/UriParser/Parsers/SegmentKeyHandler.cs
```diff
@@ -163,8 +163,9 @@ private static KeySegment CreateKeySegment(ODataPathSegment segment, KeySegment
...
                 }
 
-                // if we still didn't find any keys, then throw an error.
-                if (keyProperties.Count != key.ValueCount && resolver.GetType() == typeof(ODataUriResolver))
+                // If we still didn't find any keys, throw an error, except for case of
+                // AlternateKeysODataUriResolver which can handle alternate keys and can throw later for unresolvable case.
+                if (keyProperties.Count != key.ValueCount && !(resolver is AlternateKeysODataUriResolver))
                 {
                     throw ExceptionUtil.CreateBadRequestError(ErrorStrings.BadRequest_KeyCountMismatch(targetEntityType.FullName()));
                 }
```
mikepizzo: This would break an existing custom resolver, like the ALternateKeysODataUriResolver, that handled other key resolution logic. What if we moved this check to the virtual ResolveKeys in the ODataUriParser? The default logic for that method assumes the keyvaluepairs being compared match the key values defined for the type and would have to be overridden by a derived ODataUriParser that inserted custom logic.

mikepizzo: yes.

biaol-odata: @mikepizzo : For other resolver provided by client, having to override the default ResolveKeys is a breaking change(otherwise it won't work) is also a breaking change as well. Right?

mikepizzo: The logic in the virtual ResolveKeys has a hard requirement that the key segments specified in the URL are 1:1 with the defined key for the type. Any custom OData URI resolver that supports key segments that are not 1:1 with the defined key must already override this method.  So putting the logic here, in the method that requires that the segments are 1:1, is safe and makes sense.

biaol-odata:
Merged. Thanks!

biaol-odata:
Almost forgot to merge it. Now it can be closed.

Issue #1155, opened 2018-04-26
FunctionCallBinder issue when case-insensitive is enabled


biaol-odata:
In ODL, when case-insensitive is enabled, FunctionCallBinder.BindAsUriFunction() essentically does a lower-case search for matching function names.
https://github.com/OData/odata.net/blob/3fb1497e8c21cba1e5f8c34322476d2e40b28d9b/src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs#L269

This does not work for custom uri functions, returning error message with incorrect function name.
The following WebAPI test fails when enableCaseInsensitive is set to true.
https://github.com/OData/WebApi/blob/master/test/E2ETest/Microsoft.Test.E2E.AspNet.OData/UnboundOperation/UnboundOperationTest.cs#L326

### Assemblies affected

*7.4.4 and below.*

### Reproduce steps

*See description.*

### Expected result

*Test UnboundOperationTest pass when case-insensitive is enabled.*

### Actual result

*Test UnboundOperationTest failed with error message indicating incorrect function name.*

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

biaol-odata:
Merged.

PR #1156, opened 2018-04-26
Fix case insensitive search for custom URI functions.

biaol-odata:
### Issues

*This pull request fixes issue #1155.*

### Description

*Update the FunctionCallBinder.GetUriFunctionSignatures() function accommodate enableCaseInsensitive logic. First scan by exact match for custom Uri functions, followed by case-insensitive match if case-insensitive is enabled. Ensure function name is correctly set in the message when exception is thrown. Updated related logic to make case-insensitive search work along with looking for best signature matched for overloaded functions.*

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
@@ -5,6 +5,7 @@
 //---------------------------------------------------------------------
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Collections.ObjectModel;
 using System.Diagnostics;
@@ -88,33 +89,16 @@ namespace Microsoft.OData.UriParser
         }
 
         /// <summary>
-        /// Overloaded function for 'MatchSignatureToUriFunction' without utilizing output parameter 'matchedIndex'.
-        /// </summary>
-        /// <param name="functionName">The name of the function</param>
-        /// <param name="argumentNodes">The nodes of the arguments, can be new {null,null}.</param>
-        /// <param name="signatures">The signatures to match against</param>
-        /// <returns>Returns the matching signature or throws</returns>
-        internal static FunctionSignatureWithReturnType MatchSignatureToUriFunction(string functionName,
-            SingleValueNode[] argumentNodes,
-            FunctionSignatureWithReturnType[] signatures)
-        {
-            int matchedIndexNotUsed = int.MinValue;
-            return MatchSignatureToUriFunction(functionName, argumentNodes, signatures, out matchedIndexNotUsed);
-        }
-
-        /// <summary>
         /// Finds the signature that best matches the arguments
         /// </summary>
-        /// <param name="functionName">The name of the function</param>
+        /// <param name="functionCallToken">The name of the function</param>
         /// <param name="argumentNodes">The nodes of the arguments, can be new {null,null}.</param>
-        /// <param name="signatures">The signatures to match against</param>
-        /// <param name="matchedIndex">Output for the index of the matched signature.</param>
+        /// <param name="nameSignatures">The name-signature pairs to match against</param>
         /// <returns>Returns the matching signature or throws</returns>
-        internal static FunctionSignatureWithReturnType MatchSignatureToUriFunction(string functionName, SingleValueNode[] argumentNodes,
-            FunctionSignatureWithReturnType[] signatures, out int matchedIndex)
+        internal static KeyValuePair<string, FunctionSignatureWithReturnType> MatchSignatureToUriFunction(string functionCallToken, SingleValueNode[] argumentNodes,
+            IList<KeyValuePair<string, FunctionSignatureWithReturnType>> nameSignatures)
         {
-            FunctionSignatureWithReturnType signature = null;
-            matchedIndex = -1;
+            KeyValuePair<string, FunctionSignatureWithReturnType> nameSignature;
 
             IEdmTypeReference[] argumentTypes = argumentNodes.Select(s => s.TypeReference).ToArray();
 
@@ -125,53 +109,32 @@ namespace Microsoft.OData.UriParser
                 // we specifically want to find just the first function that matches the number of arguments, we don't care about
                 // ambiguity here because we're already in an ambiguous case where we don't know what kind of types
                 // those arguments are.
-                for (int i = 0; i < signatures.Count(); i++)
+                var found = nameSignatures.FirstOrDefault(pair => pair.Value.ArgumentTypes.Count() == argumentCount);
+                if (found.Equals(TypePromotionUtils.NotFoundKeyValuePair))
                 {
-                    if (signatures[i].ArgumentTypes.Count() == argumentCount)
-                    {
-                        // found matching arguments count.
-                        matchedIndex = i;
-                        break;
-                    }
-                }
-
-                if (matchedIndex == -1)
-                {
-                    throw new ODataException(ODataErrorStrings.FunctionCallBinder_CannotFindASuitableOverload(functionName, argumentTypes.Count()));
+                    throw new ODataException(ODataErrorStrings.FunctionCallBinder_CannotFindASuitableOverload(functionCallToken, argumentTypes.Count()));
                 }
                 else
                 {
                     // in this case we can't assert the return type, we can only assert that a function exists... so
                     // we need to set the return type to null.
-                    signature = new FunctionSignatureWithReturnType(null, signatures[matchedIndex].ArgumentTypes);
+                    nameSignature = new KeyValuePair<string, FunctionSignatureWithReturnType>(
+                        found.Key, new FunctionSignatureWithReturnType(null, found.Value.ArgumentTypes));
                 }
             }
             else
             {
-                signature = TypePromotionUtils.FindBestFunctionSignature(signatures, argumentNodes, out matchedIndex);
-                if (signature == null)
+                nameSignature =
+                    TypePromotionUtils.FindBestFunctionSignature(nameSignatures, argumentNodes, functionCallToken);
+                if (nameSignature.Equals(TypePromotionUtils.NotFoundKeyValuePair))
                 {
                     throw new ODataException(ODataErrorStrings.MetadataBinder_NoApplicableFunctionFound(
-                        functionName,
-                        UriFunctionsHelper.BuildFunctionSignatureListDescription(functionName, signatures)));
+                        functionCallToken,
+                        UriFunctionsHelper.BuildFunctionSignatureListDescription(functionCallToken, nameSignatures.Select(sig => sig.Value))));
                 }
             }
 
-            return signature;
-        }
-
-        /// <summary>
-        /// Overloaded function with case insensitive matching disabled and output candidateNames ignored.
-        /// Finds all signatures for the given function name.
-        /// Search in both BuiltIn uri functions and Custom uri functions.
-        /// Combine and return the signatures overloads of the results.
-        /// </summary>
-        /// <param name="functionName">The function to get the signatures for.</param>
-        /// <returns>The signatures which match the supplied function name.</returns>
-        internal static FunctionSignatureWithReturnType[] GetUriFunctionSignatures(string functionName)
-        {
-            string[] candidateNames = null;
-            return GetUriFunctionSignatures(functionName, out candidateNames, /*enableCaseInsensitive*/false);
+            return nameSignature;
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


biaol-odata:
This PR fixes an dependency issue for https://github.com/OData/WebApi/pull/1409.

biaol-odata:
@mikepizzo I have re-factored the code in commit 3bf94c8 per our discussion in design meeting. Can you please take a look? Thanks.

src/Microsoft.OData.Core/UriParser/CustomUriFunctions.cs
```diff
@@ -142,18 +142,62 @@ public static bool RemoveCustomUriFunction(string functionName)
...
             {
-                return CustomFunctions.TryGetValue(name, out signatures);
+                IList<KeyValuePair<string, FunctionSignatureWithReturnType>> bufferedKeyValuePairs
+                    = new List<KeyValuePair<string, FunctionSignatureWithReturnType>>();
+
+                // Do case-sensitive search first.
+                if (CustomFunctions.ContainsKey(functionCallToken))
+                {
+                    FunctionSignatureWithReturnType[] signatureGroup = null;
+                    CustomFunctions.TryGetValue(functionCallToken, out signatureGroup);
```
mikepizzo: Rather than look up functionCallToken twice, why not use boolean returned from TryGetValue, as in:

                // Do case-sensitive search first.
                FunctionSignatureWithReturnType[] signatureGroup = null;
                if (CustomFunctions.TryGetValue(functionCallToken, out signatureGroup))
                {
                       foreach (FunctionSignatureWithReturnType sig in signatureGroup)
                      ...

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
```diff
@@ -597,13 +633,24 @@ private static ICollection<FunctionParameterToken> HandleComplexOrCollectionPara
...
+        private string IsUnboundFunction(string functionName)
         {
-            return UnboundFunctionNames.Contains(functionName);
+            foreach (string key in UnboundFunctionNames)
+            {
+                if (key.Equals(functionName,
+                    this.state.Configuration.EnableCaseInsensitiveUriFunctionIdentifier
+                        ? StringComparison.OrdinalIgnoreCase
+                        : StringComparison.Ordinal))
+                {
```
mikepizzo: since unbound function names are lower case, would be more efficient to cast the value you're looking for to lower case for case insensitive search:

functionName = this.state.Configuration.EnableCaseInsensitiveUriFunctionIdentifier ? functionName.ToLowerInvariant() : functionName
return UnboundFunctionNames.FirstOrDefault(k => k.Equals(functionName, StringComparison.Ordinal));

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
```diff
@@ -106,67 +109,95 @@ internal static FunctionSignatureWithReturnType MatchSignatureToUriFunction(stri
                 // we specifically want to find just the first function that matches the number of arguments, we don't care about
                 // ambiguity here because we're already in an ambiguous case where we don't know what kind of types
                 // those arguments are.
-                signature = signatures.FirstOrDefault(candidateFunction => candidateFunction.ArgumentTypes.Count() == argumentCount);
-                if (signature == null)
+                var found = nameSignatures.FirstOrDefault(pair => pair.Value.ArgumentTypes.Count() == argumentCount);
```
mikepizzo: >var [](start = 16, length = 3)

Please replace var with type (KeyValuePair<string, FunctionSignatureWithReturnType>)

src/Microsoft.OData.Core/UriParser/TypePromotionUtils.cs
```diff
@@ -370,28 +376,33 @@ internal static bool PromoteOperandType(UnaryOperatorKind operatorKind, ref IEdm
...
+        /// <param name="functionCallToken">Function call token used for case-sensitive matching to resolve ambiguous cases.</param>
         /// <returns>The best fitting function; null if none found or ambiguous.</returns>
-        internal static FunctionSignatureWithReturnType FindBestFunctionSignature(FunctionSignatureWithReturnType[] functions, SingleValueNode[] argumentNodes)
+        internal static KeyValuePair<string, FunctionSignatureWithReturnType> FindBestFunctionSignature(
+            IList<KeyValuePair<string, FunctionSignatureWithReturnType>> nameFunctions,
+            SingleValueNode[] argumentNodes, string functionCallToken)
         {
             IEdmTypeReference[] argumentTypes = argumentNodes.Select(s => s.TypeReference).ToArray();
-            Debug.Assert(functions != null, "functions != null");
+            Debug.Assert(nameFunctions != null, "functions != null");
```
mikepizzo: >functions [](start = 49, length = 9)

nit: should be "nameFunctions != null"

src/Microsoft.OData.Core/UriParser/TypePromotionUtils.cs
```diff
@@ -432,24 +444,39 @@ internal static FunctionSignatureWithReturnType FindBestFunctionSignature(Functi
...
                 {
-                    return null;
+                    // Best match count = 1.
+                    result = equallyArgumentsMatchingNameFunctions[0];
+                }
+                else
+                {
+                    // For multiple best matches based on argument types
+                    if (equallyArgumentsMatchingNameFunctions.Count(
+                            _ => _.Key.Equals(functionCallToken, StringComparison.Ordinal)) == 1)
```
mikepizzo: >_ [](start = 28, length = 1)

please use something besides underscore.

src/Microsoft.OData.Core/UriParser/TypePromotionUtils.cs
```diff
@@ -432,24 +444,39 @@ internal static FunctionSignatureWithReturnType FindBestFunctionSignature(Functi
...
+                    {
+                        // There is only one case-sensitive function name match among the equally argument-types matching functions.
+                        result = equallyArgumentsMatchingNameFunctions.First(_ => _.Key.Equals(functionCallToken, StringComparison.Ordinal));
+                    }
```
mikepizzo: You currently loop through the list twice; once to find if there is exactly one match, and once to return it if there is.  Would be more efficient to rewrite L452-L477 as something like:

                KeyValuePair<string, FunctionSignatureWithReturnType> result = NotFoundKeyValuePair;
                if (equallyArgumentsMatchingNameFunctions.Count == 1)
                {
                    // Best match count = 1.
                    result = equallyArgumentsMatchingNameFunctions[0];
                }
                else
                {
                    // For multiple best matches based on argument types
                    foreach (KeyValuePair<string,FunctionSignatureWithReturnType> candidate in equallyArgumentsMatchingNameFunctions)
                    {
                        if (candidate.Key.Equals(functionCallToken, StringComparison.Ordinal))
                        {
                            result = candidate;
                            break;
                        }
                    }
                }

note that this also handles the Count == 0 case.

src/Microsoft.OData.Core/UriParser/CustomUriFunctions.cs
```diff
@@ -142,18 +142,62 @@ public static bool RemoveCustomUriFunction(string functionName)
...
+                            && !func.Key.Equals(functionCallToken, StringComparison.Ordinal))
+                        {
+                            foreach (FunctionSignatureWithReturnType sig in func.Value)
+                            {
+                                // case-insensitive match: canonical name is the one from the CustomFunctions dictionary.
+                                bufferedKeyValuePairs.Add(new KeyValuePair<string, FunctionSignatureWithReturnType>(func.Key, sig));
+                            }
+                        }
+                    }
+                }
```
mikepizzo: Current code loops through CustomFunctions multiple times (once to find if the key exists, once to get the value if it does, and once again for case-insensitive, doing two comparisons for each item in the list). Wouldn't it be simpler (and more efficient, especially for default case of case insensitive) to replace 164-195 with something like:

                foreach (KeyValuePair<string, FunctionSignatureWithReturnType[]> func in CustomFunctions)
                {
                    if (func.Key.Equals(functionCallToken, enableCaseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
                    {
                        foreach (FunctionSignatureWithReturnType sig in func.Value)
                        {
                            bufferedKeyValuePairs.Add(new KeyValuePair<string, FunctionSignatureWithReturnType>(func.Key, sig));
                        }
                    }
                }

biaol-odata:
Thanks for taking a look, Mike!

---
In reply to: [390311565](https://github.com/OData/odata.net/pull/1156#issuecomment-390311565) [](ancestors = 390311565)

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
biaol-odata: Make sense. Thanks.

---
In reply to: [190097377](https://github.com/OData/odata.net/pull/1156#discussion_r190097377) [](ancestors = 190097377)

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
biaol-odata: Fixed other "var "in dozen other places in this file as well. But I heard that using "var " is encouraged these days since type inference was introduced as new C# feature. I am OK with either way, understanding that many developers (including good ones) favors strong type.
I don't think we need to cleanup all files related to this since we have time crunch. We might need team discussion when topic related to coding standard comes up.

---
In reply to: [190097866](https://github.com/OData/odata.net/pull/1156#discussion_r190097866) [](ancestors = 190097866)

src/Microsoft.OData.Core/UriParser/TypePromotionUtils.cs
biaol-odata: This is a tricky case. In the case of multiple case-sensitive best matches, we should return invalid resolution ( = NotFoundKeyValuePair) instead of returning the first match.
The computational complexity should be the same O(N), and linear should be the minimum in this case.

---
In reply to: [190104069](https://github.com/OData/odata.net/pull/1156#discussion_r190104069) [](ancestors = 190104069)

biaol-odata:
:clock1:

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/CustomUriFunctionsTests.cs
```diff
@@ -509,6 +510,141 @@ FunctionSignatureWithReturnType myStringFunction
...
+                startsWithArgs[1].ShouldBeConstantQueryNode("BlaBla");
+            }
+            finally
+            {
+                CustomUriFunctions.RemoveCustomUriFunction("myFirstMixedCasestringfunction").Should().BeTrue();
+            }
+        }
+
+        [Fact]
+        public void ParseWithExactMatchCustomUriFunction_EnableCaseInsensitive_ShouldWorkForMultipleEquiventArgumentsMatches()
```
AlanWong-MS: typo: 
ShouldWorkForMultipleEquiventArgumentsMatches
ShouldWorkForMultipleEquiv**al**entArgumentsMatches

src/Microsoft.OData.Core/UriParser/TypePromotionUtils.cs
```diff
@@ -370,28 +376,33 @@ internal static bool PromoteOperandType(UnaryOperatorKind operatorKind, ref IEdm
...
+            IList<KeyValuePair<string, FunctionSignatureWithReturnType>> nameFunctions,
+            SingleValueNode[] argumentNodes, string functionCallToken)
         {
             IEdmTypeReference[] argumentTypes = argumentNodes.Select(s => s.TypeReference).ToArray();
-            Debug.Assert(functions != null, "functions != null");
+            Debug.Assert(nameFunctions != null, "nameFunctions != null");
             Debug.Assert(argumentTypes != null, "argumentTypes != null");
-            List<FunctionSignatureWithReturnType> applicableFunctions = new List<FunctionSignatureWithReturnType>(functions.Length);
+            Debug.Assert(functionCallToken != null, "functionCallToken != null");
+            IList<KeyValuePair<string,FunctionSignatureWithReturnType>> applicableNameFunctions
```
AlanWong-MS: nit: missing space between <string,FunctionSignatureWithReturnType>

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/CustomUriFunctionsTests.cs
biaol-odata: good catch. Thanks.

---
In reply to: [190790144](https://github.com/OData/odata.net/pull/1156#discussion_r190790144) [](ancestors = 190790144)

biaol-odata:
:clock1:

biaol-odata:
:clock1:

src/Microsoft.OData.Core/UriParser/TypePromotionUtils.cs
mikepizzo: In what scenario would we have multiple case-sensitive best matches?

In order to handle multiple case-sensitive best matches, could you remove the "break" in the suggested code above and throw your exception if result!=NotFoundKeyValuePair prior to setting result=candidate? That would still be more efficient than looping through the list twice.

---
In reply to: [190715400](https://github.com/OData/odata.net/pull/1156#discussion_r190715400) [](ancestors = 190715400,190104069)

biaol-odata: Sorry the ambiguous result case should be for the multiple case insensitive match, not case-sensitive matches.
Given that there can be at most one case-sensitive match candidate, the multiple matches resolution can be simpler.

---
In reply to: [191602830](https://github.com/OData/odata.net/pull/1156#discussion_r191602830) [](ancestors = 191602830,190715400,190104069)

biaol-odata:
:clock1:

biaol-odata:
Merged. Thanks for the code review!

Issue #1157, opened 2018-05-02
ODataMessageReader doesn't honor the absense of ValidationKinds.ThrowIfTypeConflictsWithMetadata flag


dozer75:
If a `ODataMessageReader` is created with settings where `ValidationKinds.ThrowIfTypeConflictsWithMetadata` flag isn't set in `ODataMessageReaderSettings.Validations` option it still checks for nulls (if configured by metadata) and fails.

### Assemblies affected

Several versions of `Microsoft.OData.Core` including the latest one 7.4.4.

### Reproduce steps

This **[gist](https://gist.github.com/dozer75/4cf219e041681de687087eb2d66b8743)** shows the problem, create a .NET Core Console app, add this code and add the `Microsoft.OData.Core` nuget package to the project and run. Line 45 will throw an error even though it shouldn't since `ODataMessageReaderSettings.Validations` doesn't have the `ValidationKinds.ThrowIfTypeConflictsWithMetadata` flag set.

### Expected result

When `ODataMessageReaderSettings.Validations` is missing the `ValidationKinds.ThrowIfTypeConflictsWithMetadata` flag no metadata checks should be done at all any places.

### Actual result

An exception is thrown that states `A null value was found for the property named 'Name', which has the expected type 'Edm.String[Nullable=False]'. The expected type 'Edm.String[Nullable=False]' does not allow null values.`

### Additional detail

The problem seems to be at **[this](https://github.com/OData/odata.net/blob/bf1f8c2a3a9b72d9fba05853ecb01f425abfeaca/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs#L1578)** line, it should check for `JsonLightInputContext.MessageReaderSettings.ThrowIfTypeConflictsWithMetadata` too and everything works.

Frankly, really I don't see the point with the section starting with the link above, those checks are already done within `ODataJsonReaderCoreUtils.TryReadNullValue` anyway, so it seems to be unnecessary code. I tried to find out why this has been added but couldn't find it.

AlanWong-MS:
@dozer75, thanks for finding this issue and providing a PR! It looks like there's a failed regression test `Microsoft.Test.Taupo.OData.Reader.Tests.Reader.ComplexValueReaderTests.ComplexValueIgnorePropertyNullValuesTest`; it could be that we need to update an expected result somewhere. Can you please take a look at this within the `OData.Tests.E2E.sln` solution file? Let me know if you need assistance.

[...]

PR #1158, opened 2018-05-02
Added check if the MessageReaderSettings.ThrowIfTypeConflictsWithMetadata

dozer75:
... is set to ensure that any metadata related checks are done if the flag is set, otherwise ignored.

### Issues

*This pull request fixes issue #1157 .*

### Description

Fixes an issue where metadata related checks on null was done even if the validation settings explicit states that such checks should be done.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
@@ -1575,7 +1575,7 @@ namespace Microsoft.OData.JsonLight
             // Try to read a null value
             if (ODataJsonReaderCoreUtils.TryReadNullValue(this.JsonReader, this.JsonLightInputContext, targetTypeReference, validateNullValue, propertyName, isDynamicProperty))
             {
-                if (validateNullValue && (targetTypeReference != null) && (!targetTypeReference.IsNullable))
+                if (this.JsonLightInputContext.MessageReaderSettings.ThrowIfTypeConflictsWithMetadata && validateNullValue && targetTypeReference != null && !targetTypeReference.IsNullable)
                 {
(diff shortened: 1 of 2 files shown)


dozer75:
Sure! I ran the unit tests and they was ok (except that they don't support non-US regional settings, but that's a different story), but for some reason E2E fails totally for me to compile, I suspect that it is something I am missing on my computer(s), the strange thing is that I get different errors on different computers.

On this computer I get a build log looking like the this file [msbuild.log](https://github.com/OData/odata.net/files/1968967/msbuild.log)

If you can point me in the right direction on how to fix this I'll fix the E2E tests...

AlanWong-MS:
For the machine that has the build errors from the log file file you attached, can you unload the DataWebRules.csproj in the solution? I have a separate PR that addresses that build issue and is waiting for commit. Hopefully disabling it will resolve your build issues.

dozer75:
@AlanWong-MS I fixed the E2E test, however the more I look at this, I think that the `if` statement at line 1578 should be removed there is no scenario what I can see that it will be triggered. I have a commit with it that I can push into this PR if you don't see any issue with it.

AlanWong-MS:
Let's keep the `if` statement. I tried to look for the original commit that added the check to find more information but it seems deep in our commit history (before 6.x version); I think it's safer to leave it there. Also, based on the conditions
`if (validateNullValue && (targetTypeReference != null) && (!targetTypeReference.IsNullable))`

`validateNullValue` is passed into the function as a parameter and I see that there are cases where it is true and false. It indicates to me that there is value to the `if` statement.

Otherwise, just a quick style comment in the code and everything looks good. :)

dozer75:
Style change fixed and committed. Ok, let's keep the if then, but I don't think it actually will be executed anymore.

AlanWong-MS:
Thanks for the note! Merged.

PR #1159, opened 2018-05-02
remove the Document element from Edm lib

xuzhg:
### Issues

* See issue https://github.com/OData/WebApi/issues/26

### Description

OData V3 spec has:

2.3 The edm:Documentation Element
The edm:Documentation element allows service authors to provide documentation for most model elements.
A model element MUST NOT contain more than one documentation element.
Refer to the [XML schema][csdl19] for details on which model elements support documentation.
A documentation element MUST contain zero or one edm:Summary and zero or one edm:LongDescription elements. The summary and long description elements MAY contain text that serves as the summary or long description. If both a summary and long description are provided, the summary MUST precede the long description.
For example:

```xml
<EntityType Name="Product"> 
 <Documentation>
  <Summary>Product names, suppliers, prices, and units in stock.</Summary>
  <LongDescription>...</LongDescription>
 </Documentation>
 ...
</EntityType>
```

However, that's not in OData Spec v4.

This PR is trying to remove the corresponding codes.

- [x] *Test cases added*

### Additional work necessary


mikepizzo:
Just to confirm; this is just code clean-up with no change to public surface area, right?  I didn't see any change to public classes or to the api baseline file, but I just wanted to verify.

Issue #1147, opened 2018-04-09
Add support for IN operator


timpengu:
The [OData Version 4.01](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html) specification adds support for [IN expressions](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_in) such as:

`http://host/service/Products?$filter=Name in ('Milk', 'Cheese')`

ODataLib should support this operator syntax.

### Assemblies affected

ODataLib 7.4.4

### Additional details

Although this can currently be be achieved using multiple OR clauses, some benefits of IN expressions are:
* much shorter query string when filtering on a large set of values (max URI length is limited)
* often simpler to translate to efficient queries on an underlying data source (e.g. I am translating OData requests to a MongoDB backend where a single $in query executes much faster than the equivalent $or queries).

AlanWong-MS:
This one's in progress by me. :) It's currently in the design phase. I'll use this issue as a way to maintain updates. Thanks for opening it!

NetTecture:
Has anything happened on this in the last 6 weeks? I consider this one of the most "wtf" shortcomings in the OData Specs (how the heck could they overlook that) which finally is closed - and am desperately waiting on being able to use it.

NetTecture:
@AlanWong-MS Well, face it - OData (the protocol) is quite interesting in capabilities. Nice how it took so long actually for someone to realize that remove functions (even in REST shape) re worth nearly nothing unless you ass query and projection syntax to the definition, so you do not end up with dozens of nearly redundant methods ;) The reason I am so behind this is that next week I set the basics for a new project and start implementing the API There in dotnet core 2.1 odata - at least until I run into too many problems that don't get fixes to make sense.... which I di not expect ;) Good to know WebApi OData 7 is getting along.

mikepizzo:
Yes, this is a feature that we are certainly interested in getting added -- thanks for all of your enthusiasm!

evilB:
I'm looking forward to being able to use the in clause, currently running into maxnode isssues due to using multiple or sequences in the filter. 

i've subscribed to remain informed

AlanWong-MS:
This feature has been merged into master via #1165. Our release with the feature is coming up :)

timpengu:
Excellent, thanks Alan!  Looking forward to integrating this into our system, and persuading our clients to  rewrite their queries :)

PR #1165, opened 2018-05-15
[Feature] IN operator

AlanWong-MS:
### Issues
#1147

### Description
Implementation of the IN operator as part of the [OData 4.01 spec](http://docs.oasis-open.org/odata/new-in-odata/v4.01/cn01/new-in-odata-v4.01-cn01.html#_Toc485385090) (see [ABNF](http://docs.oasis-open.org/odata/odata/v4.01/csprd04/abnf/odata-abnf-construction-rules.txt)).

This PR includes the IN operator feature, which includes a new class InNode, modifications to the expression lexer to recognize the "in" keyword, and unit and functional test cases.

**Exclusions of implementation:**

- The ABNF indicates that brackets are supported for enclosing the lists; however, this particular feature is currently excluded from implementation due to some breaking changes in code. For now, lists are supported only with parentheses.

- [x] *Test cases added*


src/Microsoft.OData.Core/Json/JsonReader.cs
```diff
@@ -285,7 +285,7 @@ public virtual bool Read()
                     }
 
                     // We might see end of array here
-                    if (this.characterBuffer[this.tokenStartIndex] == ']')
+                    if (this.characterBuffer[this.tokenStartIndex] == ']' || this.characterBuffer[this.tokenStartIndex] == ')')
                     {
```
mikepizzo: Will this also allow me to terminate a JSON array with a closing parens?   i.e., does "names": ["bob","sue") now parse as valid?

mikepizzo:
>         public virtual bool Read()

Let's talk through how/where JsonReader gets called for this case; it would be nice not to have to add parens parsing to the JSON reader, as it is not a JSON concept.

---
Refers to: src/Microsoft.OData.Core/Json/JsonReader.cs:220 in becc74f. [](commit_id = becc74f0549b797c9537a2cd450c96b77cdfb2f2, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/SemanticAst/InNode.cs
```diff
@@ -0,0 +1,117 @@
...
+        /// <exception cref="System.ArgumentNullException">Throws if the left or right inputs are null.</exception>
+        /// <exception cref="ODataException">Throws if the right operand single item type isn't the same type as the left operand.</exception>
+        public InNode(SingleValueNode left, CollectionNode right)
+        {
+            ExceptionUtils.CheckArgumentNotNull(left, "left");
+            ExceptionUtils.CheckArgumentNotNull(right, "right");
+            this.left = left;
+            this.right = right;
+
+            if (this.left.GetEdmTypeReference().Definition != this.right.ItemType.Definition)
```
mikepizzo: >this.left.GetEdmTypeReference().Definition != this.right.ItemType.Definition [](start = 16, length = 76)

We should support the type of the right node collection being comparable to the type of the left node (including being derived from the type of the left node).  See EdmLibraryExtensions.IsAssignableFrom.

src/Microsoft.OData.Core/Json/JsonReader.cs
AlanWong-MS: Great question. The answer is no and I don't have an existing test case for this but I've tested it. I have a test case for the vice-versa scenario (since we currently don't allow bracketed collections in IN expression):

```csharp
public void FilterWithInOperationWithMismatchedClosureCollection()
{
    Action parse = () => ParseFilter("ID in (1,2,3]", HardCodedTestModel.TestModel, HardCodedTestModel.GetPersonType());
    parse.ShouldThrow<ODataException>().WithMessage(ODataErrorStrings.ExpressionLexer_UnbalancedBracketExpression);
}
```
I added the test case that you described with the closing parens and will update the PR after we discuss the other items that you have. No product code changes needed as the parser already takes care of this case.

src/Microsoft.OData.Core/UriParser/SemanticAst/InNode.cs
AlanWong-MS: Thanks for the catch, I think I had forgotten to jot down the point regarding derived types during code review. Regarding supporting comparable types, is there a function or API that you're specifically referring to or is what I currently have in the code of comparing the definitions sufficient?

Issue #513, opened 2016-03-23
$select with complexCol/prop should be supported


chinadragon0515:
Per specification, $select with complexCol/prop should be supported
### Assemblies affected

OData .Net lib 6.15
### Reproduce steps

Use URLParser to parse a URL with $select=ComplexCol/prop
### Expected result

parsing succeed
### Actual result

exception thrown

brjohnstmsft:
This fix is necessary to get $select working with properties under complex collection properties through Web API too. See related issues: https://github.com/OData/WebApi/issues/989 https://github.com/OData/WebApi/issues/196

brjohnstmsft:
I've determined through testing that this issue was fixed by this commit from July 2016: https://github.com/OData/odata.net/commit/7f8c2f30c02fdfb8773a6319fe24cbbd6277dc39

I'll be sending a PR soon with test coverage to prevent regressions of this scenario.

PR #1166, opened 2018-05-15
Adding test coverage for selecting complex collection properties

brjohnstmsft:
### Issues

This pull request adds test coverage to confirm that this issue has been fixed: https://github.com/OData/odata.net/issues/513

### Description

This change adds new test coverage for the scenario of selecting a property of a complex collection. For example: `$select=PreviousAddresses/City` where `PreviousAddress` is of type `Collection(Address)` and `Address` is a complex type.

Originally I thought this scenario was still broken, but it looks like it was fixed by an unrelated change almost two years ago. I think it would be prudent to add test coverage for this case to prevent future regressions.

- [X] *Test cases added*


brjohnstmsft:
The build failure looks unrelated to my changes.

mikepizzo:
Looks good; thanks Bruce!

Issue #1145, opened 2018-04-06
Memory Leak in Library?


rbarten:
We have a Windows Service which reads out messages from a MSMQ en uses the OData.Net libraries to Add and Update data in an external system. The Memory Allocation of the OData.Net is increasing fast when we do batch actions using the OData.Net libary.
Especcialy the Dictionary with LinkInfo within the 'EntityDescriptor' and the List of ODataNestedResourceInfo' with MaterializerEntry (FeedAndEntryMaterializerAdapter). (See screenshot below)
This memory is never released and after some time a few GB of memory is allocated.
Is there a way to overcome this? Or should this be fixed within the library?

### Assemblies affected

OData .Net lib 7.4.4 and earlier.

### Reproduce steps

Create an application/service which is doing a lot of CRUD actions using the library.

### Expected result

Memory Allocation should be released when DataServiceContext is disposed.

### Actual result

This memory is never released and after some time a few GB of memory is allocated.
See screenshots below when application was running for a short time:
![memory allocation1](https://user-images.githubusercontent.com/2169036/38407330-4b7f8bee-3979-11e8-8c74-c22e5df7898f.png)

![memory allocation2](https://user-images.githubusercontent.com/2169036/38407341-56ab2ef6-3979-11e8-9cd7-5871b35374d1.png)

mprigge:
I am also having a similar problem with 7.4.4 with an application written to do something similar (sync from an external source to an upstream OData API). I haven't been able to revert to 6.x.x to test that, but I'm relatively certain this became a problem after upgrading to 7.x. I can provide whatever level of detail you'd find helpful in troubleshooting.

kgjersda:
Having a similar problem. W3WP uses many gigabytes of ram.

davidmilligan:
I can also confirm this memory leak. The static dictionary in ODataAnnotatableExtensions seems to be the source of the problem (though there may be others). It looks like this bug was possibly introduced in commit 0aad86f912aa87264ad0fa98ba6596b96731cd37

mathiasbl:
I'm also noticing an increased memory usage with finally an out of memory exception.

AartBluestoke:
the old replaced code in 0aad86f had a codepath of { public void SetAnnotation(...) {
            if (annotation == null)	
            {	
                RemoveAnnotation<T>();	
            }	
...
    }
The new code has no code paths for removing annotations, and infact asserts that the request is never null, meaning that once you searilize an entity for transmistion, creating the 
[...]

PR #1171, opened 2018-05-29
fix memory leak #1145

AartBluestoke:
### Issues

*This pull request fixes issue #1145.*

### Description

the memory leak occurs because the Annotatable object can never be GC'd once an annotation is attached.

There is a change in semantics here, if SetAnnotation is called twice, the second will exception, rather than updating.
This was previously a condition which would cause an error anyway, as it would orphan the previously set annotation inside CreateLink()

At the moment the only codepath to SetAnnotation is 
Microsoft.OData.Client.Materialization.MaterializerNavigationLink.CreateLink(...) 
create link is only called by FeedAndEntryMaterializerAdapter, which has an existence guard, so this should never trigger the double-insert.

- [ ] *Test cases added*

### Additional work necessary

There is a possible change in behaviour as replaceing  Dictionary[annotatable] = annotation; with Dictionary.Add(annotatable,annotation);  means that setting the same annotation twice is now an error. I do not believe we currently do that anywhere.

### DIFF
--- a/src/Microsoft.OData.Client/ODataAnnotatableExtensions.cs
+++ b/src/Microsoft.OData.Client/ODataAnnotatableExtensions.cs
@@ -4,7 +4,7 @@
 // </copyright>
 //---------------------------------------------------------------------
 
-using System.Collections.Concurrent;
+using System.Runtime.CompilerServices;
 using System.Diagnostics;
 
 namespace Microsoft.OData.Client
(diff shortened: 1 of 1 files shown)


src/Microsoft.OData.Client/ODataAnnotatableExtensions.cs
```diff
@@ -30,12 +30,12 @@ public static T GetAnnotation<T>(this ODataAnnotatable annotatable)
 
         private static class InternalDictionary<T> where T : class
         {
-            private static readonly ConcurrentDictionary<ODataAnnotatable, T> Dictionary =
-                new ConcurrentDictionary<ODataAnnotatable, T>();
+            private static readonly ConditionalWeakTable<ODataAnnotatable, T> Dictionary =
+                new ConditionalWeakTable<ODataAnnotatable, T>();
 
             public static void SetAnnotation(ODataAnnotatable annotatable, T annotation)
```
mikepizzo: >SetAnnotation [](start = 31, length = 13)

Change looks reasonable.

Since we are changing semantics on this internal class, I wonder if we should rename this to "AddAnnotation" (and perhaps add an assert that the table doesn't already contain the specified annotatable?)

Issue #1172, opened 2018-05-30
Perf: OData Property Cache Handler can avoid allocating more strings


ysanghi:
Odata.PropertyCacheHandler.GetProperty can be optimized to avoid string allocations.

### Assemblies affected

OData .NET v7.x 

### Reproduce steps
Perf optimization.

### Expected result

Less or not string allocations.

### Actual result

String allocations happening.

xuzhg:
I merged your PR. Thanks your contribution.

biaol-odata:
From PropertyCacheHandlerTests::PropertyHandlerGetProperty(), there is an assert failure at this line:
            Debug.Assert(depth >= 1, $"{nameof(depth)} should always be greater than 1");
where value of depth is 0.

Should we change L 41-42 in PropertyCacheHandler.cs with the following:
            Debug.Assert(depth >= 0, $"'depth' should always be greater than or equal to 0");
            string depthStr = depth == 0 ? string.Empty : depth.ToString(CultureInfo.InvariantCulture);

biaol-odata:
Close as merged.

PR #1173, opened 2018-05-30
Avoid use of string builder in GetProperty to reduce string allocations.

ysanghi:
### Issues

This pull request fixes issue #1172 .

### Description

Instead of using string builder moving everything to string concat.

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/PropertyCacheHandler.cs
+++ b/src/Microsoft.OData.Core/PropertyCacheHandler.cs
@@ -10,6 +10,8 @@ using Microsoft.OData.Edm;
 
 namespace Microsoft.OData
 {
+    using System.Globalization;
+
     /// <summary>
     /// Manage PropertyCache for ODataResourceSet in serialization.
     /// One ODataResourceSet has one PropertyCache.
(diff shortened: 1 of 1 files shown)


src/Microsoft.OData.Core/PropertyCacheHandler.cs
```diff
@@ -32,21 +35,14 @@ internal class PropertyCacheHandler
...
         public PropertySerializationInfo GetProperty(string name, IEdmStructuredType owningType)
         {
-            StringBuilder uniqueName = new StringBuilder();
-            if (owningType != null)
-            {
-                uniqueName.Append(owningType.FullTypeName());
-                uniqueName.Append("-");
-            }
+            int depth = this.currentResourceScopeLevel - this.resourceSetScopeLevel;
+            string depthStr = depth == 1 ? string.Empty : depth.ToString(CultureInfo.InvariantCulture);
```
mikepizzo: >depthStr = depth == 1 [](start = 19, length = 21)

Perhaps add an assert that depth >= 1?

ysanghi: Thanks Mike. Added a debug assert.

ysanghi:
Thanks 👍

PR #1174, opened 2018-05-30
Fixing missing JSON batch doc

AlanWong-MS:
### Description
JSON batch doc was missing due to the way the index.html parses for posts.

### DIFF
--- a/v7/_posts/odata features/2017-12-26-06-00-JsonBatch.md
+++ b/v7/_posts/odata features/2017-12-26-06-00-JsonBatch.md
@@ -1,8 +1,8 @@
 ---
-title: "Json batch format in ODataLib"
 layout: post
-category: "6. OData Features"
+title: "Json batch format in ODataLib"
 description: ""
+category: "5. OData Features"
 ---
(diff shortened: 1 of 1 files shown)


biaol-odata:
@AlanWong-MS Thanks for fixing this! By changing the Category to "5", will it then show up in the http://odata.github.io/odata.net or somewhere? I notice that there are other V7 posts are not shown, likely starting from https://github.com/OData/odata.net/blob/gh-pages/v7/_posts/odata%20features/2016-08-23-06-18-navigation-under-complex.md.

Wondering we are having a doc build process in the pipeline.

AlanWong-MS:
I double checked the other feature docs and they all seem to be there. The format is different for the v6 and v7 posts so I think it was just an accident to use the v6 format. The docs are all put together during runtime.

biaol-odata:
I see. Looks like the part of the gh-pages/v7/_posts/odata%20features are shown in http://odata.github.io/odata.net, the rest in http://odata.github.io/odata.net/v7.

PR #1176, opened 2018-06-01
Adding IN operator docs

AlanWong-MS:
### Description
Adding IN Operator design document and samples to support #1165.


biaol-odata:
Actually, this doc PR should be merged after the corresponding code PR is release, instead of merging it now.
My previous note is more about whether we have reached consensus about the location. Thanks.

mikepizzo:
We should talk as a team about how deeply to document features as part of gh-pages.

This is a very deep technical description of the behavior of the in operator, which is great for the design doc, but is inconsistent with how we document other features.  See, for example, the documentation on the [$apply operator](http://odata.github.io/odata.net/#06-16-aggregation-support).  If we described every operator in the depth that we describe here, the document would be pretty unwieldy (this is mitigated by the TOC, but even then reading through the document is inconsistent).  I would have expected:
1) In release notes, we describe this as a new feature and give examples of how this is used
2) In design docs we describe the feature in depth, including how it is implemented
3) Somewhere in gh_pages we list the operators we support, and point to the odata spec for deeper description on how they are used.

PR #1182, opened 2018-06-06
Fix a wrong if condition

mrdotmoon:
### Issues

*Fix a wrong if condition*

### DIFF
--- a/src/Microsoft.OData.Client/Materialization/InstanceAnnotationMaterializationPolicy.cs
+++ b/src/Microsoft.OData.Client/Materialization/InstanceAnnotationMaterializationPolicy.cs
@@ -146,7 +146,7 @@ namespace Microsoft.OData.Client.Materialization
         internal IDictionary<string, object> ConvertToClrInstanceAnnotations(ICollection<ODataInstanceAnnotation> instanceAnnotations)
         {
             var clientInstanceAnnotationValue = new Dictionary<string, object>(StringComparer.Ordinal);
-            if (instanceAnnotations != null || instanceAnnotations.Count > 0)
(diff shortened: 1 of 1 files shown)


src/Microsoft.OData.Client/Materialization/InstanceAnnotationMaterializationPolicy.cs
```diff
@@ -146,7 +146,7 @@ internal void SetInstanceAnnotations(string navigationPropertyName, ODataResourc
         internal IDictionary<string, object> ConvertToClrInstanceAnnotations(ICollection<ODataInstanceAnnotation> instanceAnnotations)
         {
             var clientInstanceAnnotationValue = new Dictionary<string, object>(StringComparer.Ordinal);
-            if (instanceAnnotations != null || instanceAnnotations.Count > 0)
+            if (instanceAnnotations != null && instanceAnnotations.Count > 0)
```
biaol-odata: Should we just remove the "&& instanceAnnotations.Count > 0" check completely?

biaol-odata:
@mrdotmoon Have you got a chance to look at the comment above and see whether it makes sense? Thanks.

src/Microsoft.OData.Client/Materialization/InstanceAnnotationMaterializationPolicy.cs
mikepizzo: Yes; we should be able to get rid of the instanceAnnotations.Count check all together.

mrdotmoon:
@biaol-odata yeah, it makes sense. Updated, thanks.

Issue #1191, opened 2018-06-13
Avoid String allocations when writing strings that require special character handling


ysanghi:
JsonValueUtils.WriteEscapedJsonString allocates additional strings when dealing with special characters.  

### Assemblies affected

OData .NET LIB

### Additional detail

JsonValueUtils WriteEscapedJsonString uses Substrings to write strings to underlying textwriter but because of the immutable nature of .NET strings it allocates new strings. To further this, strings are objects in .NET and this makes objects visible to GC. This lead high allocations and excessive work needed by GC to clean this objects up. This can be avoided by using a buffer.

AlanWong-MS:
Thanks for filing the issue and opening a PR! The fix has now been merged and should be included in our upcoming ODL release. ETA for the release is slated for end of the month.

PR #1192, opened 2018-06-13
Avoid String allocations when writing strings that require special character handling 

ysanghi:
### Issues

*This pull request fixes issue #1191 .*

### Description

1 Created a local char array buffer to buffer the read characters when dealing with strings that require special handling.
2 using char array instead of string when writing to underlying stream

- [X] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Buffers/BufferUtils.cs
+++ b/src/Microsoft.OData.Core/Buffers/BufferUtils.cs
@@ -1,0 +1,35 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="IBufferPool.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Buffers
+{
+    /// <summary>
+    /// Helpers to deal with buffers
+    /// </summary>
+
+    public static class BufferUtils
+    {
+        /// <summary>
+        /// Buffer length
+        /// </summary>
+        private const int BufferLength = 128;
+
+        /// <summary>
+        /// Checks if the buffer is not initialized and if initialized returns the same buffer or creates a new one.
+        /// </summary>
+        /// <param name="buffer">The buffer to verify</param>
+        /// <returns>The initialized buffer</returns>
+        public static char[] InitializeBufferIfRequired(char[] buffer)
+        {
+            if (buffer == null)
+            {
+                return new char[BufferUtils.BufferLength];
+            }
+
+            return buffer;
+        }
+    }
+}
--- a/src/Microsoft.OData.Core/Buffers/BufferUtils.cs
+++ b/src/Microsoft.OData.Core/Buffers/BufferUtils.cs
@@ -12,7 +12,7 @@ namespace Microsoft.OData.Buffers
     /// Helpers to deal with buffers
     /// </summary>
 
-    public static class BufferUtils
+    internal static class BufferUtils
     {
         /// <summary>
         /// Buffer length
--- a/src/Microsoft.OData.Core/Buffers/BufferUtils.cs
+++ b/src/Microsoft.OData.Core/Buffers/BufferUtils.cs
@@ -1,11 +1,13 @@
 ﻿//---------------------------------------------------------------------
-// <copyright file="IBufferPool.cs" company="Microsoft">
+// <copyright file="BufferUtils.cs" company="Microsoft">
 //      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
 // </copyright>
 //---------------------------------------------------------------------
 
 namespace Microsoft.OData.Buffers
 {
+    using System;
+
     /// <summary>
     /// Helpers to deal with buffers
     /// </summary>
--- a/src/Microsoft.OData.Core/Buffers/BufferUtils.cs
+++ b/src/Microsoft.OData.Core/Buffers/BufferUtils.cs
@@ -11,7 +11,6 @@ namespace Microsoft.OData.Buffers
     /// <summary>
     /// Helpers to deal with buffers
     /// </summary>
-
     internal static class BufferUtils
     {
         /// <summary>
--- a/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
@@ -14,7 +14,6 @@ namespace Microsoft.OData.Json
     using System.IO;
     using System.Text;
     using System.Xml;
-
     using Microsoft.OData.Buffers;
     using Microsoft.OData.Edm;
     #endregion Namespaces
--- a/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
@@ -14,6 +14,8 @@ namespace Microsoft.OData.Json
     using System.IO;
     using System.Text;
     using System.Xml;
+
+    using Microsoft.OData.Buffers;
     using Microsoft.OData.Edm;
     #endregion Namespaces
 
@@ -35,7 +37,7 @@ namespace Microsoft.OData.Json
         /// <summary>
         /// The NumberFormatInfo used in OData Json format.
         /// </summary>
-        internal static readonly NumberFormatInfo ODataNumberFormatInfo = InitializeODataNumberFormatInfo();
+        internal static readonly NumberFormatInfo ODataNumberFormatInfo = JsonValueUtils.InitializeODataNumberFormatInfo();
 
 
         /// <summary>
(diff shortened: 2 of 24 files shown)


AlanWong-MS:
@ysanghi, thanks for the PR! It turns out that the buddy build failed with the following message:

```
2018-06-13T05:20:13.5225149Z "sln\OData.Net35.sln" (default target) (1) ->
2018-06-13T05:20:13.5377441Z "src\Microsoft.OData.Core\Build.Net35\Microsoft.OData.Core.NetFX35.csproj" (default target) (2) ->
2018-06-13T05:20:13.5377441Z (CoreCompile target) -> 
2018-06-13T05:20:13.5377441Z  src\Microsoft.OData.Core\Json\JsonValueUtils.cs(18,27): error CS0234: The type or namespace name 'Buffers' does not exist in the namespace 'Microsoft.OData' (are you missing an assembly reference?) [src\Microsoft.OData.Core\Build.Net35\Microsoft.OData.Core.NetFX35.csproj]
```
We have to add the same files to `Microsoft.OData.Core.NetFX35.csproj` as well as `src\Microsoft.OData.Core\Build.NetStandard\Microsoft.OData.Core.NetStandard.csproj`.

src/Microsoft.OData.Core/Buffers/BufferUtils.cs
```diff
@@ -0,0 +1,35 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="IBufferPool.cs" company="Microsoft">
```
AlanWong-MS: Please update the filename to BufferUtils.cs

src/Microsoft.OData.Core/Buffers/BufferUtils.cs
```diff
@@ -0,0 +1,35 @@
...
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Buffers
+{
+    /// <summary>
+    /// Helpers to deal with buffers
+    /// </summary>
+
```
AlanWong-MS: Please remove extra line here

src/Microsoft.OData.Core/Buffers/BufferUtils.cs
```diff
@@ -0,0 +1,35 @@
...
+    public static class BufferUtils
```
AlanWong-MS: This should be internal

test/FunctionalTests/Microsoft.OData.Core.Tests/BufferUtilsTests.cs
```diff
@@ -0,0 +1,42 @@
...
+// <copyright file="BufferUtilsTests.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+
+namespace Microsoft.OData.Tests
+{
+    using FluentAssertions;
+
```
AlanWong-MS: Please remove the empty lines

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsTests.cs
```diff
@@ -7,8 +7,11 @@
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+
```
AlanWong-MS: Same comment regarding empty lines

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/NonIndentedTextWriterTests.cs
```diff
@@ -7,8 +7,11 @@
 using System.Collections.Generic;
 using System.Globalization;
 using System.IO;
+
```
AlanWong-MS: Same commend regarding empty lines

src/Microsoft.OData.Core/Buffers/BufferUtils.cs
ysanghi: Done.

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -433,15 +488,44 @@ internal static string GetEscapedJsonString(string inputString)
             return builder.ToString();
         }
 
+        /// <summary>
+        /// Checks if the string contains special char and returns the find the first
```
AlanWong-MS: typo: "returns the find the first" -> "returns the first"

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -433,15 +488,44 @@ internal static string GetEscapedJsonString(string inputString)
             return builder.ToString();
         }
 
+        /// <summary>
+        /// Checks if the string contains special char and returns the find the first 
+        /// index of special char if present.
+        /// </summary>
+        /// <param name="inputString">string that might contain special characters.</param>
+        /// <param name="firstIndex">first index of the special char</param>
```
AlanWong-MS: We should have a default value if no special characters are found. I think we can use -1.

AlanWong-MS: We should indicate the expected value of firstIndex if no special characters are found. I think it may be preferable to set it to -1 as it's a constant value versus the length of the input string, which changes.

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -35,7 +36,7 @@ internal static class JsonValueUtils
         /// <summary>
         /// The NumberFormatInfo used in OData Json format.
         /// </summary>
-        internal static readonly NumberFormatInfo ODataNumberFormatInfo = InitializeODataNumberFormatInfo();
+        internal static readonly NumberFormatInfo ODataNumberFormatInfo = JsonValueUtils.InitializeODataNumberFormatInfo();
```
AlanWong-MS: There's a bit of code churn from adding the JsonValueUtils namespace. For our coding style, we can leave it out if the function is within the same namespace.

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -335,42 +337,95 @@ internal static void WriteValue(TextWriter writer, byte[] value)
...
+                        buffer[bufferIndex] = c;
+                        bufferIndex++;
+                    }
+                    else
+                    {
+                        // Okay, an unhandled character was deteced.
+                        // First lets check if we can fit it in the existing buffer, if not,
+                        // flush the current buffer and reset. Add the escaped string to the buffer
+                        // and continue.
+                        int escapedStringLength = escapedString.Length;
```
AlanWong-MS: We currently don't have any escaped strings that exceed the buffer length (and I don't anticipate that we do), but in the slim chance that our assumption is not valid in the future, perhaps we can add an assert here (something like escapedStringLength >= bufferLength). Otherwise, we walk into unknown behavior from writing past the buffer boundary.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsTests.cs
```diff
@@ -98,11 +98,71 @@ public void WriteMultipleSpecialCharactersShouldWork()
...
         }
 
+        [Fact]
+        public void WriteMultipleSpecialCharactersAtStartOfBufferLengthShouldWork()
+        {
+            foreach (string specialChar in this.escapedCharMap.Keys)
+            {
+                this.TestInit();
+                char[] charBuffer = new char[10];
+                JsonValueUtils.WriteEscapedJsonString(this.writer, string.Format("StartMiddle{0}End", specialChar), ref charBuffer);
```
AlanWong-MS: It looks like this test case has only one special character. Did you intend for it to have multiple special characters at the start of the string?

src/Microsoft.OData.Core/Buffers/BufferUtils.cs
```diff
@@ -0,0 +1,37 @@
...
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Buffers
+{
+    using System;
+
+    /// <summary>
+    /// Helpers to deal with buffers
+    /// </summary>
+
```
AlanWong-MS: nit: our coding style does not contain empty spaces between summary comment and class declaration.

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
ysanghi: Is it ok to leave this for now? It's not hurting and makes it easy to distinguish between static and instance methods easily.

AlanWong-MS: I think that's fine. I don't think we strictly have any guidelines regarding keeping it or leaving it out.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsTests.cs
ysanghi: Resolved

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
ysanghi: Resolved

Issue #1187, opened 2018-06-11
Batch is broken by load balancer (proxy).


VirusQuartirus:
*When batch request comes through load balancer or proxy ODL fails to read batch request because service base uri differs from batch part uri.*

### Assemblies affected

*OData .Net lib 7.4.*

### Reproduce steps

*Use Fiddler and locally hosted service. Use http://localhost.fiddler as service host name.*

### Expected result

*Batch should work as expected.*

### Actual result

*Service crashes at ODataBatchUtils.ValidateReferenceUri() with following message:
Microsoft.OData.ODataException: 'Absolute URI http://localhost.fiddler/XXX/ is not start with the base URI [http://localhost/XXX/] specified by the operation message.'*

### Additional detail

*Typically proxies and load balancers are adding X-Forwarded-XXX headers to request. So the ODatalib should care of them. Alternatively Client should provide possibility to use BatchPayloadUriOption.RelativeUri in order to avoid such an issue*

biaol-odata:
Closed as merged.

PR #1195, opened 2018-06-21
Skip validation of reference uri that is different from base uri.

biaol-odata:
### Issues

*This pull request fixes issue #1187.*

### Description

*Skip reference url validation when it is in absolute url form and it is different from the baseUri.*

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataBatchUtils.cs
+++ b/src/Microsoft.OData.Core/ODataBatchUtils.cs
@@ -170,7 +170,7 @@ namespace Microsoft.OData
             {
                 if (baseUri == null)
                 {
-                    // The absolution Uri can contain $ character followed by name of system resource
+                    // The absolute Uri can contain $ character followed by name of system resource
                     // such as $all, $metadata, etc, along with an unspecified baseUri. In such cases there are
                     // not reference Uri to validate.
                     return;
--- a/src/Microsoft.OData.Core/ODataBatchUtils.cs
+++ b/src/Microsoft.OData.Core/ODataBatchUtils.cs
@@ -179,8 +179,8 @@ namespace Microsoft.OData
                 string baseUriString = UriUtils.UriToString(baseUri);
                 if (!uri.AbsoluteUri.StartsWith(baseUriString, StringComparison.Ordinal))
                 {
-                    throw new ODataException(Strings.ODataBatchReader_AbsoluteURINotMatchingBaseUri(
-                        uri.AbsoluteUri, baseUriString));
+                    // Skip validation if uri doesn't start with the baseUri.
+                    return;
                 }
 
                 relativePath = uri.AbsoluteUri.Substring(baseUriString.Length);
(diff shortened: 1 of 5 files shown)


mikepizzo:
>                     // The absolution Uri can contain $ character followed by name of system resource

nit: should be "absolute"

---
Refers to: src/Microsoft.OData.Core/ODataBatchUtils.cs:173 in fe86d17. [](commit_id = fe86d172e2b62858400d6c09004e3f1c62b10e74, deletion_comment = False)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonLight/SingletonBatchRoundtripJsonLightTests.cs
```diff
@@ -729,6 +797,53 @@ private enum BatchFormat
     ]
 }";
 
+        private const string ExpectedReferenceUriRequestPayload_UseDifferentBaseUri = @"
+{
+    ""requests"": [{
+            ""id"": ""1"",
+            ""atomicityGroup"": ""11d431dd-cfee-48c8-95fb-da8491644fa6"",
+            ""method"": ""PUT"",
+            ""url"": ""http://odata.org/test//MySingleton"",
```
mikepizzo: >// [](start = 44, length = 2)

Why do all of these tests use double-forward-slashes? Looks like a typo that got propagated throughout all of the tests...

mikepizzo: Looks like the source of the issue is in methods like CreateBatchRequestWithChangesetFirstAndQueryLast that create the request URI by appending a relative url starting with "/" to the servicedocumenturi which terminates in the "/". We should fix both the methods that create the requests as well as the expected output.

---
In reply to: [197298533](https://github.com/OData/odata.net/pull/1195#discussion_r197298533) [](ancestors = 197298533)

biaol-odata: Fixed. Removed the trialing '/' in sericeDocuemtnUri and updated the expected values.

---
In reply to: [197300176](https://github.com/OData/odata.net/pull/1195#discussion_r197300176) [](ancestors = 197300176,197298533)

src/Microsoft.OData.Core/ODataBatchUtils.cs
```diff
@@ -179,8 +179,8 @@ internal static void ValidateReferenceUri(Uri uri, IEnumerable<string> dependsOn
                 string baseUriString = UriUtils.UriToString(baseUri);
                 if (!uri.AbsoluteUri.StartsWith(baseUriString, StringComparison.Ordinal))
                 {
-                    throw new ODataException(Strings.ODataBatchReader_AbsoluteURINotMatchingBaseUri(
-                        uri.AbsoluteUri, baseUriString));
+                    // Skip validation if uri doesn't start with the baseUri.
```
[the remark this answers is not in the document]
biaol-odata: If the request specifies invalid ref to other operations, the request will result in error. Otherwise, it will be alright since validation would pass anyway.

---
In reply to: [197583353](https://github.com/OData/odata.net/pull/1195#discussion_r197583353) [](ancestors = 197583353)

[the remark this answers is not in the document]
AlanWong-MS: Okay thanks

PR #1204, opened 2018-07-01
Large object streaming

mikepizzo:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/ODataBinaryStreamReader.cs
```diff
@@ -0,0 +1,199 @@
...
+
+            return bytesCopied;
+        }
+
+        /// <summary>
+        /// Flush the stream; not supported for a read stream.
+        /// </summary>
+        public override void Flush()
+        {
+            throw new NotSupportedException();
```
xuzhg: >throw new NotSupportedException(); [](start = 12, length = 34)

This class is sealed, but we leave some functions NotSupported? It's strange. Maybe call the base function?

src/Microsoft.OData.Core/ODataBinaryStreamWriter.cs
```diff
@@ -0,0 +1,401 @@
...
+    using System.Diagnostics;
+    using System.IO;
+    using System.Linq;
+    using System.Text;
+    using System.Threading;
+#if PORTABLELIB
+    using System.Threading.Tasks;
+#endif
+
+    // todo (mikep): move these to their own files...
```
xuzhg: >// todo (mikep): move these to their own files... [](start = 4, length = 49)

same as for other classes

xuzhg:
> // <copyright file="ODataBatchWriter.cs" company="Microsoft">

remember to rename the copyright

---
Refers to: src/Microsoft.OData.Core/ODataBatchWriter.cs:2 in 0ecb458. [](commit_id = 0ecb4582113de2daf7a326543a24d3ed61086dfc, deletion_comment = False)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -596,6 +596,7 @@ internal static bool IsAssignableFrom(this IEdmType baseType, IEdmType subtype)
                 case EdmTypeKind.Primitive:
                     return ((IEdmPrimitiveType)baseType).IsAssignableFrom((IEdmPrimitiveType)subtype);
 
+                case EdmTypeKind.Untyped:   // fall through
```
xuzhg: >case EdmTypeKind.Untyped:   // fall through [](start = 16, length = 43)

What's the scenario for Un-type inheritance?

src/Microsoft.OData.Client/DictionaryExtensions.cs
```diff
@@ -88,9 +88,13 @@ internal static class DictionaryExtensions
         }
 #else
         /// <summary>
-        /// Convenience function that wraps ConcurrentDictionary.TryAdd() to allow same signature as IDictionary
+        /// Convenience function that wraps ConcurrentDictionary.TryAdd() to allow same signature as IDictionary.
         /// </summary>
-        public static void Add<TKey, TValue>(this ConcurrentDictionary<TKey, TValue> self, TKey key, TValue value)
+        /// <typeparam name="TKey">The type of the key.</typeparam>
+        /// <typeparam name="TValue">The type of the value.</typeparam>        /// <param name="self">The ConcurrentDictionary to which to add the item.</param>
```
xuzhg: > /// <param name="self">The ConcurrentDictionary to which to add the item.</param> [](start = 78, length = 82)

move to new line

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -355,6 +355,9 @@ internal override IEnumerable<ODataFunction> GetFunctions()
             return ODataUtilsInternal.ConcatEnumerables(this.ResourceMetadataContext.Resource.NonComputedFunctions, this.MissingOperationGenerator.GetComputedFunctions());
         }
 
+        // mikep todo: why does MarkNestedResourceInfoProcessed and GetNextUprocessedNavigationLink exist here with 
+        // apparently same implementation as in base class?
```
xuzhg: From the implementation, 

if (structuredType.IsODataEntityTypeKind())
{
    resourceState.MetadataBuilder = new ODataConventionalEntityMetadataBuilder(resourceMetadataContext, this, uriBuilder);
}
else
{
     resourceState.MetadataBuilder = new ODataConventionalResourceMetadataBuilder(resourceMetadataContext, this, uriBuilder);
}

The implementation in ODataConventionalResourceMetadataBuilder should be removed.

src/Microsoft.OData.Core/Evaluation/ODataConventionalResourceMetadataBuilder.cs
```diff
@@ -389,6 +396,44 @@ internal override ODataJsonLightReaderNestedResourceInfo GetNextUnprocessedNavig
             return null;
         }
 
+        /// <summary>
+        /// Marks the given stream property as processed.
+        /// </summary>
+        /// <param name="streamPropertyName">The nested resource info we've already processed.</param>
```
xuzhg: >The nested resource info we've already processed [](start = 45, length = 48)

rename the sentence.

src/Microsoft.OData.Core/Json/IJsonReader.cs
```diff
@@ -4,8 +4,36 @@
...
+
 namespace Microsoft.OData.Json
 {
+    using System.IO;
+    using System.Text;
+
+    /// <summary>
+    /// Interface for a class that can read arbitrary JSON and stream binary properties.
+    /// </summary>
+    public interface IJsonStreamReader : IJsonReader
```
xuzhg: >IJsonStreamReader [](start = 21, length = 17)

Maybe move the class into its own CS file.

src/Microsoft.OData.Core/Json/IJsonWriter.cs
```diff
@@ -7,8 +7,42 @@
...
 {
     using System;
+    using System.IO;
     using Microsoft.OData.Edm;
 
+    /// <summary>
+    /// Interface for writing JSON including streaming binary values.
+    /// </summary>
+    [CLSCompliant(false)]
+    public interface IJsonStreamWriter : IJsonWriter
```
xuzhg: >IJsonStreamWriter [](start = 21, length = 17)

move this class into its own CS file.

src/Microsoft.OData.Core/Json/JsonReader.cs
```diff
@@ -186,6 +200,29 @@ public virtual object Value
         {
             get
             {
+                if (this.readingStream)
+                {
+                    // mikep: make this a real exception
```
xuzhg: >// mikep: make this a real exception [](start = 20, length = 36)

+1

src/Microsoft.OData.Core/Json/JsonReader.cs
```diff
@@ -212,13 +249,47 @@ public virtual bool IsIeee754Compatible
...
         /// <summary>
         /// Reads the next node from the input.
         /// </summary>
         /// <returns>true if a new node was found, or false if end of input was reached.</returns>
         [SuppressMessage("Microsoft.Maintainability", "CA1502:AvoidExcessiveComplexity", Justification = "Not really feasible to extract code to methods without introducing unnecessary complexity.")]
         public virtual bool Read()
         {
+            if (this.readingStream)
+            {
+                // todo (mikep): create a proper error for this
```
xuzhg: >// todo (mikep): create a proper error for this [](start = 16, length = 47)

+1

src/Microsoft.OData.Core/Json/JsonReader.cs
```diff
@@ -363,6 +434,56 @@ public virtual bool Read()
...
 
+        /// <summary>
+        /// Creates a stream for reading a base64 binary value.
+        /// </summary>
+        /// <returns>A stream for reading a base64 URL encoded binary value.</returns>
+        public Stream CreateReadStream()
+        {
+            if (!this.canStream)
+            {
+                throw new Exception("CreateTextReader cannot be called in the current state");
```
xuzhg: >"CreateTextReader cannot be called in the current state"); [](start = 36, length = 58)

Exception

src/Microsoft.OData.Core/Json/JsonReader.cs
```diff
@@ -363,6 +434,56 @@ public virtual bool Read()
...
+            this.canStream = false;
+            if ((this.streamOpeningQuoteCharacter = characterBuffer[this.tokenStartIndex]) == 'n')
+            {
+                this.ParseNullPrimitiveValue();
+                this.scopes.Peek().ValueCount++;
+                this.Read();
+                return new ODataBinaryStreamReader((a, b, c) => { return 0; });
+            }
+
+            this.tokenStartIndex++;
```
xuzhg: >this.tokenStartIndex++; [](start = 12, length = 23)

what about if "startIndex" point to the last valid character in the buffer?

src/Microsoft.OData.Core/Json/JsonReader.cs
```diff
@@ -363,6 +434,56 @@ public virtual bool Read()
...
+
+        /// <summary>
+        /// Creates a TextReader for reading text values.
+        /// </summary>
+        /// <returns>A TextReader for reading a text value.</returns>
+        public TextReader CreateTextReader()
+        {
+            if (!this.canStream)
+            {
+                throw new Exception("CreateTextReader cannot be called in the current state");
```
xuzhg: >"CreateTextReader cannot be called in the current state") [](start = 36, length = 57)

Exception +1

src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
```diff
@@ -176,6 +176,21 @@ public ODataMessageQuotas MessageQuotas
...
+        /// <Remarks>
+        /// Function takes:
+        /// * Primitive type of the value being read, or null if unknown
+        /// * Whether the value being read is a collection
+        /// * The name of the property being read (null for values within a collection)
+        /// * The property being read (null for dynamic property or value within a collection)
+        /// Function returns:
+        /// * True, to have the value streamed, otherwise false
+        /// </Remarks>
+        public Func<IEdmPrimitiveType, bool, string, IEdmProperty, bool> ReadAsStream { get; set; }
```
xuzhg: >ReadAsStream [](start = 73, length = 12)

Maybe name as "ReadAsStreamFunc"

src/Microsoft.OData.Core/UriParser/SemanticAst/ExpandedNavigationSelectItem.cs
```diff
@@ -108,6 +108,7 @@ public ExpandedNavigationSelectItem(ODataExpandPath pathToNavigationProperty, IE
         /// <param name="searchOption">An levels clause for this expand (can be null)</param>
         /// <param name="levelsOption">An levels clause for this expand (can be null)</param>
         /// <param name="computeOption">A compute clause for this expand (can be null)</param>
+        /// <param name="applyOption">An apply clause for this expand (can be null)</param>
```
xuzhg: >/// <param name="applyOption">An apply clause for this expand (can be null)</param> [](start = 8, length = 83)

Seems a lot of changes unrelated to the Large object streaming. Let me create a separate PR to fix them ahead.

src/Microsoft.OData.Core/Value/ODataStreamReferenceValue.cs
```diff
@@ -8,21 +8,108 @@ namespace Microsoft.OData
 {
     #region Namespaces
     using System;
+    using System.IO;
     using Microsoft.OData.Evaluation;
+    using Edm;
 
     #endregion
 
+    // todo (mikep): move these to their own files
```
xuzhg: >// todo (mikep): move these to their own files [](start = 4, length = 46)

+1

src/Microsoft.OData.Core/Value/ODataStreamReferenceValue.cs
```diff
@@ -8,21 +8,108 @@ namespace Microsoft.OData
...
+    public class ODataStreamValue : ODataValue
+    {
         /// <summary>The name of the named stream this value belongs to; null for the default media resource.</summary>
         private string edmPropertyName;
 
+        /// <summary>Primitive Type, if known.  Either String or None.</summary>
+        private EdmPrimitiveTypeKind typeKind;
+
+        /// <summary>
+        /// Creates an ODataStringValue with a specified typeKind
```
xuzhg: >ODataStringValue [](start = 23, length = 16)

ODataStreamValue

src/Microsoft.OData.Core/Value/ODataStreamReferenceValue.cs
```diff
@@ -8,21 +8,108 @@ namespace Microsoft.OData
...
+        /// <summary>
+        /// The Stream wrapped by the ODataStreamValue
+        /// </summary>
+        public Stream Stream { get; private set; }
+    }
+
+    /// <summary>
+    /// A class to represent a stream value
+    /// </summary>
+    public class ODataStreamValue : ODataValue
```
xuzhg: >ODataStreamValue [](start = 17, length = 16)

mark it as abstract class?

src/Microsoft.OData.Core/Value/ODataStreamReferenceValue.cs
```diff
@@ -8,21 +8,108 @@ namespace Microsoft.OData
...
+    {
         /// <summary>The name of the named stream this value belongs to; null for the default media resource.</summary>
         private string edmPropertyName;
```
xuzhg: >edmPropertyName [](start = 23, length = 15)

This can be auto-property

src/Microsoft.OData.Core/Value/ODataStreamReferenceValue.cs
```diff
@@ -8,21 +8,108 @@ namespace Microsoft.OData
...
+        /// Constructor
+        /// </summary>
+        /// <param name="stream">Input stream</param>
+        public ODataBinaryStreamValue(Stream stream)
+        {
+            this.Stream = stream;
+        }
 
+        /// <summary>
+        /// The Stream wrapped by the ODataStreamValue
```
xuzhg: >ODataStreamValue [](start = 38, length = 16)

ODataBinaryStreamValue

src/Microsoft.OData.Core/Value/ODataStreamReferenceValue.cs
```diff
@@ -8,21 +8,108 @@ namespace Microsoft.OData
...
+        {
+            this.typeKind = typeKind;
+        }
+
+        /// <summary>
+        /// PrimitiveTypeKind of the stream.
+        /// </summary>
+        /// <returns>
+        /// EdmPrimitiveTypeKind.String, if the contents is a plain string;
+        /// EdmPrimitiveTypeKind.None if it could be a binary string
```
xuzhg: >None [](start = 33, length = 4)

Should be "EdmPrimitiveTypeKind.Binary"

src/Microsoft.OData.Core/Value/ODataStreamReferenceValue.cs
```diff
@@ -35,6 +122,14 @@ public sealed class ODataStreamReferenceValue : ODataValue
         /// <summary>Read link for media resource.</summary>
         private Uri computedReadLink;
 
+        /// <summary>
+        /// Creates an instance of a StreamReferenceValue for writing a
```
xuzhg: >StreamReferenceValue [](start = 37, length = 20)

<see ref="ODataStreamReferenceValue" />

src/Microsoft.OData.Core/Uri/ODataUri.cs
```diff
@@ -52,7 +52,8 @@ public ODataUri()
         /// <param name="search">Any $search option for this uri. Can be null</param>
         /// <param name="apply">Any $apply option for this uri. Can be null</param>
         /// <param name="skip">Any $skip option for this uri. Can be null.</param>
-        /// <param name="top">Any $top option for this uri. Can be null.</param>
+        /// <param name="top">Any $top option for this uri. Can be null.</param
```
xuzhg: ></param [](start = 72, length = 7)

</param>

src/Microsoft.OData.Core/Uri/ODataUri.cs
```diff
@@ -52,7 +52,8 @@ public ODataUri()
         /// <param name="search">Any $search option for this uri. Can be null</param>
         /// <param name="apply">Any $apply option for this uri. Can be null</param>
         /// <param name="skip">Any $skip option for this uri. Can be null.</param>
-        /// <param name="top">Any $top option for this uri. Can be null.</param>
+        /// <param name="top">Any $top option for this uri. Can be null.</param
+        /// <param name="index">Any index for this uri. Can be null.</param>
```
xuzhg: $index

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs
```diff
@@ -9,24 +9,80 @@ namespace Microsoft.OData.JsonLight
     #region Namespaces
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.IO;
+    using System.Text;
     using Microsoft.OData.Edm;
     #endregion Namespaces
 
+    // todo (mikep): move to own filie
```
xuzhg: > // todo (mikep): move to own filie [](start = 3, length = 35)

+1

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs
```diff
@@ -9,24 +9,80 @@ namespace Microsoft.OData.JsonLight
...
+        /// <param name="nestedProperty">The nested property for which the nested resource info will be reported.</param>
+        protected ODataJsonLightReaderNestedInfo(IEdmProperty nestedProperty)
+        {
+            this.nestedProperty = nestedProperty;
+        }
+
+        /// <summary>
+        /// The Edm property for which the nested resource info will be reported.
+        /// </summary>
+        internal IEdmProperty NestedProperty
```
xuzhg: >NestedProperty [](start = 30, length = 14)

It seems it can be auto-property

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs
```diff
@@ -9,24 +9,80 @@ namespace Microsoft.OData.JsonLight
...
+            }
+        }
+    }
+
+    internal sealed class ODataJsonLightReaderNestedStreamInfo : ODataJsonLightReaderNestedInfo
+    {
+        /// <summary>
+        /// The Stream associated with the nested info.
+        /// </summary>
+        private readonly ODataStreamValue nestedStreamValue;
```
xuzhg: >ODataStreamValue nestedStreamValue; [](start = 25, length = 35)

It can be auto-property

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs
```diff
@@ -115,24 +171,18 @@ internal IEdmStructuralProperty StructuralProperty
...
-                return this.nestedProperty;
-            }
-        }
+        internal bool StreamMembers { get; set; }
 
         /// <summary>
-        /// true if the nested resource info has a value.
-        /// </summary>
+                 /// true if the nested resource info has a value.
+                 /// </summary>
```
xuzhg: revert

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -23,6 +23,16 @@ namespace Microsoft.OData.JsonLight
...
     {
+        /// <summary>
+        /// An in-memory stream for writing stream properties to non-streaming json writer.
+        /// </summary>
+        internal MemoryStream binaryValueStream = null;
+
+        /// <summary>
+        /// An in-memory StringWriter for writing string properties to non-streaming json writer.
+        /// </summary>
+        internal StringWriter stringWriter = null;
```
xuzhg: where are they used?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -779,27 +789,41 @@ protected override void Dispose(bool disposing)
...
                 this.textWriter = null;
                 this.jsonWriter = null;
+                this.stringWriter = null;
             }
 
             base.Dispose(disposing);
         }
 
-        private static IJsonWriter CreateJsonWriter(IServiceProvider container, TextWriter textWriter, bool isIeee754Compatible)
+        private static IJsonWriter CreateJsonWriter(IServiceProvider container, Stream stream, TextWriter textWriter, bool isIeee754Compatible)
```
xuzhg: >Stream stream, [](start = 80, length = 14)

where does stream use?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -728,6 +703,67 @@ in resourceState.PropertyAndAnnotationCollector.GetODataPropertyAnnotations(nest
...
+            ODataNestedResourceInfo nestedResourceInfo = new ODataNestedResourceInfo()
+            {
+                Name = propertyName,
+                IsCollection = false,
+                IsComplex = false
+            };
+
+            return ODataJsonLightReaderNestedResourceInfo.CreateResourceReaderNestedResourceInfo(nestedResourceInfo, property, EdmCoreModel.Instance.GetPrimitiveType(EdmPrimitiveTypeKind.Stream));
+        }
+*/
```
xuzhg: These codes are un-used?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -728,6 +703,67 @@ in resourceState.PropertyAndAnnotationCollector.GetODataPropertyAnnotations(nest
...
+
+        /// <summary>
+        /// Reads a nested stream collection as a resource set.
```
xuzhg: >resource set [](start = 50, length = 12)

as a nested resource info?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -728,6 +703,67 @@ in resourceState.PropertyAndAnnotationCollector.GetODataPropertyAnnotations(nest
...
+        {
+            Debug.Assert(resourceState != null, "resourceState != null");
+            Debug.Assert((propertyName != null) || (collectionProperty != null),
+                "The collection property and property name shouldn't both be null");
+
+            ODataNestedResourceInfo nestedResourceInfo = new ODataNestedResourceInfo()
+            {
+                Name = propertyName,
+                IsCollection = true,
+                IsComplex = false
```
xuzhg: >IsComplex = false [](start = 16, length = 17)

is "IsComplex = false" always true?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -728,6 +703,67 @@ in resourceState.PropertyAndAnnotationCollector.GetODataPropertyAnnotations(nest
...
+        /// </summary>
+        /// <param name="resourceState">The state of the reader for resource to read.</param>
+        /// <param name="collectionProperty">The collection of stream property for which to read the nested resource info. null for undeclared property.</param>
+        /// <param name="propertyName">The property name.</param>
+        /// <param name="elementType">They primitive type of the collection element</param>
+        /// <returns>The nested resource info for the stream collection.</returns>
+        /// <remarks>
+        /// This method doesn't move the reader.
+        /// </remarks>
+        protected static ODataJsonLightReaderNestedResourceInfo ReadStreamCollectionNestedResourceInfo(IODataJsonLightReaderResourceState resourceState, IEdmStructuralProperty collectionProperty, string propertyName, IEdmType elementType)
```
xuzhg: >ODataJsonLightReaderNestedResourceInfo [](start = 25, length = 38)

This method should return "ODataJsonLightReaderNestedStreamInfo", Right?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -526,6 +527,20 @@ private void ResolveCollectionValueTypeName(ODataCollectionValue collectionValue
...
 
+        /// <summary>
+        /// Writes a stream property.
+        /// </summary>
+        /// <param name="streamValue">The stream value to be written</param>
+        /// <param name="isOpenPropertyType">If the property is open.</param>
+        private void WriteStreamProperty(ODataBinaryStreamValue streamValue, bool isOpenPropertyType)
+        {
+            // todo (mikep): take type from streamvalue and write if not (possibly simplified) Metadata.EdmConstants.EdmStringTypeName?
+            // this.WritePropertyTypeName();
```
xuzhg: +1

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -526,6 +527,20 @@ private void ResolveCollectionValueTypeName(ODataCollectionValue collectionValue
             }
         }
 
+        /// <summary>
+        /// Writes a stream property.
+        /// </summary>
+        /// <param name="streamValue">The stream value to be written</param>
+        /// <param name="isOpenPropertyType">If the property is open.</param>
+        private void WriteStreamProperty(ODataBinaryStreamValue streamValue, bool isOpenPropertyType)
```
xuzhg: >isOpenPropertyType [](start = 82, length = 18)

isOpenPropertyType is useless?

test/FunctionalTests/Tests/DataOData/Tests/OData.Reader.Tests/JsonLight/FeedReaderJsonLightTests.cs
```diff
@@ -192,7 +192,7 @@ public void FeedReaderTest()
                 });
         }
 
-        [TestMethod, TestCategory("Reader.Json"), Variation(Description = "Verifies correct error behavior for invalid feeds in Json Light.")]
+//        [TestMethod, TestCategory("Reader.Json"), Variation(Description = "Verifies correct error behavior for invalid feeds in Json Light.")]
```
xuzhg: >   [TestMethod, TestCategor [](start = 7, length = 27)

revert.

test/FunctionalTests/Tests/DataOData/Tests/OData.Reader.Tests/JsonLight/FeedReaderJsonLightTests.cs
```diff
@@ -241,18 +241,18 @@ public void FeedReaderErrorTest()
...
+                //    DebugDescription = "Feed with primitive item in feed property value.",
+                //    PayloadElement = PayloadBuilder.EntitySet()
+                //        .JsonRepresentation("{" +
+                //            "\"" + JsonLightConstants.ODataPropertyAnnotationSeparator + JsonLightConstants.ODataContextAnnotationName + "\":\"http://odata.org/test/$metadata#TestModel.DefaultContainer.Cities\"," +
+                //            "\"" + JsonLightConstants.ODataValuePropertyName + "\":[ 1 ]" +
+                //            "}")
+                //        .ExpectedEntityType(cityType, citiesEntitySet),
+                //    PayloadEdmModel = model,
+                //    ExpectedException = ODataExpectedExceptions.ODataException("ODataJsonLightResourceDeserializer_InvalidNodeTypeForItemsInResourceSet", "PrimitiveValue")
+                //},
```
xuzhg: It's invalid test or unfix test case?

test/FunctionalTests/Tests/DataOData/Tests/OData.Reader.Tests/JsonLight/StreamReferenceValueReaderJsonLightTests.cs
```diff
@@ -52,7 +52,7 @@ private sealed class StreamPropertyTestCase
             public IEdmTypeReference OwningEntityType { get; set; }
         }
 
-        [TestMethod, TestCategory("Reader.Json"), Variation(Description = "Verifies correct reading of stream properties (stream reference values) with fully specified metadata.")]
+//        [TestMethod, TestCategory("Reader.Json"), Variation(Description = "Verifies correct reading of stream properties (stream reference values) with fully specified metadata.")]
```
xuzhg: > [TestMethod, TestCategory("Reader.Json"), Variation(Description = "Verifies correct reading of stream properties (stream reference values) with fully specified metadata.")] [](start = 9, length = 173)

Do we need to fix it?

test/FunctionalTests/Tests/DataOData/Tests/OData.Reader.Tests/JsonLight/UndelcaredPropertyReaderJsonLightTests.cs
```diff
@@ -166,7 +165,7 @@ public void UndeclaredNavigationLinkTests()
                 });
         }
 
-        [TestMethod, TestCategory("Reader.Json"), Variation(Description = "Test the reading of undeclared stream properties on entry payloads.")]
+ //       [TestMethod, TestCategory("Reader.Json"), Variation(Description = "Test the reading of undeclared stream properties on entry payloads.")]
```
xuzhg: >TestCategory("Reader.Json" [](start = 23, length = 26)

Do we need to fix it?

test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/Writer/WriterFeedTests.cs
```diff
@@ -249,7 +249,7 @@ public void FeedInlineCountTests()
                 });
         }
 
-        [TestMethod, Variation(Description = "Test feeds with invalid content.")]
+//        [TestMethod, Variation(Description = "Test feeds with invalid content.")]
```
xuzhg: > [TestMethod, [](start = 9, length = 13)

Do we need to fix it?

src/Microsoft.OData.Client/DictionaryExtensions.cs
mikepizzo: I think this is fixed?

---
In reply to: [258638342](https://github.com/OData/odata.net/pull/1204#discussion_r258638342) [](ancestors = 258638342)

src/Microsoft.OData.Core/ODataBinaryStreamReader.cs
mikepizzo: Stream is an abstract class that can support reading, writing, or both.  Since we derive from Stream we have to implement the methods, but our implementation only supports reading, and flush isn't supported for a read stream. Similarly, our stream doesn't support things like SetLength; the appropriate thing to do for such a stream is to throw a NotSupportedException (client is supposed to call, i.e., CanSeek first to determine if the stream supports positioning).

---
In reply to: [258625597](https://github.com/OData/odata.net/pull/1204#discussion_r258625597) [](ancestors = 258625597)

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
mikepizzo: Actually, I'm guessing that the code was added to ODataConventionalResourceMetadataBuilder when it started supporting nav properties on complex types.  The implementation on the derived ODataConventionalEntityMetadataBuilder is now redundant and should be removed, in favor of the (identical) implementation in ODataConventionalResourceMetadataBuilder.

---
In reply to: [258642599](https://github.com/OData/odata.net/pull/1204#discussion_r258642599) [](ancestors = 258642599)

src/Microsoft.OData.Core/Value/ODataStreamReferenceValue.cs
mikepizzo: Actually, I meant to call out that, if the stream was not known to be a string, then None would be returned, meaning that it *might* be binary.  I've clarified the comment to call out all three cases. 

---
In reply to: [258749210](https://github.com/OData/odata.net/pull/1204#discussion_r258749210) [](ancestors = 258749210)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs
mikepizzo: I don't see a change on this line; what are you suggesting be reverted?

---
In reply to: [259079080](https://github.com/OData/odata.net/pull/1204#discussion_r259079080) [](ancestors = 259079080)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
mikepizzo: Good catch; this was from an earlier design. Fixed.

---
In reply to: [259080473](https://github.com/OData/odata.net/pull/1204#discussion_r259080473) [](ancestors = 259080473)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
mikepizzo: Apparently :-)

Cleaned up; thanks.

---
In reply to: [259081621](https://github.com/OData/odata.net/pull/1204#discussion_r259081621) [](ancestors = 259081621)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
mikepizzo: No, actually it's the nestedStreamInfo for the resource set that happens to contain stream values, not for the stream values.

---
In reply to: [259083359](https://github.com/OData/odata.net/pull/1204#discussion_r259083359) [](ancestors = 259083359)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
mikepizzo: Yes; because it's a stream, not a complex.  It's not an entity either, but we don't have a property for that.

---
In reply to: [259082576](https://github.com/OData/odata.net/pull/1204#discussion_r259082576) [](ancestors = 259082576)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
mikepizzo: yes, as nested resource info representing a resource set. Clarified.

---
In reply to: [259081790](https://github.com/OData/odata.net/pull/1204#discussion_r259081790) [](ancestors = 259081790)

src/Microsoft.OData.Client/DictionaryExtensions.cs
xuzhg: /// <param name="self">The ConcurrentDictionary ....
should be into new line. However, don't worry about it, it was fixed in the early PR. Please rebase to the latest codes.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
mikepizzo: Good question. I think untyped just wasn't handled, and this was getting called (maybe to check the type within a collection was valid for the collection)?

I'll verify.



---
In reply to: [258638045](https://github.com/OData/odata.net/pull/1204#discussion_r258638045) [](ancestors = 258638045)

src/Microsoft.OData.Core/Json/JsonReader.cs
mikepizzo: On line 961, before doing anything else, ReadChars checks to see if tokenStartIndex < storedCharacterCount. If not, it calls ReadInput().

---
In reply to: [258647724](https://github.com/OData/odata.net/pull/1204#discussion_r258647724) [](ancestors = 258647724)

src/Microsoft.OData.Core/UriParser/SemanticAst/ExpandedNavigationSelectItem.cs
mikepizzo: I merged the latest changes from master.

---
In reply to: [258748275](https://github.com/OData/odata.net/pull/1204#discussion_r258748275) [](ancestors = 258748275)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
mikepizzo: Yes; this is required for reading an untyped collection to make sure that the instances within the collection can be cast to the untyped type.

---
In reply to: [262169653](https://github.com/OData/odata.net/pull/1204#discussion_r262169653) [](ancestors = 262169653,258638045)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
mikepizzo: It won't be once we try to decide whether or not to write the property name.

---
In reply to: [259084247](https://github.com/OData/odata.net/pull/1204#discussion_r259084247) [](ancestors = 259084247)

src/Microsoft.OData.Client/DictionaryExtensions.cs
mikepizzo: Ah, I see. I had already rebased, which is probably why I didn't see it.

---
In reply to: [262168586](https://github.com/OData/odata.net/pull/1204#discussion_r262168586) [](ancestors = 262168586)

src/Microsoft.OData.Core/Value/ODataStreamReferenceValue.cs
mikepizzo: Changed the hierarchy. This is not an ODataStreamInfo, which is sealed.

---
In reply to: [258748878](https://github.com/OData/odata.net/pull/1204#discussion_r258748878) [](ancestors = 258748878)

src/Microsoft.OData.Core/ODataPropertyInfo.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace Microsoft.OData
+{
+    using System.Collections.Generic;
+    using System.Diagnostics.CodeAnalysis;
+    using Microsoft.OData.Edm;
+
+    /// <summary>
+    /// Represents information about a single property of a resource.
+    /// </summary>
+    public class ODataPropertyInfo : ODataItem
```
xuzhg: >ODataPropertyInfo [](start = 17, length = 17)

Maybe it's better to make it as abstract class

src/Microsoft.OData.Core/ODataPropertyInfo.cs
```diff
@@ -0,0 +1,49 @@
...
+        /// <returns>The property name.</returns>
+        public string Name
+        {
+            get;
+            set;
+        }
+
+        /// <summary>Gets or sets the kind of primitive type of the property.</summary>
+        /// <returns>The <see cref="EdmPrimitiveTypeKind"/> of the property.</returns>
+        public virtual EdmPrimitiveTypeKind PrimitiveTypeKind
```
xuzhg: >EdmPrimitiveTypeKind [](start = 23, length = 20)

Why "EdmPrimitiveTypeKind"?

src/Microsoft.OData.Core/Value/ODataStreamReferenceValue.cs
```diff
@@ -43,7 +44,7 @@ public Uri EditLink
             {
                 return this.HasNonComputedEditLink
                     ? this.editLink
-                    : (this.computedEditLink ?? (this.metadataBuilder == null ? null : this.computedEditLink = this.metadataBuilder.GetStreamEditLink(this.edmPropertyName)));
```
xuzhg: >this. [](start = 150, length = 5)

I cannot figure the reason to change this.

src/Microsoft.OData.Core/Value/ODataStreamReferenceValue.cs
```diff
@@ -61,7 +62,8 @@ public Uri ReadLink
             {
                 return this.HasNonComputedReadLink
                     ? this.readLink
-                    : (this.computedReadLink ?? (this.metadataBuilder == null ? null : this.computedReadLink = this.metadataBuilder.GetStreamReadLink(this.edmPropertyName)));
```
xuzhg: >this. [](start = 150, length = 5)

I cannot figure the reason to change this.

src/Microsoft.OData.Core/JsonLight/ReorderingJsonReader.cs
```diff
@@ -31,6 +31,38 @@ internal ReorderingJsonReader(IJsonReader innerReader, int maxInnerErrorDepth)
...
         }
 
+        /// <summary>
+        /// Creates a stream for reading a stream value.
+        /// </summary>
+        /// <returns>A Stream used to read a stream value.</returns>
+        public override Stream CreateReadStream()
+        {
+            Stream result = new MemoryStream(this.Value == null ? new byte[0] :
+                Convert.FromBase64String((string)this.Value));
```
xuzhg: >FromBase64String((string)this.Value)); [](start = 24, length = 38)

how about if the "value" can't convert from Base64 string?

src/Microsoft.OData.Core/JsonLight/ReorderingJsonReader.cs
```diff
@@ -31,6 +31,38 @@ internal ReorderingJsonReader(IJsonReader innerReader, int maxInnerErrorDepth)
             Debug.Assert(innerReader != null, "innerReader != null");
         }
 
+        /// <summary>
+        /// Creates a stream for reading a stream value.
+        /// </summary>
+        /// <returns>A Stream used to read a stream value.</returns>
+        public override Stream CreateReadStream()
+        {
+            Stream result = new MemoryStream(this.Value == null ? new byte[0] :
```
xuzhg: >MemoryStream [](start = 32, length = 12)

1. Do we need to cache the MemoryStream, always create a new one? Same for the TextReader?

2. How to dispose it? Or How to make sure customer dispose it or not?

test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/Writer/WriterFeedTests.cs
xuzhg: So, it's ready to revert?

---
In reply to: [259090571](https://github.com/OData/odata.net/pull/1204#discussion_r259090571) [](ancestors = 259090571)

test/FunctionalTests/Tests/DataOData/Tests/OData.Reader.Tests/JsonLight/UndelcaredPropertyReaderJsonLightTests.cs
xuzhg: > [TestMethod, TestCa [](start = 9, length = 20)

remove the change

src/Microsoft.OData.Core/ODataPropertyInfo.cs
mikepizzo: ODataPropertyInfo is used for writing individual primitive values (that are not streams)

---
In reply to: [266659110](https://github.com/OData/odata.net/pull/1204#discussion_r266659110) [](ancestors = 266659110)

src/Microsoft.OData.Core/ODataPropertyInfo.cs
mikepizzo: Because PropertyInfo can be used to start writing any primitive property.

---
In reply to: [266659939](https://github.com/OData/odata.net/pull/1204#discussion_r266659939) [](ancestors = 266659939)

src/Microsoft.OData.Core/Value/ODataStreamReferenceValue.cs
mikepizzo: Leftover from a partially reverted change. Fixed; thanks.

---
In reply to: [266668034](https://github.com/OData/odata.net/pull/1204#discussion_r266668034) [](ancestors = 266668034)

src/Microsoft.OData.Core/Value/ODataStreamReferenceValue.cs
mikepizzo: Fixed.

---
In reply to: [266668059](https://github.com/OData/odata.net/pull/1204#discussion_r266668059) [](ancestors = 266668059)

src/Microsoft.OData.Core/JsonLight/ReorderingJsonReader.cs
mikepizzo: We need a new MemoryStream because it will be a new value.
Streams returned from CreateReadStream() get wrapped by an ODataNotificationStream in ODataReaderCore that ensures it gets disposed.

---
In reply to: [266669354](https://github.com/OData/odata.net/pull/1204#discussion_r266669354) [](ancestors = 266669354)

test/FunctionalTests/Tests/DataOData/Tests/OData.Reader.Tests/JsonLight/FeedReaderJsonLightTests.cs
mikepizzo: The test is making sure that there is no in-line stream value, which is the whole point of the PR to enable.  So, it's invalid.  I left it here to consider a unit test for validating that it *is* valid, but there are already plenty of tests that confirm that validity, and it's really hard for me to modify these tests because I can't run them locally (I get a System.IO.FileLoadException trying to load the Microsoft.OData.Core assembly because the manifest doesn't match the assembly reference). 

---
In reply to: [259089922](https://github.com/OData/odata.net/pull/1204#discussion_r259089922) [](ancestors = 259089922)

src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
mikepizzo: Fixed.

---
In reply to: [258651417](https://github.com/OData/odata.net/pull/1204#discussion_r258651417) [](ancestors = 258651417)

test/FunctionalTests/Tests/DataOData/Tests/OData.Reader.Tests/JsonLight/FeedReaderJsonLightTests.cs
mikepizzo: I will open an issue to fix the tests.

---
In reply to: [266957966](https://github.com/OData/odata.net/pull/1204#discussion_r266957966) [](ancestors = 266957966,259089922)

test/FunctionalTests/Tests/DataOData/Tests/OData.Reader.Tests/JsonLight/FeedReaderJsonLightTests.cs
mikepizzo: Opened [https://github.com/OData/odata.net/issues/1427](https://github.com/OData/odata.net/issues/1427) to track re-enabling disabled tests.

---
In reply to: [259089699](https://github.com/OData/odata.net/pull/1204#discussion_r259089699) [](ancestors = 259089699)

test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/Writer/WriterFeedTests.cs
mikepizzo: Opened [https://github.com/OData/odata.net/issues/1427](https://github.com/OData/odata.net/issues/1427) to track re-enabling disabled tests.

---
In reply to: [266672982](https://github.com/OData/odata.net/pull/1204#discussion_r266672982) [](ancestors = 266672982,259090571)

src/Microsoft.OData.Core/JsonLight/ReorderingJsonReader.cs
mikepizzo: Added a binary format exception for attempting to read a non-Base64 stream as a stream.

---
In reply to: [266668885](https://github.com/OData/odata.net/pull/1204#discussion_r266668885) [](ancestors = 266668885)

src/Microsoft.OData.Core/Value/ODataBinaryStreamValue.cs
```diff
@@ -0,0 +1,32 @@
...
+    /// </summary>
+    public sealed class ODataBinaryStreamValue : ODataValue
+    {
+        /// <summary>
+        /// Constructor
+        /// </summary>
+        /// <param name="stream">Input stream</param>
+        public ODataBinaryStreamValue(Stream stream)
+        {
+            this.Stream = stream;
```
xuzhg: >this.Stream = stream; [](start = 12, length = 21)

shall we check the input stream nullable?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Streaming/ODataJsonLightStreamReadingTests.cs
```diff
@@ -0,0 +1,1312 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using FluentAssertions;
+using Microsoft.OData.Edm;
+using Microsoft.OData.JsonLight;
+using Microsoft.Test.OData.DependencyInjection;
+using Xunit;
+using System.Threading.Tasks;
```
xuzhg: It seems you should order by using.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Streaming/ODataJsonLightStreamReadingTests.cs
```diff
@@ -0,0 +1,1312 @@
...
+                using (Stream stream = reader.CreateReadStream())
+                {
+                    result = new BinaryReader(stream).ReadString();
+                }
+            }
+
+            return result;
+        }
+
+        private void ReadJson(ODataReader reader)
```
xuzhg: >private void ReadJson(ODataReader reader) [](start = 8, length = 41)

It seems no place this method is using?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Streaming/ODataJsonLightStreamReadingTests.cs
mikepizzo: Fixed here and in ODataJsonLightStreamWritingTests

---
In reply to: [268351497](https://github.com/OData/odata.net/pull/1204#discussion_r268351497) [](ancestors = 268351497)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Streaming/ODataJsonLightStreamReadingTests.cs
```diff
@@ -0,0 +1,1312 @@
...
+                        case ODataReaderState.ResourceStart:
+                            resource = reader.Item as ODataResource;
+                            break;
+                    }
+                }
+
+                resource.Should().NotBeNull();
+                resource.Properties.Count().Should().Be(expectedPropertyCount);
+                resource.Properties.FirstOrDefault(p => p.Name == "id").Should().NotBeNull();
+                var comments = resource.Properties.FirstOrDefault(p => p.Name == "comments").ODataValue as ODataCollectionValue; ;
```
xuzhg: > ; [](start = 128, length = 2)

clean the duplicated ';'

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Streaming/ODataJsonLightStreamReadingTests.cs
```diff
@@ -0,0 +1,1312 @@
...
+                    {
+                        switch (reader.State)
+                        {
+                            case ODataReaderState.ResourceStart:
+                                break;
+
+                            case ODataReaderState.Stream:
+                                // do not create nor read the stream
+                                break;
+                        }
```
xuzhg: can we remove all "switch" in this test case?

src/Microsoft.OData.Core/ODataStreamItem.cs
```diff
@@ -0,0 +1,81 @@
...
+        private EdmPrimitiveTypeKind typeKind;
+
+        /// <summary>
+        /// Creates an ODataStreamValue with a specified typeKind.
+        /// </summary>
+        /// <param name="typeKind">The type of stream value.</param>
+        public ODataStreamItem(EdmPrimitiveTypeKind typeKind)
+        {
+            this.PrimitiveTypeKind = typeKind;
+        }
```
xuzhg: Should we call the second Constructor? 

Or, this one is useless?

src/Microsoft.OData.Core/ODataStreamItem.cs
```diff
@@ -0,0 +1,81 @@
...
+            {
+                if (typeKind != EdmPrimitiveTypeKind.String &&
+                    typeKind != EdmPrimitiveTypeKind.Binary &&
+                    typeKind != EdmPrimitiveTypeKind.None)
+                {
+                    throw new ODataException(Strings.StreamItemInvalidPrimitiveKind(value));
+                }
+
+                this.typeKind = value;
+            }
```
xuzhg: why do we allow to set the property which is set in the constructor?

src/Microsoft.OData.Core/ODataPropertyInfo.cs
xuzhg: It looks wired when i create an ODataProperty and Such property maybe not "primitive property".

---
In reply to: [266948740](https://github.com/OData/odata.net/pull/1204#discussion_r266948740) [](ancestors = 266948740,266659939)

src/Microsoft.OData.Core/ODataReader.cs
```diff
@@ -30,10 +32,38 @@ public abstract class ODataReader
         /// <returns>true if more items were read; otherwise false.</returns>
         public abstract bool Read();
 
+        /// <summary>Creates a stream for reading an inline stream property. </summary>
+        /// <returns>A stream for reading the stream property.</returns>
+        public virtual Stream CreateReadStream()
```
xuzhg: >CreateReadStream [](start = 30, length = 16)

Maybe rename it as "CreateStream()"

src/Microsoft.OData.Core/ODataStreamItem.cs
mikepizzo: In most cases, there will not be a content type.  Fixed to have this one call the second with a null content type.

---
In reply to: [268357951](https://github.com/OData/odata.net/pull/1204#discussion_r268357951) [](ancestors = 268357951)

src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -372,6 +384,84 @@ public override sealed Task<bool> ReadAsync()
...
+                if (scope.StreamingState != StreamingState.None)
+                {
+                    throw new ODataException(Strings.ODataReaderCore_CreateReadStreamCalledInInvalidState);
+                }
+
+                scope.StreamingState = StreamingState.Streaming;
+                return new ODataNotificationStream(this.InterceptException(this.CreateReadStreamImplementation), this);
+            }
+            else
+            {
```
xuzhg: Can remove "else" and throw new ODataException(...) directly

xuzhg: Maybe we can do "

if (this.State != ODataReaderState.Stream)
{  throw...}

// do other things...

---
In reply to: [268359523](https://github.com/OData/odata.net/pull/1204#discussion_r268359523) [](ancestors = 268359523)

src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -372,6 +384,84 @@ public override sealed Task<bool> ReadAsync()
...
+                throw new ODataException(Strings.ODataReaderCore_CreateReadStreamCalledInInvalidState);
+            }
+        }
+
+        /// <summary>
+        /// Creates a stream for reading an inline stream property.
```
xuzhg: >Creates a stream for reading an inline stream property. [](start = 12, length = 55)

rename it to refresh the method syntax.

src/Microsoft.OData.Core/ODataReaderCore.cs
mikepizzo: sure; i like it.

---
In reply to: [268359707](https://github.com/OData/odata.net/pull/1204#discussion_r268359707) [](ancestors = 268359707,268359523)

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -903,13 +1059,13 @@ protected virtual void WritePrimitiveValue(ODataPrimitiveValue primitiveValue)
...
+        /// <param name="itemType">The structured type for the items in the resource set to be written (or null if the entity set base type should be used).</param>
         /// <param name="skipWriting">true if the content of the scope to create should not be written.</param>
         /// <param name="selectedProperties">The selected properties of this scope.</param>
         /// <param name="odataUri">The ODataUri info of this scope.</param>
         /// <param name="isUndeclared">true if the resource set is for an undeclared property</param>
         /// <returns>The newly create scope.</returns>
-        protected abstract ResourceSetScope CreateResourceSetScope(ODataResourceSet resourceSet, IEdmNavigationSource navigationSource, IEdmStructuredType resourceType, bool skipWriting, SelectedPropertiesNode selectedProperties, ODataUri odataUri, bool isUndeclared);
+        protected abstract ResourceSetScope CreateResourceSetScope(ODataResourceSet resourceSet, IEdmNavigationSource navigationSource, IEdmType itemType, bool skipWriting, SelectedPropertiesNode selectedProperties, ODataUri odataUri, bool isUndeclared);
```
xuzhg: >IEdmType itemType [](start = 136, length = 17)

To CreateResourceSetScope using "IEdmType" looks not sense.

src/Microsoft.OData.Core/StreamExtensions.cs
```diff
@@ -0,0 +1,33 @@
...
+    internal static class StreamExtensions
+    {
+#if !PORTABLELIB
+        /// Extension method to copy one stream to another
+        /// <param name="source">Stream to copy from</param>
+        /// <param name="target">Stream to copy to</param>
+        internal static void CopyTo(this Stream source, Stream target)
+        {
+            int BUFFER_SIZE = 81920;
+            byte[] buffer = new byte[BUFFER_SIZE];
```
xuzhg: >new byte[BUFFER_SIZE]; [](start = 28, length = 22)

I don't know how frequently does this method call. If it's high, it maybe another performance problem?

src/Microsoft.OData.Core/ODataWriterCore.cs
mikepizzo: ResourceSet is used for any collection, including (now) collections of primitives that are written individually.  Exchange requirement that not all (primitive) elements of a collection would be written at once.

---
In reply to: [268361587](https://github.com/OData/odata.net/pull/1204#discussion_r268361587) [](ancestors = 268361587)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Streaming/ODataJsonLightStreamWritingTests.cs
```diff
@@ -0,0 +1,712 @@
...
+                writer.WriteEnd(); // property
+                writer.WriteEnd(); // resource
+            },
+            null);
+
+            writeWithExtraValue.ShouldThrow<ODataException>().WithMessage(Strings.ODataWriterCore_PropertyValueAlreadyWritten("textAsStream"));
+        }
+
+        [Fact]
+        public void CannotWriteAndStreamValue()
```
xuzhg: >CannotWriteAndStreamValue [](start = 20, length = 25)

This test case looks doing different things?

src/Microsoft.OData.Core/ODataReader.cs
mikepizzo: I like CreateReadStream because it is more explicit that the stream is used for reading the value.

---
In reply to: [268359194](https://github.com/OData/odata.net/pull/1204#discussion_r268359194) [](ancestors = 268359194)

src/Microsoft.OData.Core/StreamExtensions.cs
mikepizzo: This is in the .NET 3.5, where streams don't have a CopyTo method, but still probably better to allocate the buffer just once.

---
In reply to: [268362716](https://github.com/OData/odata.net/pull/1204#discussion_r268362716) [](ancestors = 268362716)

mikepizzo: Made the buffer static and allocated on first use.

---
In reply to: [268671381](https://github.com/OData/odata.net/pull/1204#discussion_r268671381) [](ancestors = 268671381,268362716)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Streaming/ODataJsonLightStreamWritingTests.cs
mikepizzo: Good catch. I renamed this as "CannotWriteValueTwice" and added two new tests: "CannotWriteAndStreamProperty" and "CannotStreamAndWriteProperty" (which try to write the same property twice, once as stream and once as property, in the order indicated).

---
In reply to: [268365098](https://github.com/OData/odata.net/pull/1204#discussion_r268365098) [](ancestors = 268365098)

src/Microsoft.OData.Core/ODataStreamItem.cs
mikepizzo: Excellent question.

Originally I made it settable only in the constructor, but then I added ContentType. I initially had one constructor that took PrimitiveTypeKind and a second that took just ContentType, and set the primitivetypekind based on the content type. But that was just a guess, and could be wrong, so then I allowed explicitly setting the content type.  But then I decided just to pass both primitivetypekind and contenttype to the second constructor, so it was no longer necessary to explicitly set either the content type or primitive type, so I've now made both setters private.

---
In reply to: [268358264](https://github.com/OData/odata.net/pull/1204#discussion_r268358264) [](ancestors = 268358264)

src/Microsoft.OData.Core/Value/ODataBinaryStreamValue.cs
mikepizzo: Added an check here, and asserts in the other (internal) wrapping classes.

---
In reply to: [268350996](https://github.com/OData/odata.net/pull/1204#discussion_r268350996) [](ancestors = 268350996)

Issue #1208, opened 2018-07-07
ODataLib Client fails to deserialize payload with @odata.context relative URL.


VirusQuartirus:
*When JSON Lite payload contains @odata.context annotation with relative URL instead of absolute URL OData .NET client fails to deserialize response.*

### Assemblies affected

*OData .Net lib 7.5.0.*

### Reproduce steps

*Read response with relative URL in @odata.context annotation. (Use Fiddler Autoresponder?).*

### Expected result

*Client should deserialize response as usual.*

### Actual result

*Client fails with following error:
Microsoft.OData.ODataException: 'The top level context URL '$metadata#FooBar' should be an absolute Uri.'.*

### Additional detail

*According to [OData V4 spec](http://docs.oasis-open.org/odata/odata-json-format/v4.01/cs01/odata-json-format-v4.01-cs01.html#_Toc499720585):
The context annotation returns the context URL (see [OData-Protocol]) for the payload. This URL can be absolute or **relative**.
This output comes from Java Olingo implementation and we are blocked by this issue right now.*

Tiberriver256:
SAP Netweaver Gateway OData v4 APIs also use relative Uri in their responses

PR #1211, opened 2018-07-13
Fix relative context Uri resolution

biaol-odata:
… uri.

### Issues

*This pull request fixes issue #1208.*

### Description

*Fix error for relative context Uri resolution using the request uri passed in from the reader settings.*

- [x] *Test cases added*

### Additional work necessary


test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Reader/JsonLight/RelativeUriReaderTests.cs
```diff
@@ -0,0 +1,588 @@
...
+                RequestUri = requestUri
+            };
+
+            using (var msgReader = new ODataMessageReader((IODataResponseMessage)message, readerSettings, edmModel))
+            {
+                test(msgReader);
+            }
+        }
+
+        private ODataResource CreateResource1(Uri id = null)
```
AlanWong-MS: Can we consolidate the CreateResource# into one method?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
```diff
@@ -61,27 +61,30 @@ private ODataJsonLightContextUriParser(IEdmModel model, Uri contextUriFromPayloa
...
             if (contextUriFromPayload == null)
             {
                 throw new ODataException(ODataErrorStrings.ODataJsonLightContextUriParser_NullMetadataDocumentUri);
             }
 
             // Create an absolute URI from the payload string
-            // TODO: Support relative context uri and resolving other relative uris
             Uri contextUri;
-            if (!Uri.TryCreate(contextUriFromPayload, UriKind.Absolute, out contextUri))
+            if (!Uri.TryCreate(baseUri, contextUriFromPayload, out contextUri))
```
AlanWong-MS: What will happen if contextUriFromPayload is absolute? I'm reading the [API here](https://msdn.microsoft.com/en-us/library/ms131573(v=vs.110).aspx) and it says that the second Uri is relative.

biaol-odata: I believe it might be some doc issue. The actual implementation of Uri.TryCreate(Uri, string, out Uri) first tries to convert the string to absolution uri first if possible; otherwise it tries to use the baseUri to resolve the string representing relative url to get the result. See here: [https://github.com/Microsoft/referencesource/blob/master/System/net/System/UriExt.cs#L315](https://github.com/Microsoft/referencesource/blob/master/System/net/System/UriExt.cs#L315)


---
In reply to: [202220739](https://github.com/OData/odata.net/pull/1211#discussion_r202220739) [](ancestors = 202220739)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Reader/JsonLight/RelativeUriReaderTests.cs
biaol-odata: done.

---
In reply to: [202220090](https://github.com/OData/odata.net/pull/1211#discussion_r202220090) [](ancestors = 202220090)

AlanWong-MS (dismissed):

src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
```diff
@@ -176,6 +176,11 @@ public ODataMessageQuotas MessageQuotas
         /// </summary>
         public bool ReadUntypedAsString { get; set; }
 
+        /// <summary>
+        /// Gets or sets current request Uri.
+        /// </summary>
+        public Uri RequestUri { get; set; }
```
mikepizzo: What's the relationship between the new RequestUri property and the existing BaseUri property?  When does RequestUri get set? Can we just use BaseUri?

mikepizzo:
Does this fix the issue with OData Client?  I don't see where RelativeUrl gets set in MessageWriterSettings.

Could it use BaseUri instead?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
```diff
@@ -61,29 +61,35 @@ private ODataJsonLightContextUriParser(IEdmModel model, Uri contextUriFromPayloa
         /// <param name="model">The model to use when resolving the target of the URI.</param>
         /// <param name="contextUriFromPayload">The string value of the odata.metadata annotation read from the payload.</param>
         /// <param name="payloadKind">The payload kind we expect the context URI to conform to.</param>
-        /// <param name="clientCustomTypeResolver">The function of client cuetom type resolver.</param>
+        /// <param name="clientCustomTypeResolver">The function of client custom type resolver.</param>
         /// <param name="needParseFragment">Whether the fragment after $metadata should be parsed, if set to false, only MetadataDocumentUri is parsed.</param>
         /// <param name="throwIfMetadataConflict">Whether to throw if a type specified in the ContextUri is not found in metadata.</param>
+        /// <param name="baseUri">Optional value (with default value of null) of base Uri used for resolving the context url.
+        /// It should be a non-null value when resolving a top-level relative context Uri.
```
mikepizzo: >a top-level  [](start = 57, length = 12)

Why call out "a top level"? Shouldn't it be non-null for resolving *any* relative context Url?  Note that relative nested context urls should be resolved using the parent absolute contextUrl as a base; is this supported?

biaol-odata: Should be any relative context url. Removed "a top level". 

---
In reply to: [203898403](https://github.com/OData/odata.net/pull/1211#discussion_r203898403) [](ancestors = 203898403)

src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
biaol-odata: BaseUri is for Atom format and may be deprecated in the future. See [https://github.com/OData/odata.net/blob/9d625ffdece85ea51c7e0823029f4597bc288624/src/Microsoft.OData.Core/ODataMessageReaderSettings.cs#L98](https://github.com/OData/odata.net/blob/9d625ffdece85ea51c7e0823029f4597bc288624/src/Microsoft.OData.Core/ODataMessageReaderSettings.cs#L98).

---
In reply to: [203896122](https://github.com/OData/odata.net/pull/1211#discussion_r203896122) [](ancestors = 203896122)

biaol-odata:
I believe the the ODataJsonLightContextUriParser.Parse() is invoked when client is reading the payload using reader settings. So, yes, this change does fix the issue with OData Client.
See above related to BaseUri vs. RequestUri.

---
In reply to: [406439921](https://github.com/OData/odata.net/pull/1211#issuecomment-406439921) [](ancestors = 406439921)

mikepizzo:
Sorry, I meant MessageReaderSettings.

How/where does RequestUri get set when using ODataClient to read a result with relative URLs?

---
In reply to: [406729387](https://github.com/OData/odata.net/pull/1211#issuecomment-406729387) [](ancestors = 406729387,406439921)

src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
mikepizzo: The reason it was to be deprecated was because it was only used in Atom. The reason it was only used in Atom was because we previously didn't support relative URLs in JSON.  Now that we support relative URLs in JSON, it seems like we should use this (existing) property, rather than deprecate it and add a new one for the same purpose.

---
In reply to: [204170332](https://github.com/OData/odata.net/pull/1211#discussion_r204170332) [](ancestors = 204170332,203896122)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
mikepizzo: But changing the comment doesn't change how this works.
If I have a contextUrl nested within another contextUrl "scope", that nested context URL needs to be resolved relative to the outer contextUrl.

For example:

~~~json
{
   "@context" : "[http://sampleservice/$externalMetadata#employees](http://sampleservice/$externalMetadata#employees)",
   "value" : [
     {
       "id":"00001",
       "name":"Kathy Smith",
       ...
       "company" :
          {
              "@context": "#company",
              "name":"FirstCompany"
          }
     }
   ]
}
~~~
or
~~~json
{
   "@context" : "[http://sampleservice/externalMetadata#company](http://sampleservice/externalMetadata#company)",
   "name":"FirstCompany",
   "employees@context": "#employees"
   "value" : [
     {
       "id":"00001",
       "name":"Kathy Smith",
       ...
     }
   ]
}

It's possible (likely) that our stack doesn't support this type of nested contextUrls, but I still want to make sure that we can support relative urls if we add support for nested context Urls in the future.

---
In reply to: [204168642](https://github.com/OData/odata.net/pull/1211#discussion_r204168642) [](ancestors = 204168642,203898403)

mikepizzo:
This fix appears to be specific to the contextUrl.  how do we resolve relative urls in nextLinks, edit/readLinks, mediaLInks, relationshipLinks, etc.?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
biaol-odata: Since we are not sure how exactly nested contextUrl is going to be supported, its relative url support should be part of the future task when it comes up. Right?

---
In reply to: [205245558](https://github.com/OData/odata.net/pull/1211#discussion_r205245558) [](ancestors = 205245558,204168642,203898403)

biaol-odata:
See RelativeUriReaderTests.cs L249 as an example.

---
In reply to: [407874612](https://github.com/OData/odata.net/pull/1211#issuecomment-407874612) [](ancestors = 407874612,406729387,406439921)

biaol-odata:
This issue is for relative contextUrl only, not including other relative urls.

---
In reply to: [407879722](https://github.com/OData/odata.net/pull/1211#issuecomment-407879722) [](ancestors = 407879722)

src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
biaol-odata: I tried using BaseUrl but it fails in multiple cases when the baseurl is the $metadata url, which is common, among other failing cases. 

---
In reply to: [205241623](https://github.com/OData/odata.net/pull/1211#discussion_r205241623) [](ancestors = 205241623,204170332,203896122)

nils:
I'm also affected by this issue, is there any way I can help unblocking this pull request?

Tiberriver256:
Anything we can help out on with this PR @mikepizzo?

ruza-kavanka:
Hi guys, I face the same issue, is there any progress? Thanks a lot!

Issue #1216, opened 2018-07-24
Calling an action with optional parameter not present in payload should not fail even if not nullable


VatroslavSrzic:
When there is an OData Action that accepts optional parameters that are not nullable and client sends request without that parameter in requests payload, an `ODataException `is thrown with message:
**"One or more parameters of the operation '<operation_name>' are missing from the request payload. The missing parameters are: <parameter_name>.".**

However, I do not see anything illegal in **not sending an optional parameter**; regardless if it is nullable or not.

### Assemblies affected

Microsoft.OData.Core 7.5.0.

### Additional detail

The trouble seems to be in `ODataParameterReaderCore` in method `EnterScope(ODataParameterReaderState state, string name, object value)` in place where missing parameters are checked and only bounding parameter is skipped, not the optional ones.
However, I am not sure that in this place you have information on if the parameter is optional or not (I do not see that info on `IEdmOperationParameter`.

### Workaround

At present time I have decided to put all my optional parameters to be nullable too. I do not like it, but there are no major drawbacks either.

xuzhg:
Merged

PR #1222, opened 2018-08-02
Fix issue #1216, parameter reader should not throw exception if missing optional parameter

xuzhg:
### Issues

* OData parameter reader should not throw exception if missing the optional parameter.

### Description

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataParameterReaderCore.cs
+++ b/src/Microsoft.OData.Core/ODataParameterReaderCore.cs
@@ -280,9 +280,14 @@ this.State == ODataParameterReaderState.Collection,
                     // Note that the binding parameter will be specified on the Uri rather than the payload, skip the binding parameter.
                     foreach (IEdmOperationParameter parameter in this.Operation.Parameters.Skip(this.Operation.IsBound ? 1 : 0))
                     {
-                        if (!this.parametersRead.Contains(parameter.Name) && !this.inputContext.EdmTypeResolver.GetParameterType(parameter).IsNullable)
+                        bool isOptionalPameter = parameter is IEdmOptionalParameter;
+                        if (!isOptionalPameter)
                         {
-                            missingParameters.Add(parameter.Name);
+                            // only check the non-optional parameter
+                            if (!this.parametersRead.Contains(parameter.Name) && !this.inputContext.EdmTypeResolver.GetParameterType(parameter).IsNullable)
+                            {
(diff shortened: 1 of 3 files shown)


src/Microsoft.OData.Core/ODataParameterReaderCore.cs
```diff
@@ -280,9 +280,14 @@ protected internal void EnterScope(ODataParameterReaderState state, string name,
                     // Note that the binding parameter will be specified on the Uri rather than the payload, skip the binding parameter.
                     foreach (IEdmOperationParameter parameter in this.Operation.Parameters.Skip(this.Operation.IsBound ? 1 : 0))
                     {
-                        if (!this.parametersRead.Contains(parameter.Name) && !this.inputContext.EdmTypeResolver.GetParameterType(parameter).IsNullable)
+                        bool isOptionalPameter = parameter is IEdmOptionalParameter;
+                        if (!isOptionalPameter)
                         {
-                            missingParameters.Add(parameter.Name);
+                            // only check the non-optional parameter
+                            if (!this.parametersRead.Contains(parameter.Name) && !this.inputContext.EdmTypeResolver.GetParameterType(parameter).IsNullable)
```
AlanWong-MS: Combine !isOptionalParameter in this statement?

src/Microsoft.OData.Core/ODataParameterReaderCore.cs
```diff
@@ -280,9 +280,14 @@ protected internal void EnterScope(ODataParameterReaderState state, string name,
                     // Note that the binding parameter will be specified on the Uri rather than the payload, skip the binding parameter.
                     foreach (IEdmOperationParameter parameter in this.Operation.Parameters.Skip(this.Operation.IsBound ? 1 : 0))
                     {
-                        if (!this.parametersRead.Contains(parameter.Name) && !this.inputContext.EdmTypeResolver.GetParameterType(parameter).IsNullable)
+                        bool isOptionalPameter = parameter is IEdmOptionalParameter;
```
AlanWong-MS: typo: isOptionalPameter -> isOptionalPa**ra**meter

src/Microsoft.OData.Core/ODataParameterReaderCore.cs
biaol-odata: +1

PR #1223, opened 2018-08-02
use V7 ODL content as the default site for on odata.github.io

biaol-odata:
### Issues

*This pull request updates the odata.net gh-page doc root index to serve V7 content.*

### Description

*Updates for using V7 categories & content.*

- [x] *Test cases added*

### Additional work necessary


biaol-odata:
Thanks for the CR!
Merged.

Issue #1226, opened 2018-08-09
[PERF] EDM is allocating string by recalculating the full type name every time it is being asked. 


ysanghi:
[PERF] EDM is allocating string by recalculating the full type name every time it is being asked. 
### Assemblies affected

OData .NET EDM Lib

### Reproduce steps

Run EDM and take allocation traces

### Expected result

Less string allocations

### Actual result

More string allocations

biaol-odata:
@ysanghi looks like you have a PR. Assigned accordingly.

xuzhg:
Merged and released in 7.5.1.

PR #1227, opened 2018-08-09
Cache full name for edm types

ysanghi:
### Issues

*This pull request fixes issue #1226.*

### Description

Add a new interface for holding the Full Type Name, later in GetFullName method check if the type is inheriting from the new interface. If yes, use the cached value or fall back.

- [X] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsComplexTypeDefinition.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsComplexTypeDefinition.cs
@@ -13,7 +13,7 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
     /// <summary>
     /// Provides semantics for CsdlComplexType.
     /// </summary>
-    internal class CsdlSemanticsComplexTypeDefinition : CsdlSemanticsStructuredTypeDefinition, IEdmComplexType
+    internal class CsdlSemanticsComplexTypeDefinition : CsdlSemanticsStructuredTypeDefinition, IEdmComplexType, IEdmFullNamedElement
     {
         private readonly string fullName;
         private readonly CsdlComplexType complex;
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsComplexTypeDefinition.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsComplexTypeDefinition.cs
@@ -15,6 +15,7 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
     /// </summary>
     internal class CsdlSemanticsComplexTypeDefinition : CsdlSemanticsStructuredTypeDefinition, IEdmComplexType
     {
+        private readonly string fullName;
         private readonly CsdlComplexType complex;
 
         private readonly Cache<CsdlSemanticsComplexTypeDefinition, IEdmComplexType> baseTypeCache = new Cache<CsdlSemanticsComplexTypeDefinition, IEdmComplexType>();
@@ -25,6 +26,7 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
             : base(context, complex)
         {
             this.complex = complex;
+            this.fullName = EdmUtil.GetFullNameForSchemaElement(context?.Namespace, this.complex?.Name);
         }
 
         public override IEdmStructuredType BaseType
@@ -50,6 +52,14 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
         public string Name
         {
             get { return this.complex.Name; }
+        }
+
+        /// <summary>
+        /// Gets the full name of this schema element.
+        /// </summary>
+        public string FullName
+        {
+            get { return this.fullName; }
         }
 
         protected override CsdlStructuredType MyStructured
--- a/src/Microsoft.OData.Edm/EdmUtil.cs
+++ b/src/Microsoft.OData.Edm/EdmUtil.cs
@@ -10,6 +10,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
+
 using Microsoft.OData.Edm.Csdl;
 using Microsoft.OData.Edm.Csdl.CsdlSemantics;
 using Microsoft.OData.Edm.Vocabularies;
@@ -487,6 +488,9 @@ namespace Microsoft.OData.Edm
         /// <summary>
         /// Gets full name for the schema element with the provided namespace and name
         /// </summary>
+        /// <param name="elementNamespace">Namespace of the element</param>
+        /// <param name="elementName">The element name</param>
+        /// <returns>The full name of the element</returns>
         internal static string GetFullNameForSchemaElement(string elementNamespace, string elementName)
         {
             if (elementName == null)
--- a/src/Microsoft.OData.Edm/EdmUtil.cs
+++ b/src/Microsoft.OData.Edm/EdmUtil.cs
@@ -260,12 +260,14 @@ namespace Microsoft.OData.Edm
             return sb.ToString();
         }
 
-        internal static bool TryGetNamespaceNameFromQualifiedName(string qualifiedName, out string namespaceName, out string name)
+        internal static bool TryGetNamespaceNameFromQualifiedName(string qualifiedName, out string namespaceName, out string name, out string fullName)
         {
             // Qualified name can be a operation import name which is separated by '/'
             int lastSlash = qualifiedName.LastIndexOf('/');
             if (lastSlash < 0)
             {
+                fullName = qualifiedName;
+
                 // Not a OperationImport
                 int lastDot = qualifiedName.LastIndexOf('.');
                 if (lastDot < 0)
@@ -282,6 +284,7 @@ namespace Microsoft.OData.Edm
 
             namespaceName = qualifiedName.Substring(0, lastSlash);
             name = qualifiedName.Substring(lastSlash + 1);
+            fullName = EdmUtil.GetFullNameForSchemaElement(namespaceName, name);
             return true;
         }
 
@@ -479,6 +482,24 @@ namespace Microsoft.OData.Edm
             }
 
             return val;
+        }
+
+        /// <summary>
+        /// Gets full name for the schema element with the provided namespace and name
+        /// </summary>
+        internal static string GetFullNameForSchemaElement(string elementNamespace, string elementName)
+        {
+            if (elementName == null)
+            {
+                return string.Empty;
+            }
+
+            if (elementNamespace == null)
+            {
+                return elementName;
+            }
+
+            return elementNamespace + "." + elementName;
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Edm/EdmUtil.cs
+++ b/src/Microsoft.OData.Edm/EdmUtil.cs
@@ -10,6 +10,7 @@ using System.Diagnostics;
 using System.Linq;
 using System.Text;
 using System.Text.RegularExpressions;
+
 using Microsoft.OData.Edm.Csdl;
 using Microsoft.OData.Edm.Csdl.CsdlSemantics;
 using Microsoft.OData.Edm.Vocabularies;
@@ -262,12 +263,12 @@ namespace Microsoft.OData.Edm
 
         internal static bool TryGetNamespaceNameFromQualifiedName(string qualifiedName, out string namespaceName, out string name, out string fullName)
         {
+            bool foundNamespace;
+
             // Qualified name can be a operation import name which is separated by '/'
             int lastSlash = qualifiedName.LastIndexOf('/');
             if (lastSlash < 0)
             {
-                fullName = qualifiedName;
-
                 // Not a OperationImport
                 int lastDot = qualifiedName.LastIndexOf('.');
                 if (lastDot < 0)
@@ -275,24 +276,24 @@ namespace Microsoft.OData.Edm
                     namespaceName = String.Empty;
                     name = qualifiedName;
 
-                    // For backwards compatibility reasons, if name is empty and namespace is empty set the full name to "."
-                    if (name.Length == 0)
-                    {
-                        fullName = ".";
-                    }
-
-                    return false;
+                    foundNamespace = false;
                 }
-
-                namespaceName = qualifiedName.Substring(0, lastDot);
-                name = qualifiedName.Substring(lastDot + 1);
-                return true;
+                else
+                {
+                    namespaceName = qualifiedName.Substring(0, lastDot);
+                    name = qualifiedName.Substring(lastDot + 1);
+                    foundNamespace = true;
+                }
+            }
+            else
+            {
+                foundNamespace = true;
+                namespaceName = qualifiedName.Substring(0, lastSlash);
+                name = qualifiedName.Substring(lastSlash + 1);
             }
 
-            namespaceName = qualifiedName.Substring(0, lastSlash);
-            name = qualifiedName.Substring(lastSlash + 1);
             fullName = EdmUtil.GetFullNameForSchemaElement(namespaceName, name);
-            return true;
+            return foundNamespace;
         }
 
         internal static string FullyQualifiedName(IEdmVocabularyAnnotatable element)
--- a/src/Microsoft.OData.Edm/EdmUtil.cs
+++ b/src/Microsoft.OData.Edm/EdmUtil.cs
@@ -271,8 +271,6 @@ namespace Microsoft.OData.Edm
 
         internal static bool TryGetNamespaceNameFromQualifiedName(string qualifiedName, out string namespaceName, out string name)
         {
-            bool foundNamespace;
-
             // Qualified name can be a operation import name which is separated by '/'
             int lastSlash = qualifiedName.LastIndexOf('/');
             if (lastSlash < 0)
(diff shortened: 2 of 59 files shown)


biaol-odata:
@ysanghi In case you haven't looked at the build error:
2018-08-09T00:11:26.1812790Z ##[error]src\Microsoft.OData.Edm\Csdl\Semantics\BadElements\UnresolvedOperation.cs(16,89): Error CS0246: The type or namespace name 'IEdmFullNamedElement' could not be found (are you missing a using directive or an assembly reference?)

src/Microsoft.OData.Edm/EdmUtil.cs
```diff
@@ -481,6 +492,27 @@ internal static bool TryGetAnnotation(this IEdmModel model, IEdmElement annotata
...
 
+        /// <summary>
+        /// Gets full name for the schema element with the provided namespace and name
+        /// </summary>
+        /// <param name="elementNamespace">Namespace of the element</param>
+        /// <param name="elementName">The element name</param>
+        /// <returns>The full name of the element</returns>
+        internal static string GetFullNameForSchemaElement(string elementNamespace, string elementName)
+        {
+            if (elementName == null)
```
biaol-odata: how about String.IsNullOrEmpty(elementName) to take care of case of empty string?
Same for the elementNameSpace below.

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1387,14 +1387,19 @@ public static string FullName(this IEdmSchemaElement element)
...
-            else if (element.Namespace == null)
+
+            if (element.Namespace == null)
             {
                 return element.Name;
             }
-            else
+
+            IEdmFullNamedElement fullNamedElement = element as IEdmFullNamedElement;
+            if (fullNamedElement != null)
```
biaol-odata: nit: more compact to use ternary operator ?:

tools/FxCopRules/DataWebRules.csproj
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
-<Project ToolsVersion="4.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+<Project ToolsVersion="12.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
biaol-odata: Is this change intentional?

src/Microsoft.OData.Edm/Schema/Interfaces/IEdmFullNamedElement.cs
```diff
@@ -0,0 +1,19 @@
...
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Edm
+{
+    /// <summary>
+    /// Common base interface for all named EDM elements.
+    /// </summary>
+    public interface IEdmFullNamedElement : IEdmNamedElement
```
biaol-odata: Any reason why base type of IEdmNamedElement is required?

src/Microsoft.OData.Edm/EdmUtil.cs
ysanghi: I tried that and bunch of test cases failed. I wouldn't want to introduce a breaking change even if it's seemingly the right thing to do.

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
ysanghi: It looks a bit ugly in this case

return fullNamedElement?.FullName ??  (element.Namespace + "." + element.Name);

I prefer what is existing

src/Microsoft.OData.Edm/Schema/Interfaces/IEdmFullNamedElement.cs
ysanghi: Only Named Elements should have FullName.

tools/FxCopRules/DataWebRules.csproj
ysanghi: No it is not. Will revert.

src/Microsoft.OData.Edm/EdmUtil.cs
biaol-odata: Are those test errors for test case only or would introduce breaking changes?

tools/FxCopRules/DataWebRules.csproj
AlanWong-MS: @ysanghi, out of curiosity, which version of Visual Studio are you loading the projects with? If VS2017, then there is a known issue with building this particular project and I recommend just unloading it from your solutions if it's bothering you. I have an open PR that addresses this issue but it hasn't been checked in yet.

src/Microsoft.OData.Edm/EdmUtil.cs
```diff
@@ -272,17 +275,25 @@ internal static bool TryGetNamespaceNameFromQualifiedName(string qualifiedName,
...
+            {
+                foundNamespace = true;
+                namespaceName = qualifiedName.Substring(0, lastSlash);
+                name = qualifiedName.Substring(lastSlash + 1);
             }
 
-            namespaceName = qualifiedName.Substring(0, lastSlash);
-            name = qualifiedName.Substring(lastSlash + 1);
-            return true;
+            fullName = EdmUtil.GetFullNameForSchemaElement(namespaceName, name);
```
AlanWong-MS: From looking at each of the if-else statements above, I think that the `GetFullNameForSchemaElement` function is effectively changing any '/' in the qualifiedName to '.'. In all other cases, `fullName` seems to be equal to `qualifiedName`.

Please correct me if I'm wrong, but is `qualifiedName` essentially the "full type name" that you're looking for? If so, you can cache the `qualifiedName` instead of calculating the `fullName`.

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
AlanWong-MS: I think the current code Yogesh has would be easier to read.

src/Microsoft.OData.Edm/EdmUtil.cs
AlanWong-MS: +1 to Biao's question, I'm curious on what cases failed

src/Microsoft.OData.Edm/Schema/Interfaces/IEdmFullNamedElement.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        /// <summary>
+        /// Gets the full name of this element.
+        /// </summary>
+        string FullName { get; }
```
AlanWong-MS: This might actually be called QualifiedName. Please see my other comment regarding whether we can just use the `qualifiedName` string.

src/Microsoft.OData.Edm/Schema/Interfaces/IEdmFullNamedElement.cs
AlanWong-MS: Are there cases where a named element wouldn't have a full name (qualified name?) associated as well? I'm wondering if the new FullName property would logically belong in IEdmNamedElement.

Might need input from @xuzhg / @mikepizzo on this.

src/Microsoft.OData.Edm/EdmUtil.cs
```diff
@@ -481,6 +492,27 @@ internal static bool TryGetAnnotation(this IEdmModel model, IEdmElement annotata
...
+            {
+                return string.Empty;
+            }
+
+            if (elementNamespace == null)
+            {
+                return elementName;
+            }
+
+            return elementNamespace + "." + elementName;
```
biaol-odata: At least here we need to do empty-string-test for elementNamespace, otherwise could get sth like ".MyElemName"?

ysanghi: Gist of my changes if performance. So I am not comfortable making changes that lead to any functional changes. For the large part I am replicating the exact code that existed before and instead of calculating the full name every time, I am caching it on the object and reusing it over and again.


I do not like to mix and match perf and functional changes. I can open a new issue for you to look at this.

src/Microsoft.OData.Edm/Schema/Interfaces/IEdmFullNamedElement.cs
ysanghi: It was called FullName before, however, instead of caching it previously, it was calculating it over and over again. Now I am simply caching it.

src/Microsoft.OData.Edm/EdmUtil.cs
ysanghi: I don't believe this is completely accurate. FullName is supposed to be delimited by "." instead of "/".

This is to keep backward compatibility. I am afraid I don't know why qualified name is with "/" and full Name is with "." 

I am attempting a perf fix so afraid to do functional changes.

src/Microsoft.OData.Edm/Schema/Interfaces/IEdmFullNamedElement.cs
ysanghi: Cant change public interface as it will be breaking change.

tools/FxCopRules/DataWebRules.csproj
ysanghi: Will do it next time. For now will revert this change. As long as there are no other comments will make this commit.

src/Microsoft.OData.Edm/EdmUtil.cs
ysanghi: Gist of my changes if performance. So I am not comfortable making changes that lead to any functional changes. For the large part I am replicating the exact code that existed before and instead of calculating the full name every time, I am caching it on the object and reusing it over and again.


I do not like to mix and match perf and functional changes. I can open a new issue for you to look at this.

tools/FxCopRules/DataWebRules.csproj
```diff
@@ -40,11 +69,16 @@
     </EmbeddedResource>
   </ItemGroup>
   <ItemGroup>
+    <Reference Include="FxCopSdk">
+      <HintPath>C:\Program Files (x86)\Microsoft Visual Studio\2017\Enterprise\Common7\IDE\PrivateAssemblies\FxCopSdk.dll</HintPath>
```
AlanWong-MS: I think this change may break for devs using VS2015. We should revert the file if it is not involved with the code change.

ysanghi: Reverted it.

src/Microsoft.OData.Edm/Schema/Interfaces/IEdmFullNamedElement.cs
xuzhg: I try to think why we want to add a new interface like this? 

ODL.Edm lib has a full defined structured for element. And I don't think we should have this new interface to cache the full name.





---
In reply to: [209776151](https://github.com/OData/odata.net/pull/1227#discussion_r209776151) [](ancestors = 209776151)

ysanghi: Is there a proposed alternative?

xuzhg: I am still reading the PR. Can we use the property bag in the model to cache the mapping between element and full name?

---
In reply to: [211420888](https://github.com/OData/odata.net/pull/1227#discussion_r211420888) [](ancestors = 211420888)

ysanghi: We can but then it doesn't put us in the path forward. Ideally when ODL does it's next breaking change release, the interface I added should be removed and this property should be moved to one of the existing interfaces (read INamedElement). This way it belong to where it needs to belong. No need add additional property bag and stuff. My change puts towards this path forward.

mikepizzo: As Yogesh says, ideally we would put fullname in IEdmNamedElement and just use that, but since we can't make breaking changes it makes sense to have it in a derived interface.

---
In reply to: [211419623](https://github.com/OData/odata.net/pull/1227#discussion_r211419623) [](ancestors = 211419623,209776151)

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
mikepizzo: I agree.


---
In reply to: [209406678](https://github.com/OData/odata.net/pull/1227#discussion_r209406678) [](ancestors = 209406678)

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsComplexTypeDefinition.cs
```diff
@@ -13,8 +13,9 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
     /// <summary>
     /// Provides semantics for CsdlComplexType.
     /// </summary>
-    internal class CsdlSemanticsComplexTypeDefinition : CsdlSemanticsStructuredTypeDefinition, IEdmComplexType
+    internal class CsdlSemanticsComplexTypeDefinition : CsdlSemanticsStructuredTypeDefinition, IEdmComplexType, IEdmFullNamedElement
```
mikepizzo: >IEdmFullNamedElement [](start = 112, length = 20)

Rather than add to CsdlSemanticsComplexTypeDefinition and CsdlSemanticsEntityTypeDefinition, why don't we just add to CsdlSemanticsStructuredTypeDefinition?

src/Microsoft.OData.Edm/EdmUtil.cs
```diff
@@ -260,8 +261,10 @@ internal static string ParameterizedName(IEdmOperation operation)
             return sb.ToString();
         }
 
-        internal static bool TryGetNamespaceNameFromQualifiedName(string qualifiedName, out string namespaceName, out string name)
+        internal static bool TryGetNamespaceNameFromQualifiedName(string qualifiedName, out string namespaceName, out string name, out string fullName)
```
mikepizzo: >TryGetNamespaceNameFromQualifiedName [](start = 29, length = 36)

Rather than requiring callers that don't care about full name (i.e., FindVocabularyAnnotations and CsdlSemanticsModel.AddSchema (which iterates over every element in the schema) to pass an output variable and compute a value that they never use, should we instead just create a separate overload to return the fullname for those that will use it?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsComplexTypeDefinition.cs
ysanghi: TBH, I didn't give much thought into this. Name property was defined on the derived type (CsdlSemanticsComplexTypeDefinition) and hence I added FullName there as well. If we need to clean up things then we should pick this as a separate work item to move properties that are defined on the derived class and move then to base. Agree?

src/Microsoft.OData.Edm/EdmUtil.cs
ysanghi: I feel this method is the right place to get the fullname as well since it's already returning the namespace and name and getting the full name feels like the logical thing to do here.

src/Microsoft.OData.Edm/Schema/Interfaces/IEdmFullNamedElement.cs
mikepizzo: Agree. Keeping the full name on the type instance is much simpler/more straight forward than creating a side mechanism.

---
In reply to: [211429169](https://github.com/OData/odata.net/pull/1227#discussion_r211429169) [](ancestors = 211429169)

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsComplexTypeDefinition.cs
mikepizzo: Agreed. Strange that Name is defined on the derived types, but makes sense to have Name and FullName co-located.

---
In reply to: [215790246](https://github.com/OData/odata.net/pull/1227#discussion_r215790246) [](ancestors = 215790246)

src/Microsoft.OData.Edm/EdmUtil.cs
mikepizzo: Yes; please open an issue.  Agree that this looks suspect, but also agree that this is exactly mimicking current behavior. 

---
In reply to: [209781364](https://github.com/OData/odata.net/pull/1227#discussion_r209781364) [](ancestors = 209781364)

ysanghi: Done: https://github.com/OData/odata.net/issues/1248

mikepizzo: Thanks!

---
In reply to: [216468441](https://github.com/OData/odata.net/pull/1227#discussion_r216468441) [](ancestors = 216468441)

src/Microsoft.OData.Edm/EdmUtil.cs
```diff
@@ -260,8 +261,18 @@ internal static string ParameterizedName(IEdmOperation operation)
...
+        {
+            bool foundNamespace = EdmUtil.TryGetNamespaceNameFromQualifiedName(qualifiedName, out namespaceName, out name);
+
+            fullName = EdmUtil.GetFullNameForSchemaElement(namespaceName, name);
+            return foundNamespace;
+        }
+
         internal static bool TryGetNamespaceNameFromQualifiedName(string qualifiedName, out string namespaceName, out string name)
         {
```
mikepizzo: We should be able to revert changes to this version of the overload, yes?  i.e., rather than introduce the foundNamespace, boolean, we can just return true or false in each case below.

biaol-odata:
@ysanghi please resolve the conflict before this PR can be merged. Thanks.

biaol-odata:
image

biaol-odata:
@ysanghi sorry, closed inadvertently. Reopened. Please resolve the conflict so that we can merge this.

Issue #1220, opened 2018-07-31
OData v4 client code generator - Combined length of user strings used by the program exceeds allowed limit


javism1:
OData v4 client code generator - Get error [Error code: CS8103] "Combined length of user strings used by the program exceeds allowed limit. Try to decrease use of string literals." when building project with oDataClient.cs file

### Assemblies affected

Using:
OData v4 client code generator version 7.5.0 
Microsoft.OData.Client 7.5.0
Microsoft.OData.Core 7.5.0
Microsoft.OData.Edm 7.5.0
Microsoft.Spatial 7.5.0

### Reproduce steps

1. Point the oDataClient.tt file to use the Dynamics 365 for Finance and Operations metadata file
2. Generate oDataClient.cs file by running custom tool on oDataClient.tt file
3. Build project

### Expected result

Successful build

### Actual result

Get the error message: "Combined length of user strings used by the program exceeds allowed limit"

### Additional detail

This is currently happening when using the MS Client code generator with the EDMX file from Dynamics 365 for Finance and Operations (App. version 8.0). It seems like the list of data entities returned on the EDMX file is too large to be handled by Visual Studio. 

Other users did some further investigation:
https://github.com/Microsoft/Dynamics-AX-Integration/issues/17

biaol-odata:
Looks like a string length limit issue imposed by Visual Studio. Can this issue be avoided by reducing the string length?

fraga:
+1 on this one. We're also having issues generating metadata. File is too big and looking for alternatives.

fraga:
@biaol-odata thanks for the attention. D365 provides us the metadata and currently, as far as I know, we have no means to select different entities to make the file smaller. So, as specified by @javism1 D365 v8.0 PU15 creates a huge edmx metadata (around 15mb) and it gives us trouble when generating the odata client failing with _"Combined length of user strings used by the program exceeds allowed limit. Try to decrease use of string literals."_. I do have a sample edmx file if you want and if you allow I can email it to you.

KanishManuja-MS:
@fraga can you email me the sample edmx file at kamanuja@microsoft.com?

fraga:
Hi @KanishManuja-MS, email sent. Thanks.

anhhainguyen:
i'm having the same issue for OData D365. Is there a workaround for this?

fraga:
[...]

PR #1234, opened 2018-08-20
Add an optional file path for large metadata files

KanishManuja-MS:
Add another field to the configuration class to specify the filepath.

### Issues

This pull request fixes issue #1220.

### Description
This pull request introduces another parameter to the client code generator that specifies a temporary file path where the metadata can be stored for the client proxy classes. Alternatively, if nothing is specified, it gets stored in a string literal. 

Previous unit tests passed with nothing specified in the parameter. 

- [ ] *Test cases added*

### DIFF
--- a/src/CodeGen/ODataT4CodeGenerator.tt
+++ b/src/CodeGen/ODataT4CodeGenerator.tt
@@ -17,6 +17,10 @@ public static class Configuration
 
 	// The target language of the generated client code. The value must be set to "CSharp" or "VB".
 	public const string TargetLanguage = "OutputLanguage";
+	
+	// The path for the temporary file where the metadata xml document can be stored. Use this if your metadata is too big to be stored in a string literal. Ensure that you have write permission for this path.
+	// For example - "C:\\temp\\Test.xml"	
+	public const string TempFilePath = "";
 
 	// This flag indicates whether to enable naming alias. The value must be set to true or false.
 	public const bool EnableNamingAlias = true;
--- a/src/CodeGen/ODataT4CodeGenerator.tt
+++ b/src/CodeGen/ODataT4CodeGenerator.tt
@@ -20,7 +20,7 @@ public static class Configuration
 	
 	// The path for the temporary file where the metadata xml document can be stored. Use this if your metadata is too big to be stored in a string literal. Ensure that you have write permission for this path.
 	// For example - "C:\\temp\\Test.xml"	
-	public const string TempFilePath = "";
+	public const string TempFilePath = string.Empty;
 
 	// This flag indicates whether to enable naming alias. The value must be set to true or false.
 	public const bool EnableNamingAlias = true;
--- a/src/CodeGen/ODataT4CodeGenerator.tt
+++ b/src/CodeGen/ODataT4CodeGenerator.tt
@@ -20,7 +20,7 @@ public static class Configuration
 	
 	// The path for the temporary file where the metadata xml document can be stored. Use this if your metadata is too big to be stored in a string literal. Ensure that you have write permission for this path.
 	// For example - "C:\\temp\\Test.xml"	
-	public const string TempFilePath = string.Empty;
+	public const string TempFilePath = "";
 
 	// This flag indicates whether to enable naming alias. The value must be set to true or false.
 	public const bool EnableNamingAlias = true;
--- a/src/CodeGen/ODataT4CodeGenerator.ttinclude
+++ b/src/CodeGen/ODataT4CodeGenerator.ttinclude
@@ -3440,9 +3440,10 @@ namespace <#= fullNamespace #>
 
     internal override void WriteGeneratedEdmModel(string escapedEdmxString)
     {
+
 		string path = this.context.TempFilePath; 
         
-		if(String.IsNullOrEmpty(path))
+		if(!String.IsNullOrEmpty(path))
 		{
 			using (StreamWriter writer = new StreamWriter(path, true))
 			{
--- a/src/CodeGen/ODataT4CodeGenerator.ttinclude
+++ b/src/CodeGen/ODataT4CodeGenerator.ttinclude
@@ -695,7 +695,7 @@ public class CodeGenerationContext
         get;
         set;
     }
-	
+    
     /// <summary>
     /// The path for the temporary file where the metadata xml document can be stored.
     /// </summary>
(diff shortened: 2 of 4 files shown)


src/CodeGen/ODataT4CodeGenerator.ttinclude
```diff
@@ -48,6 +48,7 @@ THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
             UseDataServiceCollection = this.UseDataServiceCollection,
             TargetLanguage = this.TargetLanguage,
             EnableNamingAlias = this.EnableNamingAlias,
+			TempFilePath = this.TempFilePath,
```
AlanWong-MS: nit for consistency: should be white space

src/CodeGen/ODataT4CodeGenerator.ttinclude
```diff
@@ -316,7 +327,8 @@ private void ApplyParametersFromConfigurationClass()
     this.UseDataServiceCollection = Configuration.UseDataServiceCollection;
     this.ValidateAndSetTargetLanguageFromString(Configuration.TargetLanguage);
     this.EnableNamingAlias = Configuration.EnableNamingAlias;
-    this.IgnoreUnexpectedElementsAndAttributes = Configuration.IgnoreUnexpectedElementsAndAttributes;
+    this.TempFilePath = Configuration.TempFilePath;
+	this.IgnoreUnexpectedElementsAndAttributes = Configuration.IgnoreUnexpectedElementsAndAttributes;
```
AlanWong-MS: whitespace nit here as well

src/CodeGen/ODataT4CodeGenerator.ttinclude
```diff
@@ -3420,6 +3446,18 @@ namespace <#= fullNamespace #>
 
     internal override void WriteGeneratedEdmModel(string escapedEdmxString)
     {
+		string path = this.context.TempFilePath; 
+        
+		if(String.IsNullOrEmpty(path))
+		{
+			using (StreamWriter writer =
+            new StreamWriter(path, true))
```
AlanWong-MS: This line should be in the one above as a single line

src/CodeGen/ODataT4CodeGenerator.tt
```diff
@@ -17,6 +17,10 @@ public static class Configuration
 
 	// The target language of the generated client code. The value must be set to "CSharp" or "VB".
 	public const string TargetLanguage = "OutputLanguage";
+	
+	// The path for the temporary file where the metadata xml document can be stored. Use this if your metadata is too big to be stored in a string literal. Ensure that you have write permission for this path.
+	// For example - "C:\\temp\\Test.xml"	
+	public const string TempFilePath = "";
```
AlanWong-MS: It sounds like this file shares the same purpose as MetadataDocumentUri with an additional purpose of letting the code gen know whether to read it as a local file. Perhaps we can have a boolean instead for that additional purpose? Please correct me if I'm misunderstanding this.

src/CodeGen/ODataT4CodeGenerator.ttinclude
```diff
@@ -3420,6 +3446,18 @@ namespace <#= fullNamespace #>
...
+		if(String.IsNullOrEmpty(path))
+		{
+			using (StreamWriter writer =
+            new StreamWriter(path, true))
+			{
+				writer.WriteLine(escapedEdmxString);
+			}
+		}
+
+		bool useTempFile = !String.IsNullOrEmpty(path) && System.IO.File.Exists(path);
```
AlanWong-MS: Please see my comment regarding the difference between TempFilePath and MetadataDocumentUri. Would it make sense have a useTempFile boolean that the user can set?

src/CodeGen/ODataT4CodeGenerator.ttinclude
KanishManuja-MS: Fixed.

src/CodeGen/ODataT4CodeGenerator.tt
KanishManuja-MS: I think they are different in the sense that MetadataDocumentUri can be a public endpoint or it can be a local file whereas the client code will not work with this as a public endpoint.

src/CodeGen/ODataT4CodeGenerator.ttinclude
KanishManuja-MS: Replied to the comment above and for the same reason, this should not be a boolean as we need a location where writes can be made.

src/CodeGen/ODataT4CodeGenerator.tt
xuzhg: MetadataDocumentUri can be a public endpoint or it can be a local file. Why can't we use it as the TempFilePath?

KanishManuja-MS: One difference is that the header is stripped out in the temp file. The other reason is that if we tried using MetadataDocumentUri, on top of adding the stripping the header code, we would need more code to read from both file and public endpoint. Since it seemed like a low frequency issue, I did not want to change a ton for it.

AlanWong-MS: Got it, makes sense that it's different from MetadataDocumentUri then. Instead of = "", we can use = string.Empty.

src/CodeGen/ODataT4CodeGenerator.ttinclude
```diff
@@ -3420,6 +3440,18 @@ namespace <#= fullNamespace #>
 
     internal override void WriteGeneratedEdmModel(string escapedEdmxString)
     {
+
+        string path = this.context.TempFilePath; 
+        
+        if(!String.IsNullOrEmpty(path))
```
AlanWong-MS: We usually have space between if and (. Same comment for the other cases in this file

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTestDescriptors.cs
```diff
@@ -79,26 +79,31 @@ private static void VerifyGeneratedCode(string actualCode, Dictionary<string, st
...
+                bool equal = rawExpectedCode == rawActualCode;
 
                 if (!equal)
                 {
                     string filename = key + (useDSC ? "DSC" : "") + (isCSharp ? ".cs" : ".vb");
                     string currentFolder = Directory.GetCurrentDirectory();
                     string path = Path.Combine(currentFolder, filename);
                     File.WriteAllText(path, actualBak);
-                    string basePath = string.Format(currentFolder + @"\..\CodeGen\{0}", filename);
+                    string basePath = string.Format(Path.Combine(currentFolder,"Expected" + filename));
```
AlanWong-MS: nit: space after `currentFolder,`

src/CodeGen/ODataT4CodeGenerator.tt
AlanWong-MS: Discussed with Kanish and we'll be keeping "" instead of string.Empty due to code churn in test cases.

Issue #1181, opened 2018-06-06
Remove unneeded json string escaping in JsonValueUtils


maximpashuk:
I am using Microsoft.AspNetCore.OData beta4, but my issue related to Microsoft.OData.Core.

I see differences in json response containing russian language symbols in normal odata query and batched odata query.
Batched odata response always escaping russian symbols while normal response has no escaping.
```
"country": "Россия" <-- no escaping in GET request

"country":"\u0420\u043e\u0441\u0441\u0438\u044f" <-- escaping in $batch request
```

I realize from the docs ([OData JSON Format Version 4.0](http://docs.oasis-open.org/odata/odata-json-format/v4.0/cs01/odata-json-format-v4.0-cs01.html#_Toc365464699) and [RFC4627](https://www.ietf.org/rfc/rfc4627.txt)) that escaping of non-ASCII symbols is not required.
Docs says just that "Any character MAY be escaped"

From source code i realized that $batch endpoint uses [JsonValueUtils](https://github.com/OData/odata.net/blob/fb4630ec1cf49bb10ef41a33bd81863f8b765220/src/Microsoft.OData.Core/Json/JsonValueUtils.cs#L476) there ALL characters with code > 127 ALWAYS escaped.

I think forcing escaping is not optimal - 6 bytes (\uXXXX) on each character is growing json size, also deserializing logic is more complicated.
Also ODataOutputFormatter from ODATA WebApi has to forced escaping.

So please refactor JsonValueUtils to escape symbols only required to be escaped (", \, etc.)

mikepizzo:
Can you clarify where the escaping is occurring? Is it just escaping the URL, or is it escaping properties within a (request? response?) payload?

I was able to write a simple test that produced the following output:

~~~json
{
  "responses": [
    {
      "id": "1",
      "status": 200,
      "headers": {
        "odata-version": "4.0",
        "content-type": "application/json;odata.metadata=minimal;odata.streaming=true;IEEE754Compatible=false;charset=utf-8"
      },
      "body": {
        "@odata.context": "http://testservice/$metadata#customers",
        "value": [
          {
            "@odata.id": "customers/1",
            "id": "Россия"
          }
        ]
      }
    }
  ]
}
~~~

maximpashuk:
Hi.
Escaping is occurred in properties in response payload when using **/$batch** odata endpoint.
Escaping not occurred when using simple GET/POST requests.

Batched responses has special format, different from format you just posted.

[...]

PR #1242, opened 2018-08-31
Fix issue #1181, refactor for the json string escaping in JsonValueUt…

xuzhg:
…ils class

### Issues

*This pull request fixes issue #1181.*

### Description

Add configuration option for the ODataMessageWriterSetting using the new added `ODataStringEscapeOption`, internally, the JsonValueUtils will use the escape setting to write the string value.

By default, we use "EscapeNonAscii" option to make sure without breaking change for the default behavior. 

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
@@ -391,7 +391,7 @@ namespace Microsoft.OData.Json
                     char c = inputString[currentIndex];
                     string escapedString = null;
 
-                    if (stringEscapeOption == ODataStringEscapeOption.EscapeNonAscii || c < 0x7F)
+                    if (stringEscapeOption == ODataStringEscapeOption.EscapeNonAscii || c <= 0x7F)
                     {
                         escapedString = JsonValueUtils.SpecialCharToEscapedStringMap[c];
                     }
--- a/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
@@ -38,7 +38,6 @@ namespace Microsoft.OData.Json
         /// </summary>
         internal static readonly NumberFormatInfo ODataNumberFormatInfo = JsonValueUtils.InitializeODataNumberFormatInfo();
 
-
         /// <summary>
         /// Const tick value for calculating tick values.
         /// </summary>
@@ -296,8 +295,9 @@ namespace Microsoft.OData.Json
         /// </summary>
         /// <param name="writer">The text writer to write the output to.</param>
         /// <param name="value">String value to be written.</param>
+        /// <param name="stringEscapeOption">The string escape option.</param>
         /// <param name="buffer">Char buffer to use for streaming data</param>
-        internal static void WriteValue(TextWriter writer, string value, ref char[] buffer)
+        internal static void WriteValue(TextWriter writer, string value, ODataStringEscapeOption stringEscapeOption, ref char[] buffer)
         {
             Debug.Assert(writer != null, "writer != null");
 
@@ -307,7 +307,7 @@ namespace Microsoft.OData.Json
             }
             else
             {
-                JsonValueUtils.WriteEscapedJsonString(writer, value, ref buffer);
+                JsonValueUtils.WriteEscapedJsonString(writer, value, stringEscapeOption, ref buffer);
             }
         }
 
@@ -337,8 +337,10 @@ namespace Microsoft.OData.Json
         /// </summary>
         /// <param name="writer">The text writer to write the output to.</param>
         /// <param name="inputString">Input string value.</param>
+        /// <param name="stringEscapeOption">The string escape option.</param>
         /// <param name="buffer">Char buffer to use for streaming data</param>
-        internal static void WriteEscapedJsonString(TextWriter writer, string inputString, ref char[] buffer)
+        internal static void WriteEscapedJsonString(TextWriter writer, string inputString,
+            ODataStringEscapeOption stringEscapeOption, ref char[] buffer)
         {
             Debug.Assert(writer != null, "writer != null");
             Debug.Assert(inputString != null, "The string value must not be null.");
@@ -346,7 +348,7 @@ namespace Microsoft.OData.Json
             writer.Write(JsonConstants.QuoteCharacter);
 
             int firstIndex;
-            if (!JsonValueUtils.CheckIfStringHasSpecialChars(inputString, out firstIndex))
+            if (!JsonValueUtils.CheckIfStringHasSpecialChars(inputString, stringEscapeOption, out firstIndex))
             {
                 writer.Write(inputString);
             }
@@ -387,7 +389,12 @@ namespace Microsoft.OData.Json
                 for (; currentIndex < inputStringLength; currentIndex++)
                 {
                     char c = inputString[currentIndex];
-                    string escapedString = JsonValueUtils.SpecialCharToEscapedStringMap[c];
+                    string escapedString = null;
+
+                    if (stringEscapeOption == ODataStringEscapeOption.EscapeNonAscii || c < 0x7F)
+                    {
+                        escapedString = JsonValueUtils.SpecialCharToEscapedStringMap[c];
+                    }
 
                     // Append the unhandled characters (that do not require special treament)
                     // to the buffer.
@@ -491,13 +498,14 @@ namespace Microsoft.OData.Json
         }
 
         /// <summary>
-        /// Checks if the string contains special char and returns the first index 
+        /// Checks if the string contains special char and returns the first index
         /// of special char if present.
         /// </summary>
         /// <param name="inputString">string that might contain special characters.</param>
+        /// <param name="stringEscapeOption">The string escape option.</param>
         /// <param name="firstIndex">first index of the special char</param>
         /// <returns>A value indicating whether the string contains special character</returns>
-        private static bool CheckIfStringHasSpecialChars(string inputString, out int firstIndex)
+        private static bool CheckIfStringHasSpecialChars(string inputString, ODataStringEscapeOption stringEscapeOption, out int firstIndex)
         {
             Debug.Assert(inputString != null, "The string value must not be null.");
 
@@ -506,6 +514,11 @@ namespace Microsoft.OData.Json
             for (int currentIndex = 0; currentIndex < inputStringLength; currentIndex++)
             {
                 char c = inputString[currentIndex];
+
+                if (stringEscapeOption == ODataStringEscapeOption.EscapeOnlyControls && c >= 0x7F)
+                {
+                    continue;
+                }
 
                 // Append the un-handled characters (that do not require special treatment)
                 // to the string builder when special characters are detected.
--- a/src/Microsoft.OData.Core/Json/JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriter.cs
@@ -39,6 +39,11 @@ namespace Microsoft.OData.Json
         private readonly bool isIeee754Compatible;
 
         /// <summary>
+        /// Gets or sets a value indicating how to escape the string when writing JSON string.
+        /// </summary>
+        private readonly ODataStringEscapeOption stringEscapeOption;
+
+        /// <summary>
         /// The buffer to help with streaming responses.
         /// </summary>
         private char[] buffer;
@@ -47,12 +52,24 @@ namespace Microsoft.OData.Json
         /// Creates a new instance of Json writer.
         /// </summary>
         /// <param name="writer">Writer to which text needs to be written.</param>
-        /// <param name="isIeee754Compatible">if it is IEEE754Compatible</param>
+        /// <param name="isIeee754Compatible">if it is IEEE754Compatible.</param>
         internal JsonWriter(TextWriter writer, bool isIeee754Compatible)
+            : this(writer, isIeee754Compatible, ODataStringEscapeOption.EscapeNonAscii)
+        {
+        }
+
+        /// <summary>
+        /// Creates a new instance of Json writer.
+        /// </summary>
+        /// <param name="writer">Writer to which text needs to be written.</param>
+        /// <param name="isIeee754Compatible">if it is IEEE754Compatible.</param>
+        /// <param name="stringEscapeOption">Specifies how to escape string.</param>
+        internal JsonWriter(TextWriter writer, bool isIeee754Compatible, ODataStringEscapeOption stringEscapeOption)
         {
             this.writer = new NonIndentedTextWriter(writer);
             this.scopes = new Stack<Scope>();
             this.isIeee754Compatible = isIeee754Compatible;
+            this.stringEscapeOption = stringEscapeOption;
         }
 
         /// <summary>
@@ -167,7 +184,7 @@ namespace Microsoft.OData.Json
 
             currentScope.ObjectCount++;
 
-            JsonValueUtils.WriteEscapedJsonString(this.writer, name, ref this.buffer);
+            JsonValueUtils.WriteEscapedJsonString(this.writer, name, this.stringEscapeOption, ref this.buffer);
             this.writer.Write(JsonConstants.NameValueSeparator);
         }
 
@@ -231,7 +248,8 @@ namespace Microsoft.OData.Json
             // if it is IEEE754Compatible, write numbers with quotes; otherwise, write numbers directly.
             if (isIeee754Compatible)
             {
-                JsonValueUtils.WriteValue(this.writer, value.ToString(CultureInfo.InvariantCulture), ref this.buffer);
+                JsonValueUtils.WriteValue(this.writer, value.ToString(CultureInfo.InvariantCulture),
+                    this.stringEscapeOption, ref this.buffer);
             }
             else
             {
@@ -270,7 +288,8 @@ namespace Microsoft.OData.Json
             // if it is not IEEE754Compatible, write numbers directly without quotes;
             if (isIeee754Compatible)
             {
-                JsonValueUtils.WriteValue(this.writer, value.ToString(CultureInfo.InvariantCulture), ref this.buffer);
+                JsonValueUtils.WriteValue(this.writer, value.ToString(CultureInfo.InvariantCulture),
+                    this.stringEscapeOption, ref this.buffer);
             }
             else
             {
@@ -345,7 +364,7 @@ namespace Microsoft.OData.Json
         public void WriteValue(string value)
         {
             this.WriteValueSeparator();
-            JsonValueUtils.WriteValue(this.writer, value, ref this.buffer);
+            JsonValueUtils.WriteValue(this.writer, value, this.stringEscapeOption, ref this.buffer);
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Core/ODataStringEscapeOption.cs
+++ b/src/Microsoft.OData.Core/ODataStringEscapeOption.cs
@@ -1,0 +1,25 @@
+//---------------------------------------------------------------------
+// <copyright file="ODataStringEscapeOption.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData
+{
+    /// <summary>
+    /// Enum type to specify how to escape string value when writing JSON text.
+    /// </summary>
+    public enum ODataStringEscapeOption
+    {
+        /// <summary>
+        /// All non-ASCII and control characters (e.g. newline) are escaped.
+        /// </summary>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704", Justification = "Ascii is meaningful")]
+        EscapeNonAscii,
+
+        /// <summary>
+        /// Only control characters (e.g. newline) are escaped.
+        /// </summary>
+        EscapeOnlyControls
+    }
+}
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsTests.cs
@@ -18,34 +18,80 @@ namespace Microsoft.OData.Tests.Json
         private MemoryStream stream;
         private NonIndentedTextWriter writer;
         private char[] buffer;
-        private Dictionary<string, string> escapedCharMap = new Dictionary<string, string>()
-        {
-            {"\r\n","\\r\\n"},
-            {"\r","\\r"},
-            {"\t","\\t"},
-            {"\"","\\\""},
-            {"\\","\\\\"},
-            {"\n","\\n"},
-            {"\b","\\b"},
-            {"\f","\\f"},
-            {string.Format("{0}",(char)0),string.Format(CultureInfo.InvariantCulture, "\\u{0:x4}", 0)},
-            {string.Format("{0}",(char)0x80), string.Format(CultureInfo.InvariantCulture, "\\u{0:x4}", 0x80)},
-        };
+        private IDictionary<string, string> escapedCharMap;
+        private IDictionary<string, string> controlCharsMap;
 
-        [Fact]
-        public void WriteEmptyStringShouldWork()
+        public JsonValueUtilsTests()
+        {
+            controlCharsMap = new Dictionary<string, string>
+            {
+                {string.Format("{0}",(char)0),string.Format(CultureInfo.InvariantCulture, "\\u{0:x4}", 0)},
+                {"\r\n","\\r\\n"},
+                {"\r","\\r"},
+                {"\t","\\t"},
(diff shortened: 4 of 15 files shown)


src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -387,7 +389,13 @@ internal static void WriteEscapedJsonString(TextWriter writer, string inputStrin
                 for (; currentIndex < inputStringLength; currentIndex++)
                 {
                     char c = inputString[currentIndex];
-                    string escapedString = JsonValueUtils.SpecialCharToEscapedStringMap[c];
+                    string escapedString = null;
+
+                    if (stringEscapeOption == ODataStringEscapeOption.EscapeNonAscii ||
+                        (c < 0x7F && stringEscapeOption == ODataStringEscapeOption.EscapeOnlyControls))
```
mikepizzo: >&& stringEscapeOption == ODataStringEscapeOption.EscapeOnlyControls [](start = 34, length = 67)

Is this part of the condition required, or could we just check (stringEscapeOption == ODataStringEscapeOption.EscapeNonAscii || c < 0x7F)?

xuzhg: Yes. Will change, thanks.

src/Microsoft.OData.Client/EntityTracker.cs
```diff
@@ -540,24 +540,24 @@ private void EnsureIdentityToResource()
         {
             if (null == this.identityToDescriptor)
             {
-            #if PORTABLELIB && WINDOWSPHONE
+#if PORTABLELIB && WINDOWSPHONE
```
biaol-odata: Anything else other than indentation changed in this PR? No logic changes, correct?

xuzhg: Won't change. 

There's a build warning complaining about this. Why do we wait to another PR that maybe never happen?

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -52,7 +51,7 @@ internal static class JsonValueUtils
         /// <summary>
         /// Map of special characters to strings.
         /// </summary>
-        private static readonly string[] SpecialCharToEscapedStringMap = JsonValueUtils.CreateSpecialCharToEscapedStringMap();
+        private static readonly string[] SpecialCharToEscapedStringMap = CreateSpecialCharToEscapedStringMap();
```
biaol-odata: nit: This changed line is not consistent with line #39.

biaol-odata:
I am still not seeing PR-triggered build being done automatically. Can you please check?

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -507,9 +515,15 @@ private static bool CheckIfStringHasSpecialChars(string inputString, out int fir
             {
                 char c = inputString[currentIndex];
 
+                if (escapeOption == ODataStringEscapeOption.EscapeOnlyControls && c >= 0x7F)
+                {
+                    // control character is ahead of whitespace
```
biaol-odata: Not sure how this comment is related to "c>=0x7F"?

src/Microsoft.OData.Core/Json/JsonWriter.cs
```diff
@@ -76,6 +88,11 @@ internal enum ScopeType
             Padding = 2,
         }
 
+        /// <summary>
+        /// Gets or sets a value indicating how to escape the string when writing JSON string.
+        /// </summary>
+        internal ODataStringEscapeOption StringEscapeOption { get; set; }
```
biaol-odata: This field is not supposed to be set after the JsonWriter instance is instantiated. We should remove the setter method and make it read-only, similar to other fields such as isIeee754Compatible.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -791,17 +792,26 @@ protected override void Dispose(bool disposing)
...
             {
-                return new JsonWriter(textWriter, isIeee754Compatible);
+                return new JsonWriter(textWriter, isIeee754Compatible, stringEscapeOption);
             }
 
             var jsonWriterFactory = container.GetRequiredService<IJsonWriterFactory>();
             var jsonWriter = jsonWriterFactory.CreateJsonWriter(textWriter, isIeee754Compatible);
             Debug.Assert(jsonWriter != null, "jsonWriter != null");
 
+            // Owing to the breaking change, we can't change the public interface "IJsonWriterFactory".
```
biaol-odata: -> "To avoid breaking change"?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -791,17 +792,26 @@ protected override void Dispose(bool disposing)
...
+            // So, use the following working around.
```
biaol-odata: -> "work around"

src/Microsoft.OData.Core/ODataStringEscapeOption.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    /// <summary>
+    /// Enum type to specify how to escape string value when writing JSON text.
+    /// </summary>
+    public enum ODataStringEscapeOption
+    {
+        /// <summary>
+        /// All non-ASCII and control characters (e.g. newline) are escaped.
+        /// </summary>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1704", Justification = "Ascii is meaningful")]
```
biaol-odata: -> "EscapeNonAsciiAndControls", because description above indicates that non-ASCII doesn't contains control chars.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsTests.cs
```diff
@@ -134,23 +187,43 @@ public void WriteSpecialCharactersAtEndOfBufferLengthShouldWork()
...
+                JsonValueUtils.WriteEscapedJsonString(this.writer, string.Format("S{0}M{0}E{0}", specialChar),
+                    escapeOption, ref charBuffer);
+                this.StreamToString().Should().Be(string.Format("\"S{0}M{0}E{0}\"", this.controlCharsMap[specialChar]));
+            }
+        }
+
+        [Theory]
+        [InlineData(ODataStringEscapeOption.EscapeOnlyControls)]
+        [InlineData(ODataStringEscapeOption.EscapeNonAscii)]
+        public void WriteStringWithNoSpecialCharShouldLeaveBufferUntouched(ODataStringEscapeOption escapeOption)
```
biaol-odata: -> "stringEscapeOption", see above.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsTests.cs
```diff
@@ -134,23 +187,43 @@ public void WriteSpecialCharactersAtEndOfBufferLengthShouldWork()
...
             this.StreamToString().Should().Be("\"StartMiddleEnd\"");
             charBuffer.Should().BeNull("Char Buffer for cases with zero special characters should need to use buffer");
         }
 
-        [Fact]
-        public void WriteStringShouldIgnoreExistingContentsOfTheBuffer()
+        [Theory]
+        [InlineData(ODataStringEscapeOption.EscapeOnlyControls)]
+        [InlineData(ODataStringEscapeOption.EscapeNonAscii)]
+        public void WriteStringShouldIgnoreExistingContentsOfTheBuffer(ODataStringEscapeOption escapeOption)
```
biaol-odata: -> "stringEscapeOption", see above.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/NonIndentedTextWriterTests.cs
```diff
@@ -32,19 +32,23 @@ public class NonIndentedTextWriterTests
             {string.Format("{0}", (char)0x80), string.Format(CultureInfo.InvariantCulture, "\\u{0:x4}", 0x80)},
         };
 
-        [Fact]
-        public void WriteEmptyStringShouldWork()
+        [Theory]
+        [InlineData(ODataStringEscapeOption.EscapeOnlyControls)]
+        [InlineData(ODataStringEscapeOption.EscapeNonAscii)]
+        public void WriteEmptyStringShouldWork(ODataStringEscapeOption escapeOption)
```
biaol-odata: -> "stringEscapeOption", see above.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/NonIndentedTextWriterTests.cs
```diff
@@ -32,19 +32,23 @@ public class NonIndentedTextWriterTests
...
+            JsonValueUtils.WriteEscapedJsonString(this.writer, string.Empty, escapeOption, ref this.buffer);
             this.StreamToString().Should().Be("\"\"");
         }
 
-        [Fact]
-        public void WriteNonSpecialCharactersShouldWork()
+        [Theory]
+        [InlineData(ODataStringEscapeOption.EscapeOnlyControls)]
+        [InlineData(ODataStringEscapeOption.EscapeNonAscii)]
+        public void WriteNonSpecialCharactersShouldWork(ODataStringEscapeOption escapeOption)
```
biaol-odata: -> "stringEscapeOption", see above.

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -387,7 +389,12 @@ internal static void WriteEscapedJsonString(TextWriter writer, string inputStrin
                 for (; currentIndex < inputStringLength; currentIndex++)
                 {
                     char c = inputString[currentIndex];
-                    string escapedString = JsonValueUtils.SpecialCharToEscapedStringMap[c];
+                    string escapedString = null;
+
+                    if (stringEscapeOption == ODataStringEscapeOption.EscapeNonAscii || c < 0x7F)
```
biaol-odata: how about c==0x7F ??

src/Microsoft.OData.Client/EntityTracker.cs
biaol-odata: I don't see any related build warning in the currently merged build.
https://identitydivision.visualstudio.com/OData/_build/results?buildId=177022

If this is one of the warnings show above, we can definitely fix it.

xuzhg:
Build is passed. See https://identitydivision.visualstudio.com/OData/OData%20Team/_build/results?buildId=182895

xuzhg:
@biaol-odata I spent the whole day on working the build trigger issue (Investigating and debugging) . I also sent out the email to illustrate the root cause (maybe).  Do you think to fix the Build trigger issue is high-priority?

biaol-odata:
IMHO, yes. Build system issue should be of high priority because it affects the whole development process. Do you agree?

xuzhg:
@mikepizzo 

Yes. JSON Writer always encodes property values containing non-ascii characters. ( characters less than ' ' and bigger than 0x7F). As I said, our client tools (Postman, Fiddle) cheat us that they will help to show the normal character (decoded characters) if the tools detect the response payload is a JSON text. 

NewtonSoft.json has the similiar enum type at https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/StringEscapeHandling.cs#L41
From the code in the Newtonsoft.json at https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Utilities/JavaScriptUtils.cs#L216-L238,  it looks escape the non-ascii character also. 

My change is based on the Newtonsoft.json and try to avoid behaviour change for our current customer.

biaol-odata:
@xuzhg You mentioned "bigger than 0x7F", but some of the code change above treat 0x7F as special character (please search my comments above). Can you check for potential off-by-one error?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
xuzhg: #Resolved.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
xuzhg: #Resolved.

src/Microsoft.OData.Core/ODataStringEscapeOption.cs
xuzhg: #Won't fix.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsTests.cs
xuzhg: #Resolved.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsTests.cs
xuzhg: #resolved.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/NonIndentedTextWriterTests.cs
xuzhg: #Resolved.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/NonIndentedTextWriterTests.cs
xuzhg: #Resolved.

src/Microsoft.OData.Core/ODataStringEscapeOption.cs
biaol-odata: why won't fix? I believe "EscapeNonAsciiAndControls" is more aligned with the description above.

xuzhg: Use the same enum member in JSON.NET.

src/Microsoft.OData.Core/Json/JsonWriter.cs
xuzhg: See the new added comments in "ODataJsonLightOutputContext.cs"

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
biaol-odata: There is a misalignment for 0x7F.
When SpecialCharToEscapedStringMap is created (https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Json/JsonValueUtils.cs#L567-L569): 0x7F is considered non-special character. But, by using "<" here, 0x7F is considered special character.

src/Microsoft.OData.Core/ODataStringEscapeOption.cs
biaol-odata: well, for people not referencing JSON.NET, the discrepancy between this enum value and it description is confusing.

src/Microsoft.OData.Core/Json/JsonWriter.cs
biaol-odata: link?

xuzhg:
```C#
// We can't change the public interface "IJsonWriterFactory" to avoid breaking changes. 
// So, use the following work around. 
           JsonWriter defaultJsonWriter = jsonWriter as JsonWriter; 
            if (defaultJsonWriter != null) 
            { 
                defaultJsonWriter.StringEscapeOption = stringEscapeOption; 
            } 
```

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
xuzhg: You can consider "0x7F" here as a threshold. A char is a special char or not is not identified from the value, it's from "specialCharToEscapedStringMap". 

owing that specialCharToEscapedStringMap[0x7f] == null, so it's not a special char.

src/Microsoft.OData.Core/Json/JsonWriter.cs
biaol-odata: IMHO, this is a bit hacky.

xuzhg: Maybe. Do you have a better solution?

src/Microsoft.OData.Core/ODataStringEscapeOption.cs
xuzhg: Where's the confusing? See the same comments: https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/StringEscapeHandling.cs#L38-L41

src/Microsoft.OData.Core/Json/JsonWriter.cs
biaol-odata: please check with @mikepizzo

biaol-odata:
I agree with Mike's comment [above](https://github.com/OData/odata.net/pull/1242#issuecomment-417812330):
"Currently, JSON Writer always encodes property values containing non-ascii characters."
And issue discussed here is not related to batch.

Per [OData spec for json format](http://docs.oasis-open.org/odata/odata-json-format/v4.01/cs01/odata-json-format-v4.01-cs01.html#sec_PrimitiveValue), JSON string escaping rules should ***always*** be applied:
        _Values of type Edm.String are represented as JSON strings, using the JSON string escaping rules._

As for JSON escaping rules, [this link](https://stackoverflow.com/questions/19176024/how-to-escape-special-characters-in-building-a-json-string) lists three official references:

- [json.org](json.org)

- [ecma-international.org](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf)

- [RFC 7519](https://tools.ietf.org/html/rfc7159#page-8)

These references all specify that ASCII value > 0x7F should be escaped.

image

So, current JsonWriter is actually implemented per spec?

igor-moskvitin:
@biaol-odata [rfc 7159](https://tools.ietf.org/html/rfc7159#page-8) specify that `..Any character MAY BE escaped..` but not `should be escaped`. 
[ecma-international](http://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf) points the same. Even a picture, that you posted says - `any`, but not `all`.
I think that if [Json.NET](https://www.newtonsoft.com) provide options for configuring this behaviour, OData can provide the same. Otherwise we had some extra traffic when its no needed.

mikepizzo:
Okay; mostly wanted to be sure I understood the mitigation.

Alternative design would be to make the JSONWriter constructor that takes the ODataStringEscapeOption public, and allow the user to create our JSONWriter with that option and pass it in through DI.  That way we wouldn't need an ODataMessageWriterSetting that only applied to construction of our JSON Writer.

---
In reply to: [417814637](https://github.com/OData/odata.net/pull/1242#issuecomment-417814637) [](ancestors = 417814637)

mikepizzo:
As per above diagram, any UNICODE character except " or a control character can be sent through unescaped.  For both size and simplicity, we should provide a way to not escape by default (similar to JSON.NET.)

---
In reply to: [418545585](https://github.com/OData/odata.net/pull/1242#issuecomment-418545585) [](ancestors = 418545585)

mikepizzo:
Let's track build issues outside of this PR.

---
In reply to: [417796273](https://github.com/OData/odata.net/pull/1242#issuecomment-417796273) [](ancestors = 417796273)

mikepizzo:
Let's track build issues outside of this PR.

---
In reply to: [417812230](https://github.com/OData/odata.net/pull/1242#issuecomment-417812230) [](ancestors = 417812230)

mikepizzo:
Let's track build issues outside of this PR.

---
In reply to: [417811261](https://github.com/OData/odata.net/pull/1242#issuecomment-417811261) [](ancestors = 417811261)

mikepizzo:
Let's track build issues outside of this PR.

---
In reply to: [417813301](https://github.com/OData/odata.net/pull/1242#issuecomment-417813301) [](ancestors = 417813301)

mikepizzo:
Let's track build issues outside of this PR.

---
In reply to: [417814145](https://github.com/OData/odata.net/pull/1242#issuecomment-417814145) [](ancestors = 417814145)

src/Microsoft.OData.Core/Json/JsonWriter.cs
mikepizzo: What if we let the user create the JSON Writer (and specify the appropriate escape option) and then give it to us through DI?  Then we wouldn't have to add the property to ODataMessageWriterSettings.

---
In reply to: [215032187](https://github.com/OData/odata.net/pull/1242#discussion_r215032187) [](ancestors = 215032187)

mikepizzo: Added comments to previous issue.

---
In reply to: [215051592](https://github.com/OData/odata.net/pull/1242#discussion_r215051592) [](ancestors = 215051592)

mikepizzo: Added comments to previous issue.

---
In reply to: [215057928](https://github.com/OData/odata.net/pull/1242#discussion_r215057928) [](ancestors = 215057928)

mikepizzo: This is internal, so I don't mind making it r/w and having it internally set. I don't like having a messagewriter setting that only applies in this case, though.  See my comments below on an potential alternative design.

---
In reply to: [215038804](https://github.com/OData/odata.net/pull/1242#discussion_r215038804) [](ancestors = 215038804)

src/Microsoft.OData.Client/EntityTracker.cs
mikepizzo: I think Biao was just confirming that there were no material changes.  I am fine with fixing indentation issues that we find along the way.

---
In reply to: [214475994](https://github.com/OData/odata.net/pull/1242#discussion_r214475994) [](ancestors = 214475994)

mikepizzo: I'm fine with fixing indentation as part of this PR.

---
In reply to: [214489682](https://github.com/OData/odata.net/pull/1242#discussion_r214489682) [](ancestors = 214489682)

src/Microsoft.OData.Core/ODataStringEscapeOption.cs
mikepizzo: I am fine with the current name; control characters are always escaped; there is no way to change that.  so the flag only affects ascii characters.  Since there is already a precedent with NewtonSoft might as well follow that.

---
In reply to: [214485980](https://github.com/OData/odata.net/pull/1242#discussion_r214485980) [](ancestors = 214485980)

mikepizzo: I'm fine with the current name.

---
In reply to: [215054974](https://github.com/OData/odata.net/pull/1242#discussion_r215054974) [](ancestors = 215054974)

biaol-odata: In our changes, {EscapeNonAscii, EscapeOnlyControls} are used.
In JSON.NET, {Default, EscapeNonAscii, ...} are used, without specifying explicitly 'Controls'.
That is why I think EscapeNonAsciiAndControls is more clear for our changes, also aligned with description.
This Enum is part of public API, where we should be as clear as possible.

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
biaol-odata: Why changing it to a different style by removing JsonValueUtils prefix everywhere in this file? I believe the original prefixed style is the primary style for all other files in the odata.net project that is helpful for auto-complete / readability, even though it is not required for syntax.
I would recommend keeping the original style.

xuzhg:
@mikepizzo I updated the PR directly with your suggestion. Please take a look the changes. I will continue on the Unit test.

src/Microsoft.OData.Core/ODataStringEscapeOption.cs
```diff
@@ -0,0 +1,25 @@
+//---------------------------------------------------------------------
+// <copyright file="ODataStringEscapeOption.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData
```
mikepizzo: >Microsoft.OData [](start = 10, length = 15)

Should this class be in Microsoft.OData.Json?

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
biaol-odata: Is this one resolved?

xuzhg: changed to use <=

Issue #1262, opened 2018-09-25
OutOfLine Annotation for EnumMember throw exception when write to CSDL


xuzhg:
EnumMember is annotatable, it works for "Inline" serialization location. However, it can't work for "OutOfLine" serialization location.

### Assemblies affected

* ODL 7.x.*

### Reproduce steps

For example, I have the following codes to add an enum type with a
```
EdmModel model = new EdmModel();
EdmEnumType appliance = new EdmEnumType("NS", "Appliance", EdmPrimitiveTypeKind.Int64, isFlags: true);
model.AddElement(appliance);

var stove = new EdmEnumMember(appliance, "Stove", new EdmEnumMemberValue(1));
appliance.AddMember(stove);
```
If I use the following code to add annotation for `stove` member, it can work.

```
EdmVocabularyAnnotation annotation = new EdmVocabularyAnnotation(stove , CoreVocabularyModel.LongDescriptionTerm, new EdmStringConstant("Stove Top"));
annotation.SetSerializationLocation(model, EdmVocabularyAnnotationSerializationLocation.Inline);
model.SetVocabularyAnnotation(annotation);
```
if I change to use 
```
annotation.SetSerializationLocation(model, EdmVocabularyAnnotationSerializationLocation.OutOfLine);
```
It will throw exception 

```c#
Message:	System.NullReferenceException :Object reference not set to an instance of an object.

Result StackTrace:	
at Microsoft.OData.Edm.EdmUtil.IsQualifiedName(String name) in E:\github\xuzhg\odata.net\src\Microsoft.OData.Edm\EdmUtil.cs:line 178
   at Microsoft.OData.Edm.Csdl.Serialization.SerializationValidator.<>c.<.cctor>b__9_4(ValidationContext context, IEdmVocabularyAnnotation annotation) in E:\github\xuzhg\odata.net\src\Microsoft.OData.Edm\Csdl\Serialization\SerializationValidator.cs:line 97
   at Microsoft.OData.Edm.Validation.ValidationRule`1.Evaluate(ValidationContext context, Object item) in E:\github\xuzhg\odata.net\src\Microsoft.OData.Edm\Validation\ValidationRule.cs:line 49
[...]

PR #1263, opened 2018-09-25
Fix the writer/reader for out of line annotation for Enum member

xuzhg:
### Issues

*This pull request fixes issue #1262.*

### Description

* Fix the writer for out of line annotation for enum member
* Fix the reader for out of line annotation for enum member

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -281,7 +281,9 @@ public override IEnumerable<IEdmVocabularyAnnotation> FindDeclaredVocabularyAnno
                 {
                     foreach (CsdlAnnotation annotation in annotations.Annotations.Annotations)
                     {
-                        result.Add(this.WrapVocabularyAnnotation(annotation, annotations.Context, null, annotations, annotations.Annotations.Qualifier));
+                        IEdmVocabularyAnnotation vocabAnnotation = this.WrapVocabularyAnnotation(annotation, annotations.Context, null, annotations, annotations.Annotations.Qualifier);
+                        vocabAnnotation.SetSerializationLocation(this, EdmVocabularyAnnotationSerializationLocation.OutOfLine);
+                        result.Add(vocabAnnotation);
```
xuzhg: The change here is straightforward to call "SetSerializationLocation" to set "OutOfLine" for the annotation.

src/Microsoft.OData.Edm/EdmUtil.cs
```diff
@@ -322,16 +324,20 @@ internal static string FullyQualifiedName(IEdmVocabularyAnnotatable element)
                             }
                         }
                     }
-                    else
+                    else if (parameter != null)
                     {
-                        IEdmOperationParameter parameter = element as IEdmOperationParameter;
-                        if (parameter != null)
```
xuzhg: The changes here is straightforward to consider the "EnumMember" scenario, same as other two scenarios: (property, parameter).

For we have the following sub-items:

1. Property, the full name is "fullTypeName/propertyName"
2. Parameter, the full name is  "qualified operation name/parameterName".
3. New added for EnumMember, the full name is "fullEnumTypeName/EnumMemberName".

src/Microsoft.OData.Edm/EdmUtil.cs
```diff
@@ -318,6 +318,8 @@ internal static string FullyQualifiedName(IEdmVocabularyAnnotatable element)
                 else
                 {
                     IEdmProperty property = element as IEdmProperty;
+                    IEdmOperationParameter parameter = element as IEdmOperationParameter;
+                    IEdmEnumMember enumMember = element as IEdmEnumMember;
```
mikepizzo: Since casts can be relatively expensive, probably better to try each one in turn, and only attempt cast if the previous cast fails.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
mikepizzo: Help me understand why we're doing this? Why are we changing all of the annotations to OutOflLine?  Does this mean that, if I read in some inline annotations, then write them back out, they will all be out of line?

---
In reply to: [221688408](https://github.com/OData/odata.net/pull/1263#discussion_r221688408) [](ancestors = 221688408)

xuzhg: Ok. 
The function `FindDeclaredVocabularyAnnotations(element)` is to return all annotations related to the given element, include all "**InLine**" and "**OutOfLine**" annotations. 
For the "**InLine**" annotation, we have the codes to set `SerializationLocation` as `InLine`, see codes [here](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs#L507-L509)

But, for the **OutOfLine** annotation, we miss to set `SerializationLocation` for them. If we call "GetSerializationLocation", it will give you `null`. So, I changed the codes to set location as "**OutOfLine**" for them, same as for "**InLine**" annotation.

That's straightforward because the annotations are from the "**outOfLineAnnotations**" collection, see the following `if` clause.

```C#
 if (fullName != null && this.outOfLineAnnotations.TryGetValue(fullName, out elementAnnotations))
 {
         List<IEdmVocabularyAnnotation> result = new List<IEdmVocabularyAnnotation>();
         foreach (CsdlSemanticsAnnotations annotations in elementAnnotations)
         {
               foreach (CsdlAnnotation annotation in annotations.Annotations.Annotations)
                {
                     IEdmVocabularyAnnotation vocabAnnotation = this.WrapVocabularyAnnotation(annotation, annotations.Context, null, annotations, annotations.Annotations.Qualifier);
                     vocabAnnotation.SetSerializationLocation(this, EdmVocabularyAnnotationSerializationLocation.OutOfLine);
                     result.Add(vocabAnnotation);
                }
           }

         return inlineAnnotations.Concat(result);
  }

src/Microsoft.OData.Edm/EdmUtil.cs
xuzhg: @mikepizzo Changed the codes. Thanks.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
mikepizzo: I see; thanks.  Part of what was confusing me was the fact that we are concatenating the annotations to an enumerable named "inlineAnnotations"; I thought that we were setting all of the annotations to out of line, but now see that we were concatenating the outOfLine annotations to the inlineAnnotations and returning that concatenated list...

xuzhg:
@mikepizzo Also updated. Thanks.

Issue #1259, opened 2018-09-21
Navigation links not materialized for response with minimal metadata and expand-sub-select clause


biaol-odata:
*Navigation links not materialized for response with minimal metadata and expand-sub-select clause. The expand token in context url of the response is interpreted incorrectly as partial sub-tree, which should be interpreted as entire sub-tree.*

### Assemblies affected

*Microsoft.OData.Core 7.5.1*

### Reproduce steps

*Using a query of $expand with sub-$select option, examine the context-url in the response, which contains selected list containing one expand token. When the reader tries to materialize the payload, the other navigation links are not materialized except for the one related to the $expand.*

### Expected result

*All navigation links are expected to be materialized.*

### Actual result

*All other navigation links are not materialized, due to SelectedPropertiesNode treats expand token as regular property, which cannot be resolved to navigation link correctly.*

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

biaol-odata:
This issue is a prerequisite for the $expand context url: OData/odata.net #1265

biaol-odata:
PR #1286 has been merged. Issue can be closed as fixed.

PR #1264, opened 2018-09-25
Set the selectionType to EntireSubtree when the selected list contain…

biaol-odata:
…s only expand tokens.

### Issues

*This pull request fixes issue #1259.*

### Description

*During response deserialization, mark the selected properties node as EntireSubtree when the selected list contains only expand tokens, even if the list is not empty.*

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/SelectedPropertiesNode.cs
+++ b/src/Microsoft.OData.Core/SelectedPropertiesNode.cs
@@ -41,6 +41,9 @@ namespace Microsoft.OData
         /// <summary>The type of the current node.</summary>
         private readonly SelectionType selectionType;
 
+        /// <summary>The Edm structured type of the current node.</summary>
+        private readonly IEdmStructuredType structuredType;
+
         /// <summary>The separator character used to separate property names in a path.</summary>
         private const char PathSeparator = '/';
 
@@ -68,10 +71,13 @@ namespace Microsoft.OData
         /// </summary>
         /// <param name="selectClause">The string representation of the selected property hierarchy using
         /// the same format as in the $select query option.</param>
-        internal SelectedPropertiesNode(string selectClause)
+        /// <param name="structuredType">The Edm structural type of the node.</param>
+        internal SelectedPropertiesNode(string selectClause, IEdmStructuredType structuredType)
             : this(SelectionType.PartialSubtree)
         {
             Debug.Assert(!string.IsNullOrEmpty(selectClause), "!string.IsNullOrEmpty(selectClause)");
+
+            this.structuredType = structuredType;
 
             // NOTE: The select clause is a list of paths separated by comma (',').
             string[] paths = selectClause.Split(ItemSeparator);
@@ -80,6 +86,14 @@ namespace Microsoft.OData
             {
                 string[] segments = t.Split(PathSeparator);
                 this.ParsePathSegment(segments, 0);
+            }
+
+            if (this.selectedProperties != null
+                && this.selectedProperties.Count > 0
+                && this.selectedProperties.All(IsValidExpandToken))
+            {
+                // select list is not empty and only contains expand tokens
+                this.selectionType = SelectionType.EntireSubtree;
             }
         }
 
@@ -115,11 +129,12 @@ namespace Microsoft.OData
         }
 
         /// <summary>
-        /// Creates a node from the given raw $select query option value.
+        /// Creates a node from the given raw $select query option value and structural type information.
         /// </summary>
         /// <param name="selectQueryOption">The value of the $select query option.</param>
+        /// <param name="structuredType">The structural type of this node.</param>
         /// <returns>A tree representation of the selected properties specified in the query option.</returns>
-        internal static SelectedPropertiesNode Create(string selectQueryOption)
+        internal static SelectedPropertiesNode Create(string selectQueryOption, IEdmStructuredType structuredType)
         {
             if (selectQueryOption == null)
             {
@@ -134,7 +149,17 @@ namespace Microsoft.OData
                 return Empty;
             }
 
-            return new SelectedPropertiesNode(selectQueryOption);
+            return new SelectedPropertiesNode(selectQueryOption, structuredType);
+        }
+
+        /// <summary>
+        /// Creates a node from the given raw $select query option value without structural type information.
+        /// </summary>
+        /// <param name="selectQueryOption">The value of the $select query option.</param>
+        /// <returns>A tree representation of the selected properties specified in the query option.</returns>
+        internal static SelectedPropertiesNode Create(string selectQueryOption)
+        {
+            return Create(selectQueryOption, /*structuredType*/ null);
         }
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


biaol-odata:
Passed build: https://identitydivision.visualstudio.com/OData/_build/results?buildId=197850

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -81,6 +87,14 @@ internal SelectedPropertiesNode(string selectClause)
                 string[] segments = t.Split(PathSeparator);
                 this.ParsePathSegment(segments, 0);
             }
+
+            if (this.selectedProperties != null
+                && this.selectedProperties.Count > 0
+                && this.selectedProperties.All(IsValidExpandToken))
+            {
+                // select list is not empty and only contains expand tokens
+                this.selectionType = SelectionType.EntireSubtree;
```
xuzhg: >this.selectionType = SelectionType.EntireSubtree; [](start = 16, length = 49)

Line 76: call private constructor to set "selectionType" as PartialSubtree, meantime, this constructor reset it EntireSubtree if meet some condition. 

So, in this case, why don't you remove the private constructor and move the content in this constructor?

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -68,11 +71,14 @@ internal sealed class SelectedPropertiesNode
...
         /// <param name="selectClause">The string representation of the selected property hierarchy using
         /// the same format as in the $select query option.</param>
-        internal SelectedPropertiesNode(string selectClause)
+        /// <param name="structuredType">The Edm structural type of the node.</param>
+        internal SelectedPropertiesNode(string selectClause, IEdmStructuredType structuredType)
             : this(SelectionType.PartialSubtree)
         {
             Debug.Assert(!string.IsNullOrEmpty(selectClause), "!string.IsNullOrEmpty(selectClause)");
 
+            this.structuredType = structuredType;
```
xuzhg: >structuredType [](start = 34, length = 14)

need to Debug.Assert for the type?

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
xuzhg: If that's not the case, i think the line 90-97 should move out of the constructor into the static Create(...) method?

---
In reply to: [221684657](https://github.com/OData/odata.net/pull/1264#discussion_r221684657) [](ancestors = 221684657)

test/EndToEndTests/Tests/Client/Build.Desktop/ContainmentTest/ContainmentTest.cs
```diff
@@ -1197,9 +1198,9 @@ public void InvokeActionFromODataClientWhichReturnsContainedEntity()
         [TestMethod]
         public void CreateContainedEntityFromODataClientUsingAddRelatedObject()
         {
-
```
xuzhg: revert the unused changes.

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
mikepizzo: +1

---
In reply to: [221684657](https://github.com/OData/odata.net/pull/1264#discussion_r221684657) [](ancestors = 221684657)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -81,6 +87,14 @@ internal SelectedPropertiesNode(string selectClause)
                 string[] segments = t.Split(PathSeparator);
                 this.ParsePathSegment(segments, 0);
             }
+
+            if (this.selectedProperties != null
+                && this.selectedProperties.Count > 0
```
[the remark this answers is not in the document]
mikepizzo: Yes, Enumerable.All() will return true for an empty sequence, in which case selectionType should be EntireSubtree. Right?

---
In reply to: [222417081](https://github.com/OData/odata.net/pull/1264#discussion_r222417081) [](ancestors = 222417081)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -469,6 +503,44 @@ private static IEnumerable<string> GetPossibleMatchesForSelectedOperation(IEdmOp
...
+                    !(this.structuredType.FindProperty(item.Substring(0, idxRP)) is IEdmNavigationProperty)
+                    )
+                )
+            {
+                // selected item does not contain '(',
+                // or is not navigation property (for example when an Action is selected).
+                return false;
+            }
+
+            // Check #2: paretheses must be balanced.
```
mikepizzo: >// Check #2: paretheses must be balanced. [](start = 12, length = 41)

I'm not sure this check is necessary; I don't think we have any valid entries in the selectlist that allow unbalanced parens.  If we do need this check, then we should probably also verify that the last character was a closing paren?

biaol-odata: Agree. That is just a safeguard/double-check that is meaningful in the context of this function, but can be omitted for sake of optimization.

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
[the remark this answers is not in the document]
biaol-odata: My previous response might be a bit off. I remember I started with without the Count>0 check and ran into some error, but I don't remember exact what the case is. This area needs to be revisited for issue #1268, which is WIP.

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -469,6 +503,44 @@ private static IEnumerable<string> GetPossibleMatchesForSelectedOperation(IEdmOp
             yield return qualifiedContainerName + operationNameWithParameters;
         }
 
+        private bool IsValidExpandToken(string item)
+        {
+            // Check #1: Expand token must contain '(', must be related to navigation property.
+            int idxRP = item.IndexOf('(');
+            if (idxRP == -1
+                || (this.structuredType != null &&
+                    !(this.structuredType.FindProperty(item.Substring(0, idxRP)) is IEdmNavigationProperty)
```
mikepizzo: >!(this.structuredType.FindProperty(item.Substring(0, idxRP)) [](start = 20, length = 60)

As discussed, this does not allow for "dynamic" navigation properties.  It's probably sufficient to say that, if it contains an open paren, then it represents an expand (and must terminate in a balanced closing paren.)

biaol-odata: The check for "is IEdmNavigationProperty" was put in to guard against Action()/Function() alike (see comment in L517 below), which has parens. Looks like we need to further resolve potential ambiguity between navProp() and Action()/Function() alike.

biaol-odata: Per our discussion offline today, in order to avoid introducing other ambiguities related to Function/Action resolution, we don't support dynamic navigation property for now.

biaol-odata:
This PR has been superseded by the overall PR https://github.com/OData/odata.net/pull/1286, which contains the changes here.
Cancelled.

Issue #855, opened 2017-06-27
Support OptionalParameter as an external annotation


mikepizzo:
*PR #828 added support for optional parameters by looking for the "OptonalParameter" annotation term.  However, the code for the OnParameterElement method in CsdlDocumentParser only looks for this term applied as an in-line annotation. Although this is where it will be most common, for completeness we should also support it as an external annotation.*

### Assemblies affected

*OData.NET V7.x and greater.*

### Reproduce steps

*Apply the OptionalParameter term to a parameter as an external annotation.*
~~~xml
<Annotations Target="MyFunction/MyOptionalParameter"/>
    <Annotation Term="Org.OData.Core.V1.OptionalParameter" />
</Annotations>
~~~

### Expected result

*When read in, MyOptionalParameter should be read as an EdmOptionalParameter*

### Actual result

*When read in, MyOptionalParameter is read as a (required) EdmOperationParameter*

### Additional detail

*Probably a low priority, since I would expect that, in the bulk of cases, OptionalParameter will be applied as an inline annotation.*

mikepizzo:
Will re-open if we get requests/scenarios for this functionality.

xuzhg:
Fixed and Merged. Thanks.

PR #1267, opened 2018-09-29
Issue #855: enable out of line optional parameter works

xuzhg:
### Issues

* Issue #855: enable out of line optional parameter works

### Description

If we create a non-optional (general) parameter, and apply the “OptionalParameter” term on it, we also can get the similar result:

# 1
```C#
var requiredParam = new EdmOperationParameter(function, "requiredParam", stringTypeReference);

IEdmRecordExpression optionalParameterRecord = new EdmRecordExpression(.., new EdmPropertyConstructor("DefaultValue", new EdmStringConstant("Smith")));

EdmVocabularyAnnotation annotation = new EdmVocabularyAnnotation(requiredParam, CoreVocabularyModel.OptionalParameterTerm, optionalParameterRecord);
annotation.SetSerializationLocation(model, EdmVocabularyAnnotationSerializationLocation.OutOfLine);
model.SetVocabularyAnnotation(annotation);
```
The above will get the result same as the following codes:

# 2.
```C#
var requiredParam = new EdmOptionalParameter(function, "requiredParam", stringTypeReference, "Smith");
```
So, #1 and #2 can get the same result, and #1 is more generic process in the ODL (#2 will do some process to add the annotation at backend), I don’t think we should introduce the **IEdmOptionalParameter**.

However, this PR is following up the design with **IEdmOptionalParameter** and to make the out of line optional parameter work.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsOperation.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsOperation.cs
@@ -7,8 +7,10 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using Microsoft.OData.Edm.Csdl.Parsing.Ast;
 using Microsoft.OData.Edm.Vocabularies;
+using Microsoft.OData.Edm.Vocabularies.V1;
 
 namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
 {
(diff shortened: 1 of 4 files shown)


src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsOperation.cs
```diff
@@ -132,7 +134,114 @@ private IEnumerable<IEdmOperationParameter> ComputeParameters()
...
+            string fullName = ParameterizedTargetName(Namespace + "." + Name, parameters);
+
+            List<IEdmOperationParameter> newParameters = new List<IEdmOperationParameter>();
+            foreach (var parameter in parameters)
+            {
+                // for example: NS.TestFunction(Edm.String, Edm.String, Edm.String)/OptionalParameter
+                string fullTargetName = fullName + "/" + parameter.Name;
+                string defaultValue;
+
+                if (TryGetOptionalParameterAnnotation(fullTargetName, out defaultValue))
```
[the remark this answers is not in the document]
xuzhg: Which "IsOptional" property do you mean?  
If you mean the “IsOptional" property on "CsdlOperationParameter", I don't think we should set it (even though we can't set it, except we create a new instance of CsdlOperationParameter.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsOperation.cs
```diff
@@ -132,7 +134,114 @@ private IEnumerable<IEdmOperationParameter> ComputeParameters()
                 }
             }
 
-            return parameters;
```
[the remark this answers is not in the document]
xuzhg: "IsOptional" is set from InLine annotation. "IsOptional" is useless in the final "CsdlSemanticsOperationParameter". 
Please pay attention to the difference between "CsdlOperationParameter" vs "CsdlSemanticsOperationParameter" or "CsdlSemanticsOptionalParameter".

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsOperation.cs
```diff
@@ -132,7 +134,114 @@ private IEnumerable<IEdmOperationParameter> ComputeParameters()
...
+
+        private bool TryGetOptionalParameterAnnotation(string fullTargetName, out string defaultValue)
+        {
+            defaultValue = null;
+            bool isOptional = false;
+
+            List<CsdlSemanticsAnnotations> annotations;
+            if (Model.OutOfLineAnnotations.TryGetValue(fullTargetName, out annotations))
+            {
+                foreach (var annotation in annotations)
```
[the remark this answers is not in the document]
xuzhg: It depends. maybe 2 ~ 3 are the normal scenario. 

The whole "ComputeParameters()" method only be called once. Once the parameters are created from the CSDL, it's cached. So, the performance here is ok.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsOperation.cs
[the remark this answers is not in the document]
mikepizzo: I think what Kanish is suggesting is that, rather than create the (non) optional parameter in line 133, and then loop through the entire collection of parameters calling TryGetOptionalParameterAnnotation on each, we could nest the TryGetOptionalParameterAnnotation in the else on line 131 so that, if the IsOptionalFlag is not set, we look for the external annotation *prior* to creating the CsdlSemanticsOperationParameter, and then just create right one to start with.  i.e., nest lines 149-160 within the else statement on line 131 (except replace newParameters with parameters). 

---
In reply to: [222054166](https://github.com/OData/odata.net/pull/1267#discussion_r222054166) [](ancestors = 222054166)

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsOperation.cs
```diff
@@ -132,7 +134,114 @@ private IEnumerable<IEdmOperationParameter> ComputeParameters()
...
+            return newParameters;
+        }
+
+        private bool TryGetOptionalParameterAnnotation(string fullTargetName, out string defaultValue)
+        {
+            defaultValue = null;
+            bool isOptional = false;
+
+            List<CsdlSemanticsAnnotations> annotations;
+            if (Model.OutOfLineAnnotations.TryGetValue(fullTargetName, out annotations))
```
mikepizzo: >fullTargetName [](start = 55, length = 14)

Note that OData 4.0 applies annotations to all overloads of a function or action. It isn't until recently, in [ODATA-1178](https://issues.oasis-open.org/browse/ODATA-1178?jql=project%20%3D%20ODATA%20AND%20text%20~%20%22overload%22), that we added support for externally annotating different overloads, but we still support annotating the non-overloaded version (which, for this case, I think would be the common scenario).  SO, we should probably first check the fullTargetName and, if that doesn't match, check just the function name.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsOperation.cs
```diff
@@ -132,7 +134,114 @@ private IEnumerable<IEdmOperationParameter> ComputeParameters()
...
+                        CsdlRecordExpression optionalValueExpression = optionalParameterAnnotation.Expression as CsdlRecordExpression;
+                        if (optionalValueExpression != null)
+                        {
+                            foreach (CsdlPropertyValue property in optionalValueExpression.PropertyValues)
+                            {
+                                CsdlConstantExpression propertyValue = property.Expression as CsdlConstantExpression;
+                                if (propertyValue != null)
+                                {
+                                    if (property.Property == CsdlConstants.Attribute_DefaultValue)
+                                    {
```
mikepizzo: should probably first check to see if property.Property == CsdlConstants.Attribute_DefaultValue, and then try to cast property.Expression to CsdlConstantExpression (since the comparison is cheaper than the cast).  i.e., move 192-193 before 189.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
```diff
@@ -711,6 +711,105 @@ private static IEdmModel GetEdmModel(string types = "", string properties = "")
...
+                 "</Function>" +
+                 "<EntityContainer Name=\"Default\">" +
+                   "<FunctionImport Name=\"TestFunction\" Function=\"test.TestFunction\" />" +
+                 "</EntityContainer>" +
+                 "<Annotations Target=\"NS.TestFunction(Edm.String, Edm.String, Edm.String)/optionalParam\" >" +
+                   "<Annotation Term=\"Org.OData.Core.V1.OptionalParameter\" >" +
+                     "<Record />" +
+                  "</Annotation> " +
+                 "</Annotations>" +
+                 "<Annotations Target=\"NS.TestFunction(Edm.String, Edm.String, Edm.String)/optionalParamWithDefault\" >" +
```
mikepizzo: >\"NS.TestFunction(Edm.String, Edm.String, Edm.String) [](start = 38, length = 53)

should add test cases for NS.TestFunction/optionalParam and NS.TestFunction/optionalParamWithDefault.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsOperation.cs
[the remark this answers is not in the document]
xuzhg: @mikepizzo @KanishManuja-MS Ok. I got it. However, I need the `parameters` created above (foreach) to generate the "fullName" at line 143. See Line 143. Otherwise, it's hard to create the `fullname` because it's hard to get the each parameter type name form the `CsdlOperationParameter`

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsOperation.cs
xuzhg: Resolved.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsOperation.cs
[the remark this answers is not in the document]
mikepizzo: I see.

---
In reply to: [222392493](https://github.com/OData/odata.net/pull/1267#discussion_r222392493) [](ancestors = 222392493)

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
xuzhg: Added. Thanks.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsOperation.cs
```diff
@@ -132,7 +134,125 @@ private IEnumerable<IEdmOperationParameter> ComputeParameters()
...
+            defaultValue = null;
+            bool isOptional = false;
+
+            List<CsdlSemanticsAnnotations> annotations;
+
+            // OData 4.0 applies annotations to all overloads of a function or action.
+            // We still need to support annotating the non-overloaded version.
+            // So, we should probably first check the fullTargetName and, if that doesn't match, check just the function or action name.
+            bool found = Model.OutOfLineAnnotations.TryGetValue(fullTargetName, out annotations) ? true :
+                Model.OutOfLineAnnotations.TryGetValue(targetName, out annotations);
```
mikepizzo: What if the function has some annotations against the fullTarget (not including the OptionalParameter annotation) and some against the non-overloaded name (including OptionalParameter the optional parameter annotation).  It appears if there are *any* annotations against the fully qualified name then we don't even look at the non-overloaded version.  We should look to see if the OptionalParameter annotation is in the fullTarget and, if not, see if it is in targetName (and add a test for this case).

xuzhg: Synced with @Michael Pizzo, the codes here is to search the out of line annotation using the "FullTargetName" and "TargetName" (FullTargetName is the qualified operation name with the parameter type list, while TargetName is the qualified operation name). It's ok here.

---
In reply to: [223550455](https://github.com/OData/odata.net/pull/1267#discussion_r223550455) [](ancestors = 223550455)

Issue #1275, opened 2018-10-03
Support $apply in $expand


xuzhg:
Copy from the https://stackoverflow.microsoft.com/questions/117909/apply-aggregations-support-for-expand

I’m looking specifically for .NET Odata parsing support for aggregations for entities exposed inside the $expand system query option.
Based on the spec 3.17 found here: http://docs.oasis-open.org/odata/odata-data-aggregation-ext/v4.0/cs02/odata-data-aggregation-ext-v4.0-cs02.html#_Toc435016591
Example given from doc: Products?$expand=Sales($apply=aggregate(Amount with sum as Total))
My scenario requires a group by with count on the expanded entities.
There should be support based on the protocol. However, I was unable to find actual support for it in the latest .NET release, seems like today there is only support for filter, orderby, top, skip, count, levels, search, select, compute, and expand.

xuzhg:
Merged.

PR #1277, opened 2018-10-05
Fix issue #1275: Support nested  $apply in $expand

xuzhg:
### Issues

*This pull request fixes issue #1275.*

### Description

* Enable nested $apply in $expand, for example: 
    `$expand=Orders($apply=aggregate(Weight with sum as Total))`

* Fix issues for $compute in $expand, for example
   `$expand=Orders($compute=Weight mul 8 as Ratio)`

* Add tests to cover nested $compute and nested $apply in $expand

- [x] *Test cases added*

### Additional work necessary


sw47:
Thanks for supporting this! When do you plan for this to be merged and released for be consumed?

Issue #1271, opened 2018-10-01
Client side dependency injection support via IODataClientFactory, similar to IHttpClientFactory


chuanboz:
ASP.NET Core 2.1 includes a new IHttpClientFactory service that makes it easier to configure and consume instances of HttpClient in apps, see [Initiate HTTP Requests](https://docs.microsoft.com/en-us/aspnet/core/fundamentals/http-requests?view=aspnetcore-2.1)

The request here is introduce something similar to ODataClient, with the IODataClientFactory and corresponding IODataClientHandler, and the builder pattern IODataClientBuilder to use asp.net core dependency injection to construct the factory independently.
### Assemblies affected
Microsoft.OData.Client

To avoid impact existing consumer of the client assembly, might consider introduce a new Microsoft.Extensions.OData package for this new functionality, this is similar to IHttpClientFactory which is in Microsoft.Extensions.Http package.

### Reproduce steps
1. Generate an OData client with visual studio.
2. View the generated client code.

### Expected result
The generated class support extensibility and plugin model similar to IHttpClientFactory. 

Something like this:
```C#
    /// <summary>
    /// Factory to create OData client containers to an ODATA service, similar to IHttpClientFactory.
    /// Type T will be the concreate class generated from visual studio in standard usage.
    /// </summary>
    /// <remarks>
    public interface IODataClientFactory<T> where T : DataServiceContext
    {
        /// <summary>
        /// create the client container to an OData service, specifying a logical named the client
        /// </summary>
        /// <param name="name">the logic name of the client to use, including both HttpClient and ODataClient.</param>
        /// <returns>the container client with the handling logic injected from dependency injection container.</returns>
        T CreateClient(string name);
    }
```
and also introduce the OData handler similar to HttpMessageHandler, that could hook into the DataServiceContext to hook up with built-in event to impact OData Client behavior. Some example of such handler, like retry, metric, exception handling, versioning, sending request via HttpClient, etc.
```C#
    /// <summary>
    /// A single handler that can alter the behavior of odataclient
    /// </summary>
    public interface IODataClientHandler
    {
        /// <summary>
        /// Called within IODataClientFactory{T}.CreateClient(string) after the DataServiceContext created.
        /// If multiple handlers are registered, then they are called in reverse order of registration.
        /// </summary>
        /// <param name="context">the default client container created, implementation could hook up with the event to change its behavior.</param>
        void OnClientCreated(DataServiceContext context);
    }
```

[...]

PR #1279, opened 2018-10-07
Fix the issue #1271 to add IODataClientFactory support with Microsoft.Extensions.ODataClient packages

chuanboz:
### Issues

This pull request fixes issue #1271 to add IODataClientFactory DI support

### Description

followed similar pattern as IHttpClientFactory and added following change:
1. IODataClientFactory, IODataClientBuilder, IODataClientHandler and IODataClientActivator
2. add link with IHttpClientFactory and implemented request/response message.
3. Added Properties to DataServiceClientConfigurations.
4. Introduced 2 packages Microsoft.Extensions.ODataClient packages and Microsoft.Extensions.ODataClient.Abstractions packages
5. Above 2 packages only compile with .net standard 2.0 because it's dependency IHttpClientFactory only support .net standard 2.0 and above.

- [ Y ] *Test cases added and passed*
have some issue with one-click build but passed in Visual Studio 2017.

### Additional work necessary
Docs needed, and might need more scenario and performance testing.
Might also need to consider adding the handler filters to control the sequence of the handler execution.


chuanboz:
@NeilScalesTransformUK , thanks for your help to review my change, I've addressed your comment in latest PR, please take a look and see if you have additional feedback.

There is some issue in the continuous-integration/vsts checks related to CI workflow and I'm working with @xuzhg to solve that before this PR could pass the check.

src/Microsoft.Extensions.ODataClient.Abstractions/IODataClientHandler.cs
```diff
@@ -0,0 +1,26 @@
+﻿//------------------------------------------------------------------
+// <copyright file="IODataClientHandler.cs" company="Microsoft Corporation">
+// Copyright © Microsoft Corporation. All rights reserved.
+// </copyright>
+//------------------------------------------------------------------
+//---------------------------------------------------------------------
```
NeilScalesTransformUK: Duplicate header?

src/Microsoft.Extensions.ODataClient.Abstractions/Microsoft.Extensions.ODataClient.Abstractions.csproj
```diff
@@ -0,0 +1,29 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <AssemblyName>Microsoft.Extensions.ODataClient.Abstraction</AssemblyName>
```
rifuller: Microsoft.Extensions.ODataClient.Abstraction**s**  (plural)?

src/Microsoft.Extensions.ODataClient.Abstractions/ClientCreatedArgs.cs
```diff
@@ -0,0 +1,38 @@
...
+    public sealed class ClientCreatedArgs 
+    {
+        /// <summary>
+        /// Create a new instance of <see cref="ClientCreatedArgs"/>
+        /// </summary>
+        /// <param name="name">the logical name of the client.</param>
+        /// <param name="container">the instance of data service context for client communication.</param>
+        public ClientCreatedArgs(string name, DataServiceContext container)
+        {
+            Name = name;
```
rifuller: this.

src/Microsoft.Extensions.ODataClient.Abstractions/IODataClientActivator.cs
```diff
@@ -0,0 +1,24 @@
...
+//---------------------------------------------------------------------
+
+namespace Microsoft.Extensions.ODataClient
+{
+    using Microsoft.OData.Client;
+    using System;
+    using System.Net.Http;
+
+    /// <summary>
+    /// An activator that's responsbile soly to create the raw <see cref="DataServiceContext"/> instances.
```
rifuller: typo: "solely"

rifuller: also "responsible"

src/Microsoft.Extensions.ODataClient.Abstractions/ODataClientFactoryExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+        /// Each call to <see cref="CreateClient(Uri, string)"/> is guaranteed to return a new <see cref="HttpClient"/>
+        /// instance. Callers may cache the returned <see cref="HttpClient"/> instance indefinitely or surround
+        /// its use in a <langword>using</langword> block to dispose it when desired.
+        /// </para>
+        /// <para>
+        /// Callers are also free to mutate the returned <see cref="DataServiceContext"/> instance's public properties
+        /// as desired.
+        /// </para>
+        /// </remarks>
+        public static T CreateClient<T>(this IODataClientFactory factory, Uri serviceRoot) where T : DataServiceContext
```
rifuller: Why have this as an extension method instead of being on the interface?

src/Microsoft.Extensions.ODataClient.Abstractions/ODataClientOptions.cs
```diff
@@ -0,0 +1,26 @@
...
+{
+    /// <summary>
+    /// An options class for configuring the default IODataClientFactory.
+    /// </summary>
+    public class ODataClientOptions
+    {
+        /// <summary>
+        /// The default logic name if no name is specified for the client when creating client from factory.
+        /// </summary>
+        public const string DefaultName = "";
```
rifuller: string.Empty

src/Microsoft.Extensions.ODataClient.Abstractions/ODataClientFactoryExtensions.cs
rifuller: Both are in the same DLL so the calling pattern for consumers would be identical but are there any pros/cons to this approach?

src/Microsoft.Extensions.ODataClient.Abstractions/ClientCreatedArgs.cs
```diff
@@ -0,0 +1,38 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ClientCreatedArgs.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.Extensions.ODataClient
```
rifuller: Should these be in the .Abstractions namespace?

rifuller:
Added some general comments.  I'm not at all familiar with the structure of this lib or the DI extensions.  I can spend more time reviewing with more context or I can defer to someone else with more experience.

src/Microsoft.Extensions.ODataClient.Abstractions/ODataClientFactoryExtensions.cs
chuanboz: This would make the interface much simpler, and if others implement this interface, it does not needs to repeat the same implementation for this.

src/Microsoft.Extensions.ODataClient.Abstractions/ODataClientOptions.cs
chuanboz: I could not use string.Empty because that's static and not const, and for the DefaultName to show up in default parameter it has to be constant and could not be static, so it's on purpose here to use "".

src/Microsoft.Extensions.ODataClient.Abstractions/ClientCreatedArgs.cs
chuanboz: No, that's the namespace conventions all other Microsoft.Extensions.*.Absractions use, and I would agree on that because that make user experience simpler also.

xuzhg:
@chuanboz Can we close this PR, right?

chuanboz:
close it, all the changes (except minor change of adding Properties that's not needed yet) are checked in to the OData\Extensions repo

PR #1282, opened 2018-10-08
fix the poli-check problem for country/region

xuzhg:
### Issues

* Fix the poli-check problem, that's from the PR: https://github.com/OData/odata.net/pull/1277

### Description

* Change the string "Country" to "CountryRegion" to meet the poli-check.

- [ ] *Test cases added*

### Additional work necessary


xuzhg:
Pass the policheck task:
image

KanishManuja-MS:
Which PR caused this problem? Can you describe the issue in detail - 
String containing 'Country' causes poli-check to fail.
Description should be - Changed the string to use CountryRegions instead.

xuzhg:
@KanishManuja-MS I updated the description.

Issue #1287, opened 2018-10-10
Edm Library: Add support for Validation Vocabulary


xuzhg:
Add support for the validation vocabulary, see : https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Validation.V1.md.

It's a sub task of union type supporting. In particular, the `DerivedTypeConstraint` term allows us to define a restricted set of unioned types.

PR #1288, opened 2018-10-10
Add Validation vocabulary annotation

xuzhg:
### Issues

*This pull request fixes issue https://github.com/OData/odata.net/issues/1287 *

### Description

* Add Validation vocabulary annotation definition
* Add validation vocabulary annotation model and insert into model reference.
* Owning validation vocabulary refers some new elements defined in Core vocabulary, for example: "QualifiedTypeName" type definition, so update the Core vocabulary
* Add test cases to test the added validation vocabulary
* "Run custom tool" on "ODataT4CodeGenerator.tt" to refresh the latest changes in "ttinclude" file.

- [x] *Test cases added*

### Additional work necessary


src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -140,6 +142,10 @@ public static class Configuration
 
 	// The target language of the generated client code. The value must be set to "CSharp" or "VB".
 	public const string TargetLanguage = "OutputLanguage";
+
```
xuzhg: TempFilePath related codes are not my changes. It's introduced from running "The Custom Tool" for the tt file.

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -60,6 +60,11 @@ protected EdmModelBase(IEnumerable<IEdmModel> referencedModels, IEdmDirectValueA
                 this.referencedEdmModels.Add(AuthorizationVocabularyModel.Instance);
             }
 
+            if (ValidationVocabularyModel.Instance != null)
+            {
+                this.referencedEdmModels.Add(ValidationVocabularyModel.Instance);
+            }
+
```
mikepizzo: I thought we weren't going to add ValidationVocabulary by default?  I would also remove AuthorizationVocabulary (I don't think anyone is using it yet).  I would also remove AlternateKeys if it hadn't already been around long enough that I was afraid people were already taking dependencies on it.

xuzhg: AuthorizationVocabulary is using by our openapi library, Microsoft.OpenApi.OData. :)

Adding the model by default is to support the Edm model reading from CSDL, otherwise, it can't work. So, we should find a solution to make all the vocabulary models custom-able.

xuzhg:
>     public static class CoreVocabularyModel

I 'd prefer not to add the new added terms, because customers can query the terms from the model directly.
Because customers can query the term from the model directly. Thoughts?

---
In reply to: [429139242](https://github.com/OData/odata.net/pull/1288#issuecomment-429139242) [](ancestors = 429139242)

---
Refers to: src/Microsoft.OData.Edm/Vocabularies/CoreVocabularyModel.cs:22 in 5c5d8ae. [](commit_id = 5c5d8aec6a664a95f2c78f9f450d5f8cc3907fa4, deletion_comment = False)

xuzhg:
@mikepizzo I will separate this PR into two parts. One is for the Core vocabulary changes, see PR: #1293. 
I will update this PR to refresh the remaining part for the validation vocabulary. Thanks.

mikepizzo:
>     public static class CoreVocabularyModel

That's fine with me. We can always add later, if we decide it's useful.

---
In reply to: [430003964](https://github.com/OData/odata.net/pull/1288#issuecomment-430003964) [](ancestors = 430003964,429139242)

---
Refers to: src/Microsoft.OData.Edm/Vocabularies/CoreVocabularyModel.cs:22 in 5c5d8ae. [](commit_id = 5c5d8aec6a664a95f2c78f9f450d5f8cc3907fa4, deletion_comment = False)

mikepizzo:
Either way is fine with me.

---
In reply to: [430807173](https://github.com/OData/odata.net/pull/1288#issuecomment-430807173) [](ancestors = 430807173)

PR #1293, opened 2018-10-17
Update the Core vocabulary xml

xuzhg:
### Description

Continue to update the Core vocabulary model.

1. Update the Core vocabulary xml
2. Update the existing Core vocabulary tests
3. Add three new test cases to cover the type and term definition in Core vocabulary

- [x] *Test cases added*

### Additional work necessary


xuzhg:
@mikepizzo I want to use this PR to update the remaining Core vocabulary model. With this change, now, it's align with the https://github.com/oasis-tcs/odata-vocabularies/blob/master/vocabularies/Org.OData.Core.V1.xml

PR #1295, opened 2018-10-18
Uri parsing on type cast on property allowed in validation vocabulary

xuzhg:
### Issues

*This pull request is part of Union type supporting.*

### Description

* To parse the type cast on a property (structural or navigation property), the allowed type case segment should within the Validation vocabulary [DerivedTypeConstraint](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Edm/Vocabularies/ValidationVocabularies.xml#L91-L94) annotation if it applied on such property, otherwise the Uri parse should throw exception.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Microsoft.OData.Core.txt
+++ b/src/Microsoft.OData.Core/Microsoft.OData.Core.txt
@@ -819,7 +819,7 @@ PathParser_EntityReferenceNotSupported=The request URI is not valid. $ref cannot
 PathParser_CannotUseValueOnCollection=$value cannot be applied to a collection.
 PathParser_TypeMustBeRelatedToSet=The type '{0}' does not inherit from and is not a base type of '{1}'. The type of '{2}' must be related to the Type of the EntitySet.
 PathParser_TypeCastOnlyAllowedAfterStructuralCollection=Type cast segment '{0}' after a collection which is not of entity or complex type is not allowed.
-PathParser_TypeCastOnlyAllowedInDerivedTypeConstraint=Type cast segment '{0}' on {1} '{2}' is not allowed from the Org.OData.Validation.V1.DerivedTypeConstraint annotation.
+PathParser_TypeCastOnlyAllowedInDerivedTypeConstraint=Type cast segment '{0}' on {1} '{2}' is not allowed due to an Org.OData.Validation.V1.DerivedTypeConstraint annotation.
 
 ODataResourceSet_MustNotContainBothNextPageLinkAndDeltaLink=A resource set may contain a next page link, a delta link or neither, but must not contain both.
 
--- a/src/Microsoft.OData.Core/Microsoft.OData.Core.txt
+++ b/src/Microsoft.OData.Core/Microsoft.OData.Core.txt
@@ -819,6 +819,7 @@ PathParser_EntityReferenceNotSupported=The request URI is not valid. $ref cannot
 PathParser_CannotUseValueOnCollection=$value cannot be applied to a collection.
 PathParser_TypeMustBeRelatedToSet=The type '{0}' does not inherit from and is not a base type of '{1}'. The type of '{2}' must be related to the Type of the EntitySet.
 PathParser_TypeCastOnlyAllowedAfterStructuralCollection=Type cast segment '{0}' after a collection which is not of entity or complex type is not allowed.
+PathParser_TypeCastOnlyAllowedInDerivedTypeConstraint=Type cast segment '{0}' on {1} '{2}' is not allowed from the Org.OData.Validation.V1.DerivedTypeConstraint annotation.
 
 ODataResourceSet_MustNotContainBothNextPageLinkAndDeltaLink=A resource set may contain a next page link, a delta link or neither, but must not contain both.
 
@@ -881,7 +882,6 @@ RequestUriProcessor_ResourceNotFound=Resource not found for the segment '{0}'.
 RequestUriProcessor_BatchedActionOnEntityCreatedInSameChangeset=Batched service action '{0}' cannot be invoked because it was bound to an entity created in the same changeset.
 RequestUriProcessor_Forbidden=Forbidden
 RequestUriProcessor_OperationSegmentBoundToANonEntityType=Found an operation bound to a non-entity type.
-
 
 ; Note: The below list of error messages are common to both the OData and the OData.Query project.
 General_InternalError=An internal error '{0}' occurred.
--- a/src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
@@ -14,6 +14,7 @@ using System.Text.RegularExpressions;
 using Microsoft.OData.Edm;
 using Microsoft.OData.Metadata;
 using ODataErrorStrings = Microsoft.OData.Strings;
+using Microsoft.OData.Edm.Vocabularies;
 
 namespace Microsoft.OData.UriParser
 {
@@ -1043,6 +1044,8 @@ namespace Microsoft.OData.UriParser
                 throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_BatchedActionOnEntityCreatedInSameChangeset(identifier));
             }
 
+            CheckOperationTypeCastSegmentRestriction(singleOperation);
+
             ODataPathSegment segment = new OperationSegment(singleOperation, resolvedParameters, targetset)
             {
                 Identifier = identifier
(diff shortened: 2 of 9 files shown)


src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -757,6 +757,30 @@ public static string GetLongDescriptionAnnotation(this IEdmModel model, IEdmVoca
...
+        public static IList<string> GetDerivedTypeConstraintAnnotation(this IEdmModel model, IEdmVocabularyAnnotatable target)
+        {
+            EdmUtil.CheckArgumentNull(model, "model");
+            EdmUtil.CheckArgumentNull(target, "target");
+
+            IEdmVocabularyAnnotation annotation = model.FindVocabularyAnnotations<IEdmVocabularyAnnotation>(target, ValidationVocabularyModel.DerivedTypeConstraintTerm).FirstOrDefault();
+            if (annotation != null)
+            {
+                IEdmCollectionExpression collectionExpression = annotation.Value as IEdmCollectionExpression;
+                if (collectionExpression != null && collectionExpression.Elements != null)
```
mikepizzo: >collectionExpression.Elements != null [](start = 52, length = 37)

If the DerivedTypeConstraint annotation is present, but there are no terms listed, do we interpret that as no derived types are defined (that it must be exactly the expected type)? Do we have a test for that case?

mikepizzo: >collectionExpression.Elements != null [](start = 52, length = 37)

It should be legal to have the annotation with an empty (or omitted?) list of derived types, in which case the type of the element must be exactly the expected type (no derived types allowed).  Can we add a test for this case?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1432,5 +1434,50 @@ private void CheckDollarEachSegmentRestrictions(int index)
                 }
             }
         }
+
+        private static void CheckTypeCastSegmentRestrictions(IEdmModel model, ODataPathSegment previous, IEdmType targetEdmType)
```
mikepizzo: >CheckTypeCastSegmentRestrictions [](start = 28, length = 32)

Note; this doesn't support scenarios where the immediately previous segment was not a property or a navigation property;  for example, Customers/1/ns.PreferredCustomer.  Probably okay just to note this in the code.

mikepizzo: If the underlying type of the TypeDefinition is Edm.Primitive, then the DerivedTypeConstraint can be used to restrict the set of primitive types (i.e., to string and int).  If the DerivedTypeConstraint is defined on both a TypeDefinition and a Property where that DerivedTypeConstraint is used, then only those types that are in both lists are valid.

---
In reply to: [227070272](https://github.com/OData/odata.net/pull/1295#discussion_r227070272) [](ancestors = 227070272)

xuzhg: @mikepizzo Thanks for the explanation. In this PR, i left some commented codes for type definition, because ODL doesn't support the Uri parse for the type definition, see issue at: https://github.com/OData/odata.net/issues/1326. So, i will un-comment type definition codes after i fix #1326. Is it ok?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1432,5 +1437,150 @@ private void CheckDollarEachSegmentRestrictions(int index)
...
+
+                return;
+            }
+        }
+
+        private void CheckOperationTypeCastSegmentRestriction(IEdmOperation operation)
+        {
+            Debug.Assert(operation != null);
+
+            if (this.parsedSegments == null || !this.parsedSegments.Any())
```
mikepizzo: >|| !this.parsedSegments.Any() [](start = 44, length = 29)

This would also be caught in the this.parsedSegments.LastOrDefault() check on line 1546 below, right?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1432,5 +1437,150 @@ private void CheckDollarEachSegmentRestrictions(int index)
                 }
             }
         }
+
+        private void CheckTypeCastSegmentRestriction(ODataPathSegment previous, IEdmType targetEdmType)
```
mikepizzo: >CheckTypeCastSegmentRestriction [](start = 21, length = 31)

We should also validate, if the previous segment is an operation, that the return type of the operation can be cast to the specified type.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
xuzhg: Yes. You are right.

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -757,6 +757,30 @@ public static string GetLongDescriptionAnnotation(this IEdmModel model, IEdmVoca
             return null;
         }
 
+        /// <summary>
+        /// Gets the collection of qualified type name for term Org.OData.Validation.V1.DerivedTypeConstraint from a target annotatable.
+        /// </summary>
+        /// <param name="model">The model referenced to.</param>
+        /// <param name="target">The target annotatable to find annotation.</param>
+        /// <returns>Null or a collection string of qualifed type name.</returns>
+        public static IList<string> GetDerivedTypeConstraints(this IEdmModel model, IEdmVocabularyAnnotatable target)
```
mikepizzo: >IList [](start = 22, length = 5)

Should this be an IEnumerable<string>, rather than an IList<string>?  We don't expect the caller to add/remove from the list.

xuzhg: Yes. # Fixed.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
xuzhg: # resolved.

test/FunctionalTests/Tests/DataOData/Tests/OData.Common.Tests/PublicApi/PublicApi.bsl
```diff
@@ -1932,6 +1932,11 @@ public sealed class Microsoft.OData.Edm.ExtensionMethods {
 	]
 	public static object[] GetAnnotationValues (Microsoft.OData.Edm.IEdmModel model, System.Collections.Generic.IEnumerable`1[[Microsoft.OData.Edm.Vocabularies.IEdmDirectValueAnnotationBinding]] annotations)
 
+	[
+	ExtensionAttribute(),
+	]
+	public static System.Collections.Generic.IList`1[[System.String]] GetDerivedTypeConstraints (Microsoft.OData.Edm.IEdmModel model, Microsoft.OData.Edm.Vocabularies.IEdmVocabularyAnnotatable target)
```
mikepizzo: >IList [](start = 42, length = 5)

This should be updated to IEnumerable<String> rather than IList<String> as per the updated code.

src/Microsoft.OData.Core/Microsoft.OData.Core.txt
```diff
@@ -819,6 +819,7 @@ PathParser_EntityReferenceNotSupported=The request URI is not valid. $ref cannot
 PathParser_CannotUseValueOnCollection=$value cannot be applied to a collection.
 PathParser_TypeMustBeRelatedToSet=The type '{0}' does not inherit from and is not a base type of '{1}'. The type of '{2}' must be related to the Type of the EntitySet.
 PathParser_TypeCastOnlyAllowedAfterStructuralCollection=Type cast segment '{0}' after a collection which is not of entity or complex type is not allowed.
+PathParser_TypeCastOnlyAllowedInDerivedTypeConstraint=Type cast segment '{0}' on {1} '{2}' is not allowed from the Org.OData.Validation.V1.DerivedTypeConstraint annotation.
```
mikepizzo: >from the [](start = 106, length = 8)

how about "due to an"

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
mikepizzo: Separate offline discussion on how to support annotating ReturnType.

---
In reply to: [236788996](https://github.com/OData/odata.net/pull/1295#discussion_r236788996) [](ancestors = 236788996)

src/Microsoft.OData.Core/Microsoft.OData.Core.txt
xuzhg: Changed. Thanks.

Issue #1260, opened 2018-09-22
Nulls being allowed in complex collections even when Nullable=false


brjohnstmsft:
When de-serializing a payload with a JSON array containing null, where the array is being assigned to a complex collection property with `Nullable=false` in the EDM model, deserialization succeeds even though nulls are not allowed in this context. This is a behavioral change since ODataLib 6.19, which would throw `ODataException` in this case.

When used in conjunction with Web API OData, the result is that invalid payloads no longer fail model validation as they should.

### Assemblies affected

Microsoft.OData.Core 7.5.1

### Reproduce steps

Run the tests in [this project](https://github.com/brjohnstmsft/odata-repros/tree/master/NullInComplexCollection).

### Expected result

With Web API OData 5.9/ODataLib 6.19, all tests pass. Model validation of a complex collection containing null fails with this error message:

```
A null value was found with the expected type 'search.address[Nullable=False]'. The expected type 'search.address[Nullable=False]' does not allow null values.
```

### Actual result

The test `NullsInDynamicComplexCollectionsAreDisallowed` fails because the POST operation returns 200 instead of 400, indicating that model validation succeeded.

### Additional detail

Looking at the code, this appears to be another consequence of unifying complex values and entities. The code path through the JSON de-serializer appears to be completely different. Whereas before, this situation was caught here:

https://github.com/OData/odata.net/blob/c48822976a7c8a302bbcc3dfe526cbbf47c6c25c/src/Microsoft.OData.Core/Json/ODataJsonReaderCoreUtils.cs#L112

now this code doesn't even run.

Things appear to start going wrong here:

https://github.com/OData/odata.net/blob/8e47d6e1efe209b472865645b08334f53b55c3f8/src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs#L1512

Perhaps this could be the right place to perform this validation now, but I'm not sure:

https://github.com/OData/odata.net/blob/8e47d6e1efe209b472865645b08334f53b55c3f8/src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs#L1713

brjohnstmsft:
I've implemented a workaround for Web API OData here: https://github.com/brjohnstmsft/odata-repros/blob/9ca287023d0b0308b80770f20f2bab7b98f1ebd0/NullInComplexCollection/Repro/Startup.cs#L42

PR #1322, opened 2018-11-09
Throw error when null value passed for collection of non-nullable complex type

KanishManuja-MS:
### Issues

*This pull request fixes issue #1260 *

### Description
Previously, OData 7.x would not throw an exception during serializing a null value for a collection of non-nullable complex type. Now, OData rightly throws an exception.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
@@ -911,7 +911,7 @@ namespace Microsoft.OData.JsonLight
                 ? null
                 : this.jsonLightResourceDeserializer.ContextUriParseResult.SelectQueryOption;
 
-            SelectedPropertiesNode selectedProperties = SelectedPropertiesNode.Create(selectQueryOption, this.CurrentResourceTypeReference.AsStructured().StructuredDefinition(), this.jsonLightInputContext.Model);
+            SelectedPropertiesNode selectedProperties = SelectedPropertiesNode.Create(selectQueryOption, (this.CurrentResourceTypeReference !=null) ? this.CurrentResourceTypeReference.AsStructured().StructuredDefinition() : null, this.jsonLightInputContext.Model);
 
             if (this.ReadingResourceSet)
             {
@@ -1858,7 +1858,7 @@ namespace Microsoft.OData.JsonLight
                 else
                 {
                     // null resource
-                    if (this.CurrentResourceTypeReference.IsComplex() || this.CurrentResourceTypeReference.IsUntyped())
+                    if (resourceTypeReference.IsComplex() || resourceTypeReference.IsUntyped())
                     {
                         this.jsonLightResourceDeserializer.MessageReaderSettings.Validator.ValidateNullValue(this.CurrentResourceTypeReference, true, "", null);
                     }
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
@@ -911,7 +911,7 @@ namespace Microsoft.OData.JsonLight
                 ? null
                 : this.jsonLightResourceDeserializer.ContextUriParseResult.SelectQueryOption;
 
-            SelectedPropertiesNode selectedProperties = SelectedPropertiesNode.Create(selectQueryOption, this.CurrentResourceType, this.jsonLightInputContext.Model);
+            SelectedPropertiesNode selectedProperties = SelectedPropertiesNode.Create(selectQueryOption, this.CurrentResourceTypeReference.AsStructured().StructuredDefinition(), this.jsonLightInputContext.Model);
 
             if (this.ReadingResourceSet)
             {
@@ -1698,7 +1698,7 @@ namespace Microsoft.OData.JsonLight
             }
 
             this.EnterScope(new JsonLightResourceSetScope(resourceSet, this.CurrentNavigationSource,
-                this.CurrentScope.ResourceType, selectedProperties, this.CurrentScope.ODataUri, /*isDelta*/ false));
+                this.CurrentScope.ResourceTypeReference, selectedProperties, this.CurrentScope.ODataUri, /*isDelta*/ false));
         }
 
         /// <summary>
@@ -1776,7 +1776,7 @@ namespace Microsoft.OData.JsonLight
                 // Expanded null resource
                 // The expected type and expected navigation source for an expanded resource are the same as for the nested resource info around it.
                 this.EnterScope(new JsonLightResourceScope(ODataReaderState.ResourceStart, /*resource*/ null,
-                    this.CurrentNavigationSource, this.CurrentResourceType, /*propertyAndAnnotationCollector*/null,
+                    this.CurrentNavigationSource, this.CurrentResourceTypeReference, /*propertyAndAnnotationCollector*/null,
                     /*projectedProperties*/null, this.CurrentScope.ODataUri));
             }
             else
@@ -1834,7 +1834,7 @@ namespace Microsoft.OData.JsonLight
             SelectedPropertiesNode selectedProperties)
         {
             IEdmNavigationSource source = this.CurrentNavigationSource;
-            IEdmStructuredType resourceType = this.CurrentResourceType;
+            IEdmTypeReference resourceTypeReference = this.CurrentResourceTypeReference;
 
             this.jsonLightResourceDeserializer.AssertJsonCondition(JsonNodeType.StartObject, JsonNodeType.Property,
                 JsonNodeType.EndObject, JsonNodeType.PrimitiveValue);
@@ -1848,7 +1848,7 @@ namespace Microsoft.OData.JsonLight
                     if (this.CurrentResourceType.TypeKind == EdmTypeKind.Untyped)
                     {
                         this.EnterScope(new JsonLightPrimitiveScope(new ODataPrimitiveValue(primitiveValue),
-                            this.CurrentNavigationSource, this.CurrentResourceType, this.CurrentScope.ODataUri));
+                            this.CurrentNavigationSource, this.CurrentResourceTypeReference, this.CurrentScope.ODataUri));
                     }
                     else
                     {
@@ -1858,8 +1858,13 @@ namespace Microsoft.OData.JsonLight
                 else
                 {
                     // null resource
+                    if (this.CurrentResourceTypeReference.IsComplex() || this.CurrentResourceTypeReference.IsUntyped())
+                    {
+                        this.jsonLightResourceDeserializer.MessageReaderSettings.Validator.ValidateNullValue(this.CurrentResourceTypeReference, true, "", null);
+                    }
+
                     this.EnterScope(new JsonLightResourceScope(ODataReaderState.ResourceStart, /*resource*/ null,
-                        this.CurrentNavigationSource, this.CurrentResourceType, /*propertyAndAnnotationCollector*/null,
+                        this.CurrentNavigationSource, this.CurrentResourceTypeReference, /*propertyAndAnnotationCollector*/null,
                         /*projectedProperties*/null, this.CurrentScope.ODataUri));
                 }
 
@@ -1898,7 +1903,7 @@ namespace Microsoft.OData.JsonLight
                             IEdmStructuredType parsedType = parseResult.EdmType as IEdmStructuredType;
                             if (parsedType != null)
                             {
-                                resourceType = parsedType;
+                                resourceTypeReference = parsedType.ToTypeReference(true);
                                 source = parseResult.NavigationSource;
                             }
                         }
@@ -1927,7 +1932,7 @@ namespace Microsoft.OData.JsonLight
                 case ODataDeltaKind.None:
                 case ODataDeltaKind.Resource:
                     // Setup the new resource state
-                    this.StartResource(source, resourceType, propertyAndAnnotationCollector, selectedProperties);
+                    this.StartResource(source, resourceTypeReference, propertyAndAnnotationCollector, selectedProperties);
 
                     // Start reading the resource up to the first nested resource info
                     this.StartReadingResource();
@@ -1946,7 +1951,7 @@ namespace Microsoft.OData.JsonLight
                         this.StartDeletedResource(
                             deletedResource,
                             source,
-                            resourceType,
+                            resourceTypeReference,
                             propertyAndAnnotationCollector,
                             selectedProperties,
                             true /*is 4.0 Deleted Resource*/);
@@ -1956,7 +1961,7 @@ namespace Microsoft.OData.JsonLight
                         this.StartDeletedResource(
                             deletedResource,
                             source,
-                            resourceType,
+                            resourceTypeReference,
                             propertyAndAnnotationCollector,
                             selectedProperties);
 
@@ -2009,7 +2014,7 @@ namespace Microsoft.OData.JsonLight
             this.EnterScope(new JsonLightResourceSetScope(
                 deltaResourceSet,
                 this.CurrentNavigationSource,
-                this.CurrentResourceType as IEdmEntityType,
+                this.CurrentResourceTypeReference as IEdmEntityTypeReference,
                 selectedProperties,
                 this.CurrentScope.ODataUri,
                 /*isDelta*/ true));
@@ -2094,7 +2099,7 @@ namespace Microsoft.OData.JsonLight
                         if (primitiveValue != null)
                         {
                             this.EnterScope(new JsonLightPrimitiveScope(new ODataPrimitiveValue(primitiveValue),
-                                this.CurrentNavigationSource, this.CurrentResourceType, this.CurrentScope.ODataUri));
+                                this.CurrentNavigationSource, this.CurrentResourceTypeReference, this.CurrentScope.ODataUri));
                         }
                         else
                         {
@@ -2102,7 +2107,7 @@ namespace Microsoft.OData.JsonLight
                             {
                                 // null primitive
                                 this.EnterScope(new JsonLightPrimitiveScope(new ODataNullValue(),
-                                    this.CurrentNavigationSource, this.CurrentResourceType, this.CurrentScope.ODataUri));
+                                    this.CurrentNavigationSource, this.CurrentResourceTypeReference, this.CurrentScope.ODataUri));
                             }
                             else
                             {
@@ -2173,8 +2178,7 @@ namespace Microsoft.OData.JsonLight
                 this.CurrentJsonLightNestedResourceInfoScope.ReaderNestedResourceInfo;
             if (nestedResourceInfo.HasEntityReferenceLink)
             {
-                this.EnterScope(new Scope(ODataReaderState.EntityReferenceLink,
-                    nestedResourceInfo.ReportEntityReferenceLink(), null, null, this.CurrentScope.ODataUri));
+                this.EnterScope(new Scope(ODataReaderState.EntityReferenceLink, nestedResourceInfo.ReportEntityReferenceLink(), this.CurrentScope.ODataUri));
             }
             else if (nestedResourceInfo.HasValue)
             {
@@ -2213,7 +2217,7 @@ namespace Microsoft.OData.JsonLight
         /// <param name="propertyAndAnnotationCollector">The duplicate property names checker to use for the resource;
         /// or null if a new one should be created.</param>
         /// <param name="selectedProperties">The selected properties node capturing what properties should be expanded during template evaluation.</param>
-        private void StartResource(IEdmNavigationSource source, IEdmStructuredType resourceType, PropertyAndAnnotationCollector propertyAndAnnotationCollector,
+        private void StartResource(IEdmNavigationSource source, IEdmTypeReference resourceType, PropertyAndAnnotationCollector propertyAndAnnotationCollector,
             SelectedPropertiesNode selectedProperties)
         {
             this.EnterScope(new JsonLightResourceScope(
@@ -2236,7 +2240,7 @@ namespace Microsoft.OData.JsonLight
         /// or null if a new one should be created.</param>
         /// <param name="selectedProperties">The selected properties node capturing what properties should be expanded during template evaluation.</param>
         /// <param name="is40DeletedResource">Whether the current resource being read is a 4.0-style deleted resource.</param>
-        private void StartDeletedResource(ODataDeletedResource deletedResource, IEdmNavigationSource source, IEdmStructuredType resourceType, PropertyAndAnnotationCollector propertyAndAnnotationCollector,
+        private void StartDeletedResource(ODataDeletedResource deletedResource, IEdmNavigationSource source, IEdmTypeReference resourceType, PropertyAndAnnotationCollector propertyAndAnnotationCollector,
             SelectedPropertiesNode selectedProperties, bool is40DeletedResource = false)
         {
             this.EnterScope(new JsonLightDeletedResourceScope(
@@ -2302,7 +2306,7 @@ namespace Microsoft.OData.JsonLight
             Debug.Assert(readerNestedResourceInfo != null, "readerNestedResourceInfo != null");
             ODataNestedResourceInfo nestedResourceInfo = readerNestedResourceInfo.NestedResourceInfo;
             IEdmProperty nestedProperty = readerNestedResourceInfo.NestedProperty;
-            IEdmType targetResourceType = readerNestedResourceInfo.NestedResourceType;
+            IEdmTypeReference targetResourceTypeReference = readerNestedResourceInfo.NestedResourceTypeReference;
 
             Debug.Assert(
                 this.jsonLightResourceDeserializer.JsonReader.NodeType == JsonNodeType.Property ||
@@ -2318,19 +2322,19 @@ namespace Microsoft.OData.JsonLight
                 "The navigation property must match the nested resource info.");
 
             // we are at the beginning of a link
-            if (targetResourceType == null && nestedProperty != null)
+            if (targetResourceTypeReference == null && nestedProperty != null)
             {
                 IEdmTypeReference nestedPropertyType = nestedProperty.Type;
-                targetResourceType = nestedPropertyType.IsCollection()
-                    ? nestedPropertyType.AsCollection().ElementType().AsStructured().StructuredDefinition()
-                    : nestedPropertyType.AsStructured().StructuredDefinition();
+                targetResourceTypeReference = nestedPropertyType.IsCollection()
+                    ? nestedPropertyType.AsCollection().ElementType().AsStructured()
+                    : nestedPropertyType.AsStructured();
             }
 
             // Since we don't have the entity metadata builder for the resource read out from a nested payload
             // as stated in ReadAtResourceSetEndImplementationSynchronously(), we cannot access it here which otherwise
             // would lead to an exception.
             if (this.jsonLightInputContext.ReadingResponse && !this.IsReadingNestedPayload
-                && (targetResourceType == null || targetResourceType.IsStructuredOrStructuredCollectionType()))
+                && (targetResourceTypeReference == null || targetResourceTypeReference.Definition.IsStructuredOrStructuredCollectionType()))
             {
                 // Hookup the metadata builder to the nested resource info.
                 // Note that we set the metadata builder even when navigationProperty is null, which is the case when the link is undeclared.
@@ -2405,7 +2409,7 @@ namespace Microsoft.OData.JsonLight
             odataUri.Path = odataPath;
 
             JsonLightNestedResourceInfoScope newScope = new JsonLightNestedResourceInfoScope(readerNestedResourceInfo, navigationSource,
-                targetResourceType, odataUri);
+                targetResourceTypeReference, odataUri);
 
             var derivedTypeConstraints = this.jsonLightInputContext.Model.GetDerivedTypeConstraints(nestedProperty);
             if (derivedTypeConstraints != null)
@@ -2475,7 +2479,7 @@ namespace Microsoft.OData.JsonLight
         /// <param name="state">The <see cref="ODataReaderState"/> to use for the new scope.</param>
         private void ReplaceScope(ODataReaderState state)
         {
-            this.ReplaceScope(new Scope(state, this.Item, this.CurrentNavigationSource, this.CurrentResourceType,
+            this.ReplaceScope(new Scope(state, this.Item, this.CurrentNavigationSource, this.CurrentResourceTypeReference,
                 this.CurrentScope.ODataUri));
         }
 
@@ -2528,7 +2532,7 @@ namespace Microsoft.OData.JsonLight
                         ODataReaderState.ResourceEnd,
                         (ODataResource)this.Item,
                         this.CurrentNavigationSource,
-                        this.CurrentResourceType,
+                        this.CurrentResourceTypeReference,
                         this.CurrentResourceState.PropertyAndAnnotationCollector,
                         this.CurrentResourceState.SelectedProperties,
                         this.CurrentScope.ODataUri));
@@ -2540,7 +2544,7 @@ namespace Microsoft.OData.JsonLight
                         ODataReaderState.DeletedResourceEnd,
                         (ODataDeletedResource)this.Item,
                         this.CurrentNavigationSource,
-                        this.CurrentResourceType,
+                        this.CurrentResourceTypeReference,
                         this.CurrentResourceState.PropertyAndAnnotationCollector,
                         this.CurrentResourceState.SelectedProperties,
                         this.CurrentScope.ODataUri));
@@ -2584,7 +2588,7 @@ namespace Microsoft.OData.JsonLight
                             this.jsonLightResourceDeserializer.ContextUriParseResult.EdmType = resourceType;
                         }
 
-                        this.CurrentScope.ResourceType = resourceType;
+                        this.CurrentScope.ResourceTypeReference = resourceType.ToTypeReference(true).AsStructured();
                     }
                 }
             }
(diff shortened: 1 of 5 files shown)


src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2381,8 +2389,9 @@ private abstract class JsonLightResourceBaseScope : Scope, IODataJsonLightReader
                 IEdmStructuredType expectedResourceType,
                 PropertyAndAnnotationCollector propertyAndAnnotationCollector,
                 SelectedPropertiesNode selectedProperties,
-                ODataUri odataUri)
-                : base(readerState, resource, navigationSource, expectedResourceType, odataUri)
+                ODataUri odataUri,
+                bool resourceNullable = true)
```
biaol-odata: <param> tag for resourceNullable needs to be added.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2571,8 +2581,8 @@ private sealed class JsonLightResourceSetScope : Scope
             ///   it's the expected base type of the entries in the resource set.
             ///   note that it might be a more derived type than the base type of the entity set for the resource set.
             /// In all cases the specified type must be an entity type.</remarks>
-            internal JsonLightResourceSetScope(ODataResourceSetBase resourceSet, IEdmNavigationSource navigationSource, IEdmStructuredType expectedResourceType, SelectedPropertiesNode selectedProperties, ODataUri odataUri, bool isDelta)
-                : base(isDelta ? ODataReaderState.DeltaResourceSetStart : ODataReaderState.ResourceSetStart, resourceSet, navigationSource, expectedResourceType, odataUri)
+            internal JsonLightResourceSetScope(ODataResourceSetBase resourceSet, IEdmNavigationSource navigationSource, IEdmStructuredType expectedResourceType, SelectedPropertiesNode selectedProperties, ODataUri odataUri, bool isDelta, bool resourceNullable = false)
```
biaol-odata: tag for resourceNullable needs to be added.
Also, looks like this cstor is used only twice in this file, can we make it not optional by updating the two references?

KanishManuja-MS: I believe making it optional conveys that by default, the resourcetype is nullable if not specified.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2381,8 +2390,9 @@ private abstract class JsonLightResourceBaseScope : Scope, IODataJsonLightReader
                 IEdmStructuredType expectedResourceType,
```
xuzhg: I think you should change "expectedResourceType" to accept "IEdmStructureTypeReference".

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2148,7 +2156,7 @@ private void StartNestedResourceInfo(ODataJsonLightReaderNestedResourceInfo read
             odataUri.Path = odataPath;
 
             this.EnterScope(new JsonLightNestedResourceInfoScope(readerNestedResourceInfo, navigationSource,
-                targetResourceType, odataUri));
+                targetResourceType, odataUri, resourceNullable));
```
xuzhg: Can we get the "targetResourceTypeReference"?

xuzhg: I believe it can: See https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs#L1193

So, the structuredPropertyTypeOrItemType is from the property type. the property type is type reference. I do think you should use "structuredProperty.Type" directly.

KanishManuja-MS: This is on the serializer. We don't have access to the property on the reader. Again will check with you in person.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
mikepizzo: Agree that this would be the preferred design; TypeReference encompasses Type and nullability; it would be nice if we could use that, rather than passing individually.

---
In reply to: [235490462](https://github.com/OData/odata.net/pull/1322#discussion_r235490462) [](ancestors = 235490462)

mikepizzo: Looks like you were able to make it work without too much churn?

---
In reply to: [237206849](https://github.com/OData/odata.net/pull/1322#discussion_r237206849) [](ancestors = 237206849)

mikepizzo: If you pass an IEdmStructuredTypeReference in to the function, you can still have an IEdmStructuredType property (IEdmStructuredTypeReference wraps the IEdmStructuredType).  But changing the signature to pass in an IEdmStructuredTypeReference would give you a more concise way of passing in the type and the nullability.

---
In reply to: [237206849](https://github.com/OData/odata.net/pull/1322#discussion_r237206849) [](ancestors = 237206849)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2381,7 +2389,8 @@ private abstract class JsonLightResourceBaseScope : Scope, IODataJsonLightReader
                 IEdmStructuredType expectedResourceType,
                 PropertyAndAnnotationCollector propertyAndAnnotationCollector,
                 SelectedPropertiesNode selectedProperties,
-                ODataUri odataUri)
+                ODataUri odataUri,
+                bool resourceNullable = true)
```
mikepizzo: > = true [](start = 37, length = 7)

Similar to Biao's comment, I'm not generally a fan of optional params if we're not trying to support backward compatibility and it's easy enough to pass the default.

mikepizzo: >bool resourceNullable = true [](start = 16, length = 28)

Where is resourceNullable used?

src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -922,7 +965,30 @@ internal ODataUri ODataUri
...
+                    {
+                        return (IEdmStructuredType)ResourceTypeReference.Definition;
+                    }
+                    return null;
+                }
+                set
+                {
+                    if (value != null)
+                    {
+                        this.ResourceTypeReference = (IEdmStructuredTypeReference) value.ToTypeReference(true);
```
mikepizzo: >true [](start = 105, length = 4)

Should the default be nullable or not nullable?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
KanishManuja-MS: The get method for the nestedresourcetype ges the structured type from the property itself.

---
In reply to: [235493340](https://github.com/OData/odata.net/pull/1322#discussion_r235493340) [](ancestors = 235493340)

src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -939,7 +978,23 @@ internal ODataUri ODataUri
...
-            internal IEdmStructuredType ResourceType { get; set; }
+            internal IEdmStructuredType ResourceType
+            {
+                get
+                {
+                    if (this.ResourceTypeReference != null)
+                    {
+                        return (IEdmStructuredType)ResourceTypeReference.Definition;
+                    }
+                    return null;
```
xuzhg: >return null; [](start = 20, length = 12)

Insert a white line otherwise will have another build warning.

src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -869,6 +883,31 @@ protected internal class Scope
...
+            {
+            }
+
+            /// <summary>
+            /// Constructor creating a new reader scope.
+            /// </summary>
+            /// <param name="state">The reader state of this scope.</param>
+            /// <param name="item">The item attached to this scope.</param>
+            /// <param name="navigationSource">The navigation source we are going to read entities for.</param>
+            /// <param name="expectedResourceType">The expected resource type for the scope.</param>
```
xuzhg: >/// <param name="expectedResourceType">The expected resource type for the scope.</param> [](start = 12, length = 88)

change it.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -1932,7 +1937,7 @@ private void ReadNextNestedResourceInfoContentItemInRequest()
             if (nestedResourceInfo.HasEntityReferenceLink)
             {
                 this.EnterScope(new Scope(ODataReaderState.EntityReferenceLink,
-                    nestedResourceInfo.ReportEntityReferenceLink(), null, null, this.CurrentScope.ODataUri));
+                    nestedResourceInfo.ReportEntityReferenceLink(), null, (IEdmStructuredTypeReference) null, this.CurrentScope.ODataUri));
```
xuzhg: >(IEdmStructuredTypeReference) [](start = 74, length = 29)

I don't like it. Do we have other solutions?

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
```diff
@@ -773,6 +773,82 @@ public void ReadingNullValueForDeclaredCollectionPropertyInComplexTypeShouldFail
...
+                    if (reader.State == ODataReaderState.ResourceEnd)
+                    {
+                        resources.Add(reader.Item);
+                    }
+                });
+            test.ShouldThrow<ODataException>().WithMessage("A null value was found with the expected type 'NS.Car[Nullable=False]'. The expected type 'NS.Car[Nullable=False]' does not allow null values.");
+        }
+
+        [Fact]
+        public void ReadingNullValueForNonNullableCollectionofPrimitiveShouldThrow()
```
xuzhg: >ReadingNullValueForNonNullableCollectionofPrimitiveShouldThrow [](start = 20, length = 62)

Maybe add the test case for non-collection property?

src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -191,10 +191,24 @@ protected IEdmStructuredType CurrentResourceType
                 Debug.Assert(resourceType == null || this.inputContext.Model.IsUserModel(), "We can only have structured type if we also have metadata.");
                 return resourceType;
             }
+        }
 
+        /// <summary>
+        /// Returns the expected resource type reference for the current scope.
```
xuzhg: >Returns [](start = 12, length = 7)

Get/Sets ....

src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -368,7 +382,7 @@ public override sealed bool Read()
         public override sealed Task<bool> ReadAsync()
         {
             this.VerifyCanRead(false);
-            return this.ReadAsynchronously().FollowOnFaultWith(t => this.EnterScope(new Scope(ODataReaderState.Exception, null, null, null, null)));
+            return this.ReadAsynchronously().FollowOnFaultWith(t => this.EnterScope(new Scope(ODataReaderState.Exception, null, null,(IEdmStructuredTypeReference) null, null)));
```
xuzhg: >IEdmStructuredTypeReference [](start = 134, length = 27)

I don't like to call using (IEdmStructuredTypeReference)null. It maybe a design refactor.

src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -869,6 +883,31 @@ protected internal class Scope
             /// In all cases the specified type must be an structured type.</remarks>
             [SuppressMessage("Microsoft.Performance", "CA1800:DoNotCastUnnecessarily", Justification = "Debug.Assert check only.")]
             internal Scope(ODataReaderState state, ODataItem item, IEdmNavigationSource navigationSource, IEdmStructuredType expectedResourceType, ODataUri odataUri)
```
xuzhg: >IEdmStructuredType expectedResourceType [](start = 106, length = 39)

maybe you can do like this:

Add a bool argument as "bool isNullable = true"

src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -939,7 +978,23 @@ internal ODataUri ODataUri
...
             /// was not found yet, or the one specified in the payload itself (the real one).
             /// </summary>
-            internal IEdmStructuredType ResourceType { get; set; }
+            internal IEdmStructuredType ResourceType
+            {
+                get
+                {
+                    if (this.ResourceTypeReference != null)
+                    {
+                        return (IEdmStructuredType)ResourceTypeReference.Definition;
```
xuzhg: >ResourceTypeReference.Definition [](start = 51, length = 32)

There's an extension method call "StructuralDefinition()" ?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2061,6 +2066,13 @@ private void StartNestedResourceInfo(ODataJsonLightReaderNestedResourceInfo read
             ODataNestedResourceInfo nestedResourceInfo = readerNestedResourceInfo.NestedResourceInfo;
             IEdmProperty nestedProperty = readerNestedResourceInfo.NestedProperty;
             IEdmStructuredType targetResourceType = readerNestedResourceInfo.NestedResourceType;
```
xuzhg: >NestedResourceType [](start = 77, length = 18)

maybe do the same change for the "NestedResourceType", I mean it should be "NestedResourceTypeReference", then you don't need to call "ToTypeReference(...)".

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
KanishManuja-MS: Added another constructor.

---
In reply to: [263957299](https://github.com/OData/odata.net/pull/1322#discussion_r263957299) [](ancestors = 263957299)

KanishManuja-MS:
@xuzhg Can you please verify the changes and then I can assign this to mike?

src/Microsoft.OData.Core/ODataReaderCore.cs
xuzhg: what's that?

---
In reply to: [264344042](https://github.com/OData/odata.net/pull/1322#discussion_r264344042) [](ancestors = 264344042)

xuzhg:
I went through the changes, Can you close the comments that you addressed?

---
In reply to: [474038708](https://github.com/OData/odata.net/pull/1322#issuecomment-474038708) [](ancestors = 474038708)

src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -853,22 +867,9 @@ protected internal class Scope
...
-            /// EntryStart -          it's the expected base type of the resource. If the resource has no type name specified
-            ///                       this type will be assumed. Otherwise the specified type name must be
-            ///                       the expected type or a more derived type.
-            /// NestedResourceInfoStart - it's the expected base type the entries in the expanded link (either the single resource
-            ///                       or entries in the expanded resource set).
-            /// EntityReferenceLink - it's null, no need for types on entity reference links.
-            /// In all cases the specified type must be an structured type.</remarks>
             [SuppressMessage("Microsoft.Performance", "CA1800:DoNotCastUnnecessarily", Justification = "Debug.Assert check only.")]
-            internal Scope(ODataReaderState state, ODataItem item, IEdmNavigationSource navigationSource, IEdmStructuredType expectedResourceType, ODataUri odataUri)
+            internal Scope(ODataReaderState state, ODataItem item, ODataUri odataUri)
```
xuzhg: >Scope [](start = 21, length = 5)

do we have the out call for this constructor? If no, i don't think we should create a one like this.

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
```diff
@@ -773,6 +773,82 @@ public void ReadingNullValueForDeclaredCollectionPropertyInComplexTypeShouldFail
             read.ShouldThrow<ODataException>().WithMessage("A null value was found for the property named 'CountriesOrRegions', which has the expected type 'Collection(Edm.String)[Nullable=False]'. The expected type 'Collection(Edm.String)[Nullable=False]' does not allow null values.");
         }
 
+        [Fact]
+        public void ReadingNullValueForNonNullableCollectionofComplexShouldThrow()
```
xuzhg: >of [](start = 60, length = 2)

..CollectionOfComplex...

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
```diff
@@ -773,6 +773,82 @@ public void ReadingNullValueForDeclaredCollectionPropertyInComplexTypeShouldFail
             read.ShouldThrow<ODataException>().WithMessage("A null value was found for the property named 'CountriesOrRegions', which has the expected type 'Collection(Edm.String)[Nullable=False]'. The expected type 'Collection(Edm.String)[Nullable=False]' does not allow null values.");
         }
 
+        [Fact]
+        public void ReadingNullValueForNonNullableCollectionofComplexShouldThrow()
+        {
+            EdmModel model = new EdmModel();
+            EdmComplexType complexType = new EdmComplexType("NS", "Car");
+            EdmComplexType door = new EdmComplexType("NS", "Door");
+            EdmComplexTypeReference frontDoor = new EdmComplexTypeReference(complexType, false);
```
xuzhg: >frontDoor [](start = 36, length = 9)

The name of "frontDoor" sounds like a sub class name for "Door", but it looks not. Can you rename it?

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
```diff
@@ -773,6 +773,82 @@ public void ReadingNullValueForDeclaredCollectionPropertyInComplexTypeShouldFail
             read.ShouldThrow<ODataException>().WithMessage("A null value was found for the property named 'CountriesOrRegions', which has the expected type 'Collection(Edm.String)[Nullable=False]'. The expected type 'Collection(Edm.String)[Nullable=False]' does not allow null values.");
         }
 
+        [Fact]
+        public void ReadingNullValueForNonNullableCollectionofComplexShouldThrow()
+        {
+            EdmModel model = new EdmModel();
+            EdmComplexType complexType = new EdmComplexType("NS", "Car");
+            EdmComplexType door = new EdmComplexType("NS", "Door");
```
xuzhg: >EdmComplexType door = new EdmComplexType("NS", "Door"); [](start = 12, length = 55)

So, "door" is useless?

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
```diff
@@ -773,6 +773,82 @@ public void ReadingNullValueForDeclaredCollectionPropertyInComplexTypeShouldFail
...
+            List<ODataItem> resources = new List<ODataItem>();
+            Action test = () => this.ReadEntryPayload(model, payload, entitySet, entityType,
+                reader =>
+                {
+                    if (reader.State == ODataReaderState.ResourceEnd)
+                    {
+                        resources.Add(reader.Item);
+                    }
+                });
+            test.ShouldThrow<ODataException>().WithMessage("A null value was found with the expected type 'NS.Car[Nullable=False]'. The expected type 'NS.Car[Nullable=False]' does not allow null values.");
```
xuzhg: >NS.Car [](start = 107, length = 6)

Maybe "NS.Door..."

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
xuzhg: >ReadingNullValueForNonNullableCollectionofPrimitiveShouldThrow [](start = 20, length = 62)

...CollectionOfPrimitive...

xuzhg: >ReadingNullValueForNonNullableCollectionofPrimitiveShouldThrow() [](start = 20, length = 64)

Maybe add a test for "CollectionOfEnum"

src/Microsoft.OData.Core/Metadata/MetadataUtilsCommon.cs
```diff
@@ -125,6 +125,31 @@ internal static bool IsODataTypeDefinitionTypeKind(this IEdmType type)
...
+
+            return typeReference.Definition.IsODataUnTypedTypeKind();
+        }
+
+        /// <summary>
+        /// Checks whether a type refers to an OData Untyped definition.
+        /// </summary>
+        /// <param name="type">The (non-null) <see cref="IEdmType"/> to check.</param>
+        /// <returns>true if the <paramref name="type"/> is an OData Untyped definition; otherwise false.</returns>
+        internal static bool IsODataUnTypedTypeKind(this IEdmType type)
```
xuzhg: >IsODataUnTypedTypeKind [](start = 29, length = 22)

Do we have "IsUntyped(this IEdmType type)" extension method?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs
```diff
@@ -168,6 +174,33 @@ internal IEdmStructuredType NestedResourceType
...
+                    }
+
+                    bool resourceTypeNullable = true;
+                    if (nestedProperty != null && nestedProperty.Type != null)
+                    {
+                        resourceTypeNullable = nestedProperty.Type.IsNullable;
+                    }
+
+                    this.nestedResourceTypeReference = (nestedResourceType != null) ? nestedResourceType.ToTypeReference(resourceTypeNullable).AsStructured() : null;
+                }
```
xuzhg: My suggestion:

1. Remove private fields:  "nestedResourceType" & "nestedResourceTypeReference".


2. Move these or similar codes into corresponding constructor, in first constructor, call " NestedResourceTypeReference = this.nestedProperty.Type.AsStructured(); (the name maybe different).

   in second construcotr, do same as above if the nested property not null, otherwise, 
NestedResourceTypeReference =nestedResourceType.ToTypeReference(true).AsStructured();


3. Add "internal IEdmStructuredTypeReference NestedResourceTypeReference { get; private set;}

4. Change "internal IEdmStrucutruedType NestedResourceType { get { return NestedResourceTypeReference.StructuredDefinition();}

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -1721,6 +1721,11 @@ private void ReadExpandedNestedResourceInfoEnd(bool isCollection)
                 else
                 {
                     // null resource
+                    if (this.CurrentResourceTypeReference.IsODataComplexTypeKind() || this.CurrentResourceTypeReference.IsODataUnTypedTypeKind())
```
xuzhg: >IsODataComplexTypeKind [](start = 58, length = 22)

shall we add "Enum" check?

src/Microsoft.OData.Core/Metadata/MetadataUtilsCommon.cs
mikepizzo: I have added one as part of the LargeObjectStreaming PR.

---
In reply to: [266676418](https://github.com/OData/odata.net/pull/1322#discussion_r266676418) [](ancestors = 266676418)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
KanishManuja-MS: Collection of primitives and enums do not go through this check.

---
In reply to: [266678953](https://github.com/OData/odata.net/pull/1322#discussion_r266678953) [](ancestors = 266678953)

src/Microsoft.OData.Core/ODataReaderCore.cs
KanishManuja-MS: I decided to add another constructor as I had an IsNullable optional parameter to begin with but we wanted to use type reference instead.

---
In reply to: [266614311](https://github.com/OData/odata.net/pull/1322#discussion_r266614311) [](ancestors = 266614311,264344042)

src/Microsoft.OData.Core/ODataReaderCore.cs
KanishManuja-MS: This is called in a couple of other places.

---
In reply to: [266674779](https://github.com/OData/odata.net/pull/1322#discussion_r266674779) [](ancestors = 266674779)

src/Microsoft.OData.Core/Metadata/MetadataUtilsCommon.cs
KanishManuja-MS: I tried looking for this in your LOS PR but I was unable to find it. 

---
In reply to: [267020044](https://github.com/OData/odata.net/pull/1322#discussion_r267020044) [](ancestors = 267020044,266676418)

KanishManuja-MS: I tried finding this in your PR but I was unable to find it. Can you point me to the class to which this method was added?

---
In reply to: [267020044](https://github.com/OData/odata.net/pull/1322#discussion_r267020044) [](ancestors = 267020044,266676418)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs
```diff
@@ -65,6 +61,7 @@ private ODataJsonLightReaderNestedResourceInfo(ODataNestedResourceInfo nestedRes
 
             this.nestedResourceInfo = nestedResourceInfo;
             this.nestedProperty = nestedProperty;
+            NestedResourceTypeReference = (nestedProperty != null) ? nestedProperty.Type.ToStructuredType().ToTypeReference(nestedProperty.Type.IsNullable).AsStructured() : null;
```
xuzhg: >nestedProperty.Type.ToStructuredType().ToTypeReference(nestedProperty.Type.IsNullable).AsStructured() [](start = 69, length = 101)

You can just use "nestedProperty.Type..AsStructured()" directly

xuzhg: >(nestedProperty != null) [](start = 42, length = 24)

Needn't to check, right.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs
```diff
@@ -82,7 +79,16 @@ private ODataJsonLightReaderNestedResourceInfo(ODataNestedResourceInfo nestedRes
...
+
+            bool resourceTypeNullable = true;
+            if (nestedProperty != null && nestedProperty.Type != null)
+            {
+                resourceTypeNullable = nestedProperty.Type.IsNullable;
+            }
+
+            IEdmStructuredType nestedType = (nestedResourceType != null) ? nestedResourceType : (nestedProperty != null) ? nestedProperty.Type.ToStructuredType() : null;
+
+            NestedResourceTypeReference = (nestedType != null) ? nestedType.ToTypeReference(resourceTypeNullable).AsStructured() :  null;
```
xuzhg: Just do like:


if (nestedProperty != null)
{
                NestedResourceTypeReference = nestedProperty.Type.AsStructured();
}
else
 {
       NestedResourceTypeReference = nestedResourceType.ToTypeReference(true);
 }

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs
```diff
@@ -159,15 +165,20 @@ internal IEdmStructuredType NestedResourceType
         {
             get
             {
-                if (this.nestedResourceType == null && this.nestedProperty != null)
+               if (NestedResourceTypeReference != null)
```
xuzhg: >NestedResourceTypeReference [](start = 19, length = 27)

NestedResourceTypeReference should never be "null", right?

src/Microsoft.OData.Core/Metadata/MetadataUtilsCommon.cs
xuzhg: It's in EdmTypeSemantics.cs 

---
In reply to: [267131123](https://github.com/OData/odata.net/pull/1322#discussion_r267131123) [](ancestors = 267131123,267020044,266676418)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs
KanishManuja-MS: This is what I tried first but a bunch of test cases failed. 
I can clean up the coding style but the logic will slightly be different. 

---
In reply to: [267147384](https://github.com/OData/odata.net/pull/1322#discussion_r267147384) [](ancestors = 267147384)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs
KanishManuja-MS: I need to use ToStructuredType to correctly handle collections. I have cleaned up the code little though. 

---
In reply to: [267146833](https://github.com/OData/odata.net/pull/1322#discussion_r267146833) [](ancestors = 267146833)

KanishManuja-MS:
@mikepizzo I know you looked at this.  I am not sure if you wanted to take a look again, I am waiting on your approval to merge this.

mikepizzo:
>         private ODataJsonLightReaderNestedResourceInfo(ODataNestedResourceInfo nestedResourceInfo, IEdmProperty nestedProperty, IEdmStructuredType nestedResourceType, bool isExpanded)

Was it too expensive to change this to take an IEdmStructuredTypeReference?  Does the caller of this method have information about the nullability that may get lost in translation (especially if nestedProperty is null)?

---
Refers to: src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs:76 in e9251c2. [](commit_id = e9251c298de05c7baf7c71e0d47a9c764a255b49, deletion_comment = False)

KanishManuja-MS:
>         private ODataJsonLightReaderNestedResourceInfo(ODataNestedResourceInfo nestedResourceInfo, IEdmProperty nestedProperty, IEdmStructuredType nestedResourceType, bool isExpanded)

It would have changed the signature of a protected method which would have been a breaking change. Yes, the calling method does have information about nullability when the nestedProperty is null but that is an existing bug/behavior which I can track separately. 

---
In reply to: [475761426](https://github.com/OData/odata.net/pull/1322#issuecomment-475761426) [](ancestors = 475761426)

---
Refers to: src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs:76 in e9251c2. [](commit_id = e9251c298de05c7baf7c71e0d47a9c764a255b49, deletion_comment = False)

mikepizzo:
>         private ODataJsonLightReaderNestedResourceInfo(ODataNestedResourceInfo nestedResourceInfo, IEdmProperty nestedProperty, IEdmStructuredType nestedResourceType, bool isExpanded)

What method signature would have changed?  This is a private class, sealed class with a private constructor.  Is it that the method that calls this constructor is protected, and that class is public or protected/not sealed?

---
In reply to: [475782785](https://github.com/OData/odata.net/pull/1322#issuecomment-475782785) [](ancestors = 475782785,475761426)

---
Refers to: src/Microsoft.OData.Core/JsonLight/ODataJsonLightReaderNestedResourceInfo.cs:76 in e9251c2. [](commit_id = e9251c298de05c7baf7c71e0d47a9c764a255b49, deletion_comment = False)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -911,7 +911,7 @@ protected override Task<bool> ReadAtDeltaDeletedLinkImplementationAsync()
                 ? null
                 : this.jsonLightResourceDeserializer.ContextUriParseResult.SelectQueryOption;
 
-            SelectedPropertiesNode selectedProperties = SelectedPropertiesNode.Create(selectQueryOption, this.CurrentResourceType, this.jsonLightInputContext.Model);
+            SelectedPropertiesNode selectedProperties = SelectedPropertiesNode.Create(selectQueryOption, (this.CurrentResourceTypeReference !=null) ? this.CurrentResourceTypeReference.AsStructured().StructuredDefinition() : null, this.jsonLightInputContext.Model);
```
mikepizzo: >(this.CurrentResourceTypeReference !=null) [](start = 105, length = 42)

So we are asserting that, if the CurrentResourceTypeReference is non-null, then it is a structured type, right? Seems reasonable.

Issue #1313, opened 2018-11-02
IN operator containing single quote (apostrophe) within a string literal throws an error


adzhiljano:
According to the [OData ABNF v4.0](http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/abnf/odata-abnf-construction-rules.txt) (I don't think this will be changed in v4.01):

> string           = SQUOTE *( SQUOTE-in-string / pchar-no-SQUOTE ) SQUOTE
SQUOTE-in-string = SQUOTE SQUOTE ; two consecutive single quotes represent one within a string literal

**two consecutive single quotes represent one within a string literal**, but when trying to use a single quote in a string literal inside IN operator  like `$filter=StringProperty in ('i''m', 'test')` - a parsing error is produced

### Assemblies affected

Microsoft.OData.Core version=7.5.0
Microsoft.AspNet.OData version=7.0.1

### Reproduce steps

`$filter=StringProperty in ('i''m')`
`$filter=StringProperty in ('i''m', 'test')`

### Expected result

No error will occur

### Actual result

An error with the following message occurs:
`The query specified in the URI is not valid. Invalid JSON. A comma character ',' was expected in scope 'Array'. Every two elements in an array and properties of an object must be separated by commas.`

### Additional detail

The following work as expected (haven't checked all the functions and operators though, but I guess they are being parsed in a common manner):
`$filter=contains(StringProperty, 'i''m')`
`$filter=StringProperty eq 'i''m'`
`$filter=startswith(StringProperty ,'i''m')`
`$filter=concat(StringProperty,'i''m') eq 'sometexti''m'`

adzhiljano:
Forgot to mention something else which is related:

`$filter=StringProperty eq "foo"`(**notice the double quotes**) results in :

`The query specified in the URI is not valid. Syntax error: character '"' is not valid at position 18 in 'StringProperty eq "foo"'.`

but the following is fine:

`$filter=StringProperty in ("foo", "bar")`(**notice the double quotes**)

Correct me if I'm wrong, but the string literals inside the IN operator are not parsed according to the ABNF.(may be this should be the title of this issue)

biaol-odata:
looks like escape character issue related to single/double quotes.

biaol-odata:
Potential related to issue on WebAPI: https://github.com/OData/WebApi/issues/1596

PR #1323, opened 2018-11-14
IN operator fixes for escaped single-quote and Guid literals

biaol-odata:
### Issues

This pull request fixes the following issues related to IN operator:
- https://github.com/OData/odata.net/issues/1313: Invalid Json due to single quotes.
- https://github.com/OData/WebApi/issues/1596: Invalid Json due to Guid literals.

### Description

*Fix the escaped single quote for collection of strings for IN operator*
*Fix for collection of Guid literals for IN operator*

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -4,6 +4,8 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using System.Globalization;
+
 namespace Microsoft.OData.UriParser
 {
     using System;
@@ -92,6 +94,39 @@ namespace Microsoft.OData.UriParser
                     replacedText[0] = '[';
                     replacedText[replacedText.Length - 1] = ']';
                     bracketLiteralText = replacedText.ToString();
+
+                    Debug.Assert(expectedType.IsCollection());
+                    if (expectedType.Definition.AsElementType().FullTypeName().Equals("Edm.String"))
+                    {
+                        // For collection of strings, need to convert single-quoted string to double-quoted string,
+                        // and also, per ABNF, two consecutive single quotes  to one single quote.
+                        // Sample: ['a''bc','''def','xyz'''] ==> ["a'bc","'def","xyz'"], which is legitimate Json format.
+                        string[] items = bracketLiteralText.Substring(1, bracketLiteralText.Length - 2).Split(',');
+
+                        // Skip conversion if the items are already in double-quote format (for backward compatibility).
+                        // Note that per ABNF, query option strings should use single quotes.
+                        if (items.Length > 0 && items[0][0] == '\'')
+                        {
+                            StringBuilder builder = new StringBuilder();
+                            for (int i = 0; i < items.Length; i++)
+                            {
+                                string convertedItem = UriParserHelper.RemoveQuotes(items[i]);
+                                if (i != items.Length - 1)
+                                {
+                                    builder.AppendFormat(CultureInfo.InvariantCulture, "\"{0}\",", convertedItem);
+                                }
+                                else
+                                {
+                                    // No trailing comma separator for last item of the collection.
+                                    builder.AppendFormat(CultureInfo.InvariantCulture, "\"{0}\"", convertedItem);
+                                }
+                            }
+
+                            bracketLiteralText =
+                                String.Format(CultureInfo.InvariantCulture, "[{0}]", builder.ToString());
+                        }
+                    }
+
                 }
 
                 object collection = ODataUriConversionUtils.ConvertFromCollectionValue(bracketLiteralText, model, expectedType);
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -170,10 +170,16 @@ namespace Microsoft.OData.UriParser
         /// <returns>The double-quoted string with single quotes properly escaped.</returns>
         private static string NormalizeStringItem(string str)
         {
+            // Validate the string item is quoted properly.
+            if ( !(    str[0] == '\'' && str[str.Length-1] == '\''
+                    || str[0] == '"'  && str[str.Length-1] == '"')
+                )
+            {
+                throw new ODataException(ODataErrorStrings.StringItemShouldBeQuoted(str));
+            }
+
             // Skip conversion if the items are already in double-quote format (for backward compatibility).
             // Note that per ABNF, query option strings should use single quotes.
-            Debug.Assert(str[0] == '\'' || str[0] == '"', "string needs to be single/double quoted.");
-
             string convertedString = str;
             if (str[0] == '\'')
             {
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -9,6 +9,7 @@ namespace Microsoft.OData.UriParser
     using System;
     using System.Diagnostics;
     using System.Globalization;
+    using System.Linq;
     using System.Text;
     using Microsoft.OData.Edm;
     using ODataErrorStrings = Microsoft.OData.Strings;
@@ -22,6 +23,14 @@ namespace Microsoft.OData.UriParser
         /// Method to use for binding the parent node, if needed.
         /// </summary>
         private readonly Func<QueryToken, QueryNode> bindMethod;
+
+        /// <summary>
+        /// Delegate for a function that normalizes a string item representing a certain type.
+        /// Each type should define a different implementation of the delegate.
+        /// </summary>
+        /// <param name="item">The item to be normalized.</param>
+        /// <returns>Normalized string of the item.</returns>
+        private delegate string NormalizeFunction(string item);
 
         /// <summary>
         /// Constructs a InBinder with the given method to be used binding the parent token if needed.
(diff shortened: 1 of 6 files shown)


src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -92,6 +94,39 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
...
+                        string[] items = bracketLiteralText.Substring(1, bracketLiteralText.Length - 2).Split(',');
+
+                        // Skip conversion if the items are already in double-quote format (for backward compatibility).
+                        // Note that per ABNF, query option strings should use single quotes.
+                        if (items.Length > 0 && items[0][0] == '\'')
+                        {
+                            StringBuilder builder = new StringBuilder();
+                            for (int i = 0; i < items.Length; i++)
+                            {
+                                string convertedItem = UriParserHelper.RemoveQuotes(items[i]);
```
AlanWong-MS: What happens if the client is trying to find strings that actually have single quotes? For example, `StringItem in ('Alan's computer', 'Biao's mouse', 'Sam's keyboard')`.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -92,6 +94,39 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
...
+                    if (expectedType.Definition.AsElementType().FullTypeName().Equals("Edm.String"))
+                    {
+                        // For collection of strings, need to convert single-quoted string to double-quoted string,
+                        // and also, per ABNF, two consecutive single quotes  to one single quote.
+                        // Sample: ['a''bc','''def','xyz'''] ==> ["a'bc","'def","xyz'"], which is legitimate Json format.
+                        string[] items = bracketLiteralText.Substring(1, bracketLiteralText.Length - 2).Split(',');
+
+                        // Skip conversion if the items are already in double-quote format (for backward compatibility).
+                        // Note that per ABNF, query option strings should use single quotes.
+                        if (items.Length > 0 && items[0][0] == '\'')
```
AlanWong-MS: Does the ABNF say that the use of single and double quotes have to be consistent? For example, is the client allowed to do `StringItem in ("Alan's computer", "Biao''s mouse, 'Sam''s keyboard')`?

AlanWong-MS: This is probably a nitpick comment and not a common scenario.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -4,6 +4,8 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using System.Globalization;
```
AlanWong-MS: nit: this should be under the namespace

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
biaol-odata: Then the single-quote should be escaped by using two consecutive single-quotes, per ABNF.
> string           = SQUOTE *( SQUOTE-in-string / pchar-no-SQUOTE ) SQUOTE
SQUOTE-in-string = SQUOTE SQUOTE ; two consecutive single quotes represent one within a string literal

For example: ('Alan''s computer', 'Biao''s mouse', 'Sam''s keyboard')

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
biaol-odata: Better be. :-)
Actually ABNF specifies using single-quote only in query options, per my understanding. But we allows double-quote now for collection used by IN operator, so we are trying to keep backward-compatibility.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
biaol-odata: Yeah, will fix. Thanks.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
AlanWong-MS: Makes sense. In that case, if we are checking `items[0][0] == '\''`, does that mean if I use `StringItem in ("Alan's computer", 'Sam''s keyboard')`, it will miss the second item in the list?

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
AlanWong-MS: ah I see, so this was indeed a bug. Thanks for the pointing this out from the ABNF!

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
biaol-odata: Correct, it will miss the second item. This is only when inconsistent styles are used for items in the collection. :-) It would be an overkill to check for inconsistent styles?

AlanWong-MS: I guess since we are already diverging from the ABNF specification by allowing double-quote for backward-compatibility, it is up to us on how we want to treat inconsistencies. I think it's fine to just check based on the first item in that case.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -92,6 +94,39 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
...
                     replacedText[replacedText.Length - 1] = ']';
                     bracketLiteralText = replacedText.ToString();
+
+                    Debug.Assert(expectedType.IsCollection());
+                    if (expectedType.Definition.AsElementType().FullTypeName().Equals("Edm.String"))
+                    {
+                        // For collection of strings, need to convert single-quoted string to double-quoted string,
+                        // and also, per ABNF, two consecutive single quotes  to one single quote.
+                        // Sample: ['a''bc','''def','xyz'''] ==> ["a'bc","'def","xyz'"], which is legitimate Json format.
+                        string[] items = bracketLiteralText.Substring(1, bracketLiteralText.Length - 2).Split(',');
```
AlanWong-MS: Just a heads-up, when hardcoded complex type collections are allowed in the future, I think this line will have to be changed.

biaol-odata: I think I know what you are referring to. I heard Mike mentioned that IN operator collection will be changed to using brackets instead of parenthesis. Thanks for the heads up.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -110,5 +138,69 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
...
+        /// Function to normalize quoted string, ensuring single quotes are escaped properly.
+        /// If the string is double-quoted, no op since single quote doesn't need to be escaped.
+        /// </summary>
+        /// <param name="str">The quoted string item to be normalized.</param>
+        /// <returns>The double-quoted string with single quotes properly escaped.</returns>
+        private static string NormalizeStringItem(string str)
+        {
+            // Skip conversion if the items are already in double-quote format (for backward compatibility).
+            // Note that per ABNF, query option strings should use single quotes.
+            Debug.Assert(str[0] == '\'' || str[0] == '"', "string needs to be single/double quoted.");
```
KanishManuja-MS: Should this be an OData error instead of a Debug.Assert?

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -22,6 +24,14 @@ internal sealed class InBinder
         /// </summary>
         private readonly Func<QueryToken, QueryNode> bindMethod;
 
+        /// <summary>
+        /// Delegate for a function that normalizes a string item representing a certain type.
+        /// Each type should define a different implementation of the delegate.
+        /// </summary>
+        /// <param name="item">The item to be normalized.</param>
+        /// <returns>Normalized string of the item.</returns>
+        private delegate string NormalizeFunction(string item);
```
mikepizzo: Is this used anywhere?

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -92,6 +102,24 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
                     replacedText[0] = '[';
                     replacedText[replacedText.Length - 1] = ']';
                     bracketLiteralText = replacedText.ToString();
+
+                    Debug.Assert(expectedType.IsCollection());
+                    string expectedTypeFullName = expectedType.Definition.AsElementType().FullTypeName();
+                    if (expectedTypeFullName.Equals("Edm.String"))
```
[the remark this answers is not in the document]
biaol-odata: I would like to keep the scope of this PR to fixing the single-quote and GUID issue. We can examine DateTime primitive value later on and add / adjust logic.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
biaol-odata: This should be a Debug.Assert since the expected type is string, therefore the value in the json should be single-quote or double-quote.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -1772,6 +1772,34 @@ public void FilterWithInOperationWithParensCollection()
             filter.Expression.As<InNode>().Right.As<CollectionConstantNode>().LiteralText.Should().Be("(1,2,3)");
         }
 
+       [Theory]
+       [InlineData("('abc','xyz')")]
```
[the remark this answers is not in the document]
biaol-odata: Good suggestion. Added these two test cases.

PR #1325, opened 2018-11-16
[Step3] Support to read ODataResourceValue

xuzhg:
### Issues

*This pull request is the step3 for #1331, #1338 

### Description

Support read ODataResourceValue
1. Read the property value as ODataResourceValue
2. Read the instance annotation value as ODataResourceValue
3. Read the collection value item as ODataResourceValue
4. Add reading Unit tests
    
### Be noted: 
for the parameter value reading, ODL recommends to use the
corresponding resource, resourceset, collection reader to read the
  parameter value if the value is not primitive or enum. So, this change doesn't change the parameter reader.

- [x] *Test cases added*

### Additional work necessary


xuzhg:
It's a whole view of the "ODataResourceValue". However, It's too huge to review. I want to separate it into three parts. First part can be found at #1331 . Thanks.

xuzhg:

Issue #601, opened 2016-06-20
Remove  ODataComplexValue


xuzhg:
Remove  ODataComplexValue and the related codes

seanrodda:
Hi, I know this is closed, however I need some help as this broke a the ability to remove complex values properties from the OnEntryStarting event.
If an entity has a complex property, that particular property is not listed in the Properties of the Entry.
Is the a workaround or solution to remove a complex property from the Entry?

Issue #1258, opened 2018-09-21
what about ODataComplexValue in 7.x branch?


XenoAmess:
Hi.
I'm changing some codes from using your 5.x codes to 7.x branch, to deal with OData V4.
I see some of my codes are using a class ODataComplexValue. 
But I cannot find ODataComplexValue in 7.x.
Or , I found it but it is removed from product codes.
So I want to ask about this.
Why is it deleted? Is it deleted in the protocol, thus I can delete my codes related to it, Or I have to implement it again?
And if I didn't use [public IEnumerable<ODataProperty> Properties] and [public string TypeName]
in my code, is it necessary to implement them?
Thanks.
related class : [ODataComplexValue](https://github.com/OData/odata.net/blob/ODataV4-7.x/test/FunctionalTests/Taupo/Source/Taupo.Astoria/Contracts/OData/ODataComplexValue.cs)

xuzhg:
@XenoAmess We re-introduced a new class named "ODataResourceValue" to play like the "ODataComplexValue". Please try and let us know any questions.

Issue #593, opened 2016-06-08
Get complex parameter value inline from function returns ODataComplexValue


xuzhg:
Used:  \Bld-fs-im-g2-02\tfs\ODataV7x\V7xNightlySigned\20160606.2

If a function with complex type parameter, and complex type parameter value sends in-line, for example:

~/Default.MyFunction(address={\"@odata.type\":\"System.Web.OData.Routing.Address\",\"Street\":\"NE 24th St.\",\"City\":\"Redmond\"})

It's be parsed, but if you try to get the parameter value, it returns the "ODataComplexValue" object.

Please help verify and do fix.

xuzhg:
Close it because we introduced a new class named "ODataResourceValue" to play like the "ODataComplexValue".

PR #1331, opened 2018-11-21
[Step1] Add ODataResourceValue (complex and entity) class

xuzhg:
### Issues

*This pull request fixes issue #601, #1258, #593 *

### Description

1. Move all `ODataValue` related classes into `Value` subfolder
2. Add `ODataResourceValue` class derived from `ODataValue` into `Value`
subfolder
3. Update the PublicApi
4. Verify the property value in `ODataResource`
5. Add the unit tests related to `ODataResourceValue`

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/ODataResource.cs
```diff
@@ -183,7 +184,11 @@ public IEnumerable<ODataFunction> Functions
         public IEnumerable<ODataProperty> Properties
         {
             get { return this.MetadataBuilder.GetProperties(this.properties); }
-            set { this.properties = value; }
+            set
+            {
+                VerifyProperties(value);
+                this.properties = value;
+            }
```
mikepizzo: This will prevent me from passing in a list of properties that contains an ODataResourceValue (or collection containing an ODataResourceValue), but it won't prevent me from passing in a List<ODataProperty> that I later add an ODataResourceValue to, correct? Should we also validate on get, to make sure an ODataResourceValue hasn't been added to the list or a collection property after it was set?

mikepizzo:
So this introduces the ODataResourceValue class, but doesn't use it anywhere, right? Also, it's not fully supported in the stack in all the places other ODataValues are supported (i.e., TryConvertToClrInstanceAnnotation in InstanceAnnotationMaterializationPolicy, ODataPropertyConverter, ODataValueUtils, etc.), not to mention reading/writing (which I understand will be a separate PR)

xuzhg:
@mikepizzo You are correct. I titled this PR as [Step1]. In this PR, it only introduces the "ODataResourceValue". In the coming PRs, (at least one for writing, one for reading) will introduce the *usage* of "ODataResourceValue" class.

src/Microsoft.OData.Core/ODataResource.cs
xuzhg: @mikepizzo I updated this PR with a new commit. Would you please take a look?

mikepizzo: Still seems to have same issue; it appears that I can do:

  List<Properties> properties = new List<Properties>();
  myODataResource.Properties = properties; //passes validation as there are not (yet) any ODataResourceValue properties
  properties.Add(new ODataProperty("foo",new ODataResourceValue(....));   //now I've added an ODataResourceValue property; does this get flagged anywhere?

Will we add checks, i.e. when writing a top-level resource, that again validates that the Properties collection does not contain an ODataResourceValues?


---
In reply to: [236774022](https://github.com/OData/odata.net/pull/1331#discussion_r236774022) [](ancestors = 236774022)

src/Microsoft.OData.Core/Value/ODataResourceValue.cs
```diff
@@ -0,0 +1,38 @@
...
+        /// <summary>
+        /// Collection of custom instance annotations.
+        /// </summary>
+        [SuppressMessage("Microsoft.Usage", "CA2227:CollectionPropertiesShouldBeReadOnly",
+            Justification = "We want to allow the same instance annotation collection instance to be shared across ODataLib OM instances.")]
+        public ICollection<ODataInstanceAnnotation> InstanceAnnotations
+        {
+            get { return this.GetInstanceAnnotations(); }
+            set { this.SetInstanceAnnotations(value); }
+        }
```
mikepizzo: Seems inconsistent that I have instance annotations on ODataResourceValue and not other ODataValues.  I understand why we wouldn't want to add annotations to all of the ODataValues (since, in the non-annotation case, the annotations are applied to the property), but how would I write an annotation on a primitive value used in an annotation?

src/Microsoft.OData.Core/Value/ODataResourceValue.cs
```diff
@@ -0,0 +1,38 @@
...
+// </copyright>
+//---------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Diagnostics.CodeAnalysis;
+
+namespace Microsoft.OData
+{
+    /// <summary>
+    /// Represents the value of a resource (complex or entity).
```
mikepizzo: >Represents the value of a resource (complex or entity). [](start = 8, length = 55)

should probably clarify where this is used; i.e., in an annotation.

xuzhg: #Resolved.

src/Microsoft.OData.Core/Value/ODataResourceValue.cs
xuzhg: Discussed it offline.
So far, ODL doesn't allow to annotation on ODataPrimitiveValue or other ODataValue. 
However, the discussion is open. We should discuss it more later. And I filed an issue at https://github.com/OData/odata.net/issues/1353 to track on it.

xuzhg:
@mikepizzo I resolved the comments. Thanks.

xuzhg:

PR #1338, opened 2018-11-26
[Step2] Support to write ODataResourceValue

xuzhg:
### Issues

*This pull request is the second step for the #1331. It supports to write the ODataResourceValue.

### Description

Support to write ODataResourceValue:
        1. Write property value as ODataResourceValue
        2. Write Instance annotaiton value as ODataResourceValue
        3. Write Collection value item as ODataResourceValue
        4. Convert ODataResource to Uri literal
        5. Add the Unit test cases
    
### Be noted:
   so far, for parameter value writing,  if the parameter value is resource,
    resourceset or collection, ODL recommends to use corresponding writer (ResourceWriter, ResourceSetWriter, CollectionWriter) to write. so, in this change set, we don't include the support for the
    parameter value as ODataResourceValue writing.

- [x] *Test cases added*

### Additional work necessary


xuzhg:

Issue #52, opened 2015-01-14
Support of retrieving annotations on parameters and return types of function & actions


dingeins:
One scenario of it would be that the client can know if to pass an entity or an entity ref for a parameter

congysu:
@dingeins Can we ask a contribution for this task?

dingeins:
I think we can firstly figure out which annotations are appropriate for the scenario described above.

xuzhg:
Reopen it that we should support return type annotation on Edm operation.

PR #1340, opened 2018-11-27

xuzhg:
### Issues

### Description

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/EdmModelVisitor.cs
+++ b/src/Microsoft.OData.Edm/EdmModelVisitor.cs
@@ -843,7 +843,7 @@ namespace Microsoft.OData.Edm
         {
             if (operation.ReturnType != null)
             {
-                this.VisitTypeReference(operation.ReturnType);
+                this.VisitOperationReturnType(operation.ReturnType as IEdmOperationReturnType);
             }
 
             // Do not visit vocabularyAnnotatable because functions and operation imports are always going to be either a schema element or a container element and will be visited through those paths.
@@ -855,6 +855,12 @@ namespace Microsoft.OData.Edm
             this.ProcessVocabularyAnnotatable(parameter);
             this.ProcessNamedElement(parameter);
             this.VisitTypeReference(parameter.Type);
+        }
+
+        public virtual void VisitOperationReturnType(IEdmOperationReturnType returnType)
+        {
+            this.ProcessVocabularyAnnotatable(returnType);
+            this.VisitTypeReference(returnType.ReturnType);
         }
 
         #endregion
--- a/src/Microsoft.OData.Edm/EdmModelVisitor.cs
+++ b/src/Microsoft.OData.Edm/EdmModelVisitor.cs
@@ -843,7 +843,15 @@ namespace Microsoft.OData.Edm
         {
             if (operation.ReturnType != null)
             {
-                this.VisitOperationReturnType(operation.ReturnType as IEdmOperationReturnType);
+                IEdmOperationReturnType returnType = operation.ReturnType as IEdmOperationReturnType;
+                if (returnType != null)
+                {
+                    this.VisitOperationReturnType(returnType);
+                }
+                else
+                {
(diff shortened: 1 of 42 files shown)


src/Microsoft.OData.Edm/EdmModelVisitor.cs
```diff
@@ -843,7 +843,7 @@ protected virtual void ProcessOperation(IEdmOperation operation)
         {
             if (operation.ReturnType != null)
             {
-                this.VisitTypeReference(operation.ReturnType);
+                this.VisitOperationReturnType(operation.ReturnType as IEdmOperationReturnType);
             }
```
mikepizzo: For backward compatibility, it seems like we should attempt to cast operation.ReturnType to IEdmOperationReturnType and, if that failed, just call VisitTypeReference.

src/Microsoft.OData.Edm/Schema/EdmOperation.cs
```diff
@@ -29,15 +29,15 @@ protected EdmOperation(string namespaceName, string name, IEdmTypeReference retu
         {
             EdmUtil.CheckArgumentNull(namespaceName, "namespaceName");
 
-            this.ReturnType = returnType;
+            this.ReturnType = returnType == null ? null : new EdmOperationReturnType(this, returnType);
```
mikepizzo: What if the passed returnType is already an IEdmOperationReturnType? Should we check for that here and only create a new EdmOperationReturnType if casting returnType to IEdmOperationReturnType fails?

mikepizzo:
Generally looks great. Please make sure that an external implementation of the model that returns a returntype of IEdmTypeReference (not IEdmOperationTypeReference) still works throughout (i.e., that we don't *assume* when we get the returnType that cast to IEdmOperationTypeReference will succeed).

src/Microsoft.OData.Edm/Schema/EdmOperation.cs
mikepizzo: We should probably check to see if the passed in returnType was already an IEdmOperationReturnType

mikepizzo:
Also, before creating an EdmOperationReturnType to wrap an IEdmTypeReference passed from the outside, we should see if that IEdmTypeReference is already an IEdmOperationReturnType.

src/Microsoft.OData.Edm/EdmModelVisitor.cs
xuzhg: Basically, it looks safe because we can make sure the "operation.ReturnType" will return IEdmOperationReturnType object for ever with this change. We make that in the "EdmOperation" constructor method.

But,  it's also ok to keep backward compatibility to check the object type first. So, #Resolved.

src/Microsoft.OData.Edm/Schema/EdmOperation.cs
xuzhg: I thought the code itself ok if the input "returnType" is already an "IEdmOperationReturnType"?
Because:

1. Supposed the "returnType" is from the operation itself. That's not a valid case. Because, we can't create an "EdmOperationReturnType" ahead of its declaring "EdmOperation", because, to create a "EdmOperationReturnType" should have the "EdmOperation" created first (see EdmOperationReturnType constructor). So, that's a conflict. 

2. Supposed the "returnType" is from other operation. So, the above code is ok. Because, we create a new "EdmOperationReturnType" and point the declaring operation to "this" and use the input "typeReference" as the type of the return type.

So, i want to keep the code unchanged. Please share your thoughts.

xuzhg:
@mikepizzo I sent out a new PR at #1360 based this PR, in the new PR, i implemented based on our discussion. In order to compare between the new change and old change, i want to keep the old open for reference.  Will close it once i merge the new PR. Please take a look for the new PR #1360, thanks.

xuzhg:
Close it because #1360 is out for review.

Issue #1344, opened 2018-11-29
Support expand transformation in $apply


kosinsky:
Aggregation extension specification (http://docs.oasis-open.org/odata/odata-data-aggregation-ext/v4.0/cs02/odata-data-aggregation-ext-v4.0-cs02.html#_Toc435016586) defines expand transformation

**3.12 Transformation expand **
The expand transformation takes a navigation property path that could also be passed as a $expand system query option as its first parameter. The optional second parameter can be either a filter transformation that will be applied to the related entities or an expand transformation. An arbitrary number of expand transformations can be passed as additional parameters to achieve multi-level expansion.
The result set is the input set with the specified navigation property expanded according to the specified expand options.

xuzhg:
Close it as merged.

PR #1345, opened 2018-11-29
expand transformation in $apply

kosinsky:
### Issues

*This pull request fixes issue #1344.*

### Description

Added parsing and binding for expand inside $apply.
For example $apply=expand(Sales, filter(Amount gt 10), expand(Customers))

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/Microsoft.OData.Core.txt
+++ b/src/Microsoft.OData.Core/Microsoft.OData.Core.txt
@@ -687,6 +687,7 @@ UriQueryExpressionParser_WithExpected='with' expected at position {0} in '{1}'.
 UriQueryExpressionParser_UnrecognizedWithMethod=Unrecognized with '{0}' at '{1}' in '{2}'.
 UriQueryExpressionParser_PropertyPathExpected=Expression expected at position {0} in '{1}'.
 UriQueryExpressionParser_KeywordOrIdentifierExpected='{0}' expected at position {1} in '{2}'.
+UriQueryExpressionParser_InnerMostExpandRequireFilter=inner most expand require filter() at position {0} in '{1}'.
 
 UriQueryPathParser_RequestUriDoesNotHaveTheCorrectBaseUri=The URI '{0}' is not valid because it is not based on '{1}'.
 UriQueryPathParser_SyntaxError=Bad Request: there was an error in the query syntax.
--- a/src/Microsoft.OData.Core/Microsoft.OData.Core.txt
+++ b/src/Microsoft.OData.Core/Microsoft.OData.Core.txt
@@ -687,7 +687,7 @@ UriQueryExpressionParser_WithExpected='with' expected at position {0} in '{1}'.
 UriQueryExpressionParser_UnrecognizedWithMethod=Unrecognized with '{0}' at '{1}' in '{2}'.
 UriQueryExpressionParser_PropertyPathExpected=Expression expected at position {0} in '{1}'.
 UriQueryExpressionParser_KeywordOrIdentifierExpected='{0}' expected at position {1} in '{2}'.
-UriQueryExpressionParser_InnerMostExpandRequireFilter=inner most expand require filter() at position {0} in '{1}'.
+UriQueryExpressionParser_InnerMostExpandRequireFilter=The inner most expand transformation requires a filter transformation at position {0} in '{1}'.
 
 UriQueryPathParser_RequestUriDoesNotHaveTheCorrectBaseUri=The URI '{0}' is not valid because it is not based on '{1}'.
 UriQueryPathParser_SyntaxError=Bad Request: there was an error in the query syntax.
--- a/src/Microsoft.OData.Core/Uri/ApplyClauseToStringBuilder.cs
+++ b/src/Microsoft.OData.Core/Uri/ApplyClauseToStringBuilder.cs
@@ -157,8 +157,15 @@ namespace Microsoft.OData
         private void Translate(ExpandTransformationNode transformation)
         {
             ExpandedNavigationSelectItem expandedNavigation = transformation.ExpandClause.SelectedItems.Single() as ExpandedNavigationSelectItem;
+            AppendExpandExpression(expandedNavigation);
+        }
+
+        private void AppendExpandExpression(ExpandedNavigationSelectItem expandedNavigation)
+        {
             AppendExpression(expandedNavigation.PathToNavigationProperty);
-            if (expandedNavigation.FilterOption != null || expandedNavigation.SelectAndExpand != null)
+
+            // Append filter
+            if (expandedNavigation.FilterOption != null)
             {
                 AppendComma(true);
                 if (expandedNavigation.FilterOption != null)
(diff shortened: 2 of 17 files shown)


mikepizzo:
@kosinsky; note that the syntax for the expand transformation was refined in June to require at least two parameters, with the innermost requiring a filter.  See https://issues.oasis-open.org/browse/ODATA-1068.

In your example, the third parameter (the second expand) should not be required (or allowed).

mikepizzo:
:thumbsup: 

---
In reply to: [444977019](https://github.com/OData/odata.net/pull/1345#issuecomment-444977019) [](ancestors = 444977019)

src/Microsoft.OData.Core/Uri/ApplyClauseToStringBuilder.cs
```diff
@@ -148,6 +154,24 @@ private void Translate(ComputeTransformationNode transformation)
...
+            {
+                AppendComma(true);
+                if (expandedNavigation.FilterOption != null)
+                {
+                    query.Append(ExpressionConstants.SymbolEscapedSpace);
+                    query.Append(ExpressionConstants.KeywordFilter);
+                    query.Append(ExpressionConstants.SymbolOpenParen);
+                    AppendExpression(expandedNavigation.FilterOption.Expression);
+                    query.Append(ExpressionConstants.SymbolClosedParen);
+                }
```
mikepizzo: Please check this logic; it appears that, if expandedNavigation.FilterOption is null but expandedNavigation.SelectAndExpand is not null, we'll append a comma but nothing else. Otherwise, if expandedNavigation.FilterOption is non-null, we'll append the comma, followed by the parens expression and no trailing comma. Is this intended?

src/Microsoft.OData.Core/Microsoft.OData.Core.txt
```diff
@@ -687,6 +687,7 @@ UriQueryExpressionParser_WithExpected='with' expected at position {0} in '{1}'.
 UriQueryExpressionParser_UnrecognizedWithMethod=Unrecognized with '{0}' at '{1}' in '{2}'.
 UriQueryExpressionParser_PropertyPathExpected=Expression expected at position {0} in '{1}'.
 UriQueryExpressionParser_KeywordOrIdentifierExpected='{0}' expected at position {1} in '{2}'.
+UriQueryExpressionParser_InnerMostExpandRequireFilter=inner most expand require filter() at position {0} in '{1}'.
```
mikepizzo: >inner most expand require  [](start = 54, length = 26)

(minor editorial) should start with a capital letter. Also, wording suggestion:
 The inner most expand transformation requires a filter transformation at position {0} in '{1}'.

mikepizzo:
Please verify logic for null FilterOption and non-null SelectAndExpand, and see my comments on the error message. Otherwise, looks good.

src/Microsoft.OData.Core/Uri/ApplyClauseToStringBuilder.cs
kosinsky: Fixed. Also added support for nested expands (I missed that initially)

src/Microsoft.OData.Core/Uri/ApplyClauseToStringBuilder.cs
```diff
@@ -148,6 +154,45 @@ private void Translate(ComputeTransformationNode transformation)
...
+        private void AppendExpandExpression(ExpandedNavigationSelectItem expandedNavigation)
+        {
+            AppendExpression(expandedNavigation.PathToNavigationProperty);
+
+            // Append filter
+            if (expandedNavigation.FilterOption != null)
+            {
+                AppendComma(true);
+                if (expandedNavigation.FilterOption != null)
+                {
```
mikepizzo: Can get rid of this inner check now, right? Since you broke out the SelectAndExpand case it's redundant with the outer conditional.

mikepizzo:
I think we have an extra conditional on line 171 of ApplyClauseToStringBinder.cs; otherwise looks good!

src/Microsoft.OData.Core/Uri/ApplyClauseToStringBuilder.cs
```diff
@@ -148,6 +154,45 @@ private void Translate(ComputeTransformationNode transformation)
...
+
+        private void AppendExpandExpression(ExpandedNavigationSelectItem expandedNavigation)
+        {
+            AppendExpression(expandedNavigation.PathToNavigationProperty);
+
+            // Append filter
+            if (expandedNavigation.FilterOption != null)
+            {
+                AppendComma(true);
+                if (expandedNavigation.FilterOption != null)
```
mikepizzo: Can get rid of this inner check now, right? Since you broke out the SelectAndExpand case it's redundant with the outer conditional.

src/Microsoft.OData.Core/UriParser/Aggregation/ApplyBinder.cs
```diff
@@ -19,14 +19,18 @@ internal sealed class ApplyBinder
         private BindingState state;
 
         private FilterBinder filterBinder;
+        private ODataUriParserConfiguration configuration;
+        private ODataPathInfo odataPathInfo;
 
         private IEnumerable<AggregateExpressionBase> aggregateExpressionsCache;
 
-        public ApplyBinder(MetadataBinder.QueryTokenVisitor bindMethod, BindingState state)
+        public ApplyBinder(MetadataBinder.QueryTokenVisitor bindMethod, BindingState state, ODataUriParserConfiguration configuration = null, ODataPathInfo odataPathInfo = null)
```
xuzhg: >ODataUriParserConfiguration configuration = null, ODataPathInfo odataPathInfo = null [](start = 92, length = 84)

maybe it's better to create a new Constructor

src/Microsoft.OData.Core/UriParser/Aggregation/ExpandTransformationNode.cs
```diff
@@ -0,0 +1,51 @@
...
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.UriParser.Aggregation
+{
+    using Microsoft.OData.UriParser;
+
+    /// <summary>
+    /// Node representing a filter transformation.
```
xuzhg: >filter [](start = 28, length = 6)

filter => expand?

src/Microsoft.OData.Core/UriParser/Aggregation/ExpandTransformationNode.cs
```diff
@@ -0,0 +1,51 @@
...
+    /// </summary>
+    public sealed class ExpandTransformationNode : TransformationNode
+    {
+        private readonly SelectExpandClause expandClause;
+
+        /// <summary>
+        /// Create a ExpandTransformationNode.
+        /// </summary>
+        /// <param name="filterClause">A <see cref="SelectExpandClause"/> representing the metadata bound expand expression.</param>
```
xuzhg: >filterClause [](start = 25, length = 12)

It should be "expandClause" ?

PR #1346, opened 2018-11-29
Resource value build test

xuzhg:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/ODataResource.cs
```diff
@@ -391,5 +396,27 @@ public void AddFunction(ODataFunction function)
...
+                ODataCollectionValue collection = null;
+                foreach (var property in properties)
+                {
+                    if (property.Value is ODataResourceValue)
+                    {
+                        throw new ODataException(Strings.ODataResource_PropertyValueCannotBeODataResourceValue(property.Name));
+                    }
+                    else if ((collection = property.Value as ODataCollectionValue) != null)
+                    {
+                        if (collection.Items != null && collection.Items != null && collection.Items.Any(t => t is ODataResourceValue))
```
mikepizzo: >collection.Items != null [](start = 56, length = 24)

You check this twice -- you really don't want collection.Items to be null! :-)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1468,15 +1470,192 @@ private object ReadEnumValue(bool insideJsonObjectValue, IEdmEnumTypeReference e
...
+        /// <param name="structuredTypeReference">The expected type reference of the value.</param>
+        /// <param name="payloadTypeName">The type name read from the payload.</param>
+        /// <param name="propertyAndAnnotationCollector">The duplicate property names checker.</param>
+        /// <returns>The value of the resource value.</returns>
+        /// <remarks>
+        /// Pre-Condition:  JsonNodeType.Property - the first property of the resource value object, or the second one if the first one was odata.type.
+        ///                 JsonNodeType.EndObject - the end object of the resource value object.
+        /// Post-Condition: almost anything - the node after the resource value (after the EndObject)
+        /// </remarks>
+        private ODataResourceValue ReadResourceValue(
```
mikepizzo: >private ODataResourceValue ReadResourceValue [](start = 8, length = 44)

 This is the duplicated code that concerns me. What changes to regular resource would need to be made here as well?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -389,6 +396,30 @@ private void WriteStreamReferenceProperty(string propertyName, ODataStreamRefere
...
+        /// </summary>
+        /// <param name="property">The property to write out.</param>
+        /// <param name="resourceValue">The resource value to be written</param>
+        /// <param name="isOpenPropertyType">If the property is open.</param>
+        private void WriteResourceProperty(
+            ODataProperty property,
+            ODataResourceValue resourceValue,
+            bool isOpenPropertyType)
+        {
+            if (!this.currentPropertyInfo.IsTopLevel)
```
mikepizzo: would an ODataResourceValue ever be toplevel?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
```diff
@@ -75,6 +97,87 @@ public virtual void WriteNullValue()
...
+        /// <param name="isTopLevel">true when writing a top-level property; false for nested properties.</param>
+        /// <param name="isOpenPropertyType">true if the type name belongs to an open (dynamic) property.</param>
+        /// <param name="duplicatePropertyNamesChecker">The checker instance for duplicate property names.</param>
+        /// <remarks>The current recursion depth should be a value, measured by the number of resource and collection values between
+        /// this resource value and the top-level payload, not including this one.</remarks>
+        [SuppressMessage("Microsoft.Naming", "CA2204:LiteralsShouldBeSpelledCorrectly", Justification = "Names are correct. String can't be localized after string freeze.")]
+        public virtual void WriteResourceValue(
+            ODataResourceValue resourceValue,
+            IEdmTypeReference metadataTypeReference,
+            bool isTopLevel,
```
mikepizzo: is an ODataResource ever top level?

src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
```diff
@@ -61,6 +61,22 @@ internal static string ConvertToUriEnumLiteral(ODataEnumValue value, ODataVersio
             return string.Format(CultureInfo.InvariantCulture, "{0}'{1}'", value.TypeName, value.Value);
         }
 
+        /// <summary>
+        /// Converts the given string <paramref name="value"/> to an ODataResourceValue and returns it.
+        /// Tries in both JSON light and Verbose JSON.
```
mikepizzo: We should remove support for Verbose JSON (and certainly not advertise it in documentation)

src/Microsoft.OData.Core/ODataResource.cs
xuzhg: copied and pasted in the wrong place. Thanks for catching

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
xuzhg: We don't limit end user to create a property with value as ODataResourceValue. 
In my opinion, We should not limit end user to write a top level property with value as "ODataResourceValue".

src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
xuzhg: Same.  We don't limit end user to create a property with value as ODataResourceValue.
In my opinion, We should not limit end user to write a top level property with value as "ODataResourceValue".

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
xuzhg: I had your concerns, but I think they are two different using scenarios. 
Basically, reading a value is very low level process, if we delegate the value reading to high level, in return, the high level will also call the low level process again. So, I am worry about the feasibility.

Of course, I did experiment using the top-level `ODataReader`. For example as below:

```C#
private ODataResourceValue ReadResourceValue(...)
{
   ODataReader reader = this.JsonLightInputContext.CreateUriParameterResourceReader(null, 
   structuredTypeReference == null ? null : structuredTypeReference.StructuredDefinition());
   return ReadResourceValue(reader);
 }

private static ODataResourceValue ReadResourceValue(ODataReader reader)
{
    while (reader.Read())
    {
         switch (reader.State)
          {
               case ODataReaderState.ResourceStart:
                   ....
          }
   }
}
```
but, So far, i can't make the above to work.

I think we are on the same page that we should have a method named "ReadResourceValue", the concern is about how to implement this private method.

Maybe we can postpone this detail implementation and focus on the step1 (adding class) and step2 (writing), hope we are on the same page about step1 and step2.

mikepizzo: If we have the right interface, we can optimize implementation later.

---
In reply to: [241615554](https://github.com/OData/odata.net/pull/1346#discussion_r241615554) [](ancestors = 241615554)

xuzhg:
@mikepizzo I added the validation for the top level property.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
mikepizzo: Now that we've prohibited writing resource values at the top level, can we get rid of this code?

---
In reply to: [241587501](https://github.com/OData/odata.net/pull/1346#discussion_r241587501) [](ancestors = 241587501)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
mikepizzo: Now that we prohibit writing a resource value at the top level, can we get rid of this code that checks to see if it's top level?

---
In reply to: [241587425](https://github.com/OData/odata.net/pull/1346#discussion_r241587425) [](ancestors = 241587425)

mikepizzo:
Code to handle validation looks good, but the methods that we call still seem to have code to handle writing a resource value at the top level; can we now safely remove that code?

---
In reply to: [448054198](https://github.com/OData/odata.net/pull/1346#issuecomment-448054198) [](ancestors = 448054198)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
xuzhg: @mikepizzo removed. #Resolved.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
xuzhg: @mikepizzo removed. #Resolved.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
ve9u: > In my opinion, We should not limit end user to write a top level property with value as "ODataResourceValue". 

Hi @xuzhg, we are heavily depending on creating "ODataResourceValue" at top level, but after upgrading to 7.6.4 we are unable to perform this operation. 

Our example URL format will look like below.

**URL**: root-url/Mailbox('4f89d8bc-4aaa-4aec-a459-5b5a7335b955')/Exchange.GetMailboxStatistics()

Here, the idea is Exchange.GetMailboxStatistics() operational segment should return MailboxStatistics, ODataResourceValue at top level. 

Our mail goal is not to change the URL format as lot of customers are rely on mentioned format.

Any suggested work around for this?  Please suggest me if I need to create an issue for this.

Thanks, 
Venu.

xuzhg: @ve9u Why don't use binding (deserializer) to accept the top level complex ?

ve9u: Hi @xuzhg, Thanks for your help on this. writing the complex entity same as the writing entity set solves my issue. Later I found [this ](https://docs.microsoft.com/en-us/odata/odatalib/merge-complex-and-entity)blog post which explains the same thing.

PR #1348, opened 2018-12-03
Add custom headers while code generation

paule96:
### Issues

*This pull request fixes issue #xxx.*
#1315 

### Description

Add support to add custom headers for the code generation. Now you can set the property `CustomHttpHeaders` in the `Configuration` class with your custom headers. The code generation use this for the HTTP requests to your OData API.

### Checklist (Uncheck if it is not completed)

- [ ] *Test cases added*
- [ ] *Build and test with one-click build and test script passed*

### Additional work necessary

*If documentation update is needed, please add "Docs Needed" label to the issue and provide details about the required document change in the issue.*

PR #1356, opened 2018-12-09
add vscode support

paule96:
### Issues

This have no Issue.

### Description

This add the build assets to hit F5 in vscode too.
This comes from the PR #1348 and has been discussed with @mikepizzo .

- [ ] *Test cases added*

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,28 @@
+{
+   // Use IntelliSense to find out which attributes exist for C# debugging
+   // Use hover for the description of the existing attributes
+   // For further information visit https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger-launchjson.md
+   "version": "0.2.0",
+   "configurations": [
+        {
+            "name": ".NET Core Launch (console)",
+            "type": "coreclr",
+            "request": "launch",
+            "preLaunchTask": "build",
+            // If you have changed target frameworks, make sure to update the program path.
+            "program": "${workspaceFolder}/test/FunctionalTests/Microsoft.OData.Core.Tests/bin/Debug/netcoreapp1.0/Microsoft.OData.Core.Tests.dll",
+            "args": [],
+            "cwd": "${workspaceFolder}/test/FunctionalTests/Microsoft.OData.Core.Tests",
+            // For more information about the 'console' field, see https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger-launchjson.md#console-terminal-window
+            "console": "internalConsole",
+            "stopAtEntry": false,
+            "internalConsoleOptions": "openOnSessionStart"
+        },
+        {
+            "name": ".NET Core Attach",
+            "type": "coreclr",
+            "request": "attach",
+            "processId": "${command:pickProcess}"
+        }
+    ,]
+}
--- a/.vscode/tasks.json
+++ b/.vscode/tasks.json
@@ -1,0 +1,15 @@
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "build",
+            "command": "dotnet",
+            "type": "process",
+            "args": [
+                "build",
+                "${workspaceFolder}/test/FunctionalTests/Microsoft.OData.Core.Tests/Microsoft.OData.Core.Tests.NetCore.csproj"
+            ],
+            "problemMatcher": "$msCompile"
+        }
+    ]
+}


.vscode/launch.json
```diff
@@ -0,0 +1,28 @@
...
+   // For further information visit https://github.com/OmniSharp/omnisharp-vscode/blob/master/debugger-launchjson.md
+   "version": "0.2.0",
+   "configurations": [
+        {
+            "name": ".NET Core Launch (console)",
+            "type": "coreclr",
+            "request": "launch",
+            "preLaunchTask": "build",
+            // If you have changed target frameworks, make sure to update the program path.
+            "program": "${workspaceFolder}/test/FunctionalTests/Microsoft.OData.Core.Tests/bin/Debug/netcoreapp1.0/Microsoft.OData.Core.Tests.dll",
```
mikepizzo: should we have configurations other than core?

.vscode/tasks.json
```diff
@@ -0,0 +1,15 @@
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "build",
+            "command": "dotnet",
+            "type": "process",
+            "args": [
+                "build",
+                "${workspaceFolder}/test/FunctionalTests/Microsoft.OData.Core.Tests/Microsoft.OData.Core.Tests.NetCore.csproj"
```
mikepizzo: should we have tasks for projects other than netcore.csproj?

paule96:
@mikepizzo your are right. I doesn't check what vscode has created. (in my normal projects it's works)
I'm sorry for that. 

I have try to get this working but have some questions. There is a `build.p1` in the root of this repo. But this always fails by the build of `OData.Tests.E2E.NetCore.VS2017.sln`. 

If I look in the build log I see that the restoring of the dotnet core projects not working right. It doesn't create a `project.assets.json`.

So I search for this error in the WWW and found [this](https://developercommunity.visualstudio.com/content/problem/336885/error-netsdk1004-since-4th-september-without-code.html). This means maybe we use the wrong nuget.exe. So I looked in the `sln\.nuget` folder and check the version. The nuget version there is currently the `3.5.0.1996`. So I have tryed some nuget versions but only `3.5.0.1996` will work wih the rest of the project. So I'm not shure what todo with this.

But this is not my main problem. My main problem is that I can't run the dotnet core projects because there are everywhere `\` instead of `/`. So dotnet build fails with this error:

```bash
Illegales Zeichen im Pfad.
```

in englisch it means:

```bash
Illegal character in path
```

So if I replace all `\` with `/` in all build files (*.csproj, *.sln, *.props, *.config, *.json, *.targets) the build process works fine instead of the `OData.Tests.E2E.NetCore.VS2017.sln`. But if I now try to run `dotnet build test/FunctionalTests/Microsoft.OData.Core.Tests/Microsoft.OData.Core.Tests.NetCore.csproj` this fails with this error message:

```bash
H:\develop\odata.net\sln\packages\StyleCop.MSBuild.4.7.49.0\build\StyleCop.MSBuild.Targets(98,5): error MSB4062: Die StyleCopTask-Aufgabe konnte nicht aus der H:\develop\odata.net\sln\packages\StyleCop.MSBuild.4.7.49.0\build\..\tools\StyleCop.dll-Assembly geladen werden. Could not load file or assembly 'System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. Das System kann die angegebene Datei nicht finden. Stellen Sie sicher, dass die <UsingTask>-Deklaration korrekt ist, die Assembly und alle zugehörigen Abhängigkeiten verfügbar sind und die Aufgabe eine öffentliche Klasse enthält, die Microsoft.Build.Framework.ITask implementiert. [H:\develop\odata.net\src\Microsoft.Spatial\Microsoft.Spatial.NetStandard.VS2017.csproj]
H:\develop\odata.net\sln\packages\StyleCop.MSBuild.4.7.49.0\build\StyleCop.MSBuild.Targets(98,5): error MSB4062: Die StyleCopTask-Aufgabe konnte nicht aus der H:\develop\odata.net\sln\packages\StyleCop.MSBuild.4.7.49.0\build\..\tools\StyleCop.dll-Assembly geladen werden. Could not load file or assembly 'System.Windows.Forms, Version=2.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089'. Das System kann die angegebene Datei nicht finden. Stellen Sie sicher, dass die <UsingTask>-Deklaration korrekt ist, die Assembly und alle zugehörigen Abhängigkeiten verfügbar sind und die Aufgabe eine öffentliche Klasse enthält, die Microsoft.Build.Framework.ITask implementiert. [H:\develop\odata.net\src\Microsoft.OData.Edm\Microsoft.OData.Edm.NetStandard.VS2017.csproj]
```


so in conclusion I have to question:

1. Is it normal that build.ps1 throws an error?
2. How to build the dotnet core stuff in this project?


Befor this isn't clear I can't add the vscode support to this project for build and debug.

mikepizzo:
Thanks for the follow-up @paule96. Assigning this to @KanishManuja-MS, who is on-call this week, to follow up on the build issues.

PR #1357, opened 2018-12-11
Change date parse to date try parse

ysanghi:
### Description

When parsing the URI, ODataURI uses Date.Parse to check if the literal is a Date which throws exceptions. This can be avoided by using Date.TryParse and can be micro-optimized.

- [X] *Test cases added*


src/Microsoft.OData.Core/UriUtils.cs
```diff
@@ -4,13 +4,14 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using Microsoft.OData.Edm;
+
 using System;
```
mikepizzo: This should be moved back to immediately following the System using directives.

src/Microsoft.OData.Edm/Csdl/EdmValueParser.cs
```diff
@@ -4,10 +4,11 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using Microsoft.OData.Edm;
+
 using System;
```
mikepizzo: This should be moved back to immediately following the System using directives.

mikepizzo:
It looks like the using statements for Microsoft.OData.Edm got moved in UriUtils.cs and EdmValueParser.cs, which is going to cause our code analysis to complain.  If we can move those back to immediately following the System using statements, then the rest looks good -- Thanks!

PR #1358, opened 2018-12-11
Add the writing derived type constraint validation

xuzhg:
### Issues

*This pull request continues on Union type feature.

### Description

Enable writing derived type constraint validation on:
1. Resource
2. Nested resource info
3. Edm.PrimitiveType property

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/PropertyMetadataTypeInfo.cs
+++ b/src/Microsoft.OData.Core/PropertyMetadataTypeInfo.cs
@@ -14,7 +14,7 @@ namespace Microsoft.OData
     /// </summary>
     internal class PropertyMetadataTypeInfo
     {
-        private bool loadDerivedTypeConstraints;
+        private bool derivedTypeConstraintsLoaded;
         private IEnumerable<string> derivedTypeConstraints;
         private IEdmModel model;
 
@@ -29,7 +29,7 @@ namespace Microsoft.OData
             this.FullName = this.TypeReference == null ? null : this.TypeReference.Definition.AsActualType().FullTypeName();
 
             this.model = model;
-            loadDerivedTypeConstraints = false;
+            derivedTypeConstraintsLoaded = false;
         }
 
         public string PropertyName { get; private set; }
@@ -50,10 +50,10 @@ namespace Microsoft.OData
         {
             get
             {
-                if (!loadDerivedTypeConstraints)
+                if (!derivedTypeConstraintsLoaded)
                 {
                     this.derivedTypeConstraints = EdmProperty == null ? null : this.model.GetDerivedTypeConstraints(EdmProperty);
-                    loadDerivedTypeConstraints = true;
+                    derivedTypeConstraintsLoaded = true;
                 }
 
                 return this.derivedTypeConstraints;
--- a/src/Microsoft.OData.Core/PropertyMetadataTypeInfo.cs
+++ b/src/Microsoft.OData.Core/PropertyMetadataTypeInfo.cs
@@ -4,6 +4,7 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using System.Collections.Generic;
 using Microsoft.OData.Edm;
 
 namespace Microsoft.OData
@@ -13,7 +14,11 @@ namespace Microsoft.OData
     /// </summary>
     internal class PropertyMetadataTypeInfo
     {
-        public PropertyMetadataTypeInfo(string name, IEdmStructuredType owningType)
+        private bool loadDerivedTypeConstraints;
+        private IEnumerable<string> derivedTypeConstraints;
+        private IEdmModel model;
+
+        public PropertyMetadataTypeInfo(IEdmModel model, string name, IEdmStructuredType owningType)
         {
             this.PropertyName = name;
             this.OwningType = owningType;
@@ -22,6 +27,9 @@ namespace Microsoft.OData
             this.IsOpenProperty = (owningType != null && owningType.IsOpen && this.IsUndeclaredProperty);
             this.TypeReference = this.IsUndeclaredProperty ? null : this.EdmProperty.Type;
             this.FullName = this.TypeReference == null ? null : this.TypeReference.Definition.AsActualType().FullTypeName();
+
+            this.model = model;
+            loadDerivedTypeConstraints = false;
         }
 
         public string PropertyName { get; private set; }
@@ -37,5 +45,19 @@ namespace Microsoft.OData
         public IEdmTypeReference TypeReference { get; private set; }
 
         public string FullName { get; private set; }
+
+        public IEnumerable<string> DerivedTypeConstraints
+        {
+            get
+            {
+                if (!loadDerivedTypeConstraints)
+                {
+                    this.derivedTypeConstraints = EdmProperty == null ? null : this.model.GetDerivedTypeConstraints(EdmProperty);
+                    loadDerivedTypeConstraints = true;
+                }
+
+                return this.derivedTypeConstraints;
+            }
+        }
     }
 }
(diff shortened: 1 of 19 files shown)


src/Microsoft.OData.Core/PropertyMetadataTypeInfo.cs
```diff
@@ -13,7 +14,11 @@ namespace Microsoft.OData
     /// </summary>
     internal class PropertyMetadataTypeInfo
     {
-        public PropertyMetadataTypeInfo(string name, IEdmStructuredType owningType)
+        private bool loadDerivedTypeConstraints;
```
mikepizzo: >loadDerivedTypeConstraints [](start = 21, length = 26)

nit: sounds like directive (to load or not to load) rather than state. perhaps rename to something like "derivedTypeConstraintsLoaded"

src/Microsoft.OData.Core/WriterValidationUtils.cs
```diff
@@ -433,6 +512,52 @@ internal static void ValidateEntityReferenceLink(ODataEntityReferenceLink entity
...
+        }
+
+        /// <summary>
+        /// Validates the input <see cref="IEdmStructuredType"/> meets the derived type constaints on the <see cref="IEdmNavigationSource"/>.
+        /// </summary>
+        /// <param name="resourceType">The input resource type.</param>
+        /// <param name="metadataType">The type from metadata.</param>
+        /// <param name="navigationSource">The navigation source.</param>
+        /// <param name="derivedTypeConstraints">The derived type constraints on the nested resource.</param>
+        internal static void ValidateDerivedTypeConstraintOnNavigationSource(IEdmStructuredType resourceType, IEdmStructuredType metadataType, IEdmNavigationSource navigationSource, IEnumerable<string> derivedTypeConstraints)
```
mikepizzo: >ValidateDerivedTypeConstraintOnNavigationSource [](start = 29, length = 47)

ValidateDerivedTypeConstraintOnNavigationSource and ValidateDerivedTypeConstraintOnNestedResourceInfo have the same logic, except that one uses the name of the navigation source in the exception and the other uses the name from the nestedresourceinfo.  Instead of two overloads, unless you think you'll need the navigationsource (or additional information out of the nestedResourceInfo) in the future, how about just having a single overload that takes a string property "navigationSourceName", and pass the appropriate string from the caller?

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataWriterDerivedTypeConstraintTests.cs
```diff
@@ -0,0 +1,735 @@
...
+                writer.WriteStart(this.odataCustomerResource);
+                writer.WriteEnd();
+            };
+
+            // Singleton doesn't have the derived type constraint.
+            string customerActual = GetWriterOutput(this.edmModel, writeCustomerAction);
+            Assert.Contains("/$metadata#Me\",\"Id\":7}", customerActual);
+
+            // Singleton has the derived type constraint.
+            SetDerivedTypeAnnotation(this.edmModel, this.edmMe, "NS.VipCustomer");
```
mikepizzo: >SetDerivedTypeAnnotation [](start = 12, length = 24)

does a new instance of edmModel get created with each test case run, or is the derived type constraint set for all subsequent tests?

src/Microsoft.OData.Core/PropertyMetadataTypeInfo.cs
xuzhg: #resolved.

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataWriterDerivedTypeConstraintTests.cs
xuzhg: @mikepizzo For every single test, xUnit.net creates a new instance of the test class, so, in each test method, the "this.edmModel" is the basic edm model created in the constructor. 

Below is the description from xUnit document at: https://xunit.github.io/docs/shared-context

_**xUnit.net creates a new instance of the test class for every test that is run, so any code which is placed into the constructor of the test class will be run for every single test**. This makes the constructor a convenient place to put reusable context setup code where you want to share the code without sharing object instances (meaning, you get a clean copy of the context object(s) for every test that is run)._

mikepizzo: 👍 Thanks for the research @xuzhg!

src/Microsoft.OData.Core/WriterValidationUtils.cs
xuzhg: #resolved.

src/Microsoft.OData.Core/WriterValidationUtils.cs
```diff
@@ -314,6 +316,83 @@ internal static void ValidateStreamReferenceProperty(ODataProperty streamPropert
...
+
+            throw new ODataException(Strings.WriterValidationUtils_ValueTypeNotAllowedInDerivedTypeConstraint(fullTypeName, "property", propertySerializationInfo.PropertyName));
+        }
+
+        /// <summary>
+        /// Validates the value type of a property meets the derived type constraints.
+        /// </summary>
+        /// <param name="propertySerializationInfo">The property serialization info.</param>
+        /// <remarks>This does NOT validate the value of the property, just the type of property.</remarks>
+        internal static void ValidateCollectionPropertyDerivedTypeConstraint(ODataCollectionValue collectionValue,
```
mikepizzo: >ValidateCollectionPropertyDerivedTypeConstraint [](start = 29, length = 47)

Can this be combined with ValidatePropertyDerivedTypeConstraint? It looks like the logic is exactly the same, and that the ODataCollectionValue is not used.

src/Microsoft.OData.Core/WriterValidationUtils.cs
```diff
@@ -314,6 +316,83 @@ internal static void ValidateStreamReferenceProperty(ODataProperty streamPropert
...
+                return;
+            }
+
+            throw new ODataException(Strings.WriterValidationUtils_ValueTypeNotAllowedInDerivedTypeConstraint(fullTypeName, "property", propertySerializationInfo.PropertyName));
+        }
+
+        /// <summary>
+        /// Validates the value type of a property meets the derived type constraints.
+        /// </summary>
+        /// <param name="propertySerializationInfo">The property serialization info.</param>
```
mikepizzo: >param name [](start = 13, length = 10)

parameters in comments don't match parameters in function signature (ODataCollectionValue is missing). If we do need this overload to take an ODataCollectionValue, then we need to add that parameter to the comments.

src/Microsoft.OData.Core/WriterValidationUtils.cs
```diff
@@ -314,6 +316,83 @@ internal static void ValidateStreamReferenceProperty(ODataProperty streamPropert
...
+
+            throw new ODataException(Strings.WriterValidationUtils_PropertyValueTypeNotAllowedInDerivedTypeConstraint(fullTypeName, propertySerializationInfo.PropertyName));
+        }
+
+        /// <summary>
+        /// Validates the value type of a property meets the derived type constraints.
+        /// </summary>
+        /// <param name="propertySerializationInfo">The property serialization info.</param>
+        /// <remarks>This does NOT validate the value of the property, just the type of property.</remarks>
+        internal static void ValidateCollectionPropertyDerivedTypeConstraint(ODataCollectionValue collectionValue,
```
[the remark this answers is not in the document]
xuzhg: Sorry, forgot to clean the collection. #Resolved.

src/Microsoft.OData.Core/WriterValidationUtils.cs
xuzhg: #Resolved. That's old codes to verify the collection(Edm.PrimitiveType). Not, it's not valid. And I removed. Thanks.

PR #1286, opened 2018-10-09
Context url for expand

biaol-odata:
### Issues

*This pull request fixes issue ##1265 ($expand in context url), including the two prerequisite issues #1259 and #1268. It also includes fix for issue #1104.*
*Please note that the PR #1264 associated with issue #1259 has been superseded by this PR, which contains the changes needed. PR #1264 will be cancelled in favor of this PR.*

### Description

*Changes include the followings: 
1). Add parenthesized token for navigation property for $expand clause in context url; 
2). Context url select list parsing implementation update from ABNF V3 to ABNF V4.01 (#1268); 
3). Ensure explicit selected navigation properties are present in context url (#1104); 
4). Ensure navigation links are materialized for case of minimal metadata with expand-sub-select clause (#1259) .*

### Checklist (Uncheck if it is not completed)

- [x] *Test cases added*
- [x] *Build and test with one-click build and test script passed*

### Additional work necessary

*If documentation update is needed, please add "Docs Needed" label to the issue and provide details about the required document change in the issue.*

PR #1359, opened 2018-12-12
Fix SelectExpandNode for navigation properties on derived/complex types

mikepizzo:
### Issues

PR #1286 fixed a number of issues related to how non-selected expanded navigation properties were represented in the ContextUrl. However, in the new code that attempts to determine if the entire subtree is selected, it looks only at direct children to see if they are navigation properties and not deeper.  

For example, prior to this fix the following URL would return a SelectExpandNode with SelectionType=PartialSubtree.

             Employees?$expand=ns.Manager/DirectReports

After the PR, this is correctly returned as EntireSubtree

### Description

The primary purpose of this PR is to set whether the SelectNode is entiresubtree based on a deep traverse of child paths, rather than only looking at direct children.

Other changes, in addition to new tests, include some test clean-up as well as code optimizations around deferring creation of hashsets until required.

- [x ] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/SelectedPropertiesNode.cs
+++ b/src/Microsoft.OData.Core/SelectedPropertiesNode.cs
@@ -117,23 +117,23 @@ namespace Microsoft.OData
                 this.ParsePathSegment(segments, 0);
             }
 
-            DetermineSelectionType(this);
+            DetermineSelectionType();
         }
 
-        private static void DetermineSelectionType(SelectedPropertiesNode node)
+        private void DetermineSelectionType()
         {
-            if (node.children != null)
+            if (this.children != null)
             {
-                foreach (SelectedPropertiesNode childNode in node.children.Values)
+                foreach (SelectedPropertiesNode childNode in this.children.Values)
                 {
-                    DetermineSelectionType(childNode);
+                    childNode.DetermineSelectionType();
                 }
             }
 
-            if ((node.selectedProperties == null || node.selectedProperties.Count == 0)
-                && (node.children == null || node.children.Values.All(n => n.selectionType == SelectionType.EntireSubtree)))
+            if ((this.selectedProperties == null || this.selectedProperties.Count == 0)
+                && (this.children == null || this.children.Values.All(n => n.selectionType == SelectionType.EntireSubtree)))
             {
-                node.selectionType = SelectionType.EntireSubtree;
+                this.selectionType = SelectionType.EntireSubtree;
             }
         }
 
@@ -799,7 +799,6 @@ namespace Microsoft.OData
             {
                 return false;
             }
-
 
             return GetPossibleMatchesForSelectedOperation(operation, mustBeNamespaceQualified).Any(possibleMatch => this.selectedProperties.Contains(possibleMatch));
         }
--- a/src/Microsoft.OData.Core/SelectedPropertiesNode.cs
+++ b/src/Microsoft.OData.Core/SelectedPropertiesNode.cs
@@ -101,7 +101,7 @@ namespace Microsoft.OData
             {
                 string[] segments = null;
                 int idxLP = t.IndexOf('(');
-                if ( -1 == idxLP )
+                if (-1 == idxLP)
                 {
                     segments = t.Split(PathSeparator);
                 }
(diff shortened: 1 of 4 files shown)


src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -117,10 +117,23 @@ private void Parse(string selectClause)
                 this.ParsePathSegment(segments, 0);
             }
 
-            if ( (this.selectedProperties == null || this.selectedProperties.Count == 0)
-                && this.children.Values.All( n => n.selectionType == SelectionType.EntireSubtree))
+            DetermineSelectionType(this);
+        }
+
+        private static void DetermineSelectionType(SelectedPropertiesNode node)
```
xuzhg: >DetermineSelectionType [](start = 28, length = 22)

I'd suppose to move the private static method end of this class.

xuzhg: >SelectedPropertiesNode node [](start = 51, length = 27)

It seems we also can use instance method, not the static. But, either way is ok for me. Only to call this static method using "this" a little bit confused me.

test/FunctionalTests/Microsoft.OData.Core.Tests/SelectedPropertiesNodeTests.cs
```diff
@@ -63,8 +63,16 @@ public SelectedPropertiesNodeTests()
             this.defaultContainer.AddEntitySet("Cities", cityType);
             this.defaultContainer.AddEntitySet("Districts", districtType);
 
+            EdmComplexType airportType = new EdmComplexType("TestModel", "Airport");
```
xuzhg: >airportType [](start = 27, length = 11)

should we add "airportType" into edm model? maybe i miss that.

test/FunctionalTests/Microsoft.OData.Core.Tests/SelectedPropertiesNodeTests.cs
```diff
@@ -442,7 +462,51 @@ public void ExpandTokenForCollectionOfNavigationPropertyShouldHaveEntireSubtree(
...
+        }
+
+        [Fact]
+        public void ExpandTokenForNavigationPropertyOnDerivedTypeShouldHaveEntireSubtree()
+        {
+            SelectedPropertiesNode.Create("TestModel.Metropolis/MetropolisNavigation()", this.cityType, this.edmModel).Should().HaveEntireSubtree();
+        }
+
+        [Fact]
+        public void ExpandTokenForNavigationPropertyDerivedOnComplexShouldHaveEntireSubtree()
```
xuzhg: >DerivedOn [](start = 52, length = 9)

Maybe "....OnDerivedComplex...."

test/FunctionalTests/Microsoft.OData.Core.Tests/SelectedPropertiesNodeTests.cs
```diff
@@ -442,7 +462,51 @@ public void ExpandTokenForCollectionOfNavigationPropertyShouldHaveEntireSubtree(
...
+        }
+
+        [Fact]
+        public void ExpandTokenForNavigationPropertyOnComplexShouldHaveEntireSubtree()
+        {
+            SelectedPropertiesNode.Create("NearestAirport/City()", this.metropolisType, this.edmModel).Should().HaveEntireSubtree();
+        }
+
+        [Fact]
+        public void ExpandTokenForNavigationPropertyDerivedOnComplexWithSelectShouldHavePartialSubtree()
```
xuzhg: >DerivedOnComplex [](start = 52, length = 16)

Maybe "...OnDerivedComplex..."

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -791,16 +790,17 @@ private bool IsOperationSelectedAtThisLevel(IEdmOperation operation, bool mustBe
...
             }
 
-            if (this.selectionType == SelectionType.EntireSubtree)
+            if (this.selectionType == SelectionType.Empty || this.selectedProperties == null )
             {
-                return true;
+                return false;
             }
 
+
```
xuzhg: remove the un-used new line

xuzhg:
The change looks good to me,  but left some small comments. Thanks

test/FunctionalTests/Microsoft.OData.Core.Tests/SelectedPropertiesNodeTests.cs
mikepizzo: This and regionalAirportType seemed to be implicitly added to the model, but I made them explicit.

---
In reply to: [242373747](https://github.com/OData/odata.net/pull/1359#discussion_r242373747) [](ancestors = 242373747)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
mikepizzo: Yes, I like that better.

---
In reply to: [242370658](https://github.com/OData/odata.net/pull/1359#discussion_r242370658) [](ancestors = 242370658)

PR #1360, opened 2018-12-13

xuzhg:
### Issues

### Description

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlOperationReturnType.cs
```diff
@@ -4,26 +4,31 @@
...
-            this.returnType = returnType;
+            this.ReturnType = returnType;
+        }
+
+        public CsdlTypeReference ReturnType { get; }
+
+        public override void AddAnnotation(CsdlDirectValueAnnotation annotation)
+        {
+            base.AddAnnotation(annotation);
+            ReturnType.AddAnnotation(annotation);
```
mikepizzo: Is this right? Should we be annotating both the OperationReturnType and the CsdlTypeReference?  I would have expected that we only annotate the CsdlOperationReturnType...

src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlOperationReturnType.cs
```diff
@@ -4,26 +4,31 @@
...
         }
 
-        public CsdlTypeReference ReturnType
+        public override void AddAnnotation(CsdlAnnotation annotation)
         {
-            get { return this.returnType; }
+            base.AddAnnotation(annotation);
+            ReturnType.AddAnnotation(annotation);
```
mikepizzo: As above; should we be annotating *both* the CsdlTypeReference and the CsdlOperationReturnType?

src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlOperationReturnType.cs
xuzhg: I have to do like this, and I double check that's ok. 
The reason is that:

1. Only the "CsdlTypeReference" will be used to constructor the "CsdlSemanticsOperation....ReturnType"'.

2. CsdlOperationReturnType is only a middle wrapper class, it's not used in later to constructor the "CsdlSemanticsOperation....ReturnType"'.

src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlOperationReturnType.cs
xuzhg: Basically, we should annotate "CsdlOperationReturnType" only, however, we use the "CsdlTypeReference returnType" to constructor the other ""CsdlSemanticsOperation....ReturnType"'. 

So, I have to do like this. Of course, we can use "base.AddAnnotation(annotation), but for safety, it 's better to keep the base call.

xuzhg:
Close this because the #1414 enables this functionality. Thanks.

Issue #1373, opened 2019-01-08
IN operator not working with null value on nullable properties


sebastienrob:
An NullReferenceException is thrown while parsing a filter expression with the IN operator and a null value for a nullable property.

The problem seems to be related to this line of code:
https://github.com/OData/odata.net/blob/eb0fc9a1e5a7ef58de205f99818e446b6e5c85cd/src/Microsoft.OData.Core/UriParser/SemanticAst/CollectionConstantNode.cs#L57
If item is NULL, then the call to ToString() will fail with a NullReferenceException.

### Assemblies affected

ODataLib 7.5.3

### Reproduce steps

Sample Model
```c#
public class TestEntity
{
   [Key]
   public int PropertyKey { get; set; }

   public int PropertyInt { get; set; }

   public int? NullablePropertyInt { get; set; }
}
```
Sample query
```
http://host/service/TestEntity?$filter=NullablePropertyInt in (1,2,null)
```
### Expected result
The list of all the TestEntity with 1, 2 or NULL as the value of the property NullablePropertyInt.

### Actual result

An NullReferenceException is thrown:
```
System.NullReferenceException: Object reference not set to an instance of an object.
   at Microsoft.OData.UriParser.CollectionConstantNode..ctor(IEnumerable`1 objectCollection, String literalText, IEdmCollectionTypeReference collectionType)
   at Microsoft.OData.UriParser.LiteralBinder.BindLiteral(LiteralToken literalToken)
   at Microsoft.OData.UriParser.MetadataBinder.Bind(QueryToken token)
   at Microsoft.OData.UriParser.InBinder.GetCollectionOperandFromToken(QueryToken queryToken, IEdmTypeReference expectedType, IEdmModel model)
   at Microsoft.OData.UriParser.InBinder.BindInOperator(InToken inToken, BindingState state)
   at Microsoft.OData.UriParser.MetadataBinder.Bind(QueryToken token)
   at Microsoft.OData.UriParser.FilterBinder.BindFilter(QueryToken filter)
   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseFilterImplementation(String filter, ODataUriParserConfiguration configuration, ODataPathInfo odataPathInfo)
   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseFilter()
```

### Additional detail
if we try the same query on the non nullable property, we are receiving an ODataException explaining the NULL value is not allowed, which is correct.

```
[...]

PR #1375, opened 2019-01-10
Allow null values in constant collections

KanishManuja-MS:
### Issues

*This pull request fixes issue #1373 .*

### Description

CollectionConstantNode did not account for null values in the collection even if the item type is nullable. 
 
- [x] *Test cases added*

### Additional work necessary
I would like to test the nightly build against a sample WebAPI service to ensure that the filter works correctly all the way through for constant collections with null values in it.

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/SemanticAst/CollectionConstantNode.cs
+++ b/src/Microsoft.OData.Core/UriParser/SemanticAst/CollectionConstantNode.cs
@@ -54,7 +54,7 @@ namespace Microsoft.OData.UriParser
 
             foreach (object item in objectCollection)
             {
-                this.collection.Add(new ConstantNode(item, item.ToString(), this.itemType));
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/UriParser/SemanticAst/CollectionConstantNode.cs
```diff
@@ -54,7 +54,7 @@ public CollectionConstantNode(IEnumerable<object> objectCollection, string liter
 
             foreach (object item in objectCollection)
             {
-                this.collection.Add(new ConstantNode(item, item.ToString(), this.itemType));
+                this.collection.Add(new ConstantNode(item, item != null ? item.ToString() : "null", this.itemType));
```
xuzhg: May we replace the hard code "null" as other code?

PR #1376, opened 2019-01-11
Enable reading validation for derived type constraint annotation

xuzhg:
### Issues

*This pull request fixes enable the resource reading derived type constraint annotation validation.*

### Description

*Enable the resource reading derived type constraint annotation validation.*

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/DerivedTypeValidator.cs
+++ b/src/Microsoft.OData.Core/DerivedTypeValidator.cs
@@ -1,0 +1,76 @@
+//---------------------------------------------------------------------
+// <copyright file="DerivedTypeValidator.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData
+{
+    using System.Collections.Generic;
+    using System.Linq;
+    using Microsoft.OData.Edm;
+
+    /// <summary>
+    ///  Helper class to verify the resource type meets the drived type contraints.
+    /// </summary>
+    internal sealed class DerivedTypeValidator
+    {
+        private IEnumerable<string> derivedTypeConstraints;
+        private IEdmType expectedType;
+        private string resourceKind;
+        private string resourceName;
+
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="expectedType">The exptected type.</param>
+        /// <param name="derivedTypeConstraints">The derived type contraints.</param>
+        /// <param name="resourceKind">The resource type, be used at error message.</param>
+        /// <param name="resourceName">The resource name, be used at error message.</param>
+        public DerivedTypeValidator(IEdmType expectedType, IEnumerable<string> derivedTypeConstraints, string resourceKind, string resourceName)
+        {
+            this.derivedTypeConstraints = derivedTypeConstraints;
+            this.expectedType = expectedType;
+            this.resourceKind = resourceKind;
+            this.resourceName = resourceName;
+        }
+
+        /// <summary>
+        /// Validates the type of a resource.
+        /// </summary>
+        /// <param name="resourceType">The type of the resource.</param>
+        internal void ValidateResourceType(IEdmType resourceType)
+        {
+            if (resourceType == null)
+            {
+                return;
+            }
+
+            ValidateResourceType(resourceType.FullTypeName());
+        }
+
+        /// <summary>
+        /// Validates the full type name of a resource.
+        /// </summary>
+        /// <param name="resourceTypeName">The full type name of the resource.</param>
+        internal void ValidateResourceType(string resourceTypeName)
+        {
+            if (resourceTypeName == null)
+            {
+                return;
+            }
+
+            if (this.expectedType != null && this.expectedType.FullTypeName() == resourceTypeName)
+            {
+                return;
+            }
+
+            if (derivedTypeConstraints == null || derivedTypeConstraints.Any(c => c == resourceTypeName))
+            {
+                return;
+            }
+
+            throw new ODataException(Strings.ReaderValidationUtils_ValueTypeNotAllowedInDerivedTypeConstraint(resourceTypeName, resourceKind, resourceName));
+        }
+    }
+}
--- a/src/Microsoft.OData.Core/DerivedTypeValidator.cs
+++ b/src/Microsoft.OData.Core/DerivedTypeValidator.cs
@@ -23,7 +23,7 @@ namespace Microsoft.OData
         /// <summary>
         /// Constructor.
         /// </summary>
-        /// <param name="expectedType">The exptected type.</param>
+        /// <param name="expectedType">The expected type.</param>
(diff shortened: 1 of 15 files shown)


src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -871,6 +871,15 @@ in propertyAndAnnotationCollector.GetODataPropertyAnnotations(propertyName))
                 propertyTypeName = (string)propertyAnnotation.Value;
             }
 
+            if (propertyTypeName != null)
+            {
+                var validater = propertyAndAnnotationCollector.GetDerivedTypeValidator(propertyName);
```
KanishManuja-MS: >validater [](start = 20, length = 9)

It should be validator.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -1889,6 +1899,12 @@ private void StartReadingResource()
             // Resolve the type name
             this.ApplyResourceTypeNameFromPayload(currentResource.TypeName);
 
+            // Validate type with derived type validator if available
+            if (this.CurrentDerivedTypeValidator != null && !(this.ReadingDelta && this.CurrentResourceDepth == 0))
```
mikepizzo: >this.ReadingDelta [](start = 62, length = 17)

why don't we validate derived type for delta?

src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -338,6 +338,18 @@ protected ResourceSetWithoutExpectedTypeValidator CurrentResourceSetValidator
...
 
+        /// <summary>
+        /// Validator to validate the derived type constraint.
+        /// </summary>
+        protected DerivedTypeValidator CurrentDerivedTypeValidator
+        {
+            get
+            {
+                Debug.Assert(this.State == ODataReaderState.ResourceStart, "CurrentDerivedTypeValidator should only be called while reading a resource.");
+                return this.ParentScope == null ? null : this.ParentScope.DerivedTypeValidator;
```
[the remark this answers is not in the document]
xuzhg: 1.  We do the resource type validation when we are reading the resource itself.
2. The "@odata.type" is read at the ResourceStart, we should validate it as early as possible.
3. The derived type annotation is from either top level navigation source, or the nested resource (nested property). For all the resource at a certain navigation source, we use one validator.

src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -513,6 +525,11 @@ protected void EnterScope(Scope scope)
                 scope.ResourceTypeValidator = new ResourceSetWithoutExpectedTypeValidator(scope.ResourceType);
             }
 
+            if (scope.State == ODataReaderState.ResourceSetStart)
```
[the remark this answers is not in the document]
xuzhg: I need more discussion about the delta scenario.

src/Microsoft.OData.Core/DerivedTypeValidator.cs
```diff
@@ -0,0 +1,76 @@
...
+    {
+        private IEnumerable<string> derivedTypeConstraints;
+        private IEdmType expectedType;
+        private string resourceKind;
+        private string resourceName;
+
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="expectedType">The exptected type.</param>
```
mikepizzo: >exptected [](start = 43, length = 9)

nit: spelling ("expected")

src/Microsoft.OData.Core/DerivedTypeValidator.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// </summary>
+        /// <param name="resourceTypeName">The full type name of the resource.</param>
+        internal void ValidateResourceType(string resourceTypeName)
+        {
+            if (resourceTypeName == null)
+            {
+                return;
+            }
+
+            if (this.expectedType != null && this.expectedType.FullTypeName() == resourceTypeName)
```
mikepizzo: >expectedType [](start = 21, length = 12)

Do we have cases where expectedType is null?

src/Microsoft.OData.Core/DerivedTypeValidator.cs
```diff
@@ -0,0 +1,76 @@
...
+            ValidateResourceType(resourceType.FullTypeName());
+        }
+
+        /// <summary>
+        /// Validates the full type name of a resource.
+        /// </summary>
+        /// <param name="resourceTypeName">The full type name of the resource.</param>
+        internal void ValidateResourceType(string resourceTypeName)
+        {
+            if (resourceTypeName == null)
```
mikepizzo: >resourceTypeName [](start = 16, length = 16)

Do we have cases where we would call this with a null resourceTypeName?

src/Microsoft.OData.Core/DerivedTypeValidator.cs
xuzhg: #Resolved.

src/Microsoft.OData.Core/DerivedTypeValidator.cs
xuzhg: At top level , ODL supports to create reader without the expected resource type. for example, see https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/ODataMessageReader.cs#L316-L319:

        /// <summary>Creates an <see cref="T:Microsoft.OData.ODataReader" /> to read a resource set.</summary>
        /// <returns>The created reader.</returns>
        public ODataReader CreateODataResourceSetReader()
        {
            return this.CreateODataResourceSetReader(/*entitySet*/null, /*expectedResourceType*/null);
        }

mikepizzo:
A couple questions for you inline, but I'm good with this.

src/Microsoft.OData.Core/DerivedTypeValidator.cs
xuzhg: At top level , ODL supports to create reader without the expected resource type. for example, see https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/ODataMessageReader.cs#L316-L319:

        /// <summary>Creates an <see cref="T:Microsoft.OData.ODataReader" /> to read a resource set.</summary>
        /// <returns>The created reader.</returns>
        public ODataReader CreateODataResourceSetReader()
        {
            return this.CreateODataResourceSetReader(/*entitySet*/null, /*expectedResourceType*/null);
        }

src/Microsoft.OData.Core/DerivedTypeValidator.cs
mikepizzo: :thumbsup: 

---
In reply to: [252894010](https://github.com/OData/odata.net/pull/1376#discussion_r252894010) [](ancestors = 252894010)

mikepizzo:
Looks good to me; thanks for the follow-up!

Issue #1385, opened 2019-01-25
Support square brackets with in operator


mikepizzo:
When we added the IN operator, we specified the use of parens to make it more similar, i.e., to the SQL in operator:

   name in ('a','b','c')

However, due to possible ambiguity, OASIS has changed the syntax to use square brackets:

   name in ['a','b','c']

We should accept square brackets with the in operator, and generate square brackets in building URLs. 

### Assemblies affected

ODL

KanishManuja-MS:
@mikepizzo Wouldn't this be a breaking change if we replace the brackets for the generation as well?

KanishManuja-MS:
@xuzhg @mikepizzo #1164 is one of the reasons why square brackets were not originally supported for IN operator. I would love to get your thoughts on how to proceed with this as the scope of breaking change is more than I had anticipated.

xuzhg:
It was merged. close it.

Issue #1164, opened 2018-05-15
Bracket collection literals should be represented separately from ConstantNode


AlanWong-MS:
As of ODL 7.4.x, bracket collection literals (e.g. [1,2,3]) are represented as ConstantNode.

```
/// <summary>
/// Node representing a constant value, can either be primitive, complex, entity, or collection value.
/// </summary>
public sealed class ConstantNode : SingleValueNode
```

This creates ambiguity as the literal is a collection, yet it's created as a SingleValueNode. The ideal approach is to have a separate CollectionConstantNode that derives from CollectionNode. However, adjusting this logic for bracket collection literals would be a breaking change, so we should consider this change for the next version of ODL.

### Assemblies affected
ODL 7.x

### Reproduce steps
See ConstantNode implementation.

KanishManuja-MS:
@mikepizzo This is another breaking change associated with changing parentheses to brackets.

KanishManuja-MS:
While this may not be fully fixed with regards to collections in general. It is fixed for the In operator and therefore is not really concerning.

PR #1394, opened 2019-02-04
Support parantheses and brackets in a CollectionConstantNode for IN operator

KanishManuja-MS:
### Issues

*This pull request fixes issue #1385 and #1164 .*

### Description
This pull request modifies the bind method for the InNode to handle collection constants with a special method to support for both parentheses and brackets for a collection constant. 
#### Parsing
Previously, bracketed collection constants were parsed as constant nodes but the In operator is only compatible with collection constant nodes.  Now, we use a different method to bind literals for the InNode that parses bracketed collection as collection nodes as well.
The downside is that previously we allowed custom implementations to bind literals by overriding the bind method but that will not be possible for the In operator.
#### Generation
For the generation, we always returned the literal's text. Therefore, we will support both parentheses and brackets for the generation. 

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/LiteralBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/LiteralBinder.cs
@@ -26,29 +26,40 @@ namespace Microsoft.OData.UriParser
             {
                 if (literalToken.ExpectedEdmTypeReference != null)
                 {
-                    OData.Edm.IEdmCollectionTypeReference collectionReference =
-                        literalToken.ExpectedEdmTypeReference as OData.Edm.IEdmCollectionTypeReference;
-                    if (collectionReference != null && literalToken.OriginalText[0] == '(')
-                    {
-                        // CollectionConstantNode currently supports only parenthesis-based literals.
-                        // See https://github.com/OData/odata.net/issues/1164. When we are ready to move bracket-based
-                        // literals to CollectionConstantNode, simply remove the '(' check in the if-statement.
-                        ODataCollectionValue collectionValue = literalToken.Value as ODataCollectionValue;
-                        if (collectionValue != null)
-                        {
-                            return new CollectionConstantNode(collectionValue.Items, literalToken.OriginalText, collectionReference);
-                        }
-
-                        Debug.Assert(collectionValue != null,
-                            "Literal token creates ODataCollectionValue. If this is not the case anymore, update the code path.");
-                    }
-
                     return new ConstantNode(literalToken.Value, literalToken.OriginalText, literalToken.ExpectedEdmTypeReference);
                 }
 
                 return new ConstantNode(literalToken.Value, literalToken.OriginalText);
             }
 
+            return new ConstantNode(literalToken.Value);
+        }
+
+        /// <summary>
+        /// Binds a literal value to a ConstantNode
+        /// </summary>
+        /// <param name="literalToken">Literal token to bind.</param>
+        /// <returns>Bound query node.</returns>
+        internal static QueryNode BindInLiteral(LiteralToken literalToken)
+        {
+            ExceptionUtils.CheckArgumentNotNull(literalToken, "literalToken");
+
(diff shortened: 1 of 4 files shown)


src/Microsoft.OData.Core/UriParser/Binders/LiteralBinder.cs
```diff
@@ -22,33 +22,46 @@ internal static QueryNode BindLiteral(LiteralToken literalToken)
         {
             ExceptionUtils.CheckArgumentNotNull(literalToken, "literalToken");
 
+            if (!string.IsNullOrEmpty(literalToken.OriginalText))
+            {
+                if (literalToken.ExpectedEdmTypeReference != null)
+                {
+                    return new ConstantNode(literalToken.Value, literalToken.OriginalText, literalToken.ExpectedEdmTypeReference);
+                }
```
mikepizzo: Do we need to do this check, or can we just pass literalToken.ExpectedEdmTypeReference to the ConstantNode constructor, even if it's null?

src/Microsoft.OData.Core/UriParser/Binders/LiteralBinder.cs
```diff
@@ -22,33 +22,46 @@ internal static QueryNode BindLiteral(LiteralToken literalToken)
         {
             ExceptionUtils.CheckArgumentNotNull(literalToken, "literalToken");
 
+            if (!string.IsNullOrEmpty(literalToken.OriginalText))
```
mikepizzo: >if (!string.IsNullOrEmpty(literalToken.OriginalText)) [](start = 12, length = 53)

This means that we won't set the type reference if the originalText is null.  Is that the desired behavior? Would we ever know the typereference but not the originaltext?

mikepizzo:
These changes look good to me.

Did we decide that we'd change how we write a Uri with a collectionConstantNode (to use square brackets) in a separate PR?

src/Microsoft.OData.Core/UriParser/Binders/LiteralBinder.cs
KanishManuja-MS: Yes. We need this check. Explicitly setting the expected type to null causes problems. I can share the details if you are interested but in short, this check is necessary.

src/Microsoft.OData.Core/UriParser/Binders/LiteralBinder.cs
KanishManuja-MS: We will set the type reference even if the OriginalText is null but by getting the type of the value from Reflection. 

None of our tests have the scenario where we have the TypeReference but not the OriginalText. We will keep the original behavior, because it is likely either it is not a valid case or a very obscured case since there are no known issues around this.

src/Microsoft.OData.Core/UriParser/Binders/MetadataBinder.cs
```diff
@@ -354,7 +355,15 @@ protected virtual QueryNode BindStringLiteral(StringLiteralToken stringLiteralTo
...
         {
-            InBinder inBinder = new InBinder(this.Bind);
+            Func<QueryToken, QueryNode> InBinderMethod = (queryToken) =>
+             {
+                 if (queryToken.Kind == QueryTokenKind.Literal)
+                 {
+                     return LiteralBinder.BindInLiteral((LiteralToken)queryToken);
+                 }
+                 return this.Bind(queryToken);
+             };
```
xuzhg: shall we move it into a separate protected virtual method?

KanishManuja-MS: The entire InBind method is virtual and protected. There will always be an option to override the entire method instead of creating another virtual method.

mikepizzo:
Great; thanks for the follow-up!

---
In reply to: [461630997](https://github.com/OData/odata.net/pull/1394#issuecomment-461630997) [](ancestors = 461630997)

PR #1400, opened 2019-02-08
Enable Uri parser to parse escape function

xuzhg:
### Description

Currently, OneDrive supports path segments following any driveItem, and maps it to a bound function.  i.e.;

     ` root:/photos/2018/February:/permissions`

would be translated to:

       `root/oneDrive.getByPath(path='photos/2018/February')/permissions`

while:

        `root:/photos/2018/February`

would be translated to:

  ` root/oneDrive.getByTerminalPath(path='photos/2018/February')`

This PR enables the above translation. 

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1012,6 +1012,11 @@ private bool TryCreateSegmentForOperation(ODataPathSegment previousSegment, stri
                 bindingType = (previousSegment is EachSegment) ? previousSegment.TargetEdmType : previousSegment.EdmType;
             }
 
+            if (identifier != null && identifier.Length >= 1 && identifier[0] == ':' && bindingType != null)
```
mikepizzo: >identifier != null && identifier.Length >= 1 [](start = 16, length = 44)

Is this the same as !String.IsNullOrEmpty(identifier)? Is one better than the other?

xuzhg: Yes. !String.IsNullOrEmpty(identifier) is better of readability.

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
```diff
@@ -1035,5 +1037,95 @@ public void ParseFullyQualifiedEnumWithResolver()
...
         }
+
+        [Fact]
+        public void ParseEscapeFunctionUrlReturnsTheSameODataPath()
+        {
+            // Arrange
+            IEdmModel model = GetEdmModelWithEscapeFunction(escape: true);
+
+            // Act
+            string fullUriString = ServiceRoot + "/root/NS.Function(path='photos%2F2018%2FFebruray')";
```
mikepizzo: >"/root/NS.Function(path='photos%2F2018%2FFebruray') [](start = 49, length = 51)

Can we add tests for the escape function followed by a nav prop, key segment, additional escape path, or regular function?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
```diff
@@ -1035,5 +1037,95 @@ public void ParseFullyQualifiedEnumWithResolver()
...
+            ODataUriParser parser = new ODataUriParser(model, ServiceRoot, new Uri(fullUriString));
+            var functionPath = parser.ParsePath();
+
+            fullUriString = ServiceRoot + "/root:/photos/2018/Februray";
```
mikepizzo: >Februray [](start = 62, length = 8)

Nit: spelling error ("February")

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1593,5 +1598,45 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
...
+            IEdmFunction function = operations.OfType<IEdmFunction>().FirstOrDefault(f => IsUrlEscapeFunction(model, f));
+            if (function == null)
+            {
+                throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_NoBoundEscapeFunctionSupported(bindingType.FullTypeName()));
+            }
+
+            if (function.Parameters == null || function.Parameters.Count() != 2 || !function.Parameters.ElementAt(1).Type.IsString())
+            {
+                throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_EscapeFunctionMustHaveOneStringParameter(function.FullName()));
+            }
```
mikepizzo: if we can't resolve the function as an escape function, would it be more backward compatible just to return identifier? What would have happened previously if a segment started with a colon? Would this ever have been legal previously?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1593,5 +1598,45 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
...
+
+            IEdmVocabularyAnnotation annotation = model.FindVocabularyAnnotations<IEdmVocabularyAnnotation>(function,
+                Edm.Vocabularies.Community.V1.CommunityVocabularyModel.UrlEscapeFunctionTerm).FirstOrDefault();
+            if (annotation != null)
+            {
+                IEdmBooleanConstantExpression tagConstant = annotation.Value as IEdmBooleanConstantExpression;
+                if (tagConstant != null)
+                {
+                    return tagConstant.Value;
+                }
```
mikepizzo: if tagConstant == null, shouldn't we return true?

xuzhg: I think we should limit the annotation only boolean type, right. otherwise, we should ignore the annotation because maybe it's other meaning.

---
In reply to: [255711329](https://github.com/OData/odata.net/pull/1400#discussion_r255711329) [](ancestors = 255711329)

xuzhg:
@KanishManuja-MS @mikepizzo It's ready to review. Thanks.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
mikepizzo: For tag constants, if the annotation is applied but a value is not specified then the value is assumed to be true.

---
In reply to: [256082113](https://github.com/OData/odata.net/pull/1400#discussion_r256082113) [](ancestors = 256082113,255711329)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
mikepizzo: i.e., should we make this TryResolveEscapeFunction and, if it fails, continue existing processing? Or are we sure existing request processing would fail, so we might as well fail here.

---
In reply to: [255710193](https://github.com/OData/odata.net/pull/1400#discussion_r255710193) [](ancestors = 255710193)

src/Microsoft.OData.Edm/Validation/EdmErrorCode.cs
```diff
@@ -1372,5 +1372,15 @@ public enum EdmErrorCode
...
+
+        /// <summary>
+        /// The composable escape bound function should not declare more than one.
+        /// </summary>
+        EntityComposableBoundEscapeFunctionMustBeLessOne,
+
+        /// <summary>
+        /// The non-composable escape bound function should not declare more than one.
+        /// </summary>
+        EntityNoncomposableBoundEscapeFunctionMustBeLessOne,
```
mikepizzo: The other member names all have assigned values; any reason not to continue that pattern?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
mikepizzo: Thanks for all of the added tests!

---
In reply to: [255706009](https://github.com/OData/odata.net/pull/1400#discussion_r255706009) [](ancestors = 255706009)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1595,5 +1600,56 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
...
+            Debug.Assert(identifier != null && identifier.Length >= 1 && identifier[0] == ':');
+            Debug.Assert(bindingType != null);
+
+            bool isComposableRequired = identifier.Length >= 2 && identifier[identifier.Length - 1] == ':';
+            IEnumerable<IEdmFunction> functions = model.FindBoundOperations(bindingType).OfType<IEdmFunction>().Where(f => IsUrlEscapeFunction(model, f));
+
+            IEdmFunction function = functions.FirstOrDefault(f => f.IsComposable == isComposableRequired);
+            if (function == null)
+            {
+                // If no non-composable escape function exist, use the composable function
```
mikepizzo: Did we decide the right behavior was to fall back to the composable function, or to error?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1595,5 +1600,56 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
...
+                if (!isComposableRequired)
```
mikepizzo: Should we just make this part of the previous condition? i.e., if (function == null && !isComposableRequired)

src/Microsoft.OData.Core/Parameterized.Microsoft.OData.Core.cs
```diff
@@ -4427,6 +4427,13 @@ internal static class Strings {
             }
         }
 
+        /// <summary>
+        /// A string like "The URI part '{0}' is not valid because there's no escape leading delimiter before."
```
mikepizzo: >there's no escape leading delimiter before." [](start = 67, length = 44)

suggest reword to "there's no leading escape character."

src/Microsoft.OData.Core/Parameterized.Microsoft.OData.Core.cs
```diff
@@ -5674,6 +5681,20 @@ internal static class Strings {
...
 
+        /// <summary>
+        /// A string like "The request URI is not valid. The bound function binding to '{0}' does not support the escape function annotation."
+        /// </summary>
+        internal static string RequestUriProcessor_NoBoundEscapeFunctionSupported(object p0) {
+            return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.RequestUriProcessor_NoBoundEscapeFunctionSupported, p0);
+        }
+
+        /// <summary>
+        /// A string like "The UrlEscape function '{0}' must have and only have one 'Edm.String' parameter."
```
mikepizzo: >must have and only have one 'Edm.String' parameter [](start = 56, length = 50)

must have exactly one non-binding parameter of type 'Edm.String'.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
xuzhg: I saw the email you forward to me. Let's disallow this behaviour.

---
In reply to: [267994664](https://github.com/OData/odata.net/pull/1400#discussion_r267994664) [](ancestors = 267994664)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
xuzhg: Ok, Thanks.

---
In reply to: [267886718](https://github.com/OData/odata.net/pull/1400#discussion_r267886718) [](ancestors = 267886718,256082113,255711329)

src/Microsoft.OData.Edm/Validation/EdmErrorCode.cs
xuzhg: What i think is that to assign a value is not feasible to insert new member at middle. 

But, i am ok to add a value now.

---
In reply to: [267993749](https://github.com/OData/odata.net/pull/1400#discussion_r267993749) [](ancestors = 267993749)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
mikepizzo: Note: i suspect that the parse would fail if it contained this syntax, so I'm okay not making this TryResolveEscapeFunction and just failing, I just wanted to be sure we weren't making something that was previously valid now an error.

---
In reply to: [267887753](https://github.com/OData/odata.net/pull/1400#discussion_r267887753) [](ancestors = 267887753,255710193)

xuzhg: I just tested:

a: "/Customers(2):/abc:xyz:" and "Customers(2)/MyOrders:/xyz/abc:", it returns "Bad Request - Error in query syntax.".

b: "/root:/photos%2F2018%2F/:February:/Name", it returns "Resource not found for the segment ':photos%2F2018%2F/:February:'."

So, I think to throw the better exception message in "TryResolveEscapeFunction" is good. 




---
In reply to: [268329867](https://github.com/OData/odata.net/pull/1400#discussion_r268329867) [](ancestors = 268329867,267887753,255710193)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
xuzhg: Owing to disallowing to back to composable, i have to refactor these codes. 

---
In reply to: [268309133](https://github.com/OData/odata.net/pull/1400#discussion_r268309133) [](ancestors = 268309133)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
mikepizzo: To handle null value == true, I think you want return true; on line 1645.

---
In reply to: [268326230](https://github.com/OData/odata.net/pull/1400#discussion_r268326230) [](ancestors = 268326230,267886718,256082113,255711329)

Issue #1383, opened 2019-01-24
Revert writing empty parens for expanded nav prop in 4.0 response


mikepizzo:
PR #1286 fixed issue #1265 (along with others) by including an empty parens in the ContextUrl for expanded entities with no expand options specified.

This is required in 4.01 and suggested in 4.0.  

Although empty parens should be supported for 4.0, and clients using our 4.0 libraries are able to consume this, we have at least two incidents of customers parsing the contextUrl on their own which are not expecting the empty parens in this case.  

To be safe, we should make adding the empty parens written only for 4.01 responses.

mikepizzo:
Doesn't seem worth fixing at this point.

PR #1402, opened 2019-02-12
Revert empty parentheses for expanded navigation properties without su…

KanishManuja-MS:
### Issues

*This pull request fixes issue #1383 .*

### Description

This pull request reverts some of the changes made by #1286 to fix the above-mentioned issue.
The reader and writer will not support empty parentheses unless the version is specified to be greater than v4. 

The controversial piece is regarding this change is when a navigation property is selected as well as expanded. We now will include the property twice in the context URI, which is not as obvious as it was with empty parentheses. Due to this, we may omit writing some navigational links for v4 which was a preexisting bug prior to # 1286.  

- [x] *Test cases added*

### Note for the reviewers
My expectation was that simply reverting the changes made by the PR should restore the expected behavior. However, it ended up being more than that.

### DIFF
--- a/src/Microsoft.OData.Core/ODataContextUrlInfo.cs
+++ b/src/Microsoft.OData.Core/ODataContextUrlInfo.cs
@@ -287,7 +287,7 @@ namespace Microsoft.OData
                 }
                 else
                 {
-                    return CreateSelectExpandContextUriSegment(odataUri.SelectAndExpand);
+                    return CreateSelectExpandContextUriSegment(odataUri.SelectAndExpand, version);
                 }
             }
 
@@ -385,13 +385,14 @@ namespace Microsoft.OData
         /// Build the expand clause for a given level in the selectExpandClause
         /// </summary>
         /// <param name="selectExpandClause">the current level select expand clause</param>
+        /// <param name="version">OData Version of the response</param>
         /// <returns>the select and expand segment for context url in this level.</returns>
-        private static string CreateSelectExpandContextUriSegment(SelectExpandClause selectExpandClause)
+        private static string CreateSelectExpandContextUriSegment(SelectExpandClause selectExpandClause, ODataVersion version)
         {
             if (selectExpandClause != null)
             {
                 string contextUri;
-                selectExpandClause.Traverse(ProcessSubExpand, CombineSelectAndExpandResult, out contextUri);
+                selectExpandClause.Traverse(ProcessSubExpand, CombineSelectAndExpandResult, version, out contextUri);
                 if (!string.IsNullOrEmpty(contextUri))
                 {
                     return ODataConstants.ContextUriProjectionStart + contextUri + ODataConstants.ContextUriProjectionEnd;
@@ -404,11 +405,12 @@ namespace Microsoft.OData
         /// <summary>Process sub expand node, contact with subexpand result</summary>
         /// <param name="expandNode">The current expanded node.</param>
         /// <param name="subExpand">Generated sub expand node.</param>
+        /// <param name="version">OData Version of the generated response.</param>
         /// <returns>The generated expand string.</returns>
-        private static string ProcessSubExpand(string expandNode, string subExpand)
+        private static string ProcessSubExpand(string expandNode, string subExpand, ODataVersion version)
         {
-
-            return expandNode + ODataConstants.ContextUriProjectionStart + subExpand + ODataConstants.ContextUriProjectionEnd;
+            return string.IsNullOrEmpty(subExpand) && version <= ODataVersion.V4 ? expandNode :
+                            expandNode + ODataConstants.ContextUriProjectionStart + subExpand + ODataConstants.ContextUriProjectionEnd;
         }
 
         /// <summary>Create combined result string using selected items list and expand items list.</summary>
--- a/src/Microsoft.OData.Core/ODataContextUrlInfo.cs
+++ b/src/Microsoft.OData.Core/ODataContextUrlInfo.cs
@@ -417,13 +417,33 @@ namespace Microsoft.OData
         /// <param name="selectList">A list of selected item names.</param>
         /// <param name="expandList">A list of sub expanded item names.</param>
         /// <returns>The generated expand string.</returns>
-        private static string CombineSelectAndExpandResult(IList<string> selectList, IList<string> expandList)
+        private static string CombineSelectAndExpandResult(IList<string> selectList, IList<string> expandList, ODataVersion version)
         {
             string currentExpandClause = string.Empty;
-
+            IList<string> RemoveFromExpandList = new List<string>();
             if (selectList.Any())
             {
+                if (version <= ODataVersion.V4)
+                {
+                    foreach (var item in expandList)
+                    {
+                        //For backwards compatibility, we want to remove duplicate properties in select and expand list for v4
+                        int idx = item.IndexOf(ODataConstants.ContextUriProjectionStart);
+                        if (idx >= 0)
+                        {
+                            selectList.Remove(item.Substring(0, idx));
+                        }
+                        else if(selectList.Contains(item))
+                        {
+                            RemoveFromExpandList.Add(item);
+                        }
+                    }
 
+                    foreach (string s in RemoveFromExpandList)
+                    {
+                        expandList.Remove(s);
+                    }
+                }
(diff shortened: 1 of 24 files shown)


src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
```diff
@@ -283,7 +283,7 @@ internal SelectedPropertiesNode SelectedProperties
             get
             {
                 return this.SelectExpandClause != null
-                    ? SelectedPropertiesNode.Create(this.SelectExpandClause)
+                    ? SelectedPropertiesNode.Create(this.SelectExpandClause, this.Version ?? ODataVersion.V4)
```
xuzhg: ?? maybe can't pass .net 3.5 compile.

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataContextUriBuilderTests.cs
```diff
@@ -267,9 +277,10 @@ public void EntryContextUriWithExpandNestedSelectString()
         [InlineData(ODataVersion.V401)]
         public void EntryContextUriWithExpandNestedExpandString(ODataVersion version)
         {
+            string emptyParens = version == ODataVersion.V401 ? "()" : "";
             // Without inner $select, $expand=A($expand=B($expand=C))
             string expandClause = "Districts($expand=City($expand=Districts))";
-            string expectedClause = "Districts(City(Districts()))";
+            string expectedClause = "Districts(City(Districts" + emptyParens + "))";
```
xuzhg: >ng expectedClause = "Districts(City(Districts" [](start = 16, length = 46)

It's better to put the expectedClause into InlineData

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataContextUriBuilderTests.cs
```diff
@@ -296,7 +307,7 @@ public void FeedContextUriWithMixedSelectAndExpandString(ODataVersion version)
         {
             const string selectClause = "Size,Name";
             const string expandClause = "Districts($select=Zip,City;$expand=City($expand=Districts;$select=Name))";
-            const string expectedClause = "Size,Name,Districts(Zip,City,City(Name,Districts()))";
```
xuzhg: >Size,Name,Districts(Zip,City,City(Name,Districts()))"; [](start = 43, length = 54)

put it into the InlineDatah

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/ContextUrlWriterReaderTests.cs
```diff
@@ -741,9 +741,9 @@ public void CollectionOfProjectedExpandedEntities()
         [Fact]
         public void CollectionOfExpandedEntities()
         {
-            string contextString = "\"{0}$metadata#Employees(AssociatedCompany())\"";
             foreach (ODataVersion version in Versions)
```
xuzhg: >foreach (ODataVersion version in Versions) [](start = 12, length = 42)

maybe change the cases to use "[Theory]"

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/ContextUrlWriterReaderTests.cs
```diff
@@ -871,11 +872,12 @@ public void ProjectedExpandedEntity()
         [Fact]
         public void ExpandedEntity()
         {
+
             ODataResource entry = new ODataResource()
             {
                 TypeName = TestNameSpace + ".Employee",
                 Properties = new[]
-                {
```
xuzhg: >      [](start = 5, length = 5)

revert

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/ContextUrlWriterReaderTests.cs
```diff
@@ -926,11 +928,10 @@ public void ExpandedEntity()
         // V4 Protocol Spec Chapter 10.10: Projected Expanded Entities (not contain select)
         // Sample Request: http://host/service/Employees(0)?$expand=AssociatedCompany
         [Theory]
-        [InlineData(ODataVersion.V4)]
-        [InlineData(ODataVersion.V401)]
-        public void ExpandedMultiSegmentEntity(ODataVersion version)
+        [InlineData(ODataVersion.V4, "\"{0}$metadata#Employees(AssociatedCompany,AssociatedCompany)/$entity\"")]
+        [InlineData(ODataVersion.V401, "\"{0}$metadata#Employees(AssociatedCompany,AssociatedCompany())/$entity\"")]
```
xuzhg: I like that, it's clearer

xuzhg:
>         internal static ODataContextUrlInfo Create(ODataValue value, ODataVersion version, ODataUri odataUri = null, IEdmModel model = null)

can we save version into the class, and use it later

---
Refers to: src/Microsoft.OData.Core/ODataContextUrlInfo.cs:79 in caefbf9. [](commit_id = caefbf9c4a4846bacc5051064b12c74aec6eb7f8, deletion_comment = False)

src/Microsoft.OData.Core/ODataContextUrlInfo.cs
```diff
@@ -398,11 +399,12 @@ private static string CreateSelectExpandContextUriSegment(SelectExpandClause sel
...
         /// <param name="expandNode">The current expanded node.</param>
         /// <param name="subExpand">Generated sub expand node.</param>
+        /// <param name="version">OData Version of the generated response.</param>
         /// <returns>The generated expand string.</returns>
-        private static string ProcessSubExpand(string expandNode, string subExpand)
+        private static string ProcessSubExpand(string expandNode, string subExpand, ODataVersion version)
         {
-
-            return expandNode + ODataConstants.ContextUriProjectionStart + subExpand + ODataConstants.ContextUriProjectionEnd;
+            return string.IsNullOrEmpty(subExpand) && version <= ODataVersion.V4 ? expandNode :
```
xuzhg: >string.IsNullOrEmpty(subExpand) [](start = 19, length = 31)

why add IsNullOrEmpty(subExpand) test?

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
KanishManuja-MS: It is supported for 3.5 as well. At least it compiled just fine. 

---
In reply to: [256175616](https://github.com/OData/odata.net/pull/1402#discussion_r256175616) [](ancestors = 256175616)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/ContextUrlWriterReaderTests.cs
KanishManuja-MS: can you elaborate? What needs to be reverted?


---
In reply to: [256178335](https://github.com/OData/odata.net/pull/1402#discussion_r256178335) [](ancestors = 256178335)

src/Microsoft.OData.Core/ODataContextUrlInfo.cs
KanishManuja-MS: In V4.01, we need to have empty parentheses in the context URI for an expanded navigation property even if there are not nested select or expand option. With Biao's change, we started adding the empty parentheses for V4 as well. The check in question basically determines if there was any nested expand or select for the current expand in the select expand.

---
In reply to: [256179761](https://github.com/OData/odata.net/pull/1402#discussion_r256179761) [](ancestors = 256179761)

KanishManuja-MS:
>         internal static ODataContextUrlInfo Create(ODataValue value, ODataVersion version, ODataUri odataUri = null, IEdmModel model = null)

Right now, all these are static methods and if we save the version into the class, ComputeQueryClause will not longer be static.

---
In reply to: [462968518](https://github.com/OData/odata.net/pull/1402#issuecomment-462968518) [](ancestors = 462968518)

---
Refers to: src/Microsoft.OData.Core/ODataContextUrlInfo.cs:79 in caefbf9. [](commit_id = caefbf9c4a4846bacc5051064b12c74aec6eb7f8, deletion_comment = False)

PR #1407, opened 2019-02-15
OData context uri for v4 - revert empty parentheses for navigation properties

KanishManuja-MS:
### Issues

*This pull request fixes issue #1383 .*

### Description

This pull request reverts some of the changes made by #1286 to fix the above-mentioned issue.
The reader and writer will not support empty parentheses unless the version is specified to be greater than v4. 

The controversial piece is regarding this change is when a navigation property is selected as well as expanded. We now will include the property twice in the context URI, which is not as obvious as it was with empty parentheses. Due to this, we may omit writing some navigational links for v4 which was a preexisting bug prior to #1286.  

- [x] *Test cases added*


src/Microsoft.OData.Core/ODataContextUrlInfo.cs
```diff
@@ -385,13 +385,14 @@ private static string CreateApplyUriSegment(ApplyClause applyClause)
         /// Build the expand clause for a given level in the selectExpandClause
         /// </summary>
         /// <param name="selectExpandClause">the current level select expand clause</param>
+        /// <param name="version">OData Version of the response</param>
```
xuzhg: >OData Version of the response [](start = 34, length = 29)

maybe resentence this.

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -223,12 +263,13 @@ internal enum SelectionType
         /// <param name="selectQueryOption">The value of the $select query option.</param>
         /// <param name="structuredType">The structured type of this node.</param>
         /// <param name="edmModel">The Edm model.</param>
+        /// <param name="version"> The OData version. Default value is <see cref="ODataVersion.V4"/></param>
         /// <returns>A tree representation of the selected properties specified in the query option.</returns>
-        internal static SelectedPropertiesNode Create(string selectQueryOption, IEdmStructuredType structuredType, IEdmModel edmModel)
+        internal static SelectedPropertiesNode Create(string selectQueryOption, IEdmStructuredType structuredType, IEdmModel edmModel, ODataVersion? version = ODataVersion.V4)
```
xuzhg: >ODataVersion? [](start = 135, length = 13)

It seems you don't need the nullable type. 

otherwise, it seems the default value "null" is better.

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -831,7 +875,7 @@ private static SelectedPropertiesNode ProcessSubExpand(string nodeName, Selected
         /// <param name="selectList">An enumerable of selected item names.</param>
         /// <param name="expandList">An enumerable of sub expanded nodes.</param>
         /// <returns>The generated SelectedPropertiesNode.</returns>
-        private static SelectedPropertiesNode CombineSelectAndExpandResult(IEnumerable<string> selectList, IEnumerable<SelectedPropertiesNode> expandList)
+        private static SelectedPropertiesNode CombineSelectAndExpandResult(IList<string> selectList, IList<SelectedPropertiesNode> expandList, ODataVersion version)
```
xuzhg: >IList<string> [](start = 75, length = 13)

what's the reason to change from "IEnumerable<T>" to "IList<T>"

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/ContextUrlWriterReaderTests.cs
```diff
@@ -871,6 +871,7 @@ public void ProjectedExpandedEntity()
         [Fact]
         public void ExpandedEntity()
         {
+
             ODataResource entry = new ODataResource()
```
xuzhg: revert.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/ContextUrlWriterReaderTests.cs
```diff
@@ -911,7 +912,7 @@ public void ExpandedEntity()
                     writer.WriteStart(entry);
                     writer.WriteEnd();
                 },
-                string.Format("\"{0}$metadata#Employees(AssociatedCompany,AssociatedCompany())/$entity\"", TestBaseUri),
```
xuzhg: >,AssociatedCompany()) [](start = 73, length = 21)

Is it in v401?

xuzhg:
It seems ok to me.

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
KanishManuja-MS: Got reverted because of an outdated fork.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/ContextUrlWriterReaderTests.cs
KanishManuja-MS: The default version is v4 for context uri. For this test, since the version is not explicitly passed, the context uri is for v4.  

---
In reply to: [258205555](https://github.com/OData/odata.net/pull/1407#discussion_r258205555) [](ancestors = 258205555)

KanishManuja-MS:
@mikepizzo Should I close this?

mikepizzo:
Can probably close, but leave the repo. We might decide to bring this back under a compatability flag.

raheph:
Design is changed. Close it and need more discussion.

PR #1408, opened 2019-02-21
Fix typos, build warings, comments, etc

xuzhg:
### Description

This PR try to fix part of:
1. Typos
2. Build warings
3. Comments

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/DefaultJsonWriterFactory.cs
+++ b/src/Microsoft.OData.Core/Json/DefaultJsonWriterFactory.cs
@@ -14,14 +14,15 @@ namespace Microsoft.OData.Json
     /// </summary>
     public sealed class DefaultJsonWriterFactory : IJsonWriterFactory
     {
-        private ODataStringEscapeOption _stringEscapeOption;
+        private ODataStringEscapeOption stringEscapeOption;
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DefaultJsonWriterFactory" />.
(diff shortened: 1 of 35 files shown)


src/Microsoft.OData.Core/Json/DefaultJsonWriterFactory.cs
```diff
@@ -14,22 +14,23 @@ namespace Microsoft.OData.Json
...
+        }
 
         /// <summary>
         /// Initializes a new instance of the <see cref="DefaultJsonWriterFactory" />.
         /// </summary>
         /// <param name="stringEscapeOption">The string escape option.</param>
         public DefaultJsonWriterFactory(ODataStringEscapeOption stringEscapeOption)
         {
-            _stringEscapeOption = stringEscapeOption;
+            this.stringEscapeOption = stringEscapeOption;
```
KanishManuja-MS: Some stacks prefer to use underscore for private members. If we don't use this convention then please fix all these occurrences as well - 
ExpressionNormalizer.cs(55), AuthorizationVocabularyTests.cs(23), ValidationVocabularyTests.cs(23), 
ApplyBinderTests.cs(22):        
ApplyBinderTests.cs(24):       
ApplyBinderTests.cs(372)

xuzhg: #Resolved.

PR #1414, opened 2019-03-01

xuzhg:
### Issues

### Description

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Schema/EdmOperation.cs
+++ b/src/Microsoft.OData.Edm/Schema/EdmOperation.cs
@@ -17,7 +17,7 @@ namespace Microsoft.OData.Edm
         private readonly List<IEdmOperationParameter> parameters = new List<IEdmOperationParameter>();
 
         /// <summary>
-        /// Initializes a new instance of the <see cref="EdmAction"/> class.
+        /// Initializes a new instance of the <see cref="EdmOperation"/> class.
         /// </summary>
         /// <param name="namespaceName">Name of the namespace.</param>
         /// <param name="name">The name.</param>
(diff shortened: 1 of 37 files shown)


xuzhg:
@mikepizzo @KanishManuja-MS This is the third version of the operation return type (no breaking changes, no extra type classes).  Please take a look.

You can find the second version #1360.

src/Microsoft.OData.Edm/Schema/EdmOperation.cs
```diff
@@ -83,20 +83,28 @@ public string FullName
...
         public IEnumerable<IEdmOperationParameter> Parameters
         {
             get { return this.parameters; }
         }
 
         /// <summary>
-        /// Searches for a parameter with the given name in this function and returns null if no such parameter exists.
+        /// Gets the return of this operation.
+        /// </summary>
+        internal IEdmOperationReturn Return { get; private set; }
```
mikepizzo: >interna [](start = 8, length = 7)

Might we want to make this public? If not, how would someone get/add annotations to the operation return?

xuzhg: Call the extension method to add. See the CsdlWriterTest.

mikepizzo: Since the extension method is public, that works.  Thanks.

---
In reply to: [262675031](https://github.com/OData/odata.net/pull/1414#discussion_r262675031) [](ancestors = 262675031)

mikepizzo:
I like it!

---
In reply to: [468504257](https://github.com/OData/odata.net/pull/1414#issuecomment-468504257) [](ancestors = 468504257)

Issue #1415, opened 2019-03-05
Building of a filter with both "any" and "or" fails on keeping operations priority


AndreyAndreev:
*The `ODataUri.BuildUrl()` method fails when a filter looks like this: `any() and (condition1 or condition2)`*

### Assemblies affected

*OData .Net lib 7.6.0.30305*

### Reproduce steps

The test case to demonstrate the bug:

```
public void BuildFilterWithOpenPropertyInsideAnyAndTwoOrs()
{
    Uri queryUri = new Uri("People?$filter=MyPaintings%2Fany(p%3Ap%2FOpenProperty) and (ID eq 124) and ((ID eq 125) or (ID eq 126))", UriKind.Relative);
    Uri actualUri = UriBuilder(queryUri, ODataUrlKeyDelimiter.Parentheses, settings);
    Assert.Equal(new Uri("http://gobbledygook/People?$filter=MyPaintings%2Fany(p%3Ap%2FOpenProperty) and ID eq 124 and (ID eq 125 or ID eq 126)"), actualUri);
}
```

### Expected result

The parentheses at the end of the filter shouldn't be removed. The result is expected to be like this:
`http://gobbledygook/People?$filter=MyPaintings/any(p:p/OpenProperty) and ID eq 124 and (ID eq 125 or ID eq 126)`

### Actual result

All the parentheses are removed which ruins the filter breaking operations priorities.

`http://gobbledygook/People?$filter=MyPaintings/any(p:p/OpenProperty) and ID eq 124 and ID eq 125 or ID eq 126`

AndreyAndreev:
Did some investigations on that.

The root cause of the issue actually is the return type of the `any` operator which is a `nullable bool`. All the other operators seems to have a type `non-nullable bool`. That's the reason for conversion all the `BinaryOperatorNode` into `ConvertNode`. So the filter expression tree has this `ConvertNode`'s on top of the `BinaryOpertorNode`'s. That leads to not adding a priority parens in here: 
https://github.com/OData/odata.net/blob/ce39b817127c198f236f1dba9e0440180efbfb4d/src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs#L75-L79
as the `Right.Kind` not a `BinaryOperator` anymore but a `Convert`.

Hope this helps.

xuzhg:
Merged, thanks for contribution.

PR #1416, opened 2019-03-06
Fix issue #1415, keep conditions priority for queries with Convert node in NodeToStringBuilder

AndreyAndreev:
### Issues

*This pull request fixes issue #1415 .*

### Description

The problem in the `NodeToStringBuilder` appears when the filter query contains a condition of type `nullable boolean` such as `any` as a part of `BinaryOperatorNode` with another `BinaryOperatorNode` as another part. The latter `BinaryOperatorNode` is converted into `ConvertNode` in that case, and it's not tested for the priority of the internal binary operator. 

- [x] *Test cases added*


xuzhg:
Thanks @AndreyAndreev, this change makes sense to me.

AndreyAndreev:
> Thanks @AndreyAndreev, this change makes sense to me.

Thank you @xuzhg , when it's planned to make a new release by the way?

PR #1418, opened 2019-03-07
Change to use buffer when writing binary/byte[]

xuzhg:
### Issues

* Improve the json writer process for the binary/byte[].

### Description

* Improve the json writer process for the binary/byte[].

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
@@ -316,7 +316,8 @@ namespace Microsoft.OData.Json
         /// </summary>
         /// <param name="writer">The text writer to write the output to.</param>
         /// <param name="value">Byte array to be written.</param>
-        internal static void WriteValue(TextWriter writer, byte[] value)
+        /// <param name="buffer">Char buffer to use for streaming data.</param>
+        internal static void WriteValue(TextWriter writer, byte[] value, ref char[] buffer)
         {
             Debug.Assert(writer != null, "writer != null");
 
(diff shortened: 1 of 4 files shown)


xuzhg:
@ysanghi Would you please take a look?

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -327,11 +328,43 @@ internal static void WriteValue(TextWriter writer, byte[] value)
...
+            for (int i = 0; i < value.Length; i += bufferByteSize)
+            {
+                int length = bufferByteSize;
+                if (i + length > value.Length)
+                {
+                    length = value.Length - i;
+                }
+
+                int output = Convert.ToBase64CharArray(value, i, length, buffer, 0);
+                writer.Write(new string(buffer, 0, output));
```
ysanghi: Why create a string here. Can we avoid this?

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsTests.cs
```diff
@@ -241,24 +241,38 @@ public void WriteByteShouldWork()
         {
             this.TestInit();
             var byteArray = new byte[] { 1, 2, 3 };
-            JsonValueUtils.WriteValue(this.writer, byteArray);
+            JsonValueUtils.WriteValue(this.writer, byteArray, ref this.buffer);
             this.StreamToString().Should().Be("\"AQID\"");
         }
 
+        [Fact]
+        public void WriteLongBytesShouldWork()
```
ysanghi: Write edge case unit tests where size of the buffer is equal to the length of the string, etc..

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
xuzhg: Yes. We can. I update to call `writer.Write(buffer, 0, output);` #Resolved.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsTests.cs
xuzhg: # Added.

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -327,11 +328,43 @@ internal static void WriteValue(TextWriter writer, byte[] value)
...
+
+            buffer = BufferUtils.InitializeBufferIfRequired(buffer);
+            Debug.Assert(buffer != null);
+
+            int bufferLength = buffer.Length;
+
+            // Try to hold base64 string as much as possible in one converting.
+            int bufferByteSize = bufferLength * 3 / 4;
+
+            for (int i = 0; i < value.Length; i += bufferByteSize)
```
mikepizzo: >i [](start = 21, length = 1)

Consider using a more meaningful name, like "startPos"

xuzhg: Thanks, I change to use "offsetIn", which is used in 'ToBase64CharArray`

mikepizzo: perfect

PR #1420, opened 2019-03-09
Use the array pool in Json writer

xuzhg:
### Issues

* Use the array pool in Json writer

### Description

* Use the array pool in Json writer

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Buffers/BufferUtils.cs
+++ b/src/Microsoft.OData.Core/Buffers/BufferUtils.cs
@@ -23,12 +23,23 @@ namespace Microsoft.OData.Buffers
         /// <returns>The initialized buffer</returns>
         public static char[] InitializeBufferIfRequired(char[] buffer)
         {
-            if (buffer == null)
+            return InitializeBufferIfRequired(null, buffer);
+        }
+
+        /// <summary>
+        /// Checks if the buffer is not initialized and if initialized returns the same buffer or creates a new one.
+        /// </summary>
+        /// <param name="bufferPool">The character pool.</param>
+        /// <param name="buffer">The buffer to verify</param>
+        /// <returns>The initialized buffer.</returns>
+        public static char[] InitializeBufferIfRequired(ICharArrayPool bufferPool, char[] buffer)
+        {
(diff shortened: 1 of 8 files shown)


xuzhg:
@ysanghi Would you please also take a look this PR? thanks.

src/Microsoft.OData.Core/Buffers/BufferUtils.cs
```diff
@@ -23,12 +23,23 @@ internal static class BufferUtils
         /// <returns>The initialized buffer</returns>
         public static char[] InitializeBufferIfRequired(char[] buffer)
         {
-            if (buffer == null)
+            return InitializeBufferIfRequired(null, buffer);
```
raheph: Why  do we need this method? can't we do something like this?
if (buffer == null) {
       return  RentFromBuffer(null, buffer);
}
return buffer;

xuzhg: That's avoiding a duplicated code. At later, if we change something, we only change the second method, don't need to change all methods. Thanks.

src/Microsoft.OData.Core/Json/JsonWriter.cs
```diff
@@ -392,6 +398,7 @@ public void WriteRawValue(string rawValue)
         /// </summary>
         public void Flush()
         {
+            BufferUtils.ReturnToBuffer(this.ArrayPool, this.buffer);
             this.writer.Flush();
```
mikepizzo: Should we return to pool on Flush or Dispose?

xuzhg: @mikepizzo I changed to use the "Dispose".

src/Microsoft.OData.Core/Buffers/BufferUtils.cs
mikepizzo: I'm okay with that pattern.

---
In reply to: [265697058](https://github.com/OData/odata.net/pull/1420#discussion_r265697058) [](ancestors = 265697058)

PR #1425, opened 2019-03-19
Properties defined in $compute and $apply can be used in a following query options ($select, $compute, $filter or $orderby)

kosinsky:
### Issues

*This pull request fixes issue #1426.*

### Description

*Makes dynamic properties generated in $apply and $compute visible for query options evaluated after them ($select, $filter, $expand, $orderby etc.)*

Also, current implementation doesn't block queries like
`$apply=groupby((Field1))&$filter=Fiield2 eq 1` during parsing and binding that leads to cryptic exceptions during IQueryable building

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandSemanticBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandSemanticBinder.cs
@@ -5,7 +5,6 @@
 //---------------------------------------------------------------------
 
 using Microsoft.OData.Edm;
-using System.Collections.Generic;
 
 namespace Microsoft.OData.UriParser
 {
@@ -21,6 +20,7 @@ namespace Microsoft.OData.UriParser
         /// <param name="expandToken">the syntactically parsed expand token</param>
         /// <param name="selectToken">the syntactically parsed select token</param>
         /// <param name="configuration">The configuration to use for parsing.</param>
+        /// <param name="state">The state of binding.</param>
         /// <returns>A select expand clause bound to metadata.</returns>
         public static SelectExpandClause Bind(
             ODataPathInfo odataPathInfo,
--- a/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandSemanticBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandSemanticBinder.cs
@@ -27,14 +27,15 @@ namespace Microsoft.OData.UriParser
             ExpandToken expandToken,
             SelectToken selectToken,
             ODataUriParserConfiguration configuration,
-            List<string> generatedProperties)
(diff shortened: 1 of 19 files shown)


src/Microsoft.OData.Core/UriParser/SyntacticAst/PathToken.cs
```diff
@@ -27,5 +27,37 @@ public abstract class PathToken : QueryToken
         /// The name of the property to access.
         /// </summary>
         public abstract string Identifier { get; }
+
+        public override bool Equals(object obj)
```
xuzhg: >public override bool Equals(object obj) [](start = 8, length = 39)

Would you please add the comments

xuzhg:
@mikepizzo It looks good to me for the changes. Would you please take a look quickly and I can merge it and release it in the 7.6 beta.

src/Microsoft.OData.Core/UriParser/ODataQueryOptionParser.cs
```diff
@@ -474,19 +469,43 @@ private OrderByClause ParseOrderByImplementation(string orderBy, ODataUriParserC
...
+            BindingState state = CreateBindingState(configuration, odataPathInfo);
+
+            MetadataBinder binder = new MetadataBinder(state);
+            OrderByBinder orderByBinder = new OrderByBinder(binder.Bind);
+            OrderByClause orderByClause = orderByBinder.BindOrderBy(state, orderByQueryTokens);
+
+            return orderByClause;
+        }
+
+        private BindingState CreateBindingState(ODataUriParserConfiguration configuration, ODataPathInfo odataPathInfo)
```
mikepizzo: >CreateBindingState [](start = 29, length = 18)

What state from QueryOptionParser does CreateBindingState require that forces it to be an instance method, rather than a static method? applyClause and computeClause?

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandSemanticBinder.cs
```diff
@@ -5,6 +5,7 @@
 //---------------------------------------------------------------------
 
 using Microsoft.OData.Edm;
+using System.Collections.Generic;
```
mikepizzo: Why is this new using required?

mikepizzo:
Did a quick pass and didn't see anything too concerning; we should go ahead and take this for beta.

---
In reply to: [475426851](https://github.com/OData/odata.net/pull/1425#issuecomment-475426851) [](ancestors = 475426851)

src/Microsoft.OData.Core/UriParser/ODataQueryOptionParser.cs
kosinsky: Yes, applyCluase and computeClause are instance level

PR #1432, opened 2019-03-27
$count is a valid path after collapsed inputs

kosinsky:
### Issues

Allow up PR for https://github.com/OData/odata.net/pull/1425

### Description

$count should be allowed after groupby in expressions like `groupby((ID))/aggregate($count as Count)`
Before fix it was failing with Microsoft.OData.ODataException : $apply/groupby grouping expression '$count' must evaluate to a property access value. 

Fix moves check for $coutn count special case before lookup in the list of properties that are available after collapsing inputs.

- [x] *Test cases added*

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Extensions/Binders/ApplyBinderTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Extensions/Binders/ApplyBinderTests.cs
@@ -461,6 +461,30 @@ namespace Microsoft.OData.Tests.UriParser.Extensions.Binders
             expandItem1.FilterOption.Should().NotBeNull();
         }
 
+        [Fact]
+        public void BindVirtualPropertiesAfterCollapseReturnsApplyClause()
+        {
+            IEnumerable<QueryToken> tokens =
+                _parser.ParseApply(
+                    "groupby((ID))/aggregate($count as Count)");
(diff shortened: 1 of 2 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Extensions/Binders/ApplyBinderTests.cs
```diff
@@ -461,6 +461,30 @@ public void BindApplyWithNestedExpandReturnApplyClause()
...
+            ApplyClause actual = binder.BindApply(tokens);
+
+            actual.Should().NotBeNull();
+            actual.Transformations.Should().HaveCount(2);
+            GroupByTransformationNode groupby = actual.Transformations.First() as GroupByTransformationNode;
+            groupby.Should().NotBeNull();
+            AggregateTransformationNode aggregate = actual.Transformations.Last() as AggregateTransformationNode;
+            aggregate.Should().NotBeNull();
+            aggregate.AggregateExpressions.Should().HaveCount(1);
+            aggregate.AggregateExpressions.Single().As<AggregateExpression>().Method.ShouldBeEquivalentTo(AggregationMethod.VirtualPropertyCount);
```
xuzhg: Remove the useless white line. but it's ok.

xuzhg:
Merged. Thanks.

PR #1440, opened 2019-04-03
Add the Example Term into Core vocabulary

xuzhg:
### Issues

* Add the Example Term into Core vocabulary

### Description

* Add the Example Term into Core vocabulary

- [x] *Test cases added*

### Additional work necessary


src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -713,7 +713,8 @@ public bool ModelHasInheritance
                 Debug.Assert(this.EdmModel != null, "this.EdmModel != null");
                 this.modelHasInheritance = this.EdmModel.SchemaElementsAcrossModels()
                             .OfType<IEdmStructuredType>().Any(t => !t.FullTypeName().StartsWith("Org.OData.Authorization.V1") &&
-                             !t.FullTypeName().StartsWith("Org.OData.Capabilities.V1") && t.BaseType != null);
```
madansr7: @xuzhg we removed the basetype null check?

xuzhg: Yes. Skip it in the Code Generator.

PR #1448, opened 2019-04-16
Refactor & Improve the ODataMediaTypeResolver

xuzhg:
### Issues

* Refactor & Improve the ODataMediaTypeResolver

### Description

If i use the following codes to snapshot the memory usage:
image

I can get two snapshots, one is at Line 15, the other is at line 17.

### Before my change

The old memory usage likes:
image

### Use my changes

The new memory usage likes:
image

I can use the following tables to illustrate the changes:

|<div align="right">Memory item</div>|  Before my change |   With my change | Percentage down |
|:------------------|--------:|--------:|--------:|
|New Object count |+405|+167| 59% |
|New Heap Size |+23.42kb|+7.93kb| 66%|

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/MediaTypeUtils.cs
+++ b/src/Microsoft.OData.Core/MediaTypeUtils.cs
@@ -58,7 +58,7 @@ namespace Microsoft.OData
         /// <summary>
         /// Max size to cache match info.
         /// </summary>
-        private const int MatchInfoCacheMaxSize = 1024;
+        private const int MatchInfoCacheMaxSize = 256;
 
         /// <summary>
         /// Concurrent cache to cache match info.
@@ -897,7 +897,7 @@ namespace Microsoft.OData
         /// <summary>
         /// Class representing key of match info cache.
         /// </summary>
-        private sealed class MatchInfoCacheKey
+        private sealed class MatchInfoCacheKey : IEquatable<MatchInfoCacheKey>
         {
             /// <summary>
             /// Constructor.
@@ -934,11 +934,29 @@ namespace Microsoft.OData
             /// <returns>true if obj is equal to this instance; otherwise, false.</returns>
             public override bool Equals(object obj)
             {
-                MatchInfoCacheKey cacheKey = obj as MatchInfoCacheKey;
-                return cacheKey != null &&
-                    this.MediaTypeResolver == cacheKey.MediaTypeResolver &&
-                    this.PayloadKind == cacheKey.PayloadKind &&
-                    this.ContentTypeName == cacheKey.ContentTypeName;
+                return this.Equals(obj as MatchInfoCacheKey);
+            }
+
+            /// <summary>
+            /// Returns a value indicating whether this instance is equal to a specified <see cref="MatchInfoCacheKey"/>.
+            /// </summary>
+            /// <param name="other">An object to compare with this instance.</param>
+            /// <returns>true if obj is equal to this instance; otherwise, false.</returns>
+            public bool Equals(MatchInfoCacheKey other)
+            {
+                if (ReferenceEquals(other, null))
+                {
+                    return false;
+                }
+
+                if (ReferenceEquals(this, other))
+                {
+                    return true;
+                }
+
+                return this.MediaTypeResolver == other.MediaTypeResolver &&
+                    this.PayloadKind == other.PayloadKind &&
+                    this.ContentTypeName == other.ContentTypeName;
             }
 
             /// <summary>
--- a/src/Microsoft.OData.Core/ODataMediaTypeResolver.cs
+++ b/src/Microsoft.OData.Core/ODataMediaTypeResolver.cs
@@ -11,127 +11,16 @@ namespace Microsoft.OData
     using System.Collections.Generic;
     using System.Linq;
     #endregion Namespaces
+
     /// <summary>
     /// Class with the responsibility of resolving media types (MIME types) into formats and payload kinds.
     /// </summary>
     public class ODataMediaTypeResolver
     {
-        /// <summary>application/xml media type</summary>
-        private static readonly ODataMediaType ApplicationXmlMediaType = new ODataMediaType(MimeConstants.MimeApplicationType, MimeConstants.MimeXmlSubType);
-
-        /// <summary>application/json media type</summary>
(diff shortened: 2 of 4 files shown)


src/Microsoft.OData.Core/ODataMediaTypeResolver.cs
```diff
@@ -142,25 +31,63 @@ public class ODataMediaTypeResolver
...
         /// <summary>
-        /// MediaTypeResolver without atom support
+        /// The special map of payload kind and supported media type.
         /// </summary>
-        private static readonly ODataMediaTypeResolver MediaTypeResolver = new ODataMediaTypeResolver();
+        private static IDictionary<ODataPayloadKind, IEnumerable<ODataMediaTypeFormat>> SpecialMediaTypeFormat = new Dictionary<ODataPayloadKind, IEnumerable<ODataMediaTypeFormat>>
+        {
+            {
+                ODataPayloadKind.Batch,
+                new List<ODataMediaTypeFormat>
```
KanishManuja-MS: >List [](start = 20, length = 4)

Why use a list here? Arrays use lesser memory.

KanishManuja-MS:
While this is an improvement, I don't think this solves the problem reported by Yogesh. The problem reported by Yogesh was with MatchInfoCache. This change should save ~50kb but the actual memory issue is way more than that (6% of the rooted memory as reported).

src/Microsoft.OData.Core/ODataMediaTypeResolver.cs
xuzhg: #Resolved.

xuzhg:
The concatenate only happens for the "Batch" payload. Not for others.
Besides, as i said before, it's one part. another part is to investigate the ConcurrentDictionary<...> used in Cache.

xuzhg:
@ysanghi Would you please take a look this part changes and leave your comments about the Cache issue

src/Microsoft.OData.Core/MediaTypeUtils.cs
```diff
@@ -58,7 +58,7 @@ internal static class MediaTypeUtils
         /// <summary>
         /// Max size to cache match info.
         /// </summary>
-        private const int MatchInfoCacheMaxSize = 1024;
+        private const int MatchInfoCacheMaxSize = 256;
```
KanishManuja-MS: Maybe keeping at 1024 is fine or making this customizable, so that customers can experiment without waiting for a new release.

src/Microsoft.OData.Core/ODataMediaTypeResolver.cs
```diff
@@ -41,41 +41,23 @@ public class ODataMediaTypeResolver
...
             },
             {
                 ODataPayloadKind.Asynchronous,
-                new List<ODataMediaTypeFormat>
-                {
-                    new ODataMediaTypeFormat(new ODataMediaType(MimeConstants.MimeApplicationType, MimeConstants.MimeHttpSubType), ODataFormat.RawValue)
-                }
-            },
-            {
-                ODataPayloadKind.Delta, new List<ODataMediaTypeFormat>()
```
KanishManuja-MS: This does not look right. Maybe omitting this is causing the test case to be stuck forever.

src/Microsoft.OData.Core/ODataMediaTypeResolver.cs
```diff
@@ -182,92 +102,75 @@ internal static ODataMediaTypeResolver GetMediaTypeResolver(IServiceProvider con
...
-        /// Clones the default media types.
-        /// </summary>
-        /// <returns>The cloned media type table.</returns>
-        private static List<ODataMediaTypeFormat>[] CloneDefaultMediaTypes()
+        private static IEnumerable<ODataMediaTypeFormat> SetJsonLightMediaTypes()
         {
-            ODataMediaTypeFormat[][] mediaTypes = defaultMediaTypes;
+            var minimal = new KeyValuePair<string, string>(MimeConstants.MimeMetadataParameterName, MimeConstants.MimeMetadataParameterValueMinimal);
+            var full = new KeyValuePair<string, string>(MimeConstants.MimeMetadataParameterName, MimeConstants.MimeMetadataParameterValueFull);
+            var non = new KeyValuePair<string, string>(MimeConstants.MimeMetadataParameterName, MimeConstants.MimeMetadataParameterValueNone);
```
mikepizzo: >non [](start = 16, length = 3)

nit: "none"?

src/Microsoft.OData.Core/MediaTypeUtils.cs
```diff
@@ -934,11 +934,34 @@ public MatchInfoCacheKey(ODataMediaTypeResolver resolver, ODataPayloadKind paylo
...
+
+                if (ReferenceEquals(this, other))
+                {
+                    return true;
+                }
+
+                if (this.GetType() != other.GetType())
+                {
+                    return false;
+                }
```
mikepizzo: What is this checking for? When would the types be different?

mikepizzo:
Couple minor comments; otherwise looks fine.

src/Microsoft.OData.Core/MediaTypeUtils.cs
xuzhg: You are right. It's unnecessary. 

---
In reply to: [285791784](https://github.com/OData/odata.net/pull/1448#discussion_r285791784) [](ancestors = 285791784)

src/Microsoft.OData.Core/MediaTypeUtils.cs
xuzhg: I 'd like to generate a nightly build for yogesh to verify the memory usage.

---
In reply to: [283560641](https://github.com/OData/odata.net/pull/1448#discussion_r283560641) [](ancestors = 283560641)

src/Microsoft.OData.Core/ODataMediaTypeResolver.cs
xuzhg: The original is wrong. ODataPayloadKind.Delta should same as ODataPayloadKind.Resource.  #Resolved.

xuzhg:
Merged at: e4be823

PR #1449, opened 2019-04-16
Support JSON Metadata writer

xuzhg:
### Issues

* Support JSON Metadata writer

### Description

* Support JSON Metadata writer
* Provide the public APIs for customer to use to output JSON metadata

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Csdl/CsdlWriter.cs
+++ b/src/Microsoft.OData.Edm/Csdl/CsdlWriter.cs
@@ -104,16 +104,22 @@ namespace Microsoft.OData.Edm.Csdl
             return true;
         }
 
-        private void WriteEdmxElement()
-        {
-            this.writer.WriteStartElement(CsdlConstants.Prefix_Edmx, CsdlConstants.Element_Edmx, this.edmxNamespace);
-            this.writer.WriteAttributeString(CsdlConstants.Attribute_Version, GetVersionString(this.edmxVersion));
-        }
-
         /// <summary>
         /// Write CSDL output.
         /// </summary>
         protected abstract void WriteCsdl();
+
+        // Gets the string form of the EdmVersion.
+        // Note that Version 4.01 needs two digits of minor version precision.
+        protected static string GetVersionString(Version version)
+        {
+            if (version == EdmConstants.EdmVersion401)
+            {
+                return EdmConstants.EdmVersion401String;
+            }
+
+            return version.ToString();
+        }
 
         private static bool Verify(IEdmModel model, out Version edmxVersion, out IEnumerable<EdmError> errors)
         {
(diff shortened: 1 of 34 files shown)


src/Microsoft.OData.Edm/Csdl/CsdlWriter.cs
```diff
@@ -7,159 +7,141 @@
...
 using Microsoft.OData.Edm.Validation;
 
 namespace Microsoft.OData.Edm.Csdl
 {
     /// <summary>
-    /// Provides CSDL serialization services for EDM models.
+    /// Provides CSDL serialization services (XML & JSON) for EDM models.
     /// </summary>
-    public class CsdlWriter
+    public abstract class CsdlWriter
```
KanishManuja-MS: Changing a concrete class to an abstract class will be a major breaking change. Will this be targetted for 8.x?

xuzhg: let me share my thoughts first and  We can talk it more later

1. First, if you check original `CsdlWriter`, the constructor is private, customer can't instantiate it. From this, make abstract is same.

2.   `CsdlWriter` hasn't any public instance methods.

From these, I think it's better to mark it as abstract.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
```diff
@@ -8,23 +8,45 @@
...
+        internal EdmModelCsdlSerializationVisitor(IEdmModel model, IEdmJsonWriter jsonWriter, Version edmVersion)
+            : base(model)
+        {
+            this.edmVersion = edmVersion;
+            this.schemaWriter = new EdmModelCsdlSchemaJsonWriter(model, jsonWriter, this.edmVersion);
+            this.namespaceAliasMappings = this.schemaWriter.NamespaceAliasMappings;
+            this.isXml = false;
+        }
+
+        internal EdmModelCsdlSerializationVisitor(IEdmModel model, EdmModelCsdlSchemaWriter edmWriter, CsdlWriterSettings settings, Version edmVersion)
```
xuzhg: >EdmModelCsdlSchemaWriter edmWriter [](start = 67, length = 34)

Maybe it's a better solution

src/Microsoft.OData.Edm/Csdl/Json/CsdlWriterSettings.cs
```diff
@@ -0,0 +1,36 @@
...
+//---------------------------------------------------------------------
+
+using System;
+
+namespace Microsoft.OData.Edm.Csdl.Json
+{
+    /// <summary>
+    /// Configuration settings for CSDL writers.
+    /// </summary>
+    public class CsdlWriterSettings
```
xuzhg: >CsdlWriterSettings [](start = 17, length = 18)

csdlJsonWriterSetting?

mikepizzo:
A few issues with the generated JSON versus the final OASIS spec:
1) Document must include a [$Version](http://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html#_Toc26353367)
2) Document must include a [$EntityContainer] (http://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html#sec_CSDLJSONDocument) property whose string value is the namespace-qualified name of the property representing the entity container (whose kind is $Kind:"EntityContainer" 
3) "IsComposable" is missing the leading "$" => Should be "$IsComposable"
4) Boolean value “true” should be without quotes => true
5) $Kind: EntitySet => should just be $Collection:true (http://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html#_Toc26353414) with $Type of the entity type.
6) $Kind:Singleton => should just be the entity type (http://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html#sec_Singleton)
7) strings must be JSON encoded: " => /"  and / => //

xuzhg:
Closed as duplicated at #1775

PR #1454, opened 2019-04-27
Support versioned preferences for reading/writing OData prefix

mikepizzo:
### Description

ODataSimplifiedOptions controls, among other things, whether the OData prefix is included when writing (or reading) JSON responses.

Currently, the preference for whether to include the prefix is version-independent. However, clients like WebAPI OData that have a single instance of ODataSimplifiedOptions, may want to have different behavior depending on whether the response is a 4.0 payload or a 4.01 payload.

This change introduces a new version-specific method for whether to read/write the odata prefix, and exposes the version specific information on the jsonlight context.

- [ x] *Test cases added*
- [ x] *Build and test with one-click build and test script passed*

### Additional work necessary


src/Microsoft.OData.Core/ODataSimplifiedOptions.cs
```diff
@@ -88,6 +110,71 @@ public ODataSimplifiedOptions Clone()
...
+
+        /// <summary>
+        /// Version-specific behavior for writing OData control information without a prefix
+        /// True if control information can be read without prefix 'odata.', otherwise false.
+        /// The default value is false for OData 4.0 and true for OData 4.01.
+        /// The option is applied during deserialization.
+        /// </summary>
+        /// <param name="version">The version of the version-specific behavior being requested.</param>
+        /// <returns>Whether to omit the OData prefix for the specified version.</returns>
+        public bool GetOmitODataPrefix(ODataVersion version)
```
xuzhg: >ODataVersion version [](start = 39, length = 20)

why cannot use the "version" in the constructor? It's a little bit strange?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightInputContext.cs
```diff
@@ -151,6 +151,23 @@ internal Stream Stream
             }
         }
 
+        /// <summary>
+        /// Returns whether to read odata control information without the odata prefix.
+        /// True for OData 4.01 and greater. Settable for OData 4.0
+        /// </summary>
+        internal bool OptionalODataPrefix
```
xuzhg: >OptionalODataPrefix [](start = 22, length = 19)

omit or optional ?

Maybe it's a little bit confusing for maintenance.

test/FunctionalTests/Microsoft.OData.Core.Tests/Build.NetFramework/Microsoft.OData.Core.Tests.csproj
```diff
@@ -157,7 +157,7 @@
     <Compile Include="..\JsonLight\ODataJsonLightUtilsTests.cs" />
     <Compile Include="..\JsonLight\ODataJsonLightValidationUtilsTests.cs" />
     <Compile Include="..\JsonLight\ODataJsonLightValueSerializerTests.cs" />
-    <Compile Include="..\JsonLight\PayloadValueConverters.cs" />
+    <Compile Include="..\JsonLight\ODataSimplifiedOptionsTests.cs" />    <Compile Include="..\JsonLight\PayloadValueConverters.cs" />
```
xuzhg: ><Compile Include="..\JsonLight\PayloadValueConverters.cs" /> [](start = 73, length = 60)

 accident?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataSimplifiedOptionsTests.cs
```diff
@@ -0,0 +1,160 @@
...
+            var options = new ODataSimplifiedOptions(version);
+            options.EnableWritingODataAnnotationWithoutPrefix = true;
+            options.GetOmitODataPrefix().Should().BeTrue();
+            options.GetOmitODataPrefix(ODataVersion.V4).Should().BeTrue();
+            options.GetOmitODataPrefix(ODataVersion.V401).Should().BeTrue();
+            options.EnableWritingODataAnnotationWithoutPrefix.Should().BeTrue();
+        }
+
+        [Fact]
+        public void SetEnableWritingODataAnnotationWithoutPrefixFalse()
```
xuzhg: >SetEnableWritingODataAnnotationWithoutPrefixFalse() [](start = 20, length = 51)

why does not it have the InlineData input? By design?

src/Microsoft.OData.Core/ODataSimplifiedOptions.cs
```diff
@@ -33,12 +42,12 @@ public ODataSimplifiedOptions(ODataVersion? version)
...
             {
                 this.EnableReadingODataAnnotationWithoutPrefix = false;
-                this.EnableWritingODataAnnotationWithoutPrefix = false;
+                this.enableWritingODataAnnotationWithoutPrefix = this.omitODataPrefix40;
             }
             else
             {
                 this.EnableReadingODataAnnotationWithoutPrefix = true;
-                this.EnableWritingODataAnnotationWithoutPrefix = true;
+                this.enableWritingODataAnnotationWithoutPrefix = omitODataPrefix;
```
KanishManuja-MS: nit: Should use this.omitODataPrefix here as well?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightInputContext.cs
KanishManuja-MS: If my understanding is correct optional is fine. Since we are going to support reading both; it is optional. Whereas while writing we support only one thing (omit).

src/Microsoft.OData.Core/ODataSimplifiedOptions.cs
mikepizzo: ODataSimplifiedOptions is used by WebAPI OData to set options for the entire service. The challenge is that the same service may support both OData 4.0 and OData 4.01, and may want different settings of omitting the prefix for each.

---
In reply to: [279946255](https://github.com/OData/odata.net/pull/1454#discussion_r279946255) [](ancestors = 279946255)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightInputContext.cs
mikepizzo: Right; see Kanish's comment: this property means that it's optional on reading whether the prefix is there or not, versus writing where we either omit or we don't.

---
In reply to: [279946882](https://github.com/OData/odata.net/pull/1454#discussion_r279946882) [](ancestors = 279946882)

mikepizzo: xactly.:grinning: 

---
In reply to: [285693803](https://github.com/OData/odata.net/pull/1454#discussion_r285693803) [](ancestors = 285693803)

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataSimplifiedOptionsTests.cs
mikepizzo: This test was intended to test the default instructor; the other tests passed in the version for the constructor. I did add a version of Set...True() that worked with the default constructor, although it was kind of already covered in the defaults test, since true is a default.  Still, good to verify.

---
In reply to: [279948328](https://github.com/OData/odata.net/pull/1454#discussion_r279948328) [](ancestors = 279948328)

test/FunctionalTests/Microsoft.OData.Core.Tests/Build.NetFramework/Microsoft.OData.Core.Tests.csproj
mikepizzo: yes; thanks.

---
In reply to: [279947483](https://github.com/OData/odata.net/pull/1454#discussion_r279947483) [](ancestors = 279947483)

src/Microsoft.OData.Core/ODataSimplifiedOptions.cs
mikepizzo: yes; thanks.

---
In reply to: [283528439](https://github.com/OData/odata.net/pull/1454#discussion_r283528439) [](ancestors = 283528439)

PR #1456, opened 2019-04-29
Fix #1455 - Remove bogus validation error for EntitySetPath

### LINKED ISSUE #1455: Bogus EntitySetPath error
*When defining a function or an action, the optional EntitySetPath specifies the collection where any returned entities live. If the return type is a collection of entities, the EntitySetPath (if specified) must reference a collection -- if the return type is a single entity, then the path will typically be a collection but may be a singleton or single-valued containment nav prop.  Currently, validating a model throws an invalid error if the return type is a single entity and the entitysetpath indicates a collection.*

### Assemblies affected

*OData .Net lib 7.5.x*

### Reproduce steps

### Expected result

*Validation would succeed with no errors.*

### Actual result

"Error: The operation 'A1' entity set path was determined to be a collection but the return type is not a collection"

mikepizzo:
### Issues

*This pull request fixes issue #1455.*

### Description

*Removes bogus validation error*

- [x ] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Edm/Microsoft.OData.Edm.txt
+++ b/src/Microsoft.OData.Edm/Microsoft.OData.Edm.txt
@@ -87,7 +87,6 @@ EdmModel_Validator_Semantic_InvalidEntitySetPathWithNonEntityBindingParameter=Th
 EdmModel_Validator_Semantic_InvalidEntitySetPathUnknownTypeCastSegment=The attribute '{0}' has an invalid value. The path '{1}' has a type cast segment '{2}' that cannot be found in the model.
 EdmModel_Validator_Semantic_OperationWithEntitySetPathReturnTypeInvalid=The operation '{0}' has an entity set path and with an invalid return type. The return type is required to be an entity type or a collection of entity type.
 EdmModel_Validator_Semantic_OperationWithEntitySetPathAndReturnTypeTypeNotAssignable=The operation '{0}' entity set path determined entity type '{1}' is not assignable to the return type '{2}'.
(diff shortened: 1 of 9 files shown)


src/Microsoft.OData.Edm/Microsoft.OData.Edm.txt
```diff
@@ -87,7 +87,6 @@ EdmModel_Validator_Semantic_InvalidEntitySetPathWithNonEntityBindingParameter=Th
 EdmModel_Validator_Semantic_InvalidEntitySetPathUnknownTypeCastSegment=The attribute '{0}' has an invalid value. The path '{1}' has a type cast segment '{2}' that cannot be found in the model.
 EdmModel_Validator_Semantic_OperationWithEntitySetPathReturnTypeInvalid=The operation '{0}' has an entity set path and with an invalid return type. The return type is required to be an entity type or a collection of entity type.
 EdmModel_Validator_Semantic_OperationWithEntitySetPathAndReturnTypeTypeNotAssignable=The operation '{0}' entity set path determined entity type '{1}' is not assignable to the return type '{2}'.
-EdmModel_Validator_Semantic_OperationWithEntitySetPathResolvesToEntityTypeMismatchesCollectionEntityTypeReturnType=The operation '{0}' entity set path was determined to be a collection but the return type is not a collection.
```
xuzhg: need to "run custom tool" on Microsoft.OData.Edm.tt file

xuzhg:
Merged. Thanks.

src/Microsoft.OData.Edm/Microsoft.OData.Edm.txt
xuzhg: I resolved it.

PR #1463, opened 2019-05-13
Validate the resource type and resource set type in the same inheritance tree

xuzhg:
### Issues

### Description

* Enable to validate the resource type and resource set type in the inheritance tree

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ResourceSetWithoutExpectedTypeValidator.cs
+++ b/src/Microsoft.OData.Core/ResourceSetWithoutExpectedTypeValidator.cs
@@ -65,6 +65,12 @@ namespace Microsoft.OData
                 throw new ODataException(Strings.ResourceSetWithoutExpectedTypeValidator_IncompatibleTypes(itemType.FullTypeName(), this.itemType.FullTypeName()));
             }
 
+            // Make sure the types are in the same inheritance tree
+            if (!(itemType.IsAssignableFrom(this.itemType) || this.itemType.IsAssignableFrom(itemType)))
+            {
+                throw new ODataException(Strings.ResourceSetWithoutExpectedTypeValidator_IncompatibleTypes(itemType.FullTypeName(), this.itemType.FullTypeName()));
+            }
+
             this.itemType = (IEdmType)commonBaseType;
         }
     }
--- a/src/Microsoft.OData.Core/ResourceSetWithoutExpectedTypeValidator.cs
+++ b/src/Microsoft.OData.Core/ResourceSetWithoutExpectedTypeValidator.cs
@@ -58,20 +58,11 @@ namespace Microsoft.OData
                 throw new ODataException(Strings.ResourceSetWithoutExpectedTypeValidator_IncompatibleTypes(itemType.FullTypeName(), this.itemType.FullTypeName()));
             }
 
-            // If the types are not equivalent, make sure they have a common base type.
-            IEdmType commonBaseType = EdmLibraryExtensions.GetCommonBaseType(thisStructuredType, structuredType);
-            if (commonBaseType == null)
+            // Make sure the resource types is same or derived type of expected type
+            if (!this.itemType.IsAssignableFrom(itemType))
             {
                 throw new ODataException(Strings.ResourceSetWithoutExpectedTypeValidator_IncompatibleTypes(itemType.FullTypeName(), this.itemType.FullTypeName()));
             }
-
-            // Make sure the types are in the same inheritance tree
-            if (!(itemType.IsAssignableFrom(this.itemType) || this.itemType.IsAssignableFrom(itemType)))
-            {
-                throw new ODataException(Strings.ResourceSetWithoutExpectedTypeValidator_IncompatibleTypes(itemType.FullTypeName(), this.itemType.FullTypeName()));
-            }
-
-            this.itemType = (IEdmType)commonBaseType;
         }
     }
 }
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/ODataWriterDerivedTypeConstraintTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/ODataWriterDerivedTypeConstraintTests.cs
@@ -751,8 +751,8 @@ namespace Microsoft.OData.Tests
             {
                 ODataWriter writer = messageWriter.CreateODataResourceSetWriter(directoryObjects, directoryObject); // entityset
                 writer.WriteStart(new ODataResourceSet());
-                writer.WriteStart(group);
-                writer.WriteEnd();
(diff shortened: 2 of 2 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/ODataWriterDerivedTypeConstraintTests.cs
```diff
@@ -731,5 +731,113 @@ private static void SetDerivedTypeAnnotation(EdmModel model, IEdmVocabularyAnnot
...
+            {
+                TypeName = "microsoft.graph.contact",
+                Properties = new[] { new ODataProperty { Name = "id", Value = "abc" }, new ODataProperty { Name = "displayName", Value = "xyz" } } /*open type*/
+            };
+
+            Action<ODataMessageWriter> writeAction = (messageWriter) =>
+            {
+                ODataWriter writer = messageWriter.CreateODataResourceSetWriter(directoryObjects, directoryObject); // entityset
+                writer.WriteStart(new ODataResourceSet());
+                    writer.WriteStart(contact);
```
KanishManuja-MS: Is this intentionally indented in this way?

xuzhg: Yes

KanishManuja-MS:
>             if (this.itemType == null)

This will never reach here. 

---
Refers to: src/Microsoft.OData.Core/ResourceSetWithoutExpectedTypeValidator.cs:48 in 94a19c5. [](commit_id = 94a19c5f10c8226f7016b09853bd0eeb595ddaff, deletion_comment = False)

KanishManuja-MS:
>             }

This statement will never get executed. 

---
Refers to: src/Microsoft.OData.Core/ResourceSetWithoutExpectedTypeValidator.cs:51 in 94a19c5. [](commit_id = 94a19c5f10c8226f7016b09853bd0eeb595ddaff, deletion_comment = False)

src/Microsoft.OData.Core/ResourceSetWithoutExpectedTypeValidator.cs
```diff
@@ -71,6 +71,12 @@ internal void ValidateResource(IEdmType itemType)
                 throw new ODataException(Strings.ResourceSetWithoutExpectedTypeValidator_IncompatibleTypes(itemType.FullTypeName(), this.itemType.FullTypeName()));
             }
 
+            // Make sure the types are in the same inheritance tree
+            if (!(itemType.IsAssignableFrom(this.itemType) || this.itemType.IsAssignableFrom(itemType)))
```
KanishManuja-MS: >| this.itemType.IsAssignableFrom(itemType) [](start = 60, length = 42)

This would be correct if the expected type was unknown but it seems like this validator is always constructed with an IEdmType. Is the class wrongly named? May be we should check this only when no item type is provided in the constructor.

KanishManuja-MS:
>                 this.itemType = itemType;

This statement will never get executed. 

---
Refers to: src/Microsoft.OData.Core/ResourceSetWithoutExpectedTypeValidator.cs:50 in 94a19c5. [](commit_id = 94a19c5f10c8226f7016b09853bd0eeb595ddaff, deletion_comment = False)

src/Microsoft.OData.Core/ResourceSetWithoutExpectedTypeValidator.cs
mikepizzo: I think there are two cases:
1) We were provided an expected type, and we want to validate that the type is assignable from the expected type (and not the other way around)
2) We "guessed" the expected type, based on the type of the first item in the collection. In this case, we should support types in the same type hierarchy (although, technically, with the introduction of the abstract base types Edm.Primitive, Edm.ComplexType, Edm.EntityType, and Edm.Untyped, even this validation may not be valid). So, we might just want to remove the logic that guesses at the type based on the type of the first element.

---
In reply to: [283918035](https://github.com/OData/odata.net/pull/1463#discussion_r283918035) [](ancestors = 283918035)

xuzhg:
@mikepizzo Changed the codes, please take a look. thanks.

mikepizzo:
>         internal void ValidateResource(IEdmType itemType)

What if the expected type (this.itemType) is Edm.PrimitiveType, Edm.ComplexType, or Edm.EntityType? Is this already handled someplace else? I would expect that, for Edm.PrimitiveType, we should allow any typekind of primitive, enum or typedefinition; for Edm.ComplexType we should allow typekind of complex, and for Edm.EntityType we should allow typekind of entity or entityreference. 

---
Refers to: src/Microsoft.OData.Core/ResourceSetWithoutExpectedTypeValidator.cs:40 in 9c90adf. [](commit_id = 9c90adf7c3d6b8dfe68525262a1b7cac30c08440, deletion_comment = False)

mikepizzo:
See my comment on expected abstract types Edm.PrimitiveType, Edm.ComplexType, and Edm.EntityType.

---
In reply to: [494512617](https://github.com/OData/odata.net/pull/1463#issuecomment-494512617) [](ancestors = 494512617)

xuzhg:
>         internal void ValidateResource(IEdmType itemType)

Edm.PrimitiveType, Edm.EntityType and Edm.ComplexType are processed in 'IsAssignableFrom(..)' method, so that's supported already.

However, the existing codes from Line (53 - 59) assumes that the validate type is structure type. I don't understand why these old codes can work for "Edm.PrimitiveType"? 

---
In reply to: [497363315](https://github.com/OData/odata.net/pull/1463#issuecomment-497363315) [](ancestors = 497363315)

---
Refers to: src/Microsoft.OData.Core/ResourceSetWithoutExpectedTypeValidator.cs:40 in 9c90adf. [](commit_id = 9c90adf7c3d6b8dfe68525262a1b7cac30c08440, deletion_comment = False)

xuzhg:
I assume the Line (53-58) should be removed? 

---
In reply to: [497363972](https://github.com/OData/odata.net/pull/1463#issuecomment-497363972) [](ancestors = 497363972,494512617)

mikepizzo:
That seems right, although I would like to understand if this code path never gets hit for collections of primitive types.

---
In reply to: [497416559](https://github.com/OData/odata.net/pull/1463#issuecomment-497416559) [](ancestors = 497416559,497363972,494512617)

xuzhg:
@mikepizzo 

OData Spec says:

· Collection(Edm.PrimitiveType)
*  cannot be used as the type of a property.
*  cannot be used as the return type of a function.

So, does it mean the "Edm.PrimitiveType" is not a valid scenario in the ResourceSet?

xuzhg:
Merged, thanks.

PR #1464, opened 2019-05-14
Enable to write the nextlink for the collection of entity reference links

xuzhg:
…inks

### Issues

### Description

* In the Response payload, we should allow to write the next link for the collection of entity reference links.

This PR is supporting this.

Type | version | Request | Response
-- | -- | -- | --
Single Navigation Property | 4.0 | {       "@odata.type":"#Northwind.Manager",       "ID": 1,     "DirectReport@odata.bind":   "http://host/service/Employees(5)"   } | {       "@odata.type":"#Northwind.Manager",       "ID": 1,     "DirectReport   ": {        \"@odata.id\":\"Orders(2)\"     }   }
Single Navigation Property | 4.01 | {       "@odata.type":"#Northwind.Manager",       "ID": 1,     "DirectReport":   {   "@id": "Employees(5)"     }   } | {       "@odata.type":"#Northwind.Manager",     "DirectReport":   {         "@id": "Orders(5)"     }   }
Collection Navigation Property | 4.0 | {       "@odata.type":"#Northwind.Manager",       "ID": 1,     "DirectReports@odata.bind":   [         "http://host/service/Employees(5)",         "http://host/service/Employees(6)"     ]   } | {       "@odata.type":"#Northwind.Manager",       "ID": 1,     "DirectReports   ": [        {\"@odata.id\":\"Orders(2)\"},        {\"@odata.id\":\"Orders(3)\"}     ]   }
Collection Navigation Property | 4.01 | {       "@odata.type":"#Northwind.Manager",       "ID": 1,     "DirectReports":   [         {"@odata.id": "Employees(5)"},         {"@odata.id": "Employees(6)"}     ]   } | {       "@odata.type":"#Northwind.Manager",     "DirectReports":   [         {"@id": "Orders(5)"},         {"@id": "Orders(6)"}     ]   }

# 05/21/2019 updates

We should make the request and response writing with the same pattern, and avoid breaking changes in the request of collection nested resource by enabling the following patterns, 

### Old pattern in request
```
WriteStart( nestedResourceInfo ) // collection
      WriteEntityReferenceLink( EntityReferenceLink)
      WriteEntityReferenceLink( EntityReferenceLink)
      WriteEntityReferenceLink( EntityReferenceLink)
WriteEnd() // end of nested resource info
```
should remove the old pattern in the next breaking change version.

### New pattern in request (Need more discussion)
```
WriteStart( nestedResourceInfo ) // collection
   WriteStart( new ODataResourceSet() ) // Add a resource set writing logic
      WriteEntityReferenceLink( EntityReferenceLink)
      WriteEntityReferenceLink( EntityReferenceLink)
      WriteEntityReferenceLink( EntityReferenceLink)
  WriteEnd() // end of resource set
WriteEnd() // end of nested resource info
```

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
@@ -988,13 +988,13 @@ namespace Microsoft.OData.JsonLight
 
             // If we wrote entity reference links for a collection navigation property but no
             // resource set afterwards, we have to now close the array of links.
-            if (navigationLinkScope.EntityReferenceLinkWritten && !navigationLinkScope.ResourceSetWritten && nestedResourceInfo.IsCollection.Value)
-            {
-                this.jsonWriter.EndArrayScope();
-            }
-
             if (!this.jsonLightOutputContext.WritingResponse)
             {
+                if (navigationLinkScope.EntityReferenceLinkWritten && !navigationLinkScope.ResourceSetWritten && nestedResourceInfo.IsCollection.Value)
+                {
+                    this.jsonWriter.EndArrayScope();
+                }
+
                 // In requests, the nested resource info may have multiple entries in multiple resource sets in it; if we
                 // wrote at least one resource set, close the resulting array here.
                 if (navigationLinkScope.ResourceSetWritten)
@@ -1018,32 +1018,51 @@ namespace Microsoft.OData.JsonLight
             // In JSON Light, we can only write entity reference links at the beginning of a navigation link in requests;
             // once we wrote a resource set, entity reference links are not allowed anymore (we require all the entity reference
             // link to come first because of the grouping in the JSON Light wire format).
-            JsonLightNestedResourceInfoScope navigationLinkScope = (JsonLightNestedResourceInfoScope)this.CurrentScope;
-            if (navigationLinkScope.ResourceSetWritten)
+            JsonLightNestedResourceInfoScope nestedResourceScope = this.CurrentScope as JsonLightNestedResourceInfoScope;
+            if (nestedResourceScope == null)
+            {
+                nestedResourceScope = this.ParentNestedResourceInfoScope as JsonLightNestedResourceInfoScope;
+            }
+
+            if (nestedResourceScope.ResourceSetWritten)
             {
                 throw new ODataException(Strings.ODataJsonLightWriter_EntityReferenceLinkAfterResourceSetInRequest);
             }
 
-            if (!navigationLinkScope.EntityReferenceLinkWritten)
+            if (!nestedResourceScope.EntityReferenceLinkWritten)
             {
-                if (!this.jsonLightOutputContext.WritingResponse &&
-                    (this.Version == null || this.Version < ODataVersion.V401))
+                // In request
+                if (!this.jsonLightOutputContext.WritingResponse)
                 {
-                    // Write the property annotation for the entity reference link(s)
-                    this.odataAnnotationWriter.WritePropertyAnnotationName(parentNestedResourceInfo.Name, ODataAnnotationNames.ODataBind);
+                    if (this.Version == null || this.Version < ODataVersion.V401)
+                    {
+                        // Write the property annotation for the entity reference link(s)
+                        this.odataAnnotationWriter.WritePropertyAnnotationName(parentNestedResourceInfo.Name, ODataAnnotationNames.ODataBind);
+                    }
+                    else
+                    {
+                        this.jsonWriter.WriteName(parentNestedResourceInfo.Name);
+                    }
+
+                    Debug.Assert(parentNestedResourceInfo.IsCollection.HasValue, "parentNestedResourceInfo.IsCollection.HasValue");
+                    if (parentNestedResourceInfo.IsCollection.Value)
+                    {
+                        // write [ for the collection
+                        this.jsonWriter.StartArrayScope();
+                    }
                 }
-                else
+                else // In response
                 {
-                    this.jsonWriter.WriteName(parentNestedResourceInfo.Name);
+                    Debug.Assert(parentNestedResourceInfo.IsCollection.HasValue, "parentNestedResourceInfo.IsCollection.HasValue");
+                    if (!parentNestedResourceInfo.IsCollection.Value)
+                    {
+                        // Write the property name for single-nested resource,
+                        // for the collection nested resource, it's write at top level when writing ODataResourceSet
+                        this.jsonWriter.WriteName(parentNestedResourceInfo.Name);
+                    }
                 }
 
-                Debug.Assert(parentNestedResourceInfo.IsCollection.HasValue, "parentNestedResourceInfo.IsCollection.HasValue");
-                if (parentNestedResourceInfo.IsCollection.Value)
-                {
-                    this.jsonWriter.StartArrayScope();
-                }
-
-                navigationLinkScope.EntityReferenceLinkWritten = true;
+                nestedResourceScope.EntityReferenceLinkWritten = true;
             }
 
             if (!this.jsonLightOutputContext.WritingResponse &&
--- a/src/Microsoft.OData.Core/ODataWriterCore.cs
+++ b/src/Microsoft.OData.Core/ODataWriterCore.cs
@@ -1800,7 +1800,7 @@ namespace Microsoft.OData
 
             this.CheckForNestedResourceInfoWithContent(ODataPayloadKind.EntityReferenceLink, null);
             Debug.Assert(
-                this.CurrentScope.Item is ODataNestedResourceInfo,
+                this.CurrentScope.Item is ODataNestedResourceInfo || this.ParentNestedResourceInfoScope.Item is ODataNestedResourceInfo,
                 "The CheckForNestedResourceInfoWithContent should have verified that entity reference link can only be written inside a nested resource info.");
 
             if (!this.SkipWriting)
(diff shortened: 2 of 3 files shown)


src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
```diff
@@ -1018,32 +1021,51 @@ protected override void WriteEntityReferenceInNavigationLinkContent(ODataNestedR
             // In JSON Light, we can only write entity reference links at the beginning of a navigation link in requests;
             // once we wrote a resource set, entity reference links are not allowed anymore (we require all the entity reference
             // link to come first because of the grouping in the JSON Light wire format).
-            JsonLightNestedResourceInfoScope navigationLinkScope = (JsonLightNestedResourceInfoScope)this.CurrentScope;
-            if (navigationLinkScope.ResourceSetWritten)
+            JsonLightNestedResourceInfoScope nestedResourceScope = this.CurrentScope as JsonLightNestedResourceInfoScope;
+            if (nestedResourceScope == null)
+            {
+                nestedResourceScope = this.ParentNestedResourceInfoScope as JsonLightNestedResourceInfoScope;
```
KanishManuja-MS: >ParentNestedResourceInfoScope [](start = 43, length = 29)

This is not clear to me; if a parentNestedResourceInfo is passed, why are we using the private member? Should the names be different then?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
```diff
@@ -990,7 +990,10 @@ protected override void EndNestedResourceInfoWithContent(ODataNestedResourceInfo
             // resource set afterwards, we have to now close the array of links.
             if (navigationLinkScope.EntityReferenceLinkWritten && !navigationLinkScope.ResourceSetWritten && nestedResourceInfo.IsCollection.Value)
             {
-                this.jsonWriter.EndArrayScope();
+                if (!this.jsonLightOutputContext.WritingResponse)
```
mikepizzo: why make this a separate if, and not part of the above condition (perhaps with line breaks inserted for readability)?

mikepizzo: Why does this depend on whether or not we're writing a response. There are valid requests that take an array of entity references, enclosed in square brackets -- are the square brackets for that case written somewhere else?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
```diff
@@ -1018,32 +1021,51 @@ protected override void WriteEntityReferenceInNavigationLinkContent(ODataNestedR
...
+            if (!nestedResourceScope.EntityReferenceLinkWritten)
             {
-                if (!this.jsonLightOutputContext.WritingResponse &&
-                    (this.Version == null || this.Version < ODataVersion.V401))
-                {
-                    // Write the property annotation for the entity reference link(s)
-                    this.odataAnnotationWriter.WritePropertyAnnotationName(parentNestedResourceInfo.Name, ODataAnnotationNames.ODataBind);
-                }
-                else
+                // In request
```
mikepizzo: >In request [](start = 19, length = 10)

I understand writing a different annotation in a request for 4.0 versus 4.01 and greater, but not sure why the logic for writing the open/close square bracket varies from request to response.

mikepizzo:
Might help to meet quickly to discuss. I'm unclear why the logic for writing square brackets in collections that contain entity references differs from request to response. If this is because we have some legacy code pattern that writes a list of entity references, then that pattern should take care of writing the []; in the normal pattern of writing the nested info and then writing either a resource or resourceset, writing the nested resource set should always take care of the open/close brackets.

mikepizzo:
The only solution I can come up with for writing both the old (4.0) and new (4.01) syntax and still support the old API is to defer writing the property name until the first element, even for the case of a nested collection following the new pattern (i.e., the property name would not be written on WriteStart(ResourceSet) but when writing the first resource, or on WriteEnd() for the resourceSet if no resource were written.)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
xuzhg: @KanishManuja-MS What do you mean? 

ParentNestedResourceInfoScope is used if there's a ResourceSet.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
xuzhg: in Request:
```
WriteStart( nestedResourceInfo ) // collection
      WriteEntityReferenceLink( EntityReferenceLink)
      WriteEntityReferenceLink( EntityReferenceLink)
      WriteEntityReferenceLink( EntityReferenceLink)
WriteEnd() // end of nested resource info
```
the open square bracket is written in first call to : "WriteEntityReferenceLink(..)

Meanwhile, in response, 

```
WriteStart( nestedResourceInfo ) // collection
   WriteStart( new ODataResourceSet() ) // Add a resource set writing logic
      WriteEntityReferenceLink( EntityReferenceLink)
      WriteEntityReferenceLink( EntityReferenceLink)
      WriteEntityReferenceLink( EntityReferenceLink)
  WriteEnd() // end of resource set
WriteEnd() // end of nested resource info
```

the open square bracket is written in the "WriteStart(new ODataResourceSet()), so, when we call the first "WriteEntityReferenceLink(...)", it needn't to write the open square bracket. 

Same for the close square bracket.



---
In reply to: [285803927](https://github.com/OData/odata.net/pull/1464#discussion_r285803927) [](ancestors = 285803927)

xuzhg:
defer the property name and open square bracket until the first element maybe a breaking change for current behavior and it's a lit complicated, because it's not only related to the property name and open square bracket, but also other instance annotations. 

---
In reply to: [497369236](https://github.com/OData/odata.net/pull/1464#issuecomment-497369236) [](ancestors = 497369236)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
xuzhg: Yes. Writing the close square bracket in response is in `WriteEnd()` for resource set. We don't need to write `[` again here. 

Besides, `WriteEnd()` for resource set will process other part, for example, to write the nextpagelink annotation. 

---
In reply to: [285801842](https://github.com/OData/odata.net/pull/1464#discussion_r285801842) [](ancestors = 285801842)

xuzhg:
@mikepizzo https://github.com/OData/odata.net/issues/1472 is created for 7.6.1 update. If others are ok, can i merge this PR?

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -1800,15 +1800,22 @@ private void WriteEntityReferenceLinkImplementation(ODataEntityReferenceLink ent
...
             {
                 this.InterceptException(() =>
                 {
                     WriterValidationUtils.ValidateEntityReferenceLink(entityReferenceLink);
-                    this.WriteEntityReferenceInNavigationLinkContent((ODataNestedResourceInfo)this.CurrentScope.Item, entityReferenceLink);
+
+                    ODataNestedResourceInfo nestedInfo = this.CurrentScope.Item as ODataNestedResourceInfo;
+                    if (nestedInfo == null)
+                    {
+                        nestedInfo = this.ParentNestedResourceInfoScope.Item as ODataNestedResourceInfo;
```
mikepizzo: >ParentNestedResourceInfoScope [](start = 42, length = 29)

Can we safely assert that ParentNestedResourceInfoScope will always be non-null?  What if the user tries to call WriteEntityReferenceLinkImplementation at the top level?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
mikepizzo: Yech. As discussed, we need to fix this to support the same pattern in request and response (the new one) and deprecate the old pattern.

---
In reply to: [289081395](https://github.com/OData/odata.net/pull/1464#discussion_r289081395) [](ancestors = 289081395,285803927)

mikepizzo:
I'm okay merging once the remaining active comments are addressed, and fixing the case for writing requests in the new pattern in 7.6.1.

---
In reply to: [497859608](https://github.com/OData/odata.net/pull/1464#issuecomment-497859608) [](ancestors = 497859608)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
xuzhg: Resolved.

---
In reply to: [289888907](https://github.com/OData/odata.net/pull/1464#discussion_r289888907) [](ancestors = 289888907,285800014)

src/Microsoft.OData.Core/ODataWriterCore.cs
xuzhg: I think we can.

---
In reply to: [289887974](https://github.com/OData/odata.net/pull/1464#discussion_r289887974) [](ancestors = 289887974)

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -1970,7 +1977,11 @@ private void CheckForNestedResourceInfoWithContent(ODataPayloadKind contentPaylo
             {
                 if (contentPayloadKind == ODataPayloadKind.EntityReferenceLink)
                 {
-                    this.ThrowODataException(Strings.ODataWriterCore_EntityReferenceLinkWithoutNavigationLink, null);
+                    Scope parenScope = this.ParentNestedResourceInfoScope;
```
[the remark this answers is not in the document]
xuzhg: Yes. We can. 

---
In reply to: [289888445](https://github.com/OData/odata.net/pull/1464#discussion_r289888445) [](ancestors = 289888445)

xuzhg:
Merged. Thanks.

Issue #1467, opened 2019-05-22
Support enum parameters for Uri function


kosinsky:
We could add custom Uri (query) functions using static CustomUriFunctions.AddCustomUriFunction method. However, functions with enum as parameters aren't fully suppored when multiple copies of the same IEdmModel (with small changes) are generated.

Following expressions should be suppored:
* `$filter=enumFunc('Rectangle')`
* `$filter=enumFunc(Fully.Qualified.Namespace.EnumType'Rectangle')`

xuzhg:
Merged. Thanks.

PR #1468, opened 2019-05-22
Support for enum in Uri functions

kosinsky:
### Issues

*This pull request fixes issue #1467.*

### Description

ODataLib requires to register signatures for custom uri functions in static class
If we generate multiple models that use the same enum we will have different object refs, as a result we have to compare fullnames

- [x] *Test cases added*

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/CustomUriFunctionsTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/CustomUriFunctionsTests.cs
@@ -670,6 +670,34 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
             }
         }
 
+        [Fact]
+        public void ParseWithCustomFunction_EnumParameter()
+        {
+            try
+            {
+                var enumType = new EdmEnumType("Fully.Qualified.Namespace", "NonFlagShape", EdmPrimitiveTypeKind.SByte, false);
+                enumType.AddMember("Rectangle", new EdmEnumMemberValue(1));
+                enumType.AddMember("Triangle", new EdmEnumMemberValue(2));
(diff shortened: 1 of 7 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/CustomUriFunctionsTests.cs
```diff
@@ -670,6 +670,34 @@ public void ParseWithCustomUriFunction_AddAsOverloadToBuiltIn()
...
+
+                var fullUri = new Uri("http://www.odata.com/OData/People" + "?$filter=enumFunc('Rectangle')");
+                ODataUriParser parser = new ODataUriParser(HardCodedTestModel.TestModel, new Uri("http://www.odata.com/OData/"), fullUri);
+
+                var enumFuncWithArgs = parser.ParseFilter().Expression.ShouldBeSingleValueFunctionCallQueryNode("enumFunc").And.Parameters.ToList();
+                enumFuncWithArgs[0].ShouldBeEnumNode(new ODataEnumValue("Rectangle", enumType.FullName));
+            }
+            finally
+            {
+                CustomUriFunctions.RemoveCustomUriFunction("enumFunc").Should().BeTrue();
```
xuzhg: >CustomUriFunctions [](start = 16, length = 18)

Why should we use "try{} finally{}"?

kosinsky: CustomUriFunctions is static class. We need to add  and then remove function registration. try/finally allows to ensure that registration will be removed even if test fails and other test will not be affected.

It's the pattern in CustomUriFunctionsTests

xuzhg:
Merge at https://github.com/OData/odata.net/commit/9c75b9fa55a6eacfbafb341831e40869d065a909

PR #1469, opened 2019-05-23
Add a validation rule about target of the annotation should be allowed in the AppliesTo of the term

xuzhg:
### Issues

* So far, vocabulary annotation is used randomly in any element. 
* However, it should be allowed in the AppliesTo of the term.

### Description

* This PR is to add a validation rule to verify the term whether it is used correctly on the corresponding target.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Validation/ValidationRuleSet.cs
+++ b/src/Microsoft.OData.Edm/Validation/ValidationRuleSet.cs
@@ -102,7 +102,6 @@ namespace Microsoft.OData.Edm.Validation
                 ValidationRules.PropertyValueBindingValueIsCorrectType,
                 ValidationRules.EnumMustHaveIntegerUnderlyingType,
                 ValidationRules.AnnotationInaccessibleTerm,
-                ValidationRules.VocabularyAnnotationTargetAllowedApplyToElement,
                 ValidationRules.ElementDirectValueAnnotationFullNameMustBeUnique,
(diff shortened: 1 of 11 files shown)


src/Microsoft.OData.Edm/Validation/ValidationRuleSet.cs
```diff
@@ -102,6 +102,7 @@ public sealed class ValidationRuleSet : IEnumerable<ValidationRule>
                 ValidationRules.PropertyValueBindingValueIsCorrectType,
                 ValidationRules.EnumMustHaveIntegerUnderlyingType,
                 ValidationRules.AnnotationInaccessibleTerm,
+                ValidationRules.VocabularyAnnotationTargetAllowedApplyToElement,
                 ValidationRules.ElementDirectValueAnnotationFullNameMustBeUnique,
```
mikepizzo: I thought we were not going to add this rule to the default ruleset, since appliesto is only a hint and clients need to be prepared to handle a term being applied to other elements?

xuzhg: #Resolved.

xuzhg:
@mikepizzo I resolved the comments.

xuzhg:
Merged https://github.com/OData/odata.net/commit/c0d04122a42897e86c10a88a504517deb929d770

PR #1473, opened 2019-05-31
Remove and change ToList<T>, Any<T>, Where<T> in Uri parser

xuzhg:
### Issues

### Description

* Remove the ToList < T > Linq call in FunctionOverloadResolver.cs
* Change Any < T > call
* Change Where < T > call 

image

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
+++ b/src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
@@ -202,12 +202,18 @@ namespace Microsoft.OData.Metadata
             int currentInheritanceLevelsFromBase = int.MaxValue;
             foreach (IEdmOperation operation in operations)
             {
-                if (!operation.IsBound || !operation.Parameters.Any())
+                if (!operation.IsBound)
                 {
                     continue;
                 }
 
-                IEdmType operationBindingType = operation.Parameters.First().Type.Definition;
+                IEdmOperationParameter parameter = operation.Parameters.FirstOrDefault();
+                if (parameter == null)
+                {
+                    continue;
+                }
+
+                IEdmType operationBindingType = parameter.Type.Definition;
                 IEdmStructuredType operationBindingStructuralType = operationBindingType as IEdmStructuredType;
 
                 if (operationBindingType.TypeKind == EdmTypeKind.Collection)
@@ -454,11 +460,11 @@ namespace Microsoft.OData.Metadata
             {
                 if (item.IsAction())
                 {
-                    actionItems.Add((IEdmAction)item);
+                    actionItems.Add(item);
                 }
                 else
                 {
-                    functions.Add((IEdmFunction)item);
+                    functions.Add(item);
                 }
             }
 
@@ -471,20 +477,20 @@ namespace Microsoft.OData.Metadata
         /// <param name="source">The source.</param>
         /// <param name="actionImportItems">The action import items.</param>
         /// <returns>Only the function imports from the operation Import sequence.</returns>
-        internal static IEnumerable<IEdmOperationImport> RemoveActionImports(this IEnumerable<IEdmOperationImport> source, out IList<IEdmActionImport> actionImportItems)
+        internal static IEnumerable<IEdmOperationImport> RemoveActionImports(this IEnumerable<IEdmOperationImport> source, out IList<IEdmOperationImport> actionImportItems)
         {
             List<IEdmOperationImport> functions = new List<IEdmOperationImport>();
 
-            actionImportItems = new List<IEdmActionImport>();
+            actionImportItems = new List<IEdmOperationImport>();
             foreach (var item in source)
             {
                 if (item.IsActionImport())
                 {
-                    actionImportItems.Add((IEdmActionImport)item);
+                    actionImportItems.Add(item);
                 }
                 else
                 {
-                    functions.Add((IEdmFunctionImport)item);
+                    functions.Add(item);
                 }
             }
 
--- a/src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
+++ b/src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
@@ -136,27 +136,27 @@ namespace Microsoft.OData.Metadata
 #if !ODATA_SERVICE && !ODATA_CLIENT
 
         /// <summary>
-        /// Filters functions by the parameter names.
+        /// Filters operations by the parameter names.
         /// </summary>
-        /// <param name="functionImports">The operation imports.</param>
+        /// <param name="operationImports">The operation imports.</param>
         /// <param name="parameterNames">The parameter names.</param>
         /// <param name="caseInsensitive">Whether to support case insensitive.</param>
         /// <returns>Return the operation imports that match the parameter names.</returns>
-        internal static IEnumerable<IEdmFunctionImport> FilterFunctionsByParameterNames(this IEnumerable<IEdmFunctionImport> functionImports, IEnumerable<string> parameterNames, bool caseInsensitive)
+        internal static IEnumerable<IEdmOperationImport> FilterOperationsByParameterNames(this IEnumerable<IEdmOperationImport> operationImports, IEnumerable<string> parameterNames, bool caseInsensitive)
         {
-            Debug.Assert(functionImports != null, "functionImports");
+            Debug.Assert(operationImports != null, "operationImports");
             Debug.Assert(parameterNames != null, "parameterNames");
 
             IList<string> parameterNameList = parameterNames.ToList();
 
-            foreach (IEdmFunctionImport functionImport in functionImports)
+            foreach (IEdmOperationImport operationImport in operationImports)
             {
-                if (!ParametersSatisfyFunction(functionImport.Operation, parameterNameList, caseInsensitive))
+                if (!ParametersSatisfyFunction(operationImport.Operation, parameterNameList, caseInsensitive))
                 {
                     continue;
                 }
 
-                yield return functionImport;
+                yield return operationImport;
             }
         }
 
@@ -305,8 +305,7 @@ namespace Microsoft.OData.Metadata
         /// Ensures that operations are bound and have a binding parameter, other wise throws an exception.
         /// </summary>
         /// <param name="operations">The operations.</param>
-        /// <returns>Bound Operations with binding parameters.</returns>
-        internal static IEnumerable<IEdmOperation> EnsureOperationsBoundWithBindingParameter(this IEnumerable<IEdmOperation> operations)
+        internal static void EnsureOperationsBoundWithBindingParameter(this IEnumerable<IEdmOperation> operations)
         {
             foreach (IEdmOperation operation in operations)
             {
@@ -315,12 +314,10 @@ namespace Microsoft.OData.Metadata
                     throw new ODataException(ErrorStrings.EdmLibraryExtensions_UnBoundOperationsFoundFromIEdmModelFindMethodIsInvalid(operation.Name));
                 }
 
-                if (!operation.Parameters.Any())
+                if (operation.Parameters.FirstOrDefault() == null)
                 {
                     throw new ODataException(ErrorStrings.EdmLibraryExtensions_NoParameterBoundOperationsFoundFromIEdmModelFindMethodIsInvalid(operation.Name));
                 }
-
-                yield return operation;
             }
         }
 
@@ -448,11 +445,11 @@ namespace Microsoft.OData.Metadata
         /// <param name="source">The source.</param>
         /// <param name="actionItems">The action items.</param>
         /// <returns>Only the functions from the operation sequence.</returns>
-        internal static IEnumerable<IEdmFunction> RemoveActions(this IEnumerable<IEdmOperation> source, out IList<IEdmAction> actionItems)
+        internal static IEnumerable<IEdmOperation> RemoveActions(this IEnumerable<IEdmOperation> source, out IList<IEdmOperation> actionItems)
         {
-            List<IEdmFunction> functions = new List<IEdmFunction>();
+            List<IEdmOperation> functions = new List<IEdmOperation>();
 
-            actionItems = new List<IEdmAction>();
+            actionItems = new List<IEdmOperation>();
             foreach (var item in source)
             {
                 if (item.IsAction())
@@ -474,9 +471,9 @@ namespace Microsoft.OData.Metadata
         /// <param name="source">The source.</param>
         /// <param name="actionImportItems">The action import items.</param>
         /// <returns>Only the function imports from the operation Import sequence.</returns>
-        internal static IEnumerable<IEdmFunctionImport> RemoveActionImports(this IEnumerable<IEdmOperationImport> source, out IList<IEdmActionImport> actionImportItems)
+        internal static IEnumerable<IEdmOperationImport> RemoveActionImports(this IEnumerable<IEdmOperationImport> source, out IList<IEdmActionImport> actionImportItems)
         {
-            List<IEdmFunctionImport> functions = new List<IEdmFunctionImport>();
+            List<IEdmOperationImport> functions = new List<IEdmOperationImport>();
 
             actionImportItems = new List<IEdmActionImport>();
             foreach (var item in source)
--- a/src/Microsoft.OData.Core/Metadata/MetadataUtils.cs
+++ b/src/Microsoft.OData.Core/Metadata/MetadataUtils.cs
@@ -172,8 +172,10 @@ namespace Microsoft.OData.Metadata
                 throw new ODataException(Strings.MetadataUtils_CalculateBindableOperationsForType(bindingType.FullTypeName()), exc);
             }
 
+            operations.EnsureOperationsBoundWithBindingParameter();
+
             List<IEdmOperation> operationsFound = new List<IEdmOperation>();
-            foreach (IEdmOperation operation in operations.EnsureOperationsBoundWithBindingParameter())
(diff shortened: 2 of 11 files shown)


src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -448,11 +445,11 @@ internal static string FullNameWithParameters(this IEdmOperationImport operation
         /// <param name="source">The source.</param>
         /// <param name="actionItems">The action items.</param>
         /// <returns>Only the functions from the operation sequence.</returns>
-        internal static IEnumerable<IEdmFunction> RemoveActions(this IEnumerable<IEdmOperation> source, out IList<IEdmAction> actionItems)
+        internal static IEnumerable<IEdmOperation> RemoveActions(this IEnumerable<IEdmOperation> source, out IList<IEdmOperation> actionItems)
```
mikepizzo: >IEnumerable<IEdmOperation> [](start = 24, length = 26)

Why is the return type changed to IEdmOperation? the method still only returns the functions in this list, and actions as an output param.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -474,9 +471,9 @@ internal static IEnumerable<IEdmFunction> RemoveActions(this IEnumerable<IEdmOpe
         /// <param name="source">The source.</param>
         /// <param name="actionImportItems">The action import items.</param>
         /// <returns>Only the function imports from the operation Import sequence.</returns>
-        internal static IEnumerable<IEdmFunctionImport> RemoveActionImports(this IEnumerable<IEdmOperationImport> source, out IList<IEdmActionImport> actionImportItems)
+        internal static IEnumerable<IEdmOperationImport> RemoveActionImports(this IEnumerable<IEdmOperationImport> source, out IList<IEdmActionImport> actionImportItems)
```
mikepizzo: >IEdmOperationImport [](start = 36, length = 19)

Why make the return type IEdmOperationImport, rather than keep it as IEdmFunctionImport, since the returned values are always function imports?

src/Microsoft.OData.Core/Metadata/MetadataUtils.cs
```diff
@@ -172,8 +172,10 @@ internal static IEdmOperation[] CalculateBindableOperationsForType(IEdmType bind
                 throw new ODataException(Strings.MetadataUtils_CalculateBindableOperationsForType(bindingType.FullTypeName()), exc);
             }
 
+            operations.EnsureOperationsBoundWithBindingParameter();
```
mikepizzo: Rather than looping through the operations here, and again in line 178, what if we changed EnsureOperationBoundWithBindingParameter() to take a single operation, and called it within the loop on line 178?

src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
```diff
@@ -177,7 +177,7 @@ internal static bool TryBindAsOperation(PathSegmentToken pathToken, IEdmModel mo
                 }
             }
 
-            possibleFunctions = possibleFunctions.EnsureOperationsBoundWithBindingParameter().ToList();
+            possibleFunctions.EnsureOperationsBoundWithBindingParameter();
```
mikepizzo: Do we really need to validate all of the  possibleFunctions here, or should we just wait and validate at line 199 after we have reduced the number of possible functions based on overloads.

mikepizzo:
>             return operationsFound.ToArray();

Why do we return as an array, rather than an enumerable?

---
Refers to: src/Microsoft.OData.Core/Metadata/MetadataUtils.cs:188 in 2d0c8c7. [](commit_id = 2d0c8c7871047a685f2f141612d02f515436c23d, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -86,24 +87,24 @@ internal static bool ResolveOperationImportFromList(string identifier, IList<str
...
             // If parameter count is zero and there is one function import whoese parameter count is zero, return this function import.
-            if (candidateMatchingOperationImports.Count > 1 && parameterNames.Count == 0)
+            if (candidateMatchingOperationImports.Count() > 1 && parameterNames.Count == 0)
             {
-                candidateMatchingOperationImports = candidateMatchingOperationImports.Where(operationImport => operationImport.Operation.Parameters.Count() == 0).ToList();
+                candidateMatchingOperationImports = candidateMatchingOperationImports.Where(operationImport => operationImport.Operation.Parameters.Count() == 0);
             }
 
-            if (candidateMatchingOperationImports.Count == 0)
+            if (candidateMatchingOperationImports.Count() == 0)
```
mikepizzo: >.Count() [](start = 49, length = 8)

is it cheaper to use FirstOrDefault() == null?

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
xuzhg: Because the return will be used to call "FilterOperationsByParameterNames", which accepts the "IEdmOperation". We don't want to cast the operation to function, then cast the function to operation again and again. That's useless and waste of time. 



---
In reply to: [289995838](https://github.com/OData/odata.net/pull/1473#discussion_r289995838) [](ancestors = 289995838)

mikepizzo:
>                 if (parameterNames.Count() != 0)

use parameterNamesCount

---
Refers to: src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs:215 in 2d0c8c7. [](commit_id = 2d0c8c7871047a685f2f141612d02f515436c23d, deletion_comment = False)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
xuzhg: Because the return will be used to call "FilterOperationsByParameterNames", which accepts the "IEdmOperation". We don't want to cast the operation to function, then cast the function to operation again and again. That's useless and waste of time. 

Check the line 45 in FunctionOverloadResolver.cs.

---
In reply to: [289996208](https://github.com/OData/odata.net/pull/1473#discussion_r289996208) [](ancestors = 289996208)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -163,42 +164,43 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
                 throw;
             }
 
-            IList<IEdmAction> foundActionsWhenLookingForFunctions = new List<IEdmAction>();
+            IList<IEdmOperation> foundActionsWhenLookingForFunctions = new List<IEdmOperation>();
             int parameterNamesCount = parameterNames.Count();
```
mikepizzo: >int parameterNamesCount = parameterNames.Count() [](start = 12, length = 48)

would it be better to make this a bool hasParameters = (parameterNames.FirstOrDefault == null) ?

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
xuzhg: From the Count() function:

```
public static int Count<TSource>(this IEnumerable<TSource> source) {
            if (source == null) throw Error.ArgumentNull("source");
            ICollection<TSource> collectionoft = source as ICollection<TSource>;
            if (collectionoft != null) return collectionoft.Count;
            ICollection collection = source as ICollection;
            if (collection != null) return collection.Count;
            int count = 0;
            using (IEnumerator<TSource> e = source.GetEnumerator()) {
                checked {
                    while (e.MoveNext()) count++;
                }
            }
            return count;
        }
```

So, if the source is ICollection< T >, it's cheaper than "FirstOrDefault< T >". In our codes, internally the IEnumerable< T > is the IList< T >, which implements the ICollection< T >, So, here the Count() is cheaper than "FirstOrDefault< T >".



---
In reply to: [290001862](https://github.com/OData/odata.net/pull/1473#discussion_r290001862) [](ancestors = 290001862)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
xuzhg: parameterNames is IList<string>, so, Count< T > is cheaper than FirstOrDefault == null ?

---
In reply to: [290003826](https://github.com/OData/odata.net/pull/1473#discussion_r290003826) [](ancestors = 290003826)

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -2815,16 +2819,16 @@ internal static bool TryGetRelativeEntitySetPath(IEdmElement element, Collection
             }
 
             // If there is no parameter then this will fail in BoundOperationMustHaveParameters rule so skip validating this here.
-            if (!parameters.Any())
+            parameter = parameters.FirstOrDefault();
+            if (parameter == null)
             {
                 return false;
             }
+            Debug.Assert(parameter != null, "Should never be null");
```
mikepizzo: not sure this assert is necessary, since previous line would have filtered out a null parameter value.

xuzhg: I think so. Removed.

---
In reply to: [290008471](https://github.com/OData/odata.net/pull/1473#discussion_r290008471) [](ancestors = 290008471)

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -3130,6 +3134,23 @@ internal static bool TryGetStaticEntitySet(this IEdmPathExpression pathExpressio
...
+                return list.Count() > 0;
+            }
+
+            T[] array = enumerable as T[];
+            if (array != null)
+            {
+                return array.Count() > 0;
+            }
+
+            return enumerable.Any();
```
mikepizzo: >.Any() [](start = 29, length = 6)

is FirstOrDefault == null better than Any()?

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -173,7 +173,25 @@ public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(IEdmType b
         /// </returns>
         public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(string qualifiedName, IEdmType bindingType)
         {
-            return this.FindDeclaredOperations(qualifiedName).Where(o => o.IsBound && o.Parameters.Any() && o.HasEquivalentBindingType(bindingType));
+            var enumerable = this.FindDeclaredOperations(qualifiedName);
```
mikepizzo: >var [](start = 12, length = 3)

Prefer not to use var, use IEnumerable<IEdmOperation> instead.

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -173,7 +173,25 @@ public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(IEdmType b
         /// </returns>
         public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(string qualifiedName, IEdmType bindingType)
         {
-            return this.FindDeclaredOperations(qualifiedName).Where(o => o.IsBound && o.Parameters.Any() && o.HasEquivalentBindingType(bindingType));
+            var enumerable = this.FindDeclaredOperations(qualifiedName);
+            var operations = enumerable as IList<IEdmOperation>;
+            if (operations != null)
+            {
+                IList<IEdmOperation> matchedOperation = new List<IEdmOperation>();
+                for (int i = 0; i < operations.Count(); i++)
```
mikepizzo: >Count() [](start = 47, length = 7)

Count or Count()?  If this is a list, isn't Count property better?

xuzhg:
>             return operationsFound.ToArray();

That's old codes. I changed. it.

---
In reply to: [498393990](https://github.com/OData/odata.net/pull/1473#issuecomment-498393990) [](ancestors = 498393990)

---
Refers to: src/Microsoft.OData.Core/Metadata/MetadataUtils.cs:188 in 2d0c8c7. [](commit_id = 2d0c8c7871047a685f2f141612d02f515436c23d, deletion_comment = False)

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
xuzhg: Here we can't say the `T` is reference type, or it maybe `Value type`. So, for some scenarios (not a list, not an array), i think it's better to delegate to use "Any()". What do you think? 

---
In reply to: [290009255](https://github.com/OData/odata.net/pull/1473#discussion_r290009255) [](ancestors = 290009255)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
mikepizzo: still probably better to make it bool hasParameters = (parameterNames.Count() > 0), since the only thing we care about is whether there are any parameters (we don't care how many), so that saves a comparison later.

---
In reply to: [290007315](https://github.com/OData/odata.net/pull/1473#discussion_r290007315) [](ancestors = 290007315,290003826)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
mikepizzo: We don't have to cast because we are passing a more derived type. Passing a derived type to a method that is typed to receive the base type should be free, as the type checking is done at compile time, not at execution.

---
In reply to: [290003738](https://github.com/OData/odata.net/pull/1473#discussion_r290003738) [](ancestors = 290003738,289996208)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
mikepizzo: Why not use your new HasAny() extension method?

---
In reply to: [290006737](https://github.com/OData/odata.net/pull/1473#discussion_r290006737) [](ancestors = 290006737,290001862)

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
mikepizzo: >.Any() [](start = 29, length = 6)

should this use FirstOrDefault() == null?

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
xuzhg: That's internal extension method defined in OData.Edm library. 

---
In reply to: [290047580](https://github.com/OData/odata.net/pull/1473#discussion_r290047580) [](ancestors = 290047580,290006737,290001862)

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
mikepizzo: I'm okay using Any() here.

---
In reply to: [290038532](https://github.com/OData/odata.net/pull/1473#discussion_r290038532) [](ancestors = 290038532,290009255)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
mikepizzo: Should we have a similar method here? For the case where the enumerator is not a collection, this method will actually count all of the members when all we really need to know is if there are any members. I realize that our implementation uses collection, but your code would be more efficient for other enumerators.

---
In reply to: [290048582](https://github.com/OData/odata.net/pull/1473#discussion_r290048582) [](ancestors = 290048582,290047580,290006737,290001862)

xuzhg: Got it.

---
In reply to: [290049890](https://github.com/OData/odata.net/pull/1473#discussion_r290049890) [](ancestors = 290049890,290048582,290047580,290006737,290001862)

xuzhg:
As discussed with Mike. So far, the concern is good. So, Merged. Thanks.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
mikepizzo: As we discussed, since:
1) .NET 3.5 requires an explicit cast to cast from IEnumerable<type> to IEnumerable<derivedType>
2) The methods that currently take the returned enumerables treat them as operations, and never have to cast to function/action
3) The current code in EdmLibraryExtensions *does* cast them to the correct derived type (even though downstream we only care about the base type), and
4) These methods are all internal, so we could change them later to return the derived type *if* we had a need to treat them as the derived type
Then I'm okay, for now, changing all of the signatures to deal with operations, rather than the appropriate derived type. 

---
In reply to: [290046571](https://github.com/OData/odata.net/pull/1473#discussion_r290046571) [](ancestors = 290046571,290003738,289996208)

PR #1488, opened 2019-06-13
Enable parsing nested query options in $select to syntactic AST

xuzhg:
### Description

OData v4.01 spec says:
Query options can be applied to a selected property by appending a semicolon-separated list of query options, enclosed in parentheses, to the property. Allowed system query options are $select and $expand, plus $filter, $search, $count, $orderby, $skip, and $top for collection-valued properties.

This PR is intended to enable parsing the nested query options in $select to syntactic AST.

•	$filter
•	$orderby
•	$top
•	$skip
•	$count
•	$search
•	$select
•	$expand

This PR is first part of the supporting of $select nested query options; The second is to binder the syntactic AST to semantic AST.

Updated: 6/18/2019:

$compute *should* be supported as a valid select option. It was there in the ABNF and URL Conventions, but was missing from the Protocol spec

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/UriParser/SyntacticAst/ExpandTermToken.cs
```diff
@@ -176,121 +116,41 @@ public ExpandTermToken(PathSegmentToken pathToNavigationProp, QueryToken filterO
...
-            this.countQueryOption = countQueryOption;
-            this.levelsOption = levelsOption;
-            this.searchOption = searchOption;
-            this.selectOption = selectOption;
-            this.computeOption = computeOption;
-            this.expandOption = expandOption;
-            this.applyOptions = applyOptions;
+            LevelsOption = levelsOption;
+            ComputeOption = computeOption;
+            ApplyOptions = applyOptions;
```
xuzhg: @mike, does select support $compute, $apply from OData spec future?

src/Microsoft.OData.Core/UriParser/SyntacticAst/ExpandTermToken.cs
```diff
@@ -176,121 +116,41 @@ public ExpandTermToken(PathSegmentToken pathToNavigationProp, QueryToken filterO
...
-        /// </summary>
-        public long? SkipOption
-        {
-            get { return this.skipOption; }
-        }
-
-        /// <summary>
-        /// the query count option for this expand term.
-        /// </summary>
-        public bool? CountQueryOption
```
xuzhg: >CountQueryOption [](start = 21, length = 16)

Avoiding breaking change, should keep it as CountQueryOption

src/Microsoft.OData.Core/UriParser/SyntacticAst/ExpandTermToken.cs
xuzhg: Updated: 6/18/2019: 
$compute should be supported as a valid select option. It was there in the ABNF and URL Conventions, but was missing from the Protocol spec

src/Microsoft.OData.Core/UriParser/Parsers/SelectExpandOptionParser.cs
```diff
@@ -102,34 +102,26 @@ internal sealed class ExpandOptionParser
...
+        /// to build up an SelectTermToken which fully represents the tree that makes up this select term.
         /// </summary>
-        /// <param name="pathToken">The PathSegmentToken representing the parsed expand path whose options we are now parsing.</param>
-        /// <param name="optionsText">A string of the text between the parenthesis after an expand option.</param>
-        /// <returns>The list of expand term tokens based on the path token, and all available expand options.</returns>
-        internal List<ExpandTermToken> BuildExpandTermToken(PathSegmentToken pathToken, string optionsText)
+        /// <param name="pathToken">The PathSegmentToken representing the parsed select path whose options we are now parsing.</param>
+        /// <param name="optionsText">A string of the text between the parenthesis after a select option.</param>
+        /// <returns>The select term token based on the path token, and all available select options.</returns>
+        internal SelectTermToken BuildSelectTermToken(PathSegmentToken pathToken, string optionsText)
```
mikepizzo: >SelectTermToken [](start = 17, length = 15)

Why does BuildSelectTermToken return a SelectTermToken, while BuildExpandTermToken returns a List<ExpandTermToken>?

mikepizzo: >BuildSelectTermToken [](start = 33, length = 20)

could more of the implementation of BuildSelectTermToken be shared with BuildExpandTermToken?

src/Microsoft.OData.Core/UriParser/Parsers/SelectExpandOptionParser.cs
```diff
@@ -427,7 +408,7 @@ private List<ExpandTermToken> BuildStarExpandTermToken(PathSegmentToken pathToke
             }
 
             // As 2016/1/8, the navigation property is only supported in entity type, and will support in ComplexType in future.
-            var entityType = parentEntityType as IEdmEntityType;
+            var entityType = this.parentStructuredType as IEdmEntityType;
```
mikepizzo: this comment seems suspect.

mikepizzo: shouldn't we support ComplexType here, and not just IEdmEntityType?

src/Microsoft.OData.Core/UriParser/Parsers/SelectExpandOptionParser.cs
```diff
@@ -456,11 +437,188 @@ private List<ExpandTermToken> BuildStarExpandTermToken(PathSegmentToken pathToke
...
+        {
+            // advance to the equal sign
+            this.lexer.NextToken();
+            string countText = this.ReadQueryOption();
+            switch (countText)
+            {
+                case ExpressionConstants.KeywordTrue:
+                    return true;
+
+                case ExpressionConstants.KeywordFalse:
```
mikepizzo: should we support case insensitivity here?

mikepizzo:
A few minor questions/comments, but overall looks really good!

src/Microsoft.OData.Core/UriParser/Parsers/SelectExpandOptionParser.cs
xuzhg: Because they are related to "*".

For `$expand=*`, it will go through all navigation properties and for each navigation property, it will create a ExpandTermToken.  So, `$expand=*` will be parsed as `$expand=NP1,NP2,NP3,...`
That's why it's a list.

For `$select=*`, it will output a `WildcardSelectTermToken`, never expand to all properties.  That's why it's a single token.

#Resolved.

src/Microsoft.OData.Core/UriParser/Parsers/SelectExpandOptionParser.cs
xuzhg: Case insensitivity supports for the key word, for the literal of "true" and "false", it's not support case insensitivity. Let me double check it.

src/Microsoft.OData.Core/UriParser/Parsers/SelectExpandOptionParser.cs
xuzhg: Maybe. However, i think current codes look more clearly. I'd like to keep them unchanged now.

---
In reply to: [296958543](https://github.com/OData/odata.net/pull/1488#discussion_r296958543) [](ancestors = 296958543)

src/Microsoft.OData.Core/UriParser/Parsers/SelectExpandOptionParser.cs
xuzhg: It's a different feature gap.  Let's file an issue to fix it. [https://github.com/OData/odata.net/issues/1493](https://github.com/OData/odata.net/issues/1493)   

---
In reply to: [296959088](https://github.com/OData/odata.net/pull/1488#discussion_r296959088) [](ancestors = 296959088)

xuzhg: It's a different feature gap.  Let's file an issue to fix it. [https://github.com/OData/odata.net/issues/1493](https://github.com/OData/odata.net/issues/1493)   

---
In reply to: [296958890](https://github.com/OData/odata.net/pull/1488#discussion_r296958890) [](ancestors = 296958890)

xuzhg:
@mikepizzo I addressed the comments. Please let me know if you have more concerns.

src/Microsoft.OData.Core/UriParser/Parsers/SelectExpandOptionParser.cs
mikepizzo: I'm not sure the semantics of parsing $expand=* into $expand=NP1,NP2,NP3,... is correct, in that there may be dynamic navigation properties that we would exclude. But, that's a separate issue.  For $select, we need to distinguish between * and the list of all properties (the former includes dynamic properties, the later does not). So, the added behavior is the correct behavior.

---
In reply to: [297359685](https://github.com/OData/odata.net/pull/1488#discussion_r297359685) [](ancestors = 297359685)

PR #1494, opened 2019-06-25
Metadata Selector for exposure control

KanishManuja-MS:
Currently, for a full metadata request, there is no way for consumers of the library to select which metadata gets written in the response. 

With this PR, the consumers will now be able to inject a Metadata Selector that will be able to add or remove navigation property, bound operations and stream property related metadata.

### DIFF
--- a/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
@@ -389,6 +389,9 @@ namespace Microsoft.OData.Evaluation
             /// </summary>
             private readonly SelectedPropertiesNode selectedProperties;
 
+            /// <summary>
+            /// The metadata selector
+            /// </summary>
             private readonly ODataMetadataSelector metadataSelector;
 
             /// <summary>
--- a/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
@@ -483,7 +483,7 @@ namespace Microsoft.OData.Evaluation
                         if (this.metadataSelector != null)
                         {
                             this.selectedNavigationProperties =
-                                this.metadataSelector.SelectNavigationProperties(this.selectedNavigationProperties);
+                                this.metadataSelector.SelectNavigationProperties(this.actualResourceType, this.selectedNavigationProperties);
                         }
                     }
 
@@ -505,7 +505,7 @@ namespace Microsoft.OData.Evaluation
                         if (metadataSelector != null)
                         {
                             this.selectedStreamProperties = this.metadataSelector
-                                .SelectStreamProperties(this.selectedStreamProperties.Values).ToDictionary(v => v.Name);
+                                .SelectStreamProperties(this.actualResourceType, this.selectedStreamProperties.Values).ToDictionary(v => v.Name);
                         }
                     }
 
@@ -528,7 +528,7 @@ namespace Microsoft.OData.Evaluation
                         if (this.metadataSelector != null)
                         {
                             this.selectedBindableOperations =
-                                this.metadataSelector.SelectBindableOperations(this.selectedBindableOperations);
+                                this.metadataSelector.SelectBindableOperations(this.actualResourceType, this.selectedBindableOperations);
                         }
                     }
 
--- a/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
@@ -128,6 +128,7 @@ namespace Microsoft.OData.Evaluation
         /// <param name="actualResourceType">The structured type of the resource.</param>
         /// <param name="metadataContext">The metadata context to use.</param>
         /// <param name="selectedProperties">The selected properties.</param>
+        /// <param name="metadataSelector">Informs the metadata builder which properties, functions, actions, links to omit.</param>
         /// <returns>A new instance of <see cref="ODataResourceMetadataContext"/>.</returns>
         internal static ODataResourceMetadataContext Create(
             ODataResourceBase resource,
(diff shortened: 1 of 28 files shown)


src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -135,14 +135,15 @@ public IODataResourceTypeContext TypeContext
             ODataResourceSerializationInfo serializationInfo,
             IEdmStructuredType actualResourceType,
             IODataMetadataContext metadataContext,
-            SelectedPropertiesNode selectedProperties)
+            SelectedPropertiesNode selectedProperties,
+            ODataMetadataSelector metadataSelector = null)
```
mikepizzo: >ODataMetadataSelector metadataSelector = null [](start = 12, length = 45)

(minor) Since this is internal (and only called in two places outside of tests) I would generally rather explicitly pass null than specify a default.

test/FunctionalTests/Microsoft.OData.Core.Tests/Build.NetFramework/Microsoft.OData.Core.Tests.csproj
```diff
@@ -453,6 +453,8 @@
     <Compile Include="..\WriterUtilsTests.cs" />
     <Compile Include="..\ScenarioTests\Streaming\ODataJsonLightStreamReadingTests.cs" />
     <Compile Include="..\ScenarioTests\Streaming\ODataJsonLightStreamWritingTests.cs" />
+    <Compile Include="TestMetadataSelector.cs" />
+    <Compile Include="ODataMetadataSelectorTests.cs" />
```
mikepizzo: These seem to be in the wrong location.  Other tests are one level up (for common/core tests), or under "ScenarioTests" or other categories.  These should be under "JsonLight" (i.e.., ..\JsonLight\TestMetadataSelector.cs).

test/FunctionalTests/Microsoft.OData.Core.Tests/Build.NetFramework/ODataMetadataSelectorTests.cs
```diff
@@ -0,0 +1,223 @@
...
+            
+            //metadataselector only allows for two HasHat functions to be written as metadata
+            Assert.True(resource.Functions.Count() == 3);
+            Assert.True(resource.Actions.Count() == 1);
+        }
+
+        [Theory]
+        [InlineData("MyDog", 7)]
+        [InlineData("MyFriendsDogs", 7)]
+        [InlineData("Doesn't Exist", 8)]
```
mikepizzo: >Doesn't Exist [](start = 21, length = 13)

Should probably use a legal nav prop name; i.e., "DoesntExist"

test/FunctionalTests/Microsoft.OData.Core.Tests/Build.NetFramework/ODataMetadataSelectorTests.cs
```diff
@@ -0,0 +1,223 @@
...
+                NavigationSourceEntityTypeName = EntityTypeName,
+                ExpectedResourceTypeName = EntityTypeName,
+                IsMediaLinkEntry = false,
+                IsFromCollection = false,
+                NavigationSourceKind = EdmNavigationSourceKind.EntitySet
+            };
+
+            managerTypeContext = new TestFeedAndEntryTypeContext
+            {
+                NavigationSourceName = DerivedEntitySetName,
```
mikepizzo: >NavigationSourceName = DerivedEntitySetName, [](start = 16, length = 44)

Should this really be a different entity set, or should it still be EntitySetName with DerivedEntityTypeName?

test/FunctionalTests/Microsoft.OData.Core.Tests/Build.NetFramework/TestMetadataSelector.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// Determines which navigation links should be selected.
+        /// </summary>
+        /// <param name="type"></param>
+        /// <param name="navigationProperties"></param>
+        /// <returns></returns>
+        public override IEnumerable<IEdmNavigationProperty> SelectNavigationProperties(IEdmStructuredType type, IEnumerable<IEdmNavigationProperty> navigationProperties)
+        {
+            if (NavigationPropertyToAdd != null)
+            {
+                return NavigationPropertyToAdd;
```
mikepizzo: >NavigationPropertyToAdd [](start = 23, length = 23)

So this returns only the NavigationpropertyToAdd, it does not add NavigationPropertyToAdd to the passed in navigationProperties, right?  Should it be instead named "NavigationPropertiesToReturn"?

test/FunctionalTests/Microsoft.OData.Core.Tests/Build.NetFramework/TestMetadataSelector.cs
```diff
@@ -0,0 +1,47 @@
...
+            }
+            return navigationProperties.Where(s => s.Name != PropertyToOmit);
+        }
+
+        public override  IEnumerable<IEdmOperation> SelectBindableOperations(IEdmStructuredType type, IEnumerable<IEdmOperation> bindableOperations)
+        {
+            //Omit all actions
```
mikepizzo: Actually, this omits all operations not named "HasHat", right?

test/FunctionalTests/Microsoft.OData.Core.Tests/Build.NetFramework/ODataMetadataSelectorTests.cs
```diff
@@ -0,0 +1,223 @@
...
+            while (resource.MetadataBuilder.GetNextUnprocessedNavigationLink() != null)
+            {
+                count++;
+            };
+
+            Assert.True(count == expected);
+        }
+
+        [Fact]
+        public void AddNavigationLinkWithFullMetadata()
```
mikepizzo: >AddNavigationLinkWithFullMetadata [](start = 20, length = 33)

Not really add, just return specified navigation link.

mikepizzo:
A few (mostly test) clean-up issues; otherwise looks good!

test/FunctionalTests/Microsoft.OData.Core.Tests/Build.NetFramework/TestMetadataSelector.cs
KanishManuja-MS: This was a left over comment. This selects only the operations named HasHat.

---
In reply to: [310766305](https://github.com/OData/odata.net/pull/1494#discussion_r310766305) [](ancestors = 310766305)

Issue #1498, opened 2019-07-03
Navigation properties through an entity with an Enum key segment does not work.


jnanneng:
Ex:  …/OData/v4.0/1.0/Company('CRONUS International Ltd.')/Page99441(Type=NS.Page99441.Type'Order',No='1')/Customer_Link

The OData EnumBinder turns the value of the ‘Type’ key segment into a ‘ConstantNode’ with a value of an ODataEnumValue instance. 

When the @odata.Context value of the response attempts to be serialized, an exception is thrown:
The type 'Microsoft.OData.UriParser.ConstantNode' is not supported when converting to a URI literal.

This appears to occur within LiteralFormatter.FormatRawLiteral as it does not support a ‘ConstantNode’, but instead coded to ‘ODataEnumValue’.

### Assemblies affected

Microsoft.OData.* 7.5.0

### Reproduce steps

Define an entity with an enum key which is capable of navigating to another entity.  Issue a request for a navigation property of that entity.

### Expected result

It would succeed.

### Actual result

Error is encountered:  The type 'Microsoft.OData.UriParser.ConstantNode' is not supported when converting to a URI literal.

### Additional detail

The OData EnumBinder turns the value of the enum key segment into a ‘ConstantNode’ with a value of an ODataEnumValue instance. 

When the @odata.Context value of the response attempts to be serialized, an exception is thrown:
The type 'Microsoft.OData.UriParser.ConstantNode' is not supported when converting to a URI literal.

This appears to occur within LiteralFormatter.FormatRawLiteral as it does not support a ‘ConstantNode’, but instead coded to ‘ODataEnumValue’.

xuzhg:
Merged at: c64b89775306a909fdf5e71287d08ba6012ae19c

PR #1507, opened 2019-07-17
1498 :: Navigation properties through an entity with an Enum key segm…

jnanneng:
…ent does not work.

The OData EnumBinder turns the value of an enum type key segment value into a ‘ConstantNode’ with a value of an ODataEnumValue instance.

LiteralFormatter.FormatRawLiteral does not support a 'ConstantNode'; instead its coded for 'ODataEnumValue'.

The fix appears to be modifying the SegmentArgumentParser to return an ODataEnumValue.

### Issues

*This pull request fixes issue #1498.*

### Description

*Modify SegmentArgumentParser to return an ODataEnumValue*

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Parsers/SegmentArgumentParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/SegmentArgumentParser.cs
@@ -241,7 +241,7 @@ namespace Microsoft.OData.UriParser
                 QueryNode enumNode = null;
                 if (EnumBinder.TryBindIdentifier(valueText, typeReference.AsEnum(), null, out enumNode))
                 {
-                    convertedValue = enumNode;
+                    convertedValue = ((ConstantNode)enumNode).Value;
                     return true;
                 }
 
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/PathFunctionalTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/PathFunctionalTests.cs
@@ -1437,8 +1437,8 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
             path.Count.Should().Be(2);
             var keyInfo = path.Last().As<KeySegment>().Keys.Single();
             keyInfo.Key.Should().Be("PetCategorysColorPattern");
-            keyInfo.Value.As<ConstantNode>().Value.As<ODataEnumValue>().TypeName.Should().Be("Fully.Qualified.Namespace.ColorPattern");
-            keyInfo.Value.As<ConstantNode>().Value.As<ODataEnumValue>().Value.Should().Be("22");
+            keyInfo.Value.As<ODataEnumValue>().TypeName.Should().Be("Fully.Qualified.Namespace.ColorPattern");
+            keyInfo.Value.As<ODataEnumValue>().Value.Should().Be("22");
         }
         #endregion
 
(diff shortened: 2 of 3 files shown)


madansr7:
@jnanneng Thanks for your contribution. However, it seems your change is a breaking changes. We'd like it to target v8.0 version. 
Alternatively, can you fix it the Uri literal converting from the ConstantNode?

jnanneng:
It’s been a while since I’ve looked at this, if I recall correctly almost all of the downstream code paths expected an ODataEnumValue, none of which handled a ConstantNode (as that would likely require re-visiting fragments to evaluate, etc…)

Adding ConstantNode awareness to all of the other areas did not seem like a correct approach.  (IIRC this flowed through a number of other areas before it got down to the URI literal)

I was aware this was technically a breaking change, however the scenario does not function today.  Anyone that had an enum type as part of an entity key and managed to parse key segments themselves for ‘special processing’ might have thought ‘w.t.h.’ when the constant node was encountered.  If they went further and exposed a navigation property off such an entity, they were ‘dead in the water’.

Thoughts?

mikepizzo:
Looking at the code change, the code only affects what is returned by TryConvertValue, which is a private method called only by the SegmentArgumentParser.TryConvertValue method in order to determine key values.  This does not affect how enum values are parsed outside of key value segments (enumBinder still returns a constant node).

TryConvertValue is only called internally by CreateKeySegment, but is a public method, so the possible breaking changes would be:
1) if a user directly called TryConvertValue on an enum key value segment and expected the resulting value would be wrapped in a constant node (and didn't handle the unwrapped enum value). 
2) if a user called CreateKeySegment, and expected the enum value to come back wrapped in a constant node (and didn't have code to handle an unwrapped enum value).

Within a key segment there is no reason to wrap the enum value in a constant node, and none of the other key segments are wrapped in constant nodes (which is why the LiteralFormatter didn't expect the enum to be wrapped in a constant node).

It could be argued that generic external code that tries to consume a parsed key segment is likely also not to expect that the enum value is wrapped in a constant node (unless they explicitly encounter this and already hardcoded the assumption that enums are wrapped in a constant node, without handling the unwrapped value).

Support that this is relatively safe comes from the fact that the only logic/tests in the build that were broken are the unit tests that explicitly checked that the segments were constant nodes.  None of the E2E tests had to change to accommodate the new behavior.

Enums in keys are fairly rare and, as the original issue shows, are today broken when trying to write the url (i.e., id/edit/readlink to the instance, or any nav prop off the instance).

I'm not totally comfortable with this, as it is a breaking change, but it's arguably more likely that someone already using an enum as a key segment is already broken than that this change would break them.

mikepizzo:
Looking at the code change, the code only affects what is returned by TryConvertValue, which is a private method called only by the SegmentArgumentParser.TryConvertValue method in order to determine key values. This does not affect how enum values are parsed outside of key value segments (enumBinder still returns a constant node).

TryConvertValue is only called internally by CreateKeySegment, but is a public method, so the possible breaking changes would be:

1) if a user directly called TryConvertValue on an enum key value segment and expected the resulting value would be wrapped in a constant node (and didn't handle the unwrapped enum value).
2) if a user called CreateKeySegment, and expected the enum value to come back wrapped in a constant node (and didn't have code to handle an unwrapped enum value).

Within a key segment there is no reason to wrap the enum value in a constant node, and none of the other key segments are wrapped in constant nodes (which is why the LiteralFormatter didn't expect the enum to be wrapped in a constant node).

It could be argued that generic external code that tries to consume a parsed key segment is likely also not to expect that the enum value is wrapped in a constant node (unless they explicitly encounter this and already hardcoded the assumption that enums are wrapped in a constant node, without handling the unwrapped value).

Support that this is relatively safe comes from the fact that the only logic/tests in the build that were broken are the unit tests that explicitly checked that the segments were constant nodes. None of the E2E tests had to change to accommodate the new behavior.

Enums in keys are fairly rare and, as the original issue shows, are today broken when trying to write the url (i.e., id/edit/readlink to the instance, or any nav prop off the instance).

I'm not totally comfortable with this, as it is a breaking change, but it's arguably more likely that someone already using an enum as a key segment is already broken than that this change would break them.

---
In reply to: [523672818](https://github.com/OData/odata.net/pull/1507#issuecomment-523672818) [](ancestors = 523672818)

xuzhg:
Merged at: c64b89775306a909fdf5e71287d08ba6012ae19c

PR #1516, opened 2019-08-01
Dont Enforce ContentID Uniqueness outside ChangeSet in MultipartMixed

mikepizzo:
### Description

Prior to the changes for JSON Batch merged in 7.4, we did not enforce uniqueness of ContentID outside of a changeset in MultipartMixed batch format.

The OData 4.0 spec states that, within a changeset, each entry must have a ContentID and that content-id MUST be unique across the batch (not just the changeset). It doesn't mention ContentIds outside of the changeset, but the previous statement implies that the scope of uniqueness is the entire batch.  In OData 4.01 we tightened the wording to clarify that all ContentIds within the batch must be unique across the batch. This is especially important for JSON Batch responses, which can come back in any order so the ContentId is required to associated a response with a request, but the statement is made as a format-independent statement in the core protocol spec.

So, while it is arguably the right thing to do to enforce uniqueness across the batch for all formats, this is technically a breaking change to add this enforcement for the multipartmixed format.

This change request removes the uniqueness check across the batch for multipart/mixed in version 4.0; Content-Id must be unique within a changeset, but not across the batch. That also means, if Version 4.0 is specified, you can't reference things outside of the current changeset. 

For 4.01, and for all versions of JSON Batch, you Content-ID must be unique across the entire batch.

- [ x] *Test cases added*
- [ x] *Build and test with one-click build and test script passed*

### Additional work necessary

*None.*

### DIFF
--- a/src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchReader.cs
+++ b/src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchReader.cs
@@ -91,6 +91,12 @@ namespace Microsoft.OData.MultipartMixed
                     }
                 }
             }
+            else
+            {
+                // For backward compatibility, don't enforce uniqueness of Content-ID 
+                // in MultipartMixed outside of a changeset
+                this.payloadUriConverter.Reset();
+            }
 
             ODataBatchOperationRequestMessage requestMessage = BuildOperationRequestMessage(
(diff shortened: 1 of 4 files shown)


mikepizzo:
Built successfully; the CI pipeline build failed because it was unable to get sources (apparently because the branch was created in the OData repo, not my private repo).

PR #1523, opened 2019-08-14
ODataError protocol compliance change

KanishManuja-MS:
Change ODataError's ODataInnerError to be open by introducing a Properties dictionary which will allow for serializing and deserializing an error object with custom property names on the error object.

### DIFF
--- a/src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
+++ b/src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
@@ -13,7 +13,6 @@ namespace Microsoft.OData.Json
     using System.IO;
     using System.Text;
 
-
     /// <summary>
     /// Extension methods for the JSON reader.
     /// </summary>
--- a/src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
+++ b/src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
@@ -7,10 +7,12 @@
 namespace Microsoft.OData.Json
 {
     using System;
+    using System.Collections.Generic;
     using System.Diagnostics;
     using System.Globalization;
     using System.IO;
     using System.Text;
+
 
     /// <summary>
     /// Extension methods for the JSON reader.
@@ -313,6 +315,40 @@ namespace Microsoft.OData.Json
             {
                 RawValue = builder.ToString(),
             };
+        }
+
+        internal static ODataValue ReadODataValue(this IJsonReader jsonReader)
+        {
+            if (jsonReader.NodeType == JsonNodeType.PrimitiveValue)
+            {
+                object primitiveValue = jsonReader.ReadPrimitiveValue();
+
+                return primitiveValue.ToODataValue();
+            }
+            else if (jsonReader.NodeType == JsonNodeType.StartObject)
+            {
+                jsonReader.ReadStartObject();
+                ODataResourceValue resourceValue = new ODataResourceValue();
+                IList<ODataProperty> propertyList = new List<ODataProperty>();
+
+                while (jsonReader.NodeType != JsonNodeType.EndObject)
+                {
+                    ODataProperty property = new ODataProperty();
+                    property.Name = jsonReader.ReadPropertyName();
+                    property.Value = jsonReader.ReadODataValue();
+                    propertyList.Add(property);
+                }
+
+                resourceValue.Properties = propertyList;
+
+                jsonReader.ReadEndObject();
+
+                return resourceValue;
+            }
+            else
+            {
+                return jsonReader.ReadAsUntypedOrNullValue();
+            }
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
+++ b/src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
@@ -344,6 +344,22 @@ namespace Microsoft.OData.Json
 
                 return resourceValue;
             }
+            else if (jsonReader.NodeType == JsonNodeType.StartArray)
+            {
+                jsonReader.ReadStartArray();
+                ODataCollectionValue collectionValue = new ODataCollectionValue();
+                IList<object> propertyList = new List<object>();
+
+                while (jsonReader.NodeType != JsonNodeType.EndArray)
+                {
+                    propertyList.Add(jsonReader.ReadODataValue());
+                }
+
+                collectionValue.Items = propertyList;
+                jsonReader.ReadEndArray();
+
+                return collectionValue;
+            }
             else
             {
                 return jsonReader.ReadAsUntypedOrNullValue();
--- a/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
@@ -249,14 +249,14 @@ namespace Microsoft.OData.Json
                 foreach (object item in collectionValue.Items)
                 {
                     // Will not be able to accurately serialize complex objects unless they are ODataValues. 
-                    if (item is ODataValue collectionItem)
+                    ODataValue collectionItem = item as ODataValue;
+                    if (item != null)
                     {
                         jsonWriter.WriteODataValue(collectionItem);
                     }
                     else
                     {
-                        throw new ODataException(
-                            "Unable to serialize an object in a collection as it is not an ODataPrimitve or ODataResourceValue.");
+                        throw new ODataException(ODataErrorStrings.ODataJsonWriter_UnsupportedValueInCollection);
                     }
                 }
 
@@ -266,7 +266,7 @@ namespace Microsoft.OData.Json
             }
 
             throw new ODataException(
-                "Unable to serialize the ODataValue.");
+                ODataErrorStrings.ODataJsonWriter_UnsupportedValueType(odataValue.GetType().FullName));
         }
     }
 }
--- a/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
@@ -181,7 +181,7 @@ namespace Microsoft.OData.Json
         /// </summary>
         /// <param name="jsonWriter">The <see cref="JsonWriter"/> to write to.</param>
         /// <param name="propertyValue">value to write.</param>
-        private static void WriteJsonValue(this IJsonWriter jsonWriter, object propertyValue)
+        internal static void WriteJsonValue(this IJsonWriter jsonWriter, object propertyValue)
         {
             if (propertyValue == null)
             {
--- a/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
@@ -8,13 +8,12 @@ namespace Microsoft.OData.Json
 {
     #region Namespaces
     using System;
-    using System.Diagnostics;
-    using System.Collections.Generic;
     using System.Collections;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using ODataErrorStrings = Microsoft.OData.Strings;
     using Microsoft.OData.Edm;
     using Microsoft.OData.Metadata;
-    using ODataErrorStrings = Microsoft.OData.Strings;
-    using ODataPlatformHelper = Microsoft.OData.PlatformHelper;
     #endregion Namespaces
 
     /// <summary>
@@ -205,6 +204,43 @@ namespace Microsoft.OData.Json
                     jsonWriter.WriteJsonArrayValue(arrayValue);
                 }
             }
+        }
+
+        /// <summary>
+        /// Writes the json value (primitive, IDictionary or IEnumerable) to the underlying json writer.
+        /// </summary>
+        /// <param name="jsonWriter">The <see cref="JsonWriter"/> to write to.</param>
+        /// <param name="propertyValue">value to write.</param>
+        internal static void WriteODataValue(this IJsonWriter jsonWriter, ODataValue odataValue)
+        {
+            if (odataValue == null)
+            {
+                jsonWriter.WriteValue((string)null);
+                return;
+            }
+
+            object objectValue = odataValue.FromODataValue();
+            if (EdmLibraryExtensions.IsPrimitiveType(objectValue.GetType()))
+            {
+                jsonWriter.WritePrimitiveValue(objectValue);
+                return;
+            }
+
+            ODataResourceValue resourceValue = odataValue as ODataResourceValue;
+            if (resourceValue != null)
+            {
+                jsonWriter.StartObjectScope();
+
+                foreach (ODataProperty property in resourceValue.Properties)
+                {
+                    jsonWriter.WriteName(property.Name);
+                    jsonWriter.WriteODataValue(property.ODataValue);
+                }
+
+                jsonWriter.EndObjectScope();
+            }
+
+            return;
         }
     }
 }
--- a/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
@@ -11,9 +11,9 @@ namespace Microsoft.OData.Json
     using System.Collections;
     using System.Collections.Generic;
     using System.Diagnostics;
-    using ODataErrorStrings = Microsoft.OData.Strings;
     using Microsoft.OData.Edm;
     using Microsoft.OData.Metadata;
+    using ODataErrorStrings = Microsoft.OData.Strings;
     #endregion Namespaces
 
     /// <summary>
--- a/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
@@ -213,7 +213,7 @@ namespace Microsoft.OData.Json
         /// <param name="propertyValue">value to write.</param>
         internal static void WriteODataValue(this IJsonWriter jsonWriter, ODataValue odataValue)
         {
-            if (odataValue == null)
+            if (odataValue == null || odataValue is ODataNullValue)
             {
                 jsonWriter.WriteValue((string)null);
                 return;
--- a/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
@@ -180,7 +180,7 @@ namespace Microsoft.OData.Json
         /// </summary>
         /// <param name="jsonWriter">The <see cref="JsonWriter"/> to write to.</param>
         /// <param name="propertyValue">value to write.</param>
-        internal static void WriteJsonValue(this IJsonWriter jsonWriter, object propertyValue)
+        private static void WriteJsonValue(this IJsonWriter jsonWriter, object propertyValue)
         {
             if (propertyValue == null)
             {
@@ -238,9 +238,35 @@ namespace Microsoft.OData.Json
                 }
 
                 jsonWriter.EndObjectScope();
+                return;
             }
 
-            return;
+            ODataCollectionValue collectionValue = odataValue as ODataCollectionValue;
+            if (collectionValue != null)
+            {
+                jsonWriter.StartArrayScope();
+
+                foreach (object item in collectionValue.Items)
+                {
+                    // Will not be able to accurately serialize complex objects unless they are ODataValues. 
+                    if (item is ODataValue collectionItem)
+                    {
+                        jsonWriter.WriteODataValue(collectionItem);
+                    }
+                    else
+                    {
+                        throw new ODataException(
+                            "Unable to serialize an object in a collection as it is not an ODataPrimitve or ODataResourceValue.");
+                    }
+                }
+
+                jsonWriter.EndArrayScope();
+
+                return; 
+            }
+
+            throw new ODataException(
+                "Unable to serialize the ODataValue.");
         }
     }
 }
--- a/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
@@ -157,60 +157,10 @@ namespace Microsoft.OData.Json
         }
 
         /// <summary>
-        /// Writes the json array value.
+        /// Writes the ODataValue (primitive, collection or resource value) to the underlying json writer.
         /// </summary>
         /// <param name="jsonWriter">The <see cref="JsonWriter"/> to write to.</param>
-        /// <param name="arrayValue">Writes the json array value to the underlying json writer.</param>
-        private static void WriteJsonArrayValue(this IJsonWriter jsonWriter, IEnumerable arrayValue)
-        {
-            Debug.Assert(arrayValue != null, "arrayValue != null");
-
-            jsonWriter.StartArrayScope();
-
-            foreach (object element in arrayValue)
-            {
-                jsonWriter.WriteJsonValue(element);
-            }
-
-            jsonWriter.EndArrayScope();
-        }
-
-        /// <summary>
-        /// Writes the json value (primitive, IDictionary or IEnumerable) to the underlying json writer.
-        /// </summary>
-        /// <param name="jsonWriter">The <see cref="JsonWriter"/> to write to.</param>
-        /// <param name="propertyValue">value to write.</param>
-        private static void WriteJsonValue(this IJsonWriter jsonWriter, object propertyValue)
-        {
-            if (propertyValue == null)
-            {
-                jsonWriter.WriteValue((string)null);
-            }
-            else if (EdmLibraryExtensions.IsPrimitiveType(propertyValue.GetType()))
-            {
-                jsonWriter.WritePrimitiveValue(propertyValue);
-            }
-            else
-            {
-                IDictionary<string, object> objectValue = propertyValue as IDictionary<string, object>;
-                if (objectValue != null)
-                {
-                    jsonWriter.WriteJsonObjectValue(objectValue, null /*typeName */);
-                }
-                else
-                {
-                    IEnumerable arrayValue = propertyValue as IEnumerable;
-                    Debug.Assert(arrayValue != null, "arrayValue != null");
-                    jsonWriter.WriteJsonArrayValue(arrayValue);
-                }
-            }
-        }
-
-        /// <summary>
-        /// Writes the json value (primitive, IDictionary or IEnumerable) to the underlying json writer.
-        /// </summary>
-        /// <param name="jsonWriter">The <see cref="JsonWriter"/> to write to.</param>
-        /// <param name="propertyValue">value to write.</param>
+        /// <param name="odataValue">value to write.</param>
         internal static void WriteODataValue(this IJsonWriter jsonWriter, ODataValue odataValue)
         {
             if (odataValue == null || odataValue is ODataNullValue)
@@ -262,11 +212,61 @@ namespace Microsoft.OData.Json
 
                 jsonWriter.EndArrayScope();
 
-                return; 
+                return;
             }
 
             throw new ODataException(
                 ODataErrorStrings.ODataJsonWriter_UnsupportedValueType(odataValue.GetType().FullName));
+        }
+
+        /// <summary>
+        /// Writes the json array value.
+        /// </summary>
+        /// <param name="jsonWriter">The <see cref="JsonWriter"/> to write to.</param>
+        /// <param name="arrayValue">Writes the json array value to the underlying json writer.</param>
+        private static void WriteJsonArrayValue(this IJsonWriter jsonWriter, IEnumerable arrayValue)
+        {
+            Debug.Assert(arrayValue != null, "arrayValue != null");
+
+            jsonWriter.StartArrayScope();
+
+            foreach (object element in arrayValue)
+            {
+                jsonWriter.WriteJsonValue(element);
+            }
+
+            jsonWriter.EndArrayScope();
+        }
+
+        /// <summary>
+        /// Writes the json value (primitive, IDictionary or IEnumerable) to the underlying json writer.
+        /// </summary>
+        /// <param name="jsonWriter">The <see cref="JsonWriter"/> to write to.</param>
+        /// <param name="propertyValue">value to write.</param>
+        private static void WriteJsonValue(this IJsonWriter jsonWriter, object propertyValue)
+        {
+            if (propertyValue == null)
+            {
+                jsonWriter.WriteValue((string)null);
+            }
+            else if (EdmLibraryExtensions.IsPrimitiveType(propertyValue.GetType()))
+            {
+                jsonWriter.WritePrimitiveValue(propertyValue);
+            }
+            else
+            {
+                IDictionary<string, object> objectValue = propertyValue as IDictionary<string, object>;
+                if (objectValue != null)
+                {
+                    jsonWriter.WriteJsonObjectValue(objectValue, null /*typeName */);
+                }
+                else
+                {
+                    IEnumerable arrayValue = propertyValue as IEnumerable;
+                    Debug.Assert(arrayValue != null, "arrayValue != null");
+                    jsonWriter.WriteJsonArrayValue(arrayValue);
+                }
+            }
         }
     }
 }
--- a/src/Microsoft.OData.Core/Json/ODataJsonReaderCoreUtils.cs
+++ b/src/Microsoft.OData.Core/Json/ODataJsonReaderCoreUtils.cs
@@ -130,7 +130,7 @@ namespace Microsoft.OData.Json
         /// <param name="inputContext">The input context with all the settings.</param>
         /// <param name="recursionDepth">The recursion depth to start with.</param>
         /// <returns>an instance of IDictionary containing the spatial value.</returns>
-        private static IDictionary<string, object> ReadObjectValue(IJsonReader jsonReader, bool insideJsonObjectValue, ODataInputContext inputContext, int recursionDepth)
+        internal static IDictionary<string, object> ReadObjectValue(IJsonReader jsonReader, bool insideJsonObjectValue, ODataInputContext inputContext, int recursionDepth)
         {
             Debug.Assert(jsonReader != null, "jsonReader != null");
             Debug.Assert(insideJsonObjectValue || jsonReader.NodeType == JsonNodeType.StartObject, "insideJsonObjectValue || jsonReader.NodeType == JsonNodeType.StartObject");
--- a/src/Microsoft.OData.Core/Json/ODataJsonReaderCoreUtils.cs
+++ b/src/Microsoft.OData.Core/Json/ODataJsonReaderCoreUtils.cs
@@ -130,7 +130,7 @@ namespace Microsoft.OData.Json
         /// <param name="inputContext">The input context with all the settings.</param>
         /// <param name="recursionDepth">The recursion depth to start with.</param>
         /// <returns>an instance of IDictionary containing the spatial value.</returns>
-        internal static IDictionary<string, object> ReadObjectValue(IJsonReader jsonReader, bool insideJsonObjectValue, ODataInputContext inputContext, int recursionDepth)
+        private static IDictionary<string, object> ReadObjectValue(IJsonReader jsonReader, bool insideJsonObjectValue, ODataInputContext inputContext, int recursionDepth)
         {
             Debug.Assert(jsonReader != null, "jsonReader != null");
             Debug.Assert(insideJsonObjectValue || jsonReader.NodeType == JsonNodeType.StartObject, "insideJsonObjectValue || jsonReader.NodeType == JsonNodeType.StartObject");
--- a/src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
+++ b/src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
@@ -262,15 +262,21 @@ namespace Microsoft.OData.Json
             {
                 sb.Append("null");
             }
-            else if (value is ODataCollectionValue collectionValue)
+
+            ODataCollectionValue collectionValue = value as ODataCollectionValue;
+            if (collectionValue != null)
             {
                 ODataCollectionValueToString(sb, collectionValue);
             }
-            else if (value is ODataResourceValue resourceValue)
+
+            ODataResourceValue resourceValue = value as ODataResourceValue;
+            if (resourceValue != null)
             {
                 ODataResourceValueToString(sb, resourceValue);
             }
-            else if (value is ODataPrimitiveValue primitiveValue)
+
+            ODataPrimitiveValue primitiveValue = value as ODataPrimitiveValue;
+            if (primitiveValue != null)
             {
                 if (primitiveValue.FromODataValue() is string)
                 {
@@ -298,13 +304,14 @@ namespace Microsoft.OData.Json
                     sb.Append(",");
                 }
 
-                if (item is ODataValue odataValue)
+                ODataValue odataValue = item as ODataValue;
+                if (odataValue != null)
                 {
                     ODataValueToString(sb, odataValue);
                 }
                 else
                 {
-                    sb.Append(item.ToString());
+                    throw new ODataException("");
                 }
             }
 
--- a/src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
+++ b/src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
@@ -269,7 +269,7 @@ namespace Microsoft.OData.Json
             if (innerError.InnerError != null)
             {
                 // "internalexception": { <nested inner error> }
-                WriteInnerError(jsonWriter, innerError.InnerError, string.IsNullOrWhiteSpace(innerError.NestedObjectName) ? JsonConstants.ODataErrorInnerErrorInnerErrorName : innerError.NestedObjectName, recursionDepth, maxInnerErrorDepth, skipNullProperties);
+                WriteInnerError(jsonWriter, innerError.InnerError, string.IsNullOrEmpty(innerError.NestedObjectName) ? JsonConstants.ODataErrorInnerErrorInnerErrorName : innerError.NestedObjectName, recursionDepth, maxInnerErrorDepth, skipNullProperties);
             }
 
             // }
--- a/src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
+++ b/src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
@@ -13,6 +13,7 @@ namespace Microsoft.OData.Json
     using System.Linq;
     using System.Text;
     using Microsoft.OData.JsonLight;
+    using ODataErrorStrings = Microsoft.OData.Strings;
     #endregion Namespaces
 
     /// <summary>
@@ -311,7 +312,7 @@ namespace Microsoft.OData.Json
                 }
                 else
                 {
-                    throw new ODataException("");
+                    throw new ODataException(ODataErrorStrings.ODataJsonWriter_UnsupportedValueInCollection);
                 }
             }
 
--- a/src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
+++ b/src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
@@ -28,6 +28,7 @@ namespace Microsoft.OData.Json
         /// <param name="includeDebugInformation">A flag indicating whether error details should be written (in debug mode only) or not.</param>
         /// <param name="maxInnerErrorDepth">The maximum number of nested inner errors to allow.</param>
         /// <param name="writingJsonLight">true if we're writing JSON lite, false if we're writing verbose JSON.</param>
+        /// <param name="skipNullProperties">if set, stacktrace, type, and message will not be written if they are empty strings on ODataInnerError.</param>
         internal static void WriteError(IJsonWriter jsonWriter,
             Action<IEnumerable<ODataInstanceAnnotation>> writeInstanceAnnotationsDelegate, ODataError error,
             bool includeDebugInformation, int maxInnerErrorDepth, bool writingJsonLight, bool skipNullProperties)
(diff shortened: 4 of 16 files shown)


src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
```diff
@@ -144,6 +144,12 @@ public Uri BaseUri
         /// </summary>
         public bool EnableCharactersCheck { get; set; }
 
+
+        /// <summary>
+        /// Flag to control reading InnerError of ODataError as an open type. 
+        /// </summary>
+        public bool EnableExtendingInnerError { get; set; }
```
mikepizzo: >EnableExtendingInnerError [](start = 20, length = 25)

What is the value in adding this flag?

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -224,27 +224,40 @@ private static void WriteInnerError(IJsonWriter jsonWriter, ODataInnerError inne
...
-            ////       to stay compatible with Astoria.
+            if (innerError.Properties != null)
+            {
+                foreach (KeyValuePair<string, string> pair in innerError.Properties)
+                {
+                    jsonWriter.WriteName(pair.Key);
+                    jsonWriter.WriteValue(pair.Value);
+                }
+            }
+            else
```
mikepizzo: >     else [](start = 7, length = 9)

Should this really be else? does that mean we either write the open properties or the defined properties, but not both?

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -4,6 +4,8 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using System.Collections.Generic;
```
mikepizzo: move to be with other namespaces

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -30,17 +32,44 @@ public ODataInnerError()
         /// <summary>Initializes a new instance of the <see cref="T:Microsoft.OData.ODataInnerError" /> class with exception object.</summary>
         /// <param name="exception">The <see cref="T:System.Exception" /> used to create the inner error.</param>
         public ODataInnerError(Exception exception)
+            : this(exception, "internalexception")
```
mikepizzo: >"internalexception" [](start = 30, length = 19)

This should be a defined constant, like JsonConstants.ODataErrorInnerErrorInnerErrorName (except not JSON specific).

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -30,17 +32,44 @@ public ODataInnerError()
...
         {
-            ExceptionUtils.CheckArgumentNotNull(exception, "exception");
 
-            this.Message = exception.Message ?? string.Empty;
-            this.TypeName = exception.GetType().FullName;
-            this.StackTrace = exception.StackTrace;
+        }
+
+        public ODataInnerError(Exception exception, string nestedObjectName)
```
mikepizzo: >nestedObjectName [](start = 59, length = 16)

if this is not allowed to be null, add a CheckArgumentNotNull; otherwise call out in XML Comments (once added) that it may be null.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightErrorDeserializer.cs
```diff
@@ -333,7 +333,24 @@ private void ReadPropertyValueInInnerError(int recursionDepth, ODataInnerError i
...
+                            innerError.InnerError = this.ReadInnerError(recursionDepth);
+                            break;
+                        }
+
+                        if (innerError.Properties == null)
+                        {
+                            innerError.Properties = new Dictionary<string, string>();
+                        }
+
+                        innerError.Properties.Add(propertyName, this.JsonReader.ReadStringValue(propertyName));
```
mikepizzo: >ReadStringValue [](start = 80, length = 15)

why must open properties be string?

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -30,17 +32,44 @@ public ODataInnerError()
...
+                this.InnerError = new ODataInnerError(exception.InnerException, nestedObjectName);
             }
+
+            Properties = new Dictionary<string, string>();
+            Properties.Add("message", exception.Message ?? string.Empty);
+            Properties.Add("type", exception.GetType().FullName ?? string.Empty);
+            Properties.Add("stacktrace", exception.Message ?? string.Empty);
+        }
+
+        public ODataInnerError(IDictionary<string, string> properties, string nestedObjectName, ODataInnerError nestedObject)
```
mikepizzo: >string [](start = 51, length = 6)

do dynamic properties have to be string-valued? why not dictionary<string,object>?

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -30,17 +32,44 @@ public ODataInnerError()
...
             {
-                this.InnerError = new ODataInnerError(exception.InnerException);
+                this.InnerError = new ODataInnerError(exception.InnerException, nestedObjectName);
             }
+
+            Properties = new Dictionary<string, string>();
+            Properties.Add("message", exception.Message ?? string.Empty);
+            Properties.Add("type", exception.GetType().FullName ?? string.Empty);
+            Properties.Add("stacktrace", exception.Message ?? string.Empty);
+        }
```
mikepizzo: So the Message, TypeName, and Stacktrace properties are no longer used?  If I have existing code that creates an inner error and then sets these, those values will no longer be used?   And if I create an instance I can no longer get those values from the properties?

mikepizzo:
Not sure how much of this code is for experiment versus intended design, but if you are moving to storing Message, TypeName, and StackTrace in the Properties collection I think it will be problematic.  I think you want to keep the top level properties as top level (and perhaps add additional properties if the protocol calls for them) and have Properties for other properties. Only (potentially) breaking change to consider is not writing TypeName and StackTrace if not specified (rather than writing as String.Empty), but that may have to be made under a flag or in a future release.

src/Microsoft.OData.Core/ODataInnerError.cs
xuzhg: It could be Dictionary<string, object>. However, the object here should be instance of built in "primitive", enum, resource, and collection. Otherwise, maybe use Dictionary<string, ODataValue>.

---
In reply to: [314483310](https://github.com/OData/odata.net/pull/1523#discussion_r314483310) [](ancestors = 314483310)

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -30,17 +32,44 @@ public ODataInnerError()
...
+        public ODataInnerError(IDictionary<string, string> properties, string nestedObjectName, ODataInnerError nestedObject)
+        {
+            Properties = properties;
+            this.NestedObjectName = nestedObjectName;
+        }
+
+        /// <summary>
+        /// Nested object's name
+        /// </summary>
+        internal string NestedObjectName;
```
xuzhg: >internal string NestedObjectName; [](start = 8, length = 33)

why do we need this string? From the OData spec, the property name should be "innererror".

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -224,27 +224,40 @@ private static void WriteInnerError(IJsonWriter jsonWriter, ODataInnerError inne
             jsonWriter.WriteName(innerErrorPropertyName);
```
xuzhg: >WriteName(innerErrorPropertyName); [](start = 23, length = 34)

Here writes the property name. "innererror".

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -224,27 +224,40 @@ private static void WriteInnerError(IJsonWriter jsonWriter, ODataInnerError inne
...
+
+                if (innerError.InnerError != null)
+                {
+                    // "internalexception": { <nested inner error> }
+                    WriteInnerError(jsonWriter, innerError.InnerError, innerError.NestedObjectName, recursionDepth, maxInnerErrorDepth);
+                }
 
-            if (innerError.InnerError != null)
-            {
-                // "internalexception": { <nested inner error> }
```
xuzhg: >"internalexception": { <nested inner error> } [](start = 19, length = 45)

internalexception is not odata spec compliance? Should remove

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -224,27 +224,40 @@ private static void WriteInnerError(IJsonWriter jsonWriter, ODataInnerError inne
...
+            }
+            else
+            {
+                //// NOTE: we add empty elements if no information is provided for the message, error type and stack trace
+                ////       to stay compatible with Astoria.
 
-            // "message": "<message>"
-            jsonWriter.WriteName(JsonConstants.ODataErrorInnerErrorMessageName);
-            jsonWriter.WriteValue(innerError.Message ?? string.Empty);
+                // "message": "<message>"
```
xuzhg: >// "message": "<message>" [](start = 16, length = 25)

if we save the "message", "type", "stacktrace" into the Properties, it seems we don't need these codes?

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -30,41 +31,80 @@ public ODataInnerError()
...
+
+        /// <summary>
+        /// Properties to be written for the inner error.
+        /// </summary>
+        public IDictionary<string, ODataValue> Properties
+        {
+            get;
+            set;
```
xuzhg: >set; [](start = 12, length = 4)

I don't think it's ok to public set this "Properties", because you set it in the constrctor

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -93,5 +133,34 @@ internal string ToJson()
                 this.StackTrace == null ? "" : JsonValueUtils.GetEscapedJsonString(this.StackTrace),
                 this.InnerError == null ? "{}" : this.InnerError.ToJson());
         }
-}
+
+        private string LookUpStringValue(string v)
```
xuzhg: >LookUpStringValue [](start = 23, length = 17)

Maybe rename it as "GetStringValue", inconsistent with SetStringValue

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -30,41 +31,80 @@ public ODataInnerError()
...
+            Properties.Add(JsonConstants.ODataErrorInnerErrorStackTraceName, exception.StackTrace.ToODataValue() ?? string.Empty.ToODataValue());
+        }
+
+        /// <summary>
+        ///  Initializes a new instance of the <see cref="T:Microsoft.OData.ODataInnerError" /> class with a dictionary of property names and corresponding ODataValues.
+        /// </summary>
+        /// <param name="properties">Dictionary of string keys with ODataValue as value. Key string indicates the property name where as the value of the property is encapsulated in ODataValue.</param>
+        /// <param name="nestedInnerErrorName">The name of the field for the nested inner error object inside inner error.</param>
+        /// <param name="nestedObject">Nested inner error object.</param>
+        public ODataInnerError(IDictionary<string, ODataValue> properties, string nestedInnerErrorName, ODataInnerError nestedObject)
```
xuzhg: >IDictionary<string, ODataValue> properties [](start = 31, length = 42)

check the null for this dictionary?

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -93,5 +133,34 @@ internal string ToJson()
...
                 this.InnerError == null ? "{}" : this.InnerError.ToJson());
         }
-}
+
+        private string LookUpStringValue(string v)
+        {
+            if (Properties == null)
+            {
+                return string.Empty;
+            }
```
xuzhg: Properties should never be null?

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -93,5 +133,34 @@ internal string ToJson()
...
+
+            foreach (KeyValuePair<string, ODataValue> pair in Properties)
+            {
+                if (pair.Key == v)
+                {
+                    string value = (string)pair.Value.FromODataValue();
```
xuzhg: >(string) [](start = 35, length = 8)

static cast to string maybe unsafe. Shall we test it using "ODataPrimitiveValue" with String type?

xuzhg: Or call "ToString()"?

---
In reply to: [320874750](https://github.com/OData/odata.net/pull/1523#discussion_r320874750) [](ancestors = 320874750)

xuzhg:
>         internal string ToJson()

Should change this method to use "Properties"

---
Refers to: src/Microsoft.OData.Core/ODataInnerError.cs:122 in 4aa33e8. [](commit_id = 4aa33e8fde27f62038c678ca4f048ed9b4c680e3, deletion_comment = False)

xuzhg:
>         public ODataInnerError InnerError

shall we "Obsolete" it?

---
Refers to: src/Microsoft.OData.Core/ODataInnerError.cs:112 in 4aa33e8. [](commit_id = 4aa33e8fde27f62038c678ca4f048ed9b4c680e3, deletion_comment = False)

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -93,5 +133,34 @@ internal string ToJson()
...
+
+            return string.Empty;
+        }
+
+        private void SetStringValue(string v, string value)
+        {
+            if (Properties == null)
+            {
+                Properties = new Dictionary<string, ODataValue>();
+            }
```
xuzhg: Properties should never be null? maybe just Debug.Assert(...)

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
```diff
@@ -134,6 +134,11 @@ public Uri BaseUri
         /// the provided function name and parenthesis for JSONP. Otherwise this value is ignored.</remarks>
         public string JsonPCallback { get; set; }
 
+        /// <summary>
+        /// Gets or sets a flag that determines if empty string values should be written for properties on ODataInnerError.
+        /// </summary>
+        public bool SkipNullPropertiesInError { get; set; }
```
xuzhg: >SkipNullPropertiesInError [](start = 20, length = 25)

I think we have the ignore null feature in progress, right? @Mike

src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
```diff
@@ -7,11 +7,13 @@
 namespace Microsoft.OData.Json
 {
     using System;
+    using System.Collections.Generic;
     using System.Diagnostics;
     using System.Globalization;
     using System.IO;
     using System.Text;
 
+
```
xuzhg: revert

src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
```diff
@@ -315,6 +317,40 @@ internal static ODataValue ReadAsUntypedOrNullValue(this IJsonReader jsonReader)
...
 
+        internal static ODataValue ReadODataValue(this IJsonReader jsonReader)
+        {
+            if (jsonReader.NodeType == JsonNodeType.PrimitiveValue)
+            {
+                object primitiveValue = jsonReader.ReadPrimitiveValue();
+
+                return primitiveValue.ToODataValue();
+            }
+            else if (jsonReader.NodeType == JsonNodeType.StartObject)
```
xuzhg: >(jsonReader.NodeType == JsonNodeType.StartObject) [](start = 20, length = 49)

how about the StartArray NodeType

src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
```diff
@@ -206,5 +205,42 @@ private static void WriteJsonValue(this IJsonWriter jsonWriter, object propertyV
...
+            }
+
+            object objectValue = odataValue.FromODataValue();
+            if (EdmLibraryExtensions.IsPrimitiveType(objectValue.GetType()))
+            {
+                jsonWriter.WritePrimitiveValue(objectValue);
+                return;
+            }
+
+            ODataResourceValue resourceValue = odataValue as ODataResourceValue;
```
xuzhg: >odataValue as ODataResourceValue; [](start = 47, length = 33)

how about ODataCollectionValue?

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -213,36 +216,60 @@ internal static void EndJsonPaddingIfRequired(IJsonWriter jsonWriter, ODataMessa
...
+                {
+                    jsonWriter.WriteName(pair.Key);
+                    jsonWriter.WriteODataValue(pair.Value);
+                }
+            }
 
-            // "message": "<message>"
-            jsonWriter.WriteName(JsonConstants.ODataErrorInnerErrorMessageName);
-            jsonWriter.WriteValue(innerError.Message ?? string.Empty);
+            if (!skipNullProperties)
```
xuzhg: >f (!skipNullProperties [](start = 13, length = 22)

If Properties has null values, the "SkipNullProperties" setting can't effect them?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightErrorDeserializer.cs
```diff
@@ -4,6 +4,8 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using System.Linq;
```
xuzhg: move it into #region Namespaces

src/Microsoft.OData.Core/JsonLight/ODataJsonLightErrorDeserializer.cs
```diff
@@ -314,6 +316,7 @@ private ODataInnerError ReadInnerError(int recursionDepth)
         /// </remarks>
         private void ReadPropertyValueInInnerError(int recursionDepth, ODataInnerError innerError, string propertyName)
         {
+
             switch (propertyName)
```
xuzhg: remove

src/Microsoft.OData.Core/JsonLight/ODataJsonLightErrorDeserializer.cs
```diff
@@ -332,9 +335,17 @@ private void ReadPropertyValueInInnerError(int recursionDepth, ODataInnerError i
...
                     break;
 
+                case JsonConstants.ODataErrorInnerErrorName:
+                    innerError.InnerError = this.ReadInnerError(recursionDepth);
+                    break;
+
                 default:
-                    // skip any unsupported properties in the inner error
-                    this.JsonReader.SkipValue();
+                    if (innerError.Properties == null)
```
xuzhg: >innerError.Properties == null [](start = 24, length = 29)

The Properties is initialized in ODataInnerError cotr?

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -93,5 +133,34 @@ internal string ToJson()
...
+                {
+                    string value = (string)pair.Value.FromODataValue();
+                    return value;
+                }
+            }
+
+            return string.Empty;
+        }
+
+        private void SetStringValue(string v, string value)
```
xuzhg: rename "v"?

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -93,5 +133,34 @@ internal string ToJson()
...
+        {
+            if (Properties == null)
+            {
+                Properties = new Dictionary<string, ODataValue>();
+            }
+
+            Properties[v] = value.ToODataValue();
```
xuzhg: >Properties[v] [](start = 12, length = 13)

shall we check "v" is duplicated?

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataJsonWriterUtilsTests.cs
```diff
@@ -4,7 +4,10 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using System.Collections;
+using System.Collections.Generic;
 using System.IO;
+using System.Security.Cryptography.X509Certificates;
```
xuzhg: >System.Security.Cryptography.X509Certificates; [](start = 6, length = 46)

where do you us this package?

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataJsonWriterUtilsTests.cs
```diff
@@ -86,10 +89,43 @@ public void WriteError_WritesTargetAndDetails()
...
+            ODataJsonWriterUtils.WriteError(
+                jsonWriter,
+                enumerable => { },
+                error,
+                includeDebugInformation: true,
+                maxInnerErrorDepth: 5,
+                writingJsonLight: false,
+                skipNullProperties: false);
+            var result = stringWriter.GetStringBuilder().ToString();
+            result.Should().Be("{\"error\":{\"code\":\"\",\"message\":\"\",\"target\":\"any target\",\"details\":[{\"code\":\"500\",\"target\":\"any target\",\"message\":\"any msg\"}],\"innererror\":{\"stacktrace\":\"NormalString\",\"MyNewObject\":{\"NestedResourcePropertyName\":\"NestedPropertyValue\"},\"message\":\"\",\"type\":\"\",\"innererror\":{\"stacktrace\":\"NormalString\",\"MyNewObject\":{\"NestedResourcePropertyName\":\"NestedPropertyValue\"},\"message\":\"\",\"type\":\"\"}}}}");
```
xuzhg: >"message\":\"\",\"type\" [](start = 306, length = 24)

If "message" or "type" is empty string or null string, should they be in the payload?

KanishManuja-MS:
>         internal string ToJson()

This is not used anywhere internally, it will be a breaking change for ODataError.ToString(). We can clean this up in a major version. 

---
In reply to: [527996900](https://github.com/OData/odata.net/pull/1523#issuecomment-527996900) [](ancestors = 527996900)

---
Refers to: src/Microsoft.OData.Core/ODataInnerError.cs:122 in 4aa33e8. [](commit_id = 4aa33e8fde27f62038c678ca4f048ed9b4c680e3, deletion_comment = False)

KanishManuja-MS:
>         public ODataInnerError InnerError

This is still used in the code. We will have to get rid of all the references before we can mark it obsolete. 

---
In reply to: [527997061](https://github.com/OData/odata.net/pull/1523#issuecomment-527997061) [](ancestors = 527997061)

---
Refers to: src/Microsoft.OData.Core/ODataInnerError.cs:112 in 4aa33e8. [](commit_id = 4aa33e8fde27f62038c678ca4f048ed9b4c680e3, deletion_comment = False)

src/Microsoft.OData.Core/ODataInnerError.cs
KanishManuja-MS: What if someone uses the default constructor. 

---
In reply to: [320874186](https://github.com/OData/odata.net/pull/1523#discussion_r320874186) [](ancestors = 320874186)

src/Microsoft.OData.Core/ODataInnerError.cs
KanishManuja-MS: What if default constructor is used?

---
In reply to: [320875481](https://github.com/OData/odata.net/pull/1523#discussion_r320875481) [](ancestors = 320875481)

src/Microsoft.OData.Core/ODataInnerError.cs
KanishManuja-MS: We should overwrite the value. Remember, we are trying to mimic normal properties. 

---
In reply to: [320879295](https://github.com/OData/odata.net/pull/1523#discussion_r320879295) [](ancestors = 320879295)

src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
KanishManuja-MS: CollectionValue is out of scope because then I would need to know the EdmType of the collection which doesn't make a lot of sense for collection values. 

---
In reply to: [320877192](https://github.com/OData/odata.net/pull/1523#discussion_r320877192) [](ancestors = 320877192)

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -33,38 +41,73 @@ public ODataInnerError(Exception exception)
...
+            Properties.Add(JsonConstants.ODataErrorInnerErrorStackTraceName, exception.StackTrace.ToODataValue() ?? new ODataNullValue());
+        }
+
+        /// <summary>
+        ///  Initializes a new instance of the <see cref="T:Microsoft.OData.ODataInnerError" /> class with a dictionary of property names and corresponding ODataValues.
+        /// </summary>
+        /// <param name="properties">Dictionary of string keys with ODataValue as value. Key string indicates the property name where as the value of the property is encapsulated in ODataValue.</param>
+        /// <param name="nestedInnerErrorName">The name of the field for the nested inner error object inside inner error.</param>
+        /// <param name="nestedObject">Nested inner error object.</param>
+        public ODataInnerError(IDictionary<string, ODataValue> properties)
```
mikepizzo: >ODataInnerError(IDictionary<string, ODataValue> properties) [](start = 15, length = 59)

Do we have a scenario where we need this constructor?

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -33,38 +41,73 @@ public ODataInnerError(Exception exception)
...
+            ExceptionUtils.CheckArgumentNotNull(properties, "properties");
+
+            Properties = properties;
+        }
+
+        /// <summary>
+        /// Creates an instance of ODataInnerError with an empty property bag.
+        /// </summary>
+        /// <returns>returns an instance of ODataInnerError with empty property bag.</returns>
+        public static ODataInnerError CreateEmptyInnerError()
```
mikepizzo: > ODataInnerError CreateEmptyInnerError() [](start = 21, length = 40)

Is this a common enough pattern that we need the method, rather than having the caller simply create a new ODataInnerError with an empty dictionary (i.e., the line below)?

src/Microsoft.OData.Core/ODataInnerError.cs
mikepizzo: Note: params don't match constructor signature.

src/Microsoft.OData.Core/ODataInnerError.cs
mikepizzo: :thumbsup:

---
In reply to: [320930828](https://github.com/OData/odata.net/pull/1523#discussion_r320930828) [](ancestors = 320930828,320879295)

src/Microsoft.OData.Core/ODataInnerError.cs
KanishManuja-MS: Yes, there is a scenario. With this constructor, stacktrace and other properties don't get auto added, therefore this will be the one getting used in the error handling lib. 

---
In reply to: [321856768](https://github.com/OData/odata.net/pull/1523#discussion_r321856768) [](ancestors = 321856768)

src/Microsoft.OData.Core/ODataInnerError.cs
KanishManuja-MS: It will be commonly used in the error handling library but I suppose that it is not a must. However, I wanted to include this as the name clarifies that there will be no auto-populated/default properties.

---
In reply to: [321857649](https://github.com/OData/odata.net/pull/1523#discussion_r321857649) [](ancestors = 321857649)

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -93,5 +136,39 @@ internal string ToJson()
...
+            }
+
+            if (value == null)
+            {
+                Properties[propertyKey] = new ODataNullValue();
+                return;
+            }
+
+            Properties[propertyKey] = value.ToODataValue();
+            return;
```
mikepizzo: >return [](start = 12, length = 6)

unneeded return. in general, this code follows the "early return pattern". I don't have a strong bias either way, but most of our code (especially for void methods) has a single (or few, especially success) exit points.  Not sure if there's a preferred style across Microsoft. We should discuss what coding pattern we want to follow.

src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
```diff
@@ -315,6 +316,40 @@ internal static ODataValue ReadAsUntypedOrNullValue(this IJsonReader jsonReader)
...
+                jsonReader.ReadStartObject();
+                ODataResourceValue resourceValue = new ODataResourceValue();
+                IList<ODataProperty> propertyList = new List<ODataProperty>();
+
+                while (jsonReader.NodeType != JsonNodeType.EndObject)
+                {
+                    ODataProperty property = new ODataProperty();
+                    property.Name = jsonReader.ReadPropertyName();
+                    property.Value = jsonReader.ReadODataValue();
+                    propertyList.Add(property);
```
mikepizzo: what about nested objects? Should this call into ReadODataValue, and have a separate case for if (jsonReader.NodeType == JsonNodeType.Property)?

src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
mikepizzo: So, if the payload contains an array, we will attempt to read using ReadAsUntypedOrNullValue -- is that the expected/desired behavior?

---
In reply to: [320876548](https://github.com/OData/odata.net/pull/1523#discussion_r320876548) [](ancestors = 320876548)

src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
mikepizzo: The code should still be prepared to handle the case, though, right? even if it returns a (not supported?) error? Currently, it looks like we don't check for an ODataCollectionValue, so we'll just silently ignore?  Is that the desired behavior?

---
In reply to: [320936971](https://github.com/OData/odata.net/pull/1523#discussion_r320936971) [](ancestors = 320936971,320877192)

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -33,38 +41,71 @@ public ODataInnerError(Exception exception)
...
+        /// <returns>returns an instance of ODataInnerError with empty property bag.</returns>
+        public static ODataInnerError CreateEmptyInnerError()
+        {
+            return new ODataInnerError(new Dictionary<string, ODataValue>());
+        }
+
+        /// <summary>
+        /// Properties to be written for the inner error.
+        /// </summary>
+        public IDictionary<string, ODataValue> Properties
```
mikepizzo: >Properties [](start = 47, length = 10)

The code doesn't seem to support properties of type ODataCollectionValue. If this is a restriction, we might want to wrap the IDictionary and throw an error if the user tries to set a value that is an ODataCollectionValue.  Might be easier just to support reading/writing the collection value...

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -213,31 +213,39 @@ internal static void EndJsonPaddingIfRequired(IJsonWriter jsonWriter, ODataMessa
...
+            if (innerError.Properties != null)
+            {
+                foreach (KeyValuePair<string, ODataValue> pair in innerError.Properties)
+                {
+                    jsonWriter.WriteName(pair.Key);
+
+                    if ((pair.Key == JsonConstants.ODataErrorInnerErrorMessageName ||
+                        pair.Key == JsonConstants.ODataErrorInnerErrorStackTraceName ||
+                        pair.Key == JsonConstants.ODataErrorInnerErrorTypeNameName) &&
+                        pair.Value is ODataNullValue)
```
mikepizzo: >pair.Value is ODataNullValue [](start = 24, length = 28)

Maybe check for null value first? That way you don't have to check the other cases if the value is non-null.

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -213,31 +213,39 @@ internal static void EndJsonPaddingIfRequired(IJsonWriter jsonWriter, ODataMessa
...
+                    {
+                        // Write empty string for null values in stacktrace, type and message properties of inner error. 
+                        jsonWriter.WriteJsonValue(string.Empty);
```
mikepizzo: >jsonWriter.WriteJsonValue(string.Empty); [](start = 24, length = 40)

Could you instead call WriteODataValue(new ODataStringValue(string.Empty)), and leave WriteJsonValue private?

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -213,31 +213,39 @@ internal static void EndJsonPaddingIfRequired(IJsonWriter jsonWriter, ODataMessa
         /// <param name="innerErrorPropertyName">The property name for the inner error property.</param>
         /// <param name="recursionDepth">The number of times this method has been called recursively.</param>
         /// <param name="maxInnerErrorDepth">The maximum number of nested inner errors to allow.</param>
+        /// <param name="skipNullProperties">if set, stacktrace, type, and message will not be written if they are empty strings on ODataInnerError.</param>
+
         private static void WriteInnerError(IJsonWriter jsonWriter, ODataInnerError innerError, string innerErrorPropertyName, int recursionDepth, int maxInnerErrorDepth)
```
mikepizzo: You no longer have this added param, so you can remove this line.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataJsonWriterUtilsTests.cs
```diff
@@ -91,5 +92,83 @@ public void WriteError_WritesTargetAndDetails()
...
+            ODataJsonWriterUtils.WriteError(
+                jsonWriter,
+                enumerable => { },
+                error,
+                includeDebugInformation: true,
+                maxInnerErrorDepth: 5,
+                writingJsonLight: false);
+             var result = stringWriter.GetStringBuilder().ToString();
+             //result.Should().Be("{\"error\":{\"code\":\"\",\"message\":\"\",\"target\":\"any target\",\"details\":[{\"code\":\"500\",\"target\":\"any target\",\"message\":\"any msg\"}],\"innererror\":{\"stacktrace\":\"NormalString\",\"MyNewObject\":{\"NestedResourcePropertyName\":\"NestedPropertyValue\"},\"internalexception\":{\"nested\":null}}}}}");
+             result.Should().Be("{\"error\":{\"code\":\"\",\"message\":\"\",\"target\":\"any target\",\"details\":[{\"code\":\"500\",\"target\":\"any target\",\"message\":\"any msg\"}],\"innererror\":{\"stacktrace\":\"NormalString\",\"MyNewObject\":{\"NestedResourcePropertyName\":\"NestedPropertyValue\"},\"internalexception\":{\"nested\":null}}}}");
```
mikepizzo: remove this line?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightErrorDeserializerTests.cs
```diff
@@ -114,6 +116,70 @@ public void ReadTopLevelErrorWithCollectionResourceValueInstanceAnnotation_Works
...
+            const string payload =
+                    "{\"error\":{" +
+                        "\"code\":\"\",\"message\":\"\"," +
+                        "\"target\":\"any target\"," +
+                        "\"details\":[{\"code\":\"500\",\"target\":\"any target\",\"message\":\"any msg\"}]," +
+                        "\"innererror\":{" +
+                            "\"stacktrace\":\"NormalString\"," +
+                            "\"message\":\"\"," +
+                            "\"type\":\"\"," +
+                            "\"innererror\":{" +
```
mikepizzo: Maybe add an inner nested object?

src/Microsoft.OData.Core/ODataInnerError.cs
mikepizzo: Then do we really need the constructor that takes an IDictionary<string,ODataValue>, or can the error handling library just use this method and then add the properties directly to the ODataInnerError.Properties dictionary?

---
In reply to: [321869037](https://github.com/OData/odata.net/pull/1523#discussion_r321869037) [](ancestors = 321869037,321857649)

src/Microsoft.OData.Core/ODataInnerError.cs
mikepizzo: See below. I don't think we need both this and the method that creates the empty ODataInnerError. Also, I'm not sure that the expected semantics would be that we would hold a reference to the dictionary, versus copy from it - does the user expect that we'll muck with their collection (i.e., if the user sets the Message property) and do they expect that if they continue to modify the dictionary that will be reflected in the inner error object?  Consider that, if we don't support ODataCollectionValue properties, we'll want to wrap the setter, which doesn't do us any good if they have direct access to the dictionary.

---
In reply to: [321868334](https://github.com/OData/odata.net/pull/1523#discussion_r321868334) [](ancestors = 321868334,321856768)

mikepizzo:
>         internal string ToJson()

Why would this be a breaking change for ODataError.ToString()? Wouldn't it only write out something different if the user added stuff to the Properties collection?

---
In reply to: [528042140](https://github.com/OData/odata.net/pull/1523#issuecomment-528042140) [](ancestors = 528042140,527996900)

---
Refers to: src/Microsoft.OData.Core/ODataInnerError.cs:122 in 4aa33e8. [](commit_id = 4aa33e8fde27f62038c678ca4f048ed9b4c680e3, deletion_comment = False)

src/Microsoft.OData.Core/Json/ODataJsonReaderCoreUtils.cs
```diff
@@ -130,7 +130,7 @@ internal static class ODataJsonReaderCoreUtils
         /// <param name="inputContext">The input context with all the settings.</param>
         /// <param name="recursionDepth">The recursion depth to start with.</param>
         /// <returns>an instance of IDictionary containing the spatial value.</returns>
-        private static IDictionary<string, object> ReadObjectValue(IJsonReader jsonReader, bool insideJsonObjectValue, ODataInputContext inputContext, int recursionDepth)
```
mikepizzo: >private [](start = 8, length = 7)

Do you still need to make this internal, or can it be private?

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
mikepizzo: Only problem is, the new code could change the order that the properties are written in the JSON payload, which may be considered a breaking change.  Perhaps we should still write the "known" properties first, and then write the remaining properties.

---
In reply to: [315445762](https://github.com/OData/odata.net/pull/1523#discussion_r315445762) [](ancestors = 315445762)

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -33,38 +41,71 @@ public ODataInnerError(Exception exception)
...
+            ExceptionUtils.CheckArgumentNotNull(properties, "properties");
+
+            Properties = properties;
+        }
+
+        /// <summary>
+        /// Creates an instance of ODataInnerError with an empty property bag.
+        /// </summary>
+        /// <returns>returns an instance of ODataInnerError with empty property bag.</returns>
+        public static ODataInnerError CreateEmptyInnerError()
```
mikepizzo: >CreateEmptyInnerError [](start = 38, length = 21)

Also note that, by spec, the ODataError must have non-empty "Message" and "Code" strings.

src/Microsoft.OData.Core/ODataInnerError.cs
KanishManuja-MS: I have fixed the part where top level dictionary does not get updated. However, if the dictionary contained references, the copy will reflect changes to those objects though. 

---
In reply to: [321906011](https://github.com/OData/odata.net/pull/1523#discussion_r321906011) [](ancestors = 321906011,321868334,321856768)

src/Microsoft.OData.Core/ODataInnerError.cs
KanishManuja-MS: I am modifying ODataInnerError here which does not have such constraints. For ODataError, we currently do not enforce such constraints in the implementation and we cannot even do it without having a breaking change. 

---
In reply to: [321910714](https://github.com/OData/odata.net/pull/1523#discussion_r321910714) [](ancestors = 321910714)

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightErrorDeserializerTests.cs
KanishManuja-MS: The MyNewObject is a nested Object. Do you want me to rename it or add another level? 

---
In reply to: [321903391](https://github.com/OData/odata.net/pull/1523#discussion_r321903391) [](ancestors = 321903391)

mikepizzo: I meant an object nested in MyNewObject.

---
In reply to: [321944035](https://github.com/OData/odata.net/pull/1523#discussion_r321944035) [](ancestors = 321944035,321903391)

src/Microsoft.OData.Core/Microsoft.OData.Core.txt
```diff
@@ -69,6 +69,7 @@ PropertyAnnotationAfterTheProperty=An annotation with name '{0}' for property '{
 AtomValueUtils_CannotConvertValueToAtomPrimitive=Cannot convert a value of type '{0}' to the string representation of an Atom primitive value.
 
 ODataJsonWriter_UnsupportedValueType=The value of type '{0}' is not supported and cannot be converted to a JSON representation.
+ODataJsonWriter_UnsupportedValueInCollection=Unable to serialize an object in a collection as it is not an ODataPrimitve or ODataResourceValue.
```
mikepizzo: >ODataPrimitve or ODataResourceValue [](start = 107, length = 35)

Or, now, ODataCollectionValue, right? Maybe just say ODataValue?

src/Microsoft.OData.Core/Json/JsonWriterExtensions.cs
```diff
@@ -8,13 +8,12 @@ namespace Microsoft.OData.Json
 {
     #region Namespaces
     using System;
-    using System.Diagnostics;
-    using System.Collections.Generic;
     using System.Collections;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using ODataErrorStrings = Microsoft.OData.Strings;
     using Microsoft.OData.Edm;
```
mikepizzo: nit: should be alphabetical

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -248,5 +255,89 @@ private static void WriteInnerError(IJsonWriter jsonWriter, ODataInnerError inne
...
+                }
+
+                ODataValue odataValue = item as ODataValue;
+                if (odataValue != null)
+                {
+                    ODataValueToString(sb, odataValue);
+                }
+                else
+                {
+                    throw new ODataException("");
```
mikepizzo: Should define an exception to use here

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -248,5 +256,89 @@ private static void WriteInnerError(IJsonWriter jsonWriter, ODataInnerError inne
...
+                {
+                    sb.Append(",");
+                }
+
+                ODataValue odataValue = item as ODataValue;
+                if (odataValue != null)
+                {
+                    ODataValueToString(sb, odataValue);
+                }
+                else
```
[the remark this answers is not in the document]
KanishManuja-MS: Things start getting complicated soon. Like for collection of guids, how do you read those? I think it is fine to only allow for ODataValues. Since it is not going to be very frequently used, I think this is fine. 

---
In reply to: [323952068](https://github.com/OData/odata.net/pull/1523#discussion_r323952068) [](ancestors = 323952068)

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -248,5 +256,89 @@ private static void WriteInnerError(IJsonWriter jsonWriter, ODataInnerError inne
             // }
             jsonWriter.EndObjectScope();
         }
+
+        internal static void ODataValueToString(StringBuilder sb, ODataValue value)
```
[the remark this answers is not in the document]
KanishManuja-MS: I think this static class is wrongly named because of the methods contained are only related to ODataError and writing the json for it. The other place is ODataInnerError itself which does not make sense to me. 


---
In reply to: [323952586](https://github.com/OData/odata.net/pull/1523#discussion_r323952586) [](ancestors = 323952586)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightErrorDeserializer.cs
```diff
@@ -333,8 +333,15 @@ private void ReadPropertyValueInInnerError(int recursionDepth, ODataInnerError i
...
                 default:
-                    // skip any unsupported properties in the inner error
-                    this.JsonReader.SkipValue();
+                    if (!innerError.Properties.ContainsKey(propertyName))
+                    {
+                        innerError.Properties.Add(propertyName, this.JsonReader.ReadODataValue());
+                    }
+                    else
+                    {
+                        innerError.Properties[propertyName] = this.JsonReader.ReadODataValue();
```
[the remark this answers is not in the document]
KanishManuja-MS: Yes. This may be undesired but to keep the behavior same, this was the only option. For instance, a json payload with duplicate properties would not throw error prior to my change. I don't want to throw an error now. 

---
In reply to: [323950924](https://github.com/OData/odata.net/pull/1523#discussion_r323950924) [](ancestors = 323950924)

xuzhg:
Merged at: https://github.com/OData/odata.net/commit/ed08ec4e39d42f75af400226410962767afb453e

PR #1532, opened 2019-09-03
Enable  $select with nested query option semantic binder

xuzhg:
### Issues

* Enable semantic binder the $select with nested query options

### Description

Main changes:

1. Refact the PathTokenRevers
2. Remove the Unifier
3. Enable $select, $filter, $orderby, $top, $skip, $count, $compute nested in $select
4. Remove SelectBinder.cs, use the SelectExpandBinder for both $expand and $select
5. Refactor/modify/add the test cases

- [x] *Test cases added*

### Additional work necessary


KanishManuja-MS:
>     {

Why not mark the class as static as well? 

---
Refers to: src/Microsoft.OData.Core/UriParser/Binders/ExpandTreeNormalizer.cs:17 in e3e6517. [](commit_id = e3e6517d326aa811985ebbebc5a14350cd25b630, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -103,78 +104,109 @@ private ODataUriParserConfiguration Configuration
...
+
+                foreach (SelectTermToken selectTermToken in selectToken.SelectTerms)
+                {
+                    AddToSelectedItems(GenerateSelectItem(selectTermToken), selectExpandItems);
+                }
             }
 
-            return topLevelExpand;
+            // It's better to return "null" if both expand and select are null.
+            // However, in order to be consitent, we returns empty "SelectExpandClause" with AllSelected = true.
```
KanishManuja-MS: >consitent [](start = 39, length = 9)

typo.

src/Microsoft.OData.Client/ALinq/AddNewEndingTokenVisitor.cs
```diff
@@ -47,7 +47,7 @@ public void Visit(NonSystemToken tokenIn)
             {
                 if (newTokenToAdd != null)
                 {
-                    tokenIn.SetNextToken(newTokenToAdd);
```
mikepizzo: Why this change? Are you planning to remove the internal SetNextToken in favor of simply setting the public property?

xuzhg:
>     {

Yes

---
In reply to: [527681564](https://github.com/OData/odata.net/pull/1532#issuecomment-527681564) [](ancestors = 527681564)

---
Refers to: src/Microsoft.OData.Core/UriParser/Binders/ExpandTreeNormalizer.cs:17 in e3e6517. [](commit_id = e3e6517d326aa811985ebbebc5a14350cd25b630, deletion_comment = False)

xuzhg:
>     {

Because it's sealed. A class can't be both static and sealed.

---
In reply to: [527681564](https://github.com/OData/odata.net/pull/1532#issuecomment-527681564) [](ancestors = 527681564)

---
Refers to: src/Microsoft.OData.Core/UriParser/Binders/ExpandTreeNormalizer.cs:17 in e3e6517. [](commit_id = e3e6517d326aa811985ebbebc5a14350cd25b630, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -423,30 +710,137 @@ private static LevelsClause ParseLevels(long? levelsOption, IEdmType sourceType,
...
+
+            // if the selected item is "*", filter the existing property selection.
+            if (itemToAdd is WildcardSelectItem)
+            {
+                var shouldFilter = selectItems.Where(s => IsStructuralSelectionItem(s)).ToList();
+                foreach (var filterItem in shouldFilter)
+                {
+                    selectItems.Remove(filterItem);
+                }
+            }
```
mikepizzo: Alternatively, iterate through the selectItems backward (by index) and remove any items for which IsStructuralSelectionItem is true:

            // iterate the list back to front, removing structural selection items
            for (int i = selectItems.Count -1; i >= 0; i--)
            {
                if(IsStructuralSelectionItem(selectItems[i]))
                {
                    selectItems.RemoveAt(i);
                }
            }

This avoids the LINQ query, the extra list, and having to go back and look up each item by value to remove it.  Probably not a big perf issue for this case, but avoiding LINQ is good practice in general.

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -236,85 +268,340 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
             }
 
             // Add the segments in select and expand to parsed segments
-            this.parsedSegments.AddRange(pathSoFar);
+            List<ODataPathSegment> parsedPath = new List<ODataPathSegment>(this.parsedSegments);
+            parsedPath = parsedPath.Concat(pathSoFar).ToList();
```
mikepizzo: Cheaper: 
    List<ODataPathSegment> parsedPath = new List<ODataPathSegment>(this.parsedSegments.Concat(pathSoFar));

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -103,78 +104,109 @@ private ODataUriParserConfiguration Configuration
...
 
-            // the top level represents $it then has only select populated..
-            List<SelectItem> expandedTerms = new List<SelectItem>();
-            if (tokenIn.ExpandTerms.Single().ExpandOption != null)
+            // $expand=
+            if (expandToken != null && expandToken.ExpandTerms.Any())
             {
-                expandedTerms.AddRange(tokenIn.ExpandTerms.Single().ExpandOption.ExpandTerms.Select(this.GenerateExpandItem).Where(expandedNavigationSelectItem => expandedNavigationSelectItem != null));
+                selectExpandItems.AddRange(expandToken.ExpandTerms.Select(this.GenerateExpandItem).Where(s => s != null));
             }
```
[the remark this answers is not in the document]
mikepizzo: Yep; just looking for opportunities to remove LINQ as we encounter it.

---
In reply to: [323993918](https://github.com/OData/odata.net/pull/1532#discussion_r323993918) [](ancestors = 323993918,323948212)

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -236,85 +268,340 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
...
+        /// <param name="orderByToken">The orderby token to visit.</param>
+        /// <returns>The null or the built orderby clause.</returns>
+        private OrderByClause BindOrderby(IEnumerable<OrderByToken> orderByToken, IEdmNavigationSource navigationSource,
+            IEdmTypeReference elementType, HashSet<EndPathToken> generatedProperties, bool collapsed = false)
+        {
+            if (orderByToken != null && orderByToken.Any())
             {
-                MetadataBinder binder = this.BuildNewMetadataBinder(targetNavigationSource, generatedProperties, collapsed);
+                MetadataBinder binder = BuildNewMetadataBinder(this.Configuration, navigationSource, elementType, generatedProperties, collapsed);
                 OrderByBinder orderByBinder = new OrderByBinder(binder.Bind);
```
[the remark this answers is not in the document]
mikepizzo: MetadataBinder has BindingState which would have to be reset (binder.BindingState = this.Configuration) in order to reuse the MetadataBinder. That seems to be the only state, but probably fine not to try and optimize this at this point.

---
In reply to: [324002063](https://github.com/OData/odata.net/pull/1532#discussion_r324002063) [](ancestors = 324002063,323952324)

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -236,85 +268,340 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
...
+            Debug.Assert(tokenIn != null, "tokenIn != null");
 
-            LevelsClause levelsOption = ParseLevels(tokenIn.LevelsOption, currentLevelEntityType, currentNavProp);
-            return new ExpandedNavigationSelectItem(pathToNavProp, targetNavigationSource, subSelectExpand, filterOption, orderbyOption, tokenIn.TopOption, tokenIn.SkipOption, tokenIn.CountQueryOption, searchOption, levelsOption, computeOption, applyOption);
+            List<ODataPathSegment> pathSoFar = new List<ODataPathSegment>();
+            IEdmStructuredType currentLevelType = this.edmType;
+
+            // first, walk through all type segments in a row, converting them from tokens into segments.
+            if (tokenIn.IsNamespaceOrContainerQualified() && !UriParserHelper.IsAnnotation(tokenIn.Identifier))
+            {
```
mikepizzo: Don't know that it would be worth it for this case, but you could avoid allocating the empty list if you just defined the variable on 474, then pathSoFar = new List<ODataPathSegment>(SelectExpandPathBinder....) in 481, and added an "else {pathSoFar=new List<ODataPathSegment>();" at 489.

mikepizzo:
Added a few general comments around different options for performance, but don't expect them to be significant for these scenarios, so they're more informational/things to thing about rather than things that necessarily need to be fixed to merge this (very well factored and thought through) PR.

xuzhg:
Merged at: https://github.com/OData/odata.net/commit/572082374a496b5855f28a9f66145f577029659c, thanks.

PR #1533, opened 2019-09-03
Fix #1493: Enable the $expand=* after the complex type

### LINKED ISSUE #1493: Fix the complex type with navigation property gap in SelectExpandOptionParser

### Assemblies affected

### Reproduce steps

* OData libraries supports navigation property on complex type. 
* However, the Uri query options has a feature gap to support the complex type with navigation property. See: https://github.com/OData/odata.net/blob/8ee92c44d5855aea785781d7bbd9b514e4c63dc5/src/Microsoft.OData.Core/UriParser/Parsers/ExpandOptionParser.cs#L430-L434

Please investigate and fix it.

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

xuzhg:
### Issues

*This pull request fixes issue #1493.*

### Description

* Enable $expand=* after the complex type.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/SelectExpandParserTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/SelectExpandParserTests.cs
@@ -380,7 +380,7 @@ namespace Microsoft.OData.Tests.UriParser.Parsers
         public void ParseStarInExpand()
         {
             // Arrange & Act
-            ExpandToken expandToken = this.StarExpandTesting("*", "Persons");
+            ExpandToken expandToken = this.StarExpandTesting("*", "Person");
 
             // Assert
             ExpandTermToken innerExpandTerm = Assert.Single(expandToken.ExpandTerms);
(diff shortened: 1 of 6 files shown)


xuzhg:
@mikepizzo @KanishManuja-MS The build was passed in https://identitydivision.visualstudio.com/OData/_build/results?buildId=393967

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/SelectExpandParserTests.cs
```diff
@@ -380,19 +380,55 @@ private SelectToken ParseSelectClause(string select)
...
+            ExpandToken expandToken = this.StarExpandTesting("*", "Person");
 
             // Assert
             ExpandTermToken innerExpandTerm = Assert.Single(expandToken.ExpandTerms);
             innerExpandTerm.PathToNavigationProp.ShouldBeNonSystemToken("Friend");
             Assert.Null(innerExpandTerm.PathToNavigationProp.NextToken);
         }
 
+        [Fact]
+        public void ParseStarInExpandWithComplexTypeAsParent()
```
mikepizzo: >ParseStarInExpandWithComplexTypeAsParent [](start = 20, length = 40)

Do we have any tests for explicit nav props on complex types (as opposed to star)?;  i.e., do we have tests for: $expand=Address/CityNav1 ?

PR #1541, opened 2019-09-17
Copy over metadata selector when creating a copy of messagewritersetting

KanishManuja-MS:
### Issues

This pull request fixes metadata selector feature if the message writer is used to create an odata writer. 

### Description

The copy method for the message writer settings was not copying over the metadata selector property.


mikepizzo:
Change looks good; can we add a test to ODataMessageWriterSettingsTests? (see other copy constructor tests in that file).

Even better would be to write a test that used reflection to ensure that all public properties were copied over by Clone() in both ODataMessageWriterSettings and ODataMessageReaderSettings (but that can be a separate work item)

PR #1564, opened 2019-10-08
Update HttpWebRequestMessage.cs to enable TLS 1.2

KanishManuja-MS:
This fixes the build failure that was happening for package testing for OData.


test/PackageTests/Microsoft.OData.Tests.Nuget.Core/HttpWebRequestMessage.cs
```diff
@@ -21,6 +21,7 @@ public class HttpWebRequestMessage : IODataRequestMessageAsync
 
         public HttpWebRequestMessage(Uri uri)
         {
+            ServicePointManager.SecurityProtocol = System.Net.SecurityProtocolType.Tls12;
```
xuzhg: Would you please leave a simple comments in the code about this change?

PR #1577, opened 2019-10-23
Specifying culture agnostic formatting for some integer string conver…

grumpydev-net:
…sions

### Issues

There is no issue logged for this. In my configuration this warning is being treated as error and breaking the build with the following error:

`CA130: Because the behavior of 'int.ToString()' could vary based on the current user's locale settings, replace this call in 'UriWriter.VisitOperationInvocation(QueryableResourceExpression)' with a call to 'int.ToString(IFormatProvider)'. If the result of 'int.ToString(IFormatProvider)' will be displayed to the user, specify 'CultureInfo.CurrentCulture' as the 'IFormatProvider' parameter. Otherwise, if the result will be stored and accessed by software, such as when it is persisted to disk or to a database, specify 'CultureInfo.InvariantCulture'.`

### Description

Fixed the warning by adding `ToString(CultureInfo.InvariantCulture)__`

- [ ] *Test cases added* __N/A, build error, not feature.__
- [x] *Build and test with one-click build and test script passed*: 

### Additional work necessary
N/A

### DIFF
--- a/Build.props
+++ b/Build.props
@@ -78,9 +78,9 @@
     <PackagesConfig>$(NuGetToolsPath)\packages.config</PackagesConfig>
   </PropertyGroup>
 
-  <Target Name="RestoreNuGetPackages" BeforeTargets="$(GenerateTextStringResourcesDependsOn)">
+  <!--<Target Name="RestoreNuGetPackages" BeforeTargets="$(GenerateTextStringResourcesDependsOn)">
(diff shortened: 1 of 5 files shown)


bdebaere:
@grumpydev-net @mikepizzo In what way is what this PR is trying to accomplish different from https://github.com/OData/odata.net/pull/1556?

grumpydev-net:
> Changes to .cs files look fine. Can you remove the changes to buildprops, OData.Client.csproj, and packages.config from the PR? -- Thanks!

Yeah, it addresses the same issue as #1556 and IDK why those changes to the packages.config where added to this PR, those were pushed to my own fork. Could that behavior be related to #1579 ??

Issue #1250, opened 2018-09-11
Add Interfaces to to OData Client to support mocking in unit tests


rifuller:
I have a function that has a dependency on the OData client to perform its task. I inject the OData client (and other dependencies) into the function so that I can unit test the behaviour of the function.  Classes in Microsoft.OData.Client do not implemented an interface making it impossible to pass a fake into the function in the context of unit tests.

### Assemblies affected

Microsoft.OData.Client

### Reproduce steps

1. Generate an OData client using: https://outlook.office.com/api/v1.0/$metadata
1. View the generated client code

### Expected result

The generated class implements an interface that allows its methods/properties to be mocked.

### Actual result

The generated class

    public partial class EntityContainer : global::Microsoft.OData.Client.DataServiceContext

does not extend an interface that would allow its implemented to be mocked.

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

kornakar:
If you have access to the Visual Studio Enterprise, you could use MS Fakes for mocking the client code.
I had a question about it a while back so it may help you set up the fakes if you need to: [https://github.com/OData/odata.net/issues/1167](https://github.com/OData/odata.net/issues/1167)

rifuller:
> If you have access to the Visual Studio Enterprise, you could use MS Fakes for mocking the client code.
> I had a question about it a while back so it may help you set up the fakes if you need to: #1167

Thanks for the suggestion.  I have a strong preference towards NSubstitute though as it works well with the dependency injection pattern.  MS Fakes, though my understanding is limited, takes a different approach than DI.

rifuller:
> @rifuller This is a valuable suggestion but not in our radar. If you can contribute to it, we would love to take a look.

I can possibly take a look at this in October.  Do you accept pull requests on github for the project?  If not, please point me to the DevOps repo.

chuanboz:
I created another proposal in #1271 , that instead of adding interface for OData Client, I'd propose to add IODataClientFactory similar to IHttpClientFactory, to provide extensibility and dependency injection support, and that will also support UT mocking by injecting a different implementation of IODataClientFactory.

See also a similar long discussion for IHttpClientFactory vs. IHttpClient https://github.com/dotnet/corefx/issues/1624

madansr7:
@rifuller , we are identifying work for our next release of OData client, if you would like to make a contribution for this issue we would be happy to review.

rschiefer:
[...]

PR #1578, opened 2019-10-23
added working example of adding interface to DataServiceQuery with ex…

rschiefer:
…ample tests

### Issues

#1250

### Description

This is a simplified example of how you might add a IDataServiceQuery interface.  There would definitely be more code to add but wanted to get some feedback before I went further.  I suspect we might want to provide OData client consumers with a good portion of the test helper code.

- [v ] *Test cases added*
- [ v] *Build and test in VS and test script passed*

### Additional work necessary


corranrogue9 (approved):

rschiefer:
@copilot  perform a review of this pull request.

PR #1583, opened 2019-10-30
Allow duplicate properties in $select with no select options

mikepizzo:
### Issues

This PR allows multiple select paths to the same property if there are no query options

### Description

When we implemented select options, we disallowed multiple select paths in the select list to the same property. This was a breaking change. This change allows multiple select paths to the same property as long as none of the instances of that property contain any select options.

- [x] *Test cases added*

### Additional work necessary

Four added tests are disabled due to a known issue--selecting properties of a complex type include that complex type in the select list.  Once that issue is resolved, the two disabled tests can be re-enabled.

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
@@ -12,7 +12,7 @@ using Microsoft.OData.Metadata;
 using Microsoft.OData.Edm;
 using Microsoft.OData.UriParser.Aggregation;
 using ODataErrorStrings = Microsoft.OData.Strings;
-
+using System.Text;
 
 namespace Microsoft.OData.UriParser
 {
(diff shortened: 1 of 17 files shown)


src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -12,7 +12,7 @@
 using Microsoft.OData.Edm;
 using Microsoft.OData.UriParser.Aggregation;
 using ODataErrorStrings = Microsoft.OData.Strings;
-
+using System.Text;
```
xuzhg: move it ahead?

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -125,10 +125,48 @@ public SelectExpandClause Bind(ExpandToken expandToken, SelectToken selectToken)
...
+
+                        // if the new one has options, raise an error
+                        if (selectPathItem != null && selectPathItem.HasOptions)
+                        {
+                            throw new ODataException(ODataErrorStrings.SelectTreeNormalizer_MultipleSelecTermWithSamePathFound(ToPathString(selectTermToken.PathToProperty)));
+                        }
+                        else
+                        {
+                            // find the previous item
+                            PathSelectItem previousItem = selectExpandItems.Cast<PathSelectItem>().FirstOrDefault(selectExpandItem =>
```
xuzhg: maybe it's better to use `OfType<T>() `, not `cast<T>()`. `OfType<PathSelectItem().FirstOrDefault(...)`

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -125,10 +125,48 @@ public SelectExpandClause Bind(ExpandToken expandToken, SelectToken selectToken)
...
                 isAllSelected = false;
+                HashSet<PathSegmentToken> selectedPaths = new HashSet<PathSegmentToken>(new PathSegmentTokenEqualityComparer());
 
                 foreach (SelectTermToken selectTermToken in selectToken.SelectTerms)
                 {
-                    AddToSelectedItems(GenerateSelectItem(selectTermToken), selectExpandItems);
+                    SelectItem selectItem = GenerateSelectItem(selectTermToken);
+                    if (selectedPaths.Contains(selectTermToken.PathToProperty))
+                    {
+                        // It's not allowed have multple select clause with same path.
```
xuzhg: please help resolve the typo "multiple", :)

src/Microsoft.OData.Core/UriParser/SemanticAst/PathSelectItem.cs
```diff
@@ -133,5 +133,24 @@ public override void HandleWith(SelectItemHandler handler)
         {
             handler.Handle(this);
         }
+
+        /// <summary>
+        /// Returns whether or not the path select item has any options applied
+        /// </summary>
+        internal bool HasOptions
```
xuzhg: discuss:   consider to public it? It seems useful and better to add a property into the SelectItem (base class). Anyway, i can do it later.

src/Microsoft.Spatial/Microsoft.Spatial.csproj
```diff
@@ -163,6 +163,7 @@
       <DependentUpon>Microsoft.Spatial.tt</DependentUpon>
       <ExcludeFromStyleCop>true</ExcludeFromStyleCop>
     </Compile>
+    <None Include="packages.config" />
```
xuzhg: do you change anything for Microsoft.Spatial project? I think no, right? please revert them

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/ExpandAndSelectPathExtractingTests.cs
```diff
@@ -168,12 +176,107 @@ public void SelectPrimitiveAndNavigationPropertyThatIsAlsoExpanded()
...
+
+        [Fact]
+        public void SelectDuplicatePropertyWithOptionsFails()
+        {
+            const string selectClauseText = "Address($select=City),Address($select=Region)";
+            System.Action test = () => this.ParseAndExtract(selectClauseText: selectClauseText, expandClauseText: null, expectedSelectClauseFromOM: null, expectedExpandClauseFromOM: null);
+            test.Throws<ODataException>("Found multiple select terms with same select path 'Address' at one $select, please combine them together.");
+        }
+
+        [Fact(Skip = "Currently succeeds with extra top-level 'Address' node due to known issue.")]
```
xuzhg: do we fix it in the coming release?

xuzhg:
Looks good to me, but leave some comments.

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
mikepizzo: I was ignoring instances that weren't PathSelectItem in my filter. Not sure which is more efficient, but using cast is certainly simpler and may avoid an extra cast.

---
In reply to: [340909397](https://github.com/OData/odata.net/pull/1583#discussion_r340909397) [](ancestors = 340909397)

src/Microsoft.Spatial/Microsoft.Spatial.csproj
mikepizzo: I had to update packages to get it to build, but if others aren't having the problem I'll revert.

---
In reply to: [340910638](https://github.com/OData/odata.net/pull/1583#discussion_r340910638) [](ancestors = 340910638)

PR #1586, opened 2019-10-31
Fixes for Csdl.GetNamespaceAlias extension method.

syprieur:
### Issues

This pull request fixes issue #1587 .

### Description

- GetNamespace now returns null if the namespace doesn't have an alias, as advertised in the documentation (the actual behavior was to throw a KeyNotFoundException instead, as it was calling VersioningDictionary.Get).
- GetNamespace now returns null if the model doesn't define any alias (the actual behavior was to throw NullReferenceException, as GetAnnotationValue returned null and it wasn't checked).

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Edm.Tests/ExtensionMethods/ExtensionMethodTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Edm.Tests/ExtensionMethods/ExtensionMethodTests.cs
@@ -950,6 +950,13 @@ namespace Microsoft.OData.Edm.Tests.ExtensionMethods
             Assert.Null(model.GetNamespaceAlias("SomeNamespace"));
         }
 
+        [Fact]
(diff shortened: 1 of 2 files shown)


test/FunctionalTests/Microsoft.OData.Edm.Tests/ExtensionMethods/ExtensionMethodTests.cs
```diff
@@ -937,6 +937,19 @@ public void FindTypeForUndefinedTypeDoesnotGetIntoInfiniteSearchLoop()
             Assert.Null(unknownType);
         }
 
+        [Fact]
```
xuzhg: Shall we add positive test cases?

syprieur: Done in commit 4.

PR #1590, opened 2019-11-01
Implement int to enum deserialization.

bdebaere:
### Issues

This pull request fixes issue https://github.com/OData/WebApi/issues/1902.

### Description

Reads a PrimitiveValue instead of a StringValue and tries to convert the value to an enum.

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
@@ -1523,8 +1523,25 @@ namespace Microsoft.OData.JsonLight
                 throw new ODataException(ODataErrorStrings.JsonReaderExtensions_UnexpectedNodeDetectedWithPropertyName(JsonNodeType.PrimitiveValue, JsonNodeType.StartObject, propertyName));
             }
 
-            string enumStr = this.JsonReader.ReadStringValue();
-            return new ODataEnumValue(enumStr, expectedValueTypeReference.FullName());
+            object enumValue = this.JsonReader.ReadPrimitiveValue();
+            string enumString;
+            if (enumValue is int enumIntValue)
+            {
+                IEdmEnumType edmEnumType = expectedValueTypeReference.EnumDefinition();
+                IEdmEnumMember selectedMember = edmEnumType.Members.FirstOrDefault(member => member.Value.Value == enumIntValue);
+                enumString = selectedMember.Name;
+            }
+            else if (enumValue is string enumStringValue)
+            {
+                enumString = enumStringValue;
+            }
+            else
+            {
+                throw new ODataException(ODataErrorStrings.ReaderValidationUtils_CannotConvertPrimitiveValue(enumValue, typeof(IEdmEnumTypeReference)));
+            }
+
+
+            return new ODataEnumValue(enumString, expectedValueTypeReference.FullName());
         }
 
         /// <summary>


adzhilyanov:
Hey,

I posted an issue about enum deserialization, where the read value is not checked against the metadata (#1593). I was about to make a PR about it, but i saw this PR, which is touching the same method. It would be easier to merge if the fix is applied as part of this PR, if everybody is ok with it of course. @bdebaere what do you think about adding something like:

```cs
if (selectedMember == null && this.MessageReaderSettings.ThrowIfTypeConflictsWithMetadata)
{
	throw new ODataException(Strings.ValidationUtils_EnumMemberDoesNotExistOnType(enumValue, propertyName, edmEnumType.FullTypeName()));
}
```
and you must always try to get the `selectedMember` and
where ValidationUtils_EnumMemberDoesNotExistOnType is something like:
_"The value '{0}' for property '{1}' does not exist on enum type '{2}'. Make sure to only use enum members that are defined by the type."_

or we can extend ReaderValidationUtils with a similar to https://github.com/OData/odata.net/blob/1e094dcfcabf81d9f668cbc6666945618b1e9023/src/Microsoft.OData.Core/ReaderValidationUtils.cs#L118 method and call it when deserializing the enum value.

Thanks!

bdebaere:
@adzhilyanov I'd be happy to merge, I haven't gotten around to it yet.

odero:
@xuzhg , @KanishManuja-MS could we get your review on this?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1523,8 +1523,25 @@ private object ReadEnumValue(bool insideJsonObjectValue, IEdmEnumTypeReference e
                 throw new ODataException(ODataErrorStrings.JsonReaderExtensions_UnexpectedNodeDetectedWithPropertyName(JsonNodeType.PrimitiveValue, JsonNodeType.StartObject, propertyName));
             }
 
-            string enumStr = this.JsonReader.ReadStringValue();
-            return new ODataEnumValue(enumStr, expectedValueTypeReference.FullName());
+            object enumValue = this.JsonReader.ReadPrimitiveValue();
+            string enumString;
+            if (enumValue is int enumIntValue)
```
xuzhg: >enumValue is int enumIntValue [](start = 16, length = 29)

This new syntax maybe not pass .NET 4.5 and .NET 3.5 compiler. Please change it to normal.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1523,8 +1523,25 @@ private object ReadEnumValue(bool insideJsonObjectValue, IEdmEnumTypeReference e
...
+                enumString = selectedMember.Name;
+            }
+            else if (enumValue is string enumStringValue)
+            {
+                enumString = enumStringValue;
+            }
+            else
+            {
+                throw new ODataException(ODataErrorStrings.ReaderValidationUtils_CannotConvertPrimitiveValue(enumValue, typeof(IEdmEnumTypeReference)));
+            }
```
xuzhg: I'd like to see Unit test to cover the changes.

SenyaMur:
When will be done this issue ?

KenitoInc:
@bdebaere Kindly work on review comments.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1523,8 +1523,25 @@ private object ReadEnumValue(bool insideJsonObjectValue, IEdmEnumTypeReference e
...
 
-            string enumStr = this.JsonReader.ReadStringValue();
-            return new ODataEnumValue(enumStr, expectedValueTypeReference.FullName());
+            object enumValue = this.JsonReader.ReadPrimitiveValue();
+            string enumString;
+            if (enumValue is int enumIntValue)
+            {
+                IEdmEnumType edmEnumType = expectedValueTypeReference.EnumDefinition();
+                IEdmEnumMember selectedMember = edmEnumType.Members.FirstOrDefault(member => member.Value.Value == enumIntValue);
+                enumString = selectedMember.Name;
```
corranrogue9, line 1532: Can't `selectedMember` be `null` here because of `FirstOrDefault` potentially not finding an element?

Issue #1594, opened 2019-11-04
Namespace aliases support in the ODataUriParser


syprieur:
*Ability to use namespace aliases in URIs given to the ODataUriParser.*

### Assemblies affected

*All.*

### Reproduce steps

*Try to give the OData parser an URI:* 
   https://service.com/entitySet('id')/namespaceAlias.functionName
*Instead of:*
   https://service.com/entitySet('id')/fullNamespace.functionName

### Expected result

*The call would work with namespace aliases, as if the full namespace had been given.*

### Actual result

*The ODataUriParser fails to parse the path, as it doesn't recognize the type/segment.*

xuzhg:
Merged, thanks.

PR #1597, opened 2019-11-19
Support the namespace alias for operation in Uri parser

xuzhg:
### Issues

This PR is related to issue #1594. Thanks.

### Description

This PR is to enable the namespace alias for operation in the Uri parser.

- [x] *Test cases added*

### Additional work necessary


mikepizzo:
Looks good; thanks for the quick turn-around!

Issue #1576, opened 2019-10-22
OData.net Client throws InvalidOperationException when calling 'Move' on a DataServiceCollection


davidmilligan:
```
System.InvalidOperationException: Unexpected action 'Move' on the OnCollectionChanged event raised by DataServiceCollection.​
   at Microsoft.OData.Client.BindingObserver.OnDataServiceCollectionChanged(Object collection, NotifyCollectionChangedEventArgs eventArgs)​
   at System.Collections.ObjectModel.ObservableCollection`1.OnCollectionChanged(NotifyCollectionChangedEventArgs e)​
   at System.Collections.ObjectModel.ObservableCollection`1.MoveItem(Int32 oldIndex, Int32 newIndex)​
   at System.Collections.ObjectModel.ObservableCollection`1.Move(Int32 oldIndex, Int32 newIndex)​
```

### Assemblies affected

Microsoft.OData.Client 7.6.1

### Reproduce steps

This issue appears to only be occurring on .net core 3 (we recently migrated from .net 4.7 to .net core 3 and this appears as a regression)

Call the `Move()` method on the DataServiceCollection that is tracking changes (e.g. when attempting to sort the collection for display purposes).

### Expected result

The 'Move' action should just be ignored by the change tracker.

### Actual result

An exception is thrown

### Additional detail

The issue appears to be this swtich statement in https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Client/Binding/BindingObserver.cs line 378:

```
#if !PORTABLELIB
                case NotifyCollectionChangedAction.Move:
                    // Do Nothing. Added for completeness.
                    break;
#endif

                default:
                    throw new InvalidOperationException(Strings.DataBinding_DataServiceCollectionChangedUnknownActionCollection(eventArgs.Action));
```
The conditional compilation appears to be wrong, because there are platforms besides `!PORTABLELIB` (e.g. .net core) that also have the Move operation, but it would be better to simply get rid of all this code. Why would you throw an exception if the action is unknown? This makes this code incredibly brittle to any future changes or differences between platforms for no good reason or real benefit.

xuzhg:
Merged at c3b63c0b81f94aa4bb15d6603fac95e6c2c33474

PR #1601, opened 2019-11-28
Remove exception for unknown action

KenitoInc:
cc @habbes

### Issues

*This pull request fixes issue #1576 .*

### Description
Ignore 'Move' action in NET Core 3.0.

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Client/Binding/BindingObserver.cs
+++ b/src/Microsoft.OData.Client/Binding/BindingObserver.cs
@@ -375,7 +375,7 @@ namespace Microsoft.OData.Client
 
                     break;
 
-#if !PORTABLELIB
+#if (!PORTABLELIB && !NETCOREAPP3_0)
                 case NotifyCollectionChangedAction.Move:
                     // Do Nothing. Added for completeness.
                     break;
@@ -419,7 +419,7 @@ namespace Microsoft.OData.Client
                         this.OnRemoveFromComplexTypeCollection(sender, e.OldItems);
                         this.OnAddToComplexTypeCollection(sender, e.NewItems);
                         break;
-#if !PORTABLELIB
+#if (!PORTABLELIB && !NETCOREAPP3_0)
                     case NotifyCollectionChangedAction.Move:
                         // Do Nothing. Added for completeness.
                         break;
--- a/src/Microsoft.OData.Client/Binding/BindingObserver.cs
+++ b/src/Microsoft.OData.Client/Binding/BindingObserver.cs
@@ -375,14 +375,14 @@ namespace Microsoft.OData.Client
 
                     break;
 
-#if (!PORTABLELIB && !NETCOREAPP3_0)
+#if !PORTABLELIB
                 case NotifyCollectionChangedAction.Move:
                     // Do Nothing. Added for completeness.
                     break;
 #endif
 
-                default:
-                    throw new InvalidOperationException(Strings.DataBinding_DataServiceCollectionChangedUnknownActionCollection(eventArgs.Action));
+                default: // No Action
+                    break;
             }
         }
 
@@ -419,7 +419,7 @@ namespace Microsoft.OData.Client
                         this.OnRemoveFromComplexTypeCollection(sender, e.OldItems);
                         this.OnAddToComplexTypeCollection(sender, e.NewItems);
                         break;
-#if (!PORTABLELIB && !NETCOREAPP3_0)
+#if !PORTABLELIB
                     case NotifyCollectionChangedAction.Move:
                         // Do Nothing. Added for completeness.
                         break;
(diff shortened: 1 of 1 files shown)


habbes:
Hey @KenitoInc 

The exception is also thrown when using .NET core 2.1, I suspect it could be the same for other versions of .NET core as well. So maybe it would be better to whitelist the `NETCOREAPP` symbol instead of `NETCOREAPP3_0`

As a follow up to our discussion and the comments from OP of the issue, here are a couple of questions and concerns that could help us understand the background and assess the different possible solutions (cc @mikepizzo  @KanishManuja-MS @xuzhg)

- What is the rationale of throwing an exception when the action is unknown, instead of just ignoring it, regardless of the library version?
- What does the `PORTABLELIB` symbol signify? Is it intended to be used for just the .NET Portable Class library, or any versions that runs on multiple versions? Does it include .NET core?
- Does the `DataServiceCollection` or `ObservableCollection` that it inherits differ in different versions of .NET (portable, core, framework, etc.) by whether the `Move()` method is implemented or not or by whether the Move action event is triggered? Just trying to understand why the Move action seems to be acknowledge `!PORTABLELIB` and throws exceptions in `PORTABLELIB`. And is it the responsibility of the event handler to throw the exception?

mikepizzo:
I suspect the code throws on an unknown action because such an action *might* indicate a change that we need to react to, so it's safer to throw. But, it does make the code more brittle.

I suspect the original issue that led to the condition compile was that some platform that we supported (silverlight maybe?) didn't define the Move action, so the conditional was required in order to compile against that platform.  

If both the OData.NET35.sln and OData.NET45.sln compile against the code without the conditional, then we can probably get rid of the #if condition all together. Otherwise, rather than whitelisting those platforms that we are aware of today that support Move, we should explicitly exclude only those that do not.

KenitoInc:
At the moment it isn’t easy to identify platforms that support the `Move` action hence we are unable to explicitly exclude them. The best approach for now is to remove the exception for unknown action.

src/Microsoft.OData.Client/Binding/BindingObserver.cs
```diff
@@ -381,8 +381,8 @@ internal void OnDataServiceCollectionChanged(object collection, NotifyCollection
                     break;
 #endif
 
-                default:
-                    throw new InvalidOperationException(Strings.DataBinding_DataServiceCollectionChangedUnknownActionCollection(eventArgs.Action));
+                default: // No Action
```
KanishManuja-MS: Shouldn't we just remove the conditional compile statement if we want to not throw an exception for move? 
This is a broader change than just move.

xuzhg: It seems it is a breaking change. Could we consider it:
1) have a throw exception config
2) If no config or  throw exception config, throw as previous
3) if config to non-exception, no exception throw
?

src/Microsoft.OData.Client/Binding/BindingObserver.cs
```diff
@@ -427,8 +427,8 @@ internal void OnPrimitiveOrComplexCollectionChanged(object sender, NotifyCollect
                     case NotifyCollectionChangedAction.Reset:
                         this.bindingGraph.RemoveCollection(sender);
                         break;
-                    default:
-                        throw new InvalidOperationException(Strings.DataBinding_CollectionChangedUnknownActionCollection(e.Action, sender.GetType()));
+                    default: // No Action
```
KanishManuja-MS: Ditto. Same comment as above.

src/Microsoft.OData.Client/Binding/BindingObserver.cs
KenitoInc: Discussed offline

mikepizzo: I would rather not add config for things like this as they just clutter code and confuse customers. In this case, going from an error state to an ignore state in the unlikely event that .NET extends the enum seems a pretty safe change.

---
In reply to: [355600478](https://github.com/OData/odata.net/pull/1601#discussion_r355600478) [](ancestors = 355600478)

mikepizzo: I believe that, right now, Move is the only enum value that we don't have a conditional for. So the broader change would only be if .NET added new enum values in the future, and the change would be that we would ignore the new values rather than throw an exception. If we had a good way to exclude just Move across frameworks, then I do slightly prefer the default of throwing the exception for unknown values, in case a new enum value represents changes that we should be aware of, but if the (hypothetical) new enum values are correctly defined such that they represent states that are distinct from the existing change values (not returning existing values for existing states would be a breaking change on the .NET side), then we should be able to safely ignore them as well. Given the difficulty in precisely determining which targets do/do not support Move, and the relative unlikelihood of defining a new enum value that breaks existing semantics, I'm okay with the change.

---
In reply to: [355600338](https://github.com/OData/odata.net/pull/1601#discussion_r355600338) [](ancestors = 355600338)

src/Microsoft.OData.Client/Binding/BindingObserver.cs
mikepizzo: As Above, I think this is okay.

---
In reply to: [355600737](https://github.com/OData/odata.net/pull/1601#discussion_r355600737) [](ancestors = 355600737)

xuzhg:
Merged at: c3b63c0b81f94aa4bb15d6603fac95e6c2c33474

Issue #1392, opened 2019-02-01
Performance Improvement for one of the methods in ODataUriParser


KanishManuja-MS:
*Short summary (3-5 sentences) describing the issue.*
![image](https://user-images.githubusercontent.com/41647051/52141442-9ef69780-260a-11e9-96c8-c53a88e07d9d.png)

While LINQ operations provide a concise and elegant way to handle IList, there are overheads at times with using LINQ. Azure profiler guidelines  recommend to avoid LINQ but that is not possible for ODL. However, some of the harding hitting methods like ResolveOperationImportFromList can be simplified to do the same thing by creating a List while doing the same work that the method does.

### Assemblies affected
ODL 7.x

*Which assemblies and versions are known to be affected e.g. OData .Net lib 6.15-beta.*

### Reproduce steps

Open AGS logs for Azure profiler and verify that this function appears in the stack on multiple occasions.

### Expected result
The horizontal length should be shorter for the function call with respect to containing AGS call.
### Actual result
A lot of LINQ operations taking place for this call.
### Additional detail
This issue was discussed in the performance training session.

PR #1602, opened 2019-12-01
Perf improvements in FunctionOverloadResolver

KanishManuja-MS:
### Issues

This pull request fixes issue #1392 

### Description

This PR removes some of the expensive LINQ calls without breaking the public interface. 

- [ ] *Test cases added*

### Results
HasAny previous for enumerables - 0.3372341
HasAny now for enumerables - 0.2167854

### Additional work necessary
I would like to add tests for the changes once the initial work is reviewed. More performance improvement related data should be attached with this PR.

### DIFF
--- a/src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
+++ b/src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
@@ -8,6 +8,7 @@ namespace Microsoft.OData.Metadata
 {
     #region Namespaces
     using System;
+    using System.Collections;
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
@@ -25,6 +26,7 @@ namespace Microsoft.OData.Metadata
 #endif
 #if !ODATA_SERVICE && !ODATA_CLIENT
     using Microsoft.OData.JsonLight;
+    using Microsoft.OData.UriParser;
     using ErrorStrings = Microsoft.OData.Strings;
     using PlatformHelper = Microsoft.OData.PlatformHelper;
 #endif
@@ -252,17 +254,117 @@ namespace Microsoft.OData.Metadata
         }
 
         /// <summary>
-        /// Filters the operations by parameter names.
+        /// Filters the operations by parameter count.
         /// </summary>
-        /// <param name="functions">The operations.</param>
-        /// <param name="parameters">The list of non-binding parameter names to match.</param>
-        /// <param name="caseInsensitive">Whether case insensitive.</param>
+        /// <param name="operations">The operations.</param>
+        /// <param name="parameterCount">The count of non-binding parameter names to match.</param>
         /// <returns>The best matching operations based on parameters.</returns>
-        internal static IEnumerable<IEdmOperation> FindBestOverloadBasedOnParameters(this IEnumerable<IEdmOperation> functions, IEnumerable<string> parameters, bool caseInsensitive = false)
+        internal static IEnumerable<IEdmOperation> FilterOverloadsBasedOnParameterCount(this IEnumerable<IEdmOperation> operations, int parameterCount)
         {
             // The best match out of a list of candidates is the one that has the same number of (non-binding) parameters as specified.
-            IEnumerable<IEdmOperation> exactMatches = functions.Where(f => f.Parameters.Count() == parameters.Count() + (f.IsBound ? 1 : 0));
-            return exactMatches.Count() > 0 ? exactMatches : functions;
+            IEnumerable<IEdmOperation> exactMatches = operations.Where(f => f.Parameters.Count() == parameterCount + (f.IsBound ? 1 : 0));
+            return exactMatches.Count() > 0 ? exactMatches : operations;
+        }
+
+        /// <summary>
+        /// Filters the operations by parameter count.
+        /// </summary>
+        /// <param name="operations">The operations.</param>
+        /// <param name="parameterCount">The count of non-binding parameter names to match.</param>
+        /// <returns>The best matching operations based on parameters.</returns>
+        internal static IList<IEdmOperation> FilterOverloadsBasedOnParameterCount(this IList<IEdmOperation> operations, int parameterCount)
+        {
+            IList<IEdmOperation> exactMatches = null;
+
+            // The best match out of a list of candidates is the one that has the same number of (non-binding) parameters as specified.
+            for (int i = 0; i < operations.Count; i++)
+            {
+                if (operations[i].Parameters.Count() == parameterCount + (operations[i].IsBound ? 1 : 0))
+                {
+                    if (exactMatches == null)
+                    {
+                        exactMatches = new List<IEdmOperation>();
+                    }
+
+                    exactMatches.Add(operations[i]);
+                }
+            }
+
+            return exactMatches == null ? operations : exactMatches;
+        }
+
+        /// <summary>
+        /// Filters the operations candidates based on parameters.
+        /// </summary>
+        /// <param name="operations">The operations.</param>
+        /// <param name="bindingType">Binding type for the operation.</param>
+        /// <param name="parameterNameList">The parameters.</param>
+        /// <param name="caseInsensitive">Whether case insensitive.</param>
+        /// <param name="actionItems">outputs the actions found while matching functions</param>
+        /// <returns>Operations filtered by parameter.</returns>
+        internal static IList<IEdmOperation> FilterOperationCandidatesBasedOnParameterList(this IEnumerable<IEdmOperation> operations, IEdmType bindingType, IList<string> parameterNameList, bool caseInsensitive, out bool actionItemsExists)
+        {
+            Debug.Assert(operations != null, "operations");
+            Debug.Assert(parameterNameList != null, "parameters");
+
+            bool hasParameters = parameterNameList.Count > 0;
+            actionItemsExists = false;
+            IList<IEdmOperation> operationCandidates = new List<IEdmOperation>();
+
+            foreach (var operation in operations)
+            {
+                Debug.Assert(bindingType == null || (bindingType != null && operation.IsBound && operation.Parameters.Any()));
+                bool suitableCandidate = true;
+
+                // If the number of parameters specified in the url is > 0 then we should only keep functions as actions can't have parameters on the uri, only in the payload. Filter further by parameters in this case, otherwise don't.
+                if (hasParameters)
+                {
+                    if (operation.IsAction())
+                    {
+                        actionItemsExists = true;
+                        suitableCandidate = false;
+                    }
+                    else if (!ParametersSatisfyFunction(operation, parameterNameList, caseInsensitive))
+                    {
+                        suitableCandidate = false;
+                    }
+                }
+                else
+                {
+                    // If it is a bound function without required parameters, then the parameter list should only have one parameter and any others as optional.
+                    if (operation.IsFunction())
+                    {
+                        // Binding type requires atleast one parameter for a function. 
+                        bool parameterFound = bindingType != null ? false : true;
+                        foreach (IEdmOperationParameter param in operation.Parameters)
+                        {
+                            // If multiple required parameters found then mark as unsuitable. 
+                            bool isParamOptional = param is IEdmOptionalParameter;
+                            if (parameterFound && !(isParamOptional))
+                            {
+                                suitableCandidate = false;
+                                break;
+                            }
+                            else if (!(isParamOptional))
+                            {
+                                parameterFound = true;
+                            }
+                        }
+
+                        if (!parameterFound)
+                        {
+                            suitableCandidate = false;
+                        }
+                    }
+                }
+
+                if (suitableCandidate)
+                {
+                    operationCandidates.Add(operation);
+                }
+            }
+
+            return operationCandidates;
         }
 
         /// <summary>
@@ -292,22 +394,38 @@ namespace Microsoft.OData.Metadata
         }
 
         /// <summary>
+        /// Filters the operations by parameter names.
+        /// </summary>
+        /// <param name="operations">The operations.</param>
+        /// <param name="parameterNameList">The parameters.</param>
+        /// <param name="caseInsensitive">Whether case insensitive.</param>
+        /// <returns>Operations filtered by parameter.</returns>
+        internal static IEnumerable<IEdmOperation> FilterOperationsByParameterNames(this IList<IEdmOperation> operations, IList<string> parameterNameList, bool caseInsensitive)
+        {
+            Debug.Assert(operations != null, "operations");
+            Debug.Assert(parameterNameList != null, "parameters");
+
+            // TODO: update code that is duplicate between operation and operation import, add more tests.
+            for (int i = 0; i < operations.Count; i++)
+            {
+                if (!ParametersSatisfyFunction(operations[i], parameterNameList, caseInsensitive))
+                {
+                    continue;
+                }
+
+                yield return operations[i];
+            }
+        }
+
+        /// <summary>
         /// Ensures that operations are bound and have a binding parameter, other wise throws an exception.
         /// </summary>
         /// <param name="operations">The operations.</param>
         internal static void EnsureOperationsBoundWithBindingParameter(this IEnumerable<IEdmOperation> operations)
         {
-            foreach (IEdmOperation operation in operations)
+            foreach (var operation in operations)
             {
-                if (!operation.IsBound)
-                {
-                    throw new ODataException(ErrorStrings.EdmLibraryExtensions_UnBoundOperationsFoundFromIEdmModelFindMethodIsInvalid(operation.Name));
-                }
-
-                if (operation.Parameters.FirstOrDefault() == null)
-                {
-                    throw new ODataException(ErrorStrings.EdmLibraryExtensions_NoParameterBoundOperationsFoundFromIEdmModelFindMethodIsInvalid(operation.Name));
-                }
+                operation.EnsureOperationBoundWithBindingParameter();
             }
         }
 
@@ -427,32 +545,6 @@ namespace Microsoft.OData.Metadata
             Debug.Assert(operationImport != null, "operationImport != null");
 
             return operationImport.FullName() + operationImport.ParameterTypesToString();
-        }
-
-        /// <summary>
-        /// Removes the actions.
-        /// </summary>
-        /// <param name="source">The source.</param>
-        /// <param name="actionItems">The action items.</param>
-        /// <returns>Only the functions from the operation sequence.</returns>
-        internal static IEnumerable<IEdmOperation> RemoveActions(this IEnumerable<IEdmOperation> source, out IList<IEdmOperation> actionItems)
-        {
-            List<IEdmOperation> functions = new List<IEdmOperation>();
-
-            actionItems = new List<IEdmOperation>();
-            foreach (var item in source)
-            {
-                if (item.IsAction())
-                {
-                    actionItems.Add(item);
-                }
-                else
-                {
-                    functions.Add(item);
-                }
-            }
-
-            return functions;
         }
 
         /// <summary>
@@ -1826,23 +1918,33 @@ namespace Microsoft.OData.Metadata
                 parametersToMatch = parametersToMatch.Skip(1);
             }
 
-            List<IEdmOperationParameter> functionParameters = parametersToMatch.ToList();
-
             // if any required parameters are missing, don't consider it a match.
-            if (functionParameters.Where(
-                p => !(p is IEdmOptionalParameter)).Any(
-                    p => parameterNameList.All(
-                        k => !string.Equals(k, p.Name, caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))))
+            BitArray matchedUrlParameters = new BitArray(parameterNameList.Count);
+            foreach (var functionParameter in parametersToMatch)
             {
-                return false;
+                bool matched = false;
+                for (int j = 0; j < parameterNameList.Count; j++)
+                {
+                    if (string.Equals(parameterNameList[j], functionParameter.Name, caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
+                    {
+                        matched = true;
+                        matchedUrlParameters[j] = true;
+                    }
+                }
+
+                if (!matched && !(functionParameter is IEdmOptionalParameter))
+                {
+                    return false;
+                }
             }
 
             // if any specified parameters don't match, don't consider it a match.
-            if (parameterNameList.Any(
-                k => functionParameters.All(
-                    p => !string.Equals(k, p.Name, caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))))
+            for (int i = 0; i < parameterNameList.Count; i++)
             {
-                return false;
+                if (!matchedUrlParameters[i])
+                {
+                    return false;
+                }
             }
 
             return true;
@@ -1867,6 +1969,19 @@ namespace Microsoft.OData.Metadata
             }
 
             return inheritanceLevelsFromBase;
+        }
+
+        private static void EnsureOperationBoundWithBindingParameter(this IEdmOperation operation)
+        {
+            if (!operation.IsBound)
+            {
+                throw new ODataException(ErrorStrings.EdmLibraryExtensions_UnBoundOperationsFoundFromIEdmModelFindMethodIsInvalid(operation.Name));
+            }
+
+            if (!operation.Parameters.Any())
+            {
+                throw new ODataException(ErrorStrings.EdmLibraryExtensions_NoParameterBoundOperationsFoundFromIEdmModelFindMethodIsInvalid(operation.Name));
+            }
         }
 #endif
         #endregion
--- a/src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
@@ -186,7 +186,7 @@ namespace Microsoft.OData.UriParser
             // If more than one overload matches, try to select based on optional parameters
             if (possibleFunctions.Count() > 1 && parameterNames.Count > 0)
             {
-                possibleFunctions = possibleFunctions.FindBestOverloadBasedOnParameters(parameterNames);
+                possibleFunctions = possibleFunctions.FilterOverloadsBasedOnParameterCount(parameterNames.Count);
             }
 
             if (!possibleFunctions.HasAny())
--- a/src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
@@ -12,6 +12,7 @@ namespace Microsoft.OData.UriParser
     using Microsoft.OData.Metadata;
     using Microsoft.OData.Edm;
     using ODataErrorStrings = Microsoft.OData.Strings;
+    using System.Diagnostics;
 
     /// <summary>
     /// Helper class to help bind function overloads.
@@ -124,22 +125,22 @@ namespace Microsoft.OData.UriParser
         /// <param name="matchingOperation">The single matching function found.</param>
         /// <param name="resolver">Resolver to be used.</param>
         /// <returns>True if a function was matched, false otherwise. Will throw if the model has illegal operation imports.</returns>
-        internal static bool ResolveOperationFromList(string identifier, IEnumerable<string> parameterNames, IEdmType bindingType, IEdmModel model, out IEdmOperation matchingOperation, ODataUriResolver resolver)
+        internal static bool ResolveOperationFromList(string identifier, IList<string> parameterNames, IEdmType bindingType, IEdmModel model, out IEdmOperation matchingOperation, ODataUriResolver resolver)
         {
             // TODO: update code that is duplicate between operation and operation import, add more tests.
             // If the previous segment is an open type, the service action name is required to be fully qualified or else we always treat it as an open property name.
+            matchingOperation = null;
             if (bindingType != null)
             {
                 // TODO: look up actual container names here?
                 // When using extension, there may be function call with unqualified name. So loose the restriction here.
                 if (bindingType.IsOpen() && !identifier.Contains(".") && resolver.GetType() == typeof(ODataUriResolver))
                 {
-                    matchingOperation = null;
                     return false;
                 }
             }
 
-            IEnumerable<IEdmOperation> candidateMatchingOperations = null;
+            IEnumerable<IEdmOperation> operationsFromModel;
 
             // The extension method FindBoundOperations & FindOperations call IEdmModel.FindDeclaredBoundOperations which can be implemented by anyone and it could throw any type of exception
             // so catching all of them and simply putting it in the inner exception.
@@ -147,11 +148,11 @@ namespace Microsoft.OData.UriParser
             {
                 if (bindingType != null)
                 {
-                    candidateMatchingOperations = resolver.ResolveBoundOperations(model, identifier, bindingType);
+                    operationsFromModel = resolver.ResolveBoundOperations(model, identifier, bindingType);
                 }
                 else
                 {
-                    candidateMatchingOperations = resolver.ResolveUnboundOperations(model, identifier);
+                    operationsFromModel = resolver.ResolveUnboundOperations(model, identifier);
                 }
             }
             catch (Exception exc)
@@ -164,98 +165,109 @@ namespace Microsoft.OData.UriParser
                 throw;
             }
 
-            IList<IEdmOperation> foundActionsWhenLookingForFunctions = new List<IEdmOperation>();
-            bool hasParameters = parameterNames.Count() > 0;
-
-            if (bindingType != null)
-            {
-                candidateMatchingOperations.EnsureOperationsBoundWithBindingParameter();
-            }
-
-            // If the number of parameters > 0 then this has to be a function as actions can't have parameters on the uri only in the payload. Filter further by parameters in this case, otherwise don't.
-            if (hasParameters)
-            {
-                // can only be a function as only functions have parameters on the uri.
-                candidateMatchingOperations = candidateMatchingOperations.RemoveActions(out foundActionsWhenLookingForFunctions)
-                    .FilterOperationsByParameterNames(parameterNames, resolver.EnableCaseInsensitive);
-            }
-            else if (bindingType != null)
-            {
-                // Filter out functions with more than one parameter. Actions should not be filtered as the parameters are in the payload not the uri
-                candidateMatchingOperations = candidateMatchingOperations.Where(o =>
-                (o.IsFunction() && (o.Parameters.Count() == 1 || o.Parameters.Skip(1).All(p => p is IEdmOptionalParameter))) || o.IsAction());
-            }
-            else
-            {
-                // Filter out functions with any parameters
-                candidateMatchingOperations = candidateMatchingOperations.Where(o => (o.IsFunction() && !o.Parameters.Any()) || o.IsAction());
-            }
+            bool foundActionsWhenLookingForFunctions;
+            // Filters candidates based on the parameter names specified in the uri, removes actions if there were parameters specified in the uri but set the out bool to indicate that.
+            // If no parameters specified, then matches based on binding type or matches with operations with no parameters.
+            IList<IEdmOperation> candidatesMatchingOperations = operationsFromModel.FilterOperationCandidatesBasedOnParameterList(bindingType, parameterNames, resolver.EnableCaseInsensitive, out foundActionsWhenLookingForFunctions);
 
             // Only filter if there is more than one and its needed.
-            if (candidateMatchingOperations.Count() > 1)
+            if (candidatesMatchingOperations.Count > 1)
             {
-                candidateMatchingOperations = candidateMatchingOperations.FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType(bindingType);
+                candidatesMatchingOperations = candidatesMatchingOperations.FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType(bindingType) as IList<IEdmOperation>;
+                
+                // This will be only null when no candidates are left. In that case, we can return false here. 
+                if (candidatesMatchingOperations == null)
+                {
+                    if (foundActionsWhenLookingForFunctions)
+                    {
+                        throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_SegmentDoesNotSupportKeyPredicates(identifier));
+                    }
+
+                    return false; 
+                }
             }
 
-            // If any of the things returned are an action, it better be the only thing returned, and there can't be parameters in the URL
-            if (candidateMatchingOperations.Any(f => f.IsAction()))
+            // If any of the candidates are an action, it better be the only thing returned, and there can't be parameters in the URL
+            if (ResolveActionFromCandidates(candidatesMatchingOperations, identifier, parameterNames.Count > 0,  out matchingOperation))
             {
-                if (candidateMatchingOperations.Count() > 1)
-                {
-                    if (candidateMatchingOperations.Any(o => o.IsFunction()))
-                    {
-                        throw new ODataException(ODataErrorStrings.FunctionOverloadResolver_MultipleOperationOverloads(identifier));
-                    }
-                    else
-                    {
-                        throw new ODataException(ODataErrorStrings.FunctionOverloadResolver_MultipleActionOverloads(identifier));
-                    }
-                }
-
-                if (hasParameters)
-                {
-                    throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_SegmentDoesNotSupportKeyPredicates(identifier));
-                }
-
-                matchingOperation = candidateMatchingOperations.Single();
                 return true;
             }
 
-            if (foundActionsWhenLookingForFunctions.Count > 0)
+            if (foundActionsWhenLookingForFunctions)
             {
                 throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_SegmentDoesNotSupportKeyPredicates(identifier));
             }
 
             // If more than one overload matches, try to select based on optional parameters
-            if (candidateMatchingOperations.Count() > 1)
+            if (candidatesMatchingOperations.Count > 1)
             {
-                candidateMatchingOperations = candidateMatchingOperations.FindBestOverloadBasedOnParameters(parameterNames);
+                candidatesMatchingOperations = candidatesMatchingOperations.FilterOverloadsBasedOnParameterCount(parameterNames.Count);
             }
 
-            if (candidateMatchingOperations.Count() > 1)
+            if (candidatesMatchingOperations.Count > 1)
             {
                 throw new ODataException(ODataErrorStrings.FunctionOverloadResolver_NoSingleMatchFound(identifier, string.Join(",", parameterNames.ToArray())));
             }
 
-            matchingOperation = candidateMatchingOperations.SingleOrDefault();
+            matchingOperation = candidatesMatchingOperations.Count > 0 ? candidatesMatchingOperations[0] : null;
             return matchingOperation != null;
         }
 
+        private static bool ResolveActionFromCandidates(IList<IEdmOperation> candidatesMatchingOperations, string identifier, bool hasParameters, out IEdmOperation matchingOperation)
+        {
+            bool actionExists = false;
+            bool functionExists = false;
+            matchingOperation = null;
+            for (int i = 0; i < candidatesMatchingOperations.Count; i++)
+            {
+                if (candidatesMatchingOperations[i].IsFunction())
+                {
+                    if (actionExists)
+                    {
+                        throw new ODataException(ODataErrorStrings.FunctionOverloadResolver_MultipleOperationOverloads(identifier));
+                    }
+
+                    functionExists = true;
+                }
+
+                if (candidatesMatchingOperations[i].IsAction())
+                {
+                    if (functionExists)
+                    {
+                        throw new ODataException(ODataErrorStrings.FunctionOverloadResolver_MultipleOperationOverloads(identifier));
+                    }
+
+                    actionExists = true;
+                }
+            }
+
+            if (actionExists)
+            {
+                if (candidatesMatchingOperations.Count > 1)
+                {
+                    throw new ODataException(ODataErrorStrings.FunctionOverloadResolver_MultipleActionOverloads(identifier));
+                }
+
+                Debug.Assert(!hasParameters);
+                matchingOperation = candidatesMatchingOperations.Count > 0 ? candidatesMatchingOperations[0] : null;
+                return true;
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Returns true if there are elements in the enumerable. Performance is better for pure enumerables.
+        /// Use the other HasAny extension method if the underlying type is likely to be a list. 
+        /// </summary>
         internal static bool HasAny<T>(this IEnumerable<T> enumerable) where T : class
         {
-            IList<T> list = enumerable as IList<T>;
-            if (list != null)
+            if (enumerable != null)
             {
-                return list.Count > 0;
+                return enumerable.GetEnumerator().MoveNext();
             }
 
-            T[] array = enumerable as T[];
-            if (array != null)
-            {
-                return array.Length > 0;
-            }
-
-            return enumerable.FirstOrDefault() != null;
+            return false;
         }
     }
 }
--- a/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
@@ -5,6 +5,7 @@
 //---------------------------------------------------------------------
 
 using System;
+using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
@@ -99,14 +100,23 @@ namespace Microsoft.OData.UriParser
             }
 
             var result = container.Elements.OfType<IEdmNavigationSource>()
-                .Where(source => string.Equals(identifier, source.Name, StringComparison.OrdinalIgnoreCase)).ToList();
+                .Where(source => string.Equals(identifier, source.Name, StringComparison.OrdinalIgnoreCase));
 
-            if (result.Count > 1)
+            IEdmNavigationSource resolvedNavigationSource = null;
+
+            foreach (IEdmNavigationSource candidate in result)
             {
-                throw new ODataException(Strings.UriParserMetadata_MultipleMatchingNavigationSourcesFound(identifier));
+                if (resolvedNavigationSource == null)
+                {
+                    resolvedNavigationSource = candidate;
+                }
+                else
+                {
+                    throw new ODataException(Strings.UriParserMetadata_MultipleMatchingNavigationSourcesFound(identifier));
+                }
             }
 
-            return result.SingleOrDefault();
+            return resolvedNavigationSource;
         }
 
         /// <summary>
@@ -124,15 +134,23 @@ namespace Microsoft.OData.UriParser
             }
 
             var result = type.Properties()
-            .Where(_ => string.Equals(propertyName, _.Name, StringComparison.OrdinalIgnoreCase))
-            .ToList();
+            .Where(_ => string.Equals(propertyName, _.Name, StringComparison.OrdinalIgnoreCase));
 
-            if (result.Count > 1)
+            IEdmProperty resolvedProperty = null;
+
+            foreach(IEdmProperty candidate in result)
             {
-                throw new ODataException(Strings.UriParserMetadata_MultipleMatchingPropertiesFound(propertyName, type.FullTypeName()));
+                if (resolvedProperty == null)
+                {
+                    resolvedProperty = candidate;
+                }
+                else
+                {
+                    throw new ODataException(Strings.UriParserMetadata_MultipleMatchingPropertiesFound(propertyName, type.FullTypeName()));
+                }
             }
 
-            return result.SingleOrDefault();
+            return resolvedProperty;
         }
 
         /// <summary>
@@ -151,12 +169,17 @@ namespace Microsoft.OData.UriParser
 
             IList<IEdmTerm> results = FindAcrossModels<IEdmTerm>(model, termName, /*caseInsensitive*/ true);
 
+            if (results == null || results.Count == 0)
+            {
+                return null;
+            }
+
             if (results.Count > 1)
             {
                 throw new ODataException(Strings.UriParserMetadata_MultipleMatchingTypesFound(termName));
             }
 
-            return results.SingleOrDefault();
+            return results[0];
         }
 
         /// <summary>
@@ -174,12 +197,18 @@ namespace Microsoft.OData.UriParser
             }
 
             IList<IEdmSchemaType> results = FindAcrossModels<IEdmSchemaType>(model, typeName, /*caseInsensitive*/ true);
+
+            if (results == null || results.Count == 0)
+            {
+                return null;
+            }
+
             if (results.Count > 1)
             {
                 throw new ODataException(Strings.UriParserMetadata_MultipleMatchingTypesFound(typeName));
             }
 
-            return results.SingleOrDefault();
+            return results[0];
         }
 
         /// <summary>
@@ -198,10 +227,10 @@ namespace Microsoft.OData.UriParser
             }
 
             IList<IEdmOperation> operations = FindAcrossModels<IEdmOperation>(model, identifier, /*caseInsensitive*/ true);
-            if (operations != null && operations.Count() > 0)
+            if (operations != null && operations.Count > 0)
             {
                 IList<IEdmOperation> matchedOperation = new List<IEdmOperation>();
-                for (int i = 0; i < operations.Count(); i++)
+                for (int i = 0; i < operations.Count; i++)
                 {
                     if (operations[i].HasEquivalentBindingType(bindingType))
                     {
@@ -230,10 +259,10 @@ namespace Microsoft.OData.UriParser
             }
 
             IList<IEdmOperation> operations = FindAcrossModels<IEdmOperation>(model, identifier, /*caseInsensitive*/ true);
-            if (operations != null && operations.Count() > 0)
+            if (operations != null && operations.Count > 0)
             {
                 IList<IEdmOperation> matchedOperation = new List<IEdmOperation>();
-                for (int i = 0; i < operations.Count(); i++)
+                for (int i = 0; i < operations.Count; i++)
                 {
                     if (!operations[i].IsBound)
                     {
@@ -313,22 +342,21 @@ namespace Microsoft.OData.UriParser
         /// <returns>The resolved key list.</returns>
         public virtual IEnumerable<KeyValuePair<string, object>> ResolveKeys(IEdmEntityType type, IList<string> positionalValues, Func<IEdmTypeReference, string, object> convertFunc)
         {
-            var keyProperties = type.Key().ToList();
-
             // Throw an error if key size from url doesn't match that from model.
             // Other derived ODataUriResolver intended for alternative key resolution, such as the built in AlternateKeysODataUriResolver,
             // should override this ResolveKeys method.
-            if (keyProperties.Count != positionalValues.Count)
+            IEnumerable<IEdmStructuralProperty> keys = type.Key();
+            if (keys.Count() != positionalValues.Count)
             {
                 throw ExceptionUtil.CreateBadRequestError(Strings.BadRequest_KeyCountMismatch(type.FullName()));
             }
 
             var keyPairList = new List<KeyValuePair<string, object>>(positionalValues.Count);
 
-            for (int i = 0; i < keyProperties.Count; i++)
+            int i = 0;
+            foreach (IEdmProperty keyProperty in keys)
             {
-                string valueText = positionalValues[i];
-                IEdmProperty keyProperty = keyProperties[i];
+                string valueText = positionalValues[i++];
                 object convertedValue = convertFunc(keyProperty.Type, valueText);
                 if (convertedValue == null)
                 {
@@ -351,17 +379,17 @@ namespace Microsoft.OData.UriParser
         public virtual IEnumerable<KeyValuePair<string, object>> ResolveKeys(IEdmEntityType type, IDictionary<string, string> namedValues, Func<IEdmTypeReference, string, object> convertFunc)
         {
             var convertedPairs = new Dictionary<string, object>(StringComparer.Ordinal);
-            var keyProperties = type.Key().ToList();
 
             // Throw an error if key size from url doesn't match that from model.
             // Other derived ODataUriResolver intended for alternative key resolution, such as the built in AlternateKeysODataUriResolver,
             // should override this ResolveKeys method.
-            if (keyProperties.Count != namedValues.Count)
+            IEnumerable<IEdmStructuralProperty> keys = type.Key();
+            if (keys.Count() != namedValues.Count)
             {
                 throw ExceptionUtil.CreateBadRequestError(Strings.BadRequest_KeyCountMismatch(type.FullName()));
             }
 
-            foreach (IEdmStructuralProperty property in keyProperties)
+            foreach (IEdmStructuralProperty property in keys)
             {
                 string valueText;
 
@@ -369,17 +397,27 @@ namespace Microsoft.OData.UriParser
                 {
                     if (EnableCaseInsensitive)
                     {
-                        var list = namedValues.Keys.Where(key => string.Equals(property.Name, key, StringComparison.OrdinalIgnoreCase)).ToList();
-                        if (list.Count > 1)
+                        var list = namedValues.Keys.Where(key => string.Equals(property.Name, key, StringComparison.OrdinalIgnoreCase));
+
+                        string caseInsensitiveKey = string.Empty;
+                        bool keyFound = false;
+                        foreach (string key in list)
                         {
-                            throw new ODataException(Strings.UriParserMetadata_MultipleMatchingKeysFound(property.Name));
+                            if (keyFound)
+                            {
+                                throw new ODataException(Strings.UriParserMetadata_MultipleMatchingKeysFound(property.Name));
+                            }
+
+                            caseInsensitiveKey = key;
+                            keyFound = true;
                         }
-                        else if (list.Count == 0)
+
+                        if (!keyFound)
                         {
                             throw ExceptionUtil.CreateSyntaxError();
                         }
 
-                        valueText = namedValues[list.Single()];
+                        valueText = namedValues[caseInsensitiveKey];
                     }
                     else
                     {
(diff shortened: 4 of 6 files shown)


src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -1829,20 +1926,32 @@ private static bool ParametersSatisfyFunction(IEdmOperation operation, IList<str
             List<IEdmOperationParameter> functionParameters = parametersToMatch.ToList();
```
joaocpaiva: >List<IEdmOperationParameter> functionParameters = parametersToMatch.ToList(); [](start = 12, length = 77)

We can avoid this ToList allocation and use the enumerator in the below loop, since making the List copy is not giving us much in this method and the allocation is the most expensive here.


foreach (var functionParemeter in parametersToMatch)
{
 ...
}

joaocpaiva:
>             .ToList();

Save this allocation, not needed. We can just call SingleOrDefault (if null throw ODataException).

---
Refers to: src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs:129 in 5adb8e8. [](commit_id = 5adb8e8309ae55795b60ab4a6e2c34ad58d5d304, deletion_comment = False)

joaocpaiva:
>                 .Where(source => string.Equals(identifier, source.Name, StringComparison.OrdinalIgnoreCase)).ToList();

Ditto, save this allocation.

---
Refers to: src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs:103 in 5adb8e8. [](commit_id = 5adb8e8309ae55795b60ab4a6e2c34ad58d5d304, deletion_comment = False)

joaocpaiva:
>             var keyProperties = type.Key().ToList();

Save this allocation, use foreach loop below.

---
Refers to: src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs:317 in 5adb8e8. [](commit_id = 5adb8e8309ae55795b60ab4a6e2c34ad58d5d304, deletion_comment = False)

joaocpaiva:
>             var list = operation.Parameters.Where(parameter => string.Equals(identifier, parameter.Name, StringComparison.Ordinal)).ToList();

I think we can save these allocations in these method. Something like:

var list = operation.Parameters.Where(parameter => string.Equals(identifier, parameter.Name, StringComparison.Ordinal));
            if (list.Any())
            {
                // if no case sensitive, try case-insensitive
                list = operation.Parameters.Where(parameter => string.Equals(identifier, parameter.Name, StringComparison.OrdinalIgnoreCase));
            }

            var result = list.FirstOrDefault();

            if (result != null && result == list.Last())
            {
                throw new ODataException(Strings.UriParserMetadata_MultipleMatchingParametersFound(identifier));
            }

            return result;

---
Refers to: src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs:412 in 5adb8e8. [](commit_id = 5adb8e8309ae55795b60ab4a6e2c34ad58d5d304, deletion_comment = False)

joaocpaiva:
>                 list = operation.Parameters.Where(parameter => string.Equals(identifier, parameter.Name, StringComparison.OrdinalIgnoreCase)).ToList();

Can't we always do IgnoreCase to avoid two O(n) searches?

---
Refers to: src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs:416 in 5adb8e8. [](commit_id = 5adb8e8309ae55795b60ab4a6e2c34ad58d5d304, deletion_comment = False)

joaocpaiva:
>                         var list = namedValues.Keys.Where(key => string.Equals(property.Name, key, StringComparison.OrdinalIgnoreCase)).ToList();

We can also save this allocation. Calling Any(), SingleOrDefault, LastOrDefault will be better than copying an array.



---
Refers to: src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs:373 in 5adb8e8. [](commit_id = 5adb8e8309ae55795b60ab4a6e2c34ad58d5d304, deletion_comment = False)

joaocpaiva:
>             var keyProperties = type.Key().ToList();

If this can be large, it will still be better to save ToList() and instead resort to Count() to compare the length of the arrays. Or do the check as you go through the loop below.

---
Refers to: src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs:355 in 5adb8e8. [](commit_id = 5adb8e8309ae55795b60ab4a6e2c34ad58d5d304, deletion_comment = False)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -265,6 +267,111 @@ internal static IEnumerable<IEdmOperation> FindBestOverloadBasedOnParameters(thi
...
+        /// <param name="caseInsensitive">Whether case insensitive.</param>
+        /// <returns>The best matching operations based on parameters.</returns>
+        internal static IList<IEdmOperation> FindBestOverloadBasedOnParameters(this IList<IEdmOperation> functions, IList<string> parameters, bool caseInsensitive = false)
+        {
+            IList<IEdmOperation> exactMatches = new List<IEdmOperation>();
+
+            // The best match out of a list of candidates is the one that has the same number of (non-binding) parameters as specified.
+            for(int i=0; i < functions.Count; i++)
+            {
+                if (functions[i].Parameters.Count() == parameters.Count + (functions[i].IsBound ? 1 : 0))
```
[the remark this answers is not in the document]
KanishManuja-MS: IEnumerable does not have a count property. A cast to List will also be expensive. Am I missing something here? 

---
In reply to: [352772232](https://github.com/OData/odata.net/pull/1602#discussion_r352772232) [](ancestors = 352772232)

KanishManuja-MS:
>             var keyProperties = type.Key().ToList();

Saved this allocation but mind you, to avoid a breaking change, we will have to call Count() on the IEnumerable. However, since the these are keys, the length should be fairly small to iterate. 
Expensive part here would be to get the enumerator twice. 

---
In reply to: [561378338](https://github.com/OData/odata.net/pull/1602#issuecomment-561378338) [](ancestors = 561378338)

---
Refers to: src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs:317 in 5adb8e8. [](commit_id = 5adb8e8309ae55795b60ab4a6e2c34ad58d5d304, deletion_comment = False)

KanishManuja-MS:
>             var list = operation.Parameters.Where(parameter => string.Equals(identifier, parameter.Name, StringComparison.Ordinal)).ToList();

Slightly different implementation but I think it might be more performant. 

---
In reply to: [561381264](https://github.com/OData/odata.net/pull/1602#issuecomment-561381264) [](ancestors = 561381264)

---
Refers to: src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs:412 in 5adb8e8. [](commit_id = 5adb8e8309ae55795b60ab4a6e2c34ad58d5d304, deletion_comment = False)

KanishManuja-MS:
>                 list = operation.Parameters.Where(parameter => string.Equals(identifier, parameter.Name, StringComparison.OrdinalIgnoreCase)).ToList();

I suppose there is scope for improvement. However, the logic is that we first try and find the exact case matching parameter and then if nothing is found then find case insensitive parameter. If multiple are found in either of the two searching, we throw an exception. We would need to write additional logic to distinguish between multiple case sensitive and insensitive results; we can probably do that by allocating another list. Since the count of these enumerables will be small, the only cost is for getting the enumerator. The question comes down to 2 gets on enumerators vs 1 list allocation. 

---
In reply to: [561381545](https://github.com/OData/odata.net/pull/1602#issuecomment-561381545) [](ancestors = 561381545)

---
Refers to: src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs:416 in 5adb8e8. [](commit_id = 5adb8e8309ae55795b60ab4a6e2c34ad58d5d304, deletion_comment = False)

joaocpaiva:
>             if (possibleFunctions.Count() > 1)

We seem to have multiple places where we check Count() > 1 which for collections that cannot cast to IList<> will result in O(n) algorithm.

It would be more effective to create an extension that leverages GetEnumerator to move 2 positions to find whether the array size is >1. It will also save the casts Count() does.

---
Refers to: src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs:180 in 8b1a788. [](commit_id = 8b1a788e4c877a51439feb53c22eca51bde764b8, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
```diff
@@ -185,7 +185,7 @@ internal static bool TryBindAsOperation(PathSegmentToken pathToken, IEdmModel mo
             // If more than one overload matches, try to select based on optional parameters
             if (possibleFunctions.Count() > 1 && parameterNames.Count > 0)
```
joaocpaiva: >Count [](start = 34, length = 5)

Like this too.

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -164,45 +165,38 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
...
-                candidateMatchingOperations = candidateMatchingOperations.Where(o => (o.IsFunction() && !o.Parameters.Any()) || o.IsAction());
+                candidatesMatchingOperations = candidatesMatchingOperations.FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType(bindingType) as IList<IEdmOperation>;
             }
 
-            // Only filter if there is more than one and its needed.
-            if (candidateMatchingOperations.Count() > 1)
+            // If any of the things returned are an action, it better be the only thing returned, and there can't be parameters in the URL
+            bool actionExists = false;
+            bool functionExists = false;
+            for (int i = 0; i < candidatesMatchingOperations.Count; i++)
```
joaocpaiva: >for [](start = 12, length = 3)

Should we break the loop as soon as functionExists and/or actionExists ?

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -313,22 +342,20 @@ public virtual IEnumerable<IEdmOperationImport> ResolveOperationImports(IEdmMode
...
-            var keyProperties = type.Key().ToList();
-
             // Throw an error if key size from url doesn't match that from model.
             // Other derived ODataUriResolver intended for alternative key resolution, such as the built in AlternateKeysODataUriResolver,
             // should override this ResolveKeys method.
-            if (keyProperties.Count != positionalValues.Count)
+            if (type.Key().Count() != positionalValues.Count)
             {
                 throw ExceptionUtil.CreateBadRequestError(Strings.BadRequest_KeyCountMismatch(type.FullName()));
             }
```
joaocpaiva: If we optimize for the success scenario I think we could avoid the Count().

1. first do the foreach loop while counting how many elements there are in Key().
2. after the foreach loop, check if the count of Key() != than positionalValues.Count. Throw exception if so.

Wouldn't this work?

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -351,35 +378,44 @@ public virtual IEnumerable<IEdmOperationImport> ResolveOperationImports(IEdmMode
         public virtual IEnumerable<KeyValuePair<string, object>> ResolveKeys(IEdmEntityType type, IDictionary<string, string> namedValues, Func<IEdmTypeReference, string, object> convertFunc)
         {
             var convertedPairs = new Dictionary<string, object>(StringComparer.Ordinal);
-            var keyProperties = type.Key().ToList();
 
             // Throw an error if key size from url doesn't match that from model.
             // Other derived ODataUriResolver intended for alternative key resolution, such as the built in AlternateKeysODataUriResolver,
             // should override this ResolveKeys method.
-            if (keyProperties.Count != namedValues.Count)
+            if (type.Key().Count() != namedValues.Count)
```
joaocpaiva: >Count()  [](start = 27, length = 8)

Same comment applies here.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
KanishManuja-MS: @joaocpaiva It might be counted as a breaking change. In that case, we will end up throwing the other exception before we could throw key count mismatch. 
I know this sounds like a bad request scenario anyway and people shouldn't care about it but we have had customers complain because their tests started failing after an upgrade.

joaocpaiva: Yeah that makes sense. It is ok to leave as is then. Alternatively, we could also move the Syntax Error exception after the loop but it does not seem worth the trouble.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -99,14 +100,23 @@ public virtual IEdmNavigationSource ResolveNavigationSource(IEdmModel model, str
             }
 
             var result = container.Elements.OfType<IEdmNavigationSource>()
-                .Where(source => string.Equals(identifier, source.Name, StringComparison.OrdinalIgnoreCase)).ToList();
+                .Where(source => string.Equals(identifier, source.Name, StringComparison.OrdinalIgnoreCase));
+
+            IEdmNavigationSource resolvedNavigationSource = default(IEdmNavigationSource);
```
xuzhg: >default(IEdmNavigationSource); [](start = 60, length = 30)

I don't understand what's the performance advantage to use "default(interface)" and "null"?

xuzhg:
>             IEdmNavigationSource navSource = model.FindDeclaredNavigationSource(identifier);

Currently, the "resolve" process is two steps:
1) using case sensitive to resolve
2) not find at #1, search using case insensitive.

I think consider to combine 2 steps into one can improve the performance a lot. 

---
Refers to: src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs:90 in 8b1a788. [](commit_id = 8b1a788e4c877a51439feb53c22eca51bde764b8, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -313,22 +342,20 @@ public virtual IEnumerable<IEdmOperationImport> ResolveOperationImports(IEdmMode
...
             var keyPairList = new List<KeyValuePair<string, object>>(positionalValues.Count);
-
-            for (int i = 0; i < keyProperties.Count; i++)
+            
+            int i = 0; 
+            foreach (IEdmProperty keyProperty in type.Key())
             {
-                string valueText = positionalValues[i];
-                IEdmProperty keyProperty = keyProperties[i];
+                string valueText = positionalValues[i++];
```
xuzhg: >[i++] [](start = 51, length = 5)

random acess using '[]' in the list maybe a performance problem

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -313,22 +342,20 @@ public virtual IEnumerable<IEdmOperationImport> ResolveOperationImports(IEdmMode
...
             {
                 throw ExceptionUtil.CreateBadRequestError(Strings.BadRequest_KeyCountMismatch(type.FullName()));
             }
 
             var keyPairList = new List<KeyValuePair<string, object>>(positionalValues.Count);
-
-            for (int i = 0; i < keyProperties.Count; i++)
+            
+            int i = 0; 
+            foreach (IEdmProperty keyProperty in type.Key())
```
xuzhg: >type.Key() [](start = 49, length = 10)

why call Key() twice in one method. It's also performance

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -408,24 +444,28 @@ public virtual IEnumerable<IEdmOperationImport> ResolveOperationImports(IEdmMode
...
             // first look for a case-sensitive match
-            var list = operation.Parameters.Where(parameter => string.Equals(identifier, parameter.Name, StringComparison.Ordinal)).ToList();
-            if (list.Count == 0)
+            var list = operation.Parameters.Where(parameter => string.Equals(identifier, parameter.Name, StringComparison.Ordinal));
+            if (!list.Any())
             {
                 // if no case sensitive, try case-insensitive
-                list = operation.Parameters.Where(parameter => string.Equals(identifier, parameter.Name, StringComparison.OrdinalIgnoreCase)).ToList();
+                list = operation.Parameters.Where(parameter => string.Equals(identifier, parameter.Name, StringComparison.OrdinalIgnoreCase));
             }
```
xuzhg: can we combine them as one?

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -3182,19 +3182,12 @@ internal static bool TryGetStaticEntitySet(this IEdmPathExpression pathExpressio
 
         internal static bool HasAny<T>(this IEnumerable<T> enumerable) where T : class
         {
-            IList<T> list = enumerable as IList<T>;
-            if (list != null)
+            if (enumerable != null && enumerable.GetEnumerator().MoveNext())
```
xuzhg: >&& enumerable.GetEnumerator().MoveNext()) [](start = 35, length = 41)

can you point me the materials to say to use this to test "HasAny"

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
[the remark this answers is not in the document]
xuzhg: Sorry. here i mean the perf about the "[i]" on the IList<T>.



---
In reply to: [353939228](https://github.com/OData/odata.net/pull/1602#discussion_r353939228) [](ancestors = 353939228,352772232)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -1826,23 +1934,33 @@ private static bool ParametersSatisfyFunction(IEdmOperation operation, IList<str
...
-                    p => parameterNameList.All(
-                        k => !string.Equals(k, p.Name, caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))))
+            BitArray bitmask = new BitArray(parameterNameList.Count);
+            foreach (var functionParameter in parametersToMatch)
             {
-                return false;
+                bool matched = false;
+                for (int j = 0; j < parameterNameList.Count; j++)
+                {
+                    if (string.Equals(parameterNameList[j], functionParameter.Name, caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
```
xuzhg: >parameterNameList[j] [](start = 38, length = 20)

same here for the [j] on IList<T>, it's not random access. What's the benefit to access using "index"?  Maybe I miss something.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -265,6 +267,111 @@ internal static IEnumerable<IEdmOperation> FindBestOverloadBasedOnParameters(thi
...
         }
 
+        /// <summary>
+        /// Filters the operations by parameter names.
+        /// </summary>
+        /// <param name="functions">The operations.</param>
+        /// <param name="parameters">The list of non-binding parameter names to match.</param>
+        /// <param name="caseInsensitive">Whether case insensitive.</param>
+        /// <returns>The best matching operations based on parameters.</returns>
+        internal static IList<IEdmOperation> FindBestOverloadBasedOnParameters(this IList<IEdmOperation> functions, IList<string> parameters, bool caseInsensitive = false)
```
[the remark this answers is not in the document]
KanishManuja-MS: No, I will get that once the service adopt the new version. I can try creating new perf tests but that will not be accurate. 

---
In reply to: [352770976](https://github.com/OData/odata.net/pull/1602#discussion_r352770976) [](ancestors = 352770976)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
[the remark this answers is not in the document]
KanishManuja-MS: It is random access. underlying structure for List is an array. Since this is internally used, we know that IList is implemented by an array based backing field. 

---
In reply to: [363453414](https://github.com/OData/odata.net/pull/1602#discussion_r363453414) [](ancestors = 363453414,353939228,352772232)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
KanishManuja-MS: We can break when both exists. 

---
In reply to: [354092013](https://github.com/OData/odata.net/pull/1602#discussion_r354092013) [](ancestors = 354092013)

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
KanishManuja-MS: Random access is allowed in lists. It does not iterate through the entire list. LinkedList would be the one where random access would be a problem. 

---
In reply to: [362108940](https://github.com/OData/odata.net/pull/1602#discussion_r362108940) [](ancestors = 362108940)

KanishManuja-MS:
>             IEdmNavigationSource navSource = model.FindDeclaredNavigationSource(identifier);

I think we had discussed this offline. Ignore case is twice as expensive check compared to case sensitive. Let's not degrade the performance when someone is specifying the correct case. 

---
In reply to: [569810729](https://github.com/OData/odata.net/pull/1602#issuecomment-569810729) [](ancestors = 569810729)

---
Refers to: src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs:90 in 8b1a788. [](commit_id = 8b1a788e4c877a51439feb53c22eca51bde764b8, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
KanishManuja-MS: As per previous reply, Ordinal string comparison will be cheaper than Ignore case. 

---
In reply to: [362109307](https://github.com/OData/odata.net/pull/1602#discussion_r362109307) [](ancestors = 362109307)

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
KanishManuja-MS: Refer to Line 804 in ExtensionMethodTests. 

---
In reply to: [362109873](https://github.com/OData/odata.net/pull/1602#discussion_r362109873) [](ancestors = 362109873)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
KanishManuja-MS: random access is faster. 

---
In reply to: [363456913](https://github.com/OData/odata.net/pull/1602#discussion_r363456913) [](ancestors = 363456913)

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
KanishManuja-MS: To keep the parity. Previously, it was SingleOrDefault, to avoid any breaking changes, I relied on Default here rather than null. 

---
In reply to: [362107858](https://github.com/OData/odata.net/pull/1602#discussion_r362107858) [](ancestors = 362107858)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,19 +254,135 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+                    if (exactMatches == null)
+                    {
+                        exactMatches = new List<IEdmOperation>();
+                    }
+
+                    exactMatches.Add(functions[i]);
+                }
+            }
+
+            return exactMatches == null ? functions : exactMatches;
```
xuzhg: >? functions  [](start = 40, length = 12)

why return all when no exact matches?

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -217,45 +216,58 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
...
 
-            T[] array = enumerable as T[];
-            if (array != null)
+            return false;
+        }
+
+        internal static bool CountGreaterThan<T>(this IEnumerable<T> enumerable, int count) where T : class
+        {
+            int currentCount = 0;
+            foreach (T element in enumerable)
```
xuzhg: can you create a simple console app to test the benchmark between "CountGreaterThan<T> " and the ">0" operator?

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -124,15 +134,23 @@ public virtual IEdmProperty ResolveProperty(IEdmStructuredType type, string prop
...
+            IEdmProperty resolvedProperty = default(IEdmProperty);
 
-            if (result.Count > 1)
+            foreach(var candidate in result)
             {
-                throw new ODataException(Strings.UriParserMetadata_MultipleMatchingPropertiesFound(propertyName, type.FullTypeName()));
+                if (resolvedProperty == default(IEdmProperty))
+                {
+                    resolvedProperty = candidate;
+                }
```
xuzhg: 1. the codes here look unreadable. 

2. we can't use "null" to replace default(IEdmProperty)

3) SingelOrDefault will evaluate the Linq, and it's readable and maintainable, so, what's the percentage to replace it like the codes here?

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -151,12 +169,22 @@ public virtual IEdmTerm ResolveTerm(IEdmModel model, string termName)
...
 
+            if (results == null)
+            {
+                return null;
+            }
+
+            if (results == null)
+            {
+                return null;
+            }
```
xuzhg: what's the purpose to test Results twice?

xuzhg:
I'd to see reviews/comments from @michael Pizzo.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
KanishManuja-MS: Keeping the behavior same as the earlier function call. 

---
In reply to: [370381493](https://github.com/OData/odata.net/pull/1602#discussion_r370381493) [](ancestors = 370381493)

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
joaocpaiva: That's a good point @xuzhg - random access is preferable when IList objects boil down to List, since it implements random access as O(1) operation. However, when we are not absolutely sure what's the underlying implementation we should probably do foreach(...) instead of for(...) loop. Moreover, foreach(...) VS for(...) loop is a micro optimization that I don't feel too strong about it either.

To help with this, we should start using the underlying class, instead of the interface (unless when there is a need to send down the interface).

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -217,45 +216,58 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
...
             {
-                return array.Length > 0;
+                if (element == null)
+                {
```
joaocpaiva: Should we skip null elements? The previous logic that was using Count() > 1 didn't really skip null elements.

KanishManuja-MS: I expect none of the elements to be null. If they are null, we should not be counting those anyway. I explicitly had to add this check as the element value was not being used otherwise. 

---
In reply to: [370402281](https://github.com/OData/odata.net/pull/1602#discussion_r370402281) [](ancestors = 370402281)

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
KanishManuja-MS: Merge error. 

---
In reply to: [370388471](https://github.com/OData/odata.net/pull/1602#discussion_r370388471) [](ancestors = 370388471)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -1868,6 +1986,19 @@ private static int InheritanceLevelFromSpecifiedInheritedType(this IEdmStructure
...
         }
+
+        private static void EnsureOperationBoundWithBindingParameter(this IEdmOperation operation)
+        {
+            if (!operation.IsBound)
+            {
+                throw new ODataException(ErrorStrings.EdmLibraryExtensions_UnBoundOperationsFoundFromIEdmModelFindMethodIsInvalid(operation.Name));
+            }
+
+            if (operation.Parameters.FirstOrDefault() == null)
```
mikepizzo: >.FirstOrDefault() == null [](start = 36, length = 25)

Is FirstOrDefault() more efficient than Any()?  You don't actually need the parameter, you just need to know if one exists, right?

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,19 +254,135 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
             return exactMatches.Count() > 0 ? exactMatches : functions;
         }
 
+        /// <summary>
+        /// Filters the operations by parameter count.
+        /// </summary>
+        /// <param name="functions">The operations.</param>
+        /// <param name="parameterCount">The count of non-binding parameter names to match.</param>
+        /// <returns>The best matching operations based on parameters.</returns>
+        internal static IList<IEdmOperation> FindBestOverloadBasedOnParameterCount(this IList<IEdmOperation> functions, int parameterCount)
```
mikepizzo: >functions [](start = 109, length = 9)

The title implies operations (which include functions and actions) but this parameter name implies just functions.  Does this get used for both actions and functions, or just functions? (yes, I realize you just copied the signature from the previous IEnumerable version, but we should fix it both places).

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,19 +254,135 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+
+            bool hasParameters = parameterNameList.Count > 0;
+            actionItems = null;
+            IList<IEdmOperation> operationCandidates = new List<IEdmOperation>();
+
+            foreach (var operation in operations)
+            {
+                if (bindingType != null)
+                {
+                    operation.EnsureOperationBoundWithBindingParameter();
```
mikepizzo: >EnsureOperationBoundWithBindingParameter [](start = 30, length = 40)

should this also validate that the binding type is the same as the first parameter, or is that done elsewhere?

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,19 +254,135 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+                }
+
+                bool suitableCandidate = true;
+
+                // If the number of parameters > 0 then this has to be a function as actions can't have parameters on the uri only in the payload. Filter further by parameters in this case, otherwise don't.
```
mikepizzo: >// If the number of parameters > 0 then this has to be a function as actions can't have parameters on the uri only in the payload. Filter further by parameters in this case, otherwise don't. [](start = 16, length = 190)

This seems to contradict the subsequent line that checks to see if an operation is an action.  I think what you mean is that, if the operation is an action, it can have at most one parameter (the binding parameter).

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,19 +254,135 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+                        if (count == 0)
+                        {
+                            suitableCandidate = false;
+                        }
+                    }
+                    else if (!operation.IsAction())
+                    {
+                        // If IsFunction and !IsAction are the same then this may not be required. To be confirmed.
+                        suitableCandidate = false;
+                    }
```
mikepizzo: yes; seems like this would never be called. maybe replace with an assert?

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,19 +254,135 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+                    if (operation.IsAction())
+                    {
+                        if (actionItems == null)
+                        {
+                            actionItems = new List<IEdmOperation>();
+                        }
+
+                        actionItems.Add(operation);
+                        suitableCandidate = false;
+                    }
```
mikepizzo: why is this nested under if(hasParameters). If it's an action, don't we always want to add it to actionItems and move on? Maybe do this check before initializing suitableCandidates = true, and make suitableCandidates = true an else for operation.IsAction().

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,19 +254,135 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+                {
+                    // If it is a bound function without required parameters, then the parameter list should only have one parameter and any others as optional.
+                    if (operation.IsFunction())
+                    {
+                        int count = 0;
+                        foreach (var param in operation.Parameters)
+                        {
+                            if (count > 0 && !(param is IEdmOptionalParameter))
+                            {
+                                suitableCandidate = false;
```
mikepizzo: you should be able to break out of the foreach as soon as you find a non-optional parameter.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,19 +254,135 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+                            }
+
+                            count++;
+                        }
+
+                        if (count == 0)
+                        {
+                            suitableCandidate = false;
+                        }
```
mikepizzo: You should be able to break out of the foreach as soon as you find a non-optional parameter.  Can you simplify to something like:

                        foreach (var param in operation.Parameters)
                        {
                            if (!(param is IEdmOptionalParameter))
                            {
                                suitableCandidate = false;
                               break;
                            }
                        }

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,19 +254,135 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+                    }
+                    else if (!operation.IsAction())
+                    {
+                        // If IsFunction and !IsAction are the same then this may not be required. To be confirmed.
+                        suitableCandidate = false;
+                    }
+                }
+                else
+                {
+                    if (operation.IsFunction() && operation.Parameters.HasAny())
```
mikepizzo: >operation.IsFunction() &&  [](start = 24, length = 26)

if you deal with actions at the beginning then you don't have to keep checking.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,19 +254,135 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+                    }
+
+                    if (!ParametersSatisfyFunction(operation, parameterNameList, caseInsensitive))
+                    {
+                        suitableCandidate = false;
+                    }
+                }
+                else if (bindingType != null)
+                {
+                    // If it is a bound function without required parameters, then the parameter list should only have one parameter and any others as optional.
```
mikepizzo: >If it is a bound function without required parameters, then the parameter list should only have one parameter and any others as optiona [](start = 23, length = 135)

This is a very special case of a single required parameter because it's bound; can't we easily generalize this to the function having more required parameters than there are in parameterNameList?

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
KanishManuja-MS: They are the same. They just differ in their return types. Any is better suited though. 

---
In reply to: [370725688](https://github.com/OData/odata.net/pull/1602#discussion_r370725688) [](ancestors = 370725688)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
KanishManuja-MS: Yes, although I don't like it as much but the calling function calls this extension method only with already filtered candidates if a binding type exists. 

---
In reply to: [370730518](https://github.com/OData/odata.net/pull/1602#discussion_r370730518) [](ancestors = 370730518)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -1826,23 +1934,33 @@ private static bool ParametersSatisfyFunction(IEdmOperation operation, IList<str
...
             }
 
-            List<IEdmOperationParameter> functionParameters = parametersToMatch.ToList();
-
             // if any required parameters are missing, don't consider it a match.
-            if (functionParameters.Where(
-                p => !(p is IEdmOptionalParameter)).Any(
-                    p => parameterNameList.All(
-                        k => !string.Equals(k, p.Name, caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))))
+            BitArray bitmask = new BitArray(parameterNameList.Count);
```
mikepizzo: >bitmask [](start = 21, length = 7)

Please name to convey what it represents (i.e. matchedUrlParameters)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -139,19 +139,20 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
                 }
             }
 
-            IEnumerable<IEdmOperation> candidateMatchingOperations = null;
+            IEnumerable<IEdmOperation> operationsFromModel = null;
 
             // The extension method FindBoundOperations & FindOperations call IEdmModel.FindDeclaredBoundOperations which can be implemented by anyone and it could throw any type of exception
             // so catching all of them and simply putting it in the inner exception.
+            // It is better to cast the results of ResolveBoundOperations as IList instead of looping over this for basic things such as count.
```
mikepizzo: is this comment outdated? You don't cast here.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
KanishManuja-MS: I tried to maintain parity between the logic and the current one. I did not try to fix any bugs with this PR. We can walk through these to see if this should be improved. 

---
In reply to: [370785652](https://github.com/OData/odata.net/pull/1602#discussion_r370785652) [](ancestors = 370785652)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
KanishManuja-MS: Yes. I improved the logic to create a list while filtering instead of doing a ToList here. 

---
In reply to: [370842929](https://github.com/OData/odata.net/pull/1602#discussion_r370842929) [](ancestors = 370842929)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -164,45 +165,43 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
...
-                candidateMatchingOperations = candidateMatchingOperations.Where(o =>
-                (o.IsFunction() && (o.Parameters.Count() == 1 || o.Parameters.Skip(1).All(p => p is IEdmOptionalParameter))) || o.IsAction());
-            }
-            else
+            // Only filter if there is more than one and its needed.
+            if (candidatesMatchingOperations.Count > 1)
             {
-                // Filter out functions with any parameters
-                candidateMatchingOperations = candidateMatchingOperations.Where(o => (o.IsFunction() && !o.Parameters.Any()) || o.IsAction());
+                candidatesMatchingOperations = candidatesMatchingOperations.FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType(bindingType) as IList<IEdmOperation>;
```
mikepizzo: >FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType(bindingType) as IList<IEdmOperation>; [](start = 76, length = 106)

will FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType always return an IList<EdmOperation>? If so, can we just change the signature and avoid the cast? If not, can we change it to always return an IList?

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -217,45 +216,58 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
...
-                return list.Count > 0;
+                return true;
             }
 
-            T[] array = enumerable as T[];
-            if (array != null)
+            return false;
+        }
+
+        internal static bool CountGreaterThan<T>(this IEnumerable<T> enumerable, int count) where T : class
```
mikepizzo: >CountGreaterThan [](start = 29, length = 16)

This assume you can enumerate the enumerable multiple times (without copying).  works for a list, but may not work for other enumerables.

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
KanishManuja-MS: This is closely coupled with SelectPathSegmentTokenBinder and I wanted to limit the scope. We can decide to improve the SelectPathSegmentTokenBinder and then fix this with it. 

---
In reply to: [370851480](https://github.com/OData/odata.net/pull/1602#discussion_r370851480) [](ancestors = 370851480)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
KanishManuja-MS: This is not intended to be used for lists. This is intended for Enumerable, here is the implementation of count in Linq, this is an improvment only for IEnumerables ->
        public static int Count<TSource>(this IEnumerable<TSource> source)
        {
            if (source == null)
            {
                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.source);
            }

            if (source is ICollection<TSource> collectionoft)
            {
                return collectionoft.Count;
            }

            if (source is IIListProvider<TSource> listProv)
            {
                return listProv.GetCount(onlyIfCheap: false);
            }

            if (source is ICollection collection)
            {
                return collection.Count;
            }

            int count = 0;
            using (IEnumerator<TSource> e = source.GetEnumerator())
            {
                checked
                {
                    while (e.MoveNext())
                    {
                        count++;
                    }
                }
            }

            return count;
        }
 

---
In reply to: [370855165](https://github.com/OData/odata.net/pull/1602#discussion_r370855165) [](ancestors = 370855165)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -139,19 +139,19 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
                 }
             }
 
-            IEnumerable<IEdmOperation> candidateMatchingOperations = null;
+            IEnumerable<IEdmOperation> operationsFromModel = null;
```
KanishManuja-MS: >null [](start = 61, length = 4)

don't need to initialize;

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -1868,6 +1982,19 @@ private static int InheritanceLevelFromSpecifiedInheritedType(this IEdmStructure
 
             return inheritanceLevelsFromBase;
         }
+
+        private static void EnsureOperationBoundWithBindingParameter(this IEdmOperation operation)
+        {
+            if (!operation.IsBound)
```
KanishManuja-MS: Make an assert instead.

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -217,45 +215,58 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
                     throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_SegmentDoesNotSupportKeyPredicates(identifier));
                 }
 
-                matchingOperation = candidateMatchingOperations.Single();
+                matchingOperation = candidatesMatchingOperations.Count > 0 ? candidatesMatchingOperations[0] : null;
                 return true;
             }
 
-            if (foundActionsWhenLookingForFunctions.Count > 0)
+            if (foundActionsWhenLookingForFunctions != null && foundActionsWhenLookingForFunctions.Count > 0)
```
KanishManuja-MS: >foundActionsWhenLookingForFunctions [](start = 16, length = 35)

Make this a boolean instead; verify.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,17 +254,129 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+                // If the number of binding parameters > 0 then we should only keep functions as actions can't have parameters on the uri only in the payload. Filter further by parameters in this case, otherwise don't.
+                // This logic might be a potential bug. 
+                if (hasParameters)
+                {
+                    if (operation.IsAction())
+                    {
+                        if (actionItems == null)
+                        {
+                            actionItems = new List<IEdmOperation>();
+                        }
```
KanishManuja-MS: Set a boolean here.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,17 +254,129 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+                {
+                    // If it is a bound function without required parameters, then the parameter list should only have one parameter and any others as optional.
+                    if (operation.IsFunction())
+                    {
+                        int count = 0;
+                        foreach (var param in operation.Parameters)
+                        {
+                            if (count > 0 && !(param is IEdmOptionalParameter))
+                            {
+                                suitableCandidate = false;
```
KanishManuja-MS: See if edmlib enforces optional parameters at the end then just look for the first optional parameter.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,17 +254,129 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+                        }
+                    }
+                }
+                else
+                {
+                    if (operation.IsFunction() && operation.Parameters.HasAny())
+                    {
+                        suitableCandidate = false;
+                    }
+                }
```
KanishManuja-MS: Combine as per Mike's previous comment.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -1826,23 +1930,33 @@ private static bool ParametersSatisfyFunction(IEdmOperation operation, IList<str
...
+            foreach (var functionParameter in parametersToMatch)
             {
-                return false;
+                bool matched = false;
+                for (int j = 0; j < parameterNameList.Count; j++)
+                {
+                    if (string.Equals(parameterNameList[j], functionParameter.Name, caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
+                    {
+                        matched = true;
+                        parameterExists[j] = true;
```
KanishManuja-MS: break;

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -1826,23 +1930,33 @@ private static bool ParametersSatisfyFunction(IEdmOperation operation, IList<str
...
             }
 
             // if any specified parameters don't match, don't consider it a match.
-            if (parameterNameList.Any(
-                k => functionParameters.All(
-                    p => !string.Equals(k, p.Name, caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))))
+            for (int i = 0; i < parameterNameList.Count; i++)
             {
-                return false;
+                if (!parameterExists[i])
```
KanishManuja-MS: >parameterExists [](start = 21, length = 15)

Can i do it in O(1)?

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -164,45 +164,43 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
...
-                (o.IsFunction() && (o.Parameters.Count() == 1 || o.Parameters.Skip(1).All(p => p is IEdmOptionalParameter))) || o.IsAction());
-            }
-            else
+            // Only filter if there is more than one and its needed.
+            if (candidatesMatchingOperations.Count > 1)
             {
-                // Filter out functions with any parameters
-                candidateMatchingOperations = candidateMatchingOperations.Where(o => (o.IsFunction() && !o.Parameters.Any()) || o.IsAction());
+                candidatesMatchingOperations = candidatesMatchingOperations.FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType(bindingType) as IList<IEdmOperation>;
             }
```
KanishManuja-MS: Add null check if list casting fails.

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -164,45 +164,43 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
...
+
+                if (candidatesMatchingOperations[i].IsAction())
+                {
+                    actionExists = true;
+                }
+
+                if (functionExists && actionExists)
+                {
+                    break;
+                }
```
KanishManuja-MS: split and move to 185; and 190?

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -217,45 +215,58 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
                     throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_SegmentDoesNotSupportKeyPredicates(identifier));
                 }
 
-                matchingOperation = candidateMatchingOperations.Single();
+                matchingOperation = candidatesMatchingOperations.Count > 0 ? candidatesMatchingOperations[0] : null;
                 return true;
             }
```
KanishManuja-MS: Move these up.

KanishManuja-MS:
>                 {

This should be an assert. 

---
Refers to: src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs:214 in 1795eb9. [](commit_id = 1795eb90b874b5e59530e4191b412b127bad14c9, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -217,45 +215,58 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
...
-                matchingOperation = candidateMatchingOperations.Single();
+                matchingOperation = candidatesMatchingOperations.Count > 0 ? candidatesMatchingOperations[0] : null;
                 return true;
             }
 
-            if (foundActionsWhenLookingForFunctions.Count > 0)
+            if (foundActionsWhenLookingForFunctions != null && foundActionsWhenLookingForFunctions.Count > 0)
             {
                 throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_SegmentDoesNotSupportKeyPredicates(identifier));
             }
```
KanishManuja-MS: Check the validity of this in terms of the position.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,17 +254,129 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+                    {
+                        exactMatches = new List<IEdmOperation>();
+                    }
+
+                    exactMatches.Add(operations[i]);
+                }
+            }
+
+            return exactMatches == null ? operations : exactMatches;
+        }
```
KanishManuja-MS: Should return a single item and throw an exception if multiple operations match?

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -151,12 +169,17 @@ public virtual IEdmTerm ResolveTerm(IEdmModel model, string termName)
 
             IList<IEdmTerm> results = FindAcrossModels<IEdmTerm>(model, termName, /*caseInsensitive*/ true);
 
+            if (results == null)
```
KanishManuja-MS: >) [](start = 31, length = 1)

Add check for count == 0 and make the return statement cleaner.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -351,35 +379,45 @@ public virtual IEnumerable<IEdmOperationImport> ResolveOperationImports(IEdmMode
...
                 {
                     if (EnableCaseInsensitive)
                     {
-                        var list = namedValues.Keys.Where(key => string.Equals(property.Name, key, StringComparison.OrdinalIgnoreCase)).ToList();
-                        if (list.Count > 1)
+                        var list = namedValues.Keys.Where(key => string.Equals(property.Name, key, StringComparison.OrdinalIgnoreCase));
+
+                        string caseInsensitiveKey = string.Empty;
+                        int count = 0;
+                        foreach (var key in list)
```
KanishManuja-MS: make it a boolean.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -438,22 +480,36 @@ internal static ODataUriResolver GetUriResolver(IServiceProvider container)
...
 
             return results;
         }
 
-        private static IEnumerable<T> FindSchemaElements<T>(IEdmModel model, string qualifiedName, bool caseInsensitive) where T : IEdmSchemaElement
+        private static void FindSchemaElements<T>(IEdmModel model, string qualifiedName, bool caseInsensitive, ref IList<T> results) where T : IEdmSchemaElement
         {
-            return model.SchemaElements.OfType<T>()
-            .Where(e => string.Equals(qualifiedName, e.FullName(), caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal));
+            foreach (var schema in model.SchemaElements)
```
KanishManuja-MS: >var [](start = 21, length = 3)

IEDMSchemaElement, look for type in above changes as well.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -438,22 +480,36 @@ internal static ODataUriResolver GetUriResolver(IServiceProvider container)
             return container.GetRequiredService<ODataUriResolver>();
         }
 
-        private static List<T> FindAcrossModels<T>(IEdmModel model, String qualifiedName, bool caseInsensitive) where T : IEdmSchemaElement
+        private static IList<T> FindAcrossModels<T>(IEdmModel model, String qualifiedName, bool caseInsensitive) where T : IEdmSchemaElement
```
KanishManuja-MS: >private [](start = 8, length = 7)

Is it okay to return null vs an empty list?

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -438,22 +480,36 @@ internal static ODataUriResolver GetUriResolver(IServiceProvider container)
             return container.GetRequiredService<ODataUriResolver>();
         }
 
-        private static List<T> FindAcrossModels<T>(IEdmModel model, String qualifiedName, bool caseInsensitive) where T : IEdmSchemaElement
+        private static IList<T> FindAcrossModels<T>(IEdmModel model, String qualifiedName, bool caseInsensitive) where T : IEdmSchemaElement
         {
-            List<T> results = FindSchemaElements<T>(model, qualifiedName, caseInsensitive).ToList();
+            IList<T> results = null;
```
KanishManuja-MS: Remove empty list.

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
KanishManuja-MS: >enumerable.GetEnumerator().MoveNext() [](start = 38, length = 37)

return the movenext.

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -217,45 +215,58 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
...
-                return list.Count > 0;
+                return true;
             }
 
-            T[] array = enumerable as T[];
-            if (array != null)
+            return false;
+        }
+
+        internal static bool CountGreaterThan<T>(this IEnumerable<T> enumerable, int count) where T : class
```
KanishManuja-MS: move as private.

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -217,45 +215,58 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
...
+        {
+            int currentCount = 0;
+            foreach (T element in enumerable)
             {
```
KanishManuja-MS: While move next to avoid the null check.

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
```diff
@@ -217,45 +215,58 @@ internal static bool ResolveOperationFromList(string identifier, IEnumerable<str
...
-                return array.Length > 0;
+                if (element == null)
+                {
+                    continue;
+                }
+
+                currentCount++;
```
KanishManuja-MS: make it ++currentCount > count instead.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -438,22 +480,36 @@ internal static ODataUriResolver GetUriResolver(IServiceProvider container)
...
+                if (string.Equals(qualifiedName, schema.FullName(), caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
+                {
+                    if (schema is T)
+                    {
+                        if (results == null)
+                        {
+                            results = new List<T>();
+                        }
+
+                        results.Add((T)schema);
```
KanishManuja-MS: >(( [](start = 35, length = 2)

Try and get rid of this repeated cast.

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
KanishManuja-MS: Converted to assert for functionoverload resolver but did not revert for select segment code path. 

---
In reply to: [370876004](https://github.com/OData/odata.net/pull/1602#discussion_r370876004) [](ancestors = 370876004)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
KanishManuja-MS: Nope. 

---
In reply to: [370886886](https://github.com/OData/odata.net/pull/1602#discussion_r370886886) [](ancestors = 370886886)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
KanishManuja-MS: Can't move both the exceptions to above as we need to traverse the entire list before deciding if both action and function exist vs only actions exists. 

---
In reply to: [370888279](https://github.com/OData/odata.net/pull/1602#discussion_r370888279) [](ancestors = 370888279)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
KanishManuja-MS: Can't, refactored into a separate function now. Moving both of the exceptions does not improve the code as we need to iterate through all the operations anyway. 

---
In reply to: [370888720](https://github.com/OData/odata.net/pull/1602#discussion_r370888720) [](ancestors = 370888720)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
KanishManuja-MS: Instead of being smart about not matching an action when request uri has parameters, this code throws even if there might be functions matching. Now, we only throw when there are no functions matched as well. 

---
In reply to: [370889470](https://github.com/OData/odata.net/pull/1602#discussion_r370889470) [](ancestors = 370889470)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
KanishManuja-MS: Instead of throwing, I decided to change the name of the function to indicate filtering rather than finding the best overload. 

---
In reply to: [370893088](https://github.com/OData/odata.net/pull/1602#discussion_r370893088) [](ancestors = 370893088)

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
KanishManuja-MS: Does not help. If it is a class, we can but not with an interface. 

---
In reply to: [370899362](https://github.com/OData/odata.net/pull/1602#discussion_r370899362) [](ancestors = 370899362)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -252,17 +254,117 @@ internal static IEnumerable<IEdmOperation> FilterBoundOperationsWithSameTypeHier
...
+                    }
+                }
+                else
+                {
+                    // If it is a bound function without required parameters, then the parameter list should only have one parameter and any others as optional.
+                    if (operation.IsFunction())
+                    {
+                        // Binding type requires atleast one parameter for a function. 
+                        bool parameterFound = bindingType != null ? false : true;
+                        foreach (var param in operation.Parameters)
```
mikepizzo: >var [](start = 33, length = 3)

try to avoid use of var

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
mikepizzo: It doesn't look like you added the break; is there a reason you have to continue looping through the remaining parameter names once you find one that matches?

---
In reply to: [370884636](https://github.com/OData/odata.net/pull/1602#discussion_r370884636) [](ancestors = 370884636)

mikepizzo:
>                 list = operation.Parameters.Where(parameter => string.Equals(identifier, parameter.Name, StringComparison.OrdinalIgnoreCase)).ToList();

Right. The logic for case insensitive is to first match case sensitive. If that fails, attempt to match case insensitive, in which case there can be only one case-insensitive match.  This is by protocol (and best practice for handling case-insensitive comparison)

---
In reply to: [561816744](https://github.com/OData/odata.net/pull/1602#issuecomment-561816744) [](ancestors = 561816744,561381545)

---
Refers to: src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs:416 in 5adb8e8. [](commit_id = 5adb8e8309ae55795b60ab4a6e2c34ad58d5d304, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
mikepizzo: not marked as private.

---
In reply to: [370898092](https://github.com/OData/odata.net/pull/1602#discussion_r370898092) [](ancestors = 370898092)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
mikepizzo: could also do if (enumerable != null) { return enumerable.GetEnumerator().MoveNext();}

---
In reply to: [370898326](https://github.com/OData/odata.net/pull/1602#discussion_r370898326) [](ancestors = 370898326)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
KanishManuja-MS: Discussed. 

---
In reply to: [371602187](https://github.com/OData/odata.net/pull/1602#discussion_r371602187) [](ancestors = 371602187,370884636)

mikepizzo: More specifically; this matches the current (flawed) behavior of flagging *all* case-insensitive matches. Kanish to open a separate issue to track fixing this existing logic.

---
In reply to: [371616803](https://github.com/OData/odata.net/pull/1602#discussion_r371616803) [](ancestors = 371616803,371602187,370884636)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
KanishManuja-MS: EdmLib does not validate the ordering. Just checking for the first optional parameter fails some of our tests. 

---
In reply to: [370881673](https://github.com/OData/odata.net/pull/1602#discussion_r370881673) [](ancestors = 370881673)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionOverloadResolver.cs
KanishManuja-MS: I changed CountGreater than to private but ended up removing that. This is a shared function that gets used in SelectPathSegmentTokenBinder, therefore keeping as internal. 

---
In reply to: [371615196](https://github.com/OData/odata.net/pull/1602#discussion_r371615196) [](ancestors = 371615196,370898092)

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
KanishManuja-MS: [https://github.com/OData/odata.net/issues/1649](https://github.com/OData/odata.net/issues/1649)

---
In reply to: [371618735](https://github.com/OData/odata.net/pull/1602#discussion_r371618735) [](ancestors = 371618735,371616803,371602187,370884636)

KanishManuja-MS:
>             if (possibleFunctions.Count() > 1)

Reverted the CountGreaterThan change. Experimental data showed that .Count() outperforms CountGreaterThan for smaller values. 

---
In reply to: [561946259](https://github.com/OData/odata.net/pull/1602#issuecomment-561946259) [](ancestors = 561946259)

---
Refers to: src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs:180 in 8b1a788. [](commit_id = 8b1a788e4c877a51439feb53c22eca51bde764b8, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
KanishManuja-MS: Reverted the CountGreaterThan change. Experimental data showed that .Count() outperforms CountGreaterThan for smaller values. 

---
In reply to: [354091432](https://github.com/OData/odata.net/pull/1602#discussion_r354091432) [](ancestors = 354091432)

PR #1603, opened 2019-12-01
Fix CrmProfile performance issues

KanishManuja-MS:
### Issues

This pull request fixes the issues that were reported in CRM Profiling. 

### Description
Unnecessary ToList calls were replaced by List property. Since the function consuming this property was public, the signature of the function could not be changed and therefore a concrete List was added to the class as a workaround. 

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/BindingPathHelper.cs
+++ b/src/Microsoft.OData.Core/BindingPathHelper.cs
@@ -4,6 +4,7 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using Microsoft.OData.Edm;
@@ -20,7 +21,7 @@ namespace Microsoft.OData
         /// <param name="bindingPath">The binding path.</param>
         /// <param name="parsedSegments">The list of segments in Uri path.</param>
         /// <returns>True if the path of navigation property in current scope is matching the <paramref name="bindingPath"/>.</returns>
-        public static bool MatchBindingPath(IEdmPathExpression bindingPath, List<ODataPathSegment> parsedSegments)
+        public static bool MatchBindingPath(IEdmPathExpression bindingPath, IList<ODataPathSegment> parsedSegments)
         {
             List<string> paths = bindingPath.PathSegments.ToList();
 
@@ -53,7 +54,7 @@ namespace Microsoft.OData
                 else if (segment is TypeSegment)
                 {
                     // May need match type if the binding path contains type cast.
-                    if (pathIndex >= 0 && paths[pathIndex].Contains("."))
+                    if (pathIndex >= 0 && paths[pathIndex].IndexOf('.') >= 0)
                     {
                         if (string.CompareOrdinal(paths[pathIndex], segment.EdmType.AsElementType().FullTypeName()) != 0)
                         {
@@ -63,9 +64,9 @@ namespace Microsoft.OData
                         pathIndex--;
                     }
                 }
-                else if (segment is EntitySetSegment
-                      || segment is SingletonSegment
-                      || segmentIsNavigationPropertySegment)
+                else if (segmentIsNavigationPropertySegment
+                      || segment is EntitySetSegment
+                      || segment is SingletonSegment)
                 {
                     // Containment navigation property in first if statement for NavigationPropertySegment.
                     break;
--- a/src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
+++ b/src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
@@ -83,6 +83,14 @@ namespace Microsoft.OData.UriParser
         }
 
         /// <summary>
+        /// Get the List of ODataPathSegments.
+        /// </summary>
+        internal IList<ODataPathSegment> Segments
+        {
+            get { return this.segments; }
+        }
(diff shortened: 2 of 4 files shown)


src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -82,6 +82,14 @@ public int Count
             get { return this.segments.Count; }
         }
 
+        /// <summary>
+        /// Get the List of ODataPathSegments.
+        /// </summary>
+        public List<ODataPathSegment> List
```
xuzhg: >List [](start = 38, length = 4)

It's meaningless to name it as "List"

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -25,7 +25,7 @@ public class ODataPath : IEnumerable<ODataPathSegment>
         /// <summary>
         /// The segments that make up this path.
         /// </summary>
-        private readonly IList<ODataPathSegment> segments;
```
xuzhg: what's the benefit to use "List<T>" not "IList<T>"?

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
xuzhg: >List<ODataPathSegment> [](start = 15, length = 22)

In my opinion, to expose List<T> from a public class is not a good design.

1. List<T> exposes many more members than the IList<T>
2. If we change the type of private filed "segments" as other IEnumerable<T> implementation, what's the design goal about this public property?

src/Microsoft.OData.Core/BindingPathHelper.cs
```diff
@@ -53,19 +54,19 @@ public static bool MatchBindingPath(IEdmPathExpression bindingPath, List<ODataPa
                 else if (segment is TypeSegment)
                 {
                     // May need match type if the binding path contains type cast.
-                    if (pathIndex >= 0 && paths[pathIndex].Contains("."))
```
xuzhg: Contains(...) vs IndexOf(...) ? What's the different? 
I prefer to "Contains(...)", it's more readable, maintainable.

src/Microsoft.OData.Core/BindingPathHelper.cs
```diff
@@ -43,7 +44,7 @@ public static bool MatchBindingPath(IEdmPathExpression bindingPath, List<ODataPa
                 // Containment navigation property or complex property in binding path.
                 if (segment is PropertySegment || (segmentIsNavigationPropertySegment && segment.TargetEdmNavigationSource is IEdmContainedEntitySet))
                 {
-                    if (pathIndex < 0 || string.CompareOrdinal(paths[pathIndex], segment.Identifier) != 0)
```
xuzhg: >CompareOrdinal [](start = 48, length = 14)

I don't know which one is better. But from [https://www.dotnetjalps.com/2007/11/what-is-diffrenece-between.html](https://www.dotnetjalps.com/2007/11/what-is-diffrenece-between.html), it seems CompareOrdinal is more faster. (I don't know). 

But, I prefer to "CompareOrdinal", it's more readable, maintainable.

xuzhg:
>     public class ODataPath : IEnumerable<ODataPathSegment>

Maybe we can consider to change this class:

public class ODataPath : IList<ODataPathSegment>

---
Refers to: src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs:23 in bb0d17e. [](commit_id = bb0d17eb7ee4effa974660ca1af586b32e78ac09, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
mikepizzo: Would ReadOnlyCollection<T> give you the functionality you need, without allowing the user to change the collection (which could mess things up)?  You can call segments.AsReadOnly() the first time it's accessed to get the ReadOnlyCollection and then return the same instance in subsequent requests to avoid additional allocations.

---
In reply to: [352913193](https://github.com/OData/odata.net/pull/1603#discussion_r352913193) [](ancestors = 352913193)

mikepizzo:
>     public class ODataPath : IEnumerable<ODataPathSegment>

+1

---
In reply to: [569818829](https://github.com/OData/odata.net/pull/1603#issuecomment-569818829) [](ancestors = 569818829)

---
Refers to: src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs:23 in bb0d17e. [](commit_id = bb0d17eb7ee4effa974660ca1af586b32e78ac09, deletion_comment = False)

src/Microsoft.OData.Core/BindingPathHelper.cs
KanishManuja-MS: IndexOf is more performant. I verified it using a console app and measured the numbers.

src/Microsoft.OData.Core/BindingPathHelper.cs
KanishManuja-MS: After experiments, CompareOrdinal was faster.

Issue #1611, opened 2019-12-13
Error messages returned from an OData V4 service throw an exception and no data is returned


ihtfp69:
If the odata service returns an error object like below, the library throws a DataServiceRequestException and no data is returned as a result. If the service returns a status code 404, it does not even attempt to parse the error. If I force the status code to 200, it parses the error, sets the Error property of the ChangeOperationResponse, but then changes the status code to 500 and throws the same error as before. In both cases, nothing gets returned.

`{
  "error": {
    "code": "",
    "message": "Unsupported functionality"
}`

### Assemblies affected

Microsoft.OData.Client 7.6.2
SaveResult.cs

Line 349:
	`throw new DataServiceRequestException(Strings.DataServiceException_GeneralError, ex, service);`

### Reproduce steps

Create a query that returns several entities. Make changes to some of them, mark them as updated and call SaveChangesAsync.

### Expected result

It should return a DataServiceResponse with a list of ChangeOperationResponse and it should include those that have their Error property set.

### Actual result

An exception is thrown and nothing at all gets returned except the JSON error message from the first PATCH result that returned an error. The responses from entities that were successfully updated or additional ones that produced errors are lost.

KenitoInc:
@ihtfp69 Have you tried this `SaveChangesAsync(SaveChangesOptions.BatchWithIndependentOperations)`
The `BatchWithIndependentOperations` flag saves each change independently in a batch request and returns a ChangeOperationResponse for all succesful and failed requests

KenitoInc:
@ihtfp69 Did you manage to try out the code above? We would like to close this issue

KenitoInc:
@ihtfp69 Closing this issue for now. You can reopen it if the above code doesn't work for you.

ihtfp69:
@KenitoInc sorry, I never got an email about your comment. I did try every SaveChangesOption and none of them worked. It always throws an exception on the thread instead of returning data.

KenitoInc:
@ihtfp69 Kindly share the sample code you are trying to run

PR #1613, opened 2019-12-18
Remove code throwing DataServiceRequestException so as to properly return ChangeOperationResponse objects

KenitoInc:
### Issues

*This pull request fixes issue #1611 .*

### Description
I have opted to remove the code throwing the DataServiceRequestException since errors are enumerated in the  operation responses returned by the DataServiceResponse class. This will allow both failed and successful operation responses to be returned. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/SaveResult.cs
+++ b/src/Microsoft.OData.Client/SaveResult.cs
@@ -307,41 +307,28 @@ namespace Microsoft.OData.Client
         {
             List<OperationResponse> responses = new List<OperationResponse>(this.cachedResponses != null ? this.cachedResponses.Count : 0);
             DataServiceResponse service = new DataServiceResponse(null, -1, responses, false /*isBatch*/);
-            Exception ex = null;
-
-            try
+            
+            foreach (CachedResponse response in this.cachedResponses)
             {
-                foreach (CachedResponse response in this.cachedResponses)
+                Descriptor descriptor = response.Descriptor;
+                this.SaveResultProcessed(descriptor);
+                OperationResponse operationResponse = new ChangeOperationResponse(response.Headers, descriptor);
+                operationResponse.StatusCode = (int)response.StatusCode;
+                if (response.Exception != null)
                 {
-                    Descriptor descriptor = response.Descriptor;
-                    this.SaveResultProcessed(descriptor);
-                    OperationResponse operationResponse = new ChangeOperationResponse(response.Headers, descriptor);
-                    operationResponse.StatusCode = (int)response.StatusCode;
-                    if (response.Exception != null)
-                    {
-                        operationResponse.Error = response.Exception;
-
-                        if (ex == null)
-                        {
-                            ex = response.Exception;
-                        }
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Client/SaveResult.cs
```diff
@@ -344,11 +344,6 @@ protected override DataServiceResponse HandleResponse()
                 ex = e;
             }
 
-            if (ex != null)
-            {
-                throw new DataServiceRequestException(Strings.DataServiceException_GeneralError, ex, service);
-            }
```
xuzhg: 1) so, why do you still keep the variable "ex" ?
2) Why do you still keep the try{}catch{} ?

xuzhg: at least one test case to cover your change

test/EndToEndTests/Tests/Client/Build.Desktop/TransportLayerTests/RequestMessageArgsTests.cs
```diff
@@ -130,7 +130,7 @@ public void HttpMergeIsNoLongerSupported()
             {
                 var ar2 = ctx.BeginSaveChanges(null, null).EnqueueWait(this);
                 ctx.EndSaveChanges(ar2);
-                Assert.Fail("Expected error not thrown");
+                // Assert.Fail("Expected error not thrown"); // Commented since errors should be added to COR not thrown
```
KanishManuja-MS: Can you add an assert for the error response? You can access the COR after end save changes right?

src/Microsoft.OData.Client/SaveResult.cs
```diff
@@ -320,11 +320,6 @@ protected override DataServiceResponse HandleResponse()
                     if (response.Exception != null)
                     {
                         operationResponse.Error = response.Exception;
-
-                        if (ex == null)
-                        {
-                            ex = response.Exception;
-                        }
```
KanishManuja-MS: Isn't this a breaking change? The existing clients may serialize error responses now. While I agree that throwing exception is not the right thing to do, this may lead to error responses being shown in places where it wasn't being shown.

odero: Spec: http://docs.oasis-open.org/odata/odata/v4.0/os/part1-protocol/odata-v4.0-os-part1-protocol.html#_Toc372793627

> The odata.continue-on-error preference on a batch request is used to request that, upon encountering a request within the batch that returns an error, the service return the error for that request and continue processing additional requests within the batch.
If not specified, upon encountering an error the service MUST return the error within the batch and stop processing additional requests within the batch.

Shouldn't the client then mimic the service's behavior such that if the request specifies the `odata.continue-on-error` preference and the service supports it then return error via COR otherwise throw the exception. Support for this can be implicitly determined by whether the entire batch request fails or not.

KenitoInc:
Closing this PR since the issue was closed

PR #1618, opened 2020-01-07
Made some of the public DataServiCeContext methods virtual to support mocking

ElizabethOkerio:
### Issues

*This pull request fixes issue #1250*

### Description

*Made some of the DataServiceContext public methods virtual to support mocking*

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceClientFormat.cs
+++ b/src/Microsoft.OData.Client/DataServiceClientFormat.cs
@@ -6,12 +6,17 @@
 
 namespace Microsoft.OData.Client
 {
+    #region namespaces
     using System;
     using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
+    using System.IO;
+    using System.Threading.Tasks;
+    using System.Xml;
     using Microsoft.OData;
     using Microsoft.OData.Edm;
-
+    using Microsoft.OData.Edm.Csdl;
+    #endregion
     /// <summary>
     /// Tracks the user-preferred format which the client should use when making requests.
     /// </summary>
@@ -80,14 +85,28 @@ namespace Microsoft.OData.Client
         {
             get
             {
-                if (serviceModel == null && LoadServiceModel != null)
+                if (serviceModel != null)
+                {
+                    return serviceModel;
+                }
+
+                if (LoadServiceModel != null)
                 {
                     serviceModel = LoadServiceModel();
                 }
-
+                else
+                {
+                    serviceModel = LoadServiceModelFromNetwork();
+                }
                 return serviceModel;
             }
         }
+
+        /// <summary>
+        /// Invoked during test cases to fake out network calls to get the metadata
+        /// returns a string that is passed to the csdl parser and is used to bypass the network call while testing
+        /// </summary>
+        internal Func<HttpWebRequestMessage> InjectMetadataHttpNetworkRequest { get; set; }
 
         /// <summary>
         /// Indicates that the client should use the efficient JSON format.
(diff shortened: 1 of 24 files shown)


xuzhg:
>         public bool Detach(object entity)

how about this method ?
 or others?

why don't make them as virtual?

---
Refers to: src/Microsoft.OData.Client/DataServiceContext.cs:2229 in 5234ce9. [](commit_id = 5234ce9252f2e13e1b24b96553e7084596468153, deletion_comment = False)

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2749,7 +2749,7 @@ internal BuildingRequestEventArgs CreateRequestArgsAndFireBuildingRequest(string
         /// <param name="languageDependentNamespace">Namespace of the type, can be different in VB than the fullNamespace.</param>
         /// <returns>Type that the name resolved to or null if none found.</returns>
         /// <remarks>Function was added for Portable Lib support to handle the differences in accessing the assembly of the context.</remarks>
-        protected internal Type DefaultResolveType(string typeName, string fullNamespace, string languageDependentNamespace)
```
xuzhg: >protected internal [](start = 8, length = 18)

I don't think to change this "internal" as public? What's the reason why we should do like this.

xuzhg:
>         public IEnumerable EndExecute(IAsyncResult asyncResult)

what about this method?

---
Refers to: src/Microsoft.OData.Client/DataServiceQuery.cs:78 in 5234ce9. [](commit_id = 5234ce9252f2e13e1b24b96553e7084596468153, deletion_comment = False)

xuzhg:
>         public bool IsComposable { get; private set; }

how about other public properties?

---
Refers to: src/Microsoft.OData.Client/DataServiceQueryOfT.cs:125 in 5234ce9. [](commit_id = 5234ce9252f2e13e1b24b96553e7084596468153, deletion_comment = False)

test/FunctionalTests/Microsoft.OData.Client.Tests/Microsoft.OData.Client.Tests.Netcore.csproj
```diff
@@ -15,6 +15,7 @@
     <PackageReference Include="FluentAssertions" Version="4.19.2" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
     <PackageReference Include="Microsoft.NETCore.Portable.Compatibility" Version="1.0.1" />
+    <PackageReference Include="Moq" Version="4.6.38-alpha" />
```
xuzhg: >"4.6.38-alpha" /> [](start = 44, length = 17)

why do we use the "alpha" version? please use the latest RTM version

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/DataServiceContextMethodMockSampleTest.cs
```diff
@@ -0,0 +1,33 @@
...
+using Moq;
+using Xunit;
+
+namespace Microsoft.OData.Client.Tests.ALinq
+{
+    /// <summary>
+    /// How to mock the DataServiceContext class methods using moq framework example.
+    /// </summary>
+
+    public class DataServiceContextMethodMockSampleTest
```
xuzhg: >DataServiceContextMethodMockSampleTest [](start = 17, length = 38)

I don't think the test class is meaningful.

What we are testing is that we can test some functionalities using mock, not test mock itself.

For example, we can test some method is actually called when we call a certain method. And we don't care about that called method returns, we can use mock to mock that called method.

tools/CustomMSBuild/Versioning.props
```diff
@@ -18,7 +18,7 @@
     -->
   <PropertyGroup>
     <VersionStartYear Condition="'$(VersionStartYear)' == ''">2017</VersionStartYear>
-    <VersionDateCode>$([System.Convert]::ToInt16('$([MSBuild]::Add(1, $([MSBuild]::Subtract($([System.DateTime]::Now.Year), $(VersionStartYear)))))$([System.DateTime]::Now.ToString("MMdd"))'))</VersionDateCode>
+    <VersionDateCode>$([System.Convert]::ToInt32('$([MSBuild]::Add(1, $([MSBuild]::Subtract($([System.DateTime]::Now.Year), $(VersionStartYear)))))$([System.DateTime]::Now.ToString("MMdd"))'))</VersionDateCode>
```
xuzhg: >ToInt32 [](start = 41, length = 7)

This problem is fixed. Please rebase your PR to the latest master branch

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/DataServiceContextMethodMockSampleTest.cs
```diff
@@ -0,0 +1,56 @@
...
+            DoSomething doSomething = new DoSomething() ;
+            string result = doSomething.SaveUser(mockObject.Object);
+            Assert.Equal(str, result);
+        }
+    }
+
+    /// <summary>
+    /// A sample implementation class with methods calling the DataserviceContext.cs methods
+    /// </summary>
+    public class DoSomething
```
odero: While I realize this is just a test class I'd avoid namings like `DoSomething` in publicly consumed code

ElizabethOkerio: fixed and well noted

test/FunctionalTests/Microsoft.OData.Client.Tests/Microsoft.OData.Client.Tests.Netcore.csproj
ElizabethOkerio: It is the only version that is compatible with the .Netcore project.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/DataServiceContextMethodMockSampleTest.cs
```diff
@@ -0,0 +1,56 @@
...
+            SaveUserDetails saveUserDetails = new SaveUserDetails() ;
+            string result = saveUserDetails.SaveUser(mockObject.Object);
+            Assert.Equal(str, result);
+        }
+    }
+
+    /// <summary>
+    /// A sample implementation class with methods calling the DataserviceContext.cs methods
+    /// </summary>
+    public class SaveUserDetails
```
wachugamaina: Class Names, variable names and property names ideally should be Nouns, only functions, actions and methods should be verbs.

PR #1621, opened 2020-01-08
One drive escape function for key as segment staring with colon

KanishManuja-MS:
### Issues

This pull request fixes the escape function uri. 

### Description

The previous implementation of escape function uri was not able to distinguish if a request intended to use an escape function or a key-value began with a  ':' when the key is expressed as a segment. Therefore, the parsing would fail for the requests with key as segment with value starting with ':' with an error that no escape function was found for the type. 

Now, we change how we parse segments and give precedence to an escape function if one exists for the type otherwise treat it as a key-value. 

This pull requests 
### Additional work necessary

Release notes need to mention these changes when merged.

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
@@ -10,6 +10,7 @@ using System.Collections.ObjectModel;
 using System.Diagnostics;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
+using System.Text;
 using System.Text.RegularExpressions;
 using Microsoft.OData.Edm;
 using Microsoft.OData.Edm.Vocabularies;
@@ -829,12 +830,15 @@ namespace Microsoft.OData.UriParser
         /// <param name="segmentText">The text of the segment.</param>
         private void CreateFirstSegment(string segmentText)
         {
+            if (segmentText[segmentText.Length - 1] == ':' && this.TryCreateEscapeFunctionSegment(segmentText))
+            {
+                return;
+            }
+
             string identifier;
             string parenthesisExpression;
             ExtractSegmentIdentifierAndParenthesisExpression(segmentText, out identifier, out parenthesisExpression);
-
             Debug.Assert(identifier != null, "identifier != null");
-
             // Look for well-known system resource points.
             if (this.IdentifierIs(UriQueryConstants.MetadataSegment, identifier))
             {
@@ -908,6 +912,101 @@ namespace Microsoft.OData.UriParser
             }
 
             this.CreateDynamicPathSegment(null, identifier, parenthesisExpression);
+        }
+
+        /// <summary>Creates a <see cref="ODataPathSegment"/> that an escape function can bind to, which can be a
+        /// a navigation source or an operation import it is the first segment to be created. Otherwise, it can be a 
+        /// filter segment, property segment, typecast segment, operation segment or a key segment.
+        /// </summary>
+        /// <param name="segmentText">The text of the segment.</param>
+        private bool BindSegmentBeforeEscapeFunction(string segmentText)
+        {
+            string identifier;
+            string parenthesisExpression;
+            ExtractSegmentIdentifierAndParenthesisExpression(segmentText, out identifier, out parenthesisExpression);
+
+            if (this.parsedSegments.Count == 0)
+            {
+                if (this.TryCreateSegmentForNavigationSource(identifier, parenthesisExpression))
+                {
+                    return true;
+                }
+
+                if (this.TryCreateSegmentForOperationImport(identifier, parenthesisExpression))
+                {
+                    return true;
+                }
+            }
+            else
+            {
+                if (this.TryCreateFilterSegment(segmentText))
+                {
+                    return true;
+                }
+
+                ODataPathSegment previous = this.parsedSegments[this.parsedSegments.Count - 1];
+
+                if (this.TryCreatePropertySegment(previous, identifier, parenthesisExpression))
+                {
+                    return true;
+                }
+
+                // Type cast
+                if (this.TryCreateTypeNameSegment(previous, identifier, parenthesisExpression))
+                {
+                    return true;
+                }
+
+                // Operation
+                if (this.TryCreateSegmentForOperation(previous, identifier, parenthesisExpression))
+                {
+                    return true;
+                }
+
+                // For KeyAsSegment, try to handle as key segment
+                if (this.configuration.UrlKeyDelimiter.EnableKeyAsSegment && this.TryHandleAsKeySegment(segmentText))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        private bool TryCreatePropertySegment(ODataPathSegment previous, string identifier, string parenthesisExpression)
+        {
+            // property if previous is single
+            if (previous.SingleResult)
+            {
+                // if its not one of the recognized special segments, then it must be a property, type-segment, or key value.
+                Debug.Assert(
+                    previous.TargetKind == RequestTargetKind.Resource
+                    || previous.TargetKind == RequestTargetKind.Dynamic,
+                    "previous.TargetKind(" + previous.TargetKind + ") can have properties");
+
+                if (previous.TargetEdmType == null)
+                {
+                    // A segment will correspond to a property in the object model;
+                    // if we are processing an open type, anything further in the
+                    // URI also represents an open type property.
+                    Debug.Assert(previous.TargetKind == RequestTargetKind.Dynamic, "For open properties, the target resource type must be null");
+                }
+                else
+                {
+                    // if the segment corresponds to a declared property, handle it
+                    // otherwise, fall back to type-segments, actions, and dynamic/open properties
+                    IEdmProperty projectedProperty;
+                    if (this.TryBindProperty(identifier, out projectedProperty))
+                    {
+                        CheckSingleResult(previous.SingleResult, previous.Identifier);
+                        this.CreatePropertySegment(previous, projectedProperty, parenthesisExpression);
+
+                        return true;
+                    }
+                }
+            }
+
+            return false;
         }
 
         /// <summary>
@@ -1014,11 +1113,6 @@ namespace Microsoft.OData.UriParser
                 bindingType = (previousSegment is EachSegment) ? previousSegment.TargetEdmType : previousSegment.EdmType;
             }
 
-            if (!String.IsNullOrEmpty(identifier) && identifier[0] == ':' && bindingType != null)
-            {
-                identifier = ResolveEscapeFunction(identifier, bindingType, configuration.Model, out parenthesisExpression);
-            }
-
             ICollection<OperationSegmentParameter> resolvedParameters;
             IEdmOperation singleOperation;
             if (!TryBindingParametersAndMatchingOperation(identifier, parenthesisExpression, bindingType, this.configuration, out resolvedParameters, out singleOperation))
@@ -1036,6 +1130,13 @@ namespace Microsoft.OData.UriParser
                 throw new ODataException(ODataErrorStrings.FunctionCallBinder_CallingFunctionOnOpenProperty(identifier));
             }
 
+            CreateOperationSegment(previousSegment, singleOperation, resolvedParameters, identifier, parenthesisExpression);
+
+            return true;
+        }
+
+        private void CreateOperationSegment(ODataPathSegment previousSegment, IEdmOperation singleOperation, ICollection<OperationSegmentParameter> resolvedParameters, string identifier, string parenthesisExpression)
+        {
             IEdmTypeReference returnType = singleOperation.ReturnType;
             IEdmEntitySetBase targetset = null;
 
@@ -1059,10 +1160,9 @@ namespace Microsoft.OData.UriParser
             };
 
             this.parsedSegments.Add(segment);
-
             this.TryBindKeySegmentIfNoResolvedParametersAndParenthesisValueExists(parenthesisExpression, returnType, resolvedParameters, segment);
 
-            return true;
+            return;
         }
 
         /// <summary>
@@ -1071,10 +1171,14 @@ namespace Microsoft.OData.UriParser
         /// <param name="text">The text for the next segment.</param>
         private void CreateNextSegment(string text)
         {
+            if (text[text.Length - 1] == ':' && this.TryCreateEscapeFunctionSegment(text))
+            {
+                return;
+            }
+
             string identifier;
             string parenthesisExpression;
             ExtractSegmentIdentifierAndParenthesisExpression(text, out identifier, out parenthesisExpression);
-
             /*
              * For Non-KeyAsSegment, try to handle it as a key property value, unless it was preceded by an escape - marker segment('$').
              * For KeyAsSegment, the following precedence rules should be supported[ODATA - 799]:
@@ -1123,39 +1227,13 @@ namespace Microsoft.OData.UriParser
                 return;
             }
 
-            // property if previous is single
-            if (previous.SingleResult)
+            if (this.TryCreatePropertySegment(previous, identifier, parenthesisExpression))
             {
-                // if its not one of the recognized special segments, then it must be a property, type-segment, or key value.
-                Debug.Assert(
-                    previous.TargetKind == RequestTargetKind.Resource
-                    || previous.TargetKind == RequestTargetKind.Dynamic,
-                    "previous.TargetKind(" + previous.TargetKind + ") can have properties");
-
-                if (previous.TargetEdmType == null)
-                {
-                    // A segment will correspond to a property in the object model;
-                    // if we are processing an open type, anything further in the
-                    // URI also represents an open type property.
-                    Debug.Assert(previous.TargetKind == RequestTargetKind.Dynamic, "For open properties, the target resource type must be null");
-                }
-                else
-                {
-                    // if the segment corresponds to a declared property, handle it
-                    // otherwise, fall back to type-segments, actions, and dynamic/open properties
-                    IEdmProperty projectedProperty;
-                    if (this.TryBindProperty(identifier, out projectedProperty))
-                    {
-                        CheckSingleResult(previous.SingleResult, previous.Identifier);
-                        this.CreatePropertySegment(previous, projectedProperty, parenthesisExpression);
-                        return;
-                    }
-                }
+                return;
             }
 
             // Type cast
-            if (text.IndexOf('.') >= 0 && // type-cast should use qualified type names
-                this.TryCreateTypeNameSegment(previous, identifier, parenthesisExpression))
+            if (this.TryCreateTypeNameSegment(previous, identifier, parenthesisExpression))
             {
                 return;
             }
@@ -1181,6 +1259,37 @@ namespace Microsoft.OData.UriParser
 
             // Dynamic property
             this.CreateDynamicPathSegment(previous, identifier, parenthesisExpression);
+        }
+
+        private bool TryCreateEscapeFunctionSegment(string segmentText)
+        {
+            Debug.Assert(segmentText[segmentText.Length - 1] == ':');
+
+            int numberOfSegmentsParsed = this.parsedSegments.Count;
+            string newSegmentText = segmentText.Substring(0, segmentText.Length - 1);
+
+            // Try to binding the segment optimisitically to which the escape function must bind to
+            if (newSegmentText.Length > 0 && !BindSegmentBeforeEscapeFunction(newSegmentText))
+            {
+                return false;
+            }
+
+            if (this.TryBindEscapeFunction())
+            {
+                return true;
+            }
+            else
+            {
+                // The caller of this function will try binding the original segment as key value.
+                // We need to roll back the optimistic binding that we did in this step.
+                while (this.parsedSegments.Count > numberOfSegmentsParsed)
+                {
+                    // Keep on poping the last segment. 
+                    this.parsedSegments.RemoveAt(this.parsedSegments.Count - 1);
+                }
+            }
+
+            return false;
         }
 
         /// <summary>
@@ -1218,6 +1327,39 @@ namespace Microsoft.OData.UriParser
             return projectedProperty != null;
         }
 
+        private bool TryBindEscapeFunction()
+        {
+            ODataPathSegment previous = null;
+            if (this.parsedSegments.Count > 0)
+            {
+                previous = this.parsedSegments[this.parsedSegments.Count - 1];
+            }
+
+            string newIdentifier, newParenthesisExpression;
+            bool anotherEscapeFunctionStarts;
+            IEdmFunction escapeFunction;
+            if (this.TryResolveEscapeFunction(previous, out newIdentifier, out newParenthesisExpression, out anotherEscapeFunctionStarts, out escapeFunction))
+            {
+                ICollection<FunctionParameterToken> splitParameters;
+                FunctionParameterParser.TrySplitOperationParameters(newParenthesisExpression, configuration, out splitParameters);
+                ICollection<OperationSegmentParameter> resolvedParameters = FunctionCallBinder.BindSegmentParameters(configuration, escapeFunction, splitParameters);
+                CreateOperationSegment(previous, escapeFunction, resolvedParameters, newIdentifier, newParenthesisExpression);
+                if (anotherEscapeFunctionStarts)
+                {
+                    // When we encounter an invalid escape function as a parameter, we should throw.
+                    // i.e. we should throw for entitySet(key):/ComposableEscapeFunctionPath::/InvalidEscapeFunction
+                    if (!TryBindEscapeFunction())
+                    {
+                        throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_ComposableEscapeFunctionShouldHaveValidParameter);
+                    }
+                }
+
+                return true;
+            }
+
+            return false;
+        }
+
         /// <summary>
         /// Tries to create a type name segment if the given identifier refers to a known type.
         /// </summary>
@@ -1227,6 +1369,11 @@ namespace Microsoft.OData.UriParser
         /// <returns>Whether or not a type segment was created for the identifier.</returns>
         private bool TryCreateTypeNameSegment(ODataPathSegment previous, string identifier, string parenthesisExpression)
         {
+            if (identifier.IndexOf('.') < 0)
+            {
+                return false;
+            }
+
             IEdmType targetEdmType;
             if (previous.TargetEdmType == null || (targetEdmType = UriEdmHelpers.FindTypeFromModel(this.configuration.Model, identifier, this.configuration.Resolver)) == null)
             {
@@ -1603,26 +1750,139 @@ namespace Microsoft.OData.UriParser
             throw new ODataException(Strings.PathParser_TypeCastOnlyAllowedInDerivedTypeConstraint(fullTypeName, kind, name));
         }
 
-        private static string ResolveEscapeFunction(string identifier, IEdmType bindingType, IEdmModel model, out string parenthesisExpression)
+        private bool TryResolveEscapeFunction(ODataPathSegment previous, out string qualifiedName, out string parenthesisExpression, out bool anotherEscapeFunctionStarts, out IEdmFunction function)
         {
-            Debug.Assert(identifier != null && identifier.Length >= 1 && identifier[0] == ':');
-            Debug.Assert(bindingType != null);
+            qualifiedName = null;
+            parenthesisExpression = null;
+            anotherEscapeFunctionStarts = false;
+            IEdmType bindingType = null;
+            IEdmModel model = configuration.Model;
+            function = null;
 
-            bool isComposableRequired = identifier.Length >= 2 && identifier[identifier.Length - 1] == ':';
-            IEdmFunction function = model.FindBoundOperations(bindingType)
-                .OfType<IEdmFunction>().FirstOrDefault(f => f.IsComposable == isComposableRequired && IsUrlEscapeFunction(model, f));
+            if (previous != null)
+            {
+                bindingType = previous.TargetEdmType;
+            }
+
+            if (bindingType == null || model == null)  // escape function is only for bound functions
+            {
+                return false;
+            }
+
+            IEnumerable<IEdmFunction> candidates = model.FindBoundOperations(bindingType).OfType<IEdmFunction>().Where(f => IsUrlEscapeFunction(model, f));
+
+            if (!candidates.HasAny())
+            {
+                return false;
+            }
+
+            string nextPiece = null;
+            StringBuilder sb = new StringBuilder();
+
+            while (this.TryGetNextSegmentText(out nextPiece))
+            {
+                if (sb.Length >= 1)
+                {
+                    sb.Append('/');
+                }
+
+                sb.Append(nextPiece);
+
+                if (nextPiece[nextPiece.Length - 1] == ':')
+                {
+                    break;
+                }
+            }
+
+            string identifier = sb.ToString();
+
+            if (identifier != null && identifier.Length >= 2 && identifier[identifier.Length - 2] == ':')
+            {
+                anotherEscapeFunctionStarts = true;
+            }
+
+            bool isComposableRequired = identifier != null && identifier.Length >= 1 && identifier[identifier.Length - 1] == ':';
+
+            function = FindBestMatchForEscapeFunction(candidates, isComposableRequired, bindingType);
+
             if (function == null)
             {
+                // We need to throw because we have consumed segments from the queue and we don't put them back. This is fair because early checks did show an escape function bound to the type. 
                 throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_NoBoundEscapeFunctionSupported(bindingType.FullTypeName()));
             }
 
-            if (function.Parameters == null || function.Parameters.Count() != 2 || !function.Parameters.ElementAt(1).Type.IsString())
+            parenthesisExpression = function.Parameters.ElementAt(1).Name + "='" + (isComposableRequired ? identifier.Substring(0, identifier.Length - 1) : identifier) + "'";
+            qualifiedName = function.FullName();
+
+            return true;
+        }
+
+        private static IEdmFunction FindBestMatchForEscapeFunction(IEnumerable<IEdmFunction> candidates, bool isComposable, IEdmType bindingType)
+        {
+            IEdmFunction bestCandidate = null;
+            foreach (IEdmFunction f in candidates)
             {
-                throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_EscapeFunctionMustHaveOneStringParameter(function.FullName()));
+                // If the composability of the above found function does not match, we cannot use the function
+                if (f.IsComposable != isComposable)
+                {
+                    continue;
+                }
+
+                IEdmOperationParameter firstParameter;
+                if (ParametersMatchEscapeFunction(f.Parameters, out firstParameter))
+                {
+                    // The first parameter of the function is not an exact match for the binding type then try to find a better match.
+                    if (firstParameter.Type.Definition == bindingType)
+                    {
+                        return f;
+                    }
+                    else if (bestCandidate != null)
+                    {
+                        // If the binding parameter for the  bestcandidate so far is a base type of first parameter of the current candidate
+                        // then we should use the more specific function, which is the current function. 
+                        if (bestCandidate.HasEquivalentBindingType(firstParameter.Type.Definition))
+                        {
+                            bestCandidate = f;
+                        }
+                    }
+                    else
+                    {
+                        bestCandidate = f;
+                    }
+                }
             }
 
-            parenthesisExpression = function.Parameters.ElementAt(1).Name + "='" + (isComposableRequired ? identifier.Substring(1, identifier.Length - 2) : identifier.Substring(1)) + "'";
-            return function.FullName();
+            return bestCandidate;
+        }
+
+        private static bool ParametersMatchEscapeFunction(IEnumerable<IEdmOperationParameter> parameters, out IEdmOperationParameter firstParameter)
+        {
+            firstParameter = null;
+            if (parameters == null)
+            {
+                return false;
+            }
+
+            int count = 0; 
+            foreach(IEdmOperationParameter p in parameters)
+            {
+                if (++count > 2)
+                {
+                    return false;
+                }
+
+                if (count == 1)
+                {
+                    firstParameter = p;
+                }
+
+                if (count == 2 && !p.Type.IsString())
+                {
+                    return false;
+                }
+            }
+
+            return true;
         }
 
         internal static bool IsUrlEscapeFunction(IEdmModel model, IEdmFunction function)
--- a/src/Microsoft.OData.Core/UriParser/Parsers/UriPathParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/UriPathParser.cs
@@ -75,7 +75,6 @@ namespace Microsoft.OData.UriParser
                 string[] uriSegments = uri.Segments;
 #endif
 
-                int escapedStart = -1;
                 List<string> segments = new List<string>();
                 for (int i = numberOfSegmentsToSkip; i < uriSegments.Length; i++)
                 {
@@ -99,57 +98,7 @@ namespace Microsoft.OData.UriParser
                         throw new ODataException(Strings.UriQueryPathParser_TooManySegments);
                     }
 
-                    // Handle the "root...::/{xyz}"
-                    if (segment.Length >= 2 && segment.EndsWith("::", StringComparison.Ordinal))
-                    {
-                        // It should be the terminal of the previous escape segment and the start of next escape segment.
-                        // Otherwise, it's an invalid Uri.
-                        if (escapedStart == -1)
-                        {
-                            throw new ODataException(Strings.UriQueryPathParser_InvalidEscapeUri(segment));
-                        }
-                        else
-                        {
-                            string value = String.Join("/", uriSegments, escapedStart, i - escapedStart + 1);
-                            segments.Add(":" + value.Substring(0, value.Length - 1)); // because the last one has "::", remove one.
-                            escapedStart = i + 1;
-                        }
-                    }
-                    else if (segment.Length >= 1 && segment[segment.Length - 1] == ':')
-                    {
-                        // root:/{abc}....
-                        if (escapedStart == -1)
-                        {
-                            if (segment != ":")
-                            {
-                                segments.Add(segment.Substring(0, segment.Length - 1)); // remove the last ':'
-                            }
-
-                            escapedStart = i + 1;
-                        }
-                        else
-                        {
-                            // root:/{abc}:....
-                            string escapedSegment = ":" + String.Join("/", uriSegments, escapedStart, i - escapedStart + 1); // the last has one ":";
-                            segments.Add(escapedSegment);
-                            escapedStart = -1;
-                        }
-                    }
-                    else
-                    {
-                        // if we didn't find a starting escape, the current segment is normal segment, accept it.
-                        // otherwise, it's part of the escape, skip it and process it when we find the ending delimiter.
-                        if (escapedStart == -1)
-                        {
-                            segments.Add(Uri.UnescapeDataString(segment));
-                        }
-                    }
-                }
-
-                if (escapedStart != -1 && escapedStart < uriSegments.Length)
-                {
-                    string escapedSegment = ":" + String.Join("/", uriSegments, escapedStart, uriSegments.Length - escapedStart);
-                    segments.Add(escapedSegment); // We should not use "segments.Add(Uri.UnescapeDataString(escapedSegment));" to keep the original string.
+                    segments.Add(Uri.UnescapeDataString(segment));
                 }
 
                 return segments.ToArray();
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
@@ -1183,6 +1183,124 @@ namespace Microsoft.OData.Tests.UriParser
             Assert.Equal("Name", proSegment.Property.Name);
         }
 
+        [Theory]
+        [InlineData("/entitySetEscaped/32:/photos%2F2018%2F%2fFebruary:/Name", "/entitySetEscaped/32/NS.ComposableFunction(arg='photos%2F2018%2F%2fFebruary')/Name", true)]
+        [InlineData("/entitySetEscaped/32:/:/Name", "/entitySetEscaped/32/NS.ComposableFunction(arg='')/Name", true)]
+        [InlineData("/entitySetEscaped/32/:/:/Name", "/entitySetEscaped/32/NS.ComposableFunction(arg='')/Name", true)]
+        [InlineData("/entitySetEscaped/32:/", "/entitySetEscaped/32/NS.NormalFunction(path='')", false)]
+        [InlineData("/entitySetEscaped/32/:/", "/entitySetEscaped/32/NS.NormalFunction(path='')", false)]
+        [InlineData("/entitySetEscaped/32/:", "/entitySetEscaped/32/NS.NormalFunction(path='')", false)]
+        [InlineData("/entitySetEscaped/32:/photos%2F2018%2F%2fFebruary", "/entitySetEscaped/32/NS.NormalFunction(path='photos%2F2018%2F%2fFebruary')", false)]
+        public void ParseEscapeFunctionWithColonInKeyValue(string escapeFunctionUri, string functionUri, bool isComposable)
+        {
+            // Arrange
+            IEdmModel model = GetEdmModelWithEscapeFunction(escape: true);
+
+            // Act
+            string fullUriString = ServiceRoot + functionUri;
+            ODataUriParser parser = new ODataUriParser(model, ServiceRoot, new Uri(fullUriString));
+            var functionPath = parser.ParsePath();
+
+            fullUriString = ServiceRoot + escapeFunctionUri;
+            parser = new ODataUriParser(model, ServiceRoot, new Uri(fullUriString));
+            var escapePath = parser.ParsePath();
+
+            // Assert
+            Assert.True(functionPath.Equals(escapePath));
+            Assert.Equal(functionPath.FirstOrDefault(s => s is KeySegment).Identifier, escapePath.FirstOrDefault(s => s is KeySegment).Identifier);
+            OperationSegment segment = escapePath.First(c => c is OperationSegment) as OperationSegment;
+            Assert.Equal(isComposable ? "NS.ComposableFunction": "NS.NormalFunction", segment.Operations.First().FullName());
+
+            if (isComposable)
+            {
+                PropertySegment proSegment = Assert.IsType<PropertySegment>(escapePath.Last());
+                Assert.Equal("Name", proSegment.Property.Name);
+            }
+        }
+
+        [Theory]
+        [InlineData("/entitySetEscaped/32:/Name", "/entitySetEscaped/32:/Name")]
+        [InlineData("/entitySetEscaped/:32:/Name", "/entitySetEscaped/:32:/Name")]
+        public void ParseKeyValueWithColonWithoutEscapeFunction(string escapeFunctionUri, string functionUri)
+        {
+            // Arrange
+            IEdmModel model = GetEdmModelWithEscapeFunction(escape: false);
+
+            // Act
+            string fullUriString = ServiceRoot + functionUri;
+            ODataUriParser parser = new ODataUriParser(model, ServiceRoot, new Uri(fullUriString));
+            var functionPath = parser.ParsePath();
+
+            fullUriString = ServiceRoot + escapeFunctionUri;
+            parser = new ODataUriParser(model, ServiceRoot, new Uri(fullUriString));
+            var escapePath = parser.ParsePath();
+
+            // Assert
+            Assert.True(functionPath.Equals(escapePath));
+            Assert.IsType<KeySegment>(escapePath.ElementAt(1));
+            PropertySegment proSegment = Assert.IsType<PropertySegment>(escapePath.Last());       
+            Assert.Equal("Name", proSegment.Property.Name);
+        }
+
+        [Theory]
+        [InlineData("/entitySetEscaped/32/NS.SpecialDrive:/ComposableParameter:", "/entitySetEscaped/32/NS.SpecialDrive/NS.SpecialOrders(path ='ComposableParameter')")]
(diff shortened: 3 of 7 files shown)


src/Microsoft.OData.Core/UriParser/Parsers/UriPathParser.cs
```diff
@@ -99,58 +99,58 @@ public virtual ICollection<string> ParsePathIntoSegments(Uri fullUri, Uri servic
...
+                    //    else
+                    //    {
+                    //        // root:/{abc}:....
+                    //        string escapedSegment = ":" + String.Join("/", uriSegments, escapedStart, i - escapedStart + 1); // the last has one ":";
+                    //        segments.Add(escapedSegment);
+                    //        escapedStart = -1;
+                    //    }
+                    //}
+                    //else
+                    //{
```
xuzhg: get rid of all these codes

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -86,6 +87,20 @@ internal static void ExtractSegmentIdentifierAndParenthesisExpression(string seg
                 identifier = segmentText;
                 parenthesisExpression = null;
             }
+            else if (segmentText[segmentText.Length - 1] == ':')
+            {
```
mikepizzo: add a comment describing logic -- is this to support syntax such as:
 entityset(key):/escape-function-path?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -86,6 +87,20 @@ internal static void ExtractSegmentIdentifierAndParenthesisExpression(string seg
                 identifier = segmentText;
                 parenthesisExpression = null;
             }
+            else if (segmentText[segmentText.Length - 1] == ':')
+            {
+
+                if (segmentText.Length < 2 ||
```
mikepizzo: remove empty blank line

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -360,12 +375,39 @@ private bool TryHandleAsKeySegment(string segmentText)
...
+                {
+                    this.TryBindEscapeFunction();
+                }
+
                 return true;
             }
 
+
+
             return false;
```
mikepizzo: remove extra blank lines

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -360,12 +375,39 @@ private bool TryHandleAsKeySegment(string segmentText)
             KeySegment previousKeySegment = this.FindPreviousKeySegment();
 
             KeySegment keySegment;
+
+            bool trybindingEscapeFunction = false;
+
+            if (segmentText.Length > 0 && segmentText[segmentText.Length - 1] == ':' && previous.EdmType != null)
```
mikepizzo: >segmentText.Length > 0 && [](start = 16, length = 25)

is this ever called with a zero-length segmentText, or should this just be an assert?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1218,6 +1322,34 @@ private bool TryBindProperty(string identifier, out IEdmProperty projectedProper
             return projectedProperty != null;
         }
 
+        private bool TryBindEscapeFunction()
+        {
+            ODataPathSegment previous = this.parsedSegments[this.parsedSegments.Count - 1];
+            IEdmType bindingType = null;
+            if (previous != null)
+            {
+                bindingType = (previous is EachSegment) ? previous.TargetEdmType : previous.EdmType;
```
mikepizzo: >(previous is EachSegment) [](start = 30, length = 25)

Do we have a test to validate this? In general, I don't think we should support an escape function (or any other function) following $each. The only thing that should validly follow $each is a bound action.

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
```diff
@@ -1305,22 +1289,6 @@ public void ParseEscapeFunctionUrlWithPropertyAndEscapeFunctionReturnsCorrectODa
             Assert.Equal("xyz/abc", ((ConstantNode)parameter.Value).Value);
         }
 
-        [Fact]
```
mikepizzo: Is this test no longer valid?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
```diff
@@ -1183,22 +1183,6 @@ public void ParseComposableEscapeFunctionUrlAndPropertyReturnsTheSameODataPath()
...
-            // Act
-            var fullUriString = ServiceRoot + "/root:/photos/2018/February";
-            var parser = new ODataUriParser(model, ServiceRoot, new Uri(fullUriString));
-            Action test = () => parser.ParsePath();
-
-            // Assert
-            var odataException = Assert.Throws<ODataException>(test);
-            Assert.Equal(ODataErrorStrings.RequestUriProcessor_NoBoundEscapeFunctionSupported("NS.OneDrive"), odataException.Message);
-        }
-
```
mikepizzo: What is the behavior for this url? It should still throw an invalid url, right? because root: is not a valid identifier?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/UriPathParserTests.cs
```diff
@@ -198,68 +198,6 @@ public void ParsePathIgnoresQueryStringEvenWhenItHasSlashes()
...
-        [InlineData("EntitySet('key'):/xyz::/perm", new[] { "EntitySet('key')", ":xyz:", ":perm" } )]
-        [InlineData("EntitySet('key'):/xyz/abc:", new[] { "EntitySet('key')", ":xyz/abc:" } )]
-        [InlineData("EntitySet('key%2fvalue'):/xyz", new[] { "EntitySet('key%2fvalue')", ":xyz" })]
-        [InlineData("EntitySet/key%2fvalue:/xyz", new[] { "EntitySet", "key%2fvalue", ":xyz" })]
-        [InlineData("EntitySet/key:%2fvalue:/xyz", new[] { "EntitySet", "key:%2fvalue", ":xyz" })]
-        [InlineData("EntitySet('key'):/xyz::/perm", new[] { "EntitySet('key')", ":xyz:", ":perm" })]
-        [InlineData("EntitySet('key'):/xyz:/:/perm", new[] { "EntitySet('key')", ":xyz:", ":perm" })]
-        [InlineData(":/xyz::/perm", new[] { ":xyz:", ":perm" })]
-        [InlineData(":/xyz:/:/perm", new[] { ":xyz:", ":perm" })]
-        [InlineData(":/xyz://///:/perm", new[] { ":xyz:", ":perm" })]
```
mikepizzo: Why are all variations of this test removed? Aren't most of these still valid? Any that are now invalid should have added negative tests.

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/UriPathParserTests.cs
```diff
@@ -198,68 +198,6 @@ public void ParsePathIgnoresQueryStringEvenWhenItHasSlashes()
...
-            var actual = this.pathParser.ParsePathIntoSegments(fullUrl, this.baseUri);
-
-            // Assert
-            Assert.Equal(expected, actual);
-        }
-
-        [Theory]
-        [InlineData("root::/abc", "root::")]
-        [InlineData("EntitySet('key')::/abc", "EntitySet('key')::")]
-        public void ParseInvalidEscapeUriPathShouldThrow(string pattern, string segment)
```
mikepizzo: Are these now both valid? What is the behavior?

mikepizzo:
> 

remove trailing space (and trailing blank line)

---
Refers to: src/Microsoft.OData.Core/UriParser/Parsers/UriPathParser.cs:117 in 267863a. [](commit_id = 267863afa1b0f2ece6d6dce9f9bd2cba60dd9e61, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
mikepizzo: >segmentText.Length > 0 [](start = 16, length = 22)

Should this be an assert? Didn't we validate that segmentText.Length > 0 in ExtractSegmentIdentifierAndParenthesisExpression on line 118? Do we not call ExtractSegmentIdentifierAndParenthesisExpression before calling TryHanbdleAsKeySegment?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -360,12 +375,39 @@ private bool TryHandleAsKeySegment(string segmentText)
...
+                IEdmFunction function = configuration.Model.FindBoundOperations(previous.EdmType).OfType<IEdmFunction>().FirstOrDefault(f => IsUrlEscapeFunction(configuration.Model, f));
+                if (function != null)
+                {
+                    trybindingEscapeFunction = true;
+                    segmentText = segmentText.Substring(0, segmentText.Length - 1);
+
+                    // If nothing to bind for key then bind the escape function directly. 
+                    if (segmentText.Length == 0)
+                    {
+                        return this.TryBindEscapeFunction();
```
mikepizzo: You have already done a somewhat expensive operation to figure out what the escape function is. You shouldn't have to redo that work when you call TryBindEscapeFunction. Can you split TryBindEscapeFunction into two pieces, the first the determines the function and the second that takes that information and does the actual binding, and call the second function from here and line 403?

mikepizzo: >return [](start = 24, length = 6)

is this the right semantics, to base the return of TryHandleAsKeySegment on the return from TryBindEscapeFunction? don't you know at this point that TryBindEscapeFunction will return true (since you already verified the escape function), and wouldn't it be an error (not false for evaluating this segment as a key) if TryBindEscapeFunction returned false?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1603,26 +1735,74 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
...
         {
-            Debug.Assert(identifier != null && identifier.Length >= 1 && identifier[0] == ':');
-            Debug.Assert(bindingType != null);
+            qualifiedName = null;
+            parenthesisExpression = null;
+            anotherEscapeFunctionStarts = false;
+
+            if (bindingType == null)  // escape function is only for bind function
+
+            {
```
mikepizzo: remove blank line

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1603,26 +1735,74 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
...
             if (function.Parameters == null || function.Parameters.Count() != 2 || !function.Parameters.ElementAt(1).Type.IsString())
             {
                 throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_EscapeFunctionMustHaveOneStringParameter(function.FullName()));
             }
 
-            parenthesisExpression = function.Parameters.ElementAt(1).Name + "='" + (isComposableRequired ? identifier.Substring(1, identifier.Length - 2) : identifier.Substring(1)) + "'";
-            return function.FullName();
+            parenthesisExpression = function.Parameters.ElementAt(1).Name + "='" + (isComposableRequired ? identifier.Substring(0, identifier.Length - 1) : identifier.Substring(0)) + "'";
+            qualifiedName = function.FullName();
+
```
mikepizzo: remove space at end of line.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1603,26 +1735,74 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
...
-            bool isComposableRequired = identifier.Length >= 2 && identifier[identifier.Length - 1] == ':';
-            IEdmFunction function = model.FindBoundOperations(bindingType)
-                .OfType<IEdmFunction>().FirstOrDefault(f => f.IsComposable == isComposableRequired && IsUrlEscapeFunction(model, f));
             if (function == null)
             {
-                throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_NoBoundEscapeFunctionSupported(bindingType.FullTypeName()));
+                return false;
+                // should be an exception instead. 
+                //throw ExceptionUtil.CreateBadRequestError("Bound function is not composable.");
             }
```
mikepizzo: is this work to be done?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -922,6 +964,14 @@ private bool TryCreateSegmentForNavigationSource(string identifier, string paren
             IEdmEntitySet targetEdmEntitySet;
             IEdmSingleton targetEdmSingleton;
 
+            bool tryBindingEscapeFunction = false;
+
+            if (identifier[identifier.Length - 1] == ':')
+            {
+                identifier = identifier.Substring(0, identifier.Length - 1);
+                tryBindingEscapeFunction = true;
+            }
```
mikepizzo: same logic appears multiple times -- perhaps encapsulate in a common function?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -937,6 +987,21 @@ private bool TryCreateSegmentForNavigationSource(string identifier, string paren
             {
                 this.parsedSegments.Add(segment);
                 this.TryBindKeyFromParentheses(parenthesisExpression);
+
+                if (tryBindingEscapeFunction && !this.TryBindEscapeFunction())
+                {
```
mikepizzo: So the logic is, for each segment that could be followed by an escape function, if the identifier ends in ':', try binding the identifier up to the colon as that segment type, and see if the you can bind a function to it. If not, revert binding the segment and continue.  Right?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: This is the remains from a previous iteration. In my previous logic, I had permitted empty segment text. This is not needed now. 

---
In reply to: [366148775](https://github.com/OData/odata.net/pull/1621#discussion_r366148775) [](ancestors = 366148775)

mikepizzo:
>             // Type cast

Need to add code (and test) for a type cast segment followed by an escape.  i.e., Orders/NS.Order:/xyz/abc: and /Customers(2)/NS.Customer:/xyz/abc:

---
Refers to: src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs:1251 in 267863a. [](commit_id = 267863afa1b0f2ece6d6dce9f9bd2cba60dd9e61, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1062,6 +1130,16 @@ private bool TryCreateSegmentForOperation(ODataPathSegment previousSegment, stri
 
             this.TryBindKeySegmentIfNoResolvedParametersAndParenthesisValueExists(parenthesisExpression, returnType, resolvedParameters, segment);
 
+            if (tryBindingEscapeFunction && !this.TryBindEscapeFunction())
```
mikepizzo: >!this.TryBindEscapeFunction() [](start = 44, length = 29)

shouldn't this return false if bindescapefunction returns false, so the segment can be evaluated as a key?  i.e., "/Customers(2)/NS.FindOrderComposable(orderName='foo'):/xyz/abc:" currently succeeds, ignoring everything after Customers(2), and returns a path with 2 segments (Customers entity set and key) if there is no escape function defined that takes an order.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: I have removed the check of $each at this point for escape functions if it should not be supported. 

---
In reply to: [366143523](https://github.com/OData/odata.net/pull/1621#discussion_r366143523) [](ancestors = 366143523)

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/UriPathParserTests.cs
xuzhg: +1, we can change the expected to make the test valid again. 
Because all the patterns are valid case, we can use this test cases to see the ParsePath works expected for these patterns. 


---
In reply to: [366144313](https://github.com/OData/odata.net/pull/1621#discussion_r366144313) [](ancestors = 366144313)

mikepizzo:
>         private bool TryCreateSegmentForOperationImport(string identifier, string parenthesisExpression)

This needs to handle a function import followed by an escape function. 

---
Refers to: src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs:1018 in 267863a. [](commit_id = 267863afa1b0f2ece6d6dce9f9bd2cba60dd9e61, deletion_comment = False)

tools/CustomMSBuild/Versioning.props
```diff
@@ -43,7 +43,7 @@
     <T4VersionFullSemantic>$(T4VersionMajor).$(T4VersionMinor).$(T4VersionBuildNumber)</T4VersionFullSemantic>
     <VersionT4Semantic>$(T4VersionFullSemantic)</VersionT4Semantic>
   </PropertyGroup>
-  
+
   <!-- For NuGet Packages -->
```
xuzhg: revert

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
KanishManuja-MS: Correct. Added the test back. 

---
In reply to: [366144030](https://github.com/OData/odata.net/pull/1621#discussion_r366144030) [](ancestors = 366144030)

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
```diff
@@ -1348,7 +1316,7 @@ internal static IEdmModel GetCustomerOrderEdmModelWithEscapeFunction()
             EdmModel model = new EdmModel();
 
             EdmEntityType orderType = new EdmEntityType("NS", "Order");
-            orderType.AddKeys(orderType.AddStructuralProperty("Id", EdmPrimitiveTypeKind.Int32));
```
xuzhg: >Int32 [](start = 89, length = 5)

why do we have to change it?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/UriPathParserTests.cs
KanishManuja-MS: Yes, both are valid for this step. These will throw an Unrecogonized Path Exception in the E2E scenario while binding. 

---
In reply to: [366144414](https://github.com/OData/odata.net/pull/1621#discussion_r366144414) [](ancestors = 366144414)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: Instead of splitting it into two functions I decided to have this take in an optional function.

---
In reply to: [366150654](https://github.com/OData/odata.net/pull/1621#discussion_r366150654) [](ancestors = 366150654)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: Yes, you understanding is correct regarding what I am trying to do here. However, the segment may successfully bind as type which may not have escape function bound but the segment can then be bound to a type which may have an escape function. Therefore, it is incorrect to say that it will only bind to the key segment but yes, it is very likely that it will bind to the key segment. 

Even if I take that assumption, the code will be simplified but not to a great extent.

---
In reply to: [366156392](https://github.com/OData/odata.net/pull/1621#discussion_r366156392) [](ancestors = 366156392)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: It can throw an error even if there is an escape function bound to this type. It will never return false as of now. More details in comments, let me know how you would like me to proceed. 
I changed the code as I thought was the best possible way to handle it. 
PS: I dislike the int passed to tell how many segments to remove from the list.

---
In reply to: [366150964](https://github.com/OData/odata.net/pull/1621#discussion_r366150964) [](ancestors = 366150964)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -86,6 +87,20 @@ internal static void ExtractSegmentIdentifierAndParenthesisExpression(string seg
                 identifier = segmentText;
                 parenthesisExpression = null;
             }
+            else if (segmentText[segmentText.Length - 1] == ':')
+            {
+
```
Sreejithpin: Remove Blank Space

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -86,6 +87,20 @@ internal static void ExtractSegmentIdentifierAndParenthesisExpression(string seg
...
+            else if (segmentText[segmentText.Length - 1] == ':')
+            {
+
+                if (segmentText.Length < 2 ||
+                    segmentText[segmentText.Length - 2] != ')')
+                {
+                    throw ExceptionUtil.CreateSyntaxError();
+                }
+
+
```
Sreejithpin: Remove Blank Space

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -86,6 +87,20 @@ internal static void ExtractSegmentIdentifierAndParenthesisExpression(string seg
...
+                identifier = segmentText.Substring(0, parenthesisStart) + ":";
+                parenthesisExpression =
+                   segmentText.Substring(parenthesisStart + 1, segmentText.Length - identifier.Length - 2);
```
Sreejithpin: Is the identifier used only for getting the length? if so it will be parenthesisStart+1 , right? which can directly be used. Or am I missing something?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -360,12 +375,39 @@ private bool TryHandleAsKeySegment(string segmentText)
...
+
+                if (trybindingEscapeFunction)
+                {
+                    this.TryBindEscapeFunction();
+                }
+
                 return true;
             }
 
+
```
Sreejithpin: Remove blank spaces

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
```diff
@@ -1183,6 +1183,94 @@ public void ParseComposableEscapeFunctionUrlAndPropertyReturnsTheSameODataPath()
             Assert.Equal("Name", proSegment.Property.Name);
         }
 
+        [Theory]
+        [InlineData("/entitySetEscaped/32:/photos%2F2018%2F%2fFebruary:/Name", "/entitySetEscaped/32/NS.ComposableFunction(arg='photos%2F2018%2F%2fFebruary')/Name", true)]
```
xuzhg: >%2F2018%2F%2 [](start = 49, length = 12)

does unescaped "/" work as well? or Do we only allowe customer to use "escaped"?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/UriPathParserTests.cs
```diff
@@ -200,38 +200,40 @@ public void ParsePathIgnoresQueryStringEvenWhenItHasSlashes()
...
+        [InlineData("EntitySet('key%2fvalue'):/xyz", new[] { "EntitySet('key/value'):", "xyz" })]
+        [InlineData("EntitySet/key%2fvalue:/xyz", new[] { "EntitySet", "key/value:", "xyz" })]
+        [InlineData("EntitySet/key:%2fvalue:/xyz", new[] { "EntitySet", "key:/value:", "xyz" })]
+        [InlineData("EntitySet('key'):/xyz::/perm", new[] { "EntitySet('key'):", "xyz::", "perm" })]
+        [InlineData("EntitySet('key'):/xyz:/:/perm", new[] { "EntitySet('key'):", "xyz:", ":", "perm" })]
+        [InlineData(":/xyz::/perm", new[] { ":", "xyz::", "perm" })]
+        [InlineData(":/xyz:/:/perm", new[] { ":", "xyz:", ":", "perm" })]
+        [InlineData(":/xyz://///:/perm", new[] { ":", "xyz:", ":", "perm" })]
+        [InlineData("root::/abc", new[] { "root::", "abc" })]
+        [InlineData("EntitySet('key')::/abc", new[] { "EntitySet('key')::", "abc" })]
```
xuzhg: >[InlineData("EntitySet('key')::/abc", new[] { "EntitySet('key')::", "abc" })] [](start = 8, length = 77)

can you add the new pattern with colon in "key" and key as segment

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
```diff
@@ -1183,6 +1183,94 @@ public void ParseComposableEscapeFunctionUrlAndPropertyReturnsTheSameODataPath()
...
+            var functionPath = parser.ParsePath();
+
+            fullUriString = ServiceRoot + escapeFunctionUri;
+            parser = new ODataUriParser(model, ServiceRoot, new Uri(fullUriString));
+            var escapePath = parser.ParsePath();
+
+            // Assert
+            Assert.True(functionPath.Equals(escapePath));
+
+            OperationSegment segment = escapePath.First(c => c is OperationSegment) as OperationSegment;
```
xuzhg: >OperationSegment segment = escapePath.First(c => c is OperationSegment) as OperationSegment; [](start = 12, length = 92)

maybe add codes to check the "KeySegment" has the correct parsed value?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
```diff
@@ -1183,6 +1183,94 @@ public void ParseComposableEscapeFunctionUrlAndPropertyReturnsTheSameODataPath()
...
+            var functionPath = parser.ParsePath();
+
+            fullUriString = ServiceRoot + escapeFunctionUri;
+            parser = new ODataUriParser(model, ServiceRoot, new Uri(fullUriString));
+            var escapePath = parser.ParsePath();
+
+            // Assert
+            Assert.True(functionPath.Equals(escapePath));
+
+            PropertySegment proSegment = Assert.IsType<PropertySegment>(escapePath.Last());
```
xuzhg: > Assert.IsType<PropertySegment>(escapePath.Last()); [](start = 40, length = 51)

add the codes to verify the key segment with correct value. for example: the key value has the ":" in .

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -360,9 +376,47 @@ private bool TryHandleAsKeySegment(string segmentText)
...
 
             KeySegment keySegment;
+
+            bool trybindingEscapeFunction = false;
+            IEdmFunction function = null;
+
+            if (segmentText[segmentText.Length - 1] == ':' && previous.EdmType != null)
+            {
+
+                IEdmType bindingType = segmentText.Length == 1 ? previous.EdmType : previous.EdmType.AsElementType();
```
mikepizzo: >IEdmType bindingType = segmentText.Length == 1 ? previous.EdmType : previous.EdmType.AsElementType(); [](start = 16, length = 101)

the logic for this line seems obscure; please describe in a comment.

this says that, if the segment contains only a colon, and the previous segment was a collection, return the collection type; otherwise return the element type.  Right? What situation is this trying to handle?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
KanishManuja-MS: Yes, it would work as well for escaped functions.

---
In reply to: [367060428](https://github.com/OData/odata.net/pull/1621#discussion_r367060428) [](ancestors = 367060428)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: @Sreejithpin We usually follow closing parenthesis with a blank line if it is the last else clause. Look at some of the other existing code

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: It is an output parameter as well.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -86,6 +87,21 @@ internal static void ExtractSegmentIdentifierAndParenthesisExpression(string seg
                 identifier = segmentText;
                 parenthesisExpression = null;
             }
+            else if (segmentText[segmentText.Length - 1] == ':')
+            {
+                if (segmentText.Length < 2 ||
+                    segmentText[segmentText.Length - 2] != ')')
```
KanishManuja-MS: What about open paren in a key value?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1062,6 +1141,11 @@ private bool TryCreateSegmentForOperation(ODataPathSegment previousSegment, stri
 
             this.TryBindKeySegmentIfNoResolvedParametersAndParenthesisValueExists(parenthesisExpression, returnType, resolvedParameters, segment);
```
KanishManuja-MS: >             [](start = 0, length = 12)

What about (paren1)(keyParen) scenario?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: Not changing at this point to mitigate risk. The approach I was considering was to pass the entire segment text as identifier if the parentheses didn't match (here if text did not end in parenthesis), however, I was not confident on how TryCreateSegment functions would handle such input.
To reduce scope, I am keeping this out of the PR and creating separate issue for that. 

---
In reply to: [367191420](https://github.com/OData/odata.net/pull/1621#discussion_r367191420) [](ancestors = 367191420)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -910,6 +933,90 @@ private void CreateFirstSegment(string segmentText)
             this.CreateDynamicPathSegment(null, identifier, parenthesisExpression);
         }
 
+        /// <summary>Creates the first <see cref="ODataPathSegment"/> for a request.</summary>
```
mikepizzo: Update comment

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -910,6 +933,90 @@ private void CreateFirstSegment(string segmentText)
             this.CreateDynamicPathSegment(null, identifier, parenthesisExpression);
         }
 
+        /// <summary>Creates the first <see cref="ODataPathSegment"/> for a request.</summary>
+        /// <param name="segmentText">The text of the segment.</param>
+        private bool CreateSegmentForEscapeFunction(string segmentText)
```
mikepizzo: >CreateSegmentForEscapeFunction [](start = 21, length = 30)

This is not really creating a segment for an escape function, right? It's creating a segment that an escape function may be bound to?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -910,6 +933,90 @@ private void CreateFirstSegment(string segmentText)
...
+                        IEdmProperty projectedProperty;
+                        if (this.TryBindProperty(identifier, out projectedProperty))
+                        {
+                            CheckSingleResult(previous.SingleResult, previous.Identifier);
+                            this.CreatePropertySegment(previous, projectedProperty, parenthesisExpression);
+
+                            return true;
+                        }
+                    }
+                }
```
mikepizzo: Maybe put this logic in a separate TryCreatePropertySegment to reduce duplicate code?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -910,6 +933,90 @@ private void CreateFirstSegment(string segmentText)
...
+
+                // Type cast
+                if (segmentText.IndexOf('.') >= 0 && // type-cast should use qualified type names
```
mikepizzo: >segmentText.IndexOf('.') >= 0 [](start = 20, length = 29)

should we move this check into TryCreateTypeNameSegment?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1183,6 +1269,35 @@ private void CreateNextSegment(string text)
             this.CreateDynamicPathSegment(previous, identifier, parenthesisExpression);
         }
 
+        private bool TryCreateEscapeFunctionSegment(string segmentText)
+        {
+            int numberOfSegmentsParsed = this.parsedSegments.Count;
+            string newSegmentText = segmentText.Substring(0, segmentText.Length - 1);
```
xuzhg: >string newSegmentText = segmentText.Substring(0, segmentText.Length - 1); [](start = 12, length = 73)

this line is abrupt. Maybe add a comment or a Debug.Assert(...) to make sure the input segment text has a ":" at the end?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1183,6 +1269,35 @@ private void CreateNextSegment(string text)
             this.CreateDynamicPathSegment(previous, identifier, parenthesisExpression);
         }
 
+        private bool TryCreateEscapeFunctionSegment(string segmentText)
+        {
+            int numberOfSegmentsParsed = this.parsedSegments.Count;
```
xuzhg: >int numberOfSegmentsParsed = this.parsedSegments.Count; [](start = 12, length = 55)

line 1274 can move to Line 1290, right?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1621,8 +1862,10 @@ private static string ResolveEscapeFunction(string identifier, IEdmType bindingT
                 throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_EscapeFunctionMustHaveOneStringParameter(function.FullName()));
             }
 
-            parenthesisExpression = function.Parameters.ElementAt(1).Name + "='" + (isComposableRequired ? identifier.Substring(1, identifier.Length - 2) : identifier.Substring(1)) + "'";
-            return function.FullName();
+            parenthesisExpression = function.Parameters.ElementAt(1).Name + "='" + (isComposableRequired ? identifier.Substring(0, identifier.Length - 1) : identifier.Substring(0)) + "'";
```
mikepizzo: >identifier.Substring(0) [](start = 156, length = 23)

isn't this just "identifier"?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -910,6 +914,89 @@ private void CreateFirstSegment(string segmentText)
             this.CreateDynamicPathSegment(null, identifier, parenthesisExpression);
         }
 
+        /// <summary>Binds a <see cref="ODataPathSegment"/> for an escape function segment.</summary>
+        /// <param name="segmentText">The text of the segment.</param>
+        private bool BindSegmentForEscapeFunction(string segmentText)
```
xuzhg: >BindSegmentForEscapeFunction [](start = 21, length = 28)

maybe name it as "BindEscapeFunctionSegment", follow up the others

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -910,6 +914,89 @@ private void CreateFirstSegment(string segmentText)
...
+            {
+                if (this.TryCreateSegmentForNavigationSource(identifier, parenthesisExpression))
+                {
+                    return true;
+                }
+
+                if (this.TryCreateSegmentForOperationImport(identifier, parenthesisExpression))
+                {
+                    return true;
+                }
```
xuzhg: here looks duplicated a lot of codes, should be just call "CreateFirstSegment"... and in else to call "CreateNexteSegment"?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1059,7 +1141,7 @@ private bool TryCreateSegmentForOperation(ODataPathSegment previousSegment, stri
             };
 
             this.parsedSegments.Add(segment);
-
+
```
xuzhg: revert

xuzhg:
I have concerns about the duplicated codes in BindSegmentForEscapeFunction. It will double the cost to maintain. Others look good to me. So i approved it. Thanks.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1002,25 +1111,21 @@ private void TryBindKeySegmentIfNoResolvedParametersAndParenthesisValueExists(st
         /// <param name="previousSegment">The previous segment before the operation to be invoked.</param>
         /// <param name="identifier">The name of the segment</param>
         /// <param name="parenthesisExpression">The query portion</param>
+        /// <param name="bindingType">Type to which the operation is bound. It is an optional parameter that can be evalulated from the previous segment if it is null.</param>
         /// <returns>Whether or not the identifier referred to an action.</returns>
```
mikepizzo: In the case of the escape function, you actually know more than the binding type -- you know the actual specific function overload that you want to bind, parameters to pass, etc.  Rather than passing in (possibly null) values to this method, I was suggesting that you factor this method into two.  The first part of this method has the current signature (without the added bindingType), and it determines the binding type and specific function (i.e., everything down to 1143).  And then this method calls a new method (also called by TryBindEscapeFunction) that takes the previous segment, resolved function, and resolved parameters and does the bind.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1218,6 +1359,44 @@ private bool TryBindProperty(string identifier, out IEdmProperty projectedProper
...
+            IEdmType bindingType = null;
+            if (previous != null)
+            {
+                bindingType = previous.EdmType;
+            }
+
+            string newIdentifier, newParenthesisExpression;
+            bool anotherEscapeFunctionStarts = false;
+
+            if (this.TryResolveEscapeFunction(bindingType, configuration.Model, out newIdentifier, out newParenthesisExpression, out anotherEscapeFunctionStarts))
```
mikepizzo: this.configuration.Model, to be clear where configuration is coming from?  Note that you don't really have to pass this, since TryResolveEscapeFunction is an instance method and has access to the model through the same configuration property.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: Correct. 

---
In reply to: [369792392](https://github.com/OData/odata.net/pull/1621#discussion_r369792392) [](ancestors = 369792392)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1603,14 +1783,75 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
             throw new ODataException(Strings.PathParser_TypeCastOnlyAllowedInDerivedTypeConstraint(fullTypeName, kind, name));
         }
 
-        private static string ResolveEscapeFunction(string identifier, IEdmType bindingType, IEdmModel model, out string parenthesisExpression)
+        private bool TryResolveEscapeFunction(IEdmType bindingType, IEdmModel model, out string qualifiedName, out string parenthesisExpression, out bool anotherEscapeFunctionStarts)
         {
```
mikepizzo: Note that this function was originally written to be called within TryCreateSegmentForOperation, so its inputs/outputs may be a little different now that you are calling this before trying to bind the operation (you now have the actual function you need to bind).  So think about what input/output parameters make sense in this new usage, and whether it should be called from TryBindEscapeFunction or the logic should be merged with TryBindEscapeFunction.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1086,13 +1191,14 @@ private void CreateNextSegment(string text)
              * Treat as a key.
              */
 
+
+            ODataPathSegment previous = this.parsedSegments[this.parsedSegments.Count - 1];
             // $value
```
mikepizzo: Why move this?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1183,6 +1295,35 @@ private void CreateNextSegment(string text)
             this.CreateDynamicPathSegment(previous, identifier, parenthesisExpression);
         }
 
+        private bool TryCreateEscapeFunctionSegment(string segmentText)
+        {
```
mikepizzo: assert that segmentText ends in ":"?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: I would have agreed with you. I had two thoughts stopping me - 
1. Either I change the parameters of the function to take in the segment text or do the same check on identifier. I am worried as there might be a case where two are not the same.
2. Adding another function to the call stack on the hot path will add to COGS. I have a better idea for performance, which is to check for this in extractsegmentidentifier method, I will add that in other PR. 

---
In reply to: [369793950](https://github.com/OData/odata.net/pull/1621#discussion_r369793950) [](ancestors = 369793950)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: Nope. BindSegmentForEscapeFunction changes the list of segments and it's count. 

---
In reply to: [369798577](https://github.com/OData/odata.net/pull/1621#discussion_r369798577) [](ancestors = 369798577)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1603,14 +1783,75 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
...
+            }
+
+            string identifier = sb.ToString();
+
+            if (identifier != null && identifier.Length >= 2 && identifier[identifier.Length - 2] == ':')
+            {
+                anotherEscapeFunctionStarts = true;
+            }
+
+            bool isComposableRequired = identifier.Length >= 1 && identifier[identifier.Length - 1] == ':';
```
mikepizzo: >identifier.Length [](start = 40, length = 17)

add check for identifier != null.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: You mean why move the TryCreateEscapeFunctionSegment? I did that to not ExtractIdentifierAndParenthesis proactively for escape function segment. If it is successful, we may not need to at this point. 

---
In reply to: [369818816](https://github.com/OData/odata.net/pull/1621#discussion_r369818816) [](ancestors = 369818816)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1603,14 +1783,75 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
...
+
+            if ((function.Parameters.FirstOrDefault().Type.Definition != bindingType) || (isComposableRequired && !function.IsComposable) || (!isComposableRequired && function.IsComposable))
```
mikepizzo: add comment describing logic.

mikepizzo:
>             }

if function == null (for example, because there were escape functions that matched the binding parameter, but they didn't meet the composability requirements), shouldn't we return false, rather than throw here?

---
Refers to: src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs:1858 in da28656. [](commit_id = da28656ac2836f639bd22ecb85da0517edd2203a, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1621,8 +1862,10 @@ private static string ResolveEscapeFunction(string identifier, IEdmType bindingT
                 throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_EscapeFunctionMustHaveOneStringParameter(function.FullName()));
             }
```
mikepizzo: isn't this already validated in IsUrlEscapeFunction?  If so, maybe just assert here?

KanishManuja-MS: Nope, IsUrlEscapeFunction just returns the functions based on the annotation. Although it is unlikely that this will ever get thrown if people annotate their escape functions carefully. 

---
In reply to: [369831446](https://github.com/OData/odata.net/pull/1621#discussion_r369831446) [](ancestors = 369831446)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1603,14 +1783,75 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
...
+                        function = null;
+                    }
+                }
+            }
+
+            // If no exact match found for binding type, use base type to find bound functions. 
+            if (function == null)
+            {
+                function = model.FindBoundOperations(bindingType).OfType<IEdmFunction>().FirstOrDefault(f => f.IsComposable == isComposableRequired && IsUrlEscapeFunction(model, f));
+            }
```
mikepizzo: Yikes! This is the third time we've called a fairly expensive FindBoundOperations and run a fairly expensive (and similar) LINQ query over the results. Any way we can optimize this? Maybe first determine the list of candidates:
   List<IEdmFunction> candidates = model.FindBoundOperations(bindingType).OfType<IEdmFunction>().Where(f=>IsUrlEscapeFunction(model, f)).ToList();
and then pick the best from that (presumably rather small) list?
   -exact match of bindingType and composability
   -any (base) type and composability
   -return false? (I think we decided composability had to exactly match?)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
mikepizzo: See my comments in iteration 22. I still think we can do a little refactoring to make the code cleaner and more performant.

---
In reply to: [366662603](https://github.com/OData/odata.net/pull/1621#discussion_r366662603) [](ancestors = 366662603,366150654)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
mikepizzo: Interesting... only case where segmentText would be different from identifier would be if it included parens, right?  In that case, shouldn't we be be checking identifier, rather than segmentText anyway?

---
In reply to: [369824859](https://github.com/OData/odata.net/pull/1621#discussion_r369824859) [](ancestors = 369824859,369793950)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
mikepizzo: See iteration 22: you had moved ODataPathSegment previous = this.parsedSegments[this.parsedSegments.Count - 1]; from line 1202 (immediately before using it to check previous.TargetKind == RequestTargetKind.Primitive, where is is first used) to line 1195 (before TryCreateValueSegment). It looks like it moved back in iteration 23.

---
In reply to: [369827716](https://github.com/OData/odata.net/pull/1621#discussion_r369827716) [](ancestors = 369827716,369818816)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: slightly different but did what we talked about. 

---
In reply to: [369837925](https://github.com/OData/odata.net/pull/1621#discussion_r369837925) [](ancestors = 369837925)

KanishManuja-MS:
I agree. Can't think of a better solution though. 

---
In reply to: [577390017](https://github.com/OData/odata.net/pull/1621#issuecomment-577390017) [](ancestors = 577390017)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
KanishManuja-MS: That is what the previous iteration was doing. It added a lot of checks to existing cases. I did not want to degrade existing performance in the hot path.

---
In reply to: [369804646](https://github.com/OData/odata.net/pull/1621#discussion_r369804646) [](ancestors = 369804646)

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1218,6 +1327,40 @@ private bool TryBindProperty(string identifier, out IEdmProperty projectedProper
...
+                    // i.e. we should throw for entitySet(key):/ComposableEscapeFunctionPath::/InvalidEscapeFunction
+                    if (!TryBindEscapeFunction())
+                    {
+                        throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_ComposableEscapeFunctionShouldHaveValidParameter);
+                    }
+                }
+
+                return true;
+
+            }
```
mikepizzo: nit: remove blank line

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1603,26 +1751,99 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
...
+            parenthesisExpression = function.Parameters.ElementAt(1).Name + "='" + (isComposableRequired ? identifier.Substring(0, identifier.Length - 1) : identifier) + "'";
+            qualifiedName = function.FullName();
+
+            return true;
+        }
+
+        private static IEdmFunction FindBestMatchForEscapeFunction(IEnumerable<IEdmFunction> candidtates, bool isComposable, IEdmType bindingType)
+        {
+            IEdmFunction bestCandidate = null;
+            foreach (IEdmFunction f in candidtates)
```
mikepizzo: >candidtates [](start = 39, length = 11)

nit: spelling

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1603,26 +1751,99 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
...
             {
-                throw ExceptionUtil.CreateBadRequestError(ODataErrorStrings.RequestUriProcessor_EscapeFunctionMustHaveOneStringParameter(function.FullName()));
+                // If the composability of the above found function does not match, we cannot use the function
+                if (f.IsComposable != isComposable)
+                {
+                    continue;
+                }
+
+                if (f.Parameters != null && f.Parameters.Count() == 2 && f.Parameters.ElementAt(1).Type.IsString())
```
mikepizzo: >Parameters.Count() [](start = 46, length = 18)

nit: rather than repeatedly re-enumerating, probably more efficient to do ToList; then you can index for the first and second parameters.

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -1603,26 +1751,99 @@ private static void VerifyDerivedTypeConstraints(IEdmModel model, IEdmVocabulary
...
+                {
+                    // The first parameter of the function is not an exact match for the binding type then try to find a better match.
+                    if (f.Parameters.FirstOrDefault().Type.Definition == bindingType)
+                    {
+                        return f;
+                    }
+                    else
+                    {
+                        bestCandidate = f;
```
mikepizzo: You are resetting the bestCandidate each time, which means that, if there is not an exact match, you'll pick the last one that matches (which, *I think*, is likely to be the base type, rather than the most derived type).  You could instead check to see if the current candidate has the current bestCandidate (if any) as a base type (recursively) and, if so, make the current candidate your new bestCandidate.

Issue #1570, opened 2019-10-16
oDataClient.cs too large to use


fbruemmer73:
Hello,

I'm using the client generator to generate an odata client for D365FO. This used to work in the past. However, since D365FO is constantly evolving and growing, so does the number of data entities and the size of the metadata. I believe we have reached a point, where the generated client code is too large to be compiled by Visual Studio. The only way to make this work is to remove parts of the metadata manually. This however is very error-prone and takes forever.

It would be nice to be able to select a specific number of data entities for client code generation in a future release..

paulodero:
@fbruemmer73 I would recommend you use [OData Connected service](https://marketplace.visualstudio.com/items?itemName=laylaliu.ODataConnectedService) to generate client code since the tool is now up to date and has a feature to generate entity classes in multiple files. Please also watch out future release of connected service which will come  with a feature to store service metadata in a different file instead of inside the generated code itself. The coming release will also allow you to select entities to generate code. All these features will make it a lot easier to generate code for a large metadata with many entities.

PR #1631, opened 2020-01-14
Allow OData Code generator to generate multiple source code files

paulodero:
### Issues

*This pull request fixes issue #1570.*

### Description

*Allow OData Code generator to generate multiple source code files.*

- [x] *Test cases added*

### Additional work necessary


src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,195 @@
...
+    public virtual String GetCustomToolNamespace(String fileName) {
+        return null;
+    }
+
+    public virtual String DefaultProjectNamespace {
+        get { return null; }
+    }
+
+    protected bool IsFileContentDifferent(String fileName, String newContent) {
+        return !(File.Exists(fileName) && File.ReadAllText(fileName) == newContent);
```
odero:
```suggestion
        return File.Exists(fileName) && File.ReadAllText(fileName) != newContent;
```
Easier to read

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,195 @@
...
+        set {
+            if (CurrentBlock != null)
+                EndBlock();
+            if (value != null)
+                value.Start = template.Length;
+            currentBlock = value;
+        }
+    }
+
+    private class VSManager: FilesManager {
```
odero:
```suggestion
    private class VSManager : FilesManager {
```

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,195 @@
...
+
+        public override String GetCustomToolNamespace(string fileName) {
+            return dte.Solution.FindProjectItem(fileName).Properties.Item("CustomToolNamespace").Value.ToString();
+        }
+
+        public override void Process(bool split) {
+            if (templateProjectItem.ProjectItems == null)
+                return;
+            base.Process(split);
+            projectSyncAction.EndInvoke(projectSyncAction.BeginInvoke(generatedFileNames, null, null));
```
odero: What does this line do?
Why not just
`projectSyncAction.Invoke(generatedFileNames, null, null)`
or
`projectSyncAction(generatedFileNames, null, null)`

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,195 @@
...
+                String fileName = Path.Combine(outputPath, block.Name);
+                String content = headerText + template.ToString(block.Start, block.Length) + footerText;
+                generatedFileNames.Add(fileName);
+                CreateFile(fileName, content);
+                template.Remove(block.Start, block.Length);
+            }
+        }
+    }
+
+    protected virtual void CreateFile(String fileName, String content) {
```
odero:
```suggestion
    protected virtual void CreateFile(string fileName, string content) {
```

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,195 @@
...
+        protected override void CreateFile(String fileName, String content) {
+            if (IsFileContentDifferent(fileName, content)) {
+                CheckoutFileIfRequired(fileName);
+                File.WriteAllText(fileName, content);
+            }
+        }
+
+        internal VSManager(ITextTemplatingEngineHost host, StringBuilder template)
+            : base(host, template) {
+            var hostServiceProvider = (IServiceProvider) host;
```
odero:
```suggestion
            var hostServiceProvider = host as IServiceProvider;
```
Wouldn't it be better to do this so that in case the casting fails `null` is returned instead of exception being thrown - which is what I think you are trying to do here

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,195 @@
...
+            if (hostServiceProvider == null)
+                throw new ArgumentNullException("Could not obtain IServiceProvider");
+            dte = (EnvDTE.DTE) hostServiceProvider.GetService(typeof(EnvDTE.DTE));
```
odero:
```suggestion
            dte = hostServiceProvider.GetService(typeof(EnvDTE.DTE)) as EnvDTE.DTE;
```
Same here

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,195 @@
...
+        }
+
+        private static void ProjectSync(EnvDTE.ProjectItem templateProjectItem, IEnumerable<String> keepFileNames) {
+            var keepFileNameSet = new HashSet<String>(keepFileNames);
+            var projectFiles = new Dictionary<String, EnvDTE.ProjectItem>();
+            var originalFilePrefix = Path.GetFileNameWithoutExtension(templateProjectItem.get_FileNames(0)) + ".";
+            foreach(EnvDTE.ProjectItem projectItem in templateProjectItem.ProjectItems)
+                projectFiles.Add(projectItem.get_FileNames(0), projectItem);
+
+            foreach(var pair in projectFiles)
```
odero: Consistency in using either var or type inside `foreach`
```suggestion
            foreach(KeyValuePair<string, EnvDTE.ProjectItem> pair in projectFiles)
```

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,195 @@
...
+                if (!keepFileNames.Contains(pair.Key) && !(Path.GetFileNameWithoutExtension(pair.Key) + ".").StartsWith(originalFilePrefix))
```
odero:
```suggestion
                if (!keepFileNameSet.Contains(pair.Key) && !(Path.GetFileNameWithoutExtension(pair.Key) + ".").StartsWith(originalFilePrefix))
```

odero: I don't get why you need to do add the "." here and in `originalFilePrefix`:
`!(Path.GetFileNameWithoutExtension(pair.Key) + ".").StartsWith(originalFilePrefix))`
Why not just
`Path.GetFileNameWithoutExtension(pair.Key) != originalFilePrefix`

src/CodeGen/ODataT4CodeGenerator.tt
```diff
@@ -28,6 +28,10 @@ public static class Configuration
 	// This flag indicates whether to ignore unexpected elements and attributes in the metadata document and generate
 	// the client code if any. The value must be set to true or false.
 	public const bool IgnoreUnexpectedElementsAndAttributes = true;
+
+    //This files indicates whether to generate the files into multiple files or single.
```
odero: Is this comment still relevant? It's not linked to any field declaration

src/CodeGen/ODataT4CodeGenerator.tt
```diff
@@ -28,6 +28,10 @@ public static class Configuration
 	// This flag indicates whether to ignore unexpected elements and attributes in the metadata document and generate
 	// the client code if any. The value must be set to true or false.
 	public const bool IgnoreUnexpectedElementsAndAttributes = true;
+
+    //This files indicates whether to generate the files into multiple files or single.
+    //If set to true then multiple files will be generated. Otherwise only a single file is generated.
```
odero:
```suggestion
    // If set to true then multiple files will be generated. Otherwise only a single file is generated.
```

src/CodeGen/ODataT4CodeGenerator.tt
paulodero: It is relevant. It describes the field SplitGeneratedFileIntoMultipleFiles.

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
paulodero: Either way an exception is thrown.

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
paulodero: Either way an exception is thrown. if a `null` is returned `ArgumentNullException` is thrown.

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
paulodero: Your suggested change will return `false` if the file does not exist. That is not what we expect.

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,195 @@
...
+    public void EndBlock() {
+        if (CurrentBlock == null)
+            return;
+        CurrentBlock.Length = template.Length - CurrentBlock.Start;
+        if (CurrentBlock != header && CurrentBlock != footer)
+            files.Add(CurrentBlock);
+        currentBlock = null;
+    }
+
+    public virtual void Process(bool split) {
```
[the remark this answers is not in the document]
paulodero: Any suggestions?

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
paulodero: Its Async process

odero: In this case there's actually no point in attempting to make it async since EndInvoke will block immediately

src/CodeGen/ODataT4CodeGenTemplate.ttinclude
```diff
@@ -0,0 +1,1333 @@
...
+
+    internal void WriteNamespaces()
+    {
+        foreach(string fullNamespace in context.NamespacesInModel)
+        {
+            this.WriteNamespace(fullNamespace);
+        }
+    }
+
+    internal void WriteNamespace(string fullNamespace)
```
wachugamaina: This function is really big, and I think it can be broken down to smaller functions.

src/CodeGen/ODataT4CodeGenTemplate.ttinclude
```diff
@@ -0,0 +1,1333 @@
...
+            if (opeartion.IsBound)
+            {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    internal void WriteEntityContainer(IEdmEntityContainer container, string fullNamespace)
```
wachugamaina: This function can also be broken down into smaller functions.

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,180 @@
...
+            get {
+                return templateProjectItem.ContainingProject.Properties.Item("DefaultNamespace").Value.ToString();
+            }
+        }
+
+        public override String GetCustomToolNamespace(string fileName) {
+            return dte.Solution.FindProjectItem(fileName).Properties.Item("CustomToolNamespace").Value.ToString();
+        }
+
+        public override void Process(bool split) {
```
wachugamaina: What are we processing?

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,180 @@
...
+                if (!keepFileNameSet.Contains(pair.Key) && !(Path.GetFileNameWithoutExtension(pair.Key) + ".").StartsWith(originalFilePrefix))
+                    pair.Value.Delete();
+
+            foreach(String fileName in keepFileNameSet)
+                if (!projectFiles.ContainsKey(fileName))
+                    templateProjectItem.ProjectItems.AddFromFile(fileName);
+        }
+
+        private void CheckoutFileIfRequired(String fileName) {
+            var sc = dte.SourceControl;
```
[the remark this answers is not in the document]
paulodero: variable changed to `sourceControl`

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,180 @@
...
+            files.Add(CurrentBlock);
+        currentBlock = null;
+    }
+
+    public virtual void Process(bool split) {
+        if (split) {
+            EndBlock();
+            string headerText = template.ToString(header.Start, header.Length);
+            String footerText = template.ToString(footer.Start, footer.Length);
+            String outputPath = Path.GetDirectoryName(host.TemplateFile);
```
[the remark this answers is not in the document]
paulodero: changed to `string`

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
paulodero: changed to GenerateFiles

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
```diff
@@ -0,0 +1,180 @@
...
+        if (CurrentBlock != header && CurrentBlock != footer)
+            files.Add(CurrentBlock);
+        currentBlock = null;
+    }
+
+    public virtual void Process(bool split) {
+        if (split) {
+            EndBlock();
+            string headerText = template.ToString(header.Start, header.Length);
+            String footerText = template.ToString(footer.Start, footer.Length);
```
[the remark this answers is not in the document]
paulodero: changed to `string`

src/CodeGen/ODataT4CodeGenFilesManager.ttinclude
paulodero: Changed to synchronous

xuzhg:
@paulodero If it's ready for review, please label it as "Read for review"

src/CodeGen/ODataT4CodeGenTemplate.ttinclude
paulodero: Fixed

src/CodeGen/ODataT4CodeGenTemplate.ttinclude
paulodero: Fixed

ElizabethOkerio:
The generated .cs file with all the entities has some spaces at the beginning of the file. Remove the unnecessary spacing.

habbes:
@paulodero since many of the of the new filenames have the same prefix (i.e.`ODataT4CodeGen`), maybe you should move them to their own subfolder and name the files based on their class names (e.g. `ODataT4CodeGen\FilesManager.cs` instead of `ODataT4CodeGenFilesManager.cs`)

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTests.cs
```diff
@@ -113,6 +131,17 @@ public void CodeGenSimpleEdmx()
             ODataT4CodeGeneratorTestDescriptors.Simple.Verify(code, false/*isCSharp*/, true/*useDSC*/);
         }
 
+        [TestMethod]
+        public void CodeGenSimpleEdmxMultipleFiles()
+        {
+            SplitFiles = true;
```
habbes: I think `SplitFiles` should be passed as a config option to the code generator like the way the other config options are handled. What's the purpose for setting is a class member instead of a function parameter?

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTests.cs
```diff
@@ -113,6 +131,17 @@ public void CodeGenSimpleEdmx()
             ODataT4CodeGeneratorTestDescriptors.Simple.Verify(code, false/*isCSharp*/, true/*useDSC*/);
         }
 
+        [TestMethod]
+        public void CodeGenSimpleEdmxMultipleFiles()
+        {
+            SplitFiles = true;
+            CompileGeneratedCode = false;
```
habbes: What's the purpose of setting `CompilerGeneratedCode` to `false` here?

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTests.cs
```diff
@@ -113,6 +131,17 @@ public void CodeGenSimpleEdmx()
...
 
+        [TestMethod]
+        public void CodeGenSimpleEdmxMultipleFiles()
+        {
+            SplitFiles = true;
+            CompileGeneratedCode = false;
+            string code = CodeGenWithT4Template(ODataT4CodeGeneratorTestDescriptors.SimpleMultipleFiles.Metadata, null, true, false);
+            ODataT4CodeGeneratorTestDescriptors.SimpleMultipleFiles.Verify(code, true/*isCSharp*/, false/*useDSC*/);
+            SplitFiles = false;
+            CompileGeneratedCode = true;
```
habbes: I think it would be better to pass these arguments to the relevant methods instead of setting the class members to some value before the test case and to another after the test case.

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTestDescriptors.cs
```diff
@@ -196,6 +196,16 @@ private static string LoadContentFromBaseline(string key)
...
         };
         #endregion
+        #region SimpleMultipleFiles
+        public static string EdmxSimpleMultipleFiles = LoadContentFromBaseline("SimpleMultipleFiles.xml");
+        public static string SimpleMultipleFilesCSharp = LoadContentFromBaseline("SimpleMultipleMain.cs");
+        public static ODataT4CodeGeneratorTestsDescriptor SimpleMultipleFiles = new ODataT4CodeGeneratorTestsDescriptor()
+        {
+            Metadata = EdmxSimpleMultipleFiles,
+            ExpectedResults = new Dictionary<string, string>() { { ExpectedCSharp, SimpleMultipleFilesCSharp }, { ExpectedCSharpUseDSC, SimpleCSharpUseDSC }, { ExpectedVB, SimpleVB }, { ExpectedVBUseDSC, SimpleVBUseDSC } },
+            Verify = (code, isCSharp, useDSC) => VerifyGeneratedCode(code, SimpleMultipleFiles.ExpectedResults, isCSharp, useDSC, "SimpleMultipleFiles"),
```
habbes: Does this actual verify that multiple files were generated as expected? It seems to me that it's only verifying `SimpleMultipleMain.cs` was, what about `SimpleMultipleFilesInclude.cs`?

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTests.cs
paulodero: trues

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTests.cs
paulodero: At this point I wanted to content of the file without compiling the code.

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTests.cs
paulodero: True

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTestDescriptors.cs
paulodero: Correct. It shows the main file does not contain all the classes. Multiple files in this case is implied by the correct content of the main file.

src/CodeGen/ODataT4CodeGenerator.tt
```diff
@@ -28,6 +28,10 @@ public static class Configuration
 	// This flag indicates whether to ignore unexpected elements and attributes in the metadata document and generate
 	// the client code if any. The value must be set to true or false.
 	public const bool IgnoreUnexpectedElementsAndAttributes = true;
+
+    //This files indicates whether to generate the files into multiple files or single.
+    //If set to true then multiple files will be generated. Otherwise only a single file is generated.
+    public const bool SplitGeneratedFileIntoMultipleFiles = false;
```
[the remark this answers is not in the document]
paulodero: `GenerateMultipleFiles ` would be more appropriate

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTestDescriptors.cs
habbes: I don't think that correct content of the main file implies that the other files were generated correctly. It may be the case that the other files were truncated, or were mixed up in one file instead of separate file, or some other behaviour. I think it would be worthwhile to also test the other files, and better yet to have at least 2 entity types to at least verify that the files are split into  separate file per class, and each with the correct content.

paulodero: I understand your concerns. Changing the implementation.

src/CodeGen/ODataT4CodeGenerator.tt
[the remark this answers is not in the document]
paulodero: Fixed

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTests.cs
paulodero: Fixed

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTestDescriptors.cs
paulodero: Fixed

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTests.cs
paulodero: Fixed

src/CodeGen/ODataT4CodeGenerator.tt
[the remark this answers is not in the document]
paulodero: Fixed

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTestDescriptors.cs
habbes: I have not been able to spot the tests that verify that multiple files were generated correctly, in this case the `SimpleMultipleFilesInclude.cs` still does not seem to have tests. Also if the test csdl could contain at least 2 entity types in order to demonstrate that the files are split one per type, that would be great in my opinion.

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/ODataT4CodeGeneratorTests.cs
habbes: I also think it ought to be refactored, I find that the current pattern (adding more optional parameters to the `.Verify()` method) is not sustainable, or readable. I think passing an on options object with all the relevant options explicitly set would be better. But I think that's the subject of a separate discussion/PR.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/ResourceBinderTests.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Xunit;
+
+namespace Microsoft.OData.Client.Tests.ALinq
+{
+    class ResourceBinderTests
+    {
+        [Fact]
+        public void VerifyAnalyzeCountMethodDoesNotModifyOriginalExpression()
+        {
+            //ResourceSetExpression rse = new ResourceSetExpression(null, rse.Source, rse.MemberExpression, rse.ResourceType, rse.ExpandPaths, rse.CountOption, rse.CustomQueryOptions, rse.Projection, rse.ResourceTypeAs, rse.UriVersion, rse.OperationName, rse.OperationParameters, rse.IsAction);
```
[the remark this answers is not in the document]
paulodero: removed.

paulodero:
This PR is being closed since the implementation has been done on [OData Connected Service v 0.6.1](https://marketplace.visualstudio.com/items?itemName=laylaliu.ODataConnectedService) and above. It is therefore recommended to use OData Connected Service for generating client code for OData services.

PR #1633, opened 2020-01-15
Change conversion of Int16 to UInt16.

bdebaere:
Instead of using [System.Convert]::ToInt16, [System.Convert]::ToUInt16 can be used.

Related commit: https://github.com/OData/odata.net/commit/3e30e05190dbb38f5ce54e175ee10174d7391349

### DIFF
--- a/tools/CustomMSBuild/Versioning.props
+++ b/tools/CustomMSBuild/Versioning.props
@@ -18,7 +18,7 @@
     -->
   <PropertyGroup>
     <VersionStartYear Condition="'$(VersionStartYear)' == ''">2020</VersionStartYear>
-    <VersionDateCode>$([System.Convert]::ToInt16('$([MSBuild]::Add(1, $([MSBuild]::Subtract($([System.DateTime]::Now.Year), $(VersionStartYear)))))$([System.DateTime]::Now.ToString("MMdd"))'))</VersionDateCode>
+    <VersionDateCode>$([System.Convert]::ToUInt16('$([MSBuild]::Add(1, $([MSBuild]::Subtract($([System.DateTime]::Now.Year), $(VersionStartYear)))))$([System.DateTime]::Now.ToString("MMdd"))'))</VersionDateCode>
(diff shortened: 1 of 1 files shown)


tools/CustomMSBuild/Versioning.props
```diff
@@ -18,7 +18,7 @@
     -->
   <PropertyGroup>
     <VersionStartYear Condition="'$(VersionStartYear)' == ''">2020</VersionStartYear>
-    <VersionDateCode>$([System.Convert]::ToInt16('$([MSBuild]::Add(1, $([MSBuild]::Subtract($([System.DateTime]::Now.Year), $(VersionStartYear)))))$([System.DateTime]::Now.ToString("MMdd"))'))</VersionDateCode>
+    <VersionDateCode>$([System.Convert]::ToUInt16('$([MSBuild]::Add(1, $([MSBuild]::Subtract($([System.DateTime]::Now.Year), $(VersionStartYear)))))$([System.DateTime]::Now.ToString("MMdd"))'))</VersionDateCode>
```
xuzhg: ToInt16 can't work at your side?

bdebaere: @xuzhg It works, but this will last twice as long.

paulodero: I think its better to change the function to "ToUInt16(...) Just like @bdebaere has done or to "ToInt32()". This ensures no future conflict of VersionDateCode value.

xuzhg: ToInt16 or ToUInt16 both has the limitation. Sreejith is working on to change the project structure. I'd like to change this logic also in the new project structure.

bdebaere: @xuzhg I understand that you want to change the project structure so as to solve this completely. But why not make it a little better in the meantime? I don't see a reason why they should be mutually exclusive, as the possibility exists that this project structure rework will get postponed.

xuzhg: @bdebaere @paulodero I'd agree that this change is a better improvement. It's fine to me to merge it.

Issue #1623, opened 2020-01-09
IN operator fails on strings with commas in them


kenneth-hansen:
ODataUriParser.ParseFilter throws if any of the IN operator's string parameters contains commas.

### Assemblies affected

7.6.2 (but probably all versions)

### Reproduce steps

Just create a filter ala: "?$filter=propertyName in ('a','x,y,z')"

### Expected result

Should parse normally.

### Actual result

ODataUriParser.ParseFilter throws an exception.

Microsoft.OData.ODataException
  HResult=0x80131509
  Message=String item should be single/double quoted: ''x'.
  Source=Microsoft.OData.Core
  StackTrace:
   at Microsoft.OData.UriParser.InBinder.NormalizeStringItem(String str)
   at Microsoft.OData.UriParser.InBinder.NormalizeCollectionItems(String bracketLiteralText, NormalizeFunction normalizeFunc)
   at Microsoft.OData.UriParser.InBinder.GetCollectionOperandFromToken(QueryToken queryToken, IEdmTypeReference expectedType, IEdmModel model)
   at Microsoft.OData.UriParser.InBinder.BindInOperator(InToken inToken, BindingState state)
   at Microsoft.OData.UriParser.MetadataBinder.Bind(QueryToken token)
   at Microsoft.OData.UriParser.FilterBinder.BindFilter(QueryToken filter)
   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseFilterImplementation(String filter, ODataUriParserConfiguration configuration, ODataPathInfo odataPathInfo)
   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseFilter()

### Additional detail

Problem seems to be in: https://github.com/OData/odata.net/blob/2cf5247da1fde1bfde63d3daf8acdd56ec8a156e/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs#L142
It splits on comma with no regard to whether it is inside a string or not

The line that throws:
https://github.com/OData/odata.net/blob/2cf5247da1fde1bfde63d3daf8acdd56ec8a156e/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs#L178

xuzhg:
Fixed and merged. Thanks.

PR #1635, opened 2020-01-16
Fix issue #1623: In operators fails on string collection with commas in the string value

xuzhg:
…n the string value

### Issues

*This pull request fixes issue #1623.*

### Description

* a filter clause like `?$filter=propertyName in ('a','x,y,z')` can't work because the old codes simply split the literal using ','.

This PR fix this issue

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -34,14 +34,6 @@ namespace Microsoft.OData.UriParser
         }
 
         /// <summary>
-        /// Delegate for a function that normalizes a string item representing a certain type.
-        /// Each type should define a different implementation of the delegate.
-        /// </summary>
-        /// <param name="item">The item to be normalized.</param>
-        /// <returns>Normalized string of the item.</returns>
-        private delegate string NormalizeFunction(string item);
-
-        /// <summary>
         /// Binds an In operator token.
         /// </summary>
         /// <param name="inToken">The In operator token to bind.</param>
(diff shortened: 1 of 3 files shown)


src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -139,72 +132,144 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
...
+            char[] updated = new char[length + 2];
+            updated[0] = '[';
+            char startingChar = normalizedText[0];
+            int j = 1;
+            for (int i = 0; i < length; i++)
             {
-                string convertedItem = normalizeFunc(items[i]);
-                if (i != items.Length - 1)
+                char character = normalizedText[i];
+                if (character == '"')
```
KanishManuja-MS: I feel this is overly complex. I suspected the fix should have been simpler - 
1. We should not loop over the string again and again. You loop once to find item start and end index, then iterate over with the SubString method. My assumption is that SubString itself is O(n).

2. The logic should have been simple, create a string builder, if starting character is ' or ", keep on appending to sb until we encounter the same closing quote, unless it is double escaped. You can check by looking ahead one character. Once that an items end, a comma should follow if it is not the end of normalized text.

kenneth-hansen: About point 2.: as far as I can tell OData allows whitespace between the string terminator and the comma, so "... in ('alice'  ,  'bob')" should be legal. Is this correct?

xuzhg: 1) To loop over the string again and again is the original goal. In the implementation, there are only one loop. Which one do you think we loop it twice. For the SubString, you can check it only happens when there's a mismatch, and throw exception.  In order to give the customer which item is wrong, i have to use "substring" to retrieve the "sub-string".

2) For the logic, i think my logic is same as your suggestion, but I use a string[] and not a stringbuilder. I don't want to use stringbuilder because stringbuilder is not fast than string[]. Besides, the key problem:

1) double quote can be wrapper of the string, or can be part of the string value.
 for example:   "(\" afbc'efg\")" and "('   afbc\"efg')"  are valid in "In" operator
   but, "(\" afbc'efg')', "(' afbc'efg\")" are not valid.

2) single quote can be wrapper of the string, or can be part of the string value. But, if single quote is par of the string, it should be "encoded" as double single quote. 

3) Comma can be separator of items, or can be part of the string value.  
        



---
In reply to: [367803756](https://github.com/OData/odata.net/pull/1635#discussion_r367803756) [](ancestors = 367803756)

xuzhg: @Kenneth Aas Hansen, the white space is allowed between string terminator. that's my thought. So, in my implementation, i remove the whitespace at Line 233.

I want to note that:

1) If you double commas, for example:  "... in ('alice'  , ,, 'bob'), i will let it go and let the low level process to identify the error or ignore the items between the commas.

2) for all other character between terminator, i throw exception, at Line 243.




---
In reply to: [367809728](https://github.com/OData/odata.net/pull/1635#discussion_r367809728) [](ancestors = 367809728)

xuzhg:
@KanishManuja-MS The build is triggered at https://identitydivision.visualstudio.com/OData/_build/results?buildId=458665&view=logs. 

You can click the second check link to access it.

xuzhg:
Move to #1642. Thanks.

Issue #1315, opened 2018-11-06
Client side Generator add Auth-Header to the request


paule96:
*If you have protected your metadata site of your odata api, you can't generate the Client Code without the Manipulation of the file `{funyfilename}.ttinclude`. It would be nice if a user of the Client Code Generator can set an Key-Value dictionary of headers.*

### Assemblies affected

*OData T4 client code Generator - 7.5.1*

### Reproduce steps

1. Protected your Metadata URL with an Header. (this is only an example. I don't compile this Code... )
```csharp
 app.Use(async (context, next) =>
   {
      var headerSet = context.Response.Headers..GetValues("Authorization").Any(d => d != null);
      if(headerSet){
          await next();
      }
      context.Response.StatusCode = (int)HttpStatusCode.Unauthorized;
      await context.Response.WriteAsync("No Header Auth please.");
   });
```
2. Create a new Console app in visual Studio
3. Add OData Model like [this tutorial](https://blogs.msdn.microsoft.com/odatateam/2014/03/11/tutorial-sample-how-to-use-odata-client-code-generator-to-generate-client-side-proxy-class/) describes
4. Step 3 will fail!

### Expected result

I can configure my headers and then generated my Client Code.

### Actual result

I can't configure my headers.

### Additional detail

The current work around is to add the following lines under the line 875 in the file `{funyfilename}.ttinclude`

```csharp
webRequest.Headers.Add(HttpRequestHeader.Authorization, "Bearer JWT-Token");
```

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

raheph:
@mikepizzo Can you share your thoughts?

mikepizzo:
So the proposal is to add a dictionary of headers to the Configuration class that get passed to the service when downloading metadata (essentially, adds Headers to the webRequest under 875 of the ttinclude, similar to what you are doing manually)?

Seems reasonable.

@paule96; would you want/be able to provide a pull request with this functionality?

KenitoInc:
Hey @paule96 
Am closing this issue. This feature was added to OData Connected Service v0.6

PR #1637, opened 2020-01-17
Fix Issue #1315 Add custom header feature

KenitoInc:
### Issues

*This pull request fixes issue #1315.*

### Description

Add support to add custom headers for the code generation. Now you can set the property CustomHttpHeaders in the Configuration class with your custom headers.
CustomHttpHeader are set in this format "{key:value},{key:value}"

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/CodeGen/ODataT4CodeGenerator.cs
+++ b/src/CodeGen/ODataT4CodeGenerator.cs
@@ -441,10 +441,22 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
 }
 
 /// <summary>
-/// Set the CustomHttpHeaders property from the custom http headers string.
+/// Validate the supplied custom http header string.
 /// </summary>
-/// <param name="stringValue">The value to set.</param>
-public void ValidateAndSetCustomHttpHeadersFromString(string headersValue) 
+/// <param name="header">Custom http header string.</param>
+public void ValidateCustomHttpHeaderString(string header) 
+{
+    if (!header.Contains(':'))
+    {
+        throw new ArgumentException("A http header string must have a colon delimeter");
+    }
+}
+
+/// <summary>
+/// Set the CustomHttpHeaders property using supplied custom http headers string.
+/// </summary>
+/// <param name="headersValue">Custom http headers string.</param>
+public void SetCustomHttpHeadersFromString(string headersValue) 
 {
     if (String.IsNullOrWhiteSpace(headersValue))
     {
@@ -458,16 +470,8 @@ public void ValidateAndSetCustomHttpHeadersFromString(string headersValue)
     {
         // Trim header for empty spaces
         var header = headerElement.Trim();
-
-        // A header string must have atleast one colon
-        if (header.Contains(':'))
-        {
-            CustomHttpHeaders.Add(header);
-        }
-        else
-        {
-            throw new ArgumentException("A http header string must have a colon delimeter");
-        }
+        ValidateCustomHttpHeaderString(header);
+        CustomHttpHeaders.Add(header);
     }
 
 }
@@ -484,7 +488,7 @@ private void ApplyParametersFromConfigurationClass()
     this.EnableNamingAlias = Configuration.EnableNamingAlias;
     this.TempFilePath = Configuration.TempFilePath;
     this.IgnoreUnexpectedElementsAndAttributes = Configuration.IgnoreUnexpectedElementsAndAttributes;
-    this.ValidateAndSetCustomHttpHeadersFromString(Configuration.CustomHttpHeaders);
+    this.SetCustomHttpHeadersFromString(Configuration.CustomHttpHeaders);
 }
 
 /// <summary>
--- a/src/CodeGen/ODataT4CodeGenerator.cs
+++ b/src/CodeGen/ODataT4CodeGenerator.cs
@@ -70,7 +70,7 @@ THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
             this.ApplyParametersFromConfigurationClass();
         }
 
-        context = new CodeGenerationContext(new Uri(this.MetadataDocumentUri, UriKind.Absolute), this.NamespacePrefix)
+        context = new CodeGenerationContext(new Uri(this.MetadataDocumentUri, UriKind.Absolute), this.NamespacePrefix, this.CustomHttpHeaders)
         {
             UseDataServiceCollection = this.UseDataServiceCollection,
             TargetLanguage = this.TargetLanguage,
@@ -153,6 +153,12 @@ public static class Configuration
 	// This flag indicates whether to ignore unexpected elements and attributes in the metadata document and generate
 	// the client code if any. The value must be set to true or false.
 	public const bool IgnoreUnexpectedElementsAndAttributes = true;
+
+	// (Optional) Custom http headers as a verbatim string in the format below: 
+	// CustomHttpHeaders = @"headerKey:headerValue
+	// headerKey:headerValue
+	// headerKey:headerValue";
+	public const string CustomHttpHeaders = "";
 }
 
 public static class Customization
@@ -346,6 +352,15 @@ public enum LanguageOption
 }
 
 /// <summary>
+/// Stores Custom Http Headers to be added to the WebRequest.Headers property.
+/// </summary>
+public virtual IList<string> CustomHttpHeaders
+{
+    get;
+    set;
+}
+
+/// <summary>
 /// Set the UseDataServiceCollection property with the given value.
 /// </summary>
 /// <param name="stringValue">The value to set.</param>
@@ -426,6 +441,38 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
 }
 
 /// <summary>
+/// Set the CustomHttpHeaders property from the custom http headers string.
+/// </summary>
+/// <param name="stringValue">The value to set.</param>
+public void ValidateAndSetCustomHttpHeadersFromString(string headersValue) 
+{
+    if (String.IsNullOrWhiteSpace(headersValue))
+    {
+        return;
+    }
+
+    this.CustomHttpHeaders = new List<string>();
+
+    string[] headerElements = headersValue.Split(new string[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+    foreach (var headerElement in headerElements)
+    {
+        // Trim header for empty spaces
+        var header = headerElement.Trim();
+
+        // A header string must have atleast one colon
+        if (header.Contains(':'))
+        {
+            CustomHttpHeaders.Add(header);
+        }
+        else
+        {
+            throw new ArgumentException("A http header string must have a colon delimeter");
+        }
+    }
+
+}
+
+/// <summary>
 /// Reads the parameter values from the Configuration class and applies them.
 /// </summary>
 private void ApplyParametersFromConfigurationClass()
@@ -437,6 +484,7 @@ private void ApplyParametersFromConfigurationClass()
     this.EnableNamingAlias = Configuration.EnableNamingAlias;
     this.TempFilePath = Configuration.TempFilePath;
     this.IgnoreUnexpectedElementsAndAttributes = Configuration.IgnoreUnexpectedElementsAndAttributes;
+    this.ValidateAndSetCustomHttpHeadersFromString(Configuration.CustomHttpHeaders);
 }
 
 /// <summary>
@@ -558,8 +606,8 @@ public class CodeGenerationContext
     /// Constructs an instance of <see cref="CodeGenerationContext"/>.
     /// </summary>
     /// <param name="metadataUri">The Uri to the metadata document. The supported scheme are File, http and https.</param>
-    public CodeGenerationContext(Uri metadataUri, string namespacePrefix)
-        : this(GetEdmxStringFromMetadataPath(metadataUri), namespacePrefix)
+    public CodeGenerationContext(Uri metadataUri, string namespacePrefix, IList<string> CustomHttpHeaders = null)
+        : this(GetEdmxStringFromMetadataPath(metadataUri, CustomHttpHeaders), namespacePrefix)
     {
     }
 
@@ -955,10 +1003,10 @@ public class CodeGenerationContext
     /// Reads the edmx string from a file path or a http/https path.
     /// </summary>
     /// <param name="metadataUri">The Uri to the metadata document. The supported scheme are File, http and https.</param>
-    private static string GetEdmxStringFromMetadataPath(Uri metadataUri)
+    private static string GetEdmxStringFromMetadataPath(Uri metadataUri, IList<string> CustomHttpHeaders = null)
     {
         string content = null;
-        using (StreamReader streamReader = new StreamReader(GetEdmxStreamFromUri(metadataUri)))
+        using (StreamReader streamReader = new StreamReader(GetEdmxStreamFromUri(metadataUri, CustomHttpHeaders)))
         {
             content = streamReader.ReadToEnd();
         }
@@ -970,7 +1018,7 @@ public class CodeGenerationContext
     /// Get the metadata stream from a file path or a http/https path.
     /// </summary>
     /// <param name="metadataUri">The Uri to the stream. The supported scheme are File, http and https.</param>
-    private static Stream GetEdmxStreamFromUri(Uri metadataUri)
+    private static Stream GetEdmxStreamFromUri(Uri metadataUri, IList<string> CustomHttpHeaders  = null)
     {
         Debug.Assert(metadataUri != null, "metadataUri != null");
         Stream metadataStream = null;
@@ -983,6 +1031,13 @@ public class CodeGenerationContext
             try
             {
                 HttpWebRequest webRequest = (HttpWebRequest)WebRequest.Create(metadataUri);
+                if (CustomHttpHeaders != null)
+                {
+                    foreach (var header in CustomHttpHeaders)
+                    {
+                        webRequest.Headers.Add(header);
+                    }
+                }
                 WebResponse webResponse = webRequest.GetResponse();
                 metadataStream = webResponse.GetResponseStream();
             }
--- a/src/CodeGen/ODataT4CodeGenerator.ttinclude
+++ b/src/CodeGen/ODataT4CodeGenerator.ttinclude
@@ -60,7 +60,7 @@ THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
             this.ApplyParametersFromConfigurationClass();
         }
 
-        context = new CodeGenerationContext(new Uri(this.MetadataDocumentUri, UriKind.Absolute), this.NamespacePrefix)
+        context = new CodeGenerationContext(new Uri(this.MetadataDocumentUri, UriKind.Absolute), this.NamespacePrefix, this.CustomHttpHeaders)
         {
             UseDataServiceCollection = this.UseDataServiceCollection,
             TargetLanguage = this.TargetLanguage,
@@ -238,6 +238,15 @@ public enum LanguageOption
 }
 
 /// <summary>
+/// Stores Custom Http Headers to be added to the WebRequest.Headers property.
+/// </summary>
+public virtual IList<string> CustomHttpHeaders
+{
+    get;
+    set;
+}
+
+/// <summary>
 /// Set the UseDataServiceCollection property with the given value.
 /// </summary>
 /// <param name="stringValue">The value to set.</param>
@@ -318,6 +327,38 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
 }
 
 /// <summary>
+/// Set the CustomHttpHeaders property from the custom http headers string.
+/// </summary>
+/// <param name="stringValue">The value to set.</param>
+public void ValidateAndSetCustomHttpHeadersFromString(string headersValue) 
+{
+    if (String.IsNullOrWhiteSpace(headersValue))
+    {
+        return;
+    }
+
+    this.CustomHttpHeaders = new List<string>();
+
+    string[] headerElements = headersValue.Split(new string[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
+    foreach (var headerElement in headerElements)
+    {
+        // Trim header for empty spaces
+        var header = headerElement.Trim();
+
+        // A header string must have atleast one colon
+        if (header.Contains(':'))
+        {
+            CustomHttpHeaders.Add(header);
+        }
+        else
+        {
+            throw new ArgumentException("A http header string must have a colon delimeter");
+        }
+    }
+
+}
+
+/// <summary>
 /// Reads the parameter values from the Configuration class and applies them.
 /// </summary>
 private void ApplyParametersFromConfigurationClass()
@@ -329,6 +370,7 @@ private void ApplyParametersFromConfigurationClass()
     this.EnableNamingAlias = Configuration.EnableNamingAlias;
     this.TempFilePath = Configuration.TempFilePath;
     this.IgnoreUnexpectedElementsAndAttributes = Configuration.IgnoreUnexpectedElementsAndAttributes;
+    this.ValidateAndSetCustomHttpHeadersFromString(Configuration.CustomHttpHeaders);
 }
 
 /// <summary>
@@ -450,8 +492,8 @@ public class CodeGenerationContext
     /// Constructs an instance of <see cref="CodeGenerationContext"/>.
     /// </summary>
     /// <param name="metadataUri">The Uri to the metadata document. The supported scheme are File, http and https.</param>
-    public CodeGenerationContext(Uri metadataUri, string namespacePrefix)
-        : this(GetEdmxStringFromMetadataPath(metadataUri), namespacePrefix)
+    public CodeGenerationContext(Uri metadataUri, string namespacePrefix, IList<string> CustomHttpHeaders = null)
+        : this(GetEdmxStringFromMetadataPath(metadataUri, CustomHttpHeaders), namespacePrefix)
     {
     }
 
@@ -847,10 +889,10 @@ public class CodeGenerationContext
     /// Reads the edmx string from a file path or a http/https path.
     /// </summary>
     /// <param name="metadataUri">The Uri to the metadata document. The supported scheme are File, http and https.</param>
-    private static string GetEdmxStringFromMetadataPath(Uri metadataUri)
+    private static string GetEdmxStringFromMetadataPath(Uri metadataUri, IList<string> CustomHttpHeaders = null)
     {
         string content = null;
-        using (StreamReader streamReader = new StreamReader(GetEdmxStreamFromUri(metadataUri)))
+        using (StreamReader streamReader = new StreamReader(GetEdmxStreamFromUri(metadataUri, CustomHttpHeaders)))
         {
             content = streamReader.ReadToEnd();
         }
@@ -862,7 +904,7 @@ public class CodeGenerationContext
     /// Get the metadata stream from a file path or a http/https path.
     /// </summary>
     /// <param name="metadataUri">The Uri to the stream. The supported scheme are File, http and https.</param>
-    private static Stream GetEdmxStreamFromUri(Uri metadataUri)
+    private static Stream GetEdmxStreamFromUri(Uri metadataUri, IList<string> CustomHttpHeaders  = null)
     {
         Debug.Assert(metadataUri != null, "metadataUri != null");
         Stream metadataStream = null;
@@ -875,6 +917,13 @@ public class CodeGenerationContext
             try
             {
                 HttpWebRequest webRequest = (HttpWebRequest)WebRequest.Create(metadataUri);
+                if (CustomHttpHeaders != null)
+                {
+                    foreach (var header in CustomHttpHeaders)
+                    {
+                        webRequest.Headers.Add(header);
+                    }
+                }
                 WebResponse webResponse = webRequest.GetResponse();
                 metadataStream = webResponse.GetResponseStream();
             }
--- a/src/CodeGen/ODataT4CodeGenerator.ttinclude
+++ b/src/CodeGen/ODataT4CodeGenerator.ttinclude
@@ -327,10 +327,22 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
 }
 
 /// <summary>
-/// Set the CustomHttpHeaders property from the custom http headers string.
+/// Validate the supplied custom http header string.
 /// </summary>
-/// <param name="stringValue">The value to set.</param>
-public void ValidateAndSetCustomHttpHeadersFromString(string headersValue) 
+/// <param name="header">Custom http header string.</param>
+public void ValidateCustomHttpHeaderString(string header) 
+{
+    if (!header.Contains(':'))
+    {
+        throw new ArgumentException("A http header string must have a colon delimeter");
+    }
+}
+
+/// <summary>
+/// Set the CustomHttpHeaders property using supplied custom http headers string.
+/// </summary>
+/// <param name="headersValue">Custom http headers string.</param>
+public void SetCustomHttpHeadersFromString(string headersValue) 
 {
     if (String.IsNullOrWhiteSpace(headersValue))
     {
@@ -344,16 +356,8 @@ public void ValidateAndSetCustomHttpHeadersFromString(string headersValue)
     {
         // Trim header for empty spaces
         var header = headerElement.Trim();
-
-        // A header string must have atleast one colon
-        if (header.Contains(':'))
-        {
-            CustomHttpHeaders.Add(header);
-        }
-        else
-        {
-            throw new ArgumentException("A http header string must have a colon delimeter");
-        }
+        ValidateCustomHttpHeaderString(header);
+        CustomHttpHeaders.Add(header);
     }
 
 }
@@ -370,7 +374,7 @@ private void ApplyParametersFromConfigurationClass()
     this.EnableNamingAlias = Configuration.EnableNamingAlias;
     this.TempFilePath = Configuration.TempFilePath;
     this.IgnoreUnexpectedElementsAndAttributes = Configuration.IgnoreUnexpectedElementsAndAttributes;
-    this.ValidateAndSetCustomHttpHeadersFromString(Configuration.CustomHttpHeaders);
+    this.SetCustomHttpHeadersFromString(Configuration.CustomHttpHeaders);
 }
 
 /// <summary>
--- a/test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
+++ b/test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
@@ -294,5 +294,84 @@ namespace Microsoft.OData.Client.Design.T4.UnitTests
                 Assert.AreEqual(ex.Message, "Only file, http, https schemes are supported for paths to metadata source locations.");
             }
         }
+
+        [TestMethod]
+        public void SetOneCustomHeaderCorrectlyShouldNotThrowException()
+        {
+            ODataT4CodeGenerator codegen = new ODataT4CodeGenerator();
+            string customHeaderString = @"Authorization:Bearer bearer-token";
+            codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            Assert.IsNotNull(codegen.CustomHttpHeaders);
+            Assert.AreEqual(1, codegen.CustomHttpHeaders.Count);
+            Assert.AreEqual("Authorization:Bearer bearer-token", codegen.CustomHttpHeaders.FirstOrDefault());
+        }
+
+        [TestMethod]
+        public void SetMultipleCustomHeadersCorrectlyShouldNotThrowException()
+        {
+            ODataT4CodeGenerator codegen = new ODataT4CodeGenerator();
+            string customHeaderString = @"Authorization:Bearer bearer-token
+                                          odata.continue-on-error:true";
+            codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            Assert.IsNotNull(codegen.CustomHttpHeaders);
+            Assert.AreEqual(2, codegen.CustomHttpHeaders.Count);
+            Assert.AreEqual("Authorization:Bearer bearer-token", codegen.CustomHttpHeaders.FirstOrDefault());
+            Assert.AreEqual("odata.continue-on-error:true", codegen.CustomHttpHeaders.LastOrDefault());
+        }
+
+        [TestMethod]
+        public void SetCustomHeaderInCorrectlyShouldThrowException()
+        {
+            ODataT4CodeGenerator codegen = new ODataT4CodeGenerator();
+            string customHeaderString = @"Authorization Bearer bearer-token";
+            
+            try
+            {
+                codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            }
+            catch(ArgumentException exception)
+            {
+                StringAssert.Contains(exception.Message, "A http header string must have a colon delimeter");
+                return;
+            }
+            Assert.Fail("Expected ArgumentException is not thrown");
+            
+        }
+
+        [TestMethod]
+        public void SetNullCustomHeaderShouldNotThrowException()
+        {
+            ODataT4CodeGenerator codegen = new ODataT4CodeGenerator();
+            string customHeaderString = null;
+            codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            Assert.IsNull(codegen.CustomHttpHeaders);
+        }
+
+        [TestMethod]
+        public void SetCustomHeadersWithQuotesShouldNotThrowException()
+        {
+            // Quotes are sent as part of the header value
+            ODataT4CodeGenerator codegen = new ODataT4CodeGenerator();
+            string customHeaderString = @"If-Match:'67ab43'";
+            codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            Assert.IsNotNull(codegen.CustomHttpHeaders);
+            Assert.AreEqual(1, codegen.CustomHttpHeaders.Count);
+            Assert.AreEqual("If-Match:'67ab43'", codegen.CustomHttpHeaders.FirstOrDefault());
+        }
+
+        [TestMethod]
+        public void SetMultipleCustomHeadersWithEmptyNewLinesShouldNotThrowError()
+        {
+            ODataT4CodeGenerator codegen = new ODataT4CodeGenerator();
+            string customHeaderString = @"Authorization:Bearer bearer-token
+
+
+                                          odata.continue-on-error:true";
+            codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            Assert.IsNotNull(codegen.CustomHttpHeaders);
+            Assert.AreEqual(2, codegen.CustomHttpHeaders.Count);
+            Assert.AreEqual("Authorization:Bearer bearer-token", codegen.CustomHttpHeaders.FirstOrDefault());
+            Assert.AreEqual("odata.continue-on-error:true", codegen.CustomHttpHeaders.LastOrDefault());
+        }
     }
 }
--- a/test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
+++ b/test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
@@ -300,10 +300,10 @@ namespace Microsoft.OData.Client.Design.T4.UnitTests
         {
             ODataT4CodeGenerator codegen = new ODataT4CodeGenerator();
             string customHeaderString = @"Authorization:Bearer bearer-token";
-            codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            codegen.SetCustomHttpHeadersFromString(customHeaderString);
             Assert.IsNotNull(codegen.CustomHttpHeaders);
             Assert.AreEqual(1, codegen.CustomHttpHeaders.Count);
-            Assert.AreEqual("Authorization:Bearer bearer-token", codegen.CustomHttpHeaders.FirstOrDefault());
+            Assert.AreEqual("Authorization:Bearer bearer-token", codegen.CustomHttpHeaders.First());
         }
 
         [TestMethod]
@@ -312,11 +312,11 @@ namespace Microsoft.OData.Client.Design.T4.UnitTests
             ODataT4CodeGenerator codegen = new ODataT4CodeGenerator();
             string customHeaderString = @"Authorization:Bearer bearer-token
                                           odata.continue-on-error:true";
-            codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            codegen.SetCustomHttpHeadersFromString(customHeaderString);
             Assert.IsNotNull(codegen.CustomHttpHeaders);
             Assert.AreEqual(2, codegen.CustomHttpHeaders.Count);
-            Assert.AreEqual("Authorization:Bearer bearer-token", codegen.CustomHttpHeaders.FirstOrDefault());
-            Assert.AreEqual("odata.continue-on-error:true", codegen.CustomHttpHeaders.LastOrDefault());
+            Assert.AreEqual("Authorization:Bearer bearer-token", codegen.CustomHttpHeaders.First());
+            Assert.AreEqual("odata.continue-on-error:true", codegen.CustomHttpHeaders.Last());
         }
 
         [TestMethod]
@@ -324,7 +324,7 @@ namespace Microsoft.OData.Client.Design.T4.UnitTests
         {
             ODataT4CodeGenerator codegen = new ODataT4CodeGenerator();
             string customHeaderString = @"Authorization Bearer bearer-token";
-            Action act = () => codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            Action act = () => codegen.SetCustomHttpHeadersFromString(customHeaderString);
             act.ShouldThrow<ArgumentException>().WithMessage("A http header string must have a colon delimeter");
         }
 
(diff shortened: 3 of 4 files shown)


habbes:
@mikepizzo  @KenitoInc  Out of curiosity, what's the rationale for using a new ad-hoc JSON-like DSL to specify the headers for the configuration instead of a using actual JSON for which we can find an existing parser. Or instead of just using the plain HTTP headers syntax?

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -982,8 +1110,15 @@ private static Stream GetEdmxStreamFromUri(Uri metadataUri)
...
             try
             {
-                HttpWebRequest webRequest = (HttpWebRequest)WebRequest.Create(metadataUri);
-                WebResponse webResponse = webRequest.GetResponse();
+                HttpWebRequest webRequest = (HttpWebRequest) WebRequest.Create(metadataUri);
+                if (CustomHttpHeaders != null)
+                {
+	                foreach(var header in CustomHttpHeaders)
+	                {
+		                webRequest.Headers.Add(header.Key, header.Value);
```
xuzhg: align the codes

src/CodeGen/ODataT4CodeGenerator.ttinclude
```diff
@@ -38,6 +38,7 @@ THE SOFTWARE IS PROVIDED *AS IS*, WITHOUT WARRANTY OF ANY KIND, EXPRESS OR IMPLI
 <#@ Import Namespace="Microsoft.OData.Edm.Vocabularies.V1" #>
 <#@ Import Namespace="Microsoft.OData.Edm.Vocabularies.Community.V1" #>
 <#@ Import Namespace="System.Text"#>
+<#@ Import Namespace="System.Text.RegularExpressions"#>
```
xuzhg: please search the security warning to use "RegularExpression " and make sure it's ok to use it.

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -153,6 +154,9 @@ public static class Configuration
 	// This flag indicates whether to ignore unexpected elements and attributes in the metadata document and generate
 	// the client code if any. The value must be set to true or false.
 	public const bool IgnoreUnexpectedElementsAndAttributes = true;
+
+	// (Optional) Custom http headers in this format: {headerName:value},{headerName:value}
```
xuzhg: where's from this pattern " {key:value},{key:value}  "?

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -345,6 +349,12 @@ public enum LanguageOption
     VB = 1,
 }
 
+public Dictionary<string, string> CustomHttpHeaders
```
xuzhg: Maybe use "IDictionary<string, string> is better? and mark it as virtual?

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +435,123 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
     this.IgnoreUnexpectedElementsAndAttributes = boolValue;
 }
 
+public void ValidateAndSetCustomHttpHeadersFromString(string stringValue) 
+{
+    this.CustomHttpHeaders = new Dictionary <string, string> ();
```
xuzhg: pay attention to the stylecop. for example, remove unused whitespace after Dictionary

xuzhg: move line 440 to  line 445?

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +435,123 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
     this.IgnoreUnexpectedElementsAndAttributes = boolValue;
 }
 
+public void ValidateAndSetCustomHttpHeadersFromString(string stringValue)
```
xuzhg: do you want customer to provider customize parsing for the header string?

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +435,123 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+    this.CustomHttpHeaders = new Dictionary <string, string> ();
+
+    if (String.IsNullOrWhiteSpace(stringValue)) {
+        return;
+    }
+
+    if (ValidateParantheses(stringValue) && ValidateFirstLast(stringValue) && ValidateDelimeters(stringValue)) 
+    {
+        string header = stringValue.TrimEnd('}', ' ');
+        string[] elements = header.Split('}');
```
xuzhg: the value can have "}", right? how about it?

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +435,123 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+        foreach (var element in elements)
+        {
+            string _element = element.TrimStart(',', '{', ' ');
+            string[] array = _element.Split(':');
```
xuzhg: the string value can have ":", right? how about it?

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +435,123 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+            CustomHttpHeaders.Add(array[0], array[1]);
+        }
+    }
+    else 
+    {
+        throw new ArgumentException("Please check your custom http headers string format! You have set the custom http header but not use the format '{headerName:value},{headerName:value}'");
```
xuzhg: does code gen has the central error message management?

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +435,123 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+
+private static bool ValidateParantheses(string _input)
+{
+    Dictionary<char, char> paraPairs = new Dictionary<char, char>() {
+        { '{', '}' }
+    };
+
+    Stack<char> stack = new Stack<char>();
+
+    foreach (char i in _input)
```
xuzhg: "i" looks an integer. maybe use "c" or "character" is better?

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +435,123 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+    {
+        if(stack.Count > 1)
+        {
+            return false;
+        }
+        if (paraPairs.Keys.Contains(i))
```
xuzhg: what's the benefit to use Dictionary and call "Contains?
why don't directly compare if (c == '{') ?

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +435,123 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+            else
+            {
+                return false;
+            }
+        }
+    }
+    return stack.Count() == 0 ? true : false;
+}
+
+private static bool ValidateFirstLast(string _input)
```
xuzhg: This method looks unnecessary. Maybe use the codes in the main method and re-use the result of _input.Trim()?

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +435,123 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+private static int GetMatchedDelimeters(string _input)
+{
+    // Remove white spaces
+    string headers = Regex.Replace(_input, @"\s", "");
+
+    // Regex pattern },{
+    string delimeterPattern = @"[\}][,][\{]";
+    Regex rgx = new Regex(delimeterPattern);
+    // Get all matches  
+    MatchCollection matchedHeaders = rgx.Matches(headers);
```
xuzhg: check the security to use Regx?

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -437,6 +564,7 @@ private void ApplyParametersFromConfigurationClass()
     this.EnableNamingAlias = Configuration.EnableNamingAlias;
     this.TempFilePath = Configuration.TempFilePath;
     this.IgnoreUnexpectedElementsAndAttributes = Configuration.IgnoreUnexpectedElementsAndAttributes;
+	this.ValidateAndSetCustomHttpHeadersFromString(Configuration.CustomHttpHeaders);
```
xuzhg: align the codes

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: @mikepizzo gave a detailed explanation in a previous incomplete PR https://github.com/OData/odata.net/pull/1348#issuecomment-447116560

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: The customer will be providing one or more custom headers in form of a string that follows this pattern *"{key:value},{key:value}"*
This function validates the input string so that we can be able to split the string and properly add the headers to the WebRequest.Headers property

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: We are trimming the last "}" so that we don't have an empty string at the end when we split the string via the remaining "}"

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: The input string has key value pairs e.g
"{Accept:application/json},{Authorization: Bearer xxxyssh-token}";

KenitoInc:
> @mikepizzo  @KenitoInc  Out of curiosity, what's the rationale for using a new ad-hoc JSON-like DSL to specify the headers for the configuration instead of a using actual JSON for which we can find an existing parser. Or instead of just using the plain HTTP headers syntax?

There is a detailed explanation done by Mike here -> https://github.com/OData/odata.net/pull/1348#issuecomment-447116560

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: No it doesn't

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Done

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: I have made the necessary changes

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: I have made the necessary changes

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Done

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Done

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: This regex pattern is evaluated in a linear manner instead of backtracking. Therefore it isn't prone to ReDOS (Regex Denial Of Service)

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: I have replaced with `virtual IDictionary<string, string>` so as to allow *mocking*

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: I have opted to leave it that way for now to ensure code is a bit clean

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
```diff
@@ -294,5 +294,72 @@ public void GetEdmxStringFromMetadataPathShouldThrowExceptionIfInputIsInvalidUri
...
+            Assert.AreEqual(1, codegen.CustomHttpHeaders.Count);
+            Assert.AreEqual("Authorization", codegen.CustomHttpHeaders.FirstOrDefault().Key);
+            Assert.AreEqual("Bearer bearer-token", codegen.CustomHttpHeaders.FirstOrDefault().Value);
+        }
+
+        [TestMethod]
+        public void SetMultipleCustomHeadersCorrectlyShouldNotThrowException()
+        {
+            var codegen = new ODataT4CodeGenerator();
+            string customHeaderString = "{Authorization:Bearer bearer-token},{odata.continue-on-error:true}";
```
odero:
```suggestion
            var customHeaderString = "{Authorization:Bearer bearer-token},{odata.continue-on-error:true}";
```

src/CodeGen/ODataT4CodeGenerator.cs
xuzhg: [https://github.com/OData/odata.net/pull/1348#issuecomment-447116560](https://github.com/OData/odata.net/pull/1348#issuecomment-447116560)

---
In reply to: [368252373](https://github.com/OData/odata.net/pull/1637#discussion_r368252373) [](ancestors = 368252373)

xuzhg: Thanks, but how about if the value has ":" itself? should we escaped it?

---
In reply to: [368404438](https://github.com/OData/odata.net/pull/1637#discussion_r368404438) [](ancestors = 368404438)

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -345,6 +349,15 @@ public enum LanguageOption
     VB = 1,
 }
 
+/// <summary>
+/// Stores Custom Http Headers to be added to the WebRequest.Headers property.
+/// </summary>
+public virtual IDictionary<string, string> CustomHttpHeaders
+{
+	get;
+	set;
```
xuzhg: replace "tab" as 4 whitespaces.

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +438,124 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
     this.IgnoreUnexpectedElementsAndAttributes = boolValue;
 }
 
+/// <summary>
+/// Set the CustomHttpHeaders property from the custom http headers string.
+/// </summary>
+/// <param name="stringValue">The value to set.</param>
+public void ValidateAndSetCustomHttpHeadersFromString(string stringValue) 
+{
+    if (String.IsNullOrWhiteSpace(stringValue)) {
```
xuzhg: >)) { [](start = 45, length = 4)

put "{" in a single line.

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
```diff
@@ -294,5 +294,72 @@ public void GetEdmxStringFromMetadataPathShouldThrowExceptionIfInputIsInvalidUri
...
+            try
+            {
+                codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            }
+            catch(ArgumentException exception)
+            {
+                StringAssert.Contains(exception.Message, "Custom http headers string format should be '{headerName:value},{headerName:value}'");
+                return;
+            }
+            Assert.Fail("Expected ArgumentException is not thrown");
```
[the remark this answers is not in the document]
xuzhg: +1

---
In reply to: [369074844](https://github.com/OData/odata.net/pull/1637#discussion_r369074844) [](ancestors = 369074844)

src/CodeGen/ODataT4CodeGenerator.cs
xuzhg: So, it's not allowed to "{Acce}:pt:application/json:abc}" ?

---
In reply to: [368422032](https://github.com/OData/odata.net/pull/1637#discussion_r368422032) [](ancestors = 368422032)

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +438,124 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+        }
+    }
+    else 
+    {
+        throw new ArgumentException("Custom http headers string format should be '{headerName:value},{headerName:value}'");
+    }
+
+}
+
+private static bool ValidateParantheses(string _input)
```
xuzhg: >string _input [](start = 40, length = 13)

It's not conventional to name a parameter name with "underscore".

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +438,124 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+{
+
+    Stack<char> stack = new Stack<char>();
```
xuzhg: remove unused line

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +438,124 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+    foreach (char c in _input)
+    {
+        if(stack.Count > 1)
+        {
+            return false;
+        }
+        if (c == '{')
+        {
+            stack.Push(c);
+        }
```
xuzhg: >                     [](start = 9, length = 20)

remove unused whitespaces

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +438,124 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+            {
+                stack.Pop();
+            }
+            else
+            {
+                return false;
+            }
+        }
+    }
+    return stack.Count() == 0 ? true : false;
```
xuzhg: >stack.Count() == 0 ? true : false; [](start = 11, length = 34)

If the parentheses can't match, can you throw exception directly here?

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Done

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
```diff
@@ -294,5 +294,72 @@ public void GetEdmxStringFromMetadataPathShouldThrowExceptionIfInputIsInvalidUri
                 Assert.AreEqual(ex.Message, "Only file, http, https schemes are supported for paths to metadata source locations.");
             }
         }
+
+        [TestMethod]
+        public void SetOneCustomHeaderCorrectlyShouldNotThrowException()
+        {
+            var codegen = new ODataT4CodeGenerator();
+            string customHeaderString = "{Authorization:Bearer bearer-token}";
```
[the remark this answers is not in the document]
KenitoInc: I thought `var customHeaderString` will still be compile as `string customHeaderString`

mikepizzo:
@habbes/@KenitoInc: there's no clear, or clearly superior, answer here. The syntax of a header value allows virtually any character, and common delimiters (like comma, colon, space, single/double quote) are commonly used in header values, so some other convention is required.

Making this JSON or XML is probably overkill. Also, strings in JSON/XML are double-quoted, so within a C# declaration they would have to be doubled, which is ugly.

I think the curly braces implies JSON, which leads us down the wrong path.

We already have a simple escaping rule for strings in a URL (string is enclosed in single quotes, single quotes are doubled), so maybe we should just apply that here.

The following would be unambiguous, simple to parse, and consistent with handling strings in urls:
        Accept: 'application/json', If-Match: 'W/"x,y''z\"', Content-Type: 'application/json'
into:
       Accept: application/json
       If-Match: W/"x,y'z"
       Content-Type: application/json

we could also say that you only had to quote the value if it contained a comma, so we could simplify above to:
        Accept: application/json, If-Match: 'W/"x,y''z\"', Content-Type: application/json

Thoughts?

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Done

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Done

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Fixed this

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Done

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Done

habbes:
@mikepizzo thanks for the clarification. My other question was why not use the standard header format:

e.g.
```c#
string CustomHeaders = @"
Accept: application/json
If-Match: W/""x,y'z""
Content-Type: application/json
"
```
would return headers:
```
Accept: application/json
If-Match: W/"x,y'z"
Content-Type: application/json
```
I think this allows the user to include any value that's accepted in a header value without escaping. And it also makes parsing easier since you can simply split fields by line breaks. According to the [RFC 2822](https://tools.ietf.org/html/rfc2822#section-2.2.3), field value can only contain `CRLF` when folding to the next line. If we say we don't support folding, then I think this would be the ideal syntax in my opinion. And if at some point down the line users request support for folding, I don't think it would be too nerve-wrecking to add support for it.

But I do concede that quoting double quotes with extra double quotes in verbatim strings is not that pretty. But that's a csharp issue, not an issue with the headers syntax.

src/CodeGen/ODataT4CodeGenerator.ttinclude
```diff
@@ -317,6 +327,70 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
     this.IgnoreUnexpectedElementsAndAttributes = boolValue;
 }
 
+/// <summary>
+/// Set the CustomHttpHeaders property from the custom http headers string.
+/// </summary>
+/// <param name="stringValue">The value to set.</param>
+public void ValidateAndSetCustomHttpHeadersFromString(string stringValue)
```
wachugamaina: Ideally, functions should only do one thing, just the name ValidateAndSetCustomHttpHeadersFromString(string stringValue) suggests the function may be doing too much.

mikepizzo:
@habbes - I considered newline separated, but (in addition to the wrapping, which I agree is not common) it's a bit harder to format, i.e., in code or when presented as an edit box.  But I agree that sticking closer to the standard is probably good, and multi-line edit boxes can be used in UI, so I'm fine with that solution.

KenitoInc:
@habbes @mikepizzo I decided to go with a verbatim string which allows the user to input the standard http header format as a string. I have update my code and pushed to this PR

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +441,50 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+        }
+        else
+        {
+            throw new ArgumentException("A http header string must have a semi-colon delimeter");
+        }
+    }
+
+}
+
+private static int GetSemiColons(string input)
```
g2mula: The name does not indicate what this method actually does, GetSemiColonCount would probably be a more appropriate name.

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +441,50 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+{
+    // Remove white spaces
+    string headers = Regex.Replace(input, @"\s", "");
+
```
g2mula: This seems like unnecessary code.

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +441,50 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+    //Count matched closing parantheses
+    int matchedSemiColons = Regex.Matches(headers, @"\:").Count;
```
g2mula: Are we counting colons ```:``` or semi-colons ```;``` either way, Regex seems like overkill here.
If we are counting colons we might as well use ```input.Count(c => c == ':')```

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +441,50 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+    this.CustomHttpHeaders = new List<string>();
+
+    string[] headerElements = headersValue.Split(new string[] { Environment.NewLine },StringSplitOptions.None);
+    foreach (var headerElement in headerElements)
+    {
+        // Trim header for empty spaces
+        var header = headerElement.Trim();
+
+        // A header string must have atleast one semi-colon
+        if (GetSemiColons(header) > 0)
```
g2mula: It would seem like we don't care about the number of ~semi~colons in the header, rather we only care whether there is a ~semi~colon in the header, if so we could just do ```if (header.Contains(':'))``` and then drop the ```GetSemiColons()``` method.

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -983,6 +1044,13 @@ private static Stream GetEdmxStreamFromUri(Uri metadataUri)
             try
             {
                 HttpWebRequest webRequest = (HttpWebRequest)WebRequest.Create(metadataUri);
+                if(CustomHttpHeaders !=null)
```
g2mula:
```suggestion
                if (CustomHttpHeaders != null)
``` 
whitespacing

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -983,6 +1044,13 @@ private static Stream GetEdmxStreamFromUri(Uri metadataUri)
             try
             {
                 HttpWebRequest webRequest = (HttpWebRequest)WebRequest.Create(metadataUri);
+                if(CustomHttpHeaders !=null)
+                {
+                    foreach(var header in CustomHttpHeaders)
```
g2mula:
```suggestion
                    foreach (var header in CustomHttpHeaders)
```
Whitespacing.

src/CodeGen/ODataT4CodeGenerator.ttinclude
```diff
@@ -317,6 +327,50 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+    // Remove white spaces
+    string headers = Regex.Replace(input, @"\s", "");
+
+    //Count matched closing parantheses
+    int matchedSemiColons = Regex.Matches(headers, @"\:").Count;
+
+    // If valid, delimeters = closing parantheses - 1
+    return matchedSemiColons;
+}
+
```
g2mula: Check the comments left on ODataT4CodeGenerator.cs file

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +441,50 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+        var header = headerElement.Trim();
+
+        // A header string must have atleast one semi-colon
+        if (GetSemiColons(header) > 0)
+        {
+            CustomHttpHeaders.Add(header);
+        }
+        else
+        {
+            throw new ArgumentException("A http header string must have a semi-colon delimeter");
```
g2mula:
```suggestion
            throw new ArgumentException("A http header string must have a colon delimeter");
```

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
```diff
@@ -294,5 +294,69 @@ public void GetEdmxStringFromMetadataPathShouldThrowExceptionIfInputIsInvalidUri
...
+            var codegen = new ODataT4CodeGenerator();
+            string customHeaderString = @"Authorization Bearer bearer-token";
+            
+            try
+            {
+                codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            }
+            catch(ArgumentException exception)
+            {
+                StringAssert.Contains(exception.Message, "A http header string must have a semi-colon delimeter");
```
g2mula:
```suggestion
                StringAssert.Contains(exception.Message, "A http header string must have a colon delimeter");
```

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +441,50 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+public void ValidateAndSetCustomHttpHeadersFromString(string headersValue) 
+{
+    if (String.IsNullOrWhiteSpace(headersValue))
+    {
+        return;
+    }
+
+    this.CustomHttpHeaders = new List<string>();
+
+    string[] headerElements = headersValue.Split(new string[] { Environment.NewLine },StringSplitOptions.None);
```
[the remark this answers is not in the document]
KenitoInc: Done

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: I have removed the entire GetSemiColons method

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: I have removed the entire GetSemiColons method

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: I have removed the entire GetSemiColons method

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Fixed

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Fixed

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Fixed

src/CodeGen/ODataT4CodeGenerator.ttinclude
KenitoInc: Fixed

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Fixed

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
KenitoInc: Fixed

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -23,6 +23,7 @@ namespace Microsoft.OData.Client.Design.T4
     using Microsoft.OData.Edm.Vocabularies.V1;
     using Microsoft.OData.Edm.Vocabularies.Community.V1;
     using System.Text;
+    using System.Text.RegularExpressions;
```
g2mula:
```suggestion
```
Seems like this import is no longer necessary

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +441,38 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+
+    this.CustomHttpHeaders = new List<string>();
+
+    string[] headerElements = headersValue.Split(new string[] { Environment.NewLine }, StringSplitOptions.None);
+    foreach (var headerElement in headerElements)
+    {
+        // Trim header for empty spaces
+        var header = headerElement.Trim();
+
+        // A header string must have atleast one semi-colon
```
g2mula:
```suggestion
        // A header string must have atleast one colon

```
Colon-> :
Semi-colon-> ;

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +441,38 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+public void ValidateAndSetCustomHttpHeadersFromString(string headersValue) 
+{
+    if (String.IsNullOrWhiteSpace(headersValue))
+    {
+        return;
+    }
+
+    this.CustomHttpHeaders = new List<string>();
+
+    string[] headerElements = headersValue.Split(new string[] { Environment.NewLine }, StringSplitOptions.None);
```
g2mula: Question, is it possible for the headersValue to contain a newline somewhere? start, end, middle? If so, then ```StringSplitOptions.RemoveEmptyEntries``` might be better so that the tests below don't throw for an empty string. If having empty lines is not desired, then it works as is.

src/CodeGen/ODataT4CodeGenerator.ttinclude
g2mula:
```suggestion
```
Seems to no longer be in use.

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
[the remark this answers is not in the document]
g2mula: https://fluentassertions.com/exceptions/

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
```diff
@@ -294,5 +294,69 @@ public void GetEdmxStringFromMetadataPathShouldThrowExceptionIfInputIsInvalidUri
...
+            var codegen = new ODataT4CodeGenerator();
+            string customHeaderString = @"Authorization:Bearer bearer-token";
+            codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            Assert.IsNotNull(codegen.CustomHttpHeaders);
+            Assert.AreEqual(1, codegen.CustomHttpHeaders.Count);
+            Assert.AreEqual("Authorization:Bearer bearer-token", codegen.CustomHttpHeaders.FirstOrDefault());
+        }
+
+        [TestMethod]
+        public void SetMultipleCustomHeadersCorrectlyShouldNotThrowException()
```
g2mula: Would be nice to add a test for ```SetMultipleCustomHeadersWithEmptyNewLinesShouldEitherThrowOrNot()``` with a sample input like
```
var customHeaderString = @"
Auth: Yes

odata.continue-on-error: false
";
```
To document the choice of StringSplitOptions.None when splitting headerValue.

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
```diff
@@ -294,5 +294,69 @@ public void GetEdmxStringFromMetadataPathShouldThrowExceptionIfInputIsInvalidUri
                 Assert.AreEqual(ex.Message, "Only file, http, https schemes are supported for paths to metadata source locations.");
             }
         }
+
+        [TestMethod]
+        public void SetOneCustomHeaderCorrectlyShouldNotThrowException()
+        {
+            var codegen = new ODataT4CodeGenerator();
+            string customHeaderString = @"Authorization:Bearer bearer-token";
```
g2mula: Would be better to be consistent in either preferring ```var``` or ```type```.
The style in this file seems to prefer ```type``` over ```var``` would be much better to follow that.

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Done

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Updated

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Changes to RemoveEmptyEntries

src/CodeGen/ODataT4CodeGenerator.ttinclude
KenitoInc: Removed

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
KenitoInc: Added

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
KenitoInc: I have change all to type

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +435,123 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+        }
+    }
+    else 
+    {
+        throw new ArgumentException("Please check your custom http headers string format! You have set the custom http header but not use the format '{headerName:value},{headerName:value}'");
+    }
+
+}
+
+private static bool ValidateParantheses(string _input)
```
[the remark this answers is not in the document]
KenitoInc: This was fixed after I changes the input to be a verbatim string

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +435,123 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+{
+    string input = _input.Trim();
+    char firstElement = input[0];
+    char lastElement = input[input.Length - 1];
+
+    if(firstElement == '{' && lastElement == '}')
+    {
+        return true;
+    }
+    return false;
```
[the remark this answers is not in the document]
KenitoInc: This was fixed after I changes the input to be a verbatim string

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: This was fixed after I changed the input to be a verbatim string

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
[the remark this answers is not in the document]
KenitoInc: I have updated the code to use fluent assertions

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
KenitoInc: @odero The style in this file seems to prefer `type` over `var`, so I have used string for consistency

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +440,38 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
     this.IgnoreUnexpectedElementsAndAttributes = boolValue;
 }
 
+/// <summary>
+/// Set the CustomHttpHeaders property from the custom http headers string.
```
odero:
```suggestion
/// Set the CustomHttpHeaders property from using supplied custom http headers string.

```

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +440,38 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
     this.IgnoreUnexpectedElementsAndAttributes = boolValue;
 }
 
+/// <summary>
+/// Set the CustomHttpHeaders property from the custom http headers string.
+/// </summary>
+/// <param name="stringValue">The value to set.</param>
```
odero:
```suggestion
/// <param name="headersValue">Custom http headers string.</param>

```

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +440,38 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+public void ValidateAndSetCustomHttpHeadersFromString(string headersValue) 
+{
+    if (String.IsNullOrWhiteSpace(headersValue))
+    {
+        return;
+    }
+
+    this.CustomHttpHeaders = new List<string>();
+
+    string[] headerElements = headersValue.Split(new string[] { Environment.NewLine }, StringSplitOptions.RemoveEmptyEntries);
```
odero:
```suggestion
    string[] headerElements = headersValue.Split(Environment.NewLine, StringSplitOptions.RemoveEmptyEntries);
```

Any reason for passing the separator as as `string[]` instead of just `string`?

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +440,38 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
...
+        var header = headerElement.Trim();
+
+        // A header string must have atleast one colon
+        if (header.Contains(':'))
+        {
+            CustomHttpHeaders.Add(header);
+        }
+        else
+        {
+            throw new ArgumentException("A http header string must have a colon delimeter");
```
odero:
```suggestion
            throw new ArgumentException($"The http header string {header} does not have a colon separator");

```

src/CodeGen/ODataT4CodeGenerator.cs
```diff
@@ -425,6 +440,38 @@ public void ValidateAndSetIgnoreUnexpectedElementsAndAttributesFromString(string
     this.IgnoreUnexpectedElementsAndAttributes = boolValue;
 }
 
+/// <summary>
+/// Set the CustomHttpHeaders property from the custom http headers string.
+/// </summary>
+/// <param name="stringValue">The value to set.</param>
+public void ValidateAndSetCustomHttpHeadersFromString(string headersValue)
```
odero: It might make sense to split this into 2 separator methods one to validate and another to set

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
```diff
@@ -294,5 +294,74 @@ public void GetEdmxStringFromMetadataPathShouldThrowExceptionIfInputIsInvalidUri
...
+
+        [TestMethod]
+        public void SetOneCustomHeaderCorrectlyShouldNotThrowException()
+        {
+            ODataT4CodeGenerator codegen = new ODataT4CodeGenerator();
+            string customHeaderString = @"Authorization:Bearer bearer-token";
+            codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            Assert.IsNotNull(codegen.CustomHttpHeaders);
+            Assert.AreEqual(1, codegen.CustomHttpHeaders.Count);
+            Assert.AreEqual("Authorization:Bearer bearer-token", codegen.CustomHttpHeaders.FirstOrDefault());
```
odero:
```suggestion
            Assert.AreEqual("Authorization:Bearer bearer-token", codegen.CustomHttpHeaders.First());

```
No reason why it should ever return default

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
```diff
@@ -294,5 +294,74 @@ public void GetEdmxStringFromMetadataPathShouldThrowExceptionIfInputIsInvalidUri
...
+        [TestMethod]
+        public void SetMultipleCustomHeadersCorrectlyShouldNotThrowException()
+        {
+            ODataT4CodeGenerator codegen = new ODataT4CodeGenerator();
+            string customHeaderString = @"Authorization:Bearer bearer-token
+                                          odata.continue-on-error:true";
+            codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            Assert.IsNotNull(codegen.CustomHttpHeaders);
+            Assert.AreEqual(2, codegen.CustomHttpHeaders.Count);
+            Assert.AreEqual("Authorization:Bearer bearer-token", codegen.CustomHttpHeaders.FirstOrDefault());
```
odero:
```suggestion
            Assert.AreEqual("Authorization:Bearer bearer-token", codegen.CustomHttpHeaders.First());

```

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
```diff
@@ -294,5 +294,74 @@ public void GetEdmxStringFromMetadataPathShouldThrowExceptionIfInputIsInvalidUri
...
+            Assert.AreEqual("odata.continue-on-error:true", codegen.CustomHttpHeaders.LastOrDefault());
```
odero:
```suggestion
            Assert.AreEqual("odata.continue-on-error:true", codegen.CustomHttpHeaders.Last());

```

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
```diff
@@ -294,5 +294,74 @@ public void GetEdmxStringFromMetadataPathShouldThrowExceptionIfInputIsInvalidUri
...
+        [TestMethod]
+        public void SetCustomHeadersWithQuotesShouldNotThrowException()
+        {
+            // Quotes are sent as part of the header value
+            ODataT4CodeGenerator codegen = new ODataT4CodeGenerator();
+            string customHeaderString = @"If-Match:'67ab43'";
+            codegen.ValidateAndSetCustomHttpHeadersFromString(customHeaderString);
+            Assert.IsNotNull(codegen.CustomHttpHeaders);
+            Assert.AreEqual(1, codegen.CustomHttpHeaders.Count);
+            Assert.AreEqual("If-Match:'67ab43'", codegen.CustomHttpHeaders.FirstOrDefault());
```
odero:
```suggestion
            Assert.AreEqual("If-Match:'67ab43'", codegen.CustomHttpHeaders.First());

```

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: The split method only receives char[] and string[]
https://docs.microsoft.com/en-us/dotnet/api/system.string.split?view=netframework-4.8

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Done

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Done

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Done

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
KenitoInc: Done

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
KenitoInc: Done

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
KenitoInc: Done

test/FunctionalTests/Tests/DataServices/UnitTests/DesignT4UnitTests/CodeGenerationContextTest.cs
KenitoInc: Done

src/CodeGen/ODataT4CodeGenerator.cs
KenitoInc: Considering a header string can contain any character and spaces, adding it to the error message can be very messy

odero: Not sure I see why that is an issue - unless a header is really long and complicated, which is unlikely (you could also truncate it if you're worried about the length). The reason I propose this approach is to make the error message more informative by giving  hint and the dev immediately knows which header is erroneous instead of hunting for it

src/CodeGen/ODataT4CodeGenerator.cs
odero: Interesting. Seems only .netframework doesn't have it as an option. Core and standard both have it

odero:
lgtm though I'd still consider giving the more informative error message

src/CodeGen/ODataT4CodeGenerator.ttinclude
KenitoInc: I have split the functions

paulodero:
A similar pull request #1348 . Would be nice to close both when this PR is approved and merged.

KenitoInc:
Closing this PR since we are no longer maintaining Codegen. This feature was added to OData Connected Service v0.6

Issue #1636, opened 2020-01-17
Backslashes in string literals in IN expressions need to be escaped but not in other expressions like EQ


kenneth-hansen:
Backslashes in string literals in IN expressions need to be escaped but not in other expressions like EQ.  So the string in "x eq '\k'" will become "\k" but "x in ('\k')" will throw because \k is not a valid escape sequence.  Conversely the string in "x eq '\\\\'" will become "\\\\" whereas "x in ('\\\\')" becomes "\\".

### Assemblies affected

7.6.2 (but probably all versions).

### Reproduce steps

Use inputs similar to the description above and observe resulting ConstantNodes or exceptions.

### Expected result

Parsing behavior should be consistent for strings.

### Actual result

Strings in IN expressions needs escaping where other places does not.

xuzhg:
Fixed and merged. Thanks.

PR #1642, opened 2020-01-23
Fix issue #1623: In operators fails on string collection with commas in the string value

xuzhg:
…n the string value

### Issues

*This pull request fixes issue #1623. #1636 *

### Description

* a filter clause like `?$filter=propertyName in ('a','x,y,z')` can't work because the old codes simply split the literal using ','.

* backslash in single quote string should escape to double backslash

* backslash in double quote string should un-escape

This PR fix this issue

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -140,120 +140,92 @@ namespace Microsoft.OData.UriParser
             // 2) single quote could not be part of string value
             // 3) double quote could be part of string value, double quote also could be the starting and ending character.
 
-            // remove the "[" and "]"
+            // remove the '[' and ']'
             string normalizedText = literalText.Substring(1, literalText.Length - 2).Trim();
             int length = normalizedText.Length;
-            bool itemStarted = false;
-            int itemIndex = 0;
-            char[] updated = new char[length + 2];
-            updated[0] = '[';
-            char startingChar = normalizedText[0];
-            int j = 1;
+            StringBuilder sb = new StringBuilder(length + 2);
+            sb.Append('[');
             for (int i = 0; i < length; i++)
             {
-                char character = normalizedText[i];
-                if (character == '"')
+                char ch = normalizedText[i];
+                switch (ch)
                 {
-                    // double quote could be part of the string value
-                    // double quote could the enclosed character, back compatitible
-                    updated[j++] = '"';
-                    if (itemStarted)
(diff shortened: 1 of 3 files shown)


src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -139,72 +132,116 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
...
+                        i = k;
+                        break;
+
+                    case ' ':
+                        // ignore all whitespaces between items
+                        break;
+
+                    case ',':
+                        // for multiple comma(s) between items, for example ('abc',,,'xyz'),
+                        // we let it go and leave the next layer to identify the problem, or remove the duplicated silently?
```
mikepizzo: > or remove the duplicated silently? [](start = 89, length = 35)

Yes; I would leave it to the next layer to handle, not remove silently.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -108,17 +100,18 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
                     if (expectedTypeFullName.Equals("Edm.String"))
```
mikepizzo: >if (expectedTypeFullName.Equals("Edm.String")) [](start = 20, length = 46)

Please open an issue to track the fact that this logic doesn't support use of in with a Collection(Edm.Primitive). Admittedly an edge case, and not what this issue is trying to fix, but something that we should track in case we need to support it in the future.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
xuzhg: Cool. I will change the comments.

---
In reply to: [370714947](https://github.com/OData/odata.net/pull/1642#discussion_r370714947) [](ancestors = 370714947)

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
xuzhg: Done at: [https://github.com/OData/odata.net/issues/1644](https://github.com/OData/odata.net/issues/1644)

---
In reply to: [370718208](https://github.com/OData/odata.net/pull/1642#discussion_r370718208) [](ancestors = 370718208)

Issue #794, opened 2017-04-13
OData v4, Query Options for Initial client query to Bound Functions is carried over to the next query


mikitaka:
When calling Controller bound functions from Native client and we assign Container.Controller.Function() to a local query variable and the query execute the first time, that query carries over the queryoptions (such as $count) to the next subsequent queries.

1st Query : query.Count();
2nd Query: query.Top(10); // here the query variable remembers the first Count() and produces .../$count?$orderby=PK. By right, the second query should be only .../$orderby=PK

This only happens when we use a local variable assigned with the original query.

### Assemblies affected
OData .Net lib 7.0.0  - server side
OData 6.15 - client side

### Reproduce steps

```
//Client Side using T4 template and container
Container dataService = new Container(App.ODataUrl);
var myquery = dataService.Classes.GetTree(101, 728544);
var c = myquery.Count(); // after this query, the variable myquery always carry $count options
var list = myquery.OrderBy(x => x.PK).ToList(); // this will result in /$count?$orderby=PK

```
### Expected result
Every query should be clean without remembering the previously issued query options

### Actual result
A local query variable remembers the query issued before.
This results in strange query uri such as 
http://localhost:55432/odata4/Classes/JEdixOData4Service.GetTree(FKTenant=101,FKParent=728544)**/$count?$top=1**

mikitaka:
To be clearer, linq query is handled properly, just that when it translates to odata uri service string, the previous string is carried over to the next query, resulting in accumulating uri string.

1st time /$count
2nd time /$count?$orderBy=PK ... This is strange queryoption string

See attached.

[ODataClientTest201704.zip](https://github.com/OData/odata.net/files/921670/ODataClientTest201704.zip)
[ODataServerTest201704.zip](https://github.com/OData/odata.net/files/921671/ODataServerTest201704.zip)

Benjamin-Berry:
I'm seeing this with Microsoft.OData.Client 6.17.0

PR #1656, opened 2020-01-30
Avoid carrying over Count query option to the next query.

paulodero:
### Issues

*This pull request fixes issue #794.*

### Description

*When executing `$count` query option, the query option is carried over to the next query as explained in the issue #794.*

- [x] *Test cases added*

### Additional work necessary


habbes:
@paulodero just curious to know whether this issue occurs only for count queries or whether it occurs for other operations that return an aggregate value or even whether it occurs with just all operations.

For example, if you have:
```
var query = container.Classes;
var count = query.Count();
var order = query.OrderBy(x => x.PK);
var where = query.Where(x => x.Foo > 2);
```
Would the query uri generated for `where` contain both a `$count` and `$order` segment in addition to the `$filter` segment (suggesting that the reported issue could affect all operations)? Or would it only contain `$count` and `$filter` (suggesting that the reported issue only affect counts)? Or would it only contain `$filter`?

paulodero:
@habbes No. Only `$count` segment is carried over to the next query.

src/Microsoft.OData.Client/ALinq/NavigationPropertySingletonExpression.cs
```diff
@@ -109,6 +109,25 @@ internal override bool IsOperationInvocation
             get { return false; }
         }
 
+        /// <summary>
+        /// clone the ResourceExpression
```
g2mula:
```suggestion
        /// Clone the ResourceExpression
```
:-)

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
```diff
@@ -8153,6 +8153,34 @@ public void QueryRowCount_UriBuilder()
...
+            string expectedLongCountFunctionUri = "/service.svc/Leagues/LongCountFunction(Param1='Param1',Param2='Param2')?$orderby=ID";
+            string expectedCountFunctionUri = "/service.svc/Leagues/CountFunction(Param1='Param1',Param2='Param2')?$orderby=ID";
+
+            var countQuery = context.CreateQuery<League>("Leagues");
+            var longcountQuery = context.CreateQuery<League>("Leagues");
+
+            var countFunctionQuery = countQuery.CreateFunctionQuery<League>("CountFunction", true, new UriOperationParameter("Param1", "Param1"),
+                    new UriOperationParameter("Param2", "Param2"));
+            var longCountFunctionQuery = longcountQuery.CreateFunctionQuery<League>("LongCountFunction", true, new UriOperationParameter("Param1", "Param1"),
+                    new UriOperationParameter("Param2", "Param2"));
```
g2mula: I personally prefer vars over types but the style in this file is to use the type.
Var could also be used if the whole method adhered to this but can see that we started by declaring a couple of ```string```s. Maybe those could be replaced and then the ```[TestMethod]VerifyCountIsNotCarriedOver()``` becomes internally consistent.

test/FunctionalTests/Tests/DataServices/UnitTests/ServerUnitTests/Stubs/TempDbDataService/TempDbDataService.cs
```diff
@@ -134,7 +134,7 @@ private static EntityConnection GetConnectionForType(BaseModelType type)
             if (!connections.ContainsKey(type) || connections[type] == null)
             {
                 string dbName = Enum.GetName(typeof(BaseModelType), type);
-                string dataSource = DataUtil.DefaultDataSource;
+                string dataSource = DataUtil.DefaultDataSource;
```
g2mula: Maybe this was not meant for checkin?

sln/OData.Net45.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio 15
+VisualStudioVersion = 15.0.28307.1000
```
g2mula: Maybe this was not meant for checkin, this kinda upgrades the solution version for everyone...

sln/OData.Tests.E2E.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio 15
+VisualStudioVersion = 15.0.28307.1000
```
g2mula: Maybe this was not meant for checkin, this kinda upgrades the solution version for everyone...

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
```diff
@@ -8153,6 +8153,34 @@ public void QueryRowCount_UriBuilder()
...
+            try
+            {
+                //check for both count an longcount
+                countFunctionQuery.Count();
+                longCountFunctionQuery.LongCount();
+
+            }
+            catch(DataServiceQueryException)
```
odero: I don't understand this try/catch inside the test. Is it necessary? What would cause the this to fail?

paulodero: The exception is caught because there is not real service being called but the URI is formed. Our interest in this case is the URI constructed which we are able to check within the catch clause.

odero: Shouldn't Count and LongCount be lazy-evaluated though? Calling them shouldn't trigger a network request

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
```diff
@@ -8153,6 +8153,34 @@ public void QueryRowCount_UriBuilder()
...
+            {
+                var countOrderQuery = countFunctionQuery.OrderBy(x => x.ID);
+                var longCountOrderQuery = longCountFunctionQuery.OrderBy(x => x.ID);
+                Assert.IsTrue(countOrderQuery.ToString().EndsWith(expectedCountFunctionUri), "Wrong uri generated with count option");
+                Assert.IsTrue(longCountOrderQuery.ToString().EndsWith(expectedLongCountFunctionUri), "Wrong uri generated with longcount option");
```
odero: Why do a .EndsWith instead of a .Contains?

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
paulodero: They are evaluated while making the call. They are not evaluated beforehand.

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
paulodero: The Query contains the full URL including the host. We are not checking the host. Remember the port of the inmemory service changes on every service instance.

sln/OData.Net45.sln
paulodero: True. Fixed.

sln/OData.Tests.E2E.sln
paulodero: True. Fixed.

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -1313,11 +1313,17 @@ private static Expression AnalyzeCountMethod(MethodCallExpression mce)
                 return mce;
             }
 
+           
+            Console.WriteLine(rse.ToString());
```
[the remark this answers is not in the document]
paulodero: Done. Thanks.

test/FunctionalTests/Microsoft.OData.Client.Tests/Microsoft.OData.Client.Tests.csproj
```diff
@@ -67,6 +67,7 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="ALinq\AddNewEndingTokenVisitorTests.cs" />
+    <Compile Include="ALinq\ResourceBinderTests.cs" />
```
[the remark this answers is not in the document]
paulodero: Unnecessary reference. Removed. Thanks.

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -1313,11 +1313,17 @@ private static Expression AnalyzeCountMethod(MethodCallExpression mce)
...
 
+           
+            Console.WriteLine(rse.ToString());
             ValidationRules.RequireCanAddCount(rse);
             rse.ConvertKeyToFilterExpression();
-            rse.CountOption = CountOption.CountSegment;
 
-            return rse;
+            //Create a copy of ResourceExpression to prevent carrying over 
+            //of count an longcount functions to subsequent queries.
```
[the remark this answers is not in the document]
paulodero: Fixed.

src/Microsoft.OData.Client/ALinq/NavigationPropertySingletonExpression.cs
paulodero: Fixed.

test/FunctionalTests/Tests/DataServices/UnitTests/ServerUnitTests/Stubs/TempDbDataService/TempDbDataService.cs
paulodero: True. No idea what the change means.

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -1313,11 +1313,17 @@ private static Expression AnalyzeCountMethod(MethodCallExpression mce)
...
+            QueryableResourceExpression rseCopy = rse.CreateCloneResourceExpression() as QueryableResourceExpression;
+            rseCopy.CountOption = CountOption.CountSegment;
```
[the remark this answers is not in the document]
paulodero: Fixed

test/FunctionalTests/Tests/DataServices/UnitTests/ServerUnitTests/Stubs/TempDbDataService/TempDbDataService.cs
mikepizzo: Looks like this removes an empty trailing space, which CA would have complained about.  So it's good clean-up in order for the check-in to pass, even if it's not material to the PR.

---
In reply to: [385690196](https://github.com/OData/odata.net/pull/1656#discussion_r385690196) [](ancestors = 385690196)

src/Microsoft.OData.Client/ALinq/ResourceSetExpression.cs
```diff
@@ -110,5 +110,27 @@ protected override QueryableResourceExpression CreateCloneWithNewTypes(Type newT
                 this.OperationParameters,
                 this.IsAction);
         }
+
+        /// <summary>
+        /// Creates a copy of the current ResourceSetExpression. Object references remain the same.
+        /// </summary>
+        /// <returns>A copy of this</returns>
+        internal override QueryableResourceExpression CreateClone()
```
xuzhg: >CreateClone [](start = 54, length = 11)

can this method call the method above?

return CreateClone(this.Type, this.ResourceType)?

xuzhg: Or just to call the Above method in the place where to call "CreateClone()" ?

---
In reply to: [396908061](https://github.com/OData/odata.net/pull/1656#discussion_r396908061) [](ancestors = 396908061)

src/Microsoft.OData.Client/ALinq/SingletonResourceExpression.cs
```diff
@@ -104,5 +104,27 @@ protected override QueryableResourceExpression CreateCloneWithNewTypes(Type newT
                 this.OperationParameters,
                 this.IsAction);
         }
+
+        /// <summary>
+        /// Creates a clone of the current SingletResourceExpression
+        /// </summary>
+        /// <returns>A copy of this </returns>
+        internal override QueryableResourceExpression CreateClone()
```
xuzhg: >reateClone() [](start = 55, length = 12)

same as here?

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
```diff
@@ -8154,6 +8154,34 @@ public void QueryRowCount_UriBuilder()
...
+                    new UriOperationParameter("Param2", "Param2"));
+            var longCountFunctionQuery = longcountQuery.CreateFunctionQuery<League>("LongCountFunction", true, new UriOperationParameter("Param1", "Param1"),
+                    new UriOperationParameter("Param2", "Param2"));
+            try
+            {
+                //check for both count an longcount
+                countFunctionQuery.Count();
+                longCountFunctionQuery.LongCount();
+
+            }
```
xuzhg: Use Assert.Throws<T> , don't use the try/catch

xuzhg:
> namespace AstoriaUnitTests.Stubs

no changes for this file? remove it from the changeset

---
Refers to: test/FunctionalTests/Tests/DataServices/UnitTests/ServerUnitTests/Stubs/TempDbDataService/TempDbDataService.cs:7 in f905a7e. [](commit_id = f905a7e6cd2ecf436cf71608613552595a2d0366, deletion_comment = False)

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
paulodero: Fixed.

src/Microsoft.OData.Client/ALinq/ResourceSetExpression.cs
paulodero: I have called method above as suggested.

src/Microsoft.OData.Client/ALinq/SingletonResourceExpression.cs
paulodero: Fixed as suggested.

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
paulodero: @xuzhg I could not figure out how Assert.Throws could be used in this case since the test seeks to verify the url formed although an exception is expected to be thrown since the Service does not exist. `Count()` and `LongCount()` unfortunately are executed when a service is being called.

xuzhg:
Rebase, squash merge,

---
In reply to: [380007774](https://github.com/OData/odata.net/pull/1656#issuecomment-380007774) [](ancestors = 380007774)

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
odero: @paulodero for a unit test you should not even be making an actual network call. You should be able to fix that by using the `OnMessageCreating` client hook to return a fake response message and avoid the n/w call altogether:

```c#
    // create a custom request message
    class CustomReqMsg : HttpWebRequestMessage
    {
        public CustomReqMsg(DataServiceClientRequestMessageArgs args) : base(args) { }

        public override IODataResponseMessage GetResponse()
        {
            return new HttpWebResponseMessage(
                headers: new Dictionary<string, string>(),
                statusCode: 200,
                getResponseStream: () => new MemoryStream(Encoding.UTF8.GetBytes("2"))  // mimic a .Count response of 2
                );
        }
    }
   ...
   // use the custom request message to override OnMessageCreating
   context.Configurations.RequestPipeline.OnMessageCreating = args => new CustomReqMsg(args);
   // now when you make your `.Count` and `.LongCount` calls they shouldnt hit the n/w
```

paulodero: Better option. On it.

paulodero: Perfect solution! Done!

Issue #1334, opened 2018-11-22
Missing overload of 'IncludeTotalCount' that takes a 'bool'


julealgon:
This is a feature request issue.

`DataServiceQuery<T>` includes a parameterless `IncludeTotalCount` method, which is responsible for adding `$count=true` to the underlying query.

Using this approach, it is impossible to create a `$count=false` query unless one uses the general purpose `AddQueryOption` method explicitly and hardcodes the values.

It is also clunky to conditionally set the count via an outside boolean parameter, as you have to break the query with an `if` check like:
```csharp
var query = new DataServiceContext(...)
    .CreateQuery<Type>("types");
    
if (includeTotalCount)
    query = query.IncludeTotalCount();

query = query
    .Skip(...)
    .Take(...)
    .Select(...);
```

Versus:
```csharp
var query = new DataServiceContext(...)
    .CreateQuery<Type>("types")
    .IncludeTotalCount(includeTotalCount)
    .Skip(...)
    .Take(...)
    .Select(...);
```

### Assemblies affected

Microsoft.OData.Client 7.5.2

### Expected result

New overload of `IncludeTotalCount` that takes a `bool` parameter.
- `IncludeTotalCount(true)` should be identical in behavior to `IncludeTotalCount()`
- `IncludeTotalCount(false)` should set `$count=false` on the underlying query

### Actual result

Impossible to generate `$count=false` unless using the generic and hardcoded `AddQueryOption`.

### Additional detail

Having a way to explicitly set `$count` to false mimics the possible values on the server, as the count property on `ODataQueryOptions` can be one of:
- `null` 
  Equivalent to not calling `IncludeTotalCount()`

- `true`
  Equivalent to `IncludeTotalCount()` or `IncludeTotalCount(true)`

- `false`
  Equivalent to `IncludeTotalCount(false)`

This can be relevant to people manually handling OData query options on the server, as they have full control of the behavior on all 3 branches.

I needed to create an extension myself to have this in place so that I could stress-test our implementation with all 3 possibilities. My version actually takes a `bool?` as that led to easier parameterized testing (but I think a `bool` version would be more fitting on this API):
```csharp
public static DataServiceQuery<T> IncludeTotalCount<T>(
    this DataServiceQuery<T> dataServiceQuery,
    bool? value)
{
    if (dataServiceQuery == null)
        throw new ArgumentNullException(nameof(dataServiceQuery));

    switch (value)
    {
        case true:
[...]

PR #1658, opened 2020-02-04
Add IncludeTotalCount overloading method

KenitoInc:
### Issues

*This pull request fixes issue #1334 .*

### Description
By default, IncludeTotalCount() method handles $count=true only.
We have added a new method IncludeTotalCount(bool countQuery) that accepts a bool parameter. This parameter handles $count=true and $count=false in query options.

*Support for OData v4*
In OData v4, we return `Count` instead of `TotalCount`, we have added functionality to get `Count` from `QueryOperationResponse<T>`. We have deprecated `TotalCount`
We have added methods IncludeCount() and IncludeCount(bool countQuery) and deprecated `IncludeTotalCount`

So the customer should execute the code as follows:
```
DataServiceQuery<Person> query = dsc.CreateQuery<Person>("People");
query = query.IncludeCount(true);
QueryOperationResponse<Person> response = await query.ExecuteAsync() as QueryOperationResponse<Person>;
Console.WriteLine($"There are a total of {response.Count} People.");
```

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Client/DataServiceQueryOfT.cs
```diff
@@ -314,14 +314,27 @@ public DataServiceQuery<TElement> Expand<TTarget>(Expression<Func<TElement, TTar
...
             return (DataServiceQuery<TElement>)this.Provider.CreateQuery<TElement>(
                 Expression.Call(
                     Expression.Convert(this.Expression, typeof(DataServiceQuery<TElement>.DataServiceOrderedQuery)),
                     mi));
         }
 
+        /// <summary>Requests that the count of all entities in the entity set be returned inline with the query results.</summary>
+        /// <returns>A new <see cref="T:Microsoft.OData.Client.DataServiceQuery`1" /> object that has the inline count option set.</returns>
+        /// <param name="includeTotal">Whether to include total count.</param>
+        public DataServiceQuery<TElement> IncludeTotalCount(bool includeTotal)
```
g2mula: The 
C# provides an optional arguments feature which may have some candidate use here
The signature could be
```public DataServiceQuery<TElement> IncludeTotalCount(bool? includeTotal = false)```
[More info](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments#optional-arguments)
However considering that you are doing some reflection stuff to attach the parameterless method above to the expression, maybe old school overloading is how to do it...

File: src/Microsoft.OData.Client/DataServiceQueryOfT.cs [hunk 2]
```diff
@@ -314,14 +314,27 @@ public DataServiceQuery<TElement> Expand<TTarget>(Expression<Func<TElement, TTar
         /// <returns>A new <see cref="T:Microsoft.OData.Client.DataServiceQuery`1" /> object that has the inline count option set.</returns>
         public DataServiceQuery<TElement> IncludeTotalCount()
         {
-            MethodInfo mi = typeof(DataServiceQuery<TElement>).GetMethod("IncludeTotalCount");
+            MethodInfo mi = typeof(DataServiceQuery<TElement>).GetMethods()
+                .FirstOrDefault(m => m.Name == "IncludeTotalCount" && m.GetParameters().Length == 0);
```
g2mula: Consider using ```nameof(IncludeTotalCount)``` as opposed to the string ```"IncludeTotalCount"``` if the name were ever to change, a refactor will be caught and fixed.
[More info](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/nameof)

g2mula: After opting for ```.FirstOrDefault``` you would need to take care of the ```OrDefault``` part of the result, what if ```mi == null```?

xuzhg: +1

---
In reply to: [375320794](https://github.com/OData/odata.net/pull/1658#discussion_r375320794) [](ancestors = 375320794)

src/Microsoft.OData.Client/DataServiceQueryOfT.cs
```diff
@@ -314,14 +314,27 @@ public DataServiceQuery<TElement> Expand<TTarget>(Expression<Func<TElement, TTar
...
         public DataServiceQuery<TElement> IncludeTotalCount()
         {
-            MethodInfo mi = typeof(DataServiceQuery<TElement>).GetMethod("IncludeTotalCount");
+            MethodInfo mi = typeof(DataServiceQuery<TElement>).GetMethods()
+                .FirstOrDefault(m => m.Name == "IncludeTotalCount" && m.GetParameters().Length == 0);
 
             return (DataServiceQuery<TElement>)this.Provider.CreateQuery<TElement>(
                 Expression.Call(
                     Expression.Convert(this.Expression, typeof(DataServiceQuery<TElement>.DataServiceOrderedQuery)),
                     mi));
```
xuzhg: It's better to move all the code into the new overload method. In this old method, call the new one with the default "bool" value.

test/EndToEndTests/Tests/Client/Build.Desktop/QueryCountTests.cs
```diff
@@ -56,6 +56,21 @@ public void CountLinqTest()
...
 
+        /// <summary>
+        /// IncludeTotalCount(true) Test
+        /// </summary>
+        [TestMethod]
+        public void CountWithBoolTrueParamLinqTest()
+        {
+            var context = this.CreateWrappedContext<DefaultContainer>().Context;
+
+            var query = context.Customer.IncludeTotalCount(true);
```
xuzhg: >var query = context.Customer.IncludeTotalCount(true); [](start = 12, length = 53)

should have negative test case to test "false"

src/Microsoft.OData.Client/DataServiceQueryOfT.cs
```diff
@@ -314,14 +314,27 @@ public DataServiceQuery<TElement> Expand<TTarget>(Expression<Func<TElement, TTar
         /// <returns>A new <see cref="T:Microsoft.OData.Client.DataServiceQuery`1" /> object that has the inline count option set.</returns>
         public DataServiceQuery<TElement> IncludeTotalCount()
```
xuzhg: >IncludeTotalCount [](start = 42, length = 17)

TotalCount is OData V3 term, OData V4 is using Count. Consider to refactor as "IncludeCount()"?

src/Microsoft.OData.Client/DataServiceQueryOfT.cs
KenitoInc: Done

KenitoInc: Switched to using `First`

src/Microsoft.OData.Client/DataServiceQueryOfT.cs
KenitoInc: Done

test/EndToEndTests/Tests/Client/Build.Desktop/QueryCountTests.cs
KenitoInc: There is one

src/Microsoft.OData.Client/DataServiceQueryOfT.cs
KenitoInc: @xuzhg Do we still keep the IncludeTotalCount() for backwards compatibility? cc @mikepizzo

src/Microsoft.OData.Client/DataServiceQueryOfT.cs
KenitoInc: I have updated the code to ensure we add `$count=false` to the request Uri whenever we call `IncludeTotal(false)`

src/Microsoft.OData.Client/DataServiceQueryOfT.cs
odero: @KenitoInc That will just throw an exception if there's no "First". Might be better to just stick to `FirstOrDefault `and handle `mi == null`

src/Microsoft.OData.Client/DataServiceQueryOfT.cs
mikepizzo: Assuming that, for OData V4, setting IncludeTotalCount() sets $count=true, then we should keep it for backward compatibility.

xuzhg: Or maybe we can mark the old one as [obsoleted] and add new one as "IncludeCount()". 

I agree both. 

---
In reply to: [381424145](https://github.com/OData/odata.net/pull/1658#discussion_r381424145) [](ancestors = 381424145)

test/EndToEndTests/Tests/Client/Build.Desktop/QueryCountTests.cs
```diff
@@ -278,6 +293,28 @@ public void GetTotalCountInvalidTest()
...
+            {
+                var response = query.Execute() as QueryOperationResponse<Customer>;
+
+                var count = response.TotalCount;
+                Assert.Fail("Should fail on getting customer count");
+            }
+            catch (Exception ex)
+            {
+                Assert.AreEqual(true, ex.Message.Contains("Count value is not part of the response stream"));
+            }
```
xuzhg: We can't use "Assert.Throws(...)" assert?

src/Microsoft.OData.Client/DataServiceQueryOfT.cs
KenitoInc: @mikepizzo @xuzhg 
I have added `QueryOperationResponse<T>.Count` and `IncludeCount`
I have deprecated `QueryOperationResponse<T>.TotalCount` and `IncludeTotalCount`

test/FunctionalTests/Tests/DataOData/Tests/OData.Common.Tests/PublicApi/PublicApi.bsl
```diff
@@ -7973,7 +7973,11 @@ public class Microsoft.OData.Client.MessageWriterSettingsArgs {
 }
 
 public class Microsoft.OData.Client.QueryOperationResponse : Microsoft.OData.Client.OperationResponse, IEnumerable {
+	long Count  { public virtual get; }
 	Microsoft.OData.Client.DataServiceRequest Query  { public get; }
+	[
+	ObsoleteAttribute(),
+	]
```
g2mula: When adding the obsolete attribute, it is good to add a message to be displayed and whether to classify the "obsoletion" as an error or a warning ie
```[ObsoleteAttribute("This property is obsolete, please use Count", false)]```
and have the attribute in one line.
also please specify the access level of the prop: public, private, internal?

Comment (KenitoInc) [reply test/FunctionalTests/Tests/DataOData/Tests/OData.Common.Tests/PublicApi/PublicApi.bsl]: @g2mula This file is generated. It's used by the Astoria end to end tests

File: src/Microsoft.OData.Client/ALinq/ResourceBinder.cs [hunk 8]
```diff
@@ -1061,7 +1061,16 @@ private static Expression AnalyzeAddCountOption(MethodCallExpression mce, CountO
...
+
+            ConstantExpression countQueryOption = StripTo<ConstantExpression>(mce.Arguments[0]);
+            if ((bool)countQueryOption.Value == true)
+            {
+                rse.CountOption = CountOption.CountQueryTrue;
+            }
+            else
+            {
+                rse.CountOption = CountOption.CountQueryFalse;
+            }
```
habbes: What about when count option is neither true or false? i.e. when it should not be explicitly set in the query?

src/Microsoft.OData.Client/DataServiceQueryOfT.cs
```diff
@@ -312,14 +312,41 @@ public DataServiceQuery<TElement> Expand<TTarget>(Expression<Func<TElement, TTar
...
         {
-            MethodInfo mi = typeof(DataServiceQuery<TElement>).GetMethod("IncludeTotalCount");
+            return this.IncludeCount(true);
+        }
+
+        /// <summary>Requests that the count of all entities in the entity set be returned inline with the query results.</summary>
+        /// <returns>A new <see cref="T:Microsoft.OData.Client.DataServiceQuery`1" /> object that has the inline count option set.</returns>
+        /// <param name="countQuery">Whether to include total count.</param>
+        [Obsolete("Please use includeCount(bool countQuery)")]
```
habbes: use the right case of the method name in the Obsolete attribute message
```diff
- Please use includeCount(bool countQuery)
+ Please use IncludeCount(bool countQuery)
```

src/Microsoft.OData.Client/QueryOperationResponseOfT.cs
```diff
@@ -42,6 +42,7 @@ internal QueryOperationResponse(HeaderCollection headers, DataServiceRequest que
 
         /// <summary>The server result set count value from a query, if the query has requested the value.</summary>
         /// <returns>The return value can be either zero or a positive value equal to the number of entities in the set on the server.</returns>
+        [Obsolete("Please use Count")]
         public override long TotalCount
         {
             get
```
habbes: I think you should change this getter to
```c#
[Obsolete(...)]
public override long TotalCount
{
  get
  {
    return Count;
  }
}
```

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
KenitoInc: If it's not set in the query, the call doesn't reach here

src/Microsoft.OData.Client/DataServiceQueryOfT.cs
KenitoInc: fixed

src/Microsoft.OData.Client/QueryOperationResponseOfT.cs
KenitoInc: fixed

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
KenitoInc: `IncludeCount(false)` test case is included. #Resolved

Issue #725, opened 2016-11-21
Microsoft.OData.Client request failure when calling a function requiring array parameter


jamesra:
I am using Microsoft.OData.Client to call a function requiring a collection of integers as a parameter.  The client and server appear to disagree how the array should be represented in the request's URI.

I first asked Stack Overflow with this issue without luck, but here is a link in case an answer shows up later:
[http://stackoverflow.com/questions/40687528/microsoft-odata-client-request-failure-when-calling-a-function-requiring-array-p](url)

### Assemblies affected
Microsoft.OData.Client v6.15 (Client)
Microsoft.OData.Core v7.0 (Server)

### Reproduce steps
My service defines a function that is not bound to a collection: ODataModelBuilder builder;

```
/*Removed unrelated model definitions…*/
var NetworkCellsFuncConfig = builder.Function("NetworkCells");
NetworkCellsFuncConfig.CollectionParameter<long>("IDs");
NetworkCellsFuncConfig.Parameter<int>("Hops");
NetworkCellsFuncConfig.ReturnsCollectionFromEntitySet<Structure>("Structures");
NetworkCellsFuncConfig.Namespace = null;
```
In my controller I define this function:
```

[HttpGet]
[EnableQuery(PageSize = 2048)]
[ODataRoute("NetworkCells(IDs={IDs},Hops={Hops})")]
public IQueryable<Structure> GetNetworkCells([FromODataUri] ICollection<long> IDs, [FromODataUri] int Hops)
{
    Request.ODataProperties().Path = DefaultODataPathHandler().Parse(System.Web.HttpContext.Current.Request.Url.GetLeftPart(System.UriPartial.Path), "Structures", Request.GetRequestContainer()) 
    return db.SelectNetworkStructures(IDs, Hops);
}
```

### Expected result
If I call my service using the URL fragment below I get a correct response:

> …/OData/NetworkCells(IDs=@IDs,Hops=1)?@IDs=[476]

### Actual result
The response is 404 - not found.  When I call the function from the client using Microsoft.OData.Client v6.15 this URL fragment is generated:

> ../OData/NetworkCells(IDs=@IDs,Hops=1)?@IDs=["476"]

### Additional details
The difference is the quotations around “476”. Removing the quotes returns a valid response.

Below is the auto-generated client code from the template:
```
/// <summary>
/// There are no comments for NetworkCells in the schema.
/// </summary>
[global::Microsoft.OData.Client.OriginalNameAttribute("NetworkCells")]
[...]

Issue #522, opened 2016-03-31
OData client always serialize Int64 and Decimal parameters in IEEE754-compatible format.


dingpl716:
_Short summary (3-5 sentences) describing the issue._  
OData client always serialize Int64 and Decimal parameters into string surrounded by double quotes.
Example URL:
https://serverroot/Products(22)/GetSlotComponents(channelId=56,slotId=87,selectedComponents=@selectedComponents)?@selectedComponents=[{"@odata.type":"#ComponentInSlotRelation","ComponentId@odata.type":"#Int64","ComponentId":"22565429523”,”Amount@odata.type":"#Decimal","Amount":"687.19"}]
The value of ComponentId and Amount are surrounded by double quotes which makes the server throw exceptions.  
### Assemblies affected

_Which assemblies and versions are known to be affected e.g. OData .Net lib 6.15-beta._  
### Reproduce steps

_The simplest set of steps to reproduce the issue. If possible, reference a commit that demonstrates the issue._  
1. Create a Function with complex type parameter.
2. Send the request by using following way
Task.Factory.FromAsync<Uri, IEnumerable<T>>(
    (uri, callback, state) => this.BeginExecute<T>(uri, callback, state, httpMethod, false, parameters),
     this.EndExecute<T>,
     requestUri,
     null);
### Expected result

_What would happen if there wasn't a bug._  
The value of ComponentId and Amount are not surrounded by double quotes. It should look like:
"#ComponentInSlotRelation","ComponentId@odata.type":"#Int64","ComponentId":22565429523,”Amount@odata.type":"#Decimal","Amount":687.19
### Actual result

_What is actually happening._  
### Additional details

_Optional, details of the root cause if known. Delete this section if you have no additional details to add._

LaylaLiu:
#725

paulodero:
@dingpl716 to avoid serializing Int64 and Decimal parameters in `IEEE754-compatible` format you will need to pass in `IEEE754-compatible=false` as part of the `Content-Type` header in your request.

PR #1659, opened 2020-02-05
Enable OData client to send IEEE754Compatible parameter in the reques…

paulodero:
…t header to an OData service.

### Issues

*This pull request fixes issue #725 and  #522  *

### Description

*Enable OData client to send IEEE754Compatible parameter in the request header. This is to ensure that both the service and the client are in sync on how long and decimal values in a collection are serialized. In the header, this value can be set to true or false. If set to true then long and decimals will be serialized as strings.*

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -3007,6 +3007,10 @@ namespace Microsoft.OData.Client
                 {
                     this.IsIeee754Compatible = true;
                 }
+                else
+                {
+                    this.IsIeee754Compatible = false;
+                }
 
                 // The reason to clone it is so that users can change the
                 // value after this event is fired.
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -60,6 +60,14 @@ namespace Microsoft.OData.Client
         /// </summary>
         private const string ServiceRootParameterName = "serviceRoot";
 
+        /// <summary>
+        /// string constant for the 'ContentType' header
+        /// </summary>
+        private const string ContentType = "Content-Type";
+
+        /// <summary>string constant for 'IEEE754Compatible' header set to true.</summary>
+        private const string MimeIeee754CompatibleHeaderTrue = "IEEE754Compatible=true";
+
         /// <summary>The client model for the current context instance.</summary>
         private readonly ClientEdmModel model;
 
@@ -693,6 +701,12 @@ namespace Microsoft.OData.Client
         }
 
         /// <summary>
+        /// Gets IsIeee754Compatible header value.
+        /// </summary>
+        internal bool IsIeee754Compatible { get; private set; }
+
+
+        /// <summary>
         /// Indicates whether user is using <see cref="T:Microsoft.OData.Client.DataServiceCollection`1" /> to track changes.
         /// </summary>
         internal bool UsingDataServiceCollection { get; set; }
@@ -1012,6 +1026,7 @@ namespace Microsoft.OData.Client
         /// <returns>A new <see cref="T:Microsoft.OData.Client.DataServiceQuery`1" /> instance that represents the function call.</returns>
         public DataServiceQuery<T> CreateFunctionQuery<T>(string path, string functionName, bool isComposable, params UriOperationParameter[] parameters)
         {
+            this.CreateRequestArgsAndFireBuildingRequest(null, null, new HeaderCollection(), this.HttpStack, null /*descriptor*/);
             Dictionary<string, string> operationParameters = this.SerializeOperationParameters(parameters);
             ResourceSetExpression rse = new ResourceSetExpression(typeof(IOrderedQueryable<T>), null, Expression.Constant(path), typeof(T), null, CountOption.None, null, null, null, null, functionName, operationParameters, false);
             return new DataServiceQuery<T>.DataServiceOrderedQuery(rse, new DataServiceQueryProvider(this), isComposable);
@@ -1026,6 +1041,7 @@ namespace Microsoft.OData.Client
         /// <returns>A new <see cref="T:Microsoft.OData.Client.DataServiceQuerySingle`1" /> instance that represents the function call.</returns>
         public DataServiceQuerySingle<T> CreateFunctionQuerySingle<T>(string path, string functionName, bool isComposable, params UriOperationParameter[] parameters)
         {
+            this.CreateRequestArgsAndFireBuildingRequest(null, null, new HeaderCollection(), this.HttpStack, null /*descriptor*/);
             Dictionary<string, string> operationParameters = this.SerializeOperationParameters(parameters);
             SingletonResourceExpression rse = new SingletonResourceExpression(typeof(IOrderedQueryable<T>), null, Expression.Constant(path), typeof(T), null, CountOption.None, null, null, null, null, functionName, operationParameters, false);
             return new DataServiceQuerySingle<T>(new DataServiceQuery<T>.DataServiceOrderedQuery(rse, new DataServiceQueryProvider(this)), isComposable);
@@ -2982,6 +2998,15 @@ namespace Microsoft.OData.Client
             if (this.HasBuildingRequestEventHandlers)
             {
                 this.InnerBuildingRequest(this, buildingRequestEventArgs);
+
+                string contentType = buildingRequestEventArgs.HeaderCollection.GetHeader(ContentType);
+
+                
+                if (!string.IsNullOrEmpty(contentType) && contentType.ToUpperInvariant().Contains(MimeIeee754CompatibleHeaderTrue.ToUpperInvariant()))
+
+                {
+                    this.IsIeee754Compatible = true;
+                }
 
                 // The reason to clone it is so that users can change the
                 // value after this event is fired.
--- a/src/Microsoft.OData.Client/Serialization/Serializer.cs
+++ b/src/Microsoft.OData.Client/Serialization/Serializer.cs
@@ -750,7 +750,8 @@ namespace Microsoft.OData.Client
             // to determine which IEdmOperationImport to pass to the ODL writer. So the ODL writer is
             // serializing the parameter payload without metadata. Setting the model to null so ODL doesn't
             // do unnecessary validations when writing without metadata.
-            string literal = ODataUriUtils.ConvertToUriLiteral(valueInODataFormat, CommonUtil.ConvertToODataVersion(this.requestInfo.MaxProtocolVersionAsVersion), null /* edmModel */);
+            bool isIeee754Compatible = this.requestInfo.IsIeee754Compatible;
+            string literal = ODataUriUtils.ConvertToUriLiteral(valueInODataFormat, CommonUtil.ConvertToODataVersion(this.requestInfo.MaxProtocolVersionAsVersion), null /* edmModel */, isIeee754Compatible);
 
             // The value from ConvertToUriValue will not be escaped, but will already contain literal delimiters like single quotes, so we
             // need to use our own escape method that will preserve those characters instead of directly calling Uri.EscapeDataString that may escape them.
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
@@ -132,7 +132,8 @@ namespace Microsoft.OData.JsonLight
             Debug.Assert(messageWriterSettings != null, "messageWriterSettings != null");
 
             this.textWriter = textWriter;
-            this.jsonWriter = CreateJsonWriter(messageInfo.Container, textWriter, true /*isIeee754Compatible*/, messageWriterSettings);
+            bool ieee754CompatibleSetToTrue = (messageInfo.MediaType != null) ? messageInfo.MediaType.HasIeee754CompatibleSetToTrue() : false;
+            this.jsonWriter = CreateJsonWriter(messageInfo.Container, textWriter, ieee754CompatibleSetToTrue, messageWriterSettings);
             this.metadataLevel = new JsonMinimalMetadataLevel();
             this.propertyCacheHandler = new PropertyCacheHandler();
         }
--- a/src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
+++ b/src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
@@ -426,13 +426,13 @@ namespace Microsoft.OData
             this.Version = other.Version;
             this.LibraryCompatibility = other.LibraryCompatibility;
             this.MetadataSelector = other.MetadataSelector;
+            this.IsIeee754Compatible = other.IsIeee754Compatible;
 
             this.validations = other.validations;
             this.ThrowIfTypeConflictsWithMetadata = other.ThrowIfTypeConflictsWithMetadata;
             this.ThrowOnDuplicatePropertyNames = other.ThrowOnDuplicatePropertyNames;
             this.ThrowOnUndeclaredPropertyForNonOpenType = other.ThrowOnUndeclaredPropertyForNonOpenType;
             this.ArrayPool = other.ArrayPool;
-            this.IsIeee754Compatible = other.IsIeee754Compatible;
         }
     }
 }
--- a/src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
+++ b/src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
@@ -216,6 +216,13 @@ namespace Microsoft.OData
         }
 
         /// <summary>
+        /// isIEEE754Compatible is used to determine if the message follows IEEE 754 standard.
+
+        /// If this is set to true then long and decimals will be serialized as strings.
+        /// </summary>
+        internal bool IsIeee754Compatible { get; set; }
+
+        /// <summary>
         /// The acceptable charsets used to the determine the encoding of the message.
         /// This is a comma separated list of charsets as specified in RFC 2616, Section 14.2
         /// </summary>
@@ -425,6 +432,7 @@ namespace Microsoft.OData
             this.ThrowOnDuplicatePropertyNames = other.ThrowOnDuplicatePropertyNames;
             this.ThrowOnUndeclaredPropertyForNonOpenType = other.ThrowOnUndeclaredPropertyForNonOpenType;
             this.ArrayPool = other.ArrayPool;
+            this.IsIeee754Compatible = other.IsIeee754Compatible;
         }
     }
 }
--- a/src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
+++ b/src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
@@ -280,6 +280,19 @@ namespace Microsoft.OData
         /// <returns>A string representation of <paramref name="collectionValue"/> to be added to a Url.</returns>
         internal static string ConvertToUriCollectionLiteral(ODataCollectionValue collectionValue, IEdmModel model, ODataVersion version)
         {
+            return ConvertToUriCollectionLiteral(collectionValue, model, version, true);
+        }
+
+        /// <summary>
+        /// Converts a <see cref="ODataCollectionValue"/> to a string for use in a Url.
+        /// </summary>
+        /// <param name="collectionValue">Instance to convert.</param>
+        /// <param name="model">Model to be used for validation. User model is optional. The EdmLib core model is expected as a minimum.</param>
+        /// <param name="version">Version to be compliant with. Collection requires >= V3.</param>
+        /// <param name="isIeee754Compatible">true if value should be IEEE 754 compatible.</param>
+        /// <returns>A string representation of <paramref name="collectionValue"/> to be added to a Url.</returns>
+        internal static string ConvertToUriCollectionLiteral(ODataCollectionValue collectionValue, IEdmModel model, ODataVersion version, bool isIeee754Compatible)
+        {
             ExceptionUtils.CheckArgumentNotNull(collectionValue, "collectionValue");
             ExceptionUtils.CheckArgumentNotNull(model, "model");
 
@@ -292,7 +305,8 @@ namespace Microsoft.OData
                     Validations = ~ValidationKinds.ThrowOnUndeclaredPropertyForNonOpenType,
 
                     // TBD: Should write instance annotations for the literal???
-                    ShouldIncludeAnnotation = ODataUtils.CreateAnnotationFilter("*")
+                    ShouldIncludeAnnotation = ODataUtils.CreateAnnotationFilter("*"),
+                    IsIeee754Compatible = isIeee754Compatible
                 };
 
                 WriteJsonLightLiteral(
@@ -536,6 +550,13 @@ namespace Microsoft.OData
         /// <param name="writeValue">Delegate to use to actually write the value.</param>
         private static void WriteJsonLightLiteral(IEdmModel model, ODataMessageWriterSettings messageWriterSettings, TextWriter textWriter, Action<ODataJsonLightValueSerializer> writeValue)
         {
+
+            IEnumerable<KeyValuePair<string,string>> parameters = new Dictionary<string, string>
+            {
+                { MimeConstants.MimeIeee754CompatibleParameterName, messageWriterSettings.IsIeee754Compatible.ToString() }
+            };
+            ODataMediaType mediaType = new ODataMediaType(MimeConstants.MimeApplicationType, MimeConstants.MimeJsonSubType, parameters);
+
             // Calling dispose since it's the right thing to do, but when created from a custom-built TextWriter
             // the output context Dispose will not actually dispose anything, it will just cleanup itself.
             // TODO: URI parser will also support DI container in the future but set the container to null at this moment.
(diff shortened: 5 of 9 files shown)


src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -132,7 +132,8 @@ internal sealed class ODataJsonLightOutputContext : ODataOutputContext
             Debug.Assert(messageWriterSettings != null, "messageWriterSettings != null");
 
             this.textWriter = textWriter;
-            this.jsonWriter = CreateJsonWriter(messageInfo.Container, textWriter, true /*isIeee754Compatible*/, messageWriterSettings);
+            bool ieee754CompatibleSetToTrue = (messageInfo.MediaType != null) ? messageInfo.MediaType.HasIeee754CompatibleSetToTrue() : false;
```
xuzhg: >: false; [](start = 134, length = 8)

default as "true" for non-regression?

test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/ConvertToUriLiteralTests.cs
```diff
@@ -466,24 +466,37 @@ public void ConvertToUriLiteralCollectionTest()
...
+        [TestMethod]
+        public void ConvertToUriLiteralWhenIeee754CompatibleSetTrue()
+        {
+            string expected = "[\"-9223372036854775808\",\"9223372036854775807\"]";
+
+            var parameter = new ODataCollectionValue() { TypeName = EntityModelUtils.GetCollectionTypeName("Edm.Int64"), Items = new object[] { Int64.MinValue, Int64.MaxValue } };
+
+            string actual = ODataUriUtils.ConvertToUriLiteral(parameter, ODataVersion.V4, null, "true");
+
+            Assert.IsTrue(expected.Equals(actual), "Test failed; Ieee754Compatible was not properly set");
```
xuzhg: >Assert.IsTrue(expected.Equals(actual) [](start = 12, length = 37)

Can't use "Assert.Equal(...)"?

src/Microsoft.OData.Core/Uri/ODataUriUtils.cs
```diff
@@ -95,15 +95,29 @@ public static object ConvertFromUriLiteral(string value, ODataVersion version, I
         /// </summary>
         /// <param name="value">Value to be converted.</param>
         /// <param name="version">Version to be compliant with.</param>
+        /// <param name="isIeee754Compatible">true if value should be IEEE 754 compatible.</param>
+        /// <returns>A string representation of <paramref name="value"/> for use in a Url.</returns>
+        [SuppressMessage("Microsoft.Design", "CA1055:UriReturnValuesShouldNotBeStrings", Justification = "designed to aid the creation on a URI, not create a full one")]
+        public static string ConvertToUriLiteral(object value, ODataVersion version, string isIeee754Compatible)
```
xuzhg: >string isIeee754Compatible) [](start = 85, length = 27)

true/false right? why don't use "bool" type?

src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
```diff
@@ -279,6 +279,19 @@ internal static string ConvertToResourceLiteral(ODataResourceValue resourceValue
...
+
+        /// <summary>
+        /// Converts a <see cref="ODataCollectionValue"/> to a string for use in a Url.
+        /// </summary>
+        /// <param name="collectionValue">Instance to convert.</param>
+        /// <param name="model">Model to be used for validation. User model is optional. The EdmLib core model is expected as a minimum.</param>
+        /// <param name="version">Version to be compliant with. Collection requires >= V3.</param>
+        /// <param name="isIeee754Compatible">true if value should be IEEE 754 compatible.</param>
+        /// <returns>A string representation of <paramref name="collectionValue"/> to be added to a Url.</returns>
+        internal static string ConvertToUriCollectionLiteral(ODataCollectionValue collectionValue, IEdmModel model, ODataVersion version, string isIeee754Compatible)
```
xuzhg: >string isIeee754Compatible [](start = 138, length = 26)

use bool type?

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
```diff
@@ -215,6 +215,12 @@ internal string AcceptableMediaTypes
             }
         }
 
+        /// <summary>
+        /// The is IEEE 754 Compatible is used to determine if the message follows IEEE 754 standard.
+        /// If this is set to true then long and decimals will be serialized as strings.
+        /// </summary>
+        internal string IsIeee754Compatible { get; set; }
```
xuzhg: >string [](start = 17, length = 6)

bool?

xuzhg: or bool? ?

---
In reply to: [375396724](https://github.com/OData/odata.net/pull/1659#discussion_r375396724) [](ancestors = 375396724)

src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
```diff
@@ -506,7 +506,7 @@ internal override Expression VisitConstant(ConstantExpression c)
                 MemberInfo member = typeAnnotation.ElementType.GetField(c.Value.ToString());
                 string memberValue = ClientTypeUtil.GetServerDefinedName(member);
                 ODataEnumValue enumValue = new ODataEnumValue(memberValue, typeNameInEdm ?? typeAnnotation.ElementTypeName);
-                result = ODataUriUtils.ConvertToUriLiteral(enumValue, CommonUtil.ConvertToODataVersion(this.uriVersion), null);
+                result = ODataUriUtils.ConvertToUriLiteral(enumValue, CommonUtil.ConvertToODataVersion(this.uriVersion), null as IEdmModel);
```
xuzhg: >null as IEdmModel [](start = 121, length = 17)

if we change the Ieee... as bool? we can get rid of this.

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2945,6 +2963,13 @@ private BuildingRequestEventArgs FireBuildingRequest(BuildingRequestEventArgs bu
             {
                 this.InnerBuildingRequest(this, buildingRequestEventArgs);
 
+                string contentType = buildingRequestEventArgs.HeaderCollection.GetHeader(ContentType);
+                
+                if(!string.IsNullOrEmpty(contentType) && contentType.Contains(MimeIeee754CompatibleHeaderTrue))
```
g2mula: This performs a case sensitive search for the IEEE754 parameter... from limited [research here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types) it is the value part's case which may or may not have specific meaning.

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -692,6 +703,11 @@ internal ClientEdmModel Model
             get { return this.model; }
         }
 
+        /// <summary>
+        /// Gets the client model.
+        /// </summary>
+        internal string IsIeee754Compatible { get; set; }
```
g2mula: Could use a bool or a bool? (Nullable<bool>) if you also want to distinguish between the parameter not being set at all and the parameter being set to false.

src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
g2mula: could also use [named parameters](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/named-and-optional-arguments#named-arguments) to resolve the ambiguity over which null you mean; something like 
```ConvertToUriLiteral(enumValue, CommonUtil.ConvertToODataVersion(this.uriVersion), model: null)```

Comment (paulodero) [reply src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs]: Noted. changing IsIeee754Compatible parameter to bool fixes the problem.

File: src/Microsoft.OData.Client/DataServiceContext.cs (hunk 7 above)
Comment (paulodero) [reply src/Microsoft.OData.Client/DataServiceContext.cs]: noted and fixed.

File: src/Microsoft.OData.Client/DataServiceContext.cs (hunk 8 above)
Comment (paulodero) [reply src/Microsoft.OData.Client/DataServiceContext.cs]: Fixed.

File: src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs (hunk 6 above)
Comment (paulodero) [reply src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs]: Using `bool `fixes this.

Comment (paulodero) [reply src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs]: Fixed.

File: src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs (hunk 4 above)
Comment (paulodero) [reply src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs]: Fixed.

File: test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/ConvertToUriLiteralTests.cs (hunk 2 above)
Comment (paulodero) [reply test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/ConvertToUriLiteralTests.cs]: fixed.

File: src/Microsoft.OData.Core/Uri/ODataUriUtils.cs [hunk 9]
```diff
@@ -95,15 +95,29 @@ public static object ConvertFromUriLiteral(string value, ODataVersion version, I
         /// </summary>
         /// <param name="value">Value to be converted.</param>
         /// <param name="version">Version to be compliant with.</param>
+        /// <param name="isIeee754Compatible">true if value should be IEEE 754 compatible.</param>
+        /// <returns>A string representation of <paramref name="value"/> for use in a Url.</returns>
+        [SuppressMessage("Microsoft.Design", "CA1055:UriReturnValuesShouldNotBeStrings", Justification = "designed to aid the creation on a URI, not create a full one")]
+        public static string ConvertToUriLiteral(object value, ODataVersion version, bool isIeee754Compatible)
```
xuzhg: >bool isIeee754Compatible [](start = 85, length = 24)

Can we consider to create a "ConvertSetting" class and add "IsIeee754Compatible" as one of this setting.

The advantage is that, in the future if we have more configuration, we don't need to change the syntax of the method, we can just modify the ConvertSetting ?

test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/ConvertToUriLiteralTests.cs
```diff
@@ -484,6 +485,21 @@ public void ConvertToUriLiteralCollectionTest()
...
+            string expectedWhenieee754ParamSetFalse = "[-9223372036854775808,9223372036854775807]";
+            string expectedWhenieee754ParamSetTrue = "[\"-9223372036854775808\",\"9223372036854775807\"]";
+
+            var parameter = new ODataCollectionValue() { TypeName = EntityModelUtils.GetCollectionTypeName("Edm.Int64"), Items = new object[] { Int64.MinValue, Int64.MaxValue } };
+
+            string actualWhenieee754ParamSetFalse = ODataUriUtils.ConvertToUriLiteral(parameter, ODataVersion.V4, null, false);
+            string actualWhenieee754ParamSetTrue = ODataUriUtils.ConvertToUriLiteral(parameter, ODataVersion.V4, null, true);
+
+            Assert.AreEqual(expectedWhenieee754ParamSetFalse, actualWhenieee754ParamSetFalse, "Ieee754Compatible was not properly set");
+            Assert.AreEqual(expectedWhenieee754ParamSetTrue, expectedWhenieee754ParamSetTrue, "Ieee754Compatible was not properly set");
```
xuzhg: >expectedWhenieee754ParamSetTrue, expectedWhenieee754ParamSetTrue [](start = 28, length = 64)

compare the same value?

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
xuzhg: Here i am think it's better to use "bool?" or Nullable<bool> . So we can distinguish between
1) set from customer
2) not set from customer

---
In reply to: [375396872](https://github.com/OData/odata.net/pull/1659#discussion_r375396872) [](ancestors = 375396872,375396724)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
xuzhg: how about this?

---
In reply to: [375393605](https://github.com/OData/odata.net/pull/1659#discussion_r375393605) [](ancestors = 375393605)

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2945,6 +2963,13 @@ private BuildingRequestEventArgs FireBuildingRequest(BuildingRequestEventArgs bu
             {
                 this.InnerBuildingRequest(this, buildingRequestEventArgs);
 
+                String contentType = buildingRequestEventArgs.HeaderCollection.GetHeader(ContentType);
```
odero:
```suggestion
                string contentType = buildingRequestEventArgs.HeaderCollection.GetHeader(ContentType);

```

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
odero: I cant see where `HasIeee754CompatibleSetToTrue` is defined. It doesn't exist in MediaType

src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
```diff
@@ -536,14 +550,23 @@ private static void WriteStartResource(ODataWriter writer, ODataResourceBase res
...
             // Calling dispose since it's the right thing to do, but when created from a custom-built TextWriter
             // the output context Dispose will not actually dispose anything, it will just cleanup itself.
             // TODO: URI parser will also support DI container in the future but set the container to null at this moment.
             ODataMessageInfo messageInfo = new ODataMessageInfo
             {
                 Model = model,
                 IsAsync = false,
-                IsResponse = false
+                IsResponse = false,
+                MediaType = mediaType
```
odero: I'd go with only adding the header + MediaType only when it's necessary - in this case only when `IsIeee754Compatible == true` instead of always having it there. No need adding response headers when not required e.g. when `IsIeee754Compatible  == false`

test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/ConvertToUriLiteralTests.cs
odero:
```suggestion
            Assert.AreEqual(expectedWhenieee754ParamSetTrue, actualWhenieee754ParamSetTrue, "Ieee754Compatible was not properly set");

```

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
paulodero: It is an ODataMediaType extension method defined in [MediaTypeUtils.cs ](https://github.com/paulodero/odata.net/blob/31573526cf9a9359c35f101875fd2f39f5341b26/src/Microsoft.OData.Core/MediaTypeUtils.cs#L324)class

src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
paulodero: I think it is good to be explicit especially when the header has been provided.

test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/ConvertToUriLiteralTests.cs
paulodero: Fixed.

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -692,6 +703,11 @@ internal ClientEdmModel Model
             get { return this.model; }
         }
 
+        /// <summary>
+        /// Gets the client model.
```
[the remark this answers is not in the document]
paulodero: Fixed

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
```diff
@@ -215,6 +215,12 @@ internal string AcceptableMediaTypes
             }
         }
 
+        /// <summary>
+        /// The is IEEE 754 Compatible is used to determine if the message follows IEEE 754 standard.
+        /// If this is set to true then long and decimals will be serialized as strings.
+        /// </summary>
+        internal bool IsIeee754Compatible { get; set; }
```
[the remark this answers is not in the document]
paulodero: Fixed.

test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/ConvertToUriLiteralTests.cs
paulodero: Fixed.

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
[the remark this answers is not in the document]
paulodero: Fixed.

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2945,6 +2963,13 @@ private BuildingRequestEventArgs FireBuildingRequest(BuildingRequestEventArgs bu
             {
                 this.InnerBuildingRequest(this, buildingRequestEventArgs);
 
+                String contentType = buildingRequestEventArgs.HeaderCollection.GetHeader(ContentType);
+                
+                if(!string.IsNullOrEmpty(contentType) && contentType.ToUpperInvariant().Contains(MimeIeee754CompatibleHeaderTrue.ToUpperInvariant()))
+                {
+                    this.IsIeee754Compatible = Ieee754CompatibleHeaderValueTrue;
```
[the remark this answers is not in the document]
paulodero: Fixed.

paulodero:
> You've linked to #725 which doesn’t seem to be the issue this PR fixes
> What about handling Accept headers? You only seem to check content-type
> 
> "If the IEEE754Compatible is specified in the Accept header, we should always honor that." - mikepizzo (OData/WebApi#53 (comment)) Maybe 'Accept' is already supported. Would be good to confirm

@odero That is the correct issue this PR is fixing.

src/Microsoft.OData.Core/Uri/ODataUriUtils.cs
paulodero: @odero an enum would restrict the type of settings. Remember a good design requires an enum to hold related values like days in a week. In this case we are bound to have a number of varied types/values.

paulodero: @xuzhg Do we see an increase in these configurations in future? At the moment we have very few configs for the function.

test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/ConvertToUriLiteralTests.cs
```diff
@@ -484,6 +485,21 @@ public void ConvertToUriLiteralCollectionTest()
...
         }
 
+
+        [TestMethod]
+        public void ConvertToUriLiteralWhenIeee754CompatibleSetTrue()
+        {
+            string expectedWhenieee754ParamSetFalse = "[-9223372036854775808,9223372036854775807]";
+            string expectedWhenieee754ParamSetTrue = "[\"-9223372036854775808\",\"9223372036854775807\"]";
+
+            var parameter = new ODataCollectionValue() { TypeName = EntityModelUtils.GetCollectionTypeName("Edm.Int64"), Items = new object[] { Int64.MinValue, Int64.MaxValue } };
```
g2mula:
```suggestion
            ODataCollectionValue parameter = new ODataCollectionValue
            {
                TypeName = EntityModelUtils.GetCollectionTypeName("Edm.Int64"),
                Items = new object[] { Int64.MinValue, Int64.MaxValue }
            };
```
Style here prefers types over vars
Some newlines also adds some visibility.

src/Microsoft.OData.Core/Uri/ODataUriUtils.cs
```diff
@@ -112,6 +126,22 @@ public static string ConvertToUriLiteral(object value, ODataVersion version)
         /// <returns>A string representation of <paramref name="value"/> for use in a Url.</returns>
         [SuppressMessage("Microsoft.Design", "CA1055:UriReturnValuesShouldNotBeStrings", Justification = "designed to aid the creation on a URI, not create a full one")]
         public static string ConvertToUriLiteral(object value, ODataVersion version, IEdmModel model)
+        {
+            return ODataUriUtils.ConvertToUriLiteral(value, version, model, true /*isIeee754Compatible*/);
+
```
g2mula:
```suggestion
```
Removing excess whitespace.

src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
```diff
@@ -536,14 +550,23 @@ private static void WriteStartResource(ODataWriter writer, ODataResourceBase res
         /// <param name="writeValue">Delegate to use to actually write the value.</param>
         private static void WriteJsonLightLiteral(IEdmModel model, ODataMessageWriterSettings messageWriterSettings, TextWriter textWriter, Action<ODataJsonLightValueSerializer> writeValue)
         {
+
+            IEnumerable<KeyValuePair<string,string>> parameters = new Dictionary<string, string>
+            {
+                {MimeConstants.MimeIeee754CompatibleParameterName, messageWriterSettings.IsIeee754Compatible.ToString() }
```
g2mula:
```suggestion
                { MimeConstants.MimeIeee754CompatibleParameterName, messageWriterSettings.IsIeee754Compatible.ToString() }
```
Whitespace

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -692,6 +700,12 @@ internal ClientEdmModel Model
             get { return this.model; }
         }
 
+        /// <summary>
+        /// Gets IsIeee754Compatible header value.
+        /// </summary>
+        internal bool IsIeee754Compatible { get; private set; }
```
xuzhg: I am wondering why do we need to save "Ieee...Compatible" in a ServiceContext?

A service context is a context for all potential request. The state should not change. Now, you have to change this member every time for every request. I don't think it's a good place for it. 
@Michael Pizzo

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -750,7 +750,9 @@ private string ConvertToEscapedUriValue(string paramName, object value, bool use
             // to determine which IEdmOperationImport to pass to the ODL writer. So the ODL writer is
             // serializing the parameter payload without metadata. Setting the model to null so ODL doesn't
             // do unnecessary validations when writing without metadata.
-            string literal = ODataUriUtils.ConvertToUriLiteral(valueInODataFormat, CommonUtil.ConvertToODataVersion(this.requestInfo.MaxProtocolVersionAsVersion), null /* edmModel */);
+            bool isIeee754Compatible = this.requestInfo.Context.IsIeee754Compatible;
```
xuzhg: >requestInfo [](start = 44, length = 11)

I'd like to add "IsIeee754Com..." into "RequestInfo"?

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
```diff
@@ -215,6 +215,13 @@ internal string AcceptableMediaTypes
             }
         }
 
+        /// <summary>
+        /// isIEEE754Compatible is used to determine if the message follows IEEE 754 standard.
+
+        /// If this is set to true then long and decimals will be serialized as strings.
+        /// </summary>
+        internal bool IsIeee754Compatible { get; set; }
```
xuzhg: >internal bool IsIeee754Compatible { get; set; } [](start = 8, length = 47)

add this into "Clone" method?

src/Microsoft.OData.Core/Uri/ODataUriUtils.cs
xuzhg: So far no, :)


---
In reply to: [383678255](https://github.com/OData/odata.net/pull/1659#discussion_r383678255) [](ancestors = 383678255)

habbes: @paulodero  I think a settings object would also be a better idea even if we only have one option. The reason I think so is because `IsIeee754Compatible` accounts for special case, applies only for numbers. In that sense, it doesn't seem like a "main" option and seem more suited to be part of a larger settings object than standalone parameter.

paulodero: Will be good to create an object for it if we have more settings.

src/Microsoft.OData.Client/DataServiceContext.cs
paulodero: @mikepizzo Do you have any reservations on this?

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
paulodero: Fixed.

xuzhg:
rebase, squash and merge, thanks.

Issue #843, opened 2017-06-19
DataServiceClientException messages should be improved


0xced:
The DataServiceClientException.Message property is a serialized [OData error response](http://docs.oasis-open.org/odata/odata-json-format/v4.0/os/odata-json-format-v4.0-os.html#_Toc372793091). I think this should be improved.

Excerpt from [Exception.Message Property documentation](https://docs.microsoft.com/en-us/dotnet/api/system.exception.message?view=netframework-4.7) (emphasis mine):

> Error messages target the developer who is handling the exception. The text of the Message property should completely describe the error and, when possible, should also explain how to correct the error. Top-level exception handlers may display the message to end-users, so **you should ensure that it is grammatically correct and that each sentence of the message ends with a period**. Do not use question marks or exclamation points. If your application uses localized exception messages, you should ensure that they are accurately translated.

Clearly, a serialized JSON object is not a grammatically correct sentence.

### Assemblies affected
Microsoft.OData.Client version 6.x and 7.x

### Additional details
The Microsoft.OData.Core assembly already has everything needed to decode an OData error response, i.e. `Microsoft.OData.ODataMessageReader.ReadError()` which returns a `Microsoft.OData.ODataError` object. Unfortunately, these mechanisms are not used at all when creating a `DataServiceClientException` which simply [reads the raw stream](https://github.com/OData/odata.net/blob/c10013f444e5b52f5150c47fb49c79d8f0599917/src/Microsoft.OData.Client/BaseSaveResult.cs#L236-L254) as string thus creating a `Microsoft.OData.Client.DataServiceClientException` with terrible `Message` property.

[...]

PR #1664, opened 2020-02-13
Fixing issue #843

paulodero:
### Issues

*This pull request fixes issue #843.*

### Description

*Improving `DataServiceClientException` messages. Currently `DataServiceClientException.Message` property is set to a serialized Odata response error which means the consumer has to parse it before use. Since changing the property `DataServiceClientException.Message`  is a breaking change we would not like to do that at this point in time. Instead, `DataServiceClientException.InnerException` has been set with `Microsoft.OData.ODataErrorException`. This makes it easier to get `Microsoft.OData.ODataError` object which can be useful in getting the error message *

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/BaseSaveResult.cs
+++ b/src/Microsoft.OData.Client/BaseSaveResult.cs
@@ -244,24 +244,20 @@ namespace Microsoft.OData.Client
 
             IDictionary<string, string> headers = new Dictionary<string, string>();
 
-
             using (Stream stream = getResponseStream())
             {
                 if ((null != stream) && stream.CanRead && stream.CanSeek)
                 {
-                    Stream streamClone = new MemoryStream();
-                    stream.CopyTo(streamClone);
-                    stream.Position = 0;
-                    streamClone.Position = 0;
                     // this StreamReader can go out of scope without dispose because the underlying stream is disposed of
-                    message = new StreamReader(streamClone).ReadToEnd();
+                    message = new StreamReader(stream).ReadToEnd();
+                    stream.Position = 0;
                 }
 
                 if (string.IsNullOrEmpty(message))
                 {
                     message = statusCode.ToString();
                 }
-                else if (Util.CanBeJson(message))
+                else if (Util.MayBeJson(message))
                 {
                     headers.Add(ODataConstants.ContentTypeHeader, JsonContentTypeHeader);
                     httpWebResponseMessage = new HttpWebResponseMessage(headers, (int)statusCode, getResponseStream);
--- a/src/Microsoft.OData.Client/BaseSaveResult.cs
+++ b/src/Microsoft.OData.Client/BaseSaveResult.cs
@@ -249,11 +249,8 @@ namespace Microsoft.OData.Client
                 if ((null != stream) && stream.CanRead && stream.CanSeek)
                 {
                     // this StreamReader can go out of scope without dispose because the underlying stream is disposed of
-                    using (StreamReader streamReader = new StreamReader(stream))
-                    {
-                        message = streamReader.ReadToEnd();
-                        stream.Position = 0;
-                    }
+                    message = new StreamReader(stream).ReadToEnd();
+                    stream.Position = 0;
                 }
 
                 if (string.IsNullOrEmpty(message))
--- a/src/Microsoft.OData.Client/BaseSaveResult.cs
+++ b/src/Microsoft.OData.Client/BaseSaveResult.cs
@@ -63,6 +63,9 @@ namespace Microsoft.OData.Client
         /// </summary>
         protected Stream mediaResourceRequestStream;
 
+        /// <summary>application/json Content-Type header</summary>
+        private const string JsonContentTypeHeader = "application/json";
+
         /// <summary>temporary buffer when cache results from CUD op in non-batching save changes</summary>
         protected byte[] buildBatchBuffer;
 
@@ -235,23 +238,41 @@ namespace Microsoft.OData.Client
         internal static DataServiceClientException GetResponseText(Func<Stream> getResponseStream, HttpStatusCode statusCode)
         {
             string message = null;
+            HttpWebResponseMessage httpWebResponseMessage = null;
+            DataServiceClientException dataServiceClientException;
+
+            IDictionary<string, string> headers = new Dictionary<string, string>()
+            {
+                { ODataConstants.ContentTypeHeader,JsonContentTypeHeader}
+            };
+
+            ODataErrorException oDataErrorException = null;
             using (Stream stream = getResponseStream())
             {
                 if ((null != stream) && stream.CanRead)
                 {
+                    Stream streamClone = new MemoryStream();
+                    stream.CopyTo(streamClone);
+                    stream.Position = 0;
+                    streamClone.Position = 0;
                     // this StreamReader can go out of scope without dispose because the underlying stream is disposed of
-                    message = new StreamReader(stream).ReadToEnd();
+                    message = new StreamReader(streamClone).ReadToEnd();
                 }
-            }
 
-            if (string.IsNullOrEmpty(message))
-            {
-                message = statusCode.ToString();
-            }
+                if (string.IsNullOrEmpty(message))
+                {
+                    message = statusCode.ToString();
+                }
+                else
+                {
+                    httpWebResponseMessage = new HttpWebResponseMessage(headers, (int)statusCode, getResponseStream);
+                    ODataMessageReader oDataMessageReader = new ODataMessageReader(httpWebResponseMessage);
+                    oDataErrorException = new ODataErrorException(oDataMessageReader.ReadError());
+                }
 
-            HttpWebResponseMessage httpWebResponseMessage = new HttpWebResponseMessage(null as IDictionary<string,string>,(int)statusCode, getResponseStream);
-            ODataMessageReader oDataMessageReader = new ODataMessageReader(httpWebResponseMessage);
-            return new DataServiceClientException(message, new ODataErrorException(oDataMessageReader.ReadError()), httpWebResponseMessage.StatusCode);
+                dataServiceClientException = new DataServiceClientException(message, oDataErrorException, (int)statusCode);
+            }
+            return dataServiceClientException;
         }
 
         /// <summary>process the batch</summary>
--- a/src/Microsoft.OData.Client/BaseSaveResult.cs
+++ b/src/Microsoft.OData.Client/BaseSaveResult.cs
@@ -240,13 +240,11 @@ namespace Microsoft.OData.Client
             string message = null;
             HttpWebResponseMessage httpWebResponseMessage = null;
             DataServiceClientException dataServiceClientException;
-
-            IDictionary<string, string> headers = new Dictionary<string, string>()
-            {
-                { ODataConstants.ContentTypeHeader,JsonContentTypeHeader}
-            };
-
             ODataErrorException oDataErrorException = null;
+
+            IDictionary<string, string> headers = new Dictionary<string, string>();
+
+
             using (Stream stream = getResponseStream())
             {
                 if ((null != stream) && stream.CanRead)
@@ -263,8 +261,9 @@ namespace Microsoft.OData.Client
                 {
                     message = statusCode.ToString();
                 }
-                else
+                else if (Util.IsJson(message))
                 {
+                    headers.Add(ODataConstants.ContentTypeHeader, JsonContentTypeHeader);
                     httpWebResponseMessage = new HttpWebResponseMessage(headers, (int)statusCode, getResponseStream);
                     ODataMessageReader oDataMessageReader = new ODataMessageReader(httpWebResponseMessage);
                     oDataErrorException = new ODataErrorException(oDataMessageReader.ReadError());
--- a/src/Microsoft.OData.Client/BaseSaveResult.cs
+++ b/src/Microsoft.OData.Client/BaseSaveResult.cs
@@ -261,7 +261,7 @@ namespace Microsoft.OData.Client
                 {
                     message = statusCode.ToString();
                 }
-                else if (Util.IsJson(message))
+                else if (Util.CanBeJson(message))
                 {
                     headers.Add(ODataConstants.ContentTypeHeader, JsonContentTypeHeader);
                     httpWebResponseMessage = new HttpWebResponseMessage(headers, (int)statusCode, getResponseStream);
--- a/src/Microsoft.OData.Client/BaseSaveResult.cs
+++ b/src/Microsoft.OData.Client/BaseSaveResult.cs
@@ -249,7 +249,9 @@ namespace Microsoft.OData.Client
                 message = statusCode.ToString();
             }
 
-            return new DataServiceClientException(message, (int)statusCode);
+            HttpWebResponseMessage httpWebResponseMessage = new HttpWebResponseMessage(null as IDictionary<string,string>,(int)statusCode, getResponseStream);
+            ODataMessageReader oDataMessageReader = new ODataMessageReader(httpWebResponseMessage);
+            return new DataServiceClientException(message, new ODataErrorException(oDataMessageReader.ReadError()), httpWebResponseMessage.StatusCode);
         }
 
         /// <summary>process the batch</summary>
--- a/src/Microsoft.OData.Client/BaseSaveResult.cs
+++ b/src/Microsoft.OData.Client/BaseSaveResult.cs
@@ -248,7 +248,7 @@ namespace Microsoft.OData.Client
             {
                 if ((null != stream) && stream.CanRead && stream.CanSeek)
                 {
-                    // this StreamReader can go out of scope without dispose because the underlying stream is disposed of
+                    // This StreamReader can go out of scope without dispose because the underlying stream is disposed of.
                     message = new StreamReader(stream).ReadToEnd();
                     stream.Position = 0;
                 }
--- a/src/Microsoft.OData.Client/BaseSaveResult.cs
+++ b/src/Microsoft.OData.Client/BaseSaveResult.cs
@@ -249,8 +249,11 @@ namespace Microsoft.OData.Client
                 if ((null != stream) && stream.CanRead && stream.CanSeek)
                 {
                     // this StreamReader can go out of scope without dispose because the underlying stream is disposed of
-                    message = new StreamReader(stream).ReadToEnd();
-                    stream.Position = 0;
+                    using (StreamReader streamReader = new StreamReader(stream))
+                    {
+                        message = streamReader.ReadToEnd();
+                        stream.Position = 0;
+                    }
                 }
 
                 if (string.IsNullOrEmpty(message))
--- a/src/Microsoft.OData.Client/BaseSaveResult.cs
+++ b/src/Microsoft.OData.Client/BaseSaveResult.cs
@@ -247,7 +247,7 @@ namespace Microsoft.OData.Client
 
             using (Stream stream = getResponseStream())
             {
-                if ((null != stream) && stream.CanRead)
+                if ((null != stream) && stream.CanRead && stream.CanSeek)
                 {
                     Stream streamClone = new MemoryStream();
                     stream.CopyTo(streamClone);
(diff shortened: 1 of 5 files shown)


src/Microsoft.OData.Client/Materialization/HttpWebResponseMessage.cs
```diff
@@ -159,11 +154,6 @@ public virtual void SetHeader(string headerName, string headerValue)
         /// <returns>Stream from which the response payload can be read.</returns>
         public virtual Stream GetStream()
         {
-#if DEBUG
```
marabooy: Not sure if these should be removed as they could have been used by certain (users) for debugging purposes only.

src/Microsoft.OData.Client/Materialization/HttpWebResponseMessage.cs
```diff
@@ -26,16 +26,11 @@ public class HttpWebResponseMessage : IODataResponseMessage, IDisposable
         private readonly Func<Stream> getResponseStream;
 
         /// <summary>The response status code.</summary>
-        private readonly int statusCode;
+        private readonly int statusCode;
```
habbes: Remove the empty whitespace.

src/Microsoft.OData.Client/Util.cs
```diff
@@ -597,6 +597,18 @@ internal static Version ToVersion(this ODataProtocolVersion protocolVersion)
...
+        /// <summary>
+        /// Quick check if the <paramref name="input"> is Json.
+        /// Note that this does not check if the input is a well formed Json.
+        /// This can be used to check verify that the string is not xml
+        /// </summary>
+        /// <param name="input">the input string.</param>
+        /// <returns>Return true if the input string can be Json.</returns>
+        public static bool CanBeJson(string input)
+        {
+            return input.Trim().Substring(0, 1).IndexOfAny(new[] { '[', '{' }) == 0;
```
habbes: What if the JSON starts with whitespace, or contains only a number, string, etc.? i.e. what if it's valid JSON that does not begin with `[` or `{`?

test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientQueryTests.cs
```diff
@@ -175,6 +175,8 @@ public void QueryEntityNavigationWithImplicitKeys()
                     Assert.IsTrue(testCase.Value);
                     Assert.AreEqual(400, ex.Response.StatusCode);
                     StringResourceUtil.VerifyDataServicesString(ClientExceptionUtil.ExtractServerErrorMessage(ex), "BadRequest_KeyCountMismatch", "Microsoft.Test.OData.Services.AstoriaDefaultService.Message");
+                    //InnerException for DataServiceClientException must be set with the exception response frpom the server.
+                    Assert.IsTrue(ex.InnerException.InnerException != null, "InnerException for DataServiceClientException has not been set.");
```
habbes: Maybe use `Assert.NotNull` or equivalent if available?

habbes: Is it possible to also test whether the `ex.InnerException.InnerException` contains the expected information?

paulodero:
Docs have been added.

test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientQueryTests.cs
paulodero: > or equivalent if availab

Serves the same purpose.

src/Microsoft.OData.Client/Util.cs
paulodero: Note the `Trim()` function. If it does not begin with `[` or `{` therefore it is not a valid Json object that can be parsed.

test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientQueryTests.cs
paulodero: > Is it possible to also test whether the ex.InnerException.InnerException contains the expected information?

Good point. In this case we will check for a well formed `ODataErrorException ` Object. content is not necessary.

src/Microsoft.OData.Client/Materialization/HttpWebResponseMessage.cs
paulodero: fixed

src/Microsoft.OData.Client/BaseSaveResult.cs
```diff
@@ -235,21 +238,40 @@ internal static BaseSaveResult CreateSaveResult(DataServiceContext context, stri
...
+            IDictionary<string, string> headers = new Dictionary<string, string>();
+
+
             using (Stream stream = getResponseStream())
             {
                 if ((null != stream) && stream.CanRead)
                 {
+                    Stream streamClone = new MemoryStream();
+                    stream.CopyTo(streamClone);
+                    stream.Position = 0;
```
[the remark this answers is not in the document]
paulodero: fixed.

src/Microsoft.OData.Client/Materialization/HttpWebResponseMessage.cs
paulodero: It is not necesary , you should be able to call the function many times. You  can handle stream reads after the call.

src/Microsoft.OData.Client/BaseSaveResult.cs
```diff
@@ -247,7 +247,7 @@ internal static DataServiceClientException GetResponseText(Func<Stream> getRespo
 
             using (Stream stream = getResponseStream())
             {
-                if ((null != stream) && stream.CanRead)
+                if ((null != stream) && stream.CanRead && stream.CanSeek)
```
habbes: I think another alternative would be to simply pass along the clone memory stream to the odata message reader, so that you don't have to bother checking whether the original stream is seekable

e.g.
```c#
message = new StreamReader(cloneStream).ReadToEnd();
cloneStream.Position = 0;
Func<Stream> getCloneResponseStream = () => cloneStream;
// ...
httpWebResponseMessage  = new HttpWebResponseMessage(headers, (int) statusCode, getCloneResponseStream);
```

But if the original `getResponseStream` delegate always returns a new stream instance, then maybe there's no need in going through rewinding or cloning in the first place. But I don't know if that's the case.

test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientQueryTests.cs
```diff
@@ -176,7 +177,10 @@ public void QueryEntityNavigationWithImplicitKeys()
                     Assert.AreEqual(400, ex.Response.StatusCode);
                     StringResourceUtil.VerifyDataServicesString(ClientExceptionUtil.ExtractServerErrorMessage(ex), "BadRequest_KeyCountMismatch", "Microsoft.Test.OData.Services.AstoriaDefaultService.Message");
                     //InnerException for DataServiceClientException must be set with the exception response frpom the server.
-                    Assert.IsTrue(ex.InnerException.InnerException != null, "InnerException for DataServiceClientException has not been set.");
+                    ODataErrorException oDataErrorException = ex.InnerException.InnerException as ODataErrorException;
+                    Assert.IsTrue(oDataErrorException != null, "InnerException for DataServiceClientException has not been set.");
```
odero: Can you used `Assert.ThrowsException` instead?

src/Microsoft.OData.Client/BaseSaveResult.cs
paulodero: getResponseStream  Returns the same instance. The current design of Odata message classes does not allow passing the stream object just like that.

src/Microsoft.OData.Client/BaseSaveResult.cs
```diff
@@ -235,21 +238,40 @@ internal static BaseSaveResult CreateSaveResult(DataServiceContext context, stri
...
+
             using (Stream stream = getResponseStream())
             {
-                if ((null != stream) && stream.CanRead)
+                if ((null != stream) && stream.CanRead && stream.CanSeek)
                 {
+                    Stream streamClone = new MemoryStream();
+                    stream.CopyTo(streamClone);
+                    stream.Position = 0;
+                    streamClone.Position = 0;
```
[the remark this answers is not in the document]
paulodero: The stream is used/read twice.

src/Microsoft.OData.Client/DataServiceRequest.cs
```diff
@@ -114,12 +114,15 @@ internal static IEnumerable<TElement> EndExecute<TElement>(object source, DataSe
             catch (DataServiceQueryException ex)
             {
                 Exception inEx = ex;
+                Exception inExPrev = null;
                 while (inEx.InnerException != null)
                 {
+                    inExPrev = inEx;
```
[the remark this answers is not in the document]
paulodero: The level of `InnerException ` is not predetermined. In the world of `Exceptions`, this can go very deep.

src/Microsoft.OData.Client/Util.cs
```diff
@@ -597,6 +597,18 @@ internal static Version ToVersion(this ODataProtocolVersion protocolVersion)
...
         }
 
+        /// <summary>
+        /// Quick check if the <paramref name="input"> is Json.
+        /// Note that this does not check if the input is a well formed Json.
+        /// This can be used to check verify that the string is not xml
+        /// </summary>
+        /// <param name="input">the input string.</param>
+        /// <returns>Return true if the input string can be Json.</returns>
+        public static bool CanBeJson(string input)
```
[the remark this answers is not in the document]
paulodero: No need to parse Json at this point just to check if the string is not XML. Remember OData response messages can be JSON or XML. This is a simple/Lazy check to eliminate possibility of the string being XML.

test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientQueryTests.cs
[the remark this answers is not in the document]
paulodero: Done.

paulodero: At this point exception is already thrown. What we are checking is the composition of the `Exception `object.

src/Microsoft.OData.Client/DataServiceRequest.cs
```diff
@@ -114,12 +114,15 @@ internal static IEnumerable<TElement> EndExecute<TElement>(object source, DataSe
             catch (DataServiceQueryException ex)
             {
                 Exception inEx = ex;
+                Exception inExPrev = null;
```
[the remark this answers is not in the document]
paulodero: `inExPrev ` is the previous innerException. Any suggestions?

test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientQueryTests.cs
odero: Check how `Assert.ThrowsException` works

src/Microsoft.OData.Client/DataServiceRequest.cs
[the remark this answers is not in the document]
odero: But this doesnt go deeper, it looks like it's looping through exceptions until it finds one. Is inner exception optional? Can one outer exception have multiple inner exceptions?

src/Microsoft.OData.Client/DataServiceRequest.cs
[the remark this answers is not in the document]
odero: Maybe just call it `[current|previous]InnerException` :)

src/Microsoft.OData.Client/BaseSaveResult.cs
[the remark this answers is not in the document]
odero: Dont think you need to clone it for that. If it's seekable you can just reset position to 0 and reread it

src/Microsoft.OData.Client/BaseSaveResult.cs
```diff
@@ -235,21 +238,40 @@ internal static BaseSaveResult CreateSaveResult(DataServiceContext context, stri
...
+            DataServiceClientException dataServiceClientException;
+            ODataErrorException oDataErrorException = null;
+
+            IDictionary<string, string> headers = new Dictionary<string, string>();
+
+
             using (Stream stream = getResponseStream())
             {
-                if ((null != stream) && stream.CanRead)
+                if ((null != stream) && stream.CanRead && stream.CanSeek)
```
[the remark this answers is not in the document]
paulodero: This is a syntactic sugar. Do we have a requirement to use syntactic sugar?

src/Microsoft.OData.Client/Util.cs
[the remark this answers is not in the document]
habbes: Since this check does not guarantee that the payload is indeed json, or that parsing will succeed, why not just let it be parsed as json, and if it's not, an exception will be thrown? Because in this case an exception will still be thrown if the json is not valid.

src/Microsoft.OData.Client/BaseSaveResult.cs
habbes: But the `HttpWebResponseMessage` does not care whether it's the same instance. So if you passed a delegate that returns a different/cloned stream with the same data, it should still work I believe.

Regarding the design of odata messages, it seems like a bit of hack to create a `HttpWebResponseMessage` just so that you can parse the error out of the stream. Maybe this could be something worth refactoring down the line.

src/Microsoft.OData.Client/BaseSaveResult.cs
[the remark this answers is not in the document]
odero: It's the c# way :)

src/Microsoft.OData.Client/Materialization/HttpWebResponseMessage.cs
xuzhg: I think this code here to make sure the "Response Stream" is only used once. You can not use stream multiple time.

---
In reply to: [380228582](https://github.com/OData/odata.net/pull/1664#discussion_r380228582) [](ancestors = 380228582)

src/Microsoft.OData.Client/Util.cs
xuzhg: >CanBeJson [](start = 27, length = 9)

MaybeJson

test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientQueryTests.cs
```diff
@@ -175,6 +176,11 @@ public void QueryEntityNavigationWithImplicitKeys()
                     Assert.IsTrue(testCase.Value);
                     Assert.AreEqual(400, ex.Response.StatusCode);
                     StringResourceUtil.VerifyDataServicesString(ClientExceptionUtil.ExtractServerErrorMessage(ex), "BadRequest_KeyCountMismatch", "Microsoft.Test.OData.Services.AstoriaDefaultService.Message");
+                    //InnerException for DataServiceClientException must be set with the exception response frpom the server.
```
xuzhg: >frpom [](start = 108, length = 5)

typo?

src/Microsoft.OData.Client/BaseSaveResult.cs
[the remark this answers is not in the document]
paulodero: @odero you know why we get the error `Error	CS0019	Operator '&&' cannot be applied to operands of type 'bool?' and 'bool?'` when we try to use your recommended syntax?

src/Microsoft.OData.Client/BaseSaveResult.cs
```diff
@@ -235,21 +238,40 @@ internal static BaseSaveResult CreateSaveResult(DataServiceContext context, stri
...
-            }
+                if (string.IsNullOrEmpty(message))
+                {
+                    message = statusCode.ToString();
+                }
+                else if (Util.CanBeJson(message))
+                {
+                    headers.Add(ODataConstants.ContentTypeHeader, JsonContentTypeHeader);
+                    httpWebResponseMessage = new HttpWebResponseMessage(headers, (int)statusCode, getResponseStream);
+                    ODataMessageReader oDataMessageReader = new ODataMessageReader(httpWebResponseMessage);
```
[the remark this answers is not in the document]
paulodero: Note: // this StreamReader can go out of scope without dispose because the underlying stream is disposed of

src/Microsoft.OData.Client/DataServiceRequest.cs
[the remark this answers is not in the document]
paulodero: It loops through `InnerException ` until it finds the last one. Its like finding the root cause.

src/Microsoft.OData.Client/Util.cs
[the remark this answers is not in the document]
paulodero: At this point in time we do not need to parse the whole string. We only need to eliminate the fact that the response is from the service is not in `OData Atom` format. OData service response can either be `JSON `or `OData Atom` format.

test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientQueryTests.cs
paulodero: I have checked. But I don't think there is any added benefit to restructure the initial design of the whole `try/catch` block.

test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientQueryTests.cs
paulodero: Fixed.

src/Microsoft.OData.Client/BaseSaveResult.cs
```diff
@@ -235,21 +238,40 @@ internal static BaseSaveResult CreateSaveResult(DataServiceContext context, stri
         internal static DataServiceClientException GetResponseText(Func<Stream> getResponseStream, HttpStatusCode statusCode)
         {
             string message = null;
+            HttpWebResponseMessage httpWebResponseMessage = null;
+            DataServiceClientException dataServiceClientException;
+            ODataErrorException oDataErrorException = null;
+
+            IDictionary<string, string> headers = new Dictionary<string, string>();
+
+
```
paulodero: Fixed.

src/Microsoft.OData.Client/BaseSaveResult.cs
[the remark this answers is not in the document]
paulodero: Fixed.

src/Microsoft.OData.Client/DataServiceRequest.cs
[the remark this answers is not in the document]
paulodero: Fixed

src/Microsoft.OData.Client/Util.cs
paulodero: Changed.

src/Microsoft.OData.Client/BaseSaveResult.cs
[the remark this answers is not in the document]
odero: Good catch - hadn't realized that. Makes sense though. I suspect it's because a `bool?` type can also be null whereas `&&` only accepts true/false

src/Microsoft.OData.Client/BaseSaveResult.cs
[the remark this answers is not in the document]
odero: Are you sure about that? Which underlying stream are you referring to here that is being disposed of?
The use case in LN#247 seems different to this one. 
`getResponseStream` is used here to generate the web response message which is then passed into and creates a new message reader. I'd assume they are independently disposed in this case.
In any case, if you're absolutely sure and have tested that it's auto-disposed then no problem. If not, then I'd rather err on the side of caution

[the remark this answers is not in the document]
paulodero: Fixed.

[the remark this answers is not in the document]
paulodero: Reverted back. Adding using causes `stream `to be disposed as well when disposing the `StreamReader `.

Issue #1305, opened 2018-10-26
Unable to call action with complex parameter in complex parameter.


NetTecture:
7.5.1 - the bad tests on complex types continues, actions now totally useless outside simple usages.

Metadata first.

The following action:

```
<Action Name="BulkChange" IsBound="true">
<Parameter Name="bindingParameter" Type="Collection(PropertyCore.InspectionDuty)"/>
<Parameter Name="Updates" Type="PropertyCore.InspectionDutyChanges"/>
</Action>
```

uses complex types:

```
<ComplexType Name="InspectionDutyChanges">
<Property Name="Comment" Type="Edm.String"/>
<Property Name="Changes" Type="Collection(PropertyCore.InspectionDutyChange)"/>
</ComplexType>

<ComplexType Name="InspectionDutyChange">
<Property Name="Operation" Type="Operations.Operations" Nullable="false"/>
<Property Name="Identity" Type="Edm.Guid" Nullable="false"/>
<Property Name="IsDisabled" Type="Edm.Boolean" Nullable="false"/>
<Property Name="SeriesStart" Type="Edm.DateTimeOffset"/>
<Property Name="Interval" Type="Common.Interval"/>
<NavigationProperty Name="Module" Type="PropertyCore.Module"/>
<NavigationProperty Name="Equipment" Type="PropertyCore.Equipment"/>
<NavigationProperty Name="OperatorTask" Type="PropertyCore.OperatorTask"/>
</ComplexType>
```
Saving this change object:

```
            var change = new {
                Updates = new InspectionDutyChanges() {
                    Comment = "comment try",

                    Changes = new InspectionDutyChange [] {
                            new InspectionDutyChange() { Identity = Guid.NewGuid(), IsDisabled = false },
                            new InspectionDutyChange() { Identity = Guid.NewGuid(), IsDisabled = false },
                        },

                }
            };
```

should work. Except...

Microsoft.OData.ODataException: The parameter 'Updates' is of Edm type kind 'Complex'. You cannot call WriteValue on a parameter that is not of Edm type kinds 'Primitive', 'Enum' or 'Complex'.

Ah, yes. It is complex. it can only be complex. Exception. Nope, no.

Stack trace:

Result StackTrace:	
at Microsoft.OData.ODataParameterWriterCore.VerifyCanWriteValueParameter(Boolean synchronousCall, String parameterName, Object parameterValue)
[...]

PR #1669, opened 2020-02-19
Changing CannotWriteValueOnNonValueType error message

paulodero:
### Issues

*This pull request fixes issue #1305 .*

### Description

*You cannot call `WriteValue(..)` function in `ODataParameterWriter` on Complex parameter. Unfortunately the error message you get when you try to do this is confusing. Therefore this pull request rectifies the error message received.*

- [ ] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/Microsoft.OData.Core.txt
```diff
@@ -120,7 +120,7 @@ ODataParameterWriterCore_CannotWriteParameter=WriteValue and CreateCollectionWri
 ODataParameterWriterCore_CannotWriteEnd=WriteEnd can only be called after WriteStart and after the previously created sub-writer has completed.
 ODataParameterWriterCore_CannotWriteInErrorOrCompletedState=The writer is in either the 'Error' or 'Completed' state. No further writes can be performed on this writer.
 ODataParameterWriterCore_DuplicatedParameterNameNotAllowed=The parameter '{0}' has already been written. Duplicate parameter names are not allowed in the parameter payload.
-ODataParameterWriterCore_CannotWriteValueOnNonValueTypeKind=The parameter '{0}' is of Edm type kind '{1}'. You cannot call WriteValue on a parameter that is not of Edm type kinds 'Primitive', 'Enum' or 'Complex'.
+ODataParameterWriterCore_CannotWriteValueOnNonValueTypeKind=The parameter '{0}' is of Edm type kind '{1}'. You cannot call WriteValue on a parameter that is not of Edm type kinds 'Primitive' or 'Enum'.
```
xuzhg: >ODataParameterWriterCore_Canno [](start = 0, length = 30)

if you change the "txt' file, please right click the 
Microsoft.OData.Core.tt 
and
Parameterized.Microsoft.OData.Core.tt
in the Visual studio, select "run the custom tool" to re-generate the corresponding CS file. otherwise, the error message changes can't refresh into the assembly.

Meanwhile, is there any Unit test cases in "OData.Core.Test" project to verify the error message. If no, please add one to verify that.

paulodero: @xuzhg We have not added or removed any string variable. We have only changed the value therefore running the `.tt` files do not change the content of the corresponding `.cs` files.


[Yes, we have tests to verify exception being thrown.](https://github.com/OData/odata.net/blob/a0fc496c360d4feda0f2a27cb1494c52a07e7fe4/test/FunctionalTests/Tests/DataOData/Tests/OData.Writer.Tests/ParameterWriter/ParameterWriterTests.cs#L874-L1207)

Issue #1671, opened 2020-02-20
ArgumentNullException on Any in ODataUriResolver


PatrickHofman:
Running an OData server on .NET Core 3.1. When I call an endpoint (table) that doesn't exist, it gives a 404 as expected. However, the code also raises a ArgumentNullException, which is seemingly ignored.

### Assemblies affected

Microsoft.OData.Core version 7.6.3.

### Reproduce steps

Still not exactly sure what the minimal required steps are to reproduce, I need to narrow down a lot of code for that, which I can't do now. The bug seems obvious enough, see details below.

### Expected result

A HTTP 404 response, but without the underlying exception.

### Actual result

A HTTP 404, with an ignored ArgumentNullException.

### Additional detail

The exception occurs in https://github.com/OData/odata.net/blob/c1c234498c4deaf08c8c4718db760e6ec3356c28/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs#L288 on the call to `Any`. It seems `result` is `null`, which can be true since https://github.com/OData/odata.net/blob/c1c234498c4deaf08c8c4718db760e6ec3356c28/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs#L1083-L1097 can return `null`.

Suggest to add a null check on `result`.

PR #1672, opened 2020-02-20
Fix ArgumentNullException on Any in ODataUriResolver #1671

PatrickHofman:
### Issues

This pull request fixes issue #1671.

### Description

Fix the argument exception.

- [ ] *Test cases added*

### Additional work necessary

None.

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
@@ -285,7 +285,7 @@ namespace Microsoft.OData.UriParser
         public virtual IEnumerable<IEdmOperationImport> ResolveOperationImports(IEdmModel model, string identifier)
         {
             IEnumerable<IEdmOperationImport> results = model.FindDeclaredOperationImports(identifier);
-            if (results.Any() || !EnableCaseInsensitive)
+            if ((results != null && results.Any()) || !EnableCaseInsensitive)
             {
                 return results;
             }


gathogojr:
@PatrickHofman Would you be willing and available to implement the change like @g2mula suggested while also adding the test? We discussed internally and agreed that is a pragmatic and least contentious fix to the issue. If you're not available to follow it through we can slot it into our TODO

gathogojr:
@PatrickHofman Kindly note that this issue is being looked into at the moment. One observation made though is that the only scenario that will result into a `null` being returned from `FindDeclaredOperationImports` is if the Edm model has no entity container...
```
            new EdmModel().FindDeclaredOperationImports("NonExistingOperationImports");
```

Otherwise if the entity container is set and the passed operation import is not found, the call to `container.FindOperationImportsExtended(qualifiedName);` returns a collection for all execution paths. Hopefully the result you're observing is not caused by a subtle bug in your code. For now the fix will involve ensuring the method never returns a `null`.
If you have made any further observation that point to any other scenario that would result into a `null` being returned, kindly share with us so we can do further investigation

PatrickHofman:
@gathogojr Sorry I didn't have any time to respond the last few days. It has been quite busy here.

The problem occurred on my end when I requested a table that didn't exist. As far as I can recollect the exception was ignored later on, but I wondered why it was thrown in the first place and so I found this problem. Thanks for looking in to it. I don't think I will have the time to fix this soon, also because I am not familiar with the project itself. If you can pick this up, it would be highly appreciated.

gathogojr:
@PatrickHofman No problem. Already did a PR to in the list ensure that the method never responds with a null https://github.com/OData/odata.net/pull/1690. I'll merge it as soon as I get approvals

gathogojr:
Fix implemented in this PR https://github.com/OData/odata.net/pull/1690

Issue #1332, opened 2018-11-22
LINQ 'Select()' call not producing expected '$select' query option on expression


julealgon:
Calling `.Select` on a `DataServiceQuery<T>` for a bound function is not correctly generating the OData `$select` in the created Uri. All that is generated in the Uri is the `?` query string character and nothing else.

### Assemblies affected

Microsoft.OData.Client 7.5.2

### Reproduce steps

1. Create an API with a simple bound function on top of a single resource set. 
Sample:

Simple models:
```csharp
public class ProductDetails
{
    public string NonStandardId { get; set; }
    public string OtherField { get; set; }
}

public enum ProductType
{
    Type1,
    Type2
}
```

OData config (Startup.cs):
```csharp
IEdmModel GetEdmModel()
{
    var builder = new ODataConventionModelBuilder();
    
    var productDetailEntityConfig = builder
        .EntitySet<ProductDetail>("ProductDetails")
        .EntityType
        .HasKey(p => p.NonStandardId)
        .Count()
        .Select()
        .Page(null, 100);

    var function = builder.EntityType<ProductDetail>().Collection.Function("All");
    function.Parameter<string>("Param1");
    function.Parameter<string>("Param2");
    function.Parameter<string>("Param3");
    function.Parameter<ProductType>("Param4");
    function.ReturnsCollectionFromEntitySet<ProductDetail>("ProductDetails");
    
    return builder.GetEdmModel()
}

app.UseMvc(routes =>
{
    routes.EnableDependencyInjection();

    routes.MapODataServiceRoute("odata", "api/odata", GetEdmModel());
});
```

Controller method:
```csharp
[ODataRoutePrefix("ProductDetails")]
public class ProductDetailController : ODataController
{
    [AllowAnonymous]
    [ODataRoute("Default.All(Param1={param1},Param2={param2},Param3={param3},Param4={param4})")]
    public ActionResult<IEnumerable<ProductDetail>> Get(
        string param1,
        string param2,
        string param3,
        ProductType param4,
        ODataQueryOptions<ProductDetail> oDataQueryOptions)
    {
        ...
    }
}
```
2. Point into this API from a `DataServiceContext` instance:
```csharp

public static class HttpResponseMessageExtensions
{
    public static async Task<HttpWebResponseMessage> ToHttpWebResponseMessageAsync(this HttpResponseMessage httpResponseMessage)
    {
        var stream = await httpResponseMessage.Content.ReadAsStreamAsync();

[...]

Issue #543, opened 2016-04-25
OData Client Requires Microsoft.OData.Client.Key attribute for the entity, does not support the common key.


rjjkathy:
OData client doesn't support the System.ComponentModel.DataAnnotations.Key

tfabian:
Hi, we are sharing our POCO classes on the server and client side (OData web service and WPF client).  This means for us that the domain layer has a dependency on Microsoft.OData.Client.dll. Entity Framework already uses KeyAttribute from DataAnnotations. So we have to duplicate the key attributes on property level and class level.
In my opinion DataServiceContext should be able to infer the entity key, based on metadata (EDM Key element) or based on convention (*Id). Thx

newclaus:
@ElizabethOkerio 
https://github.com/OData/odata.net/commit/7bbc068193a299ab11316eb7f4bcc273e87b7394#r44804113

PR #1674, opened 2020-02-21
Key attribute support on Primary Key field when using POCO classes with the client

ElizabethOkerio:
### Issues

*This pull request fixes issue #1332 and #543*

### Description
OData client supports 3 ways of defining an entity key:
1. ID
2. type+ID
3. KeyAttribute
 Currently, the client does not support the use of the Key  attribute annotation on a type property when using POCO classes.  Like below: 
<pre>
//POCO class
public class ProductDetails
{
    [Key]
    public string NonStandardId { get; set; }
    public string OtherField { get; set; }
}</pre>
This PR adds support for the use of the Key attribute on a primary key field. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
+++ b/src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
@@ -18,7 +18,7 @@ namespace Microsoft.OData.Client.Metadata
     using Microsoft.OData.Edm;
     using c = Microsoft.OData.Client;
 
-    #endregion Namespaces.
+#endregion Namespaces.
 
     /// <summary>
     /// Utility methods for client types.
@@ -439,8 +439,8 @@ namespace Microsoft.OData.Client.Metadata
             KeyAttribute dataServiceKeyAttribute = customAttributes.OfType<KeyAttribute>().FirstOrDefault();
             List<PropertyInfo> keyProperties = new List<PropertyInfo>();
             PropertyInfo[] properties = ClientTypeUtil.GetPropertiesOnType(type, false /*declaredOnly*/).ToArray();
+ 
             hasProperties = properties.Length > 0;
-
             KeyKind currentKeyKind = KeyKind.NotKey;
             KeyKind newKeyKind = KeyKind.NotKey;
             foreach (PropertyInfo propertyInfo in properties)
@@ -478,7 +478,7 @@ namespace Microsoft.OData.Client.Metadata
                 }
             }
 
-            if (newKeyKind == KeyKind.AttributedKey && keyProperties.Count != dataServiceKeyAttribute.KeyNames.Count)
+            if (newKeyKind == KeyKind.AttributedKey && keyProperties.Count != dataServiceKeyAttribute?.KeyNames.Count)
             {
                 var m = (from string a in dataServiceKeyAttribute.KeyNames
                          where null == (from b in properties
@@ -704,6 +704,12 @@ namespace Microsoft.OData.Client.Metadata
             {
                 keyKind = KeyKind.AttributedKey;
             }
+#if !PORTABLELIB
+            else if (propertyInfo.GetCustomAttributes().OfType<System.ComponentModel.DataAnnotations.KeyAttribute>().Any())
+            {
+                keyKind = KeyKind.AttributedKey;              
+            }
+#endif
             else if (propertyName.EndsWith("ID", StringComparison.Ordinal))
             {
                 string declaringTypeName = propertyInfo.DeclaringType.Name;
--- a/test/FunctionalTests/Microsoft.OData.Client.Tests/Metadata/ClientTypeUtilTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Client.Tests/Metadata/ClientTypeUtilTests.cs
@@ -1,0 +1,55 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ClientTypeUtilTests.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using Microsoft.OData.Client.Metadata;
+using System;
+using Xunit;
+
+namespace Microsoft.OData.Client.Tests.Metadata
+{
+    /// <summary>
+    /// A class test to test whether the use of Key Attribute on a Type property can identify a type as an entity when the conventional key naming methods are not used.
+    /// </summary>
+    public class ClientTypeUtilTests
+    {
+#if !PORTABLELIB
+        [Fact]
+        public void IFTypeProperty_HasKeyAttribute_TypeIsEntity()
+        {
+            //Arrange
+            Type person = typeof(Person);
+            //Act
+            bool actualResult = ClientTypeUtil.TypeOrElementTypeIsEntity(person);
+            //Assert
+            Assert.True(actualResult);
+        }
+
+        [Fact]
+        public void IFTypeProperty_HasNoKeyAttribute_TypeIsNotEntity()
+        {
+            //Arrange
+            Type student = typeof(Student);
+            //Act
+            bool actualResult = ClientTypeUtil.TypeOrElementTypeIsEntity(student);
+            //Assert
+            Assert.False(actualResult);
+        }
+
+        public class Person
+        {
+            [System.ComponentModel.DataAnnotations.Key]
+            public int PId { get; set; }
+            public string Name { get; set; }
+        }
+
+        public class Student
+        {
+            public int SId { get; set; }
+            public string Name { get; set; }
+        }
+#endif
+    }
+}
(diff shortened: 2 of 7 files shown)


src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -704,6 +706,10 @@ private static KeyKind IsKeyProperty(PropertyInfo propertyInfo, KeyAttribute dat
             {
                 keyKind = KeyKind.AttributedKey;
             }
+            else if (propertyInfo.GetCustomAttributes()?.OfType<cm.KeyAttribute>().FirstOrDefault() != null)
```
gathogojr: @ElizabethOkerio I could be wrong but I think `GetCustomAttributes` will at worst return an empty collection so `?` might be redundant here

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -478,7 +480,7 @@ internal static PropertyInfo[] GetKeyPropertiesOnType(Type type, out bool hasPro
                 }
             }
 
-            if (newKeyKind == KeyKind.AttributedKey && keyProperties.Count != dataServiceKeyAttribute.KeyNames.Count)
+            if (newKeyKind == KeyKind.AttributedKey && keyProperties.Count != dataServiceKeyAttribute?.KeyNames.Count)
```
gathogojr: @ElizabethOkerio Is there a specific reason for introducing the `?` operator on this line?

ElizabethOkerio: Yes. So if someone used the property key attribute instead of the type attribute, then this statement will be executed  as the keykind is AttributedKey and the dataserviceKeyAttribute for a key property attribute is null. So I limited this check to only when a type attribute is used.

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
ElizabethOkerio: Fixed

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -709,7 +705,7 @@ private static KeyKind IsKeyProperty(PropertyInfo propertyInfo, KeyAttribute dat
                 keyKind = KeyKind.AttributedKey;
             }
 #if !PORTABLELIB
-            else if (propertyInfo.GetCustomAttributes().OfType<cm.KeyAttribute>().FirstOrDefault() != null)
+            else if (propertyInfo.GetCustomAttributes().OfType<System.ComponentModel.DataAnnotations.KeyAttribute>().FirstOrDefault() != null)
```
g2mula:
```suggestion
            else if (propertyInfo.GetCustomAttributes().OfType<System.ComponentModel.DataAnnotations.KeyAttribute>().Any())
```

test/FunctionalTests/Microsoft.OData.Client.Tests/metadata/ClientTypeUtilTests.cs
```diff
@@ -38,11 +38,12 @@ public void IFTypeProperty_HasNoKeyAttribute_TypeIsNotEntity()
             //Assert
             Assert.Equal(expectedResult, actualResult);
         }
+#endif
 
     }
     public class Person
```
g2mula: Could move this class inside the ClientTypeUtilTests class if you don't want to create it in its own file..

test/FunctionalTests/Microsoft.OData.Client.Tests/metadata/ClientTypeUtilTests.cs
```diff
@@ -38,11 +38,12 @@ public void IFTypeProperty_HasNoKeyAttribute_TypeIsNotEntity()
             //Assert
             Assert.Equal(expectedResult, actualResult);
         }
+#endif
 
     }
     public class Person
     {
-        [m.Key]
+        [System.ComponentModel.DataAnnotations.Key]
```
g2mula: This annotation is only available for ```#if !PORTABLELIB``` so the whole class probably needs to be wrapped in the ```#if```

test/FunctionalTests/Microsoft.OData.Client.Tests/Metadata/ClientTypeUtiTests.cs
```diff
@@ -0,0 +1,58 @@
...
+using Microsoft.OData.Client.Metadata;
+using System;
+using Xunit;
+
+namespace Microsoft.OData.Client.Tests.Metadata
+{
+    /// <summary>
+    /// A class test to test whether the use of Key Attribute on a Type property can identify a type as an entity when the conventional key naming methods are not used.
+    /// </summary>
+    public class ClientTypeUtilTests
```
g2mula: The class name does not match the Filename

test/FunctionalTests/Microsoft.OData.Client.Tests/Microsoft.OData.Client.Tests.Netcore.csproj
```diff
@@ -9,8 +9,8 @@
     <SignAssembly>True</SignAssembly>
     <AssemblyOriginatorKeyFile>..\..\..\tools\StrongNamePublicKeys\testkey.snk</AssemblyOriginatorKeyFile>
     <OutputPath>..\..\..\bin\AnyCPU\$(Configuration)\Test\.NETPortable</OutputPath>
+    <DefineConstants>$(DefineConstants);PORTABLELIB;</DefineConstants>
```
g2mula: Maybe this wasn't meant for checkin?

test/FunctionalTests/Microsoft.OData.Client.Tests/Microsoft.OData.Client.Tests.csproj
```diff
@@ -21,7 +21,7 @@
     <DebugType>full</DebugType>
     <Optimize>false</Optimize>
     <OutputPath>..\..\..\bin\AnyCPU\Debug\Test\Desktop\</OutputPath>
-    <DefineConstants>DEBUG;TRACE</DefineConstants>
+    <DefineConstants>TRACE;DEBUG</DefineConstants>
```
g2mula: This change doesn't seem to do anything, maybe it wasn't meant for checkin?

test/FunctionalTests/Microsoft.OData.Client.Tests/Metadata/ClientTypeUtiTests.cs
```diff
@@ -0,0 +1,58 @@
...
+        [Fact]
+        public void IFTypeProperty_HasKeyAttribute_TypeIsEntity()
+        {
+            //Arrange
+            Type person = typeof(Person);
+            bool expectedResult = true;
+            //Act
+            bool actualResult = ClientTypeUtil.TypeOrElementTypeIsEntity(person);
+            //Assert
+            Assert.Equal(expectedResult, actualResult);
```
g2mula:
```suggestion
            Assert.IsTrue(actualResult);
```
Could assert directly to the expected value...

test/FunctionalTests/Microsoft.OData.Client.Tests/Metadata/ClientTypeUtiTests.cs
```diff
@@ -0,0 +1,58 @@
...
+        [Fact]
+        public void IFTypeProperty_HasNoKeyAttribute_TypeIsNotEntity()
+        {
+            //Arrange
+            Type student = typeof(Student);
+            bool expectedResult = false;
+            //Act
+            bool actualResult = ClientTypeUtil.TypeOrElementTypeIsEntity(student);
+            //Assert
+            Assert.Equal(expectedResult, actualResult);
```
g2mula:
```suggestion
            Assert.IsFalse(expectedResult, actualResult);
```
Could assert directly to expected false value...

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -18,7 +18,7 @@ namespace Microsoft.OData.Client.Metadata
     using Microsoft.OData.Edm;
     using c = Microsoft.OData.Client;
 
-    #endregion Namespaces.
+#endregion Namespaces.
```
marabooy: Probably wasn't meant for checkin

test/FunctionalTests/Microsoft.OData.Client.Tests/Metadata/ClientTypeUtilTests.cs
```diff
@@ -0,0 +1,55 @@
...
+using Microsoft.OData.Client.Metadata;
+using System;
+using Xunit;
+
+namespace Microsoft.OData.Client.Tests.Metadata
+{
+    /// <summary>
+    /// A class test to test whether the use of Key Attribute on a Type property can identify a type as an entity when the conventional key naming methods are not used.
+    /// </summary>
+    public class ClientTypeUtilTests
```
marabooy: Test should include the method to see if the client will still read the key property and give out the right value for it. In this case there should be a test to get if it gets the Key as PId and would love to see a test that tests for a key name which is non conventional as the Attribute should sort this out.

PR #1690, opened 2020-03-03
Return empty enumerable from FindDeclaredOperationImports instead of null

gathogojr:
### Issues

*This pull request fixes issue #1671 .*

### Description

Based on the docstring for `FindDeclaredOperationImports` function, an empty enumerable should be returned if no operation imports are found.
```<returns>All operation imports that can be found by the specified name, returns an empty enumerable if no operation import exists.</returns>```
In a scenario where the method is invoked against an Edm model that has no entity container, a `null` is returned instead. This PR addresses that issue such that all execution paths for the method always return a non-null enumerable.

- [x] *Test cases added*


gathogojr:
Build passes in ADO but status was not reported to GitHub. Merging PR

gathogojr:
PR recreated here https://github.com/OData/odata.net/pull/1696

PR #1695, opened 2020-03-09
Support Mocking in OData Client.

ElizabethOkerio:
### Issues

*This pull request fixes issue #1250.*

### Description
The DataServiceContext, DataServiceQuery, DataServiceQueryOfT and DataServiceQuerySingleOfT classes cannot be mocked as they are. In this PR, I have made these classes' public methods and properties virtual to support mocking. 
Virtual methods can be mocked using any Mock Framework. Example, using the Moq Framework
You can mock the <code>AddObject() </code> method in the <code> Microsoft.OData.Client.DataServiceContext </code> class like below: 

<pre>
using Moq; 
using Xunit
namespace SampleMock
{
   /// How to mock the DataServiceContext class methods using moq framework example. 
    public class DataServiceContextMockTests 
    { 
        [Fact] 
        public void VerifyIfDataServiceContextAddObjectMethodIsMockable() 
        { 
            string str = "user"; 
            Mock<DataServiceContext> mockObject = new Mock<DataServiceContext>(); 
            mockObject.CallBase = true; 
            mockObject.Setup(x => x.AddObject(It.IsAny<string>(), It.IsAny<object>())).Callback(() => { str = "Saved Successfully"; }); 
            Users users = new Users(); 
            string result = users.SaveUser(mockObject.Object); 
            Assert.Equal(str, result); 
        }  
        /// A sample implementation class with methods calling the DataserviceContext.cs methods 
        public class Users 
        { 
            /// A simple method to save a user's details  
            /// A concrete class in the Microsoft.OData.Client library with implemented methods that this method calls.         
            public string SaveUser(DataServiceContext dataServiceContext) 
            { 
                string str = "user"; 
                object obj = null; 
                dataServiceContext.AddObject(str, obj); 
                return "Saved Successfully"; 
            } 
        } 
    }
</pre>
- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -273,7 +273,7 @@ namespace Microsoft.OData.Client
         /// When calling BeginSaveChanges and not using SaveChangesOptions.BatchWithSingleChangeset and SaveChangesOptions.BatchWithIndependentOperations,
         /// this event may be raised from a different thread.
         /// </remarks>
-        public event EventHandler<SendingRequest2EventArgs> SendingRequest2;
+        public virtual event EventHandler<SendingRequest2EventArgs> SendingRequest2;
 
         /// <summary>
         /// This event is fired before a request message object is built, giving
@@ -287,7 +287,7 @@ namespace Microsoft.OData.Client
         /// When calling BeginSaveChanges and not using SaveChangesOptions.BatchWithSingleChangeset and SaveChangesOptions.BatchWithIndependentOperations,
         /// this event may be raised from a different thread.
         /// </remarks>
-        public event EventHandler<BuildingRequestEventArgs> BuildingRequest
+        public virtual event EventHandler<BuildingRequestEventArgs> BuildingRequest
         {
             add
             {
@@ -309,7 +309,7 @@ namespace Microsoft.OData.Client
         /// For inner batch operations, the event is also fired before any processing is done, with
         /// the exception that the content-ID of a changeset operation will be read before the event is fired.
         /// </remarks>
-        public event EventHandler<ReceivingResponseEventArgs> ReceivingResponse;
+        public virtual event EventHandler<ReceivingResponseEventArgs> ReceivingResponse;
 
         /// <summary>
         /// This event fires when SaveChanges or EndSaveChanges is called
@@ -330,7 +330,7 @@ namespace Microsoft.OData.Client
 
         /// <summary>Gets or sets the delegate method that is used to resolve the entity set URI when the value cannot be determined from an edit-link or self-link URI.</summary>
         /// <returns>A delegate that takes a <see cref="T:System.String" /> and returns a <see cref="T:System.Uri" /> value.</returns>
-        public Func<String, Uri> ResolveEntitySet
+        public virtual Func<String, Uri> ResolveEntitySet
         {
             get
             {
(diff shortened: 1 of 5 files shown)


test/FunctionalTests/Microsoft.OData.Client.Tests/DataServiceContextMockTests.cs
```diff
@@ -0,0 +1,52 @@
...
+        public class Users
+        {
+            /// <summary>
+            /// A simple method to save a user's details
+            /// </summary>
+            /// <param name="DataServiceContext">
+            /// A concrete class in the Microsoft.OData.Client library with implemented methods that this method calls.        
+            /// </param>
+            /// <returns>
+            /// return "Saved Successfully if the AddObject method runs successfully".
```
paulodero:
```suggestion
            /// return "Saved Successfully" if the AddObject method runs successfully.
```

test/FunctionalTests/Microsoft.OData.Client.Tests/Microsoft.OData.Client.Tests.Netcore.csproj
```diff
@@ -15,6 +15,7 @@
     <PackageReference Include="FluentAssertions" Version="4.19.2" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
     <PackageReference Include="Microsoft.NETCore.Portable.Compatibility" Version="1.0.1" />
+    <PackageReference Include="Moq" Version="4.10.1" />
```
paulodero: Moq version should be the latest `4.13.1` just like the one you have used in `Microsoft.OData.Client.Tests` project I believe. And again, do we have any test case added in this project that requires `Moq`?

ElizabethOkerio: Yes. the lastest Moq version(4.13.1) supports netcoreapp2.0 but this project targets netcoreapp1.0 and this is the most up to date Moq version that is compatible with netcoreapp1.0

ElizabethOkerio: Yes there is a sample test to shoe how these methods can be mocked.

paulodero: In that case is it good to consistently use version `4.10.1` in all projects? I can only see tests added for `Microsoft.OData.Client.Tests` project and not `Microsoft.OData.Client.Tests.Netcore` project. This brings the question if we need `Moq `in `Microsoft.OData.Client.Tests.Netcore` project.

ElizabethOkerio: Basically for the test to run in a .netcore application, then the <code>Moq framework </code> 
is needed in <code>Microsoft.OData.Clieny.Tests.Netcore</code>

ElizabethOkerio: I don't see why we cannot use the latest <code>Moq Framework </code> in <code>
Microsoft.OData.Client.Tests</code> if it is supported.

weitzhandler:
Related: https://github.com/OData/ODataConnectedService/issues/93

PR #1700, opened 2020-03-10
Fix null reference error when matching binding type

KanishManuja-MS:
### Issues

This pull request fixes issue reported by Microsoft Graph

### Description

This is an experiment. The nuget package from this build will be used to verify if the problem stops from happening.

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
@@ -134,8 +134,8 @@ namespace Microsoft.OData.UriParser
         /// <returns>True if the token was bound successfully, or false otherwise.</returns>
         internal static bool TryBindAsOperation(PathSegmentToken pathToken, IEdmModel model, IEdmStructuredType entityType, out ODataPathSegment segment)
         {
-            Debug.Assert(pathToken != null, "pathToken != null");
-            Debug.Assert(entityType != null, "bindingType != null");
+            ExceptionUtils.CheckArgumentNotNull(pathToken, "pathToken");
+            ExceptionUtils.CheckArgumentNotNull(entityType, "entityType");
 
             IEnumerable<IEdmOperation> possibleFunctions = Enumerable.Empty<IEdmOperation>();
             IList<string> parameterNames = new List<string>();
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ModelBuildingHelpers.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ModelBuildingHelpers.cs
@@ -184,21 +184,17 @@ namespace Microsoft.OData.Tests.UriParser
             EdmCollectionTypeReference collectionTypeReference = new EdmCollectionTypeReference(new EdmCollectionType(new EdmEntityTypeReference(vegetableType, false)));
 
             EdmEntityContainer container = new EdmEntityContainer("Test", "Container");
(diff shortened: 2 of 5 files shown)


xuzhg:
rebase & squash

src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
```diff
@@ -178,7 +178,7 @@ internal static bool TryBindAsOperation(PathSegmentToken pathToken, IEdmModel mo
             }
 
             // Only filter if there is more than one and its needed.
-            if (possibleFunctions.Count() > 1)
+            if (possibleFunctions.Count() > 1 && entityType != null)
```
xuzhg: >&& entityType != null [](start = 46, length = 21)

That's test at beginning of the function. It seems unnecessary?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/FunctionOverloadResolverTests.cs
```diff
@@ -54,6 +54,17 @@ public void ActionWithMultipleOverloadsForTheSameBindingParameter()
         }
 
         [Fact]
+        public void OperationOverloadsWithSameNameWithoutBindingTypeS()
```
xuzhg: >S [](start = 68, length = 1)

typo?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ModelBuildingHelpers.cs
```diff
@@ -176,6 +176,33 @@ public static EdmModel GetModelWithIllegalActionOverloads()
...
+            EdmEntityContainer container = new EdmEntityContainer("Test", "Container");
+            var set = container.AddEntitySet("Vegetables", vegetableType);
+
+            var function1 = new EdmFunction("Test", "Function", new EdmEntityTypeReference(vegetableType, true), false /*isBound*/, null /*entitySetPath*/, false);
+            function1.AddParameter("p1", collectionTypeReference);
+            function1.AddParameter("p2", EdmCoreModel.Instance.GetInt32(false));
+
+            var function2 = new EdmFunction("Test", "Function", new EdmEntityTypeReference(vegetableType, true), false /*isBound*/, null /*entitySetPath*/, false);
+            function2.AddParameter("p1", new EdmEntityTypeReference(vegetableType, true));
+            function2.AddParameter("p2", EdmCoreModel.Instance.GetInt32(false));
```
mikepizzo: Does this pass model validation? It should not -- you shouldn't be able to have two overloads of the same unbound function with the same parameter names.

mikepizzo:
I'm okay with the changes as being more protective, but the added test case seems dubious; by spec, we should not allow multiple overloads of the same nonbinding function name with the same parameter names.  I assume this is just testing the logic for unbound functions and *not* a real life scenario, but would also be interested if calling Validate() on the model succeeded (if so, we should file a separate issue to add this check to model validation).

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ModelBuildingHelpers.cs
KanishManuja-MS: Yes. I will follow up on that. In the meanwhile, it is a valid scenario for action overloads which differ by the binding type and there the parameter name does not even matter. Hypothetically, can the parameter name be the same but differ in type for a bound action.

To be on the safe side, I changed the above test to be a valid scenario with a different name. I will follow up with the behavior in model validation separately.

Issue #1398, opened 2019-02-07
[T4] Add CodeGen support for emitting dynamic properties on open types


pedropaulovc:
In OData v4 Web API, the [docs](http://odata.github.io/WebApi/#open-complex-type) instruct developers that
> In order to build an open complex type, you should change the CLR class by adding an `IDictionary<string, object>` property, the property name can be any name.

This property can then be interpreted by an EDM builder to emit `OpenType="true"` for such complex type.

The OData Client Code Generator does not have support for performing the opposite operation: from a complex type annotated with `OpenType="true"`, emit code like

```cs
public Dictionary<string, object> DynamicProperties { get; set; }
```

### Assemblies affected

All versions of the OData Client Code Generator tool.

### Reproduce steps
1. Create a CSDL document like the one below

```xml
<?xml version="1.0" encoding="utf-8"?>
<edmx:Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
  <edmx:DataServices>
    <Schema Namespace="OpenTypeRepro" xmlns="http://docs.oasis-open.org/odata/ns/edm">
      <ComplexType Name="MyOpenType" OpenType="true" />
    </Schema>
  </edmx:DataServices>
</edmx:Edmx>
```

2. Run the OData Client Code Generator tool

### Expected result

The code generator's configuration would let developers optionally enable the generation of dynamic properties:

```cs
// This flag indicates whether to generate an additional Dictionary<string, object> property
// to store dynamic properties in open types.  The value must be set to true or false.
public const bool GenerateDynamicPropertiesCollection = true;

// This defines the name of Dictionary<string, object> properties
// that store dynamic properties in open types. This property is only applicable if
// GenerateDynamicPropertiesCollection is set to true. This value must be
// a valid identifier name.
public const string DynamicPropertiesCollectionName = "DynamicProperties";
```

If enabled, the generated code would look like:

```cs
namespace OpenTypeRepro
{
    /// <summary>
    /// There are no comments for MyOpenType in the schema.
    /// </summary>
    [global::Microsoft.OData.Client.OriginalNameAttribute("MyOpenType")]
    public partial class MyOpenType
    {
        /// <summary>
        /// Holds dynamically generated properties for open type <see cref="global::OpenTypeRepro.MyOpenType"/>.
        /// </summary>
        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.OData.Client.Design.T4", "7.5.1")]
[...]

PR #1712, opened 2020-03-18
Implement client support for dynamic properties dictionary

gathogojr:
### Issues

*This pull request implements support for feature request #1398 .*

### Description

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Client/Materialization/StructuralValueMaterializationPolicy.cs
```diff
@@ -275,5 +275,135 @@ internal void MaterializeDataValues(ClientTypeAnnotation actualType, IEnumerable
...
+            // Dictionary not found or key with matching name already exists
+            if (!ClientTypeUtil.TryGetDynamicPropertiesDictionary(instance, out dynamicPropertiesDictionary) 
+                || dynamicPropertiesDictionary.ContainsKey(property.Name))
+            {
+                return;
+            }
+
+            object value = property.Value;
+
+            // Handle primitive type
```
xuzhg, line 346: > Handle primitive type [](start = 14, length = 22)

untyped or primitive?

src/Microsoft.OData.Client/Materialization/StructuralValueMaterializationPolicy.cs
```diff
@@ -275,5 +275,135 @@ internal void MaterializeDataValues(ClientTypeAnnotation actualType, IEnumerable
...
+            ODataUntypedValue untypedVal = value as ODataUntypedValue;
+            if (untypedVal != null)
+            {
+                value = CommonUtil.ParseJsonToPrimitiveValue(untypedVal.RawValue);
+                dynamicPropertiesDictionary.Add(property.Name, value);
+                return;
+            }
+
+            // Handle enum value
+            ODataEnumValue enumVal = property.Value as ODataEnumValue;
```
xuzhg, line 356: >property.Value [](start = 37, length = 14)

use "value"

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -765,5 +776,59 @@ private static bool IsOverride(Type type, PropertyInfo propertyInfo)
...
+            {
+                return false;
+            }
+
+            dynamicPropertiesDictionary = (IDictionary<string, object>)propertyInfo.GetValue(instance);
+
+            // Is property initialized?
+            if (dynamicPropertiesDictionary == null)
+            {
+                return false;
```
xuzhg, line 828: >return false; [](start = 16, length = 13)

if it's not initialized, can you create the dict and return it?

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -765,5 +776,59 @@ private static bool IsOverride(Type type, PropertyInfo propertyInfo)
...
+        internal static bool TryGetDynamicPropertiesDictionary(object instance, out IDictionary<string, object> dynamicPropertiesDictionary)
+        {
+            Debug.Assert(instance != null, $"{nameof(instance)} != null");
+
+            dynamicPropertiesDictionary = default(IDictionary<string, object>);
+
+            PropertyInfo propertyInfo = instance.GetType().GetProperties().Where(p =>
+                                !p.GetCustomAttributes(typeof(IgnoreClientPropertyAttribute), true).Any() &&
+                                typeof(IDictionary<string, object>).IsAssignableFrom(p.PropertyType)
+                            ).FirstOrDefault();
```
xuzhg, line 816: Can we consider to distinguish the container from customized dict? for example, there are two "IDictionary<string, object>", so far, you pick the first one, can we have an attribute on the property to identify A is the container, B is not the container?

xuzhg:
Looks good to me

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
gathogojr: @xuzhg I'll look into this. Initializing the dictionary though would only be relevant during materialization. In a serialization scenario, the developer would have had to initialize the property in order to add the dynamic properties

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
gathogojr: @xuzhg Thank you for the suggestion. I'll look into it

gathogojr:
@unchase Based on your comment [here](https://github.com/OData/odata.net/issues/1398#issuecomment-533782691), you may want to keep track of this PR, plus [this](https://github.com/OData/ODataConnectedService/pull/83) related PR in OData Connected Service repo.

Unchase OData Connected Service extension already emits the dynamic properties dictionary and users could be using it in different ways - e.g. they could be stashing dynamic properties there and riding on `OnEntryStarting` hook to convert those properties to `ODataProperty`s. However, there is [this](https://github.com/OData/odata.net/pull/1712#discussion_r405683898) suggestion which if implemented would ensure that this change would never break their code.

In addition, any feedback you may have on this change would be welcome. Expect some more changes to the PR (hopefully not much) plus feature docs in the coming days

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
gathogojr: @xuzhg Addressed your review feedback

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
gathogojr: @xuzhg Addressed your review feedback

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/ODataDynamicPropertyTests.cs
```diff
@@ -0,0 +1,785 @@
...
+
+            var materializedObject = MaterializeEntity<Director>(rawJsonResponse);
+
+            Assert.IsNotNull(materializedObject);
+            Assert.AreEqual(materializedObject.Id, 1);
+            Assert.AreEqual(materializedObject.Name, "Director 1");
+            Assert.AreEqual(materializedObject.DynamicProperties.Count, 0);
+        }
+
+        private void AssertShared(Director materializedObject, Type dynamicPropertyType, string dynamicPropertyName)
```
habbes, line 493: Why the name `AssertShared`?

gathogojr: @habbes It happens that the first set of asserts in each test scenario were exactly the same and I thought that instead of repeating them over and over for each test I'd put them in a shared method that each test calls.

habbes: Ooh. I thought it was "shared" in the context of static members or shared data. Maybe "AssertCommon"? But it's non-issue

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/ODataDynamicPropertyTests.cs
```diff
@@ -0,0 +1,785 @@
...
+            };
+
+            var query = context.CreateQuery<T>(entitySetName);
+
+#if !(NETCOREAPP1_0 || NETCOREAPP2_0)
+            T materializedObject = query.Execute().FirstOrDefault();
+#else
+            var asyncResult = query.BeginExecute(null, null);
+            asyncResult.AsyncWaitHandle.WaitOne();
+            var queryResult = query.EndExecute(asyncResult);
```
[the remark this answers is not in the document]
gathogojr: @habbes Apparently, and I learnt this through sweat and blood, that method is not shared across NETCOREAPP1_0, NETCOREAPP2_0 and now NETCOREAPP3_0. Is why I did it this old-fashioned way. Also why you might find the commit message "Make netcoreapp 1.0 and 2.0 frameworks happy" repeated a couple of times!

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/ODataDynamicPropertyTests.cs
gathogojr: @habbes I struggled with the naming. I could name it `AssertCommon` as I'm adding the test you requested

gathogojr:
@habbes I can add a test for this
https://github.com/OData/odata.net/pull/1712#pullrequestreview-394807105

gathogojr:
@habbes I have committed a test for the scenario you highlighted https://github.com/OData/odata.net/pull/1712#pullrequestreview-394807105

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/ODataDynamicPropertyTests.cs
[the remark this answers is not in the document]
habbes: Interesting, maybe we should document this somewhere to avoid other people paying the same sweat and blood.

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
```diff
@@ -66,9 +66,58 @@ internal IEnumerable<ODataProperty> PopulateProperties(object resource, string s
                 }
             }
 
+            // Process non-structured dynamic properties stored in the open type dictionary
```
paulodero, line 69: What happens to the structured type dynamic properties (i.e complex types) ?

src/Microsoft.OData.Client/Materialization/CollectionValueMaterializationPolicy.cs
```diff
@@ -15,6 +15,7 @@ namespace Microsoft.OData.Client.Materialization
     using Microsoft.OData;
     using Microsoft.OData.Edm;
     using DSClient = Microsoft.OData.Client;
+    using System.Collections.ObjectModel;
```
g2mula, line 18: Nit:This should probably be somewhere on line 12...

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
```diff
@@ -575,7 +576,7 @@ private void MaterializeResolvedEntry(MaterializerEntry entry, bool includeLinks
             // entries. This keeps this code compatible with V1 behavior.
             this.MaterializeDataValues(actualType, entry.Properties, this.MaterializerContext.UndeclaredPropertyBehavior);
 
-            if (entry.NestedResourceInfos != null)
+            if (entry.NestedResourceInfos.Any())
```
g2mula, line 579: If you are no longer testing for null, then there is no reason to test for Any() since the loop inside the if will just not run if there is !.Any()

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
```diff
@@ -66,9 +66,58 @@ internal IEnumerable<ODataProperty> PopulateProperties(object resource, string s
...
 
+        /// <summary>
+        /// Populates list of odata properties with non-structured dynamic properties on an open type
+        /// </summary>
+        /// <param name="resource">Instance of the resource which is getting serialized.</param>
+        /// <param name="serverTypeName">The server type name of the entity whose properties are being populated.</param>
+        /// <param name="odataProperties">List of ODataProperty</param>
+        private void PopulateDynamicProperties(object resource, string serverTypeName, List<ODataProperty> odataProperties)
+        {
+            Debug.Assert(resource != null, "resource !=null");
```
paulodero, line 86: What happens when in production and these values are null?

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
```diff
@@ -687,6 +943,6 @@ private void AddTypeAnnotationNotDeclaredOnServer(string serverTypeName, ClientP
                 // and spatial values always contain their specific type inside the GeoJSON/GML representation.
                 primitiveValue.TypeAnnotation = new ODataTypeAnnotation(property.EdmProperty.Type.FullName());
             }
-        }
+        }
```
g2mula, line 946:
```suggestion
        }
```
Probably wasn't meant for checkin.

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/ODataDynamicPropertyTests.cs
```diff
@@ -0,0 +1,868 @@
...
+        private string producerEntityTypeName;
+        private string producerEntitySetName;
+        private string actorEntityTypeName;
+        private string actorEntitySetName;
+        private string awardEntityTypeName;
+        private string awardEntitySetName;
+
+        private ClientEdmModel clientEdmModel;
+        private DataServiceContext context;
+        private IDictionary<string, string> typeNameToEntitySetMapping;
```
g2mula, line 46: Most of these seem to be set in the constructor, maybe need to have them as readonly...?

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/ODataDynamicPropertyTests.cs
```diff
@@ -0,0 +1,868 @@
...
+            entityDescriptor.Entity = entityObject;
+            var requestMessageArgs = new BuildingRequestEventArgs("POST", new Uri("http://tempuri.org"), headers, entityDescriptor, HttpStack.Auto);
+            var odataRequestMessageWrapper = ODataRequestMessageWrapper.CreateRequestMessageWrapper(requestMessageArgs, requestInfo);
+            var linkDescriptors = new LinkDescriptor[0];
+
+            serializer.WriteEntry(entityDescriptor, linkDescriptors, odataRequestMessageWrapper);
+
+            var stream = (MemoryStream)odataRequestMessageWrapper.CachedRequestStream.Stream;
+            var streamReader = new StreamReader(stream);
+            var body = streamReader.ReadToEnd();
```
g2mula, line 194: The coding style in this project prefers ```Type```s over ```var```s

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
gathogojr: @g2mula Good catch... I should change this to `?.Any()`

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
```diff
@@ -611,6 +802,33 @@ private bool TryConvertPropertyToResourceOrResourceSet(ClientPropertyAnnotation
...
+        /// <summary>
+        /// Tries to convert the given value into an instance of <see cref="ODataItemWrapper"/>.
+        /// </summary>
+        /// <param name="propertyName">Name of the dynamic property.</param>
+        /// <param name="propertyValue">The property value to convert.</param>
+        /// <param name="serverTypeName">The server type name of the entity whose properties are being populated.</param>
+        /// <param name="visitedComplexTypeObjects">Set of instances of complex types encountered in the hierarchy. Used to detect cycles.</param>
+        /// <param name="odataItem">The odata resource or resource set if one was created.</param>
+        /// <returns>Whether or not the value was converted.</returns>
+        private bool TryConvertDynamicPropertyToResourceOrResourceSet(string propertyName, object propertyValue, string serverTypeName, HashSet<object> visitedComplexTypeObjects, out ODataItemWrapper odataItem)
```
paulodero, line 814: This function will always return true. When does it get to return false?

paulodero, line 814: This function will always return true. When does it get to return false? Probably the Debug.Assert(...) statements should be modified to return false in case the objects are null.

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
```diff
@@ -611,6 +802,33 @@ private bool TryConvertPropertyToResourceOrResourceSet(ClientPropertyAnnotation
...
+        {
+            Debug.Assert(!string.IsNullOrEmpty(propertyName), "!string.IsNullOrEmpty(propertyName)");
+            Debug.Assert(propertyValue != null, "propertyValue != null");
+            Debug.Assert(serverTypeName != null, "serverTypeName != null");
```
paulodero, line 818: Why do we need to check this only in debug mode?

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
```diff
@@ -664,6 +900,26 @@ private ODataResourceSetWrapper CreateODataComplexCollectionPropertyResourceSet(
...
+        /// <param name="propertyName">Name of the collection dynamic property.</param>
+        /// <param name="propertyValue">Collection instance.</param>
+        /// <param name="serverTypeName">The server type name of the entity whose properties are being populated.</param>
+        /// <param name="visitedComplexTypeObjects">List of instances of complex types encountered in the hierarchy. Used to detect cycles.</param>
+        /// <returns>An instance of ODataResourceSetWrapper representing the value of the property.</returns>
+        private ODataResourceSetWrapper CreateODataComplexCollectionDynamicPropertyResourceSet(string propertyName, object propertyValue, string serverTypeName, HashSet<object> visitedComplexTypeObjects)
+        {
+            Debug.Assert(!string.IsNullOrEmpty(propertyName), "!string.IsNullOrEmpty(propertyName)");
+            Debug.Assert(propertyValue != null, "propertyValue != null");
+            Debug.Assert(propertyValue is ICollection, "propertyValue must be a collection");
```
paulodero, line 915: Same case for these. Could we check this in release mode as well?

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -687,6 +698,72 @@ internal static string GetEnumValuesString(string enumString, Type enumType)
...
+            Debug.Assert(instance != null, "instance !=null");
+            Debug.Assert(model != null, "model !=null");
+
+            Type clientType = instance.GetType();
+            ClientTypeAnnotation clientTypeAnnotation = model.GetClientTypeAnnotation(clientType);
+
+            Debug.Assert(clientTypeAnnotation != null, "clientTypeAnnotation != null");
+
+            // TODO: Populate dynamic properties only if this is an open (entity or complex) type
+            return clientTypeAnnotation.EdmType.IsOpen();
```
paulodero, line 718: Correct me if I'm wrong. I thought we use //TODO to track tasks in visual studio that is pending.

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -687,6 +698,72 @@ internal static string GetEnumValuesString(string enumString, Type enumType)
...
+        /// <returns>true if expected dictionary is found</returns>
+        internal static bool TryGetDynamicPropertiesDictionary(object instance, out IDictionary<string, object> dynamicPropertiesDictionary)
+        {
+            Debug.Assert(instance != null, "instance != null");
+
+            dynamicPropertiesDictionary = default(IDictionary<string, object>);
+
+            PropertyInfo propertyInfo = instance.GetType().GetPublicProperties(true /* instanceOnly */).Where(p =>
+                                p.GetCustomAttributes(typeof(ContainerPropertyAttribute), true).Any() &&
+                                typeof(IDictionary<string, object>).IsAssignableFrom(p.PropertyType)).FirstOrDefault();
```
paulodero, line 736: What happens if indeed the object instance has properties of type  IDictionary<string,object> which are not intended to be dynamic properties? In this case it gets .FirstOrDefault(). what happens in such a scenario?

gathogojr: @paulodero Note that beyond looking for a property of type IDictionary<string, object> or that implements that interface, we also check that its decorated with an attribute `Microsoft.OData.Client.ContainerProperty`. Adding that attribute is signaling that the dictionary property is the dynamic properties container.

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
gathogojr: @paulodero I might have forgotten to purge some TODOs I had included during development. This is one of them and I'll remove it. I had added it there before implementing that check

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
gathogojr: @paulodero It's common practice to apply asserts in debug mode and not release mode for performance reasons - though in Release you'd typically use `Trace.Assert()`. Assertions will help you verify your assumptions during development.  An assertion in Release mode can't save you. Your program would crash either way, same way it'll crash if without the assert.
`Debug.Assert(customer != null)` will yield the same outcome as try to access a property from the null reference - `customer.Name` in Release mode. In Debug you can do something about it

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
gathogojr: @paulodero Responded to this [here](https://github.com/OData/odata.net/pull/1712#discussion_r412895515)

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/ODataDynamicPropertyTests.cs
gathogojr: @g2mula I'll be the first to admit that I prefer `Type`s over `var`s in a codebase with multiple contributors. I considered this but I saw that this particular project had a mix of both `var`s and `Type`s, e.g. [this](https://github.com/OData/odata.net/blob/master/test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/ClientSerializerTests.cs) file and [this](https://github.com/OData/odata.net/blob/master/test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DateTimeClientSupportTest.cs) other one in the same project. My observation is that code styles are not strictly being applied to test projects. I'd say it was more of a coin flip that a deliberate choice to use `var`s here

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
gathogojr: Good observation

gathogojr: Good observation

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
gathogojr: @paulodero Responded to this [here](https://github.com/OData/odata.net/pull/1712#discussion_r412895515)

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/ODataDynamicPropertyTests.cs
gathogojr: @g2mula They're not in a constructor but rather in a method decorated with `TestInitialize` so C# will not allow them to be `readonly`

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
gathogojr: @paulodero In the `WriteEntry` method of `Serializer` class [here](https://github.com/OData/odata.net/blob/a398670872ef5694cafe0f5630537b47b911b66e/src/Microsoft.OData.Client/Serialization/Serializer.cs#L309), you'll notice a call to `PopulateProperties` and another call to `WriteNestedComplexProperties`. `PopulateProperties` only deals with non-structured types while `WriteNestedComplexProperties` deals with structured types (specifically complex types). Dynamic complex type properties are handled in a similar way as declared complex type properties. The call is made [here](https://github.com/OData/odata.net/pull/1712/files#diff-555a2001db6de39abd06516406c5fdc3R212)

src/Microsoft.OData.Client/Materialization/CollectionValueMaterializationPolicy.cs
gathogojr: Fixed

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
gathogojr: Fixed

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
gathogojr: Fixed

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
gathogojr: Fixed

gathogojr: Fixed

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
paulodero: I get it. Thanks.

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
paulodero: Now that we know the object can be null, can the function handle this to ensure the program does not crash?

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
paulodero: I get it John. Thanks.

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
```diff
@@ -611,6 +801,31 @@ private bool TryConvertPropertyToResourceOrResourceSet(ClientPropertyAnnotation
...
+            Debug.Assert(propertyValue != null, "propertyValue != null");
+            Debug.Assert(serverTypeName != null, "serverTypeName != null");
+
+            if (propertyValue is ICollection)
+            {
+                odataItem = this.CreateODataComplexCollectionDynamicPropertyResourceSet(propertyName, propertyValue, serverTypeName, visitedComplexTypeObjects);
+            }
+            else
+            {
+                odataItem = CreateODataComplexDynamicPropertyResouce(propertyName, propertyValue, visitedComplexTypeObjects);
```
gathogojr, line 825: Use this...

src/Microsoft.OData.Client/Serialization/ODataPropertyConverter.cs
paulodero: Got it.

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
```diff
@@ -675,5 +684,84 @@ private void MaterializeResolvedEntry(MaterializerEntry entry, bool includeLinks
...
+
+            MaterializerNavigationLink linkState = MaterializerNavigationLink.GetLink(link);
+            if (linkState == null || (linkState.Entry == null && linkState.Feed == null))
+            {
+                return;
+            }
+
+            // NOTE: ODL (and OData WebApi) support navigational property on complex types
+            // That support has not yet been implemented in OData client
+
```
ElizabethOkerio, line 713: There is an open PR for adding that support.

PR #1714, opened 2020-03-23
ODataMessageWriter can't dispose the stream if there's no write method called

xuzhg:
### Issues

### Description

Exchange reports an issue that:
**Looks like there are certain conditions under which ODataMessageWriter is not disposing the stream returned by IODataResponseMessage.GetStream. Are there any known bugs in this area?**

After investigation, we found this edge case. This is an edge case only happens when customer creates the message writer without writing anything. However, if there's some control to prevent the
message writer from writing anything, the outside stream still needs to
dispose. This PR is to resolve this edge problem.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataMessageWriter.cs
+++ b/src/Microsoft.OData.Core/ODataMessageWriter.cs
@@ -1212,6 +1212,13 @@ namespace Microsoft.OData
                     {
                         this.outputContext.Dispose();
                     }
+                    else if (!this.writeMethodCalled)
+                    {
+                        if (this.settings.EnableMessageStreamDisposal)
+                        {
+                            this.message.GetStream().Dispose();
+                        }
+                    }
                 }
                 finally
                 {
--- a/src/Microsoft.OData.Core/ODataMessageWriter.cs
+++ b/src/Microsoft.OData.Core/ODataMessageWriter.cs
@@ -1212,11 +1212,12 @@ namespace Microsoft.OData
                     {
                         this.outputContext.Dispose();
                     }
-                    else if (!this.writeMethodCalled)
+                    else if (this.settings.EnableMessageStreamDisposal)
                     {
-                        if (this.settings.EnableMessageStreamDisposal)
+                        Stream stream = this.message.GetStream();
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/ODataMessageWriter.cs
```diff
@@ -1212,6 +1212,13 @@ private void Dispose(bool disposing)
                     {
                         this.outputContext.Dispose();
                     }
+                    else if (!this.writeMethodCalled)
```
mikepizzo: is the !this.writeMethodCalled check needed? Is there a scenario where outputContext == null and this.writeMethodCalled == true?

src/Microsoft.OData.Core/ODataMessageWriter.cs
```diff
@@ -1212,6 +1212,13 @@ private void Dispose(bool disposing)
                     {
                         this.outputContext.Dispose();
                     }
+                    else if (!this.writeMethodCalled)
+                    {
+                        if (this.settings.EnableMessageStreamDisposal)
+                        {
+                            this.message.GetStream().Dispose();
+                        }
```
mikepizzo: Do we need to check for GetStream() returning null?

src/Microsoft.OData.Core/ODataMessageWriter.cs
```diff
@@ -1212,6 +1212,14 @@ private void Dispose(bool disposing)
                     {
                         this.outputContext.Dispose();
                     }
+                    else if (this.settings.EnableMessageStreamDisposal)
+                    {
+                        Stream stream = this.message.GetStream();
```
KanishManuja-MS: Do we need to always call GetStream? Instead, should we keep a reference to it in the ODataMessage and dispose of that if it is not null and call GetStream as part of else. 
 else if (this.settings.EnableMessageStreamDisposal && this.message.InternalStream != null)
                    {
                        this.message.InternalStream.dispose();
                    }
else if (EnableMessageStreamDisposal)
{
GetStream()
.Dispose();
}

Issue #1726, opened 2020-03-31
Exception when serializing open type with undeclared property


gathogojr:
A bug exists in OData client library where open types `IsOpen` property get initialized `false` specifically **in a Post scenario where no query operation to the service has been done prior**. This later causes an exception to be thrown during serialization since the owning type isn't recognized as open and the undeclared property won't also be interpreted as an open property. This bug had been dissected by a user back in 2017 but the [issue](https://github.com/OData/odata.net/issues/272#issuecomment-284971690) had been closed with a **not-fixed** label.

### Assemblies affected
Microsoft.OData.Client 7.6.4

### Reproduce steps

1. Create an OData V4 service with an open type and a controller

**Product** class
```
public class Product
{
    [Key]
    public int Id { get; set; }
    public string Name { get; set; }
    public IDictionary<string, object> DynamicProperties { get; set; }
}
```

**ProductsController**
```
[ApiController]
[Route("[controller]")]
public class ProductsController : ControllerBase
{
    [EnableQuery]
    public IQueryable<Product> Get()
    {
        return new[] { new Product { Id = 1, Name = "Product 1" } }.AsQueryable();
    }

    [EnableQuery]
    public SingleResult<Product> Get([FromODataUri]int key)
    {
        return SingleResult.Create(new[] { new Product { Id = key, Name = "Product " + key } }.AsQueryable());
    }

    public async Task<IActionResult> Post([FromBody]Product product)
    {
        await Task.Run(() =>
        {
            // ...            
        });

        return Created(new Uri($"{Request.Scheme}://{Request.Host}{Request.Path}/{product.Id}", UriKind.Absolute), product);
    }
}
```

2. Create an OData client
3. Add a Connected Service to the client using OData Connected Service extension, making use of service endpoint for the OData service created in step 1
4. Add code to perform a post operation, making sure NOT to include any code that does any query operation to the service. Also add an undeclared property to the post payload - one way to achieve this is by riding on the `OnEntryStarting` hook supported by client, as shown below:

```
static void Main(string[] args)
{
    var serviceUri = new Uri("<ServiceEndpoint>");
    var context = new Default.Container(serviceUri);

    var product = new Product
    {
        Id = 99,
        Name = "Product 99"
    };

    context.Configurations.RequestPipeline.OnEntryStarting(args =>
    {
        if (product.GetType().IsAssignableFrom(args.Entity.GetType()))
        {
[...]

PR #161, opened 2015-04-22
Undeclared properties should be allowed on Open Types

AdamCaviness:
Added support to ClientEdmModel to consult IEdmEntityType's IsOpen
property when instantiating EdmEntityTypeWithDelayLoadedProperties and
EdmComplexTypeWithDelayLoadedProperties.  ClientEdmModel was always
setting isOpen to false for these instances. This fixes an issue where
sending arbitrary/undeclared ODataProperty instances up would throw in
WriterValidationUtil.ValidatePropertyDefined().  This addresses
https://github.com/OData/odata.net/issues/143

PR #1727, opened 2020-03-31
Fix bug where open types are not identified as such during serialization

gathogojr:
### Issues

*This pull request fixes issue #1726.*

### Description

This PR fixes a bug  in OData client library where open types `IsOpen` property get initialized to `false` specifically in a Post scenario where no query operation to the service has been done prior. This later causes an exception to be thrown during serialization since the owning type isn't recognized as open and the undeclared property won't also be interpreted as an open property.

The fix involved moving the code for populating the `EdmStructuredSchemaElements` property of `ClientEdmModel` to a position right after the service model has been loaded

Bug was inadvertently introduced in this [PR](https://github.com/OData/odata.net/pull/161) and only occurs if query operation is done prior to a post operation.
Another user dissected and raised the issue here

- [x] *Test cases added*

### DIFF
--- a/test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientOpenTypeUpdateTests.cs
+++ b/test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientOpenTypeUpdateTests.cs
@@ -44,6 +44,20 @@ namespace Microsoft.Test.OData.Tests.Client
             // No more check, this case is to make sure that client doesn't throw exception.
         }
 
+        [TestMethod]
+        public void AddOpenTypeWithUndeclaredProperties()
+        {
+            SetContextWrapper();
+            
+            var row = Row.CreateRow(Guid.NewGuid());
+
+            contextWrapper.Configurations.RequestPipeline.OnEntryStarting(ea => EntryStarting(ea));
+
+            contextWrapper.AddObject("Row", row);
+            contextWrapper.SaveChanges();
+            // All clear if no exception is thrown
+        }
+
         private void EntryStarting(WritingEntryArgs ea)
         {
             var odataProps = ea.Entry.Properties as List<ODataProperty>;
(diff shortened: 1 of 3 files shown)


test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientOpenTypeUpdateTests.cs
```diff
@@ -44,6 +44,20 @@ public void UpdateOpenTypeWithUndeclaredProperties()
...
+        public void AddOpenTypeWithUndeclaredProperties()
+        {
+            SetContextWrapper();
+            
+            var row = Row.CreateRow(Guid.NewGuid());
+
+            contextWrapper.Configurations.RequestPipeline.OnEntryStarting(ea => EntryStarting(ea));
+
+            contextWrapper.AddObject("Row", row);
+            contextWrapper.SaveChanges();
```
xuzhg, line 57: cann't use Assert.NotThrows<...>(....) pattern?

gathogojr: Hey @xuzhg I assume you're talking something like [this](https://github.com/OData/odata.net/blob/962aa97d616b34243bbddc9c7bf6cead4a8fe275/test/EndToEndTests/Tests/Client/Build.Desktop/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs#L483). Top of my head, I don't think **Microsoft.VisualStudio.QualityTools.UnitTestFramework** provides `Assert.NotThrows`. Be that as it may, asserting that an exception is not thrown is frowned upon. They actually got rid of it in xUnit - see [this thread](https://github.com/xunit/xunit/issues/188). Any uncaught exception will fail the test hence I'd understand the argument against it. I can however improvise if you feel strongly about it.

xuzhg: I'd like to get rid of the current [TestMethod] and use "xunit.net" for all. Spatial, Edm, Core are moved to xunit already.
So, @odero, can you create a task to switch to xunit for the client test cases? 

---
In reply to: [401829651](https://github.com/OData/odata.net/pull/1727#discussion_r401829651) [](ancestors = 401829651)

Issue #1333, opened 2018-11-22
Inconsistent behavior on 'DataServiceQuery.AddQueryOption' related to '$' character


julealgon:
The `AddQueryOption` method on `DataServiceQuery` appears to perform specific validation depending on the name supplied in the call by looking for the `$` character in the beginning of the option name. This check is done in an inconsistent manner, conditionally causing an exception for some entries, but not for others.

This is incredibly misleading and unnecessary as far as I can tell.

### Assemblies affected

Microsoft.OData.Client 7.5.2

### Reproduce steps

Create any query using `DataServiceContext`, and then call `AddQueryOptions` with different values:
1. `AddQueryOptions("$select", "field")`
2. `AddQueryOptions("$expand", "field")`
3. `AddQueryOptions("$orderby", "field")`
4. `AddQueryOptions("$skip", 10)`
5. `AddQueryOptions("$skiptoken", "something")`

Then inspect the `RequestUri` property on the `DataServiceContext`.

### Expected result

Either of:
1. All calls should work correctly and append the requested query option to the Expression Url.
2. An exception should be thrown with a specific message telling why a given option is not supported.

### Actual result

Call 5 above causes an exception to be thrown when inspecting the `RequestUri` property on the `DataServiceQuery` object with an overly broad message:
> System.NotSupportedException: Can't add query option '$skiptoken' **because it begins with reserved character '$'**.
at Microsoft.OData.Client.ResourceBinder.ValidationRules.RequireLegalCustomQueryOption(Expression e, ResourceExpression target)
   at Microsoft.OData.Client.ResourceBinder.AnalyzeAddCustomQueryOption(MethodCallExpression mce)
   at Microsoft.OData.Client.ResourceBinder.VisitMethodCall(MethodCallExpression mce)
   at Microsoft.OData.Client.ALinqExpressionVisitor.Visit(Expression exp)
   at Microsoft.OData.Client.ResourceBinder.Bind(Expression e, DataServiceContext context)
   at Microsoft.OData.Client.DataServiceQueryProvider.Translate(Expression e)
   at Microsoft.OData.Client.DataServiceQuery`1.get_RequestUri()

Error message itself is completely misleading since I just provided several samples of options beginning with the `$` character that worked fine, going against what is stated in the bolded section. If "anything that began with ´$` was illegal", then for sure I'd get the very same error on the other calls?

### Additional detail

I actually found this out when creating an integration test for one of my OData routes:
```csharp
[DataTestMethod]
[...]

PR #1729, opened 2020-03-31
Update error message when adding unsupported query option

habbes:
### Issues

*This pull request fixes issue #1333 

### Description

Currently, when add a query option like `$custom`, you get the error:
```
 Can't add query option '$custom' because it begins with reserved character '$'.
```
to let you know that you can't add custom query options starting with `$`, since this is reserved for standard OData query options. However this error is also displayed for standard query options that are not supported by the client or not yet implemented on the client e.g.:
```
 Can't add query option '$apply' because it begins with reserved character '$'.
```
In such cases, the error is misleading.

This PR changes the error message to:
```
The query option '$apply' is not supported.
```
also apply for user-defined/non-standard query options:
```
The query option '$custom` is not supported.
```
I thought of giving the user-defined case a different error message to distinguish it from known standard query options which are not supported, but that would mean updating these checks every time there's a new query option in the standard that's not yet implemented on the client.

- [x] *Test cases added*

### Additional work necessary

The `LinqAddQueryOption` tests in `LinqTests.cs` were disabled (the `[TestMethod]` attribute was commented out), not sure why this is the case. I enabled them and added a few more tests there.


test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
```diff
@@ -7899,7 +7899,7 @@ internal string ReturnString(string value)
             return value;
         }
 
-        // [TestMethod]
+        [TestMethod]
```
gathogojr, line 7902: @habbes I think this test could be simplified like this:
```
foreach (var option in new[] {"$custom", "$apply", "$skiptoken", "$delta"})
{
    var query = ((DataServiceQuery<League>)from l in context.CreateQuery<League>("Leagues") select l)
                                                .AddQueryOption(option, 1);

    try
    {
        query.GetEnumerator();
    }
    catch (Exception e)
    {
        if (e.Message != string.Format(System.Globalization.CultureInfo.InvariantCulture, "The query option {} is not supported.", option)
        {
            throw new Exception("Test Failed");
        }
    }
}
```

src/Microsoft.OData.Client/Build.Portable/Microsoft.OData.Client.Portable.cs
```diff
@@ -21,7 +21,7 @@ namespace Microsoft.OData.Client {
     /// <summary>
     ///    AutoGenerated resource class. Usage:
     ///
-    ///        string s = TextRes.GetString(TextRes.MyIdentifier);
+    ///        string s = TextRes.GetString(TextRes.MyIdenfitier);
```
odero, line 24:
```suggestion
    ///        string s = TextRes.GetString(TextRes.MyIdentifier);

```

src/Microsoft.OData.Client/Microsoft.OData.Client.Common.txt
```diff
@@ -176,6 +176,7 @@ ALinq_CantAddQueryOption=Custom query option not allowed.
 ALinq_CantAddDuplicateQueryOption=Can't add duplicate query option '{0}'.
 ALinq_CantAddAstoriaQueryOption=Can't add query option '{0}' because it would conflict with the query options from the translated Linq expression.
 ALinq_CantAddQueryOptionStartingWithDollarSign=Can't add query option '{0}' because it begins with reserved character '$'.
+ALinq_QueryOptionNotSupported=The query option '{0}' is not supported.
```
odero, line 179: Based on the originally filed issue, this error message still seems just as non-descriptive as the previous one. As the user who filed the original issue this new message still doesn't let me know that the reason I cant use the options is because they are controlled by the server and should not be set by the client.
Maybe something along the lines of: "The query option '{0}' is not supported or is controlled by the odata service"

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
```diff
@@ -8035,24 +8035,73 @@ orderby t.TeamID
...
             catch (Exception e)
             {
-                if (e.Message != "Can't add query option '$custom' because it begins with reserved character '$'.")
+                if (e.Message != "The query option '$custom' is not supported.")
                 {
                     throw new Exception("Test Failed");
                 }
             }
 
+            var queryableApplyOption = ((DataServiceQuery<League>)from l in context.CreateQuery<League>("Leagues")
```
odero, line 8044: Shouldnt $apply be supported?

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
```diff
@@ -8035,24 +8035,73 @@ orderby t.TeamID
...
+                                                        select l)
+                                                        .AddQueryOption("$apply", 1);
+
+            Trace.WriteLine("known, but unsupported query  options");
+
+            try
```
odero, line 8050: It should be possible to convert these `try/catch`es to be `Assert.Throws` right?

src/Microsoft.OData.Client/Build.Portable/Microsoft.OData.Client.Portable.cs
```diff
@@ -175,6 +175,7 @@ internal sealed class TextRes {
         internal const string ALinq_CantAddDuplicateQueryOption = "ALinq_CantAddDuplicateQueryOption";
         internal const string ALinq_CantAddAstoriaQueryOption = "ALinq_CantAddAstoriaQueryOption";
         internal const string ALinq_CantAddQueryOptionStartingWithDollarSign = "ALinq_CantAddQueryOptionStartingWithDollarSign";
```
odero, line 177: After you replaced this in the ResourceBinder, is it still used anywhere else other than in the test?

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
habbes: When you allow `$apply` to be added, a `PlatformNotSupported` exception is thrown with the error message "Secure binary serialization is not supported on this platform". I have not investigated this exception in depth.

src/Microsoft.OData.Client/Build.Portable/Microsoft.OData.Client.Portable.cs
habbes: Didn't see it use anywhere else, I've removed it.

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
habbes: I still tested $custom separately in order to make a distinction between unknown/custom and official query options.

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
habbes: I don't think the assertion library used in these particular tests provides that method.

src/Microsoft.OData.Client/Build.Portable/Microsoft.OData.Client.Portable.cs
odero: Reminder to fix this before you merge

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/LinqTests.cs
KenitoInc: I don't think $apply is currently supported in Client

Issue #1581, opened 2019-10-28
Unable to create a new entity with a nested complex type collection


Sayan751:
I am unable to POST a request for a new entity object creation with a collection of derived complex types, nested in a navigational property, from the generated C# OData client. 

### Assemblies affected

7.6.0

### Reproduce steps

```csharp
/**approx EDM**/
abstract class BaseThing { }

class DerivedThing1 : BaseThing { }

class DerivedThing2 : BaseThing { }

class Configuration {
  public IEnumerable<BaseThing> Things {get; set;}
}
class Entity {
  public Configuration Configuration {get; set;}
}

/**OData Client**/
var entity = new Entity {
  Configuration = new Configuration {
    Things =  new ObservableCollection<BaseThing>(new List<BaseThing>{ new DerivedThing1()}))
  }
};

context.AddToEntities(entity);
context.SaveChanges();
```

### Expected result

Expected the entity to be saved.

### Actual result

> Unexpected error 500 - The type 'Odata.DerivedThing1' of a resource in an expanded link is not compatible with the element type 'Odata.BaseThing' of the expanded link. Entries in an expanded link must have entity
types that are assignable to the element type of the expanded link..

### Additional detail

- The service is working perfectly when POSTed from a REST client, such as Insomnia, or a JS Client. This is happening only with the generated C# client.
- The error is thrown from here: https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/WriterValidationUtils.cs#L159-L173

Sayan751:
This works when I do following.

```csharp
var configuration = new Configuration();
configuration.Things.AddRange(new List<BaseThing>{ new DerivedThing1() });
var entity = new Entity {  Configuration = configuration };

context.AddToEntities(entity);
context.SaveChanges();
```

ElizabethOkerio:
Hey @Sayan751, can you please share a link to a repo of your implementation that we can look at and advise accordingly because we have been able to create a new entity with a nested complex type collection.

Sayan751:
@ElizabethOkerio It is bit difficult to provide a reproduction repo. However, I will try to create one, and will get back to you. Sorry for the late reply.

ElizabethOkerio:
Hey @Sayan751, Thanks for getting back. Will really appreciate it if you create one.

Sayan751:
[...]

PR #1731, opened 2020-04-03
Support for abstract complex types with no properties

ElizabethOkerio:
### Issues

*This pull request fixes issue #1581.*

### Description
Since Web API OData V5.5-beta, it is allowed to:

1. define abstract type (entity & complex) without any properties.

But in this case, when you define an abstract complex type without any properties  an exception is thrown. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/ClientEdmModel.cs
+++ b/src/Microsoft.OData.Client/ClientEdmModel.cs
@@ -323,12 +323,12 @@ namespace Microsoft.OData.Client
             }
             else
             {
-                // type is a complex type. Return all types on the hierarchy where there are properties defined.
+                // type is a complex type. Return all types on the hierarchy with or without properties defined.
                 do
                 {
                     hierarchy.Insert(0, type);
                 }
-                while ((type = c.PlatformHelper.GetBaseType(type)) != null && ClientTypeUtil.GetPropertiesOnType(type, false /*declaredOnly*/).Any());
+                while ((type = c.PlatformHelper.GetBaseType(type)) != null);
             }
 
             return hierarchy.ToArray();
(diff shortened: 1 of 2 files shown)


xuzhg:
@ElizabethOkerio It's better to have a test case to cover the changes.

src/Microsoft.OData.Client/ClientEdmModel.cs
```diff
@@ -328,7 +328,7 @@ private static Type[] GetTypeHierarchy(Type type, out PropertyInfo[] keyProperti
                 {
                     hierarchy.Insert(0, type);
                 }
-                while ((type = c.PlatformHelper.GetBaseType(type)) != null && ClientTypeUtil.GetPropertiesOnType(type, false /*declaredOnly*/).Any());
+                while ((type = c.PlatformHelper.GetBaseType(type)) != null );
```
odero, line 331: You'll want to update the comment on L326 as well otherwise it might be very confusing for someone reading this file later on.
Also add some tests for this

test/FunctionalTests/Microsoft.OData.Client.Tests/ClientEdmModelTests.cs
```diff
@@ -0,0 +1,67 @@
...
+    {
+        /// <summary>
+        /// A test to show that an abstract complex type with no properties defined is 
+        /// assignable from a concrete complex type that extends it.
+        /// Product is the abstract complex type with no properties defined and
+        /// ProductA is the concrete complex type that extends the abstract 
+        /// complex type(Product) with no properties defined. 
+        /// </summary>
+        [Fact]
+        public void BaseTypeOfProductAIsProduct()
```
odero, line 23: I'd suggest 2 tests to show that the empty abstract works for ComplextTypes but fails for EntityTypes

ElizabethOkerio: it should not fail for entity types. You can have an empty abstract type for entity types.

odero: Oh right. My mistake

Issue #1692, opened 2020-03-05
GetReadStreamAsync not available with NoTracking - No way to get Streams with NoTracking


norcino:
It is not possible to consume streams when connecting to a read only API with MergeOption set to NoTracking.

### Assemblies affected

Microsoft.OData.Client 7.6.3

### Reproduce steps

1) Create a Container (service reference)
2) Set the MergeOptions to NoTracking
3) Load an entity which supports streams

at this stage if you try to get the stream for the given entity it will fail, even if you try to attach it

```
var stream = (await apiReference.GetReadStreamAsync(qqml, new DataServiceRequestArgs())).Stream;
var sr = new StreamReader(stream).ReadToEnd();
Console.WriteLine(sr);
```

### Expected result

1) I would expect to be able to get a stream by Id of the entity which is not possible
2) I would like to disable tracking for memory and performance benefit, given that I don't need it for a read only api. Along with it I would still like to be able to consume streams.

### Actual result

1) Is not possible to get a stream straight away (not documented at least)
2) When getting a stream for an entity not tracked, even if you try to manually track it, it throws an exception as shown below.

When the entity is not attached (NoTracking):

> System.InvalidOperationException: 'The context is not currently tracking the entity.'

When the entity is manually attached:

> System.ArgumentException: 'This operation requires that the specified entity be a Media Link Entry and that the ReadStreamUri be available. However, the specified entity either is not a Media Link Entry or does not have a valid ReadStreamUri value. If the entity is a Media Link Entry, re-query the data service for this entity to obtain a valid ReadStreamUri value. '

### Notes
I think this is important because for me would not be worth enable tracking for all requests only to support a minimal percentage of requests targeting streams.

marabooy:
@norcino  This is a limitation in the current code. We are happy to accept any contributions on the above as well.

marabooy:
@norcino Would you be fine with a solution that tracks the stream entities only. You can later remove the tracked entities once you have consumed the stream.

norcino:
@marabooy Ideally I would keep things as simple as possible. Therefore I rather allow the streaming without the need to track the entity.
I cannot see how tracking would help in a read-only API while accessing streams.
Maybe the underlying code makes easier to still require the entity to be tracked, I am not sure about that, I can only speak as a user.

PR #1738, opened 2020-04-08
Support for Media Link entities with no tracking

marabooy:
### Issues

*This pull request fixes issue #1692*

### Description
This adds a stream descriptor to the base entity to allow entities that have streams to continue working even when the Context is in non tracking mode. In non tracking mode the stream properties are tracked with the object without any breaking changes. 

Allowing users to perform the following actions in no tracking mode for media link entries and any other stream properties.
```
var stream = (await context.GetReadStreamAsync(object, new DataServiceRequestArgs())).Stream;
var sr = new StreamReader(stream).ReadToEnd();
Console.WriteLine(sr);
```

- [ ] *Test cases added*

### Additional work necessary
None


test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextTrackOnlyMleNonAsyncApiTests.cs
```diff
@@ -0,0 +1,373 @@
+﻿
```
xuzhg, line 1: copyright?

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextTrackOnlyMleNonAsyncApiTests.cs
```diff
@@ -0,0 +1,373 @@
...
+                    });
+            };
+
+
+            var streamResponse = dataServiceContext.GetReadStream(entity, new DataServiceRequestArgs());
+            return streamResponse.Stream;
+
+        }
+
+        [Theory]
```
xuzhg, line 212: only one scenario, why don't use the [Fact] directly?

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceTrackingMLEOnly.cs
```diff
@@ -0,0 +1,366 @@
+﻿
```
xuzhg, line 1: copy right?

src/Microsoft.OData.Client/BaseEntityType.cs
```diff
@@ -13,5 +13,9 @@ public class BaseEntityType
     {
         /// <summary>DataServiceContext for query provider</summary>
         protected internal DataServiceContext Context { get; set; }
+        /// <summary>
+        /// Stream descriptor for the entity  containing entity stream links
+        /// </summary>
+        protected internal StreamDescriptor StreamDescriptor { get; set; }
```
xuzhg, line 19: I didn't find the place to set this Descriptor?

src/Microsoft.OData.Client/MaterializeFromAtom.cs
```diff
@@ -4,6 +4,8 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using Microsoft.OData.Edm;
+
```
xuzhg, line 8: So, Client still has "Atom" related codes, Is there any plan to remove them?

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextTrackOnlyMleNonAsyncApiTests.cs
```diff
@@ -0,0 +1,373 @@
...
+
+                        });
+            }
+
+        }
+
+
+        private void SetupContextWithRequestPipeline(DataServiceContext[] dataServiceContexts, bool forMle)
+        {
+
```
xuzhg, line 290: remove all unused white space line

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceTrackingMLEOnly.cs
```diff
@@ -0,0 +1,366 @@
+﻿
+namespace Microsoft.OData.Client.Tests.Tracking
+{
```
xuzhg, line 3: Shall we add this file into the csproj compiler list?

src/Microsoft.OData.Client/BaseEntityType.cs
marabooy: It is set inside the materializer, when certain conditions are met, in this case an entity that has a stream and  the context is in NoTracking mode. In normal tracking mode the descriptor is available in the entity tracker.

src/Microsoft.OData.Client/MaterializeFromAtom.cs
marabooy: This specific class seems to have evolved to handled more than just Atom and I find the name to be a bit misleading due to its usage. It handles all feeds with the right materializer being selected and provided to provide the right representation.

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceTrackingMLEOnly.cs
marabooy: I believe it is already added.

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextTrackOnlyMleNonAsyncApiTests.cs
```diff
@@ -0,0 +1,335 @@
...
+            Assert.Equal(0, NonTrackingContext.Entities.Count);
+            Assert.Equal(3, DefaultTrackingContext.Entities.Count);
+        }
+
+        [Fact]
+        public void TrackMediaEntitiesShouldBePopulatedWithMediaEntities()
+        {
+            SetupContextWithRequestPipeline(new DataServiceContext[] { NonTrackingContext }, true);
+
+            var mleNoTracking = NonTrackingContext.Documents.ToList();
```
habbes, line 147: What does `mle` mean?

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextTrackOnlyMleNonAsyncApiTests.cs
```diff
@@ -0,0 +1,335 @@
...
+            foreach (var entity in entities)
+            {
+                var doc = entity as BaseEntityType;
+                Assert.Null(doc.StreamDescriptor);
+                Action act1 = () => NonTrackingContext.GetReadStreamUri(entity);
+
+                act1.ShouldThrow<InvalidOperationException>(Strings.Context_EntityInNonTrackedContextLacksMediaLinks);
+                //try and get the content and verify that the content matches the values
+                Action act2 = () => GetTestReadStreamResult(NonTrackingContext, entity);
+                act2.ShouldThrow<InvalidOperationException>().WithMessage(Strings.Context_EntityInNonTrackedContextLacksMediaLinks);
```
habbes, line 183: This comment does not seem to reflect what is happening. You're not comparing results, just checking for an exception.

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextTrackOnlyMleNonAsyncApiTests.cs
```diff
@@ -0,0 +1,335 @@
...
+
+        public string Name { get; set; }
+        public int FileLength { get; set; }
+    }
+    [Key("Id")]
+    internal class User : BaseEntityType
+    {
+        public int Id { get; set; }
+        public string Name { get; set; }
+    }
```
habbes, line 301: Would it be worthwile to add a test using a class that does not inherit from `BaseEntityType` to explicitly test what its expected behaviour is?

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextTrackOnlyMleNonAsyncApiTests.cs
```diff
@@ -0,0 +1,335 @@
...
+                var baseEntity = entity as BaseEntityType;
+                // if we read the wrong link then reply with gibberish
+                var resp = (args.RequestUri == baseEntity?.StreamDescriptor?.SelfLink) ? AttatchmentResponse : "gibberish";
+
+                return new CustomizedRequestMessage(
+                    args,
+                    resp,
+                    new Dictionary<string, string>
+                    {
+                            {"Content-Type", "text/plain"}
```
habbes, line 200: Why not use the same content type as on the stream descriptor, for those that have a content type?

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextTrackOnlyMleNonAsyncApiTests.cs
```diff
@@ -0,0 +1,335 @@
...
+            var mleNoTracking = NonTrackingContext.Documents.ToList();
+
+            Assert.Equal(0, NonTrackingContext.EntityTracker.Entities.Count());
+
+            // verify that the stream links are equal 
+            foreach (var document in mleNoTracking)
+            {
+                var doc = document as BaseEntityType;
+                Assert.NotNull(doc.StreamDescriptor.SelfLink);
+                Assert.NotNull(NonTrackingContext.GetReadStreamUri(document));
```
habbes, line 156: Does this actually verify that the links are equal? Or it only verifies that they are not null?

habbes: On top of that, I think it would be beneficial to add tests that explicitly test whether the links and content type on the descriptor are the same ones that are in the response payload.

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextTrackOnlyMleNonAsyncApiTests.cs
marabooy: Wont cause any issues on the client, the server could in theory send a different content type. But will change it to be consistent to reduce confusion.

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextTrackOnlyMleNonAsyncApiTests.cs
marabooy: media link entry

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -1248,9 +1248,40 @@ public QueryOperationResponse<T> LoadProperty<T>(object entity, string propertyN
         public Uri GetReadStreamUri(object entity)
         {
             Util.CheckArgumentNull(entity, "entity");
-            EntityDescriptor box = this.entityTracker.GetEntityDescriptor(entity);
+            EntityDescriptor box = GetEntityDescriptorForStreamEntity(entity);
```
g2mula, line 1251: I believe the style is to append ```this.``` and looking at the build errors, stylecop would have caught this.

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -1248,9 +1248,40 @@ public QueryOperationResponse<T> LoadProperty<T>(object entity, string propertyN
...
+            EntityDescriptor box;
+            if (MergeOption == MergeOption.NoTracking)
+            {
+                BaseEntityType baseEntity = entity as BaseEntityType;
+                if (baseEntity == null)
+                {
+                    throw Error.InvalidOperation(Strings.Context_EntityMediaLinksNotTrackedInEntity);
+                }
+
+                box = baseEntity?.EntityDescriptor;
```
g2mula, line 1271:
```suggestion
                box = baseEntity.EntityDescriptor;
```
The null check has already happened in line 1266

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -1264,7 +1295,7 @@ public Uri GetReadStreamUri(object entity, string name)
             Util.CheckArgumentNull(entity, "entity");
             Util.CheckArgumentNullAndEmpty(name, "name");
             this.EnsureMinimumProtocolVersionV3();
-            EntityDescriptor entityDescriptor = this.entityTracker.GetEntityDescriptor(entity);
+            EntityDescriptor entityDescriptor = GetEntityDescriptorForStreamEntity(entity);
```
g2mula, line 1298:
```suggestion
            EntityDescriptor entityDescriptor = this.GetEntityDescriptorForStreamEntity(entity);
```
Project style.

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3284,8 +3315,9 @@ private bool EnsureRelatable(object source, string sourceProperty, object target
             Util.CheckArgumentNull(entity, "entity");
             Util.CheckArgumentNull(args, "args");
 
-            EntityDescriptor entityDescriptor = this.entityTracker.GetEntityDescriptor(entity);
+            EntityDescriptor entityDescriptor = GetEntityDescriptorForStreamEntity(entity);
```
g2mula, line 3318: nit: Styleguide is to use ```this.```

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
```diff
@@ -671,6 +673,19 @@ private void MaterializeResolvedEntry(MaterializerEntry entry, bool includeLinks
             if (entity != null)
             {
                 entity.Context = this.EntityTrackingAdapter.Context;
+                // if NoTracking
+                if (!entry.IsTracking)
+                {
+                    int? namedStreamDescriptors = entry.EntityDescriptor.StreamDescriptors?.Count;
+                    IEdmEntityType entityType =
+                        this.EntityTrackingAdapter.Model.FindDeclaredType(entry.Entry.TypeName) as IEdmEntityType;
+                    if ((namedStreamDescriptors.HasValue && namedStreamDescriptors.Value > 0) ||
```
g2mula, line 682:
```suggestion
                    if ((namedStreamDescriptors > 0) ||
```

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
```diff
@@ -671,6 +673,19 @@ private void MaterializeResolvedEntry(MaterializerEntry entry, bool includeLinks
...
             {
                 entity.Context = this.EntityTrackingAdapter.Context;
+                // if NoTracking
+                if (!entry.IsTracking)
+                {
+                    int? namedStreamDescriptors = entry.EntityDescriptor.StreamDescriptors?.Count;
+                    IEdmEntityType entityType =
+                        this.EntityTrackingAdapter.Model.FindDeclaredType(entry.Entry.TypeName) as IEdmEntityType;
+                    if ((namedStreamDescriptors.HasValue && namedStreamDescriptors.Value > 0) ||
+                        ( entityType != null && entityType.HasStream))
```
g2mula, line 683:
```suggestion
                        ( entityType?.HasStream == true))
```

src/Microsoft.OData.Client/Materialization/MaterializerEntry.cs
```diff
@@ -313,6 +313,24 @@ public void UpdateEntityDescriptor()
                         streamInfo.ContentType = streamValue.ContentType;
                     }
                 }
+                // We want this to be populated as well
```
g2mula, line 316:
```suggestion

                // We want this to be populated as well
```

src/Microsoft.OData.Client/MaterializeFromAtom.cs
```diff
@@ -345,6 +346,7 @@ private bool MoveNextInternal()
                     if (result)
                     {
                         BaseEntityType entity = this.materializer.CurrentValue as BaseEntityType;
+
```
g2mula, line 349: nit: probably not meant for checkin?

src/Microsoft.OData.Client/Microsoft.OData.Client.cs
```diff
@@ -21,7 +21,7 @@ namespace Microsoft.OData.Client {
     /// <summary>
     ///    AutoGenerated resource class. Usage:
     ///
-    ///        string s = TextRes.GetString(TextRes.MyIdenfitier);
+    ///        string s = TextRes.GetString(TextRes.MyIdentifier);
```
g2mula, line 24: 👍

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextTrackOnlyMleNonAsyncApiTests.cs
g2mula: ```var mediaLinkEntries...``` would be even better :-)

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceTrackingMLEOnly.cs
```diff
@@ -0,0 +1,413 @@
...
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Linq;
+    using System.Text;
+    using System.Threading.Tasks;
+    using System.Xml;
+    using Xunit;
+
+    public class DataServiceContextNoTrackingStreamsTests
```
g2mula, line 20: q: Does this class name need to match the filename?

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -1248,10 +1248,36 @@ public QueryOperationResponse<T> LoadProperty<T>(object entity, string propertyN
         public Uri GetReadStreamUri(object entity)
         {
             Util.CheckArgumentNull(entity, "entity");
-            EntityDescriptor box = this.entityTracker.GetEntityDescriptor(entity);
+            var box = GetEntityDescriptorForStreamEntity(entity);
             return box.ReadStreamUri;
         }
 
+        private EntityDescriptor GetEntityDescriptorForStreamEntity(object entity)
```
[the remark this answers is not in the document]
marabooy: There is another one that has the same name but updating it may cause some undefined behavior.

src/Microsoft.OData.Client/Materialization/MaterializerEntry.cs
```diff
@@ -314,6 +314,25 @@ public void UpdateEntityDescriptor()
                     }
                 }
 
+                // We need this to be populated as well for entities that may contain this
+                if (this.entry.MediaResource != null)
+                {
+                    if (this.entry.MediaResource.ReadLink != null)
+                    {
+                        this.EntityDescriptor.ReadStreamUri = this.entry.MediaResource.ReadLink;
```
g2mula, line 322: Is there a default value (non null) that you are protecting by needing to test the new value for non null first?
Same for the other ifs below this...

src/Microsoft.OData.Client/MaterializeFromAtom.cs
```diff
@@ -11,11 +11,12 @@ namespace Microsoft.OData.Client
     using System;
     using System.Collections;
     using System.Collections.Generic;
-    using Microsoft.OData.Client.Materialization;
-    using Microsoft.OData.Client.Metadata;
     using System.Diagnostics;
     using System.Xml;
     using Microsoft.OData;
+    using Microsoft.OData.Client.Materialization;
```
g2mula, line 17: This file (non-)edit was probably not meant for checkin...

src/Microsoft.OData.Client/Microsoft.OData.Client.cs
```diff
@@ -12,16 +12,13 @@
 using System.Reflection;
 using System.Resources;
 using System.Threading;
-#if !PORTABLELIB
```
g2mula, line 15: Yaaaaay!!!

test/FunctionalTests/Microsoft.OData.Client.Tests/Microsoft.OData.Client.Tests.Netcore.csproj
```diff
@@ -11,6 +11,10 @@
     <OutputPath>..\..\..\bin\AnyCPU\$(Configuration)\Test\.NETPortable</OutputPath>
     <DefineConstants>$(DefineConstants);PORTABLELIB;</DefineConstants>
   </PropertyGroup>
+
+  <ItemGroup>
+    <Compile Remove="Tracking\DataServiceContextNoTrackingModeStreamsTests.cs" />
```
g2mula, line 16: Any reason for removing this from the project?

marabooy: I removed this class and forgot to edit this. The reason for removal was this class tested the non-async api which wouldn't work in .net core

src/Microsoft.OData.Client/Microsoft.OData.Client.cs
```diff
@@ -12,16 +12,13 @@
...
-using System.Security.Permissions;
-#endif
 
 namespace Microsoft.OData.Client
 {
     /// <summary>
     ///    AutoGenerated resource class. Usage:
     ///
-    ///        string s = TextRes.GetString(TextRes.MyIdentifier);
+    ///        string s = TextRes.GetString(TextRes.MyIdenfitier);
```
habbes, line 21: Why is getting changed from `MyIdentifier` to `MyIdenfitier`?

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextNoTrackingStreamsTests.cs
```diff
@@ -0,0 +1,450 @@
...
+        public void TestBehaviourShouldRemainTheSameForTheDataContext()
+        {
+            SetupContextWithRequestPipeline(new DataServiceContext[] { DefaultTrackingContext, NonTrackingContext }, false);
+
+            var users = DefaultTrackingContext.Users.ExecuteAsync().GetAwaiter().GetResult();
+            var users2 = NonTrackingContext.Users.ExecuteAsync().GetAwaiter().GetResult();
+            // data should be fetched and the count should be the same
+            Assert.Equal(users.ToList().Count, users2.ToList().Count);
+
+            //  entities should be tracked by no tracking context
```
habbes, line 172: Did you mean "Entities should **not** be tracked by no-tracking context"?

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextNoTrackingStreamsTests.cs
```diff
@@ -0,0 +1,450 @@
...
+            };
+
+            DefaultTrackingContext.AddObject("Users", user);
+            NonTrackingContext.AddObject("Users", user);
+            Assert.NotNull(NonTrackingContext.GetEntityDescriptor(user));
+            Assert.NotNull(DefaultTrackingContext.GetEntityDescriptor(user));
+
+            SaveContextChanges(new DataServiceContext[] { DefaultTrackingContext, NonTrackingContext });
+            Assert.Equal(1, DefaultTrackingContext.Entities.Count);
+            Assert.Equal(1, NonTrackingContext.Entities.Count);
```
habbes, line 239: Does this mean that the non-tracking context is actually tracking the user entities? (which don't have a stream)

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextNoTrackingStreamsTests.cs
```diff
@@ -0,0 +1,450 @@
...
+        [Fact]
+        public async Task GetNamedStreamsShouldNotWorkInNoTrackingModeIfNotExtendsBaseEntity()
+        {
+            SetupContextWithRequestPipeline(new DataServiceContext[] { NonTrackingContext }, true, true);
+
+            var customers = await NonTrackingContext.Customers2.ExecuteAsync();
+
+            Action action = () => NonTrackingContext.GetReadStreamUri(customers.First(), "ProfilePhoto");
+            action.ShouldThrow<InvalidOperationException>();
+        }
```
habbes, line 295: Why should it not work without extending `BaseEntity`?

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextNoTrackingStreamsTests.cs
```diff
@@ -0,0 +1,450 @@
...
+        {
+            foreach (var dataServiceContext in dataServiceContexts)
+            {
+                await dataServiceContext.SaveChangesAsync();
+            }
+        }
+
+        private void SetupContextWithRequestPipelineForSaving(DataServiceContext[] dataServiceContexts, bool forMle)
+        {
+            var response = forMle ? DOCUMENT_RESPONSE : USER_RESPONSE;
```
habbes, line 307: Does `forMle` stand for `forMediaLinkEntities`? I think it's worthwhile to just spell it all out instead of abbreviating.

src/Microsoft.OData.Client/Microsoft.OData.Client.cs
marabooy: seems like the generator got changed again thanks for noticing this

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextNoTrackingStreamsTests.cs
marabooy: for addition it seems that all entities are tracked. only fetched entities are not

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextNoTrackingStreamsTests.cs
marabooy: Adding should behaviour stays the same as the entities are tracked. but fetched entities are not

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextNoTrackingStreamsTests.cs
habbes: But in this case you're testing fetching, not addition. Even your assertions test that the no-tracking context has 0 entities in its entity tracker:

```c#
Assert.Equal(0, NonTrackingContext.EntityTracker.Entities.ToList().Count);
```

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextNoTrackingStreamsTests.cs
marabooy: I think using base entity provides a way to opt-in to the behavior. Once the new changes are done to have these in an interface then this could be refactored a bit to introduce some interface for this and document. probably this would be a great candidate for the new version 8.0 👍

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextNoTrackingStreamsTests.cs
marabooy: Fixed

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -1266,9 +1266,40 @@ public QueryOperationResponse<T> LoadProperty<T>(object entity, string propertyN
         public Uri GetReadStreamUri(object entity)
         {
             Util.CheckArgumentNull(entity, "entity");
-            EntityDescriptor box = this.entityTracker.GetEntityDescriptor(entity);
+            EntityDescriptor box = this.GetEntityDescriptorForStreamEntity(entity);
             return box.ReadStreamUri;
         }
+        /// <summary>
```
g2mula, line 1272:
```suggestion

        /// <summary>
```
This actually produces a warning
```SA1513 : CSharp.Layout : Statements or elements wrapped in curly brackets must be followed by a blank line.```

File: src/Microsoft.OData.Client/DataServiceContext.cs [hunk 34]
```diff
@@ -1266,9 +1266,40 @@ public QueryOperationResponse<T> LoadProperty<T>(object entity, string propertyN
...
         }
+        /// <summary>
+        /// This function is used to get the entity descriptor that is to be checked when trying to resolve streams.
+        /// For other use cases use the <see cref="GetEntityDescriptor"/> function
+        /// </summary>
+        /// <param name="entity"></param>
+        /// <returns></returns>
+        private EntityDescriptor GetEntityDescriptorForStreamEntity(object entity)
+        {
+            EntityDescriptor box;
```
g2mula, line 1280: nit: naming: ```box```?

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
```diff
@@ -671,6 +673,18 @@ private void MaterializeResolvedEntry(MaterializerEntry entry, bool includeLinks
             if (entity != null)
             {
                 entity.Context = this.EntityTrackingAdapter.Context;
+                // if NoTracking
```
g2mula, line 676:
```suggestion

                // if NoTracking
```
Produces warning
```CSharp.Layout : A single-line comment must be preceded by a blank line or another single-line comment, or must be the first item in its scope. To ignore this error when commenting out a line of code, begin the comment with '////' rather than '//'.```

File: src/Microsoft.OData.Client/MaterializeFromAtom.cs [hunk 36]
```diff
@@ -11,11 +11,11 @@ namespace Microsoft.OData.Client
     using System;
     using System.Collections;
     using System.Collections.Generic;
-    using Microsoft.OData.Client.Materialization;
-    using Microsoft.OData.Client.Metadata;
     using System.Diagnostics;
     using System.Xml;
+    using Materialization;
```
g2mula, line 16: Probably didn't mean to check this file in?

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextNoTrackingStreamsTests.cs
```diff
@@ -0,0 +1,450 @@
...
+                    new Dictionary<string, string>()
+                    {
+                            {"Content-Type", "text/plain"},
+                    });
+            };
+
+            var streamResponse = await dataServiceContext.GetReadStreamAsync(document, new DataServiceRequestArgs());
+            return streamResponse.Stream;
+        }
+#if NET452
```
g2mula, line 220: But... Why....?
Our package targets .net standard 1.1 why the #ifs...

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
```diff
@@ -671,6 +673,18 @@ private void MaterializeResolvedEntry(MaterializerEntry entry, bool includeLinks
...
             {
                 entity.Context = this.EntityTrackingAdapter.Context;
+
+                if (!entry.IsTracking)
+                {
+                    int? streamDescriptorsCount = entry.EntityDescriptor.StreamDescriptors?.Count;
+                    IEdmEntityType entityType =
+                        this.EntityTrackingAdapter.Model.FindDeclaredType(entry.Entry.TypeName) as IEdmEntityType;
+                    if ( streamDescriptorsCount != null && (streamDescriptorsCount.Value > 0 ||
+                                                            ( entityType != null && entityType.HasStream)))
```
g2mula, line 683:
```suggestion
                    if (streamDescriptorsCount.Value > 0 || entityType?.HasStream == true)
```
The spacings after the opening brackets would create warnings...
The checks for null ca be simplified where necessary.

src/Microsoft.OData.Client/Materialization/FeedAndEntryMaterializerAdapter.cs
```diff
@@ -9,8 +9,10 @@ namespace Microsoft.OData.Client.Materialization
     using System;
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.Linq;
```
g2mula, line 12: This file is probably not meant for checkin...

KanishManuja-MS:
@xuzhg Since you reviewed this, is it ready to be shipped? Are you done with the review?

xuzhg:
>                     }

so, maybe it's better to have a "Debug.Assert(this.Id == null)"

---
Refers to: src/Microsoft.OData.Client/Materialization/MaterializerEntry.cs:342 in 79f8055. [](commit_id = 79f80555201f855ba34f67d4aef300039639eb82, deletion_comment = False)

xuzhg:
@David W, please rebase to the latest master

marabooy:
@xuzhg Updated to the latest master. 
not sure I understand this comment 
> so, maybe it's better to have a "Debug.Assert(this.Id == null)"

xuzhg:
>                     }

There's a comments: "Remove these lines since complex type doesn't have Id."
In my understanding, Complex type doesn't have the Id setting. So, If that's the case here, we should make sure the "this.Id == null", if that's the complex type resource. 
So, is this code only for the complex type resource.  That is  '(this)' is not only related to complex type resource, but entity type resource?

If it's for all type of resource, we can' remove this If clause, and leave a comment like "We don't need to check the Id because it maybe a complex type resource" ?

---
In reply to: [630985001](https://github.com/OData/odata.net/pull/1738#issuecomment-630985001) [](ancestors = 630985001)

---
Refers to: src/Microsoft.OData.Client/Materialization/MaterializerEntry.cs:342 in 6decb8d. [](commit_id = 6decb8dfe118f7ed86614b2245af3a93efc82e2f, deletion_comment = False)

marabooy:
@xuzhg done. I removed the offending line.

xuzhg:
@marabooy please rebase and squash merge the PR. Thanks.

PR #1741, opened 2020-04-09
Convert BaseEntityType to interface

### LINKED ISSUE #1735: Convert BaseEntityType to interface
Currently, all entities in MS OData client are based on `BaseEntityType` - a concrete class lying in an external assembly.
This is very limiting, as it can't be marked as partial, and requires OData users to base their types on it rather than POCO classes.

Please introduce an interface `IBaseEntityType`, and have `BaseEntityType` implement it.
Throughout the entire library, only require the interface `IBaseEntityType`, rather than the concrete class.
The `BaseEntityType` should remain there for backwards compatibility or for other scenarios.
All other features that we might want to add in the future can be accomplished with extension methods or external services or using DI.

### Assemblies affected

Microsoft.OData.Client

### Reproduce steps

`BaseEntityType` should be an interface, not a concrete class, to allow for multi-inheritance.

### Expected result

We should be able to use our own entity base types.

###

This first needs to be dealt in this repo, then, the it can be addressed in the generator as well: https://github.com/OData/ODataConnectedService/issues/88.

weitzhandler:
Fixes #1735

### DIFF
--- a/src/Microsoft.OData.Client/BaseEntityType.cs
+++ b/src/Microsoft.OData.Client/BaseEntityType.cs
@@ -7,7 +7,7 @@
 namespace Microsoft.OData.Client
 {
     /// <summary>
-    /// Base type of entity type to include <see cref="T:Microsoft.OData.Client.DataServiceContext" /> for function and action invocation
+    /// Base type of entity type to include <see cref="Microsoft.OData.Client.DataServiceContext" /> for function and action invocation
     /// </summary>
     public class BaseEntityType : IBaseEntityType
     {
--- a/src/Microsoft.OData.Client/IBaseEntityType.cs
+++ b/src/Microsoft.OData.Client/IBaseEntityType.cs
@@ -7,7 +7,8 @@
 namespace Microsoft.OData.Client
 {
     /// <summary>
-    /// Base type of entity type to include <see cref="T:Microsoft.OData.Client.DataServiceContext" /> for function and action invocation
+    /// Base type of entity type to include <see cref="Microsoft.OData.Client.DataServiceContext" /> for function and action invocation
+
     /// </summary>
     public interface IBaseEntityType
     {
--- a/src/Microsoft.OData.Client/IBaseEntityType.cs
+++ b/src/Microsoft.OData.Client/IBaseEntityType.cs
@@ -1,4 +1,4 @@
-//---------------------------------------------------------------------
+﻿//---------------------------------------------------------------------
 // <copyright file="BaseEntityType.cs" company="Microsoft">
 //      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
 // </copyright>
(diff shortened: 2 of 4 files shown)


src/Microsoft.OData.Client/BaseEntityType.cs
```diff
@@ -6,12 +6,17 @@
...
-        /// <summary>DataServiceContext for query provider</summary>
-        protected internal DataServiceContext Context { get; set; }
-    }
+	/// <summary>
+	/// Base type of entity type to include <see cref="T:Microsoft.OData.Client.DataServiceContext" /> for function and action invocation
+	/// </summary>
+	public class BaseEntityType : IBaseEntityType
+	{
+		/// <inheritdoc/>
+		protected internal DataServiceContext Context { get; set; }
```
chrisspre, line 15: Can't this be simplified to just `public DataServiceContext Context { get; set; }` instead of the explicit interface implementation?

Sreejithpin:
Hi,
 We recently had changed our project structure.  Can you please either rebase your code with out new master and update your PR, or create a new branch from master and make your updates in that branch and put a PR. Thanks a lot

KenitoInc:
@weitzhandler 
Hey,
Kindly rebase your PR and fix conflicts.

weitzhandler:
@KenitoInc sorry for the delay.
Rebase completed.

weitzhandler:
@Sreejithpin?

src/Microsoft.OData.Client/BaseEntityType.cs
```diff
@@ -7,16 +7,30 @@
 namespace Microsoft.OData.Client
 {
     /// <summary>
-    /// Base type of entity type to include <see cref="Microsoft.OData.Client.DataServiceContext" /> for function and action invocation
+    /// Base type of entity type to include <see cref="T:Microsoft.OData.Client.DataServiceContext" /> for function and action invocation
```
KenitoInc, line 10: We shouldn't be adding the `T:`

src/Microsoft.OData.Client/IBaseEntityType.cs
```diff
@@ -0,0 +1,24 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="BaseEntityType.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Client
+{
+    /// <summary>
+    /// Base type of entity type to include <see cref="T:Microsoft.OData.Client.DataServiceContext" /> for function and action invocation
```
KenitoInc, line 10:
```suggestion
    /// Base type of entity type to include <see cref="Microsoft.OData.Client.DataServiceContext" /> for function and action invocation

```

KenitoInc (changes requested):

src/Microsoft.OData.Client/BaseEntityType.cs
weitzhandler: Done.

src/Microsoft.OData.Client/IBaseEntityType.cs
weitzhandler: Done.

gathogojr (approved):
LGTM

KenitoInc (approved):
We next to cleanup this in the next breaking change release.

src/Microsoft.OData.Client/BaseEntityType.cs
```diff
@@ -9,14 +9,28 @@ namespace Microsoft.OData.Client
...
+        /// <summary>
+        /// DataServiceContext for query provider.
+        /// </summary>
         protected internal DataServiceContext Context { get; set; }
 
+        DataServiceContext IBaseEntityType.Context
+        {
+            get => Context;
+            set => Context = value;
+        }
```
habbes, line 23: Why make these explicitly implement the interface instead of just making the backing properties public?

habbes (approved):

Issue #1678, opened 2020-02-25
Problem with navigation property in complex type


hugener:
I originally posted my issue at: https://github.com/unchase/Unchase.Odata.Connectedservice/issues/38, but was referred here.

I'm trying to consume an OData service and the code generation (Unchase) worked just fine.
However, when I execute a query I get the following error:
The property 'X' is of entity type and it cannot be a property of the type 'Y', which is not of entity type. Only entity types can contain navigation properties.

I checked the definition and there is a complex type, which has a navigation property to an entity type. As the message states this is not supported.

However, according to the following link, it should be supported in OData Core Lib V7: https://docs.microsoft.com/en-us/odata/odatalib/navigation-under-complex
I'm not sure though, if that version maps to the dlls used by the generated connected service:
Microsoft.OData.Client 7.6.3
Microsoft.OData.Core 7.6.3
Microsoft.OData.Edm 7.6.3
etc.

I don't control the service I'm consuming, so I can't change that.
Does anyone know if it's possible to get this to work?

### Assemblies affected

*Microsoft.OData.Client 7.6.3
Microsoft.OData.Core 7.6.3
Microsoft.OData.Edm 7.6.3*

### Reproduce steps

*Unfortunately the service I'm trying to query is provided by a third party and there is not a lot of documentation available. The only reference I find is from this training material:
https://www.celum.com/en/event/celum-development-training-200420/*

Below is an excerpt of the metadata:
```
<ComplexType Name="AssetInformationFieldValues">
...
    <NavigationProperty Name="additionalkeywords" Type="Collection(com.celum.cora.Node)"/>
...
</ComplexType>
```

```
<EntityType Name="Node">
    <Key>
    <PropertyRef Name="id"/>
    </Key>
    <Property Name="id" Type="Edm.Int64"/>
    <Property Name="name" Type="Collection(com.celum.cora.LocalizedValue)"/>
...
</EntityType>
```

### Expected result

*Successful query*

### Actual result

*InvalidOperationException:
The property 'additionalkeywords' is of entity type and it cannot be a property of the type 'AssetInformationFieldValues', which is not of entity type. Only entity types can contain navigation properties.*

hugener:
Great to hear it'll be implement. Is it possible to get an ETA on when it'll be out?
I'm just trying to plan whether we should wait for it or start implementing a client manually.

PR #1743, opened 2020-04-15
entity navigation property support in complex types

ElizabethOkerio:
### Issues

*This pull request fixes issue #1678.*

### Description
<pre>
   //EntityType
    public class Person 
    { 
        [Key]
        public string UserName { get; set; } 
        public Address Address { get; set; } 
        public List<Address> Addresses { get; set; } 
    } 
    //EntityType
    public class City 
    { 
        [Key]
    } 
    //ComplexType
    public class Address 
    { 
        public string Road { get; set; } 
        public City City { get; set; } 
    } 
}
</pre>

Currently, if you have such kind of a model where a complex type has an entity type navigation, you'll not be able to create a new entity using <code> Microsoft.OData.Client </code>. This is because
<code> Microsoft.OData.Client </code>  does not support entity navigations in complex types.

However, according to the following link, it is supported in OData Core Lib V7: https://docs.microsoft.com/en-us/odata/odatalib/navigation-under-complex 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/ClientEdmModel.cs
+++ b/src/Microsoft.OData.Client/ClientEdmModel.cs
@@ -591,32 +591,34 @@ namespace Microsoft.OData.Client
                 propertyEdmType.TypeKind == EdmTypeKind.Collection,
                 "Property kind should be Entity, Complex, Enum, Primitive or Collection.");
 
-            IEdmProperty edmProperty;
+            IEdmProperty edmProperty = null;
             bool isPropertyNullable = ClientTypeUtil.CanAssignNull(propertyInfo.PropertyType);
             if (propertyEdmType.TypeKind == EdmTypeKind.Entity || (propertyEdmType.TypeKind == EdmTypeKind.Collection && ((IEdmCollectionType)propertyEdmType).ElementType.TypeKind() == EdmTypeKind.Entity))
             {
-                IEdmEntityType declaringEntityType = declaringType as IEdmEntityType;
-                if (declaringEntityType == null)
+                if (declaringType.TypeKind == EdmTypeKind.Entity || declaringType.TypeKind == EdmTypeKind.Complex)
                 {
-                    throw c.Error.InvalidOperation(c.Strings.ClientTypeCache_NonEntityTypeCannotContainEntityProperties(propertyInfo.Name, propertyInfo.DeclaringType.ToString()));
-                }
+                    if (declaringType as IEdmEntityType == null && declaringType as IEdmComplexType == null)
+                    {
+                        throw c.Error.InvalidOperation(c.Strings.ClientTypeCache_NonEntityTypeOrNonComplexTypeCannotContainEntityProperties(propertyInfo.Name, propertyInfo.DeclaringType.ToString()));
+                    }
 
-                // Create a navigation property representing one side of an association.
-                // The partner representing the other side exists only inside this property and is not added to the target entity type,
-                // so it should not cause any name collisions.
-                edmProperty = EdmNavigationProperty.CreateNavigationPropertyWithPartner(
-                    ClientTypeUtil.GetServerDefinedName(propertyInfo),
-                    propertyEdmType.ToEdmTypeReference(isPropertyNullable),
-                    /*dependentProperties*/ null,
-                    /*principalProperties*/ null,
-                    /*containsTarget*/ false,
-                    EdmOnDeleteAction.None,
-                    "Partner",
-                    declaringEntityType.ToEdmTypeReference(true),
-                    /*partnerDependentProperties*/ null,
-                    /*partnerPrincipalProperties*/ null,
-                    /*partnerContainsTarget*/ false,
-                    EdmOnDeleteAction.None);
+                    // Create a navigation property representing one side of an association.
+                    // The partner representing the other side exists only inside this property and is not added to the target entity type,
+                    // so it should not cause any name collisions.
+                    edmProperty = EdmNavigationProperty.CreateNavigationPropertyWithPartner(
+                        ClientTypeUtil.GetServerDefinedName(propertyInfo),
+                        propertyEdmType.ToEdmTypeReference(isPropertyNullable),
+                        /*dependentProperties*/ null,
+                        /*principalProperties*/ null,
+                        /*containsTarget*/ false,
+                        EdmOnDeleteAction.None,
+                        "Partner",
+                        declaringType.ToEdmTypeReference(true),
+                        /*partnerDependentProperties*/ null,
+                        /*partnerPrincipalProperties*/ null,
+                        /*partnerContainsTarget*/ false,
+                        EdmOnDeleteAction.None);
+                }
             }
             else
             {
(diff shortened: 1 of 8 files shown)


src/Microsoft.OData.Client/ClientEdmModel.cs
```diff
@@ -591,32 +591,60 @@ private IEdmProperty CreateEdmProperty(IEdmStructuredType declaringType, Propert
...
+                        /*principalProperties*/ null,
+                        /*containsTarget*/ false,
+                        EdmOnDeleteAction.None,
+                        "Partner",
+                        declaringComplexType.ToEdmTypeReference(true),
+                        /*partnerDependentProperties*/ null,
+                        /*partnerPrincipalProperties*/ null,
+                        /*partnerContainsTarget*/ false,
+                        EdmOnDeleteAction.None);
+                }
```
habbes, line 647: What's the difference between this block of code and the if-block above?

src/Microsoft.OData.Edm/Schema/EdmNavigationProperty.cs
```diff
@@ -249,6 +260,24 @@ private static IEdmEntityType GetEntityType(IEdmTypeReference type)
...
+            {
+                type = ((IEdmCollectionType)type.Definition).ElementType;
+                if (type.IsComplex())
+                {
+                    complexType = (IEdmComplexType)type.Definition;
+                }
+            }
+
+            return complexType;
+        }
```
habbes, line 280: I think it would be easier to follow this function if you returned the complex type immediately in each branch, and returned null at the end.

```c#
if (type.IsComplex())
{
  return (IEdmComplexType) type.Definition;
}
if (type.IsCollection())
{
   // ...
   if (type.IsComplex())
   {
      return ...
    }
}
return null
```
That way the reader doesn't have to keep each branch path in the mind after it's return value is already known. It's also easier to tell that null is returned if the type is not a complex type.

test/FunctionalTests/Microsoft.OData.Client.Tests/ComplexNavigations/ComplexNavigationTests.cs
```diff
@@ -0,0 +1,65 @@
...
+        /// <summary>
+        /// This test checks whether the declaring type kind of the navigation property created is complex.
+        /// </summary>
+        [Fact]
+        public void DeclaringTypeOFAnEntityNavigationCanBeAComplexType()
+        {
+            //Arrange
+            Type complexDeclaringType = typeof(Address);          
+            Type entityNavigationType = typeof(City);        
+            EdmTypeKind expecteddeclaringTypeKind = EdmTypeKind.Complex;
```
habbes, line 29: Could you use camel case here? `expectedDeclaringTypeKind`

src/Microsoft.OData.Client/ClientEdmModel.cs
```diff
@@ -591,32 +591,60 @@ private IEdmProperty CreateEdmProperty(IEdmStructuredType declaringType, Propert
...
             {
-                IEdmEntityType declaringEntityType = declaringType as IEdmEntityType;
-                if (declaringEntityType == null)
+                if (declaringType.TypeKind == EdmTypeKind.Entity)
                 {
-                    throw c.Error.InvalidOperation(c.Strings.ClientTypeCache_NonEntityTypeCannotContainEntityProperties(propertyInfo.Name, propertyInfo.DeclaringType.ToString()));
+                    IEdmEntityType declaringEntityType = declaringType as IEdmEntityType;
+                    if (declaringEntityType == null)
+                    {
+                        throw c.Error.InvalidOperation(c.Strings.ClientTypeCache_NonEntityTypeCannotContainEntityProperties(propertyInfo.Name, propertyInfo.DeclaringType.ToString()));
```
g2mula, line 603: This exception message no longer seems valid since a ```ComplexTypesCanAlsoContainEntityProperties```

src/Microsoft.OData.Client/ClientEdmModel.cs
```diff
@@ -591,32 +591,60 @@ private IEdmProperty CreateEdmProperty(IEdmStructuredType declaringType, Propert
...
+                    // so it should not cause any name collisions.
+                    edmProperty = EdmNavigationProperty.CreateNavigationPropertyWithPartner(
+                        ClientTypeUtil.GetServerDefinedName(propertyInfo),
+                        propertyEdmType.ToEdmTypeReference(isPropertyNullable),
+                        /*dependentProperties*/ null,
+                        /*principalProperties*/ null,
+                        /*containsTarget*/ false,
+                        EdmOnDeleteAction.None,
+                        "Partner",
+                        declaringEntityType.ToEdmTypeReference(true),
```
g2mula, line 617: This seems the only difference between this block and the else block below... Could have a test to create it for both complex and entity types and if the containing type is none of the two then throw an exception with a more correct(?) message...

src/Microsoft.OData.Client/ClientEdmModel.cs
```diff
@@ -625,7 +653,8 @@ private IEdmProperty CreateEdmProperty(IEdmStructuredType declaringType, Propert
 
             edmProperty.SetClientPropertyAnnotation(new ClientPropertyAnnotation(edmProperty, propertyInfo, this));
             return edmProperty;
-        }
+
```
g2mula, line 656: nit: probably not meant for checkin...

ElizabethOkerio:
> Looks good to me.
> However, not sure whether this is meant ot add support for collection of complex types as well. If it does, then kindly a test case for that too.

@habbes  what do you mean by adding support for collection of complex types

gathogojr:
@ElizabethOkerio @g2mula Regarding the suggestion to use `nameof`, it so happens that its just a syntactic sugar for `string.Format()`. The issue is, it makes use of the overload without the `IFormatProvider` parameter. The build pipeline for the repo reports warnings when `string.Format` is used without the `IFormatProvider` parameter... unless something has changed between when I last tried to use it and now.
Kindly verify otherwise we'll end up with more warnings and a less happy G2 :)

ElizabethOkerio:
> @ElizabethOkerio @g2mula Regarding the suggestion to use nameof, it so happens that its just a syntactic sugar for string.Format(). The issue is, it makes use of the overload without the IFormatProvider parameter. The build pipeline for the repo reports warnings when string.Format is used without the IFormatProvider parameter... unless something has changed between when I last tried to use it and now.
> Kindly verify otherwise we'll end up with more warnings and a less happy G2 :)

fixed.

src/Microsoft.OData.Client/ClientEdmModel.cs
```diff
@@ -591,32 +591,34 @@ private IEdmProperty CreateEdmProperty(IEdmStructuredType declaringType, Propert
...
             bool isPropertyNullable = ClientTypeUtil.CanAssignNull(propertyInfo.PropertyType);
             if (propertyEdmType.TypeKind == EdmTypeKind.Entity || (propertyEdmType.TypeKind == EdmTypeKind.Collection && ((IEdmCollectionType)propertyEdmType).ElementType.TypeKind() == EdmTypeKind.Entity))
             {
-                IEdmEntityType declaringEntityType = declaringType as IEdmEntityType;
-                if (declaringEntityType == null)
+                if (declaringType.TypeKind == EdmTypeKind.Entity || declaringType.TypeKind == EdmTypeKind.Complex)
                 {
-                    throw c.Error.InvalidOperation(c.Strings.ClientTypeCache_NonEntityTypeCannotContainEntityProperties(propertyInfo.Name, propertyInfo.DeclaringType.ToString()));
-                }
+                    if (declaringType as IEdmEntityType == null && declaringType as IEdmComplexType == null)
```
KanishManuja-MS, line 600: Can probably use is instead of as.

PR #1749, opened 2020-04-21
Add support for Json Batch Requests in Odata Client

KenitoInc:
### Description

Currently Odata Client only supports multipart/mixed batch requests while ODL has support for both json batch requests and multipart/mixed batch requests. We are adding support for OData Json Batch Requests in OData Client.

We are adding a new `SaveChangesOption` flag named `UseJsonBatch`. When you add the flag in batch requests when calling SaveChangesAsync, the request and response will have Content-Type as `application/json`

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/BatchSaveResult.cs
+++ b/src/Microsoft.OData.Client/BatchSaveResult.cs
@@ -54,6 +54,9 @@ namespace Microsoft.OData.Client
         /// <summary>If batch request is allowed to use Relative Uri.</summary>
         private bool useRelativeUri = false;
 
+        /// <summary>If we are making a json batch request.</summary>
+        private bool useJsonBatch = false;
+
         #endregion
 
         /// <summary>
@@ -72,6 +75,7 @@ namespace Microsoft.OData.Client
             this.Queries = queries;
             this.streamCopyBuffer = new byte[StreamCopyBufferSize];
             this.useRelativeUri = Util.UseRelativeUri(options);
+            this.useJsonBatch = Util.UseJsonBatch(options);
         }
 
         /// <summary>returns true since this class handles batch requests.</summary>
@@ -280,6 +284,15 @@ namespace Microsoft.OData.Client
         }
 
         /// <summary>
+        /// Creates the type of the application/json content.
+        /// </summary>
+        /// <returns>An application/json header</returns>
+        private static string CreateApplicationJsonContentType()
+        {
+            return XmlConstants.MimeApplicationType + "/" + XmlConstants.MimeJsonSubType;
+        }
+
+        /// <summary>
         /// Creates a ODataRequestMessage for batch request.
         /// </summary>
         /// <returns>Returns an instance of ODataRequestMessage for the batch request.</returns>
@@ -288,7 +301,14 @@ namespace Microsoft.OData.Client
             Uri requestUri = UriUtil.CreateUri(this.RequestInfo.BaseUriResolver.GetBaseUriWithSlash(), UriUtil.CreateUri("$batch", UriKind.Relative));
             HeaderCollection headers = new HeaderCollection();
             headers.SetRequestVersion(Util.ODataVersion4, this.RequestInfo.MaxProtocolVersionAsVersion);
-            headers.SetHeader(XmlConstants.HttpContentType, CreateMultiPartMimeContentType());
+            if(useJsonBatch)
+            {
+                headers.SetHeader(XmlConstants.HttpContentType, CreateApplicationJsonContentType());
+            }
+            else
+            {
+                headers.SetHeader(XmlConstants.HttpContentType, CreateMultiPartMimeContentType());
+            }
             this.RequestInfo.Format.SetRequestAcceptHeaderForBatch(headers);
 
             return this.CreateTopLevelRequest(XmlConstants.HttpMethodPost, requestUri, headers, this.RequestInfo.HttpStack, null /*descriptor*/);
@@ -753,7 +773,11 @@ namespace Microsoft.OData.Client
                         #endregion
 
                         default:
-                            Error.ThrowBatchExpectedResponse(InternalError.UnexpectedBatchState);
+                            // In ODataJsonLightBatchReader, readerState remains Initial after calling Read() the first time
+                            if (!this.useJsonBatch)
+                            {
+                                Error.ThrowBatchExpectedResponse(InternalError.UnexpectedBatchState);
+                            }
                             break;
                     }
                 }
--- a/src/Microsoft.OData.Client/DataServiceClientFormat.cs
+++ b/src/Microsoft.OData.Client/DataServiceClientFormat.cs
@@ -26,6 +26,9 @@ namespace Microsoft.OData.Client
         /// <summary>MIME type for JSON bodies in light mode (http://www.iana.org/assignments/media-types/application/).</summary>
         private const string MimeApplicationJsonODataLight = "application/json;odata.metadata=minimal";
 
+        /// <summary>MIME type for JSON (https://www.ietf.org/rfc/rfc4627.txt).</summary>
+        private const string MimeApplicationJson = "application/json";
+
         /// <summary>MIME type for JSON bodies in light mode with all metadata.</summary>
         private const string MimeApplicationJsonODataLightWithAllMetadata = "application/json;odata.metadata=full";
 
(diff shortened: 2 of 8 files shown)


test/EndToEndTests/Tests/Client/AsynchronousTests/AsyncMethodTests.cs
```diff
@@ -110,6 +110,13 @@ public async Task SaveChangesTest()
             var dataServiceResponse = await context.SaveChangesAsync(SaveChangesOptions.BatchWithIndependentOperations | SaveChangesOptions.UseRelativeUri);
             Assert.AreEqual((dataServiceResponse.First() as ChangeOperationResponse).StatusCode, 201, "StatusCode == 201");
 
+            // UseJsonBatch
```
mikepizzo, line 113: Note that this doesn't actually appear to test that the batch is created (or read) as JSON -- just that setting the option doesn't break anything.

src/Microsoft.OData.Client/BatchSaveResult.cs
```diff
@@ -753,7 +773,11 @@ private IEnumerable<OperationResponse> HandleBatchResponse(ODataBatchReader batc
                         #endregion
 
                         default:
-                            Error.ThrowBatchExpectedResponse(InternalError.UnexpectedBatchState);
+                            // In ODataJsonLightBatchReader, readerState remains Initial after calling Read() the first time
+                            if (!this.useJsonBatch)
+                            {
+                                Error.ThrowBatchExpectedResponse(InternalError.UnexpectedBatchState);
+                            }
```
mikepizzo, line 780: Really? This seems like a bug in the ODataJsonLightBatchReader -- your state shouldn't depend on the format.

src/Microsoft.OData.Client/BatchSaveResult.cs
```diff
@@ -279,6 +283,15 @@ private static string CreateMultiPartMimeContentType()
             return string.Format(CultureInfo.InvariantCulture, "{0}; {1}={2}_{3}", XmlConstants.MimeMultiPartMixed, XmlConstants.HttpMultipartBoundary, XmlConstants.HttpMultipartBoundaryBatch, Guid.NewGuid());
         }
 
+        /// <summary>
+        /// Creates the type of the application/json content.
+        /// </summary>
+        /// <returns>A application/json header</returns>
```
odero, line 289:
```suggestion
        /// <returns>An application/json header</returns>

```

src/Microsoft.OData.Client/DataServiceClientFormat.cs
```diff
@@ -175,12 +175,14 @@ internal void SetRequestAcceptHeaderForCount(HeaderCollection headers)
         }
 
         /// <summary>
-        /// Sets the value of the Accept header for a count request (will set it to 'multipart/mixed').
+        /// Sets the value of the Accept header for a batch request
+        /// Will set it to 'multipart/mixed' for a multipart batch request
+        /// Will set it to 'application/json;odata.metadata=minimal' for a json batch request
```
odero, line 180: It's important to note that multipart batch request header does not automatically imply multipart batch response (same for json batch request & json batch response). It is possible to send application/json content-type and receive multipart response. In fact for a default webapi setup sending a json batch response returns a multipart response unless you explicitly set accept headers as json as well.
Maybe we could have it as the default behaviour but then give the user the option to [un]set it
cc @mikepizzo for confirmation on whether this is by design or an oversight on the odata service side

src/Microsoft.OData.Client/DataServiceClientFormat.cs
```diff
@@ -175,12 +175,14 @@ internal void SetRequestAcceptHeaderForCount(HeaderCollection headers)
...
 
         /// <summary>
-        /// Sets the value of the Accept header for a count request (will set it to 'multipart/mixed').
+        /// Sets the value of the Accept header for a batch request
+        /// Will set it to 'multipart/mixed' for a multipart batch request
+        /// Will set it to 'application/json;odata.metadata=minimal' for a json batch request
         /// </summary>
         /// <param name="headers">The headers to modify.</param>
-        internal void SetRequestAcceptHeaderForBatch(HeaderCollection headers)
+        internal void SetRequestAcceptHeaderForBatch(HeaderCollection headers, bool useJsonBatch)
```
odero, line 183: Do you really need this bool param? It's possible to infer if it's json based on the headers collection

test/EndToEndTests/Tests/Client/AsynchronousTests/AsyncMethodTests.cs
odero: +1 You actually want to check the both request and response to see if they are json

src/Microsoft.OData.Client/BatchSaveResult.cs
KenitoInc: @mikepizzo Definitely a bug in `ODataJsonLightBatchReader`

src/Microsoft.OData.Client/DataServiceClientFormat.cs
KenitoInc: I found this abit straight forward

odero: It seems unnecessary to modify the method signature since inferring within the method is also quite straightforward

KenitoInc: I have fixed this.

src/Microsoft.OData.Client/DataServiceClientFormat.cs
mikepizzo: It is certainly possible for a mulitpart batch request to return a json batch response and vice-versa.  I *thought* we specified that the default response was the same as the request, but I didn't find that doing a quick search and it sounds from the comment above like that's not the behavior of the webapi stack.

For the client, I would suggest that we always specify both Content-Type and Accept headers; in the UseJson case we would set *both* to application/json, and in the multipart scenario set them both to multipart. I don't think there's a benefit in providing an option for the data service client to send a request as multipart and get back JSON response, or vice-versa.

src/Microsoft.OData.Client/BatchSaveResult.cs
```diff
@@ -288,7 +301,14 @@ private ODataRequestMessageWrapper CreateBatchRequest()
...
             headers.SetRequestVersion(Util.ODataVersion4, this.RequestInfo.MaxProtocolVersionAsVersion);
-            headers.SetHeader(XmlConstants.HttpContentType, CreateMultiPartMimeContentType());
+            if(useJsonBatch)
+            {
+                headers.SetHeader(XmlConstants.HttpContentType, CreateApplicationJsonContentType());
+            }
+            else
+            {
+                headers.SetHeader(XmlConstants.HttpContentType, CreateMultiPartMimeContentType());
+            }
```
paulodero, line 311: What happens when the client sends `IEEE754Compatible` `Json `request?

src/Microsoft.OData.Client/BatchSaveResult.cs
```diff
@@ -279,6 +283,15 @@ private static string CreateMultiPartMimeContentType()
             return string.Format(CultureInfo.InvariantCulture, "{0}; {1}={2}_{3}", XmlConstants.MimeMultiPartMixed, XmlConstants.HttpMultipartBoundary, XmlConstants.HttpMultipartBoundaryBatch, Guid.NewGuid());
         }
 
+        /// <summary>
+        /// Creates the type of the application/json content.
+        /// </summary>
+        /// <returns>An application/json header</returns>
+        private static string CreateApplicationJsonContentType()
```
gathogojr, line 290: @KenitoInc Any reason you can't use a `const string` to achieve the same purpose? Use of `string.Format` just introduces a performance penalty IMO.

src/Microsoft.OData.Client/BatchSaveResult.cs
KenitoInc: @paulodero In the above code, we are setting headers for the batch `top level request`.
`IEEE754Compatible Json` request headers are in the individual requests

paulodero: I get it.

src/Microsoft.OData.Client/BatchSaveResult.cs
KenitoInc: #Resolved

src/Microsoft.OData.Client/DataServiceClientFormat.cs
KenitoInc: #Resolved

src/Microsoft.OData.Client/BatchSaveResult.cs
KenitoInc: #Resolved

test/EndToEndTests/Tests/Client/AsynchronousTests/AsyncMethodTests.cs
KenitoInc: #Resolved

PR #1757, opened 2020-04-28
Support JSON Metadata writer using System.Text.Json

xuzhg:
###  Description
Using System.Text.Json

1. Support JSON Metadata writer
2. Provide the public APIs for customer to use to output JSON metadata

Noted:
1) only support in .net standard 2.0, so it's not supported in .NET 45
2) only provide the API using System.Text.Json

- [ ] *Test cases added*

### Additional work necessary


xuzhg:
Closed as duplicated at #1775

PR #1760, opened 2020-05-04
Fix WebApi issue https://github.com/OData/WebApi/issues/2136: IN operator with double quote fails

xuzhg:
…ator with double quote fails

### Issues

*This pull request fixes issue https://github.com/OData/WebApi/issues/2136.*

### Description

IN operator can not pass the double quote. This PR try to fix it in single quote string and double quote string.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -151,61 +151,11 @@ namespace Microsoft.OData.UriParser
                 switch (ch)
                 {
                     case '"':
+                        i = ProcessDoubleQuotedStringItem(i, normalizedText, sb);
+                        break;
+
                     case '\'':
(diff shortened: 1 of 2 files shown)


xuzhg:
@mikepizzo @KanishManuja-MS @Sreejithpin @marabooy @gathogojr The CI is pass at https://identitydivision.visualstudio.com/OData/_build/results?buildId=521661&view=results. Please take a look this PR.

gathogojr:
@xuzhg Hopefully the code and test coverage for ODL is good and comprehensive but I wonder whether there's any setup that would make it possible to run tests in OData Web Api repo against the output  libraries from any PR that affects ODL

xuzhg:
@gathogojr That's good idea.
OData.net repo has the package verifications at: https://github.com/OData/odata.net/tree/master/test/PackageTests
But, it can't cover the WebAPI. 
Besides, we have the Up-level testing within release at OData Onenote -> Build -> Release-> Up-level testing. However, it's not covered in recently release.

Issue #851, opened 2017-06-22
DataServiceQuery makes a "by key" request when Where clause compares just the ID, causing exception instead of empty result if the entity is not found.


fpintos:
### Summary

When using an ODATA client built with the [Microsoft.OData.Client](https://www.nuget.org/packages/Microsoft.OData.Client/) package, a Where clause that just compares the id property becomes a by-key request instead of using $filter. If an entity with that id is not found, the query ends up with an exception instead of returning empty results.

### Assemblies affected

[Microsoft.OData.Client](https://www.nuget.org/packages/Microsoft.OData.Client/) 6.13.0

### Reproduce steps
Consider the following queries:
1. [http://services.odata.org/V4/OData/OData.svc/Products(-1)](http://services.odata.org/V4/OData/OData.svc/Products(-1))
2. [http://services.odata.org/V4/OData/OData.svc/Products?$filter=ID eq -1 or ID eq -2&$top=1](http://services.odata.org/V4/OData/OData.svc/Products?$filter=ID%20eq%20-1%20or%20ID%20eq%20-2&$top=1)
3. [http://services.odata.org/V4/OData/OData.svc/Products?$filter=ID eq -1&$top=1](http://services.odata.org/V4/OData/OData.svc/Products?$filter=ID%20eq%20-1&$top=1)

The first one returns 404 and the other two return 200 with empty results.
So far this is expected.

Install LINQPad and setup a connection to the reference service:
*. Setup [LINQPad](http://www.linqpad.net/).
*. Install the [ODATAv4 driver](https://github.com/meancrazy/LINQPadOData4) by clicking "Add Connection" followed by "View more drivers".
*. Add an ODATAv4 connection to http://services.odata.org/V4/OData/OData.svc/
*. Open a new LINQPad query, select the connection above and select "C# Expression" language.

Now issue the following queries:
```c#
this.Products.ByKey(-1).GetValue()
```
This correctly executes the query using the 1st URL above and the result is the expected exception, indicating the entity was not found.

```c#
this.Products.Where(x => x.ID == -1 || x.ID == -2)
```
This correctly executes the query using the 2nd URL above and the result is an expected empty list.

**However**
```c#
this.Products.Where(x => x.ID == -1)
```
This results in an unexpected exception.

### Expected result
One would expect this query to issue a $filter, like URL 3 above, in which case it would return an empty list, well within the semantics of "Where". 

### Actual results
[...]

PR #1762, opened 2020-05-05
Enable Where clause to generate $filter query options for key predicates

KenitoInc:
### Issues

*This pull request fixes issue #851 .*

### Description
The `Where` clause generates a Uri with a $filter where we have a non-key predicate e.g
`var books = dsc.Books.Where(b => b.Title == "B1");` creates the Uri below
`https://serviceRoot/Books?$filter=Title eq 'B1'`
In this case, if the `Title` is not found, the query will return an empty collection.

When we have a key in the predicate, the `Where` clause generates a Uri with a ByKey resource path e.g 
`var books = dsc.Books.Where(b => b.Id == 1);` creates the Uri below
`https://serviceRoot/Books(1)`
In this case, if the `Id` is not found, the query will throw an exception.

By design `Where` should not throw an exception whenever the predicate does not match any value/element in the source collection.

### Solution
Changing the current behavior will be a breaking change. So I have added a `DataServiceContext.KeyComparisonGeneratesFilterQuery` property which is `false` by default.
So if a customer want to ensure that a Uri with a $filter query option is generated for key predicates in the `Where` clause, they need to set the `KeyComparisonGeneratesFilterQuery` property as `true`

```csharp
Container context = new Container(uri);
context.KeyComparisonGeneratesFilterQuery= true;
```

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
+++ b/src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
@@ -45,9 +45,9 @@ namespace Microsoft.OData.Client
         private readonly DataServiceContext context;
 
         /// <summary>
-        /// /// <summary>Whether a Where clause that just compares the id property becomes a by-key request instead of using $filter.</summary>
+        /// /// <summary>Whether a Where clause that just compares the id property generates a $filter query option.</summary>
         /// </summary>
-        private static bool makeIdPredicateByKey;
+        private static bool idPredicateGeneratesFilterQueryOption;
 
         /// <summary>Convenience property: model.</summary>
         private ClientEdmModel Model
@@ -62,7 +62,7 @@ namespace Microsoft.OData.Client
         private ResourceBinder(DataServiceContext context)
         {
             this.context = context;
-            makeIdPredicateByKey = context.MakeIdPredicateByKey;
+            idPredicateGeneratesFilterQueryOption = context.IdPredicateGeneratesFilterQueryOption;
         }
 
         /// <summary>Analyzes and binds the specified expression.</summary>
@@ -280,8 +280,8 @@ namespace Microsoft.OData.Client
                     keyPredicates = ExtractKeyPredicate(input, currentPredicates, model, out nonKeyPredicates);
                 }
 
-                // A key predicate can only be applied if makeIdPredicateByKey=true
-                if (keyPredicates != null && makeIdPredicateByKey)
+                // A key predicate can only be applied if IdPredicateGeneratesFilterQueryOption=false
+                if (keyPredicates != null && !idPredicateGeneratesFilterQueryOption)
                 {
                     input.SetKeyPredicate(keyPredicates);
                     input.RemoveFilterExpression();
@@ -294,7 +294,7 @@ namespace Microsoft.OData.Client
                     input.ConvertKeyToFilterExpression();
                 }
 
-                if (keyPredicates == null || !makeIdPredicateByKey)
+                if (keyPredicates == null || idPredicateGeneratesFilterQueryOption)
                 {
                     input.ConvertKeyToFilterExpression();
                     input.AddFilter(inputPredicates);
--- a/src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
+++ b/src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
@@ -44,11 +44,6 @@ namespace Microsoft.OData.Client
         /// </summary>
         private readonly DataServiceContext context;
 
-        /// <summary>
-        /// If true, a Where clause that compares only the key property, will generate a $filter query option.
-        /// </summary>
-        private static bool keyComparisonGeneratesFilterQuery;
-
         /// <summary>Convenience property: model.</summary>
         private ClientEdmModel Model
         {
@@ -62,7 +57,6 @@ namespace Microsoft.OData.Client
         private ResourceBinder(DataServiceContext context)
         {
             this.context = context;
-            keyComparisonGeneratesFilterQuery = context.KeyComparisonGeneratesFilterQuery;
         }
 
         /// <summary>Analyzes and binds the specified expression.</summary>
@@ -163,7 +157,7 @@ namespace Microsoft.OData.Client
         /// <returns>
         /// An equivalent expression to <paramref name="mce"/>, possibly a different one with additional annotations.
         /// </returns>
-        private static Expression AnalyzePredicate(MethodCallExpression mce, ClientEdmModel model)
+        private static Expression AnalyzePredicate(MethodCallExpression mce, ClientEdmModel model, DataServiceContext context)
         {
             Debug.Assert(mce != null, "mce != null -- caller couldn't have know the expression kind otherwise");
             Debug.Assert(mce.Method.Name == "Where", "mce.Method.Name == 'Where' -- otherwise this isn't a predicate");
@@ -281,7 +275,7 @@ namespace Microsoft.OData.Client
                 }
 
                 // A key predicate should only be applied if keyComparisonGeneratesFilterQuery=false
-                if (keyPredicates != null && !keyComparisonGeneratesFilterQuery)
+                if (keyPredicates != null && !context.KeyComparisonGeneratesFilterQuery)
                 {
                     input.SetKeyPredicate(keyPredicates);
                     input.RemoveFilterExpression();
@@ -294,7 +288,7 @@ namespace Microsoft.OData.Client
                     input.ConvertKeyToFilterExpression();
                 }
 
-                if (keyPredicates == null || keyComparisonGeneratesFilterQuery)
+                if (keyPredicates == null || context.KeyComparisonGeneratesFilterQuery)
                 {
                     input.ConvertKeyToFilterExpression();
                     input.AddFilter(inputPredicates);
@@ -1452,7 +1446,7 @@ namespace Microsoft.OData.Client
                     switch (sequenceMethod)
                     {
                         case SequenceMethod.Where:
-                            return AnalyzePredicate(mce, this.Model);
+                            return AnalyzePredicate(mce, this.Model, this.context);
                         case SequenceMethod.Select:
                             return AnalyzeNavigation(mce, this.context);
                         case SequenceMethod.SelectMany:
--- a/src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
+++ b/src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
@@ -269,13 +269,12 @@ namespace Microsoft.OData.Client
                     currentPredicates = currentPredicates.Union(inputPredicates).ToList();
                 }
 
-                if (!input.UseFilterAsPredicate)
+                if (!input.UseFilterAsPredicate && !context.KeyComparisonGeneratesFilterQuery)
                 {
                     keyPredicates = ExtractKeyPredicate(input, currentPredicates, model, out nonKeyPredicates);
                 }
 
-                // A key predicate should only be applied if keyComparisonGeneratesFilterQuery=false
-                if (keyPredicates != null && !context.KeyComparisonGeneratesFilterQuery)
+                if (keyPredicates != null)
                 {
                     input.SetKeyPredicate(keyPredicates);
                     input.RemoveFilterExpression();
@@ -288,7 +287,7 @@ namespace Microsoft.OData.Client
                     input.ConvertKeyToFilterExpression();
                 }
 
-                if (keyPredicates == null || context.KeyComparisonGeneratesFilterQuery)
+                if (keyPredicates == null)
                 {
                     input.ConvertKeyToFilterExpression();
                     input.AddFilter(inputPredicates);
--- a/src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
+++ b/src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
@@ -16,13 +16,11 @@ namespace Microsoft.OData.Client
     using System.Linq;
     using System.Linq.Expressions;
     using System.Reflection;
-    using Microsoft.OData.Client.ALinq.UriParser;
-    using Microsoft.OData.Client.Metadata;
     using Microsoft.OData;
-    using Microsoft.OData.UriParser;
+    using Microsoft.OData.Client.Metadata;
     using Microsoft.OData.Edm;
-    using PathSegmentToken = Microsoft.OData.Client.ALinq.UriParser.PathSegmentToken;
     using NonSystemToken = Microsoft.OData.Client.ALinq.UriParser.NonSystemToken;
+    using PathSegmentToken = Microsoft.OData.Client.ALinq.UriParser.PathSegmentToken;
     #endregion Namespaces
 
     /// <summary>
@@ -46,6 +44,11 @@ namespace Microsoft.OData.Client
         /// </summary>
         private readonly DataServiceContext context;
 
+        /// <summary>
+        /// /// <summary>Whether a Where clause that just compares the id property becomes a by-key request instead of using $filter.</summary>
+        /// </summary>
+        private static bool makeIdPredicateByKey;
+
         /// <summary>Convenience property: model.</summary>
         private ClientEdmModel Model
         {
@@ -59,6 +62,7 @@ namespace Microsoft.OData.Client
         private ResourceBinder(DataServiceContext context)
         {
             this.context = context;
+            makeIdPredicateByKey = context.MakeIdPredicateByKey;
         }
 
         /// <summary>Analyzes and binds the specified expression.</summary>
@@ -276,7 +280,8 @@ namespace Microsoft.OData.Client
                     keyPredicates = ExtractKeyPredicate(input, currentPredicates, model, out nonKeyPredicates);
                 }
 
-                if (keyPredicates != null)
+                // A key predicate can only be applied if makeIdPredicateByKey=true
+                if (keyPredicates != null && makeIdPredicateByKey)
                 {
                     input.SetKeyPredicate(keyPredicates);
                     input.RemoveFilterExpression();
@@ -289,7 +294,7 @@ namespace Microsoft.OData.Client
                     input.ConvertKeyToFilterExpression();
                 }
 
-                if (keyPredicates == null)
+                if (keyPredicates == null || !makeIdPredicateByKey)
                 {
                     input.ConvertKeyToFilterExpression();
                     input.AddFilter(inputPredicates);
--- a/src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
+++ b/src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
@@ -45,9 +45,9 @@ namespace Microsoft.OData.Client
         private readonly DataServiceContext context;
 
         /// <summary>
-        /// /// <summary>Whether a Where clause that just compares the id property generates a $filter query option.</summary>
+        /// If true, a Where clause that compares only the key property, will generate a $filter query option.
         /// </summary>
-        private static bool idPredicateGeneratesFilterQueryOption;
+        private static bool keyComparisonGeneratesFilterQuery;
 
         /// <summary>Convenience property: model.</summary>
         private ClientEdmModel Model
@@ -62,7 +62,7 @@ namespace Microsoft.OData.Client
         private ResourceBinder(DataServiceContext context)
         {
             this.context = context;
-            idPredicateGeneratesFilterQueryOption = context.IdPredicateGeneratesFilterQueryOption;
+            keyComparisonGeneratesFilterQuery = context.KeyComparisonGeneratesFilterQuery;
         }
 
         /// <summary>Analyzes and binds the specified expression.</summary>
@@ -280,8 +280,8 @@ namespace Microsoft.OData.Client
                     keyPredicates = ExtractKeyPredicate(input, currentPredicates, model, out nonKeyPredicates);
                 }
 
-                // A key predicate can only be applied if IdPredicateGeneratesFilterQueryOption=false
-                if (keyPredicates != null && !idPredicateGeneratesFilterQueryOption)
+                // A key predicate should only be applied if keyComparisonGeneratesFilterQuery=false
+                if (keyPredicates != null && !keyComparisonGeneratesFilterQuery)
                 {
                     input.SetKeyPredicate(keyPredicates);
                     input.RemoveFilterExpression();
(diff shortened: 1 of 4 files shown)


KanishManuja-MS:
@KenitoInc Can you please rebase the PR and get someone to approve it from the Nairobi team?

KenitoInc:
@KanishManuja-MS I need to redesign a small feature so as not to break exisiting functionality. In the meantime I have removed this PR from v7.7 milestone

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
```diff
@@ -1197,6 +1197,22 @@ public void Linq_Where_GreaterThan()
...
+        /// <summary>
+        /// LINQ - An expression with an Id as a predicate should create an Uri with a $filter in query options
+        /// </summary>
+        [Fact]
+        public void Linq_Filter_By_Id_Test()
+        {
+            var context = this.CreateWrappedContext<DefaultContainer>().Context;
+            var query = (from c in context.Customer
+                         where c.CustomerId == -10
+                         select new Customer { Name = c.Name, CustomerId = c.CustomerId }) as DataServiceQuery<Customer>;
```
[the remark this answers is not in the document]
KenitoInc: A `DataServiceQuery` must have either `Select` or `GroupBy` statements

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
```diff
@@ -1197,6 +1197,22 @@ public void Linq_Where_GreaterThan()
...
+            var uri = query.ToString();
+            Assert.Contains("$filter", uri);
```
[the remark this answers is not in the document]
KenitoInc: Fixed

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
```diff
@@ -1197,6 +1197,22 @@ public void Linq_Where_GreaterThan()
...
+
+            this.WaitForTestToComplete();
```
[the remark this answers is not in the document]
KenitoInc: Fixed

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
```diff
@@ -1197,6 +1197,36 @@ public void Linq_Where_GreaterThan()
...
+        public void Linq_Where_Generates_Filter_By_Id_When_MakeIdPredicateByKey_Is_True()
+        {
+            var context = this.CreateWrappedContext<DefaultContainer>().Context;
+            context.MakeIdPredicateByKey = false;
+            var query = (from c in context.Customer
+                         where c.CustomerId == -10
+                         select new Customer { Name = c.Name, CustomerId = c.CustomerId }) as DataServiceQuery<Customer>;
+            var uri = query.ToString();
+            Assert.Contains("Customer(-10)", uri);
+        }
```
habbes, line 1228: Judging from the method name and the comment, the intention was to set it true `context.MakeIdPredicateByKey` to true.

However, the fact that this test is passing regardless indicates that there's probably a bug. Maybe the uri contains both the "Customer(-10)" segment and the "$filter=Id eq -10" section. Just to be safe, I think your assertion should check for full uri equality instead of just checking whether the id segment is contained. Or at least specifically assert that it does not contain contain the filter section. But I think the full url check would be safer.

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
```diff
@@ -1197,6 +1197,36 @@ public void Linq_Where_GreaterThan()
...
+        [Fact]
+        public void Linq_Where_Generates_Filter_By_Id_When_MakeIdPredicateByKey_Is_False()
+        {
+            var context = this.CreateWrappedContext<DefaultContainer>().Context;
+            context.MakeIdPredicateByKey = false;
+            var query = (from c in context.Customer
+                         where c.CustomerId == -10
+                         select new Customer { Name = c.Name, CustomerId = c.CustomerId }) as DataServiceQuery<Customer>;
+            var uri = query.ToString();
+            Assert.Contains("$filter=CustomerId eq -10", uri);
```
habbes, line 1212: I think you should also explicitly assert that it does not `Customers(-10)`. Or better yet, verify the entire url with an equality assertion (see related comments on the test below).

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
gathogojr: @habbes I don't think there's any problem with this test. It does confirm that when the value of `MakeIdPredicateKey` is `false`, the key is included in the resource path: `Customer(-10)` while the previous test demonstrated that when the value if `true`, the key field appears on the filter: `$filter=CustomerId eq -10`

habbes: @gathogojr @KenitoInc  I may have gotten things mixed up. But looking at the updated code now, I think I'm still confused. The option `MakeIdPredicateByKey` is supposed to result in `Customer(-10)` when it's set to `true`. However, he had set it to `false` and it still contained `Customer(-10)`. Now looking at the updated test, he has set it to `true` and it still contains `Customer(-10)`. It's as if the key segment is in the url regardless of the value of `MakeIdPredicateByKey`.

Also regarding your statement

> while the previous test demonstrated that when the value if true, the key field appears on the filter: $filter=CustomerId eq -10

Note that in the previous, the `MakeIdPredicateKey` is actually set to `false`, not `true` as you state. I could be wrong, but it may even be possible that it's setting both the `$filter` option and the key segment regardless of the value of the setting.

Just to be sure, I propose that @KenitoInc modify both tests such that they perform an `Assert.Equal()` instead of `Assert.Contains()`. In any case this wouldn't hurt, and it's more faithful to the intention since the current tests don't verify whether the filter or key is excluded from the url when the other is included.

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
```diff
@@ -1197,6 +1197,36 @@ public void Linq_Where_GreaterThan()
...
+        }
+
+        /// <summary>
+        /// When DataServiceContext.MakeIdPredicateByKey=true, An expression with an Id as a predicate should create a ByKey Uri
+        /// </summary>
+        [Fact]
+        public void Linq_Where_Generates_Filter_By_Id_When_MakeIdPredicateByKey_Is_True()
+        {
+            var context = this.CreateWrappedContext<DefaultContainer>().Context;
+            context.MakeIdPredicateByKey = true;
```
gathogojr, line 1222: @KenitoInc @habbes I think there's a need to rethink this, starting with the naming for this new property. Based on the current behaviour, a predicate with key field results into a ByKey query. Whatever new property we introduce, we need a value of `false` (default for boolean) to map to this current behaviour. This makes me think that the new property should be named in a way that it conveys the meaning "Id predicate should generate a filter query". In other words the reverse of the property that was added - something like `IdPredicateGeneratesFilterQueryOption` (in not such tacky form).

habbes: I think your "not so tacky" form is actually more readable and unambiguous than the shorter "MakeIdPredicateByKey". I think @KenitoInc opted to for the value to be `true` when reflecting the current behaviour and then setting it to `true` by default. In my opinion either way works so long as the default value reflects the current behaviour.

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -59,6 +62,7 @@ private ClientEdmModel Model
         private ResourceBinder(DataServiceContext context)
         {
             this.context = context;
+            idPredicateGeneratesFilterQueryOption = context.IdPredicateGeneratesFilterQueryOption;
```
odero, line 65: For consistency I'd say either use `this.<field>` or update both fields to use `_field` syntax. Option 2 is preferred of course.
Doesn't seem like `idPredicateGeneratesFilterQueryOption` is being set anywhere else other than here so I'd define it as `readonly`

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -46,6 +44,11 @@ internal class ResourceBinder : DataServiceALinqExpressionVisitor
         /// </summary>
         private readonly DataServiceContext context;
 
+        /// <summary>
+        /// /// <summary>Whether a Where clause that just compares the id property generates a $filter query option.</summary>
+        /// </summary>
```
odero, line 49:
```suggestion
        /// <summary>
        /// If true, a Where clause that compares only the key property, will generate a $filter query option.
        /// </summary>

```

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -289,7 +294,7 @@ private static Expression AnalyzePredicate(MethodCallExpression mce, ClientEdmMo
                     input.ConvertKeyToFilterExpression();
                 }
 
-                if (keyPredicates == null)
+                if (keyPredicates == null || idPredicateGeneratesFilterQueryOption)
```
odero, line 297: Why do you need to check `idPredicateGeneratesFilterQueryOption` in this case?

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -46,6 +44,11 @@ internal class ResourceBinder : DataServiceALinqExpressionVisitor
         /// </summary>
         private readonly DataServiceContext context;
 
+        /// <summary>
+        /// /// <summary>Whether a Where clause that just compares the id property generates a $filter query option.</summary>
+        /// </summary>
+        private static bool idPredicateGeneratesFilterQueryOption;
```
odero, line 50:
```suggestion
        private static bool keyPredicateGeneratesFilterQueryOption;

```

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -276,7 +280,8 @@ private static Expression AnalyzePredicate(MethodCallExpression mce, ClientEdmMo
                     keyPredicates = ExtractKeyPredicate(input, currentPredicates, model, out nonKeyPredicates);
                 }
 
-                if (keyPredicates != null)
+                // A key predicate can only be applied if IdPredicateGeneratesFilterQueryOption=false
```
odero, line 283:
```suggestion
                // A key predicate should only be applied if IdPredicateGeneratesFilterQueryOption=false

```

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -140,6 +140,9 @@ public class DataServiceContext
         /// <summary>The HTTP stack to use for requests.</summary>
         private HttpStack httpStack;
 
+        /// <summary>Whether a Where clause that just compares the id property generates a $filter query option.</summary>
+        private bool idPredicateGeneratesFilterQueryOption;
```
odero, line 144:
```suggestion
        private bool keyPredicateGeneratesFilterQueryOption;

```

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
```diff
@@ -1197,6 +1197,38 @@ public void Linq_Where_GreaterThan()
...
+        public void Linq_Where_Generates_ByKey_When_IdPredicateGeneratesFilterQueryOption_Is_False()
+        {
+            var context = this.CreateWrappedContext<DefaultContainer>().Context;
+            context.IdPredicateGeneratesFilterQueryOption = false;
+            var query = (from c in context.Customer
+                         where c.CustomerId == -10
+                         select new Customer { Name = c.Name, CustomerId = c.CustomerId }) as DataServiceQuery<Customer>;
+            var uri = query.ToString();
+            Assert.Contains("Customer(-10)", uri);
+            Assert.Equal("Customer(-10)", uri);
```
odero, line 1229: `Assert.Equal` should be adequate. Same for previous test

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -631,6 +635,14 @@ public virtual EntityTracker EntityTracker
         /// </summary>
         public virtual bool EnableWritingODataAnnotationWithoutPrefix { get; set; }
 
+        /// <summary>
+        /// Indicates whether a Where clause that just compares the id property generates a $filter query option.
+        /// </summary>
+        public virtual bool IdPredicateGeneratesFilterQueryOption
```
odero, line 641: Devs don't necessarily have any knowledge of what a "predicate" is wrt DSC - such vocabulary seems internal to the DSC implementation.
I'd consider renaming this to something like
`GenerateKeyCheckAsFilterQuery` or `GenerateKeyComparisonAsFilterQuery`

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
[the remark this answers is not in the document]
odero: And why not use the `context.Customer.Where()` syntax in this case?

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: I have switched to `KeyComparisonGeneratesFilterQuery`

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
KenitoInc: Fixed

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
KenitoInc: Fixed

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: Fixed

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
KenitoInc: Fixed

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
KenitoInc: In both conditions, we need to convert Key to filter expression, incase there was one.

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
KenitoInc: Fixed

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
```diff
@@ -1197,6 +1197,38 @@ public void Linq_Where_GreaterThan()
...
         }
 
+        /// <summary>
+        /// When DataServiceContext.IdPredicateGeneratesFilterQueryOption=true, An expression with an Id as a predicate should create an Uri with a $filter in query options
+        /// </summary>
+        [Fact]
+        public void Linq_Where_Generates_Filter_When_IdPredicateGeneratesFilterQueryOption_Is_True()
+        {
+            var context = this.CreateWrappedContext<DefaultContainer>().Context;
+            context.IdPredicateGeneratesFilterQueryOption = true;
```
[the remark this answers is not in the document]
KenitoInc: Actually it should be false by default. I have changed in the description and updated the relevant test.

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -46,6 +44,11 @@ internal class ResourceBinder : DataServiceALinqExpressionVisitor
         /// </summary>
         private readonly DataServiceContext context;
 
+        /// <summary>
+        /// If true, a Where clause that compares only the key property, will generate a $filter query option.
+        /// </summary>
+        private static bool keyComparisonGeneratesFilterQuery;
```
habbes, line 50: Why is this field static?

KenitoInc: @habbes It's being used within a static method

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
KenitoInc: Their usage is different:
`keyComparisonGeneratesFilterQuery` is static since it's used within a static method, hence cannot be accessed using an instance reference.

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
mikepizzo, line 50: Defining a static property on a non-static seems sketchy. Maybe just add context as a parameter to the AnalyzePredicate method and remove the property on the binder?

mikepizzo:
>                 if (!input.UseFilterAsPredicate)

trying to understand how UseFilterAsPredicate relates to the new keyComparisonGeneratesFilterQuery.  Do they do the same thing, but are set in different places? In what cases/why do we set UseFilterAsPredicate today? Would we have the same logic if line 278 was !input.UseFilterAsPredicate && !keyComparisonGeneratesFilterQuery, and removed the additional checks below?

---
Refers to: src/Microsoft.OData.Client/ALinq/ResourceBinder.cs:278 in 8368f37. [](commit_id = 8368f37d99c6a855abb374551d67a2503cf62982, deletion_comment = False)

mikepizzo:
@ - See a couple comments. If we can get those resolved I'd love to merge. Also, please mark other comments as resolved as appropriate (i.e., in codeflow) to make it easier to track status of comments. Thanks!

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
KenitoInc: #Resolved

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
KenitoInc: #Resolved

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -251,6 +254,7 @@ internal DataServiceContext(Uri serviceRoot, ODataProtocolVersion maxProtocolVer
             this.httpStack = HttpStack.Auto;
             this.UsingDataServiceCollection = false;
             this.UsePostTunneling = false;
+            this.idPredicateGeneratesFilterQueryOption = false;
 
             // Need to use the same defaults when running sl in portable lib as when running in SL normally.
 #if PORTABLELIB
```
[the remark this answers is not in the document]
KenitoInc: It's affecting the enclosed code. We can discuss and remove them in future PRs.

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
KenitoInc: Resolved by passing the context in the `AnalyzePredicate` method

test/EndToEndTests/Microsoft.OData.E2E.Portable/AsynchronousTests/AsynchronousQueryTests.cs
KenitoInc: Used `Assert.EndsWith`

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
KenitoInc: Resolved by passing the context in the `AnalyzePredicate` method

KenitoInc:
> > ```
> >             if (!input.UseFilterAsPredicate)
> > ```
> 
> trying to understand how UseFilterAsPredicate relates to the new keyComparisonGeneratesFilterQuery. Do they do the same thing, but are set in different places? In what cases/why do we set UseFilterAsPredicate today? Would we have the same logic if line 278 was !input.UseFilterAsPredicate && !keyComparisonGeneratesFilterQuery, and removed the additional checks below?
> 
> Refers to: src/Microsoft.OData.Client/ALinq/ResourceBinder.cs:278 in [8368f37](https://github.com/OData/odata.net/commit/8368f37d99c6a855abb374551d67a2503cf62982). [](commit_id = [8368f37](https://github.com/OData/odata.net/commit/8368f37d99c6a855abb374551d67a2503cf62982), deletion_comment = False)

@mikepizzo 
`UseFilterAsPredicate` is used create a $filter query option when we have both key and non-key predicates in the expression.
However in scenarios where we have only a key predicate in the expression, we generate a Uri with a ByKey resource path. `KeyComparisonGeneratesFilterQuery` allows us to specify if we want a $filter or ByKey.

We can clean up the logic so as to work well with both.

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
KenitoInc: #Fixed

mikepizzo:
It would be, which is why we have introduced a property on the DataServiceContext to control it.

---
In reply to: [406936949](https://github.com/OData/odata.net/pull/1762#issuecomment-406936949) [](ancestors = 406936949)

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -253,6 +256,7 @@ internal DataServiceContext(Uri serviceRoot, ODataProtocolVersion maxProtocolVer
             this.httpStack = HttpStack.Auto;
             this.UsingDataServiceCollection = false;
             this.UsePostTunneling = false;
+            this.keyComparisonGeneratesFilterQuery = false;
```
mikepizzo, line 259: >false [](start = 53, length = 5)

Should this initially be true (for backward compatibility) and then set to false (or removed) in our next breaking change release (8.0)?

PR #1763, opened 2020-05-06
Fix #1640 - Set IEEE754Compatible parameter from the batch request

### LINKED ISSUE #1640: ODataJsonLightBatchBodyContentReaderStream assumes that batch part body is IEEE754Compatible
When sending a batch request that includes PATCH on entity with decimal property, the oData .Net lib automatically assumes that body of request item in in IEEE 754 compatible format with regards to actual item `Content-Type` header.

### Assemblies affected

OData .Net lib 7.6.1

### Reproduce steps

https://github.com/xsmerek/BatchDecimalRepro

Steps:

1. Run IIS Express.
2. Send POST request to `http://localhost:52169/odata/$batch`
with header `Content-Type = application/json`
and body 
```
{
    "requests": [
        {
            "url":"/odata/Examples(1)",
            "method":"PATCH",
            "headers":{"Content-Type":"application/json"},
            "body":{
                "Price": 1.3
            },
            "id":"1",
            "atomicityGroup": "g1"
        }
    ]
}
```

Please note that the app uses Microsoft.OData.Core indirectly via Microsoft.AspNetCore.OData but the issue is caused by this library (please see Additional detail for explanation).

### Expected result

Controller should return success HTTP code.

### Actual result

`Microsoft.OData.ODataException: 'Cannot convert a value to target type 'Edm.Decimal' because of conflict between input format string/number and parameter 'IEEE754Compatible' false/true.'`

is thrown in `ODataJsonLightPropertyAndValueDeserializer`, method `ReadPrimitiveValue`.

### Additional detail

The root cause is most likely in [ODataJsonLightBatchBodyContentReaderStream.cs, line 254](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchBodyContentReaderStream.cs#L254) where `jsonWriter` is created with `isIeee754Compatible=true` without regards for actual `Content-Type` header of batch item, hence, item header and body are incompatible.

Please note that if we set `Content-Type` header to `application/json;IEEE754Compatible=true`, the request is successful (even if the item body is not compatible with IEEE 754).

Also, if I send PATCH request with the given body, the request is successful.

paulodero:
### Issues

*This pull request fixes #1640 .*

### Description

*Set IEEE754Compatible parameter from the batch request.*

- [ ] *Test cases added*

### Additional work necessary

*ODataJsonLightBatchBodyContentReaderStream assumes that batch part body is IEEE754Compatible. This pull request ensures IEEE754Compatible parameter is set from the batch request sent from the client*

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchBodyContentReaderStream.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchBodyContentReaderStream.cs
@@ -253,7 +253,7 @@ namespace Microsoft.OData.JsonLight
             // Reader is on the value node after the "body" property name node.
             IJsonWriter jsonWriter = new JsonWriter(
                 new StreamWriter(this),
-                true /*isIeee754Compatible*/);
+                reader.IsIeee754Compatible);
 
             WriteCurrentJsonObject(reader, jsonWriter);
             this.Flush();


xuzhg:
@paulodero Can you make the title more meaningful?

paulodero:
> @paulodero Can you make the title more meaningful?

Done.

KanishManuja-MS:
@paulodero Can you please re-trigger/rebase the build for this? The current rolling build status is not upto date?

Issue #1604, opened 2019-12-04
Async functionality in ODL broken


mikepizzo:
There are a few issues in ODL wrt async support:
1) Apparent bug in AsyncBufferedStream; calls FlushSync() on asyncstream, rather than FlushAsync
2) Async functions in ODataWriterCore that call implementations that call writer.WriteEnd() are broken.
	-WriteDeltaLink, WriteDeltaDeletedLink, WritePrimitive
3) ODataMessageWriter.CreateODataResourceSetWriterAsync and CreateODataDeltaResourceSetWriterAsync should take an IEdmStructuredType, not IEdmEntityType

### Assemblies affected

OData.NET 7.6.1

PR #1770, opened 2020-05-08
Fix how composite async methods invoke lower level methods.

mikepizzo:
### Issues

*This pull request fixes issue #1604.*
*This pull request fixes issue #1776.*

### Description

*Some of the async fluent APIs incorrectly call lower level APIs.*

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/ODataMessageWriter.cs
+++ b/src/Microsoft.OData.Core/ODataMessageWriter.cs
@@ -242,14 +242,14 @@ namespace Microsoft.OData
         /// Asynchronously creates an <see cref="ODataWriter" /> to write a resource set.
         /// </summary>
         /// <param name="entitySet">The entity set we are going to write entities for.</param>
-        /// <param name="entityType">The entity type for the entries in the resource set to be written (or null if the entity set base type should be used).</param>
+        /// <param name="structuredType">The structured type for the entries in the resource set to be written (or null if the entity set base type should be used).</param>
         /// <returns>A running task for the created writer.</returns>
-        public Task<ODataWriter> CreateODataResourceSetWriterAsync(IEdmEntitySetBase entitySet, IEdmEntityType entityType)
+        public Task<ODataWriter> CreateODataResourceSetWriterAsync(IEdmEntitySetBase entitySet, IEdmStructuredType structuredType)
         {
             this.VerifyCanCreateODataResourceSetWriter();
             return this.WriteToOutputAsync(
                 ODataPayloadKind.ResourceSet,
-                (context) => context.CreateODataResourceSetWriterAsync(entitySet, entityType));
+                (context) => context.CreateODataResourceSetWriterAsync(entitySet, structuredType));
         }
 
         /// <summary> Creates an <see cref="T:Microsoft.OData.ODataWriter" /> to write a delta resource set. </summary>
@@ -306,14 +306,14 @@ namespace Microsoft.OData
         /// Asynchronously creates an <see cref="ODataWriter" /> to write a delta resource set.
         /// </summary>
         /// <param name="entitySet">The entity set we are going to write entities for.</param>
-        /// <param name="entityType">The entity type for the entries in the resource set to be written (or null if the entity set base type should be used).</param>
+        /// <param name="structuredType">The structured type for the entries in the resource set to be written (or null if the entity set base type should be used).</param>
         /// <returns>A running task for the created writer.</returns>
-        public Task<ODataWriter> CreateODataDeltaResourceSetWriterAsync(IEdmEntitySetBase entitySet, IEdmEntityType entityType)
+        public Task<ODataWriter> CreateODataDeltaResourceSetWriterAsync(IEdmEntitySetBase entitySet, IEdmStructuredType structuredType)
         {
             this.VerifyCanCreateODataResourceSetWriter();
             return this.WriteToOutputAsync(
                 ODataPayloadKind.ResourceSet,
-                (context) => context.CreateODataDeltaResourceSetWriterAsync(entitySet, entityType));
+                (context) => context.CreateODataDeltaResourceSetWriterAsync(entitySet, structuredType));
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.OData.Core/ODataMessageWriter.cs
```diff
@@ -306,14 +306,14 @@ public Task<ODataWriter> CreateODataDeltaResourceSetWriterAsync(IEdmEntitySetBas
         /// Asynchronously creates an <see cref="ODataWriter" /> to write a delta resource set.
         /// </summary>
         /// <param name="entitySet">The entity set we are going to write entities for.</param>
-        /// <param name="entityType">The entity type for the entries in the resource set to be written (or null if the entity set base type should be used).</param>
+        /// <param name="structuredType">The structured type for the entries in the resource set to be written (or null if the entity set base type should be used).</param>
```
xuzhg, line 309: >entity [](start = 127, length = 6)

also change to "resource set"

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Writer/JsonLight/ODataJsonLightInheritComplexCollectionWriterTests.cs
```diff
@@ -11,6 +11,9 @@
 using Microsoft.OData.JsonLight;
 using Microsoft.OData.Edm;
 using Xunit;
+using System.Globalization;
+using System.ComponentModel;
+using System.Linq;
```
xuzhg, line 16: sorting it?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Writer/JsonLight/ODataJsonLightInheritComplexCollectionWriterTests.cs
```diff
@@ -221,23 +226,83 @@ private static void WriteAndValidateSync(IEdmTypeReference itemTypeReference, OD
...
+            {
+                odataWriter.WriteStartAsync(item).Wait();
+                odataWriter.WriteEndAsync().Wait();
+            }
+
+            odataWriter.WriteEndAsync().Wait();
+            ValidateWrittenPayload(stream, expectedPayload);
+        }
+
+        private static void WriteAndValidatePrimitivesSync(IEdmTypeReference itemTypeReference, ODataResourceSet collectionStart, IEnumerable<ODataResource> items, string expectedPayload, bool writingResponse)
```
xuzhg, line 247: >Sync [](start = 54, length = 4)

shall we use 'Sync" suffix ?

xuzhg, line 247: >collectionStart [](start = 113, length = 15)

resourceSet ?

src/Microsoft.OData.Core/ODataWriter.cs
```diff
@@ -324,7 +324,7 @@ public ODataWriter Write(ODataProperty primitiveProperty, Action nestedAction)
         /// <summary> Asynchronously write a primitive property within a resource. </summary>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
         /// <param name="primitiveProperty">The primitive property to write.</param>
-        public virtual Task WriteStartAsync(ODataProperty primitiveProperty)
+        public virtual Task WriteStartAsync(ODataPropertyInfo primitiveProperty)
```
xuzhg, line 327: >ODataPropertyInfo [](start = 44, length = 17)

ODataPropertyInfo is the base class of ODataProperty, So it's not a breaking change, right?

How about other Write(...) method?  For example: Line 315?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Writer/JsonLight/ODataJsonLightInheritComplexCollectionWriterTests.cs
mikepizzo: Actually, those are unnecessary (not sure how they got added) so I removed them -- thanks for the catch!

---
In reply to: [427635169](https://github.com/OData/odata.net/pull/1770#discussion_r427635169) [](ancestors = 427635169)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Writer/JsonLight/ODataJsonLightInheritComplexCollectionWriterTests.cs
mikepizzo: The "Sync" suffix was there previously for "WriteAndValidateSync" and "WriteAndValidateAsync" -- I just duplicated the pattern for WriteAndValidatePrimitive..

---
In reply to: [427635649](https://github.com/OData/odata.net/pull/1770#discussion_r427635649) [](ancestors = 427635649)

mikepizzo: I just copied the pattern that was there before. The arguments that get passed in are named "collectionStartWithSerializationInfo" and "collectionStartWithoutSerializationInfo", which I think is why this parameter was originally named "collectionStart".

---
In reply to: [427635707](https://github.com/OData/odata.net/pull/1770#discussion_r427635707) [](ancestors = 427635707)

src/Microsoft.OData.Core/ODataMessageWriter.cs
mikepizzo: actually, I reverted this change. The delta resource set always contains entity types, so it's correct to say entity set and entitytype.. Should probably also change CreateODataDeltaResourceSetWriter on the context in 8.0, but it would be breaking to do so in a 7.x release.

---
In reply to: [427634277](https://github.com/OData/odata.net/pull/1770#discussion_r427634277) [](ancestors = 427634277)

src/Microsoft.OData.Core/ODataWriter.cs
mikepizzo: Right.

---
In reply to: [427636669](https://github.com/OData/odata.net/pull/1770#discussion_r427636669) [](ancestors = 427636669)

mikepizzo: In 315, we are writing the entire property (name and value) so it needs the property, not just a property info.

---
In reply to: [428232034](https://github.com/OData/odata.net/pull/1770#discussion_r428232034) [](ancestors = 428232034,427636669)

Issue #1715, opened 2020-03-24
Null validation logic differentiates between collection property with primitive values and collection property with complex values


t-bzhan:
I've observed that there are some difference between the null validation logic for collection property with primitive values and collection property with complex values. While a JSON request payload is received with the collection property explicitly set to 'null', the behavior is not unified.

For collection property with primitive values, it will complain:

>  A null value was found for the property named 'XXX', which has the expected type 'Collection(Edm.String)[Nullable=False]'. The expected type 'Collection(Edm.String)[Nullable=False]' does not allow null values.

While for collection property with complex values, it will complain:

> A node of type 'PrimitiveValue' was read from the JSON reader when trying to read the contents of the property 'XXX'; however, a 'StartArray' node was expected

### Assemblies affected
Microsoft.AspNetCore.OData 7.3.0

### Reproduce steps
Using below code snippet:
```
using System.Collections.Generic;
using System.IO;
using System.Text;
using Microsoft.AspNet.OData.Builder;
using Microsoft.AspNet.OData.Formatter.Deserialization;
using Microsoft.OData;
using Microsoft.OData.Edm;

namespace ODataTest
{
    public class Org
    {
        public List<Policy> Policies { get; set; }
    }

    public class Policy
    {
        public int Order { get; set; }

        public string Describtion { get; set; }
    }

    public class Org1
    {
        public List<string> Members { get; set; }
    }

    class Program
    {
        static void TestCollectionComplex()
        {
            ODataModelBuilder builder = new ODataModelBuilder();
            ComplexTypeConfiguration<Org> orgType = builder.ComplexType<Org>();
            orgType.CollectionProperty<Policy>(x => x.Policies);

            IEdmModel model = builder.GetEdmModel();

            Stream stream = new MemoryStream(Encoding.UTF8.GetBytes("{" +
            "  \"Policies\": null" +
            "}"));

            IODataRequestMessage requestMessage = new InMemoryMessage() { Stream = stream };
            requestMessage.SetHeader("Content-Type", "application/json");
            ODataMessageReader reader = new ODataMessageReader(
                                                requestMessage,
                                                new ODataMessageReaderSettings()
                                                {
                                                    ReadUntypedAsString = true,
                                                    Validations = ValidationKinds.All
                                                },
                                                model);

            var odataReader = reader.CreateODataResourceReader(model.FindDeclaredType("ODataTest.Org") as IEdmStructuredType);
[...]

PR #1793, opened 2020-05-28
different null validation messages for complex and primitive collections

ElizabethOkerio:
### Issues

*This pull request fixes issue #1715.*

### Description

The error message returned by the null validation of a collection of primitive values and a collection of complex values is different. The expected result is for the error message for the null validation of both collections to be the same. 

A null collection of primitive values returns this error message: 
<pre>
A null value was found for the property named 'XXX', 
which has the expected type 'Collection(Edm.String)[Nullable=False]'.
The expected type 'Collection(Edm.String)[Nullable=False]' 
does not allow null values.
</pre>

while a null collections of complex values returns: 
<pre>
A node of type 'PrimitiveValue' was read from the JSON reader
when trying to read the contents of the property 'XXX';
however, a 'StartArray' node was expected
</pre>

The expected result is for the error message for both collections to be: 
<pre>
A null value was found for the property named 'XXX', 
which has the expected type 'Collection(XXX)[Nullable=False]'. 
The expected type 'Collection(XXX)[Nullable=False]' does not allow null values
</pre>

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
@@ -422,7 +422,7 @@ namespace Microsoft.OData.JsonLight
 
                 // Complex property or collection of complex property.
                 bool isCollection = payloadTypeReference.IsCollection();
-                ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, isCollection, propertyName);
+                ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, isCollection, propertyName, payloadTypeReference);
                 if (isCollection)
                 {
                     readerNestedResourceInfo = this.ReadingResponse
@@ -473,7 +473,8 @@ namespace Microsoft.OData.JsonLight
         /// <param name="jsonReader">The IJsonReader.</param>
         /// <param name="isCollection">true if the property is entity set reference property; false for a resource reference property, null if unknown.</param>
         /// <param name="propertyName">Name for the navigation property, used in error message only.</param>
-        protected static void ValidateExpandedNestedResourceInfoPropertyValue(IJsonReader jsonReader, bool? isCollection, string propertyName)
+        /// <param name="typeReference">Type of navigation property</param>
+        protected static void ValidateExpandedNestedResourceInfoPropertyValue(IJsonReader jsonReader, bool? isCollection, string propertyName, IEdmTypeReference typeReference)
         {
             // an expanded link with resource requires a StartObject node here;
             // an expanded link with resource set requires a StartArray node here;
@@ -491,7 +492,16 @@ namespace Microsoft.OData.JsonLight
                 // Expanded resource (null or non-null)
                 if (isCollection == true)
                 {
-                    throw new ODataException(ODataErrorStrings.ODataJsonLightResourceDeserializer_CannotReadCollectionNestedResource(nodeType, propertyName));
+                    if (typeReference.IsNonEntityCollectionType())
+                    {
+                        ReaderValidationUtils.ValidateNullValue(typeReference, true,
+                                                  true, propertyName, false);
+                    }
+                    else
+                    {
+                        throw new ODataException(ODataErrorStrings.ODataJsonLightResourceDeserializer_CannotReadCollectionNestedResource(nodeType, propertyName));
+                    }
+                    
                 }
             }
             else
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
@@ -1232,7 +1232,7 @@ namespace Microsoft.OData.JsonLight
                     ODataJsonLightReaderNestedResourceInfo readerNestedResourceInfo = null;
 
                     // Complex property or collection of complex property.
-                    ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, isCollection, propertyName);
+                    ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, isCollection, propertyName, edmProperty.Type);
 
                     if (isCollection)
                     {
@@ -1251,7 +1251,7 @@ namespace Microsoft.OData.JsonLight
                     ODataJsonLightReaderNestedResourceInfo readerNestedResourceInfo = null;
 
                     // Expanded link
-                    ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, isCollection, propertyName);
+                    ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, isCollection, propertyName, edmProperty.Type);
                     if (isCollection)
                     {
                         readerNestedResourceInfo = this.ReadingResponse || isDeltaResourceSet
@@ -1548,7 +1548,7 @@ namespace Microsoft.OData.JsonLight
             if (payloadTypeOrItemType != null)
             {
                 // Complex property or collection of complex property.
-                ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, isCollection, propertyName);
+                ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, isCollection, propertyName, payloadTypeReference);
                 if (isCollection)
                 {
                     return ReadNonExpandedResourceSetNestedResourceInfo(resourceState, null, payloadTypeOrItemType, propertyName);
@@ -1668,7 +1668,7 @@ namespace Microsoft.OData.JsonLight
                 // If the property is expanded, ignore the content if we're asked to do so.
                 if (propertyWithValue)
                 {
-                    ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, null, propertyName);
+                    ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, null, propertyName, resourceState.ResourceType.ToTypeReference());
 
                     // Since we marked the nested resource info as deferred the reader will not try to read its content
                     // instead it will behave as if it was a real deferred link (without a property value).
@@ -1679,7 +1679,6 @@ namespace Microsoft.OData.JsonLight
 
                 return navigationLinkInfo;
             }
-
 
             if (resourceState.ResourceType.IsOpen)
             {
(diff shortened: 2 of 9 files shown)


src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -491,7 +494,21 @@ protected static void ValidateExpandedNestedResourceInfoPropertyValue(IJsonReade
...
                 {
-                    throw new ODataException(ODataErrorStrings.ODataJsonLightResourceDeserializer_CannotReadCollectionNestedResource(nodeType, propertyName));
+                    if (typeReference.IsODataPrimitiveTypeKind() ||
+                        typeReference.IsODataTypeDefinitionTypeKind() ||
+                        typeReference.IsODataEnumTypeKind() ||
+                        typeReference.IsODataComplexTypeKind() ||
+                        typeReference.IsUntyped() ||
+                        typeReference.IsNonEntityCollectionType())
+                    {
+                        ReaderValidationUtils.ValidateNullValue(typeReference, true,
```
odero, line 504: Any reason to use `ValidateNullValue` directly via `ReaderValidationUtils` instead of using it via `ReaderValidator` which binds to `ODataMessageReaderSettings`?
`ReaderValidator` uses settings from `ODataMessageReaderSettings` to determine the value for `enablePrimitiveTypeConversion`, which in this case you are ignoring and forcing to always be true

ElizabethOkerio: Have used the <code> ValidateNullValue </code> directly via <code> ReaderValidationUtils </code> because at this time I know am validating a collection of a complex type and not of a primitive type. It really does not matter the value that I pass for the 
<code> enablePrimitiveTypeConversion </code> at this point. It can as well be false. it won't have any effect on the method.

test/FunctionalTests/Microsoft.OData.Core.Tests/NavigationPropertyOnComplexTests.cs
```diff
@@ -1652,5 +1652,27 @@ private static IEdmModel ContainmentComplexModel()
...
+        [Fact]
+        public void NullValidationErrorMessageForCollectionsofComplexTypes()
+        {
+            string expectedErrorMessage = "A null value was found for the property named 'Addresses', which has the expected type 'Collection(DefaultNs.Address)[Nullable=False]'. The expected type 'Collection(DefaultNs.Address)[Nullable=False]' does not allow null values.";
+            string actualErrorMessage = "";
+            const string payload =
+                "{" +
+                    "\"@odata.context\":\"http://host/$metadata#People/$entity\"," +
+                    "\"UserName\":\"abc\"," +
+                    "\"Addresses\":null" +
```
xuzhg, line 1664: Nullable at collection is for the items in the collection. Collection itself is not nullalble. 
So, your test case is not correct.
It can be :  "Addresses": [ ]   // empty collection, but can't be null
or It can be: "Addresses": [ null  ]  // non empty collection with null item.

@mikepizzo need mike's input.

mikepizzo: As Sam says, nullable on a collection means it can *contain* a null value, not that it can *be* a null value. 

This test *correctly* validates that the collection value cannot be null, and the same should be true if Addresses is defined as a collection with nullable=true. I don't know if we have such a test, and it would probably be a breaking change to make that an error if it's not today, but that would be the expected behavior.

ElizabethOkerio: @mikepizzo there was no such test. Have added the test though. Please review.

test/FunctionalTests/Microsoft.OData.Core.Tests/NavigationPropertyOnComplexTests.cs
```diff
@@ -1674,5 +1675,28 @@ public void NullValidationErrorMessageForCollectionsofComplexTypes()
             }
             Assert.Equal(expectedErrorMessage, actualErrorMessage);
         }
+
+        [Fact]
+        public void NullValidationErrorMessageForANullableCollectionsofComplexTypes()
+        {
+            string expectedErrorMessage = "A null value was found for the property named 'AddressesNullable', which has the expected type 'Collection(DefaultNs.Address)[Nullable=True]'. The expected type 'Collection(DefaultNs.Address)[Nullable=True]' does not allow null values.";
```
odero, line 1682:
```suggestion
            string expectedErrorMessage = "A null value was found for the property named 'AddressesNullable', which has the expected type 'Collection(DefaultNs.Address)[Nullable=True]'. For an empty collection of type 'Collection(DefaultNs.Address)[Nullable=True]', set the value to '[]' instead of null.";

```
Could you change this message? It doesn't make itclear why a field that is marked nullable cannot not be set to null.

ElizabethOkerio: Noted.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -491,7 +492,21 @@ protected static void ValidateExpandedNestedResourceInfoPropertyValue(IJsonReade
                 // Expanded resource (null or non-null)
                 if (isCollection == true)
                 {
-                    throw new ODataException(ODataErrorStrings.ODataJsonLightResourceDeserializer_CannotReadCollectionNestedResource(nodeType, propertyName));
+                    if (typeReference.IsODataPrimitiveTypeKind() ||
+                        typeReference.IsODataTypeDefinitionTypeKind() ||
+                        typeReference.IsODataEnumTypeKind() ||
+                        typeReference.IsODataComplexTypeKind() ||
+                        typeReference.IsUntyped() ||
+                        typeReference.IsNonEntityCollectionType())
```
paulodero, line 500: All these is either a structured type or just an edm type definition. Any chance the checks could be compressed?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -1667,8 +1667,8 @@ private ODataJsonLightReaderNestedInfo ReadUndeclaredProperty(IODataJsonLightRea
 
                 // If the property is expanded, ignore the content if we're asked to do so.
                 if (propertyWithValue)
-                {
-                    ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, null, propertyName);
+                {
```
KenitoInc, line 1670: There is an unnecessary spacing

src/Microsoft.OData.Core/Microsoft.OData.Core.cs
```diff
@@ -18,7 +18,7 @@ namespace Microsoft.OData
     /// <summary>
     ///    AutoGenerated resource class. Usage:
     ///
-    ///        string s = TextRes.GetString(TextRes.MyIdentifier);
+    ///        string s = TextRes.GetString(TextRes.MyIdenfitier);
```
KenitoInc, line 21: Why does this look like a typo?
`MyIdentifier` looks like the logical spelling

src/Microsoft.OData.Core/Parameterized.Microsoft.OData.Core.cs
```diff
@@ -1801,13 +1801,27 @@ internal static class Strings {
             return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.ReaderValidationUtils_NullValueForNonNullableType, p0);
         }
 
+        /// <summary>
+        /// A string like "A null value was found with the expected type '{0}[Nullable=True]'. The expected type '{0}[Nullable=True]' does not allow null values."
+        /// </summary>
+        internal static string ReaderValidationUtils_NullValueForNullableType(object p0) {
+            return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.ReaderValidationUtils_NullValueForNullableType, p0);
+        }
+
```
xuzhg, line 1810: should rebase to the master, because this file has been changed using different format

src/Microsoft.OData.Core/Parameterized.Microsoft.OData.Core.cs
```diff
@@ -2147,6 +2147,13 @@ internal static string ReaderValidationUtils_NullValueForNonNullableType(object
             return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.ReaderValidationUtils_NullValueForNonNullableType, p0);
         }
 
+        /// <summary>
+        /// A string like "A null value was found with the expected type '{0}[Nullable=True]'. The expected type '{0}[Nullable=True]' does not allow null values."
```
mikepizzo, line 2151: I was trying to understand the difference between this and ReaderValidationUtils_NullValueForNonNullableType, but the only difference is whether or not the type is nullable (and it wasn't clear from this message why a nullable type wouldn't allow null values). please update to match the string in Microsoft.OData.Core.txt.

mikepizzo, line 2151: >The expected type '{0}[Nullable=True]' does not allow null values. [](start = 95, length = 66)

This message is confusing to me.  Is it used exclusively when a collection-valued property is null? 

Can we make it something like "A null value was found for a collection of type '{0}[Nullable=True]. Collection-valued properties with Nullable=True can contain null values, but collection-valued properties cannot themselves be null."

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
mikepizzo: +1 

what is the semantic you're checking for here? is there an easier way to check?

---
In reply to: [439973000](https://github.com/OData/odata.net/pull/1793#discussion_r439973000) [](ancestors = 439973000)

src/Microsoft.OData.Core/Parameterized.Microsoft.OData.Core.cs
ElizabethOkerio: resolved

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
ElizabethOkerio: resolved

src/Microsoft.OData.Core/Microsoft.OData.Core.cs
mikepizzo: +1

---
In reply to: [441495806](https://github.com/OData/odata.net/pull/1793#discussion_r441495806) [](ancestors = 441495806)

test/FunctionalTests/Microsoft.OData.Core.Tests/NavigationPropertyOnComplexTests.cs
mikepizzo: :thumbs up:

---
In reply to: [439432201](https://github.com/OData/odata.net/pull/1793#discussion_r439432201) [](ancestors = 439432201)

odero (approved):

KenitoInc (approved):
LGTM

Issue #1753, opened 2020-04-22
Strange Timeout behavior when requesting entities from D365


obutov:
When certain D365 entities take longer than 5 minutes to return, the OData client fails to receive any response. 

### Assemblies affected

"Microsoft.OData.Client" version="7.6.2"

### Reproduce steps

Generate OData client proxy by consuming D365 edmx file. 
Set context timeout to 10 minutes
Request a slow-performing entity from D365 (e.g. entity that takes longer than 5 minutes to return the response). For example, _context.CustomersV2.Execute();
The response never arrives and request times out after 10 minutes.
Set timeout to 30 minutes or 60 minutes - the result would be the same. 
### Expected result

The response should be received and processed by OData client, even if it takes more than 5 minutes for D365 to send the response. 

### Actual result

OData client seems to fail to properly receive and process any long-running responses (that take longer than 5 minutes).

### Additional detail

N/A

marabooy:
Hi could you try and use fiddler or a trace client to see what is happening? I suspect that the request could be closed before the response comes in.

obutov:
I used fiddler to trace the request, but could not see any response. The request is sent to D365, but then it just "sits" there. Even when the client throws the context timeout, nothing happens on the fiddler side. When I send a request using Postman, the response arrives after 8-9 minutes. When I send it using OData client, nothing happens. It almost seems like there is some kind of "internal" timeout that doesn't bubble up properly, nor can it be adjusted. Just a wild speculation at this point.

ElizabethOkerio:
@obutov The <code>DataServiceContext </code> class in <code>Microsoft.OData.Client </code> library has a Timeout property that you can set before you make any request with <code>Microsoft.OData.Client </code>. You can set the timeout value to whichever value you want to ensure that the request does not time out before a response is returned. Normally, if you do not set this value before making a request, the request uses the default timeout value which is 100 seconds. So if you are making a request that will take more than a 100 seconds to return a value then it is important that you set the timeout value like below: 
<pre>
DefaultContainer context = new DefaultContainer(new Uri("https://services.odata.org/v4/TripPinServiceRW/"));
            context.Timeout = 1200000;
var cust = context.Customers.FirstOrDefault();

</pre>

ElizabethOkerio:
Closing this issue. Feel free to reopen it if this does not solve your issue. Thanks.

obutov:
[...]

PR #1796, opened 2020-05-29
update read write timeout value when timeout value is set

ElizabethOkerio:
### Issues

*This pull request fixes issue #1753.*

### Description
Currently, It is not possible to set the ReadWriteTimeout property of a request in <code> Microsoft.OData.Client </code> but you can set the Timeout property. This means that each request uses the default <code>HttpWebRequest ReadWriteTimeout </code> property which is 300seconds or 5 minutes. 

If an endpoint takes more than 5 minutes to return a response to <code> Microsoft.OData.Client </code> that response will not be received and processed by <code> Microsoft.OData.Client</code>. This is because the ReadWriteTimeout property times out after 5 mins hence the ReadWrite connection times out. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -112,10 +112,11 @@ namespace Microsoft.OData.Client
         /// <summary>resolve typename from a type</summary>
         private Func<string, Type> resolveType;
 
-#if !PORTABLELIB // Timeout not available
         /// <summary>time-out value in seconds, 0 for default</summary>
         private int timeout;
-#endif
+
+        /// <summary>read or write time-out value in seconds, 0 for default</summary>
+        private int readWriteTimeout;
         /// <summary>whether to use post-tunneling for PUT/DELETE</summary>
         private bool postTunneling;
 
@@ -480,7 +481,6 @@ namespace Microsoft.OData.Client
             set { this.resolveType = value; }
         }
 
-#if !PORTABLELIB // Timeout not available
         /// <summary>Gets or sets the time-out option (in seconds) that is used for the underlying HTTP request to the data service.</summary>
         /// <returns>An integer that indicates the time interval (in seconds) before time-out of a service request.</returns>
         /// <remarks>
@@ -507,7 +507,33 @@ namespace Microsoft.OData.Client
                 this.timeout = value;
             }
         }
-#endif
+
+        /// <summary>Gets or sets the readwrite time-out option (in seconds) that is used for the underlying HTTP request to the data service.</summary>
+        /// <returns>An integer that indicates the time interval (in seconds) before readwritetime-out of a service request.</returns>
+        /// <remarks>
+        /// A value of 0 will use the default readwritetimeout of the underlying HTTP request.
+        /// This value must be set before executing any query or update operations against
+        /// the target data service for it to have effect on the on the request.
+        /// The value may be changed between requests to a data service and the new value
+        /// will be picked up by the next data service request.
+        /// </remarks>
+        public virtual int ReadWriteTimeout
+        {
+            get
+            {
+                return this.readWriteTimeout;
+            }
+
+            set
+            {
+                if (value < 0)
+                {
+                    throw Error.ArgumentOutOfRange("ReadWriteTimeout");
+                }
+
+                this.readWriteTimeout = value;
+            }
+        }
 
         /// <summary>Gets or sets a Boolean value that indicates whether to use post tunneling.</summary>
         /// <returns>A Boolean value that indicates whether to use post tunneling.</returns>
--- a/src/Microsoft.OData.Client/InternalODataRequestMessage.cs
+++ b/src/Microsoft.OData.Client/InternalODataRequestMessage.cs
@@ -105,11 +105,19 @@ namespace Microsoft.OData.Client
             set { throw new NotSupportedException(); }
         }
 
-#if !PORTABLELIB
         /// <summary>
         /// Gets or sets the timeout (in seconds) for this request.
         /// </summary>
         public override int Timeout
+        {
+            get { throw new NotSupportedException(); }
+            set { throw new NotSupportedException(); }
+        }
+
+        /// <summary>
+        /// Gets or sets the read and write timeout (in seconds) for this request.
+        /// </summary>
+        public override int ReadWriteTimeout
         {
             get { throw new NotSupportedException(); }
             set { throw new NotSupportedException(); }
@@ -131,7 +139,6 @@ namespace Microsoft.OData.Client
                 throw new NotImplementedException();
             }
         }
-#endif
 
         /// <summary>
         /// internal headers dictionary
--- a/src/Microsoft.OData.Client/ODataRequestMessageWrapper.cs
+++ b/src/Microsoft.OData.Client/ODataRequestMessageWrapper.cs
@@ -157,12 +157,16 @@ namespace Microsoft.OData.Client
                 requestMessage.Credentials = requestInfo.Credentials;
             }
 
-#if !PORTABLELIB // Timeout not available
             if (requestInfo.Timeout != 0)
             {
                 requestMessage.Timeout = requestInfo.Timeout;
+               
             }
-#endif
+
+            if (requestInfo.ReadWriteTimeout != 0)
+            {
+                requestMessage.ReadWriteTimeout = requestInfo.ReadWriteTimeout;
+            }
 
             return new TopLevelRequestMessageWrapper(requestMessage, requestInfo, requestMessageArgs.Descriptor);
         }
--- a/src/Microsoft.OData.Client/Serialization/HttpWebRequestMessage.cs
+++ b/src/Microsoft.OData.Client/Serialization/HttpWebRequestMessage.cs
@@ -177,7 +177,6 @@ namespace Microsoft.OData.Client
             set { this.httpRequest.Credentials = value; }
         }
 
-#if !PORTABLELIB
         /// <summary>
         /// Gets or sets the timeout (in seconds) for this request.
         /// </summary>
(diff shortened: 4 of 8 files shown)


src/Microsoft.OData.Client/Serialization/HttpWebRequestMessage.cs
```diff
@@ -187,6 +187,15 @@ public override int Timeout
             set { this.httpRequest.Timeout = (int)Math.Min(Int32.MaxValue, new TimeSpan(0, 0, value).TotalMilliseconds); }
         }
 
+        /// <summary>
+        /// Gets or sets the read and write timeout (in seconds) for this request.
```
marabooy, line 191: Although this says it is in seconds, the set value isn't the same value that is read out. 
if i set to 100s when reading it seems like the returned value will be in milliseconds

ElizabethOkerio: @marabooy yes you set the value in seconds.. but when it is being passed to HttpWebRequest it is converted to milliseconds.

src/Microsoft.OData.Client/ODataRequestMessageWrapper.cs
```diff
@@ -161,6 +161,8 @@ internal static ODataRequestMessageWrapper CreateRequestMessageWrapper(BuildingR
             if (requestInfo.Timeout != 0)
             {
                 requestMessage.Timeout = requestInfo.Timeout;
+                //set readwritetimeout value as the set timeout value
+                requestMessage.ReadWriteTimeout = requestInfo.Timeout;
```
odero, line 165: I'm wondering if it would probably make sense to allow the user to set the 2 values separately and default to `ReadWriteTimeout = Timeout` only when `ReadWriteTimeout` is not separately defined

src/Microsoft.OData.Client/InternalODataRequestMessage.cs
```diff
@@ -115,6 +115,15 @@ public override int Timeout
             set { throw new NotSupportedException(); }
         }
 
+        /// <summary>
+        /// Gets or sets the read and write timeout (in seconds) for this request.
+        /// </summary>
+        public override int ReadWriteTimeout
```
g2mula, line 121: What is this class for? It just throws exceptions...? Should it probably be removed?

src/Microsoft.OData.Client/Serialization/HttpWebRequestMessage.cs
```diff
@@ -187,6 +187,15 @@ public override int Timeout
             set { this.httpRequest.Timeout = (int)Math.Min(Int32.MaxValue, new TimeSpan(0, 0, value).TotalMilliseconds); }
         }
 
+        /// <summary>
+        /// Gets or sets the read and write timeout (in seconds) for this request.
+        /// </summary>
+        public override int ReadWriteTimeout
+        {
+            get { return this.httpRequest.Timeout; }
+            set { this.httpRequest.Timeout = (int)Math.Min(Int32.MaxValue, new TimeSpan(0, 0, value).TotalMilliseconds); }
```
g2mula, line 196:
```suggestion
            get { return this.Timeout; }
            set { this.Timeout = value); }
```
Seems to be repeating the Timeout code, might as well just forward these requests to that.

src/Microsoft.OData.Client/ODataRequestMessageWrapper.cs
ElizabethOkerio: resolved.

src/Microsoft.OData.Client/InternalODataRequestMessage.cs
ElizabethOkerio: @g2mula, the new property is another timeout property of an <code> HttpWebRequest </code>. Basically, the <code> HttpWebRequest </code> that <code> Microsoft.OData.Client </code> uses has a Timeout property and a ReadWriteTimeout property. These properties have default values but you can update the default values if your request takes more than the default values to evaluate. Currently you can update the Timeout property but cannot update the ReadWriteTimeout property.

src/Microsoft.OData.Client/Serialization/HttpWebRequestMessage.cs
ElizabethOkerio: fixed

src/Microsoft.OData.Client/InternalODataRequestMessage.cs
g2mula: Asking about this class, the InternalODataRequestMessage class, is it used anywhere?

test/EndToEndTests/Tests/Client/Build.Desktop/TransportLayerTests/HttpClientRequestMessage.cs
```diff
@@ -137,6 +137,18 @@ public override int Timeout
...
 
+        public override int ReadWriteTimeout
+        {
+            get
+            {
+                return (int)this.client.Timeout.TotalSeconds;
+            }
+            set
+            {
+                this.client.Timeout = new TimeSpan(0, 0, value);
```
g2mula, line 148: Needs an summary documentation explaining the timeout is in seconds.

test/FunctionalTests/Tests/DataServices/UnitTests/ClientExtensions/TestDataServiceClientRequestMessage.cs
```diff
@@ -85,6 +85,18 @@ public override int Timeout
                 throw new NotImplementedException();
             }
         }
+
+        public override int ReadWriteTimeout
```
g2mula, line 89: Is this class used for anything? It just throws exceptions...

src/Microsoft.OData.Client/Serialization/HttpWebRequestMessage.cs
g2mula: This is not consistent and would be a source of bugs :-(
```
var message = new HttpWebRequestMessage();
var expected = 1; // second
message.ReadWriteTimeout = expected;
var actual = message.ReadWriteTimeout; // => 1000ms?!
Assert.AreEqual(expected, actual, "Property set should generally be equal to property read");
```
Options for this could be
- Use various properties for millisecond get/set, second get/set, minutes get/set... Or simply change the type to a Timespan
- Use ```SetReadWriteTimeoutInSeconds(int seconds)``` and ```GetReadWriteTimeoutInMilliseconds()```
- ?

src/Microsoft.OData.Client/InternalODataRequestMessage.cs
ElizabethOkerio: Yes there is a place where that class is required to throw exceptions if called.

src/Microsoft.OData.Client/Serialization/HttpWebRequestMessage.cs
ElizabethOkerio: Ok. so the ReadWriteTimeout property has been added to the <code> DataServiceContext </code> class. This means that when a user is configuring the context that is where the set a value for this property in seconds(it is easier for users). But within the code when this value is being set to <code> HttpWebRequest</code> it is converted to milliseconds. This means that what the <HttpWebRequest </code> gets is in milliseconds. In your case there, that is not how it works. 
<pre>
var valfromcontext = 1; //second
var message = new HttpWebRequestMessage();
var expected = 1000; //millisecond when passing to webrequestmessage we convert this value to milliseconds
message.ReadWriteTimeout = expected;
var actual = message.ReadWriteTimeout; // => 1000ms?!
Assert.AreEqual(expected, actual, "Property set should generally be equal to property read");

</pre>

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -115,6 +115,9 @@ public class DataServiceContext
 #if !PORTABLELIB // Timeout not available
```
xuzhg, line 115: Can you also help to remove the "#if !PORTABLELIB"?

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -507,6 +510,33 @@ public virtual int Timeout
...
+        public virtual int ReadWriteTimeout
+        {
+            get
+            {
+                return this.readWriteTimeout;
+            }
+
+            set
+            {
+                if (value < 0)
```
xuzhg, line 532: Can we consider "<0" means infinity

src/Microsoft.OData.Client/ODataRequestMessageWrapper.cs
```diff
@@ -161,6 +161,12 @@ internal static ODataRequestMessageWrapper CreateRequestMessageWrapper(BuildingR
             if (requestInfo.Timeout != 0)
             {
                 requestMessage.Timeout = requestInfo.Timeout;
+               
+            }
+
+            if (requestInfo.ReadWriteTimeout != 0)
```
xuzhg, line 167: why shall we exclude the "0"?
I mean what's the effect if we directly copy the timeout ?

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -115,6 +115,9 @@ public class DataServiceContext
 #if !PORTABLELIB // Timeout not available
         /// <summary>time-out value in seconds, 0 for default</summary>
         private int timeout;
```
xuzhg, line 117: what's the difference between "timeout" and "readWriteTimeout" ?

src/Microsoft.OData.Client/ODataRequestMessageWrapper.cs
ElizabethOkerio: so we check if the user set this value,, if they did not set it then it is zero. so if the user set the value then that value is the one that is used. Otherwise <code> HttpWebRequest </code> uses its default value.

src/Microsoft.OData.Client/DataServiceContext.cs
ElizabethOkerio: timeout  is the number of milliseconds to wait before the request(basically the connection) timesout while readWriteTimeout is the number of milliseconds to wait before the writing/reading of a request/response timeouts.

PR #1806, opened 2020-06-12
Sequencing Json Batch requests with the dependsOn property

KenitoInc:
### Description
Individual requests can be executed in a specified order by using the dependsOn property. 
The value of dependsOn is an array of strings whose values MUST be values of either id or atomicityGroup of preceding request objects; forward references are not allowed. https://docs.oasis-open.org/odata/odata-json-format/v4.01/odata-json-format-v4.01.html#sec_BatchRequest

This PR introduces a `dependsOnIds` property in the `Descriptor` and an overload of the `DataServiceContext.UpdateObject` method.
`UpdateObject(object entity, params object[] dependsOnIds)`

From the client, one make the calls as shown in the sample below
```csharp
Container context = new Container(uri);
var book1 = context.Books.Where(b => b.Id == 1).First();
var book2 = context.Books.Where(b => b.Id == 2).First();
book1.Year = 2001;
book2.Year = 2002;

context.UpdateObject(book1);
// update of book2 depends on book1
context.UpdateObject(book2, book1);
DataServiceResponse response = await context.SaveChangesAsync(SaveChangesOptions.BatchWithIndependentOperations | SaveChangesOptions.UseJsonBatch);
```

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/BatchSaveResult.cs
+++ b/src/Microsoft.OData.Client/BatchSaveResult.cs
@@ -370,12 +370,27 @@ namespace Microsoft.OData.Client
 
                     for (int i = 0; i < this.ChangedEntries.Count; ++i)
                     {
+                        Descriptor descriptor = this.ChangedEntries[i];
+
                         if (Util.IsBatchWithIndependentOperations(this.Options))
                         {
-                            this.batchWriter.WriteStartChangeset();
+                            if(descriptor.ChangeSetId != null)
+                            {
+                                this.batchWriter.WriteStartChangeset(descriptor.ChangeSetId);
+                            }
+                            else
+                            {
+                                this.batchWriter.WriteStartChangeset();
+                            }
                         }
 
-                        Descriptor descriptor = this.ChangedEntries[i];
+                        // In BatchWithIndependentOperations, we use atomicityGroup(in this case changeSetId)
+                        // as the dependsOn property
+                        if(descriptor.DependsOnIds != null && Util.IsBatchWithIndependentOperations(this.Options))
+                        {
+                            descriptor.DependsOnIds = descriptor.DependsOnChangeSetIds;
+                        }
+
                         if (descriptor.ContentGeneratedForSave)
                         {
                             continue;
--- a/src/Microsoft.OData.Client/BatchSaveResult.cs
+++ b/src/Microsoft.OData.Client/BatchSaveResult.cs
@@ -371,8 +371,9 @@ namespace Microsoft.OData.Client
                     for (int i = 0; i < this.ChangedEntries.Count; ++i)
                     {
                         Descriptor descriptor = this.ChangedEntries[i];
+                        bool isBatchWithIndependentOperations = Util.IsBatchWithIndependentOperations(this.Options);
 
-                        if (Util.IsBatchWithIndependentOperations(this.Options))
+                        if (isBatchWithIndependentOperations)
                         {
                             if(descriptor.ChangeSetId != null)
                             {
@@ -386,7 +387,7 @@ namespace Microsoft.OData.Client
 
                         // In BatchWithIndependentOperations, we use atomicityGroup(in this case changeSetId)
                         // as the dependsOn property
-                        if(descriptor.DependsOnIds != null && Util.IsBatchWithIndependentOperations(this.Options))
+                        if(descriptor.DependsOnIds != null && isBatchWithIndependentOperations)
                         {
                             descriptor.DependsOnIds = descriptor.DependsOnChangeSetIds;
                         }
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -2533,6 +2533,28 @@ namespace Microsoft.OData.Client
 
         /// <summary>Changes the state of the specified object to be deleted in the <see cref="Microsoft.OData.Client.DataServiceContext" />.</summary>
         /// <param name="entity">The tracked entity to be changed to the deleted state.</param>
+        /// <param name="dependsOnIds">DependsOnIds for the deleted entity.</param>
+        /// <exception cref="System.ArgumentNullException">When <paramref name="entity" /> is null.</exception>
+        /// <exception cref="System.InvalidOperationException">When the object is not being tracked by the <see cref="Microsoft.OData.Client.DataServiceContext" />.</exception>
+        /// <remarks>
+        /// Existing objects in the Added state become detached.
+        /// </remarks>
+        public virtual void DeleteObject(object entity, params object[] dependsOnIds)
+        {
+            var dependsOn = new List<string>();
+            if (dependsOnIds != null)
+            {
+                foreach (var id in dependsOnIds)
+                {
+                    dependsOn.Add(id.ToString());
+                }
+            }
+
+            this.DeleteObjectInternal(entity, false /*failIfInAddedState*/, dependsOn);
+        }
+
+        /// <summary>Changes the state of the specified object to be deleted in the <see cref="Microsoft.OData.Client.DataServiceContext" />.</summary>
+        /// <param name="entity">The tracked entity to be changed to the deleted state.</param>
         /// <exception cref="System.ArgumentNullException">When <paramref name="entity" /> is null.</exception>
         /// <exception cref="System.InvalidOperationException">When the object is not being tracked by the <see cref="Microsoft.OData.Client.DataServiceContext" />.</exception>
         /// <remarks>
@@ -2584,7 +2606,7 @@ namespace Microsoft.OData.Client
         /// <exception cref="System.ArgumentException">When <paramref name="entity" /> is in the <see cref="Microsoft.OData.Client.EntityStates.Detached" /> state.</exception>
         public virtual void UpdateObject(object entity)
         {
-            this.UpdateObject(entity, null);
+            this.UpdateObjectInternal(entity, false /*failIfInAddedState*/);
         }
 
         /// <summary>Update a related object to the context.</summary>
@@ -3926,24 +3948,14 @@ namespace Microsoft.OData.Client
                 //We will extract ChangeOrder from the entity descriptor based on the Id.
                 //This is because we pass the ChangeOrder as Content-ID to the batch writer.
                 //We will pass the Content-ID(s) as dependsOnIds to the batch writer.
-                var dependsOnIdsAsChangeOrders = new List<string>();
-                var dependsOnChangeSetIds = new List<string>();
-                foreach (var dependOnId in dependsOnIds)
-                {
-                    foreach (var entityDescriptor in this.EntityTracker.Entities)
-                    {
-                        var e = (object)entityDescriptor.Entity;
-                        string entityName = e.GetType().FullName;
-                        var key = GetKeyPropertyName(this.Model, entityName);
-                        var id = e.GetType().GetProperty(key).GetValue(e, null);
-                        if(dependOnId == id.ToString())
-                        {
-                            var changeOrder = entityDescriptor.ChangeOrder;
-                            dependsOnIdsAsChangeOrders.Add(changeOrder.ToString(CultureInfo.CurrentCulture));
-                            dependsOnChangeSetIds.Add(entityDescriptor.ChangeSetId);
-                        }
-                    }
-                }
+                List<string> dependsOnIdsAsChangeOrders, dependsOnChangeSetIds;
+
+                GetDependsOnChangeOrdersAndChangeSetIds(
+                    dependsOnIds,
+                    this.EntityTracker.Entities,
+                    this.Model,
+                    out dependsOnIdsAsChangeOrders,
+                    out dependsOnChangeSetIds);
                 resource.DependsOnIds = dependsOnIdsAsChangeOrders;
                 resource.DependsOnChangeSetIds = dependsOnChangeSetIds.Distinct().ToList();
             }
@@ -3951,18 +3963,6 @@ namespace Microsoft.OData.Client
             resource.State = EntityStates.Modified;
             resource.ChangeSetId = Guid.NewGuid().ToString();
             this.entityTracker.IncrementChange(resource);
-        }
-
-        /// <summary>
-        /// Returns the name of the Declared Key for an entity.
-        /// </summary>
-        /// <param name="model"></param>
-        /// <param name="entityName"></param>
-        /// <returns>Declared key property name.</returns>
-        private static string GetKeyPropertyName(ClientEdmModel model, string entityName)
-        {
-            EdmEntityTypeWithDelayLoadedProperties schemaType = (EdmEntityTypeWithDelayLoadedProperties)model.FindDeclaredType(entityName);
-            return schemaType.DeclaredKey.First().Name;
         }
 
         /// <summary>
@@ -3983,12 +3983,13 @@ namespace Microsoft.OData.Client
         /// </summary>
         /// <param name="entity">entity to be mark deleted</param>
         /// <param name="failIfInAddedState">If true, then an exception will be thrown if the entity is in the added state.</param>
+        /// <param name="dependsOnIds">DependsOnIds for the deleted entity.</param>
         /// <exception cref="ArgumentNullException">if entity is null</exception>
         /// <exception cref="InvalidOperationException">if entity is not being tracked by the context, or if the entity is in the added state and <paramref name="failIfInAddedState"/> is true.</exception>
         /// <remarks>
         /// Existing objects in the Added state become detached if <paramref name="failIfInAddedState"/> is false.
         /// </remarks>
-        private void DeleteObjectInternal(object entity, bool failIfInAddedState)
+        private void DeleteObjectInternal(object entity, bool failIfInAddedState, List<string> dependsOnIds)
         {
             Util.CheckArgumentNull(entity, "entity");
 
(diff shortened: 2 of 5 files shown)


src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2529,6 +2531,28 @@ public virtual void AttachTo(string entitySetName, object entity, string etag)
...
+        /// Existing objects in the Added state become detached.
+        /// </remarks>
+        public virtual void DeleteObject(object entity, params object[] dependsOnIds)
+        {
+            var dependsOn = new List<string>();
+            if (dependsOnIds != null)
+            {
+                foreach (var id in dependsOnIds)
+                {
+                    dependsOn.Add(id.ToString());
```
marabooy, line 2549: This is repeated in the file in `UpdateObject` refactor to reduce repetition.

marabooy, line 2549: [Question] Are we always sure that the depends on can always be converted to a string? What types are usually used for this, Guids? ints?

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3899,21 +3942,54 @@ private void UpdateObjectInternal(object entity, bool failIfNotUnchanged)
...
+                List<string> dependsOnIdsAsChangeOrders, dependsOnChangeSetIds;
+
+                GetDependsOnChangeOrdersAndChangeSetIds(
+                    dependsOnIds,
+                    this.EntityTracker.Entities,
+                    this.Model,
+                    out dependsOnIdsAsChangeOrders,
+                    out dependsOnChangeSetIds);
+                resource.DependsOnIds = dependsOnIdsAsChangeOrders;
+                resource.DependsOnChangeSetIds = dependsOnChangeSetIds.Distinct().ToList();
```
marabooy, line 3960: To avoid this check maybe change the container to a hashset.

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3935,13 +4011,88 @@ private void DeleteObjectInternal(object entity, bool failIfInAddedState)
...
+
+                    GetDependsOnChangeOrdersAndChangeSetIds(
+                        dependsOnIds,
+                        this.EntityTracker.Entities,
+                        this.Model,
+                        out dependsOnIdsAsChangeOrders,
+                        out dependsOnChangeSetIds);
+
+                    resource.DependsOnIds = dependsOnIdsAsChangeOrders;
+                    resource.DependsOnChangeSetIds = dependsOnChangeSetIds.Distinct().ToList();
```
marabooy, line 4030: Maybe a hashset would be better here. So as to avoid the Distinct checkup.

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3935,13 +4011,88 @@ private void DeleteObjectInternal(object entity, bool failIfInAddedState)
...
+                foreach (var entityDescriptor in entities)
+                {
+                    var e = (object)entityDescriptor.Entity;
+                    string entityName = e.GetType().FullName;
+                    var key = GetKeyPropertyName(model, entityName);
+                    var id = e.GetType().GetProperty(key).GetValue(e, null);
+                    if (dependOnId == id.ToString())
+                    {
+                        var changeOrder = entityDescriptor.ChangeOrder;
+                        dependsOnIdsAsChangeOrders.Add(changeOrder.ToString(CultureInfo.CurrentCulture));
```
marabooy, line 4077: [Question] What would happen if we had InvariantCulture instead of current. Would this affect the generation of keys some languages?

test/EndToEndTests/Tests/Client/AsynchronousTests/AsyncMethodTests.cs
```diff
@@ -135,6 +136,51 @@ public async Task SaveChangesTest()
...
+            await context.SaveChangesAsync(SaveChangesOptions.BatchWithSingleChangeset);
+
+            c1.Name = "customerOne updated name";
+            c2.Name = "customerTwo updated name";
+
+            context.UpdateObject(c1);
+            context.UpdateObject(c2, c1.CustomerId);
+
+            var dscResponse = await context.SaveChangesAsync(SaveChangesOptions.BatchWithIndependentOperations | SaveChangesOptions.UseJsonBatch);
+            Assert.Equal(204, (dscResponse.First() as ChangeOperationResponse).StatusCode);
```
marabooy, line 156: [Question] Shouldn't we be checking the second item as well as it had the depends on property?

marabooy (changes requested):

test/EndToEndTests/Tests/Client/AsynchronousTests/AsyncMethodTests.cs
KenitoInc: Good catch!

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: This is the basic assumption since `ODataBatchWriter.CreateOperationRequestMessage` only accepts `IEnumerable<string>` as the dependsOnIds.
https://github.com/OData/odata.net/blob/25ea344714d291d169b48ebf71c09721014989aa/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs#L294-L295

test/EndToEndTests/Tests/Client/AsynchronousTests/AsyncMethodTests.cs
KenitoInc: Resolved

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: Fixed to `InvariantCulture`

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: Done

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: Done

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: Done!

src/Microsoft.OData.Client/BatchSaveResult.cs
```diff
@@ -370,12 +370,27 @@ private ODataRequestMessageWrapper GenerateBatchRequest()
...
+                            }
+                            else
+                            {
+                                this.batchWriter.WriteStartChangeset();
+                            }
+                        }
+
+                        // In BatchWithIndependentOperations, we use atomicityGroup(in this case changeSetId)
+                        // as the dependsOn property
+                        if(descriptor.DependsOnIds != null && Util.IsBatchWithIndependentOperations(this.Options))
```
gathogojr, line 389: @KenitoInc Since the `if` block above this one also calls `Util.IsBatchWithIndependentOperations(this.Options)`, you could store the result in a variable declared in the same scope as the `descriptor` variable.

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2558,6 +2575,18 @@ public virtual bool Detach(object entity)
...
+        /// <summary>Changes the state of the specified object in the <see cref="Microsoft.OData.Client.DataServiceContext" /> to <see cref="Microsoft.OData.Client.EntityStates.Modified" />.</summary>
+        /// <param name="entity">The tracked entity to be assigned to the <see cref="Microsoft.OData.Client.EntityStates.Modified" /> state.</param>
+        /// <param name="dependsOnIds">DependsOnIds for the modified entity.</param>
+        /// <exception cref="System.ArgumentNullException">When <paramref name="entity" /> is null.</exception>
+        /// <exception cref="System.ArgumentException">When <paramref name="entity" /> is in the <see cref="Microsoft.OData.Client.EntityStates.Detached" /> state.</exception>
+        public virtual void UpdateObject(object entity, params object[] dependsOnIds)
+        {
+            List<string> dependsOn = ConvertDependOnIdsToString(dependsOnIds);
+
+            this.UpdateObjectInternal(entity, false /*failIfNotUnchanged*/, dependsOn.Count>0 ? dependsOn:null);
```
gathogojr, line 2587:
```suggestion
            this.UpdateObjectInternal(entity, false /*failIfNotUnchanged*/, dependsOn.Count > 0 ? dependsOn : null);
```

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3899,21 +3948,55 @@ private void UpdateObjectInternal(object entity, bool failIfNotUnchanged)
...
+                    this.EntityTracker.Entities,
+                    this.Model,
+                    out dependsOnIdsAsChangeOrders,
+                    out dependsOnChangeSetIds);
+                resource.DependsOnIds = dependsOnIdsAsChangeOrders;
+                resource.DependsOnChangeSetIds = dependsOnChangeSetIds.ToList();
+            }
+
             resource.State = EntityStates.Modified;
+            resource.ChangeSetId = Guid.NewGuid().ToString();
```
gathogojr, line 3971: @KenitoInc Is the assumption here that `ChangeSetId` was null? Would there be side effects if there was a value and it gets overwritten?

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3935,13 +4018,89 @@ private void DeleteObjectInternal(object entity, bool failIfInAddedState)
                     Util.IncludeLinkState(state),
                     "bad state transition to deleted");
 
+                if (dependsOnIds != null)
```
gathogojr, line 4021: @KenitoInc This block of code mirrors the other in `UpdateObjectInternal`. Refactor into a shared object?

gathogojr: As a matter of fact you could redesign and rename `GetDependsOnChangeOrdersAndChangeSetIds` such that you pass `resource` into it and that method populates `DependsOnIds` and `DependsOnChangeSetIds` properties

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3935,13 +4018,89 @@ private void DeleteObjectInternal(object entity, bool failIfInAddedState)
...
+            }
+        }
+
+        /// <summary>
+        /// Returns the name of the Declared Key for an entity.
+        /// </summary>
+        /// <param name="model"></param>
+        /// <param name="entityName"></param>
+        /// <returns>Declared key property name.</returns>
+        private static string GetKeyPropertyName(ClientEdmModel model, string entityName)
```
gathogojr, line 4098: @KenitoInc They way you're going about identifying the key properties here seems very direct and I have to wonder about the logic in [`GetKeyPropertiesOnType` method](https://github.com/OData/odata.net/blob/325e28394a473e579ac7573e0928ebb3635b5765/src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs#L428). Would the existing method do the job efficiently?

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3935,13 +4018,89 @@ private void DeleteObjectInternal(object entity, bool failIfInAddedState)
...
+        {
+            EdmEntityTypeWithDelayLoadedProperties schemaType = (EdmEntityTypeWithDelayLoadedProperties)model.FindDeclaredType(entityName);
+            return schemaType.DeclaredKey.First().Name;
```
gathogojr, line 4101: @KenitoInc If we only pick the first property, would the `dependOn` feature be foolproof when the entity has a compound key?

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2567,6 +2596,25 @@ public virtual void UpdateObject(object entity)
             this.UpdateObjectInternal(entity, false /*failIfNotUnchanged*/);
         }
 
+        /// <summary>
+        /// Convert array object to List of strings.
+        /// </summary>
+        /// <param name="dependsOnIds">Object array of DependsOnIds.</param>
+        /// <returns>List of strings.</returns>
+        private static List<string> ConvertDependOnIdsToString(object[] dependsOnIds)
```
marabooy, line 2604: since this list is not expanded could we switch this out to something that won't allocate any list. Since we are only doing conversion I believe yield return id.toString() could give us better performance memory wise as it had no allocations. 
```suggestion
        private static IEnumerable<string> ConvertDependOnIdsToString(object[] dependsOnIds)

```

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2567,6 +2596,25 @@ public virtual void UpdateObject(object entity)
...
+        /// <param name="dependsOnIds">Object array of DependsOnIds.</param>
+        /// <returns>List of strings.</returns>
+        private static List<string> ConvertDependOnIdsToString(object[] dependsOnIds)
+        {
+            var dependsOn = new List<string>();
+            if (dependsOnIds != null)
+            {
+                foreach (var id in dependsOnIds)
+                {
+                    dependsOn.Add(id.ToString());
```
marabooy, line 2611:
```suggestion
                    yield return id.ToString();

```

marabooy (commented):
Looks good may need a bit of small changes to reduce allocations.

src/Microsoft.OData.Client/Descriptor.cs
```diff
@@ -125,6 +135,27 @@ internal virtual bool IsModified
             }
         }
 
+        /// <summary>DependsOnIds for the modified entity.</summary>
```
gathogojr, line 138: @KenitoInc How about using auto-properties?

src/Microsoft.OData.Client/ODataRequestMessageWrapper.cs
```diff
@@ -127,7 +127,9 @@ internal bool SendChunked
                 requestMessageArgs.Method,
                 requestMessageArgs.RequestUri,
                 contentId,
-                isRelativeUri ? BatchPayloadUriOption.RelativeUri : BatchPayloadUriOption.AbsoluteUri);
+                isRelativeUri ? BatchPayloadUriOption.RelativeUri : BatchPayloadUriOption.AbsoluteUri,
+                requestMessageArgs.Descriptor == null ? null : requestMessageArgs.Descriptor.DependsOnIds
```
gathogojr, line 131:
```suggestion
                requestMessageArgs.Descriptor?.DependsOnIds
```

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3935,13 +4018,89 @@ private void DeleteObjectInternal(object entity, bool failIfInAddedState)
...
+
+            foreach (string dependOnId in dependsOnIds)
+            {
+                foreach (EntityDescriptor entityDescriptor in entities)
+                {
+                    object entity = (object)entityDescriptor.Entity;
+                    string entityName = entity.GetType().FullName;
+                    string key = GetKeyPropertyName(model, entityName);
+                    object id = entity.GetType().GetProperty(key).GetValue(entity, null);
+                    if (dependOnId == id.ToString())
```
gathogojr, line 4082: @KenitoInc Should we do a case-insensitive string comparison here? To mitigate scenarios like:
- `dependsOnId = 0d599e52-ab31-4084-8b16-f5bebfcc70ef`, and
- `id = 0D599E52-AB31-4084-8B16-F5BEBFCC70EF`...

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3935,13 +4018,89 @@ private void DeleteObjectInternal(object entity, bool failIfInAddedState)
...
+                    {
+                        uint changeOrder = entityDescriptor.ChangeOrder;
+                        dependsOnIdsAsChangeOrders.Add(changeOrder.ToString(CultureInfo.InvariantCulture));
+                        dependsOnChangeSetIds.Add(entityDescriptor.ChangeSetId);
```
gathogojr, line 4086: @KenitoInc Given that we're comparing `dependsOnId` against the key property, shouldn't we break from the `for` loop after entering this `if` block?

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3935,13 +4018,89 @@ private void DeleteObjectInternal(object entity, bool failIfInAddedState)
...
+            List<string> dependsOnIds,
+            IEnumerable<EntityDescriptor> entities,
+            ClientEdmModel model,
+            out List<string>  dependsOnIdsAsChangeOrders,
+            out HashSet<string> dependsOnChangeSetIds)
+        {
+            dependsOnIdsAsChangeOrders = new List<string>();
+            dependsOnChangeSetIds = new HashSet<string>();
+
+            foreach (string dependOnId in dependsOnIds)
```
gathogojr, line 4074:
```suggestion
            foreach (string dependsOnId in dependsOnIds)
```

src/Microsoft.OData.Client/DataServiceContext.cs
marabooy: If we do make this into an IEnumerable `this.UpdateObjectInternal(entity, false /*failIfNotUnchanged*/, dependsOn.Count > 0 ? dependsOn : null);` this check which is done after the conversion could be done using the dependsOnIds.

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3935,13 +4018,89 @@ private void DeleteObjectInternal(object entity, bool failIfInAddedState)
...
+            {
+                foreach (EntityDescriptor entityDescriptor in entities)
+                {
+                    object entity = (object)entityDescriptor.Entity;
+                    string entityName = entity.GetType().FullName;
+                    string key = GetKeyPropertyName(model, entityName);
+                    object id = entity.GetType().GetProperty(key).GetValue(entity, null);
```
gathogojr, line 4081:
```suggestion
                    object id = entity.GetType().GetProperty(key).GetValue(entity);
```

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3935,13 +4018,89 @@ private void DeleteObjectInternal(object entity, bool failIfInAddedState)
...
+
+                    GetDependsOnChangeOrdersAndChangeSetIds(
+                        dependsOnIds,
+                        this.EntityTracker.Entities,
+                        this.Model,
+                        out dependsOnIdsAsChangeOrders,
+                        out dependsOnChangeSetIds);
+
+                    resource.DependsOnIds = dependsOnIdsAsChangeOrders;
+                    resource.DependsOnChangeSetIds = dependsOnChangeSetIds.ToList();
```
marabooy, line 4038: [Question] Once this value is set? do we loop through it multiple times or could we even refactor it out so it can take either hashset or an IEnumerable? Is this property exposed to the world at the moment?

src/Microsoft.OData.Client/DataServiceContext.cs
gathogojr: Later found out that this is a property that you added...

src/Microsoft.OData.Client/BatchSaveResult.cs
KenitoInc: Resolved

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: I updated the design of this feature.
We now have a straight forward way of getting the `ChangeOrder` without relying on the key property.

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: See this comment https://github.com/OData/odata.net/pull/1806#discussion_r610376573

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: See this comment https://github.com/OData/odata.net/pull/1806#discussion_r610376573

src/Microsoft.OData.Client/Descriptor.cs
KenitoInc: I had thought of the same, however other properties in the same file don't used autoproperties.
Maybe we can cleanup this code later. What do you think?

src/Microsoft.OData.Client/ODataRequestMessageWrapper.cs
KenitoInc: Resolved

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: Once I changed the design, this method was removed

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: We pass the value as it is when calling `ODataBatchWriter`.`CreateOperationsRequestMessage` https://github.com/OData/odata.net/blob/4e263ca4b94d01f2d105d8282c62ec272cfcd1bb/src/Microsoft.OData.Client/ODataRequestMessageWrapper.cs#L126

The non-async `CreateOperationsRequestMessage` accepts `IEnumerable<string>` https://github.com/OData/odata.net/blob/4e263ca4b94d01f2d105d8282c62ec272cfcd1bb/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs#L294-L295
However the async method accepts `IList<string>` https://github.com/OData/odata.net/blob/4e263ca4b94d01f2d105d8282c62ec272cfcd1bb/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs#L339-L340
That's why I decided to stick with `List<string>`

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: Resolved

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: We are no longer using this code section after design change

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: We are no longer using this code section after design change

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: We are no longer using this code section after design change

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: We are no longer using this code section after design change

marabooy (approved):

gathogojr (approved):
LGTM... I hope the tests you added cover the diff @KenitoInc

Issue #1809, opened 2020-06-16
Bug affecting $apply/aggregate when using average aggregation method against floats and nullable floats


gathogojr:
It appears there might an ODL bug affecting $apply/aggregate expression specifically when using average aggregation method against floats and nullable floats

### Assemblies affected

- Microsoft.OData.Core 7.6.4
- Microsoft.AspNetCore.OData 7.4.1

### Reproduce steps

1. Create an OData Web API project
2. Create a CLR type with float and nullable float properties
    ```c#
    public class Number
    {
        public int Id { get; set; }
        public float SingleProp { get; set; }
        public float? NullableSingleProp { get; set; }
    }
    ```
3. Create a controller with a `Get` action
    ```c#
    public class NumbersController : ODataController
    {
        // ...
        [EnableQuery]
        public IQueryable<Number> Get()
        {
            return _db.Numbers;
        }
        // ...
    }
4. Bootstrap the service
    ```c#
    // ...
    ODataModelBuilder modelBuilder = new ODataConventionModelBuilder();
    modelBuilder.EntitySet<Number>("Numbers");

    app.UseMvc(routeBuilder =>
    {
        routeBuilder.Select().Filter().Expand().Count().OrderBy().SkipToken().MaxTop(null);
        routeBuilder.MapODataServiceRoute("odata", "odata", modelBuilder.GetEdmModel());
    });
    // ...
    ```
5. Try to query for averages for the float and nullable float properties using $apply/aggregate expression
    a. `http://ServiceRoot/Numbers?$apply=aggregate(SingleProp with average as AverageSingleProp)`
    b. `http://ServiceRoot/Numbers?$apply=aggregate(NullableSingleProp with average as AverageNullableSingleProp)`

### Expected result

The averages values for both properties should be returned in the respective cases

### Actual result

An exception it thrown

> {
>     "error": {
>         "code": "",
>         "message": "The query specified in the URI is not valid. $apply/aggregate expression 'Microsoft.OData.UriParser.SingleValuePropertyAccessNode' operation does not support value type 'Single'.",
>         "details": [],
>         "innererror": {
>             "message": "$apply/aggregate expression 'Microsoft.OData.UriParser.SingleValuePropertyAccessNode' operation does not support value type 'Single'.",
>             "type": "Microsoft.OData.ODataException",
[...]

PR #1811, opened 2020-06-17
Fix parsing of floats in $apply/aggregate/average expression

gathogojr:
### Issues

*This pull request fixes issue #1809.*

### Description

Fixes a bug where `ApplyBinder` does not cater for float/single for the average aggregate method. This causes the following **average** expression to be rejected:
```
$apply=aggregate(SingleProp with average as AverageSingleProp)
```
while related **sum**, **min**, and **max** are accepted
```
$apply=aggregate(SingleProp with sum as SumSingleProp)
$apply=aggregate(SingleProp with min as MinSingleProp)
$apply=aggregate(SingleProp with min as MaxSingleProp)
```

**Note**: The return value from averaging floats should be a float
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.average?view=netcore-3.1#System_Linq_Enumerable_Average__1_System_Collections_Generic_IEnumerable___0__System_Func___0_System_Single__
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.average?view=netcore-3.1#System_Linq_Enumerable_Average__1_System_Collections_Generic_IEnumerable___0__System_Func___0_System_Nullable_System_Single___

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/FakeBindMethods.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/FakeBindMethods.cs
@@ -40,7 +40,7 @@ namespace Microsoft.OData.Tests.UriParser.Binders
         public static readonly SingleValueNode FakeSingleIntPrimitive =
             new ConstantNode(100);
 
-        public static readonly SingleValueNode FakeSingleSinglePrimitive =
+        public static readonly SingleValueNode FakeSingleFloatPrimitive =
             new ConstantNode(100.50f);
 
         public static readonly SingleValueNode FakeSingleOpenProperty =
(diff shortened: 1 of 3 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/FakeBindMethods.cs
```diff
@@ -40,6 +40,9 @@ internal class FakeBindMethods
         public static readonly SingleValueNode FakeSingleIntPrimitive =
             new ConstantNode(100);
 
+        public static readonly SingleValueNode FakeSingleSinglePrimitive =
```
KenitoInc, line 43: Even though a float is `EdmPrimitiveTypeKind.Single`, Maybe for a better description, you could have used `FakeSingleFloatPrimitive`

gathogojr: I could do that

Issue #1822, opened 2020-06-24
Fix bug on atomicityGroup as a dependsOn value in Odata v4


KenitoInc:
`atomicityGroup` as a `dependsOn` value fails for `Odata v4` requests, but succeeds in `Odata v4.01` requests.
In Json batch, we use atomicGroup Id in dependsOn for 2 reasons:
1. To extract Request ids for use in reference Uris (in v4.01).
2. Sequence requests so that they are executed in a certain order (in both v4 and v4.01).

The current implementation of dependsOn atomicityGroup is aimed at extracting request Ids for use in reference uris. 
This works well in Odata v4.01 but not in Odata v4. This is because in Odata v4, the last request Id in a group is not added to the contentId cache,hence validation fails in the snippets below. https://github.com/OData/odata.net/blob/d6cebf5cd1355e49fc8dd89545dcba0b5b4f0b47/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs#L518-L528
https://github.com/OData/odata.net/blob/f375e468b77c4c16445e6e15c0b9afcbe54d387b/src/Microsoft.OData.Core/Batch/ODataBatchReader.cs#L319-L328

The validations above should not be applicable to v4 since we may want to use atomicity groups to sequence requests without using reference Uris.

### Assemblies affected

Microsoft.OData.Core v7.6.4

### Reproduce steps

Using postman, make a batch request to an odata service with a payload similar to the one below
`POST http://localhost:6143/odata/$batch`
```json
{
  "requests": [
    {
      "id": "3",
      "atomicityGroup": "ac92c60b-bbb5-438b-ba1f-db6fbe2ee053",
      "method": "PATCH",
      "url": "http://localhost:6143/odata/Books(1)",
      "headers": {
        "odata-version": "4.0",
        "odata-maxversion": "4.0",
        "content-type": "application/json;odata.metadata=minimal",
        "accept": "application/json;odata.metadata=minimal",
        "accept-charset": "UTF-8",
        "user-agent": "Microsoft.OData.Client/7.7.0"
      },
      "body": {
        "@odata.type": "#OdataService.Book",
        "Id": 1,
        "Isbn": null,
        "Title": "B1",
        "Year": 2001
      }
    },
    {
      "id": "4",
      "atomicityGroup": "a7224f21-60e4-44b7-9c3d-a6718f3f7181",
      "dependsOn": [
        "ac92c60b-bbb5-438b-ba1f-db6fbe2ee053"
      ],
[...]

PR #1831, opened 2020-07-01
Fix/1822 depends on atomicity group id v4

KenitoInc:
### Issues

*This pull request fixes issue #1822 .*

### Description

Moved validations from `ODataBatchWriter` and `ODataBatchReader` to individual implementations of `ODataJsonLightBatchWriter` , `ODataJsonLightBatchReader`, `ODataMultipartMixedBatchWriter` and `ODataMultipartMixedBatchReader`

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchReader.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchReader.cs
@@ -54,6 +54,11 @@ namespace Microsoft.OData.JsonLight
         private ODataJsonLightBatchPayloadItemPropertiesCache messagePropertiesCache = null;
 
         /// <summary>
+        /// Collection for keeping track of unique atomic group ids and member request ids.
+        /// </summary>
+        private HashSet<string> requestIds = new HashSet<string>();
+
+        /// <summary>
         /// Constructor.
         /// </summary>
         /// <param name="inputContext">The input context to read the content from.</param>
@@ -123,6 +128,16 @@ namespace Microsoft.OData.JsonLight
             // atomicityGroup
             string atomicityGroupId = (string)this.messagePropertiesCache.GetPropertyValue(
                 ODataJsonLightBatchPayloadItemPropertiesCache.PropertyNameAtomicityGroup);
+
+            if (id != null)
+            {
+                this.requestIds.Add(id);
+            }
+
+            if (atomicityGroupId != null)
+            {
+                this.requestIds.Add(atomicityGroupId);
+            }
 
             // dependsOn
             // Flatten the dependsOn list by converting every groupId into request Ids, so that the caller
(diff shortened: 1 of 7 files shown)


src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchReader.cs
```diff
@@ -124,6 +136,21 @@ protected override ODataBatchOperationRequestMessage CreateOperationRequestMessa
             string atomicityGroupId = (string)this.messagePropertiesCache.GetPropertyValue(
                 ODataJsonLightBatchPayloadItemPropertiesCache.PropertyNameAtomicityGroup);
 
+            if (id != null && !this.requestIdToAtomicGroupId.ContainsKey(id))
+            {
+                this.requestIdToAtomicGroupId.Add(id, atomicityGroupId);
+            }
```
mikepizzo, line 142: Rather than do an explicit ContainsKey check before calling Add, why not just do:
if (id != null)
{
    this.requestIdToAtomicGroupId.TryAdd(id, atomicityGroupId);
}

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchReader.cs
```diff
@@ -53,6 +53,18 @@ internal sealed class ODataJsonLightBatchReader : ODataBatchReader
...
+        /// </summary>
+        private Dictionary<string, string> requestIdToAtomicGroupId = new Dictionary<string, string>();
+
+        /// <summary>
+        /// Dictionary for keeping track of each atomic group's member request id. This is optimization
+        /// for reversed lookup.
+        /// </summary>
+        private Dictionary<string, IList<string>> atomicityGroupIdToRequestId = new Dictionary<string, IList<string>>();
+
         /// <summary>
```
mikepizzo, line 68: Why do you need dictionaries, it looks like you just check to see if the dictionary contains the key, not the relationship between the id and the atomicity group. Can you just use a list?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchReader.cs
```diff
@@ -53,6 +53,18 @@ internal sealed class ODataJsonLightBatchReader : ODataBatchReader
...
+        /// Dictionary for keeping track of each request's associated atomic group id, which is null
+        /// for request that does not belong to atomic group.
+        /// </summary>
+        private Dictionary<string, string> requestIdToAtomicGroupId = new Dictionary<string, string>();
+
+        /// <summary>
+        /// Dictionary for keeping track of each atomic group's member request id. This is optimization
+        /// for reversed lookup.
+        /// </summary>
+        private Dictionary<string, IList<string>> atomicityGroupIdToRequestId = new Dictionary<string, IList<string>>();
```
mikepizzo, line 66: >atomicityGroupIdToRequestId [](start = 50, length = 27)

Do you need a separate list for atomicity group versus ids, or can you simply look up to see if the id has already been specified?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchReader.cs
KenitoInc: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchReader.cs
KenitoInc: I have changed it to a HashSet.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchReader.cs
KenitoInc: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchReader.cs
mikepizzo: Should you use TryAdd(id) rather than Add(id) to avoid possible exceptions if the id already exists, or are we asserting that the id doesn't already exist? (same for below)

---
In reply to: [454179343](https://github.com/OData/odata.net/pull/1831#discussion_r454179343) [](ancestors = 454179343)

mikepizzo: Never mind; I see that you moved to HashSet, which is perfect.

---
In reply to: [467600534](https://github.com/OData/odata.net/pull/1831#discussion_r467600534) [](ancestors = 467600534,454179343)

Issue #1842, opened 2020-07-16
Line break codes of batch response  are not CRLF  on Linux server.


iwate:
Batch response of multipart has LF line break code on Linux server.

### Assemblies affected

Microsoft.OData.Core 7.4.1

### Reproduce steps

I wrote a sample.

https://github.com/iwate/ODataBatchResponseSample/

### Expected result

All platform must return `CRLF`

### Actual result

https://github.com/iwate/ODataBatchResponseSample/actions/runs/171318359

Windows: Line break code = `CRLF`

Linux: Line break code = `LF`

### Additional detail

I think this error is due to using StreamWriter.
https://github.com/OData/odata.net/blob/182964dd4de0d9271d7e749ff3afde898430326c/src/Microsoft.OData.Core/RawValueWriter.cs#L177

StreamWrite use `Environment.NewLineConst`. 
https://source.dot.net/#System.Private.CoreLib/TextWriter.cs,f3a16a4f25483a3f,references

`Environment.NewLineConst` is `LF` on Unix
https://source.dot.net/#System.Private.CoreLib/Environment.UnixOrBrowser.cs,5fbfd9f39ec095a7

PR #1844, opened 2020-07-31
Add setting to control RawValueWriter's line ending

chrisspre:
### Issues

This pull request fixes issue https://github.com/OData/odata.net/issues/1842 .

### Description

By default RawValueWriter create a TextWriter with the standard line ending which is OS specific.
RFC2064 requires this to be CRLF  https://tools.ietf.org/html/rfc2046#section-5.1.1
Added setting to ODataMessageWriterSettings to control how RawValueWriter creates the underlying TextWriter.

### Checklist 

- [ ] *Test cases added*

### Additional work necessary

none

### DIFF
--- a/src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
+++ b/src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
@@ -77,6 +77,7 @@ namespace Microsoft.OData
             this.Validations = ValidationKinds.All;
             this.Validator = new WriterValidator(this);
             this.LibraryCompatibility = ODataLibraryCompatibility.Latest;
+            this.MultipartNewLine = "\r\n";
         }
 
         /// <summary>
@@ -180,6 +181,13 @@ namespace Microsoft.OData
         /// Informs the metadata builder which properties, functions, actions, links to omit.
         /// </summary>
         public ODataMetadataSelector MetadataSelector { get; set; }
+
+        /// <summary>
+        /// Gets or sets the new line character sequence used when writing multipart messages
+        /// see https://tools.ietf.org/html/rfc2046#section-5.1.1
+        /// A TextWriter uses OS specific newline but rfc2046 requires it to be CRLF.
+        /// </summary>
+        public string MultipartNewLine { get; set; }
 
         /// <summary>
         /// Gets the validator corresponding to the validation settings.
--- a/src/Microsoft.OData.Core/RawValueWriter.cs
+++ b/src/Microsoft.OData.Core/RawValueWriter.cs
@@ -173,8 +173,8 @@ namespace Microsoft.OData
             {
                 nonDisposingStream = MessageStreamWrapper.CreateNonDisposingStream(this.stream);
             }
-
-            this.textWriter = new StreamWriter(nonDisposingStream, this.encoding);
+            
+            this.textWriter = new StreamWriter(nonDisposingStream, this.encoding) { NewLine = this.settings.MultipartNewLine };
             this.jsonWriter = new JsonWriter(this.textWriter, isIeee754Compatible: false);
         }
     }


src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
```diff
@@ -181,6 +182,13 @@ public ODataUri ODataUri
         /// </summary>
         public ODataMetadataSelector MetadataSelector { get; set; }
 
+        /// <summary>
+        /// Gets or sets the new line character sequence used when writing multipart messages
+        /// see https://tools.ietf.org/html/rfc2046#section-5.1.1
+        /// A TextWriter uses OS specific newline but rfc2046 requires it to be CRLF.
+        /// </summary>
+        public string MultipartNewLine { get; set; }
```
xuzhg, line 190: >MultipartNewLine [](start = 22, length = 16)

maybe just "NewLine"

xuzhg: By the way, it's only for the law value writer, what about other writer, for the JsonWriter? 

---
In reply to: [463717992](https://github.com/OData/odata.net/pull/1844#discussion_r463717992) [](ancestors = 463717992)

src/Microsoft.OData.Core/RawValueWriter.cs
```diff
@@ -173,8 +173,8 @@ private void InitializeTextWriter()
             {
                 nonDisposingStream = MessageStreamWrapper.CreateNonDisposingStream(this.stream);
             }
-
-            this.textWriter = new StreamWriter(nonDisposingStream, this.encoding);
+            
+            this.textWriter = new StreamWriter(nonDisposingStream, this.encoding) { NewLine = this.settings.MultipartNewLine };
```
xuzhg, line 177: >(nonDisposingStream, this.encoding) { NewLine = this.settings.MultipartNewLine }; [](start = 46, length = 81)

RawValueWriter only writes "Primitive, Enum" value, both of them are single line value. (correct me @mikepizzo).

src/Microsoft.OData.Core/RawValueWriter.cs
```diff
@@ -173,8 +173,8 @@ private void InitializeTextWriter()
             {
                 nonDisposingStream = MessageStreamWrapper.CreateNonDisposingStream(this.stream);
             }
-
-            this.textWriter = new StreamWriter(nonDisposingStream, this.encoding);
+            
+            this.textWriter = new StreamWriter(nonDisposingStream, this.encoding) { NewLine = this.settings.MultipartNewLine };
             this.jsonWriter = new JsonWriter(this.textWriter, isIeee754Compatible: false);
```
xuzhg, line 178: I'd like to see the test cases to cover this change. Maybe we can use "|" as a new Line to test the result.

src/Microsoft.OData.Core/RawValueWriter.cs
chrisspre: This is how Raw Text writer is coming into the picture when using ODataBatchWriter:


`abstract class ODataBatchWriter` has an abstract method `abstract void WriteStartChangesetImplementation ();`

Which is overwritten in the `internal sealed class ODataMultipartMixedBatchWriter : ODataBatchWriter`

ODataMultipartMixedBatchWriter derives from ODataRawOutputContext whic has a TextWriter property

`class ODataMultipartMixedBatchOutputContext : ODataRawOutputContext`
` internal TextWriter TextWriter { get    { return this.rawValueWriter.TextWriter; }

Which is initialized in line 259
`   this.rawValueWriter = new RawValueWriter( … `

The implementation of for example WriteStartChangesetImplementation uses the static ( ☹ ) `ODataMultipartMixedBatchWriterUtils.WriteStartBoundary`

Line 194 of ODataMultipartMixedBatchWriterUtils
`       ODataMultipartMixedBatchWriterUtils.WriteStartBoundary(this.RawOutputContext.TextWriter, ` …

and this calls for example:  ` writer.WriteLine("--{0}", boundary);`

src/Microsoft.OData.Core/RawValueWriter.cs
chrisspre: Correct me if I am wrong but that would only test that System.IO.TextWriter is indeed honoring the NewLine setting. 
To test that the setting flows from `new ODataBatchWriter ` into the TextWriter would require to be able to mock a TextWriter which is to my understand way out of the scope of this.

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
chrisspre: > > MultipartNewLine [](start = 22, length = 16)
> 
> maybe just "NewLine"

This is specifically a setting for ODataBatchWriter and how it writes multipart messages, I would prefer to qualify it that way.

chrisspre: > By the way, it's only for the law value writer, what about other writer, for the JsonWriter?
> 
> In reply to: [463717992](https://github.com/OData/odata.net/pull/1844#discussion_r463717992) [](ancestors = 463717992)

Json Writer is a wrapper arround the same TextWriter

line 177 RawValueWriter
```
    this.textWriter = new StreamWriter(nonDisposingStream, this.encoding) { NewLine = this.settings.MultipartNewLine };
    this.jsonWriter = new JsonWriter(this.textWriter, isIeee754Compatible: false);
```

mikepizzo, line 190: Is it useful to have this as a general setting on the MessageWriterSettings? Do we expect it ever to be used except for the Multipart/Mixed batch case? Is it ever valid to set it to anything other than /r/n for that case?

Are we afraid that, if we change to always writing /r/n for multipart/mixed (as per spec), that we'll break existing batch payloads on LINUX?

src/Microsoft.OData.Core/RawValueWriter.cs
mikepizzo: If we don't think it would break clients, the ideal fix would be to have an internal lineEnding property on rawValueWriter, which defaults to the current platform-dependent line ending, and set that to /r/n when called from ODataMultipartMixedBatchWriterUtils. What we are currently writing is not compliant to the http multipart/mixed specification.

---
In reply to: [463729331](https://github.com/OData/odata.net/pull/1844#discussion_r463729331) [](ancestors = 463729331)

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
chrisspre: My thinking is no. there is no reasonable scenario to change it to something else. 
It also is used for all line breaks in RawValueWriter (which handles the multipart messages). Having mixed line breaks woudl require a lot more refactoring and there is no reason at the moment to expect it to break clients..

But I did not want to add this hard coded to the RawValueWriter but "centralize" it on the settings. A compromise might be to make this a constant (or readonly) on MessageWriterSettings.

src/Microsoft.OData.Core/RawValueWriter.cs
chrisspre: This is roughly the direction I was taking. 
But I did not want to have a property on RawValueWriter but pass it in the constructor. Instead of adding a new parameter I found it appropriate to pass the flag in as part of the settings.

marabooy (approved):

Issue #1833, opened 2020-07-08
PlatformNotSupportedException thrown when DataServiceTransportException is about to be raised


ddosfd:
Using:
Microsoft.OData.Client v7.7.0
dotnetcore 2.1

While using batch operation to unauthorized OData service instead of DataServiceTransportException I get PlatformNotSupportedException with no details of an exception

ElizabethOkerio:
@ddosfd  please provide more info on this issue for us to dig more

AdamCaviness:
@ElizabethOkerio I believe I hit the same issue as @ddosfd did but in a different class, `DataServiceClientException`. I am on netcore3.1 but the same exception is raised. Here's the portion of the StackTrace as it transitions from my code to the OData Client code:

> System.PlatformNotSupportedException
  HResult=0x80131539
  Message=Secure binary serialization is not supported on this platform.
  Source=System.Private.CoreLib
  StackTrace:
   at System.Exception.add_SerializeObjectState(EventHandler \`1 value)
   at Microsoft.OData.Client.DataServiceClientException..ctor(String message, Exception innerException, Int32 statusCode) in C:\Users\acaviness\source\repos\ssp\Source\Microsoft.OData.Client\DataServiceClientException.cs:line 72
   at Microsoft.OData.Client.DataServiceClientException..ctor(String message, Int32 statusCode) in C:\Users\acaviness\source\repos\ssp\Source\Microsoft.OData.Client\DataServiceClientException.cs:line 58
   at Microsoft.OData.Client.BaseSaveResult.GetResponseText(Func \`1 getResponseStream, HttpStatusCode statusCode) in C:\Users\acaviness\source\repos\ssp\Source\Microsoft.OData.Client\BaseSaveResult.cs:line 252
   at Microsoft.OData.Client.BaseSaveResult.HandleResponse(RequestInfo requestInfo, HttpStatusCode statusCode, String responseVersion, Func`1 getResponseStream, Boolean throwOnFailure, Version& parsedResponseVersion) in C:\Users\acaviness\source\repos\ssp\Source\Microsoft.OData.Client\BaseSaveResult.cs:line 217
   at Microsoft.OData.Client.BatchSaveResult.ProcessCurrentOperationResponse(ODataBatchReader batchReader, Boolean isChangesetOperation) in C:\Users\acaviness\source\repos\ssp\Source\Microsoft.OData.Client\BatchSaveResult.cs:line 841
[...]

PR #1848, opened 2020-08-03
fix platform not supported exception

ElizabethOkerio:
### Issues

*This pull request fixes issue #1833*

### Description

<code> Microsoft.OData.Client</code> uses <code> Exception.SerializeObjectState </code> to serialize/deserealize request exceptions. 
<code>Exception.SerializeObjectState </code> is not supported in <code>.NET CORE</code> as per this document https://docs.microsoft.com/en-us/dotnet/core/compatibility/unsupported-apis. 

Leveraging custom serialization in this PR. 

- [ ] *Test cases added* [ ] *Build and test with one-click build and test script passed*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceClientException.cs
+++ b/src/Microsoft.OData.Client/DataServiceClientException.cs
@@ -9,25 +9,17 @@ namespace Microsoft.OData.Client
     using System;
     using System.Diagnostics.CodeAnalysis;
     using System.Runtime.Serialization;
+    using System.Security.Permissions;
 
     /// <summary>
     /// The exception that is thrown when the server returns an error.
     /// </summary>
-#if !PORTABLELIB
     [Serializable]
-#endif
     [System.Diagnostics.DebuggerDisplay("{Message}")]
     [SuppressMessage("Microsoft.Design", "CA1032:ImplementStandardExceptionConstructors", Justification = "No longer relevant after .NET 4 introduction of SerializeObjectState event and ISafeSerializationData interface.")]
     public sealed class DataServiceClientException : InvalidOperationException
     {
-        /// <summary>
-        /// Contains the state for this exception.
-        /// </summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Client/DataServiceClientException.cs
```diff
@@ -66,53 +58,45 @@ public DataServiceClientException(string message, int statusCode)
...
-            this.state.StatusCode = statusCode;
+            this.statusCode = statusCode;
+        }
 
-#if !PORTABLELIB
-            this.SerializeObjectState += (sender, e) => e.AddSerializedState(this.state);
-#endif
+        [SecurityPermissionAttribute(SecurityAction.Demand, SerializationFormatter = true)]
+#pragma warning disable CS0628 // New protected member declared in sealed class
+        protected DataServiceClientException(SerializationInfo info, StreamingContext context)
```
mikepizzo, line 66: Can you add a comment explaining why we are adding a protected constructor to a sealed class? Does the serializer call this somehow?

mikepizzo: from https://docs.microsoft.com/en-us/dotnet/api/system.runtime.serialization.iserializable?view=netcore-3.1:

The ISerializable interface implies a constructor with the signature constructor (SerializationInfo information, StreamingContext context). At deserialization time, the current constructor is called only after the data in the SerializationInfo has been deserialized by the formatter. In general, this constructor should be protected **if the class is not sealed**.

Unclear whether, for a sealed class, it should still be protected. I assume this doesn't work if it's marked as private.

I'm okay with the current code, just please add a comment to the new constructors that they are for serialization and marked protected as per .NET guidelines.

ElizabethOkerio: @mikepizzo making this constructor private works fine.

PR #1849, opened 2020-08-03
Add "RemovalDate" property to deprecation annotation.

mikepizzo:
### Description

Add support for "RemovalDate" property to Revisions annotation for deprecated properties.
Also add to deprecation tests to validate extended properties.

- [x] *Test cases added*

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/UrlValidation/DeprecationTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/UrlValidation/DeprecationTests.cs
@@ -34,9 +34,13 @@ namespace Microsoft.OData.Tests
         [InlineData(@"competitors?$filter=contains(name,'sprocket')", "competitors", "name","state","name")]
         public static void WithDeprecatedElementsGeneratesErrors(String request, params string[] expectedErrors)
         {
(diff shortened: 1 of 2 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/UrlValidation/DeprecationTests.cs
```diff
@@ -97,9 +112,11 @@ private static IEdmModel GetModel()
           <Annotation Term = ""Core.Revisions"" >
             <Collection>
               <Record>
-                <PropertyValue Property=""Version"" String=""2022-03-30""/>
+                <PropertyValue Property=""Version"" String=""2020-03-30""/>
```
xuzhg, line 115: 2022 is not good?
and the property is named "Version", but the value is a DateTime?

mikepizzo: Version is a string. In this example the string looks like a date, but it can be anything.

I made the string consistent across the example so that the test could validate that it got back the expected result.

---
In reply to: [468049254](https://github.com/OData/odata.net/pull/1849#discussion_r468049254) [](ancestors = 468049254)

Issue #1725, opened 2020-03-30
 ODataReader: high CPU usage and slow performance when reading larger OData payloads


korygin:
High CPU usage and slow performance when reading larger OData payloads.

### Assemblies affected

Microsoft.OData.Core 7.6.1

### Reproduce steps

During profiling of an OData client app we have noticed a significantly degraded performance and high CPU usage when reading larger OData payloads. I have attached a zip file with sample EDM model, payload and unit test to showcase the problem. Reading the ~ 500Kb resource set payload with 1000 resources in it takes approximately 3.5 - 4 secs. 

[ODataReaderPerformance.zip](https://github.com/OData/odata.net/files/4404214/ODataReaderPerformance.zip)

### Actual result

*What is actually happening.*

![image](https://user-images.githubusercontent.com/14077250/77930458-e90ff300-7278-11ea-9ce1-6e0b6d3fdd6e.png)

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

KenitoInc:
@korygin We are investigating this. We will get back later.

PR #1854, opened 2020-08-07
OData path fixes

marabooy:
### Issues

*This pull request fixes issue #1725.*

### Description

Updates the way OData Path and OData path extensions work to reduce the impact they have on the OData reader.

- [ ] *Test cases added*

### Additional work necessary

N/A

### DIFF
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
@@ -657,17 +657,17 @@ namespace Microsoft.OData.Evaluation
             segments = segments.GetRange(0, segmentsCount);
 
             // Remove the unnecessary type cast segment.
-            ODataPathSegment nextToLastSegment = segments.Last();
             int lastIndex = segments.Count - 1;
+            ODataPathSegment nextToLastSegment = segments[lastIndex];
             while (nextToLastSegment is TypeSegment)
             {
-                ODataPathSegment previousSegment = segments[segments.Count - 2];
+                ODataPathSegment previousSegment = segments[lastIndex - 1];
                 IEdmStructuredType owningType = previousSegment.TargetEdmType as IEdmStructuredType;
                 if (owningType != null && owningType.FindProperty(lastSegment.Identifier) != null)
                 {
                     segments.RemoveAt(lastIndex);
-                    nextToLastSegment = segments.Last();
-                    lastIndex = segments.Count - 1;
+                    --lastIndex;
+                    nextToLastSegment = segments[lastIndex];
                 }
                 else
                 {
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
@@ -644,31 +644,30 @@ namespace Microsoft.OData.Evaluation
         private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
         {
             ODataPath path = odataUri.Path;
-            // List<ODataPathSegment> segments = path.ToList();
             List<ODataPathSegment> segments = path.ToList();
-            // Reversing list so we can have better performance since List.Remove uses list.IndexOf internally which is faster if items are in-front of the list
-            segments.Reverse();
-            ODataPathSegment lastSegment = segments.First();
+            int segmentsCount = segments.Count - 1;
+            ODataPathSegment lastSegment = segments.Last();
             while (!(lastSegment is NavigationPropertySegment) && !(lastSegment is OperationSegment))
             {
-                segments.Remove(lastSegment);
-                lastSegment = segments.First();
+                segmentsCount--;
+                lastSegment = segments[segmentsCount];
             }
 
             // also removed the last navigation property segment
-            segments.Remove(lastSegment);
+            segments = segments.GetRange(0, segmentsCount);
 
             // Remove the unnecessary type cast segment.
             ODataPathSegment nextToLastSegment = segments.Last();
-
+            int lastIndex = segments.Count - 1;
             while (nextToLastSegment is TypeSegment)
             {
-                ODataPathSegment previousSegment = segments[1];
+                ODataPathSegment previousSegment = segments[segments.Count - 2];
                 IEdmStructuredType owningType = previousSegment.TargetEdmType as IEdmStructuredType;
-                if (owningType?.FindProperty(lastSegment.Identifier) != null)
+                if (owningType != null && owningType.FindProperty(lastSegment.Identifier) != null)
                 {
-                    segments.Remove(nextToLastSegment);
-                    nextToLastSegment = segments.First();
+                    segments.RemoveAt(lastIndex);
+                    nextToLastSegment = segments.Last();
+                    lastIndex = segments.Count - 1;
                 }
                 else
                 {
@@ -676,10 +675,9 @@ namespace Microsoft.OData.Evaluation
                 }
             }
 
-            segments.Reverse();
             // append each segment to base uri
             Uri uri = baseUri;
-            foreach (var segment in segments)
+            foreach (ODataPathSegment segment in segments)
             {
                 var keySegment = segment as KeySegment;
                 if (keySegment == null)
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
@@ -646,18 +646,12 @@ namespace Microsoft.OData.Evaluation
             ODataPath path = odataUri.Path;
             List<ODataPathSegment> segments = path.ToList();
             int lastIndex = segments.Count - 1;
-            ODataPathSegment lastSegment = segments.Last();
+            ODataPathSegment lastSegment = segments[lastIndex];
             while (!(lastSegment is NavigationPropertySegment) && !(lastSegment is OperationSegment))
             {
-                lastIndex--;
-                lastSegment = segments[lastIndex];
+                lastSegment = segments[lastIndex--];
             }
 
-            // also removed the last navigation property segment
-            segments = segments.GetRange(0, lastIndex);
-
-            // Remove the unnecessary type cast segment.
-            lastIndex = segments.Count - 1;
             lastSegment = segments[lastIndex];
             while (lastSegment is TypeSegment)
             {
@@ -665,15 +659,16 @@ namespace Microsoft.OData.Evaluation
                 IEdmStructuredType owningType = previousSegment.TargetEdmType as IEdmStructuredType;
                 if (owningType != null && owningType.FindProperty(lastSegment.Identifier) != null)
                 {
-                    segments.RemoveAt(lastIndex);
-                    lastIndex--;
-                    lastSegment = segments[lastIndex];
+                    lastSegment = segments[lastIndex--];
                 }
                 else
                 {
                     break;
                 }
             }
+
+            // trim all the un-needed segments 
+            segments = segments.GetRange(0, lastIndex);
 
             // append each segment to base uri
             Uri uri = baseUri;
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
@@ -645,21 +645,21 @@ namespace Microsoft.OData.Evaluation
         {
             ODataPath path = odataUri.Path;
             List<ODataPathSegment> segments = path.ToList();
-            int segmentsCount = segments.Count - 1;
+            int lastIndex = segments.Count - 1;
             ODataPathSegment lastSegment = segments.Last();
             while (!(lastSegment is NavigationPropertySegment) && !(lastSegment is OperationSegment))
             {
-                segmentsCount--;
-                lastSegment = segments[segmentsCount];
+                lastIndex--;
+                lastSegment = segments[lastIndex];
             }
 
             // also removed the last navigation property segment
-            segments = segments.GetRange(0, segmentsCount);
+            segments = segments.GetRange(0, lastIndex);
 
             // Remove the unnecessary type cast segment.
-            int lastIndex = segments.Count - 1;
-            ODataPathSegment nextToLastSegment = segments[lastIndex];
-            while (nextToLastSegment is TypeSegment)
+            lastIndex = segments.Count - 1;
+            lastSegment = segments[lastIndex];
+            while (lastSegment is TypeSegment)
             {
                 ODataPathSegment previousSegment = segments[lastIndex - 1];
                 IEdmStructuredType owningType = previousSegment.TargetEdmType as IEdmStructuredType;
@@ -667,7 +667,7 @@ namespace Microsoft.OData.Evaluation
                 {
                     segments.RemoveAt(lastIndex);
                     --lastIndex;
-                    nextToLastSegment = segments[lastIndex];
+                    lastSegment = segments[lastIndex];
                 }
                 else
                 {
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
@@ -644,12 +644,15 @@ namespace Microsoft.OData.Evaluation
         private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
         {
             ODataPath path = odataUri.Path;
+            // List<ODataPathSegment> segments = path.ToList();
             List<ODataPathSegment> segments = path.ToList();
-            ODataPathSegment lastSegment = segments.Last();
+            // Reversing list so we can have better performance since List.Remove uses list.IndexOf internally which is faster if items are in-front of the list
+            segments.Reverse();
+            ODataPathSegment lastSegment = segments.First();
             while (!(lastSegment is NavigationPropertySegment) && !(lastSegment is OperationSegment))
             {
                 segments.Remove(lastSegment);
-                lastSegment = segments.Last();
+                lastSegment = segments.First();
             }
 
             // also removed the last navigation property segment
@@ -657,14 +660,15 @@ namespace Microsoft.OData.Evaluation
 
             // Remove the unnecessary type cast segment.
             ODataPathSegment nextToLastSegment = segments.Last();
+
             while (nextToLastSegment is TypeSegment)
             {
-                ODataPathSegment previousSegment = segments[segments.Count - 2];
+                ODataPathSegment previousSegment = segments[1];
                 IEdmStructuredType owningType = previousSegment.TargetEdmType as IEdmStructuredType;
-                if (owningType != null && owningType.FindProperty(lastSegment.Identifier) != null)
+                if (owningType?.FindProperty(lastSegment.Identifier) != null)
                 {
                     segments.Remove(nextToLastSegment);
-                    nextToLastSegment = segments.Last();
+                    nextToLastSegment = segments.First();
                 }
                 else
                 {
@@ -672,9 +676,10 @@ namespace Microsoft.OData.Evaluation
                 }
             }
 
+            segments.Reverse();
             // append each segment to base uri
             Uri uri = baseUri;
-            foreach (ODataPathSegment segment in segments)
+            foreach (var segment in segments)
             {
                 var keySegment = segment as KeySegment;
                 if (keySegment == null)
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
@@ -649,17 +649,16 @@ namespace Microsoft.OData.Evaluation
             ODataPathSegment lastSegment = segments[lastIndex];
             while (!(lastSegment is NavigationPropertySegment) && !(lastSegment is OperationSegment))
             {
-                lastSegment = segments[lastIndex--];
+                lastSegment = segments[--lastIndex];
             }
 
-            lastSegment = segments[lastIndex];
             while (lastSegment is TypeSegment)
             {
                 ODataPathSegment previousSegment = segments[lastIndex - 1];
                 IEdmStructuredType owningType = previousSegment.TargetEdmType as IEdmStructuredType;
                 if (owningType != null && owningType.FindProperty(lastSegment.Identifier) != null)
                 {
-                    lastSegment = segments[lastIndex--];
+                    lastSegment = segments[--lastIndex];
                 }
                 else
                 {
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
@@ -666,7 +666,7 @@ namespace Microsoft.OData.Evaluation
                 if (owningType != null && owningType.FindProperty(lastSegment.Identifier) != null)
                 {
                     segments.RemoveAt(lastIndex);
-                    --lastIndex;
+                    lastIndex--;
                     lastSegment = segments[lastIndex];
                 }
                 else
--- a/src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
+++ b/src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
@@ -22,9 +22,9 @@ namespace Microsoft.OData.UriParser
     [SuppressMessage("Microsoft.Naming", "CA1710:IdentifiersShouldHaveCorrectSuffix", Justification = "ODataPathCollection just doesn't sound right")]
     public class ODataPath : IEnumerable<ODataPathSegment>
     {
-        /// <summary>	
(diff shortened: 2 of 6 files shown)


marabooy:
Initial contribution of Collections to the performance
image

User code hotspots before the changes (1000 Items)
image

After the optimizations 
image

Hotspots after the change for the same (1000 items)
image

KenitoInc:
@marabooy There is improvement in performance of some method calls e.g `ODataPath constructor` and `ExceptionUtils.CheckArgumentNotNull`. But also there has been degraded performance of `ODataPath.WalkWith`.

What is the general performance improvement?

marabooy:
@KenitoInc I will have to revisit the walk with as it seems to be just an foreach loop not sure why it degraded. Generally the whole reading took 4.5s on average and now it takes <2s on average.

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -142,23 +130,23 @@ public void WalkWith(PathSegmentHandler handler)
...
+            return !this.Where((t, i) => !t.Equals(other[i])).Any();
         }
 
         /// <summary>
         /// Add a segment to this path.
         /// </summary>
         /// <param name="newSegment">the segment to add</param>
         /// <exception cref="System.ArgumentNullException">Throws if the input newSegment is null.</exception>
-        internal void Add(ODataPathSegment newSegment)
+        internal new void Add(ODataPathSegment newSegment)
```
gathogojr, line 146: @marabooy Why did you remove `new` modifier from here? Given that its used to instruct the compiler to use the child class implementation, you could easily change behaviour altogether

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPathExtensions.cs
```diff
@@ -101,14 +101,16 @@ public static ODataPath AppendKeySegment(this ODataPath path, IEnumerable<KeyVal
             KeySegment keySegment = new KeySegment(keys, edmType, navigationSource);
             ODataPath newPath = handler.FirstPart;
             newPath.Add(keySegment);
-            foreach (var segment in handler.LastPart)
-            {
-                newPath.Add(segment);
-            }
-
+            AppendLastSegment(handler, newPath);
```
gathogojr, line 104: @marabooy By making this change, you effectively changed the logic here from `foreach (var segment in handler.LastPart)` to `foreach (var segment in handler.FirstPart)`

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/SemanticAst/ODataPathTests.cs
```diff
@@ -62,15 +62,24 @@ public void PathsShouldNotAllowSegmentsToBeNull()
         {
             Action createWithNull = () => new ODataPath((IEnumerable<ODataPathSegment>)null);
             Assert.Throws<ArgumentNullException>("segments", createWithNull);
+
```
gathogojr, line 65: nit: I assume this change wasn't intended

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/SemanticAst/ODataPathTests.cs
```diff
@@ -62,15 +62,24 @@ public void PathsShouldNotAllowSegmentsToBeNull()
...
+            List<ODataPathSegment> segments = new List<ODataPathSegment>(){
+                null
+            };
+
+            createWithNull = () => new ODataPath(segments);
+            Assert.Throws<ArgumentNullException>("segments", createWithNull);
         }
 
+
```
gathogojr, line 82: nit: I assume this change wasn't intended

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPathExtensions.cs
```diff
@@ -123,11 +125,7 @@ public static ODataPath TrimEndingKeySegment(this ODataPath path)
             path.WalkWith(typeHandler);
             typeHandler.FirstPart.WalkWith(keyHandler);
             ODataPath newPath = keyHandler.FirstPart;
-            foreach (var segment in typeHandler.LastPart)
-            {
-                newPath.Add(segment);
-            }
-
+            AppendLastSegment(typeHandler,newPath);
```
gathogojr, line 128:
```suggestion
            AppendLastSegment(typeHandler, newPath);

```

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -645,26 +645,29 @@ private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
...
             {
-                segments.Remove(lastSegment);
-                lastSegment = segments.Last();
+                segmentsCount--;
+                lastSegment = segments[segmentsCount];
             }
 
             // also removed the last navigation property segment
-            segments.Remove(lastSegment);
+            segments = segments.GetRange(0, segmentsCount);
```
gathogojr, line 657: @marabooy I wish there was a way the change starting from [this](https://github.com/OData/odata.net/pull/1854/files#diff-5c1c8c475ffa3bfbcec9695a7554881aR648) line to here could be isolated and unit tested. This is the stuff off-by-one bugs are made off - if you got your arithmetic wrong!

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -645,26 +645,29 @@ private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
...
 
             // Remove the unnecessary type cast segment.
             ODataPathSegment nextToLastSegment = segments.Last();
+            int lastIndex = segments.Count - 1;
```
gathogojr, line 661: @marabooy You could initialize `nextToLastSegment` on this line and forgo use of `.Last()` - unless there's no obvious benefit in doing that, i.e.
```
ODataPathSegment nextToLastSegment = segments[lastIndex];
```
Actually, why is the variable named `nextToLastSegment` yet its seems to me like it's assigned the last segment. Maybe it should be `typecastSegment`

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -645,26 +645,29 @@ private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
...
             while (nextToLastSegment is TypeSegment)
             {
                 ODataPathSegment previousSegment = segments[segments.Count - 2];
                 IEdmStructuredType owningType = previousSegment.TargetEdmType as IEdmStructuredType;
                 if (owningType != null && owningType.FindProperty(lastSegment.Identifier) != null)
                 {
-                    segments.Remove(nextToLastSegment);
+                    segments.RemoveAt(lastIndex);
                     nextToLastSegment = segments.Last();
+                    lastIndex = segments.Count - 1;
```
gathogojr, line 670: @marabooy [This](https://github.com/OData/odata.net/pull/1854/files#r470436103) comment could apply here too

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -20,28 +20,31 @@ namespace Microsoft.OData.UriParser
     /// A representation of the path portion of an OData URI which is made up of <see cref="ODataPathSegment"/>s.
     /// </summary>
     [SuppressMessage("Microsoft.Naming", "CA1710:IdentifiersShouldHaveCorrectSuffix", Justification = "ODataPathCollection just doesn't sound right")]
-    public class ODataPath : IEnumerable<ODataPathSegment>
+    public class ODataPath : List<ODataPathSegment>, IEnumerable<ODataPathSegment>
```
gathogojr, line 23: @marabooy Just curious, by having `OData` path inherit from `List<ODataPathSegment>` here, might we end up inadvertently surfacing behaviour available to `List<ODataPathSegment>` that is not applicable/relevant to `ODataPath`?

marabooy: For this maybe i could use a backing list inside to avoid polluting the api this way. 👍

mikepizzo:
@gathogojr -- what's the status of this PR? I'd love to get it merged once the feedback is addressed -- I'm particularly interested in understanding why OdataPath.WalkWith degraded, and if we can fix that without losing the overall improvements.

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -645,26 +645,29 @@ private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
...
             // also removed the last navigation property segment
-            segments.Remove(lastSegment);
+            segments = segments.GetRange(0, segmentsCount);
 
             // Remove the unnecessary type cast segment.
             ODataPathSegment nextToLastSegment = segments.Last();
+            int lastIndex = segments.Count - 1;
             while (nextToLastSegment is TypeSegment)
             {
                 ODataPathSegment previousSegment = segments[segments.Count - 2];
```
mikepizzo, line 664: >segments.Count [](start = 60, length = 14)

Can we use lastIndex-1 here?

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
mikepizzo, line 670: >segments.Count [](start = 32, length = 14)

can we just use lastIndex-1 and remove the .Count?

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
mikepizzo, line 664: >segments.Count - 2 [](start = 60, length = 18)

can we use lastIndex-1 and remove the .Count?

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -645,26 +645,29 @@ private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
...
                 IEdmStructuredType owningType = previousSegment.TargetEdmType as IEdmStructuredType;
                 if (owningType != null && owningType.FindProperty(lastSegment.Identifier) != null)
                 {
-                    segments.Remove(nextToLastSegment);
+                    segments.RemoveAt(lastIndex);
                     nextToLastSegment = segments.Last();
```
mikepizzo, line 669: >segments.Last() [](start = 40, length = 15)

is it more efficient to do segments[lastIndex]?

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
mikepizzo, line 661: >segments.Count - 1 [](start = 28, length = 18)

can we just use segmentsCount, rather than recomputing by calling .Count?

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -59,7 +62,7 @@ public ODataPathSegment FirstSegment
         {
             get
             {
-                return this.segments.Count == 0 ? null : this.segments[0];
+                return this.FirstOrDefault();
```
mikepizzo, line 65: >s.FirstOrDefault() [](start = 26, length = 18)

is this[0] more efficient than this.FirstOrDefault()?

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
marabooy: Let me evaluate this, my assumption is that that segments.Last() and segments[lastindex] should give use the same performance esp on lists.

mikepizzo: According to [this](https://stackoverflow.com/questions/1377864/what-is-the-performance-of-the-last-extension-method-for-listt), there is an optimization in .Last() that checks to see if it's a list, but that incurs a cast. Not a huge deal, but if you already have the lastIndex, seems it would be faster just to use it.

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
marabooy: I added a test to verify that the url is not impacted, earlier while working the wrong arithmetic could be observed on nested navigation properties where some of the segments repeated themselved. The test case will catch this if it happens again.

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
marabooy: reverted to using a backing list.

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -22,10 +22,10 @@ namespace Microsoft.OData.UriParser
     [SuppressMessage("Microsoft.Naming", "CA1710:IdentifiersShouldHaveCorrectSuffix", Justification = "ODataPathCollection just doesn't sound right")]
     public class ODataPath : IEnumerable<ODataPathSegment>
     {
-        /// <summary>
-        /// The segments that make up this path.
-        /// </summary>
-        private readonly IList<ODataPathSegment> segments;
+        /// <summary>	
+        /// The segments that make up this path.	
+        /// </summary>
```
KenitoInc, line 27: Remove unnecessary spaces

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -22,10 +22,10 @@ namespace Microsoft.OData.UriParser
...
     public class ODataPath : IEnumerable<ODataPathSegment>
     {
-        /// <summary>
-        /// The segments that make up this path.
-        /// </summary>
-        private readonly IList<ODataPathSegment> segments;
+        /// <summary>	
+        /// The segments that make up this path.	
+        /// </summary>	
+        private readonly List<ODataPathSegment> segments;
```
KenitoInc, line 28: Why did you replace the IList with a concrete type?

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -70,18 +71,17 @@ public ODataPathSegment LastSegment
...
+                return this.segments.LastOrDefault();
             }
         }
 
-        /// <summary>
-        /// Get the number of segments in this path.
-        /// </summary>
+        /// <summary>	
+        /// Get the number of segments in this path.	
+        /// </summary>
```
KenitoInc, line 80: Remove unnecessary spaces

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -90,10 +90,7 @@ internal IList<ODataPathSegment> Segments
             get { return this.segments; }
         }
 
-        /// <summary>
-        /// Get the segments enumerator
-        /// </summary>
-        /// <returns>The segments enumerator</returns>
+
```
KenitoInc, line 93: Since `GetEnumerator` is a public method, you shouldn't had deleted the `<summary>` and `<returns>` tags

KenitoInc (changes requested):
A few change requests

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
marabooy: List exposes extra methods that are more performant which is something I am relying on.

KenitoInc: #Resolved

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -90,7 +91,6 @@ internal IList<ODataPathSegment> Segments
             get { return this.segments; }
         }
 
-        /// <summary>
         /// Get the segments enumerator
```
KenitoInc, line 94: Why did you delete `/// <summary>` ?

KenitoInc (changes requested):

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -645,26 +645,29 @@ private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
...
             {
-                ODataPathSegment previousSegment = segments[segments.Count - 2];
+                ODataPathSegment previousSegment = segments[lastIndex - 1];
                 IEdmStructuredType owningType = previousSegment.TargetEdmType as IEdmStructuredType;
                 if (owningType != null && owningType.FindProperty(lastSegment.Identifier) != null)
                 {
-                    segments.Remove(nextToLastSegment);
-                    nextToLastSegment = segments.Last();
+                    segments.RemoveAt(lastIndex);
+                    --lastIndex;
```
gathogojr, line 669: @marabooy Any particular reason why you chose to use `--lastIndex` here but `lastIndex--` [here](https://github.com/OData/odata.net/pull/1854/files#diff-5c1c8c475ffa3bfbcec9695a7554881aR652)?

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -117,7 +118,7 @@ IEnumerator IEnumerable.GetEnumerator()
         [SuppressMessage("Microsoft.Design", "CA1011:ConsiderPassingBaseTypesAsParameters", Justification = "We would rather ship the PathSegmentTranslator so that its more extensible later")]
         public IEnumerable<T> WalkWith<T>(PathSegmentTranslator<T> translator)
         {
-            return this.segments.Select(segment => segment.TranslateWith(translator));
+            return this.Select(segment => segment.TranslateWith(translator));
```
gathogojr, line 121: Don't we need `return this.segments.Select(segment => segment.TranslateWith(translator));` here?

marabooy: Nice catch John, I had missed this when i was reverting from the original change with List

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -160,5 +162,14 @@ internal void Add(ODataPathSegment newSegment)
             ExceptionUtils.CheckArgumentNotNull(newSegment, "newSegment");
             this.segments.Add(newSegment);
         }
+
+        /// <summary>
+        /// Adds a range of segements to the current path
+        /// </summary>
+        /// <param name="oDataPath"></param>
+        public void AddRange(ODataPath oDataPath)
```
mikepizzo, line 170: >public [](start = 8, length = 6)

Should this be public or internal? Do we have customer scenarios for this?  If we start out making it internal we can make it public later, but we can't go the other way.

Also, if we change the public API we'll need to update the baseline in the E2E tests.

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -645,26 +645,29 @@ private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
...
             ODataPath path = odataUri.Path;
             List<ODataPathSegment> segments = path.ToList();
+            int lastIndex = segments.Count - 1;
             ODataPathSegment lastSegment = segments.Last();
             while (!(lastSegment is NavigationPropertySegment) && !(lastSegment is OperationSegment))
             {
-                segments.Remove(lastSegment);
-                lastSegment = segments.Last();
+                lastIndex--;
+                lastSegment = segments[lastIndex];
```
mikepizzo, line 653: >lastIndex [](start = 39, length = 9)

could make this --lastIndex and remove 652.

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -645,26 +645,29 @@ private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
...
+            while (lastSegment is TypeSegment)
             {
-                ODataPathSegment previousSegment = segments[segments.Count - 2];
+                ODataPathSegment previousSegment = segments[lastIndex - 1];
                 IEdmStructuredType owningType = previousSegment.TargetEdmType as IEdmStructuredType;
                 if (owningType != null && owningType.FindProperty(lastSegment.Identifier) != null)
                 {
-                    segments.Remove(nextToLastSegment);
-                    nextToLastSegment = segments.Last();
+                    segments.RemoveAt(lastIndex);
```
mikepizzo, line 668: >lastIndex [](start = 38, length = 9)

could make this lastIndex-- and remove 669.

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -35,13 +35,14 @@ public class ODataPath : IEnumerable<ODataPathSegment>
...
-            if (this.segments.Any(s => s == null))
+            this.segments = new List<ODataPathSegment>(segments);
+            if (this.segments.Contains(null))
             {
                 throw Error.ArgumentNull(nameof(segments));
             }
         }
 
+
         /// <summary>
```
mikepizzo, line 46: nit: extra blank line

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -127,8 +128,9 @@ public IEnumerable<T> WalkWith<T>(PathSegmentTranslator<T> translator)
         [SuppressMessage("Microsoft.Design", "CA1011:ConsiderPassingBaseTypesAsParameters", Justification = "We would rather ship the PathSegmentHandler so that its more extensible later")]
         public void WalkWith(PathSegmentHandler handler)
         {
-            foreach (ODataPathSegment segment in this.segments)
+            for (var index = 0; index < this.segments.Count; index++)
```
mikepizzo, line 131: >this.segments.Count [](start = 40, length = 19)

perhaps move the computation of the count outside of the loop?
  int segments = this.segments.Count;
  for (int index=0; index < segments; index++)
   ...

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -645,26 +645,29 @@ private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
...
             }
 
             // also removed the last navigation property segment
-            segments.Remove(lastSegment);
+            segments = segments.GetRange(0, lastIndex);
 
             // Remove the unnecessary type cast segment.
-            ODataPathSegment nextToLastSegment = segments.Last();
-            while (nextToLastSegment is TypeSegment)
+            lastIndex = segments.Count - 1;
```
mikepizzo, line 660: >segments.Count [](start = 24, length = 14)

do we need to recompute lastIndex? isn't it already segments.Count - 1 (appropriately adjusted for any nav prop/operation segments above)?

mikepizzo:
I really like where this PR is going. Just a few comments below, and I'd like to get it merged.

Are there patterns that you discovered that we could apply in other performance-critical areas?

marabooy:
@mikepizzo Not too sure if there is anything else except for whenever the virtual GetEnumerator gets invoked in most cases the performance drops drastically. Also bunching up additions and removals from the collections seem to take less of a performance hit which In my opinion could be attributded to the internal array resizing.

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -645,33 +645,31 @@ private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
...
             ODataPath path = odataUri.Path;
             List<ODataPathSegment> segments = path.ToList();
-            ODataPathSegment lastSegment = segments.Last();
+            int lastIndex = segments.Count - 1;
+            ODataPathSegment lastSegment = segments[lastIndex];
             while (!(lastSegment is NavigationPropertySegment) && !(lastSegment is OperationSegment))
             {
-                segments.Remove(lastSegment);
-                lastSegment = segments.Last();
+                lastSegment = segments[lastIndex--];
```
mikepizzo, line 652: >lastIndex-- [](start = 39, length = 11)

Shouldn't this be --lastIndex?  I.e., don't we need to decrement *before* using the value?

mikepizzo: I went ahead and made this change, as well as the similar change when looping through TypeSegments, and removed the redundant lastSegment=segments[lastIndex].

---
In reply to: [480401681](https://github.com/OData/odata.net/pull/1854#discussion_r480401681) [](ancestors = 480401681)

PR #1865, opened 2020-08-26
Fix Enum as string used with Has Issue

Sreejithpin:
### Issues
Fix Enum as string used with Has Issue

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/ODataUriResolverTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/ODataUriResolverTests.cs
@@ -78,14 +78,6 @@ namespace Microsoft.OData.Tests.UriParser.Metadata
             bin.Right.ShouldBeEnumNode(enumtypeRef.EnumDefinition(), "2");
         }
 
-        private static void VerifyEnumVsStringFilterExpressionforHas(FilterClause filter)
-        {
-            var enumtypeRef = new EdmEnumTypeReference(UriEdmHelpers.FindEnumTypeFromModel(HardCodedTestModel.TestModel, "Fully.Qualified.Namespace.ColorPattern"), true);
-            var bin = filter.Expression.ShouldBeBinaryOperatorNode(BinaryOperatorKind.Has);
-            bin.Left.ShouldBeSingleValuePropertyAccessQueryNode(HardCodedTestModel.GetPet2PetColorPatternProperty());
-            bin.Right.ShouldBeStringCompatibleEnumNode(enumtypeRef.EnumDefinition(), "2");
-        }
-
         private static void VerifyEnumVsStringFilterExpressionReverse(FilterClause filter)
         {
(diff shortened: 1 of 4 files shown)


mikepizzo:
>                 "MoonSet/'Blue'",

proper syntax should be "MoonSet/Blue (without the quotes). Is this validated anywhere?

---
Refers to: test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/ODataUriResolverTests.cs:235 in 45735c8. [](commit_id = 45735c8f36d95f9d5bd975c0b04e2157b0e76561, deletion_comment = False)

mikepizzo:
>                 "MoonSet2(id=1,color='Blue')",

does this succeed or return an error?

---
Refers to: test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/ODataUriResolverTests.cs:255 in 45735c8. [](commit_id = 45735c8f36d95f9d5bd975c0b04e2157b0e76561, deletion_comment = False)

mikepizzo:
>                 "MoonSet('Blue')",

does this succeed, or is it expected to return an error?

---
Refers to: test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/ODataUriResolverTests.cs:218 in 45735c8. [](commit_id = 45735c8f36d95f9d5bd975c0b04e2157b0e76561, deletion_comment = False)

mikepizzo:
Fix to the original issue looks good, but I'm curious whether the same issue happens with key values. are the tests no expecting unqualified enum values to be valid when used as key values?

xuzhg:
>                 "MoonSet('Blue')",

I think it is correct, because it plays like a string, and a string should have single quote wrapped? 

---
In reply to: [681187915](https://github.com/OData/odata.net/pull/1865#issuecomment-681187915) [](ancestors = 681187915)

---
Refers to: test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/ODataUriResolverTests.cs:218 in 45735c8. [](commit_id = 45735c8f36d95f9d5bd975c0b04e2157b0e76561, deletion_comment = False)

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/ODataUriResolverTests.cs
```diff
@@ -78,6 +78,14 @@ private static void VerifyEnumVsStringFilterExpression(FilterClause filter)
             bin.Right.ShouldBeEnumNode(enumtypeRef.EnumDefinition(), "2");
         }
 
+        private static void VerifyEnumVsStringFilterExpressionforHas(FilterClause filter)
```
xuzhg, line 81: If this method is only used at one place, maybe remove this function and add the codes into the test directly?

Sreejithpin:
> See my comments. Do we need a separate issue to track use of enum as string in keys?

Lets discuss in the call

Issue #609, opened 2016-06-29
.Contains() should be supported


nickkin-msft:
Any .Contains() in a .Where() clause, could easily be decomposed by the OData client's IQueryable expression tree visitor into a list of "ORed" conditionals, even if .Contains() is not strictly supported by OData directly with an equivalent URI.

This is at least, or especially, true if the list is a simple value type list rather than a list of complex.
### Assemblies affected

Microsoft.OData.Client v4.0.30319
Microsoft.OData.Core v4.0.30319
Microsoft.OData.Edm v4.0.30319
### Reproduce steps

Execute the following or equivalent against any OData endpoint using the standard T4 generated client.

var values = new string[] { "val1", "val2" };
odataClient.MyCollection.Where(x => list.Contains(x.prop))
### Expected result

The list.Contains() clause should be converted into a series of "ORed" conditionals in the generated URI which is fully supported by OData and allows easier use of the client. 
### Actual result

Depending on how you execute .Contains(), either NotSupportedException or a more confusing variant (see issues/608).

StevenLiekens:
I just ran into this, trying to filter an entity set by a collection of in-memory values.

Per the example:

``` csharp
var values = new string[] { "val1", "val2" };
odataClient.MyCollection.Where(x => list.Contains(x.prop))
```

My expectation was that this gets translated to _in_.

```
GET /odata/entities?$filter=x.prop in ('val1', 'val2')
```

Is is possible to add this behavior? Is there a workaround?

StevenLiekens:
There are two _Contains_ methods that result in different behavior:

- `ICollection<T>.Contains()` which is implemented by `List<T>`
- `Enumerable.Contains<TSource>(this IEnumerable<TSource> source, TSource value)` for arrays and other enumerables

The latter was easy to implement but _ICollection<T>.Contains()_ is not so easy to parse because the method itself isn't a generic method.

gathogojr:
Fixed by https://github.com/OData/odata.net/pull/1868

PR #1868, opened 2020-08-28
Add IN operator support to client

StevenLiekens:
### Issues

This pull request fixes issues #609.

### Description

This commit adds support for the IN operator to the Client project.
<https://docs.microsoft.com/en-us/odata/webapi/in-operator>

After this change, Client users will be able to use _Enumerable.Contains()_ in LINQ expressions.

``` csharp
var productNames = new[] { "Milk", "Cheese", "Donut" };
var products = ctx.CreateQuery<Product>().Where(product => productNames.Contains(product.Name)).ToList();
```

Which translates to **IN**.

```
/Products()?$filter=Name in ('Milk','Cheese','Donut')
```

- [x] *Test cases added*

### Additional work necessary

This commit only adds support for the extension method, i.e.

- `Enumerable.Contains<TSource>(this IEnumerable<TSource> source, TSource value)`

IF desirable, more work is required to add support for other _Contains_ methods, such as `ICollection<T>.Contains()`. The rationale for not implementing other _Contains_ methods is that you can always use _AsEnumerable()_ in client code as a workaround.

Additionally, this won't work with expressions that resolve to a collection, e.g.

```
~/Products?$filter=Name in RelevantProductNames
~/Products?$filter=ShipToAddress/CountryCode in MyShippers/Regions
~/Products?$filter=Name in Fully.Qualified.Namespace.MostPopularItemNames
```

I wouldn't know where to begin to make that work.


src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
```diff
@@ -372,6 +373,22 @@ internal override Expression VisitMethodCall(MethodCallExpression m)
                         this.builder.Append(UriHelper.FORWARDSLASH).Append(UriHelper.DOLLARSIGN).Append(UriHelper.COUNT);
                         return m;
                     }
+                    else if (sequenceMethod == SequenceMethod.Contains)
```
wachugamaina, line 376: The if on this method is way too big, perhaps it can be extracted into smaller specialized methods.

StevenLiekens: I'm not sure if that's a valid concern here? It's 8 lines of actual code and the rest are comments. If you look at the surrounding if-blocks, this if-block is actually one of the smaller blocks.

StevenLiekens:
Update: test cases added for String.Contains and Enumerable.Contains

marabooy (approved):

src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
StevenLiekens: I may have misunderstood. Did you mean that the entire method should be broken up? Because then I agree 100%. But still, I don't think it's up to me as a first-time contributor to refactor all of _VisitMethodCall_ so I'll mark this as resolved.

KenitoInc (approved):
LGTM
I would suggest you add a test case where you combine `contains` with other `clauses`.
e.g
`context.Books.Where(b => new[] { "Title 1", "Title 2" }.Contains(b.Title) && b.Price > 5);`

StevenLiekens:
Update:

- Added test for `Contains()` combined with another clause
- Added test for `Contains()` with a single item
- Added test for `Contains()` with an empty collection

What should be the expected behavior for the empty collection scenario?  
Right now, it translates to `$filter=Value in ()` which is meaningless.

ElizabethOkerio (approved):
LGTM

gathogojr (changes requested):
@StevenLiekens Thank you for your contribution. There are some commonly used collections that I believe we should support:
```c#
// 1.
var list = new List<string> { "Milk" };
context.Products.Where(d => list.Contains(d.Name));
// &
context.Products.Where(d => new List<string> { "Milk" }.Contains(d.Name));

// 2.
var collection = new Collection<string> { "Milk" };
context.Products.Where(d => collection.Contains(d.Name));
// &
context.Products.Where(d => new Collection<string> { "Milk" }.Contains(d.Name));
```

In addition, note that the following request will result into an error
```
GET http://ServiceRoot/Products?$filter=Name equals ()
```
Error Message:
> The query specified in the URI is not valid. The right operand for the IN operation is not a collection value. IN operations require the left operand to be a single value and the right operand to be a collection value.

For this reason, I believe the right thing to do when the collection/array is empty is leave out the `$filter` query option altogether if `Contains` expression was the only predicate in the `Where` expression, or, to leave out the `contains` clause from the `$filter` expression if there were other predicates.

gathogojr:
@StevenLiekens My argument for supporting the above scenarios is that I don't believe it'd take much effort to add that. Plus it'd improve user experience since its intuitive. Requiring users to use `.AsEnumerable()` is much less intuitive

gathogojr:
Just to note also that we could support the following scenarios:

```c#
// 1.
public string[] GetNames()
{
	return new [] { "Milk" };
}
context.Products.Where(d => GetNames().Contains(d.Name));

// 2.
public List<string> GetNames()
{
	return new List<string> { "Milk" };
}
context.Products.Where(d => GetNames().Contains(d.Name));

// 3.
public List<string> GetNames()
{
	return new Collection<string> { "Milk" };
}
context.Products.Where(d => GetNames().Contains(d.Name));
```

But I'd agree with you that it'd make sense to leave this for the future.

StevenLiekens:
@gathogojr Thanks for the comprehensive review. I'm weeding through some of these points to see if I can implement them.

---

RE: 

> There are some commonly used collections that I believe we should support:

I agree and I did make an attempt to add them, but I got stuck. The code uses generic method definitions to map method signatures to a member of the `SequenceMethod` enum.

https://github.com/OData/odata.net/blob/087d9a8647fe9df75da6052968f1c92b8c8cafc7/src/Microsoft.OData.Client/ALinq/ReflectionUtil.cs#L185-L186

The problem is that `ICollection<T>.Contains()` is not a generic method: the method definition changes according to the type of `T` in `ICollection<T>`:

One possible solution is to add maps for all the well-known primitive types. This way I predict I can make it work for primitive types, without adding more special cases to the code. What do you think?

``` csharp
map.Add(@"Contains(System.String)->Boolean", SequenceMethod.Contains);
map.Add(@"Contains(System.Int32)->Boolean", SequenceMethod.Contains);
map.Add(@"Contains(System.Int64)->Boolean", SequenceMethod.Contains);
// etc
```

---

RE:

> For this reason, I believe the right thing to do when the collection/array is empty is leave out the `$filter` query option altogether

I don't think I can do that without enumerating collections twice. Once to count the number of elements, and again to build the actual URI. Is it okay to assume that re-enumerating will be cheap and won't crash?

gathogojr:
> The problem is that ICollection<T>.Contains() is not a generic method: the method definition changes according to the type of T in ICollection<T>:
>
> One possible solution is to add maps for all the well-known primitive types. This way I predict I can make it work for collections of primitive types, without adding more special cases to the code. What do you think?

I agree with your reasoning here. Our current mapping for caters for sequence methods that act on `IQueryable` and `IEnumerable` types and we'd need to think more deliberately before introducing support for `ICollection` and its derived types.

gathogojr:
> I don't think I can do that without enumerating collections twice. Once to count the number of elements, and again to build the actual URI. Is it okay to assume that re-enumerating will be cheap and won't crash?

One way or the other, we'll need to handle the empty collection scenario since the expression generated is invalid and fails when executed against the service. You could use `.FirstOrDefault()` instead of `.Count()` to determine if the collection has at least one element

src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
```diff
@@ -372,6 +373,22 @@ internal override Expression VisitMethodCall(MethodCallExpression m)
                         this.builder.Append(UriHelper.FORWARDSLASH).Append(UriHelper.DOLLARSIGN).Append(UriHelper.COUNT);
                         return m;
                     }
+                    else if (sequenceMethod == SequenceMethod.Contains)
+                    {
+                        // First argument is the collection expression
+                        // Second argument is the value expression
+                        // Note that arguments must be reordered for the IN operator
+                        // e.g. ctx.CreateQuery<Product>("Products").Where(p => (new [] { "Milk", "Cheese", "Donut"}).Contains(p.Name))
```
xuzhg, line 381: >(p => (new [] { "Milk", "Cheese", "Donut"}).Contains(p.Name)) [](start = 74, length = 61)

can we do:

Where(p => p.CollectionProperties.Contains('abc'))?

test/FunctionalTests/Microsoft.OData.Client.Tests/DataServiceQueryProviderTests.cs
```diff
@@ -0,0 +1,121 @@
...
+
+            // Assert
+            Assert.Equal(@"http://root/Products?$filter=Name in ('Milk','Flour') and Price lt 5", queryComponents.Uri.ToString());
+        }
+
+        #endregion
+
+        [EntityType]
+        [Key(nameof(Id))]
+        private class Product
```
mikepizzo, line 112: >Product [](start = 22, length = 7)

to Sam's comment, can you add a collection-valued property (i.e., public string[] comments {get; set;} and then add a test for:
            var products = dsc.CreateQuery<Product>("Products")
                .Where(product => comments.Contains("buy");

and:
            var products = dsc.CreateQuery<Product>("Products")
                .Where(product => comments.Contains(product.Name);

src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
mikepizzo: See my suggested added test case...

---
In reply to: [539597874](https://github.com/OData/odata.net/pull/1868#discussion_r539597874) [](ancestors = 539597874)

src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
mikepizzo: Agree.

---
In reply to: [498118908](https://github.com/OData/odata.net/pull/1868#discussion_r498118908) [](ancestors = 498118908)

src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
```diff
@@ -508,6 +525,23 @@ internal override Expression VisitConstant(ConstantExpression c)
                 ODataEnumValue enumValue = new ODataEnumValue(memberValue, typeNameInEdm ?? typeAnnotation.ElementTypeName);
                 result = ODataUriUtils.ConvertToUriLiteral(enumValue, CommonUtil.ConvertToODataVersion(this.uriVersion), null);
             }
+            else if (m != null && ReflectionUtil.IsSequenceMethod(m.Method, SequenceMethod.Contains))
+            {
+                var csvBuilder = new StringBuilder();
```
mikepizzo, line 530: >var [](start = 16, length = 3)

In this library, we generally prefer to use actual types instead of "var" as it tends to increase readability and validate expectations.  Could you please replace the occurances of "var" with the actual type?

mikepizzo:
According to the OData ABNF ([http://docs.oasis-open.org/odata/odata/v4.01/os/abnf/odata-abnf-construction-rules.txt](http://docs.oasis-open.org/odata/odata/v4.01/os/abnf/odata-abnf-construction-rules.txt)), this should be an error:

inExpr = RWS "in" RWS ( listExpr / commonExpr )
listExpr  = OPEN BWS primitiveLiteral BWS *( COMMA BWS primitiveLiteral BWS ) CLOSE


---
In reply to: [706676144](https://github.com/OData/odata.net/pull/1868#issuecomment-706676144) [](ancestors = 706676144)

mikepizzo:
Thanks for your contributions @StevenLiekens -- this will be a great addition for OData Client.

We are planning to do a release this week (Wed/Thurs). It looks like there are just a few comments left to resolve -- If you can get those addressed we'd love to include it in this release.

Thanks!

test/FunctionalTests/Microsoft.OData.Client.Tests/DataServiceQueryProviderTests.cs
StevenLiekens: Oof... I'm having a hard time with this one.

Before my change, that would just throw.

> System.NotSupportedException
> The method 'Contains' is not supported.

After my change, it generates nonsensical URIs or throws more obscure exceptions. 😅

StevenLiekens:
> According to the OData ABNF (http://docs.oasis-open.org/odata/odata/v4.01/os/abnf/odata-abnf-construction-rules.txt), this should be an error:
> 
> inExpr = RWS "in" RWS ( listExpr / commonExpr )
> listExpr = OPEN BWS primitiveLiteral BWS *( COMMA BWS primitiveLiteral BWS ) CLOSE
> 
> In reply to: [706676144](https://github.com/OData/odata.net/pull/1868#issuecomment-706676144) [](ancestors = 706676144)

There is a suggestion that such expressions should not be translated (or partially skipped if it's a multi-part filter expression). Can I just throw InvalidOperationException instead? That seems a bit easier, and possibly less surprising behavior.

mikepizzo:
Yes; please raise an exception, don't silently ignore.  Ignoring a predicate is basically data corruption as it returns data that should not have been in the requested set and can lead to bad decisions.

---
In reply to: [745631134](https://github.com/OData/odata.net/pull/1868#issuecomment-745631134) [](ancestors = 745631134)

src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
mikepizzo: Would be great to refactor if, but not something to address in this PR.

---
In reply to: [485146911](https://github.com/OData/odata.net/pull/1868#discussion_r485146911) [](ancestors = 485146911)

test/FunctionalTests/Microsoft.OData.Client.Tests/DataServiceQueryProviderTests.cs
mikepizzo: What's the status of these cases?

Preference would be to support these, but min bar would be to raise a useful exception. I don't really want to ship generating invalid URIs...

---
In reply to: [543716880](https://github.com/OData/odata.net/pull/1868#discussion_r543716880) [](ancestors = 543716880)

mikepizzo:
Looks like this was resolved without enumerating twice by checking the length of the generated list.

---
In reply to: [727932429](https://github.com/OData/odata.net/pull/1868#issuecomment-727932429) [](ancestors = 727932429)

StevenLiekens:
@mikepizzo I tried my best to restore the old behavior Contains() with collection-valued properties, while keeping support for Contains() on static collections.

- Contains() on a static collection translates to IN
- Contains() on a collection property throws NotSupportedException

src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
StevenLiekens: I added the suggested test cases but they simply reflect the old behavior, because I never meant to add support for these type of expressions.

test/FunctionalTests/Microsoft.OData.Client.Tests/DataServiceQueryProviderTests.cs
StevenLiekens: I added the suggested test cases but they simply reflect the old behavior, because I never meant to add support for these type of expressions.

src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
mikepizzo: fair enough -- we'll create a work item to add the other form.

---
In reply to: [544603939](https://github.com/OData/odata.net/pull/1868#discussion_r544603939) [](ancestors = 544603939)

mikepizzo (approved):
:shipit:

mikepizzo:
We don't want to silently ignore the expression as that would return the wrong results (which is worse than an error as it could lead to making a bad choice...)

---
In reply to: [520431031](https://github.com/OData/odata.net/pull/1868#issuecomment-520431031) [](ancestors = 520431031)

mikepizzo:
Makes sense - hopefully we can pick up the additional scenario next time around.

I've merged your PR and it will be in this week's build -- Thanks for all of your work getting this in!

PR #1875, opened 2020-09-07
Group by on a date(System.DateTime) results in an exception

ElizabethOkerio:
### Issues

*This pull request fixes issue #1874 *

### Description

OData doesn't include DateTime as a primitive type.  If a user defines a model as 
<pre>
public class Customer
{
    public int Id { get; set; }
    public DateTime Birthday { get; set; }
}
</pre>

then uses the <code> AsDate()</code> method to translate the DateTime field to <code> Edm.Date</code> as explained here https://docs.microsoft.com/en-us/odata/webapi/date-timeofday-with-entity-framework,<code> Microsoft.OData.Core</code> throws an exception that the primitive type (System.DateTime) is unsupported. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
+++ b/src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
@@ -601,6 +601,11 @@ namespace Microsoft.OData.Metadata
                 return true;
             }
 
+            if (clrType == typeof(DateTime))
+            {
+                return PrimitiveTypeReferenceMap.ContainsKey(typeof(Date));
+            }
+
             return PrimitiveTypeReferenceMap.ContainsKey(clrType) || typeof(ISpatial).IsAssignableFrom(clrType);
         }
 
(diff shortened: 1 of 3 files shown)


paulodero (commented):
Any test cases added for this fix?

src/Microsoft.OData.Core/Metadata/EdmLibraryExtensions.cs
```diff
@@ -601,6 +601,11 @@ internal static bool IsPrimitiveType(Type clrType)
                 return true;
             }
 
+            if (clrType == typeof(System.DateTime))
```
paulodero, line 604: Any reason to use a fully qualified name here?

ElizabethOkerio: @paulodero what is the implication of using a fully qualified name?

paulodero: @ElizabethOkerio Going by [C# naming conventions](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/inside-a-program/coding-conventions#naming-conventions) you do not have to use `FQN` if the namespace is already imported. In this case `System `is already imported and its redundant to do this. You only use FQN in simple code that do not include `using directives` or when you want to resolve namespace conflict.

Issue #1878, opened 2020-09-08
Downgrade minimum required version of System.Text.Json to 4.0


mikepizzo:
### Overview 
When we added support to EdmLib for JSON Metadata serialization, we referenced the System.Text.Json 4.0.1.1 library. The default .NETCore app application references System.Text.Json 4.0, so when a developer creates a .NET Core application that references ODL they will get an error that there is a version mismatch.

This is fixed by updating the nuget reference to System.Text.Json for the project, but that's a poor developer experience.

By downgrading EdmLib dependency to align with the default 4.0.1.0 version we can avoid that poor experience.

### Steps to repo:
Create a new .NET Core console application.
Add a nuget reference to Microsoft.OData.Edm 7.7.1.
Enter the following:
~~~c#
            IEdmModel model;
            IEnumerable<EdmError> errors;
            CsdlReader.TryParse(XmlReader.Create("myMetadata.xml"), out model, out errors);
~~~
Attempt to comple

### Expected Result:
No Error.

Actual Result:
Error CS1705:	Assembly 'Microsoft.OData.Edm' with identity 'Microsoft.OData.Edm, Version=7.7.1.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' uses 'System.Text.Json, Version=4.0.1.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' which has a higher version than referenced assembly 'System.Text.Json' with identity 'System.Text.Json, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'

xuzhg:
Fixed and merged in 7.7.2 version.

jmajoor:
Not fixed in 7.7.2:
Error	CS1705	Assembly 'Microsoft.OData.Edm' with identity 'Microsoft.OData.Edm, Version=7.7.2.0, Culture=neutral, PublicKeyToken=31bf3856ad364e35' uses 'System.Text.Json, Version=4.0.1.2, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51' which has a higher version than referenced assembly 'System.Text.Json' with identity 'System.Text.Json, Version=4.0.1.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'

xuzhg:
> Not fixed in 7.7.2:
[...]

PR #1880, opened 2020-09-10
downgrade the System.Text.Json package reference version

xuzhg:
### Issues

*This pull request fixes issue #1878.*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Microsoft.OData.Edm.csproj
+++ b/src/Microsoft.OData.Edm/Microsoft.OData.Edm.csproj
@@ -94,9 +94,7 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="System.Text.Json">
-      <Version>4.7.1</Version>
-    </PackageReference>
+    <PackageReference Include="System.Text.Json" Version="4.*" />
   </ItemGroup>
 
   <ItemGroup>


habbes (approved):

marabooy (approved):

src/Microsoft.OData.Edm/Microsoft.OData.Edm.csproj
```diff
@@ -94,9 +94,7 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="System.Text.Json">
-      <Version>4.7.1</Version>
-    </PackageReference>
+    <PackageReference Include="System.Text.Json" Version="4.*" />
```
gathogojr, line 97: @xuzhg Curious how `Version=4.*` is interpreted. Wouldn't 4.7.1 be validated against 4.*?

xuzhg: Here's the context:

Floating versions
A floating dependency version is specified with the * character. For example, 6.0.*. This version specification says "use the latest 6.0.x version"; 4.* means "use the latest 4.x version." Using a floating version reduces changes to the project file, while keeping up to date with the latest version of a dependency.

When using a floating version, NuGet resolves the highest version of a package that matches the version pattern, for example 6.0.* gets the highest version of a package that starts with 6.0:

Choosing version 6.0.1 when a floating version 6.0.* is requested

xuzhg: @gathogojr Besides, I use this PR and output a nightly, That nightly can be used directly without the build error.

gathogojr: @xuzhg Got it. So having the version pinned to 4.7.1 was the issue

gathogojr (approved):

gathogojr:
@xuzhg Any idea why ADO is reporting building failure for [OData-master-pipeline-Rolling] in spite of this:
image

habbes (approved):

habbes:
@xuzhg are the builds failing as a result of this change?

xuzhg:
> @xuzhg are the builds failing as a result of this change?

No, i run the rolling manually, it can pass. See https://identitydivision.visualstudio.com/OData/_build/results?buildId=621841&view=results

Issue #1885, opened 2020-09-16
Performance: OData.metadata=full is very CPU intensive


joaocpaiva:
Recently found a large increase in CPU in one of our DCs and root caused that to requests that accept odata.metadata=full.

The number of requests with odata.metadata=full increased from 1.13% to 3.4% and as result, the CPU cost of Microsoft.OData.ODataWriterCore!WriteEnd jumped exponentially from ~1.3% to ~13.5%. This is a huge increase.

If more clients started requesting odata.metadata=full that would quickly blow up the service costs.

### Assemblies affected
Microsoft.OData.ODataWriterCore.dll 7.7.1

### Reproduce steps
Multiple requests with odata.metadata=full. Incoming URI and other details provided offline.

### Expected result
CPU cost should grow linearly, in the worst case.

### Actual result
CPU cost grows exponentially, which suggests opportunity to significantly optimize this code path.

![image](https://user-images.githubusercontent.com/24846248/93276758-50f9bd80-f775-11ea-8a10-856980dff3b2.png)

gathogojr:
Fixed by [1888](https://github.com/OData/odata.net/pull/1888)

PR #1888, opened 2020-09-19
Performance fixes

Sreejithpin:
### Issues
Performance fixes
*This pull request fixes issue #1885.*

### Description

- [ ] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -175,23 +197,27 @@ public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(string qua
         {
             IEnumerable<IEdmOperation> enumerable = this.FindDeclaredOperations(qualifiedName);
```
mikepizzo, line 198: >FindDeclaredOperations [](start = 57, length = 22)

Rather than getting operations based on name, and filtering by binding type, now that we are caching the declared operations by binding type, we should call that method and filter the results by name.

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -154,13 +160,29 @@ public IEnumerable<IEdmOperation> FindDeclaredOperations(string qualifiedName)
...
+
+                foreach (IList<IEdmOperation> operations in this.functionDictionary.Values)
                 {
-                    yield return operation;
+                    for (int i = 0; i < operations.Count; i++)
+                    {
+                        if (!operationList.Contains(operations[i]) && operations[i].HasEquivalentBindingType(bindingType))
+                        {
+                            bindableOperations.Add(operations[i]);
+                            operationList.Add(operations[i]);
```
mikepizzo, line 177: Move this check outside of loop:

    if (operationList.TryAdd(operations[i])
    {
        for (int i = 0; i < operations.Count; i++)
        {
            if (!operationList.Contains(operations[i]) && operations[i].HasEquivalentBindingType(bindingType))
            {
                bindableOperations.Add(operations[i]);
             }
    }

src/Microsoft.OData.Core/Metadata/MetadataUtils.cs
```diff
@@ -174,22 +174,7 @@ internal static IList<IEdmOperation> CalculateBindableOperationsForType(IEdmType
             List<IEdmOperation> operationsFound = new List<IEdmOperation>();
             foreach (IEdmOperation operation in operations)
```
mikepizzo, line 175: First, check to see if the returned operations is an IList<IEdmOperation> (as it will be in most cases.  If so, return it directly, rather than creating a copy of the list to return.

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
```diff
@@ -15,6 +15,15 @@ namespace Microsoft.OData.Edm
     /// </summary>
     public static class EdmTypeSemantics
     {
+
+        private static IDictionary<IEdmType, IDictionary<IEdmType, bool>> typeDictionaryCache;
+
+#pragma warning disable CA1810 // Initialize reference type static fields inline
+        static EdmTypeSemantics()
```
mikepizzo, line 22: why is this done this way, rather than inline?

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
```diff
@@ -15,6 +15,15 @@ namespace Microsoft.OData.Edm
     /// </summary>
     public static class EdmTypeSemantics
     {
+
+        private static IDictionary<IEdmType, IDictionary<IEdmType, bool>> typeDictionaryCache;
```
mikepizzo, line 19: >IDictionary<IEdmType, IDictionary<IEdmType, bool>> [](start = 23, length = 50)

why is this not just an IDictionary<IEdmType, HashSet<IEdmType>> where the HashSet contains all of the types that the first type is or inherits from?

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -23,6 +23,11 @@ public abstract class EdmModelBase : EdmElement, IEdmModel
         private readonly Dictionary<string, IEdmTerm> termDictionary = new Dictionary<string, IEdmTerm>();
         private readonly Dictionary<string, IList<IEdmOperation>> functionDictionary = new Dictionary<string, IList<IEdmOperation>>();
 
+        /// <summary>
+        /// Cache of operations that are bindable to entity types.
+        /// </summary>
+        private readonly Dictionary<string, IList<IEdmOperation>> bindableOperationsCache;
```
mikepizzo, line 29: >Dictionary [](start = 25, length = 10)

If we are adding to this dynamically, based on usage of the model, we probably need to make it a concurrent dictionary.  The other dictionaries don't need to be concurrent because they are populated before the model is consumed.

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
mikepizzo, line 19: >IDictionary [](start = 23, length = 11)

Probably needs to be a concurrent dictionary.

src/Microsoft.OData.Edm/ReferenceEqualityComparer.cs
```diff
@@ -0,0 +1,86 @@
+//---------------------------------------------------------------------
+// <copyright file="ReferenceEqualityComparer.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
```
mikepizzo, line 4: is this class used anywhere?  how does it differ from the default ReferenceEquityComparer class?

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
Sreejithpin: Sorry didnt get that, use tryadd outside loop? operations[i] will be there only inside loop right? or am i missing something

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
Sreejithpin: For some it may not be inheritable , for some it is. we can only have hashset contain either. I we put all the edmtypes which is or can be inherited in the hashset, and if a type is not in hashset it could either mean it is not processed or it could be its not inheritable , so we have to go through that again and again.

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
mikepizzo: Sorry; typo. I meant: operationList.TryAdd(operations):

if (operationList.TryAdd(operations))
{
    for (int i = 0; i < operations.Count; i++)
    {
        if (!operationList.Contains(operations[i]) && operations[i].HasEquivalentBindingType(bindingType))
        {
            bindableOperations.Add(operations[i]);
         }
} 

I believe this has the same effect as the previous Distinct().



---
In reply to: [491743298](https://github.com/OData/odata.net/pull/1888#discussion_r491743298) [](ancestors = 491743298)

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
mikepizzo: Right; I'm not suggesting we lazily load which types each type does and does not inherit from -- when asked if type a inherits from type b, we should should load all of the types that type a inherits from and cache that for type a.  Then, if we have an entry for type a, we can just see if type b is in that list.  Type hierarchies are not likely to be deep, so computing the inherited types to the base type should be quick. In fact, for each base type we should check to see if we already have the base types and populate them recursively. 

---
In reply to: [491743487](https://github.com/OData/odata.net/pull/1888#discussion_r491743487) [](ancestors = 491743487)

src/Microsoft.OData.Edm/ReferenceEqualityComparer.cs
Sreejithpin: This was in existing code

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
Sreejithpin: I guess the multithreading scenario is handled using ReferenceEquityComparer here. So do we need to use Concurrent dict, or rather just use concurrent dict and remove ReferenceEquityComparer unless there is some other intention

mikepizzo:
>         private readonly Dictionary<IEdmType, IList<IEdmOperation>> bindableOperationsCache;

should this be removed now? Is it still used anywhere?

---
Refers to: src/Microsoft.OData.Core/Evaluation/ODataMetadataContext.cs:83 in c1bf887. [](commit_id = c1bf887743ccaee7eea930666d038daab7dd9989, deletion_comment = False)

mikepizzo:
>             namedDefinition = collectionType.ElementType.Definition as IEdmSchemaElement;

can we check type.TypeKind on any of these before doing casts?

---
Refers to: src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs:1610 in c1bf887. [](commit_id = c1bf887743ccaee7eea930666d038daab7dd9989, deletion_comment = False)

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
```diff
@@ -1217,17 +1222,29 @@ public static EdmPrimitiveTypeKind PrimitiveKind(this IEdmTypeReference type)
         /// <returns>True if and only if the type inherits from the potential base type.</returns>
         public static bool InheritsFrom(this IEdmStructuredType type, IEdmStructuredType potentialBaseType)
         {
+            HashSet<string> lstTypes;
```
mikepizzo, line 1225: >lstTypes [](start = 28, length = 8)

pick a more descriptive name -- isn't this "baseTypes"?

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
```diff
@@ -15,6 +17,9 @@ namespace Microsoft.OData.Edm
     /// </summary>
     public static class EdmTypeSemantics
     {
+
+        private static ConcurrentDictionary<string, HashSet<string>> typeDictionaryCache = new ConcurrentDictionary<string, HashSet<string>>();
```
mikepizzo, line 21: >typeDictionaryCache [](start = 69, length = 19)

pick a more descriptive name --- what types of types are in this dictionary? what's it used for?  maybe something like "baseTypeCache?

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
```diff
@@ -1217,17 +1222,29 @@ public static EdmPrimitiveTypeKind PrimitiveKind(this IEdmTypeReference type)
...
-                    return true;
+                    lstTypes.Add(type.FullTypeName());
                 }
             }
             while (type != null);
 
-            return false;
+            typeDictionaryCache.TryAdd(fullName, lstTypes);
+
+            return lstTypes.Contains(potentialBaseType.FullTypeName());
```
mikepizzo, line 1247: I would factor out a "GetBaseTypes" static function that, given a type, returns a hashset of base types.  Then, in line 1229, if the value is not in the cache in line 1228, call this function.  Then, when you exit the check, you know you have the hashset (either from the cache or just created) and you can use it to see if the base type is in the list.

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -154,13 +161,31 @@ public IEnumerable<IEdmOperation> FindDeclaredOperations(string qualifiedName)
...
-            foreach (IEnumerable<IEdmOperation> operations in this.functionDictionary.Values.Distinct())
+            IList<IEdmOperation> bindableOperations;
+
+            if (!this.bindableOperationsCache.TryGetValue(bindingType.FullTypeName(), out bindableOperations))
             {
-                foreach (IEdmOperation operation in operations.Where(o => o.HasEquivalentBindingType(bindingType)))
+                HashSet<IList<IEdmOperation>> operationList = new HashSet<IList<IEdmOperation>>();
+                bindableOperations = new List<IEdmOperation>();
+
+                foreach (IList<IEdmOperation> operations in this.functionDictionary.Values)
```
mikepizzo, line 171: you know that this.functionDictionary is an IList, so rather than foreaching over the enumerator, just write a loop that indexes the values by ordinal.

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -154,13 +161,31 @@ public IEnumerable<IEdmOperation> FindDeclaredOperations(string qualifiedName)
...
                 {
-                    yield return operation;
+                    if (operationList.Add(operations))
```
mikepizzo, line 173: >if (operationList.Add(operations)) [](start = 20, length = 34)

maybe add a comment that this is in order to ensure the list of operations is unique (do we know, under what conditions, it will not be unique?)

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -173,25 +198,21 @@ public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(IEdmType b
         /// </returns>
         public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(string qualifiedName, IEdmType bindingType)
         {
-            IEnumerable<IEdmOperation> enumerable = this.FindDeclaredOperations(qualifiedName);
-            IList<IEdmOperation> operations = enumerable as IList<IEdmOperation>;
+            IList<IEdmOperation> operations = this.FindDeclaredBoundOperations(bindingType) as IList<IEdmOperation>;
```
mikepizzo, line 201: > as IList<IEdmOperation> [](start = 91, length = 24)

since FindDeclaredBoundOperations is virtual, probably can't assume it is an IList -- instead, try casting and, if that fails, fall back to enumerable logic.

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -173,25 +198,21 @@ public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(IEdmType b
         /// </returns>
         public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(string qualifiedName, IEdmType bindingType)
         {
-            IEnumerable<IEdmOperation> enumerable = this.FindDeclaredOperations(qualifiedName);
-            IList<IEdmOperation> operations = enumerable as IList<IEdmOperation>;
+            IList<IEdmOperation> operations = this.FindDeclaredBoundOperations(bindingType) as IList<IEdmOperation>;
+            IList<IEdmOperation> matchedOperation = new List<IEdmOperation>();
```
mikepizzo, line 202: >new List<IEdmOperation>(); [](start = 52, length = 26)

rather than create the list here (and allocate the default size which may not be used in the common case where there are no bound operations) allocate the list after checking if operations != null, and add an else{return new IEdmOperation[]{};}

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -173,25 +198,21 @@ public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(IEdmType b
...
+
             if (operations != null)
-            {
-                IList<IEdmOperation> matchedOperation = new List<IEdmOperation>();
+            {                
                 for (int i = 0; i < operations.Count; i++)
                 {
-                    if (operations[i].HasEquivalentBindingType(bindingType))
+                    if ( string.Equals(operations[i].FullName(),qualifiedName,System.StringComparison.Ordinal) )
```
mikepizzo, line 208: >string.Equals(operations[i].FullName(),qualifiedName,System.StringComparison.Ordinal) [](start = 25, length = 85)

space after commas. no space before "string.Equals"

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
mikepizzo, line 202: >matchedOperation [](start = 33, length = 16)

should be plural: "matchedOperations"

mikepizzo (changes requested):
:clock1:

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
```diff
@@ -1216,18 +1221,33 @@ public static EdmPrimitiveTypeKind PrimitiveKind(this IEdmTypeReference type)
         /// <param name="potentialBaseType">The potential base type of the type being tested.</param>
         /// <returns>True if and only if the type inherits from the potential base type.</returns>
         public static bool InheritsFrom(this IEdmStructuredType type, IEdmStructuredType potentialBaseType)
+        {
+            HashSet<string> baseTypes;
+            string fullName = type.FullTypeName();
+
+            if (!baseTypeCache.TryGetValue(fullName, out baseTypes))
+            {
+                baseTypes = new HashSet<string>();
```
mikepizzo, line 1230: why pass in the HashSet rather than creating it in GetBaseTypes and returning it?

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
```diff
@@ -1216,18 +1221,33 @@ public static EdmPrimitiveTypeKind PrimitiveKind(this IEdmTypeReference type)
...
+        {
+            HashSet<string> baseTypes;
+            string fullName = type.FullTypeName();
+
+            if (!baseTypeCache.TryGetValue(fullName, out baseTypes))
+            {
+                baseTypes = new HashSet<string>();
+
+                GetBaseTypes(type, baseTypes, fullName);
+            }
```
mikepizzo, line 1233: actually, I would add this logic to GetBaseTypes as well; GetBaseTypes should check the cache to see if it has it and if not, create it and add it to the cache.  InheritsFrom should just check to see if GetBaseTypes(type).Contains(potentialBaseType)

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -173,25 +199,39 @@ public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(IEdmType b
         /// </returns>
         public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(string qualifiedName, IEdmType bindingType)
         {
-            IEnumerable<IEdmOperation> enumerable = this.FindDeclaredOperations(qualifiedName);
-            IList<IEdmOperation> operations = enumerable as IList<IEdmOperation>;
-            if (operations != null)
+            IEnumerable<IEdmOperation> enumarable = this.FindDeclaredBoundOperations(bindingType);  
+            
+            if(enumarable == null || !enumarable.Any())
```
mikepizzo, line 204: >|| !enumarable.Any()) [](start = 34, length = 21)

no; don't use enumerable.Any() -- that's linq, and is expensive.  Instead, optimize for common case by checking to see if enumerable is an IList and, if so, using it as such.

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -173,25 +199,39 @@ public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(IEdmType b
         /// </returns>
         public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(string qualifiedName, IEdmType bindingType)
         {
-            IEnumerable<IEdmOperation> enumerable = this.FindDeclaredOperations(qualifiedName);
-            IList<IEdmOperation> operations = enumerable as IList<IEdmOperation>;
-            if (operations != null)
+            IEnumerable<IEdmOperation> enumarable = this.FindDeclaredBoundOperations(bindingType);
```
mikepizzo, line 202: >enumarable [](start = 39, length = 10)

spelling: "enumerable"

src/Microsoft.OData.Core/Metadata/MetadataUtils.cs
```diff
@@ -171,25 +171,17 @@ internal static IList<IEdmOperation> CalculateBindableOperationsForType(IEdmType
...
+                return operationsFound;
+            }
 
-                IEdmOperationParameter bindingParameter = operation.Parameters.FirstOrDefault();
-                IEdmType resolvedBindingType = edmTypeResolver.GetParameterType(bindingParameter).Definition;
-                if (resolvedBindingType.IsAssignableFrom(bindingType))
-                {
-                    operationsFound.Add(operation);
-                }
+            operationsFound = new List<IEdmOperation>();
```
mikepizzo, line 181: >List<IEdmOperation>() [](start = 34, length = 21)

you can just pass operations into constructor: operationsFound = new List<IEdmOperation>(operations);

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
```diff
@@ -1216,18 +1221,36 @@ public static EdmPrimitiveTypeKind PrimitiveKind(this IEdmTypeReference type)
...
-            do
+            HashSet<IEdmStructuredType> baseTypes;
+
+            if (baseTypeCache.TryGetValue(type, out baseTypes))
+            {
+                return baseTypes;
+            }
+
+            baseTypes = new HashSet<IEdmStructuredType>();
+            baseTypeCache.TryAdd(type, baseTypes);
```
mikepizzo, line 1241: We generally add the item to the hashset after it's fully populated. Hashset is not concurrent, so if we are adding items to baseTypes while people are trying to read it that could be a problem.

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1599,17 +1599,18 @@ public static string FullTypeName(this IEdmType type)
                 return primitiveType.FullName;
             }
```
mikepizzo, line 1601: enclose 1596-1600 in an if (type.TypeKind == EdmTypeKind.Primitive) to avoid the failed cast for non-primitive types

mikepizzo (approved):
:shipit:

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -23,6 +24,11 @@ public abstract class EdmModelBase : EdmElement, IEdmModel
         private readonly Dictionary<string, IEdmTerm> termDictionary = new Dictionary<string, IEdmTerm>();
         private readonly Dictionary<string, IList<IEdmOperation>> functionDictionary = new Dictionary<string, IList<IEdmOperation>>();
 
+        /// <summary>
+        /// Cache of operations that are bindable to entity types.
+        /// </summary>
+        private readonly ConcurrentDictionary<string, IList<IEdmOperation>> bindableOperationsCache;
```
xuzhg, line 30: >bindableOperationsCache [](start = 76, length = 23)

What's the difference between "functionDictionary" and new "bindableOperationsCache"?

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1618,9 +1622,8 @@ public static string FullTypeName(this IEdmType type)
         /// <param name="type">Reference to the calling object.</param>
         /// <returns>The element type of this references definition.</returns>
         public static IEdmType AsElementType(this IEdmType type)
-        {
-            IEdmCollectionType collectionType = type as IEdmCollectionType;
-            return (collectionType != null) ? collectionType.ElementType.Definition : type;
+        {   
+            return (type.TypeKind == EdmTypeKind.Collection) ? (type as IEdmCollectionType).ElementType.Definition : type;
```
xuzhg, line 1626: >type.TypeKind [](start = 20, length = 13)

if the input "type = null", the previous one can return null,
but now, the method will throw "null reference exception"

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
```diff
@@ -1216,18 +1221,39 @@ public static EdmPrimitiveTypeKind PrimitiveKind(this IEdmTypeReference type)
         /// <param name="potentialBaseType">The potential base type of the type being tested.</param>
         /// <returns>True if and only if the type inherits from the potential base type.</returns>
         public static bool InheritsFrom(this IEdmStructuredType type, IEdmStructuredType potentialBaseType)
+        {            
+            string fullName = type.FullTypeName();
```
xuzhg, line 1225: >fullName [](start = 19, length = 8)

why do you use the "fullName"?

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
```diff
@@ -1216,18 +1221,39 @@ public static EdmPrimitiveTypeKind PrimitiveKind(this IEdmTypeReference type)
...
+            HashSet<IEdmStructuredType> baseTypes;
+
+            if (baseTypeCache.TryGetValue(type, out baseTypes))
             {
-                type = type.BaseType;
-                if (type != null && type.IsEquivalentTo(potentialBaseType))
+                return baseTypes;
+            }
+
+            baseTypes = new HashSet<IEdmStructuredType>();
```
xuzhg, line 1240: >baseTypes = new HashSet<IEdmStructuredType>(); [](start = 12, length = 46)

If a type hasn't any  base type, why do we need to create a HashSet?

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
```diff
@@ -1216,18 +1221,39 @@ public static EdmPrimitiveTypeKind PrimitiveKind(this IEdmTypeReference type)
         /// <param name="potentialBaseType">The potential base type of the type being tested.</param>
         /// <returns>True if and only if the type inherits from the potential base type.</returns>
         public static bool InheritsFrom(this IEdmStructuredType type, IEdmStructuredType potentialBaseType)
```
xuzhg, line 1223: >InheritsFrom [](start = 27, length = 12)

I 'd like to see the perf telemetry about this change.

It's easy to test, create an edm model with a big inheritance type true,

A => B => C => D => ....

Call this method multiple time and record the time duration.

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
mikepizzo: functionDictionary is all functions, indexed by name. This is a cache of all bindable functions, indexed by binding type.  could perhaps make name or comment clearer?

---
In reply to: [492997426](https://github.com/OData/odata.net/pull/1888#discussion_r492997426) [](ancestors = 492997426)

xuzhg: key: full type name of the binding type. 

Key in the functionDictionary is the full operation name.

---
In reply to: [492997426](https://github.com/OData/odata.net/pull/1888#discussion_r492997426) [](ancestors = 492997426)

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
mikepizzo: Good catch. We should add handling for null.

---
In reply to: [492999471](https://github.com/OData/odata.net/pull/1888#discussion_r492999471) [](ancestors = 492999471)

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
xuzhg: If it's binding to collection, the key is string like "Collection(Namespace.Customer)" ?

---
In reply to: [493025946](https://github.com/OData/odata.net/pull/1888#discussion_r493025946) [](ancestors = 493025946,492997426)

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
mikepizzo: so that we have something to put in the dictionary (in this case, an empty hashset)

---
In reply to: [493002182](https://github.com/OData/odata.net/pull/1888#discussion_r493002182) [](ancestors = 493002182)

xuzhg: Maybe to use "null" directly? We don't need to create an empty hashet? 

---
In reply to: [493027004](https://github.com/OData/odata.net/pull/1888#discussion_r493027004) [](ancestors = 493027004,493002182)

xuzhg: Make null so we don't need to call Contains at line 1228

---
In reply to: [493028268](https://github.com/OData/odata.net/pull/1888#discussion_r493028268) [](ancestors = 493028268,493027004,493002182)

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
mikepizzo: Good catch; I think this is leftover code from a previous iteration that can be removed.

---
In reply to: [493000379](https://github.com/OData/odata.net/pull/1888#discussion_r493000379) [](ancestors = 493000379)

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
mikepizzo: I would be okay with that, but then we might consider changing:

    HashSet<IEdmStructuredType> GetBaseTypes(IEdmStructuredType type)

to:

   bool TryGetBaseTypes(IEdmStructuredType type, out HashSet<IEdmStructuredType>) 

and return false if there are no base types.

Either way works, since it's a private method whichever is easier is fine with me, but if we exposed this publicly I would rather make the empty base type explicit rather than returning a null and having the caller have to check for null (or not, which would likely lead to unexpected null ref exceptions)

---
In reply to: [493029669](https://github.com/OData/odata.net/pull/1888#discussion_r493029669) [](ancestors = 493029669,493028268,493027004,493002182)

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
Sreejithpin: Had tested using  a test , fulltypename coming as
bindingType.FullTypeName()
"Collection(TestModel.Movie)"

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -173,25 +199,39 @@ public virtual IEnumerable<IEdmOperation> FindDeclaredBoundOperations(IEdmType b
...
         {
-            IEnumerable<IEdmOperation> enumerable = this.FindDeclaredOperations(qualifiedName);
+            IEnumerable<IEdmOperation> enumerable = this.FindDeclaredBoundOperations(bindingType);  
+            
+            if(enumerable == null)
+            {
+                return new IEdmOperation[] { };
+            }
+
+            IList<IEdmOperation> matchedOperations = new List<IEdmOperation>();
```
joaocpaiva, line 209: In hot path, something that can also have some performance impact (CPU and memory/GC), is to provide the initial capacity in the List constructor, when possible/available of course.

The default constructor initializes with Capacity of 4 and it doubles the size (which requires allocation and copy) every time needed when Add is invoked.

Even if we just know the MAX capacity expected and it is not crazy high, that is something worth providing because then we avoid all allocation and copies that come with adding elements to the list.

joaocpaiva (commented):
The changes look very promising!

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -23,6 +24,11 @@ public abstract class EdmModelBase : EdmElement, IEdmModel
         private readonly Dictionary<string, IEdmTerm> termDictionary = new Dictionary<string, IEdmTerm>();
         private readonly Dictionary<string, IList<IEdmOperation>> functionDictionary = new Dictionary<string, IList<IEdmOperation>>();
 
+        /// <summary>
+        /// Cache of operations that are bindable to entity types, its a cache of all bindable functions, indexed by binding type.
+        /// </summary>
+        private readonly ConcurrentDictionary<string, IList<IEdmOperation>> bindableOperationsCache;
```
joaocpaiva, line 30: With these new caches introduced, there shouldn't be a problem with it growing too big during process lifetime, where we need to start considering a eviction policy or something of that sort right?

xuzhg (approved):
:shipit:

src/Microsoft.OData.Core/Metadata/MetadataUtils.cs
```diff
@@ -171,27 +171,15 @@ internal static IList<IEdmOperation> CalculateBindableOperationsForType(IEdmType
...
-                if (resolvedBindingType.IsAssignableFrom(bindingType))
-                {
-                    operationsFound.Add(operation);
-                }
+            if(operationsFound != null)
+            {
+                return operationsFound;
             }
 
+            operationsFound = new List<IEdmOperation>(operations);
```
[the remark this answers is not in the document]
Sreejithpin: This is the fall back option when the returned IEnumarable cannot be casted to IList. We are hoping this is a less common scenario

src/Microsoft.OData.Edm/ExtensionMethods/EdmTypeSemantics.cs
```diff
@@ -1217,15 +1222,44 @@ public static EdmPrimitiveTypeKind PrimitiveKind(this IEdmTypeReference type)
         /// <returns>True if and only if the type inherits from the potential base type.</returns>
         public static bool InheritsFrom(this IEdmStructuredType type, IEdmStructuredType potentialBaseType)
         {
-            do
+            HashSet<IEdmStructuredType> baseTypes;
+            
+            if(TryGetBaseTypes(type, out baseTypes))
+            {
+                return baseTypes.Contains(potentialBaseType);
```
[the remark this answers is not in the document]
Sreejithpin: Yes

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
```diff
@@ -154,13 +161,32 @@ public IEnumerable<IEdmOperation> FindDeclaredOperations(string qualifiedName)
...
             {
-                foreach (IEdmOperation operation in operations.Where(o => o.HasEquivalentBindingType(bindingType)))
+                HashSet<IList<IEdmOperation>> operationList = new HashSet<IList<IEdmOperation>>();
+                bindableOperations = new List<IEdmOperation>();
+
+                foreach (IList<IEdmOperation> operations in this.functionDictionary.Values)
                 {
-                    yield return operation;
+                    //To ensure the Listof operations is distinct, as the same list of operations could be there for another qualified name in dictionary
+                    if (operationList.Add(operations))
```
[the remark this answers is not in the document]
Sreejithpin: Yes

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
Sreejithpin: Shouldnt be the case as this cache will contain only a subset of operations

src/Microsoft.OData.Core/Metadata/MetadataUtils.cs
```diff
@@ -171,27 +171,15 @@ internal static IList<IEdmOperation> CalculateBindableOperationsForType(IEdmType
...
             }
 
-            List<IEdmOperation> operationsFound = new List<IEdmOperation>();
-            foreach (IEdmOperation operation in operations)
-            {
-                if (!operation.IsBound)
-                {
-                    throw new ODataException(Strings.EdmLibraryExtensions_UnBoundOperationsFoundFromIEdmModelFindMethodIsInvalid(operation.Name));
-                }
+            IList<IEdmOperation> operationsFound = operations as IList<IEdmOperation>;
```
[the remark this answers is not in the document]
mikepizzo: xactly.

---
In reply to: [493160266](https://github.com/OData/odata.net/pull/1888#discussion_r493160266) [](ancestors = 493160266)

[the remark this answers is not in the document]
mikepizzo: as per below, this will usually be a (cached) IList<T>, so should have a significant perf improvement.

---
In reply to: [493150453](https://github.com/OData/odata.net/pull/1888#discussion_r493150453) [](ancestors = 493150453)

src/Microsoft.OData.Core/Metadata/MetadataUtils.cs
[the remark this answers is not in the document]
mikepizzo: With this fix, FindBoundOperations should both be less expensive, be cached, and be returned as an IList<T>, so (unless the developer overwrites FindBoundOperations) this code should never get hit.

---
In reply to: [493156324](https://github.com/OData/odata.net/pull/1888#discussion_r493156324) [](ancestors = 493156324)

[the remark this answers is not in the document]
mikepizzo: As above, the only time the returned IEnumerable will not be an IList<T> is if the developer overwrites FindBoundOperations(bindingType) and returns a non-IEnumerable IList<T>. This would be a very uncommon scenario, and the developer could address any perf issues  by returning an IList<T>.

---
In reply to: [493184421](https://github.com/OData/odata.net/pull/1888#discussion_r493184421) [](ancestors = 493184421)

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
mikepizzo: nope; max size will be the number of bindable operations in the model; they are just partitioned by type.

---
In reply to: [493171544](https://github.com/OData/odata.net/pull/1888#discussion_r493171544) [](ancestors = 493171544)

src/Microsoft.OData.Edm/Schema/EdmModelBase.cs
mikepizzo: An initial guess of 4 is probably a pretty good guess.

---
In reply to: [493170697](https://github.com/OData/odata.net/pull/1888#discussion_r493170697) [](ancestors = 493170697)

Issue #1890, opened 2020-09-21
Deprecated types used in NavigationProperties report deprecation twice


mikepizzo:
*If an entity type marked as deprecated is used in a navigation property, then the deprecation is reported twice when validating a URL containing that navigation property*

### Assemblies affected

*OData .Net lib 7.7.1*

### Reproduce steps

Mark a type as deprecated.
Use that type in a navigation property.
Using the deprecation rules, validate a URL referencing that navigation property.

### Expected result

There should be one deprecation warning.

### Actual result

There are two deprecation warnings.

### Additional detail

This occurs because, when validating an NavigationPropertySegment, we validate both the NavigationProperty and the NavigationSource. 

When validating either a NavigationProperty or a NavigationSource, we also validate the type of the navigation property/navigation source.  Thus, the type is validated twice (and raises two warning messages).

mikepizzo:
Fixed by #1891

PR #1891, opened 2020-09-21
Fix logic to validate URL segment by type

mikepizzo:
### Issues

*This pull request fixes issue #1890.*

### Description

When validating a segment, we also validate the type. However, if the type of the segment is a collection, we pass the typereference, rather than the type.

Because this resulted in not validating the type for a collection-valued navigation property in the path, the DeprecationRule had logic to explicitly validate the type of the segment being validated.

For navigation property segments, the validator validates both the navigation property and the navigation source, so the deprecation rule gets called twice (once for the navigation property and once for the navigation source).

Which is fine, except that the deprecation rule had explicit logic to also validate the type, which is the same for the navigation property and the navigation source, so the rule was called twice and two messages were logged.

The fix is to:
1) Fix handling within the ODataUriValidator to pass the EdmType, rather than the EdmTypeReference, in the case of a collection-valued segment.
2) Remove the logic from the deprecation rule to separately validate the type
3) Also removed a redundant check for used types. 

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UrlValidation/Rules/DeprecationRules.cs
+++ b/src/Microsoft.OData.Core/UrlValidation/Rules/DeprecationRules.cs
@@ -118,10 +118,6 @@ namespace Microsoft.OData.UriParser.Validation.Rules
             {
                 context.Messages.Add(CreateUrlValidationMessage(element.Name, message, version, date, removalDate));
             }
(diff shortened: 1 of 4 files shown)


xuzhg (approved):
:shipit:

Issue #1911, opened 2020-10-21
Support JSON Metadata in OData.Core lib


xuzhg:
*Short summary (3-5 sentences) describing the issue.*

### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData .Net lib 7.x*

### Reproduce steps

Since OData.Edm lib supports JSON metadata reader/writer for .NET Standard 2.0.

It's time to enable it in the OData.Core lib.

For example, a request 

1) `~/odata/$metadata` returns the xml CSDL.
2) `~/odata/$metadata?$format=application/json` returns JSON CSDL

xuzhg:
Merged. Thanks.

PR #1913, opened 2020-10-22
Enable JSON metadata reader, writer in OData Core

xuzhg:
### Issues

*This pull request fixes issue #1911*

### Description

OData.Edm lib supports to read/write JSON metadata.

This PR is to enable read/write JSON metadata in OData.Core lib.

So, if a request like:

*  `~/odata/$metadata `  => return XML CSDL
* `~/odata/$metadata?$format=application/json`  => return JSON CSDL

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
+++ b/src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
@@ -82,4 +82,8 @@
     </Compile>
   </ItemGroup>
 
+  <ItemGroup>
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
+  </ItemGroup>
+
 </Project>
--- a/src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
+++ b/src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
@@ -82,8 +82,4 @@
     </Compile>
   </ItemGroup>
 
-  <ItemGroup>
-    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
-  </ItemGroup>
-
 </Project>
--- a/src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
+++ b/src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
@@ -9,6 +9,7 @@ namespace Microsoft.OData
     using System;
     using Microsoft.OData.Buffers;
     using Microsoft.OData.Edm;
+    using Microsoft.OData.Edm.Csdl;
 
     /// <summary>
     /// Configuration settings for OData message readers.
@@ -203,6 +204,12 @@ namespace Microsoft.OData
         /// </summary>
         public Func<string, bool> ShouldIncludeAnnotation { get; set; }
 
+#if NETSTANDARD2_0
+        /// <summary>
+        /// Gets/sets the JSON-CSDL reader settings.
+        /// </summary>
+        public CsdlJsonReaderSettings JsonCsdlReaderSettings { get; set; }
+#endif
         /// <summary>
         /// Gets the bound validator.
         /// </summary>
--- a/src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
+++ b/src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
@@ -204,12 +204,6 @@ namespace Microsoft.OData
         /// </summary>
         public Func<string, bool> ShouldIncludeAnnotation { get; set; }
 
-#if NETSTANDARD2_0
-        /// <summary>
-        /// Gets/sets the JSON-CSDL reader settings.
-        /// </summary>
-        public CsdlJsonReaderSettings JsonCsdlReaderSettings { get; set; }
-#endif
         /// <summary>
         /// Gets the bound validator.
         /// </summary>
--- a/src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
+++ b/src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
@@ -9,7 +9,6 @@ namespace Microsoft.OData
     using System;
     using Microsoft.OData.Buffers;
     using Microsoft.OData.Edm;
-    using Microsoft.OData.Edm.Csdl;
 
     /// <summary>
     /// Configuration settings for OData message readers.
--- a/src/Microsoft.OData.Core/ODataMetadataFormat.cs
+++ b/src/Microsoft.OData.Core/ODataMetadataFormat.cs
@@ -60,7 +60,25 @@ namespace Microsoft.OData
             ExceptionUtils.CheckArgumentNotNull(messageInfo, "messageInfo");
             ExceptionUtils.CheckArgumentNotNull(messageReaderSettings, "messageReaderSettings");
 
+            bool isJson = IsJsonMetadata(messageInfo.MediaType);
+
+#if NETSTANDARD2_0
+            if (isJson)
+            {
+                return new ODataMetadataJsonInputContext(messageInfo, messageReaderSettings);
+            }
+            else
+            {
+                return new ODataMetadataInputContext(messageInfo, messageReaderSettings);
+            }
+#else
+            if (isJson)
+            {
+                throw new ODataException(Strings.ODataMetadataOutputContext_NotSupportJsonMetadata);
+            }
+
             return new ODataMetadataInputContext(messageInfo, messageReaderSettings);
+#endif
         }
 
         /// <summary>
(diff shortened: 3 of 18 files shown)


src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
```diff
@@ -82,4 +82,8 @@
     </Compile>
   </ItemGroup>
 
+  <ItemGroup>
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
+  </ItemGroup>
+
```
mikepizzo, line 88: what is this?

mikepizzo:
>         }

Is there a similar overload for reading a json metadata document that references other json metadata documents?

---
Refers to: src/Microsoft.OData.Core/ODataMessageReader.cs:881 in 5c1aa37. [](commit_id = 5c1aa372e3d5f42586ed0980e6f8846c36165aa7, deletion_comment = False)

src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
xuzhg: It seems it's VS autogenerated.

---
In reply to: [518468018](https://github.com/OData/odata.net/pull/1913#discussion_r518468018) [](ancestors = 518468018)

src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
```diff
@@ -203,6 +204,12 @@ public ODataMessageQuotas MessageQuotas
         /// </summary>
         public Func<string, bool> ShouldIncludeAnnotation { get; set; }
 
+#if NETSTANDARD2_0
+        /// <summary>
+        /// Gets/sets the JSON-CSDL reader settings.
+        /// </summary>
+        public CsdlJsonReaderSettings JsonCsdlReaderSettings { get; set; }
+#endif
```
mikepizzo, line 212: would rather not have a JSON-format specific property on MessageReaderSettings. Is there somewhere we can put this that is specific to the JSON format?

xuzhg:
>         }

No, Json metadata reader use the setting.  See the ODataMessageReaderSettings.cs changes with "CsdlJsonReaderSettings".

---
In reply to: [722747779](https://github.com/OData/odata.net/pull/1913#issuecomment-722747779) [](ancestors = 722747779)

---
Refers to: src/Microsoft.OData.Core/ODataMessageReader.cs:881 in 5c1aa37. [](commit_id = 5c1aa372e3d5f42586ed0980e6f8846c36165aa7, deletion_comment = False)

src/Microsoft.OData.Core/ODataMetadataFormat.cs
```diff
@@ -128,6 +164,23 @@ public override string ToString()
             throw new ODataException(Strings.General_InternalError(InternalErrorCodes.ODataMetadataFormat_CreateOutputContextAsync));
         }
 
+        private static bool IsJsonMetadata(ODataMediaType contentType)
+        {
+            // by default, it's XML metadata
+            if (contentType == null)
+            {
+                return false;
+            }
```
mikepizzo, line 173: you could put contentType!=null as part of the check in 175, but I'm okay if you want to leave as is for readability.

src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
xuzhg: Otherwise is to create overload "ReadMetadataDocument" to accept the setting? 

public IEdmModel ReadMetadataDocument(CsdlJsonReaderSettings jsonReaderSetting)

---
In reply to: [518470116](https://github.com/OData/odata.net/pull/1913#discussion_r518470116) [](ancestors = 518470116)

src/Microsoft.OData.Core/ODataMetadataJsonInputContext.cs
```diff
@@ -0,0 +1,99 @@
...
+        /// <summary>
+        /// Read a metadata document.
+        /// This method reads the metadata document from the input and returns
+        /// an <see cref="IEdmModel"/> that represents the read metadata document.
+        /// </summary>
+        /// <returns>An <see cref="IEdmModel"/> representing the read metadata document.</returns>
+        internal override IEdmModel ReadMetadataDocument()
+        {
+            // We can't use stream.Read(Span<byte> buffer), this method is introduced since .NET Core 2.1. :(
+            byte[] bytes = this.messageStream.ReadAllBytes();
```
mikepizzo, line 49: Is it worth having an #IFDEF for later versions of .NET Core?

src/Microsoft.OData.Core/ODataMetadataFormat.cs
xuzhg: Yes, put it as it for readability 

---
In reply to: [518471298](https://github.com/OData/odata.net/pull/1913#discussion_r518471298) [](ancestors = 518471298)

src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs
```diff
@@ -0,0 +1,174 @@
...
+        /// </summary>
+        /// <param name="messageInfo">The context information for the message.</param>
+        /// <param name="messageWriterSettings">Configuration settings of the OData writer.</param>
+        internal ODataMetadataJsonOutputContext(
+            ODataMessageInfo messageInfo,
+            ODataMessageWriterSettings messageWriterSettings)
+            : base(ODataFormat.Metadata, messageInfo, messageWriterSettings)
+        {
+            Debug.Assert(messageInfo.MessageStream != null, "messageInfo.MessageStream != null");
+            Debug.Assert(!messageInfo.IsAsync, "Metadata output context is only supported in synchronous operations.");
```
mikepizzo, line 42: :-(

src/Microsoft.OData.Core/ODataMetadataJsonInputContext.cs
xuzhg: At top this class, we have "#if NETSTANDARD2_0".

Besides, we don't output for other version of .NET Core. So, can we leave it as it?

---
In reply to: [518471948](https://github.com/OData/odata.net/pull/1913#discussion_r518471948) [](ancestors = 518471948)

src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
mikepizzo: I like that slightly better, but it still ties the new overload to JSON.  If we did something like that, I'd rather define a base CsdlReaderSettings that CsdlJsonReaderSettings derived from, and have the new overload take an instance of that base class.  So then if we supported other formats in the future, we could pass the appropriate settings in for that format.

We could even define an XmlReaderSettings that has a Func<uri,xmlreader> property and in 8.0 we can remove the ReadMetadataDocument overload that takes Func<uri,xmlreader>.

---
In reply to: [518471455](https://github.com/OData/odata.net/pull/1913#discussion_r518471455) [](ancestors = 518471455,518470116)

xuzhg: Bad news is that we already have "CsdlReaderSettings" and that's for XML specified, sealed. 
When we designed the JSON metadata in OData.Edm, I think we discussed it already.
So, Let me do "public IEdmModel ReadMetadataDocument(CsdlJsonReaderSettings jsonReaderSetting)".

---
In reply to: [518478832](https://github.com/OData/odata.net/pull/1913#discussion_r518478832) [](ancestors = 518478832,518471455,518470116)

mikepizzo:
I really don't want to add a JSON-specific method to the core reader.

How about if we define a base class for CsdlReaderSettings that we make the current (Xml) CsdlReaderSettings derive from, and then derive JsonCsdlReaderSettings from that same base class.

xuzhg:
Hi, @Michael Pizzo do you have other concerns?

---
In reply to: [723236195](https://github.com/OData/odata.net/pull/1913#issuecomment-723236195) [](ancestors = 723236195)

src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
mikepizzo: Can we remove it?

---
In reply to: [518470007](https://github.com/OData/odata.net/pull/1913#discussion_r518470007) [](ancestors = 518470007,518468018)

xuzhg: Yes, it's removed

mikepizzo (approved):
:shipit:

Issue #1907, opened 2020-10-16
IN clause not compatible with EQ


kksrivatsa:
*Short summary (3-5 sentences) describing the issue.*
OData library expects the IN clause to be a valid JSON array. This breaks the compatibility with simple equals check. I expect the two queries below to be equivalent:
$filter=propertyName eq '\\[abc\\]'
$filter=propertyName in ('\\[abc\\]')

However the second query breaks because it isn't a valid JSON array.

### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData .Net lib 7.x*
Microsoft.OData.Core.dll; version 7.7.2.

### Reproduce steps

*The simplest set of steps to reproduce the issue. If possible, reference a commit that demonstrates the issue.*
Run queries with filters below. The second query fails.
$filter=propertyName eq '\\[abc\\]'
$filter=propertyName in ('\\[abc\\]')

The issue is with the backslash. Without the backslash, both the queries are equivalent.

### Expected result

*What would happen if there wasn't a bug.*
The second query would work and give similar results as the first.

### Actual result

*What is actually happening.*
The request fails with the error below. Please see additional details for stack trace.

       Invalid JSON. An unrecognized escape sequence '\[' was found in a JSON string value.

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*
Stack trace:

   at Microsoft.OData.Json.JsonReader.ParseStringPrimitiveValue(Boolean& hasLeadingBackslash)
   at Microsoft.OData.Json.JsonReader.get_Value()
   at Microsoft.OData.Json.BufferingJsonReader.get_Value()
   at Microsoft.OData.Json.ODataJsonReaderCoreUtils.TryReadNullValue(IJsonReader jsonReader, ODataInputContext inputContext, IEdmTypeReference expectedTypeReference, Boolean validateNullValue, String propertyName, Nullable`1 isDynamicProperty)
   at Microsoft.OData.JsonLight.ODataJsonLightPropertyAndValueDeserializer.ReadNonEntityValueImplementation(String payloadTypeName, IEdmTypeReference expectedTypeReference, PropertyAndAnnotationCollector propertyAndAnnotationCollector, CollectionWithoutExpectedTypeValidator collectionValidator, Boolean validateNullValue, Boolean isTopLevelPropertyValue, Boolean insideResourceValue, String propertyName, Nullable`1 isDynamicProperty)
[...]

PR #1916, opened 2020-11-02
Apply same backslash escaping to string literal within "IN" clause as Equals

mikepizzo:
### Issues

*This pull request fixes issue #1907.*

### Description

*Single backslashes within an In clause should be handled the same as within a equals expression and not require doubling.*

- [x] *Test cases added*


xuzhg (approved):
:shipit:

PR #1917, opened 2020-11-03
Improve UrlValidation Performance

mikepizzo:
### Issues

*This pull request improves performance of UrlValidation*

### Description

Optimize how UrlValidator eliminates necessary tests.
Make ODataUrlValidator public in order to support caching/re-use of detected optimizations

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/UrlValidation/ValidationEngine/ODataUrlValidator.cs
+++ b/src/Microsoft.OData.Core/UrlValidation/ValidationEngine/ODataUrlValidator.cs
@@ -148,8 +148,11 @@ namespace Microsoft.OData.UriParser.Validation
             // GetInterfaces() returns derived interfaces, so only do this at the top level
             if (includeInterfaces)
             {
-                foreach (Type interfaceType in GetInterfaces(itemType))
+                Type[] interfaces = GetInterfaces(itemType);
+                int length = interfaces.Length;
+                for (int i=0; i < length; i++)
                 {
+                    Type interfaceType = interfaces[i];
                     if (validatedTypes.Add(interfaceType) && !this.unusedTypes.ContainsKey(interfaceType))
                     {
                         if (ValidateByType(item, interfaceType, validationContext, impliedProperty))
--- a/src/Microsoft.OData.Core/UrlValidation/ValidationEngine/ODataUrlValidator.cs
+++ b/src/Microsoft.OData.Core/UrlValidation/ValidationEngine/ODataUrlValidator.cs
@@ -18,11 +18,12 @@ namespace Microsoft.OData.UriParser.Validation
     /// <summary>
     /// Validation engine for running validation rules
     /// </summary>
-    internal class ODataUrlValidator
+    public sealed class ODataUrlValidator
     {
         private Dictionary<Type, List<ODataUrlValidationRule>> ruleDictionary = new Dictionary<Type, List<ODataUrlValidationRule>>();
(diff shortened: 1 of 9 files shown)


src/Microsoft.OData.Core/UrlValidation/ValidationEngine/ODataUrlValidator.cs
```diff
@@ -144,33 +145,59 @@ private bool ValidateTypeHierarchy(object item, Type itemType, ODataUrlValidatio
             validatedTypes.Add(itemType);
 
             // find any rules defined on an interface defined on the type of the item
-            foreach (Type interfaceType in itemType.GetInterfaces())
+            // GetInterfaces() returns derived interfaces, so only do this at the top level
+            if (includeInterfaces)
             {
-                if (validatedTypes.Add(interfaceType) && !this.unusedTypes.ContainsKey(interfaceType))
+                foreach (Type interfaceType in GetInterfaces(itemType))
                 {
```
Sreejithpin, line 152: Can we use for loops as its array

src/Microsoft.OData.Core/UrlValidation/ValidationEngine/ODataUrlValidator.cs
```diff
@@ -411,26 +438,29 @@ private void ValidateImpliedProperties(IEdmType segmentType, SelectExpandClause
         private void ValidateProperties(IEdmType edmType, ODataUrlValidationContext context)
```
Sreejithpin, line 438: Could this be called multiple times for same edmtype? if so can this be cahced

test/FunctionalTests/Microsoft.OData.Core.Tests/UrlValidation/ODataUrlValidatorTests.cs
```diff
@@ -55,17 +55,121 @@ public static void FailedParse()
...
+            Assert.Collection<ODataUrlValidationMessage>(errors, (error) =>
+            {
+                Assert.Equal("dummyCode", error.MessageCode);
+                Assert.Equal("dummyMessage", error.Message);
+                Assert.Equal(OData.UriParser.Validation.Severity.Warning, error.Severity);
+            });
+        }
+
+        [Fact]
+        public static void PassingValdiatorSucceeds()
```
xuzhg, line 97: >Valdiator [](start = 34, length = 9)

typo

xuzhg (approved):
:shipit:

src/Microsoft.OData.Core/UrlValidation/ValidationEngine/ODataUrlValidator.cs
mikepizzo: Good call; I like it.

---
In reply to: [518474457](https://github.com/OData/odata.net/pull/1917#discussion_r518474457) [](ancestors = 518474457)

src/Microsoft.OData.Core/UrlValidation/ValidationEngine/ODataUrlValidator.cs
mikepizzo: there's not really anything to cache (the method looks for and executes any rules that are applied to the particular type, and the rules add any exceptions to the validation context). However, we go one better, and on line 441 we skip validation if we've already run validation on this type. 

---
In reply to: [518476690](https://github.com/OData/odata.net/pull/1917#discussion_r518476690) [](ancestors = 518476690)

Issue #1910, opened 2020-10-21
Writing a Deleted Resource with ODataDeltaResourceSetWriter fails with null ref exception.


arjkailash:
*Short summary (3-5 sentences) describing the issue.*
Writing a Deleted Resource with ODataDeltaResourceSetWriter fails with null ref exception.when using writter settings OdataVersion v4.0 and when the Id of the resource is not set.

This doesnt fail in OdataVersion v4.01 since key properties were present (key properties or @id needs to be present.) 

### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData .Net lib 7.x*
Currently using Microsoft.Odata.Code(7.7.2)
### Reproduce steps

*The simplest set of steps to reproduce the issue. If possible, reference a commit that demonstrates the issue.*
Try writing the response 
{"keyproperty":"value", "reason":"deleted"}
With odata writer settings OdataVersion4.0 and an OdataDeltaResourceSetWriter.

### Expected result

*What would happen if there wasn't a bug.*
The write would succeed if there are key properties present even if the Id of the deleted resource was null.
### Actual result

*What is actually happening.*
Since Id of the deleted Resource is null, it fails with a null ref exception.
### Additional detail
Stack trace - 
**ODataJsonLightWriter.WriteDeletedResourceId(ODataDeletedResource resource)**
ODataJsonLightWriter.StartDeletedResource(ODataDeletedResource resource)
ODataWriterCore.InterceptException(Action action)
<WriteODataEntry>d__16.MoveNext() line 1770

mikepizzo:
This should be addressed in #1919 - please re-open if the issue is not resolved.

PR #1919, opened 2020-11-05
Support reading/writing OData V4 deleted entries with only key values

mikepizzo:
… supplying key properties.

### Issues

*This pull request fixes issue #1910.*

### Description

If an id is not supplied when writing an OData V4.0 Deleted entry, uses key values instead if provided

- [x] *Test cases added*

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
@@ -7,7 +7,6 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
-using System.Reflection;
 using System.Text;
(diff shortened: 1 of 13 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
```diff
@@ -389,6 +389,7 @@ public void WriteEntityInDeltaFeedWithSelectExpand(bool isResponse)
             {
                 Properties = new List<ODataProperty>
                 {
+
                     new ODataProperty { Name = "City", Value = "Shanghai" },
```
xuzhg, line 393: no needed

xuzhg (approved):
:shipit:

Sreejithpin (approved):

xuzhg (approved):

PR #1810, opened 2020-06-17
Support for $apply/groupby and $apply/aggregate

gathogojr:
### Issues

*This pull request fixes issue #596.*

### Description

This pull request implement support for $apply/aggregate in OData client. The scenarios covered include the following:
Scalar/singleton scenarios:

### Queryable.Sum(d => d.Prop)
Translated to `$apply=aggregate(Prop with sum as SumProp)`
### Queryable.Average(d => d.Prop)
Translated to `$apply=aggregate(Prop with average as SumProp)`
### Queryable.Min(d => d.Prop)
Translated to `$apply=aggregate(Prop with min as SumProp)`
### Queryable.Max(d => d.Prop)
Translated to `$apply=aggregate(Prop with max as SumProp)`
### Queryable.Select(d => d.Prop).Distinct().Count()
Translated to `$apply=aggregate(Prop with countdistinct as CountDistinct)`
_One of the ideas that came up post-implementation is introducing a new method `Queryable.CountDistinct(d -> d.Prop)` that can be used instead_
### Queryable.Count()
Ignored since this expression is already translated to `/<EntitySet>/$count`
It would translate to `$apply=aggregate($count as Count)` that achieves the same purpose albeit a little differently.

**NOTE**: Applicable overloads supported for each of the aggregate expressions
i.e., **int**, **int?**, **double**, **double?**, **decimal**, **decimal?**, **long**, **long?**, **float**, **float**?

### Group by a constant
- ResultSelector overload:
```Queryable.GroupBy(d1 => [Constant], (d2, d3) => new { SumProp = d3.Sum(d4 => d4.Prop), AverageProp = d3.Average(d4 => d4.Prop), ... })```
- Projection overload:
```Queryable.GroupBy(d1 => [Constant]).Select(d3 => new { SumProp = d3.Sum(d4 => d4.Prop), AverageProp = d3.Average(d4 => d4.Prop), ... })```

Translated to
`$apply=aggregate(Prop with sum as SumProp, Prop with average as AverageProp, ...)`
### Group by a single property
- ResultSelector overload:
```Queryable.GroupBy(d1 => d1.GroupingProp, (d2, d3) => new { GroupingProp = d2, SumProp = d3.Sum(d4 => d4.Prop), AverageProp = d3.Average(d4 => d4.Prop), ... }```
- Projection overload:
[...]

PR #1921, opened 2020-11-11
Support for $apply/aggregate

gathogojr:
### Issues

*This pull request is 1 of 3 from splitting [this](https://github.com/OData/odata.net/pull/1810) huge pull request to make it easier to review.*

### Description

The simple aggregations below are covered. The scope of this PR is aggregations that return a single scalar result.

```csharp
    // EntitySet?$apply=aggregate(Prop with average as AverageProp)
    EntitySet.Average(d1 => d1.Prop)

    // EntitySet?$apply=aggregate(Prop with sum as SumProp)
    EntitySet.Sum(d1 => d1.Prop)

    // EntitySet?$apply=aggregate(Prop with min as MinProp)
    EntitySet.Min(d1 => d1.Prop)

    // EntitySet?$apply=aggregate(Prop with max as MaxProp)
    EntitySet.Max(d1 => d1.Prop)

    // EntitySet?$apply=aggregate(Prop with countdistinct as CountDistinctProp)
    EntitySet.CountDistinct(d1 => d1.Prop)

    // EntitySet?$apply=aggregate(NavProp/Prop with average as AverageNavPro_Prop)
    EntitySet.Average(d1 => d1.NavProp.Prop)

    // EntitySet?$apply=aggregate(NavProp/Prop with sum as SumNavPro_Prop)
    EntitySet.Sum(d1 => d1.NavProp.Prop)

    // EntitySet?$apply=aggregate(NavProp/Prop with min as MinNavPro_Prop)
    EntitySet.Min(d1 => d1.NavProp.Prop)

    // EntitySet?$apply=aggregate(NavProp/Prop with max as MaxNavPro_Prop)
    EntitySet.Max(d1 => d1.NavProp.Prop)

    // EntitySet?$apply=aggregate(NavProp/Prop with countdistinct as CountDistinctNavPro_Prop)
    EntitySet.CountDistinct(d1 => d1.NavProp.Prop)

    // NOTE: We support nested navigation properties in the aggregate expression, i.e. d1 => d1.NavProp.NestedNavProp.Prop
    // E.g.
    // EntitySet?$apply=aggregate(NavProp/NestedNavProp/Prop with average as AverageNavProp_NestedNavProp_Prop)
    EntitySet.Average(d1 => d1.NavProp.NestedNavProp.Prop)
```
The [filter transformation](http://docs.oasis-open.org/odata/odata-data-aggregation-ext/v4.0/cs02/odata-data-aggregation-ext-v4.0-cs02.html#_Toc435016585) used together with aggregation is also covered. A filter transformation is used (_not $filter_) to restrict the set of data to be aggregated.
```csharp
    // EntitySet?$apply=filter(Amount gt 1)/aggregate(Prop with average as AverageProp)
    EntitySet.Where(d0 => d0.Amount > 1).Average(d1 => d1.Prop)
```

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/ALinq/ApplyQueryOptionExpression.cs
+++ b/src/Microsoft.OData.Client/ALinq/ApplyQueryOptionExpression.cs
@@ -35,7 +35,7 @@ namespace Microsoft.OData.Client
 		}
 
 		/// <summary>
-		/// Aggregations in the $apply expression
+		/// Gets the aggregations in the $apply expression
 		/// </summary>
 		internal List<Aggregation> Aggregations { get; private set; }
 
--- a/src/Microsoft.OData.Client/ALinq/ApplyQueryOptionExpression.cs
+++ b/src/Microsoft.OData.Client/ALinq/ApplyQueryOptionExpression.cs
@@ -6,130 +6,114 @@
 
 namespace Microsoft.OData.Client
 {
-	using System;
-	using System.Collections.Generic;
+    using System;
+    using System.Collections.Generic;
     using System.Collections.ObjectModel;
+    using System.Linq;
     using System.Linq.Expressions;
-	using Microsoft.OData.UriParser.Aggregation;
+    using Microsoft.OData.UriParser.Aggregation;
 
-	/// <summary>
-	/// A resource specific expression representing an apply query option.
-	/// </summary>
-	internal class ApplyQueryOptionExpression : QueryOptionExpression
-	{
-		/// <summary>
-		/// The filter expressions that make the filter predicate
-		/// </summary>
-		private readonly List<Expression> filterExpressions;
+    /// <summary>
+    /// A resource specific expression representing an apply query option.
+    /// </summary>
+    internal class ApplyQueryOptionExpression : QueryOptionExpression
+    {
+        /// <summary>
+        /// The filter expressions that make the filter predicate
+        /// </summary>
+        private readonly List<Expression> filterExpressions;
 
-		/// <summary>
-		/// Creates an <see cref="ApplyQueryOptionExpression"/> expression.
-		/// </summary>
-		/// <param name="type">the return type of the expression.</param>
-		internal ApplyQueryOptionExpression(Type type)
-			: base(type)
-		{
-			this.Aggregations = new List<Aggregation>();
-			this.filterExpressions = new List<Expression>();
-		}
+        /// <summary>
+        /// Creates an <see cref="ApplyQueryOptionExpression"/> expression.
+        /// </summary>
+        /// <param name="type">the return type of the expression.</param>
+        internal ApplyQueryOptionExpression(Type type)
+            : base(type)
+        {
+            this.Aggregations = new List<Aggregation>();
+            this.filterExpressions = new List<Expression>();
+        }
 
-		/// <summary>
-		/// The <see cref="ExpressionType"/> of the <see cref="Expression"/>.
-		public override ExpressionType NodeType
-		{
-			get { return (ExpressionType)ResourceExpressionType.ApplyQueryOption; }
-		}
+        /// <summary>
+        /// The <see cref="ExpressionType"/> of the <see cref="Expression"/>.
+        /// </summary>
+        public override ExpressionType NodeType
+        {
+            get { return (ExpressionType)ResourceExpressionType.ApplyQueryOption; }
+        }
 
-		/// <summary>
-		/// Gets the aggregations in the $apply expression
-		/// </summary>
-		internal List<Aggregation> Aggregations { get; private set; }
+        /// <summary>
+        /// Gets the aggregations in the $apply expression
+        /// </summary>
+        internal List<Aggregation> Aggregations { get; private set; }
 
-		/// <summary>
-		/// Adds the conjuncts to the filter expressions
-		/// </summary>
-		internal void AddPredicateConjuncts(IEnumerable<Expression> predicates)
-		{
-			this.filterExpressions.AddRange(predicates);
-		}
+        /// <summary>
+        /// Adds the conjuncts to the filter expressions
+        /// </summary>
+        internal void AddPredicateConjuncts(IEnumerable<Expression> predicates)
+        {
+            this.filterExpressions.AddRange(predicates);
+        }
 
-		internal ReadOnlyCollection<Expression> PredicateConjuncts
-		{
-			get
-			{
-				return new ReadOnlyCollection<Expression>(this.filterExpressions);
-			}
-		}
+        internal ReadOnlyCollection<Expression> PredicateConjuncts
+        {
+            get
+            {
+                return new ReadOnlyCollection<Expression>(this.filterExpressions);
+            }
+        }
 
-		/// <summary>
-		/// Gets filter transformation predicate.
-		/// </summary>
-		/// <returns>A predicate with all conjuncts AND'd</returns>
-		internal Expression GetPredicate()
-		{
-			Expression combinedPredicate = null;
-			bool isFirst = true;
+        /// <summary>
+        /// Gets filter transformation predicate.
+        /// </summary>
+        /// <returns>A predicate with all conjuncts AND'd</returns>
+        internal Expression GetPredicate()
+        {
+            return this.filterExpressions.Aggregate((leftExpr, rightExpr) => Expression.And(leftExpr, rightExpr));
+        }
 
-			foreach (Expression expr in this.filterExpressions)
-			{
-				if (isFirst)
-				{
-					combinedPredicate = expr;
-					isFirst = false;
-				}
-				else
-				{
-					combinedPredicate = Expression.And(combinedPredicate, expr);
-				}
-			}
+        /// <summary>
+        /// Structure for an aggregation. Holds lambda expression plus enum indicating aggregation method
+        /// </summary>
+        internal struct Aggregation
+        {
+            /// <summary>
+            /// Lambda expression for aggregation selector.
+            /// </summary>
+            internal readonly Expression Expression;
 
-			return combinedPredicate;
-		}
+            /// <summary>
+            /// Enum indicating aggregation method.
+            /// </summary>
+            internal readonly AggregationMethod AggregationMethod;
 
-		/// <summary>
-		/// Structure for an aggregation. Holds lambda expression plus enum indicating aggregation method
-		/// </summary>
-		internal struct Aggregation
-		{
-			/// <summary>
-			/// Lambda expression for aggregation selector.
-			/// </summary>
-			internal readonly Expression Expression;
+            /// <summary>
+            /// Aggregation alias.
+            /// </summary>
+            internal readonly string AggregationAlias;
 
-			/// <summary>
-			/// Enum indicating aggregation method.
-			/// </summary>
-			internal readonly AggregationMethod AggregationMethod;
+            /// <summary>
+            /// Creates an aggregation.
+            /// </summary>
+            /// <param name="expr">Lambda expression for aggregation selector.</param>
+            /// <param name="aggregationMethod">Enum indicating aggregation method.</param>
+            internal Aggregation(Expression expr, AggregationMethod aggregationMethod)
+                : this(expr, aggregationMethod, string.Empty)
+            {
+            }
 
-			/// <summary>
-			/// Aggregation alias.
-			/// </summary>
-			internal readonly string AggregationAlias;
-
-			/// <summary>
-			/// Creates an aggregation.
-			/// </summary>
-			/// <param name="exp">Lambda expression for aggregation selector.</param>
-			/// <param name="aggregationMethod">Enum indicating aggregation method.</param>
-			internal Aggregation(Expression exp, AggregationMethod aggregationMethod)
-			{
-				this.Expression = exp;
-				this.AggregationMethod = aggregationMethod;
-				this.AggregationAlias = string.Empty;
-			}
-
-			/// <summary>
-			/// Creates an aggregation.
-			/// </summary>
-			/// <param name="exp">Lambda expression for aggregation selector.</param>
-			/// <param name="aggregationMethod">Enum indicating aggregation method.</param>
-			/// <param name="aggregationAlias">Aggregation alias.</param>
-			internal Aggregation(Expression exp, AggregationMethod aggregationMethod, string aggregationAlias)
-			{
-				this.Expression = exp;
-				this.AggregationMethod = aggregationMethod;
-				this.AggregationAlias = aggregationAlias;
-			}
-		}
-	}
+            /// <summary>
+            /// Creates an aggregation.
+            /// </summary>
+            /// <param name="expr">Lambda expression for aggregation selector.</param>
+            /// <param name="aggregationMethod">Enum indicating aggregation method.</param>
+            /// <param name="aggregationAlias">Aggregation alias.</param>
+            internal Aggregation(Expression expr, AggregationMethod aggregationMethod, string aggregationAlias)
+            {
+                this.Expression = expr;
+                this.AggregationMethod = aggregationMethod;
+                this.AggregationAlias = aggregationAlias;
+            }
+        }
+    }
 }
--- a/src/Microsoft.OData.Client/ALinq/ApplyQueryOptionExpression.cs
+++ b/src/Microsoft.OData.Client/ALinq/ApplyQueryOptionExpression.cs
@@ -8,7 +8,8 @@ namespace Microsoft.OData.Client
 {
 	using System;
 	using System.Collections.Generic;
-	using System.Linq.Expressions;
+    using System.Collections.ObjectModel;
+    using System.Linq.Expressions;
 	using Microsoft.OData.UriParser.Aggregation;
 
 	/// <summary>
@@ -16,6 +17,10 @@ namespace Microsoft.OData.Client
 	/// </summary>
 	internal class ApplyQueryOptionExpression : QueryOptionExpression
 	{
+		/// <summary>
+		/// The filter expressions that make the filter predicate
+		/// </summary>
+		private readonly List<Expression> filterExpressions;
 
 		/// <summary>
 		/// Creates an <see cref="ApplyQueryOptionExpression"/> expression.
@@ -25,6 +30,7 @@ namespace Microsoft.OData.Client
 			: base(type)
 		{
 			this.Aggregations = new List<Aggregation>();
+			this.filterExpressions = new List<Expression>();
 		}
 
 		/// <summary>
@@ -38,6 +44,47 @@ namespace Microsoft.OData.Client
 		/// Gets the aggregations in the $apply expression
 		/// </summary>
 		internal List<Aggregation> Aggregations { get; private set; }
+
+		/// <summary>
+		/// Adds the conjuncts to the filter expressions
+		/// </summary>
+		internal void AddPredicateConjuncts(IEnumerable<Expression> predicates)
+		{
+			this.filterExpressions.AddRange(predicates);
+		}
+
+		internal ReadOnlyCollection<Expression> PredicateConjuncts
+		{
+			get
+			{
+				return new ReadOnlyCollection<Expression>(this.filterExpressions);
+			}
+		}
+
+		/// <summary>
+		/// Gets filter transformation predicate.
+		/// </summary>
+		/// <returns>A predicate with all conjuncts AND'd</returns>
+		internal Expression GetPredicate()
+		{
+			Expression combinedPredicate = null;
+			bool isFirst = true;
+
+			foreach (Expression expr in this.filterExpressions)
+			{
+				if (isFirst)
+				{
+					combinedPredicate = expr;
+					isFirst = false;
+				}
+				else
+				{
+					combinedPredicate = Expression.And(combinedPredicate, expr);
+				}
+			}
+
+			return combinedPredicate;
+		}
 
 		/// <summary>
 		/// Structure for an aggregation. Holds lambda expression plus enum indicating aggregation method
--- a/src/Microsoft.OData.Client/ALinq/ApplyQueryOptionExpression.cs
+++ b/src/Microsoft.OData.Client/ALinq/ApplyQueryOptionExpression.cs
@@ -1,0 +1,88 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ApplyQueryOptionExpression.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Client
+{
+	using System;
+	using System.Collections.Generic;
+	using System.Linq.Expressions;
+	using Microsoft.OData.UriParser.Aggregation;
+
+	/// <summary>
+	/// A resource specific expression representing an apply query option.
+	/// </summary>
+	internal class ApplyQueryOptionExpression : QueryOptionExpression
+	{
+
+		/// <summary>
+		/// Creates an <see cref="ApplyQueryOptionExpression"/> expression.
+		/// </summary>
+		/// <param name="type">the return type of the expression.</param>
+		internal ApplyQueryOptionExpression(Type type)
+			: base(type)
+		{
+			this.Aggregations = new List<Aggregation>();
+		}
+
+		/// <summary>
+		/// The <see cref="ExpressionType"/> of the <see cref="Expression"/>.
+		public override ExpressionType NodeType
+		{
+			get { return (ExpressionType)ResourceExpressionType.ApplyQueryOption; }
+		}
+
+		/// <summary>
+		/// Aggregations in the $apply expression
+		/// </summary>
+		internal List<Aggregation> Aggregations { get; private set; }
+
+		/// <summary>
+		/// Structure for an aggregation. Holds lambda expression plus enum indicating aggregation method
+		/// </summary>
+		internal struct Aggregation
+		{
+			/// <summary>
+			/// Lambda expression for aggregation selector.
+			/// </summary>
+			internal readonly Expression Expression;
+
+			/// <summary>
+			/// Enum indicating aggregation method.
+			/// </summary>
+			internal readonly AggregationMethod AggregationMethod;
+
+			/// <summary>
+			/// Aggregation alias.
+			/// </summary>
+			internal readonly string AggregationAlias;
+
+			/// <summary>
+			/// Creates an aggregation.
+			/// </summary>
+			/// <param name="exp">Lambda expression for aggregation selector.</param>
+			/// <param name="aggregationMethod">Enum indicating aggregation method.</param>
+			internal Aggregation(Expression exp, AggregationMethod aggregationMethod)
+			{
+				this.Expression = exp;
+				this.AggregationMethod = aggregationMethod;
+				this.AggregationAlias = string.Empty;
+			}
+
+			/// <summary>
+			/// Creates an aggregation.
+			/// </summary>
+			/// <param name="exp">Lambda expression for aggregation selector.</param>
+			/// <param name="aggregationMethod">Enum indicating aggregation method.</param>
+			/// <param name="aggregationAlias">Aggregation alias.</param>
+			internal Aggregation(Expression exp, AggregationMethod aggregationMethod, string aggregationAlias)
+			{
+				this.Expression = exp;
+				this.AggregationMethod = aggregationMethod;
+				this.AggregationAlias = aggregationAlias;
+			}
+		}
+	}
+}
--- a/src/Microsoft.OData.Client/ALinq/DataServiceQueryProvider.cs
+++ b/src/Microsoft.OData.Client/ALinq/DataServiceQueryProvider.cs
@@ -11,9 +11,11 @@ namespace Microsoft.OData.Client
     using System;
     using System.Collections.Generic;
     using System.Diagnostics;
+    using System.IO;
     using System.Linq;
     using System.Linq.Expressions;
     using System.Reflection;
+    using System.Xml;
 
     #endregion Namespaces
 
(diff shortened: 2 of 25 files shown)


src/Microsoft.OData.Client/ALinq/UriWriter.cs
```diff
@@ -582,14 +585,85 @@ internal void VisitCustomQueryOptions(Dictionary<ConstantExpression, ConstantExp
...
+            if (aqoExpr.Aggregations.Count == 0)
+            {
+                return;
+            }
+
+            StringBuilder aggregateBuilder = new StringBuilder();
+
+            aggregateBuilder.Append(UriHelper.AGGREGATE);
+            aggregateBuilder.Append(UriHelper.LEFTPAREN);
+            int aggIdx = 0;
```
marabooy, line 603: nit: Maybe give this a more descriptive name :)

src/Microsoft.OData.Client/ALinq/UriWriter.cs
```diff
@@ -582,14 +585,85 @@ internal void VisitCustomQueryOptions(Dictionary<ConstantExpression, ConstantExp
             }
         }
 
+        /// <summary>
+        /// ApplyQueryOptionExpression visit method.
+        /// </summary>
+        /// <param name="aqoExpr">ApplyQueryOptionExpression expression to visit</param>
+        internal void VisitQueryOptionExpression(ApplyQueryOptionExpression aqoExpr)
```
marabooy, line 592: nit: name

marabooy (approved):
LGTM

src/Microsoft.OData.Client/ALinq/DataServiceQueryProvider.cs
```diff
@@ -152,5 +176,82 @@ internal QueryComponents Translate(Expression e)
...
+        /// <returns></returns>
+        private TElement ParseAggregateSingletonResult<TElement>(QueryResult queryResult)
+        {
+            IDictionary<string, string> responseHeaders = new Dictionary<string, string>();
+            responseHeaders.Add(ODataConstants.ContentTypeHeader, "application/json");
+            HttpWebResponseMessage httpWebResponseMessage = new HttpWebResponseMessage(responseHeaders, (int)queryResult.StatusCode, queryResult.GetResponseStream);
+
+            ODataMessageReaderSettings messageReaderSettings = new ODataMessageReaderSettings
+            {
+                Validations = ~ValidationKinds.ThrowOnUndeclaredPropertyForNonOpenType
```
habbes, line 217: What does the `~` operator do in this case?

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -2910,6 +2988,46 @@ internal static void ValidateExpandPath(Expression input, DataServiceContext con
...
+                        throw new NotSupportedException(Strings.ALinq_InvalidAggregateExpression(expr));
+                    }
+
+                    Type collectionType = ClientTypeUtil.GetImplementationType(parentExpr.Type, typeof(ICollection<>));
+                    if (collectionType != null)
+                    {
+                        throw new NotSupportedException(Strings.ALinq_InvalidAggregateExpression(expr));
+                    }
+                }
+            }
```
habbes, line 3030: What is it that's not supported? Is it accessing the length of properties like strings or collections or is it accessing nested properties? is `d1.Prop.ChildProp` supported? and is `d1.Rectangle.Length` supported (assuming `Square` is not a collection)?

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -2910,6 +2988,46 @@ internal static void ValidateExpandPath(Expression input, DataServiceContext con
 
                 throw new NotSupportedException(Strings.ALinq_InvalidExpressionInNavigationPath(input));
             }
+
+            /// <summary>
+            /// Checks whether the specified <paramref name="expr"/> is a valid aggregate expression.
```
habbes, line 2993: It would be good to know what counts as a valid aggregate expression.

src/Microsoft.OData.Client/ALinq/TypeSystem.cs
```diff
@@ -223,6 +227,13 @@ static TypeSystem()
                 typeof(Date).GetProperty("Day", typeof(int)).GetGetMethod());
 
             Debug.Assert(propertiesAsMethodsMap.Count == 24, "propertiesAsMethodsMap.Count == 24");
+
+            aggregationMethodMap.Add(AggregationMethod.Sum, "sum");
+            aggregationMethodMap.Add(AggregationMethod.Average, "average");
+            aggregationMethodMap.Add(AggregationMethod.Min, "min");
+            aggregationMethodMap.Add(AggregationMethod.Max, "max");
+            aggregationMethodMap.Add(AggregationMethod.CountDistinct, "countdistinct");
+            aggregationMethodMap.Add(AggregationMethod.VirtualPropertyCount, "$count");
```
habbes, line 236: Out of curiosity, why doesn't this use the string constants defined in `UriHelper`?

src/Microsoft.OData.Client/ALinq/UriWriter.cs
```diff
@@ -582,14 +585,85 @@ internal void VisitCustomQueryOptions(Dictionary<ConstantExpression, ConstantExp
...
+                aggregateBuilder.Append(aggregationAlias);
+
+                if (++i == applyQueryOptionExpr.Aggregations.Count)
+                {
+                    break;
+                }
+
+                aggregateBuilder.Append(UriHelper.COMMA);
+            }
+            aggregateBuilder.Append(UriHelper.RIGHTPAREN);
```
habbes, line 653: Maybe an empty line after the close `}`?

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -2910,6 +2988,46 @@ internal static void ValidateExpandPath(Expression input, DataServiceContext con
...
             }
+
+            /// <summary>
+            /// Checks whether the specified <paramref name="expr"/> is a valid aggregate expression.
+            /// </summary>
+            /// <param name="expr">The aggregate expression</param>
+            internal static void ValidateAggregateExpression(Expression expr)
+            {
+                MemberExpression memberExpr = StripTo<MemberExpression>(expr);
+
```
KenitoInc, line 2999: What does `StripTo` do?

src/Microsoft.OData.Client/DataServiceExtensions.cs
```diff
@@ -0,0 +1,156 @@
...
+            return source.Provider.Execute<int>(countMethodExpr);
+        }
+
+        /// <summary>
+        /// Returns the distinct count of elements in a sequence after applying the projection function to each element.
+        /// </summary>
+        /// <typeparam name="TSource">The type of the elements of <paramref name="source"/></typeparam>
+        /// <typeparam name="TTarget">The type returned by the projection function represented in <paramref name="selector"/>.</typeparam>
+        /// <param name="source">A sequence of values of type <typeparamref name="TSource"/>.</param>
+        /// <param name="selector">A projection function to apply to each element.</param>
```
KenitoInc, line 80: For the `CountDistinct` methods, the `selector` params description should be different. One is an `Expression`, the other is a `Func`.

src/Microsoft.OData.Client/DataServiceExtensions.cs
```diff
@@ -0,0 +1,156 @@
...
+        /// <returns>Distinct count of elements in a sequence after applying the projection function to each element.</returns>
+        public static int CountDistinct<TSource, TTarget>(this IEnumerable<TSource> source, Func<TSource, TTarget> selector)
+        {
+            // Provide a default implementation...
+            // In the event that a developer who adds a reference to the library invokes CountDistinct() as follows:
+            // - new List<int> { 1, 2, 1 }.CountDistinct(d => d)
+
+            // Method: Select<TSource,TResult>(IEnumerable<TSource>, Func<TSource,TResult>)
+            MethodInfo selectMethod = typeof(Enumerable).GetMethods(BindingFlags.Static | BindingFlags.Public)
```
KenitoInc, line 89: Is it possible to refractor the `GetSelectMethod` and move this logic there?

KenitoInc (commented):
A few comments/questions

src/Microsoft.OData.Client/ALinq/DataServiceQueryProvider.cs
gathogojr: @habbes The ~ operator performs a bitwise complement operation on its operand. In this case, we're configuring the message reader NOT to throw on undeclared property for non-open type.

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
gathogojr: @habbes Nested properties are supported. E.g. If you have a navigation property `Product`, you can group by property `ProductName` of `Product` type as follows:
```csharp
GroupBy(d1 => d1.Product.ProductName, ... )
```
What is not supported, is invoking properties of a known primitive type in the grouping expression. E.g. say `ProductName` in my example is of type `string`, it would be invalid to have
```csharp
GroupBy(d1 => d1.Product.ProductName.Length, ... )
```
We throw a `NotSupportedExpression` if that is used in the `GroupBy` `keySelector` expression

Note that this validation does not stop you from doing this:
```csharp
GroupBy(d1 => d1.Product.ProductName, (d1, d2) => new { ProductNameLength = d1.Length, ... });
```
Invoking the `Length` property (or any other property or method of the known primitive type) is just disallowed in the GroupBy `keySelector` expression.

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
gathogojr: In the specification, the aggregate expression has to meet [one of 5 criteria](http://docs.oasis-open.org/odata/odata-data-aggregation-ext/v4.0/cs02/odata-data-aggregation-ext-v4.0-cs02.html#_Toc435016568). I'll see how I can improve the docstring to give the user a better idea. The exception message tries to capture what is wrong with the exception.

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
gathogojr: @KenitoInc It strips the expression of intermediate expressions like Expression.Convert() and Expression.Quote(). See more [here](https://github.com/OData/odata.net/blob/d853504765436a1cd7b8df2b56598433e53d101b/src/Microsoft.OData.Client/ALinq/ResourceBinder.cs#L1180)

src/Microsoft.OData.Client/DataServiceExtensions.cs
gathogojr: The logic for getting the `Select` defined in `Enumerable` type differs a little from the one defined in `Queryable` type but I'll see how I can refactor and use distinct names for each.

gathogojr:
**FYI:**

When a `Where` method call appears before the aggregation method (e.g. `Average`, `Sum`, etc.), a `filter` transformation should be used to restrict the set of data to be aggregated. `$filter` query option, as well as other query options like `$orderby`, `$expand`, `$top`, `$skip`, etc., are not allowed before the `$apply` query option.

http://docs.oasis-open.org/odata/odata-data-aggregation-ext/v4.0/cs02/odata-data-aggregation-ext-v4.0-cs02.html#_Toc435016585

Example - Involving simple aggregation scenarios
```csharp
dataServiceContext.Sales.Where(d0 => d0.Amount > 1).Average(d1 => d1.Amount);
```
should be translated into
```
/Sales?$apply=filter(Amount gt 1)/aggregate(Amount with average as AverageAmount)
```
NOT
```
/Sales?$filter=Amount gt 1&$apply=aggregate(Amount with average as AverageAmount)
```
Take note of the `filter` transformation (under `$apply` with no `$` sign). I'll be checking in the logic to support that.

The `expand` transformation will be handled in a separate pull request

habbes (approved):

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
habbes: Maybe the comment or error-message should be more clear that the scenario being guarded against is accessing properties of known primitive types.

src/Microsoft.OData.Client/DataServiceExtensions.cs
gathogojr: @KenitoInc Note that in the official docs for `Queryable` and `Enumerable` methods, the descriptions don't differ much

`selector
Func<TSource,TResult>` -  [A transform function to apply to each element.](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.select?view=net-5.0#System_Linq_Enumerable_Select__2_System_Collections_Generic_IEnumerable___0__System_Func___0___1__)
`selector
Expression<Func<TSource,TResult>>` - [A projection function to apply to each element.](https://docs.microsoft.com/en-us/dotnet/api/system.linq.queryable.select?view=net-5.0#System_Linq_Queryable_Select__2_System_Linq_IQueryable___0__System_Linq_Expressions_Expression_System_Func___0___1___)

src/Microsoft.OData.Client/ALinq/ApplyQueryOptionExpression.cs
```diff
@@ -39,6 +45,47 @@ public override ExpressionType NodeType
...
+			this.filterExpressions.AddRange(predicates);
+		}
+
+		internal ReadOnlyCollection<Expression> PredicateConjuncts
+		{
+			get
+			{
+				return new ReadOnlyCollection<Expression>(this.filterExpressions);
+			}
+		}
```
habbes, line 62: Is this property access multiple times? And if so, is it expected to return a new collection each time?

src/Microsoft.OData.Client/ALinq/QueryableResourceExpression.cs
```diff
@@ -371,6 +371,56 @@ internal void AddFilter(IEnumerable<Expression> predicateConjuncts)
...
+                // http://docs.oasis-open.org/odata/odata-data-aggregation-ext/v4.0/cs02/odata-data-aggregation-ext-v4.0-cs02.html#_Toc435016590
+                
+                // If a Where appears before an aggregation method (e.g. Average, Sum, etc) or GroupBy, 
+                // the conjuncts of the filter expression will be used to restrict the set of data to be aggregated. 
+                // They will not appear on the $filter query option. Instead, we use them to construct a filter transformation.
+                // E.g. /Sales?$apply=filter(Amount gt 1)/aggregate(Amount with average as AverageAmount)
+                
+                // If a Where appears after an aggregation method or GroupBy, the conjuncts should appear
+                // on a $filter query option after the $apply.
+                // E.g. /Sales?$apply=groupby((Product/Color),aggregate(Amount with average as AverageAmount))&$filter=Product/Color eq 'Brown'
```
habbes, line 411: So there's support for having Where clauses both before and after an aggregation? In this case the generated query would contain both a filter transformation in the $apply clause as well as a $filter clause? Like `/Sales?$apply=filter(Amount gt 1)/groupby((Product/Color),aggregae(Amount with average as Avg))$filter=Product/Color eq 'Brown'`?
Is this a correct implication of what you've explained here?

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -797,6 +799,37 @@ internal static Expression ApplyOrdering(MethodCallExpression mce, bool descendi
             return input;
         }
 
+        private static Expression AnalyzeAggregation(MethodCallExpression methodCallExpr, AggregationMethod aggregationMethod)
+        {
+            Debug.Assert(methodCallExpr != null, "methodCallExpr != null");
+            if (methodCallExpr.Arguments.Count != 2)
```
odero, line 805: Why must the number of args be == 2?

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
odero: Have to admit I can't say I fully understand this line :)
>An aggregate expression must evaluate to a single-valued property path to an aggregatable property

test/FunctionalTests/Microsoft.OData.Client.Tests/TestHttpWebRequestMessage.cs
```diff
@@ -0,0 +1,60 @@
...
+
+	public TestHttpWebRequestMessage(DataServiceClientRequestMessageArgs args, IDictionary<string, string> headers, int statusCode, Func<Stream> getResponseStream)
+		: base(args)
+	{
+		this.headers = headers;
+		this.statusCode = statusCode;
+		this.getResponseStream = getResponseStream;
+	}
+
+#if (NETCOREAPP1_0 || NETCOREAPP2_0)
```
odero, line 39: Is this check necessary? What about core 3+?

odero (approved):

src/Microsoft.OData.Client/ALinq/QueryableResourceExpression.cs
gathogojr: @habbes You understood it correctly. A `$filter` (_as well as `$order` and `$select`_) query option can appear after the `$apply`. When it does, it'll be applied to the result of the aggregation.

As a matter of fact, a `filter` transformation can also appear after `groupby` and `aggregate` expression. When it does, it'll be applied to the result of the aggregation, such that the following two expressions give you the same result:

1. `/Sales?$apply=filter(Amount gt 1)/groupby((Product/Color),aggregate(Amount with average as AverageAmount))&$filter=Product/Color eq 'Brown'`
We're dealing with 2 query params here - `$apply` and `$filter` - separated by a `&`. The `$apply` here has 2 transformations - `filter` and `groupby` - separated by a forward slash (/). The `$filter` will be applied to the aggregated result.

2. `/Sales?$apply=filter(Amount gt 1)/groupby((Product/Color),aggregate(Amount with average as AverageAmount))/filter(Product/Color eq 'Brown')`
We're dealing with a single query param here - `$apply`. It has 3 transformations separated by a forward slash (/). The `filter` transformation before `groupby` restricts the set of data to be aggregated while the one that appears after is applied to the aggregated result

That said, I deliberately decided to leave out  support for `filter` transformation **appearing after** the `groupby` and `aggregate` transformation as well as system query options **appearing after** `$apply`. The effort required to make them play is not trivial. That work will tackled in a separate work item/pull request

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
gathogojr: @odero Aggregation methods - `Average`, `Sum`, `Min`, `Max`) defined in `Queryable` and `Enumerable` classes have two sets of overloads for each aggregable type (`double`, `double?`, `float`, `float?`, `int`, `int?`, `long`, `long?`, `decimal`, `decimal?`)
- The first overload takes one parameter. Examples: `Average(IQueryable<Decimal>)` or `Average(IEnumerable<Decimal>)`.
To use this overload, you'd write the expression as follows:
    ```csharp
    context.Sales.Select(d1 => d1.Amount).Average(); // => Average(IEnumerable<Decimal>)
    ```
- The second overload takes two parameters. Example: `Average<TSource>(IQueryable<TSource>, Expression<Func<TSource,Decimal>>)`.
To use this overload, you'd write the expression as follows:
    ```csharp
    context.Sales.Average(d1 => d1.Amount);
    ```
**I implemented support for the second overload** in this pull request. That's the reason we're debug-asserting to verify that we're dealing with the second overload. It is succinct and requires the least amount of effort and almost no refactor to fit into the existing logic.

Implementing the first overload would be a lot of pain with little or no additional gain. Among other reasons, the `Select()` method is currently translated into `$select` expression. We'd need to do quite a bit of messy plumbing and refactor to handle `Select()` differently depending on whether or not its followed by an aggregation method, i.e.:
- Translate `Select(d1 => d1.Amount)` into `$select=Amount`, and
- Translate `Select(d1 => d1.Amount).Average()` into `$apply=aggregate(Amount with average as AverageAmount)`

The strategy would be to "look ahead" when doing the translation to determine which expression to translate into.
That support could be implemented at a later date. The effort would not be trivial so we'd need to have a strong justification.

We'd obviously need to document what is supported for the benefit of the users.

gathogojr:
@odero 
RE: https://github.com/OData/odata.net/pull/1921#discussion_r530794617

Consider the following class:
```c#
class Sale
{
    public string Currency { get; set; }
    public Amount Amount { get; set; }
    public Product Product { get; set; }
    public Collection<Review> Reviews { get; set; }
}
```
- `Currency`, `Amount` and `Product` are single-valued properties. `Reviews` is a collection property.
- On the subject of aggregatable (or is it aggregable?), only `Amount` fits the bill - think `average`, `sum`, `min`, and `max`. You can't aggregate `Currency` - of type string, or `Product` - of type Product.
- However, `Currency` would qualify as aggregatable property when using `countdistinct` aggregation.

The fact that I have used this many words and an illustration to explain tells you how hard it is to come up with a good docstring for the method :)
However, if you go through the method there are helpful comments that cover the scenarios we're validating for.

src/Microsoft.OData.Client/ALinq/ApplyQueryOptionExpression.cs
```diff
@@ -39,6 +45,47 @@ public override ExpressionType NodeType
...
+				if (isFirst)
+				{
+					combinedPredicate = expr;
+					isFirst = false;
+				}
+				else
+				{
+					combinedPredicate = Expression.And(combinedPredicate, expr);
+				}
+			}
```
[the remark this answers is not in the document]
gathogojr: I like it :)

src/Microsoft.OData.Client/ALinq/ApplyQueryOptionExpression.cs
gathogojr: @habbes ReadOnlyCollection is just a wrapper on top of the `filterExpressions` list. It does not create a new list
https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.readonlycollection-1?view=net-5.0#examples

src/Microsoft.OData.Client/ALinq/QueryableResourceExpression.cs
```diff
@@ -371,6 +371,56 @@ internal void AddFilter(IEnumerable<Expression> predicateConjuncts)
...
+            {
+                throw new NotSupportedException(Strings.ALinq_QueryOptionOutOfOrder("apply", "orderby"));
+            }
+            else if (this.Skip != null)
+            {
+                // $skip and $top may be used together with rollup. 
+                // However, support for rollup is currently not implemented in OData WebApi
+                // If $skip and/or $top appears before $apply, its currently ignored. 
+                // Makes sense to throw an exception to avoid giving a false impression.
+                throw new NotSupportedException(Strings.ALinq_QueryOptionOutOfOrder("apply", "skip"));
```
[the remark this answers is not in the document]
gathogojr: @habbes I'll sidebar with you on this

test/FunctionalTests/Microsoft.OData.Client.Tests/TestHttpWebRequestMessage.cs
gathogojr: @odero Maybe not... I'll remove it and see how the tests behave

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
gathogojr: @odero 
Consider the following class:
```c#
class Sale
{
    public string Currency { get; set; }
    public Amount Amount { get; set; }
    public Product Product { get; set; }
    public Collection<Review> Reviews { get; set; }
}
```
- `Currency`, `Amount` and `Product` are single-valued properties. `Reviews` is a collection property.
- On the subject of aggregatable (or is it aggregable?), only `Amount` fits the bill - think `average`, `sum`, `min`, and `max`. You can't aggregate `Currency` - of type string, or `Product` - of type Product.
- However, `Currency` would qualify as aggregatable property when using `countdistinct` aggregation.

The fact that I have used this many words and an illustration to explain tells you how hard it is to come up with a good docstring for the method :)
However, if you go through the method there are helpful comments that cover the scenarios we're validating for.

src/Microsoft.OData.Client/ALinq/ApplyQueryOptionExpression.cs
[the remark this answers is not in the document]
odero: tbf I find the previous one more readable

src/Microsoft.OData.Client/ALinq/QueryableResourceExpression.cs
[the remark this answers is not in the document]
gathogojr: @habbes @odero Just to clarify, ODL doesn't support `rollup` either.
The `rollup` grouping operator allows requesting additional levels of aggregation in addition to the most granular level defined by the grouping properties. It can be used instead of a property path in the first parameter of `groupby`. To me its a feature for the future if ever.

Here's a link to the section of the spec that describes how it works http://docs.oasis-open.org/odata/odata-data-aggregation-ext/v4.0/cs02/odata-data-aggregation-ext-v4.0-cs02.html#_Toc435016584

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -797,6 +799,37 @@ internal static Expression ApplyOrdering(MethodCallExpression mce, bool descendi
...
 
+        private static Expression AnalyzeAggregation(MethodCallExpression methodCallExpr, AggregationMethod aggregationMethod)
+        {
+            Debug.Assert(methodCallExpr != null, "methodCallExpr != null");
+            if (methodCallExpr.Arguments.Count != 2)
+            {
+                return methodCallExpr;
+            }
+
+            QueryableResourceExpression input;
```
xuzhg, line 810: >input [](start = 40, length = 5)

maybe rename it as "output", it's weird to use "input" as output. maybe other name

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -797,6 +799,37 @@ internal static Expression ApplyOrdering(MethodCallExpression mce, bool descendi
             return input;
         }
 
+        private static Expression AnalyzeAggregation(MethodCallExpression methodCallExpr, AggregationMethod aggregationMethod)
```
xuzhg, line 802: >Analyze [](start = 34, length = 7)

Apply

src/Microsoft.OData.Client/DataServiceQueryOfT.cs
```diff
@@ -386,17 +384,10 @@ public virtual DataServiceQuery<TElement> AddQueryOption(string name, object val
 
         /// <summary>Executes the query and returns the results as a collection.</summary>
         /// <returns>A typed enumerator over the results in which TElement represents the type of the query results.</returns>
-#if !PORTABLELIB // Synchronous methods not available
```
xuzhg, line 389: >#if !PORTABLELIB [](start = 0, length = 16)

thanks for removing this

src/Microsoft.OData.Client/DataServiceExtensions.cs
```diff
@@ -0,0 +1,173 @@
...
+        /// <typeparam name="TSource">The type of the elements of <paramref name="source"/></typeparam>
+        /// <typeparam name="TTarget">The type returned by the projection function represented in <paramref name="selector"/>.</typeparam>
+        /// <param name="source">A sequence of values of type <typeparamref name="TSource"/>.</param>
+        /// <param name="selector">A projection function to apply to each element.</param>
+        /// <returns>Distinct count of elements in a sequence after applying the projection function to each element.</returns>
+        public static int CountDistinct<TSource, TTarget>(this IQueryable<TSource> source, Expression<Func<TSource, TTarget>> selector)
+        {
+            MethodCallExpression countMethodExpr;
+
+            if (source.Provider.GetType().Equals(typeof(DataServiceQueryProvider)))
```
xuzhg, line 33: >source.Provider.GetType().Equals(typeof(DataServiceQueryProvider))) [](start = 16, length = 67)

Could the provider type be the sub type of "DataServiceQueryProvider"?

src/Microsoft.OData.Client/DataServiceExtensions.cs
```diff
@@ -0,0 +1,173 @@
...
+                    && d3.Parameters[0].ParameterType.GetGenericTypeDefinition().Equals(typeof(IEnumerable<>))
+                    && d3.Parameters[1].ParameterType.IsGenericType
+                    && d3.Parameters[1].ParameterType.GetGenericTypeDefinition().Equals(typeof(Func<,>)))
+                .Select(d6 => d6.Method).Single();
+        }
+
+        /// <summary>
+        /// Returns Distinct method defined in <paramref name="declaringType"/>.
+        /// </summary>
+        private static MethodInfo GetDistinctMethod(Type declaringType, Type sourceType)
```
xuzhg, line 148: >GetDistinctMethod [](start = 34, length = 17)

should be static (cache) it?

src/Microsoft.OData.Client/DataServiceExtensions.cs
```diff
@@ -0,0 +1,173 @@
...
+
+        /// <summary>
+        /// Returns the distinct count of elements in a sequence after applying the projection function to each element.
+        /// </summary>
+        /// <typeparam name="TSource">The type of the elements of <paramref name="source"/></typeparam>
+        /// <typeparam name="TTarget">The type returned by the projection function represented in <paramref name="selector"/>.</typeparam>
+        /// <param name="source">A sequence of values of type <typeparamref name="TSource"/>.</param>
+        /// <param name="selector">A transform function to apply to each element.</param>
+        /// <returns>Distinct count of elements in a sequence after applying the projection function to each element.</returns>
+        public static int CountDistinct<TSource, TTarget>(this IEnumerable<TSource> source, Func<TSource, TTarget> selector)
```
xuzhg, line 82: >IEnumerable<TSource> source [](start = 63, length = 27)

should we check the null?

src/Microsoft.OData.Client/DataServiceExtensions.cs
xuzhg, line 33: should we check the null?

test/FunctionalTests/Microsoft.OData.Client.Tests/TestHttpWebRequestMessage.cs
```diff
@@ -0,0 +1,56 @@
...
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace Microsoft.OData.Client.Tests
+{
+    public class TestHttpWebRequestMessage : HttpWebRequestMessage
+{
```
xuzhg, line 16: align

src/Microsoft.OData.Client/ALinq/UriHelper.cs
```diff
@@ -83,6 +83,21 @@ internal static class UriHelper
...
+        /// <summary>apply</summary>
+        internal const string OPTIONAPPLY = "apply";
+
+        /// <summary>groupby</summary>
+        internal const string GROUPBY = "groupby";
+
+        /// <summary>aggregate</summary>
+        internal const string AGGREGATE = "aggregate";
+
+        /// <summary>aggregate</summary>
```
xuzhg, line 95: with  
as

src/Microsoft.OData.Client/ALinq/UriHelper.cs
```diff
@@ -158,6 +173,24 @@ internal static class UriHelper
...
+        internal const string MIN = "min";
+
+        /// <summary>max</summary>
+        internal const string MAX = "max";
+
+        /// <summary>countdistinct</summary>
+        internal const string COUNTDISTINCT = "countdistinct";
+
+        /// <summary>$count</summary>
+        internal const string VIRTUALPROPERTYCOUNT = "$count";
```
xuzhg, line 192: >VIRTUALPROPERTYCOUNT [](start = 30, length = 20)

why has a name "virtual property count" equals "$count"?

src/Microsoft.OData.Client/ALinq/UriWriter.cs
```diff
@@ -582,14 +585,121 @@ internal void VisitCustomQueryOptions(Dictionary<ConstantExpression, ConstantExp
...
+        /// <returns>The aggregate tranformation.</returns>
+        private string ConstructAggregateTransformation(IList<ApplyQueryOptionExpression.Aggregation> aggregations)
+        {
+            StringBuilder aggregateBuilder = new StringBuilder();
+
+            aggregateBuilder.Append(UriHelper.AGGREGATE);
+            aggregateBuilder.Append(UriHelper.LEFTPAREN);
+            int i = 0;
+
+            while (true)
```
xuzhg, line 640: > while (true) [](start = 11, length = 13)

why don't use foreach (...) ?

xuzhg (approved):
:shipit:

xuzhg:
me too,

---
In reply to: [530944405](https://github.com/OData/odata.net/pull/1921#issuecomment-530944405) [](ancestors = 530944405)

src/Microsoft.OData.Client/ALinq/UriHelper.cs
gathogojr: @xuzhg I borrowed this one from ODL https://github.com/OData/odata.net/blob/576c4ab0337bb1cecda85f4199bcc406d55b787b/src/Microsoft.OData.Core/UriParser/Aggregation/AggregationMethod.cs#L32

src/Microsoft.OData.Client/DataServiceExtensions.cs
gathogojr: @xuzhg Thanks to your observation here. As a result of looking at it afresh I was able to significantly reduce the code in the extension method

src/Microsoft.OData.Client/DataServiceExtensions.cs
gathogojr: @xuzhg Used `SimpleLazy` class, let me know if its a good fit

src/Microsoft.OData.Client/ALinq/UriWriter.cs
gathogojr: @xuzhg This is one scenario where traversing the collection using the indexer worked better. In a single statement, we're able increment the counter and check if its time to exit to avoid adding a trailing comma. The trailing comma would make the Uri invalid. Re:
```csharp
    if (++i == aggregations.Count)
    {
        break;
    }

    aggregateBuilder.Append(UriHelper.COMMA);
```

Issue #1538, opened 2019-09-07
Type cast path segments are duplicated when serializing nested resources causing high CPU load.


dmitrylobkov:
When serializing derived types with nested complex\navigation properties TypeSegment-s are being added to ODataWriterCore.CurrentScope.ODataUri.Path on entering a corresponding scope and are never freed on leaving the scope, i.e. the segments are added for every nested resource of every entity being processed.

When dealing with models similar to msgraph beta 'group' (we actually do:), there might be about 20k-30k duplicated type segments on 1k page response serialization. Since the path is constantly being reallocated, it causes huge load on CPU. 

### Assemblies affected

Microsoft.OData.Core  <= 7.6.0

### Reproduce steps

- Define the following entities:
```
public class Entity 
{
   public string Id { get; set; }
}
public class User : Entity 
{
    public Entity[] MemberOf { get; set; }
} 
public class Group : Entity 
{
    /* 25-30 complex\navigation properties */ 
}
```

- Controller action:
```
[ODataRoute("users/{id}/memberof")]
public IActionResult GetMemberOf(string id)
{
   return Ok(Enumerable
             .Range(1, 1000)
             .Select(item => new Group() { Id = item.ToString() })
             .OfType<Entity>());
}

```
- Invoke the controller action
### Expected result
The response should be serialized in several milliseconds.

### Actual result
The response serialization takes more than 20sec on I7-8700, CPU is overloaded.

### Additional detail
Repro source:
[TestProject.zip](https://github.com/OData/odata.net/files/3587124/TestProject.zip)

We've made a quick fix for the issue, the same response is serialized in 100-200 ms. Here is the patch: 
[typecastfix.zip](https://github.com/OData/odata.net/files/3587141/typecastfix.zip)

dmitrylobkov:
@KanishManuja-MS @madansr7 Hi guys) Any change it will be fixed? I believe the issue might affect msgraph services if they are built on top of odata.net.

mikepizzo:
Sorry that we missed this.  Easy win.  Thanks for tracking down the root cause and creating a patch.

PR #1930.

PR #1930, opened 2020-11-17
Address perf issue with Appending Cast Segment

mikepizzo:
### Issues

*This pull request fixes issue #1538*

### Description

Creates a new ODatPath rather than repeatedly adding redundant cast segments to the existing path.

### DIFF
--- a/src/Microsoft.OData.Core/ODataWriterCore.cs
+++ b/src/Microsoft.OData.Core/ODataWriterCore.cs
@@ -2339,10 +2339,10 @@ namespace Microsoft.OData
 
                             if (resourceTypeCast != null)
                             {
-                                odataPath = odataPath.AppendSegment(resourceTypeCast);
+                                odataPath = odataPath.AddSegment(resourceTypeCast);
                             }
 
-                            odataPath = odataPath.AppendPropertySegment(structuredProperty);
+                            odataPath = odataPath.AddPropertySegment(structuredProperty);
 
                             derivedTypeConstraints = this.outputContext.Model.GetDerivedTypeConstraints(structuredProperty);
                         }
@@ -2369,7 +2369,7 @@ namespace Microsoft.OData
 
                                 if (resourceTypeCast != null)
                                 {
-                                    odataPath = odataPath.AppendSegment(resourceTypeCast);
+                                    odataPath = odataPath.AddSegment(resourceTypeCast);
                                 }
 
                                 navigationSource = currentNavigationSource == null
@@ -2398,12 +2398,12 @@ namespace Microsoft.OData
 
                                         if (odataPath != null && typeCastFromExpand != null)
                                         {
-                                            odataPath = odataPath.AppendSegment(typeCastFromExpand);
+                                            odataPath = odataPath.AddSegment(typeCastFromExpand);
                                         }
 
                                         Debug.Assert(navigationSource is IEdmContainedEntitySet, "If the NavigationSourceKind is ContainedEntitySet, the navigationSource must be IEdmContainedEntitySet.");
                                         IEdmContainedEntitySet containedEntitySet = (IEdmContainedEntitySet)navigationSource;
-                                        odataPath = odataPath.AppendNavigationPropertySegment(containedEntitySet.NavigationProperty, containedEntitySet);
+                                        odataPath = odataPath.AddNavigationPropertySegment(containedEntitySet.NavigationProperty, containedEntitySet);
                                         break;
                                     case EdmNavigationSourceKind.EntitySet:
                                         odataPath = new ODataPath(new EntitySetSegment(navigationSource as IEdmEntitySet));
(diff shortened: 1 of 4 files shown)


src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPathExtensions.cs
```diff
@@ -55,6 +55,19 @@ public static bool IsCollection(this ODataPath path)
...
 
+        /// <summary>
+        /// Creates a new ODataPath with the specified segment appended
+        /// </summary>
+        /// <param name="path">The path against which the segment should be applied.</param>
+        /// <param name="segment">The segment applied to the path.</param>
+        /// <returns>A new ODataPath with the segment appended.</returns>
+        public static ODataPath AppendSegment(this ODataPath path, ODataPathSegment segment)
+        {
+            var newPath = new ODataPath(path);
```
xuzhg, line 66: Append* sounds to insert the item at the end and to return the same "path"?

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -2339,7 +2339,7 @@ private void EnterScope(WriterState newState, ODataItem item)
 
                             if (resourceTypeCast != null)
                             {
-                                odataPath.Add(resourceTypeCast);
+                                odataPath = odataPath.AppendSegment(resourceTypeCast);
```
xuzhg, line 2342: >odataPath = odataPath.AppendSegment(resourceTypeCast); [](start = 32, length = 54)

what i mean is that:

If it's "append", we don't need return a new ODataPath.

Can it be directly "  odataPath.AppendSegment(resourceTypeCast);"

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPathExtensions.cs
mikepizzo: Yes' Append does sound like it just adds to the end, but this naming is consistent with AppendEntityKeySegment, AppendNavigationPropertySegment, AppendPropertySegment, and AppendKeySegment, so I kept the naming consistent, rather than changing it everywhere.

---
In reply to: [532791397](https://github.com/OData/odata.net/pull/1930#discussion_r532791397) [](ancestors = 532791397)

src/Microsoft.OData.Core/ODataWriterCore.cs
mikepizzo: See other comment.  I just kept this consistent with the existing naming convention, rather than changing it everywhere, which I could do.

---
In reply to: [539590070](https://github.com/OData/odata.net/pull/1930#discussion_r539590070) [](ancestors = 539590070)

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
```diff
@@ -627,20 +627,16 @@ public static class ValidationRules
                     IEdmSchemaType schemaType = structuredType as IEdmSchemaType;
                     if (schemaType != null)
                     {
-                        var properties = structuredType.Properties().ToList();
-                        if (properties.Count > 0)
+                        foreach (var property in structuredType.Properties())
```
Sreejithpin, line 630: Can we use IEdmProperty instead of var

src/Microsoft.OData.Core/ODataWriterCore.cs
mikepizzo: Renamed all of these to Add*.

---
In reply to: [542970463](https://github.com/OData/odata.net/pull/1930#discussion_r542970463) [](ancestors = 542970463,539590070)

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
mikepizzo: Good catch -- thanks!

---
In reply to: [543543057](https://github.com/OData/odata.net/pull/1930#discussion_r543543057) [](ancestors = 543543057)

xuzhg (approved):
:shipit:

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPathExtensions.cs
```diff
@@ -55,46 +55,55 @@ public static bool IsCollection(this ODataPath path)
...
 
+        /// <summary>
+        /// Creates a new ODataPath with the specified segment added.
+        /// </summary>
+        /// <param name="path">The path against which the segment should be applied.</param>
+        /// <param name="segment">The segment applied to the path.</param>
+        /// <returns>A new ODataPath with the segment appended.</returns>
+        public static ODataPath AddSegment(this ODataPath path, ODataPathSegment segment)
+        {
+            var newPath = new ODataPath(path);
```
xuzhg, line 66: > var newPath = new ODataPath(path); [](start = 11, length = 35)

It's not a big deal. The methods in this class don't check the nullability of the arguments?

mikepizzo: The class is internal, so an assert (if anything) would probably be more appropriate.

---
In reply to: [543923833](https://github.com/OData/odata.net/pull/1930#discussion_r543923833) [](ancestors = 543923833)

PR #1932, opened 2020-11-19
Fix reading OData error at synchronous operations for .NET Core.

tfaller:
### Issues

None

### Description

During synchronous operations like "SaveChanges" an OData error was not successfully unmarshalled when .NET Core or .NET 5 is used. If the server returns 400 and an OData error, the excepetion message is simply "Bad Request".
This PR fixes this problem. The problem did not occur at async operations or when .NET Framework was used.

The method to read OData erros needs to read the response stream twice. This was not a problem during async operations or using .NET Framework, because the response was copied to a MemoryStream so that the stream is seekable. However during synchronous operations using .NET Core or .NET 5, the response stream is not buffered. The method was not able to read the stream twice. This is because .NET Framework seems to handle HttpWebResponse streams internally a bit different than .NET Core and .NET 5.

### Additional work necessary

None

### DIFF
--- a/src/Microsoft.OData.Client/BaseSaveResult.cs
+++ b/src/Microsoft.OData.Client/BaseSaveResult.cs
@@ -244,10 +244,21 @@ namespace Microsoft.OData.Client
 
             IDictionary<string, string> headers = new Dictionary<string, string>();
 
-            using (Stream stream = getResponseStream())
+            using (Stream responseStream = getResponseStream())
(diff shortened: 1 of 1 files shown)


xuzhg (approved):

mikepizzo (approved):
:shipit:

mikepizzo (approved):
:shipit:

PR #1933, opened 2020-11-25
Fix deserialization of decimal in exponential notation

gathogojr:
### Issues

*This pull request fixes OData/WebApi#2125.*

### Description

Fix deserialization of untyped values in exponential notation into decimals

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataPayloadValueConverter.cs
+++ b/src/Microsoft.OData.Core/ODataPayloadValueConverter.cs
@@ -88,16 +88,7 @@ namespace Microsoft.OData
                 }
                 else if (value is Double)
                 {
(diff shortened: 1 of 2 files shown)


KenitoInc (approved):
If 7.5e3m is equal to 7500, you can add a test for that.

odero (approved):

gathogojr:
> If 7.5e3m is equal to 7500, you can add a test for that.

@KenitoInc Added a test for this, though when the untyped value is 7500 then it's first read as an integer before being converted to decimal. That was working okay even before. The issue affected untyped values that were first read as double

marabooy (approved):

xuzhg (approved):

PR #1941, opened 2020-12-02
add optional annotation location parameter to AddVocabularyAnnotation

chrisspre:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Schema/EdmModel.cs
+++ b/src/Microsoft.OData.Edm/Schema/EdmModel.cs
@@ -119,7 +119,7 @@ namespace Microsoft.OData.Edm
         /// Adds a vocabulary annotation to this model.
         /// </summary>
         /// <param name="annotation">The annotation to be added.</param>
-        public void AddVocabularyAnnotation(IEdmVocabularyAnnotation annotation)
+        public void AddVocabularyAnnotation(IEdmVocabularyAnnotation annotation, Csdl.EdmVocabularyAnnotationSerializationLocation location = Csdl.EdmVocabularyAnnotationSerializationLocation.OutOfLine)
         {
             EdmUtil.CheckArgumentNull(annotation, "annotation");
             if (annotation.Target == null)
@@ -135,6 +135,7 @@ namespace Microsoft.OData.Edm
             }
 
             elementAnnotations.Add(annotation);
+            Csdl.SerializationExtensionMethods.SetSerializationLocation(annotation, this, location);
         }
 
         /// <summary>


src/Microsoft.OData.Edm/Schema/EdmModel.cs
```diff
@@ -119,7 +119,7 @@ public void AddElements(IEnumerable<IEdmSchemaElement> newElements)
         /// Adds a vocabulary annotation to this model.
         /// </summary>
         /// <param name="annotation">The annotation to be added.</param>
-        public void AddVocabularyAnnotation(IEdmVocabularyAnnotation annotation)
```
xuzhg, line 122: Hi, FYI, ODL has the extension method to set the location for annotation.

public static void SetSerializationLocation(this IEdmVocabularyAnnotation annotation, IEdmModel model, EdmVocabularyAnnotationSerializationLocation? location)

If we think add a parameter for "AddVocabularyAnnotation", we can add an extension method to "EdmModel" as `public void AddVocabularyAnnotation(this EdmModel model, IEdmVocabularyAnnotation annotation, Csdl.EdmVocabularyAnnotationSerializationLocation location)`

xuzhg (changes requested):

src/Microsoft.OData.Edm/Schema/EdmModel.cs
chrisspre: I understand that the extension method is there. It is actually used in the implementation. 

Woudl you prefer the extension method? For the caller there is actually no difference due to the fact that the location parameter is optional.

chrisspre: moved to extension method

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -2027,6 +2027,18 @@ public static void AddAlternateKeyAnnotation(this EdmModel model, IEdmEntityType
             model.SetVocabularyAnnotation(annotation);
         }
 
+        /// <summary>
+        /// Adds a vocabulary annotation to this model.
+        /// </summary>
+        /// <param name="model">The model the annotation is added to.</param>
+        /// <param name="annotation">The annotation to be added.</param>
+        /// <param name="location">The location to add the allocation</param>
+        public static void AddVocabularyAnnotation(this EdmModel model, IEdmVocabularyAnnotation annotation, Csdl.EdmVocabularyAnnotationSerializationLocation location)
```
xuzhg, line 2036: >IEdmVocabularyAnnotation annotation, Csdl.EdmVocabularyAnnotationSerializationLocation location [](start = 72, length = 95)

It's not a deal for me to add such extensions. Because, there's only two line codes .

but, i have a question here:   How can handle this scenario: 

annotation itself has the Location setting, and that setting is different with the "location" given?

chrisspre: I wasn't aware the IEdmVocabularyAnnotation  has a location. Will look into it.

Issue #1943, opened 2020-12-03
Error when materializing an open entity type when a dynamic property is returned with a value of null


gathogojr:
If an OData service returns an open entity type when a dynamic property is returned with a value of `null`, attempt to materialize that value into the dynamic properties container property results into an error

### Assemblies affected
Microsoft.OData.Client 7.7.3

### Reproduce steps
1. Create an OData service that returns an open entity type with dynamic property with a `null` value
2. Create an OData client and generate the proxy using OData Connected Service extension
3. Write code to query the entity

### Expected result
The entity is materialized successfully on the client

### Actual result

An null reference exception is thrown
![image](https://user-images.githubusercontent.com/28291332/100909697-1a168680-34de-11eb-8206-e2f212f06394.png)

### Additional detail
OData WebApi currently does not write a dynamic property into the response if the value is `null`.

gathogojr:
Fix merged

PR #1944, opened 2020-12-03
Fix null dynamic property materialization

gathogojr:
### Issues

*This pull request fixes issue #1943.*

### Description

If an OData service returns an open entity type when a dynamic property is returned with a value of null, attempt to materialize that value into the dynamic properties container property results into an error. This pull request fixes that.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/ODataDynamicPropertyTests.cs
+++ b/test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/ODataDynamicPropertyTests.cs
@@ -488,6 +488,18 @@ namespace AstoriaUnitTests.TDD.Tests.Client
         }
 
         [Fact]
+        public void SerializationWithNullValueAsDynamicProperty()
+        {
+            var director = new Director { Id = 1, Name = "Director 1" };
+            director.DynamicProperties.Add("Title", null);
+
+            var messageBody = SerializeEntity(this.directorEntitySetName, director);
+            var expectedResult = "{\"@odata.type\":\"#ServiceNS.Director\",\"Id\":1,\"Name\":\"Director 1\"}";
+
+            VerifyMessageBody(expectedResult, messageBody);
+        }
+
+        [Fact]
         public void MaterializationWithNoDynamicProperty()
         {
             var rawJsonResponse = "{" +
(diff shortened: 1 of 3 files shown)


marabooy (approved):

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/ODataDynamicPropertyTests.cs
```diff
@@ -487,6 +487,18 @@ public void SerializationIgnoresContainerPropertyIfEdmTypeIsNotOpen()
...
         }
 
+        [Fact]
+        public void SerializationWithNullValueAsDynamicProperty()
+        {
+            var director = new Director { Id = 1, Name = "Director 1" };
+            director.DynamicProperties.Add("Title", null);
+
+            var messageBody = SerializeEntity(this.directorEntitySetName, director);
+            var expectedResult = "{\"@odata.type\":\"#ServiceNS.Director\",\"Id\":1,\"Name\":\"Director 1\"}";
```
xuzhg, line 497: > var expectedResult = "{\"@odata.type\":\"#ServiceNS.Director\",\"Id\":1,\"Name\":\"Director 1\"}"; [](start = 11, length = 99)

do we support to upload a dynamic property with null value in client?

xuzhg (approved):
:shipit:

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/ODataDynamicPropertyTests.cs
gathogojr: @xuzhg Based on the spec, [a missing dynamic property is defined to be the same as a dynamic property with value null](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_SetaValuetoNull). I therefore reasoned there's no reason to send dynamic properties with null values over the wire. We don't serialize dynamic properties with null values in OData WebAPI either. This test verifies that if we come across such a property in the dynamic properties container property, we just skip it.

xuzhg: For example:  if we want to remove a dynamic property value as null for a certain entity, we should "patch" that entity with "null" dynamic property. How can we do this in Client?

gathogojr: That would be achieved by using a declared property on the client side. This way the type annotation will be written in the payload for that dynamic property. Plus it'd make it easy to do property tracking for that property.

Given that the container property is a dictionary of `string` and `object`, we wouldn't know the type annotation to write if we came across `KeyValuePair<string, object>` with a `null` value during serialization. We'd have no problem serializing it as `"propertyName":null`, but it'd be minus type annotation. I'm not confident the service would handle it gracefully with type info missing. Which is why I don't think the container property is a good fit for this scenario.

PR #1949, opened 2020-12-08
key attribute with one property throws exception

ElizabethOkerio:
### Issues

If I have such a  POCO class with only one property when working with <code> Microsoft.OData.Client </code> 
image

An exception will always be thrown because this condition will always be true  as the <code> Microsoft.OData.Client.KeyAttribute </code> is null. 

<code> if (newKeyKind == KeyKind.AttributedKey && keyProperties.Count != dataServiceKeyAttribute?.KeyNames.Count) </code> 

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
+++ b/src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
@@ -489,14 +489,17 @@ namespace Microsoft.OData.Client.Metadata
                 }
             }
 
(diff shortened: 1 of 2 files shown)


KenitoInc (commented):
Is there a way we can add a test case for this?

xuzhg:
+1

---
In reply to: [547226245](https://github.com/OData/odata.net/pull/1949#issuecomment-547226245) [](ancestors = 547226245)

KenitoInc (approved):
LGTM

mikepizzo (approved):
:shipit:

Issue #1938, opened 2020-12-02
The target of navigation property binding is not valid format


xuzhg:
*Short summary (3-5 sentences) describing the issue.*

### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData .Net lib 7.x*

### Reproduce steps

```C#
var model = new EdmModel();

            var areaEntityType = new EdmEntityType("NS", "Area");
            var key = areaEntityType.AddStructuralProperty("Id", EdmPrimitiveTypeKind.String);
            areaEntityType.AddStructuralProperty("Name", EdmPrimitiveTypeKind.String);
            areaEntityType.AddKeys(key);

            var plantEntityType = new EdmEntityType("NS", "Plant");
            key = plantEntityType.AddStructuralProperty("Id", EdmPrimitiveTypeKind.String);
            plantEntityType.AddStructuralProperty("Name", EdmPrimitiveTypeKind.String);
            plantEntityType.AddKeys(key);

            var siteEntityType = new EdmEntityType("NS", "Site");
            key = siteEntityType.AddStructuralProperty("Id", EdmPrimitiveTypeKind.String);
            siteEntityType.AddStructuralProperty("Name", EdmPrimitiveTypeKind.String);
            siteEntityType.AddKeys(key);

            // Contained entity sets
            plantEntityType.AddUnidirectionalNavigation(new EdmNavigationPropertyInfo()
            {
                ContainsTarget = true,
                Name = "Areas",
                Target = areaEntityType,
                TargetMultiplicity = EdmMultiplicity.Many
            });

            var sitePlantsNavigationProperty = siteEntityType.AddUnidirectionalNavigation(new EdmNavigationPropertyInfo()
            {
                ContainsTarget = true,
                Name = "Plants",
                Target = plantEntityType,
                TargetMultiplicity = EdmMultiplicity.Many
            });

            // Non-contained navigation property back to plant
            var areaPlantNavigationProperty = areaEntityType.AddUnidirectionalNavigation(new EdmNavigationPropertyInfo()
            {
                ContainsTarget = false,
                Name = "Plant",
                Target = plantEntityType,
                TargetMultiplicity = EdmMultiplicity.One
            });

            model.AddElement(areaEntityType);
            model.AddElement(plantEntityType);
            model.AddElement(siteEntityType);

            var entityContainer = new EdmEntityContainer("NS", "MyApi");
            var sitesEntitySet = entityContainer.AddEntitySet("Sites", siteEntityType);
[...]

PR #1950, opened 2020-12-10
Fix the navigation property binding target path issue

xuzhg:
### Issues

*This pull request fixes issue #1938.*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
+++ b/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
@@ -3055,19 +3055,53 @@ namespace Microsoft.OData.Edm
         /// <returns>The navigation source found or empty if none found.</returns>
         internal static IEdmNavigationSource FindNavigationSource(this IEdmEntityContainer container, string path)
         {
-            string[] pathSegments = path.Split('.').Last().Split('/');
+            EdmUtil.CheckArgumentNull(container, "container");
+            EdmUtil.CheckArgumentNull(path, "path");
+
+            // the path could be:
+            // NS.Default.Customers/ContainedOrders  --> for back-compatible
+            // NS.Default/Customers/ContainedOrders
+            // Customers
+            // Customers/ContainedOrders
+            string[] pathSegments = path.Split('/');
+
+            string firstElementName = pathSegments[0];
+            int nextIndex = 1;
+            if (pathSegments[0].Contains("."))
+            {
+                if (pathSegments[0] == container.FullName())
+                {
+                    if (pathSegments.Length > 1)
+                    {
+                        // NS.Default/Customers/ContainedOrders
+                        firstElementName = pathSegments[1];
+                        nextIndex = 2;
+                    }
+                    else
+                    {
+                        // if path only includes the namespace, for example "NS.Default", just return null;
+                        return null;
+                    }
+                }
+                else
+                {
+                    // NS.Default.Customers/ContainedOrders
+                    // Split the first item using "." and fetch the last segment.
+                    firstElementName = pathSegments[0].Split('.').Last();
+                }
+            }
 
             // Starting segment must be a singleton or entity set
-            IEdmNavigationSource navigationSource = container.FindEntitySet(pathSegments[0]);
+            IEdmNavigationSource navigationSource = container.FindEntitySet(firstElementName);
 
             if (navigationSource == null)
             {
-                navigationSource = container.FindSingleton(pathSegments[0]);
+                navigationSource = container.FindSingleton(firstElementName);
             }
 
             // Subsequent segments may be single-valued complex or containment nav props
             List<string> subPathSegments = new List<string>();
-            for (int i = 1; i < pathSegments.Length && navigationSource != null; i++)
+            for (int i = nextIndex; i < pathSegments.Length && navigationSource != null; i++)
             {
                 subPathSegments.Add(pathSegments[i]);
                 IEdmNavigationProperty navProp = navigationSource.EntityType().FindProperty(pathSegments[i]) as IEdmNavigationProperty;
--- a/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
+++ b/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
@@ -3059,17 +3059,17 @@ namespace Microsoft.OData.Edm
             EdmUtil.CheckArgumentNull(path, "path");
 
             // the path could be:
-            // NS.Default.Customers/ContainedOrders  --> for back-compatible
-            // NS.Default/Customers/ContainedOrders
-            // Customers
-            // Customers/ContainedOrders
+            // "NS.Default.Customers/ContainedOrders"(for backward-compatibility) or
+            // "NS.Default/Customers/ContainedOrders" (for top-level entity set in the Default entity container) or
+            // "Customers" (unqualified)
+            // "Customers/ContainedOrders" (unqualified)
             string[] pathSegments = path.Split('/');
 
             string firstElementName = pathSegments[0];
             int nextIndex = 1;
-            if (pathSegments[0].Contains("."))
+            if (firstElementName.Contains("."))
             {
-                if (pathSegments[0] == container.FullName())
+                if (string.Equals(firstElementName, container.FullName(), StringComparison.OrdinalIgnoreCase))
                 {
                     if (pathSegments.Length > 1)
                     {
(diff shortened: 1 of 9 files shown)


src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -3055,19 +3055,53 @@ internal static IEdmNavigationSource FindNavigationSourceExtended(this IEdmEntit
...
+            // the path could be:
+            // NS.Default.Customers/ContainedOrders  --> for back-compatible
+            // NS.Default/Customers/ContainedOrders
+            // Customers
+            // Customers/ContainedOrders
+            string[] pathSegments = path.Split('/');
+
+            string firstElementName = pathSegments[0];
+            int nextIndex = 1;
+            if (pathSegments[0].Contains("."))
```
gathogojr, line 3070:
```suggestion
            if (firstElementName.Contains("."))

```

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -3055,19 +3055,53 @@ internal static IEdmNavigationSource FindNavigationSourceExtended(this IEdmEntit
...
+            {
+                if (pathSegments[0] == container.FullName())
```
gathogojr, line 3072:
```suggestion
                if (firstElementName == container.FullName())

```
```suggestion
                if (firstElementName == container.FullName())
```

gathogojr, line 3072: Might we want to do a case-insensitive comparison here?

gathogojr (approved):

gathogojr:
@xuzhg Based on your change, the generated metadata will have `NS/` in place of `NS.`. Your code takes care of parsing any Url in previous format and that seems sufficient to me but I'm not sure if Mike was implying that we should give the users control over the format of metadata to be generated in [this comment](https://github.com/OData/odata.net/issues/1938#issuecomment-740792461):
> For compatibility we should probably introduce a flag that controls whether the target is written in the old (incorrect) format or the new (correct) format. For 4.0 we should (unfortunately) make the old format for the default. For 4.01 metadata responses we can probably get away with changing the default to the new (correct) syntax.

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -3055,19 +3055,53 @@ internal static IEdmNavigationSource FindNavigationSourceExtended(this IEdmEntit
         /// <returns>The navigation source found or empty if none found.</returns>
         internal static IEdmNavigationSource FindNavigationSource(this IEdmEntityContainer container, string path)
         {
-            string[] pathSegments = path.Split('.').Last().Split('/');
+            EdmUtil.CheckArgumentNull(container, "container");
+            EdmUtil.CheckArgumentNull(path, "path");
+
+            // the path could be:
+            // NS.Default.Customers/ContainedOrders  --> for back-compatible
```
mikepizzo, line 3062: or even NS.Default.Customers (for backward-compatibility) or NS.Default/Customers (for top-level entity set in the Default entity container)

src/Microsoft.OData.Edm/Schema/EdmEntitySet.cs
```diff
@@ -41,7 +41,7 @@ public EdmEntitySet(IEdmEntityContainer container, string name, IEdmEntityType e
 
             this.container = container;
             this.type = new EdmCollectionType(new EdmEntityTypeReference(elementType, false));
-            this.path = new EdmPathExpression(this.container.FullName() + "." + this.Name);
+            this.path = new EdmPathExpression(this.container.FullName() + "/" + name);
```
mikepizzo, line 44: >this.container.FullName() + "/" + name [](start = 46, length = 38)

use string.Concat(this.container.FullName(), "/", name)

src/Microsoft.OData.Edm/Schema/EdmSingleton.cs
```diff
@@ -29,7 +29,7 @@ public EdmSingleton(IEdmEntityContainer container, string name, IEdmEntityType e
 
             this.container = container;
             this.entityType = entityType;
-            this.path = new EdmPathExpression(name);
+            this.path = new EdmPathExpression(this.container.FullName() + "/" + name);
```
mikepizzo, line 32: >this.container.FullName() + "/" + name [](start = 46, length = 38)

use string.Concat()

mikepizzo, line 32: So we are changing the "Path" property on a singleton to be container-qualified.  Where is this path used?

In general, we should NOT be using container-qualified paths.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
```diff
@@ -1986,7 +1986,7 @@ public void CanWriteNavigationPropertyBindingWithTargetPathOnContainmentOnSingle
                     "<EntityContainer Name=\"Default\">" +
                        "<Singleton Name=\"Me\" Type=\"NS.Customer\" />" +
                        "<EntitySet Name=\"Customers\" EntityType=\"NS.Customer\">" +
-                         "<NavigationPropertyBinding Path=\"ContainedOrders/OrderLines\" Target=\"Me/ContainedOrderLines\" />" +
+                         "<NavigationPropertyBinding Path=\"ContainedOrders/OrderLines\" Target=\"NS.Default/Me/ContainedOrderLines\" />" +
```
mikepizzo, line 1989: We should not be container-qualifying the target unless necessary (which it shouldn't be)

mikepizzo:
Looking more closely, there really isn't ever a reason we should have to write the container-qualified form.  The container-qualified form is a leftover from when we had multiple entity containers.

When writing the target we should ideally always write the unqualified entityset/singleton name. If we're making a breaking change, I would move to writing without the container-qualifier for both entityset and singleton. If we're worried about breaking someone, we could add a flag to add the entitycontainer using the old format (no reason to add it in the new format).

I like the reading logic, which supports both the old (invalid) container-qualified format, the appropriate container-qualified format, and the unqualified format.

---
In reply to: [744196418](https://github.com/OData/odata.net/pull/1950#issuecomment-744196418) [](ancestors = 744196418)

mikepizzo (changes requested):
:clock1:

mikepizzo:
Looking more closely, there really isn't ever a reason we should have to write the container-qualified form.  The container-qualified form is a leftover from when we had multiple entity containers.

When writing the target we should ideally always write the unqualified entityset/singleton name. If we're making a breaking change, I would move to writing without the container-qualifier for both entityset and singleton. If we're worried about breaking someone, we could add a flag to add the entitycontainer using the old format (no reason to add it in the new format).

I like the reading logic, which supports both the old (invalid) container-qualified format, the appropriate container-qualified format, and the unqualified format.

xuzhg:
I agreed to write the unqualifed entityset/singleton name.

---
In reply to: [745006913](https://github.com/OData/odata.net/pull/1950#issuecomment-745006913) [](ancestors = 745006913)

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
xuzhg: Revert to path without qualified name

---
In reply to: [542983048](https://github.com/OData/odata.net/pull/1950#discussion_r542983048) [](ancestors = 542983048)

src/Microsoft.OData.Edm/Schema/EdmSingleton.cs
xuzhg: revert

---
In reply to: [542982488](https://github.com/OData/odata.net/pull/1950#discussion_r542982488) [](ancestors = 542982488)

xuzhg: Not applicable

---
In reply to: [542982068](https://github.com/OData/odata.net/pull/1950#discussion_r542982068) [](ancestors = 542982068)

src/Microsoft.OData.Edm/Schema/EdmEntitySet.cs
xuzhg: change to use the name directly, so not applicable

---
In reply to: [542981676](https://github.com/OData/odata.net/pull/1950#discussion_r542981676) [](ancestors = 542981676)

xuzhg:
Yes, I agree to write the unqualified entityset/single name since we only have one entity container now.
I don't think we should introduce the flag because the old pattern is wrong pattern, and it only happens in containment binding.

---
In reply to: [744196418](https://github.com/OData/odata.net/pull/1950#issuecomment-744196418) [](ancestors = 744196418)

mikepizzo (approved):
:shipit:

PR #1954, opened 2020-12-11
Performance fixes

Sreejithpin:
Performance fixes
This Performance PR addresses 3 issues
1. Instead of + use string.Concat for concatenating strings in OData.Core
2. There is a map which generates all characters with escape strings, changed it to generate only for needed ones
3. Update a Linq expression to remove the Linq (To List)

### Issues

*This pull request fixes performance issues

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/BindingPathHelper.cs
+++ b/src/Microsoft.OData.Core/BindingPathHelper.cs
@@ -23,16 +23,15 @@ namespace Microsoft.OData
         /// <returns>True if the path of navigation property in current scope is matching the <paramref name="bindingPath"/>.</returns>
         public static bool MatchBindingPath(IEdmPathExpression bindingPath, IList<ODataPathSegment> parsedSegments)
         {
-            IEnumerable<string> paths = bindingPath.PathSegments;
-            int count = paths.Count();
+            List<string> paths = bindingPath.PathSegments.ToList();
 
             // If binding path only includes navigation property name, it matches.
-            if (count == 1)
+            if (paths.Count == 1)
             {
                 return true;
             }
 
-            int pathIndex = count - 2; // Skip the last segment which is navigation property name.
+            int pathIndex = paths.Count - 2; // Skip the last segment which is navigation property name.
             int segmentIndex = parsedSegments.Count - 1;
             // Match from tail to head.
             for (; segmentIndex >= 0; segmentIndex--)
@@ -41,11 +40,11 @@ namespace Microsoft.OData
 
                 // Cache the cast result to avoid CA1800:DoNotCastUnnecessarily.
                 bool segmentIsNavigationPropertySegment = segment is NavigationPropertySegment;
-                
+
                 // Containment navigation property or complex property in binding path.
                 if (segment is PropertySegment || (segmentIsNavigationPropertySegment && segment.TargetEdmNavigationSource is IEdmContainedEntitySet))
                 {
-                    if (pathIndex < 0 || string.CompareOrdinal(paths.ElementAt(pathIndex), segment.Identifier) != 0)
+                    if (pathIndex < 0 || string.CompareOrdinal(paths[pathIndex], segment.Identifier) != 0)
                     {
                         return false;
                     }
@@ -55,9 +54,9 @@ namespace Microsoft.OData
                 else if (segment is TypeSegment)
                 {
                     // May need match type if the binding path contains type cast.
-                    if (pathIndex >= 0 && paths.ElementAt(pathIndex).IndexOf('.') >= 0)
+                    if (pathIndex >= 0 && paths[pathIndex].IndexOf('.') >= 0)
                     {
-                        if (string.CompareOrdinal(paths.ElementAt(pathIndex), segment.EdmType.AsElementType().FullTypeName()) != 0)
+                        if (string.CompareOrdinal(paths[pathIndex], segment.EdmType.AsElementType().FullTypeName()) != 0)
                         {
                             return false;
                         }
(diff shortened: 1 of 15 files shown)


src/Microsoft.OData.Core/BindingPathHelper.cs
```diff
@@ -23,15 +23,16 @@ internal static class BindingPathHelper
         /// <returns>True if the path of navigation property in current scope is matching the <paramref name="bindingPath"/>.</returns>
         public static bool MatchBindingPath(IEdmPathExpression bindingPath, IList<ODataPathSegment> parsedSegments)
         {
-            List<string> paths = bindingPath.PathSegments.ToList();
```
mikepizzo, line 26: I would question, in the code path below, whether IEnumerable is more efficient than calling ToList() since we are looping through the results multiple times.  In particular, line 27 will loop through the enumerable to count the number of segments and then lines 48/60 will call ElementAt() for each property/type segment in the path

src/Microsoft.OData.Core/TypeUtils.cs
```diff
@@ -96,7 +96,7 @@ internal static bool AreTypesEquivalent(Type typeA, Type typeB)
         /// <param name="isCollection">Returns whether or not the returned type is a collection.</param>
         internal static void ParseQualifiedTypeName(string qualifiedTypeName, out string namespaceName, out string typeName, out bool isCollection)
         {
-            isCollection = qualifiedTypeName.StartsWith(ODataConstants.CollectionPrefix + "(", StringComparison.Ordinal);
+            isCollection = qualifiedTypeName.StartsWith(string.Concat(ODataConstants.CollectionPrefix, "("), StringComparison.Ordinal);
```
mikepizzo, line 99: >string.Concat(ODataConstants.CollectionPrefix, "(") [](start = 56, length = 51)

If this code is called frequently, since all of the strings being concatenated are static, a better optimization would be to define an ODataConstantsCollectionPrefixWithParens = string.Concat(ODataContstants.CollectionPrefix, "(")); and use that here (so the concatenation only occurs once).

Even better, in this case, it looks like in every case that ODataConstants.CollectionPrefix is used, the parens is added (except when the length is used, in which case the count is incremented by one to account for the open parens).  So you could just define the ODataConstants.CollectionPrefix to include the parens and then fix the places it is referenced to account or the fact that the open parens is already added.

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -51,7 +52,8 @@ internal static class JsonValueUtils
         /// <summary>
         /// Map of special characters to strings.
         /// </summary>
-        private static readonly string[] SpecialCharToEscapedStringMap = JsonValueUtils.CreateSpecialCharToEscapedStringMap();
+        private static Dictionary<char, string> SpecialCharToEscapedStringMap = new Dictionary<char, string>() { { '\r', "\\r" }, { '\t', "\\t" }, { '\n', "\\n" },
```
mikepizzo, line 55: >Dictionary [](start = 23, length = 10)

This is not threadsafe, as you are adding to the dictionary in line 96. Probably have to make this a concurrentdictionary. I'd want to see perf comparisons to validate that looking up/lazily adding to a concurrent dictionary would be more performant than constructing the map up-front and doing a simple array index look-up.  For frequently lookups (where performance becomes more critical) I would think that the readonly string[] would probably be faster.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
```diff
@@ -327,7 +327,7 @@ private ODataPayloadKind ParseContextUriFragment(string fragment, Func<IEdmType,
                 {
                     detectedPayloadKind = ODataPayloadKind.ServiceDocument;
                 }
-                else if (fragment.Equals(ODataConstants.CollectionPrefix + "(" + ODataConstants.EntityReferenceSegmentName + ")", StringComparison.Ordinal))
+                else if (fragment.Equals(string.Concat(ODataConstants.CollectionPrefix, "(", ODataConstants.EntityReferenceSegmentName, ")"), StringComparison.Ordinal))
```
mikepizzo, line 330: >else if (fragment.Equals(string.Concat(ODataConstants.CollectionPrefix, "(", ODataConstants.EntityReferenceSegmentName, ")"), StringComparison.Ordinal)) [](start = 16, length = 152)

Since these are all static values, it probably makes sense to define an ODataConstant.EntityReferenceCollection = string.Concat(ODataConstants.CollectionPrefix, "(", ODataConstants.EntityReferenceSegmentName, ")")

src/Microsoft.OData.Core/UrlValidation/Rules/DeprecationRules.cs
```diff
@@ -271,7 +271,7 @@ private static IEnumerable<IEdmVocabularyAnnotation> GetRevisionAnnotations(IEdm
         private static bool isRevisionsAnnotation(IEdmVocabularyAnnotation annotation)
         {
             return string.Equals(annotation.Term.FullName(), CoreVocabularyConstants.Revisions, StringComparison.OrdinalIgnoreCase) ||
-                            string.Equals(annotation.Term.FullName(), DefaultCoreAlias + "." + RevisionTerm, StringComparison.OrdinalIgnoreCase);
+                            string.Equals(annotation.Term.FullName(), string.Concat(DefaultCoreAlias, ".", RevisionTerm), StringComparison.OrdinalIgnoreCase);
```
mikepizzo, line 274: > string.Concat(DefaultCoreAlias, ".", RevisionTerm) [](start = 69, length = 51)

again, since all values are constants, consider defining a private const string QualifiedRevisionsTerm =  string.Concat(DefaultCoreAlias, ".", RevisionTerm)

mikepizzo (changes requested):
:clock1:

mikepizzo (approved):
:shipit:

mikepizzo:
Looks good -- thanks!

Issue #1955, opened 2020-12-13
CsdlReader: Unable to parse SAP Common Metadata 


Tiberriver256:
Hello,

I just ran into this issue in a downstream library. I can actually get the reader to parse the xml file by setting the ignore unknown attributes and elements to false but it would be nice to have  better support for this out of the box.

The [SAP OData common vocabulary](https://github.com/SAP/odata-vocabularies/blob/master/vocabularies/Common.xml) is maintained by the OData team at SAP and is referenced in every single one of their OData v4 APIs.

Being able to parse this file (and maybe the rest in that repo) would allow Microsoft downstream libraries to more easily integrate with SAP OData APIs.

### Assemblies affected

Microsoft.Odata.Edm

### Reproduce steps

Use the TryParse method of CsdlReader to parse the xml file.

### Expected result

The Csdl should be parsed successfully

### Actual result

An error is thrown about the attribute 'BaseTerm' which is a OData v4.01 standard attribute

mikepizzo:
See http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_SpecializedTerm

xuzhg:
https://github.com/OData/odata.net/issues/1816

Issue #1816, opened 2020-06-23
Support Term inheritance


xuzhg:
OData spec says"

```
14.1.1 Specialized Term
A term MAY specialize another term in scope by specifying it as its base type.

When applying a term with a base term, the base term MUST also be applied with the same qualifier, and so on until a term without a base term is reached.

Attribute BaseTerm

The value of BaseTerm is the qualified name of the base term.
```

ODL doesn't support it yet.

mikepizzo:
This test should pass if BaseTerm was supported:

~~~xml
        public void BaseTerm()
        {
            String schema = @"
<edmx:Edmx Version=""4.0"" xmlns:edmx=""http://docs.oasis-open.org/odata/ns/edmx"" xmlns:ags=""http://aggregator.microsoft.com/internal"">
  <edmx:DataServices>
    <Schema xmlns=""http://docs.oasis-open.org/odata/ns/edm"" Namespace=""Test"">
      <Term Name=""TagTerm"" Type=""Edm.Boolean"" DefaultValue=""true"" BaseTerm=""Core.Tag""/>
    </Schema>
  </edmx:DataServices>
</edmx:Edmx>";

            IEdmModel model;
            IEnumerable<EdmError> errors;
            Assert.True(CsdlReader.TryParse(XmlReader.Create(new StringReader(csdl)), out model, out errors), errors.FirstOrDefault()?.ErrorMessage);
            Assert.True(model.Validate(out errors), errors.FirstOrDefault()?.ErrorMessage);
        }
~~~

PR #1962, opened 2020-12-17
Enable Base term for IEdmTerm

xuzhg:
### Issues

*This pull request fixes issue #1955. & #1816*

### Description

*14.1.1 Specialized Term

A term MAY specialize another term in scope by specifying it as its base type.

When applying a term with a base term, the base term MUST also be applied with the same qualifier, and so on until a term without a base term is reached.

Attribute BaseTerm

The value of BaseTerm is the qualified name of the base term.

This PR is to enable the base term.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsTerm.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsTerm.cs
@@ -24,6 +24,9 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
 
         private readonly Cache<CsdlSemanticsTerm, IEdmTypeReference> typeCache = new Cache<CsdlSemanticsTerm, IEdmTypeReference>();
         private static readonly Func<CsdlSemanticsTerm, IEdmTypeReference> ComputeTypeFunc = (me) => me.ComputeType();
+        private readonly Cache<CsdlSemanticsTerm, IEdmTerm> baseTermCache = new Cache<CsdlSemanticsTerm, IEdmTerm>();
+        private static readonly Func<CsdlSemanticsTerm, IEdmTerm> ComputeBaseTermFunc = (me) => me.ComputeBaseTerm();
+        private static readonly Func<CsdlSemanticsTerm, IEdmTerm> OnCycleBaseTermFunc = (me) => new CyclicTerm(me.GetCyclicBaseTermName(me.term.BaseTermName), me.Location);
 
         public CsdlSemanticsTerm(CsdlSemanticsSchema context, CsdlTerm valueTerm)
             : base(valueTerm)
@@ -61,6 +64,11 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
             get { return this.typeCache.GetValue(this, ComputeTypeFunc, null); }
         }
 
+        public IEdmTerm BaseTerm
+        {
+            get { return this.baseTermCache.GetValue(this, ComputeBaseTermFunc, OnCycleBaseTermFunc); }
+        }
+
         public string AppliesTo
         {
             get { return this.term.AppliesTo; }
@@ -89,6 +97,32 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
         private IEdmTypeReference ComputeType()
         {
             return CsdlSemanticsModel.WrapTypeReference(this.Context, this.term.Type);
+        }
+
+        private IEdmTerm ComputeBaseTerm()
+        {
+            if (this.term.BaseTermName != null)
+            {
+                IEdmTerm baseTerm = this.Context.FindTerm(this.term.BaseTermName);
+                if (baseTerm != null)
+                {
+                    // Evaluate the inductive step to detect cycles.
+                    // Overriding BaseTerm getter from concrete type implementing IEdmTerm will be invoked to
+                    // detect cycles. The object assignment is required by compiler only.
+                    IEdmTerm baseTerm2 = baseTerm.BaseTerm;
+                }
+
+                return baseTerm ?? new UnresolvedVocabularyTerm(this.term.BaseTermName);
+            }
+
+            return null;
+        }
+
+        // Resolves the real name of the base term, in case it was using an alias before.
+        protected string GetCyclicBaseTermName(string baseTermName)
+        {
+            IEdmTerm schemaBaseTerm = this.Context.FindTerm(baseTermName);
+            return (schemaBaseTerm != null) ? schemaBaseTerm.FullName() : baseTermName;
         }
     }
 }
--- a/src/Microsoft.OData.Edm/Schema/CyclicTerm.cs
+++ b/src/Microsoft.OData.Edm/Schema/CyclicTerm.cs
@@ -1,0 +1,22 @@
+//---------------------------------------------------------------------
+// <copyright file="CyclicTerm.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using Microsoft.OData.Edm.Validation;
+using Microsoft.OData.Edm.Vocabularies;
+
+namespace Microsoft.OData.Edm
+{
+    /// <summary>
+    /// Represents an EDM term that cannot be determined due to a cyclic reference.
+    /// </summary>
+    internal class CyclicTerm : BadTerm
+    {
+        public CyclicTerm(string qualifiedName, EdmLocation location)
+            : base(qualifiedName, new EdmError[] { new EdmError(location, EdmErrorCode.BadCyclicTerm, Edm.Strings.Bad_CyclicTerm(qualifiedName)) })
+        {
+        }
+    }
+}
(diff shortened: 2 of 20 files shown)


test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/Parsing/SchemaJsonParserTests.cs
```diff
@@ -286,13 +286,15 @@ public void ParseCsdlTypeDefinitionWithMembersWorksAsExpected()
         #endregion
 
         #region Term
+
         [Fact]
-        public void ParseCsdlTermWithMembersWorksAsExpected()
+        public void ParseCsdlTermWithBaseTermAndMembersWorksAsExpected()
```
mikepizzo, line 291: >ParseCsdlTermWithBaseTermAndMembersWorksAsExpected [](start = 20, length = 50)

Can you also add some XML test variants of reading/writing terms with base values?

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/Serialization/EdmModelCsdlSerializationVisitorTests.cs
```diff
@@ -815,6 +815,34 @@ public void VerifyTermWrittenCorrectly()
 }");
         }
 
+        [Fact]
+        public void VerifyTermWithBaseTermWrittenCorrectly()
```
mikepizzo, line 819: >VerifyTermWithBaseTermWrittenCorrectly [](start = 20, length = 38)

Please add XML Variation as well.

xuzhg: Yes. Do you mean "Line 827 - 828"

---
In reply to: [545398455](https://github.com/OData/odata.net/pull/1962#discussion_r545398455) [](ancestors = 545398455)

src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmTerm.cs
```diff
@@ -25,5 +25,10 @@ public interface IEdmTerm : IEdmSchemaElement
         /// Gets the DefaultValue of this term.
         /// </summary>
         string DefaultValue { get; }
+
+        /// <summary>
+        /// Gets the base term of this term.
+        /// </summary>
+        IEdmTerm BaseTerm { get; }
```
mikepizzo, line 32: This is my only real concern with the PR.  We need to convince ourselves that we are the only ones implementing this interface, as anyone else providign a custom implementation will break when they update to the new library and have to modify their code to implement this property.

The fallback would be to define an IEdmTermWithBase : IEdmTerm, but that would require logic each place we consume IEdmTerm that we care about base to handle base as well, if the instance of IEdmTerm can be cast to IEdmTermWithBase.  We could add IEdmTermWithBase in 7.9 and clean it up in 8.0, or we can convince ourselves that there aren't any custom implementations of IEdmTerm that we would break.

mikepizzo:
What are the ramifications of introducing a base Term? Do consumers of the API need to do anything different?

i.e., for structured types, a base type implies that all of the properties of the base type are available on the derived type.  a client that looks only at the declared properties of the derived type will miss the properties inherited from the base type.

Are there similar ramifications for consumers of a derived term?

mikepizzo:
>     public interface IEdmTerm : IEdmSchemaElement

Need to update public api

---
Refers to: src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmTerm.cs:12 in 22945d8. [](commit_id = 22945d8514b5d8ab3b0ae247b7f3a4a4655a7ef6, deletion_comment = False)

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/Serialization/EdmModelCsdlSerializationVisitorTests.cs
mikepizzo: Yes; missed that -- thanks.

---
In reply to: [545399808](https://github.com/OData/odata.net/pull/1962#discussion_r545399808) [](ancestors = 545399808,545398455)

xuzhg:
Quick answer is that" It's the same usage of "Base type" and "Base Term".

When we create the annotation, we can create the PropertyValue for the "Property" defined in the base term.
However, OData.Edm lib doesn't have the process to verify that "PropertyName" of the Record is defined or not. 
Let me add a test case to cover it.

---
In reply to: [747701313](https://github.com/OData/odata.net/pull/1962#issuecomment-747701313) [](ancestors = 747701313)

src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmTerm.cs
xuzhg: That's also my concern. However, i have the following two reasons:

1) To add new into a public interface is less "harmful" than to remove or to change element in the public interface.
2) I can't find any reason why there's a customer who need to implement this interface.



---
In reply to: [545401071](https://github.com/OData/odata.net/pull/1962#discussion_r545401071) [](ancestors = 545401071)

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/Parsing/SchemaJsonParserTests.cs
xuzhg: There's no related test cases to test part of an XML Element.

I added a whole write test cases to cover it.

---
In reply to: [545398235](https://github.com/OData/odata.net/pull/1962#discussion_r545398235) [](ancestors = 545398235)

src/Microsoft.OData.Edm/Validation/EdmErrorCode.cs
```diff
@@ -1383,6 +1383,11 @@ public enum EdmErrorCode
         /// </summary>
         AnnotationApplyToNotAllowedAnnotatable = 400,
 
+        /// <summary>
+        /// This ter, type is part of a cycle.
```
gathogojr, line 1387: Fix docstring

src/Microsoft.OData.Edm/Schema/CyclicTerm.cs
```diff
@@ -0,0 +1,22 @@
...
+    /// Represents an EDM term that cannot be determined due to a cyclic reference.
+    /// </summary>
+    internal class CyclicTerm : BadTerm
+    {
+        public CyclicTerm(string qualifiedName, EdmLocation location)
+            : base(qualifiedName, new EdmError[] { new EdmError(location, EdmErrorCode.BadCyclicTerm, Edm.Strings.Bad_CyclicTerm(qualifiedName)) })
+        {
+        }
+    }
+}
```
gathogojr, line 22: nit: Add newline

test/PublicApiTests/BaseLine/Microsoft.OData.PublicApi.net45.bsl
```diff
@@ -7742,21 +7756,6 @@ public abstract class Microsoft.OData.Client.OperationResponse {
     int StatusCode  { public get; }
 }
 
-[
```
gathogojr, line 7745: This is from a merged PR. Have you rebased your branch?

test/PublicApiTests/BaseLine/Microsoft.OData.PublicApi.netstandard2.0.bsl
```diff
@@ -7774,21 +7788,6 @@ public abstract class Microsoft.OData.Client.OperationResponse {
     int StatusCode  { public get; }
 }
 
-[
```
gathogojr, line 7777: This is from a merged PR. Have you rebased your branch?

test/PublicApiTests/BaseLine/Microsoft.OData.PublicApi.netstandard2.0.bsl
```diff
@@ -3080,6 +3080,7 @@ public enum Microsoft.OData.Edm.Validation.EdmErrorCode : int {
     BadCyclicComplex = 227
     BadCyclicEntity = 229
     BadCyclicEntityContainer = 228
+    BadCyclicTerm = 401
```
gathogojr, line 3083: Assigning value of 401 seems to have affected other enum values. Is that okay?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsTerm.cs
```diff
@@ -90,5 +98,31 @@ private IEdmTypeReference ComputeType()
...
+            }
+
+            return null;
+        }
+
+        // Resolves the real name of the base term, in case it was using an alias before.
+        protected string GetCyclicBaseTermName(string baseTermName)
+        {
+            IEdmTerm schemaBaseTerm = this.Context.FindTerm(baseTermName);
+            return (schemaBaseTerm != null) ? schemaBaseTerm.FullName() : baseTermName;
```
corranrogue9, line 125: super nit: I prefer the null propagation syntax here: `return schemaBaseTerm?.FullName() ?? baseTermName`

src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmTerm.cs
corranrogue9: I think I'm understand the high-level concern but missing the specifics for our case. Could we not have this property be of type `BaseTerm : IEdmTerm` that has an `internal` constructor so we have full control over the behavior?

Issue #1889, opened 2020-09-21
Add WriteMetadataDocumentAsync in the ODataMessageWriter


xuzhg:
*Short summary (3-5 sentences) describing the issue.*

### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData .Net lib 7.x*

### Reproduce steps

* ODL has the async method call for most of the writer. for example:

1) WriteServiceDocumentAsync  vs WriteServiceDocument
2) WriteStartAsync(...) vs WriteStart(...)

However, `WriteMetadataDocument` doesn't have the corresponding Async version. Please add it.

PR #1966, opened 2020-12-24
Add WriteMetadataAsync API to enable write CSDL async

xuzhg:
### Issues

*This pull request fixes issue #https://github.com/OData/odata.net/issues/1889.*

### Description

* Add `WriteMetadataDocumentAsync` public API.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataMetadataFormat.cs
+++ b/src/Microsoft.OData.Core/ODataMetadataFormat.cs
@@ -67,10 +67,8 @@ namespace Microsoft.OData
             {
                 return new ODataMetadataJsonInputContext(messageInfo, messageReaderSettings);
             }
-            else
-            {
-                return new ODataMetadataInputContext(messageInfo, messageReaderSettings);
-            }
+
+            return new ODataMetadataInputContext(messageInfo, messageReaderSettings);
 #else
             if (isJson)
             {
@@ -101,10 +99,8 @@ namespace Microsoft.OData
             {
                 return new ODataMetadataJsonOutputContext(messageInfo, messageWriterSettings);
             }
-            else
-            {
-                return new ODataMetadataOutputContext(messageInfo, messageWriterSettings);
-            }
+
+            return new ODataMetadataOutputContext(messageInfo, messageWriterSettings);
 #else
             if (isJson)
             {
@@ -168,10 +164,8 @@ namespace Microsoft.OData
             {
                 return Task.FromResult<ODataOutputContext>(new ODataMetadataJsonOutputContext(messageInfo, messageWriterSettings));
             }
-            else
-            {
-                return Task.FromResult<ODataOutputContext>(new ODataMetadataOutputContext(messageInfo, messageWriterSettings));
-            }
+
+            return Task.FromResult<ODataOutputContext>(new ODataMetadataOutputContext(messageInfo, messageWriterSettings));
 #else
             if (isJson)
             {
--- a/src/Microsoft.OData.Core/ODataMetadataFormat.cs
+++ b/src/Microsoft.OData.Core/ODataMetadataFormat.cs
@@ -161,7 +161,25 @@ namespace Microsoft.OData
             ExceptionUtils.CheckArgumentNotNull(messageInfo, "messageInfo");
             ExceptionUtils.CheckArgumentNotNull(messageWriterSettings, "messageWriterSettings");
 
-            throw new ODataException(Strings.General_InternalError(InternalErrorCodes.ODataMetadataFormat_CreateOutputContextAsync));
+            bool isJson = IsJsonMetadata(messageInfo.MediaType);
+
+#if NETSTANDARD2_0
+            if (isJson)
+            {
+                return Task.FromResult<ODataOutputContext>(new ODataMetadataJsonOutputContext(messageInfo, messageWriterSettings));
+            }
+            else
+            {
+                return Task.FromResult<ODataOutputContext>(new ODataMetadataOutputContext(messageInfo, messageWriterSettings));
+            }
+#else
+            if (isJson)
+            {
+                throw new ODataException(Strings.ODataMetadataOutputContext_NotSupportJsonMetadata);
+            }
+
+            return Task.FromResult<ODataOutputContext>(new ODataMetadataOutputContext(messageInfo, messageWriterSettings));
+#endif
         }
 
         private static bool IsJsonMetadata(ODataMediaType contentType)
--- a/src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs
+++ b/src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs
@@ -12,6 +12,7 @@ using System.IO;
 using System.Text;
 using System.Text.Encodings.Web;
 using System.Text.Json;
+using System.Threading.Tasks;
 using Microsoft.OData.Edm.Csdl;
 using Microsoft.OData.Edm.Validation;
 
@@ -28,6 +29,9 @@ namespace Microsoft.OData
         /// <summary>The jsonWriter to write to.</summary>
         private Utf8JsonWriter jsonWriter;
 
+        /// <summary>The asynchronous output stream if we're writing asynchronously.</summary>
+        private AsyncBufferedStream asynchronousOutputStream;
+
         /// <summary>
         /// Constructor.
         /// </summary>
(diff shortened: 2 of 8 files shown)


src/Microsoft.OData.Core/ODataMetadataFormat.cs
```diff
@@ -161,7 +161,25 @@ public override string ToString()
...
+
+#if NETSTANDARD2_0
+            if (isJson)
+            {
+                return Task.FromResult<ODataOutputContext>(new ODataMetadataJsonOutputContext(messageInfo, messageWriterSettings));
+            }
+            else
+            {
+                return Task.FromResult<ODataOutputContext>(new ODataMetadataOutputContext(messageInfo, messageWriterSettings));
+            }
```
habbes, line 174: Is the `else` necessary?

src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs
```diff
@@ -156,19 +180,55 @@ protected override void Dispose(bool disposing)
...
                 {
-                    this.jsonWriter.Flush();
-                    this.jsonWriter.Dispose();
+                    if (this.asynchronousOutputStream != null)
+                    {
+                        this.asynchronousOutputStream.FlushSync();
+                        this.asynchronousOutputStream.Dispose();
+
+                        this.jsonWriter.FlushAsync();
+                        this.jsonWriter.DisposeAsync();
```
habbes, line 189: We're not waiting for the async operations to complete? Or is it okay to just fire and forget since the asynchronous stream has already been flushed and disposed?

src/Microsoft.OData.Core/ODataMetadataOutputContext.cs
```diff
@@ -127,8 +179,18 @@ protected override void Dispose(bool disposing)
             {
                 if (this.xmlWriter != null)
                 {
-                    // XmlWriter.Flush will call the underlying Stream.Flush.
-                    this.xmlWriter.Flush();
+                    // In the async case the underlying stream is the async buffered stream, so we have to flush that explicitly.
+                    if (this.asynchronousOutputStream != null)
+                    {
+                        this.xmlWriter.FlushAsync();
```
habbes, line 185: We don't wait for the async operation to complete?

habbes (approved):

src/Microsoft.OData.Core/ODataMetadataFormat.cs
xuzhg: If clause to create "ODataMetadataJsonOutputContext" => json
else clause to create "ODataMetadataOutputContext"   => xml

So, it's necessary.

---
In reply to: [556528819](https://github.com/OData/odata.net/pull/1966#discussion_r556528819) [](ancestors = 556528819)

src/Microsoft.OData.Core/ODataMetadataOutputContext.cs
xuzhg: I checked with others codes in this project. For example here: [https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Json/ODataJsonTextWriter.cs#L298](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Json/ODataJsonTextWriter.cs#L298)

I'd invite @Michael Pizzo to share us the reason?


---
In reply to: [556556718](https://github.com/OData/odata.net/pull/1966#discussion_r556556718) [](ancestors = 556556718)

src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs
xuzhg: Below is the code for DisposeAysnc():

```C#
        public async ValueTask DisposeAsync()
        {
            if (_stream == null)
            {
                // The conditions are ordered with stream first as that would be the most common mode
                if (_output == null)
                {
                    return;
                }
            }

            await FlushAsync().ConfigureAwait(false);
            ResetHelper();

            _stream = null;
            _arrayBufferWriter = null;
            _output = null;
        }
```

In this case, i think we don't need to call await, and we can NOT call await in a non-async method. @mikepizzo

---
In reply to: [556555576](https://github.com/OData/odata.net/pull/1966#discussion_r556555576) [](ancestors = 556555576)

xuzhg:
@mikepizzo Would you please share your thoughts about the comments in the PR?

xuzhg:
@mikepizzo @gathogojr @KenitoInc

src/Microsoft.OData.Core/ODataMetadataFormat.cs
mikepizzo: I think what Clement is saying is that, if it is json, the method will return on line 169 so we could remove the else on line 171 (and associated braces), and just let control fall through to returning xml on line 173.

---
In reply to: [558495394](https://github.com/OData/odata.net/pull/1966#discussion_r558495394) [](ancestors = 558495394,556528819)

src/Microsoft.OData.Core/ODataMetadataOutputContext.cs
```diff
@@ -57,6 +72,55 @@ internal sealed class ODataMetadataOutputContext : ODataOutputContext
...
+        /// <summary>
+        /// Writes an <see cref="ODataError"/> into the message payload.
+        /// </summary>
+        /// <param name="error">The error to write.</param>
+        /// <param name="includeDebugInformation">
+        /// A flag indicating whether debug information (e.g., the inner error from the <paramref name="error"/>) should
+        /// be included in the payload. This should only be used in debug scenarios.
+        /// </param>
+        /// <returns>Task which represents the pending write operation.</returns>
+        internal override Task WriteInStreamErrorAsync(ODataError error, bool includeDebugInformation)
```
mikepizzo, line 95: >WriteInStreamErrorAsync [](start = 31, length = 23)

Why do we have this async method on the Xml Metadata Output context but not the Json metadata output context?

src/Microsoft.OData.Core/ODataMetadataOutputContext.cs
```diff
@@ -127,8 +179,18 @@ protected override void Dispose(bool disposing)
             {
                 if (this.xmlWriter != null)
                 {
-                    // XmlWriter.Flush will call the underlying Stream.Flush.
-                    this.xmlWriter.Flush();
+                    // In the async case the underlying stream is the async buffered stream, so we have to flush that explicitly.
+                    if (this.asynchronousOutputStream != null)
+                    {
+                        this.xmlWriter.FlushAsync();
+                        this.asynchronousOutputStream.FlushSync();
```
mikepizzo, line 186: >FlushSync [](start = 54, length = 9)

Why are we calling FlushSync, rather than FlushAsync, on the asynchronous stream?

src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs
```diff
@@ -156,19 +180,55 @@ protected override void Dispose(bool disposing)
             {
                 if (this.jsonWriter != null)
                 {
-                    this.jsonWriter.Flush();
-                    this.jsonWriter.Dispose();
+                    if (this.asynchronousOutputStream != null)
+                    {
+                        this.asynchronousOutputStream.FlushSync();
```
mikepizzo, line 185: >FlushSync [](start = 54, length = 9)

Why are we calling FlushSync, rather than FlushAsync, on the asynchronous stream?

src/Microsoft.OData.Core/ODataMetadataOutputContext.cs
xuzhg: It's weird that we have such method in Metadata outputcontext, And there's no way to call this throw metadata output context.
I think we should remove this method but avoid changing existing class, i keep it unchanged.

So, For JSON metadata, i didn't add this method.

---
In reply to: [561201263](https://github.com/OData/odata.net/pull/1966#discussion_r561201263) [](ancestors = 561201263)

src/Microsoft.OData.Core/ODataMetadataFormat.cs
xuzhg: Oh, I got it.

---
In reply to: [561197162](https://github.com/OData/odata.net/pull/1966#discussion_r561197162) [](ancestors = 561197162,558495394,556528819)

xuzhg:
@mikepizzo Is it ok to merge?

src/Microsoft.OData.Core/ODataMetadataOutputContext.cs
mikepizzo: The difference is, in ODataJsonTextWriter we are returning the task from calling FlushAsync() on the internal textwriter, so the caller can wait on that result. In this case, though, we are continuing to call additional methods that may then run concurrently, potentially leading to threading issues.

---
In reply to: [560398552](https://github.com/OData/odata.net/pull/1966#discussion_r560398552) [](ancestors = 560398552,556556718)

mikepizzo (approved):
:shipit:

mikepizzo:
Looks good to me.

---
In reply to: [777888145](https://github.com/OData/odata.net/pull/1966#issuecomment-777888145) [](ancestors = 777888145)

PR #1985, opened 2021-01-21
Fix reading untyped collections containing values with @type specified.

mikepizzo:
### Issues

### Description

*When ODL reads an untyped value (for example, in a collection of untyped) that has a type specified, it fails because it doesn't understand that untyped can be cast to any type.  This PR adds support for specifying the type for an untyped value.*

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
@@ -1783,6 +1783,7 @@ namespace Microsoft.OData.JsonLight
                 }
             }
 
+            expectedTypeReference = expectedTypeReference != null && expectedTypeReference.IsUntyped() ? null : expectedTypeReference;
             ODataTypeAnnotation typeAnnotation;
             EdmTypeKind targetTypeKind;
             IEdmTypeReference targetTypeReference = this.ReaderValidator.ResolvePayloadTypeNameAndComputeTargetType(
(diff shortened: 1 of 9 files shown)


xuzhg:
>     public sealed class EdmUntypedStructuredType : EdmStructuredType, IEdmStructuredType, IEdmSchemaElement, IEdmSchemaType, IEdmFullNamedElement

I can't understand why do we have:  EdmUntypedType and EdmUntypedStructuredType? 

It could be a confusing for customer.

---
Refers to: src/Microsoft.OData.Edm/Schema/EdmUntypedStructuredType.cs:14 in 801c068. [](commit_id = 801c068cf1d2c196e9d58a5f42e2c6fd3393532e, deletion_comment = False)

xuzhg (approved):
:shipit:

mikepizzo:
>     public sealed class EdmUntypedStructuredType : EdmStructuredType, IEdmStructuredType, IEdmSchemaElement, IEdmSchemaType, IEdmFullNamedElement

Short answer is that we need to distinguish in our code between an untyped value that is structured and an untyped value that is a primitive.  So, when we read a dynamic property that has structure we return an EdmUntypedStructuredType, and the rest of the code knows that its dealing with a structured type, and if we read a dynamic property that doesn't have structure we return an EdmUntypedType and the rest of the code knows that it's dealing with a primitive type.

We probably could have made the EdmUntypedStructuredType internal -- I don't know that there is ever a case where the user needs to create an instance of an untyped structured type, or distinguish between an untyped structured type and untyped primitive.

This would be good cleanup for ODL 8.0.

---
In reply to: [777887252](https://github.com/OData/odata.net/pull/1985#issuecomment-777887252) [](ancestors = 777887252)

---
Refers to: src/Microsoft.OData.Edm/Schema/EdmUntypedStructuredType.cs:14 in 801c068. [](commit_id = 801c068cf1d2c196e9d58a5f42e2c6fd3393532e, deletion_comment = False)

PR #1987, opened 2021-01-25
Import vocabulary annotations declared in external referenced models

### LINKED ISSUE #1936: Referenced external permissions annotations don't reflect in main EDM model
Using this notation in the main XML document to reference an external permissions annotation XML document does not seem to surface the referenced permissions annotations in the main `EdmModel` object.

```
<edmx:Reference  Uri="https://raw.githubusercontent.com/irvinesunday/msgraph-metadata/master/scopes_metadata/apiPermissionsAndScopes-v1.0.xml">
    <edmx:IncludeAnnotations  TermNamespace="Org.OData.Authorization.V1"/>
    <edmx:IncludeAnnotations  TermNamespace="Org.OData.Capabilities.V1"/>
  </edmx:Reference >
```
Main XML file: https://raw.githubusercontent.com/irvinesunday/msgraph-metadata/master/clean_v10_metadata/cleanMetadataWithDescriptionsv1.0.xml
Referenced XML file (with permissions annotations): https://raw.githubusercontent.com/irvinesunday/msgraph-metadata/master/scopes_metadata/apiPermissionsAndScopes-v1.0.xml

### Assemblies affected

`Microsoft.OData.Edm`, Version=7.6.1.0

### Reproduce steps

Below is the .NET code for parsing an `EdmModel` from a CSDL (including any referenced XML in the referencing XML document)

```
// Function to load referenced model xml
XmlReader getReferencedModelReaderFunc(Uri uri)
{
  if (uri != null)
  {
    var httpClient = new HttpClient();
    var csdl = httpClient.GetStringAsync(uri.OriginalString);
    var xmlReader =  XElement.Parse(csdl).CreateReader();
    return xmlReader;
  }
  return null;
}

var edmModel = CsdlReader.Parse(XElement.Load(csdl).CreateReader(), getReferencedModelReaderFunc);
```

### Expected result

The `EdmModel.VocabularyAnnotations` property should contain the collection of vocabulary annotations that are contained in both the referencing and referenced models.

### Actual result

The `EdmModel.VocabularyAnnotations` property does not contain the collection of vocabulary annotations that are contained in the referenced model.

habbes:
### Issues

Fix #1936 

### Description

Includes annotations referenced in the `edmx:IncludeAnnotations` tag of the `edmx:Reference` to be included when loading external referenced models.

This makes the vocabulary annotations imported from referenced models available through:
```c#
IEnumerable<IEdmVocabularyAnnotation> annotations = model.FindVocabularyAnnotations(element);
```
or
```c#
IEnumerable<IEdmVocabularyAnnotation> annotations = element.VocabularyAnnotations(model);
```

**Note**, these external vocabulary annotations will not appear in `model.FindDeclaredVocabularyAnnotations(element)`  or `model.VocabularyAnnotations` since they only return the annotations declared directly in the current model.

This PR attempts to restrict the imported annotations based on the `TermNamespace`, `TargetNamespace` and `Qualifier` 
as specified in the specification: http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_IncludedAnnotations

Assuming the main model has the following reference:
```xml
<edmx:Reference  Uri=""SimplePermissions.xml"">
    <edmx:IncludeAnnotations  TermNamespace=""Org.OData.Capabilities.V1"" Qualifier=""Insert""/>
    <edmx:IncludeAnnotations  TermNamespace=""OtherIncludedNS"" TargetNamespace=""Example.Types""/>
</edmx:Reference>
```

then this feature will import vocabulary annotations from the `SimplePermissions.xml` csdl which meet either of the following criteria:
- The annotation's term is in the namespace `Org.OData.Capabilities.V1` namespace and it has the qualifier `Insert`
- The annotation's term is in the namespace `OtherIncludedNS` and it's target element is in the namespace `Example.Types`

Other vocabulary annotations in the referenced csdl will be skipped.

**Note**: This restriction only applies to "out of line" annotations, i.e. annotations with are declared inside of a top-level `<Annotations>` element rather than being declared directly inside ("in-line") the target element.

**Note**: If the reference only has `edmx:IncludeAnnotations` declarations, but not `edmx:Include` declaration (http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_IncludedSchema), then this feature will only vocabulary annotations from the referenced models and exclude other schema elements like entity types, operations, etc.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
@@ -113,15 +113,7 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
 
             foreach (var schema in referencedCsdlModel.Schemata)
             {
-                string schemaNamespace = schema.Namespace;
-                bool shouldAddTypes = referencedCsdlModel.ParentModelReferences.SelectMany(s => s.Includes).Any(s => s.Namespace == schemaNamespace);
-                IEnumerable<IEdmIncludeAnnotations> includeAnnotations = referencedCsdlModel.ParentModelReferences.SelectMany(s => s.IncludeAnnotations);
-                bool shouldAddAnnotations = includeAnnotations.Any();
-
-                if (shouldAddTypes || shouldAddAnnotations)
-                {
-                    this.AddSchema(schema, shouldAddTypes, shouldAddAnnotations, includeAnnotations);
-                }
+                this.AddSchemaIfReferenced(schema, referencedCsdlModel.ParentModelReferences);
             }
         }
 
@@ -523,20 +515,17 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
                 ann => new CsdlSemanticsVocabularyAnnotation(schema, targetContext, annotationsContext, ann, qualifier));
         }
 
-        private void AddSchema(CsdlSchema schema, bool addSchemaElements = true)
+        private void AddSchema(CsdlSchema schema)
         {
-            this.AddSchema(schema, addSchemaElements, true, null);
+            this.AddSchema(schema, true);
         }
 
-        private void AddSchema(CsdlSchema schema, bool addSchemaElements, bool addAnnotations, IEnumerable<IEdmIncludeAnnotations> includeAnnotations = null)
+        private void AddSchema(CsdlSchema schema, bool addAnnotations)
         {
             CsdlSemanticsSchema schemaWrapper = new CsdlSemanticsSchema(this, schema);
             this.schemata.Add(schemaWrapper);
 
-            if (addSchemaElements)
-            {
-                AddSchemaElements(schema, schemaWrapper);
-            }
+            AddSchemaElements(schema, schemaWrapper);
 
             if (!string.IsNullOrEmpty(schema.Alias))
             {
@@ -545,13 +534,104 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
 
             if (addAnnotations)
             {
-                AddOutOfLineAnnotationsFromSchema(schema, schemaWrapper, includeAnnotations);
+                AddOutOfLineAnnotationsFromSchema(schema, schemaWrapper, includeAnnotationsIndex: null);
             }
 
             var edmVersion = this.GetEdmVersion();
             if (edmVersion == null || edmVersion < schema.Version)
             {
                 this.SetEdmVersion(schema.Version);
+            }
+        }
+
+        private void AddSchemaIfReferenced(CsdlSchema schema, IEnumerable<IEdmReference> parentReferences)
+        {
+            CsdlSemanticsSchema schemaWrapper = new CsdlSemanticsSchema(this, schema);
+
+            bool shouldAddSchemaElements;
+            Dictionary<string, List<IEdmIncludeAnnotations>> includeAnnotationsIndex;
+            ProcessSchemaParentReferences(schema, schemaWrapper, parentReferences, out shouldAddSchemaElements, out includeAnnotationsIndex);
+
+            if (!string.IsNullOrEmpty(schema.Alias))
+            {
+                this.SetNamespaceAlias(schema.Namespace, schema.Alias);
+            }
+
+            if (shouldAddSchemaElements)
+            {
+                AddSchemaElements(schema, schemaWrapper);
+            }
+
+            if (includeAnnotationsIndex.Count > 0)
+            {
+                AddOutOfLineAnnotationsFromSchema(schema, schemaWrapper, includeAnnotationsIndex);
+            }
+
+            if (shouldAddSchemaElements || includeAnnotationsIndex.Count > 0)
+            {
+                this.schemata.Add(schemaWrapper);
+
+                var edmVersion = this.GetEdmVersion();
+                if (edmVersion == null || edmVersion < schema.Version)
+                {
+                    this.SetEdmVersion(schema.Version);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Process the <see cref="IEdmReference"/>s of the parent model to
+        /// figure out whether the schema's types (entiy types, enums, containers, etc.) and annotations should
+        /// be imported
+        /// </summary>
+        /// <param name="schema"></param>
+        /// <param name="schemaWrapper"></param>
+        /// <param name="parentReferences"></param>
+        /// <param name="shouldAddSchemaElements">Whether schema types such entity types, operations, enums, etc. should be important</param>
+        /// <param name="includeAnnotationsIndex">Cache of the schema's out-of-line <see cref="IEdmIncludeAnnotations"/>s indexed by the term namespace</param>
+        private static void ProcessSchemaParentReferences(
+            CsdlSchema schema,
+            CsdlSemanticsSchema schemaWrapper,
+            IEnumerable<IEdmReference> parentReferences,
+            out bool shouldAddSchemaElements,
+            out Dictionary<string, List<IEdmIncludeAnnotations>> includeAnnotationsIndex)
+        {
+            string schemaNamespace = schema.Namespace;
+
+            shouldAddSchemaElements = false;
+            includeAnnotationsIndex = new Dictionary<string, List<IEdmIncludeAnnotations>>();
+            
+
+            foreach (IEdmReference reference in parentReferences)
+            {
+                if (!shouldAddSchemaElements)
+                {
+                    // we should add types from this schema if there's at least
+                    // one edmx:Include that references this schema's namespace
+                    foreach (IEdmInclude include in reference.Includes)
+                    {
+                        if (include.Namespace == schemaNamespace)
+                        {
+                            shouldAddSchemaElements = true;
+                            break;
+                        }
+                    }
+                }
+
+                foreach(IEdmIncludeAnnotations includeAnnotations in reference.IncludeAnnotations)
+                {
+                    // index the edmx:IncludeAnnotations by namespace to make it
+                    // easier to filter which annotations to import with the schema
+                    string termNamespace = includeAnnotations.TermNamespace;
+                    List<IEdmIncludeAnnotations> cachedAnnotations;
+                    if (!includeAnnotationsIndex.TryGetValue(termNamespace, out cachedAnnotations))
+                    {
+                        cachedAnnotations = new List<IEdmIncludeAnnotations>();
+                        includeAnnotationsIndex[termNamespace] = cachedAnnotations;
+                    }
+
+                    cachedAnnotations.Add(includeAnnotations);
+                }
             }
         }
 
@@ -602,12 +682,15 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
             }
         }
 
-        private void AddOutOfLineAnnotationsFromSchema(CsdlSchema schema, CsdlSemanticsSchema schemaWrapper, IEnumerable<IEdmIncludeAnnotations> includeAnnotations)
+        private void AddOutOfLineAnnotationsFromSchema(
+            CsdlSchema schema,
+            CsdlSemanticsSchema schemaWrapper,
+            Dictionary<string, List<IEdmIncludeAnnotations>> includeAnnotationsIndex)
         {
             foreach (CsdlAnnotations schemaOutOfLineAnnotations in schema.OutOfLineAnnotations)
             {
                 string target = this.ReplaceAlias(schemaOutOfLineAnnotations.Target);
-
+                
                 List<CsdlSemanticsAnnotations> annotations;
                 if (!this.outOfLineAnnotations.TryGetValue(target, out annotations))
                 {
@@ -615,35 +698,51 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
                     this.outOfLineAnnotations[target] = annotations;
                 }
 
-                CsdlAnnotations filteredAnnotations = FilterIncludedAnnotations(schemaOutOfLineAnnotations, target, includeAnnotations);
+                CsdlAnnotations filteredAnnotations = FilterIncludedAnnotations(schemaOutOfLineAnnotations, target, includeAnnotationsIndex);
                 annotations.Add(new CsdlSemanticsAnnotations(schemaWrapper, filteredAnnotations));
             }
         }
 
-        private CsdlAnnotations FilterIncludedAnnotations(CsdlAnnotations csdlAnnotations, string target, IEnumerable<IEdmIncludeAnnotations> includeAnnotations)
+        private CsdlAnnotations FilterIncludedAnnotations(
+            CsdlAnnotations csdlAnnotations,
+            string target,
+            Dictionary<string, List<IEdmIncludeAnnotations>> includeAnnotationsIndex)
         {
-            if (includeAnnotations == null || !includeAnnotations.Any())
+            if (includeAnnotationsIndex == null)
             {
                 return csdlAnnotations;
             }
 
             var filteredAnnotations = csdlAnnotations.Annotations.Where(annotation =>
-                ShouldIncludeAnnotation(annotation, target, includeAnnotations));
+                ShouldIncludeAnnotation(annotation, target, includeAnnotationsIndex));
             return new CsdlAnnotations(filteredAnnotations, csdlAnnotations.Target, csdlAnnotations.Qualifier);
         }
 
-        private bool ShouldIncludeAnnotation(CsdlAnnotation annotation, string target, IEnumerable<IEdmIncludeAnnotations> includeAnnotations)
+        private bool ShouldIncludeAnnotation(
+            CsdlAnnotation annotation,
+            string target,
+            Dictionary<string, List<IEdmIncludeAnnotations>> includeAnnotationsIndex)
         {
             // include annotation if it matches one of the edmx:IncludeAnnotation references
             // we check if the namespace matches, and (if applicable) also the target namespace and the qualifier
             // see: http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_IncludedAnnotations
 
-            return includeAnnotations.Any(include =>
-                (this.ReplaceAlias(annotation.Term).StartsWith(include.TermNamespace, System.StringComparison.Ordinal))
-                    && (string.IsNullOrEmpty(include.Qualifier) || annotation.Qualifier == include.Qualifier)
-                    && (string.IsNullOrEmpty(include.TargetNamespace)
-                        || target.StartsWith(include.TargetNamespace, System.StringComparison.Ordinal))
-            );
+            string qualifiedTerm = this.ReplaceAlias(annotation.Term);
+            string targetNamespace;
+            EdmUtil.TryGetNamespaceNameFromQualifiedName(target, out targetNamespace, out string baseTarget);
+            if (EdmUtil.TryGetNamespaceNameFromQualifiedName(qualifiedTerm, out string termNamespace, out string baseTerm))
+            {
+                if (includeAnnotationsIndex.TryGetValue(termNamespace, out List<IEdmIncludeAnnotations> includeAnnotations))
+                {
+
+                    return includeAnnotations.Any(include =>
+                        (string.IsNullOrEmpty(include.Qualifier) || annotation.Qualifier == include.Qualifier)
+                        && (string.IsNullOrEmpty(include.TargetNamespace) || targetNamespace == include.TargetNamespace));
+                }
+                   
+            }
+
+            return false;
         }
     }
 }
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
@@ -115,14 +115,13 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
             {
                 string schemaNamespace = schema.Namespace;
                 IEdmInclude edmInclude = referencedCsdlModel.ParentModelReferences.SelectMany(s => s.Includes).FirstOrDefault(s => s.Namespace == schemaNamespace);
-                bool includeAnnotations = referencedCsdlModel.ParentModelReferences.SelectMany(s => s.IncludeAnnotations).Any();
-                if (edmInclude != null || includeAnnotations)
-                {
-                    this.AddSchema(schema, includeAnnotations);
-                    //this.AddSchema(schema, false);
-                }
+                IEnumerable<IEdmIncludeAnnotations> includeAnnotations = referencedCsdlModel.ParentModelReferences.SelectMany(s => s.IncludeAnnotations);
+                bool shouldAddAnnotations = includeAnnotations.Any();
 
-                // TODO: REF add annotations
+                if (edmInclude != null || shouldAddAnnotations)
+                {
+                    this.AddSchema(schema, shouldAddAnnotations, includeAnnotations);
+                }
             }
         }
 
@@ -529,7 +528,7 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
             this.AddSchema(schema, true);
         }
 
-        private void AddSchema(CsdlSchema schema, bool addAnnotations)
+        private void AddSchema(CsdlSchema schema, bool addAnnotations, IEnumerable<IEdmIncludeAnnotations> includeAnnotations = null)
         {
             CsdlSemanticsSchema schemaWrapper = new CsdlSemanticsSchema(this, schema);
             this.schemata.Add(schemaWrapper);
@@ -584,19 +583,7 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
 
             if (addAnnotations)
             {
-                foreach (CsdlAnnotations schemaOutOfLineAnnotations in schema.OutOfLineAnnotations)
-                {
-                    string target = this.ReplaceAlias(schemaOutOfLineAnnotations.Target);
-
-                    List<CsdlSemanticsAnnotations> annotations;
-                    if (!this.outOfLineAnnotations.TryGetValue(target, out annotations))
-                    {
-                        annotations = new List<CsdlSemanticsAnnotations>();
-                        this.outOfLineAnnotations[target] = annotations;
-                    }
-
-                    annotations.Add(new CsdlSemanticsAnnotations(schemaWrapper, schemaOutOfLineAnnotations));
-                }
+                AddOutOfLineAnnotationsFromSchema(schema, schemaWrapper, includeAnnotations);
             }
 
             var edmVersion = this.GetEdmVersion();
@@ -606,9 +593,49 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
             }
         }
 
-        private void AddOutOfLineAnnotationsFromSchema(CsdlSchema schema)
+        private void AddOutOfLineAnnotationsFromSchema(CsdlSchema schema, CsdlSemanticsSchema schemaWrapper, IEnumerable<IEdmIncludeAnnotations> includeAnnotations)
         {
+            foreach (CsdlAnnotations schemaOutOfLineAnnotations in schema.OutOfLineAnnotations)
+            {
+                string target = this.ReplaceAlias(schemaOutOfLineAnnotations.Target);
 
+                List<CsdlSemanticsAnnotations> annotations;
+                if (!this.outOfLineAnnotations.TryGetValue(target, out annotations))
+                {
+                    annotations = new List<CsdlSemanticsAnnotations>();
+                    this.outOfLineAnnotations[target] = annotations;
+                }
+
+                CsdlAnnotations filteredAnnotations = FilterIncludedAnnotations(schemaOutOfLineAnnotations, target, includeAnnotations);
+                annotations.Add(new CsdlSemanticsAnnotations(schemaWrapper, filteredAnnotations));
+            }
+        }
+
+        private CsdlAnnotations FilterIncludedAnnotations(CsdlAnnotations csdlAnnotations, string target, IEnumerable<IEdmIncludeAnnotations> includeAnnotations)
+        {
+            
+            if (includeAnnotations == null && !includeAnnotations.Any())
+            {
+                return csdlAnnotations;
+            }
+
+            var filteredAnnotations = csdlAnnotations.Annotations.Where(annotation =>
+                ShouldIncludeAnnotation(annotation, target, includeAnnotations));
+            return new CsdlAnnotations(filteredAnnotations, csdlAnnotations.Target, csdlAnnotations.Qualifier);
+        }
+
+        private bool ShouldIncludeAnnotation(CsdlAnnotation annotation, string target, IEnumerable<IEdmIncludeAnnotations> includeAnnotations)
+        {
+            // include annotation if it matches one of the edmx:IncludeAnnotation references
+            // we check if the namespace matches, and (if applicable) also the target namespace and the qualifier
+            // see: http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_IncludedAnnotations
+
+            return includeAnnotations.Any(include =>
+                (this.ReplaceAlias(annotation.Term).StartsWith(include.TermNamespace, System.StringComparison.Ordinal))
+                    && (string.IsNullOrEmpty(include.Qualifier) || annotation.Qualifier == include.Qualifier)
+                    && (string.IsNullOrEmpty(include.TargetNamespace)
+                        ||target.StartsWith(include.TargetNamespace, System.StringComparison.Ordinal))
+            );
         }
     }
 }
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
@@ -534,7 +534,8 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
 
             if (addAnnotations)
             {
-                AddOutOfLineAnnotationsFromSchema(schema, schemaWrapper, includeAnnotationsIndex: null);
+                // this adds all out-of-line annotations in the schema regardless of edmx:IncludeAnnotations references in the model
+                AddOutOfLineAnnotationsFromSchema(schema, schemaWrapper, /* includeAnnotationsIndex */ null);
             }
 
             var edmVersion = this.GetEdmVersion();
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -114,13 +114,14 @@ private CsdlSemanticsModel(CsdlModel referencedCsdlModel, IEdmDirectValueAnnotat
             foreach (var schema in referencedCsdlModel.Schemata)
             {
                 string schemaNamespace = schema.Namespace;
-                IEdmInclude edmInclude = referencedCsdlModel.ParentModelReferences.SelectMany(s => s.Includes).FirstOrDefault(s => s.Namespace == schemaNamespace);
-                if (edmInclude != null)
+                bool shouldAddTypes = referencedCsdlModel.ParentModelReferences.SelectMany(s => s.Includes).Any(s => s.Namespace == schemaNamespace);
```
gathogojr, line 117:
```suggestion
                bool addSchemaElements = referencedCsdlModel.ParentModelReferences.SelectMany(s => s.Includes).Any(s => s.Namespace == schemaNamespace);
```

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
```diff
@@ -1230,5 +1230,242 @@ public void ValidateEdmxVersions(string odataVersion)
...
+            var product = model.FindDeclaredType("Example.Types.Product");
+            var productAnnotations = model.FindVocabularyAnnotations(product);
+            Assert.Single(productAnnotations);
+            IEdmVocabularyAnnotation productAnnotation = productAnnotations.FirstOrDefault(a => a.Term.Name == "SomeAnnotationPath");
+            Assert.NotNull(productAnnotation);
+        }
+
+        
     }
 }
```
gathogojr, line 1471: Add newline

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -522,15 +523,40 @@ private IEdmVocabularyAnnotation WrapVocabularyAnnotation(CsdlAnnotation annotat
...
+                this.SetNamespaceAlias(schema.Namespace, schema.Alias);
+            }
+
+            if (addAnnotations)
+            {
+                AddOutOfLineAnnotationsFromSchema(schema, schemaWrapper, includeAnnotations);
+            }
+
+            var edmVersion = this.GetEdmVersion();
+            if (edmVersion == null || edmVersion < schema.Version)
```
xuzhg, line 552: >edmVersion < schema.Version [](start = 38, length = 27)

why do we have this check?

xuzhg (approved):
:shipit:

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -522,15 +523,40 @@ private IEdmVocabularyAnnotation WrapVocabularyAnnotation(CsdlAnnotation annotat
                 ann => new CsdlSemanticsVocabularyAnnotation(schema, targetContext, annotationsContext, ann, qualifier));
         }
 
-        private void AddSchema(CsdlSchema schema)
+        private void AddSchema(CsdlSchema schema, bool addSchemaElements = true)
         {
-            this.AddSchema(schema, true);
+            this.AddSchema(schema, addSchemaElements, true, null);
```
gathogojr, line 528:
```suggestion
            this.AddSchema(schema, addSchemaElements, true /*addAnnotations*/, null/*includeAnnotations*/);
```

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -574,34 +600,50 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+                CsdlAnnotations filteredAnnotations = FilterIncludedAnnotations(schemaOutOfLineAnnotations, target, includeAnnotations);
+                annotations.Add(new CsdlSemanticsAnnotations(schemaWrapper, filteredAnnotations));
             }
+        }
 
-            var edmVersion = this.GetEdmVersion();
-            if (edmVersion == null || edmVersion < schema.Version)
+        private CsdlAnnotations FilterIncludedAnnotations(CsdlAnnotations csdlAnnotations, string target, IEnumerable<IEdmIncludeAnnotations> includeAnnotations)
+        {
+            if (includeAnnotations == null || !includeAnnotations.Any())
```
gathogojr, line 625:
```suggestion
            if (includeAnnotations?.Any() != true)
```

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -574,34 +600,50 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+        {
+            // include annotation if it matches one of the edmx:IncludeAnnotation references
+            // we check if the namespace matches, and (if applicable) also the target namespace and the qualifier
+            // see: http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_IncludedAnnotations
+
+            return includeAnnotations.Any(include =>
+                (this.ReplaceAlias(annotation.Term).StartsWith(include.TermNamespace, System.StringComparison.Ordinal))
+                    && (string.IsNullOrEmpty(include.Qualifier) || annotation.Qualifier == include.Qualifier)
+                    && (string.IsNullOrEmpty(include.TargetNamespace)
+                        || target.StartsWith(include.TargetNamespace, System.StringComparison.Ordinal))
```
gathogojr, line 645: Is use of `StartsWith` here fool proof? `TargetNamespace` with a value of _a/b/c_ would be matched to the following for example
- _a/b/c/d_
- _a/b/c_

habbes: That's a good point. I'll fix that.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
habbes: @xuzhg This was there in the previous code, I didn't change it. Maybe it's meant to ensure that latest version of all the schemas is the one that's applied to the main schema.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
habbes: Maybe I could use named parameters instead, like:
```c#
this.AddSchema(schema, addSchemaElements, addAnnotations: true, includeAnnotations: null);
```

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -114,13 +114,14 @@ private CsdlSemanticsModel(CsdlModel referencedCsdlModel, IEdmDirectValueAnnotat
             foreach (var schema in referencedCsdlModel.Schemata)
             {
                 string schemaNamespace = schema.Namespace;
-                IEdmInclude edmInclude = referencedCsdlModel.ParentModelReferences.SelectMany(s => s.Includes).FirstOrDefault(s => s.Namespace == schemaNamespace);
-                if (edmInclude != null)
+                bool shouldAddTypes = referencedCsdlModel.ParentModelReferences.SelectMany(s => s.Includes).Any(s => s.Namespace == schemaNamespace);
+                IEnumerable<IEdmIncludeAnnotations> includeAnnotations = referencedCsdlModel.ParentModelReferences.SelectMany(s => s.IncludeAnnotations);
```
mikepizzo, line 118: >includeAnnotations [](start = 52, length = 18)

I'm concerned about the performance of executing this (somewhat expensive) SelectMany LINQ expression multiple times -- we execute it once on line 119, again on line 625, and then for each term on line 641.  Probably be better, especially since we are iterating multiple times, to make this a list.  removing the linq expression all together and simply iterating through building the collection would be even better, but probably not a significant savings here.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -574,34 +600,50 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+            return new CsdlAnnotations(filteredAnnotations, csdlAnnotations.Target, csdlAnnotations.Qualifier);
+        }
+
+        private bool ShouldIncludeAnnotation(CsdlAnnotation annotation, string target, IEnumerable<IEdmIncludeAnnotations> includeAnnotations)
+        {
+            // include annotation if it matches one of the edmx:IncludeAnnotation references
+            // we check if the namespace matches, and (if applicable) also the target namespace and the qualifier
+            // see: http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_IncludedAnnotations
+
+            return includeAnnotations.Any(include =>
```
mikepizzo, line 641: >Any [](start = 38, length = 3)

LINQ is expensive -- AGS has gone through a LINQ hunt and removed all use of LINQ from their code.  We should avoid it where we can.

Instead, it would be more efficient to iterate through the collection and evaluate the options.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
```diff
@@ -1230,5 +1230,242 @@ public void ValidateEdmxVersions(string odataVersion)
...
         }
+
+        [Fact]
+        public void ImportsAnnotationsFromExernalReferencedModel()
+        {
+            string permissionsCsdl = @"<?xml version=""1.0"" encoding=""utf-8""?>
+<edmx:Edmx Version=""4.0"" xmlns:edmx=""http://docs.oasis-open.org/odata/ns/edmx"">
+  <edmx:Reference  Uri=""SimpleModel.xml"">
+    <edmx:Include Namespace=""Default""/>
+    <edmx:Include Namespace=""Example.Types"" />
```
mikepizzo, line 1241: Maybe add a test where the annotation schema defines an alias for the referenced namespace that is used locally in specifying the target, and make sure that annotations restricted to that target are still (exclusively) included.

i.e., define an alias for the "Example.Types" namespace in permissions.Csdl (say "examples") and use it in the Annotations Target, and then in SimpleModel.xml specify a TargetNamespace of "Example.Types" in your include annotations and make sure it still returns the right set of annotations (I think your code handles that, just want to make sure it's in the tests...)

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -574,34 +600,50 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+        {
+            if (includeAnnotations == null || !includeAnnotations.Any())
             {
-                this.SetEdmVersion(schema.Version);
+                return csdlAnnotations;
             }
+
+            var filteredAnnotations = csdlAnnotations.Annotations.Where(annotation =>
+                ShouldIncludeAnnotation(annotation, target, includeAnnotations));
+            return new CsdlAnnotations(filteredAnnotations, csdlAnnotations.Target, csdlAnnotations.Qualifier);
```
mikepizzo, line 632: >csdlAnnotations.Target [](start = 60, length = 22)

should this use csdlAnnotations.Target, or "target" (which has the alias already resolved)?

KenitoInc (changes requested):

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -522,15 +523,40 @@ private IEdmVocabularyAnnotation WrapVocabularyAnnotation(CsdlAnnotation annotat
                 ann => new CsdlSemanticsVocabularyAnnotation(schema, targetContext, annotationsContext, ann, qualifier));
         }
 
-        private void AddSchema(CsdlSchema schema)
+        private void AddSchema(CsdlSchema schema, bool addSchemaElements = true)
```
marabooy, line 526: Would the default of adding schema elements cause any regressions in any cases and could we document how the end user would disable this incase that is the expected behaviour from their end.

habbes: The previous behaviour was adding the schema elements by default when you called `AddSchema()`. But `AddSchema()` was only ever called if you wanted to add schema elements. With this change, `AddSchema()` can also be called when you only want to add annotations but not schema elements, or both annotations and schema elements. So setting the default for the argument to `true` retains the old behaviour while providing a way to exclude schema elements when necessary.

By the way, I'm using schema elements to refer to things like entities, enums, containers, etc. defined in the schema. I'm not sure if the variable name captures that effectively.

habbes: Also, this method is private, and only used a few times in this class.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
habbes: The aim here is to return a copy of the `CsdlAnnotations` instance but with the actual annotations filtered. So every other property should be identical to the original `CsdlAnnotations` instance, hence `csdlAnnotations.Target`. The `target` with the alias already resolved is used when comparing against the `include.TargetNamespace`.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
habbes: I've added the test

habbes:
@mikepizzo I've made changes to how the `Include` and `IncludeAnnotations` are retrieve from the references to avoid potentially re-evaluating the LINQ `SelectMany()` calls.

Now I loop over the references only once using a `foreach` loop and I store the `IncludeAnnotations` in a dictionary indexed by the term namespace to make the look up a bit quicker when I am filtering the annotations to include.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -531,6 +524,119 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+
+        /// <summary>
+        /// Process the <see cref="IEdmReference"/>s of the parent model to
+        /// figure out whether the schema's types (entiy types, enums, containers, etc.) and annotations should
+        /// be imported
+        /// </summary>
+        /// <param name="schema"></param>
+        /// <param name="schemaWrapper"></param>
+        /// <param name="parentReferences"></param>
+        /// <param name="shouldAddSchemaElements">Whether schema types such entity types, operations, enums, etc. should be important</param>
```
gathogojr, line 590:
```suggestion
        /// <param name="shouldAddSchemaElements">Whether schema types such entity types, operations, enums, etc. should be add</param>
```

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -531,6 +524,119 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Process the <see cref="IEdmReference"/>s of the parent model to
+        /// figure out whether the schema's types (entiy types, enums, containers, etc.) and annotations should
+        /// be imported
+        /// </summary>
+        /// <param name="schema"></param>
```
gathogojr, line 587: Document this param and the following 2

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -531,6 +524,119 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+                if (edmVersion == null || edmVersion < schema.Version)
+                {
+                    this.SetEdmVersion(schema.Version);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Process the <see cref="IEdmReference"/>s of the parent model to
+        /// figure out whether the schema's types (entiy types, enums, containers, etc.) and annotations should
```
gathogojr, line 584:
```suggestion
        /// figure out whether the schema's types (entity types, enums, containers, etc.) and annotations should
```

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -531,6 +524,119 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+            IEnumerable<IEdmReference> parentReferences,
+            out bool shouldAddSchemaElements,
+            out Dictionary<string, List<IEdmIncludeAnnotations>> includeAnnotationsIndex)
+        {
+            string schemaNamespace = schema.Namespace;
+
+            shouldAddSchemaElements = false;
+            includeAnnotationsIndex = new Dictionary<string, List<IEdmIncludeAnnotations>>();
+            
+
```
gathogojr, line 604:
```suggestion
```

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -531,6 +524,119 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+            CsdlSchema schema,
+            CsdlSemanticsSchema schemaWrapper,
+            IEnumerable<IEdmReference> parentReferences,
+            out bool shouldAddSchemaElements,
+            out Dictionary<string, List<IEdmIncludeAnnotations>> includeAnnotationsIndex)
+        {
+            string schemaNamespace = schema.Namespace;
+
+            shouldAddSchemaElements = false;
+            includeAnnotationsIndex = new Dictionary<string, List<IEdmIncludeAnnotations>>();
```
gathogojr, line 602: Maybe apply the constructor that takes an `IEqualityComparer<T>` argument if `TermNamespace="Org.OData.Capabilities.V1"` is considered to be equivalent to `TermNamespace="org.OData.Capabilities.V1"` (i.e. case-insensitive)

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -531,6 +524,119 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+                        if (include.Namespace == schemaNamespace)
+                        {
+                            shouldAddSchemaElements = true;
+                            break;
+                        }
+                    }
+                }
+
+                foreach(IEdmIncludeAnnotations includeAnnotations in reference.IncludeAnnotations)
+                {
```
gathogojr, line 622: @habbes The logic in this block is hard to follow...

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -531,6 +524,119 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+                    // easier to filter which annotations to import with the schema
+                    string termNamespace = includeAnnotations.TermNamespace;
+                    List<IEdmIncludeAnnotations> cachedAnnotations;
+                    if (!includeAnnotationsIndex.TryGetValue(termNamespace, out cachedAnnotations))
+                    {
+                        cachedAnnotations = new List<IEdmIncludeAnnotations>();
+                        includeAnnotationsIndex[termNamespace] = cachedAnnotations;
+                    }
+
+                    cachedAnnotations.Add(includeAnnotations);
```
gathogojr, line 633: After we add `includeAnnotations` to `cachedAnnotations` what do we do with that?

gathogojr: Understood

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -574,34 +680,69 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
             }
+        }
 
-            if (addAnnotations)
+        private CsdlAnnotations FilterIncludedAnnotations(
+            CsdlAnnotations csdlAnnotations,
+            string target,
+            Dictionary<string, List<IEdmIncludeAnnotations>> includeAnnotationsIndex)
+        {
+            if (includeAnnotationsIndex == null)
```
gathogojr, line 711: Is this ever likely to evaluate to `true` given the initialization of `includeAnnotationsIndex` [here](https://github.com/OData/odata.net/pull/1987/files#diff-f5d2f568fee616e97aad18909cd5716484d0e0b873569410006f6896b0c5c6f4R602)? Maybe check for `null` or empty (or just empty if `null` is unlikely)?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -574,34 +680,69 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+            EdmUtil.TryGetNamespaceNameFromQualifiedName(target, out targetNamespace, out string baseTarget);
+            if (EdmUtil.TryGetNamespaceNameFromQualifiedName(qualifiedTerm, out string termNamespace, out string baseTerm))
+            {
+                if (includeAnnotationsIndex.TryGetValue(termNamespace, out List<IEdmIncludeAnnotations> includeAnnotations))
+                {
 
-                    annotations.Add(new CsdlSemanticsAnnotations(schemaWrapper, schemaOutOfLineAnnotations));
+                    return includeAnnotations.Any(include =>
+                        (string.IsNullOrEmpty(include.Qualifier) || annotation.Qualifier == include.Qualifier)
+                        && (string.IsNullOrEmpty(include.TargetNamespace) || targetNamespace == include.TargetNamespace));
```
gathogojr, line 740: When dealing with `IncludeAnnotations`, can `TargetNamespace` be `null` or empty?
The [spec](http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_IncludedAnnotations) says: _The edmx:IncludeAnnotations element MUST provide the TermNamespace attribute_

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -574,34 +680,69 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+            string target,
+            Dictionary<string, List<IEdmIncludeAnnotations>> includeAnnotationsIndex)
+        {
+            // include annotation if it matches one of the edmx:IncludeAnnotation references
+            // we check if the namespace matches, and (if applicable) also the target namespace and the qualifier
+            // see: http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_IncludedAnnotations
+
+            string qualifiedTerm = this.ReplaceAlias(annotation.Term);
+            string targetNamespace;
+            EdmUtil.TryGetNamespaceNameFromQualifiedName(target, out targetNamespace, out string baseTarget);
```
gathogojr, line 732:
```suggestion
            EdmUtil.TryGetNamespaceNameFromQualifiedName(target, out targetNamespace, out _);
```

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -574,34 +680,69 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+            if (EdmUtil.TryGetNamespaceNameFromQualifiedName(qualifiedTerm, out string termNamespace, out string baseTerm))
```
gathogojr, line 733:
```suggestion
            if (EdmUtil.TryGetNamespaceNameFromQualifiedName(qualifiedTerm, out string termNamespace, out _))
```

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -574,34 +680,69 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+            {
+                if (includeAnnotationsIndex.TryGetValue(termNamespace, out List<IEdmIncludeAnnotations> includeAnnotations))
+                {
 
-                    annotations.Add(new CsdlSemanticsAnnotations(schemaWrapper, schemaOutOfLineAnnotations));
+                    return includeAnnotations.Any(include =>
```
gathogojr, line 738: Just to confirm `annotation.Qualifier == include.Qualifier` and `targetNamespace == include.TargetNamespace` will be a case-sensitive comparison...

Maybe 
```csharp
(string.IsNullOrEmpty(include.Qualifier) || include.Qualifier.Equals(annotation.Qualifier, StringComparison.OrdinalIgnoreCase))
&& (string.IsNullOrEmpty(include.TargetNamespace) || include.TargetNamespace.Equals(targetNamespace, StringComparison.OrdinalIgnoreCase)));
```
?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -531,6 +524,119 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+            if (includeAnnotationsIndex.Count > 0)
+            {
+                AddOutOfLineAnnotationsFromSchema(schema, schemaWrapper, includeAnnotationsIndex);
+            }
+
+            if (shouldAddSchemaElements || includeAnnotationsIndex.Count > 0)
+            {
+                this.schemata.Add(schemaWrapper);
+
+                var edmVersion = this.GetEdmVersion();
```
gathogojr, line 574: Based on the previous logic, the block of code for setting Edm version was not conditional... I think it should move outside the `if (shouldAddSchemaElements || includeAnnotationsIndex.Count > 0)` block

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -531,6 +524,119 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+            }
+
+            if (includeAnnotationsIndex.Count > 0)
+            {
+                AddOutOfLineAnnotationsFromSchema(schema, schemaWrapper, includeAnnotationsIndex);
+            }
+
+            if (shouldAddSchemaElements || includeAnnotationsIndex.Count > 0)
+            {
+                this.schemata.Add(schemaWrapper);
```
gathogojr, line 572: Given that `schemata` field is not documented, could you add a helpful comment to this statement?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
KenitoInc: @habbes Use John's suggestion. Named parameters aren't supported in some c# versions.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
habbes: I believe namespaces are case-sensitive. At least in many cases, the spec mentions that identifiers are case-sensitive. I assume it applies to namespaces as well.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
habbes: The block was not conditional, but the `AddSchema` method was called conditionally (i.e. only when there was an eligible EdmInclude reference). Now the AddSchema method is always called, so the aim here was to make it so that it's "side-effects" are only observed if schema elements or annotations are imported.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
habbes: It's actually set to null when called from the original `AddSchema(schema, addAnnotations)` method [here](https://github.com/OData/odata.net/pull/1987/files#diff-f5d2f568fee616e97aad18909cd5716484d0e0b873569410006f6896b0c5c6f4R535-R538).

In that method, when the `addAnnotations` was set true, all the annotations in the schema were added without any filtering. In that case I call the `AddSchemaOutOfLineAnnotations` with `includeAnnotationsIndex` set to `null` to indicate that all annotations should be added without any filtering. If, on the other hand, the `includeAnnotationsIndex` is empty, then all annotations will be filtered out and not get added, since the `ShouldIncludeAnnotation` method will return false.

I think I should mention in the doc comments what the intended behaviour is when `includeAnnotationsIndex` is `null`.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
habbes: I believe they are case sensitive

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -531,6 +524,119 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+            }
+
+            var edmVersion = this.GetEdmVersion();
+            if (edmVersion == null || edmVersion < schema.Version)
+            {
+                this.SetEdmVersion(schema.Version);
+            }
+        }
+
+        private void AddSchemaIfReferenced(CsdlSchema schema, IEnumerable<IEdmReference> parentReferences)
```
KenitoInc, line 547: For better code readability and maintainability, I would suggest you add doc strings to private methods. I have seen this practice in other ODL code https://github.com/OData/odata.net/blob/6f145fdb1c24a79850f0b34b645e1163bf150fd8/src/Microsoft.OData.Core/UriParser/Binders/BinaryOperatorBinder.cs#L85

KenitoInc (changes requested):

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
habbes: I think I should make an optimization to avoid filtering the annotations if the `includeAnnotationsIndex` is empty, since in that case none of the csdl annotations will be imported.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
habbes: `schemata` is the list of schemas in this model. The term `schemata` is just the Latin plural form of `schema`.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
gathogojr: Indeed: _Names are case-sensitive, but service authors SHOULD NOT choose names that differ only in case._

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
habbes: I've added doc comments to the private methods.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -574,35 +691,104 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+        /// Adds out-of-line annotations from the specified schema
+        /// if they are referenced in the <paramref name="includeAnnotationsIndex"/> cache.
+        /// </summary>
+        /// <param name="schema">The schema to add annotations from.</param>
+        /// <param name="schemaWrapper">The <see cref="CsdlSemanticsSchema"/> wrapper for the provided schema</param>
+        /// <param name="includeAnnotationsIndex">Cache of the schema's out-of-line <see cref="IEdmIncludeAnnotations"/>s indexed by the term namespace.
+        /// If this dictionary is non-empty, then only annotations that match
+        /// the references will be added. If it's empty, no annotations will be added. If it's null, all annotations
+        /// will be added unconditionally.</param>
+        private void AddOutOfLineAnnotationsFromSchema(
```
gathogojr, line 706: @habbes RE: _If it's null, all annotations will be added unconditionally_
Does this happen from the calling method or this (`AddOutOfLineAnnotationsFromSchema`) method?

gathogojr (approved):
LGTM

KenitoInc (approved):
LGTM

KenitoInc:
I see the build is failing.
You have defined `AddSchemaElements` accepting 1 parameter, but calling it with 2 parameters.

mikepizzo:
:grinning: 

---
In reply to: [772912424](https://github.com/OData/odata.net/pull/1987#issuecomment-772912424) [](ancestors = 772912424)

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -574,35 +691,104 @@ private void AddSchema(CsdlSchema schema, bool addAnnotations)
...
+            CsdlSchema schema,
+            CsdlSemanticsSchema schemaWrapper,
+            Dictionary<string, List<IEdmIncludeAnnotations>> includeAnnotationsIndex)
+        {
+            if (includeAnnotationsIndex?.Count == 0)
```
marabooy, line 711:
```suggestion
            if (includeAnnotationsIndex?.Any() == false)

```

marabooy: or handle the null in a better way as Any can cause some issues with performance here due. I believe `null == 0` wil give you false but I could be wrong.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
```diff
@@ -1230,5 +1231,323 @@ public void ValidateEdmxVersions(string odataVersion)
             Assert.Empty(edmErrors);
             Assert.Equal(edmModel.GetEdmVersion(), odataVersion == "4.0" ? EdmConstants.EdmVersion4 : EdmConstants.EdmVersion401);
         }
+
+        [Fact]
```
mikepizzo, line 1235: >Fact [](start = 9, length = 4)

Do you have any tests that validate referenced types (not annotations) from referenced namespace.

For example, in your tests, you could test loading permissionsCsdl and validating that it imported types from "Default" and "Exmaple.Types" from main.csdl, but didn't import types from main.csdl defined in 3rd namespace.

mikepizzo (approved):
:shipit:

mikepizzo:
nice.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
habbes: That actually takes place in the `FilterIncludedAnnotations` method, which is called by this method (`AddOutOfLineAnnotationsFromSchema`). The `FilterIncludedAnnotations` method has a similar note in its doc comments.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
habbes: I've added the test.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
habbes: I didn't understand this comment very well. But I made the null check more explicit `includeAnnotationsIndex != null && includeAnnotationsIndex.Count == 0`

PR #1990, opened 2021-01-26
Fix performance tests

### LINKED ISSUE #1988: Fix performance tests
Build errors occur when you try to run ODL performance tests.

Current performance tests are based on the [Microsoft.Xunit.Performance](https://github.com/Microsoft/xunit-performance) library which seems to no longer maintained. It also doesn't seem like it ever got an official stable release on NuGet. Consider migrating performance tests to [BenchmarkDotNet](https://benchmarkdotnet.org/), which seems to be the current standard in performance testing/benchmarking in .Net and used by other Microsoft projects.

### Assemblies affected

### Reproduce steps

Try to build the `OData.Tests.Performance` solution.

Or try to run the `PerformanceBuild.ps1` script.

### Expected result

Build and run performance tests then generate performance reports.

### Actual result

A number of build errors occur.

### Additional detail

It would be great to get reports from the current performance tests before migrating to a new benchmarking framework so that we can compare the reports before and after migration. If the reports are too different it could indicate that the migration was not done correctly. However, it's been a challenge to get the current tests to run as-is.

To get Microsoft.Xunit.Performance dependency installed in VS, add the following myget feed to the NuGet sources: https://dotnet.myget.org/F/dotnet-core/api/v3/index.json

habbes:
### Issues

Fix #1988 

### Description

- Fixes performance tests and build errors such that running `PerformanceBuild.ps1` should now work.
- Migrated the ODL performance tests (`test\PerformanceTests\ComponentTests` and `test\PerformanceTests\ServiceTests`) to [BenchmarkDotNet](https://benchmarkdotnet.org/)

####  `Microsoft.Xunit.Performance` (XP) to `BenchmarkDotNet` (BDN) migration

In both frameworks, you create a class that contains your benchmarks. The benchmark runner will execute the methods annotated with the `[Benchmark]` attribute and collect their measurements.

With XP, you create your project as a class library and use a xunit runner to run your benchmarks and generate result reports. With BDN, your create your project as console application and you invoke the BDN runner from the `Main` method.

##### Setup and clean up code

An XP benchmark method has the following structure:
```c#
class MyBenchmarkTests {

[Benchmark]
public void SomeBenchmark()
{
   // any setup that should run once before all iterations  should go here
   // e.g. setting up resources
   foreach (var iteration in Benchmark.Iterations)
   {
      
      // setup code that should be executed before each iteration goes here
      using (iteration.StartMeasurement())
      {
         // actual code to measure goes here
      }
      
      // clean up code hat should be executed after each iteration
   }

   // clean up code after all iterations are complete
}

}
```

The code to benchmark is executed multiple times and measurements will be collected and summarized. XP gives you a handle for each iteration and allows you to control what code should run before and/or after the iteration. Only the code within the `iteration.StartMeasurement()` block counts towards the benchmark's result.

In BDN, the entire method annotated with `[Benchmark]` will be measured, and any setup and clean up code that should not count towards the measurement should be defined in other methods with appropriate annotations:

```c#
class MyBenchmarkTests
{
    
    [GlobalSetup]
    public void BeforeAll() {
        // any setup that should run once before all iterations  should go here
    }

   [GlobalCleanup]
    public void AfterAll() {
       // clean up code after all iterations are complete, e.g cleaning up resources created in the global setup code
    }

     [IterationSetup]
    public void BeforeEach() {
         // setup code that should be executed before each iteration goes here
    }

   [IterationCleanup]
    public void AfterEach() {
       // clean up code hat should be executed after each iteration
    }
   


    [Benchmark]
    public void SomeBenchmark()
    {
        // actual code to measure goes here
    }

}
```

#### Collecting memory usage statistics

- XP annotates each method for which GC allocation data should be collected with `[MeasureGCAllocs]`
- BDN allows you to collect memory usage and GC data by annotating the class with `[MemoryDiagnoser]`

- [ ] *Test cases added*

### Additional work necessary


test/PerformanceTests/Framework/TestService/Microsoft.Test.OData.TestService.csproj
```diff
@@ -13,12 +13,15 @@
     <AppDesignerFolder>Properties</AppDesignerFolder>
     <RootNamespace>Microsoft.Test.OData.TestService</RootNamespace>
     <AssemblyName>Microsoft.Test.OData.TestService</AssemblyName>
-    <TargetFrameworkVersion>v4.6</TargetFrameworkVersion>
```
xuzhg, line 16: shall we move to .NET Core

xuzhg (approved):

test/PerformanceTests/ServiceTests/app.config
```diff
@@ -0,0 +1,23 @@
...
+        <assemblyIdentity name="System.ValueTuple" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.0.3.0" newVersion="4.0.3.0" />
+      </dependentAssembly>
+      <dependentAssembly>
+        <assemblyIdentity name="System.Threading.Tasks.Extensions" publicKeyToken="cc7b13ffcd2ddd51" culture="neutral" />
+        <bindingRedirect oldVersion="0.0.0.0-4.2.0.0" newVersion="4.2.0.0" />
+      </dependentAssembly>
+    </assemblyBinding>
+  </runtime>
+</configuration>
```
gathogojr, line 23:
```suggestion
</configuration>

```

sln/OData.Tests.Performance.sln
```diff
@@ -1,22 +1,22 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 14
-VisualStudioVersion = 14.0.25420.1
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.30225.117
```
gathogojr, line 4: The changes here may have been inadvertent but we may want to retain Visual Studio 14 (VS 2015), or Visual Studio 15 (VS 2017) at least...

test/PerformanceTests/ComponentTests/Scenario/BinaryDataScaleTests.cs
```diff
@@ -23,39 +22,38 @@ public class BinaryDataScaleTests : WriteReadFeedTestBase
...
-                WriteStream.SetLength(0);
+            WriteFeed(WriteStream, ExchangeAttachmentModel, NumberOfEntries, Write4MB, TestEntitySet);
+        }
 
-                using (iteration.StartMeasurement())
-                {
-                    WriteFeed(WriteStream, ExchangeAttachmentModel, NumberOfEntries, writeEntry, TestEntitySet);
-                }
-            }
+        // TODO: used GlobalSetup here because the setup code appeared before the iterations in the old tests
```
gathogojr, line 40:
```suggestion
        // Used GlobalSetup here because the setup code appeared before the iterations in the old tests
```

test/PerformanceTests/ComponentTests/ODataReader/ODataReaderFeedTests.cs
```diff
@@ -6,90 +6,89 @@
...
+        
+        
+        public void SetupDataStream(string templateFile, int entryCount)
         {
-            ReadFeedTestAndMeasure("EntryIncludeSpatialWithExpansions.json", 100, false);
+            _stream = new MemoryStream(PayloadGenerator.GenerateFeed(templateFile, entryCount));
         }
 
-        private void ReadFeedTestAndMeasure(string templateFile, int entryCount, bool isFullValidation)
+        public void RunReadTest(bool isFullValidation)
```
gathogojr, line 86:
```suggestion
        public void RunReadFeedTest(bool isFullValidation)
```

PerformanceBuild.ps1
```diff
@@ -10,73 +10,43 @@ If("Debug","Release" -notcontains $Config)
...
 
-$Item = Get-ChildItem -Filter xunit.performance.analysis.exe -Recurse -path $PackagesPath
-$PerfAnalysisPath = $Item.FullName
 
-$Item = Get-ChildItem -Filter xunit.console.exe -Recurse -path $PackagesPath
-$XunitConsoleRunnerPath = $Item.FullName
+    $exeName = "$baseName.exe"
+    $exePath = "$projectRoot\bin\$config\$exeName";
+
+    Write-Host "*** Building Component tests $projectRoot***"
```
gathogojr, line 27:
```suggestion
    Write-Host "*** Building tests project $projectRoot***"
```
Given that the function is invoked for both "component" and "service" tests

gathogojr (approved):
Left some comments otherwise LGTM

test/PerformanceTests/ComponentTests/ODataWriter/ODataWriterFeedTests.cs
```diff
@@ -9,12 +9,12 @@ namespace Microsoft.OData.Performance
...
     using System.IO;
     using System.Linq;
-    using global::Xunit;
     using Microsoft.OData;
     using Microsoft.OData.Edm;
     using Microsoft.Spatial;
-    using Microsoft.Xunit.Performance;
+    using BenchmarkDotNet.Attributes;
 
+    [MemoryDiagnoser]
```
marabooy, line 17: I think we should include class comments for test cases which give an overview of what is happening.

test/PerformanceTests/ComponentTests/UriParser/UriParserTests.cs
```diff
@@ -33,11 +32,11 @@ public void ParseUri()
 
             int roundPerIteration = 1000;
 
+            // TODO: verify whether setting up these params prior to executing the method affects results
```
marabooy, line 35: Cleanup todos and leave a comment that will help a future soul here.

test/PerformanceTests/ServiceTests/Program.cs
```diff
@@ -0,0 +1,12 @@
+﻿using BenchmarkDotNet.Running;
```
marabooy, line 1: Move namespace import to inside the namespace block.

test/PerformanceTests/Framework/TestService/Microsoft.Test.OData.TestService.csproj
habbes: Moving the TestService might require a lot of rewriting. I suggest we do that in a later PR once the current tests have been fixed.

test/PerformanceTests/ComponentTests/ODataWriter/ODataWriterFeedTests.cs
habbes: Done

marabooy (approved):
LGTM

Issue #1974, opened 2021-01-07
ODataUriExtensions.BuildUri ignores $it literal


voronov-maxim:
I'm trying to use $filter query inside $expand query to filter navigation property collection. ODataUriExtensions.BuildUri does not produce $it to the uri.

### Assemblies affected
Microsoft.Odata.Core 7.8.1

### Reproduce steps
```cs
var order = new EdmEntityType("OdataToEntity", "Order");
order.AddKeys(order.AddStructuralProperty("Id", EdmPrimitiveTypeKind.String, false));
order.AddStructuralProperty("Product", EdmPrimitiveTypeKind.String, false);

var item = new EdmEntityType("OdataToEntity", "OrderItems");
item.AddKeys(item.AddStructuralProperty("Id", EdmPrimitiveTypeKind.String, false));

var itemProperty = order.AddUnidirectionalNavigation(new EdmNavigationPropertyInfo
{
    ContainsTarget = false,
    Name = "Items",
    Target = item,
    TargetMultiplicity = EdmMultiplicity.Many
});

var defaultContainer = new EdmEntityContainer("OdataToEntity", "DefaultContainer");
var orders = defaultContainer.AddEntitySet("Orders", order);
var items = defaultContainer.AddEntitySet("Items", item);
orders.AddNavigationTarget(itemProperty, items);

var model = new EdmModel();
model.AddElement(defaultContainer);

var options = new Dictionary<String, String> { { "$expand", "Items($filter=$it/Product eq 'unknown')" } };
var parser = new ODataQueryOptionParser(model, order, orders, options);
var odataUri = new ODataUri()
{
    Path = new ODataPath(),
    SelectAndExpand = parser.ParseSelectAndExpand()
};
Uri uri = odataUri.BuildUri(ODataUrlKeyDelimiter.Parentheses);
String data = Uri.UnescapeDataString(uri.OriginalString);
Debug.Assert(data == "?$expand=Items($filter=Product eq 'unknown')");
```

### Expected result
Uri must be "?$expand=Items($filter=$it/Product eq 'unknown')".

### Actual result
Uri is "?$expand=Items($filter=Product eq 'unknown')".

mikepizzo:
Probably related to #1906.

KenitoInc:
This is not a regression, but an existing bug.
Range Variables are handled outside the AST. So if an `expand clause` has an `explicit range variable`, we don't have a way of adding it to the Uri when we call `BuildUri` extension method.
[...]

PR #1991, opened 2021-01-27
Add Dollar it in BuildUri response

KenitoInc:
### Issues

*This pull request fixes issue #1974 .*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
+++ b/src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
@@ -496,9 +496,8 @@ namespace Microsoft.OData
                     if (rightBinaryNode != null && IsDifferentSource(filterClause, rightBinaryNode))
                     {
                         rightBinary = "$it/" + rightBinary;
+                        translatedNode = leftBinary + binaryOperator + rightBinary;
                     }
-
-                    translatedNode = leftBinary + binaryOperator + rightBinary;
                     break;
 
                 // $expand=Items($filter=$it/ID in ['1', '2', '3'])
@@ -514,9 +513,8 @@ namespace Microsoft.OData
                     if (leftInNode != null && IsDifferentSource(filterClause, leftInNode))
                     {
                         leftIn = "$it/" + leftIn;
+                        translatedNode = leftIn + inOperator + rightIn;
                     }
-
-                    translatedNode = leftIn + inOperator + rightIn;
                     break;
 
                 // $expand=Items($filter=endswith($it/Name, 'xyz')
@@ -544,8 +542,8 @@ namespace Microsoft.OData
                     if (leftParameterNode != null && IsDifferentSource(filterClause, leftParameterNode))
                     {
                         leftParameter = "$it/" + leftParameter;
+                        translatedNode = parameters.Length == 1 ? svfcSubtrings[0] + '(' + leftParameter + ')' : svfcSubtrings[0] + '(' + leftParameter + ',' + rightParameter + ')';
                     }
-                    translatedNode = parameters.Length == 1 ? svfcSubtrings[0] + '(' + leftParameter + ')' : svfcSubtrings[0] + '(' + leftParameter + ',' + rightParameter + ')';
                     break;
 
                 // $expand=Orders($filter=$it/Items/any(d:d/Quantity gt 100))
@@ -558,23 +556,18 @@ namespace Microsoft.OData
 
                     ResourceRangeVariable anyExpressionResourceRangeVariable = (filterClause.Expression as AnyNode).RangeVariables.Single(x => x.Name == "$it") as ResourceRangeVariable;
 
-                    ResourceRangeVariable anyFilterResourceRangeVariable = filterClause.RangeVariable as ResourceRangeVariable;
-                    bool isDifferentSource = anyExpressionResourceRangeVariable.NavigationSource != anyFilterResourceRangeVariable.NavigationSource;
-
-                    if(isDifferentSource)
+                    if(IsDifferentSource(filterClause, anyExpressionResourceRangeVariable))
                     {
                         leftAnyNodeSubstring = "/$it" + leftAnyNodeSubstring;
+                        translatedNode = leftAnyNodeSubstring + slashAny + rightAnyNodeSubstring;
                     }
-                    translatedNode = leftAnyNodeSubstring + slashAny + rightAnyNodeSubstring;
                     break;
 
+                // Products?$filter=ProductName
                 case QueryNodeKind.SingleValueOpenPropertyAccess:
                     ResourceRangeVariable svopaExpressionResourceRangeVariable = ((filterClause.Expression as SingleValueOpenPropertyAccessNode).Source as ResourceRangeVariableReferenceNode).RangeVariable;
 
-                    ResourceRangeVariable svopaFilterResourceRangeVariable = filterClause.RangeVariable as ResourceRangeVariable;
-                    bool isDifferentSource2 = svopaExpressionResourceRangeVariable.NavigationSource != svopaFilterResourceRangeVariable.NavigationSource;
-
-                    if (isDifferentSource2)
+                    if (IsDifferentSource(filterClause, svopaExpressionResourceRangeVariable))
                     {
                         translatedNode = "/$it" + translatedNode;
                     }
@@ -599,6 +592,22 @@ namespace Microsoft.OData
                 property.Source is SingleResourceNode propertySource)
             {
                 if(filterSource.NavigationSource != propertySource.NavigationSource)
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        private static bool IsDifferentSource(FilterClause filterClause, ResourceRangeVariable expressionRangeVariable)
+        {
+            Debug.Assert(filterClause != null, "filterClause != null");
+            Debug.Assert(expressionRangeVariable != null, "expressionRangeVariable != null");
+
+            if (filterClause.RangeVariable is ResourceRangeVariable filterSource)
+            {
+                if (filterSource.NavigationSource != expressionRangeVariable.NavigationSource)
                 {
                     return true;
                 }
--- a/src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
+++ b/src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
@@ -470,7 +470,88 @@ namespace Microsoft.OData
         internal String TranslateFilterClause(FilterClause filterClause)
         {
             Debug.Assert(filterClause != null, "filterClause != null");
-            return this.TranslateNode(filterClause.Expression);
+            string translatedNode = this.TranslateNode(filterClause.Expression);
+
+            switch (filterClause.Expression.Kind)
+            {
+                // $expand=Items($filter=$it/Product eq 'Some Product')
+                case QueryNodeKind.BinaryOperator:
+                    BinaryOperatorKind binaryNodeKind = (filterClause.Expression as BinaryOperatorNode).OperatorKind;
+
+                    string binaryOperator = this.BinaryOperatorNodeToString(binaryNodeKind);
+                    string[] binarySeparator = { binaryOperator };
+                    string[] binarySubstrings = translatedNode.Trim().Split(binarySeparator, StringSplitOptions.RemoveEmptyEntries);
+                    string leftBinary = binarySubstrings[0];
+                    string rightBinary = binarySubstrings[1];
+
+                    SingleValuePropertyAccessNode leftBinaryNode = ((filterClause.Expression as BinaryOperatorNode).Left as ConvertNode)?.Source as SingleValuePropertyAccessNode;
+                    SingleValuePropertyAccessNode rightBinaryNode = ((filterClause.Expression as BinaryOperatorNode).Right as ConvertNode)?.Source as SingleValuePropertyAccessNode;
+
+                    if (leftBinaryNode != null && IsDifferentSource(filterClause, leftBinaryNode))
+                    {
+                        leftBinary = "$it/" + leftBinary;
+                    }
+
+                    if (rightBinaryNode != null && IsDifferentSource(filterClause, rightBinaryNode))
+                    {
+                        rightBinary = "$it/" + rightBinary;
+                    }
+
+                    translatedNode = leftBinary + binaryOperator + rightBinary;
+                    break;
+
+                // $expand=Items($filter=$it/ID in ['1', '2', '3'])
+                case QueryNodeKind.In:
+                    string inOperator = ExpressionConstants.KeywordIn;
+                    string[] inSeparator = { inOperator };
+                    string[] inSubstrings = translatedNode.Trim().Split(inSeparator, StringSplitOptions.RemoveEmptyEntries);
+                    string leftIn = inSubstrings[0];
+                    string rightIn = inSubstrings[1];
+
+                    SingleValuePropertyAccessNode leftInNode = (filterClause.Expression as InNode).Left as SingleValuePropertyAccessNode;
+
+                    if (leftInNode != null && IsDifferentSource(filterClause, leftInNode))
+                    {
+                        leftIn = "$it/" + leftIn;
+                    }
+
+                    translatedNode = leftIn + inOperator + rightIn;
+                    break;
+
+                case QueryNodeKind.SingleValueFunctionCall:
+                    break;
+
+                case QueryNodeKind.Any:
+                    break;
+
+                case QueryNodeKind.SingleValueOpenPropertyAccess:
+                    break;
+
+                case QueryNodeKind.ParameterAlias:
+                    break;
+
+                default:
+                    break;
+            }
+
+            return translatedNode;
+        }
+
+        private static bool IsDifferentSource(FilterClause filterClause, SingleValuePropertyAccessNode property)
+        {
+            Debug.Assert(filterClause != null, "filterClause != null");
+            Debug.Assert(property != null, "property != null");
+
+            if (filterClause.RangeVariable is ResourceRangeVariable filterSource &&
+                property.Source is SingleResourceNode propertySource)
+            {
+                if(filterSource.NavigationSource != propertySource.NavigationSource)
+                {
+                    return true;
+                }
+            }
+
+            return false;
         }
 
         /// <summary>Translates a <see cref="OrderByClause"/> into a string.</summary>
--- a/src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
+++ b/src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
@@ -611,12 +611,37 @@ namespace Microsoft.OData
             Debug.Assert(filterClause != null, "filterClause != null");
             Debug.Assert(property != null, "property != null");
 
-            if (filterClause.RangeVariable is ResourceRangeVariable filterSource &&
-                property.Source is ResourceRangeVariableReferenceNode propertySource)
+            ResourceRangeVariable filterSource = filterClause.RangeVariable as ResourceRangeVariable;
+
+            if (filterSource !=null && property.Source is ResourceRangeVariableReferenceNode propertySource)
             {
                 if(filterSource.NavigationSource != propertySource.NavigationSource)
                 {
                     return true;
+                }
+            }
+
+            // EDGE case:
+            // In SingleComplexNodee, we traverse the Source properties until we get a ResourceRangeVariableReferenceNode.
+            else if (property.Source is SingleComplexNode complexPropertySource)
+            {
+                while(complexPropertySource.Source != null)
+                {
+                    if(complexPropertySource.Source is ResourceRangeVariableReferenceNode rangeVariableSource)
+                    {
+                        if (filterSource.NavigationSource != rangeVariableSource.NavigationSource)
+                        {
+                            return true;
+                        }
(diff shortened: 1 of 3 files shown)


src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +470,88 @@ internal String TranslateNode(QueryNode node)
...
+                    break;
+
+                case QueryNodeKind.Any:
+                    break;
+
+                case QueryNodeKind.SingleValueOpenPropertyAccess:
+                    break;
+
+                case QueryNodeKind.ParameterAlias:
+                    break;
```
xuzhg, line 531: If that's empty, we can remove them

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +470,88 @@ internal String TranslateNode(QueryNode node)
...
+                    SingleValuePropertyAccessNode rightBinaryNode = ((filterClause.Expression as BinaryOperatorNode).Right as ConvertNode)?.Source as SingleValuePropertyAccessNode;
+
+                    if (leftBinaryNode != null && IsDifferentSource(filterClause, leftBinaryNode))
+                    {
+                        leftBinary = "$it/" + leftBinary;
+                    }
+
+                    if (rightBinaryNode != null && IsDifferentSource(filterClause, rightBinaryNode))
+                    {
+                        rightBinary = "$it/" + rightBinary;
```
xuzhg, line 497: I am worry about the implementation here, does it only work for one level? I mean $it/Product?  Do we work:  $it/Address/City

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +470,88 @@ internal String TranslateNode(QueryNode node)
...
+                    }
+
+                    translatedNode = leftBinary + binaryOperator + rightBinary;
```
xuzhg, line 500: >translatedNode = leftBinary + binaryOperator + rightBinary; [](start = 20, length = 59)

If it's not a difference source, we should return the existing string "translateNode".

xuzhg (changes requested):
:clock1:

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
KenitoInc: It should support `$it/Address/City`
I will add tests to validate that.

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,197 @@ internal String TranslateNode(QueryNode node)
         internal String TranslateFilterClause(FilterClause filterClause)
         {
             Debug.Assert(filterClause != null, "filterClause != null");
-            return this.TranslateNode(filterClause.Expression);
+            string translatedNode = this.TranslateNode(filterClause.Expression);
+
+            switch (filterClause.Expression.Kind)
```
xuzhg, line 476: can we refactor the implementation here?
1) separate the whole into methods
2) Use the traverse, for example: https://github.com/OData/WebApi/blob/master/src/Microsoft.AspNet.OData.Shared/Query/Expressions/FilterBinder.cs#L184

KenitoInc: @xuzhg Done

marabooy (approved):
LGTM

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,240 @@ internal String TranslateNode(QueryNode node)
...
+        /// </summary>
+        /// <param name="node">node to bind.</param>
+        /// <param name="filterClauseRangeVariable">The <see cref="FilterClause"/> range variable.</param>
+        /// <returns>The translated string with $it binding.</returns>
+        private string BindAnyNode(AnyNode node, ResourceRangeVariable filterClauseRangeVariable)
+        {
+            string translatedNode = this.TranslateNode(node);
+            ResourceRangeVariableReferenceNode leftRangeVariableNode = GetResourceRangeVariableReferenceNode(node);
+
+            const string slashAny = "/any";
```
gathogojr, line 631: Maybe have this as a `SlashAny` constant in `ExpressionConstants` class?

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,240 @@ internal String TranslateNode(QueryNode node)
...
+
+            string binaryOperator = this.BinaryOperatorNodeToString(binaryNodeKind);
+            string[] binarySeparator = { binaryOperator };
+            string[] substrings = translatedNode.Trim().Split(binarySeparator, StringSplitOptions.RemoveEmptyEntries);
+            string leftBinary = substrings[0].Trim();
+            string rightBinary = substrings[1].Trim();
+
+            if (leftRangeVariableNode != null && IsDifferentSource(filterClauseRangeVariable, leftRangeVariableNode))
+            {
+                leftBinary = "$it/" + leftBinary;
```
gathogojr, line 584: Maybe have `ItSlash` as a constant in `ExpressionConstants` class?

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,240 @@ internal String TranslateNode(QueryNode node)
...
+        /// <returns>The translated string with $it binding.</returns>
+        private string BindSingleValueFunctionCallNode(SingleValueFunctionCallNode node, ResourceRangeVariable filterClauseRangeVariable)
+        {
+            string translatedNode = this.TranslateNode(node);
+            ResourceRangeVariableReferenceNode firstParameterRangeVariableNode = GetResourceRangeVariableReferenceNode(node);
+
+            char[] separators = { '(', ')' };
+            string[] subtrings = translatedNode.Trim().Split(separators, StringSplitOptions.RemoveEmptyEntries);
+            string withinBrackets = subtrings[1];
+            string[] parameters = withinBrackets.Trim().Split(',');
```
gathogojr, line 659: Are spaces possible in the parameters string, i.e. `(a, b, c)` as opposed to `(a,b,c)`, and should we bother trimming them?

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,240 @@ internal String TranslateNode(QueryNode node)
...
+            string leftParameter = parameters[0];
+            string rightParameter = parameters.Length == 2 ? parameters[1] : String.Empty;
```
gathogojr, line 661: We expect maximum two parameters?

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,240 @@ internal String TranslateNode(QueryNode node)
...
+        /// <returns>The translated string with $it binding.</returns>
+        private string BindNode(QueryNode node, ResourceRangeVariable filterClauseRangeVariable)
+        {
+            BinaryOperatorNode binaryNode = node as BinaryOperatorNode;
+            InNode inNode = node as InNode;
+            AnyNode anyNode = node as AnyNode;
+            SingleValueFunctionCallNode singleValueFunctionCallNode = node as SingleValueFunctionCallNode;
+            SingleValueOpenPropertyAccessNode singleValueOpenPropertyAccessNode = node as SingleValueOpenPropertyAccessNode;
+
+            if (binaryNode != null)
```
gathogojr, line 536: Alternatively,

```csharp
BinaryOperatorNode binaryNode;
// ... other variable declarations

if ((binaryNode = node as BinaryOperatorNode) != null)
{
   // ...
}
// else if ...
```
I however don't know if the cost of doing all 5 explicit conversions using the `as` operator regardless is significant. In this case it might not matter since the most common scenario is handled by the `else` block and all the explicit conversions will have been done either way.

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,240 @@ internal String TranslateNode(QueryNode node)
...
+            string[] inSeparator = { inOperator };
+            string[] substrings = translatedNode.Trim().Split(inSeparator, StringSplitOptions.RemoveEmptyEntries);
+            string leftIn = substrings[0];
+            string rightIn = substrings[1];
+
+            if (leftRangeVariableNode != null && IsDifferentSource(filterClauseRangeVariable, leftRangeVariableNode))
+            {
+                leftIn = "$it/" + leftIn;
+                translatedNode = leftIn + inOperator + rightIn;
+            }
```
gathogojr, line 616:
```suggestion
            }

```

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,240 @@ internal String TranslateNode(QueryNode node)
...
+            string[] anySeparator = { slashAny };
+            string[] substrings = translatedNode.Trim().Split(anySeparator, StringSplitOptions.RemoveEmptyEntries);
+            string leftAnyNodeSubstring = substrings[0];
+            string rightAnyNodeSubstring = substrings[1];
+
+            if (leftRangeVariableNode != null && IsDifferentSource(filterClauseRangeVariable, leftRangeVariableNode))
+            {
+                leftAnyNodeSubstring = "$it/" + leftAnyNodeSubstring;
+                translatedNode = leftAnyNodeSubstring + slashAny + rightAnyNodeSubstring;
+            }
```
gathogojr, line 641:
```suggestion
            }

```

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,240 @@ internal String TranslateNode(QueryNode node)
...
+            string withinBrackets = subtrings[1];
+            string[] parameters = withinBrackets.Trim().Split(',');
+            string leftParameter = parameters[0];
+            string rightParameter = parameters.Length == 2 ? parameters[1] : String.Empty;
+
+            if (firstParameterRangeVariableNode != null && IsDifferentSource(filterClauseRangeVariable, firstParameterRangeVariableNode))
+            {
+                leftParameter = "$it/" + leftParameter;
+                translatedNode = parameters.Length == 1 ? subtrings[0] + '(' + leftParameter + ')' : subtrings[0] + '(' + leftParameter + ',' + rightParameter + ')';
+            }
```
gathogojr, line 667:
```suggestion
            }

```

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,240 @@ internal String TranslateNode(QueryNode node)
...
+        /// <returns>The translated string with $it binding.</returns>
+        private string BindSingleValueOpenPropertyAccess(SingleValueOpenPropertyAccessNode node, ResourceRangeVariable filterClauseRangeVariable)
+        {
+            string translatedNode = this.TranslateNode(node);
+            ResourceRangeVariableReferenceNode rangeVariableNode = GetResourceRangeVariableReferenceNode(node);
+
+            if (rangeVariableNode != null && IsDifferentSource(filterClauseRangeVariable, rangeVariableNode))
+            {
+                translatedNode = "$it/" + translatedNode;
+            }
```
gathogojr, line 685:
```suggestion
            }

```

gathogojr (approved):

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
KenitoInc: We have `KeywordAny` in `ExpressionConstants` which represents `any`. I don't think we have more instances we will be using `/any`

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
KenitoInc: Good catch!

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
KenitoInc: Let me hear the opinions of others on this.

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
KenitoInc: Replaced it with `ExpressionConstants.It + ExpressionConstants.SymbolForwardSlash`

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
mikepizzo: I had the same thought -- we want to limit the number of casts.

Can we switch on QueryNode.Kind (or QueryNode.InternalKind)?

---
In reply to: [574336006](https://github.com/OData/odata.net/pull/1991#discussion_r574336006) [](ancestors = 574336006)

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
mikepizzo: For perf sensitive scenarios it would be better to define an ItSlash = ExpressionConstants.It + ExpressionConstants.SymbolForwardSlash, but a single call per range variable translation is not going to be a hotspot in constructing the Uri so I wouldn't worry about it here.

---
In reply to: [574337676](https://github.com/OData/odata.net/pull/1991#discussion_r574337676) [](ancestors = 574337676)

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,245 @@ internal String TranslateNode(QueryNode node)
...
+
+            BinaryOperatorKind binaryNodeKind = node.OperatorKind;
+
+            string binaryOperator = this.BinaryOperatorNodeToString(binaryNodeKind);
+            string[] binarySeparator = { binaryOperator };
+            string[] substrings = translatedNode.Trim().Split(binarySeparator, StringSplitOptions.RemoveEmptyEntries);
+            string leftBinary = substrings[0].Trim();
+            string rightBinary = substrings[1].Trim();
+
+            if (leftRangeVariableNode != null && IsDifferentSource(filterClauseRangeVariable, leftRangeVariableNode))
```
mikepizzo, line 582: >IsDifferentSource(filterClauseRangeVariable, leftRangeVariableNode) [](start = 49, length = 67)

So this logic is assuming that, if the range variable is a different source, the range variable must be $it, right? In which case, $this would work because it's the same type as the filterClauseRangeVariable so it will just fall through? How would this logic change when if we supported $root?

Minimally, we need to document these assumptions and add tests (if they don't already exist) for these different cases.

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,245 @@ internal String TranslateNode(QueryNode node)
...
+        private string BindInNode(InNode node, ResourceRangeVariable filterClauseRangeVariable)
+        {
+            string translatedNode = this.TranslateNode(node);
+            ResourceRangeVariableReferenceNode leftRangeVariableNode = GetResourceRangeVariableReferenceNode(node.Left);
+
+            string inOperator = ExpressionConstants.KeywordIn;
+            string[] inSeparator = { inOperator };
+            string[] substrings = translatedNode.Trim().Split(inSeparator, StringSplitOptions.RemoveEmptyEntries);
+            string leftIn = substrings[0];
+            string rightIn = substrings[1];
```
mikepizzo, line 610: This logic only needs to execute if the sources are different, so it would be better to move it within your if statement.

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
mikepizzo: As above, if this was a performance critical hotspot with many iterations it would be worth defining SlashAny, but in this case it's probably not worth it.

---
In reply to: [574293978](https://github.com/OData/odata.net/pull/1991#discussion_r574293978) [](ancestors = 574293978)

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,245 @@ internal String TranslateNode(QueryNode node)
...
+            string translatedNode = this.TranslateNode(node);
+            ResourceRangeVariableReferenceNode firstParameterRangeVariableNode = GetResourceRangeVariableReferenceNode(node);
+
+            char[] separators = { '(', ')' };
+            string[] subtrings = translatedNode.Trim().Split(separators, StringSplitOptions.RemoveEmptyEntries);
+            string withinBrackets = subtrings[1];
+            char[] parameterSeparators = { ',' };
+            string[] parameters = withinBrackets.Trim().Split(parameterSeparators, StringSplitOptions.RemoveEmptyEntries);
+            string leftParameter = parameters[0];
+            string rightParameter = parameters.Length == 2 ? parameters[1] : String.Empty;
```
mikepizzo, line 664: all of this logic should be moved inside the conditional and only executed in the case that we are inserting the $it.

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
```diff
@@ -470,7 +471,245 @@ internal String TranslateNode(QueryNode node)
...
+            Debug.Assert(rangeVariableReferenceNode != null, "rangeVariableReferenceNode != null");
+
+            if (filterClauseRangeVariable.NavigationSource != rangeVariableReferenceNode.NavigationSource)
+            {
+                return true;
+            }
+            else
+            {
+                return false;
+            }
```
mikepizzo, line 712: could simplify to:
   return filterClauseRangeVariable.NavigationSource != rangeVariableReferenceNode.NavigationSource;

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/SelectExpandBuilderTests.cs
```diff
@@ -652,6 +652,46 @@ public void SelectWithNestedExpandShouldWork()
...
+            Assert.Equal("http://gobbledygook/People?$expand=" + Uri.EscapeDataString("MyDog($filter=$it/PreviousAddresses/any(d:d/City eq 'Seattle'))"), actualUri.OriginalString);
+        }
+
+        [Fact]
+        public void ExpandWithDollarItInFilterFunctionCallShouldWork()
+        {
+            Uri queryUri = new Uri("People?$expand=MyDog($filter=endswith($it/Name,'abcd'))", UriKind.Relative);
+            Uri actualUri = UriBuilder(queryUri, ODataUrlKeyDelimiter.Parentheses, settings);
+            Assert.Equal("http://gobbledygook/People?$expand=" + Uri.EscapeDataString("MyDog($filter=endswith($it/Name,'abcd'))"), actualUri.OriginalString);
+        }
```
mikepizzo, line 693: do we have variants of each of these that don't use $it, and instead assume the range variable of the filter clause?

mikepizzo:
A few comments, otherwise looks good.

src/Microsoft.OData.Core/Uri/NodeToStringBuilder.cs
KenitoInc: Good catch!

xuzhg (approved):

PR #1992, opened 2021-01-31
Client Perf improvement

Sreejithpin:
### Issues
improvement for client performance
This creates a OdataTypeInfo which will cache some property and type infos to be reused

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
+++ b/src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
@@ -9,6 +9,7 @@ namespace Microsoft.OData.Client.Metadata
     #region Namespaces.
 
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.Linq;
@@ -16,8 +17,7 @@ namespace Microsoft.OData.Client.Metadata
     using System.Reflection;
     using Microsoft.OData.Metadata;
     using Microsoft.OData.Edm;
-    using c = Microsoft.OData.Client;
-    using System.Collections.Concurrent;
+    using c = Microsoft.OData.Client;    
 
     #endregion Namespaces.
 
--- a/src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
+++ b/src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
@@ -622,7 +622,7 @@ namespace Microsoft.OData.Client.Metadata
             ODataTypeInfo typeInfo = GetODataTypeInfo(t);
             string clientFieldName;
 
-            if (typeInfo.ClientDefinedNameDict.TryGetValue(serverDefinedName, out clientFieldName))
+            if (!typeInfo.ClientDefinedNameDict.TryGetValue(serverDefinedName, out clientFieldName))
             {
                 List<string> serverDefinedNames = serverDefinedName.Split(',').Select(name => name.Trim()).ToList();
                 List<string> clientMemberNames = new List<string>();
@@ -661,7 +661,7 @@ namespace Microsoft.OData.Client.Metadata
             ODataTypeInfo typeInfo = GetODataTypeInfo(t);
             PropertyInfo clientPropertyInfo;
 
-            if (typeInfo.ClientPropertyInfoDict.TryGetValue(serverDefinedName, out clientPropertyInfo))
+            if (!typeInfo.ClientPropertyInfoDict.TryGetValue(serverDefinedName, out clientPropertyInfo))
             {
                 PropertyInfo propertyInfo = t.GetProperty(serverDefinedName);
                 if (propertyInfo == null)
--- a/src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
+++ b/src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
@@ -17,8 +17,9 @@ namespace Microsoft.OData.Client.Metadata
     using Microsoft.OData.Metadata;
     using Microsoft.OData.Edm;
     using c = Microsoft.OData.Client;
+    using System.Collections.Concurrent;
 
-#endregion Namespaces.
+    #endregion Namespaces.
 
     /// <summary>
     /// Utility methods for client types.
@@ -27,6 +28,8 @@ namespace Microsoft.OData.Client.Metadata
     {
         /// <summary>A static empty PropertyInfo array.</summary>
         internal static readonly PropertyInfo[] EmptyPropertyInfoArray = new PropertyInfo[0];
+
+        internal static ConcurrentDictionary<Type, ODataTypeInfo> ODataTypeInfoCache { get; set; } = new ConcurrentDictionary<Type, ODataTypeInfo>();
 
         /// <summary>
         /// Enumeration for the kind of key
@@ -439,70 +442,80 @@ namespace Microsoft.OData.Client.Metadata
         /// <returns>Returns the list of key properties defined on <paramref name="type"/>; null if <paramref name="type"/> is complex.</returns>
         internal static PropertyInfo[] GetKeyPropertiesOnType(Type type, out bool hasProperties)
         {
-            if (CommonUtil.IsUnsupportedType(type))
-            {
-                throw new InvalidOperationException(c.Strings.ClientType_UnsupportedType(type));
-            }
+            ODataTypeInfo typeInfo = GetODataTypeInfo(type);
 
-            string typeName = type.ToString();
-            IEnumerable<object> customAttributes = type.GetCustomAttributes(true);
-            bool isEntity = customAttributes.OfType<EntityTypeAttribute>().Any();
-            KeyAttribute dataServiceKeyAttribute = customAttributes.OfType<KeyAttribute>().FirstOrDefault();
-            List<PropertyInfo> keyProperties = new List<PropertyInfo>();
-            PropertyInfo[] properties = ClientTypeUtil.GetPropertiesOnType(type, false /*declaredOnly*/).ToArray();
- 
-            hasProperties = properties.Length > 0;
-            KeyKind currentKeyKind = KeyKind.NotKey;
-            KeyKind newKeyKind = KeyKind.NotKey;
-            foreach (PropertyInfo propertyInfo in properties)
+            if (typeInfo.KeyProperties == null)
             {
-                if ((newKeyKind = ClientTypeUtil.IsKeyProperty(propertyInfo, dataServiceKeyAttribute)) != KeyKind.NotKey)
+                if (CommonUtil.IsUnsupportedType(type))
                 {
-                    if (newKeyKind > currentKeyKind)
+                    throw new InvalidOperationException(c.Strings.ClientType_UnsupportedType(type));
+                }
+
+                string typeName = type.ToString();
+                IEnumerable<object> customAttributes = type.GetCustomAttributes(true);
+                bool isEntity = customAttributes.OfType<EntityTypeAttribute>().Any();
+                KeyAttribute dataServiceKeyAttribute = customAttributes.OfType<KeyAttribute>().FirstOrDefault();
+                List<PropertyInfo> keyProperties = new List<PropertyInfo>();
+                PropertyInfo[] properties = ClientTypeUtil.GetPropertiesOnType(type, false /*declaredOnly*/).ToArray();
+
+                hasProperties = properties.Length > 0;
+                typeInfo.HasProperties = hasProperties;
+
+                KeyKind currentKeyKind = KeyKind.NotKey;
+                KeyKind newKeyKind = KeyKind.NotKey;
+                foreach (PropertyInfo propertyInfo in properties)
+                {
+                    if ((newKeyKind = ClientTypeUtil.IsKeyProperty(propertyInfo, dataServiceKeyAttribute)) != KeyKind.NotKey)
                     {
-                        keyProperties.Clear();
-                        currentKeyKind = newKeyKind;
-                        keyProperties.Add(propertyInfo);
+                        if (newKeyKind > currentKeyKind)
+                        {
+                            keyProperties.Clear();
+                            currentKeyKind = newKeyKind;
+                            keyProperties.Add(propertyInfo);
+                        }
+                        else if (newKeyKind == currentKeyKind)
+                        {
+                            keyProperties.Add(propertyInfo);
+                        }
                     }
-                    else if (newKeyKind == currentKeyKind)
+                }
+
+                Type keyPropertyDeclaringType = null;
+                foreach (PropertyInfo key in keyProperties)
+                {
+                    if (keyPropertyDeclaringType == null)
                     {
-                        keyProperties.Add(propertyInfo);
+                        keyPropertyDeclaringType = key.DeclaringType;
+                    }
+                    else if (keyPropertyDeclaringType != key.DeclaringType)
+                    {
+                        throw c.Error.InvalidOperation(c.Strings.ClientType_KeysOnDifferentDeclaredType(typeName));
+                    }
+
+                    if (!PrimitiveType.IsKnownType(key.PropertyType) && !(key.PropertyType.GetGenericTypeDefinition() == typeof(System.Nullable<>) && key.PropertyType.GetGenericArguments().First().IsEnum()))
+                    {
+                        throw c.Error.InvalidOperation(c.Strings.ClientType_KeysMustBeSimpleTypes(key.Name, typeName, key.PropertyType.FullName));
                     }
                 }
+
+                if (dataServiceKeyAttribute != null)
+                {
+                    if (newKeyKind == KeyKind.AttributedKey && keyProperties.Count != dataServiceKeyAttribute?.KeyNames.Count)
+                    {
+                        var m = (from string a in dataServiceKeyAttribute.KeyNames
+                                 where (from b in properties
+                                        where b.Name == a
+                                        select b).FirstOrDefault() == null
+                                 select a).First<string>();
+                        throw c.Error.InvalidOperation(c.Strings.ClientType_MissingProperty(typeName, m));
+                    }
+                }
+
+                typeInfo.KeyProperties = keyProperties.Count > 0 ? keyProperties.ToArray() : (isEntity ? ClientTypeUtil.EmptyPropertyInfoArray : null);
             }
 
-            Type keyPropertyDeclaringType = null;
-            foreach (PropertyInfo key in keyProperties)
-            {
-                if (keyPropertyDeclaringType == null)
-                {
-                    keyPropertyDeclaringType = key.DeclaringType;
-                }
-                else if (keyPropertyDeclaringType != key.DeclaringType)
-                {
-                    throw c.Error.InvalidOperation(c.Strings.ClientType_KeysOnDifferentDeclaredType(typeName));
-                }
-
-                if (!PrimitiveType.IsKnownType(key.PropertyType) && !(key.PropertyType.GetGenericTypeDefinition() == typeof(System.Nullable<>) && key.PropertyType.GetGenericArguments().First().IsEnum()))
-                {
-                    throw c.Error.InvalidOperation(c.Strings.ClientType_KeysMustBeSimpleTypes(key.Name, typeName, key.PropertyType.FullName));
-                }
-            }
-
-            if (dataServiceKeyAttribute != null)
-            {
-                if (newKeyKind == KeyKind.AttributedKey && keyProperties.Count != dataServiceKeyAttribute?.KeyNames.Count)
-                {
-                    var m = (from string a in dataServiceKeyAttribute.KeyNames
-                             where (from b in properties
-                                    where b.Name == a
-                                    select b).FirstOrDefault() == null
-                             select a).First<string>();
-                    throw c.Error.InvalidOperation(c.Strings.ClientType_MissingProperty(typeName, m));
-                }
-            }
-
-            return keyProperties.Count > 0 ? keyProperties.ToArray() : (isEntity ? ClientTypeUtil.EmptyPropertyInfoArray : null);
+            hasProperties = typeInfo.HasProperties.Value;
+            return typeInfo.KeyProperties;
         }
 
         /// <summary>Gets the type of the specified <paramref name="member"/>.</summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
```diff
@@ -0,0 +1,69 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ClientTypeUtil.cs" company="Microsoft">
```
xuzhg, line 2: update the copyright

src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
```diff
@@ -0,0 +1,69 @@
...
+namespace Microsoft.OData.Client.Metadata
+{
+    #region Namespaces.
+
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Linq;
+    using System.Linq.Expressions;
+    using System.Reflection;
```
xuzhg, line 16: sort and remove unused using

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -439,70 +442,80 @@ internal static PropertyInfo[] GetKeyPropertiesOnType(Type type)
         /// <returns>Returns the list of key properties defined on <paramref name="type"/>; null if <paramref name="type"/> is complex.</returns>
         internal static PropertyInfo[] GetKeyPropertiesOnType(Type type, out bool hasProperties)
         {
-            if (CommonUtil.IsUnsupportedType(type))
-            {
-                throw new InvalidOperationException(c.Strings.ClientType_UnsupportedType(type));
-            }
+            ODataTypeInfo typeInfo = GetODataTypeInfo(type);
```
xuzhg, line 445: GetODataTypeInfo() should return null if type is new

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -439,70 +442,80 @@ internal static PropertyInfo[] GetKeyPropertiesOnType(Type type)
...
-            bool isEntity = customAttributes.OfType<EntityTypeAttribute>().Any();
-            KeyAttribute dataServiceKeyAttribute = customAttributes.OfType<KeyAttribute>().FirstOrDefault();
-            List<PropertyInfo> keyProperties = new List<PropertyInfo>();
-            PropertyInfo[] properties = ClientTypeUtil.GetPropertiesOnType(type, false /*declaredOnly*/).ToArray();
- 
-            hasProperties = properties.Length > 0;
-            KeyKind currentKeyKind = KeyKind.NotKey;
-            KeyKind newKeyKind = KeyKind.NotKey;
-            foreach (PropertyInfo propertyInfo in properties)
+            if (typeInfo.KeyProperties == null)
```
xuzhg, line 447: if (typeInfo != null)
{
    return ....;
}

Keep the existing code unchanged.

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -439,70 +442,80 @@ internal static PropertyInfo[] GetKeyPropertiesOnType(Type type)
...
+                                 select a).First<string>();
+                        throw c.Error.InvalidOperation(c.Strings.ClientType_MissingProperty(typeName, m));
+                    }
                 }
+
+                typeInfo.KeyProperties = keyProperties.Count > 0 ? keyProperties.ToArray() : (isEntity ? ClientTypeUtil.EmptyPropertyInfoArray : null);
             }
 
-            return keyProperties.Count > 0 ? keyProperties.ToArray() : (isEntity ? ClientTypeUtil.EmptyPropertyInfoArray : null);
+            hasProperties = typeInfo.HasProperties.Value;
```
xuzhg, line 517: hasProperties is a local variable or the instance variable?
What's the meaning here to set its value?

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -439,70 +442,80 @@ internal static PropertyInfo[] GetKeyPropertiesOnType(Type type)
...
+            return typeInfo.KeyProperties;
```
xuzhg, line 518: Before return, Let's create the ODataTypeInfo and save into the concurrentdict?

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -28,6 +29,8 @@ internal static class ClientTypeUtil
         /// <summary>A static empty PropertyInfo array.</summary>
         internal static readonly PropertyInfo[] EmptyPropertyInfoArray = new PropertyInfo[0];
 
+        internal static ConcurrentDictionary<Type, ODataTypeInfo> ODataTypeInfoCache { get; set; } = new ConcurrentDictionary<Type, ODataTypeInfo>();
```
xuzhg, line 32: 1) one ConcurrentDictionary here
2) two ConcurrentDictionarys in each ODataTypeInfo?
It seems too much?

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -677,6 +728,19 @@ internal static MethodInfo GetClientMethod(Type t, string serverDefinedName, Typ
             return methodInfo;
         }
 
+
+        private static ODataTypeInfo GetODataTypeInfo(Type type)
```
xuzhg, line 732: Get...() should return "null" if the type is not in the dict?

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -677,6 +728,19 @@ internal static MethodInfo GetClientMethod(Type t, string serverDefinedName, Typ
             return methodInfo;
         }
 
+
+        private static ODataTypeInfo GetODataTypeInfo(Type type)
+        {
+            ODataTypeInfo typeInfo;
+            if (!ODataTypeInfoCache.TryGetValue(type, out typeInfo))
```
xuzhg, line 735: Add a new method "SetODataTypeInfo(Type type, ODataTypeInfo)" ?

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -439,70 +442,80 @@ internal static PropertyInfo[] GetKeyPropertiesOnType(Type type)
...
-                else if (keyPropertyDeclaringType != key.DeclaringType)
+                string typeName = type.ToString();
+                IEnumerable<object> customAttributes = type.GetCustomAttributes(true);
+                bool isEntity = customAttributes.OfType<EntityTypeAttribute>().Any();
+                KeyAttribute dataServiceKeyAttribute = customAttributes.OfType<KeyAttribute>().FirstOrDefault();
+                List<PropertyInfo> keyProperties = new List<PropertyInfo>();
+                PropertyInfo[] properties = ClientTypeUtil.GetPropertiesOnType(type, false /*declaredOnly*/).ToArray();
+
+                hasProperties = properties.Length > 0;
+                typeInfo.HasProperties = hasProperties;
```
xuzhg, line 462: here: typeInfo.HasProperties = hasProperties;
line 517: hasProperties = typeInfo.HasProperties.Value;

I can't understand it. Between line 462 & 517, can you let me know where do you update the "HasProperties"?

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
Sreejithpin: Get is intended to add to dict if a type is not present and then return it

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
Sreejithpin: Get is intended to add to dict if a type is not present and then return it

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
Sreejithpin: typeinfo will be not null

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
Sreejithpin: if in 447 hasproperties is set , 517 ensures its set back to hasproperties, sicne its out param

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
Sreejithpin: out param

KenitoInc (changes requested):
@Sreejithpin Add a description to the PR so that we can understand what your are trying to fix.

mikepizzo:
+1

---
In reply to: [652218599](https://github.com/OData/odata.net/pull/1992#issuecomment-652218599) [](http://example.com/codeflow?ancestors=652218599,652218599)

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
mikepizzo, line 445: >GetODataTypeInfo(type); [](http://example.com/codeflow?start=37&length=23)

Instead of implementing this method, just call GetOrAdd on the concurrent dictionary which does the same thing.

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
mikepizzo: Just use GetOrAdd on the concurrent dictionary.

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -617,23 +658,33 @@ internal static string GetClientFieldName(Type t, string serverDefinedName)
         /// <returns>Client PropertyInfo, or null if the method is not found.</returns>
         internal static PropertyInfo GetClientPropertyInfo(Type t, string serverDefinedName, UndeclaredPropertyBehavior undeclaredPropertyBehavior)
```
mikepizzo, line 659: >GetClientPropertyInfo [](http://example.com/codeflow?start=37&length=21)

Consider instead:
1) Having a Properties property (of type dictionary<string,propertyInfo>) on ODataTypeInfo.
2) The first time the Properties property is requested, generate it and then store it on the object. Since the only thread that writes to this will be the one that creates it, and it won't be set for other threads to read until it is fully populated, it does not need to be a concurrent dictionary.
3) this method just becomes:
ODataTypeInfo typeInfo = GetODataTypeInfo(t);
PropertyInfo propertyInfo = typeInfo.Properties[serverDefinedName];  
if (clientPropertyInfo == null && (undeclaredPropertyBehavior == UndeclaredPropertyBehavior.ThrowException))
    {
        throw c.Error.InvalidOperation(c.Strings.ClientType_MissingProperty(t.ToString(), serverDefinedName));
     }
 return propertyInfo;

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -617,23 +658,33 @@ internal static string GetClientFieldName(Type t, string serverDefinedName)
         /// <returns>Client PropertyInfo, or null if the method is not found.</returns>
```
mikepizzo, line 658: >or null if the method is not found [](http://example.com/codeflow?start=42&length=34)

actually, whether this returns null or throws an exception depends on undeclaredPropertyBehavior, right?

mikepizzo (changes requested):
:clock1:

mikepizzo:
>         internal static IEnumerable<PropertyInfo> GetPropertiesOnType(Type type, bool declaredOnly)

why is the set of properties for a type not cached?  i would think it would form the basis for what is cached on the ODataTypeInfo, and things like KeyProperties would be determined from the cached set of properties (and perhaps themselves cached as a further optimization as necessary).

---
Refers to: src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs:362 in cd23af5. [](commit_id = cd23af5e2a1da0e4b3d76df5c13131141a8990e6, deletion_comment = False)

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -556,27 +569,46 @@ internal static string GetServerDefinedName(MemberInfo memberInfo)
...
-            OriginalNameAttribute originalNameAttribute = (OriginalNameAttribute)type.GetCustomAttributes(typeof(OriginalNameAttribute), false).SingleOrDefault();
-            if (originalNameAttribute != null)
+            ODataTypeInfo typeInfo = GetODataTypeInfo(type);
+
+            if (typeInfo.ServerDefinedTypeFullName == null)
             {
-                return type.Namespace + "." + originalNameAttribute.OriginalName;
+                OriginalNameAttribute originalNameAttribute = (OriginalNameAttribute)type.GetCustomAttributes(typeof(OriginalNameAttribute), false).SingleOrDefault();
+                if (originalNameAttribute != null)
+                {
```
mikepizzo, line 602: Maybe call GetServiceDefinedTypeName and use it to compute ServerDefinedTypeFullName?

src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <summary>
+        /// Sertver defined type name
+        /// </summary>
+        public string ServerDefinedTypeName { get; set; }
+
+
+        /// <summary>
+        /// Sertver defined type full name
+        /// </summary>
+        public string ServerDefinedTypeFullName { get; set; }
```
mikepizzo, line 49: move logic to this type for computing and caching properties, keyproperties, hasproperties, server defined type name/full name, and mapping to client defined keys/properties.  Then you can make the dictionaries private and just call public properties on this internal class.

src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
```diff
@@ -0,0 +1,61 @@
...
+
+        /// <summary>
+        /// Concurrent Dictionary cache to save ClientDefinedName for the type and serverDefinedName
+        /// </summary>
+        public ConcurrentDictionary<string, string> ClientDefinedNameDict { get; set; }
+
+        /// <summary>
+        /// Concurrent Dictionary cache to save Propertyinfo for the type and serverDefinedName
+        /// </summary>
+        public ConcurrentDictionary<string, PropertyInfo> ClientPropertyInfoDict { get; set; }
```
mikepizzo, line 59: >public ConcurrentDictionary<string, PropertyInfo> ClientPropertyInfoDict { get; set; } [](http://example.com/codeflow?start=8&length=86)

just expose a public readonlydictionary of string, propertyInfo.  The first time get is called, this is created, populated for all properties (you are reading them all anyway), and set.

src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
```diff
@@ -0,0 +1,61 @@
...
+        /// <summary>
+        /// Sertver defined type full name
+        /// </summary>
+        public string ServerDefinedTypeFullName { get; set; }
+
+        /// <summary>
+        /// Concurrent Dictionary cache to save ClientDefinedName for the type and serverDefinedName
+        /// </summary>
+        public ConcurrentDictionary<string, string> ClientDefinedNameDict { get; set; }
+
```
mikepizzo, line 55: Expose a public method to get the client defined name for a string that uses an internal dictionary populated on first use.

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -587,25 +619,34 @@ internal static string GetServerDefinedTypeFullName(Type type)
         /// <returns>Client clr name.</returns>
         internal static string GetClientFieldName(Type t, string serverDefinedName)
         {
-            List<string> serverDefinedNames = serverDefinedName.Split(',').Select(name => name.Trim()).ToList();
-            List<string> clientMemberNames = new List<string>();
-            foreach (var serverSideName in serverDefinedNames)
+            ODataTypeInfo typeInfo = GetODataTypeInfo(t);
+            string clientFieldName;
+
+            if (!typeInfo.ClientDefinedNameDict.TryGetValue(serverDefinedName, out clientFieldName))
```
mikepizzo, line 625: >ClientDefinedNameDict [](http://example.com/codeflow?start=26&length=21)

note that serverDefinedName is a comma-separated list of property names.  So your cache is caching that set of field names, not the indidual fields, and if the code in one point asks for type a, in another point asks for type b, in another point asks for a, b, and in another point asks for b,a, those are all distinct computed entries in your cache where you really should just have 2.

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
mikepizzo, line 732: >private [](http://example.com/codeflow?start=8&length=7)

private methods should follow internal methods.

mikepizzo:
I like the idea of introducing a cache of ODataTypeInfos, but I think the code could be factored better.

I would keep the logic for computing things like the set of properties, key properties, has properties, etc., private to the ODataType info (i.e., move your logic for computing those into the ODataTypeInfo class) and have them computed on first use. Then have ClientTypeUtils get the item from the cache (creating it if it doesn't already exist) and simply request the information from the type.

Also, note that GetClientFieldName takes a comma-separated list of server defined names, which is not the best cache key.  I would refactor what is cached *and*, since this is an internal method, look at where it's called and see if there might be a more efficient way to call/return the information than a comma separated string.

There is also (still) a lot of duplicated logic that should be able to be consolidated.

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
Sreejithpin: at this point it should be in concurrent dict since the getodatatypeinfo takes care of that

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -556,27 +453,46 @@ internal static string GetServerDefinedName(MemberInfo memberInfo)
...
+                OriginalNameAttribute originalNameAttribute = (OriginalNameAttribute)type.GetCustomAttributes(typeof(OriginalNameAttribute), false).SingleOrDefault();
+                if (originalNameAttribute != null)
+                {
+                    typeInfo.ServerDefinedTypeName = originalNameAttribute.OriginalName;
+                }
+                else
+                {
+                    typeInfo.ServerDefinedTypeName = type.Name;
+                }
             }
```
mikepizzo, line 469: Why is this logic not in ODataTypeInfo?

src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
```diff
@@ -0,0 +1,292 @@
...
+                //// if PropertyType == IntPtr or UIntPtr
+
+                //// Properties overriding abstract or virtual properties on a base type
+                //// are also ignored (because they are part of the base type declaration
+                //// and not of the derived type).
+
+                OriginalNameAttribute originalNameAttribute = (OriginalNameAttribute)propertyInfo.GetCustomAttributes(typeof(OriginalNameAttribute), false).SingleOrDefault();
+                string serverDefinedName = originalNameAttribute != null? originalNameAttribute.OriginalName: propertyInfo.Name;
+
+                _propertyInfoDict[serverDefinedName] = propertyInfo;
```
mikepizzo, line 155: > _propertyInfoDict[serverDefinedName] = propertyInfo; [](http://example.com/codeflow?start=15&length=53)

There is a chance that, in a race condition, GetAllProperties could be called simultaneously on two separate threads, in which case this line could be problemmatic.  To be thread safe, instead of adding to the static dictionary, create an instance of a dictionary in GetAllProperties and do a single set of _propertyInfoDict to this dictionary at the end of GetAllProperties.

src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
```diff
@@ -0,0 +1,292 @@
...
+            {
+                throw new InvalidOperationException(c.Strings.ClientType_UnsupportedType(type));
+            }
+
+            string typeName = type.ToString();
+            IEnumerable<object> customAttributes = type.GetCustomAttributes(true);
+            bool isEntity = customAttributes.OfType<EntityTypeAttribute>().Any();
+            KeyAttribute dataServiceKeyAttribute = customAttributes.OfType<KeyAttribute>().FirstOrDefault();
+            List<PropertyInfo> keyProperties = new List<PropertyInfo>();
+            PropertyInfo[] properties = Properties.ToArray();
```
mikepizzo, line 195: >.ToArray() [](http://example.com/codeflow?start=50&length=10)

I don't think this is buying us anything

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -587,22 +503,32 @@ internal static string GetServerDefinedTypeFullName(Type type)
...
+                        throw c.Error.InvalidOperation(c.Strings.ClientType_MissingProperty(t.ToString(), serverSideName));
+                    }
+
+                    memberInfoName = memberInfo.Name;
+                    typeInfo.ServerSideNameDict[serverSideName] = memberInfoName;
                 }
 
-                clientMemberNames.Add(memberInfo.Name);
+                clientMemberNames.Add(memberInfoName);
             }
```
mikepizzo, line 533: implement a method "GetClientFieldName" in ODataTypeInfo that takes a type and a single server defined name and returns the client name.  Keep the logic for splitting the string, calling the GetClientFieldName method, and joining the results in this method.

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -677,6 +545,7 @@ internal static MethodInfo GetClientMethod(Type t, string serverDefinedName, Typ
             return methodInfo;
         }
 
+
         /// <summary>
```
mikepizzo, line 549: remove added blank line

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -556,27 +453,20 @@ internal static string GetServerDefinedName(MemberInfo memberInfo)
...
-                return originalNameAttribute.OriginalName;
-            }
+            ODataTypeInfo typeInfo = GetODataTypeInfo(type);           
 
-            return type.Name;
+            return typeInfo.ServerDefinedTypeName;
         }
 
+
         /// <summary>Gets the full server defined type name in <see cref="OriginalNameAttribute"/> of the specified <paramref name="type"/>.</summary>
```
mikepizzo, line 462: remove added blank line

src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
```diff
@@ -0,0 +1,382 @@
...
+                }
+
+                memberInfoName = memberInfo.Name;
+                ServerSideNameDict[serverSideName] = memberInfoName;
+            }
+
+            return memberInfoName;
+        }
+
+        internal PropertyInfo GetClientPropertyInfo(string serverDefinedName, UndeclaredPropertyBehavior undeclaredPropertyBehavior)
```
mikepizzo, line 366: >internal [](http://example.com/codeflow?start=8&length=8)

should be public method on internal class

src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
```diff
@@ -0,0 +1,382 @@
...
+                //// also ignored
+                //// if PropertyType.IsPointer (like byte*)
+                //// if PropertyType.IsArray except for byte[] and char[]
+                //// if PropertyType == IntPtr or UIntPtr
+
+                //// Properties overriding abstract or virtual properties on a base type
+                //// are also ignored (because they are part of the base type declaration
+                //// and not of the derived type).
+
+                propertyInfoDict[ServerDefinedTypeName] = propertyInfo;
```
mikepizzo, line 204: >ServerDefinedTypeName [](http://example.com/codeflow?start=33&length=21)

shouldn't this be the server defined name for the propertyinfo, not for the type?

mikepizzo (approved):
:shipit:

xuzhg (approved):

KenitoInc (approved):

PR #1993, opened 2021-02-01
Perf fixes for ODL- Linq fixes

Sreejithpin:
Few perf fixes

### Issues
Perf fixes for ODL- Linq fixes

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalIdMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalIdMetadataBuilder.cs
@@ -273,11 +273,12 @@ namespace Microsoft.OData.Evaluation
         private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
         {
             ODataPath path = odataUri.Path;
-            List<ODataPathSegment> segments = path.ToList();
+            IList<ODataPathSegment> segments = path.Segments;
             int lastIndex = segments.Count - 1;
             ODataPathSegment lastSegment = segments[lastIndex];
             while (!(lastSegment is NavigationPropertySegment) && !(lastSegment is OperationSegment))
             {
+                segments.RemoveAt(lastIndex);
                 lastSegment = segments[--lastIndex];
             }
 
@@ -287,6 +288,7 @@ namespace Microsoft.OData.Evaluation
                 IEdmStructuredType owningType = previousSegment.TargetEdmType as IEdmStructuredType;
                 if (owningType != null && owningType.FindProperty(lastSegment.Identifier) != null)
                 {
+                    segments.RemoveAt(lastIndex);
                     lastSegment = segments[--lastIndex];
                 }
                 else
@@ -296,7 +298,7 @@ namespace Microsoft.OData.Evaluation
             }
 
             // trim all the un-needed segments 
-            segments = segments.GetRange(0, lastIndex);
+            segments.RemoveAt(lastIndex);
 
             // append each segment to base uri
             Uri uri = baseUri;
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalIdMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalIdMetadataBuilder.cs
@@ -278,7 +278,6 @@ namespace Microsoft.OData.Evaluation
             ODataPathSegment lastSegment = segments[lastIndex];
             while (!(lastSegment is NavigationPropertySegment) && !(lastSegment is OperationSegment))
             {
-                segments.RemoveAt(lastIndex);
                 lastSegment = segments[--lastIndex];
             }
 
@@ -288,7 +287,6 @@ namespace Microsoft.OData.Evaluation
                 IEdmStructuredType owningType = previousSegment.TargetEdmType as IEdmStructuredType;
                 if (owningType != null && owningType.FindProperty(lastSegment.Identifier) != null)
                 {
-                    segments.RemoveAt(lastIndex);
                     lastSegment = segments[--lastIndex];
                 }
                 else
@@ -298,10 +296,7 @@ namespace Microsoft.OData.Evaluation
             }
 
             // trim all the un-needed segments 
-            if (lastIndex >= 0)
-            {
-                segments.RemoveAt(lastIndex);
-            }
+            segments = segments.GetRange(0, lastIndex);
 
             // append each segment to base uri
             Uri uri = baseUri;
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalIdMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalIdMetadataBuilder.cs
@@ -273,7 +273,7 @@ namespace Microsoft.OData.Evaluation
         private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
         {
             ODataPath path = odataUri.Path;
-            IList<ODataPathSegment> segments = path.Segments;
+            List<ODataPathSegment> segments = path.ToList();
             int lastIndex = segments.Count - 1;
             ODataPathSegment lastSegment = segments[lastIndex];
             while (!(lastSegment is NavigationPropertySegment) && !(lastSegment is OperationSegment))
@@ -298,7 +298,10 @@ namespace Microsoft.OData.Evaluation
             }
 
             // trim all the un-needed segments 
-            segments.RemoveAt(lastIndex);
+            if (lastIndex >= 0)
+            {
+                segments.RemoveAt(lastIndex);
+            }
 
             // append each segment to base uri
             Uri uri = baseUri;
--- a/src/Microsoft.OData.Core/ODataUtils.cs
+++ b/src/Microsoft.OData.Core/ODataUtils.cs
@@ -206,11 +206,11 @@ namespace Microsoft.OData
 
             ODataServiceDocument serviceDocument = new ODataServiceDocument();
             serviceDocument.EntitySets = model.EntityContainer.EntitySets()
-                .Select(entitySet => new ODataEntitySetInfo() { Name = entitySet.Name, Title = entitySet.Name, Url = new Uri(entitySet.Name, UriKind.RelativeOrAbsolute) }).ToList();
+                .Select(entitySet => new ODataEntitySetInfo() { Name = entitySet.Name, Title = entitySet.Name, Url = new Uri(entitySet.Name, UriKind.RelativeOrAbsolute) });
             serviceDocument.Singletons = model.EntityContainer.Singletons()
-                .Select(singleton => new ODataSingletonInfo() { Name = singleton.Name, Title = singleton.Name, Url = new Uri(singleton.Name, UriKind.RelativeOrAbsolute) }).ToList();
+                .Select(singleton => new ODataSingletonInfo() { Name = singleton.Name, Title = singleton.Name, Url = new Uri(singleton.Name, UriKind.RelativeOrAbsolute) });
             serviceDocument.FunctionImports = model.EntityContainer.OperationImports().OfType<IEdmFunctionImport>().Where(functionImport => functionImport.IncludeInServiceDocument && !functionImport.Function.Parameters.Any())
-                .Select(functionImport => new ODataFunctionImportInfo() { Name = functionImport.Name, Title = functionImport.Name, Url = new Uri(functionImport.Name, UriKind.RelativeOrAbsolute) }).ToList();
+                .Select(functionImport => new ODataFunctionImportInfo() { Name = functionImport.Name, Title = functionImport.Name, Url = new Uri(functionImport.Name, UriKind.RelativeOrAbsolute) });
 
             return serviceDocument;
         }
--- a/src/Microsoft.OData.Core/ODataUtils.cs
+++ b/src/Microsoft.OData.Core/ODataUtils.cs
@@ -204,13 +204,42 @@ namespace Microsoft.OData
                 throw new ODataException(Strings.ODataUtils_ModelDoesNotHaveContainer);
             }
 
+            IList<ODataEntitySetInfo> lstEntitySets = new List<ODataEntitySetInfo>();
+            IList<ODataSingletonInfo> lstSingletons = new List<ODataSingletonInfo>();
+            IList<ODataFunctionImportInfo> functionImports = new List<ODataFunctionImportInfo>();
+
+            foreach (var element in model.EntityContainer.Elements)
+            {
+                IEdmEntitySet entitySet = element as IEdmEntitySet;
+
+                if (entitySet != null)
+                {
+                    lstEntitySets.Add(new ODataEntitySetInfo() { Name = entitySet.Name, Title = entitySet.Name, Url = new Uri(entitySet.Name, UriKind.RelativeOrAbsolute) });
+                }
+                else
+                {
+                    IEdmSingleton singleton = element as IEdmSingleton;
+
+                    if (singleton != null)
+                    {
+                        lstSingletons.Add(new ODataSingletonInfo() { Name = singleton.Name, Title = singleton.Name, Url = new Uri(singleton.Name, UriKind.RelativeOrAbsolute) });
+                    }
+                    else
+                    {
+                        IEdmFunctionImport functionImport = element as IEdmFunctionImport;
+
+                        if(functionImport != null && functionImport.IncludeInServiceDocument && !functionImport.Function.Parameters.Any())
+                        {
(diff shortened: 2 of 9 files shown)


src/Microsoft.OData.Core/ODataUtils.cs
```diff
@@ -206,11 +206,11 @@ public static ODataServiceDocument GenerateServiceDocument(this IEdmModel model)
 
             ODataServiceDocument serviceDocument = new ODataServiceDocument();
             serviceDocument.EntitySets = model.EntityContainer.EntitySets()
-                .Select(entitySet => new ODataEntitySetInfo() { Name = entitySet.Name, Title = entitySet.Name, Url = new Uri(entitySet.Name, UriKind.RelativeOrAbsolute) }).ToList();
```
xuzhg, line 209: What's the standard about the usage of "ToList()" after a Linq?

Do you have such investigation? Would you please share detail in to PR description?

xuzhg (commented):
Neither the title nor the description can give me the context, reason, about this PR. 
Would you please share detail in the PR description?

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -849,15 +862,15 @@ private static SelectedPropertiesNode CombineSelectAndExpandResult(IEnumerable<s
                 children = new Dictionary<string, SelectedPropertiesNode>(StringComparer.Ordinal)
             };
 
-            foreach (string selectItem in rawSelect)
+            for (int i=0; i<rawSelect.Count; i++)
```
habbes, line 865:
```suggestion
            for (int i = 0; i < rawSelect.Count; i++)

```

src/Microsoft.OData.Core/ODataUtils.cs
```diff
@@ -206,11 +206,11 @@ public static ODataServiceDocument GenerateServiceDocument(this IEdmModel model)
...
             serviceDocument.Singletons = model.EntityContainer.Singletons()
-                .Select(singleton => new ODataSingletonInfo() { Name = singleton.Name, Title = singleton.Name, Url = new Uri(singleton.Name, UriKind.RelativeOrAbsolute) }).ToList();
+                .Select(singleton => new ODataSingletonInfo() { Name = singleton.Name, Title = singleton.Name, Url = new Uri(singleton.Name, UriKind.RelativeOrAbsolute) });
             serviceDocument.FunctionImports = model.EntityContainer.OperationImports().OfType<IEdmFunctionImport>().Where(functionImport => functionImport.IncludeInServiceDocument && !functionImport.Function.Parameters.Any())
-                .Select(functionImport => new ODataFunctionImportInfo() { Name = functionImport.Name, Title = functionImport.Name, Url = new Uri(functionImport.Name, UriKind.RelativeOrAbsolute) }).ToList();
+                .Select(functionImport => new ODataFunctionImportInfo() { Name = functionImport.Name, Title = functionImport.Name, Url = new Uri(functionImport.Name, UriKind.RelativeOrAbsolute) });
```
habbes, line 213: If you remove the `ToList()`, does that mean each time the caller calls `serviceDocument.EntitySets` or `serviceDocument.Singletons` it will re-evaluate the query? Wouldn't that be a penalty if those properties are accessed frequently? Or is the rationale to make it the responsibility of the caller to call `ToList()` on the property if they intend to iterate over them multiple times?

src/Microsoft.OData.Core/Evaluation/ODataConventionalIdMetadataBuilder.cs
```diff
@@ -296,7 +298,10 @@ private Uri GetContainingEntitySetUri(Uri baseUri, ODataUri odataUri)
             }
 
             // trim all the un-needed segments 
-            segments = segments.GetRange(0, lastIndex);
+            if (lastIndex >= 0)
+            {
+                segments.RemoveAt(lastIndex);
+            }
```
habbes, line 304: Would calling `segments.RemoveRange(lastIndex, countToRemove)` to trim the un-needed segments instead of calling `RemoveAt` once at a time be a better alternative?

src/Microsoft.OData.Core/ODataUtils.cs
```diff
@@ -206,11 +206,11 @@ public static ODataServiceDocument GenerateServiceDocument(this IEdmModel model)
 
             ODataServiceDocument serviceDocument = new ODataServiceDocument();
             serviceDocument.EntitySets = model.EntityContainer.EntitySets()
-                .Select(entitySet => new ODataEntitySetInfo() { Name = entitySet.Name, Title = entitySet.Name, Url = new Uri(entitySet.Name, UriKind.RelativeOrAbsolute) }).ToList();
+                .Select(entitySet => new ODataEntitySetInfo() { Name = entitySet.Name, Title = entitySet.Name, Url = new Uri(entitySet.Name, UriKind.RelativeOrAbsolute) });
```
mikepizzo, line 209: Each call to .EntitySets(), .Singletons(), .OperationImports, etc. calls through EntityContainer.Elements(),OfType<> for each element type.  Better would be to remove the LINQ all together -- build 3 lists and iterate through model.EntityContainer.Elements(), adding the appropriate element to each list.

src/Microsoft.OData.Core/ODataUtils.cs
mikepizzo: That's the trade-off.  If we only iterate once (or not at all) than the enumerable is more performant.  But, if we may iterate multiple times then it's worth building the list.  Lazyloading the lists falls somewhere in between -- you still have the benefit of the list for multiple iterations, but not the overhead if nobody ever asks for it.

@Sreejith -- does AGS ever call this method? If so, is it called just once or multiple times? How does WebAPI use it?

---
In reply to: [570816077](https://github.com/OData/odata.net/pull/1993#discussion_r570816077) [](ancestors = 570816077)

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -835,9 +835,22 @@ private static SelectedPropertiesNode ProcessSubExpand(string nodeName, Selected
...
+
+            foreach(string name in selectList)
+            {
+                if (!expandSet.Contains(name))
+                {
+                    rawSelect.Add(name);
+                }
+            }
+                       
             if (rawSelect.Count == 0 && expandList.All(n => n.IsEntireSubtree()))
```
mikepizzo, line 854: >expandList.All(n => n.IsEntireSubtree() [](start = 40, length = 39)

rather than re-iterate through expandlist, just add a boolean property "isEntireSubtree" that is initialized to true and changed to false as you iterate through the expandList on 839 building expandSet.

src/Microsoft.OData.Core/Evaluation/ODataConventionalIdMetadataBuilder.cs
mikepizzo: +1 -- it seems like the existing code would be more performant than individually removing elements.

---
In reply to: [570827746](https://github.com/OData/odata.net/pull/1993#discussion_r570827746) [](ancestors = 570827746)

mikepizzo:
>                 return new List<ExpandTermToken>();

just return Array.Empty<ExpandTermToken>()

---
Refers to: src/Microsoft.OData.Core/UriParser/Binders/ExpandTreeNormalizer.cs:140 in 8bb05ce. [](commit_id = 8bb05cead0cd470e803b1a35de33471381e3354f, deletion_comment = False)

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsElement.cs
```diff
@@ -127,7 +127,7 @@ protected IEnumerable<IEdmDirectValueAnnotation> ComputeDirectValueAnnotations()
                 return null;
             }
 
-            List<CsdlDirectValueAnnotation> annotations = this.Element.ImmediateValueAnnotations.ToList();
+            IEnumerable<CsdlDirectValueAnnotation> annotations = this.Element.ImmediateValueAnnotations;
             if (annotations.FirstOrDefault() != null)
```
mikepizzo, line 131: >FirstOrDefault [](start = 28, length = 14)

Maybe use Any() here?

mikepizzo:
>             }

alternative (so that you don't have to call Any() and then iterate) would be to iterate and allocate List<IEdmDirectValueAnnotation> on the first (if any) iteration.  i.e.;

               List<IEdmDirectValueAnnotation> wrappedAnnotations;

                foreach (CsdlDirectValueAnnotation annotation in this.Element.ImmediateValueAnnotations)
                {
                    if (wrappedAnnotation == null)
                   {
                       wrappedAnnotations = new List<IEdmDirectValueAnnotation>();
                   }

                   wrappedAnnotations.Add(new CsdlSemanticsDirectValueAnnotation(annotation, this.Model));
                }

                return wrappedAnnotations;
            }

This trades off a null check per ValueAnnotation with the Any() (or FirstOrDefault()== null) check.

---
Refers to: src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsElement.cs:141 in 8bb05ce. [](commit_id = 8bb05cead0cd470e803b1a35de33471381e3354f, deletion_comment = False)

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsNavigationSource.cs
```diff
@@ -127,7 +127,7 @@ public IEnumerable<IEdmNavigationPropertyBinding> FindNavigationPropertyBindings
         {
             if (!navigationProperty.ContainsTarget)
             {
-                return this.NavigationPropertyBindings.Where(targetMapping => targetMapping.NavigationProperty == navigationProperty).ToList();
+                return this.NavigationPropertyBindings.Where(targetMapping => targetMapping.NavigationProperty == navigationProperty);
             }
```
mikepizzo, line 131: if this gets at all hot, it would be a good candidate for caching. you could iterate the navigationpropertybindings once, creating a dictionary of navigationProperty to NavigationPropertyBinding, and then just return the value from the dictionary each request.

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -2864,8 +2864,8 @@ internal static bool TryGetRelativeEntitySetPath(IEdmElement element, Collection
             relativeNavigations = null;
             lastEntityType = null;
 
-            var pathItems = pathExpression.PathSegments.ToList();
-            if (pathItems.Count < 1)
+            var pathItems = pathExpression.PathSegments;
+            if (!pathItems.Any())
```
mikepizzo, line 2868: Since you call pathItems.First() on line 2883 to get the first path item, you could just call string bindingParameterName = pathItems.FirstOrDefault() here, then check for bindingParameterName == null to determine if you should return the error.  That way, when you got to line 2883 you would have already determined bindingParameterName and wouldn't have to call First().

src/Microsoft.OData.Edm/Schema/EdmReferentialConstraint.cs
```diff
@@ -24,7 +24,7 @@ public class EdmReferentialConstraint : IEdmReferentialConstraint
         public EdmReferentialConstraint(IEnumerable<EdmReferentialConstraintPropertyPair> propertyPairs)
         {
             EdmUtil.CheckArgumentNull(propertyPairs, "propertyPairs");
-            this.propertyPairs = propertyPairs.ToList();
+            this.propertyPairs = propertyPairs;
```
mikepizzo, line 27: this changes semantics slightly by delaying iteration until the caller calls PropertyPairs.  Are we okay with that change in semantics?  Are we certain we will be able to iterate propertyPairs multiple times after the referential constraint is constructed?

src/Microsoft.OData.Core/Evaluation/ODataConventionalIdMetadataBuilder.cs
marabooy: @Sreejithpin just wanted to drop these two here as they may provide some more context as to why this code exists here as they contain the details and comments https://github.com/OData/odata.net/issues/1725 & the pr https://github.com/OData/odata.net/pull/1854.

Sreejithpin:
>             }

IEnumerable.Any() is O(1) since it enumarets only through 1 element, but for n elements if (wrappedAnnotation == null) will be done right, 

---
In reply to: [777978725](https://github.com/OData/odata.net/pull/1993#issuecomment-777978725) [](http://example.com/codeflow?ancestors=777978725,777978725)

---
Refers to: src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsElement.cs:141 in 8bb05ce. [](commit_id = 8bb05cead0cd470e803b1a35de33471381e3354f, deletion_comment = False)

Sreejithpin:
>                 return new List<ExpandTermToken>();

Array.Empty is available only in standard2.0

---
In reply to: [777973420](https://github.com/OData/odata.net/pull/1993#issuecomment-777973420) [](http://example.com/codeflow?ancestors=777973420,777973420)

---
Refers to: src/Microsoft.OData.Core/UriParser/Binders/ExpandTreeNormalizer.cs:140 in 8bb05ce. [](commit_id = 8bb05cead0cd470e803b1a35de33471381e3354f, deletion_comment = False)

src/Microsoft.OData.Core/ODataUtils.cs
Sreejithpin: Dont see it used by WebAPI, have asked JP whether thy use it in AGS

Sreejithpin: Asked JP , he told its used only once, so better to remove TOList()

src/Microsoft.OData.Core/ODataUtils.cs
Sreejithpin: ToList() is usually expensive O(n) time and O(n) space. But if someone is iterating over the Ienumarable multiple times, its better to use ToList(), In this case we have identified (checked with aGS) that its iterated only 1 time, so better to remove to list.

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -835,10 +835,30 @@ private static SelectedPropertiesNode ProcessSubExpand(string nodeName, Selected
...
+
+            foreach(string name in selectList)
+            {
+                if (!expandSet.Contains(name))
+                {
+                    rawSelect.Add(name);
+                }
+            }
+                       
+            if (rawSelect.Count == 0 && isEntireSubTree)
```
mikepizzo, line 861: >rawSelect.Count == 0 &&  [](http://example.com/codeflow?start=16&length=24)

Rather than count the items in rawSelect to see if there are zero, just set isEntireSubTree to false between line 857 and 858 (as well as 847) and here just check isEntireSubTree.

mikepizzo:
>             }

Could avoid the Any() all together and only enumerate annotations once by allocating wrappedAnnotations on first use in your foreach:

List<IEdmDirectValueAnnotation> wrappedAnnotations = null;
foreach (CsdlDirectValueAnnotation annotation in annotations)
{
    if (wrappedAnnotations == null)
    {
        wrappedAnnotations = new List<IEdmDirectValueAnnotation>();
    }
     wrappedAnnotations.Add(new CsdlSemanticsDirectValueAnnotation(annotation, this.Model));
}

return wrappedAnnotations;

since the number of immediate value annotations is generally small, the null check in the foreach is likely to be less overhead than evaluating ImmediateValueAnnotations.Any() (which, depending on the logic, could be expensive)

---
Refers to: src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsElement.cs:141 in e47b4c8. [](commit_id = e47b4c8cf6f07383dfd59b494c11664be212e4a6, deletion_comment = False)

mikepizzo:
>             }

alternative (so that you don't have to call Any() and then iterate) would be to iterate and allocate List<IEdmDirectValueAnnotation> on the first (if any) iteration.  i.e.;
   List<IEdmDirectValueAnnotation> wrappedAnnotations;

    foreach (CsdlDirectValueAnnotation annotation in this.Element.ImmediateValueAnnotations)
    {
        if (wrappedAnnotation == null)
       {
           wrappedAnnotations = new List<IEdmDirectValueAnnotation>();
       }

       wrappedAnnotations.Add(new CsdlSemanticsDirectValueAnnotation(annotation, this.Model));
    }

    return wrappedAnnotations;
}
This trades off a null check per ValueAnnotation with the Any() (or FirstOrDefault()== null) check.

---
Refers to: src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsElement.cs:141 in e47b4c8. [](commit_id = e47b4c8cf6f07383dfd59b494c11664be212e4a6, deletion_comment = False)

mikepizzo (changes requested):
:clock1:

habbes (approved):

PR #1995, opened 2021-02-04
Retrieve the content type for stream property from vocabulary annotation by convention

xuzhg:
…ion by default

### Issues

OData spec says:

6.4 Media Entity Type
An entity type that does not specify a base type MAY specify that it is a media entity type. Media entities are entities that represent a media stream, such as a photo. Use a media entity if the out-of-band stream is the main topic of interest and the media entity is just additional structured information attached to the stream. Use a normal entity with one or more properties of type Edm.Stream if the structured data of the entity is the main topic of interest and the stream data is just additional information attached to the structured data. For more information on media entities see [OData‑Protocol].

An entity type derived from a media entity type MUST indicate that it is also a media entity type.

**Media entity types MAY specify a list of acceptable media types using an annotation with term Core.AcceptableMediaTypes, see [OData‑VocCore].**

and

**Stream properties MAY specify a list of acceptable media types using an annotation with term Core.AcceptableMediaTypes, see [OData-VocCore].**

This PR is indented to retrieve the media types from the vocabulary annotation.

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
@@ -235,12 +235,15 @@ namespace Microsoft.OData.Evaluation
 
                 // from OData spec: Media entity types MAY specify a list of acceptable media types using an annotation with term Core.AcceptableMediaTypes
                 IEdmEntityType entityType = this.ResourceMetadataContext.ActualResourceType as IEdmEntityType;
-                var mediaTypes = this.MetadataContext.Model.GetVocabularyStringCollection(entityType, CoreVocabularyModel.AcceptableMediaTypesTerm);
-                if (mediaTypes.Count() == 1)
+                if (entityType != null)
                 {
-                    // Be noted: AcceptableMediaTypes might have more than one media type,
-                    // Convention (default) behavior only works if AcceptableMediaTypes is a collection of one.
-                    this.computedMediaResource.ContentType = mediaTypes.ElementAt(0);
+                    var mediaTypes = this.MetadataContext.Model.GetVocabularyStringCollection(entityType, CoreVocabularyModel.AcceptableMediaTypesTerm);
+                    if (mediaTypes.Count() == 1)
+                    {
+                        // Be noted: AcceptableMediaTypes might have more than one media type,
+                        // Convention (default) behavior only works if AcceptableMediaTypes is a collection of one.
+                        this.computedMediaResource.ContentType = mediaTypes.ElementAt(0);
+                    }
                 }
             }
 
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
@@ -13,6 +13,7 @@ namespace Microsoft.OData.Evaluation
     using System.Diagnostics.CodeAnalysis;
     using System.Text;
     using Microsoft.OData.Edm;
+    using Microsoft.OData.Edm.Vocabularies.V1;
     #endregion
 
     /// <summary>
@@ -230,6 +231,15 @@ namespace Microsoft.OData.Evaluation
             {
                 this.computedMediaResource = new ODataStreamReferenceValue();
                 this.computedMediaResource.SetMetadataBuilder(this, /*propertyName*/ null);
+
+                // from OData spec: Media entity types MAY specify a list of acceptable media types using an annotation with term Core.AcceptableMediaTypes
+                IEdmEntityType entityType = this.ResourceMetadataContext.ActualResourceType as IEdmEntityType;
+                var mediaTypes = this.MetadataContext.Model.GetVocabularyStringCollection(entityType, CoreVocabularyModel.AcceptableMediaTypesTerm);
+                if (mediaTypes != null)
+                {
+                    // TODO: Make sure we use ',' to concatant the multiple media types or pick the first one?
+                    this.computedMediaResource.ContentType = string.Join(",", mediaTypes);
+                }
             }
 
             return this.computedMediaResource;
@@ -426,6 +436,16 @@ namespace Microsoft.OData.Evaluation
                     {
                         ODataStreamReferenceValue streamPropertyValue = new ODataStreamReferenceValue();
                         streamPropertyValue.SetMetadataBuilder(this, missingStreamPropertyName);
+
+                        // by default, let's retrieve the content type from vocabulary annotation
+                        var edmProperty = projectedStreamProperties[missingStreamPropertyName];
+                        var mediaTypes = this.MetadataContext.Model.GetVocabularyStringCollection(edmProperty, CoreVocabularyModel.AcceptableMediaTypesTerm);
+                        if (mediaTypes != null)
+                        {
+                            // TODO: Make sure we use ',' to concatant the multiple media types or pick the first one?
+                            streamPropertyValue.ContentType = string.Join(",", mediaTypes);
+                        }
+
                         this.computedStreamProperties.Add(new ODataProperty { Name = missingStreamPropertyName, Value = streamPropertyValue });
                     }
                 }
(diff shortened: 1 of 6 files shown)


src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -230,6 +231,15 @@ internal override ODataStreamReferenceValue GetMediaResource()
             {
                 this.computedMediaResource = new ODataStreamReferenceValue();
                 this.computedMediaResource.SetMetadataBuilder(this, /*propertyName*/ null);
+
+                // from OData spec: Media entity types MAY specify a list of acceptable media types using an annotation with term Core.AcceptableMediaTypes
+                IEdmEntityType entityType = this.ResourceMetadataContext.ActualResourceType as IEdmEntityType;
+                var mediaTypes = this.MetadataContext.Model.GetVocabularyStringCollection(entityType, CoreVocabularyModel.AcceptableMediaTypesTerm);
+                if (mediaTypes != null)
+                {
+                    // TODO: Make sure we use ',' to concatant the multiple media types or pick the first one?
```
garethj-msft, line 240: List doesn't make sense for the returned ContentType.
This should use the first item.
So the convention behavior could only work when the acceptable types is a collection of one, which will be a fairly common case.

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -426,6 +436,16 @@ private IEnumerable<ODataProperty> GetComputedStreamProperties(IEnumerable<OData
                     {
                         ODataStreamReferenceValue streamPropertyValue = new ODataStreamReferenceValue();
                         streamPropertyValue.SetMetadataBuilder(this, missingStreamPropertyName);
+
+                        // by default, let's retrieve the content type from vocabulary annotation
+                        var edmProperty = projectedStreamProperties[missingStreamPropertyName];
+                        var mediaTypes = this.MetadataContext.Model.GetVocabularyStringCollection(edmProperty, CoreVocabularyModel.AcceptableMediaTypesTerm);
+                        if (mediaTypes != null)
+                        {
+                            // TODO: Make sure we use ',' to concatant the multiple media types or pick the first one?
```
garethj-msft, line 445: see comment above.

src/Microsoft.OData.Core/Evaluation/ODataConventionalResourceMetadataBuilder.cs
```diff
@@ -428,13 +429,23 @@ internal override ODataProperty GetNextUnprocessedStreamProperty()
...
                 ODataStreamReferenceValue streamPropertyValue = new ODataStreamReferenceValue();
                 streamPropertyValue.SetMetadataBuilder(this, propertyName);
+
+                // by default, let's retrieve the content type from vocabulary annotation
+                var edmProperty = this.ResourceMetadataContext.SelectedStreamProperties[propertyName];
+                var mediaTypes = this.MetadataContext.Model.GetVocabularyStringCollection(edmProperty, CoreVocabularyModel.AcceptableMediaTypesTerm);
+                if (mediaTypes != null)
+                {
+                    // TODO: Make sure we use ',' to concatant the multiple media types or pick the first one?
+                    streamPropertyValue.ContentType = string.Join(",", mediaTypes);
```
garethj-msft, line 439: See previous comment.

src/Microsoft.OData.Edm/Vocabularies/VocabularyAnnotationExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+
+using System.Collections.Generic;
+using System.Linq;
+
+namespace Microsoft.OData.Edm.Vocabularies.V1
+{
+    /// <summary>
+    /// Extension methods for the vocabulary annotations.
+    /// </summary>
+    public static class VocabularyAnnotationExtensions
```
garethj-msft, line 15: So glad you made this public :-) I can use this.

src/Microsoft.OData.Edm/Vocabularies/VocabularyAnnotationExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+        /// <summary>
+        /// Gets the collection of string for a target annotatable.
+        /// </summary>
+        /// <param name="model">The model referenced to.</param>
+        /// <param name="target">The target annotatable to find annotation.</param>
+        /// <param name="term">The target annotatable to find annotation.</param>
+        /// <returns>Null or a collection string of qualified type name.</returns>
+        public static IEnumerable<string> GetVocabularyStringCollection(this IEdmModel model, IEdmVocabularyAnnotatable target, IEdmTerm term)
+        {
+            if (model == null || target == null || term == null)
```
garethj-msft, line 26: I would expect argumentnullexceptions rather than returning null in a public API.

src/Microsoft.OData.Edm/Vocabularies/VocabularyAnnotationExtensions.cs
```diff
@@ -0,0 +1,44 @@
...
+            if (annotation != null)
+            {
+                IEdmCollectionExpression collectionExpression = annotation.Value as IEdmCollectionExpression;
+                if (collectionExpression != null && collectionExpression.Elements != null)
+                {
+                    return collectionExpression.Elements.OfType<IEdmStringConstantExpression>().Select(e => e.Value);
+                }
+            }
+
+            return null;
```
garethj-msft, line 41: I would expect Enumerable.Empty rather than null for any API that returns an IEnumerable<T> when there aren't any items.

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
xuzhg: @garethj-msft & @mikepizzo 

OData spec says:  http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_UpdateStreamValues:

*Stream properties MAY specify a list of acceptable media types using an annotation with term Core.AcceptableMediaTypes, see [OData-VocCore].*

http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_MediaEntityType
: 
*Media entity types MAY specify a list of acceptable media types using an annotation with term Core.AcceptableMediaTypes, see [OData‑VocCore].*

The AcceptableMediaTypes is defined as **COLLECTION** of media type. 
```xml
<Term Name="AcceptableMediaTypes" Type="Collection(Edm.String)" Nullable="false" AppliesTo="EntityType Property Term" />
```

But, there's no **Executable** instruction about how to format it into "ContentType" of stream (entity/property).

src/Microsoft.OData.Edm/Vocabularies/VocabularyAnnotationExtensions.cs
xuzhg: return Enumerable.Empty<string>();

---
In reply to: [570055513](https://github.com/OData/odata.net/pull/1995#discussion_r570055513) [](ancestors = 570055513)

mikepizzo: +1

---
In reply to: [570055513](https://github.com/OData/odata.net/pull/1995#discussion_r570055513) [](ancestors = 570055513)

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
mikepizzo: AcceptableMediaTypes should return a collection of media types, which is the set of media types that the media entity stream/stream property may contain.  If this collection contains a single media type (which is likely to be a common case), then we should set the Content-Type to be that single media type.  As Gareth says, we should not attempt to set Content-Type to multiple mediatypes; that's invalid as per the definition of the Content-Type header.

---
In reply to: [570412053](https://github.com/OData/odata.net/pull/1995#discussion_r570412053) [](ancestors = 570412053)

mikepizzo (approved):
:shipit:

PR #2002, opened 2021-02-18
Expand TextWriterWrapper abstract class with equivalent async method

gathogojr:
### Issues

*This pull request is a partial fulfilment of issue #2019*

### Description

Expand `TextWriterWrapper` internal abstract class with equivalent async method and provide the implementation in the derived `NonIndentedTextWriter` internal class and `JsonValueUtils` internal static class

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
@@ -1,0 +1,468 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="JsonValueUtils.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Json
+{
+    #region Namespaces
+    using System;
+    using System.Diagnostics;
+    using System.Diagnostics.CodeAnalysis;
+    using System.Globalization;
+    using System.IO;
+    using System.Text;
+    using System.Threading.Tasks;
+    using System.Xml;
+    using Microsoft.OData.Buffers;
+    using Microsoft.OData.Edm;
+    #endregion Namespaces
+
+    /// <summary>
+    /// Provides helper method for converting data values to and from the OData JSON format.
+    /// </summary>
+    internal static partial class JsonValueUtils
+    {
+        /// <summary>
+        /// Write a char value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">The char value to write.</param>
+        /// <param name="stringEscapeOption">The ODataStringEscapeOption to use in escaping the string.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, char value, ODataStringEscapeOption stringEscapeOption)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            if (stringEscapeOption == ODataStringEscapeOption.EscapeNonAscii || value <= 0x7F)
+            {
+                string escapedString = SpecialCharToEscapedStringMap[value];
+                if (escapedString != null)
+                {
+                    await writer.WriteAsync(escapedString).ConfigureAwait(false);
+                    return;
+                }
+            }
+
+            await writer.WriteAsync(value).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Write a boolean value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">The boolean value to write.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, bool value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteAsync(FormatAsBooleanLiteral(value)).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Write an integer value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Integer value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, int value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteAsync(value.ToString(CultureInfo.InvariantCulture)).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Write a float value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Float value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, float value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            if (JsonSharedUtils.IsFloatValueSerializedAsString(value))
+            {
+                await writer.WriteQuotedAsync(value.ToString(ODataNumberFormatInfo)).ConfigureAwait(false);
+            }
+            else
+            {
+                // float.ToString() supports a max scale of six,
+                // whereas float.MinValue and float.MaxValue have 8 digits scale. Hence we need
+                // to use XmlConvert in all other cases, except infinity
+                await writer.WriteAsync(XmlConvert.ToString(value)).ConfigureAwait(false);
+            }
+        }
+
+        /// <summary>
+        /// Write a short value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Short value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, short value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteAsync(value.ToString(CultureInfo.InvariantCulture)).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Write a long value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Long value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, long value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteAsync(value.ToString(CultureInfo.InvariantCulture)).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Write a double value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Double value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, double value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            if (JsonSharedUtils.IsDoubleValueSerializedAsString(value))
+            {
+                await writer.WriteQuotedAsync(value.ToString(ODataNumberFormatInfo)).ConfigureAwait(false);
+            }
+            else
+            {
+                // double.ToString() supports a max scale of 14,
+                // whereas double.MinValue and double.MaxValue have 16 digits scale. Hence we need
+                // to use XmlConvert in all other cases, except infinity
+                string valueToWrite = XmlConvert.ToString(value);
+
+                await writer.WriteAsync(valueToWrite).ConfigureAwait(false);
+                if (valueToWrite.IndexOfAny(DoubleIndicatingCharacters) < 0)
+                {
+                    await writer.WriteAsync(".0").ConfigureAwait(false);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Write a Guid value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Guid value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, Guid value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteQuotedAsync(value.ToString()).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Write a decimal value
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Decimal value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, decimal value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteAsync(value.ToString(CultureInfo.InvariantCulture)).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Write a DateTimeOffset value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">DateTimeOffset value to be written.</param>
+        /// <param name="dateTimeFormat">The format to write out the DateTime value in.</param>
+        [SuppressMessage("Microsoft.Globalization", "CA1303:Do not pass literals as localized parameters", MessageId = "Microsoft.OData.Json.JsonValueUtils.WriteQuotedAsync(System.IO.TextWriter,System.String)", Justification = "Constant defined by the JSON spec.")]
+        internal static async Task WriteValueAsync(this TextWriter writer, DateTimeOffset value, ODataJsonDateTimeFormat dateTimeFormat)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            switch (dateTimeFormat)
+            {
+                case ODataJsonDateTimeFormat.ISO8601DateTime:
+                    {
+                        // Uses the same format as DateTime but with offset:
+                        // jsonDateTime= quotation-mark
+                        //  YYYY-MM-DDThh:mm:ss.sTZD
+                        //  [("+" / "-") offset]
+                        //  quotation-mark
+                        //
+                        // offset = 4DIGIT
+                        string textValue = XmlConvert.ToString(value);
+                        await writer.WriteQuotedAsync(textValue).ConfigureAwait(false);
+                    }
+
+                    break;
+
+                case ODataJsonDateTimeFormat.ODataDateTime:
+                    {
+                        // Uses the same format as DateTime but with offset:
+                        // jsonDateTime= quotation-mark
+                        //  "\/Date("
+                        //  ticks
+                        //  [("+" / "-") offset]
+                        //  ")\/"
+                        //  quotation-mark
+                        //
+                        // ticks = *DIGIT
+                        // offset = 4DIGIT
+                        string textValue = FormatDateTimeAsJsonTicksString(value);
+                        await writer.WriteQuotedAsync(textValue).ConfigureAwait(false);
+                    }
+
+                    break;
+            }
+        }
+
+        /// <summary>
+        /// Write a TimeSpan value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">TimeSpan value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, TimeSpan value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteQuotedAsync(EdmValueWriter.DurationAsXml(value)).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Write a TimeOfDay value
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">TimeOfDay value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, TimeOfDay value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteQuotedAsync(value.ToString()).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Write a Date value
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Date value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, Date value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteQuotedAsync(value.ToString()).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Write a byte value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Byte value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, byte value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteAsync(value.ToString(CultureInfo.InvariantCulture)).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Write an sbyte value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">SByte value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, sbyte value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteAsync(value.ToString(CultureInfo.InvariantCulture)).ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Write a string value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">String value to be written.</param>
+        /// <param name="stringEscapeOption">The string escape option.</param>
+        /// <param name="buffer">Char buffer to use for streaming data.</param>
+        /// <param name="arrayPool">Array pool for renting a buffer.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, string value, ODataStringEscapeOption stringEscapeOption, Ref<char[]> buffer, ICharArrayPool arrayPool = null)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            if (value == null)
+            {
+                await writer.WriteAsync(JsonConstants.JsonNullLiteral).ConfigureAwait(false);
+            }
+            else
+            {
+                await writer.WriteEscapedJsonStringAsync(value, stringEscapeOption, buffer, arrayPool).ConfigureAwait(false);
+            }
+        }
+
+        /// <summary>
+        /// Write a byte array.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Byte array to be written.</param>
+        /// <param name="buffer">Char buffer to use for streaming data.</param>
+        /// <param name="arrayPool">Array pool for renting a buffer.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, byte[] value, Ref<char[]> buffer, ICharArrayPool arrayPool = null)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            if (value == null)
+            {
+                await writer.WriteAsync(JsonConstants.JsonNullLiteral).ConfigureAwait(false);
+            }
+            else
+            {
+                await writer.WriteAsync(JsonConstants.QuoteCharacter).ConfigureAwait(false);
+                await writer.WriteBinaryStringAsync(value, buffer, arrayPool).ConfigureAwait(false);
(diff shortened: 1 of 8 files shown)


src/Microsoft.OData.Core/Json/TextWriterWrapper.cs
```diff
@@ -76,6 +77,31 @@ public override void Flush()
             this.writer.Flush();
         }
 
+        /// <summary>
+        /// Increases the level of indentation applied to the output asynchronously.
+        /// </summary>
+        public virtual Task IncreaseIndentationAsync()
```
xuzhg, line 83: >Task [](start = 23, length = 4)

not need the async keyword?

xuzhg:
>     internal sealed class NonIndentedTextWriter : TextWriterWrapper

Is this class still using?

---
Refers to: src/Microsoft.OData.Core/Json/NonIndentedTextWriter.cs:15 in 907d31f. [](commit_id = 907d31f94923f39ace2237f41d7ed06ea88881c1, deletion_comment = False)

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -0,0 +1,491 @@
...
+    /// </summary>
+    internal static partial class JsonValueUtils
+    {
+        /// <summary>
+        /// Write a char value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">The char value to write.</param>
+        /// <param name="stringEscapeOption">The ODataStringEscapeOption to use in escaping the string.</param>
+        internal static async Task WriteValueAsync(TextWriter writer, char value, ODataStringEscapeOption stringEscapeOption)
```
xuzhg, line 33: >TextWriter [](start = 51, length = 10)

maybe make these methods as extension methods to (this TextWriter writer)?

src/Microsoft.OData.Core/RefOfT.cs
```diff
@@ -0,0 +1,60 @@
...
+            this.value = value;
+        }
+
+        /// <summary>
+        /// Gets or sets the value wrapped in the object.
+        /// </summary>
+        public T Value
+        {
+            get { return this.value; }
+            set { this.value = value; }
```
xuzhg, line 39: >set { this.value = value; } [](start = 12, length = 27)

If we have the constructor to set the value, shall we keep the public set for the property?

src/Microsoft.OData.Core/RefOfT.cs
```diff
@@ -0,0 +1,60 @@
...
+        }
+
+        /// <summary>
+        /// Supports implicit conversion from a <see cref="Ref{T}"/> object to a T object/value.
+        /// </summary>
+        /// <param name="ref">The <see cref="Ref{T}"/> object.</param>
+        public static implicit operator T(Ref<T> @ref)
```
xuzhg, line 46: >@ref [](start = 49, length = 4)

maybe we can just rename it as "otherRef" to get rid of @.

src/Microsoft.OData.Core/RefOfT.cs
```diff
@@ -0,0 +1,60 @@
...
+        /// <param name="value">The value to be wrapped.</param>
+        public Ref(T value)
+        {
+            this.value = value;
+        }
+
+        /// <summary>
+        /// Gets or sets the value wrapped in the object.
+        /// </summary>
+        public T Value
```
xuzhg, line 36: >T Value [](start = 15, length = 7)

we can just use auto-property

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsAsyncTests.cs
```diff
@@ -0,0 +1,326 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="JsonValueUtilsTests.cs" company="Microsoft">
```
xuzhg, line 2: >JsonValueUtilsTests [](start = 20, length = 19)

JsonValueUtilsAsyncTests.cs

xuzhg (approved):
:shipit:

gathogojr:
> > ```
> > internal sealed class NonIndentedTextWriter : TextWriterWrapper
> > ```
> 
> Is this class still using?
> 
> Refers to: src/Microsoft.OData.Core/Json/NonIndentedTextWriter.cs:15 in [907d31f](https://github.com/OData/odata.net/commit/907d31f94923f39ace2237f41d7ed06ea88881c1). [](commit_id = [907d31f](https://github.com/OData/odata.net/commit/907d31f94923f39ace2237f41d7ed06ea88881c1), deletion_comment = False)

The `NonIndentedTextWriter` doesn't apply any indentations or newlines. It rides on the default implementations for `IncreaseIndentation`, `DecreaseIndentation`, `IncreaseIndentationAsync` and `DecreaseIndentationAsync` that [do nothing](https://github.com/OData/odata.net/blob/8444960bed25590dcdb6398b86cd5f7e3acef7a5/src/Microsoft.OData.Core/Json/TextWriterWrapper.cs#L60).
It also overrides the implementations for `WriteLine` and `WriteLineAsync` to [do nothing](https://github.com/OData/odata.net/blob/8444960bed25590dcdb6398b86cd5f7e3acef7a5/src/Microsoft.OData.Core/Json/NonIndentedTextWriter.cs#L47).

src/Microsoft.OData.Core/RefOfT.cs
gathogojr: @xuzhg Keeping the property setter public allows us do something like [this](https://github.com/OData/odata.net/pull/2002/files#diff-44795ee4fde32bffe0ac82d8b21b2ac93180517dce7d66dfeb7afce151a1fb8bR408)
```csharp
buffer.Value = BufferUtils.InitializeBufferIfRequired(bufferPool, buffer.Value);
```

marabooy (approved):

src/Microsoft.OData.Core/Json/TextWriterWrapper.cs
```diff
@@ -76,6 +77,30 @@ public override void Flush()
             this.writer.Flush();
         }
 
+        /// <summary>
+        /// Increases the level of indentation applied to the output asynchronously.
+        /// </summary>
+        public virtual async Task IncreaseIndentationAsync()
+        {
+            await TaskUtils.CompletedTask.ConfigureAwait(false);
```
mikepizzo, line 85: >.ConfigureAwait(false) [](start = 41, length = 22)

Shouldn't need to call ConfigureAwait on a completed task (recurring comment)

mikepizzo, line 85: >TaskUtils.CompletedTask [](start = 18, length = 23)

In 8.0, hopefully we can move to Task.CompletedTask/Task.FromResult instead of TaskUtils...

src/Microsoft.OData.Core/Json/NonIndentedTextWriter.cs
```diff
@@ -41,11 +52,30 @@ public override void Write(char value)
...
         {
         }
+
+        /// <summary>
+        /// Writes a new line asynchronously.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        public override async Task WriteLineAsync()
+        {
+            await TaskUtils.CompletedTask.ConfigureAwait(false);
```
mikepizzo, line 78: Doesn't this need to call WriteLineAsync on the underling writer?

mikepizzo (changes requested):
:clock1:

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -0,0 +1,491 @@
...
+        /// <summary>
+        /// Write the string value with quotes.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="text">String value to be written.</param>
+        private static async Task WriteQuotedAsync(this TextWriter writer, string text)
+        {
+            await writer.WriteAsync(JsonConstants.QuoteCharacter).ConfigureAwait(false);
+            await writer.WriteAsync(text).ConfigureAwait(false);
+            await writer.WriteAsync(JsonConstants.QuoteCharacter).ConfigureAwait(false);
```
mikepizzo, line 488: I wonder what the trade-off would be between making 3 asynchronous write calls versus doing the concatenation and then making a single async write call:
   await writer.WriteAsync(String.Concat(JsonConstants.QuoteCharacter, text, JsonConstants.QuoteCharacter)).ConfigureAwait(false);

This seems to suggest, at least if the call result in Response.WriteAsync, the concatenation would be faster.  This would be interesting to benchmark.

src/Microsoft.OData.Core/Json/NonIndentedTextWriter.cs
gathogojr: @mikepizzo For the `NonIndentedTextWriter`, we don't really apply newlines and indentations. Kindly take a look at the equivalent [`WriteLine`](https://github.com/OData/odata.net/blob/cc9d86f0069be395a4f07aabe5cad0cec7b9a2cd/src/Microsoft.OData.Core/Json/NonIndentedTextWriter.cs#L47) in the synchronous workflow.

src/Microsoft.OData.Core/Json/TextWriterWrapper.cs
gathogojr: @mikepizzo Initially I had this method without the `async` modifier as follows:
```
public virtual Task IncreaseIdentationAsync()
{
    return TaskUtils.CompletedTask;
}
```
Acting on review feedback, I added `async` and then added `.ConfigureAwait(false)` to pacify the analyzer. In your opinion, do you think its okay to ditch the `async` modifier and just return the completed task? Or do I just suppress the analyzer warning?

odero: `ConfigureAwait` on an already completed task has no effect. That said I'd stick with `async` and suppress the analyzer

odero (approved):

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
gathogojr: @mikepizzo I did some benchmarking and observed the following results that make a strong case for the concatenation

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19042
Intel Core i7-8665U CPU 1.90GHz (Coffee Lake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=5.0.103
  [Host]     : .NET Core 3.1.12 (CoreCLR 4.700.21.6504, CoreFX 4.700.21.6905), X64 RyuJIT
  Job-WIKFZT : .NET Core 3.1.12 (CoreCLR 4.700.21.6504, CoreFX 4.700.21.6905), X64 RyuJIT

InvocationCount=1  UnrollFactor=1

// Run 1

|                  Method |     Mean |   Error |  StdDev |
|------------------------ |---------:|--------:|--------:|
| WriteAsyncWithoutConcat | 202.4 ms | 3.93 ms | 3.49 ms |
|    WriteAsyncWithConcat | 120.4 ms | 2.27 ms | 1.89 ms |

// Run 2
|                  Method |     Mean |   Error |  StdDev |
|------------------------ |---------:|--------:|--------:|
| WriteAsyncWithoutConcat | 254.1 ms | 4.82 ms | 3.76 ms |
|    WriteAsyncWithConcat | 120.4 ms | 2.36 ms | 2.09 ms |

// Run 3
|                  Method |     Mean |   Error |  StdDev |
|------------------------ |---------:|--------:|--------:|
| WriteAsyncWithoutConcat | 163.1 ms | 2.42 ms | 2.02 ms |
|    WriteAsyncWithConcat | 121.3 ms | 2.03 ms | 2.49 ms |

src/Microsoft.OData.Core/Json/NonIndentedTextWriter.cs
habbes: @gathogojr maybe that should be mentioned in a comment?

habbes:
@gathogojr seems like there's a lot of duplicate logic between the async and sync apis. Is this inevitable or can code be reused? Might this cause some maintainability issues for as long we need to support both apis?

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -0,0 +1,491 @@
...
+                        string textValue = String.Format(
+                            CultureInfo.InvariantCulture,
+                            JsonConstants.ODataDateTimeOffsetFormat,
+                            DateTimeTicksToJsonTicks(value.Ticks),
+                            offsetMinutes >= 0 ? JsonConstants.ODataDateTimeOffsetPlusSign : string.Empty,
+                            offsetMinutes);
+                        await writer.WriteQuotedAsync(textValue).ConfigureAwait(false);
+                    }
+
+                    break;
```
habbes, line 223: Is it possible to abstract that logic that generates the `textValue` in a method then re-use that method in both `WriteValueAsync` and `WriteValue`?

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -0,0 +1,491 @@
...
+        /// <param name="bufferPool">Array pool for renting a buffer.</param>
+        internal static async Task WriteEscapedJsonStringAsync(this TextWriter writer, string inputString,
+            ODataStringEscapeOption stringEscapeOption, Ref<char[]> buffer, ICharArrayPool bufferPool = null)
+        {
+            Debug.Assert(writer != null, "writer != null");
+            Debug.Assert(inputString != null, "The string value must not be null.");
+
+            await writer.WriteAsync(JsonConstants.QuoteCharacter).ConfigureAwait(false);
+            await writer.WriteEscapedJsonStringValueAsync(inputString, stringEscapeOption, buffer, bufferPool).ConfigureAwait(false);
+            await writer.WriteAsync(JsonConstants.QuoteCharacter).ConfigureAwait(false);
```
habbes, line 381: Since you've created `WriteQuotedAsync` to reduce the number of async calls when writing something wrapped into quotes, can that approach be applied here as well? If it's worth while performance wise.
Don't know whether the quotes could be concatenated around the inputString, or whether the quotes could be added to the buffer that gets written.

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -0,0 +1,491 @@
...
+            Debug.Assert(writer != null, "writer != null");
+            Debug.Assert(value != null, "The value must not be null.");
+
+            buffer.Value = BufferUtils.InitializeBufferIfRequired(arrayPool, buffer.Value);
+            Debug.Assert(buffer.Value != null);
+
+            int bufferLength = buffer.Value.Length;
+
+            // Try to hold base64 string as much as possible in one converting.
+            int bufferByteSize = bufferLength * 3 / 4;
```
habbes, line 350: Why ` 3/4`? maybe that should be a constant?

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -0,0 +1,491 @@
...
+
+            if (value == null)
+            {
+                await writer.WriteAsync(JsonConstants.JsonNullLiteral).ConfigureAwait(false);
+            }
+            else
+            {
+                await writer.WriteAsync(JsonConstants.QuoteCharacter).ConfigureAwait(false);
+                await writer.WriteBinaryStringAsync(value, buffer, arrayPool).ConfigureAwait(false);
+                await writer.WriteAsync(JsonConstants.QuoteCharacter).ConfigureAwait(false);
```
habbes, line 328: Is it also possible to write the quotes here as part of the writing the buffer to reduce the async calls?

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -0,0 +1,491 @@
...
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Float value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, float value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            if (float.IsInfinity(value) || float.IsNaN(value))
+            {
+                await writer.WriteQuotedAsync(value.ToString(ODataNumberFormatInfo)).ConfigureAwait(false);
```
KenitoInc, line 85: Add Culture info in the `ToString`

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -0,0 +1,491 @@
...
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Double value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, double value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            if (JsonSharedUtils.IsDoubleValueSerializedAsString(value))
+            {
+                await writer.WriteQuotedAsync(value.ToString(ODataNumberFormatInfo)).ConfigureAwait(false);
```
KenitoInc, line 131: Add Culture Info in `ToString`
`value.ToString(ODataNumberFormatInfo, CultureInfo.InvariantCulture)`

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -0,0 +1,491 @@
...
+        /// <summary>
+        /// Write a Guid value.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Guid value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, Guid value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteQuotedAsync(value.ToString()).ConfigureAwait(false);
```
KenitoInc, line 157: Add CultureInfo

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -0,0 +1,491 @@
...
+                case ODataJsonDateTimeFormat.ISO8601DateTime:
+                    {
+                        // Uses the same format as DateTime but with offset:
+                        // jsonDateTime= quotation-mark
+                        //  YYYY-MM-DDThh:mm:ss.sTZD
+                        //  [("+" / "-") offset]
+                        //  quotation-mark
+                        //
+                        // offset = 4DIGIT
+                        string textValue = XmlConvert.ToString(value);
```
KenitoInc, line 196: Add CultureInfo

KenitoInc: I can see you you are suppressing this.

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -0,0 +1,491 @@
...
+        /// <summary>
+        /// Write a TimeOfDay value
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">TimeOfDay value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, TimeOfDay value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteQuotedAsync(value.ToString()).ConfigureAwait(false);
```
KenitoInc, line 248: Add CultureInfo

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -0,0 +1,491 @@
...
+        /// <summary>
+        /// Write a Date value
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="value">Date value to be written.</param>
+        internal static async Task WriteValueAsync(this TextWriter writer, Date value)
+        {
+            Debug.Assert(writer != null, "writer != null");
+
+            await writer.WriteQuotedAsync(value.ToString()).ConfigureAwait(false);
```
KenitoInc, line 260: Add CultureInfo

KenitoInc (changes requested):
A few FxCop change requests.

On a broader conversation on async, Is there any plan to support Cancellation Tokens in ODL? cc @mikepizzo @xuzhg

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
gathogojr: @KenitoInc
The `.ToString()` overloads available when converting a number to string are as follows:
- `.ToString()`
- `.ToString(string format)`
- `.ToString(IFormatProvider provider)`
- `.ToString(string format, IFormatProvider provider)`

We're using the 3rd overload by supplying the `IFormatProvider` - `ODataNumberFormatInfo`. There's no overload that takes `CultureInfo`. That is usually applicable when using `string.Format()`

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
gathogojr: Refer https://github.com/OData/odata.net/pull/2002#discussion_r582865181

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
gathogojr: @KenitoInc
The `.ToString()` overloads available when converting a `Guid` to string are as follows:
- `.ToString()`
- `.ToString(string format)`
- `.ToString(string format, IFormatProvider provider)`

We're using the 1st overload and it suffices for a `Guid`

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
gathogojr: Here we're formatting a `TimeOfDay` value and that is taken care of in the `Edm` library [here](https://github.com/OData/odata.net/blob/cc9d86f0069be395a4f07aabe5cad0cec7b9a2cd/src/Microsoft.OData.Edm/Schema/TimeOfDay.cs#L297). The `.ToString()` method for `TimeOfDay` does not support a `CultureInfo` argument

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
gathogojr: @KenitoInc `.ToString()` against a `byte` suffices here

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
gathogojr: @KenitoInc `System.Xml.XmlConvert.ToString(DateTimeOffset)` will do the expected thing here for ISO8601 date format

`System.Xml.XmlConvert.ToString(DateTimeOffset.Now)` => `2021-02-25T17:31:58.9858279+03:00`

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
KenitoInc: Actually this line is taking care of the CultureInfo https://github.com/OData/odata.net/blob/cc9d86f0069be395a4f07aabe5cad0cec7b9a2cd/src/Microsoft.OData.Edm/Schema/TimeOfDay.cs#L302

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
gathogojr: @KenitoInc The `value` here is of type `Microsoft.OData.Edm.Date` and the custom `.ToString()` method already takes culture into consideration [here](https://github.com/OData/odata.net/blob/8444960bed25590dcdb6398b86cd5f7e3acef7a5/src/Microsoft.OData.Edm/Schema/Date.cs#L239)

KenitoInc (approved):
LGTM

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
gathogojr: @habbes Done

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
gathogojr: @habbes Based on the fact that quotes in the input string would be escaped, concatenating the enclosing quotes would result into them getting escaped as well. Regarding the option of writing the quotes to directly to the buffer, that's a change I'd prefer to be handle separately as part of perf improvements back by an actual performance analysis.

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
gathogojr: @habbes Not a 100% sure. This was lifted from the synchronous workflow. I believe the idea behind using 3 quarters of the buffer is to leave an allowance.

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
gathogojr: @habbes Same argument as [here](https://github.com/OData/odata.net/pull/2002#discussion_r584965109). We escape special characters when writing the byte array

habbes (approved):

src/Microsoft.OData.Core/Json/TextWriterWrapper.cs
mikepizzo: I would keep the async but supress the warning (as you've done).

---
In reply to: [582566983](https://github.com/OData/odata.net/pull/2002#discussion_r582566983) [](ancestors = 582566983)

mikepizzo: :thumbsup: 

---
In reply to: [582591007](https://github.com/OData/odata.net/pull/2002#discussion_r582591007) [](ancestors = 582591007)

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
mikepizzo: If "WriteEscapedJsonStringValueAsync" is only called from here (or from other places where the string must be enclosed in quotes), then we could add the writing of the quote characters as part of that method, rather than writing them to the buffer before & after calling that method.  But, happy to take that separately, as it looks like the WriteEscapedJsonStringValueAsync method could use stringbuilder to build the string and writeasync once, but we'd want to do perf to verify that using a stringbuilder would be better than individual async writes (though I suspect it would be).

---
In reply to: [584965109](https://github.com/OData/odata.net/pull/2002#discussion_r584965109) [](ancestors = 584965109)

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
mikepizzo: Awesome -- thanks for the analysis & update.  This will be a good opportunity for improvement throughout the stack, especially in perf-sensitive areas like reading/writing values.

---
In reply to: [582601314](https://github.com/OData/odata.net/pull/2002#discussion_r582601314) [](ancestors = 582601314)

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -0,0 +1,468 @@
...
+
+        /// <summary>
+        /// Write the string value with quotes.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="text">String value to be written.</param>
+        private static async Task WriteQuotedAsync(this TextWriter writer, string text)
+        {
+            await writer.WriteAsync(
+                JsonConstants.QuoteCharacter + text + JsonConstants.QuoteCharacter)
```
mikepizzo, line 464: >JsonConstants.QuoteCharacter + text + JsonConstants.QuoteCharacter [](start = 16, length = 66)

Consider using String.Concat(JsonConstants.QuoteCharacter, text, JsonConstants.QuoteCharacter) instead of JsonConstants.QuoteCharacter + text + JsonConstants.QuoteCharacter. I think the compiler might automatically optimize this case by generating the concat, rather than do two allocations, but I'm not sure.

mikepizzo (approved):
:shipit:

PR #2014, opened 2021-03-02
Implement asynchronous support in JsonWriter

gathogojr:
### Issues

*This pull request is a partial fulfilment of issue #2019.*

### Description

Implement asynchronous support in JsonWriter
- Added `IJsonWriterAsync` interface.
- Added `IJsonStreamWriterAsync` interface.
- Added `IJsonWriterFactoryAsync` interface.
- Implemented `IJsonStreamWriterAsync` asynchronous methods in `JsonWriter` class.
- Implemented `IJsonWriterFactoryAsync` methods in `DefaultJsonWriterFactory` class.
- Implemented asynchronous methods in `ODataJsonTextWriter` class.
- Implemented asynchronous methods in `ODataBinaryStreamWriter` class.
- Added test for `JsonWriter`, `ODataJsonTextWriter` and `ODataBinaryStreamWriter` classes asynchronous methods.
- Passing around a `char` array reference in the synchronous workflow to support use of char array pool and a `char` array wrapped in an object in the asynchronous workflow started degenerating into a code smell so I opted for one that works across both workflows.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/IJsonWriterFactoryAsync.cs
+++ b/src/Microsoft.OData.Core/Json/IJsonWriterFactoryAsync.cs
@@ -22,6 +22,6 @@ namespace Microsoft.OData.Json
         /// <param name="textWriter">Writer to which text needs to be written.</param>
         /// <param name="isIeee754Compatible">True if it is IEEE754Compatible.</param>
         /// <returns>The JSON writer created.</returns>
-        IJsonWriterAsync CreateAsyncJsonWriter(TextWriter textWriter, bool isIeee754Compatible);
+        IJsonWriterAsync CreateAsynchronousJsonWriter(TextWriter textWriter, bool isIeee754Compatible);
     }
 }
--- a/src/Microsoft.OData.Core/Json/IJsonWriterFactoryAsync.cs
+++ b/src/Microsoft.OData.Core/Json/IJsonWriterFactoryAsync.cs
@@ -1,0 +1,27 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="IJsonWriterFactoryAsync.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+
+namespace Microsoft.OData.Json
+{
+    /// <summary>
+    /// Interface of the factory to create asynchronous JSON writers.
+    /// </summary>
+    [CLSCompliant(false)]
+    public interface IJsonWriterFactoryAsync
+    {
+        /// <summary>
+        /// Creates a new JSON writer of <see cref="IJsonWriterAsync"/> with support for writing asynchronously.
+        /// </summary>
+        /// <param name="textWriter">Writer to which text needs to be written.</param>
+        /// <param name="isIeee754Compatible">True if it is IEEE754Compatible.</param>
+        /// <returns>The JSON writer created.</returns>
+        IJsonWriterAsync CreateAsyncJsonWriter(TextWriter textWriter, bool isIeee754Compatible);
+    }
+}
--- a/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
@@ -314,7 +314,12 @@ namespace Microsoft.OData.Json
         /// <param name="stringEscapeOption">The string escape option.</param>
         /// <param name="buffer">Char buffer to use for streaming data.</param>
         /// <param name="arrayPool">Array pool for renting a buffer.</param>
-        internal static void WriteValue(TextWriter writer, string value, ODataStringEscapeOption stringEscapeOption, Ref<char[]> buffer, ICharArrayPool arrayPool = null)
+        internal static void WriteValue(
+            TextWriter writer,
+            string value,
+            ODataStringEscapeOption stringEscapeOption,
+            Ref<char[]> buffer,
+            ICharArrayPool arrayPool = null)
         {
             Debug.Assert(writer != null, "writer != null");
 
@@ -386,8 +391,12 @@ namespace Microsoft.OData.Json
         /// <param name="stringEscapeOption">The string escape option.</param>
         /// <param name="buffer">Char buffer to use for streaming data</param>
         /// <param name="bufferPool">Array pool for renting a buffer.</param>
-        internal static void WriteEscapedJsonString(TextWriter writer, string inputString,
-            ODataStringEscapeOption stringEscapeOption, Ref<char[]> buffer, ICharArrayPool bufferPool = null)
+        internal static void WriteEscapedJsonString(
+            TextWriter writer,
+            string inputString,
+            ODataStringEscapeOption stringEscapeOption,
+            Ref<char[]> buffer,
+            ICharArrayPool bufferPool = null)
         {
             Debug.Assert(writer != null, "writer != null");
             Debug.Assert(inputString != null, "The string value must not be null.");
@@ -405,8 +414,12 @@ namespace Microsoft.OData.Json
         /// <param name="stringEscapeOption">The string escape option.</param>
         /// <param name="buffer">Char buffer to use for streaming data.</param>
         /// <param name="bufferPool">Array pool for renting a buffer.</param>
-        internal static void WriteEscapedJsonStringValue(TextWriter writer, string inputString,
-            ODataStringEscapeOption stringEscapeOption, Ref<char[]> buffer, ICharArrayPool bufferPool)
+        internal static void WriteEscapedJsonStringValue(
+            TextWriter writer,
+            string inputString,
+            ODataStringEscapeOption stringEscapeOption,
+            Ref<char[]> buffer,
+            ICharArrayPool bufferPool)
         {
             Debug.Assert(writer != null, "writer != null");
             Debug.Assert(inputString != null, "The string value must not be null.");
@@ -469,7 +482,13 @@ namespace Microsoft.OData.Json
         /// <param name="stringEscapeOption">The string escape option.</param>
         /// <param name="buffer">Char buffer to use for streaming data.</param>
         /// <param name="bufferPool">Character buffer pool.</param>
-        internal static void WriteEscapedCharArray(TextWriter writer, char[] inputArray, int inputArrayOffset, int inputArrayCount, ODataStringEscapeOption stringEscapeOption, Ref<char[]> buffer, ICharArrayPool bufferPool)
+        internal static void WriteEscapedCharArray(
+            TextWriter writer,
+            char[] inputArray,
+            int inputArrayOffset,
+            int inputArrayCount,
+            ODataStringEscapeOption stringEscapeOption,
+            Ref<char[]> buffer, ICharArrayPool bufferPool)
         {
             int bufferIndex = 0;
             buffer.Value = BufferUtils.InitializeBufferIfRequired(bufferPool, buffer.Value);
@@ -781,7 +800,13 @@ namespace Microsoft.OData.Json
         /// IMPORTANT: After all characters have been written,
         /// caller is responsible for writing the final buffer contents to the writer.
         /// </remarks>
-        private static void WriteEscapedStringToBuffer(TextWriter writer, string inputString, ref int currentIndex, char[] buffer, ref int bufferIndex, ODataStringEscapeOption stringEscapeOption)
+        private static void WriteEscapedStringToBuffer(
+            TextWriter writer,
+            string inputString,
+            ref int currentIndex,
+            char[] buffer,
+            ref int bufferIndex,
+            ODataStringEscapeOption stringEscapeOption)
         {
             Debug.Assert(inputString != null, "inputString != null");
             Debug.Assert(buffer != null, "buffer != null");
@@ -806,7 +831,14 @@ namespace Microsoft.OData.Json
         /// IMPORTANT: After all characters have been written,
         /// caller is responsible for writing the final buffer contents to the writer.
         /// </remarks>
-        private static void WriteEscapedCharArrayToBuffer(TextWriter writer, char[] inputArray, ref int inputArrayOffset, int inputArrayCount, char[] buffer, ref int bufferIndex, ODataStringEscapeOption stringEscapeOption)
+        private static void WriteEscapedCharArrayToBuffer(
+            TextWriter writer,
+            char[] inputArray,
+            ref int inputArrayOffset,
+            int inputArrayCount,
+            char[] buffer,
+            ref int bufferIndex,
+            ODataStringEscapeOption stringEscapeOption)
         {
             Debug.Assert(inputArray != null, "inputArray != null");
             Debug.Assert(buffer != null, "buffer != null");
--- a/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
@@ -314,7 +314,7 @@ namespace Microsoft.OData.Json
         /// <param name="stringEscapeOption">The string escape option.</param>
         /// <param name="buffer">Char buffer to use for streaming data.</param>
         /// <param name="arrayPool">Array pool for renting a buffer.</param>
-        internal static void WriteValue(TextWriter writer, string value, ODataStringEscapeOption stringEscapeOption, ref char[] buffer, ICharArrayPool arrayPool = null)
+        internal static void WriteValue(TextWriter writer, string value, ODataStringEscapeOption stringEscapeOption, Ref<char[]> buffer, ICharArrayPool arrayPool = null)
         {
             Debug.Assert(writer != null, "writer != null");
 
@@ -324,7 +324,7 @@ namespace Microsoft.OData.Json
             }
             else
             {
-                JsonValueUtils.WriteEscapedJsonString(writer, value, stringEscapeOption, ref buffer, arrayPool);
+                JsonValueUtils.WriteEscapedJsonString(writer, value, stringEscapeOption, buffer, arrayPool);
             }
         }
 
@@ -335,7 +335,7 @@ namespace Microsoft.OData.Json
         /// <param name="value">Byte array to be written.</param>
         /// <param name="buffer">Char buffer to use for streaming data.</param>
         /// <param name="arrayPool">Array pool for renting a buffer.</param>
-        internal static void WriteValue(TextWriter writer, byte[] value, ref char[] buffer, ICharArrayPool arrayPool = null)
+        internal static void WriteValue(TextWriter writer, byte[] value, Ref<char[]> buffer, ICharArrayPool arrayPool = null)
         {
             Debug.Assert(writer != null, "writer != null");
 
@@ -346,7 +346,7 @@ namespace Microsoft.OData.Json
             else
             {
                 writer.Write(JsonConstants.QuoteCharacter);
-                WriteBinaryString(writer, value, ref buffer, arrayPool);
+                WriteBinaryString(writer, value, buffer, arrayPool);
                 writer.Write(JsonConstants.QuoteCharacter);
             }
         }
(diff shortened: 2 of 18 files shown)


src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -314,7 +314,7 @@ internal static void WriteValue(TextWriter writer, sbyte value)
         /// <param name="stringEscapeOption">The string escape option.</param>
         /// <param name="buffer">Char buffer to use for streaming data.</param>
         /// <param name="arrayPool">Array pool for renting a buffer.</param>
-        internal static void WriteValue(TextWriter writer, string value, ODataStringEscapeOption stringEscapeOption, ref char[] buffer, ICharArrayPool arrayPool = null)
+        internal static void WriteValue(TextWriter writer, string value, ODataStringEscapeOption stringEscapeOption, Ref<char[]> buffer, ICharArrayPool arrayPool = null)
```
habbes, line 317: Does the use Ref<char[]> affect memory allocations on the sync writer?

gathogojr: @habbes Passing around a `char` array reference in the synchronous workflow to support use of `char` array pool and a char array wrapped in an object in the asynchronous workflow started degenerating into a code smell so I opted for one that works across both workflows.
It required me to maintain both a `char[] buffer` and a `Ref<char[]> buffer` as member variables of the `JsonWriter` class, both renting  char arrays from the same pool, and returning to the same pool. The code felt brittle and likely to hide subtle bugs.
So the answer is yes, with this change both use the wrapped buffer.

src/Microsoft.OData.Core/Json/JsonWriter.cs
```diff
@@ -487,11 +489,15 @@ public void EndTextWriterValueScope()
...
+            if (this.ArrayPool != null)
             {
-                BufferUtils.ReturnToBuffer(this.ArrayPool, this.buffer);
+                if (this.wrappedBuffer.Value != null)
+                {
+                    BufferUtils.ReturnToBuffer(this.ArrayPool, this.wrappedBuffer.Value);
+                    this.wrappedBuffer.Value = null;
+                }
+
                 this.ArrayPool = null;
```
habbes, line 500: Could you explain the change in logic here?

In the original version this.ArrayPool is set to null only when the buffer is null. In the updated code, ArrayPool is set to null regardless of whether the buffer is null or not. Is this change significant? Or was it a bug in the original version?

src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
```diff
@@ -0,0 +1,448 @@
...
+
+        /// <summary>
+        /// End the padding function scope.
+        /// </summary>
+        public async Task EndPaddingFunctionScopeAsync()
+        {
+            Debug.Assert(this.scopes.Count > 0, "No scope to end.");
+
+            await this.writer.WriteLineAsync().ConfigureAwait(false);
+            await this.writer.DecreaseIndentationAsync().ConfigureAwait(false);
```
habbes, line 45: Does the `DecreaseIndentationAsync()` actually write anything? Should it be async?

gathogojr: @habbes Depending on the writer implementation, it might or might not write anything. For this reason, it's right to await. It just happens that for `NonIndentedTextWriter` it doesn't write anything.

src/Microsoft.OData.Core/Json/JsonWriter.cs
gathogojr: @habbes argh! I had rewritten this code when I was dealing with both a char buffer reference and a wrapped buffer. I'll rewrite it so the block is executed when both `this.ArrayPool` and `this.wrappedBuffer.Value` are `null`

src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
```diff
@@ -0,0 +1,448 @@
...
+            await this.writer.WriteAsync(JsonConstants.QuoteCharacter).ConfigureAwait(false);
+            await this.writer.FlushAsync().ConfigureAwait(false);
+            // IMPORTANT: The Dispose method of the returned stream does the following:
+            // - Writes trailing bytes to the writer synchronously
+            // - Flushes the buffer of the writer synchronously
+            // ODL supports net45 and netstandard1.1 (in addition to .netstandard2.0)
+            // This makes it complicated to implement IAsyncDisposable in ODataBinaryStreamWriter
+            // TODO: Can the returned stream be safely used asynchronously?
+            this.binaryValueStream = new ODataBinaryStreamWriter(writer, this.wrappedBuffer, ArrayPool);
+            return this.binaryValueStream;
```
habbes, line 349: If writing and flushing synchronously when disposing is a great concern, maybe you could extend the `ODataBinaryStreamWriter` with a version where `Dispose()` simply calls the `base.Dispose()` then you add a method like `WriteTrailingBytesAndFlushAsync()`. Then you call:
```c#
await (this.binaryValueStream as ODataBinaryStreamWriterAsync).WriteTrailingBytesAndFlushAsync();
this.binaryValueStream.Dispose();
this.binaryValueStream = null;
```

But this looks messy and maybe not worth it if it will be replaced when dropping net45.

habbes: But since the `binaryValueStream` property is private implementation detail, maybe there's no harm in changing its type to some custom `AsyncDisposableStream` that extends `Stream` and adds a `DisposeAsync` method. Then make `ODataBinaryStreamWriterAsync` extend that type. Then when `net45` is dropped, you could extend a regular `Stream`.

Or what about an extension method to the `Stream` class wrapped instead `#if NET45`?

src/Microsoft.OData.Core/Json/ODataJsonTextWriter.cs
```diff
@@ -299,54 +299,50 @@ public override Task FlushAsync()
         }
 
         /// <inheritdoc/>
-        public override Task WriteAsync(char value)
+        public override async Task WriteAsync(char value)
         {
-            return TaskUtils.GetTaskForSynchronousOperation(() =>
-                this.Write(value));
+            await this.WriteEscapedCharValueAsync(value).ConfigureAwait(false);
```
habbes, line 304: Is it necessary to async/await this given that the method being called already returns an async task? Or it doesn't make a difference? I assume we incur more overhead in this case compared to just returning the task? I could be wrong.

src/Microsoft.OData.Core/ODataBinaryStreamWriter.cs
```diff
@@ -131,37 +134,40 @@ public override long Position
...
+        public override async Task WriteAsync(byte[] bytes, int offset, int count, CancellationToken cancellationToken)
         {
-            return TaskUtils.GetTaskForSynchronousOperation(() =>
-                this.Write(buffer, offset, count));
+            Debug.Assert(this.wrappedBuffer != null, "this.wrappedBuffer != null");
+
+            // if we have less than 3 bytes, store the bytes and continue
+            if (count + trailingBytes.Length < MinBytesPerWriteEvent)
+            {
+                this.trailingBytes = this.trailingBytes.Concat(bytes.Skip(offset).Take(count)).ToArray();
```
habbes, line 161: Maybe this is not the scope of this PR, but just curious whether reallocating the `trailingBytes` all the time doesn't put pressure on the GC? Can a single array with an index to keep track of where to add new bytes? Or an array pool?

habbes: And the impact of the LINQ calls?

src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
habbes: I understand that argument for `WriteLineAsync`, but I guess I just couldn't imagine an implementation of increasing/decreasing indentation that would need to write to the stream. I assume it would update the indentation level, which would probably be stored in an int, then other `Writexxx` methods would read the indentation level to decide how many spaces to add after a new line.

src/Microsoft.OData.Core/Json/ODataJsonTextWriter.cs
habbes: According to the discussion here: https://stackoverflow.com/questions/38017016/async-task-then-await-task-vs-task-then-return-task the differences lie in exception handling and the state machine generated for async/await by the compiler. The discussions also suggest that there are optimizations to prevent unnecessary awaiting. But according to [the linked sharplab code sample](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLQFEAeAHJzkCWA9gHYBqUCBUwANigD4ACADAARMCMA3ALABQrDpwCsffgKYBmDgCY2AYTYBvAW3UcZTABwcAbGwCCyAJ4kAxoYDuUAjAAUASjYBeAHwcAnEdMWFUWnoIJ3ENNjUNaX0fM3MAJQh4BBInVw9jWP9AiGDHUI0I9VwCADdYCGiMvwCg1Pd9ADoFIgBbbHoYCAATJj1xAF8gA) it appears that the state machine instance is generated regardless. I think this could lead to unnecessary allocations.

KenitoInc: My thoughts are that we don't need async here.
But am not sure if there is extra overhead by having async

KenitoInc (commented):
A few comments

src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
gathogojr: @habbes On the contrary, since `WriteAsync`/`WriteValueAsync` methods do not concern themselves with whitespace characters, the typical implementation of `DecreaseIndentationAsync` would be responsible for working out the applicable indentation level while also writing the whitespace spaces required to apply the indentation.
For example, in `EndObjectScopeAsync` method, we do:
- `WriteLineAsync()` - Apply a new line
- `DecreaseIndentationAsync()` - Decrease indentation level and apply indentation
- `WriteAsync(scope.EndString)` - Write the `}` character to signal end of the object
From the above you can see that `DecreaseIndentationAsync` would have to be responsible for both decreasing indentation level and applying the indentation itself.

src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
```diff
@@ -0,0 +1,448 @@
...
+        {
+            Debug.Assert(this.scopes.Count > 0, "No scope to end.");
+
+            await this.writer.WriteLineAsync().ConfigureAwait(false);
+            await this.writer.DecreaseIndentationAsync().ConfigureAwait(false);
+            Scope scope = this.scopes.Pop();
+
+            Debug.Assert(scope.Type == ScopeType.Object, "Ending scope does not match.");
+
+            await this.writer.WriteAsync(scope.EndString).ConfigureAwait(false);
```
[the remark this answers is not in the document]
gathogojr: See my comment [here](https://github.com/OData/odata.net/pull/2014/files#r594085022)

src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
```diff
@@ -0,0 +1,448 @@
...
+        /// </summary>
+        /// <param name="name">Name of the object property.</param>
+        public async Task WriteNameAsync(string name)
+        {
+            Debug.Assert(!string.IsNullOrEmpty(name), "The name must be specified.");
+            Debug.Assert(this.scopes.Count > 0, "There must be an active scope for name to be written.");
+            Debug.Assert(this.scopes.Peek().Type == ScopeType.Object, "The active scope must be an object scope for name to be written.");
+
+            Scope currentScope = this.scopes.Peek();
+            if (currentScope.ObjectCount != 0)
```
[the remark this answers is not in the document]
gathogojr: @habbes Yes. In this case for example, we're verifying that we're within an object scope when writing a property name.

src/Microsoft.OData.Core/Json/ODataJsonTextWriter.cs
gathogojr: @habbes @KenitoInc Eliding `async` and `await` can lead to unexpected pitfalls/behaviour. Reference to [this](https://blog.stephencleary.com/2016/12/eliding-async-await.html) article plus [Stephen Toub's  article on the cost of `async` and `await`](https://docs.microsoft.com/en-us/archive/msdn-magazine/2011/october/asynchronous-programming-async-performance-understanding-the-costs-of-async-and-await), the state machine for `async` methods will capture exceptions from your code and place them on the returned task. Without the `async` keyword, the exception is raised directly rather than going on the task:
```csharp
public async Task<string> GetWithKeywordsAsync()
{
    string url = /* Something that can throw an exception */;
    return await DownloadStringAsync(url);
}

public Task<string> GetElidingKeywordsAsync()
{
    string url = /* Something that can throw an exception */;
    return DownloadStringAsync(url);
}
```
These methods work exactly the same as long as the calling method does something like this:
```csharp
var result = await GetWithKeywordsAsync();

var result = await GetElidingKeywordsAsync();
```
However, if the method call is separated from the `await`, then the semantics are different:
```csharp
var task = GetWithKeywordsAsync();
var result = await task; // Exception thrown here

var task = GetElidingKeywordsAsync(); // Exception thrown here
var result = await task;
```
Eliding the keywords in this case causes different (and unexpected) exception behavior.

Honestly I'd rather focus on micro optimizations separately after a analyzing the costs carefully plus first determining whether the compiler does those kind of optimizations by itself.
Plus I'd hope to receive views on the same from other reviewers before embarking on those kind of manual interventions.

src/Microsoft.OData.Core/ODataBinaryStreamWriter.cs
gathogojr: @habbes I stumbled upon this method and my thoughts were the same as yours. I did a refactor to make some of the logic shareable between `Write` and `WriteAsync` method.
What I believe is needed is a rewrite taking into consideration the suggestions you have provided. I even left a comment to that effect in the `PrepareByteArray` method where I moved the shareable code.
I'll have to make a judgement whether to create a separate issue for that or make that change in the PR. That said, I don't think this method is in a hot path based on the typical scenarios that this class is applied.

src/Microsoft.OData.Core/Json/JsonWriterAsyncExtensions.cs
```diff
@@ -0,0 +1,273 @@
...
+    /// </summary>
+    internal static class JsonWriterAsyncExtensions
+    {
+        /// <summary>
+        /// Writes the json object value to the <paramref name="jsonWriter"/> asynchronously.
+        /// </summary>
+        /// <param name="jsonWriter">The <see cref="JsonWriter"/> to write to.</param>
+        /// <param name="jsonObjectValue">Writes the given json object value to the underlying json writer.</param>
+        /// <param name="injectPropertyAction">Called when the top-level object is started to possibly inject first property into the object.</param>
+        internal static async Task WriteJsonObjectValueAsync(this IJsonWriterAsync jsonWriter, IDictionary<string, object> jsonObjectValue, Action<IJsonWriterAsync> injectPropertyAction)
```
marabooy, line 31:
```suggestion
        internal static async Task WriteJsonObjectValueAsync(
        this IJsonWriterAsync jsonWriter,
        IDictionary<string, object> jsonObjectValue, 
        Action<IJsonWriterAsync> injectPropertyAction)

```
nit: Wrap params

src/Microsoft.OData.Core/Json/IJsonWriterFactoryAsync.cs
```diff
@@ -0,0 +1,27 @@
...
+    [CLSCompliant(false)]
+    public interface IJsonWriterFactoryAsync
+    {
+        /// <summary>
+        /// Creates a new JSON writer of <see cref="IJsonWriterAsync"/> with support for writing asynchronously.
+        /// </summary>
+        /// <param name="textWriter">Writer to which text needs to be written.</param>
+        /// <param name="isIeee754Compatible">True if it is IEEE754Compatible.</param>
+        /// <returns>The JSON writer created.</returns>
+        IJsonWriterAsync CreateAsyncJsonWriter(TextWriter textWriter, bool isIeee754Compatible);
```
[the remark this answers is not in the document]
gathogojr: @xuzhg I considered that and at one point I had even named the method that way. However, methods suffixed with `Async` usually suggests that a return type is `Task` or `Task<TResult>` while this method just returns a JSON writer that supports writing asynchronously. I looked through the repo and came across methods like [`CreateAsynchronousReader`](https://github.com/OData/odata.net/blob/cc9d86f0069be395a4f07aabe5cad0cec7b9a2cd/src/Microsoft.OData.Core/ODataRawInputContext.cs#L85) and thought I could adopt the similar naming here.

KenitoInc (approved):
LGTM

src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
habbes: Hmm I see what you mean. But here is where my confusion comes from.
`IncreaseIndentationAsync()` and `DecreaseIndentationAsync()` seem to only be called once per "block" (where a block is a series of "statements" that should appear at the same indentation level).

If the indentation method was responsible for writing the spaces, then it would be called every time before writing a property for example, or every time before writing an array element. Is that currently the case? Looking at the usage, it seems that the indentation methods are only called at the beginning and end of an object, and not at the properties in between. I could be wrong. And this could be irrelevant at the end of the day. But could be worth find out.

src/Microsoft.OData.Core/Json/ODataJsonTextWriter.cs
habbes: But in his particular case we are not doing anything that could throw an exception before the await. We're simply fowarding the arguments as they are to an async call, with no intermerdiary checks or preprocessing calls. Also, since this that's called by another async method with will be called using async/await, any exception that it would throw would be caught by the async method that calls it, so it wouldn't end being unhandled. Aside from the optimization concern, I was just wondering why it was necessary to have the await there in the first place.

But if you feel like it's risky to remove it, I have not objections to keeping it there.

src/Microsoft.OData.Core/Json/IJsonWriterFactoryAsync.cs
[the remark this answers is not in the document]
habbes: Maybe renaming to `CreateAsynchronousJsonWriter` could avoid the confusion and bring more consistency with `CreateAsynchronousReader`.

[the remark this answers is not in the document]
habbes: Or maybe it would be too long?

src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
habbes: Maybe it could make sense to measure whether making IncreaseIndentation async vs sync makes a difference. If it doesn't make a difference then I think you could ignore my concern. If it does make a difference, it could be worth looking into because once this method is merged in the public API, it might be hard to change later on if we find that we need to, since the signature would have to change.

habbes: WebAPI JSON responses are indented, maybe we could see what implementation it uses, that could provide insights on how the Increase/DecreaseIndentation methods are meant to achieve. But if in doubt, I think making them async is the safer thing to do.

habbes: Actually, just remembered that they are not indented, it's just that the postman client and browser extensions auto-format the response. My bad.

gathogojr: @habbes 
> indentation methods are only called at the beginning and end of an object, and not at the properties in between
This is indeed the case. The indentation methods are called at the beginning of an array, object and padding function. The JsonWriter in ODL does not apply a newline or indentation before a property is written

src/Microsoft.OData.Core/Json/IJsonWriterFactoryAsync.cs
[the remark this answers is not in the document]
gathogojr: @habbes Done

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -436,20 +436,20 @@ internal static async Task WriteBinaryStringAsync(this TextWriter writer, byte[]
         /// <param name="inputArrayOffset">How many characters to skip in the input array.</param>
         /// <param name="inputArrayCount">How many characters to write from the input array.</param>
         /// <param name="stringEscapeOption">The string escape option.</param>
+        /// <param name="buffer">Char buffer to use for streaming data.</param>
         /// <param name="bufferPool">Character buffer pool.</param>
-        internal static async Task WriteEscapedCharArrayAsync(this TextWriter writer, char[] inputArray, int inputArrayOffset, int inputArrayCount, ODataStringEscapeOption stringEscapeOption, ICharArrayPool bufferPool)
+        internal static async Task WriteEscapedCharArrayAsync(this TextWriter writer, char[] inputArray, int inputArrayOffset,
```
marabooy, line 441:
```suggestion
        internal static async Task WriteEscapedCharArrayAsync(
        this TextWriter writer,
         char[] inputArray, 
         int inputArrayOffset, 

```

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -436,20 +436,20 @@ internal static async Task WriteBinaryStringAsync(this TextWriter writer, byte[]
         /// <param name="inputArrayOffset">How many characters to skip in the input array.</param>
         /// <param name="inputArrayCount">How many characters to write from the input array.</param>
         /// <param name="stringEscapeOption">The string escape option.</param>
+        /// <param name="buffer">Char buffer to use for streaming data.</param>
         /// <param name="bufferPool">Character buffer pool.</param>
-        internal static async Task WriteEscapedCharArrayAsync(this TextWriter writer, char[] inputArray, int inputArrayOffset, int inputArrayCount, ODataStringEscapeOption stringEscapeOption, ICharArrayPool bufferPool)
+        internal static async Task WriteEscapedCharArrayAsync(this TextWriter writer, char[] inputArray, int inputArrayOffset, 
+            int inputArrayCount, ODataStringEscapeOption stringEscapeOption, Ref<char[]> buffer, ICharArrayPool bufferPool)
```
marabooy, line 442:
```suggestion
            int inputArrayCount,
             ODataStringEscapeOption stringEscapeOption,
              Ref<char[]> buffer,
              ICharArrayPool bufferPool)

```

marabooy (approved):

PR #2017, opened 2021-03-03
Issues #2016, #1447: In-operator precedence

### LINKED ISSUE #1447: Error when I try to use "in" logical operator in an "any" lambda operator
Hi!

I'm using OData 4, and I'm trying do this filter:
**$filter=Itens/any(c: c/NotaFiscalID in (8962,8503))**
But it just not work.

When a change te filter to this work pretty good:
**$filter=Itens/any(c: c/NotaFiscalID eq 8962 or c/NotaFiscalID eq 8503)**

The problem is just when I try to use the "in" logical operator.

Thank you!

### LINKED ISSUE #2016: In Operator precedence not respected by unary not operator
`in`-operator precedence does not appear to be taken into account with respect to logical negation (unary `not`-operator).

### Assemblies affected

OData.Net 7.2.1+

### Reproduce steps

Simple.OData.Client was given ...
```
foo.Filter(x => !x.Status.Contains("NotSoGood", "NotEvenBetter"))
```
... and produced the following query string:
```
?$filter=not Status in ('NotSoGood', 'NotEvenBetter')
```

### Expected result

[Reading the operator precedence from the 4.01 spec](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#_Toc31361036), it appears that `in` should have precedence over the unary `not`.

I would expect that this parses into the following pseudo-AST given the spec:
- not
    - in
        - Status
        - ('NotSoGood', 'NotEvenBetter')

### Actual result

From https://github.com/simple-odata-client/Simple.OData.Client/pull/776#issuecomment-788087835:
> The query specified in the URI is not valid. A unary operator with an incompatible type was detected. Found operand type 'Edm.String' for operator kind 'Not'.

Which means the pseudo-AST was instead:
- not
    - Status
    - !Error!

watfordgnf:
### Issues

This pull request fixes #2016.
This pull request fixes #1447.

### Description

This PR lowers the parsing of the `in` and `has` operators to just before the other "primary" operators per [the precedence table listed in the OData 4.01 spec](https://docs.oasis-open.org/odata/odata/v4.01/os/part2-url-conventions/odata-v4.01-os-part2-url-conventions.html#_Toc31361036).

- [X] *Test cases added*

~_NOTE:_ I'm having issues building the E2E project locally and will keep this as a Draft PR until I work that out.~

### Additional work necessary

There is an open issue with this fix in that it introduces a new recursion depth to the code. This could be considered a breaking change for services which use OData.Net with queries that work today but tomorrow will not because they exist right _at_ the recursion limit.

I see two paths forward:
1. Leave this PR as-is and make a possible breaking change note explaining the new recursion depth.
2. Make this PR "pay-for-play" and only charge for recursion depths if an `in` or `has` operator is present.

Option 2 is not what the existing code does, which took me somewhat by surprise. I was unaware that the recursion limits for filters et al were based on the parsing logic rather than the depth of the resultant AST. I experimented with changing the usage of recursion limits to be based on the AST, but that has the side effect of allowing some seriously large parser call stacks when the AST ends up being rather simple.

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Parsers/UriQueryExpressionParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/UriQueryExpressionParser.cs
@@ -980,7 +980,7 @@ namespace Microsoft.OData.UriParser
         }
 
         /// <summary>
-        /// Parses the has, and in operators.
+        /// Parses the has and in operators.
         /// </summary>
         /// <returns>The lexical token representing the expression.</returns>
         private QueryToken ParseInHas()
--- a/src/Microsoft.OData.Core/UriParser/Parsers/UriQueryExpressionParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/UriQueryExpressionParser.cs
@@ -823,7 +823,7 @@ namespace Microsoft.OData.UriParser
         }
 
         /// <summary>
-        /// Parses the eq, ne, lt, gt, le, ge, has, and in operators.
+        /// Parses the eq, ne, lt, gt, le, and ge operators.
         /// </summary>
         /// <returns>The lexical token representing the expression.</returns>
         private QueryToken ParseComparison()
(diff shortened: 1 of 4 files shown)


watfordgnf:
It appears my issue is that I am running the Preview version of VS 2k19 and `build.ps1` makes certain assumptions about file locations. I'm working through modifying them locally to avoid this issue.

watfordgnf:
I'm unable to build the E2E tests at baseline due to the following errors:

1. Build.props has a lot of DefineConstants magic that do not seem to work with .vbproj files (due to ; instead of , as the separator apparently)
2. Build errors in `test\FunctionalTests\Tests\DataServices\UnitTests\ClientUnitTests\ServerDrivenPaging.vb`
    a. `error BC30002: Type 'Order' is not defined.`

I was able to work around 1 by adding the following to the end of Build.props:
```
  <!-- Fixup VBPROJ DefineConstants -->
  <PropertyGroup Condition="'$(MSBuildProjectExtension)' == '.vbproj'">
    <DefineConstants>$(DefineConstants.Replace(';', ','))</DefineConstants>
  </PropertyGroup>
```

I have been unable to determine how to fix the VB projects. I can't get Visual Studio to figure out which reference to use and hard coding it doesn't seem to help either.

src/Microsoft.OData.Core/UriParser/Parsers/UriQueryExpressionParser.cs
```diff
@@ -989,7 +976,39 @@ private QueryToken ParseUnary()
             }
 
             this.RecurseLeave();
-            return this.ParsePrimary();
+            return this.ParseInHas();
+        }
+
+        /// <summary>
+        /// Parses the has, and in operators.
```
gathogojr, line 983:
```suggestion
        /// Parses the has and in operators.
```

KenitoInc (approved):
LGTM

watfordgnf:
As a heads-up, _not a rush_: I will no longer have access to this account post 30 April 2021.

xuzhg (approved):
:shipit:

gathogojr (approved):
:shipit:

KenitoInc:
Thanks for your contribution @watfordgnf

PR #2018, opened 2021-03-03
Make sure we pass the nodollar flag through

ificator:
### Issues

The following does not work when "no dollar" is enabled:
```
select=foo(select=bar) 
```

### Description

`SelectExpandSyntacticParser` creates two instances of `SelectExpandParser` internally... one for parsing `select` and one for parsing `expand`. Unfortunately, while `EnableNoDollarQueryOptions` is passed through for the `expand` case it is NOT for the `select` case.

- [ ] *Test cases added*

### Additional work necessary

N/A

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/SelectExpandParserTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/SelectExpandParserTests.cs
@@ -375,6 +375,68 @@ namespace Microsoft.OData.Tests.UriParser.Parsers
         }
         #endregion
 
+        #region ParseSelectExpand
+        [Fact]
+        public void NoDollarNestedOptions()
+        {
+            // Act
+            SelectExpandParserTests.ParseSelectExpand(
+                select: "foo,bar(select=baz)",
+                expand: "foo,bar(select=baz;expand=qux)",
+                out SelectToken selectToken,
+                out ExpandToken expandToken,
+                enableNoDollarQueryOptions: true);
+
+            // Assert Select
+            Assert.NotNull(selectToken.Properties);
+            PathSegmentToken[] properties = selectToken.Properties.ToArray();
+            Assert.Equal(2, properties.Length);
+            Assert.Equal("foo", properties[0].Identifier);
+            Assert.Equal("bar", properties[1].Identifier);
+
+            Assert.NotNull(selectToken.SelectTerms);
+            SelectTermToken[] selectTerms = selectToken.SelectTerms.ToArray();
+            Assert.Equal(2, selectTerms.Length);
+            Assert.Null(selectTerms[0].SelectOption);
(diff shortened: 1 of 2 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/SelectExpandParserTests.cs
```diff
@@ -375,6 +375,68 @@ private SelectToken ParseSelectClause(string select)
...
+            Assert.Equal(2, expandTerms.Length);
+            Assert.Null(expandTerms[0].ExpandOption);
+            Assert.Null(expandTerms[0].SelectOption);
+            Assert.NotNull(expandTerms[1].ExpandOption);
+            Assert.NotEmpty(expandTerms[1].ExpandOption.ExpandTerms);
+            Assert.NotNull(expandTerms[1].SelectOption);
+            Assert.NotEmpty(expandTerms[1].SelectOption.SelectTerms);
+        }
+
+        private static void ParseSelectExpand(
```
KenitoInc, line 416: This private method can go to the bottom of the file where other private methods are located.

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/SelectExpandParserTests.cs
```diff
@@ -375,6 +375,68 @@ private SelectToken ParseSelectClause(string select)
         }
         #endregion
 
+        #region ParseSelectExpand
+        [Fact]
+        public void NoDollarNestedOptions()
+        {
+            // Act
+            SelectExpandParserTests.ParseSelectExpand(
+                select: "foo,bar(select=baz)",
```
KenitoInc, line 384: Look at other test cases in this file and use property names similar to those.
`foo` and `bar` in both Select and Expand is confusing.
**I understand we dont validate entities and properties in syntactic parsing, but less confusing property and entity names are better**

We can have something like
`select=RelatedSSNs,Addresses(select=City)`
`expand=Trips,Orders(select=OrderDetails; expand=Discounts)`

KenitoInc (changes requested):
A minor change request.
Otherwise LGTM

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/SelectExpandParserTests.cs
ificator: That doesn't appear to be the pattern used in this file though - see `ParseSelectClause` and `StarExpandTesting`. Each region groups public/private methods for the same scenario together. I can move this new one, but it feels like we'll now be inconsistent.

gathogojr (approved):
LGTM

habbes (approved):

KenitoInc (approved):
LGTM

PR #2021, opened 2021-03-05
HttpClientRequestMessage: Fix UnwrapAggregateException.

tfaller:
### Description

UnwrapAggregateException of HttpClientRequestMessage can currently either re-throw incomplete exception informations (by skipping an exception layer) or cause a NPE.

UnwrapAggregateException tries to unwrap an exception two layers deep. This is wrong. I got here a NPE because the exception that was wrapped in the AggregateException had no InnerException. This PR fixes this simple mistake.

### Additional work necessary
none

### DIFF
--- a/src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
+++ b/src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
@@ -470,7 +470,7 @@ namespace Microsoft.OData.Client
             {
                 if (aggregateException.InnerExceptions.Count == 1)
                 {
-                    throw ConvertToDataServiceTransportException(new WebException(aggregateException.InnerException.InnerException.ToString()));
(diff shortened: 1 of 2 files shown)


gathogojr (approved):

marabooy (approved):

KenitoInc (changes requested):
I suggest you add a test to validate this fix.

habbes:
@tfaller hi, I'm following up to check whether you'd be able to add the tests requested by @KenitoInc, or whether you need any additional support.

tfaller:
Hello @KenitoInc and @habbes , I've added a test now. It would be nice if you could pull this soon, because I'm encountering this bug many times in production. Thanks.

KenitoInc (approved):

KenitoInc:
Thanks for this PR @tfaller

PR #2024, opened 2021-03-17
Perf Fix for swallowed exception in AppendKeySegment

Sreejithpin:
### Issues
Firing performance issue in case of odata exception being swallowed. Trying to get keys in case we intend to follow without any exception

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
@@ -288,13 +288,19 @@ namespace Microsoft.OData.Evaluation
             Debug.Assert(resource != null, "resource != null");
             Debug.Assert(propertyKind == ODataPropertyKind.Key || propertyKind == ODataPropertyKind.ETag, "propertyKind == ODataPropertyKind.Key || propertyKind == ODataPropertyKind.ETag");
 
-            KeyValuePair<string, object>[] properties = EmptyProperties;
+            List<KeyValuePair<string, object>> properties = new List<KeyValuePair<string, object>>(); ;
             if (resource.NonComputedProperties != null)
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -2459,22 +2459,81 @@ private ODataPath AppendEntitySetKeySegment(ODataPath odataPath, bool throwIfFai
...
+                    else 
+                    {
+                        KeyValuePair<string, object>[] keys = ODataResourceMetadataContext.GetKeyProperties(resource,
+                            this.GetResourceSerializationInfo(resource), currentEntityType);
+
+                        path = path.AddKeySegment(keys, currentEntityType, this.CurrentScope.NavigationSource);
+                    }                    
                 }
             }
             catch (ODataException)
```
mikepizzo, line 2517: why still catch the exception if we just re-throw it? should just be able to remove the try/catch

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -288,13 +288,19 @@ private static void ValidateEntityTypeHasKeyProperties(KeyValuePair<string, obje
...
+                {
+                    if(property.SerializationInfo != null && property.SerializationInfo.PropertyKind == propertyKind)
+                    {
+                        properties.Add(new KeyValuePair<string, object>(property.Name, GetPrimitiveOrEnumPropertyValue(actualEntityTypeName, property, propertyKind == ODataPropertyKind.Key)));
+                    }
+                }                
             }
 
-            return properties;
+            return properties.ToArray();
```
mikepizzo, line 303: >.ToArray() [](start = 29, length = 10)

How is this used? Is it more efficient to return it as a list, rather than copy to an array?

Sreejithpin: In other methods the returned key property is accessed, and passed down to other methods as array (in most cases) . Thats the reason we returned it as array

---
In reply to: [600680945](https://github.com/OData/odata.net/pull/2024#discussion_r600680945) [](ancestors = 600680945)

mikepizzo (approved):
:shipit:

Issue #2019, opened 2021-03-04
Implement true async in ODL


gathogojr:
ODL currently supports async by creating a task and executing the synchronous method asynchrously on the task.
This has a negative impact for platforms that try to work asynchronously, such as ASP.NET Core 3.1 and greater.
Make the async method calls in ODL truly async down to reading/writing to the stream.

### Assemblies affected
- Microsoft.OData.Core 7.8.3

gathogojr:
Fixed by various PRs

PR #2030, opened 2021-03-23
Implement asynchronous support in ODataJsonLightSerializer

gathogojr:
### Issues

### Description

Implement asynchronous support in `ODataJsonLightSerializer`, `ODataJsonLightValueSerializer`, `ODataJsonLightPropertySerializer` and `JsonLightInstanceAnnotationWriter`

**Disclaimer:**
This PR ended up affecting 5 classes due to dependencies that were not easy to go around **without ending up with an incomplete/non-functional API**. For the following reasons:
- `ODataJsonLightSerializer` class has a hard-coded dependency on `JsonLightInstanceAnnotationWriter` and invokes particular methods to write instance annotations.
- `JsonLightInstanceAnnotationWriter` class has a hard-coded dependency on `ODataJsonLightSerializer` and `ODataJsonLightValueSerializer` classes and invokes methods from both classes in the course of writing instance annotations.
- `ODataJsonLightValueSerializer` class has a hard-coded dependency on `ODataJsonLightPropertySerializer` class
- Both `ODataJsonLightValueSerializer` and `ODataJsonLightPropertySerializer` subclass ODataJsonLightSerializer

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -235,40 +233,231 @@ internal void WriteTopLevelProperty(ODataProperty property)
...
             {
-                this.currentPropertyInfo = new PropertySerializationInfo(this.JsonLightOutputContext.Model, propertyName, owningType) { IsTopLevel = isTopLevel };
+                return;
             }
-            else
+
+            foreach (ODataProperty property in properties)
             {
-                this.currentPropertyInfo = this.JsonLightOutputContext.PropertyCacheHandler.GetProperty(this.JsonLightOutputContext.Model, propertyName, owningType);
+                await this.WritePropertyAsync(
```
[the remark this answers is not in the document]
gathogojr: @marabooy It does change the semantics. Since we're writing to a single stream, we need the first write operation to complete before the second commence

src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
```diff
@@ -382,6 +383,337 @@ public virtual void WriteStreamValue(ODataBinaryStreamValue streamValue)
...
+                    ValidationUtils.ValidateCollectionItem(item, expectedItemTypeReference.IsNullable());
+
+                    ODataResourceValue itemAsResourceValue = item as ODataResourceValue;
+                    if (itemAsResourceValue != null)
+                    {
+                        if (duplicatePropertyNamesChecker == null)
+                        {
+                            duplicatePropertyNamesChecker = this.CreateDuplicatePropertyNameChecker();
+                        }
+
```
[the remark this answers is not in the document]
gathogojr: @marabooy See my response here https://github.com/OData/odata.net/pull/2030#discussion_r610431169

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -235,40 +233,231 @@ internal void WriteTopLevelProperty(ODataProperty property)
...
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        internal Task WriteTopLevelPropertyAsync(ODataProperty property)
+        {
+            Debug.Assert(property != null, "property != null");
+            Debug.Assert(!(property.Value is ODataStreamReferenceValue), "!(property.Value is ODataStreamReferenceValue)");
+
+            return this.WriteTopLevelPayloadAsync(
+                async () =>
+                {
+                    await this.AsynchronousJsonWriter.StartObjectScopeAsync().ConfigureAwait(false);
```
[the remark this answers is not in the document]
gathogojr: @marabooy If you notice, we're eliding the call to `WriteTopLevelPayloadAsync` - since its the only _material_ statement within `WriteTopLevelPropertyAsync` method (i.e. the `WriteTopLevelPropertyAsync` is a pass-through). As for the delegate, we can't elide the await since we have multiple statements that should be executed in chronological order.
[Here](https://github.com/OData/odata.net/pull/2030/files#diff-3b9d6424a2dfe7cd670aec4af41eb01d3962575eef4efe31df83c73562ad0011R352) is another call to `WriteTopLevelPayloadAsync` and you'll notice that because there's only one method called when the delegate is invoked, we're actually eliding the await.

marabooy (approved):
LGTM

src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
```diff
@@ -16,6 +16,7 @@ namespace Microsoft.OData.JsonLight
     using Microsoft.OData.Metadata;
     using Microsoft.OData.Edm;
     using ODataErrorStrings = Microsoft.OData.Strings;
+    using System.Threading.Tasks;
```
xuzhg, line 19: sorting

gathogojr: @xuzhg I'm not sure how I missed that. Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -235,40 +233,231 @@ internal void WriteTopLevelProperty(ODataProperty property)
...
+        }
 
-            if (!this.JsonLightOutputContext.PropertyCacheHandler.InResourceSetScope())
+        /// <summary>
+        /// Asynchronously writes an <see cref="ODataProperty" /> to the given stream. This method creates an
+        /// async buffered stream and writes the property to it.
+        /// </summary>
+        /// <param name="property">The property to write.</param>
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        internal Task WriteTopLevelPropertyAsync(ODataProperty property)
```
xuzhg, line 259: >Task [](start = 17, length = 4)

why no async key word?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightSerializerTests.cs
```diff
@@ -12,6 +12,7 @@
 using Microsoft.OData.JsonLight;
 using Microsoft.OData.Edm;
 using Xunit;
+using System.Threading.Tasks;
```
xuzhg: >System.Threading.Tasks; [](start = 6, length = 23)

sort

xuzhg (approved):
:shipit:

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
gathogojr: @xuzhg Eliding is recommended when a method is a pass-through. See this article on the subject
https://blog.stephencleary.com/2016/12/eliding-async-await.html#recommended-guidelines:~:text=Recommended%20Guidelines
This method qualifies as a pass-through since it merely calls another method to do the job.
Eliding will save us the overhead of creating and later garbage-collecting the state machine.
For that reason, I'm eliding the `async` and `await` whenever the method is just a pass-through

PR #2031, opened 2021-03-23
Implement asynchronous support in ODataJsonWriterUtils

gathogojr:
### Issues

### Description

- Implement asynchronous support in ODataJsonWriterUtils
- Tests for the changes

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/Json/ODataJsonWriterUtilsAsync.cs
```diff
@@ -0,0 +1,274 @@
...
+    {
+        /// <summary>
+        /// Asynchronously writes an error message.
+        /// </summary>
+        /// <param name="jsonWriter">The JSON writer to write the error.</param>
+        /// <param name="writeInstanceAnnotationsDelegate">Delegate to write the instance annotations.</param>
+        /// <param name="error">The error instance to write.</param>
+        /// <param name="includeDebugInformation">A flag indicating whether error details should be written (in debug mode only) or not.</param>
+        /// <param name="maxInnerErrorDepth">The maximum number of nested inner errors to allow.</param>
+        /// <param name="writingJsonLight">true if we're writing JSON lite, false if we're writing verbose JSON.</param>
```
wandeg, line 31: Why not have `verbose` as the param?

wandeg: Also if you change it to `verbose` the condition may need to be flipped around

gathogojr: @wandeg While your suggestion may be valid, I'd lean towards letting the name be so as not to deviate much from the already existing synchronous equivalent defined [here](https://github.com/OData/odata.net/blob/25ea344714d291d169b48ebf71c09721014989aa/src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs#L39).
In addition, `verbose` also doesn't really tell us much about what happens when the value is `true` or `false`.
A user would need to remember that its called one thing when invoking the synchronous method and another when invoking the asynchronous equivalent.
[Here's](https://github.com/OData/odata.net/blob/25ea344714d291d169b48ebf71c09721014989aa/src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs#L202) how the flag is applied:
```csharp
if (writingJsonLight)
{
    Debug.Assert(writeInstanceAnnotationsDelegate != null, "writeInstanceAnnotations != null");
    writeInstanceAnnotationsDelegate(instanceAnnotations);
}
```
When the `writingJsonLight` is true, we invoke a delegate to write instance annotations

src/Microsoft.OData.Core/Json/ODataJsonWriterUtilsAsync.cs
```diff
@@ -0,0 +1,274 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ODataJsonWriterUtilsAsync.cs" company="Microsoft">
```
xuzhg, line 2: >ODataJsonWriterUtilsAsync.cs [](start = 20, length = 28)

since your file name is 'ODataJsonWriterUtilsAsync.cs", might be use the same name for the class name?

src/Microsoft.OData.Core/Json/ODataJsonWriterUtilsAsync.cs
```diff
@@ -0,0 +1,274 @@
...
+
+            // "message": "<message string>"
+            await jsonWriter.WriteNameAsync(JsonConstants.ODataErrorMessageName).ConfigureAwait(false);
+            await jsonWriter.WriteValueAsync(message).ConfigureAwait(false);
+
+            // For example, "target": "query",
+            if (target != null)
+            {
+                await jsonWriter.WriteNameAsync(JsonConstants.ODataErrorTargetName).ConfigureAwait(false);
+                await jsonWriter.WriteValueAsync(target).ConfigureAwait(false);
```
xuzhg, line 144: the old OData error payload is not same as the format defined in the spec. If possible, should we fix it in the new API?

src/Microsoft.OData.Core/Json/ODataJsonWriterUtilsAsync.cs
```diff
@@ -0,0 +1,274 @@
...
+            Func<IEnumerable<ODataInstanceAnnotation>, Task> writeInstanceAnnotationsDelegate, int maxInnerErrorDepth,
+            bool writingJsonLight)
+        {
+            Debug.Assert(jsonWriter != null, "jsonWriter != null");
+            Debug.Assert(code != null, "code != null");
+            Debug.Assert(message != null, "message != null");
+            Debug.Assert(instanceAnnotations != null, "instanceAnnotations != null");
+
+            // "error": {
+            await jsonWriter.StartObjectScopeAsync().ConfigureAwait(false);
```
xuzhg, line 120: >StartObjectScopeAsync [](start = 29, length = 21)

 I am wondering the performance since we have some many await in a function call/

Does each await create a thread? Could you do a research? Could we combine the write today and only call one await?

xuzhg (approved):
:shipit:

src/Microsoft.OData.Core/Json/ODataJsonWriterUtilsAsync.cs
mikepizzo: This is the async part of the partial ODataJsonWriterUtils class -- not a separate class.  Alternatively, we could just add these methods to ODataJsonWriterUtils.cs.

---
In reply to: [600024664](https://github.com/OData/odata.net/pull/2031#discussion_r600024664) [](ancestors = 600024664)

src/Microsoft.OData.Core/Json/ODataJsonWriterUtilsAsync.cs
mikepizzo: For the async methods, can we just remove the verbose option?  That is a carry-over from OData V3 -- new code should never be writing verbose (opportunity for 8.0 clean-up to synchronous methods).

---
In reply to: [599501467](https://github.com/OData/odata.net/pull/2031#discussion_r599501467) [](ancestors = 599501467)

src/Microsoft.OData.Core/Json/ODataJsonWriterUtilsAsync.cs
```diff
@@ -0,0 +1,274 @@
...
+        {
+            Debug.Assert(jsonWriter != null, "jsonWriter != null");
+            Debug.Assert(error != null, "error != null");
+
+            string code, message;
+            ErrorUtils.GetErrorDetails(error, out code, out message);
+
+            ODataInnerError innerError = includeDebugInformation ? error.InnerError : null;
+
+            await WriteErrorAsync(
```
mikepizzo, line 45: >await [](start = 12, length = 5)

pass-through where we can probably elide the await *provided we are confident that error.GetInstanceAnnotations() will never throw an exception*.

mikepizzo:
I would remove the verbose option from the async methods. otherwise, looks good.

mikepizzo (changes requested):
:clock1:

src/Microsoft.OData.Core/Json/ODataJsonWriterUtilsAsync.cs
KenitoInc: @mikepizzo @gathogojr 
We can create an issue for tracking v8.0 cleanup items

gathogojr: @mikepizzo To confirm I understood correctly, we use the `writingJsonLight` flag as a condition to invoke the delegate for writing instance annotations. If we get rid of that flag, we get rid of that delegate as well, right?

src/Microsoft.OData.Core/Json/ODataJsonWriterUtilsAsync.cs
gathogojr: @habbes FYI... This is what we were debating about in my previous PR. We can duck generation of the `async` state machine in quite a few places where the method is a passthrough and relieve the pressure on the GC in particular since that's where the higher cost it. I'll consider a PR to address the scenarios we had identified previously.

habbes: @gathogojr Yeah, I also think we should remove await from pass-through methods

habbes (approved):

src/Microsoft.OData.Core/Json/ODataJsonWriterUtilsAsync.cs
habbes: Aside from eliding, maybe there are situations when they could be combined. Maybe in a future PR we could measure impact and see whether there are opportunities for improvements.

marabooy (approved):

src/Microsoft.OData.Core/Json/ODataJsonWriterUtilsAsync.cs
mikepizzo: We should *always* write as if writingJsonLight was true.

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -356,5 +427,173 @@ private static void ODataResourceValueToString(StringBuilder sb, ODataResourceVa
...
+            int maxInnerErrorDepth)
+        {
+            Debug.Assert(jsonWriter != null, "jsonWriter != null");
+            Debug.Assert(code != null, "code != null");
+            Debug.Assert(message != null, "message != null");
+            Debug.Assert(instanceAnnotations != null, "instanceAnnotations != null");
+
+            // "error": {
+            await jsonWriter.StartObjectScopeAsync().ConfigureAwait(false);
+            await jsonWriter.WriteNameAsync(JsonConstants.ODataErrorName).ConfigureAwait(false);
```
mikepizzo, line 460: >ODataErrorName [](start = 58, length = 14)

This should write the jsonLight format, not the verbose format (i.e., it should use ODataErrorPropertyName).

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -356,5 +427,173 @@ private static void ODataResourceValueToString(StringBuilder sb, ODataResourceVa
...
+                await WriteErrorDetailsAsync(jsonWriter, details, JsonConstants.ODataErrorDetailsName).ConfigureAwait(false);
+            }
+
+            if (innerError != null)
+            {
+                await WriteInnerErrorAsync(jsonWriter, innerError, JsonConstants.ODataErrorInnerErrorName,
+                    /* recursionDepth */ 0, maxInnerErrorDepth).ConfigureAwait(false);
+            }
+
+            // } }
```
mikepizzo, line 495: should write instance annotations

mikepizzo:
Sorry -- my comment was apparently not clear.  We should remove the writingJsonLight flag and treat all methods as if writingJsonLight were true.

A bit of background: The libraries were originally written to support OData V3.  OData V3 had a JSON format that was inherited from OData V2, but then we completely changed the format to what we called "JsonLight" (because it omitted certain control information that could be inferred on the client). When we added this new "JsonLight" format, we renamed the old JSON format to "VerboseJson" (partially to discourage its use).  When we brought OData to the standards body we only defined the new ("JsonLight") format as *the* JSON format for OData V4.  So, any code in our codebase that deals with non-JSON Light (i.e., verbose) is legacy code. We should not bring such legacy code forward, and should remove the legacy code in our next (8.0) breaking change release.

So, for this change, what I was suggesting was that we remove the writingJsonLight flag and treat the code as if writingJsonLight were always true -- that is, we use the ErrorPropertyName and write annotations.

---
In reply to: [805945601](https://github.com/OData/odata.net/pull/2031#issuecomment-805945601) [](ancestors = 805945601)

gathogojr:
> Sorry -- my comment was apparently not clear. We should remove the writingJsonLight flag and treat all methods as if writingJsonLight were true.
> 
> A bit of background: The libraries were originally written to support OData V3. OData V3 had a JSON format that was inherited from OData V2, but then we completely changed the format to what we called "JsonLight" (because it omitted certain control information that could be inferred on the client). When we added this new "JsonLight" format, we renamed the old JSON format to "VerboseJson" (partially to discourage its use). When we brought OData to the standards body we only defined the new ("JsonLight") format as _the_ JSON format for OData V4. So, any code in our codebase that deals with non-JSON Light (i.e., verbose) is legacy code. We should not bring such legacy code forward, and should remove the legacy code in our next (8.0) breaking change release.
> 
> So, for this change, what I was suggesting was that we remove the writingJsonLight flag and treat the code as if writingJsonLight were always true -- that is, we use the ErrorPropertyName and write annotations.
> 
> In reply to: [805945601](https://github.com/OData/odata.net/pull/2031#issuecomment-805945601) [](ancestors = 805945601)

@mikepizzo Noted. I'll fix that

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -356,5 +430,183 @@ private static void ODataResourceValueToString(StringBuilder sb, ODataResourceVa
...
+                await WriteErrorDetailsAsync(jsonWriter, details, JsonConstants.ODataErrorDetailsName).ConfigureAwait(false);
+            }
+
+            if (innerError != null)
+            {
+                await WriteInnerErrorAsync(jsonWriter, innerError, JsonConstants.ODataErrorInnerErrorName,
+                    /* recursionDepth */ 0, maxInnerErrorDepth).ConfigureAwait(false);
+            }
+
+            if (instanceAnnotations.Any())
```
mikepizzo, line 500: >instanceAnnotations.Any() [](start = 16, length = 25)

probably better if writeInstanceAnnotationsDelegate handles the empty case so that you don't have to use the LINQ Any() method to check for instance annotations before invoking the writeInstanceAnnotationsDelegate (you would need to make the check for writeInstanceAnnotationsDelegate a real check, not just an assertion).

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -356,5 +430,183 @@ private static void ODataResourceValueToString(StringBuilder sb, ODataResourceVa
...
+                await jsonWriter.WriteValueAsync(target).ConfigureAwait(false);
+            }
+
+            // Such as, "details": [
+            //  {
+            //   "code": "301",
+            //   "target": "$search",
+            //   "message": "$search query option not supported"
+            //  }]
+            if (details?.Any() == true)
```
mikepizzo, line 489: >?.Any() == true [](start = 23, length = 15)

probably better to handle empty details in WriteErrorDetailsAsync, rather than incurring cost of the LINQ Any() method prior to calling.

gathogojr: @mikepizzo I had thought that it'd be less expensive to call Any() method instead of calling await since that in turn would trigger creation of the state machine and the even more expensive GC clean up? What are your thoughts on that?

mikepizzo: Partially depends on how details (and instanceAnnotations) are constructed.  details are probably pretty static and easy to enumerate, but instanceAnnotations may be expensive to enumerate (i.e., if the enumerable is a filter over a broader set of instance annotations).

---
In reply to: [603521432](https://github.com/OData/odata.net/pull/2031#discussion_r603521432) [](ancestors = 603521432)

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
gathogojr: @mikepizzo To handle the empty instanceAnnotations in the async invocation of the delegate, we'd have to require the delegate to always be non-null. Which is not a problem since we have [`WriteInstanceAnnotationsForError` method](https://github.com/OData/odata.net/blob/cc9d86f0069be395a4f07aabe5cad0cec7b9a2cd/src/Microsoft.OData.Core/Json/JsonLightInstanceAnnotationWriter.cs#L125) that we can pass as the delegate - defined in `JsonLightAnnotationWriter`.
If we're on the same page on that, I'll change the logic such that we throw an exception if the delegate is `null`, firstly from the `internal WriteErrorAsync` method - so we don't bother to call the `private WriteErrorAsync`. I'll add the same check in the `private` method just in case it was triggered from another method

Issue #2028, opened 2021-03-19
Nullable attribute missing for collection type reference.


chrisspre:
In [section 7.2.1 Nullable](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_Nullable) the standard states that 

> In OData 4.01 responses a collection-valued property MUST specify a value for the Nullable attribute.
> 
> If no value is specified for a collection-valued property, the client cannot assume any default value. Clients SHOULD be prepared for this situation even in OData 4.01 responses.

but the EDM library leaves out the Nullable attribute, in all cases. 

### Assemblies affected

Microsoft.OData.Edm -Version 7.8.3

### Reproduce steps

*The simplest set of steps to reproduce the issue. If possible, reference a commit that demonstrates the issue.*
The C# code 
```CSharp

      @type.AddProperty(new EdmStructuralProperty(@type, "f",
          new EdmCollectionTypeReference(new EdmCollectionType(
             new EdmDecimalTypeReference(@decimal, true)))));
``` 

produces CSDL of the form 
```XML
<Property Name="f" Type="Collection(Edm.Decimal)" />
```
that is missing the explicit Nullable attribute as stated in the standard

### Expected result

```XML
<Property Name="f" Type="Collection(Edm.Decimal)" />
```

### Actual result

```XML
<Property Name="f" Type="Collection(Edm.Decimal)" Nullable="true"/>
```

habbes:
@chrisspre I think you mixed up the "expected" and "actual" results in the issue description?

chrisspre:
> @chrisspre I think you mixed up the "expected" and "actual" results in the issue description?

Yes, sorry. fixed it.

xuzhg:
If the type is collection and nullable of element is true, we should write the nullable="true".

chrisspre:
will be addressed in https://github.com/OData/odata.net/pull/2034

PR #2034, opened 2021-03-23
add parameter to control if Nullable gets written always

chrisspre:
### Issues

*This pull request fixes issue #2028 

### Description

Added parameter to WriteNullableAttribute to indicate if the property should be written always or depending on the default value.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaJsonWriter.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaJsonWriter.cs
@@ -384,10 +384,18 @@ namespace Microsoft.OData.Edm.Csdl.Serialization
         /// 7.2.1 Nullable, A Boolean value specifying whether a value is required for the property.
         /// </summary>
         /// <param name="reference">The Edm type reference.</param>
-        internal override void WriteNullableAttribute(IEdmTypeReference reference)
+        /// <param name="alwaysWrite">Specifies if the attribute is always written or not if the value is the default value.
+        internal override void WriteNullableAttribute(IEdmTypeReference reference, bool alwaysWrite)
         {
             // The value of $Nullable is one of the Boolean literals true or false. Absence of the member means false.
-            this.jsonWriter.WriteOptionalProperty("$Nullable", reference.IsNullable, defaultValue: false);
+            if(alwaysWrite)
+            {
+                this.jsonWriter.WriteRequiredProperty("$Nullable", reference.IsNullable);
+            }
+            else
+            {
+                this.jsonWriter.WriteOptionalProperty("$Nullable", reference.IsNullable, defaultValue: false);
+            }
         }
 
         internal override void WriteTypeDefinitionAttributes(IEdmTypeDefinitionReference reference)
--- a/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
@@ -668,12 +668,12 @@ namespace Microsoft.OData.Edm.Csdl.Serialization
                     if (element.TypeKind() == EdmTypeKind.Collection)
                     {
                         IEdmCollectionTypeReference collectionElement = element.AsCollection();
-                        this.schemaWriter.WriteNullableAttribute(collectionElement.CollectionDefinition().ElementType);
+                        this.schemaWriter.WriteNullableAttribute(collectionElement.CollectionDefinition().ElementType, true);
                         VisitTypeReference(collectionElement.CollectionDefinition().ElementType);
                     }
                     else
                     {
-                        this.schemaWriter.WriteNullableAttribute(element);
+                        this.schemaWriter.WriteNullableAttribute(element, false);
                         VisitTypeReference(element);
                     }
                 }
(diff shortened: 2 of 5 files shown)


test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
```diff
@@ -311,7 +311,8 @@ public void WriteNavigationPropertyInComplexType()
       },
       ""Addresses"": {
         ""$Collection"": true,
-        ""$Type"": ""DefaultNs.Address""
+        ""$Type"": ""DefaultNs.Address"",
```
mikepizzo, line 314: Nullable facet is required for collections in XML, but I don't think it's required in JSON.

Which means our factoring is a bit off -- we shouldn't have code in the serializer to determine whether the null needs to be written. Instead, we should pass down the information that lets the XML serializer decide whether or not to write the null value.

mikepizzo (changes requested):
This PR makes the nullable attribute required for collections in both XML and JSON.  However, I think the nullable attribute requirement only applies to XML -- I believe it is valid in JSON to have a collection valued property that does not specify the nullable attribute.

This means that the factoring is a bit off. Rather than deciding in the common serialization classes whether the nullable attribute is required, we should pass down the information required for the xml serializer to decide whether or not the null should be written.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
```diff
@@ -653,7 +656,7 @@ private void ProcessReferentialConstraint(IEdmReferentialConstraint element)
             }
         }
 
-        private void ProcessFacets(IEdmTypeReference element, bool inlineType)
+        private void ProcessFacets(IEdmTypeReference element, bool inlineType, bool nullableAttributeRequiredForCollection)
```
mikepizzo, line 659: >nullableAttributeRequiredForCollection [](http://example.com/codeflow?start=84&length=38)

So this gets set true for structural property facets, and false otherwise?  Can we just call this something like "isStructuralPropertyDeclaration"?

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaJsonWriter.cs
```diff
@@ -384,10 +384,18 @@ internal override void WriteEnumMemberElementHeader(IEdmEnumMember member)
         /// 7.2.1 Nullable, A Boolean value specifying whether a value is required for the property.
         /// </summary>
         /// <param name="reference">The Edm type reference.</param>
-        internal override void WriteNullableAttribute(IEdmTypeReference reference)
+        /// <param name="alwaysWrite">Specifies if the attribute is always written or not if the value is the default value.
+        internal override void WriteNullableAttribute(IEdmTypeReference reference, bool alwaysWrite)
```
habbes, line 388: When is the `alwaysWrite` parameter ever set to true and why? what determines whether it should be set true or false?

chrisspre: It gets called from EdmModelCsdlSerializationVisitor.ProcessFacets 

And that method tests if the property is of a collection type and if the serializer wants to always serialize the nullable attributes for collection (which is true for XML, false for JSON)

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
chrisspre: the call "chain" is 
    EdmModelCsdlSerializationVisitor.{ProcessStructuralProperty|ProcessTerm|ProcessCollectionType|...)
    EdmModelCsdlSerializationVisitor.ProcessFacets -> 
    EdmModelCsdlSchemaWriter.WriteNullableAttribute 
    
and  EdmModelCsdlSchemaWriter is abstract (one concrete subclass for XML, one for JSON)
    
And yes, it is true that only ProcessStructuralProperty is passing in true. But that doesn't describe the behavior of ProcessFacets which is : instruct the writer to write the property for a collection type (-reference) even if it is false (which otherwise dosn't get written). The name is based on what the method does, not on how and when it is called. 

But I don't have a strong opinion. Happy to change it if you prefer.

mikepizzo: Let's discuss --  my concern is that, as I understand it, a component that should be format-agnostic (EdmModelCsdlSerializationVistor) would be making the decision of whether to write the nullable attribute or not based on the format (write for xml, not for json).  ideally, we wouldn't be making format-specific decisions in these common classes, they would all be made in the format-specific implementations.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/Serialization/EdmModelCsdlSerializationVisitorTests.cs
```diff
@@ -2015,6 +2016,53 @@ public void VerifyAnnotationWithRecordExpressionWrittenCorrectly()
...
+    ""$Kind"": ""ComplexType"",
+    ""test01"": {
+      ""$Collection"": true,
+      ""$Type"": ""Edm.Decimal"",
+      ""$Nullable"": true
+    },
+    ""test02"": {
+      ""$Collection"": true,
+      ""$Type"": ""Edm.Decimal"",
+      ""$Nullable"": false
```
mikepizzo, line 2052: shouldn't be written in this case for json csdl, right?

Issue #1997, opened 2021-02-11
IN operator not working with null guid properties


jonyeezs:
It doesn't detect that `null` is a valid value for a nullable guid. It throws a `System.FormatException` expecting a proper guid

### Assemblies affected

OData .Net lib 7.8.1

### Reproduce steps

sample Model

```csharp
public class TestEntity
{
   public guid? NullableProperty { get; set; }
}
```

Sample query

```url
http://host/service/TestEntity?$filter=NullableProperty in (null,2fe0f66a-da8c-452e-8310-95690e10b454)
```

### Expected result

The list of the TestEntity with guid `2fe0f66a-da8c-452e-8310-95690e10b454` and null (`00000000-0000-0000-0000-000000000000`)

### Actual result

```
Message: 
    Microsoft.OData.ODataException : Cannot convert the literal 'null' to the expected type 'Edm.Guid'.
      ----> System.FormatException : Guid should contain 32 digits with 4 dashes (xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx).
```

```
  Stack Trace: 
    ODataPayloadValueConverter.ConvertFromPayloadValue(Object value, IEdmTypeReference edmTypeReference)
    ODataJsonLightReaderUtils.ConvertValue(Object value, IEdmPrimitiveTypeReference primitiveTypeReference, ODataMessageReaderSettings messageReaderSettings, Boolean validateNullValue, String propertyName, ODataPayloadValueConverter converter)
    ODataJsonLightPropertyAndValueDeserializer.ReadPrimitiveValue(Boolean insideJsonObjectValue, IEdmPrimitiveTypeReference expectedValueTypeReference, Boolean validateNullValue, String propertyName)
    ODataJsonLightPropertyAndValueDeserializer.ReadNonEntityValueImplementation(String payloadTypeName, IEdmTypeReference expectedTypeReference, PropertyAndAnnotationCollector propertyAndAnnotationCollector, CollectionWithoutExpectedTypeValidator collectionValidator, Boolean validateNullValue, Boolean isTopLevelPropertyValue, Boolean insideResourceValue, String propertyName, Nullable`1 isDynamicProperty)
    ODataJsonLightPropertyAndValueDeserializer.ReadCollectionValue(IEdmCollectionTypeReference collectionValueTypeReference, String payloadTypeName, ODataTypeAnnotation typeAnnotation)
[...]

PR #2035, opened 2021-03-24
in operator not working with null Guid values

ElizabethOkerio:
### Issues

*This pull request fixes issue #1997 .*

### Description

For a sample model like the one shown below

```csharp
public class TestEntity
{
   public guid? NullableProperty { get; set; }
}
```
Running the following sample query throws an error. 
It is also been clear that a null GUID should not be equated to an empty GUID. 
```url
http://host/service/TestEntity?$filter=NullableProperty in (null)
```
Guid literals are converted to single-quoted form, so that they are  compatible with the Json reader used for deserialization.
``` 
Sample: [D01663CF-EB21-4A0E-88E0-361C10ACE7FD, 492CF54A-84C9-490C-A7A4-B5010FAD8104]
==>  ['D01663CF-EB21-4A0E-88E0-361C10ACE7FD', '492CF54A-84C9-490C-A7A4-B5010FAD8104']
```
This means that even null GUIDS will be converted to the single quoted form which will later lead to them being processed as null strings. ("null"). 
``` 
Sample: [null, null]
==>  ['null', 'null']
```
This fix ensures that null GUIDs are not converted to the single-quoted form and are processed as null values. 
``` 
Sample: [null] ==>  [null]
```

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -313,6 +313,7 @@ namespace Microsoft.OData.UriParser
                 {
                     items[i] = String.Format(CultureInfo.InvariantCulture, "'{0}'", items[i]);
                 }
+
             }
 
             return "[" + String.Join(",", items) + "]";
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -305,15 +305,10 @@ namespace Microsoft.OData.UriParser
 
             for (int i = 0; i < items.Length; i++)
             {
-                if (items[i] == "null")
-                {
-                    items[i] = items[i];
-                }
-                else if(items[i][0] != '\'' && items[i][0] != '"')
+                if (items[i] != "null" && items[i][0] != '\'' && items[i][0] != '"')
                 {
                     items[i] = String.Format(CultureInfo.InvariantCulture, "'{0}'", items[i]);
                 }
-
             }
 
             return "[" + String.Join(",", items) + "]";
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -305,7 +305,11 @@ namespace Microsoft.OData.UriParser
 
             for (int i = 0; i < items.Length; i++)
             {
-                if (items[i][0] != '\'' && items[i][0] != '"')
+                if (items[i] == "null")
+                {
+                    items[i] = items[i];
+                }
+                else if(items[i][0] != '\'' && items[i][0] != '"')
                 {
                     items[i] = String.Format(CultureInfo.InvariantCulture, "'{0}'", items[i]);
                 }
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
@@ -2345,12 +2345,12 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
         [Fact]
         public void FilterWithInOperationWithGuidCollection()
         {
-            FilterClause filter = ParseFilter("MyGuid in (D01663CF-EB21-4A0E-88E0-361C10ACE7FD, 492CF54A-84C9-490C-A7A4-B5010FAD8104)",
+            FilterClause filter = ParseFilter("MyGuid in (D01663CF-EB21-4A0E-88E0-361C10ACE7FD, 492CF54A-84C9-490C-A7A4-B5010FAD8104, null)",
                 HardCodedTestModel.TestModel, HardCodedTestModel.GetPersonType());
 
             var inNode = Assert.IsType<InNode>(filter.Expression);
             Assert.Equal("MyGuid", Assert.IsType<SingleValuePropertyAccessNode>(inNode.Left).Property.Name);
-            Assert.Equal("(D01663CF-EB21-4A0E-88E0-361C10ACE7FD, 492CF54A-84C9-490C-A7A4-B5010FAD8104)",
+            Assert.Equal("(D01663CF-EB21-4A0E-88E0-361C10ACE7FD, 492CF54A-84C9-490C-A7A4-B5010FAD8104, null)",
                 Assert.IsType<CollectionConstantNode>(inNode.Right).LiteralText);
         }
 


src/Microsoft.OData.Core/ODataPayloadValueConverter.cs
```diff
@@ -152,6 +152,10 @@ private static object ConvertStringValue(string stringValue, Type targetType)
 
             if (targetType == typeof(Guid))
             {
+                if (stringValue == "null")
+                {
+                    return Guid.Empty;
```
habbes, line 157: Out of curiosity, is Guid.Empty equivalent to it being null?

If you had class like 
```c#
class SomeEntity {
   int Id { get;set; }
   guid? SomeGuid {get;set}
}
```
if you set an object of this type like this
```
var entity = new SomeEntity();
entity.SomeGuid = null;
```
If you serialize and deserialize this, what value do we expect to have for entity.SomeGuid? null or Guid.Empty?

I assume if a field type is explicitly marked as nullable, then it would make sense to support null.

For nullable `int?` type, I assume 0 != null,

habbes:
Could you add test cases for this?

src/Microsoft.OData.Core/ODataPayloadValueConverter.cs
ElizabethOkerio: This is a very good discussion you've started.  Also had some questions around this.

ElizabethOkerio: @habbes  A Guid is a value type so it can't be null. You can  only check if its empty

habbes: But in this case it's a nullable Guid, i.e. the type is `guid?` not jus a plan `guid`. A nullable value type can have `null` as the value: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/builtin-types/nullable-value-types Maybe just to confirm, you could verify how we handle nullable ints?

ElizabethOkerio: @xuzhg @mikepizzo.

xuzhg: typically, it should return "return (Guid?)null";

But, my concern is, targetType is "Guid", not "Guid?"?

If that's the case, any other type has the similar problem? for example, TimeSpan?, Date?, etc.

So, why should we change the code here?  Shall we do the "null" check at other place?

---
In reply to: [603236521](https://github.com/OData/odata.net/pull/2035#discussion_r603236521) [](ancestors = 603236521)

ElizabethOkerio: @xuzhg this means running this query http://host/service/TestEntity?$filter=NullableProperty in (null) should not return entities with empty Guids but with null guids? right?

odero: Curious, how does one filter for entities with empty Guids?
`?$filter=SomeProperty eq '00000000-0000-0000-0000-000000000000'` or is there some other way?

ElizabethOkerio: @xuzhg  filtering using in operator with null  on nullable datetime, date etc works okay.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -305,9 +305,16 @@ private static string NormalizeGuidCollectionItems(string bracketLiteralText)
...
-                    items[i] = String.Format(CultureInfo.InvariantCulture, "'{0}'", items[i]);
+                    items[i] = String.Format(CultureInfo.InvariantCulture, "{0}", items[i]);
+                }
+                else
+                {
+                    if (items[i][0] != '\'' && items[i][0] != '"')
+                    {
+                        items[i] = String.Format(CultureInfo.InvariantCulture, "'{0}'", items[i]);
+                    }
                 }
```
gathogojr, line 318:
```suggestion
                else if (items[i][0] != '\'' && items[i][0] != '"')
                {
                    items[i] = String.Format(CultureInfo.InvariantCulture, "'{0}'", items[i]);
                }

```

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -305,9 +305,16 @@ private static string NormalizeGuidCollectionItems(string bracketLiteralText)
 
             for (int i = 0; i < items.Length; i++)
             {
-                if (items[i][0] != '\'' && items[i][0] != '"')
+                if (items[i] == "null")
                 {
-                    items[i] = String.Format(CultureInfo.InvariantCulture, "'{0}'", items[i]);
+                    items[i] = String.Format(CultureInfo.InvariantCulture, "{0}", items[i]);
```
gathogojr, line 310: Do you need to call `String.Format` if `items[i]` has a value of `null`? 
The above statement evaluates to:
```csharp
items[i] = items[i] = String.Format(CultureInfo.InvariantCulture, "{0}", "null");
```
and the value of `items[i]` will still be `"null"` after its executed.

gathogojr:
Please add tests

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
ElizabethOkerio: this is a part of the change..what will be returned will be "[null]" as opposed to "['null']"

gathogojr: @ElizabethOkerio In that case maybe the statement should be:
```csharp
items[i] = null;
```
?
All the more reason why you need some tests...

gathogojr: Not a 100% sure though...

ElizabethOkerio: @gathogojr  before this change, this method was transforming null GUIDS to: 
``` 
" ['null']"
```
Which led to this NULL GUIDS being read as "null" in subsequent operations: 

This change transforms a null GUID to:
``` 
" [null]"
```
which will be read as a null value in subsequent operations:

gathogojr: Can you have null and something else? E.g. [null,'844f0fb2-580d-4842-b965-bedc341736b3'] for example?

ElizabethOkerio: Yes..

gathogojr (approved):
Looks good to me but I have to wonder if the issue could have been resolved by changing this:
```csharp
if (items[i][0] != '\'' && items[i][0] != '"')
```
to
```csharp
if (items[i] != "null" && items[i][0] != '\'' && items[i][0] != '"')
```
Kindly confirm if that would do the job

xuzhg (approved):
:shipit:

ElizabethOkerio:
> Looks good to me but I have to wonder if the issue could have been resolved by changing this:
> 
> ```cs
> if (items[i][0] != '\'' && items[i][0] != '"')
> ```
> 
> to
> 
> ```cs
> if (items[i] != "null" && items[i][0] != '\'' && items[i][0] != '"')
> ```
> 
> Kindly confirm if that would do the job

We still need the else part:
 ```cs
 if (items[i] != "null" && items[i][0] != '\'' && items[i][0] != '"')
{
}
else if(items[i] == "null")
{

} 
 ```

Because we need to assign the items[i]. And for the check we have to specify that it is a null coz it could be anything else that does not meet the first criteria.

gathogojr:
> > Looks good to me but I have to wonder if the issue could have been resolved by changing this:
> > ```cs
> > if (items[i][0] != '\'' && items[i][0] != '"')
> > ```
> > 
> > 
> > to
> > ```cs
> > if (items[i] != "null" && items[i][0] != '\'' && items[i][0] != '"')
> > ```
> > 
> > 
> > Kindly confirm if that would do the job
> 
> We still need the else part:
> 
> ```cs
> if (items[i] != "null" && items[i][0] != '\'' && items[i][0] != '"')
> {
> }
> else if(items[i] == "null")
> {
> 
> } 
> ```
> 
> Because we need to assign the items[i]. And for the check we have to specify that it is a null coz it could be anything else that does not meet the first criteria.

Maybe I don't understand, but why would we need the else part? Assigning `items[i]` to `items[i]` is the same as leaving `items[i]` untouched and that would be accomplished by the check `if (items[i] != "null"...`, no?
I may not have full visibility though so I could be missing something

ElizabethOkerio:
> > > Looks good to me but I have to wonder if the issue could have been resolved by changing this:
> > > ```cs
> > > if (items[i][0] != '\'' && items[i][0] != '"')
> > > ```
> > > 
> > > 
> > > to
> > > ```cs
> > > if (items[i] != "null" && items[i][0] != '\'' && items[i][0] != '"')
> > > ```
> > > 
> > > 
> > > Kindly confirm if that would do the job
> > 
> > 
> > We still need the else part:
> > ```cs
> > if (items[i] != "null" && items[i][0] != '\'' && items[i][0] != '"')
> > {
> > }
> > else if(items[i] == "null")
> > {
> > 
> > } 
> > ```
> > 
> > 
> > Because we need to assign the items[i]. And for the check we have to specify that it is a null coz it could be anything else that does not meet the first criteria.
> 
> Maybe I don't understand, but why would we need the else part? Assigning `items[i]` to `items[i]` is the same as leaving `items[i]` untouched and that would be accomplished by the check `if (items[i] != "null"...`, no?
> I may not have full visibility though so I could be missing something

Yes I get

PR #2036, opened 2021-03-25
Add ComputedDefaultValue to CoreVocabularyModel

chrisspre:
### Issues

ComputedDefaultValue is missing from to CoreVocabularyModel

### Description

Changed corresponding XML, .cs and constants fule. 

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Edm/Vocabularies/CoreVocabularyConstants.cs
+++ b/src/Microsoft.OData.Edm/Vocabularies/CoreVocabularyConstants.cs
@@ -44,6 +44,9 @@ namespace Microsoft.OData.Edm.Vocabularies.V1
         /// <summary>Org.OData.Core.V1.Computed</summary>
         public const string Computed = "Org.OData.Core.V1.Computed";
 
+        /// <summary>Org.OData.Core.V1.Computed</summary>
+        public const string ComputedDefaultValue = "Org.OData.Core.V1.ComputedDefaultValue";
+
         /// <summary>Org.OData.Core.V1.IsURL</summary>
         public const string IsURL = "Org.OData.Core.V1.IsURL";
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.OData.Edm/Vocabularies/CoreVocabularyConstants.cs
```diff
@@ -44,6 +44,9 @@ public static class CoreVocabularyConstants
         /// <summary>Org.OData.Core.V1.Computed</summary>
         public const string Computed = "Org.OData.Core.V1.Computed";
 
+        /// <summary>Org.OData.Core.V1.Computed</summary>
```
xuzhg, line 47: >Org.OData.Core.V1.Computed [](start = 21, length = 26)

update the summary.

xuzhg:
>         public static readonly IEdmTerm OptionalParameterTerm = VocabularyModelProvider.CoreModel.FindDeclaredTerm(CoreVocabularyConstants.OptionalParameter);

Actually, i don't like to have these static members declared in this class.

1) It can't cover all
2) It's time consuming



---
Refers to: src/Microsoft.OData.Edm/Vocabularies/CoreVocabularyModel.cs:98 in c845528. [](commit_id = c845528d1cbf310d49b75245fef83c3940cedc0d, deletion_comment = False)

xuzhg (approved):
:shipit:

src/Microsoft.OData.Edm/Vocabularies/CoreVocabularyConstants.cs
chrisspre: fixed

chrisspre:
> > ```
> >     public static readonly IEdmTerm OptionalParameterTerm = VocabularyModelProvider.CoreModel.FindDeclaredTerm(CoreVocabularyConstants.OptionalParameter);
> > ```
> 
> Actually, i don't like to have these static members declared in this class.
> 
> 1. It can't cover all
> 2. It's time consuming
> 
ata/odata.net/commit/c845528d1cbf310d49b75245fef83c3940cedc0d), deletion_comment = False)

I understand. Nonetheless:
- In the context of this PR this is a large but mechanical change, that deserves a seperate PR. 
- Secondly removing them is a breaking change. that Should be in a major release
-  it is actually quite convenient for a user of the SDK since it give some form of IntelliSense. We should look into either generating them of find another way to expose this in the SDK.

I suggest to Not making a change now. and will add an Issue.

PR #2032, opened 2021-03-23
Implement asynchronous support in ODataJsonLightWriterUtils

gathogojr:
### Issues

This pull request is a partial fulfilment of issue #2019.

### Description

- Implement asynchronous support in ODataJsonLightWriterUtils
- Tests for the changes

### Checklist (Uncheck if it is not completed)

- [x] *Test cases added*
- [x] *Build and test with one-click build and test script passed*

### Additional work necessary

*If documentation update is needed, please add "Docs Needed" label to the issue and provide details about the required document change in the issue.*

PR #2038, opened 2021-03-29
Implement asynchronous support in JsonLightODataAnnotationWriter

gathogojr:
### Issues

### Description

- Implement asynchronous support in JsonLightODataAnnotationWriter
- Tests for the changes

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/JsonLightODataAnnotationWriter.cs
+++ b/src/Microsoft.OData.Core/JsonLight/JsonLightODataAnnotationWriter.cs
@@ -81,6 +81,7 @@ namespace Microsoft.OData.JsonLight
         /// <param name="writeRawValue">Whether to write the raw typeName without removing/adding prefix 'Edm.'/'#'.</param>
         public void WriteODataTypeInstanceAnnotation(string typeName, bool writeRawValue = false)
         {
+            this.AssertSynchronous();
             Debug.Assert(typeName != null, "typeName != null");
 
             // "@odata.type": "#typename"
@@ -102,6 +103,7 @@ namespace Microsoft.OData.JsonLight
         /// <param name="typeName">The type name to write.</param>
         public void WriteODataTypePropertyAnnotation(string propertyName, string typeName)
         {
+            this.AssertSynchronous();
             Debug.Assert(!string.IsNullOrEmpty(propertyName), "!string.IsNullOrEmpty(propertyName)");
             Debug.Assert(typeName != null, "typeName != null");
 
@@ -117,7 +119,9 @@ namespace Microsoft.OData.JsonLight
         /// <param name="annotationName">The name of the annotation to write.</param>
         public void WritePropertyAnnotationName(string propertyName, string annotationName)
         {
+            this.AssertSynchronous();
             Debug.Assert(!string.IsNullOrEmpty(propertyName), "!string.IsNullOrEmpty(propertyName)");
+            Debug.Assert(!string.IsNullOrEmpty(annotationName), "!string.IsNullOrEmpty(annotationName)");
             Debug.Assert(annotationName.StartsWith(JsonLightConstants.ODataAnnotationNamespacePrefix,
                 StringComparison.Ordinal), "annotationName.StartsWith(\"odata.\")");
 
@@ -130,6 +134,8 @@ namespace Microsoft.OData.JsonLight
         /// <param name="annotationName">The name of the instance annotation to write.</param>
         public void WriteInstanceAnnotationName(string annotationName)
         {
+            this.AssertSynchronous();
+            Debug.Assert(!string.IsNullOrEmpty(annotationName), "!string.IsNullOrEmpty(annotationName)");
             Debug.Assert(annotationName.StartsWith(JsonLightConstants.ODataAnnotationNamespacePrefix,
                 StringComparison.Ordinal), "annotationName.StartsWith(\"odata.\")");
 
@@ -191,6 +197,7 @@ namespace Microsoft.OData.JsonLight
         {
             this.AssertAsynchronous();
             Debug.Assert(!string.IsNullOrEmpty(propertyName), "!string.IsNullOrEmpty(propertyName)");
+            Debug.Assert(!string.IsNullOrEmpty(annotationName), "!string.IsNullOrEmpty(annotationName)");
             Debug.Assert(annotationName.StartsWith(JsonLightConstants.ODataAnnotationNamespacePrefix,
                 StringComparison.Ordinal), "annotationName.StartsWith(\"odata.\")");
 
@@ -205,6 +212,7 @@ namespace Microsoft.OData.JsonLight
         public Task WriteInstanceAnnotationNameAsync(string annotationName)
         {
             this.AssertAsynchronous();
+            Debug.Assert(!string.IsNullOrEmpty(annotationName), "!string.IsNullOrEmpty(annotationName)");
             Debug.Assert(annotationName.StartsWith(JsonLightConstants.ODataAnnotationNamespacePrefix,
                 StringComparison.Ordinal), "annotationName.StartsWith(\"odata.\")");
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/JsonLight/JsonLightODataAnnotationWriter.cs
```diff
@@ -115,6 +136,81 @@ public void WriteInstanceAnnotationName(string annotationName)
...
+                .ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Asynchronously writes a JSON property name which represents a property annotation.
+        /// </summary>
+        /// <param name="propertyName">The name of the property to annotate.</param>
+        /// <param name="annotationName">The name of the annotation to write.</param>
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        public Task WritePropertyAnnotationNameAsync(string propertyName, string annotationName)
```
xuzhg, line 190: >ublic Task [](start = 9, length = 10)

why no async ?

src/Microsoft.OData.Core/JsonLight/JsonLightODataAnnotationWriter.cs
```diff
@@ -115,6 +136,81 @@ public void WriteInstanceAnnotationName(string annotationName)
...
+
+            return asyncJsonWriter.WritePropertyAnnotationNameAsync(propertyName, SimplifyODataAnnotationName(annotationName));
+        }
+
+        /// <summary>
+        /// Asynchronously writes a JSON instance annotation name which represents a instance annotation.
+        /// </summary>
+        /// <param name="annotationName">The name of the instance annotation to write.</param>
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        public Task WriteInstanceAnnotationNameAsync(string annotationName)
```
xuzhg, line 205: > public Task [](start = 7, length = 12)

why no async ?

src/Microsoft.OData.Core/JsonLight/JsonLightODataAnnotationWriter.cs
```diff
@@ -124,5 +220,16 @@ private string SimplifyODataAnnotationName(string annotationName)
...
+
+        /// <summary>
+        /// Asserts that the annotation writer was created for asynchronous operation.
+        /// </summary>
+        [Conditional("DEBUG")]
+        private void AssertAsynchronous()
+        {
+#if DEBUG
+            Debug.Assert(this.asyncJsonWriter != null, "The method should only be called on an asynchronous annotation writer.");
+#endif
```
xuzhg, line 232: at least check the "asyncJsonWriter != null"?

xuzhg (approved):
:shipit:

src/Microsoft.OData.Core/JsonLight/JsonLightODataAnnotationWriter.cs
gathogojr: @xuzhg For performance reasons, we made the decision to elide `async` and `await` where a method is a pass-through. See Mike's [comment](https://github.com/OData/odata.net/pull/2032#discussion_r600609802) on related PR. More on the subject in this post [https://blog.stephencleary.com/2016/12/eliding-async-await.html](https://blog.stephencleary.com/2016/12/eliding-async-await.html)

src/Microsoft.OData.Core/JsonLight/JsonLightODataAnnotationWriter.cs
gathogojr: See my comment [here](https://github.com/OData/odata.net/pull/2038#discussion_r603790463)

src/Microsoft.OData.Core/JsonLight/JsonLightODataAnnotationWriter.cs
gathogojr: That is what the assertion statement is doing. I'm not sure I understand your question here @xuzhg

xuzhg: It's only in debug mode, I mean why do you have #if DEBUG, then have Debug.Assert(...)?

Debug.Assert() only runs for #if DEBUG, right?

So, remove #if DEBUG?

marabooy (approved):

src/Microsoft.OData.Core/JsonLight/JsonLightODataAnnotationWriter.cs
habbes: But is this one a pass-through call? The call to `this.AssertAsynchronous()` has not potential effects/exceptions?

habbes: I see down below that it just wraps Debug.Assert.

src/Microsoft.OData.Core/JsonLight/JsonLightODataAnnotationWriter.cs
```diff
@@ -115,6 +136,81 @@ public void WriteInstanceAnnotationName(string annotationName)
...
+        /// Asynchronously writes a JSON property name which represents a property annotation.
+        /// </summary>
+        /// <param name="propertyName">The name of the property to annotate.</param>
+        /// <param name="annotationName">The name of the annotation to write.</param>
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        public Task WritePropertyAnnotationNameAsync(string propertyName, string annotationName)
+        {
+            this.AssertAsynchronous();
+            Debug.Assert(!string.IsNullOrEmpty(propertyName), "!string.IsNullOrEmpty(propertyName)");
+            Debug.Assert(annotationName.StartsWith(JsonLightConstants.ODataAnnotationNamespacePrefix,
```
habbes, line 194: I guess this could theoretically throw an exception if annotationName is null.

src/Microsoft.OData.Core/JsonLight/JsonLightODataAnnotationWriter.cs
gathogojr: @xuzhg I finally understood your comment.
We're applying the `Conditional("DEBUG")` attribute to this method and it will not be compiled unless the `DEBUG` compilation symbol is defined.
For that reason, I believe we don't need the `#if DEBUG/#endif` statements.
Just to be sure I won't be shooting myself in the foot by removing them, would you know why we've had this practice of applying the `#if DEBUG/#endif` statements in methods with similar `ConditionalAttribute`?
- https://github.com/OData/odata.net/blob/5fc3f2d24c50c6108c053763279a92377eebc099/src/Microsoft.OData.Core/ODataInputContext.cs#L582
- https://github.com/OData/odata.net/blob/5fc3f2d24c50c6108c053763279a92377eebc099/src/Microsoft.OData.Core/ODataInputContext.cs#L594
- https://github.com/OData/odata.net/blob/a355bf73d87bcfff6d1d62d9847b3d304464cb43/src/Microsoft.OData.Core/ODataOutputContext.cs#L662
- https://github.com/OData/odata.net/blob/a355bf73d87bcfff6d1d62d9847b3d304464cb43/src/Microsoft.OData.Core/ODataOutputContext.cs#L674

src/Microsoft.OData.Core/JsonLight/JsonLightODataAnnotationWriter.cs
gathogojr: That is correct. In reality though, `annotationName` would never be null so we can cure this by adding a debug assertion in the both synchronous and asynchronous execution path

habbes (approved):

src/Microsoft.OData.Core/JsonLight/JsonLightODataAnnotationWriter.cs
xuzhg: I don't have such history.

PR #2041, opened 2021-04-01
Remember the SelectItem instances that are thrown away because of the existence of a WildcardSelectItem

ificator:
### Description

If a query like `$select=*,property` is provided, the parsed `SelectExpandClause` will contain only `WildcardSelectItem`. While this is, in general, an acceptable optimization there may be scenarios were services need to determine whether a property is explicitly selected even when the wildcard is provided. This is not currently possible.

This change introduces a new property to `WildcardSelectItem` that can be used by service implementations to see all of the `SelectItem` instances that were subsumed by the current `WildcardSelectItem`.

This also optimizes the `SelectExpandClause` so calls to `AddToSelectedItems` avoid, for the most part, allocating new lists.

- [X] *Test cases added*


src/Microsoft.OData.Core/UriParser/SemanticAst/SelectExpandClause.cs
```diff
@@ -49,7 +49,7 @@ public IEnumerable<SelectItem> SelectedItems
             get
             {
                 ////Debug.Assert(!this.usedInternalLegacyConstructor || this.selectedItems != null, "You cannot get the list of selected items until processing is complete.");
-                return this.selectedItems.AsEnumerable();
+                return new ReadOnlyCollection<SelectItem>(this.selectedItems).AsEnumerable();
```
marabooy, line 52: @ificator I would probably have the old code here as it will reduce the amout of allocations for cases where reading the selected items happens alot.

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandClauseFinisher.cs
```diff
@@ -20,19 +20,24 @@ internal sealed class SelectExpandClauseFinisher
         /// <param name="clause">the select expand clause to modify.</param>
         public static void AddExplicitNavPropLinksWhereNecessary(SelectExpandClause clause)
         {
-            IEnumerable<SelectItem> selectItems = clause.SelectedItems;
+            // Since we're going to be modifying the clause's selected items we need to make a copy here to avoid failures related to
+            // enumerating a changed collection.
+            IEnumerable<SelectItem> selectItems = clause.SelectedItems.ToList();
```
xuzhg, line 25: why don't use "IList<SelectItem>"

xuzhg (approved):

src/Microsoft.OData.Core/UriParser/SemanticAst/SelectExpandClause.cs
ificator: I debated what we wanted here - I could just return `this.selectedItems` but that would be someone could theoretically cast it back to a `IList<SelectItem>` and modify our collection. Another option would be to have to allocate the `ReadOnlyCollection<SelectItem>` once - we should be able to modify the underlying collection independently of the read-only instance.

I ended up going with this approach because I rationalized the allocation of the `ReadOnlyCollection<SelectItem>` is pretty cheap (it's just a wrapper), but I'm willing to update to one of the other options (or something else?) if there's a preference.

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandClauseFinisher.cs
ificator: I can update to be an `IList<SelectItem>`. It wasn't a conscious decision, I simply added `ToList()` to the current `IEnumerable<SelectItem>` to force it to materialize. It may be I'll need to update this anyway if `SelectedItems` ends up returning an `IList<SelectItem>` itself (see previous comment).

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -984,9 +986,11 @@ private static void AddToSelectedItems(SelectItem itemToAdd, List<SelectItem> se
             }
 
             // if the selected item is "*", filter the existing property selection.
-            if (itemToAdd is WildcardSelectItem)
+            wildcardSelectItem = itemToAdd as WildcardSelectItem;
+            if (wildcardSelectItem != null)
```
odero, line 990:
```suggestion
            if (itemToAdd is WildcardSelectItem wildcardSelectItem)

```

odero, line 990:
```suggestion
            if (itemToAdd is WildcardSelectItem wildcardSelectItem)
```

src/Microsoft.OData.Core/UriParser/SemanticAst/SelectExpandClause.cs
```diff
@@ -76,18 +76,20 @@ internal void AddToSelectedItems(SelectItem itemToAdd, bool getOnlySubset = fals
         {
             ExceptionUtils.CheckArgumentNotNull(itemToAdd, "itemToAdd");
 
-            if (this.selectedItems.Any(x => x is WildcardSelectItem) && IsStructuralOrNavigationPropertySelectionItem(itemToAdd))
+            WildcardSelectItem wildcardSelectItem = this.selectedItems.FirstOrDefault(x => x is WildcardSelectItem) as WildcardSelectItem;
```
odero, line 79:
```suggestion
            WildcardSelectItem wildcardSelectItem = this.selectedItems.OfType<WildcardSelectItem>().FirstOrDefault();

```

src/Microsoft.OData.Core/UriParser/SemanticAst/SelectExpandClause.cs
```diff
@@ -76,18 +76,20 @@ internal void AddToSelectedItems(SelectItem itemToAdd, bool getOnlySubset = fals
...
-            if (this.selectedItems.Any(x => x is WildcardSelectItem) && IsStructuralOrNavigationPropertySelectionItem(itemToAdd))
+            WildcardSelectItem wildcardSelectItem = this.selectedItems.FirstOrDefault(x => x is WildcardSelectItem) as WildcardSelectItem;
+            if (wildcardSelectItem != null && IsStructuralOrNavigationPropertySelectionItem(itemToAdd))
             {
+                wildcardSelectItem.AddSubsumed(itemToAdd);
                 return;
             }
 
-            bool isWildcard = itemToAdd is WildcardSelectItem;
+            wildcardSelectItem = itemToAdd as WildcardSelectItem;
```
KenitoInc, line 86: Why are we assigning `itemToAdd` to `wildcardSelectItem`?

KenitoInc:
@ificator 
We want to merge this PR to go out with the 7.9 release. Kindly look into the review comments.

src/Microsoft.OData.Core/UriParser/SemanticAst/SelectExpandClause.cs
mikepizzo: @ificator -- That seems reasonable to me.

KenitoInc:
> > @ificator
> > We want to merge this PR, to go out with the 7.9 release. Kindly look into the review comments.
> 
> Sorry for the delay, I've been busy. I'll see if I can work on it this weekend.

Hey @ificator 
Did you get time to look into the review comments?

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
ificator: We're reusing `wildcardSelectItem` from above so cannot redefine it here.

src/Microsoft.OData.Core/UriParser/SemanticAst/SelectExpandClause.cs
ificator: Because if `itemToAdd` is an instance of `WildcardSelectItem` we need access to it to add to the subsumed list. We could define a new variable (e.g. what @odero mentioned in one of his comments), but we already had a variable of type `WildcardSelectItem` defined so I figured I'd just use it again.

ificator:
I thought this had gone through after my latest updates, but it looks like there was a random failure during the pipeline unrelated to my change. Is there a way to rerun the failed pipeline?

marabooy:
@ificator Kindly rebase and the tests should run.

marabooy (approved):

KenitoInc (approved):
LGTM

KenitoInc:
Thanks for this PR @ificator

PR #2052, opened 2021-04-15
Implement asynchronous support in ODataJsonLightResourceSerializer

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataJsonLightResourceSerializer`**

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceSerializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceSerializer.cs
@@ -8,6 +8,7 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
+using System.Threading.Tasks;
 using Microsoft.OData.Edm;
 using Microsoft.OData.Json;
 
@@ -79,30 +80,7 @@ namespace Microsoft.OData.JsonLight
 
             ODataResourceBase resource = resourceState.Resource;
 
-            // expectedResourceTypeName : if expected is of base type. but the resource real type is derived,
-            // we need to set the resource type name.
-            //      Writing response: expected type info can be identical with context uri info.
-            //      Writing request: contextUri may not be provided, we always use the type from metadata info.
-            //                       From model: if the resource can be found in model.
-            //                       From serializationInfo: if user set the serializationInfo for the resource.
-            string expectedResourceTypeName = null;
-            if (this.WritingResponse)
-            {
-                expectedResourceTypeName = resourceState.GetOrCreateTypeContext(this.WritingResponse).ExpectedResourceTypeName;
-            }
-            else
-            {
-                if (resourceState.ResourceTypeFromMetadata == null)
-                {
-                    expectedResourceTypeName = resourceState.SerializationInfo == null
-                     ? null
-                     : resourceState.SerializationInfo.ExpectedTypeName;
-                }
-                else
-                {
-                    expectedResourceTypeName = resourceState.ResourceTypeFromMetadata.FullTypeName();
-                }
-            }
+            string expectedResourceTypeName = GetExpectedResourceTypeName(resourceState);
 
             // Write the "@odata.type": "typename"
             string typeName = this.JsonLightOutputContext.TypeNameOracle.GetResourceTypeNameForWriting(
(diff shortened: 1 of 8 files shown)


src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceSerializer.cs
```diff
@@ -372,6 +334,347 @@ internal ODataContextUrlInfo WriteResourceSetContextUri(ODataResourceTypeContext
...
+        internal async Task WriteResourceEndMetadataPropertiesAsync(
+            IODataJsonLightWriterResourceState resourceState,
+            IDuplicatePropertyNameChecker duplicatePropertyNameChecker)
+        {
+            Debug.Assert(resourceState != null, "resourceState != null");
+
+            ODataResourceBase resource = resourceState.Resource;
+
+            // Write computed navigation properties
+            var navigationLinkInfo = resource.MetadataBuilder.GetNextUnprocessedNavigationLink();
```
marabooy, line 520: use concrete types

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightResourceSerializerTests.cs
```diff
@@ -0,0 +1,766 @@
...
+            this.messageWriterSettings.SetServiceDocumentUri(new Uri(ServiceUri));
+        }
+
+        [Theory]
+        [InlineData(false, "Name", "{")]
+        [InlineData(true, null, "{\"@odata.type\":\"#Collection(NS.Category)\"")]
+        [InlineData(true, "Prop", "{\"Prop@odata.type\":\"#Collection(NS.Category)\"")]
+        public async Task WriteResourceSetStartMetadataPropertiesAsync_WritesResourceSetMetadataProperties(bool isUndeclared, string propertyName, string expected)
+        {
+            var result = await SetupJsonLightResourceSerializerAndRunTestAsync(
```
marabooy, line 71: Lets use concrete types so in future this can catch any changes to the public api.

marabooy (approved):
LGTM

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceSerializer.cs
```diff
@@ -56,7 +57,7 @@ internal void WriteResourceSetStartMetadataProperties(ODataResourceSet resourceS
                 resourceSet,
                 isUndeclared);
 
-            if (typeName != null && !typeName.Contains(ODataConstants.ContextUriFragmentUntyped))
+            if (!IsNullOrContainsContextUriFragmentUntyped(typeName))
```
habbes, line 60: I feel like this abstraction is overly specific. Is it commonly used?

gathogojr: Used in 2 places - sync and async method - and it looked and felt brittle enough to the extent that I didn't feel comfortable repeating it. I however believe this is the kind of method that the compiler would inline.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightResourceSerializerTests.cs
gathogojr: The `SetupJsonLightResourceSerializerAndRunTestAsync` is a private method within this test class. It sets up an ODataJsonLightResourceSerializer, then runs the given test code asynchronously, and then flushes and reads the stream back as a `string` for assertion. `var` felt natural to use here rather than `string`

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceSerializer.cs
```diff
@@ -505,5 +808,214 @@ private void WriteOperation(ODataOperation operation)
...
+        /// <remarks>
+        /// Expects the actions or functions scope to already be open.
+        /// </remarks>
+        /// <param name="operations">A grouping of operations that are all actions or all functions and share the same "metadata".</param>
+        private async Task WriteOperationMetadataGroupAsync(IGrouping<string, ODataOperation> operations)
+        {
+            this.ValidateOperationMetadataGroup(operations);
+            await this.JsonLightOutputContext.AsynchronousJsonWriter.WriteNameAsync(operations.Key)
+                .ConfigureAwait(false);
+            bool useArray = operations.Count() > 1;
```
habbes, line 841: Would `Any()` ? be better hear?

habbes (approved):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceSerializer.cs
habbes: It felt like the name of the method does not provide a considerable level of abstraction beyond what the statement itself does. I mean it's a matter of preference/perception so I don't think you should change it, but if I would give an analogy.

I think this abstraction makes sense:
```c#
bool isAdultAge(int age) {
   return age >= 18;
}
```
but I'd feel like the following one is gratuitous and possibly unnecessary:
```c#
bool isGreaterThanEighteen(int x) {
   return x >= 18
}
```

habbes: About the inlining, I think that's something that could be verified. But I also don't think it's a major issue at this point.

gathogojr: @habbes I got rid of the superfluous methods

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceSerializer.cs
gathogojr: @habbes @marabooy Notice that here the check is for collection having **more than** one item. `Any()` is equivalent to `Count() > 0`

xuzhg:
>     {

nothing change to this file?

---
Refers to: src/Microsoft.OData.Core/ODataResourceSetBase.cs:21 in bb3d0bf. [](commit_id = bb3d0bf6cd97e0cdcc6713718b5c758513a71e7e, deletion_comment = False)

xuzhg (approved):
:shipit:

KenitoInc (approved):

PR #2053, opened 2021-04-15
Elide async and await in pass-through asynchronous methods

gathogojr:
### Issues

*This pull request fixes issue #2019.*

### Description

Elide async and await in pass-through asynchronous methods

Affects:
- **`TextWriterWrapper`** abstract class
- **`NonIndentedTextWriter`** internal class
- **`JsonWriter`** internal class
- **`JsonValueUtils`** internal static class
- **`ODataJsonTextWriter`** internal class
- **`ODataBinaryStreamWriter`** internal class

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
@@ -76,22 +76,20 @@ namespace Microsoft.OData.Json
         /// </summary>
         /// <param name="writer">The text writer to write the output to.</param>
         /// <param name="value">Float value to be written.</param>
-        internal static async Task WriteValueAsync(this TextWriter writer, float value)
+        internal static Task WriteValueAsync(this TextWriter writer, float value)
         {
             Debug.Assert(writer != null, "writer != null");
 
             if (JsonSharedUtils.IsFloatValueSerializedAsString(value))
             {
-                await writer.WriteQuotedAsync(value.ToString(ODataNumberFormatInfo))
-                    .ConfigureAwait(false);
+                return writer.WriteQuotedAsync(value.ToString(ODataNumberFormatInfo));
             }
             else
             {
                 // float.ToString() supports a max scale of six,
                 // whereas float.MinValue and float.MaxValue have 8 digits scale. Hence we need
                 // to use XmlConvert in all other cases, except infinity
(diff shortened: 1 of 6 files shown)


src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -82,14 +82,16 @@ internal static async Task WriteValueAsync(this TextWriter writer, float value)
 
             if (JsonSharedUtils.IsFloatValueSerializedAsString(value))
             {
-                await writer.WriteQuotedAsync(value.ToString(ODataNumberFormatInfo)).ConfigureAwait(false);
+                await writer.WriteQuotedAsync(value.ToString(ODataNumberFormatInfo))
```
g2mula, line 85: Can this also be "Elided" :-)
ie 
```csharp
static Task WriteValue()
{
  Debug.Assert(writer != null, "not null")
  if (true)
  {
    return writer.WriterQuotedAsync(...);
  }
  else
  {
    return writer.WriteAsync(...)
  }
}

g2mula (commented):
Personally I prefer this, but also hope we are aware of what gets lost for this performance improvement eg as [documented here](https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#prefer-asyncawait-over-directly-returning-task)

marabooy (approved):

odero:
I get doing this for internal/private APIs where you have control over how the APIs are used, but do you want to do this for public APIs? Looking at the benefits linked to in this comment https://github.com/OData/odata.net/pull/2053#pullrequestreview-637506011

gathogojr:
> I get doing this for internal/private APIs where you have control over how the APIs are used, but do you want to do this for public APIs? Looking at the benefits linked to in this comment [#2053 (review)](https://github.com/OData/odata.net/pull/2053#pullrequestreview-637506011)

@odero All the affected 6 classes are `internal`, out of which 4 are `sealed`, 1 is `abstract` and the last one is `static`

odero (approved):

habbes (approved):

xuzhg (approved):
:shipit:

KenitoInc (approved):
LGTM

g2mula (approved):
Ship it!

PR #2058, opened 2021-04-21
Modify IODataOutputInStreamErrorListener interface to support asynchronous notifications when an in-stream error is to be written

gathogojr:
### Issues

### Description

Modify **`IODataOutputInStreamErrorListener`** interface to support asynchronous notifications when an in-stream error is to be written

#### Background
When an error occurs when writing a payload to the stream, a notification is sent to any subscribed listener to notify it that an in-stream error is to be written. The listener implements the `IODataOutputInStreamErrorListener` interface that contains a single synchronous method `OnInStreamError`. **This interface is `internal`**.
To support similar functionality in asynchronous scenarios, we need to modify the interface to support an asynchronous equivalent method - `OnInStreamErrorAsync`.

`ODataBatchWriter` is a `public abstract` class that implements the `IODataOutputInStreamErrorListener` interface. If `OnInStreamErrorAsync` method was implemented abstractly in the classes, subclasses of `ODataBatchWriter` that users of the library might have implemented would break. For that reason, the method has been implemented as a virtual method that just throws a `NotImplementedException` to prevent the breaking change.
As for other `internal` classes that implement the interface, an explicit implementation has been added. For now, it just throws a `NotImplementedException`. The actual implemententation will be provided when implementing asynchronous support in the following 6 sets of writer classes:
- `ODataWriterCore`/`ODataJsonLightWriter`
- `ODataParameterWriterCore`/`ODataJsonLightParameterWriter`
- `ODataCollectionWriterCore`/`ODataJsonLightCollectionWriter` (PR #2059)
- `ODataDeltaWriter`/`ODataJsonLightDeltaWriter`
- `ODataBatchWriter`/`ODataJsonLightBatchWriter`
- `ODataAsynchronousWriter`

Below is how that implementation looks like in **`ODataCollectionWriterCore`**. Notice the call to the asynchronous method `StartPayloadInStartStateAsync` that sets the stage for writing the error.
```csharp
async Task IODataOutputInStreamErrorListener.OnInStreamErrorAsync()
{
    this.VerifyNotDisposed();

    // We're in a completed state trying to write an error
    // We can't write error after the payload was finished as it might introduce another top-level element in XML
    if (this.State == CollectionWriterState.Completed)
    {
        throw new ODataException(Strings.ODataWriterCore_InvalidTransitionFromCompleted(this.State.ToString(), CollectionWriterState.Error.ToString()));
    }

    await this.StartPayloadInStartStateAsync()
        .ConfigureAwait(false);
    this.EnterScope(CollectionWriterState.Error, this.scopes.Peek().Item);
}
```

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
+++ b/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
@@ -419,6 +419,14 @@ namespace Microsoft.OData
         /// </remarks>
         public abstract void OnInStreamError();
 
+        /// <inheritdoc/>
+        public virtual Task OnInStreamErrorAsync()
+        {
+            // NOTE: ODataBatchWriter class is abstract and public. This method body is intended
+            // to prevent a breaking change in subclasses that provide the implementation.
+            throw new NotImplementedException();
+        }
+
         /// <summary>
         /// Flush the output.
         /// </summary>
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeltaWriter.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeltaWriter.cs
@@ -287,6 +287,12 @@ namespace Microsoft.OData.JsonLight
             this.inStreamErrorListener.OnInStreamError();
         }
 
+        /// <inheritdoc/>
+        Task IODataOutputInStreamErrorListener.OnInStreamErrorAsync()
+        {
+            throw new NotImplementedException();
+        }
+
         #endregion
     }
 }
--- a/src/Microsoft.OData.Core/IODataOutputInStreamErrorListener.cs
+++ b/src/Microsoft.OData.Core/IODataOutputInStreamErrorListener.cs
@@ -6,6 +6,10 @@
 
 namespace Microsoft.OData
 {
+    #region Namespaces
+    using System.Threading.Tasks;
+    #endregion Namespaces
+
     /// <summary>
     /// An interface that allows the implementations of the writers to get notified if an in-stream error is to be written.
     /// </summary>
@@ -19,5 +23,15 @@ namespace Microsoft.OData
         /// If the listener returns, the writer should not allow any more writing, since the in-stream error is the last thing in the payload.
         /// </remarks>
         void OnInStreamError();
+
+        /// <summary>
+        /// This method asynchronously notifies the listener, that an in-stream error is to be written.
+        /// </summary>
+        /// <remarks>
+        /// This listener can choose to fail, if the currently written payload doesn't support in-stream error at this position.
+        /// If the listener returns, the writer should not allow any more writing, since the in-stream error is the last thing in the payload.
+        /// </remarks>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        Task OnInStreamErrorAsync();
     }
 }
--- a/src/Microsoft.OData.Core/ODataAsynchronousWriter.cs
+++ b/src/Microsoft.OData.Core/ODataAsynchronousWriter.cs
@@ -101,6 +101,12 @@ namespace Microsoft.OData
             throw new ODataException(Strings.ODataAsyncWriter_CannotWriteInStreamErrorForAsync);
         }
 
+        /// <inheritdoc/>
+        Task IODataOutputInStreamErrorListener.OnInStreamErrorAsync()
+        {
+            throw new NotImplementedException();
+        }
+
         /// <summary>
         /// Validates that the async writer is not disposed.
         /// </summary>
--- a/src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
+++ b/src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
@@ -265,6 +265,12 @@ namespace Microsoft.OData
             this.EnterScope(CollectionWriterState.Error, this.scopes.Peek().Item);
         }
 
+        /// <inheritdoc/>
+        Task IODataOutputInStreamErrorListener.OnInStreamErrorAsync()
+        {
+            throw new NotImplementedException();
+        }
+
         /// <summary>
         /// Determines whether a given writer state is considered an error state.
         /// </summary>
--- a/src/Microsoft.OData.Core/ODataParameterWriterCore.cs
+++ b/src/Microsoft.OData.Core/ODataParameterWriterCore.cs
@@ -306,6 +306,12 @@ namespace Microsoft.OData
             throw new ODataException(Strings.ODataParameterWriter_InStreamErrorNotSupported);
         }
 
+        /// <inheritdoc/>
+        Task IODataOutputInStreamErrorListener.OnInStreamErrorAsync()
+        {
+            throw new NotImplementedException();
+        }
+
         /// <summary>
         /// Check if the object has been disposed; called from all public API methods. Throws an ObjectDisposedException if the object
         /// has already been disposed.
--- a/src/Microsoft.OData.Core/ODataWriterCore.cs
+++ b/src/Microsoft.OData.Core/ODataWriterCore.cs
@@ -760,6 +760,12 @@ namespace Microsoft.OData
             this.EnterScope(WriterState.Error, this.CurrentScope.Item);
         }
 
+        /// <inheritdoc/>
+        Task IODataOutputInStreamErrorListener.OnInStreamErrorAsync()
+        {
+            throw new NotImplementedException();
+        }
+
         /// <summary>
         /// This method is called when a stream is requested. It is a no-op.
         /// </summary>


KenitoInc (approved):

src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
```diff
@@ -419,6 +419,14 @@ public Task FlushAsync()
         /// </remarks>
         public abstract void OnInStreamError();
 
+        /// <inheritdoc/>
+        public virtual Task OnInStreamErrorAsync()
+        {
+            // NOTE: ODataBatchWriter class is abstract and public. This method body is intended
+            // to prevent a breaking change in subclasses that provide the implementation.
+            throw new NotImplementedException();
```
habbes, line 427: I wonder if wrapping `OnInStreamError` in a task would be a more "expected" default implementation for `OnInStreamErrorAsync`. Is that something you considered? If so, I would be curious to know what dissuaded you from that approach.

Also does the method body prevent breaking change in subclasses that provide the implementation or in those that did not provide the implementation?

gathogojr: @habbes Dealing with asynchronous wrappers over synchronous methods is actually what this "true async" initiative is all about so I don't think its a good idea to wrap `OnInStreamError`.
With regard to your second point, the method body prevents a breaking change in the event that someone had subclassed `ODataBatchWriter`. The only implementation they would have provided is for the synchronous method and we are not tinkering with that.

habbes: @gathogojr on the first point: You make a good point. But I still wonder which default is the lesser evil. My concern is whether there cases where `OnInStreamErrorAsync` is called without the knowledge or control of the developer who implemented the abstract class. I think this is usually quite likely. If this is the case, then merely updating the library without any additional changes to their program code might lead to unexpected exceptions occurring (possibly occurring in production and undetected in development). This would be quite disruptive, and a breaking change. On the other hand, if the default implementation wraps the `OnInStreamError` method, then program will retain its current behaviour even after updating the lib. While this implementation is not necessarily ideal for async, it's less disruptive and not a breaking change. If you look at the case of the built-in Stream abstract class, `Flush` is a abstract and `FlushAsync` is a virtual method which defaults to wrapping `Flush` in a task: https://source.dot.net/#System.Private.CoreLib/Stream.cs,189

habbes: If we stick with the default of throwing an exception, then I think we should have a way to ensure that developers updating the library in their programs (which could also be libraries used by others) are aware that they need to implement that method or to ensure that that method is not inadvertently called internally by ODL.

gathogojr: @habbes I think you misunderstood. The `OnStreamInErrorAsync` is not triggered automatically, just like the current `OnStreamInError` is not triggered automatically. Just adding the method does not cause it to be triggered.
[Here](https://github.com/OData/odata.net/blob/caee76d0cafbf9b107effa30c6e12920824cf406/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs#L950) is how the existing method is currently triggered.
Which is why throwing a `NotImplementedException` does no harm. FYI, most of the `public virtual` methods in the `public abstract` `ODataWriter` actually throw `NotImplementedException`. If as a developer you choose to provide an implementation, you'll need to implement the methods that you intend to trigger. It's no different here.

gathogojr: If a developer updates the library in their programs, the method will just sit there doing nothing. It'll not be triggered at any point.

marabooy (approved):

src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
habbes: @gathogojr I'm a bit confused. Are those methods triggered by the developer's code or by ODL code?
In the link you provided, the `OnStreamInError` method seems to be explicitly called by ODL code, so that leaves me a bit confused by what you mean when you say "just like the current `OnStreamInError` is not triggered automatically. In the PR changes, I can't see an instance where `OnStreamInErrorAsync` is called, who is responsible for calling it and where? If it's never called directly by ODL, then what's the point of having that method on the class? I feel like I'm missing something.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeltaWriter.cs
```diff
@@ -287,6 +287,12 @@ void IODataOutputInStreamErrorListener.OnInStreamError()
             this.inStreamErrorListener.OnInStreamError();
         }
 
+        /// <inheritdoc/>
+        Task IODataOutputInStreamErrorListener.OnInStreamErrorAsync()
```
habbes, line 291: Out of curiosity, why is this method explicitly defined for the `IODataOutputInStreamErrorListener` interface? I realize that this is also the case for the pre-existing `OnInStreamError` methods, but just curious to know why. Does this class implement multiple interfaces each defining `OnInStreamError`?

src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
gathogojr: @habbes I'll rephrase. The breaking change concern stems from the scenario where a developer has implemented the `ODataBatchWriter` abstract class. Let's say they call it `DeveloperODataBatchWriter`. What I meant by saying it is not triggered automatically is that, ODL wouldn't trigger `DeveloperODataBatchWriter.OnStreamInErrorAsync` at any time. The link I shared is to ODL `internal` `ODataJsonLightOutputContext` triggering the method. What I could check is if this internal class can resolve `ODataBatchWriter` from the DI container before triggering the method because that would be the only way it'd be a breaking change.

gathogojr: The other concern I have with your suggestion is that the synchronous and asynchronous approach are using 2 different writers. While they share the same output stream, we verify every step of the way whether the `OutputContext` instance if set for synchronous or asynchronous writing

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeltaWriter.cs
gathogojr: > Does this class implement multiple interfaces each defining OnInStreamError?
No it doesn't
> why is this method explicitly defined for the IODataOutputInStreamErrorListener interface?
I'm not 100% sure why explicit interface implementation route was taken

gathogojr: I'm guessing the benefit  is that any class that implements `ODataBatchWriter` abstract class for example can override the default `IODataOutputInStreamErrorListener` methods either as a public method, or as an explicit interface implementation.

gathogojr: @habbes A person who implements `ODataBatchWriter` abstract class would only bring it into play by implementing the `public abstract` **`ODataOutputContext`** `CreateODataBatchWriter` method just like we do in [`ODataJsonLightOutputContext`](https://github.com/OData/odata.net/blob/de68595d34fa77ab2b192b1d7c6d35682a4100dd/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs#L564).
By doing that they take the internal `ODataJsonLightOutputContext` out of play and take the responsibility of triggering the `IODataOutputInStreamErrorListener` methods.
`ODataJsonLightOutputContext` has a hard dependency of `ODataJsonLightBatchWriter` as you can see [here](https://github.com/OData/odata.net/blob/de68595d34fa77ab2b192b1d7c6d35682a4100dd/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs#L931)
```csharp
private ODataBatchWriter CreateODataBatchWriterImplementation()
{
    ODataBatchWriter batchWriter = new ODataJsonLightBatchWriter(this);
    this.outputInStreamErrorListener = batchWriter;
    return batchWriter;
}
```

src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
habbes: @gathogojr and I had an offline chat around this topic and he clarified why this is not a breaking change. Thought I'd share that here as well.

The ODL calls `OnInStreamError` for the internal implementations of the `OutputContext` class. This internal implementations only uses internal subclasses of `ODataBatchWriter`. At no point does it use an `ODataBatchWriter` that would be passed in by third-party code or via external-facing dependency-injection. So if the user creates their own subclass of the `ODataBatchWriter`, then we can guarantee that that version will not be called by the internal ODL methods. The only exception is if the user also creates their own `OutputContext` class, but if they do this, then they are responsible for all the logic, including error handling, whether or not to use error listeners, how to use invoke error listeners, etc. So they will cater for all that (note that a lot of the built-in classes, as well as the `IODataOutputInStreamErrorListener` interface are `internal` and not accessible to the developer).

A follow-up question I had was why it was necessary to add the `OnStreamInErrorAsync` method if it's never called by ODL. The reason it was necessary is because `ODataBatchWriter` implements `IODataOutputInStreamErrorListener` and this method has been added to that interface, so either `ODataBatchWriter` or its concrete subclasses need to implement it. So a basic implementation is added to `ODataBatchWriter` so that subclasses aren't required to.

habbes (approved):

PR #2059, opened 2021-04-21
Implement asynchronous support in ODataJsonLightCollectionWriter

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataJsonLightCollectionWriter`**

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
+++ b/src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
@@ -178,10 +178,11 @@ namespace Microsoft.OData
         /// </summary>
         /// <param name="collection">The <see cref="ODataCollectionStart"/> representing the collection.</param>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
-        public sealed override Task WriteStartAsync(ODataCollectionStart collection)
+        public sealed override async Task WriteStartAsync(ODataCollectionStart collection)
         {
             this.VerifyCanWriteStart(false, collection);
-            return TaskUtils.GetTaskForSynchronousOperation(() => this.WriteStartImplementation(collection));
+            await this.WriteStartImplementationAsync(collection)
+                .ConfigureAwait(false);
         }
 
         /// <summary>
@@ -199,10 +200,11 @@ namespace Microsoft.OData
         /// </summary>
         /// <param name="item">The collection item to write.</param>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
-        public sealed override Task WriteItemAsync(object item)
+        public sealed override async Task WriteItemAsync(object item)
         {
             this.VerifyCanWriteItem(false);
-            return TaskUtils.GetTaskForSynchronousOperation(() => this.WriteItemImplementation(item));
+            await this.WriteItemImplementationAsync(item)
+                .ConfigureAwait(false);
         }
 
         /// <summary>
@@ -224,10 +226,10 @@ namespace Microsoft.OData
         /// Asynchronously finish writing a collection.
         /// </summary>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
-        public sealed override Task WriteEndAsync()
+        public sealed override async Task WriteEndAsync()
         {
             this.VerifyCanWriteEnd(false);
-            return TaskUtils.GetTaskForSynchronousOperation(this.WriteEndImplementation)
+            await this.WriteEndImplementationAsync()
                 .FollowOnSuccessWithTask(
                     task =>
                     {
@@ -240,7 +242,7 @@ namespace Microsoft.OData
                         {
                             return TaskUtils.CompletedTask;
                         }
-                    });
+                    }).ConfigureAwait(false);
         }
 
         /// <summary>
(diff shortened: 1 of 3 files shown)


habbes (approved):

src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
```diff
@@ -266,9 +268,20 @@ void IODataOutputInStreamErrorListener.OnInStreamError()
         }
 
         /// <inheritdoc/>
-        Task IODataOutputInStreamErrorListener.OnInStreamErrorAsync()
+        async Task IODataOutputInStreamErrorListener.OnInStreamErrorAsync()
```
KenitoInc, line 271: This method is private?
If so, I would prefer it to have the `private` modifier

KenitoInc (changes requested):
Just a minor comment. Otherwise looks good

src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
gathogojr: @KenitoInc The `IODataOutputInStreamErrorListener` interface is implemented explicitly. Applying a `private` modifier or even a `public` modifier is not the right thing to do nor is it required.
Based on [C# specification](https://www.microsoft.com/en-us/download/details.aspx?id=7029):
_Explicit interface member implementations have different accessibility characteristics than other members. Because explicit interface member implementations are never accessible through their fully qualified name in a method invocation or a property access, they are in a sense private. However, since they can be accessed through an interface instance, they are in a sense also public._

It's accessible through the interface instance as shown below so it's not quite private:
```csharp
public interface I
{
    void M();
}

public class C : I
{
    // Explicit implementation
    void I.M();
}

I i = new C();
i.M();
```

KenitoInc (approved):

src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
```diff
@@ -178,10 +178,11 @@ public sealed override void WriteStart(ODataCollectionStart collectionStart)
         /// </summary>
         /// <param name="collection">The <see cref="ODataCollectionStart"/> representing the collection.</param>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
-        public sealed override Task WriteStartAsync(ODataCollectionStart collection)
+        public sealed override async Task WriteStartAsync(ODataCollectionStart collection)
         {
             this.VerifyCanWriteStart(false, collection);
-            return TaskUtils.GetTaskForSynchronousOperation(() => this.WriteStartImplementation(collection));
+            await this.WriteStartImplementationAsync(collection)
```
marabooy, line 184: cant we elide this?

src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
```diff
@@ -199,10 +200,11 @@ public sealed override void WriteItem(object item)
         /// </summary>
         /// <param name="item">The collection item to write.</param>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
-        public sealed override Task WriteItemAsync(object item)
+        public sealed override async Task WriteItemAsync(object item)
         {
             this.VerifyCanWriteItem(false);
-            return TaskUtils.GetTaskForSynchronousOperation(() => this.WriteItemImplementation(item));
+            await this.WriteItemImplementationAsync(item)
```
marabooy, line 206: Elidable?

src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
```diff
@@ -224,10 +226,10 @@ public sealed override void WriteEnd()
         /// Asynchronously finish writing a collection.
         /// </summary>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
-        public sealed override Task WriteEndAsync()
+        public sealed override async Task WriteEndAsync()
         {
             this.VerifyCanWriteEnd(false);
-            return TaskUtils.GetTaskForSynchronousOperation(this.WriteEndImplementation)
+            await this.WriteEndImplementationAsync()
```
marabooy, line 232: Elide

src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
gathogojr: It was being done before but its actually not a safe thing to do here. `VerifyCanWriteStart` can throw an exception and if this method is not async, an AsyncStateMachine will not be created, meaning that the exception would escape to the caller(s).

src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
gathogojr: Explained why it's not particularly recommendable to do that here https://github.com/OData/odata.net/pull/2059#discussion_r635801748

src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
gathogojr: Explained why it's not particularly recommendable to do that here https://github.com/OData/odata.net/pull/2059#discussion_r635801748

PR #2069, opened 2021-04-27
Improve $count in $filter collection properties

KenitoInc:
### Description

Currently ODL supports
`$filter=navProp/$count gt 2`
`$filter=collectionProp/$count gt 2`

It doesn't support `$filter` and `$search` within $count.
Spec: http://docs.oasis-open.org/odata/odata/v4.01/cs01/part1-protocol/odata-v4.01-cs01-part1-protocol.html#sec_RequestingtheNumberofItemsinaCollect

This PR add support for the following:

`$filter=navProp/$count($filter=prop gt 1) gt 2`
`$filter=navProp/$count($search=prop) eq 'abc'`

`$filter=collectionProp/$count($filter=prop gt 1) gt 2`
`$filter=collectionProp/$count($search=prop) eq 'abc'`

- [ ] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/UriParser/Binders/CountSegmentBinder.cs
```diff
@@ -0,0 +1,72 @@
...
+            this.state = state;
+        }
+
+        /// <summary>
+        /// Binds an Count segment token.
+        /// </summary>
+        /// <param name="countSegmentToken">The Count segment token to bind.</param>
+        /// <param name="state">State of the metadata binding.</param>
+        /// <returns>The bound Count segment token.</returns>
+        internal QueryNode BindCountSegment(CountSegmentToken countSegmentToken)
```
xuzhg, line 43: >BindCountSegment [](start = 27, length = 16)

maybe name it as Bind(...)

xuzhg (approved):
:shipit:

src/Microsoft.OData.Core/UriParser/Parsers/UriQueryExpressionParser.cs
```diff
@@ -44,7 +44,7 @@ public sealed class UriQueryExpressionParser
         /// <summary>
         /// The current recursion depth.
         /// </summary>
-        private int recursionDepth;
+        internal int recursionDepth;
```
marabooy, line 47:
```suggestion
        internal int RecursionDepth { get; set; }

```
nit: Maybe a property would be better for this.

marabooy (approved):

src/Microsoft.OData.Core/UriParser/Binders/CountSegmentBinder.cs
```diff
@@ -0,0 +1,72 @@
...
+        {
+            ExceptionUtils.CheckArgumentNotNull(countSegmentToken, "countSegmentToken");
+
+            QueryNode source = this.bindMethod(countSegmentToken.NextToken);
+            CollectionNode node = source as CollectionNode;
```
mikepizzo, line 48: >node [](http://example.com/codeflow?start=27&length=4)

any chance this could be null?  i.e., that bindMethod could return null, or something other than a CollectionNode?

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// </summary>
+        public ExpressionLexer Lexer
+        {
+            get { return this.lexer; }
+        }
+
+        /// <summary>
+        /// Create a <see cref="CountSegmentToken"/>.
+        /// </summary>
+        /// <param name="identifier">The name of this token, which in this case is always "$count".The name of this token, which in this case is always "$count".</param>
```
mikepizzo, line 59: >The name of this token, which in this case is always "$count" [](http://example.com/codeflow?start=99&length=61)

text seems to be repeated

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -400,6 +400,30 @@ public void ParseFilterWithEntityCollectionCount()
                     Source.ShouldBeCollectionNavigationNode(HardCodedTestModel.GetPersonMyFriendsDogsProp());
         }
 
+        [Fact]
+        public void ParseFilterWithEntityCollectionCountWithFilterOption()
+        {
+            var filterQueryNode = ParseFilter("MyFriendsDogs/$count($filter=Color eq 'Brown') gt 1", HardCodedTestModel.TestModel, HardCodedTestModel.GetPersonType(), HardCodedTestModel.GetPeopleSet());
```
mikepizzo, line 406: >$filter=Color eq 'Brown' [](http://example.com/codeflow?start=68&length=24)

Do you have a test for $search?  Do you have a test for both $filter and $search?

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
```diff
@@ -0,0 +1,138 @@
...
+            else
+            {
+                // Only a $filter and $search can be inside a $count()
+                throw ParseError(ODataErrorStrings.UriQueryExpressionParser_IllegalQueryOptioninDollarCount());
+            }
+
+            if (this.lexer.CurrentToken.Kind != ExpressionTokenKind.CloseParen)
+            {
+                throw ParseError(ODataErrorStrings.UriQueryExpressionParser_CloseParenExpected(this.lexer.CurrentToken.Position, this.lexer.ExpressionText));
+            }
```
mikepizzo, line 101: what if /$count includes *both* $filter and $search?

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
```diff
@@ -0,0 +1,138 @@
...
+                string searchQuery = TryGetQueryOption(ExpressionConstants.QueryOptionSearch, textWithinParenthesis);
+                searchToken = searchQuery == null ? null : this.UriQueryExpressionParser.ParseSearch(searchQuery);
+            }
+            else
+            {
+                // Only a $filter and $search can be inside a $count()
+                throw ParseError(ODataErrorStrings.UriQueryExpressionParser_IllegalQueryOptioninDollarCount());
+            }
+
+            if (this.lexer.CurrentToken.Kind != ExpressionTokenKind.CloseParen)
```
mikepizzo, line 98: >if (this.lexer.CurrentToken.Kind != ExpressionTokenKind.CloseParen) [](http://example.com/codeflow?start=12&length=67)

can you just use this.lexer.ValidateToken(ExpressionTokenKind.CloseParen) ?

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
```diff
@@ -0,0 +1,138 @@
...
+            if (this.lexer.CurrentToken.Kind != ExpressionTokenKind.OpenParen)
+            {
+                return new CountSegmentToken(identifier, countedInstance);
+            }
+            else
+            {
+                textWithinParenthesis = this.lexer.AdvanceThroughExpandOption();
+            }
+
+            // TODO: check for NoDollarQueryOptions
```
mikepizzo, line 81: please add support for nodollarqueryoption

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
```diff
@@ -0,0 +1,138 @@
...
+            int outerRecursiveDepth = this.UriQueryExpressionParser.recursionDepth;
+            ExpressionLexer outerLexer = this.UriQueryExpressionParser.Lexer;
+
+            if (this.lexer.CurrentToken.Kind != ExpressionTokenKind.OpenParen)
+            {
+                return new CountSegmentToken(identifier, countedInstance);
+            }
+            else
+            {
+                textWithinParenthesis = this.lexer.AdvanceThroughExpandOption();
```
mikepizzo, line 78: >AdvanceThroughExpandOption [](http://example.com/codeflow?start=51&length=26)

caution -- you are using AdvanceThroughExpandOption differently than how it is used in selectexpandbinder.  There it is used to read a single option value only (i.e., after the equals sign). Here you using it to read the entire option key/value pair. Special logic within AdvanceThroughExpandOption for handling things like quoting (now or in the future) might make assumptions that it is only reading the value.  Better might be to first parse the query option name and then use AdvanceThroughExpandOption to parse just the value of the query option (i.e., everything after the equals).

src/Microsoft.OData.Core/UriParser/SyntacticAst/CountSegmentToken.cs
```diff
@@ -0,0 +1,122 @@
...
+        /// The token representing $search.
+        /// </summary>
+        private QueryToken searchOption;
+
+        /// <summary>
+        /// Create a CountSegmentToken given the Identifier and the NextToken.
+        /// </summary>
+        /// <param name="identifier">The name of this token, which in this case is always "$count".</param>
+        /// <param name="nextToken">The instance to count on.</param>
+        public CountSegmentToken(string identifier, QueryToken nextToken)
```
mikepizzo, line 43: >string identifier,  [](http://example.com/codeflow?start=33&length=19)

if this is always $count, why bother passing it?  Why not just return QueryOptionCount on line 79?

src/Microsoft.OData.Core/UriParser/Parsers/UriQueryExpressionParser.cs
```diff
@@ -1219,6 +1234,20 @@ private QueryToken ParseSegment(QueryToken parent)
...
         }
 
+        /// <summary>
+        /// Parses a $count segment.
+        /// </summary>
+        /// <param name="parent">The parent of the segment node.</param>
+        /// <returns>The lexical token representing the $count segment.</returns>
+        private QueryToken ParseCountSegment(QueryToken parent)
+        {
+            string identifier = this.lexer.CurrentToken.GetIdentifier();
```
mikepizzo, line 1244: >string identifier = this.lexer.CurrentToken.GetIdentifier(); [](http://example.com/codeflow?start=12&length=60)

Do you need to do this? Don't you already know it's "$count"?

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
```diff
@@ -0,0 +1,138 @@
...
+            }
+
+            this.lexer.NextToken();
+            this.UriQueryExpressionParser.recursionDepth = outerRecursiveDepth;
+            this.UriQueryExpressionParser.Lexer = outerLexer;
+            return new CountSegmentToken(identifier, countedInstance, filterToken, searchToken);
+        }
+
+        /// <summary>
+        ///
```
mikepizzo, line 110: please add summary

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
```diff
@@ -0,0 +1,138 @@
...
+            return new CountSegmentToken(identifier, countedInstance, filterToken, searchToken);
+        }
+
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="queryOptionName"></param>
+        /// <param name="query"></param>
+        /// <returns></returns>
+        private static string TryGetQueryOption(string queryOptionName, string query)
```
mikepizzo, line 115: >TryGetQueryOption [](http://example.com/codeflow?start=30&length=17)

I think this entire method goes away if you first check to see if the query option is $filter/$search, and then use AdvanceThroughExpandOption to just read the value.

Right now it seems like you get the expression for the key value pair, check to see if it is filter or search, and then try to remove the filter/search so that you can pass it to the appropriate Parse method.

src/Microsoft.OData.Core/UriParser/SyntacticAst/CountSegmentToken.cs
```diff
@@ -0,0 +1,122 @@
...
+        {
+            get { return QueryTokenKind.CountSegment; }
+        }
+
+        /// <summary>
+        /// The name of this token, which in this case is always "$count".
+        /// </summary>
+        public override string Identifier
+        {
+            get { return this.identifier; }
```
mikepizzo, line 79: >this.identifier [](http://example.com/codeflow?start=25&length=15)

Just return ExpressionConstants.QueryOptionCount and remove the identifier private variable and constructor argument.

mikepizzo (changes requested):
:clock1:

src/Microsoft.OData.Core/UriParser/SyntacticAst/CountSegmentToken.cs
KenitoInc: Done

src/Microsoft.OData.Core/UriParser/Parsers/UriQueryExpressionParser.cs
KenitoInc: Fixed

src/Microsoft.OData.Core/UriParser/SyntacticAst/CountSegmentToken.cs
KenitoInc: Fixed

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
KenitoInc: Fixed

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
KenitoInc: Removed the `identifier` entirely since we don't need it.

src/Microsoft.OData.Core/UriParser/Parsers/UriQueryExpressionParser.cs
KenitoInc: I refractored the code and reverted this change.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
KenitoInc: Resolved

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
KenitoInc: Resolved

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
KenitoInc: Resolved

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
KenitoInc: Resolved

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
KenitoInc: Resolved

src/Microsoft.OData.Core/UriParser/Binders/CountSegmentBinder.cs
KenitoInc: I have added the code to throw an exception whenever the node is null.

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
KenitoInc: Added

src/Microsoft.OData.Core/UriParser/Binders/CountSegmentBinder.cs
mikepizzo: This is an internal method, so if this would only be null i we had a logic issue in our code we could make this an assert, rather than a null check in production code

mikepizzo (approved):
:shipit:

mikepizzo:
I like the new factoring much better -- thanks.  A few minor suggestions, but looks good to me.

src/Microsoft.OData.Core/UriParser/Binders/CountSegmentBinder.cs
KenitoInc: Since this occurs when user tries parsing `NonCollectionProperty/$count`, we need to return the exception to the user.

PR #2072, opened 2021-04-28
no error message when reading an incorrect structure of a collection …

ElizabethOkerio:
…of entities

### Issues

*This pull request fixes issue #2071 .*

### Description
If we have such a model: 
```
    public class ProductFamily
    {
        public int Id { get; set; }

        public string Description { get; set; }

        public virtual Supplier Supplier { get; set; }

        public virtual ICollection<Product> Products { get; set; } = new List<Product>();
    }
```
Then a HTTP POST request to create a `ProductFamily` that demonstrates the bug has this POST Body:
```
{
  "Id": 9,
  "Products": ["Hello"]
}
```
The above request body creates a ProductFamily successfully instead of returning a BadRequest

- [ ] *Test cases added*

### Additional work necessary


xuzhg (approved):

gathogojr:
@ElizabethOkerio Test(s)?

KenitoInc (approved):
LGTM

PR #2075, opened 2021-04-30
Modify IODataStreamListener interface to support asynchronous notifications when the content stream of an operation has been disposed

gathogojr:
### Issues

### Description

Modify **`IODataStreamListener`** interface to support asynchronous notifications when the content stream of an operation has been disposed.

#### Background
**`IODataStreamListener`** interface allows the creators of a stream to listen for status changes of the operation stream. The `StreamDisposed` method notified the implementer of this interface that the content stream of the operation has been disposed. To support similar functionality in asynchronous scenarios, we need to modify the interface to support an equivalent method - `OnStreamDisposedAsync`.

The actual implemented will be provided when implementing asynchronous support in the following 3 classes that implement the interface:
- `ODataWriterCore`
- `ODataBatchWriter`
- `ODataReaderCore`
- `ODataBatchReader`

A stubbed implementation has been provided in each of the classes to prevent a breaking change.
Here's how the actual implementation would look like in `ODataWriterCore`.
```csharp
async Task IODataStreamListener.StreamDisposedAsync()
{
    Debug.Assert(this.State == WriterState.Stream || this.State == WriterState.String, "Stream was disposed when not in WriterState.Stream state.");

    // Complete writing the stream
    if (this.State == WriterState.Stream)
    {
        await this.EndBinaryStreamAsync().ConfigureAwait(false);
    }
    else if (this.State == WriterState.String)
    {
        await this.EndTextWriterAsync().ConfigureAwait(false);
    }

    await this.LeaveScopeAsync().ConfigureAwait(false);
}
```

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
+++ b/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
@@ -411,6 +411,17 @@ namespace Microsoft.OData
         public abstract void StreamDisposed();
 
         /// <summary>
+        /// This method is called asynchronously to notify that the content stream of a batch operation has been disposed.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        public virtual Task StreamDisposedAsync()
+        {
+            // NOTE: ODataBatchWriter class is abstract and public. This method body is intended
+            // to prevent a breaking change in subclasses that provide the implementation.
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
         /// This method notifies the listener, that an in-stream error is to be written.
         /// </summary>
         /// <remarks>
--- a/src/Microsoft.OData.Core/Batch/ODataBatchReader.cs
+++ b/src/Microsoft.OData.Core/Batch/ODataBatchReader.cs
@@ -232,6 +232,16 @@ namespace Microsoft.OData
         }
 
         /// <summary>
+        /// This method is called asynchronously to notify that the content stream of a batch operation has been disposed.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        Task IODataStreamListener.StreamDisposedAsync()
+        {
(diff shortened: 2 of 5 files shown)


marabooy:
Tests?

gathogojr:
> Tests?

@marabooy I'll have a follow up PR for each of the affected classes with tests against the actual implementations. Main reason why I created an independent PR for this change is to keep the other pull requests that depend on this change leaner and cleaner.
At best, what I could do in this PR is to add tests to verify that a `NotImplementedException` is thrown when the method is invoked against classes that implement either of abstract `ODataWriterCore`, `ODataReaderCore` and `ODataBatchWriter` and that a completed task is returned by classes that implement `ODataBatchReader` - _tests I don't find that useful unless you feel strongly about it..._

marabooy (approved):

src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
```diff
@@ -410,6 +410,17 @@ public Task FlushAsync()
...
         public abstract void StreamDisposed();
 
+        /// <summary>
+        /// This method is called asynchronously to notify that the content stream of a batch operation has been disposed.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous operation.</returns>
+        public virtual Task StreamDisposedAsync()
+        {
+            // NOTE: ODataBatchWriter class is abstract and public. This method body is intended
+            // to prevent a breaking change in subclasses that provide the implementation.
```
habbes, line 420: I think this message is a bit confusing. I think it would be clearer to say something like:
`This method body is intended a breaking change in subclasses that do not override this method`.
When you say "subclasses that provide the implementation", it's not clear whether you're referring to the implementation of the abstract class, or implementation of the `StreamDisposeAsync` method.

habbes (approved):

KenitoInc (approved):
LGTM

PR #2078, opened 2021-05-03
Implement asynchronous support in ODataJsonLightWriter

gathogojr:
### Issues

### Description

Affects **`ODataWriterCore`** and **`ODataJsonLightWriter`**

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
@@ -7,14 +7,13 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
-using System.Diagnostics.CodeAnalysis;
 using System.IO;
 using System.Linq;
 using System.Threading.Tasks;
 using Microsoft.OData.Edm;
 using Microsoft.OData.Evaluation;
-using Microsoft.OData.Metadata;
 using Microsoft.OData.Json;
+using Microsoft.OData.Metadata;
 
 namespace Microsoft.OData.JsonLight
 {
@@ -64,6 +63,36 @@ namespace Microsoft.OData.JsonLight
         private readonly JsonLightODataAnnotationWriter odataAnnotationWriter;
 
         /// <summary>
+        /// The underlying asynchronous JSON writer.
+        /// </summary>
+        private readonly IJsonWriterAsync asynchronousJsonWriter;
+
+        /// <summary>
+        /// The underlying asynchronous JSON writer.
+        /// </summary>
+        private readonly IJsonStreamWriterAsync asynchronousJsonStreamWriter;
+
+        /// <summary>
+        /// OData asynchronous annotation writer.
+        /// </summary>
+        private readonly JsonLightODataAnnotationWriter asynchronousODataAnnotationWriter;
+
+        /// <summary>
+        /// OData instance annotation writer
+        /// </summary>
+        private readonly JsonLightInstanceAnnotationWriter instanceAnnotationWriter;
+
+        /// <summary>
+        /// True if the writer was created for writing a response; false otherwise.
+        /// </summary>
+        private readonly bool writingResponse;
+
+        /// <summary>
+        /// The message writer settings to be used for writing.
+        /// </summary>
+        private readonly ODataMessageWriterSettings messageWriterSettings;
+
+        /// <summary>
         /// Constructor.
         /// </summary>
         /// <param name="jsonLightOutputContext">The output context to write to.</param>
@@ -93,8 +122,13 @@ namespace Microsoft.OData.JsonLight
             this.writingParameter = writingParameter;
             this.jsonWriter = this.jsonLightOutputContext.JsonWriter;
             this.jsonStreamWriter = this.jsonWriter as IJsonStreamWriter;
-            this.odataAnnotationWriter = new JsonLightODataAnnotationWriter(this.jsonWriter,
-                this.jsonLightOutputContext.OmitODataPrefix, this.jsonLightOutputContext.MessageWriterSettings.Version);
+            this.odataAnnotationWriter = this.jsonLightValueSerializer.ODataAnnotationWriter;
+            this.asynchronousJsonWriter = this.jsonLightOutputContext.AsynchronousJsonWriter;
+            this.asynchronousJsonStreamWriter = this.asynchronousJsonWriter as IJsonStreamWriterAsync;
+            this.asynchronousODataAnnotationWriter = this.jsonLightValueSerializer.AsynchronousODataAnnotationWriter;
+            this.instanceAnnotationWriter = this.jsonLightValueSerializer.InstanceAnnotationWriter;
+            this.writingResponse = this.jsonLightOutputContext.WritingResponse;
+            this.messageWriterSettings = this.jsonLightOutputContext.MessageWriterSettings;
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightWriterTests.cs
```diff
@@ -0,0 +1,2168 @@
...
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.IO;
+using System.Text;
+using System.Threading.Tasks;
+using Microsoft.OData.Edm;
+using Microsoft.OData.JsonLight;
+using Xunit;
+
+namespace Microsoft.OData.Tests.JsonLight
```
xuzhg, line 17: >Tests [](start = 26, length = 5)

the namespace is "Microsoft.OData.Core.Tests.JsonLight"

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
```diff
@@ -1262,275 +1296,1280 @@ protected override NestedResourceInfoScope CreateNestedResourceInfoScope(WriterS
...
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        protected override async Task WriteEntityReferenceInNavigationLinkContentAsync(
+            ODataNestedResourceInfo parentNestedResourceInfo,
+            ODataEntityReferenceLink entityReferenceLink)
+        {
+            Debug.Assert(parentNestedResourceInfo != null, "parentNestedResourceInfo != null");
+            Debug.Assert(entityReferenceLink != null, "entityReferenceLink != null");
+
+            // In JSON Light, we can only write entity reference links at the beginning of a navigation link in requests;
+            // once we wrote a resource set, entity reference links are not allowed anymore (we require all the entity reference
```
KenitoInc, line 2186:
```suggestion
            // once we write a resource set, entity reference links are not allowed anymore (we require all the entity reference

```

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
```diff
@@ -1597,7 +2636,6 @@ private void WriteDeltaLinkTarget(ODataDeltaLinkBase link)
         /// Validates that the ODataResourceSet.InstanceAnnotations collection is empty for the given expanded resource set.
         /// </summary>
         /// <param name="resourceSet">The expanded resource set in question.</param>
-        [SuppressMessage("Microsoft.Performance", "CA1822:MarkMembersAsStatic", Justification = "An instance field is used in a debug assert.")]
```
KenitoInc, line 1600: Why did we stop suppressing this warning?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
```diff
@@ -1611,6 +2649,340 @@ private void ValidateNoCustomInstanceAnnotationsForExpandedResourceSet(ODataReso
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Asynchronously writes the properties for a delta resource.
+        /// </summary>
+        /// <param name="resource">The resource whose properties to write.</param>
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        private Task WriteDeltaResourcePropertiesAsync(ODataResourceBase resource)
```
KenitoInc, line 2870: I think we need to Debug Assert that the resource is not null.

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -581,16 +578,16 @@ public sealed override void WritePrimitive(ODataPrimitiveValue primitiveValue)
...
         }
 
-
         /// <summary>
         /// Asynchronously write a primitive value.
         /// </summary>
         /// <param name="primitiveValue"> Primitive value to write.</param>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
-        public sealed override Task WritePrimitiveAsync(ODataPrimitiveValue primitiveValue)
+        public sealed override async Task WritePrimitiveAsync(ODataPrimitiveValue primitiveValue)
```
KenitoInc, line 586: Why are we not eliding async in this method, like the way we are doing in `WriteDeltaDeletedLinkAsync` ?

Same question for the methods below:
`WriteStartAsync`
`CreateBinaryWriteStreamAsync`
`CreateTextWriterAsync`
`WriteEntityReferenceLinkAsync`
`WriteEntityReferenceLinkAsync`

KenitoInc: Also
I can see we have some `Verify****Async` methods. Why are some missing?

KenitoInc (changes requested):
A few questions and change requests

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
```diff
@@ -1262,275 +1296,1280 @@ protected override NestedResourceInfoScope CreateNestedResourceInfoScope(WriterS
...
+            await this.asynchronousJsonWriter.EndObjectScopeAsync()
+                .ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Asynchronously start writing a property.
+        /// </summary>
+        /// <param name="property">The property info to write.</param>
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        protected override Task StartPropertyAsync(ODataPropertyInfo property)
```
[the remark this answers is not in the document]
gathogojr: Since we defer to `ODataJsonLightPropertySerializer`'s `WritePropertyInfoAsync` method, property is null scenario would be caught by the call to `ValidatePropertyInfo` method.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
gathogojr: @KenitoInc Because the `Debug.Assert` statements refer to `this` so I reasoned that the "MarkMembersAsStatic" warning would no longer valid. Unless the alternative reasoning is that we need to retain the suppression since assert statements are stripped in non-debug scenarios? What is your opinion?

src/Microsoft.OData.Core/ODataWriterCore.cs
gathogojr: @KenitoInc The reason we're not eliding is because the statement `this.VerifyCanWritePrimitive(false, primitiveValue);` can throw an exception. By making the method async, a state machine will be generated and such an exception will be captured and placed on the returned task. If we elide `async`, the exception is raised directly rather than going on the task. The expected asynchronous semantics are that exceptions are placed on the returned task.
For that same reason I have applied `async` to `WriteDeltaDeletedLinkAsync`.

Regarding your second question about missing `Verify***Async` methods, the only scenarios where I introduced an asynchronous equivalent is where the verify method does an asynchronous operation. For example,
```csharp
        private async Task VerifyCanWriteStartResourceSetAsync(bool synchronousCall, ODataResourceSet resourceSet)
        {
            ExceptionUtils.CheckArgumentNotNull(resourceSet, "resourceSet");

            this.VerifyNotDisposed();
            this.VerifyCallAllowed(synchronousCall);

            await this.StartPayloadInStartStateAsync()
                .ConfigureAwait(false);
        }
```
As you can see here, this method is initiating an asynchronous operation to start writing the payload. One apparent observation would be that the verify method is doing a little more than it'd be expected to but the primary objective of this initiative is to achieve parity with the synchronous API.
The alternative approach would be to move the call to `StartPayloadInStartState` from `VerifyCanWriteStartResourceSet` to the calling method such that the verify method can be reused by both the synchronous and asynchronous workflow but I thought I could do that in a refactor PR.

KenitoInc (approved):
LGTM

marabooy (approved):

PR #2079, opened 2021-05-06
Support asynchronous invocation of IODataStreamListener.StreamDisposedAsync method from ODataNotificationWriter class

gathogojr:
### Issues

### Description

Support asynchronous invocation of **`IODataStreamListener.StreamDisposedAsync`** method from `Dispose`/`DisposeAsync` methods of **`ODataNotificationWriter`** class.
- In .NET Core 3.1 or later, we're able to ride on asynchronous `DisposeAsync` method defined in `IAsyncDisposable` interface to avoid blocking.
- In platforms below .NET Core 3.1, we're forced to `.Wait()` on the asynchronous method.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataNotificationWriter.cs
+++ b/src/Microsoft.OData.Core/ODataNotificationWriter.cs
@@ -21,9 +21,10 @@ namespace Microsoft.OData
     internal sealed class ODataNotificationWriter : TextWriter
 #endif
     {
-        private readonly TextWriter textWriter;
+        private TextWriter textWriter;
         private IODataStreamListener listener;
-        private bool synchronous;
+        private bool disposed = false;
+        private readonly bool synchronous;
 
         internal ODataNotificationWriter(TextWriter textWriter, IODataStreamListener listener, bool synchronous = true)
             : base(System.Globalization.CultureInfo.InvariantCulture)
@@ -326,44 +327,44 @@ namespace Microsoft.OData
         /// <param name="disposing">True if called from Dispose; false if called from the finalizer.</param>
         protected override void Dispose(bool disposing)
         {
-            if (disposing)
+            if (!this.disposed && disposing)
             {
-                if (this.listener != null)
+                // Tell the listener that the stream is being disposed.
+                if (synchronous)
                 {
-                    // Tell the listener that the stream is being disposed.
-                    if (synchronous)
-                    {
-                        this.listener.StreamDisposed();
-                    }
-                    else
-                    {
-                        this.listener.StreamDisposedAsync().Wait();
-                    }
-
-                    this.listener = null;
+                    this.listener?.StreamDisposed();
                 }
+                else
+                {
+                    this.listener?.StreamDisposedAsync().Wait();
+                }
+
+                this.listener = null;
+
+                this.textWriter?.Dispose();
+                this.textWriter = null;
             }
 
-            this.textWriter.Dispose();
+            this.disposed = true;
             base.Dispose(disposing);
         }
 
 #if NETSTANDARD2_0
         public async ValueTask DisposeAsync()
         {
-            if (this.listener != null)
+            if (!this.disposed && this.listener != null)
             {
                 await this.listener.StreamDisposedAsync()
                     .ConfigureAwait(false);
+                this.textWriter?.Dispose();
 
                 this.listener = null;
+                this.textWriter = null;
             }
 
             // Dispose unmanaged resources
-            // Pass false to ensure functional equivalence with the synchronous dispose pattern
+            // Pass `false` to ensure functional equivalence with the synchronous dispose pattern
             this.Dispose(false);
-            // Suppress finalization
-            GC.SuppressFinalize(this);
         }
 #endif
     }
--- a/src/Microsoft.OData.Core/ODataNotificationWriter.cs
+++ b/src/Microsoft.OData.Core/ODataNotificationWriter.cs
@@ -15,12 +15,17 @@ namespace Microsoft.OData
     /// <summary>
     /// Wrapper for TextWriter to listen for dispose.
     /// </summary>
+#if NETSTANDARD2_0
+    internal sealed class ODataNotificationWriter : TextWriter, IAsyncDisposable
+#else
     internal sealed class ODataNotificationWriter : TextWriter
+#endif
     {
         private readonly TextWriter textWriter;
         private IODataStreamListener listener;
+        private bool synchronous;
 
-        internal ODataNotificationWriter(TextWriter textWriter, IODataStreamListener listener)
+        internal ODataNotificationWriter(TextWriter textWriter, IODataStreamListener listener, bool synchronous = true)
             : base(System.Globalization.CultureInfo.InvariantCulture)
         {
             Debug.Assert(textWriter != null, "Creating a notification writer for a null textWriter.");
@@ -28,6 +33,7 @@ namespace Microsoft.OData
 
             this.textWriter = textWriter;
             this.listener = listener;
+            this.synchronous = synchronous;
         }
 
         /// <inheritdoc/>
@@ -81,6 +87,7 @@ namespace Microsoft.OData
         }
 
         #region Write methods
+
         /// <inheritdoc/>
         public override void Write(char value)
         {
@@ -168,6 +175,7 @@ namespace Microsoft.OData
         #endregion
 
         #region WriteLine methods
+
         /// <inheritdoc/>
         public override void WriteLine()
         {
@@ -262,7 +270,6 @@ namespace Microsoft.OData
 
         #region async methods
 
-
         /// <inheritdoc/>
         public override Task FlushAsync()
         {
@@ -324,7 +331,15 @@ namespace Microsoft.OData
                 if (this.listener != null)
                 {
                     // Tell the listener that the stream is being disposed.
-                    this.listener.StreamDisposed();
+                    if (synchronous)
+                    {
+                        this.listener.StreamDisposed();
+                    }
+                    else
+                    {
+                        this.listener.StreamDisposedAsync().Wait();
+                    }
+
                     this.listener = null;
                 }
             }
@@ -332,5 +347,24 @@ namespace Microsoft.OData
             this.textWriter.Dispose();
             base.Dispose(disposing);
         }
+
+#if NETSTANDARD2_0
+        public async ValueTask DisposeAsync()
+        {
+            if (this.listener != null)
+            {
+                await this.listener.StreamDisposedAsync()
+                    .ConfigureAwait(false);
+
+                this.listener = null;
+            }
+
+            // Dispose unmanaged resources
+            // Pass false to ensure functional equivalence with the synchronous dispose pattern
+            this.Dispose(false);
+            // Suppress finalization
+            GC.SuppressFinalize(this);
+        }
+#endif
     }
 }
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationWriterTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationWriterTests.cs
@@ -1,0 +1,122 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ODataNotificationWriterTests.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using System;
+using System.IO;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Microsoft.OData.Tests
+{
+    public class ODataNotificationWriterTests
+    {
+        private MemoryStream stream;
+        private TextWriter writer;
+        private IODataStreamListener streamListener;
+
+        public ODataNotificationWriterTests()
+        {
+            this.stream = new MemoryStream();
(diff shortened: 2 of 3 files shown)


src/Microsoft.OData.Core/ODataNotificationWriter.cs
```diff
@@ -324,13 +331,40 @@ protected override void Dispose(bool disposing)
...
+                    .ConfigureAwait(false);
+
+                this.listener = null;
+            }
+
+            // Dispose unmanaged resources
+            // Pass false to ensure functional equivalence with the synchronous dispose pattern
+            this.Dispose(false);
+            // Suppress finalization
+            GC.SuppressFinalize(this);
```
habbes, line 366: Do we need to suppress finalization if this class doesn't implement a finalizer? But I guess it doesn't hurt to call either way.

src/Microsoft.OData.Core/ODataNotificationWriter.cs
```diff
@@ -324,13 +331,40 @@ protected override void Dispose(bool disposing)
...
             base.Dispose(disposing);
         }
+
+#if NETSTANDARD2_0
+        public async ValueTask DisposeAsync()
+        {
+            if (this.listener != null)
+            {
+                await this.listener.StreamDisposedAsync()
+                    .ConfigureAwait(false);
```
habbes, line 357: Is it possible that the listener only implemented the synchronous version `StreamDisposed`?

src/Microsoft.OData.Core/ODataNotificationWriter.cs
```diff
@@ -324,13 +331,40 @@ protected override void Dispose(bool disposing)
...
+                    else
+                    {
+                        this.listener.StreamDisposedAsync().Wait();
+                    }
+
                     this.listener = null;
                 }
             }
 
             this.textWriter.Dispose();
```
habbes, line 347: Why is `this.textWriter.Dispose()` not called in `if (disposing)` block. Also, why isn't it set to `null` after being disposed?

Based on [the guidelines](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose), I assume the following pattern would have been more appropriate:
```c#
if (disposing) {
   //...
  this.textWriter?.Dispose();
}

// ...
this.textWriter = null;
```

src/Microsoft.OData.Core/ODataNotificationWriter.cs
```diff
@@ -324,13 +331,40 @@ protected override void Dispose(bool disposing)
                 if (this.listener != null)
                 {
                     // Tell the listener that the stream is being disposed.
-                    this.listener.StreamDisposed();
+                    if (synchronous)
+                    {
+                        this.listener.StreamDisposed();
```
habbes, line 336: Maybe call `this.listener?.StreamDispose()` so you can get rid of the `this.listener != null` block?

src/Microsoft.OData.Core/ODataNotificationWriter.cs
habbes, line 347: By the way `TextWriter` also implements `IAsyncDisposable`. So maybe you could call `textWriter.Dispose()` in the `Dispose(bool)` method and `textWriter.DisposeAsync()` in the `DisposeAsync()` method.

src/Microsoft.OData.Core/ODataNotificationWriter.cs
gathogojr: @habbes Previously yes, we only had the asynchronous `StreamDisposed` method. In `ODataJsonLightWriter` for example, here is how that method was implemented:
```csharp
        void IODataStreamListener.StreamDisposed()
        {
            // Complete writing the stream
            if (this.State == WriterState.Stream)
            {
                this.EndBinaryStream();
            }
            else if (this.State == WriterState.String)
            {
                this.EndTextWriter();
            }
            this.LeaveScope();
        }
```
You'll find that the method called from here actually perform some writing. Given that we want to achieve true async, that required introduction of an asynchronous equivalent - `StreamDisposedAsync` - looking something like this:
```csharp
        async Task IODataStreamListener.StreamDisposedAsync()
        {
            // Complete writing the stream
            if (this.State == WriterState.Stream)
            {
                await this.EndBinaryStreamAsync()
                    .ConfigureAwait(false);
            }
            else if (this.State == WriterState.String)
            {
                await this.EndTextWriterAsync()
                    .ConfigureAwait(false);
            }

            await this.LeaveScopeAsync()
                .ConfigureAwait(false);
        }
```
This therefore is intended to help us deliver the promise of a true asynchronous API.

src/Microsoft.OData.Core/ODataNotificationWriter.cs
gathogojr: I followed the guideline [here](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#implement-the-async-dispose-pattern). Note however that in our class we don't need to implement `DisposeAsyncCore` since the class is sealed.

src/Microsoft.OData.Core/ODataNotificationWriter.cs
gathogojr: @habbes That is a good question but that logic was introduced before by another contributor. I'd be hesitant to change unless there was a good reason to - e.g. a confirmed bug.

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationWriterTests.cs
```diff
@@ -0,0 +1,122 @@
...
+            Assert.Equal("StreamDisposed", result);
+        }
+
+#if NETCOREAPP3_1
+        [Fact]
+        public async Task NotificationWriterDisposeShouldInvokeStreamDisposedAsync()
+        {
+            await using (var notificationWriter = new ODataNotificationWriter(
+                new StreamWriter(new MemoryStream()),
+                this.streamListener)) // `synchronous` argument becomes irrelevant
```
habbes, line 50: Why is the synchronous argument irrelevant here? Is it because `await using` ensures that `DisposeAsync` is called?

gathogojr: Yes

src/Microsoft.OData.Core/ODataNotificationWriter.cs
```diff
@@ -15,19 +15,25 @@ namespace Microsoft.OData
...
+#else
     internal sealed class ODataNotificationWriter : TextWriter
+#endif
     {
         private readonly TextWriter textWriter;
         private IODataStreamListener listener;
+        private bool synchronous;
 
-        internal ODataNotificationWriter(TextWriter textWriter, IODataStreamListener listener)
+        internal ODataNotificationWriter(TextWriter textWriter, IODataStreamListener listener, bool synchronous = true)
```
habbes, line 28: Was there another way to implement this without adding the `bool synchronous` parameter?

src/Microsoft.OData.Core/ODataNotificationWriter.cs
gathogojr: @habbes There are two statements in that block:
```csharp
this.listener.StreamDisposed();
this.listener = null;
```
The second statement is important especially because if someone did something like this:
```csharp
notificationWriter.Dispose();
notificationWriter.Dispose();
// Or
await notificationWriter.DisposeAsync();
await notificationWriter.DisposeAsync();
```
Checking if listener is null guarantees we'll only ever call `StreamDisposed` or `StreamDisposedAsync` once.
That can also be guaranteed by the change you suggested in this way:
```csharp
this.listener?.StreamDisposed();
this.listener = null;
```
But I'm not particularly convinced that's a significant improvement from:
```csharp
if (this.listener != null)
{
    this.listener.StreamDisposed();
    this.listener = null;
}
```

src/Microsoft.OData.Core/ODataNotificationWriter.cs
gathogojr: I couldn't figure any other way. If we were only supporting .NET Core 3.1 or higher it'd be easy because in asynchronous situations we'd always be guaranteed of `DisposeAsync`. But for say .NET FX and .NET Core 1.x or 2.x, we honestly have no way of know from the `Dispose` method whether we're doing things synchronously or asynchronously. That's where the `synchronous` parameter comes in.

KenitoInc (commented):
One comment!

src/Microsoft.OData.Core/ODataNotificationWriter.cs
```diff
@@ -324,13 +331,40 @@ protected override void Dispose(bool disposing)
...
                 {
                     // Tell the listener that the stream is being disposed.
-                    this.listener.StreamDisposed();
+                    if (synchronous)
+                    {
+                        this.listener.StreamDisposed();
+                    }
+                    else
+                    {
+                        this.listener.StreamDisposedAsync().Wait();
```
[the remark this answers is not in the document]
KenitoInc: Just read the PR description again!
But I haven't understood clear why in pre netcore 3.1, we can't use await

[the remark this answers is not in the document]
gathogojr: @KenitoInc In 3.1 we can use `await using` which is why I have implemented `IAsyncDisposable` for `NETSTANDARD2_0`. I even added a test that uses `await using` for `NETCOREAPP3_1`

src/Microsoft.OData.Core/ODataNotificationWriter.cs
gathogojr: @habbes I'm not sure why but `.DisposeAsync()` method doesn't appear available to be invoked directly. However, `textWriter.Dispose()` should do the job since `TextWriter` is a BCL type.

gathogojr: Fixed

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationWriterTests.cs
```diff
@@ -0,0 +1,162 @@
...
+#else
+        [Fact]
+        public async Task NotificationWriterDisposeShouldInvokeStreamDisposedAsync()
+        {
+            using (var notificationWriter = new ODataNotificationWriter(
+                new StreamWriter(new MemoryStream()),
+                this.streamListener,
+                /*synchronous*/ false))
+            {
+
```
marabooy, line 109:
```suggestion
```

marabooy (approved):

src/Microsoft.OData.Core/ODataNotificationWriter.cs
```diff
@@ -319,18 +327,45 @@ public override Task WriteLineAsync(string value)
...
+                    }
+
                     this.listener = null;
                 }
+
+                this.textWriter?.Dispose();
             }
 
-            this.textWriter.Dispose();
+            this.disposed = true;
```
habbes, line 350: I think we should also set `this.textWriter = null;`?

src/Microsoft.OData.Core/ODataNotificationWriter.cs
habbes: Oh, so they will be required to implement `StreamDisposedAsync()` otherwise get a compiler error?

src/Microsoft.OData.Core/ODataNotificationWriter.cs
```diff
@@ -319,18 +327,45 @@ public override Task WriteLineAsync(string value)
...
+
+#if NETSTANDARD2_0
+        public async ValueTask DisposeAsync()
+        {
+            if (!this.disposed && this.listener != null)
+            {
+                await this.listener.StreamDisposedAsync()
+                    .ConfigureAwait(false);
+
+                this.listener = null;
```
habbes, line 362: I think you should also call `this.textWriter.DisposeAsync()` and set it to null.

habbes (approved):
Requesting minor changes. But other than that, looks good.

src/Microsoft.OData.Core/ODataNotificationWriter.cs
habbes: Interesting that it's not available on the class, yet it appears in the public API: https://docs.microsoft.com/en-us/dotnet/api/system.io.textwriter.disposeasync?view=net-5.0

Maybe we're reference an older version.

In that case I think it makes sense to call `this.textWriter?.Dispose()` outside the if-block. Alternatively, we can keep it in the if-block and duplicate it in both `Dispose()` and `DisposeAsync()`.

src/Microsoft.OData.Core/ODataNotificationWriter.cs
gathogojr: Fixed

src/Microsoft.OData.Core/ODataNotificationWriter.cs
gathogojr: As discussed, the BCL types in C# 7.1 do not have this method. The BCL package introduces the IAsyncDisposable interface

src/Microsoft.OData.Core/ODataNotificationWriter.cs
gathogojr: An earlier PR introduced the new method with a body that just throws a `NotImplementedException`. In subsequent PRs, I have been replacing that with the actual implementation

PR #2082, opened 2021-05-10
Implement asynchronous support in ODataJsonLightDeltaWriter

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataJsonLightDeltaWriter`**

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeltaWriter.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeltaWriter.cs
@@ -123,7 +123,7 @@ namespace Microsoft.OData.JsonLight
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
         public override Task WriteStartAsync(ODataDeltaResourceSet deltaResourceSet)
         {
-            return TaskUtils.GetTaskForSynchronousOperation(() => this.resourceWriter.WriteStart(deltaResourceSet));
+            return this.resourceWriter.WriteStartAsync(deltaResourceSet);
         }
 
         /// <summary>
@@ -140,7 +140,7 @@ namespace Microsoft.OData.JsonLight
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
         public override Task WriteEndAsync()
         {
-            return TaskUtils.GetTaskForSynchronousOperation(() => this.resourceWriter.WriteEnd());
+            return this.resourceWriter.WriteEndAsync();
         }
 
         /// <summary>
@@ -159,7 +159,7 @@ namespace Microsoft.OData.JsonLight
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
         public override Task WriteStartAsync(ODataNestedResourceInfo nestedResourceInfo)
         {
-            return TaskUtils.GetTaskForSynchronousOperation(() => this.resourceWriter.WriteStart(nestedResourceInfo));
+            return this.resourceWriter.WriteStartAsync(nestedResourceInfo);
         }
 
         /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
```diff
@@ -1597,7 +2636,6 @@ private void WriteDeltaLinkTarget(ODataDeltaLinkBase link)
         /// Validates that the ODataResourceSet.InstanceAnnotations collection is empty for the given expanded resource set.
         /// </summary>
         /// <param name="resourceSet">The expanded resource set in question.</param>
-        [SuppressMessage("Microsoft.Performance", "CA1822:MarkMembersAsStatic", Justification = "An instance field is used in a debug assert.")]
```
KenitoInc, line 1600: Why are we deleting the SuppressMessage?

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -773,14 +787,13 @@ void IODataStreamListener.StreamRequested()
...
 
-
         /// <summary>
         /// This method is called when an async stream is requested. It is a no-op.
         /// </summary>
         /// <returns>A task for method called when a stream is requested.</returns>
         Task IODataStreamListener.StreamRequestedAsync()
         {
-            return TaskUtils.GetTaskForSynchronousOperation(() => ((IODataStreamListener)this).StreamRequested());
+            return TaskUtils.CompletedTask;
```
KenitoInc, line 796: You can remove the `IODataStreamListener.StreamRequested()` if it's not needed anymore

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -2374,11 +2629,6 @@ private void EnterScope(WriterState newState, ODataItem item)
                                     nestedResourceInfo.IsCollection = navigationProperty.Type.IsEntityCollectionType();
                                 }
 
-                                if (!nestedResourceInfo.IsCollection.HasValue)
-                                {
```
KenitoInc, line 2378: Why was this code removed?

gathogojr: @KenitoInc Because the block was repeated
image

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
gathogojr: Because the `Debug.Assert` statements refer to `this` so I reasoned that the "MarkMembersAsStatic" warning would no longer valid.

src/Microsoft.OData.Core/ODataWriterCore.cs
gathogojr: It'd not be safe to remove the method since this class implements `IODataOutputInStreamErrorListener` hence the requirement to implement all the methods including `StreamRequested` and `StreamRequestedAsync`. Its actually an extensibility hook, just like `StreamDisposed` and `StreamDisposedAsync`

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeltaWriter.cs
```diff
@@ -290,7 +290,7 @@ void IODataOutputInStreamErrorListener.OnInStreamError()
         /// <inheritdoc/>
         Task IODataOutputInStreamErrorListener.OnInStreamErrorAsync()
         {
-            throw new NotImplementedException();
+            return this.inStreamErrorListener.OnInStreamErrorAsync();
```
marabooy, line 293: could we have a short test to test that this does get called.

marabooy (approved):
Looks good to me; might need a test for OnInStreamErrorAsync

habbes (approved):

xuzhg (approved):

KenitoInc (approved):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeltaWriter.cs
gathogojr: @marabooy I'll add the tests on a PR I'm working on for `ODataJsonLightOutputContext` since the logic for writing an error response is triggered from there.

PR #2085, opened 2021-05-12
Add System.Runtime.CompilerServices.Unsafe and System.Text.Encodings.web dependency for OData.Edm

xuzhg:
See detail at: https://github.com/dotnet/aspnetcore/issues/31725

### Issues

### Description

v2.1.27 nuget package pins a number of libraries to 4.6.0 not inclusive. This breaks downstream dependencies when combined with the OData libraries.

It's better to add "System.Runtime.CompilerServices.Unsafe and System.Text.Encodings.web" dependency to unblock customer to use latest .net core 2.1.27

- [ ] *Test cases added*

### Additional work necessary


mikepizzo (approved):
:shipit:

KenitoInc (approved):

Issue #2081, opened 2021-05-07
Context URL does not include STAR (*) when no selectors are specified


cryogen:
*As reported in https://github.com/OData/WebApi/issues/2088, the context URL is incorrect causing issues with Excel and Power BI*

### Assemblies affected

*Microsoft.OData.Core 7.x*

### Reproduce steps

*Start the OData sample project AspNetCore3xODataSample.Web
Open Excel
Go To Data -> Get Data -> From Other Sources -> OData Feed
Enter the URL to your application http://{host}/odata/Customers?$select=Name&$expand=Order*

### Expected result

*Excel has no issue in bringing back data from the OData API*

### Actual result

*An error is shown "The odata.context url 'http://{host}/odata/$metadata#Customers(Name,Order())' is invalid"*

### Additional detail

*My interpretation of the OData specification suggests that the STAR (*) character is required when there are no selectors specified. I have a fix that I can PR if this is accepted as an issue*

Sreejithpin:
Thanks for the issue and thanks for offering the PR. It would be great if you can put  a PR for this . Thanks again

DaleyKD:
@cryogen : Did you ever find a work around for this? We recently updated our nugets, and it broke our customers who used our OData API for Power BI.

cryogen:
@DaleyKD Sorry, I don't actually remember. I know that we did fix it on our end but I can't remember how and I no longer work in the job that needed this

PR #2086, opened 2021-05-12
Add a * (STAR) in the OData context URL when expanding an entity

cryogen:
### Issues

*This pull request fixes issue https://github.com/OData/WebApi/issues/2088 and #2081*

### Description

*This change adds the STAR character (*) to the OData context URL when an entity is expanded and no selectors are specified. The way I (and apparently the team behind the Excel and PowerBI OData connector) read the OData spec suggests this character is required. As a result, it is not possible to use Excel or PowerBI to expand any data from an odata.net backed API. Test cases adjusted to match the change.*

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/ODataContextUrlInfo.cs
+++ b/src/Microsoft.OData.Core/ODataContextUrlInfo.cs
@@ -420,7 +420,7 @@ namespace Microsoft.OData
         /// <returns>The generated expand string.</returns>
         private static string ProcessSubExpand(string expandNode, string subExpand)
         {
-            return expandNode + ODataConstants.ContextUriProjectionStart + subExpand + ODataConstants.ContextUriProjectionEnd;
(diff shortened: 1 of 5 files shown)


mikepizzo:
We'll need to review this ramifications of this change carefully. I would love to fix Excel/PowerBI, but want to make sure we don't end up breaking other clients that are expecting the empty parens.

One option we could consider is the introduction of a MessageWriterSetting that a service can set to control whether or not the parens contains a star in the absence of expand options.

cryogen:
I understand the concern but at the end of the day this is the OData spec and it should be compliant. I've looked into adding the option but it's beyond my understanding of the code at this point in time

mikepizzo (changes requested):
Thanks for the conribution!

The OData spec does not require the * and, unfortunately, adding it would break many existing applications that are not expecting the *.

We can look at adding it as an option, but beware that setting the option is likely to break existing clients.

cryogen:
I’ll close this in that case. Making the option is beyond my knowledge of the project and time commitment :)

thanks for the response

Issue #1531, opened 2019-09-02
Exception if the entity has Equals and GetHashcode implementation


DarkGraySun:
I want to build a project with an OData interface. There exists a server and a client for this.
Server and client use an identical library of classes with which the exchange between server and client takes place (EDMModel). The problem arises when these classes (Entities) have the functions Equals and GetHashCode.
Then the client can not perform a successful update or add on the server.

### Assemblies affected
Microsoft.OData.Client 7.6.0

### Reproduce steps

The Client (DataserviceContext) uses an entity that implements "Equals" and "GetHashcode" by using the id (key) of the class.

The server receives the entity changes the id (store the entity in the database) and returns the entity with the changed id to the client.

Then the client gets an exception when receiving the response

### Expected result

The client gets no exception..

### Actual result

The client gets an exception.

Microsoft.OData.Client.DataServiceRequestException: An error occurred while processing this request. ---> System.InvalidOperationException: The context is already tracking a different entity with the same resource Uri.

Stacktrace:
 at Microsoft.OData.Client.EntityTracker.InternalAttachEntityDescriptor(EntityDescriptor entityDescriptorFromMaterializer, Boolean failIfDuplicated)
   at Microsoft.OData.Client.AtomMaterializerLog.ApplyToContext()
   at Microsoft.OData.Client.MaterializeAtom.MoveNextInternal()
   at Microsoft.OData.Client.MaterializeAtom.MoveNext()
   at Microsoft.OData.Client.BaseSaveResult.MaterializeResponse(EntityDescriptor entityDescriptor, ResponseInfo responseInfo, String etag)
   at Microsoft.OData.Client.BaseSaveResult.HandleResponsePost(EntityDescriptor entityDescriptor, String etag)
   at Microsoft.OData.Client.BaseSaveResult.HandleResponsePost(Descriptor descriptor, HeaderCollection contentHeaders)
   at Microsoft.OData.Client.SaveResult.HandleResponse()
   --- End of inner exception stack trace ---
   at Microsoft.OData.Client.SaveResult.HandleResponse()
   at Microsoft.OData.Client.BaseSaveResult.EndRequest()
   at Microsoft.OData.Client.DataServiceContext.EndSaveChanges(IAsyncResult asyncResult)
   at System.Threading.Tasks.TaskFactory`1.FromAsyncCoreLogic(IAsyncResult iar, Func`2 endFunction, Action`1 endAction, Task`1 promise, Boolean requiresSynchronization)
[...]

PR #2090, opened 2021-05-17
entity tracker equality comparer

ElizabethOkerio:
### Issues

*This pull request fixes issue #1531 *.
Having such a POCO class when working with <code> OData Client </code> leads to errors when you try to perform save or update operations. 
```csharp
public class ExampleObject
    {
        [Key]
        public int Id { get; set; }
        public bool Equals(ExampleObject other)
        {
            if (ReferenceEquals(other, null))
            {
                return false;
            }

            return other.Id == Id;
        }
        public override bool Equals(object obj)
        {
            var other = obj as ExampleObject;

            if (ReferenceEquals(other, null))
            {
                return false;
            }

            return other.Id == Id;
        }
        public override int GetHashCode()
        {
            return Id.GetHashCode();
        }
    }

```
The reason for the exceptions is because the <code> EntityTracker </code> uses the default EqualityComparer. 

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/EntityTracker.cs
+++ b/src/Microsoft.OData.Client/EntityTracker.cs
@@ -14,6 +14,7 @@ namespace Microsoft.OData.Client
     using System.Diagnostics.CodeAnalysis;
     using System.Linq;
     using Microsoft.OData.Client.Metadata;
+    using System.Runtime.CompilerServices;
     #endregion Namespaces
 
     /// <summary>
@@ -26,28 +27,12 @@ namespace Microsoft.OData.Client
 
         #region Resource state management
 
-    #if PORTABLELIB && WINDOWSPHONE
         /// <summary>Set of tracked resources</summary>
-        private Dictionary<object, EntityDescriptor> entityDescriptors = new Dictionary<object, EntityDescriptor>(EqualityComparer<object>.Default);
-    #else
-        /// <summary>Set of tracked resources</summary>
-        private ConcurrentDictionary<object, EntityDescriptor> entityDescriptors = new ConcurrentDictionary<object, EntityDescriptor>(EqualityComparer<object>.Default);
-    #endif
-    #if PORTABLELIB && WINDOWSPHONE
-        /// <summary>Set of tracked resources by Identity</summary>
-        private Dictionary<Uri, EntityDescriptor> identityToDescriptor;
-    #else
+        private ConcurrentDictionary<object, EntityDescriptor> entityDescriptors = new ConcurrentDictionary<object, EntityDescriptor>(EntityEqualityComparer<object>.Instance);
         /// <summary>Set of tracked resources by Identity</summary>
         private ConcurrentDictionary<Uri, EntityDescriptor> identityToDescriptor;
-    #endif
-
-    #if PORTABLELIB && WINDOWSPHONE
-        /// <summary>Set of tracked bindings</summary>
-        private Dictionary<LinkDescriptor, LinkDescriptor> bindings;
-    #else
         /// <summary>Set of tracked bindings</summary>
         private ConcurrentDictionary<LinkDescriptor, LinkDescriptor> bindings;
-    #endif
 
         /// <summary>change order</summary>
         private uint nextChange;
@@ -575,5 +560,17 @@ namespace Microsoft.OData.Client
                 throw Error.InvalidOperation(Strings.Context_DifferentEntityAlreadyContained);
             }
         }
+    }
+
+    /// <summary>
+    /// An object instance equality comparer
+    /// </summary>
+    /// <typeparam name="TObject"></typeparam>
+    internal class EntityEqualityComparer<TObject> : EqualityComparer<TObject> where TObject : class
+    {
+        public static EntityEqualityComparer<TObject> Instance = new EntityEqualityComparer<TObject>();
+        private EntityEqualityComparer() { }
+        public override bool Equals(TObject x, TObject y) => ReferenceEquals(x, y);
+        public override int GetHashCode(TObject obj) => RuntimeHelpers.GetHashCode(obj);
     }
 }
--- a/test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/ClientEntityTrackerTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/ClientEntityTrackerTests.cs
@@ -1,0 +1,374 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ClientEntityTrackerTests.cs" company="Microsoft">
+// Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Client.Tests.Tracking
+{
+    using System;
+    using System.Collections.Generic;
+    using System.IO;
+    using System.Threading.Tasks;
+    using System.Xml;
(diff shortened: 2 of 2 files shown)


src/Microsoft.OData.Client/EntityTracker.cs
```diff
@@ -576,4 +576,12 @@ private void ValidateDuplicateIdentity(Uri identity, EntityDescriptor descriptor
             }
         }
     }
+
+    public class CustomComparer<TObject> : IEqualityComparer<TObject> where TObject : class
```
marabooy, line 580:
```suggestion
    public class EntityDescriptorComparer<T> : IEqualityComparer<T> where T: class
```
Since the class is public do add some class documentation for it.

marabooy (approved):

marabooy (changes requested):
Add a test for this to show the issue has been fixed.

src/Microsoft.OData.Client/EntityTracker.cs
```diff
@@ -26,12 +26,12 @@ public class EntityTracker : EntityTrackerBase
 
         #region Resource state management
 
-    #if PORTABLELIB && WINDOWSPHONE
+#if PORTABLELIB && WINDOWSPHONE
```
gathogojr, line 29: @ElizabethOkerio For some time now, we have been getting rid of these `PORTABLELIB` blocks. I think you should get rid of this too

src/Microsoft.OData.Client/EntityTracker.cs
```diff
@@ -26,12 +26,12 @@ public class EntityTracker : EntityTrackerBase
...
         /// <summary>Set of tracked resources</summary>
-        private Dictionary<object, EntityDescriptor> entityDescriptors = new Dictionary<object, EntityDescriptor>(EqualityComparer<object>.Default);
-    #else
+        private Dictionary<object, EntityDescriptor> entityDescriptors = new Dictionary<object, EntityDescriptor>(CustomComparer<object>.Instance);
+#else
         /// <summary>Set of tracked resources</summary>
-        private ConcurrentDictionary<object, EntityDescriptor> entityDescriptors = new ConcurrentDictionary<object, EntityDescriptor>(EqualityComparer<object>.Default);
+        private ConcurrentDictionary<object, EntityDescriptor> entityDescriptors = new ConcurrentDictionary<object, EntityDescriptor>(CustomComparer<object>.Instance);
     #endif
     #if PORTABLELIB && WINDOWSPHONE
```
gathogojr, line 36: Maybe remove this `PORTABLELIB` block too

src/Microsoft.OData.Client/EntityTracker.cs
gathogojr, line 580: I'd probably name this `EntityEqualityComparer` or something more contextual rather than just `CustomComparer`

src/Microsoft.OData.Client/EntityTracker.cs
```diff
@@ -576,4 +576,12 @@ private void ValidateDuplicateIdentity(Uri identity, EntityDescriptor descriptor
             }
         }
     }
+
+    public class CustomComparer<TObject> : IEqualityComparer<TObject> where TObject : class
+    {
+        public static CustomComparer<TObject> Instance = new CustomComparer<TObject>();
+        private CustomComparer() { }
+        public bool Equals(TObject x, TObject y) => ReferenceEquals(x, y);
+        public int GetHashCode(TObject obj) => System.Runtime.CompilerServices.RuntimeHelpers.GetHashCode(obj);
```
gathogojr, line 585: Import `System.Runtime.CompilerServices` namespace

gathogojr:
Top of my mind, there are three tests I'd recommend that you add to verify that the `EntityTracker` works as expected against:
1. POCO classes that override `Equals` and `GetHashCode`
2. POCO classes that implement `IEqualityComparer` interface
3. POCO classes that do neither of the above

marabooy:
> Top of my mind, there are three tests I'd recommend that you add to verify that the `EntityTracker` works as expected against:
> 
> 1. POCO classes that override `Equals` and `GetHashCode`
> 2. POCO classes that implement `IEqualityComparer` interface
> 3. POCO classes that do neither of the above
4. Classes that extend BaseEntity and have the same as above.

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/ClientEntityTrackerTests.cs
```diff
@@ -0,0 +1,243 @@
...
+            {
+                await dataServiceContext.SaveChangesAsync();
+            }
+        }
+
+        private void SetupContextWithRequestPipelineForSaving(DataServiceContext[] dataServiceContexts, string type)
+        {
+            var response = "";
+            var location = "";
+            if (type == "ExampleObject")
```
marabooy, line 144: Refactor this to move the type out of the method and pass the expected response and location so as to reduce the length of the method.

marabooy (approved):

src/Microsoft.OData.Client/EntityTracker.cs
```diff
@@ -576,4 +561,16 @@ private void ValidateDuplicateIdentity(Uri identity, EntityDescriptor descriptor
             }
         }
     }
+
+    /// <summary>
+    /// An object instance equality comparer
+    /// </summary>
+    /// <typeparam name="TObject"></typeparam>
+    public class EntityEqualityComparer<TObject> : IEqualityComparer<TObject> where TObject : class
```
KenitoInc, line 569: From the docs, they recommend deriving from `EqualityComparer<T>` instead of the `IEqualityComparer<T>`. Is it something you can consider? https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.iequalitycomparer-1?view=net-5.0

KenitoInc (commented):
Just one comment.
Overall the PR looks good.

src/Microsoft.OData.Client/EntityTracker.cs
habbes, line 569: Should this class be `public`?

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/ClientEntityTrackerTests.cs
```diff
@@ -0,0 +1,225 @@
...
+        #endregion
+
+        public ClientEntityTrackerTests()
+        {
+            var uri = new Uri("http://localhost:8000");
+            DefaultTrackingContext = new Container(uri);
+        }
+
+        [Fact]
+        public async Task SavingObject_ThatOveridesEqualsAndGetHashCodeMethods_SavesSuccessfully()
```
habbes, line 82:
```suggestion
        public async Task SavingObject_ThatOverridesEqualsAndGetHashCodeMethods_SavesSuccessfully()

```

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/ClientEntityTrackerTests.cs
```diff
@@ -0,0 +1,225 @@
...
+        }
+
+        public int GetHashCode(ExampleObject obj)
+        {
+            return obj.Id.GetHashCode();
+        }
+    }
+
+    [Key("Id")]
+    public class ExampleObjectB : BaseEntityType
```
habbes, line 220: Could you also add tests cases for POCO classes that do not inherit from `BaseEntityType`?

KenitoInc (approved):

Issue #2074, opened 2021-04-30
JsonWriter does not use ICharArrayPool when writing strings


habbes:
The [JsonWriter](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Json/JsonWriter.cs) contains an `ArrayPool` property of type [ICharArrayPool](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Buffers/ICharArrayPool.cs). It's supposed to pass this array pool to the `JsonValueUtils` `Write*` methods that write a string value. The array pool is used to rent a char buffer that's used when escaping a string or expanding binary data to base64. [If the array pool is not provided, then a new char array will be allocated instead.](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Buffers/BufferUtils.cs#L51-L65). 
The following methods in `JsonValueUtils` accept an `ICharArrayPool arrayPool` parameter:

- [WriteValue](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Json/JsonValueUtils.cs#L317-L322) that accepts a `string` value
- [WriteValue](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Json/JsonValueUtils.cs#L343) that accepts a `byte[]` value
- [WriteEscapedJsonString](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Json/JsonValueUtils.cs#L394-L399)

We can see that the `JsonWriter` correctly passed the `ArrayPool` when writing a binary string: https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Json/JsonWriter.cs#L408

However, in the following cases, it's not passing the `ArrayPool` parameter:
- https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Json/JsonWriter.cs#L218
- https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Json/JsonWriter.cs#L282-L283
- https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Json/JsonWriter.cs#L322-L323
- https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Json/JsonWriter.cs#L398

### Assemblies affected

Microsoft.OData.Core 7.x

### Reproduce steps

N/A (see linked code snippets)

### Expected result

```c#
[...]

PR #2093, opened 2021-05-20
Consistently use CharArrayPool in JsonWriter to cut cost of initializing char arrays

gathogojr:
### Issues

*This pull request fixes issue #2074.*

### Description

Consistently use `CharArrayPool` in `JsonWriter` to cut cost of initializing char arrays

**Notes:**
- `CharArrayPool` is already used in `JsonWriter` class. It was just not being passed along in some instances.
- There's no real value in passing the `CharArrayPool` parameter for the methods listed below because we only use it if the input has special characters. I decided to still pass the parameter just to keep things consistent and avoid confusion.
  - `WriteName(string)`/`WriteNameAsync(string)` - name _should_ not contain special chars
  - `WriteValue(long)`/`WriteValueAsync(long)` - - special chars not expected when the value is converted to string (IEEE 754 compatible)
  - `WriteValue(decimal)`/`WriteValueAsync(decimal)` - special chars not expected when the value is converted to string (IEEE 754 compatible)

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriter.cs
@@ -215,7 +215,7 @@ namespace Microsoft.OData.Json
 
             currentScope.ObjectCount++;
 
-            JsonValueUtils.WriteEscapedJsonString(this.writer, name, this.stringEscapeOption, this.wrappedBuffer);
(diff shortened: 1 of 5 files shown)


habbes (approved):

marabooy (approved):
nit: Wrap the arguments.

Issue #2094, opened 2021-05-20
ExecuteAsync with CancellationToken is missing on some QueryType objects


HugzM:
it seems DataServiceActionQuery and DataServiceActionQueryofT does not implement the ExecuteAsync(CancellationToken ) like DataServiceActionQuerySingleOfT does.

### Assemblies affected

Microsoft.odata.client 7.9.0.0

PR #2097, opened 2021-05-26
Fix/2094 Add cancellation token to ExecuteAsync in DataServiceActionQuery and DataServiceActionQueryofT 

KenitoInc:
### Issues

*This pull request fixes issue #2094 .*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceActionQuery.cs
+++ b/src/Microsoft.OData.Client/DataServiceActionQuery.cs
@@ -68,7 +68,7 @@ namespace Microsoft.OData.Client
         /// <returns>A task represents the result of the operation. </returns>
         public Task<OperationResponse> ExecuteAsync()
         {
-            return Context.ExecuteAsync(this.RequestUri, XmlConstants.HttpMethodPost, Parameters);
+            return this.ExecuteAsync(CancellationToken.None);
         }
 
         /// <summary>Asynchronously sends the request so that this call does not block processing while waiting for the results from the service.</summary>
--- a/src/Microsoft.OData.Client/DataServiceActionQuery.cs
+++ b/src/Microsoft.OData.Client/DataServiceActionQuery.cs
@@ -7,6 +7,7 @@
 namespace Microsoft.OData.Client
 {
     using System;
+    using System.Threading;
     using System.Threading.Tasks;
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


ElizabethOkerio (approved):

src/Microsoft.OData.Client/DataServiceActionQuery.cs
```diff
@@ -70,6 +71,14 @@ public Task<OperationResponse> ExecuteAsync()
             return Context.ExecuteAsync(this.RequestUri, XmlConstants.HttpMethodPost, Parameters);
         }
 
+        /// <summary>Asynchronously sends the request so that this call does not block processing while waiting for the results from the service.</summary>
+        /// <returns>A task represents the result of the operation. </returns>
+        /// <param name="cancellationToken">The token to monitor for cancellation requests.</param>
+        public Task<OperationResponse> ExecuteAsync(CancellationToken cancellationToken)
```
marabooy, line 77: Won't this break some existing clients? Due to the public api change?

marabooy (approved):

src/Microsoft.OData.Client/DataServiceActionQuery.cs
KenitoInc: Nope, We added an overloaded method. The original method with no parameters still exist

src/Microsoft.OData.Client/DataServiceActionQueryOfT.cs
```diff
@@ -73,6 +74,14 @@ public Task<IEnumerable<T>> ExecuteAsync()
             return Context.ExecuteAsync<T>(this.RequestUri, XmlConstants.HttpMethodPost, false, Parameters);
```
habbes, line 74: Maybe you should refactor this as:
```
public Task<IEnumerable<T>> ExecuteAsync()
{
      return ExecuteAsync(CancellationToken.None);
}
```

Or make `cancellationToken` an optional parameter for `ExecuteAsync`?

habbes (commented):
Looks good. Consider refactoring to avoid having separate/duplicate implementations of `ExecuteAsync()`.

src/Microsoft.OData.Client/DataServiceActionQueryOfT.cs
KenitoInc: I have updated based on the suggested code.

gathogojr (approved):
LGTM

PR #2109, opened 2021-06-17
Implement asynchronous support in ODataWriteStream

gathogojr:
### Issues

### Description

Implement asynchronous support in `ODataWriteStream`
- Ensures that if `Dispose` method of `ODataWriteStream` is fired when writing asynchronously, `StreamDisposedAsync` method of `IODataStreamListener` instance is fired instead of the synchronous `StreamDisposed` method.
- Implements `IAsyncDisposable` interface in `ODataWriteStream` such that its possible to use `await using` for frameworks that implement NETSTANDARD2_0

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataStream.cs
+++ b/src/Microsoft.OData.Core/ODataStream.cs
@@ -10,6 +10,7 @@ namespace Microsoft.OData
     using System;
     using System.Diagnostics;
     using System.IO;
+    using System.Threading.Tasks;
     #endregion Namespaces
 
     /// <summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.OData.Core/ODataStream.cs
```diff
@@ -50,20 +62,53 @@ public override long Seek(long offset, SeekOrigin origin)
...
+        /// </summary>
+        /// <returns>A task representing the asynchronous operation.</returns>
+        protected virtual async ValueTask DisposeAsyncCore()
+        {
+            if (!this.disposed && this.listener != null)
+            {
+                await this.listener.StreamDisposedAsync()
+                    .ConfigureAwait(false);
+
+                this.listener = null;
```
mikepizzo, line 107: In the async version you never set this.disposed=true like you do in the sync version. You do set listener to false, which ensures it won't be executed again (although the synchronous dispose would pass the check on line 65 and then be a no-op).

So I think it all works, I'm just wondering why you rely on both disposed=false and listener!=null and handle them differently between sync and async.

mikepizzo (approved):
:shipit:

src/Microsoft.OData.Core/ODataStream.cs
gathogojr: @mikepizzo `DisposeAsyncCore` is called from `DisposeAsync` and following the [guidance from Task Asynchronous Programming team](https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#implement-both-dispose-and-async-dispose-patterns), `DisposeAsync` in turn calls `Dispose` method, passing `false` as a parameter to ensure functional equivalency with the synchronous pattern. `Dispose` method in turn sets `disposed` to `true` and that way idempotency is guaranteed.

mikepizzo: I missed that DisposeAsync calls Dispose -- makes sense, thanks John.

marabooy (approved):

PR #2080, opened 2021-05-06
Support asynchronous invocation of IODataStreamListener.StreamDisposedAsync method from ODataNotificationStream class

gathogojr:
### Issues

This pull request is in partial fulfilment of issue #2019.

### Description

Support asynchronous invocation of **`IODataStreamListener.StreamDisposedAsync`** method from `Dispose`/`DisposeAsync` methods of **`ODataNotificationStream`** class.
- In .NET Core 3.1 or later, we're able to ride on asynchronous `DisposeAsync` method defined in `IAsyncDisposable` interface to avoid blocking.
- In platforms below .NET Core 3.1, we're forced to `.Wait()` on the asynchronous method.

### Checklist (Uncheck if it is not completed)

- [x] *Test cases added*
- [ ] *Build and test with one-click build and test script passed*

### Additional work necessary

*If documentation update is needed, please add "Docs Needed" label to the issue and provide details about the required document change in the issue.*

PR #2110, opened 2021-06-17
ODataNotificationWriter and ODataNotificationStream wrapper classes should not dispose the passed TextWriter and Stream respectively

gathogojr:
### Description

`ODataNotificationWriter` and `ODataNotificationStream` wrapper classes should not dispose the passed `TextWriter` and `Stream` respectively.

Reference to merged PRs [2079](https://github.com/OData/odata.net/pull/2079) and [2080](https://github.com/OData/odata.net/pull/2080) implementing support for asynchronous invocation of `IODataStreamListener.StreamDisposedAsync` from `ODataNotificationWriter` and `ODataNotificationStream` wrapper classes, a line was added such that the passed `TextWriter` and `Stream` respectively are disposed when the `Dispose` method is called.
This can cause a problem if the owner of `TextWriter` or `Stream` chose to use it after - _though that does not happen currently_. This pull request addresses that loop hole, leaving the owner to do the disposing.

**NOTE:** Confirmation of the fix is in the change made to the tests where the test class is able to read from the stream even after the `ODataNotificationWriter` or `ODataNotificationStream` instance is disposed.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationStreamTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationStreamTests.cs
@@ -32,7 +32,7 @@ namespace Microsoft.OData.Tests
             // We care about the notification stream being disposed
             // We don't care about the stream passed to the notification stream
             using (var notificationStream = new ODataNotificationStream(
-                new MemoryStream(),
+                this.stream,
                 this.streamListener,
                 synchronous))
             {
@@ -49,7 +49,7 @@ namespace Microsoft.OData.Tests
         public void NotificationStreamDisposeShouldBeIdempotent(bool synchronous, string expected)
         {
             var notificationStream = new ODataNotificationStream(
-                new MemoryStream(),
+                this.stream,
                 this.streamListener,
                 synchronous);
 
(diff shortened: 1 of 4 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationStreamTests.cs
```diff
@@ -32,7 +32,7 @@ public void NotificationStreamDisposeShouldInvokeStreamDisposed(bool synchronous
             // We care about the notification stream being disposed
             // We don't care about the stream passed to the notification stream
             using (var notificationStream = new ODataNotificationStream(
-                new MemoryStream(),
+                this.stream,
```
habbes, line 35: Why this change necessary? Is `this.stream` used in the test? Do you make a check to verify that it was not disposed?

habbes (approved):

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationStreamTests.cs
gathogojr: Please observe that I'm verifying that `StreamDisposed` or `StreamDiposedAsync` are invoked by reading the content of the stream they wrote to when they were invoked. In the previous case, the two methods were writing to a stream entirely different to the one passed to `ODataNotificationStream` constructor. Since `ODataNotifictionStream` is not disposing the stream after the change, the same stream being passed to the constructor is the same one that `StreamDisposed`/`StreamDiposedAsync` are writing to, with the assurance that it'll still be available for the  the helper `ReadStreamContents`/`ReadStreamContentsAsync` methods to read from. That is why I noted in the PR description that the change I made to the tests is the confirmation needed to verify that the stream is not disposed

KenitoInc (approved):

marabooy (approved):

PR #2111, opened 2021-06-21
Implement asynchronous support in ODataJsonLightBatchWriter

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataJsonLightBatchWriter`**

To prevent a breaking change, I added the following 6 `protected virtual` asynchronous methods to the `public abstract ODataBatchWriter` class as asynchronous wrappers over the synchronous methods.
- `WriteStartBatchImplementationAsync`
- `WriteEndBatchImplementationAsync`
- `WriteStartChangesetImplementationAsync`
- `WriteEndChangesetImplementationAsync`
- `CreateOperationRequestMessageImplementationAsync`
- `CreateOperationResponseMessageImplementationAsync`

These virtual asynchronous methods are then invoked from the public asynchronous methods. For example;

```csharp
public Task WriteStartBatchAsync()
{
    // ...
    return TaskUtils.GetTaskForSynchronousOperation(this.WriteStartBatchImplementation);
}

protected abstract void WriteEndBatchImplementation();
```
becomes
```csharp
public Task WriteStartBatchAsync()
{
    // ...
    return this.WriteStartBatchImplementationAsync();
}

protected abstract void WriteEndBatchImplementation();

protected virtual Task WriteStartBatchImplementationAsync()
{
    return TaskUtils.GetTaskForSynchronousOperation(this.WriteStartBatchImplementation);
}
```
This I believe should guarantee asynchronous methods in the abstract base class will work as before unless overridden in a derived class.
It also allows asynchronous support to be implemented by overriding the protected methods in the derived class(es).

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Batch/ODataBatchUtils.cs
+++ b/src/Microsoft.OData.Core/Batch/ODataBatchUtils.cs
@@ -110,12 +110,13 @@ namespace Microsoft.OData
         /// <returns>A new <see cref="ODataWriteStream"/> instance.</returns>
         internal static ODataWriteStream CreateBatchOperationWriteStream(
             Stream outputStream,
-            IODataStreamListener operationListener)
+            IODataStreamListener operationListener,
+            bool synchronous = true)
         {
             Debug.Assert(outputStream != null, "outputStream != null");
             Debug.Assert(operationListener != null, "operationListener != null");
 
-            return new ODataWriteStream(outputStream, operationListener);
+            return new ODataWriteStream(outputStream, operationListener, synchronous);
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
+++ b/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
@@ -56,7 +56,7 @@ namespace Microsoft.OData
         /// <summary>
         /// Whether the writer is currently processing inside a changeset or atomic group.
         /// </summary>
-        private bool isInChangset;
+        private bool isInChangeset;
 
         /// <summary>The batch-specific URL converter that stores the content IDs found in a changeset and supports resolving cross-referencing URLs.</summary>
         internal readonly ODataBatchPayloadUriConverter payloadUriConverter;
@@ -167,10 +167,11 @@ namespace Microsoft.OData
 
         /// <summary>Asynchronously starts a new batch; can be only called once and as first call.</summary>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
-        public Task WriteStartBatchAsync()
+        public async Task WriteStartBatchAsync()
         {
             this.VerifyCanWriteStartBatch(false);
-            return TaskUtils.GetTaskForSynchronousOperation(this.WriteStartBatchImplementation);
+            await this.WriteStartBatchImplementationAsync()
+                .ConfigureAwait(false);
         }
 
         /// <summary>Ends a batch; can only be called after WriteStartBatch has been called and if no other active changeset or operation exist.</summary>
@@ -185,13 +186,13 @@ namespace Microsoft.OData
 
         /// <summary>Asynchronously ends a batch; can only be called after WriteStartBatch has been called and if no other active change set or operation exist.</summary>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
-        public Task WriteEndBatchAsync()
+        public async Task WriteEndBatchAsync()
         {
             this.VerifyCanWriteEndBatch(false);
-            return TaskUtils.GetTaskForSynchronousOperation(this.WriteEndBatchImplementation)
-
+            await this.WriteEndBatchImplementationAsync()
                 // Note that we intentionally go through the public API so that if the Flush fails the writer moves to the Error state.
-                .FollowOnSuccessWithTask(task => this.FlushAsync());
+                .FollowOnSuccessWithTask(task => this.FlushAsync())
+                .ConfigureAwait(false);
         }
 
         /// <summary>
@@ -231,13 +232,14 @@ namespace Microsoft.OData
         /// <param name="changesetId"> The change set Id of the batch request. Cannot be null.</param>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
         /// <exception cref="ODataException">Thrown if the <paramref name="changesetId"/> is null.</exception>
-        public Task WriteStartChangesetAsync(string changesetId)
+        public async Task WriteStartChangesetAsync(string changesetId)
         {
             ExceptionUtils.CheckArgumentNotNull(changesetId, "changesetId");
 
             this.VerifyCanWriteStartChangeset(false);
-            return TaskUtils.GetTaskForSynchronousOperation(() => this.WriteStartChangesetImplementation(changesetId))
-                .FollowOnSuccessWith(t => this.FinishWriteStartChangeset());
+            await this.WriteStartChangesetImplementationAsync(changesetId)
+                .FollowOnSuccessWith(t => this.FinishWriteStartChangeset())
+                .ConfigureAwait(false);
         }
 
         /// <summary>Ends an active changeset; this can only be called after WriteStartChangeset and only once for each changeset.</summary>
@@ -250,11 +252,12 @@ namespace Microsoft.OData
 
         /// <summary>Asynchronously ends an active change set; this can only be called after WriteStartChangeset and only once for each change set.</summary>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
-        public Task WriteEndChangesetAsync()
+        public async Task WriteEndChangesetAsync()
         {
             this.VerifyCanWriteEndChangeset(false);
-            return TaskUtils.GetTaskForSynchronousOperation(this.WriteEndChangesetImplementation)
-                .FollowOnSuccessWith(t => this.FinishWriteEndChangeset());
+            await this.WriteEndChangesetImplementationAsync()
+                .FollowOnSuccessWith(t => this.FinishWriteEndChangeset())
+                .ConfigureAwait(false);
         }
 
         /// <summary>Creates an <see cref="Microsoft.OData.ODataBatchOperationRequestMessage" /> for writing an operation of a batch request.</summary>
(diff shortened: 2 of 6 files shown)


marabooy (approved):

KenitoInc (changes requested):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchWriter.cs
```diff
@@ -83,6 +83,11 @@ internal sealed class ODataJsonLightBatchWriter : ODataBatchWriter
         /// </summary>
         private readonly IJsonWriter jsonWriter;
 
+        /// <summary>
+        /// The underlying asynchronous JSON writer.
+        /// </summary>
+        private readonly IJsonWriterAsync asynchronousJsonWriter;
```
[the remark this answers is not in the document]
gathogojr: By the time we were having the discussion, I had already merged the PR that introduced this interface. I therefore decided to waiting for a major release opportunity to do the clean up. In any case, when the major release opportunity presents itself, `IJsonWriterAsync` will be merged into `IJsonWriter`. We only introduced `IJsonWriterAsync` to prevent a breaking change since `IJsonWriter` is a public interface.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchWriter.cs
```diff
@@ -266,18 +302,18 @@ protected override IEnumerable<string> GetDependsOnRequestIds(IEnumerable<string
         /// </summary>
         /// <param name="contentId">The content Id.</param>
         /// <param name="dependsOnIds">The dependsOn ids specifying current request's prerequisites.</param>
-        protected override void ValidateDependsOnIds(string contentId, IEnumerable<string> dependsOnIds)
+protected override void ValidateDependsOnIds(string contentId, IEnumerable<string> dependsOnIds)
```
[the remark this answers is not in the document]
gathogojr: My bad! The change was inadvertent. I'll fix it

src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
```diff
@@ -573,11 +571,91 @@ protected void SetState(BatchWriterState newState)
...
+        protected virtual Task WriteStartBatchImplementationAsync()
+        {
+            return TaskUtils.GetTaskForSynchronousOperation(this.WriteStartBatchImplementation);
+        }
+
+        /// <summary>
+        /// Asynchronously ends a batch.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        protected virtual Task WriteEndBatchImplementationAsync()
```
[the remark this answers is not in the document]
gathogojr: Because this method as defined here is a pass-through.
`GetTaskForSynchronousOperation` method defined in `TaskUtils`. The way it is implemented is such that it executes the synchronous operation in a try...catch and returns a completed task if the execution is successful and a faulted task if not (_unless the exception is not "catchable"_).
The fact the `GetTaskForSynchronousOperation` returns a task and we are doing nothing else within `WriteEndBatchImplementationAsync` apart from calling the method means its safe to elide async/await in this particular case.

KenitoInc (approved):
LGTM

wandeg (approved):

src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
```diff
@@ -185,13 +186,13 @@ public void WriteEndBatch()
...
-        public Task WriteEndBatchAsync()
+        public async Task WriteEndBatchAsync()
         {
             this.VerifyCanWriteEndBatch(false);
-            return TaskUtils.GetTaskForSynchronousOperation(this.WriteEndBatchImplementation)
-
+            await this.WriteEndBatchImplementationAsync()
                 // Note that we intentionally go through the public API so that if the Flush fails the writer moves to the Error state.
-                .FollowOnSuccessWithTask(task => this.FlushAsync());
+                .FollowOnSuccessWithTask(task => this.FlushAsync())
```
habbes, line 194: Out of curiosity, do you know why `ContinueWith` is not used here instead?

src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
```diff
@@ -231,13 +232,14 @@ public Task WriteStartChangesetAsync()
...
-        public Task WriteStartChangesetAsync(string changesetId)
+        public async Task WriteStartChangesetAsync(string changesetId)
         {
             ExceptionUtils.CheckArgumentNotNull(changesetId, "changesetId");
 
             this.VerifyCanWriteStartChangeset(false);
-            return TaskUtils.GetTaskForSynchronousOperation(() => this.WriteStartChangesetImplementation(changesetId))
-                .FollowOnSuccessWith(t => this.FinishWriteStartChangeset());
+            await this.WriteStartChangesetImplementationAsync(changesetId)
+                .FollowOnSuccessWith(t => this.FinishWriteStartChangeset())
```
habbes, line 241: Out of curiosity, what's the difference between `FollowOnSuccessWithTask` and `FollowOnSuccessWith`

habbes: Are there async versions of `this.FinishWriteStartChangeSet()`?

src/Microsoft.OData.Core/Batch/ODataBatchUtils.cs
```diff
@@ -110,12 +110,13 @@ internal static Uri CreateOperationRequestUri(Uri uri, Uri baseUri, IODataPayloa
...
             Stream outputStream,
-            IODataStreamListener operationListener)
+            IODataStreamListener operationListener,
+            bool synchronous = true)
         {
             Debug.Assert(outputStream != null, "outputStream != null");
             Debug.Assert(operationListener != null, "operationListener != null");
 
-            return new ODataWriteStream(outputStream, operationListener);
+            return new ODataWriteStream(outputStream, operationListener, synchronous);
```
habbes, line 119: Could you remind me what the `synchronous` argument does again? I remember it from a previous PR but don't remember what it was used for. I think it was used in the dispose methods, but I'm not quite sure.

src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
habbes: Or just using `await` on the second statement, i.e.

```c#
await this.WriteEndBatchImplementationAsync();
await this.FlushAsync();
```

What else is `FollowOnSuccessWithTask` doing behind the scenes?

src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
```diff
@@ -383,12 +385,14 @@ public void Flush()
...
-        public Task FlushAsync()
+        public async Task FlushAsync()
         {
             this.VerifyCanFlush(false);
 
             // Make sure we switch to writer state Error if an exception is thrown during flushing.
-            return this.FlushAsynchronously().FollowOnFaultWith(t => this.SetState(BatchWriterState.Error));
+            await this.FlushAsynchronously()
+                .FollowOnFaultWith(t => this.SetState(BatchWriterState.Error))
+                .ConfigureAwait(false);
```
habbes, line 395: What does `FollowOnFaulWith` do? Could this be safely transformed to:

```c#
try {
    await this.FlushAsynchronously();
} catch {
    this.SetState(BatchWriterState.Error);
}
```

src/Microsoft.OData.Core/Batch/ODataBatchUtils.cs
gathogojr: Reference to https://github.com/OData/odata.net/pull/2109, `ODataWriteStream` previously had no way of telling whether we're writing synchronously or asynchronously. We address that gap by introducing this argument such that when the `Dispose` method is invoked, either of `StreamDisposed` or `StreamDisposedAsync` is called depending on the value passed for the argument.
This approach is necessitated by the fact that we support `NET45` and `NETSTANDARD1_1` both of which do not support `IAsyncDisposable` that would make it feasible to use `DisposeAsync` in asynchronous scenarios.

src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
gathogojr: You can think of `FollowOnSuccessWithTask` as a "syntactic" sugar for `ContinueWith`, although it's a little more than that. We actually use `ContinueWith` in the method. The method is designed to propagate failures if the antecedent task fails. If `WriteEndBatchImplementationAsync` throws an `ODataException`, we propagate that to the caller - not an `AggregateException` like `ContinueWith` would do.
The delegate passed to `FollowOnSuccessWithTask` will not be executed if the antecedent task fails. Specifically in this case, we do not wish to execute `FlushAsync` if `WriteEndBatchImplementationAsync` fails.
The method is defined in [`TaskUtils`](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/TaskUtils.cs) class just in case you'd like to study the implementation further.

src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
gathogojr: `FollowOnSuccessWithTask` supports delegate that return another Task instance while `FollowOnSuccessWith` doesn't. As an illustration consider the following two usages:
```csharp
Task.Factory.StartNew(() =>
{
    // Do work
    return new Random().Next(1, 100);
}).FollowOnSuccessWith((antecedentTask) =>
{
    Console.WriteLine("Result: {0}", antecedentTask.Result);
});

// vs.

Task.Factory.StartNew(() =>
{
    // Do work
    // return new Random().Next(1, 100);
}).FollowOnSuccessWithTask((antecedentTask) =>
{
    return Task.Factory.StartNew(() => Console.WriteLine("Result: {0}", antecedentTask.Result));
});
```

gathogojr: There's no async version of `FinishWriteStartChangeSet`. The private method currently doesn't do any writing. It's an implementation detail so it'd be easy to introduce an async equivalent if the synchronous version evolved to doing any writing.

src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
gathogojr: `FollowOnFaultWith` is designed such that the `Action` delegate is invoked only when the antecedent task fails (i.e. task is in a Faulted state)

gathogojr: It could be transformed they way you demonstrated. But instead of littering the code with such try...catch I think it's neater to encapsulate that in a helper method

habbes (approved):

PR #2114, opened 2021-06-28
Implement asynchronous support in ODataJsonLightOutputContext

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataJsonLightOutputContext`**.

##### About `CreateJsonWriter` and `CreateAsynchronousJsonWriter` methods
**What:**
Asynchronous support is not implemented in `Microsoft.Spatial` library. To write spatial data, we rely on the synchronous `PrimitiveConverter.Instance.WriteJsonLight(object, IJsonWriter)` method. When writing asynchronously we wrap this method in a task. The `WriteJsonLight` method takes an `IJsonWriter` parameter while the asynchronous writer is declared as `IJsonWriterAsync`. Luckily, `JsonWriter` sealed internal class implements both `IJsonWriter` and `IJsonWriterAsync`. To guarantee that when the synchronous writer is called to write spatial data it shares the same scope(s) as the asynchronous writer, we initialize them to the same concrete `JsonWriter` instance. We only do this when the dependency injection container is uninitialized or when the JSON writer factory is `DefaultJsonWriterFactory` - since both `CreateJsonWriter` and `CreateAsynchronousJsonWriter` methods of that factory return an instance of `JsonWriter`.

**Why:**
Consider a scenario where we need to write a Geography collection property into the payload. The array start and end (`[` and `]`) will get written using the asynchronous writer. But the elements of the array - which are of Geography types (e.g. `GeographyPoint`) - are written using the synchronous writer. If the two writers don't share scope, the synchronous writer won't know about the array scope and therefore won't write the array element separator.

Merging `IJsonWriter` and `IJsonWriterAsync` interface in the next major release will simplify this.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
@@ -66,6 +66,14 @@ namespace Microsoft.OData.JsonLight
         private PropertyCacheHandler propertyCacheHandler;
 
         /// <summary>
+        /// The concrete <see cref="JsonWriter"/> instance initialized when
+        /// creating either synchronous or asynchronous writer.
+        /// Used to guarantee that the synchronous and asynchronous writers share the same scope(s) specifically
+        /// because the synchronous writer is used to write spatial data in both synchronous and asynchronous scenarios.
+        /// </summary>
+        private JsonWriter concreteJsonWriter;
+
+        /// <summary>
         /// Constructor.
         /// </summary>
         /// <param name="messageInfo">The context information for the message.</param>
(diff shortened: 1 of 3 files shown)


ElizabethOkerio (approved):
LGTM

marabooy (approved):

KenitoInc (approved):

habbes (approved):

gathogojr:
Added [test: WriteSpatialCollectionPropertyForInjectedJsonWriterFactoryAsync](https://github.com/OData/odata.net/pull/2114/files#diff-2c0641629571db40ce211df43a984b262bcfda3fcd78ab161ba8868c11af5fe6R549) for untested scenario and squashed changes

PR #2121, opened 2021-07-06
Implement asynchronous support in RawValueWriter

gathogojr:
### Issues

### Description

Implement asynchronous support in **`RawValueWriter`**.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/RawValueWriter.cs
+++ b/src/Microsoft.OData.Core/RawValueWriter.cs
@@ -8,9 +8,9 @@ namespace Microsoft.OData
 {
     using System;
     using System.Diagnostics;
-    using System.Diagnostics.CodeAnalysis;
     using System.IO;
     using System.Text;
+    using System.Threading.Tasks;
     using Microsoft.OData.Json;
     using Microsoft.Spatial;
 
@@ -119,9 +119,7 @@ namespace Microsoft.OData
         {
             Debug.Assert(!(value is byte[]), "!(value is byte[])");
 
-            string valueAsString;
-            ODataEnumValue enumValue = value as ODataEnumValue;
-            if (enumValue != null)
+            if (value is ODataEnumValue enumValue)
             {
                 this.textWriter.Write(enumValue.Value);
             }
@@ -129,7 +127,7 @@ namespace Microsoft.OData
             {
                 PrimitiveConverter.Instance.WriteJsonLight(value, jsonWriter);
             }
-            else if (ODataRawValueUtils.TryConvertPrimitiveToString(value, out valueAsString))
+            else if (ODataRawValueUtils.TryConvertPrimitiveToString(value, out string valueAsString))
             {
                 this.textWriter.Write(valueAsString);
             }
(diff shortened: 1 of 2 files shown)


marabooy (approved):

src/Microsoft.OData.Core/RawValueWriter.cs
```diff
@@ -154,6 +154,87 @@ internal void Flush()
...
+
+        /// <summary>
+        /// Asynchronously end the writing of a raw output. This should be the last thing called.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        internal Task EndAsync()
+        {
+            if (this.settings.HasJsonPaddingFunction())
+            {
+                return this.textWriter.WriteAsync(JsonConstants.EndPaddingFunctionScope);
```
g2mula, line 180: General question for all the Task methods in the PR; 
Do we have a general rule on where to elide and where not to elide async awaits with regards to Tasks?

src/Microsoft.OData.Core/RawValueWriter.cs
```diff
@@ -154,6 +154,87 @@ internal void Flush()
...
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        /// <remarks>We do not accept binary values here; WriteBinaryValue should be used for binary data.</remarks>
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        internal Task WriteRawValueAsync(object value)
+        {
+            Debug.Assert(!(value is byte[]), "!(value is byte[])");
+
+            string valueAsString;
+            ODataEnumValue enumValue = value as ODataEnumValue;
+            if (enumValue != null)
```
g2mula, line 200:
```suggestion
            if (value is ODataEnumValue enumValue)

```

src/Microsoft.OData.Core/RawValueWriter.cs
```diff
@@ -154,6 +153,86 @@ internal void Flush()
...
+            else if (ODataRawValueUtils.TryConvertPrimitiveToString(value, out valueAsString))
+            {
+                return this.textWriter.WriteAsync(valueAsString);
+            }
+            else
+            {
+                // Value is neither enum nor primitive
+                return TaskUtils.GetFaultedTask(
+                    new ODataException(Strings.ODataUtils_CannotConvertValueToRawString(value.GetType().FullName)));
+            }
```
g2mula, line 216:
```suggestion
            if (value is ODataEnumValue enumValue)
            {
                return this.textWriter.WriteAsync(enumValue.Value);
            }

            if (value is Geometry || value is Geography)
            {
                return TaskUtils.GetTaskForSynchronousOperation(
                    () => PrimitiveConverter.Instance.WriteJsonLight(value, jsonWriter));
            }

            if (ODataRawValueUtils.TryConvertPrimitiveToString(value, out var valueAsString))
            {
                return this.textWriter.WriteAsync(valueAsString);
            }

            // Value is neither enum nor primitive
            return TaskUtils.GetFaultedTask(
                new ODataException(Strings.ODataUtils_CannotConvertValueToRawString(value.GetType().FullName)));
```

src/Microsoft.OData.Core/RawValueWriter.cs
gathogojr: When a method is a pass-through. The reasoning here is that `HasJsonPaddingFunction` would never be expected to throw an exception. Here is how it is defined:
```csharp
        internal bool HasJsonPaddingFunction()
        {
            return !string.IsNullOrEmpty(this.JsonPCallback);
        }
```
By that reasoning, `EndAsync` method can be treated as a pass-through since we have the option of either returning the task that does the actual writing of the Json padding function closing parens (`)`) or a completed task - `TaskUtils.CompletedTask` is defined as a `static`. Either of the option saves us from creating a `Task` object hence this is an optimization.
It's unwise to elide where a function has multiple statements where any single one of them can throw. In this scenario, the function needs to the `async` modifier to cause the generation of the state machine class such that in the event of an exception being thrown, it'll be stashed in the task using the state machine's `SetException` method.

KenitoInc (approved):
LGTM

src/Microsoft.OData.Core/RawValueWriter.cs
```diff
@@ -154,6 +153,86 @@ internal void Flush()
...
+        /// </summary>
+        /// <returns>A task that represents the asynchronous write operation.</returns>
+        internal Task EndAsync()
+        {
+            if (this.settings.HasJsonPaddingFunction())
+            {
+                return this.textWriter.WriteAsync(JsonConstants.EndPaddingFunctionScope);
+            }
+
+            return TaskUtils.CompletedTask;
```
habbes, line 182: Is this better than using `Task.CompletedTask` from the standard lib?

src/Microsoft.OData.Core/RawValueWriter.cs
```diff
@@ -154,6 +153,86 @@ internal void Flush()
...
+
+            string valueAsString;
+            if (value is ODataEnumValue enumValue)
+            {
+                return this.textWriter.WriteAsync(enumValue.Value);
+            }
+            else if (value is Geometry || value is Geography)
+            {
+                return TaskUtils.GetTaskForSynchronousOperation(
+                    () => PrimitiveConverter.Instance.WriteJsonLight(value, jsonWriter));
```
habbes, line 205: Out of curiosity, does this `WriteJsonLight` method write to a stream (the `jsonWriter`)? If so, why doesn't it have an async equivalent?

src/Microsoft.OData.Core/RawValueWriter.cs
```diff
@@ -154,6 +153,86 @@ internal void Flush()
...
+            }
+            else if (ODataRawValueUtils.TryConvertPrimitiveToString(value, out valueAsString))
+            {
+                return this.textWriter.WriteAsync(valueAsString);
+            }
+            else
+            {
+                // Value is neither enum nor primitive
+                return TaskUtils.GetFaultedTask(
+                    new ODataException(Strings.ODataUtils_CannotConvertValueToRawString(value.GetType().FullName)));
```
habbes, line 215: Is this better than using `Task.FromException`?

habbes (commented):
Should this class also implement `IAsyncDisposable`?

src/Microsoft.OData.Core/RawValueWriter.cs
gathogojr: Task.Completed is not supported across all frameworks that we support.

src/Microsoft.OData.Core/RawValueWriter.cs
gathogojr: `Task.FromException` is not support across all the frameworks that ODL supports

src/Microsoft.OData.Core/RawValueWriter.cs
gathogojr: It was a deliberate decision to retain asynchronous wrappers over synchronous methods every place where `PrimitiveConverter.Instance.WriteJsonLight(object, IJsonWriter)` method has been used. If you explore deeper, you'll realize that this method is used for writing spatial data. Since extending asynchronous support to Microsoft.Spatial library at this time would have resulted into an explosion of scope, wrapping the synchronous call was the alternative option. You'll find that there's a test to verify that it works as expected.

gathogojr:
> Should this class also implement `IAsyncDisposable`?

Not necessary. `Dispose` method is still called even in asynchronous scenarios so the introducing `DisposeAsync` would be redundant since the implementation would be a mirror of `Dispose`, i.e.
```csharp
this.textWriter.Dispose();
this.textWriter = null;
```
In this case `Dispose` does all we need to be done.

habbes (approved):

g2mula (approved):
lgtm

PR #2124, opened 2021-07-08
Implement asynchronous support in ODataAsynchronousWriter

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataAsynchronousWriter`**.

#### About the constructors...
Firstly, they are all private.
Secondly, I introduced 2 new constructors.
The first new constructor doesn't take a delegate parameter to write the envelop for the inner message. I made the change such that its called from `CreateMessageForReading` method that was previously passing a `null` for `writeEnvelope` `Action` delegate parameter.
The other new constructor takes a `Func` delegate of `ODataAsynchronousResponseMessage` and `Task`. It is applied in asynchronous scenarios such that its possible to `await`. The constructor is called from `CreateMessageForWritingAsync` method.
Finally I cleaned everything up such that the 2 constructors that accept 6 parameters call the constructor that accepts 5 parameters and only initialize the additional parameter in their body.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataAsynchronousResponseMessage.cs
+++ b/src/Microsoft.OData.Core/ODataAsynchronousResponseMessage.cs
@@ -31,6 +31,9 @@ namespace Microsoft.OData
         /// <summary>The action to write envelope for the inner message before returning the stream.</summary>
         private readonly Action<ODataAsynchronousResponseMessage> writeEnvelope;
 
+        /// <summary>The action to write envelope for the inner message asynchronously before returning the stream.</summary>
+        private readonly Func<ODataAsynchronousResponseMessage, Task> writeEnvelopeAsync;
+
         /// <summary>The dependency injection container to get related services.</summary>
         private readonly IServiceProvider container;
 
@@ -49,6 +52,30 @@ namespace Microsoft.OData
         /// <param name="stream">The stream underlying the response message.</param>
         /// <param name="statusCode">The status code to use for the async response message.</param>
         /// <param name="headers">The headers to use for the async response message.</param>
+        /// <param name="writing">true if the response message is being written; false when it is read.</param>
+        /// <param name="container">The dependency injection container to get related services.</param>
+        private ODataAsynchronousResponseMessage(
+            Stream stream,
+            int statusCode,
+            IDictionary<string, string> headers,
+            bool writing,
+            IServiceProvider container)
+        {
+            Debug.Assert(stream != null, "stream != null");
+
+            this.stream = stream;
+            this.statusCode = statusCode;
+            this.headers = headers;
+            this.writing = writing;
+            this.container = container;
+        }
+
+        /// <summary>
+        /// Constructor.
+        /// </summary>
+        /// <param name="stream">The stream underlying the response message.</param>
+        /// <param name="statusCode">The status code to use for the async response message.</param>
+        /// <param name="headers">The headers to use for the async response message.</param>
         /// <param name="writeEnvelope">The action to write envelope for the inner message before returning the stream.</param>
         /// <param name="writing">true if the response message is being written; false when it is read.</param>
         /// <param name="container">The dependency injection container to get related services.</param>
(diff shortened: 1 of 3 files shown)


src/Microsoft.OData.Core/ODataAsynchronousWriter.cs
```diff
@@ -102,9 +104,13 @@ void IODataOutputInStreamErrorListener.OnInStreamError()
         }
 
         /// <inheritdoc/>
-        Task IODataOutputInStreamErrorListener.OnInStreamErrorAsync()
+        async Task IODataOutputInStreamErrorListener.OnInStreamErrorAsync()
         {
-            throw new NotImplementedException();
+            this.rawOutputContext.VerifyNotDisposed();
+            await this.rawOutputContext.TextWriter.FlushAsync()
```
marabooy, line 110: Do we need to flush if we are still going to throw the exception below?

marabooy: Is the rawOutputContext reused?

marabooy (approved):

src/Microsoft.OData.Core/ODataAsynchronousWriter.cs
gathogojr: @marabooy In particular scenarios, we return the partial payload written before things went haywire. Flushing is therefore necessary. We're not flushing so we can reuse the `RawOutputContext` instance.

src/Microsoft.OData.Core/ODataAsynchronousResponseMessage.cs
```diff
@@ -164,9 +206,21 @@ public Stream GetStream()
...
+            // If writing response, the envelope for the inner message should be written once and only once before returning the stream.
+            if (this.writing && !this.envelopeWritten)
+            {
+                if (this.writeEnvelopeAsync != null)
+                {
+                    await this.writeEnvelopeAsync(this)
+                        .ConfigureAwait(false);
+                }
+
+                this.envelopeWritten = true;
```
habbes, line 220: Out of curiosity, what if `writeEnvelopeAsync` is null but `writeEnvelope` is not null. With this change, the envelope won't be return. But in the current version, it would be written since it calls `GetStream`. With that in mind, isn't it it a breaking change to not fall back to the sync implementation if `writeEnvelope` exists?

gathogojr: `ODataAsynchronousResponseMessage` public* API comprises of:
- `StatusCode` property
- `Headers` property (readonly)
- `Container` property (readonly)
- `GetHeader(string)`
- `SetHeader(string, string)`
- `GetStream()`
- `GetStreamAsync()`
- `CreateMessageForWriting(Stream, Action<ODataAsynchronousResponseMessage>, IServiceProvider)` - internal
- `CreateMessageForWritingAsync(Stream, Func<ODataAsynchronousResponseMessage, Task>, IServiceProvider)` - internal
- `CreateMessageForReading(Stream, statusCode, IDictionary<string, string>, IServiceProvider)` - internal

`ODataAsynchronousWriter` public* API on the other hand comprises of:
- `ODataAsynchronousWriter(ODataRawOutputContext)` - internal constructor
- `CreateResponseMessage()`
- `CreateResponseMessageAsync()`
- `Flush()`
- `FlushAsync()`

**Before my change...**
The `ODataAsynchronousResponseMessage.writeEnvelope` delegate is indirectly initialized through either of the following call sequence:

`ODataAsynchronousWriter.CreateResponseMessage` [public] => `ODataAsynchronousWriter.CreateResponseMessageImplementation` [private] => `ODataAsynchronousResponseMessage.CreateMessageForWriting` [interna] => `ODataAsynchronousResponseMessage` [private constructor]

`ODataAsynchronousWriter.CreateResponseMessageAsync` [public] => `ODataAsynchronousWriter.CreateResponseMessageImplementation` [private] => `ODataAsynchronousResponseMessage.CreateMessageForWriting` [interna] => `ODataAsynchronousResponseMessage` [private constructor]

The `writeEnvelope` delegate is passed when `ODataAsynchronousResponseMessage.CreateMessageForWriting` is triggered and this internal method is only triggered from `ODataAsynchronousWriter.CreateResponseMessageImplementation` currently.

**After my change...**
The `ODataAsynchronousResponseMessage.writeEnvelope` delegate will be indirectly initialized through the following call sequence:

`ODataAsynchronousWriter.CreateResponseMessage` [public] => `ODataAsynchronousWriter.CreateResponseMessageImplementation` [private] => `ODataAsynchronousResponseMessage.CreateMessageForWriting` [interna] => `ODataAsynchronousResponseMessage` [private constructor]

`ODataAsynchronousResponseMessage.writeEnvelopeAsync` delegate will be indirectly initialized through the following call sequence:

`ODataAsynchronousWriter.CreateResponseMessageAsync` [public] => `ODataAsynchronousWriter.CreateResponseMessageImplementationAsync` [private] => `ODataAsynchronousResponseMessage.CreateMessageForWritingAsync` [interna] => `ODataAsynchronousResponseMessage` [private constructor]

Since synchronous writing involves using `ODataAsynchronousWriter.CreateResponseMessage`/`ODataAsynchronousResponseMessage.GetStream` pair, while asynchronous writing involves using `ODataAsynchronousWriter.CreateResponseMessageAsync`/`ODataAsynchronousResponseMessage.GetStreamAsync` pair, my change is not a breaking one since `writeEnvelope` will always be null when one is writing asynchronously after I merge my change.

\* _I classified `internal` methods as public in this context so it'd make the explanation easier_

KenitoInc (approved):

src/Microsoft.OData.Core/ODataAsynchronousResponseMessage.cs
gathogojr: Note also that usages of `ODataAsynchronousResponseMessage` are only in `ODataAsynchronousWriter` and `ODataAsynchronousReader` both in ODL. Furthermore, `ODataAsynchronousReader` doesn't make use of the `writeEnvelope` delegate at all.

habbes (approved):

PR #2125, opened 2021-07-08
Implement asynchronous support in ODataRawOutputContext

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataRawOutputContext`**

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataRawOutputContext.cs
+++ b/src/Microsoft.OData.Core/ODataRawOutputContext.cs
@@ -121,22 +121,20 @@ namespace Microsoft.OData
         /// </summary>
         /// <returns>Task which represents the pending flush operation.</returns>
         /// <remarks>The method should not throw directly if the flush operation itself fails, it should instead return a faulted task.</remarks>
-        internal Task FlushAsync()
+        internal async Task FlushAsync()
         {
             this.AssertAsynchronous();
 
-            return TaskUtils.GetTaskForSynchronousOperationReturningTask(
-                () =>
-                {
-                    if (this.rawValueWriter != null)
-                    {
-                        this.rawValueWriter.Flush();
-                    }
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/ODataRawOutputContext.cs
```diff
@@ -186,10 +184,11 @@ internal override Task WriteInStreamErrorAsync(ODataError error, bool includeDeb
         {
             if (this.outputInStreamErrorListener != null)
             {
-                this.outputInStreamErrorListener.OnInStreamError();
+                return this.outputInStreamErrorListener.OnInStreamErrorAsync();
```
marabooy, line 187: My guess is that we were not returning earlier due to the limited async support. right?

marabooy (approved):

KenitoInc (approved):

src/Microsoft.OData.Core/ODataRawOutputContext.cs
gathogojr: @marabooy Thank you for the question. I just realized that introducing the return statement changes the logic altogether. Funny how I made the mistake... The `outputInStreamErrorListener` variable in this class only gets initialized to an instance of `ODataAsynchronousWriter` and `OnInStreamError` or `OnInStreamErrorAsync` methods defined in that class actually throw an exception such that no statement after calling either method would ever be executed. So I introduced the return statement based on that knowledge! But that is an implementation detail that can change. The right thing is to add the `async` modifier to the method and await `this.outputInStreamErrorListener.OnInStreamErrorAsync()`. That way the exception thrown by `OnInStreamErrorAsync` or by the `throw` statement a couple of lines below will be captured and placed on the returned task.

PR #2126, opened 2021-07-08
Implement asynchronous support in ODataMultipartMixedBatchWriter

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataMultipartMixedBatchWriter`**.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
+++ b/src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
@@ -115,15 +115,17 @@ namespace Microsoft.OData.MultipartMixed
         /// A task representing any action that is running as part of the status change of the operation;
         /// null if no such action exists.
         /// </returns>
-        public override Task StreamRequestedAsync()
+        public override async Task StreamRequestedAsync()
         {
             // Write any pending data and flush the batch writer to the async buffered stream
-            this.StartBatchOperationContent();
+            await this.StartBatchOperationContentAsync()
+                .ConfigureAwait(false);
 
             // Asynchronously flush the async buffered stream to the underlying message stream (if there's any);
             // then dispose the batch writer (since we are now writing the operation content) and set the corresponding state.
-            return this.RawOutputContext.FlushBuffersAsync()
-                .FollowOnSuccessWith(task => this.DisposeBatchWriterAndSetContentStreamRequestedState());
+            await this.RawOutputContext.FlushBuffersAsync()
+                .FollowOnSuccessWith(task => this.DisposeBatchWriterAndSetContentStreamRequestedState())
+                .ConfigureAwait(false);
         }
 
         /// <summary>
@@ -158,7 +160,20 @@ namespace Microsoft.OData.MultipartMixed
             this.SetState(BatchWriterState.Error);
             this.RawOutputContext.TextWriter.Flush();
 
-            // The OData protocol spec did not defined the behavior when an exception is encountered outside of a batch operation. The batch writer
+            // The OData protocol spec does not define the behavior when an exception is encountered outside of a batch operation. The batch writer
+            // should not allow WriteError in this case. Note that WCF DS Server does serialize the error in XML format when it encounters one outside of a
+            // batch operation.
+            throw new ODataException(Strings.ODataBatchWriter_CannotWriteInStreamErrorForBatch);
+        }
+
+        public override async Task OnInStreamErrorAsync()
+        {
+            this.RawOutputContext.VerifyNotDisposed();
+            this.SetState(BatchWriterState.Error);
+            await this.RawOutputContext.TextWriter.FlushAsync()
+                .ConfigureAwait(false);
+
+            // The OData protocol spec does not define the behavior when an exception is encountered outside of a batch operation. The batch writer
             // should not allow WriteError in this case. Note that WCF DS Server does serialize the error in XML format when it encounters one outside of a
             // batch operation.
             throw new ODataException(Strings.ODataBatchWriter_CannotWriteInStreamErrorForBatch);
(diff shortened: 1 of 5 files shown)


src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriterUtils.cs
```diff
@@ -260,5 +388,86 @@ private static void WriteRequestUri(TextWriter writer, string httpMethod, Uri ur
...
             }
         }
+
+        /// <summary>
+        /// Asynchronously writes the headers.
+        /// </summary>
+        /// <param name="writer">Writer to write headers.</param>
+        /// <param name="inChangeSetBound">Whether we are in ChangeSetBound.</param>
+        /// <param name="contentId">The Content-ID value to write in ChangeSet head.</param>
+		/// <returns>A task that represents the asynchronous write operation.</returns>
```
KenitoInc, line 398: Fix indent

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriterUtils.cs
```diff
@@ -260,5 +388,86 @@ private static void WriteRequestUri(TextWriter writer, string httpMethod, Uri ur
...
+
+        /// <summary>
+        /// Asynchronously writes the request line.
+        /// </summary>
+        /// <param name="writer">Writer to write request uri.</param>
+        /// <param name="httpMethod">The Http method to be used for this request operation.</param>
+        /// <param name="uri">The Uri to be used for this request operation.</param>
+        /// <param name="baseUri">The service root Uri to be used for this request operation.</param>
+        /// <param name="payloadUriOption">The format of operation Request-URI, which could be AbsoluteUri, AbsoluteResourcePathAndHost, or RelativeResourcePath.</param>
+		/// <returns>A task that represents the asynchronous write operation.</returns>
```
KenitoInc, line 421: Fix indent

src/Microsoft.OData.Core/ODataWriteStream.cs
```diff
@@ -138,6 +138,16 @@ public override void Flush()
             this.stream.Flush();
         }
 
+        ///// <summary>
```
KenitoInc, line 141: Why did we comment out this?

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMultipartMixedBatchWriterTests.cs
```diff
@@ -0,0 +1,797 @@
...
+        public async Task WriteMultipartMixedBatchAsync()
+        {
+            var result = await SetupMultipartMixedBatchWriterAndRunTestAsync(
+                async (multipartMixedBatchWriter) =>
+                {
+                    await multipartMixedBatchWriter.WriteStartBatchAsync();
+                    await multipartMixedBatchWriter.WriteEndBatchAsync();
+                });
+
+            Assert.Equal("--batch_aed653ab--\r\n", result);
```
KenitoInc, line 56: WriteStartBatch should write `--batch_aed653ab`
WriteEndBatch should write `--batch_aed653ab--`
So what should be the result?

KenitoInc (changes requested):

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMultipartMixedBatchWriterTests.cs
gathogojr: @KenitoInc In the existing implementation, writing of the batch start boundary is triggered by either of the following 3 events
- Creation of a operation request message
- Creation of a operation response message
- Start of a changeset

This test is more for an experimental rather than practical scenario and I'd also agree with you that the logical outcome should be `--batch_aed653ab\r\n--batch_aed653ab--\r\n`.
In addition the fact that calling only `WriteStartBatch` and `WriteEndBatch` serves no practical purposes in batch writing scenario, attempting to rectifying the logic such that the start boundary is written when only the two methods are called on multipart-mixed batch writer would be considered a breaking change since `--batch_aed653ab--\r\n` is the outcome for the already existing synchronous workflow.
I have however modified the test to trigger `CreateOperationRequestMessageAsync` to reflect a more logical outcome

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
```diff
@@ -462,5 +658,95 @@ private void WritePendingMessageData(bool reportMessageCompleted)
...
+                }
+
+                // Write CRLF after the headers (or request/response line if there are no headers)
+                await this.RawOutputContext.TextWriter.WriteLineAsync()
+                    .ConfigureAwait(false);
+
+                if (reportMessageCompleted)
+                {
+                    this.CurrentOperationMessage.PartHeaderProcessingCompleted();
+                    this.CurrentOperationRequestMessage = null;
```
marabooy, line 746: Do we need to check if current operationrequest message is null? Is this method always called once?

marabooy (approved):

KenitoInc (approved):
LGTM

src/Microsoft.OData.Core/MultipartMixed/ODataMultipartMixedBatchWriter.cs
gathogojr: This logic is inside an `if (this.CurrentOperationMessage != null)` block. Here's a stripped down version of  `CurrentOperationMessage` property definition:
```csharp
private ODataBatchOperationMessage CurrentOperationMessage
{
    get
    {
        if (this.CurrentOperationRequestMessage != null)
        {
            return this.CurrentOperationRequestMessage.OperationMessage;
        }
        else if (this.CurrentOperationResponseMessage != null)
        {
            return this.CurrentOperationResponseMessage.OperationMessage;
        }
        else
        {
            return null;
        }
    }
}
```
The call to `this.CurrentOperationMessage.PartHeaderProcessingCompleted()` should be safe since it'll never be triggered unless either `CurrentOperationRequestMessage` or `CurrentOperationResponseMessage` is non-null.
Assigning null to both `CurrentOperationRequestMessage` and `CurrentOperationResponseMessage` is also safe

marabooy: Thanks @gathogojr

PR #2127, opened 2021-07-08
Fixing Linq Performance Issue for Navigation binding cache

Sreejithpin:
### Issues
Fixing Linq Performance Issue for Navigation binding cache

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Aggregation/ApplyBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Aggregation/ApplyBinder.cs
@@ -125,7 +125,12 @@ namespace Microsoft.OData.UriParser.Aggregation
                 }
             }
 
-            return mergedTokens.Concat(entitySetTokens.Values).ToList();
+            foreach(AggregateTokenBase tokenBase in entitySetTokens.Values)
+            {
+                mergedTokens.Add(tokenBase);
+            }
+
+            return mergedTokens;
         }
 
         private AggregateExpressionBase BindAggregateExpressionToken(AggregateTokenBase aggregateToken)
--- a/src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
@@ -155,7 +155,12 @@ namespace Microsoft.OData.UriParser
                     NonSystemToken nonSystemToken = pathToken as NonSystemToken;
                     if (nonSystemToken != null && nonSystemToken.NamedValues != null)
                     {
-                        parameterNames = nonSystemToken.NamedValues.Select(s => s.Name).ToList();
+                        parameterNames = new List<string>();
+
+                        foreach(NamedValue namedValue in nonSystemToken.NamedValues)
+                        {
+                            parameterNames.Add(namedValue.Name);
+                        }
                     }
 
                     if (parameterNames.Count > 0)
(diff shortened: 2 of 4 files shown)


src/Microsoft.OData.Core/UriParser/Aggregation/ApplyBinder.cs
```diff
@@ -125,7 +125,12 @@ private static IEnumerable<AggregateTokenBase> MergeEntitySetAggregates(IEnumera
                 }
             }
 
-            return mergedTokens.Concat(entitySetTokens.Values).ToList();
+            foreach(AggregateTokenBase tokenBase in entitySetTokens.Values)
+            {
+                mergedTokens.Add(tokenBase);
+            }
```
marabooy, line 131:
```suggestion   
                mergedTokens.AddRange(entitySetTokens.Values);
```

src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
```diff
@@ -155,7 +155,12 @@ internal static bool TryBindAsOperation(PathSegmentToken pathToken, IEdmModel mo
                     NonSystemToken nonSystemToken = pathToken as NonSystemToken;
                     if (nonSystemToken != null && nonSystemToken.NamedValues != null)
                     {
-                        parameterNames = nonSystemToken.NamedValues.Select(s => s.Name).ToList();
+                        parameterNames = new List<string>();
+
+                        foreach(NamedValue namedValue in nonSystemToken.NamedValues)
```
marabooy, line 160: The linq query on the right could be faster in most cases as the list may need to resize during addition.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsNavigationSource.cs
```diff
@@ -130,11 +130,20 @@ public IEnumerable<IEdmNavigationPropertyBinding> FindNavigationPropertyBindings
...
                 if (!navigationPropertyBindingCache.TryGetValue(navigationProperty, out navigationBindingList))
                 {
-                    navigationBindingList = this.NavigationPropertyBindings.Where(targetMapping => targetMapping.NavigationProperty == navigationProperty);
+                    navigationBindingList = new List<IEdmNavigationPropertyBinding>();
+
+                    foreach (IEdmNavigationPropertyBinding targetMapping in this.NavigationPropertyBindings)
+                    {
+                        if (targetMapping.NavigationProperty == navigationProperty)
+                        {
+                            navigationBindingList.Add(targetMapping);
```
marabooy, line 143: Add range should be faster in most cases as it uses a faster array copy method for the items.

src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
Sreejithpin: But it does 2 iteration right. 1 for select and another for T0List

KenitoInc:
```suggestion
                        foreach (NamedValue namedValue in nonSystemToken.NamedValues)

```

KenitoInc (changes requested):

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsNavigationSource.cs
Sreejithpin: You mean add range with a where to check for navigationproperty? that would be slower right?

src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
habbes: @Sreejithpin  does it really do 2 iterations? I think `.Select()` is lazily evaluated when `ToList()` is called, and probably leads to just one iteration? I'd be curious to see the benchmarks for this particular loop compared to the code being replaced.

habbes:
@Sreejithpin maybe you could share a screenshot of the benchmark results for these changes?

Sreejithpin:
Closing as it was already done in https://github.com/OData/odata.net/pull/2123

PR #2130, opened 2021-07-14
Fix NullReferenceException thrown during materialization of lower-camel-cased payload

### LINKED ISSUE #2116: How to avoid exposing properties in an entity to client?
Say we have an entity which has property ETag. When I add this entity to EdmModelBuilder, I want to avoid adding  ETag property in the schema so that we can hide this property from client. 
We need to define this in the entity because we want to read the ETag from storage layer.
IS there a way to hide properties of an entity from client?

```
Public class PhoneEntity
{
        /// <summary>
        /// Gets or sets the ETag
        /// </summary>
        [ConcurrencyCheck]
        [DataMember]
        string ETag { get; set; }
 
        // Other properties 
}
```

### LINKED ISSUE #2119: Error when materializing an open entity type when a dynamic complex property is returned with a value of null
If an OData service returns an open entity type when a dynamic **complex** property is returned with a value of null, attempt to materialize that value into the dynamic properties container property results into an error. This issue is related to https://github.com/OData/odata.net/issues/1943

### Assemblies affected

Microsoft.OData.Client 7.9.0

### Reproduce steps

1. Create an OData service that returns an open entity type with dynamic **complex** property with a null value
2. Create an OData client and generate the proxy using OData Connected Service extension
3. Write code to query the entity

### Expected result

The entity is materialized successfully on the client

### Actual result

**`NullReferenceException`**

### Additional detail

OData WebApi currently does not write a dynamic property into the response if the value is null.

gathogojr:
### Issues

*This pull request fixes #2119.*

### Description

When an OData service metadata and payload is camel-cased and an entity bound to base entity type collection is returned, the value for `@odata.type` property is not in the expected form leading to a `NullReferenceException` during materialization.

Consider the following service metadata for a contrived sample service comprising of two open entity types `Rectangle` and `Shape` where `Rectangle` derives from `Shape`.
```xml
<?xml version="1.0" encoding="utf-8"?>
  <edmx:DataServices>
    <Schema xmlns="http://docs.oasis-open.org/odata/ns/edm" Namespace="NS.Models">
      <EntityType Name="shape" OpenType="true">
        <Key>
          <PropertyRef Name="id" />
        </Key>
        <Property Name="id" Type="Edm.Int32" Nullable="false" />
        <Property Name="area" Type="Edm.Double" Nullable="false" />
      </EntityType>
      <EntityType Name="rectangle" BaseType="NS.Models.shape" OpenType="true">
		<Property Name="length" Type="Edm.Double" Nullable="false" />
        <Property Name="width" Type="Edm.Double" Nullable="false" />
        <Property Name="attributes" Type="Collection(Edm.String)" Nullable="false" />
      </EntityType>
      <EntityContainer Name="Container">
        <EntitySet Name="shapes" EntityType="NS.Models.shape" />
        <EntitySet Name="rectangles" EntityType="NS.Models.rectangle" />
      </EntityContainer>
    </Schema>
  </edmx:DataServices>
</edmx:Edmx>
```
In the `EdmModelBase` class, we maintain a dictionary mapping the fully qualified name of each structured type to the corresponding schema type. Given the above service metadata, we'll store the `Rectangle` entity type under the key "NS.Models.rectangle".
However, if one queries for the `Rectangle` entity against the base entity type collection "shapes", the following payload is returned:
```json
{
  "@odata.context": "http://tempuri.org/$metadata#shapes/$entity",
  "@odata.type": "#NS.Models.Rectangle",
  "id": 1,
  "area": 28,
  "length": 7,
  "width": 4,
  "attributes": []
}
```
In the above payload, the value for `@odata.type` property is written as "#NS.Models.Rectangle" rather than "#NS.Models.rectangle".
For this reason, the search for the entity type in the schema types dictionary by key **is unsuccessful**. This throws off the materialization logic and a `NullReferenceException` is thrown when trying to materialize the primitive collection property.

**Note:**
The alternative option considered was to apply a fix in ODL such that the lower-camel-case option was applied when writing `@odata.type` property but that carries a bigger risk since payloads for all workloads on AGS would suddenly change.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/ClientEdmModel.cs
+++ b/src/Microsoft.OData.Client/ClientEdmModel.cs
@@ -328,7 +328,7 @@ namespace Microsoft.OData.Client
                 {
                     hierarchy.Insert(0, type);
                 }
-                while ((type = c.PlatformHelper.GetBaseType(type)) != null);
+                while ((type = c.PlatformHelper.GetBaseType(type)) != null && type != typeof(object));
             }
 
             return hierarchy.ToArray();
@@ -413,7 +413,7 @@ namespace Microsoft.OData.Client
             }
 
             if (cachedEdmType == null)
-            {
+            { 
                 Type collectionType;
                 bool isOpen = false;
                 IEdmSchemaElement edmSchemaElement = null;
--- a/src/Microsoft.OData.Client/Common.cs
+++ b/src/Microsoft.OData.Client/Common.cs
@@ -187,6 +187,18 @@ namespace Microsoft.OData.Service
         internal static string GetModelTypeName(Type type)
         {
             Debug.Assert(type != null, "type != null");
+
+            string typeName;
+            OriginalNameAttribute originalNameAttribute = (OriginalNameAttribute)type.GetCustomAttributes(typeof(OriginalNameAttribute), false).SingleOrDefault();
+            if (originalNameAttribute != null)
+            {
+                typeName = originalNameAttribute.OriginalName;
+            }
+            else
+            {
+                typeName = type.Name;
+            }
+
 #if ODATA_CLIENT
             if (type.IsGenericType())
 #else
@@ -194,7 +206,7 @@ namespace Microsoft.OData.Service
 #endif
             {
                 Type[] genericArguments = type.GetGenericArguments();
-                StringBuilder builder = new StringBuilder(type.Name.Length * 2 * (1 + genericArguments.Length));
+                StringBuilder builder = new StringBuilder(typeName.Length * 2 * (1 + genericArguments.Length));
                 if (type.IsNested)
                 {
                     Debug.Assert(type.DeclaringType != null, "type.DeclaringType != null");
@@ -202,7 +214,7 @@ namespace Microsoft.OData.Service
                     builder.Append('_');
                 }
 
-                builder.Append(type.Name);
+                builder.Append(typeName);
                 builder.Append('[');
                 for (int i = 0; i < genericArguments.Length; i++)
                 {
@@ -234,11 +246,11 @@ namespace Microsoft.OData.Service
             else if (type.IsNested)
             {
                 Debug.Assert(type.DeclaringType != null, "type.DeclaringType != null");
-                return GetModelTypeName(type.DeclaringType) + "_" + type.Name;
+                return GetModelTypeName(type.DeclaringType) + "_" + typeName;
             }
             else
             {
-                return type.Name;
+                return typeName;
             }
         }
 
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -12,6 +12,7 @@ namespace Microsoft.OData.Client
     #region Namespaces
 
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Collections.ObjectModel;
     using System.Diagnostics;
@@ -169,6 +170,8 @@ namespace Microsoft.OData.Client
 
         /// <summary>A factory class to use in selecting the the request message transport mode implementation </summary>
         private IDataServiceRequestMessageFactory requestMessageFactory = new DataServiceRequestMessageFactory();
+
+        private ConcurrentDictionary<string, Type> resolveTypesCache = new ConcurrentDictionary<string, Type>();
 
         #region Test hooks for header and payload verification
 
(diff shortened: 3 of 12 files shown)


gathogojr:
Marked pull request to draft as there are test that already validate that type names are case-sensitive - pending decision on way forward.

mikepizzo (changes requested):
I'm trying to understand if this should be fixed in ODL or in ODataClient.

If we wanted to implement this logic in ODL, I would provide an overload of FindType (or FindDeclaredType) that took a boolean caseInsensitive flag and implemented logic similar to the ResolveType method in ODataUriResolver, and then call that from ODataClient.

If we think this is limited to ODataClient, I would put logic in ODataClient, similar to ODataUriResolver, to look up case-insensitive if the case-sensitive lookup for the type fails.

As a side note, should ODataClient be calling FindDeclaredType (which only searches the current model), or should it be calling the extension method for FindType, which searches across referenced models?

gathogojr:
@mikepizzo The observed bug can be mitigated by setting the `ResolveName` delegate property on the `DataServiceContext` instance from the OData client application. Together with its functional opposite `ResolveType`, these 2 properties enable one to override the default type resolution strategy used by the client library when materializing a payload from an OData service into an object instance, or when serializing an object instance into a payload to send to an OData service.
The user need not set those delegate properties manually when using OData Connected Service extension to generate the proxy classes. The code auto-generated by the extension takes care of that. The code will look something like this:
```csharp
public partial class Container : DataServiceContext
{
    public Container(Uri serviceRoot) : base(serviceRoot, ODataProtocolVersion.V4)
    {
        this.ResolveName = new Func<Type, string>(this.ResolveNameFromType);
        this.ResolveType = new Func<string, Type>(this.ResolveTypeFromName);
        // ...
    }
	
    // ...
	
    protected string ResolveNameFromType(Type clientType)
    {
        OriginalNameAttribute originalNameAttribute = (OriginalNameAttribute)Enumerable.SingleOrDefault(
            Utility.GetCustomAttributes(clientType, typeof(OriginalNameAttribute), true));
        if (clientType.Namespace.Equals("NS", StringComparison.Ordinal))
        {
            if (originalNameAttribute != null)
            {
                return string.Concat("NS.", originalNameAttribute.OriginalName);
            }
            return string.Concat("NS.", clientType.Name);
        }
        if (clientType.Namespace.Equals("Default", StringComparison.Ordinal))
        {
            if (originalNameAttribute != null)
            {
                return string.Concat("Default.", originalNameAttribute.OriginalName);
            }
            return string.Concat("Default.", clientType.Name);
        }
        if (originalNameAttribute != null)
        {
            return clientType.Namespace + "." + originalNameAttribute.OriginalName;
        }
        return clientType.FullName;
    }
	
    protected Type ResolveTypeFromName(string typeName)
    {
        Type resolvedType = this.DefaultResolveType(typeName, "NS", "NS");
        if (resolvedType != null)
        {
            return resolvedType;
        }
        resolveType = this.DefaultResolveType(typeName, "Default", "Default");
        if (resolvedType != null)
        {
            return resolvedType;
        }
        return null;
    }
}
```
Setting these delegates assume more importance in scenarios where the names for proxy classes generated on the client are not an exact match to the type names returned in the service metadata. For the sample service metadata I provided in the PR description for example, a class named `Shape` will be generated for the entity type named "shape", while `Rectangle` class will be generated for the entity type named "rectangle". The OData Connected Service extension takes care of this by decorating the classes with `OriginalNameAttribute`. For example,
```
namespace NS
{
    // using statements

    [Key("id")]
    [EntitySet("shapes")]
    [OriginalName("shape")]
    public class Shape
    {
        [OriginalName("id")]
        public int Id { get; set; }
        [OriginalName("area")]
        public double Area { get; set; }
    }

    [Key("id")]
    [OriginalName("rectangle")]
    public class Rectangle : Shape
    {
        public Rectangle()
        {
            Attributes = new ObservableCollection<string>();
        }

        [OriginalName("length")]
        public double Length { get; set; }
        [OriginalName("width")]
        public double Width { get; set; }
        [OriginalName("attributes")]
        public ObservableCollection<string> Attributes { get; set; }
    }
}
```
The `ResolveName` and `ResolveType` delegates will make use of `OriginalNameAttribute` to achieve the expected name and type resolution during serialization and materialization.

I'll check with the affected customers whether they're using OData Connected Service extension - hopefully the latest - since that would help mitigate the issue. If they're using POCOs, I'll advise them to apply `OriginalNameAttribute` in addition to setting the relevant delegates for type resolution.

In this PR, I have made some fixes and improvements. The `DefaultResolveType` method defined in `DataServiceContext` was ineffective since it was scanning the `Microsoft.OData.Client` assembly for client types. The code [`this.GetType().GetAssembly()`](https://github.com/OData/odata.net/blob/f7e39938f81e0b2780281a91ff2af8b5e82c1671/src/Microsoft.OData.Client/DataServiceContext.cs#L3261) used in the method resolves the library that `DataServiceContext` class belongs to. The outcome in always a miss. A more inefficient [`ResolveFromName`](https://github.com/OData/odata.net/blob/4e263ca4b94d01f2d105d8282c62ec272cfcd1bb/src/Microsoft.OData.Client/Metadata/ClientTypeCache.cs#L37) method defined in `ClientTypeCache` class almost always ends up being [triggered](https://github.com/OData/odata.net/blob/4e263ca4b94d01f2d105d8282c62ec272cfcd1bb/src/Microsoft.OData.Client/TypeResolver.cs#L438) to resolve the type. I rewrote the logic to make use of a method that scans the loaded assemblies instead. In the process I also added a cache to help with faster resolution in subsequent calls.

Another improvement I made is in matching the names of the Edm types generated on the client to the equivalent types in the service Edm model. The [`GetOrCreateEdmTypeInternal`](https://github.com/OData/odata.net/blob/4e263ca4b94d01f2d105d8282c62ec272cfcd1bb/src/Microsoft.OData.Client/ClientEdmModel.cs#L404) method defined in `ClientEdmModel` class makes use of a utility method `GetModelTypeName` that doesn't bother with `OriginalNameAttribute`. By modifying it to pay attention to the attribute, the generated Edm types will have names that mirror those of the service Edm model and this will help with quick and smooth type resolution.

marabooy (approved):

src/Microsoft.OData.Client/ClientEdmModel.cs
```diff
@@ -328,7 +328,7 @@ private static Type[] GetTypeHierarchy(Type type, out PropertyInfo[] keyProperti
                 {
                     hierarchy.Insert(0, type);
                 }
-                while ((type = c.PlatformHelper.GetBaseType(type)) != null);
+                while ((type = c.PlatformHelper.GetBaseType(type)) != null && type != typeof(object));
```
mikepizzo, line 331: >&& type != typeof(object) [](http://example.com/codeflow?start=75&length=25)

should we put this check first? is it cheaper? more likely to be true?

mikepizzo (approved):
:shipit:

src/Microsoft.OData.Client/ClientEdmModel.cs
gathogojr: @mikepizzo Because this is a do...while loop, we want to check if `type` after it has been assigned. The assignment happens in the first expression: `type = c.PlatformHelper.GetBaseType(type)`

PR #2131, opened 2021-07-14
Fix issue with async reader for Parsing the Context Uri

Sreejithpin:
### Issues
Fix issue with async reader for Parsing the Context Uri - Its there good in sync but not in async

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
@@ -1162,7 +1162,7 @@ namespace Microsoft.OData.JsonLight
                     UriUtils.UriToString(nestedInfo.ContextUrl),
                     payloadKind,
                     this.jsonLightResourceDeserializer.MessageReaderSettings.ClientCustomTypeResolver,
-                    this.jsonLightResourceDeserializer.JsonLightInputContext.ReadingResponse).Path;
+                    this.jsonLightResourceDeserializer.JsonLightInputContext.ReadingResponse || this.ReadingDelta).Path;
 
                 return new ODataUri() { Path = odataPath };
             }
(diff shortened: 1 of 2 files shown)


mikepizzo:
Please add test case(s)

mikepizzo:
>                         this.contextUriParseResult = ODataJsonLightContextUriParser.Parse(

There is still one call to Parse() in ODataJsonLightReader.ResolveODataUriFromContextUrl() that doesn't include the check for Delta -- should it be added there as well?

---
Refers to: src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs:301 in a6a33df. [](commit_id = a6a33df1006148c32463a4b405ae62dc1ad32b4e, deletion_comment = False)

mikepizzo (changes requested):
:clock1:

mikepizzo (approved):
:shipit:

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -202,11 +202,8 @@ protected override Task<bool> ReadAtStartImplementationAsync()
                 this.jsonLightInputContext.CreatePropertyAndAnnotationCollector();
 
             // Position the reader on the first node depending on whether we are reading a nested payload or not.
-            ODataPayloadKind payloadKind = this.ReadingDelta
-                ? ODataPayloadKind.Delta
-                : this.ReadingResourceSet ?
-                    ODataPayloadKind.ResourceSet
-                    : ODataPayloadKind.Resource;
+            ODataPayloadKind payloadKind = this.ReadingResourceSet ?
+                this.ReadingDelta ? ODataPayloadKind.Delta : ODataPayloadKind.ResourceSet : ODataPayloadKind.Resource;
```
habbes, line 206: I think this is less readable than the old code. Maybe a bit of nesting might help:
```suggestion
            ODataPayloadKind payloadKind = this.ReadingResourceSet ?
                this.ReadingDelta ? ODataPayloadKind.Delta : ODataPayloadKind.ResourceSet
                : ODataPayloadKind.Resource;

```

habbes (approved):
Looks good. But readability could be improved.

habbes:
@Sreejithpin since you didn't link to a specific issue, it was hard to tell what the actual issue being fixed is. Could you add more context to the description illustrating the actual bug and the difference in behaviour between the sync an async implementations?

PR #2133, opened 2021-07-16
Cache annotation full qualified name in FindDeclaredVocabularyAnnotations

joaocpaiva:
### Issues

3% of all allocations in AGS, come from this method, allocating string for full qualified name. This also represents 16% of all string allocations for AGS.

### Description

This fix attempts to reduce string allocations by caching the annotation full qualified name in CsdlSemanticsElement the first time it is evaluated.

image

- [X ] *Test cases already covering this code path*

### DIFF
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsElement.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsElement.cs
@@ -6,6 +6,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Linq;
 using Microsoft.OData.Edm.Csdl.Parsing.Ast;
 using Microsoft.OData.Edm.Validation;
@@ -95,7 +96,7 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
         /// <returns>The cached annotation full qualified name.</returns>
         public string GetAnnotationFullQualifiedName(IEdmVocabularyAnnotatable element)
         {
-            System.Diagnostics.Debug.Assert(object.ReferenceEquals(this as IEdmVocabularyAnnotatable, element));
+            Debug.Assert(object.ReferenceEquals(this as IEdmVocabularyAnnotatable, element));
             this.annotationFullName = this.annotationFullName ?? EdmUtil.FullyQualifiedName(element);
             return this.annotationFullName;
         }
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsElement.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsElement.cs
@@ -26,6 +26,8 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
 
         private static readonly IEnumerable<IEdmVocabularyAnnotation> emptyVocabularyAnnotations = Enumerable.Empty<IEdmVocabularyAnnotation>();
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -316,33 +312,32 @@ public override IEnumerable<IEdmVocabularyAnnotation> FindDeclaredVocabularyAnno
...
 
-            List<CsdlSemanticsAnnotations> elementAnnotations;
-            string fullName = EdmUtil.FullyQualifiedName(element);
-
-            if (fullName != null && this.outOfLineAnnotations.TryGetValue(fullName, out elementAnnotations))
+            if (this.outOfLineAnnotations.Count > 0)
             {
-                List<IEdmVocabularyAnnotation> result = new List<IEdmVocabularyAnnotation>();
+                List<CsdlSemanticsAnnotations> elementAnnotations;
+                string fullName = semanticsElement.GetAnnotationFullQualifiedName(element);
```
joaocpaiva, line 323: This code works under the assumption CsdlSemanticsElement instances are always the same (i.e. reused during process lifetime) and immutable, which should be the case of AGS. Not sure there are other scenarios, edge cases to cover?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
```diff
@@ -303,10 +303,6 @@ public override IEnumerable<IEdmVocabularyAnnotation> FindDeclaredVocabularyAnno
                 {
                     return semanticsReference.InlineVocabularyAnnotations;
                 }
-                else
-                {
-                    Enumerable.Empty<IEdmVocabularyAnnotation>();
```
KenitoInc, line 308: I think this was intended to return an empty enumerable.
So it should probably be `return Enumerable.Empty<IEdmVocabularyAnnotation>();`
What do you think @xuzhg @mikepizzo ?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsElement.cs
```diff
@@ -86,6 +88,18 @@ public IEnumerable<IEdmDirectValueAnnotation> DirectValueAnnotations
...
         }
 
+        /// <summary>
+        /// Gets the cached annotation full qualified name.
+        /// </summary>
+        /// <param name="element">This element as <see cref="IEdmVocabularyAnnotatable"/>.</param>
+        /// <returns>The cached annotation full qualified name.</returns>
+        public string GetAnnotationFullQualifiedName(IEdmVocabularyAnnotatable element)
+        {
+            System.Diagnostics.Debug.Assert(object.ReferenceEquals(this as IEdmVocabularyAnnotatable, element));
```
KenitoInc, line 98: Add `using System.Diagnostics`

KenitoInc (changes requested):

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsModel.cs
joaocpaiva: Done

KenitoInc (approved):
LGTM

marabooy (approved):

xuzhg (approved):
:shipit:

KenitoInc:
Thanks @joaocpaiva for the PR

PR #2138, opened 2021-07-26
ODL DI restructure

### LINKED ISSUE #2385: Getting rid of the IContainerBuilder interface in ODL and using Microsoft.Extensions.Dependency.Injection package to register the default OData servicies. This will make it easy for library users to add the default services to their projects.

ElizabethOkerio:
### Issues

*This pull request fixes #2385.*

### Description
Getting rid of the IContainerBuilder interface in ODL and using <code>Microsoft.Extensions.Dependency.Injection </code> package to register the default OData servicies. This will make it easy for library users to add the default services to their projects. 
Customer will be adding OData services like this to their projects: 

```c#
var services = new ServiceCollection();
services.AddDefaultODataServices();
```
or
`services.AddDefaultODataServices(ODataVersion.V401);`

or 
`services.AddDefaultODataServices(configureReaderAction: (reader) => reader.EnableCharactersCheck = true);`

To get the services: 
`var provider = services.BuildServiceProvider();`
`provider.GetService<IJsonReaderFactory>()`

- [ ] *Test cases added*

### Additional work necessary


test/Common/Microsoft.Test.OData.DependencyInjection/TestContainerBuilder.cs
```diff
@@ -12,37 +12,9 @@
...
-        {
-            Debug.Assert(serviceType != null, "serviceType != null");
-            Debug.Assert(implementationFactory != null, "implementationFactory != null");
-
-            services.Add(new ServiceDescriptor(
-                serviceType, implementationFactory, TranslateServiceLifetime(lifetime)));
-
-            return this;
-        }
+        public IServiceCollection services = new ServiceCollection();
```
gathogojr, line 17:
```suggestion
        public IServiceCollection Services { get; } = new ServiceCollection();
```

test/EndToEndTests/Tests/Client/Build.Desktop/PluggableFormatServiceTests/PluggableFormatServiceTests.cs
```diff
@@ -287,14 +289,17 @@ public void InvokeAvroAction()
 
         private static void SetAvroMediaTypeResolver(HttpWebRequestMessage requestMessage)
         {
-            requestMessage.Container = ContainerBuilderHelper.BuildContainer(builder =>
-                builder.AddService<ODataMediaTypeResolver, AvroMediaTypeResolver>(ServiceLifetime.Singleton));
+            IServiceCollection services = new ServiceCollection();
+            services.AddDefaultODataServices().BuildServiceProvider();
+            IServiceProvider serviceProvider = services.AddDefaultODataServices().BuildServiceProvider();
+                requestMessage.Container = ContainerBuilderHelper.BuildContainer(builder =>
```
gathogojr, line 295: Looks like there's code formatting issues here...

test/EndToEndTests/Tests/Client/Build.Desktop/PluggableFormatServiceTests/PluggableFormatServiceTests.cs
```diff
@@ -287,14 +289,17 @@ public void InvokeAvroAction()
 
         private static void SetAvroMediaTypeResolver(HttpWebRequestMessage requestMessage)
         {
-            requestMessage.Container = ContainerBuilderHelper.BuildContainer(builder =>
-                builder.AddService<ODataMediaTypeResolver, AvroMediaTypeResolver>(ServiceLifetime.Singleton));
+            IServiceCollection services = new ServiceCollection();
+            services.AddDefaultODataServices().BuildServiceProvider();
```
gathogojr, line 293: Do you need this line given that you're calling `services.AddDefaultODataServices().BuildServiceProvider();` on the line that follows?

test/Common/Microsoft.Test.OData.DependencyInjection/TestContainerBuilder.cs
gathogojr: `services` was a private readonly field. To expose the same publicly, the right thing is to make it a property and use Pascal case

src/Microsoft.OData.Core/DefaultServicesConfigurations.cs
```diff
@@ -0,0 +1,63 @@
...
+    /// </summary>
+    public static class DefaultServicesConfigurations
+    {
+        /// <summary>
+        /// Adds the default OData services to the <see cref="IServiceCollection"/>.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/> to add the services to.</param>
+        /// <returns>The <see cref="IServiceCollection"/> instance itself.</returns>
+        public static IServiceCollection AddDefaultODataServices(this IServiceCollection services)
+        {
```
gathogojr, line 26: I think you need to check `services` for null in this method as well, otherwise `services.AddDefaultODataServices(ODataVersion.V4)` might throw an exception?

src/Microsoft.OData.Core/DefaultServicesConfigurations.cs
```diff
@@ -0,0 +1,68 @@
...
+
+            services.AddSingleton<IJsonReaderFactory, DefaultJsonReaderFactory>();
+            services.AddSingleton<IJsonWriterFactory, DefaultJsonWriterFactory>();
+            services.AddSingleton<IJsonWriterFactoryAsync, DefaultJsonWriterFactory>();
+            services.AddSingleton(sp => ODataMediaTypeResolver.GetMediaTypeResolver(null));
+            services.AddScoped<ODataMessageInfo>();
+            services.AddSingleton(sp => new ServicePrototype<ODataMessageReaderSettings>(new ODataMessageReaderSettings()));
+            services.AddScoped(sp => sp.GetServicePrototype<ODataMessageReaderSettings>().Clone());
+            services.AddSingleton(sp => new ServicePrototype<ODataMessageWriterSettings>(new ODataMessageWriterSettings()));
+            services.AddScoped(sp => sp.GetServicePrototype<ODataMessageWriterSettings>().Clone());
```
habbes, line 56: I think we should consider dropping the use of `ServicePrototype` and just register new instances as scoped directly. But I guess that should be a separate PR discussion.

@xuzhg @gathogojr looked at this code and we couldn't understand why it was designed this way. And it may be potentially related to some race conditions that occur due to some shared settings state being modified concurrently by separate requests.

test/Common/Microsoft.Test.OData.DependencyInjection/TestContainerBuilder.cs
```diff
@@ -5,48 +5,18 @@
...
-using System.Diagnostics;
 using Microsoft.Extensions.DependencyInjection;
-using Microsoft.OData;
 using ServiceLifetime = Microsoft.Extensions.DependencyInjection.ServiceLifetime;
 
 namespace Microsoft.Test.OData.DependencyInjection
 {
-    public class TestContainerBuilder : IContainerBuilder
+    public class TestContainerBuilder 
     {
```
habbes, line 14: I think we can get rid of his entire file altogether since we no longer use `IContainerBuilder`.

test/Common/Microsoft.Test.OData.DependencyInjection/ContainerBuilderHelper.cs
```diff
@@ -5,21 +5,23 @@
...
     {
-        public static IServiceProvider BuildContainer(Action<IContainerBuilder> action)
+        public static IServiceProvider BuildContainer(Action<IServiceCollection> action)
         {
-            IContainerBuilder builder = new TestContainerBuilder();
-
-            builder.AddDefaultODataServices();
+            var builder = new TestContainerBuilder();
+            IServiceCollection services = builder.Services;
+            services.AddDefaultODataServices();
```
habbes, line 20: I think we can just call `new ServiceCollection()`  and `services.BuildServiceProvider()` directly and get rid of `TestContainerBuilder`.

test/EndToEndTests/Services/ODataDefaultService/DefaultWCFSvcDataSource.cs
```diff
@@ -1205,15 +1206,15 @@ protected override IEdmModel CreateModel()
             return DefaultInMemoryModel.CreateODataServiceModel("Microsoft.Test.OData.Services.ODataWCFService");
         }
 
-        protected override void ConfigureContainer(IContainerBuilder builder)
+        protected override void ConfigureContainer(IServiceCollection services)
         {
-            base.ConfigureContainer(builder);
-            builder.AddServicePrototype(new ODataSimplifiedOptions()
```
habbes, line 1211: Where is `AddServicePrototype` defined? Can we see what it's doing under the hood?

test/EndToEndTests/Services/ODataDefaultService/DefaultWCFSvcDataSource.cs
```diff
@@ -1205,15 +1206,15 @@ protected override IEdmModel CreateModel()
             return DefaultInMemoryModel.CreateODataServiceModel("Microsoft.Test.OData.Services.ODataWCFService");
         }
 
-        protected override void ConfigureContainer(IContainerBuilder builder)
+        protected override void ConfigureContainer(IServiceCollection services)
         {
-            base.ConfigureContainer(builder);
-            builder.AddServicePrototype(new ODataSimplifiedOptions()
+            base.ConfigureContainer(services);
+            services.AddSingleton(sp => new ServicePrototype<ODataSimplifiedOptions>(new ODataSimplifiedOptions()
```
habbes, line 1212: Was this registered as singleton in the old version? It's not immediately clear from the diff.

test/EndToEndTests/Services/ODataPluggableFormatService/PluggableFormatServiceDataSource.cs
```diff
@@ -118,10 +119,10 @@ protected override IEdmModel CreateModel()
             return TestHelper.GetModel("Microsoft.Test.OData.Services.PluggableFormat.Csdl.PluggableFormat.xml");
         }
 
-        protected override void ConfigureContainer(IContainerBuilder builder)
+        protected override void ConfigureContainer(IServiceCollection serviceProvider)
```
habbes, line 122:
```suggestion
        protected override void ConfigureContainer(IServiceCollection serviceCollection)

```

I think `serviceProvider` can be confusing name here since `provider` is associated with a read interface (i.e. retrieving services from the container/provider) and here we are adding a service (i.e. `IServiceCollection` vs `IServiceProvider`)

test/FunctionalTests/Microsoft.OData.Core.Tests/ContainerBuilderExtensionsTests.cs
```diff
@@ -15,44 +16,44 @@ public class ContainerBuilderExtensionsTests
         [Fact]
         public void AddServiceWithServiceType()
         {
-            IContainerBuilder builder = new TestContainerBuilder();
-            builder.AddService(ServiceLifetime.Transient, typeof(Foo));
+            var builder = new TestContainerBuilder();
```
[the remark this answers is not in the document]
habbes: @KenitoInc  it's the `TestContainerBuilder` class. But I think we should get rid of it and just use `new ServiceCollection()` directly. `TestContainerBuilder` was needed cause we were mocking `IContainerBuilder`, but since we're getting rid of `IContainerBuilder` we can get rid of `TestContainerBuilder` as well.

habbes (changes requested):

habbes:
@ElizabethOkerio is this a breaking change?

habbes:
@ElizabethOkerio could you show how consumers configured the services when ODL used `IContainerBuilder` vs how it will look now with the new changes? Or will nothing change from the consumer's end?

src/Microsoft.OData.Core/DefaultServicesConfigurations.cs
ElizabethOkerio: @habbes  I saw prototypes are used when instantiation of a service requires a lot of resources or is resource intensive. A prototype is created which is then cloned in subsequent calls so that the whole instantiation of the service is not done over and over again. I agree more discussion is required here.

test/EndToEndTests/Services/ODataDefaultService/DefaultWCFSvcDataSource.cs
ElizabethOkerio: @habbes          
''' c sharp
/// <param name="instance">The service prototype to add.</param>
        /// <returns>The <see cref="IContainerBuilder"/> instance itself.</returns>
        public static IContainerBuilder AddServicePrototype<TService>(
            this IContainerBuilder builder,
            TService instance)
        {
            Debug.Assert(builder != null, "builder != null");

            return builder.AddService(ServiceLifetime.Singleton, sp => new ServicePrototype<TService>(instance));
        }
'''

src/Microsoft.OData.Core/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,68 @@
...
+
+        /// <summary>
+        /// Adds the default OData services to the <see cref="IServiceCollection"/>.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/> to add the services to.</param>
+        /// <param name="odataVersion">ODataVersion for the default services.</param>
+        /// <returns>The <see cref="IServiceCollection"/> instance itself.</returns>
+        public static IServiceCollection AddDefaultODataServices(this IServiceCollection services, ODataVersion odataVersion)
+        {
+            if (services == null)
```
corranrogue9, line 43: I don't know if we are hyper-focused on performance, but a `private` overload would allow us to do the `null` check only once per call

corranrogue9 (dismissed):

mikepizzo:
@robertmclaws -- we are considering taking this breaking change for ODL 8. I know you have strong thoughts in this area -- can you comment on the approach?

src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
robertmclaws, line 1: - The .NET Standard 1.1 conditional references should be deleted.
- I think there is the potential to break people in unnecessary ways if .NET Standard 2.0 is not supported.
- If .NET Standard 2.0 support is added back in, the conditional references at the bottom should still be deleted.
- Is System.Buffers 4.5.1 needed on .NET 8? Should check that is properly limited where necessary.
- Is SUPPRESS_PORTABLELIB_TARGETFRAMEWORK_ATTRIBUTE still needed?
- Consider using [Directory.Build.props](https://github.com/OData/RESTier/blob/main/src/Directory.Build.props) to remove the analyzers, SourceLink, and InternalsVisibleTo from the project file.
- Please consider moving PublicAPI.txt -related files out of this project... they are testing resources and should not be shipped in the NuGet package.
- If you are changing item types (starting on Line 41) then the adjustments (Line 75 to Line 84) should appear immediately afterward.

src/Microsoft.OData.Core/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+// </copyright>
+//---------------------------------------------------------------------
+
+using System;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.OData.Edm;
+using Microsoft.OData.Json;
+using Microsoft.OData.UriParser;
+
+namespace Microsoft.OData.Core
```
robertmclaws, line 13: These extensions should actually live in `Microsoft.Extensions.DependencyInjection` so that developers don't have to add `using Microsoft.OData.Core` in the file, and the extensions magically appear to them.

Best practices are to ALWAYS include extension methods in the namespace they are extending. Making a developer futz around for the right namespace for an extension they are expecting is _no bueno_.

src/Microsoft.OData.Core/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    /// <summary>
+    /// Extension methods for <see cref="IServiceCollection"/>.
+    /// </summary>
+    public static class ServiceCollectionExtensions
```
robertmclaws, line 18: This should be named `OData_ServiceCollectionExtensions` so that you can differentiate OData-specific `IServiceCollection` extensions from other extensions using Reflection, given the namespace guidance above.

src/Microsoft.OData.Core/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        /// <summary>
+        /// Adds the default OData services to the <see cref="IServiceCollection"/>.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/> to add the services to.</param>
+        /// <param name="odataVersion">ODataVersion for the default services.</param>
+        /// <returns>The <see cref="IServiceCollection"/> instance itself.</returns>
+        public static IServiceCollection AddDefaultODataServices(this IServiceCollection services, ODataVersion odataVersion)
+        {
+            if (services == null)
```
robertmclaws, line 38: This should be `if (services is null)`. This new syntax is different than an equality check and improves performance.

src/Microsoft.OData.Core/ServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,62 @@
...
+            services.AddScoped(sp => sp.GetServicePrototype<ODataMessageWriterSettings>().Clone());
+            services.AddSingleton(sp => ODataPayloadValueConverter.GetPayloadValueConverter(null));
+            services.AddSingleton<IEdmModel>(sp => EdmCoreModel.Instance);
+            services.AddSingleton(sp => ODataUriResolver.GetUriResolver(null));
+            services.AddScoped<ODataUriParserSettings>();
+            services.AddScoped<UriPathParser>();
+            services.AddSingleton(sp => new ServicePrototype<ODataSimplifiedOptions>(new ODataSimplifiedOptions(odataVersion)));
+            services.AddScoped(sp => sp.GetServicePrototype<ODataSimplifiedOptions>().Clone());
+
+            return AddDefaultODataServices(services, ODataVersion.V4);
```
robertmclaws: Is this creating an infinite loop? Shoudn't it just be returning `services`?

src/Microsoft.OData.Core/ServicePrototype.cs
robertmclaws, line 1: This seems highly unnecessary... why are we needing an abstraction over a service here?

robertmclaws (changes requested):
Please take a look at the comments provided and let me know if we need to jump on a Teams call to discuss further.

src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
```diff
@@ -59,6 +59,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.Extensions.DependencyInjection" Version="1.1.0" />
```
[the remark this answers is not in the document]
ElizabethOkerio: updated.

src/Microsoft.OData.Core/ServiceCollectionExtensions.cs
ElizabethOkerio: updated

src/Microsoft.OData.Core/ServiceCollectionExtensions.cs
ElizabethOkerio: updated

test/Common/Microsoft.Test.OData.DependencyInjection/ServiceProviderBuilderHelper.cs
```diff
@@ -0,0 +1,25 @@
...
+// <copyright file="ServiceProviderBuilderHelper.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using System;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.OData.Core;
+
+namespace Microsoft.Test.OData.DependencyInjection
```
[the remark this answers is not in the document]
ElizabethOkerio: updated

test/EndToEndTests/Services/ODataDefaultService/DefaultWCFSvcDataSource.cs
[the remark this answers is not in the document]
ElizabethOkerio: @robertmclaws  we have instances we have such: `            services.AddSingleton(sp => new ServicePrototype<ODataMessageReaderSettings>(new ODataMessageReaderSettings()));
            services.AddScoped(sp => sp.GetServicePrototype<ODataMessageReaderSettings>().Clone());`

Are you suggesting we do away with this? I believe the cloning is to make it easy to create copies of some of the services without affecting their current state. What will be a better alternative?

src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
ElizabethOkerio: I'd like to create a new work item to track most of the items here outside this PR since this affects other project files as well and there are still ongoing discussions on which frameworks need to be supported and which don't.

test/EndToEndTests/Services/ODataDefaultService/DefaultWCFSvcDataSource.cs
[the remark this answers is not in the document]
robertmclaws: Yep, I am definitely suggesting we remove this concept all the way up the stack. It's not consistent with DI implementations in other ASP.NET subsystems.

I'll branch this PR and work on a suggested option. 

I think the method we've adopted for Restier has served us well, where [we only add things](https://github.com/OData/RESTier/blob/main/src/Microsoft.Restier.AspNet.Shared/Extensions/RestierApiServiceCollectionExtensions.cs#L42-L102) if they haven't already been added.

Ideally, you would allow someone to configure the single instance that is created, instead of registering multiple instances of the same type and then having to clone, combine, or deal with FIFO/LIFO issues.

In situations where you have different settings for different routes, then you would use DI 8.0's new Keyed Services to replace the per-route container concept, and then register different settings instances for the route's key. DI resolution would look for the Keyed service first, then fall back to the default service.

This is why I pushed us to upgrade to DI 8.0 in earlier feedback... Keyed services solve a huge problem that will make OData on .NET 8 reduce memory usage, reduce allocations, and be crazy fast.

I'll work on it over the weekend and then we can PR review my suggestions separately :). You're awesome, have a great weekend 🤜🏻

src/Microsoft.OData.Core/ServicePrototype.cs
```diff
@@ -5,7 +5,10 @@
 //---------------------------------------------------------------------
 
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 
+[assembly: InternalsVisibleTo("Microsoft.Test.OData.Services.ODataDefaultService" + AssemblyRef.TestPublicKey)]
+[assembly: InternalsVisibleTo("Microsoft.Test.OData.Services.ODataSimplifiedService" + AssemblyRef.TestPublicKey)]
```
habbes, line 11: Why are these added to this file?

src/Microsoft.OData.Core/ServicePrototype.cs
```diff
@@ -5,7 +5,10 @@
 //---------------------------------------------------------------------
 
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 
+[assembly: InternalsVisibleTo("Microsoft.Test.OData.Services.ODataDefaultService" + AssemblyRef.TestPublicKey)]
+[assembly: InternalsVisibleTo("Microsoft.Test.OData.Services.ODataSimplifiedService" + AssemblyRef.TestPublicKey)]
 namespace Microsoft.OData
 {
     internal class ServicePrototype<TService>
```
habbes, line 14: I'm not sure I understand the value or purpose of the `ServicePrototype` class. It seems to only wrap another object. What's the benefit of accessing the instance through the service prototype instead of accessing the instance directly?

`GetServicePrototype<TService>` just returns the `.Instance` property of the `ServicePrototype`. Why not just store the instance directly in the service collection and make it directly accessible via `GetService<T>` or `GetRequiredService<T>`

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchWriterTests.cs
```diff
@@ -503,9 +504,9 @@ public async Task WriteBatchResponseAsync_WithStreamCopy()
         [Fact]
         public async Task WriteBatchResponseAsync_WithStreamCopy_UsingODataUtf8JsonWriter()
         {
-            Action<IContainerBuilder> configureServices = (builder) =>
+            Action<IServiceCollection> configureServices = (builder) =>
             {
-                builder.AddService<IStreamBasedJsonWriterFactory>(ServiceLifetime.Singleton, _ =>DefaultStreamBasedJsonWriterFactory.Default);
+                builder.AddSingleton<IStreamBasedJsonWriterFactory>(_ =>DefaultStreamBasedJsonWriterFactory.Default);
```
habbes, line 509:
```suggestion
                builder.AddSingleton<IStreamBasedJsonWriterFactory>(_ => DefaultStreamBasedJsonWriterFactory.Default);

```

src/Microsoft.OData.Core/ServicePrototype.cs
ElizabethOkerio: removed.

test/EndToEndTests/Services/ODataDefaultService/DefaultWCFSvcDataSource.cs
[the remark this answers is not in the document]
ElizabethOkerio: @robertmclaws  I think we can get rid of the `ServicePrototype` and register the instances directly but leave the cloning for now. We welcome a contribution on how to do away with the cloning.

habbes (approved):

src/Microsoft.OData.Core/ODataMediaTypeResolver.cs
```diff
@@ -6,6 +6,7 @@
 
 namespace Microsoft.OData
 {
+    using Microsoft.Extensions.DependencyInjection;
```
gathogojr, line 9: Why is this one outside the `Namespaces` region?

test/Common/Microsoft.Test.OData.DependencyInjection/ServiceProviderBuilderHelper.cs
```diff
@@ -0,0 +1,24 @@
...
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using System;
+using Microsoft.Extensions.DependencyInjection;
+
+namespace Microsoft.OData.Core.Tests.DependencyInjection
+{
+    public static class ServiceProviderBuilderHelper
```
gathogojr, line 12: nit: `ServiceProviderBuilder`?

test/FunctionalTests/Microsoft.OData.Core.Tests/ContainerBuilderExtensionsTests.cs
```diff
@@ -5,55 +5,56 @@
...
 
 namespace Microsoft.OData.Tests
 {
     public class ContainerBuilderExtensionsTests
     {
+        private IServiceCollection Services;
+        public ContainerBuilderExtensionsTests()
+        {
+            Services = new ServiceCollection();
+        }
```
gathogojr, line 20:
```suggestion
        public ContainerBuilderExtensionsTests()
        {
            Services = new ServiceCollection();
        }

```

test/FunctionalTests/Microsoft.OData.Core.Tests/ContainerBuilderExtensionsTests.cs
```diff
@@ -5,55 +5,56 @@
...
 using System;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Test.OData.DependencyInjection;
 using Xunit;
 
 namespace Microsoft.OData.Tests
 {
     public class ContainerBuilderExtensionsTests
     {
+        private IServiceCollection Services;
```
gathogojr, line 16:
```suggestion
        private IServiceCollection services;
```

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
```diff
@@ -11,12 +11,10 @@
...
 using Microsoft.OData.UriParser;
 using Microsoft.OData.Edm;
-using Microsoft.Test.OData.DependencyInjection;
 using Xunit;
-using System.Xml;
-using Microsoft.OData.Edm.Validation;
-using Microsoft.OData.Edm.Csdl;
 using System.Threading.Tasks;
+using Microsoft.OData.Core.Tests.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection;
```
gathogojr, line 17: Sort usings

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
```diff
@@ -11,9 +11,10 @@
 using System.Text;
 using Microsoft.OData.Tests.JsonLight;
 using Microsoft.OData.Edm;
-using Microsoft.Test.OData.DependencyInjection;
 using Xunit;
 using ODataErrorStrings = Microsoft.OData.Strings;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.OData.Core.Tests.DependencyInjection;
```
gathogojr, line 17: Sort usings

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightValueSerializerTests.cs
```diff
@@ -9,11 +9,12 @@
...
 using System.IO;
 using System.Text;
-using Microsoft.OData.Json;
 using Microsoft.OData.JsonLight;
+using Microsoft.OData.Json;
 using Microsoft.OData.Edm;
-using Microsoft.Test.OData.DependencyInjection;
 using Xunit;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.OData.Core.Tests.DependencyInjection;
```
gathogojr, line 17: Sort usings

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -121,10 +124,9 @@ public void WriteTopLevelStringPropertyWithStringEscapeOptionShouldWork(ODataStr
 
             if (stringEscapeOption != null)
             {
-                var containerBuilder = new Test.OData.DependencyInjection.TestContainerBuilder();
-                containerBuilder.AddDefaultODataServices();
-                containerBuilder.AddService(ServiceLifetime.Singleton, sp => new DefaultJsonWriterFactory(stringEscapeOption.Value));
-                request.Container = containerBuilder.BuildContainer();
+                IServiceCollection services = new ServiceCollection().AddDefaultODataServices();
+                services.AddSingleton(sp => new DefaultJsonWriterFactory(stringEscapeOption.Value));
+                request.Container = services.BuildServiceProvider();
```
gathogojr, line 129: Why can't you use `CreateTestServiceContainer` defined in this file?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonReaderWriterInjectionTests.cs
```diff
@@ -10,11 +10,13 @@
 using System.Linq;
 using System.Text;
 using System.Threading.Tasks;
+using Microsoft.Extensions.DependencyInjection;
 using Microsoft.OData.Edm;
 using Microsoft.OData.Json;
-using Microsoft.Test.OData.DependencyInjection;
 using Microsoft.Test.OData.Utils.ODataLibTest;
 using Xunit;
+using Microsoft.OData.Core.Tests.DependencyInjection;
```
gathogojr, line 18: Sort usings

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonReaderWriterInjectionTests.cs
```diff
@@ -98,11 +100,11 @@ public void UseComplexJsonReaderSelectComplex()
         [Fact]
         public void InjectCustomJsonReader()
         {
-            RunReaderTest(builder => builder.AddService<IJsonReaderFactory, TestJsonReaderFactory>(ServiceLifetime.Transient),
+            RunReaderTest(builder => builder.AddSingleton<IJsonReaderFactory, TestJsonReaderFactory>(),
```
gathogojr, line 103: The `TestJsonReaderFactory` was being added with transient scope before

test/FunctionalTests/Microsoft.OData.Core.Tests/ServiceProviderExtensionsTests.cs
```diff
@@ -6,84 +6,81 @@
...
 
 namespace Microsoft.OData.Tests
 {
     public class ServiceProviderExtensionsTests
     {
+        private IServiceCollection Services;
+        public ServiceProviderExtensionsTests()
+        {
+            Services = new ServiceCollection();
+        }
```
gathogojr, line 20:
```suggestion
        public ServiceProviderExtensionsTests()
        {
            Services = new ServiceCollection();
        }

```

test/FunctionalTests/Microsoft.OData.Core.Tests/ServiceProviderExtensionsTests.cs
```diff
@@ -6,84 +6,81 @@
...
 using System.Linq;
-using Microsoft.Test.OData.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection;
 using Xunit;
 
 namespace Microsoft.OData.Tests
 {
     public class ServiceProviderExtensionsTests
     {
+        private IServiceCollection Services;
```
gathogojr, line 16:
```suggestion
        private IServiceCollection services;
```

src/Microsoft.OData.Core/ODataServiceColectionExtensions.cs
```diff
@@ -0,0 +1,58 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ODataServiceColectionExtensions.cs" company="Microsoft">
+// Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using System;
```
gathogojr, line 7: The style in this project appears to be to nest usings inside the namespace block

src/Microsoft.OData.Core/ODataServiceColectionExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                throw new ArgumentNullException(nameof(services));
+            }
+
+            services.AddSingleton<IJsonReaderFactory, DefaultJsonReaderFactory>();
+            services.AddSingleton<IJsonWriterFactory, DefaultJsonWriterFactory>();
+            services.AddSingleton(sp => ODataMediaTypeResolver.GetMediaTypeResolver(null));
+            services.AddScoped<ODataMessageInfo>();
+            services.AddScoped(sp => new ODataMessageReaderSettings(odataVersion));
+            services.AddScoped(sp => new ODataMessageWriterSettings(odataVersion));
```
gathogojr, line 48: Unless it's an extreme undertaking, you should collect COGS stats comparing the previous use of `ServicePrototype` to register `ODataMessageWriterSettings` and `ODataMessageReaderSettings` plus cloning vs doing it in this new way.

test/Common/Microsoft.Test.OData.DependencyInjection/ServiceProviderBuilderHelper.cs
ElizabethOkerio: I think `ServiceProviderBuilder` sounds ok.

src/Microsoft.OData.Core/ODataServiceColectionExtensions.cs
robertmclaws: Please do not put usings inside the namespace block. It's an unnecessary readability hit.

src/Microsoft.OData.Core/PublicAPI/net48/PublicAPI.Unshipped.txt
```diff
@@ -1,4 +1,7 @@
+Microsoft.OData.IContainerProvider
```
robertmclaws, line 1: I see this getting added to the public API surface, but I don't see where it is coming from. Is it necessary?

test/EndToEndTests/Services/ODataDefaultService/DefaultWCFSvcDataSource.cs
```diff
@@ -1206,22 +1207,22 @@ protected override IEdmModel CreateModel()
...
-            builder.AddServicePrototype(new ODataMessageWriterSettings()
+            base.ConfigureContainer(services);
+            services.AddScoped(new ODataMessageWriterSettings()
             {
-                EnableWritingKeyAsSegment = false,
+                EnableReadingKeyAsSegment = false,
             });
 
-            builder.AddServicePrototype(new ODataMessageReaderSettings()
+            services.AddScoped(new ODataMessageWriterSettings()
```
robertmclaws, line 1218: Was it intended to add `ODataMessageWriterSettings` twice? If so, please leave a comment in the code to explain why.

test/EndToEndTests/Services/ODataSimplifiedService/ODataSimplifiedDataSource.cs
```diff
@@ -168,17 +169,18 @@ protected override IEdmModel CreateModel()
             return ODataSimplifiedInMemoryModel.CreateModel("Microsoft.OData.SampleService.Models.ODataSimplified");
         }
 
-        protected override void ConfigureContainer(IContainerBuilder builder)
+        protected override void ConfigureContainer(IServiceCollection services)
```
robertmclaws, line 172: I want to point out that, if I'm reading this right, this whole process is a DI anti-pattern. We should not be registering multiple different types of the same settings in the DI container and relying on FIFO or LIFO to resolve what we want.

The registration process should allow people to change the settings however they want via an Action lambda with overloads to the `AddDefaultODataServices` method. That would let people set the different options the right way at the right time without multiple registrations or cloning.

If done right, those settings would even be able to be specified in appsettings.json files and not just in code, because they would be applied by the configuration system.

src/Microsoft.OData.Core/ODataServiceColectionExtensions.cs
ElizabethOkerio: @gathogojr

test/EndToEndTests/Services/ODataDefaultService/DefaultWCFSvcDataSource.cs
ElizabethOkerio: We initially instantiated ODataMessageWriterSettings as a singleton, then generated clones of this instance as scoped, rather than consistently creating a new scope for ODataMessageWriterSettings. Upon evaluating the cloning process, it appeared to have negligible impact compared to generating a fresh instance of ODataMessageWriterSettings each time.

src/Microsoft.OData.Core/PublicAPI/net48/PublicAPI.Unshipped.txt
ElizabethOkerio: Yes. It is used.

ElizabethOkerio: It is in the `Microsoft.OData.Core` library

src/Microsoft.OData.Core/ODataServiceColectionExtensions.cs
ElizabethOkerio: Ran some benchmarks for the two methods and this is what I got: 
xzz

robertmclaws: Awesome! 75-80% perf improvement and 50% allocation reduction if I'm reading that right. Nicely done!

src/Microsoft.OData.Core/ODataServiceColectionExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// Extension methods for <see cref="IServiceCollection"/>.
+    /// </summary>
+    public static class ODataServiceColectionExtensions
+    {
+        /// <summary>
+        /// Adds the default OData services to the <see cref="IServiceCollection"/>.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/> to add the services to.</param>
+        /// <returns>The <see cref="IServiceCollection"/> instance itself.</returns>
+        public static IServiceCollection AddDefaultODataServices(this IServiceCollection services)
```
xuzhg, line 25: Maybe just name it 'AddOData(this IServiceCollection services)'

xuzhg (approved):

src/Microsoft.OData.Core/ODataServiceColectionExtensions.cs
habbes: Would this not be a source of conflict or confusion with AspNetCoreOData's `AddOData` extension method?

xuzhg: AspNetCoreOData, AddOData() is the extension method on 'IMvcBuilder'

For the IServiceCollection extension method, it's named as "AddODataCore" and its internal. So, I don't think there is a conflict.

test/FunctionalTests/Microsoft.OData.Core.Tests/ContainerBuilderExtensionsTests.cs
robertmclaws, line 1: I believe this file does not actually test anything useful, because there are not OData-specific things going on here. I removed it in my PR to your branch.

src/Microsoft.OData.Core/ODataServiceColectionExtensions.cs
```diff
@@ -0,0 +1,58 @@
...
+    /// Extension methods for <see cref="IServiceCollection"/>.
+    /// </summary>
+    public static class ODataServiceColectionExtensions
+    {
+        /// <summary>
+        /// Adds the default OData services to the <see cref="IServiceCollection"/>.
+        /// </summary>
+        /// <param name="services">The <see cref="IServiceCollection"/> to add the services to.</param>
+        /// <returns>The <see cref="IServiceCollection"/> instance itself.</returns>
+        public static IServiceCollection AddOData(this IServiceCollection services)
```
robertmclaws, line 25: Please don't change it to this. I am telling you it is going to cause confusion up the stack. When a developer sees all of these methods in there together, they are not going to know which one to pick.

robertmclaws (changes requested):

test/EndToEndTests/Services/ODataDefaultService/DefaultWCFSvcDataSource.cs
```diff
@@ -1206,22 +1207,22 @@ protected override IEdmModel CreateModel()
             return DefaultInMemoryModel.CreateODataServiceModel("Microsoft.Test.OData.Services.ODataWCFService");
         }
 
-        protected override void ConfigureContainer(IContainerBuilder builder)
+        protected override void ConfigureContainer(IServiceCollection services)
         {
-            base.ConfigureContainer(builder);
-            builder.AddServicePrototype(new ODataMessageWriterSettings()
+            base.ConfigureContainer(services);
+            services.AddScoped(new ODataMessageWriterSettings()
```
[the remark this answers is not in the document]
ElizabethOkerio: The method is not used anywhere. Removed.

test/EndToEndTests/Services/ODataDefaultService/DefaultWCFSvcDataSource.cs
```diff
@@ -1206,22 +1207,22 @@ protected override IEdmModel CreateModel()
...
-        protected override void ConfigureContainer(IContainerBuilder builder)
+        protected override void ConfigureContainer(IServiceCollection services)
         {
-            base.ConfigureContainer(builder);
-            builder.AddServicePrototype(new ODataMessageWriterSettings()
+            base.ConfigureContainer(services);
+            services.AddScoped(new ODataMessageWriterSettings()
             {
-                EnableWritingKeyAsSegment = false,
+                EnableReadingKeyAsSegment = false,
```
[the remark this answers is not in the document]
ElizabethOkerio: Removed the method not used anywhere..

src/Microsoft.OData.Core/PublicAPI/net48/PublicAPI.Unshipped.txt
ElizabethOkerio: I think the name of the interface could be updated instead of doing away with it. If in future we want to add the ServiceProvider to a new class then we can still reuse the interface.

ElizabethOkerio: IServiceCollectionProvider

robertmclaws:
It's definitely looking a lot more consistent now! 

Please take a look at my PR to your branch... I streamlined `AddDefaultODataServices` and added new functionality, along with 10 new unit tests to cover all permutations of using it. https://github.com/ElizabethOkerio/odata.net/pull/3

robertmclaws:
Here's an example of why this change is necessary: https://github.com/OData/RESTier/issues/688

src/Microsoft.OData.Core/IServiceCollectionProvider.cs
```diff
@@ -9,14 +9,14 @@
 namespace Microsoft.OData
 {
     /// <summary>
-    /// An interface that provides a dependency injection container.
+    /// An interface that provides a dependency injection service collection.
```
robertmclaws, line 12: `An interface that provides a dependency injection <see cref="IServiceProvider" />.`

test/EndToEndTests/Tests/Client/Build.Desktop/PluggableFormatServiceTests/PluggableFormatServiceTests.cs
```diff
@@ -287,14 +288,16 @@ public void InvokeAvroAction()
 
         private static void SetAvroMediaTypeResolver(HttpWebRequestMessage requestMessage)
         {
-            requestMessage.Container = ContainerBuilderHelper.BuildContainer(builder =>
-                builder.AddService<ODataMediaTypeResolver, AvroMediaTypeResolver>(ServiceLifetime.Singleton));
+            IServiceCollection services = new ServiceCollection();
+            IServiceProvider serviceProvider = services.AddDefaultODataServices().BuildServiceProvider();
+            requestMessage.Container = ServiceProviderHelper.BuildServiceProvider(builder =>
```
robertmclaws, line 293: Looks like `requestMessage.Container` needs to be changed to `requestMessage.ServiceProvider`.

test/FunctionalTests/Framework/ReflectionCodeGeneration/ConstructedInterfaces.cs
```diff
@@ -630,7 +630,7 @@ public string GetMethodDeclaration()
             StringBuilder builder = new StringBuilder();
             builder.AppendLine("      public object GetService(Type serviceType)");
             builder.AppendLine("      {");
-            builder.AppendLine("          APICallLog.Current.ServiceProvider.GetService(serviceType);");
+            builder.AppendLine("          APICallLog.Current.Container.GetService(serviceType);");
```
robertmclaws, line 633: I would argue the version on the left is probably closer to being correct.

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
```diff
@@ -1366,8 +1364,8 @@ private static IEdmModel GetModelWithAbstractType(out EdmEntitySet entitySet)
         public void WritingContainedNavigationPropertyWithoutProvideKeyPropertyOnParentResourceWorks(string metadataLevel)
         {
             MemoryStream outputStream = new MemoryStream();
-            var container = ContainerBuilderHelper.BuildContainer(null);
-            IODataResponseMessage message = new InMemoryMessage() { Stream = outputStream, Container = container };
+            var container = ServiceProviderHelper.BuildServiceProvider(null);
```
robertmclaws, line 1367: I know this is nitpicky, but changing the variable name to `serviceProvider` here, and other places in this file, would potentially reduce confusion.

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/FeedAnnotationIntegrationTests.cs
```diff
@@ -534,9 +535,10 @@ private void CreateMessageReader(string payloadBody, bool forEntry, bool isRespo
             const string payloadSuffix = "}";
             string payload = payloadPrefix + payloadBody + payloadSuffix;
 
-            var container = ContainerBuilderHelper.BuildContainer(null);
+            var container = ServiceProviderHelper.BuildServiceProvider(null);
```
robertmclaws, line 538: Maybe a search and replace to find `var container = ` and replace it with `var serviceProvider = ` would be in order?

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataServiceCollectionExtensionTests.cs
```diff
@@ -0,0 +1,230 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="BufferUtilsTests.cs" company="Microsoft">
```
robertmclaws, line 2: Looks like I didn't catch this. Should be `ODataServiceCollectionExtensionsTests.cs`. Oops.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Reader/JsonLight/ODataJsonLightSingletonReaderTests.cs
```diff
@@ -406,7 +406,7 @@ private ODataResource ReadSingleton(string payload, bool odataSimplified = false
                 MediaType = new ODataMediaType("application", "json"),
                 IsAsync = false,
                 Model = this.userModel,
-                Container = ContainerBuilderHelper.BuildContainer(null)
+                Container = ServiceProviderHelper.BuildServiceProvider(null)
```
robertmclaws, line 409: Should probably change the `Container` property here to `ServiceProvider`.

test/FunctionalTests/Microsoft.OData.Core.Tests/ServiceProviderExtensionsTests.cs
```diff
@@ -6,84 +6,82 @@
 
 using System;
 using System.Linq;
-using Microsoft.Test.OData.DependencyInjection;
+using Microsoft.Extensions.DependencyInjection;
 using Xunit;
 
 namespace Microsoft.OData.Tests
 {
     public class ServiceProviderExtensionsTests
```
robertmclaws, line 14: None of these methods test relevant OData functionality. The file should probably be deleted.

robertmclaws (changes requested):
Almost there (stay on target)!

ElizabethOkerio:
@robertmclaws

robertmclaws (approved):
I found one or two places in there where there were variable names like "containerProvider", but I've been nit-picky enough. Congrats on a job well-done! 🤜🏻

image

src/Microsoft.OData.Core/Batch/ODataBatchOperationRequestMessage.cs
```diff
@@ -85,7 +85,7 @@ public sealed class ODataBatchOperationRequestMessage : IODataRequestMessageAsyn
             this.groupId = groupId;
 
             this.message = new ODataBatchOperationMessage(contentStreamCreatorFunc, headers, operationListener, payloadUriConverter, writing);
-            this.Container = container;
+            this.ServiceProvider = container;
```
gathogojr, line 88: nit:
```suggestion
            this.ServiceProvider = serviceProvider;
```

src/Microsoft.OData.Core/Batch/ODataBatchOperationResponseMessage.cs
```diff
@@ -59,7 +59,7 @@ public sealed class ODataBatchOperationResponseMessage : IODataResponseMessageAs
 
             this.message = new ODataBatchOperationMessage(contentStreamCreatorFunc, headers, operationListener, payloadUriConverter, writing);
             this.ContentId = contentId;
-            this.Container = container;
+            this.ServiceProvider = container;
```
gathogojr, line 62: nit:
```suggestion
            this.ServiceProvider = serviceProvider;
```

src/Microsoft.OData.Core/ODataMessageReader.cs
```diff
@@ -890,8 +891,8 @@ internal ODataFormat GetFormat()
         private static IServiceProvider GetContainer<T>(T message)
             where T : class
         {
-            var containerProvider = message as IContainerProvider;
-            return containerProvider == null ? null : containerProvider.Container;
+            var containerProvider = message as IServiceCollectionProvider;
```
gathogojr, line 894: nit:
```suggestion
            var serviceCollectionProvider = message as IServiceCollectionProvider;
```

src/Microsoft.OData.Core/ODataMessageReader.cs
```diff
@@ -890,8 +891,8 @@ internal ODataFormat GetFormat()
         private static IServiceProvider GetContainer<T>(T message)
             where T : class
         {
-            var containerProvider = message as IContainerProvider;
-            return containerProvider == null ? null : containerProvider.Container;
+            var containerProvider = message as IServiceCollectionProvider;
+            return containerProvider == null ? null : containerProvider.ServiceProvider;
```
gathogojr, line 895:
```suggestion
            return containerProvider?.ServiceProvider;
```

src/Microsoft.OData.Core/ODataMessageWriter.cs
```diff
@@ -144,12 +145,12 @@ public ODataMessageWriter(IODataResponseMessage responseMessage, ODataMessageWri
         {
             ExceptionUtils.CheckArgumentNotNull(responseMessage, "responseMessage");
 
-            this.container = GetContainer(responseMessage);
-            this.settings = ODataMessageWriterSettings.CreateWriterSettings(this.container, settings);
+            this.serviceProvider = GetContainer(responseMessage);
```
gathogojr, line 148: nit:
```suggestion
            this.serviceProvider = GetServiceProvider(responseMessage);
```

src/Microsoft.OData.Core/ODataMessageWriter.cs
```diff
@@ -751,8 +752,8 @@ internal ODataFormat SetHeaders(ODataPayloadKind payloadKind)
         private static IServiceProvider GetContainer<T>(T message)
             where T : class
         {
-            var containerProvider = message as IContainerProvider;
-            return containerProvider == null ? null : containerProvider.Container;
+            var containerProvider = message as IServiceCollectionProvider;
+            return containerProvider == null ? null : containerProvider.ServiceProvider;
```
gathogojr, line 756:
```suggestion
            return (message as IServiceCollectionProvider)?.ServiceProvider;
```

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
```diff
@@ -441,71 +439,71 @@ static AutoComputePayloadMetadataInJsonIntegrationTests()
                 ContainsTarget = true
             });
 
-            var container = new EdmEntityContainer("Namespace", "Container");
-            EntitySet = container.AddEntitySet("EntitySet", EntityType);
+            var serviceProvider = new EdmEntityContainer("Namespace", "Container");
```
gathogojr, line 442:
```suggestion
            var entityContainer = new EdmEntityContainer("Namespace", "Container");
```
Why did we change `container` of type `EdmEntityContainer` into `serviceProvider`?

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/InstanceAnnotationsReaderIntegrationTests.cs
```diff
@@ -165,15 +165,18 @@ private static void TopLevelFeedInstanceAnnotationTest(string feedPayload, bool
...
             var stream = new MemoryStream(Encoding.UTF8.GetBytes(payload));
-            var readerSettings = new ODataMessageReaderSettings {
+
+            var readerSettings = new ODataMessageReaderSettings 
+            { 
                 EnableMessageStreamDisposal = false,
                 EnableReadingODataAnnotationWithoutPrefix = enableReadingODataAnnotationWithoutPrefix
             };
-            var container = ContainerBuilderHelper.BuildContainer(null);
+            var container = ServiceProviderHelper.BuildServiceProvider(null);
```
gathogojr, line 174: nit:
```suggestion
            var serviceProvider = ServiceProviderHelper.BuildServiceProvider(null);
```

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/ODataJsonLightReaderIntegrationTests.cs
```diff
@@ -1159,9 +1160,11 @@ private void ReadTopLevleCollectionPayload(IEdmModel userModel, string payload,
 
         private ODataResource ReadJsonLightEntry(string payload, string contentType, bool readingResponse, bool enableReadingODataAnnotationWithoutPrefix = false)
         {
-            var container = ContainerBuilderHelper.BuildContainer(null);
+            var container = ServiceProviderHelper.BuildServiceProvider(null);
```
gathogojr, line 1163: nit:
```suggestion
            var serviceProvider = ServiceProviderHelper.BuildServiceProvider(null);
```

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Reader/JsonLight/PropertyAndValueJsonLightReaderIntegrationTests.cs
```diff
@@ -970,8 +971,8 @@ public void ReadDateTimeOffsetWithCustomFormat()
                 "}";
 
             ODataResource entry = null;
-            var diContainer = ContainerBuilderHelper.BuildContainer(
-                builder => builder.AddService<ODataPayloadValueConverter, DateTimeOffsetCustomFormatPrimitivePayloadValueConverter>(ServiceLifetime.Singleton));
+            var diContainer = ServiceProviderHelper.BuildServiceProvider(
```
gathogojr, line 974: nit:
```suggestion
            var serviceProvider = ServiceProviderHelper.BuildServiceProvider(
```

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Writer/InstanceAnnotationWriterIntegrationTests.cs
```diff
@@ -706,7 +707,9 @@ private void WriteAnnotationsAndValidatePayload(Action<ODataWriter> action, IEdm
             writerSettings.SetContentType(format);
             writerSettings.SetServiceDocumentUri(new Uri("http://www.example.com/"));
 
-            var container = ContainerBuilderHelper.BuildContainer(null);
+            var container = ServiceProviderHelper.BuildServiceProvider(null);
```
gathogojr, line 710: nit:
```suggestion
            var serviceProvider = ServiceProviderHelper.BuildServiceProvider(null);
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonReaderWriterInjectionTests.cs
```diff
@@ -115,38 +116,41 @@ private static void RunWriterTest(Action<IContainerBuilder> action, string expec
...
 
-        private static Dictionary<string, IEnumerable<ODataProperty>> RunComplexReaderTest(Action<IContainerBuilder> action, string messageContent)
+        private static Dictionary<string, IEnumerable<ODataProperty>> RunComplexReaderTest(Action<IServiceCollection> action, string messageContent)
         {
             var model = GetModel();
             var entitySet = model.FindDeclaredEntitySet("People");
             var entityType = model.GetEntityType("DefaultNs.Person");
-            var container = ContainerBuilderHelper.BuildContainer(action);
+
+            var container = ServiceProviderHelper.BuildServiceProvider(action);
```
gathogojr, line 151: nit:
```suggestion
            var serviceProvider = ServiceProviderHelper.BuildServiceProvider(action);
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Writer/JsonLight/AutoGeneratedUrlsShouldPutKeyValueInDedicatedSegmentTests.cs
```diff
@@ -88,8 +89,10 @@ public void IfKeyAsSegmentSettingIsFalseAndModelHasKeyAsSegmentAnnotationThenLin
             var settings = new ODataMessageWriterSettings { EnableWritingKeyAsSegment = useKeyAsSegment };
             settings.SetServiceDocumentUri(new Uri("http://example.com/"));
             var outputStream = new MemoryStream();
-            var container = ContainerBuilderHelper.BuildContainer(null);
-            var responseMessage = new InMemoryMessage { Stream = outputStream, Container = container };
+            var container = ServiceProviderHelper.BuildServiceProvider(null);
```
gathogojr, line 92: nit:
```suggestion
            var serviceProvider = ServiceProviderHelper.BuildServiceProvider(null);
```

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserInjectionTests.cs
```diff
@@ -174,7 +175,7 @@ public ODataUriParserInjectionTests()
         [Fact]
         public void CustomizedUriPathParserTest()
         {
-            var container = ContainerBuilderHelper.BuildContainer(builder => builder.AddService<UriPathParser, MultipleSegmentUriPathParser>(ServiceLifetime.Scoped));
+            var container = ServiceProviderHelper.BuildServiceProvider(builder => builder.AddScoped<UriPathParser, MultipleSegmentUriPathParser>());
```
gathogojr, line 178: nit:
```suggestion
            var serviceProvider = ServiceProviderHelper.BuildServiceProvider(builder => builder.AddScoped<UriPathParser, MultipleSegmentUriPathParser>());
```

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserInjectionTests.cs
```diff
@@ -187,7 +188,7 @@ public void CustomizedUriPathParserTest()
         [Fact]
         public void ParseDynamicPathSegmentFunc_ReturnDynamicPathSegment_FollowedByDynamicPathSegmentsAndProp()
         {
-            var container = ContainerBuilderHelper.BuildContainer(builder => builder.AddService<UriPathParser, MultipleSegmentUriPathParser>(ServiceLifetime.Scoped));
+            var container = ServiceProviderHelper.BuildServiceProvider(builder => builder.AddScoped<UriPathParser, MultipleSegmentUriPathParser>());
```
gathogojr, line 191: nit:
```suggestion
            var serviceProvider = ServiceProviderHelper.BuildServiceProvider(builder => builder.AddScoped<UriPathParser, MultipleSegmentUriPathParser>());
```

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserInjectionTests.cs
```diff
@@ -297,7 +298,7 @@ private ODataUriParser ParseDynamicPathSegmentFunc_ReturnDynamicPathSegment_With
 
         private ODataUriParser ParseDynamicPathSegmentFunc_ReturnDynamicPathSegment(Uri fullUri, out ODataPath odataPath)
         {
-            var container = ContainerBuilderHelper.BuildContainer(builder => builder.AddService<UriPathParser, SingleSegmentUriPathParser>(ServiceLifetime.Scoped));
+            var container = ServiceProviderHelper.BuildServiceProvider(builder => builder.AddScoped<UriPathParser, SingleSegmentUriPathParser>());
```
gathogojr, line 301: nit:
```suggestion
            var serviceProvider = ServiceProviderHelper.BuildServiceProvider(builder => builder.AddScoped<UriPathParser, SingleSegmentUriPathParser>());
```

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserInjectionTests.cs
```diff
@@ -341,7 +342,7 @@ public void ParseDynamicPathSegmentFunc_FirstSegmentReturnNavAndSegment_WithProp
 
         private ODataPath ParseDynamicPathSegmentFunc_ReturnNavAndSegment(Uri fullUri)
         {
-            var container = ContainerBuilderHelper.BuildContainer(builder => builder.AddService<UriPathParser, SingleSegmentUriPathParser>(ServiceLifetime.Scoped));
+            var container = ServiceProviderHelper.BuildServiceProvider(builder => builder.AddScoped<UriPathParser, SingleSegmentUriPathParser>());
```
gathogojr, line 345: nit:
```suggestion
            var serviceProvider = ServiceProviderHelper.BuildServiceProvider(builder => builder.AddScoped<UriPathParser, SingleSegmentUriPathParser>());
```

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserInjectionTests.cs
```diff
@@ -375,7 +376,7 @@ private ODataPath ParseDynamicPathSegmentFunc_ReturnNavAndSegment(Uri fullUri)
         [Fact]
         public void ParseDynamicPathSegmentFunc_ReturnOperationSegment_WithCollectionReturnType_WithCount()
         {
-            var container = ContainerBuilderHelper.BuildContainer(builder => builder.AddService<UriPathParser, SingleSegmentUriPathParser>(ServiceLifetime.Scoped));
+            var container = ServiceProviderHelper.BuildServiceProvider(builder => builder.AddScoped<UriPathParser, SingleSegmentUriPathParser>());
```
gathogojr, line 379: nit:
```suggestion
            var serviceProvider = ServiceProviderHelper.BuildServiceProvider(builder => builder.AddScoped<UriPathParser, SingleSegmentUriPathParser>());
```

gathogojr (approved):
LG

src/Microsoft.OData.Core/ODataServiceCollectionExtensions.cs
```diff
@@ -0,0 +1,72 @@
...
+        /// <param name="configureWriterAction">An <see cref="Action{ODataMessageWriterSettings}" /> that allows for configuring the default <see cref="ODataMessageWriterSettings" />.</param>
+        /// <param name="configureUriParserAction">An <see cref="Action{ODataUriParserSettings}" /> that allows for configuring the default <see cref="ODataUriParserSettings" />.</param>
+        /// <returns>The <see cref="IServiceCollection"/> instance we added OData services to, for fluent manipulation.</returns>
+        /// <remarks>This is usually called by OData platform libraries as a part of initializing runtime components of the OData platform.</remarks>
+        /// <exception cref="ArgumentNullException">Thrown if the <paramref name="services"/> parameter is <c>null</c>.</exception>
+        public static IServiceCollection AddDefaultODataServices(this IServiceCollection services,
+            ODataVersion odataVersion = ODataVersion.V4, 
+            Action<ODataMessageReaderSettings> configureReaderAction = default, 
+            Action<ODataMessageWriterSettings> configureWriterAction = default,
+            Action<ODataUriParserSettings> configureUriParserAction = default)
```
habbes, line 36: I wonder whether it would be better to have one Options parameter that you can use to configure the different settings instead of having all these parameters. But maybe the options parameter would lead to confusion by introducing an extra layer of settings on top of a layer of settings, etc.

src/Microsoft.OData.Core/IServiceCollectionProvider.cs
```diff
@@ -9,14 +9,14 @@
 namespace Microsoft.OData
 {
     /// <summary>
-    /// An interface that provides a dependency injection container.
+    /// An interface that provides a dependency injection <see cref="IServiceProvider"./>
     /// </summary>
-    public interface IContainerProvider
+    public interface IServiceCollectionProvider
```
habbes, line 14: Why do we need this interface?

src/Microsoft.OData.Core/ODataMessageReader.cs
```diff
@@ -887,29 +888,29 @@ internal ODataFormat GetFormat()
...
         }
 
-        private static IServiceProvider GetContainer<T>(T message)
+        private static IServiceProvider GetServiceProvider<T>(T message)
             where T : class
         {
-            var containerProvider = message as IContainerProvider;
-            return containerProvider == null ? null : containerProvider.Container;
+            var serviceCollectionProvider = message as IServiceCollectionProvider;
+            return serviceCollectionProvider?.ServiceProvider;
```
habbes, line 895: I don't know whether this is outside the scope of this PR, but I'm wondering why we retrieve the service provider from the `IODataRequest/ResponseMessage`. This seems to be the only reason why `IServiceCollectionProvider` interface exists. Is this the best way to approach this? Why not just pass the `IServiceProvider` to the `ODataMessageReader/Writer` constructor directly?

cc @ElizabethOkerio  @robertmclaws  @gathogojr

ElizabethOkerio: I think this is outside the scope of this PR.

src/Microsoft.OData.Core/ODataMessageReader.cs
```diff
@@ -887,29 +888,29 @@ internal ODataFormat GetFormat()
...
             {
-                if (this.container == null)
+                if (this.serviceProvider == null)
                 {
                     this.messageInfo = new ODataMessageInfo();
                 }
                 else
                 {
-                    this.messageInfo = this.container.GetRequiredService<ODataMessageInfo>();
+                    this.messageInfo = this.serviceProvider.GetRequiredService<ODataMessageInfo>();
```
habbes, line 913: Given that `messageInfo` is modified in this method, actually all of it's properties are modified, there seems to be reason to get this from the service provider.

If the caller provided a custom `ODataMessageInfo`, then we would overwrite all the properties they set, making it pointless (and arguably an anti-pattern).

I think we should remove this class from the service container and consider making it private.

I think customers who have passed a service provider without creating a scope per request have ran into issues because threads from different requests ended up writing to the same message stream. Of course, this service was designed to be scoped, so that's technically a customer error. But my argument is that it shouldn't be in the DI container in the first place.

src/Microsoft.OData.Core/ODataServiceCollectionExtensions.cs
habbes: Just to clarify, this is more of thinking-out-loud, not necessarily requesting a change.

src/Microsoft.OData.Core/ODataMessageReader.cs
habbes: While it may be outside the scope of this PR, would still like to get your thoughts on it. If you think the suggestion (to pass the container to the relevant constructors) is a better approach, then we could create a separate issue for it and possibly a PR for it in ODL 8 (or 9).

My issue with the current approach is that it's not intuitive at all. There's no way for a customer to expect (other than by reading the docs) that for them to pass custom settings to the `ODataMessageWriter`, they have to modify the `IODataRequest/ResponseMessage` to implement `IServiceCollectionProvider`. It also adds maintenance cost because it requires an extra public interface that in my opinion is unnecessary.

If passing the service provider to the constructor is not a favorable approach, the we could consider adding a property to the `ODataMessageReader/WriterSettings`

ElizabethOkerio: This could be a design flaw and needs to be looked into. Let's create a work item for it.

gathogojr: I think @habbes is onto something here but let's create a work item and then we can deliberate when to take it on for ODL 8 or after.

habbes: In order to avoid derailing this PR, I've created a separate issue to track and discuss this: https://github.com/OData/odata.net/issues/2907

src/Microsoft.OData.Core/ODataServiceCollectionExtensions.cs
robertmclaws: This is something that I've been trying to work out in my OData.NEXT thought experiment. Ultimately I think there should be a single Options object with multiple layers of settings that can be manipulated. That presently creates an N+1 problem that doesn't actually solve anything at the moment, so I went with this and optional parameters to give flexibility without creating something that will undoubtedly require a deeper plumbing change.

src/Microsoft.OData.Core/ODataMessageReader.cs
habbes: To avoid derailing this PR. I've created a separate issue to track and discuss this: https://github.com/OData/odata.net/issues/2908

habbes (approved):

ElizabethOkerio:
@robertmclaws Thanks for the reviews and your input in this PR.

PR #2139, opened 2021-07-26
Fix escape function binding type constraint

ificator:
### Issues

Currently the binding type constraint for escape functions comes from `TargetEdmType`, but when the escape function is rooted on a contained navigation the chosen type ends up being the collection instead of the entity.

### Description

Consider the following request (where `items` is a contained navigation from `oneDrive.drive` to `Collection(oneDrive.item)`):
`/drive/items/32:/path`

The expectation is that, if `oneDrive.getByPath` is bound to `oneDrive.item`, the escape function mechanism will translate the path to:
`/drive/items/32/oneDrive.getByPath(path='path')`

However, the current logic is trying to find a function bound to `Collection(oneDrive.item)` and so the escape logic is not translating at all. The fix is to use `EdmType` which appears to be the expected value for all scenarios.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
@@ -1763,7 +1763,7 @@ namespace Microsoft.OData.UriParser
 
             if (previous != null)
             {
-                bindingType = previous.TargetEdmType;
+                bindingType = previous.EdmType;
             }
 
             if (bindingType == null || model == null)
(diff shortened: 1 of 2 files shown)


xuzhg (approved):
:shipit:

gathogojr (approved):
`EdmType` property will be null for some segments (e.g. `BatchSegment`). However, it doesn't appear to me like this change would cause problems there since the scope is limited to binding escape function.

mikepizzo (approved):

PR #2142, opened 2021-07-28
Tests to verify congruence between synchronous and asynchronous APIs exposed by ODataMultipartMixedBatchOutputContext class

gathogojr:
### Issues

Tests to verify congruence between synchronous and asynchronous APIs exposed by `ODataMultipartMixedBatchOutputContext` class

### Description

Synchronous and asynchronous APIs are executed and the results compared to verify that the outcome is the same.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMultipartMixedBatchOutputContextApiTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMultipartMixedBatchOutputContextApiTests.cs
@@ -1,0 +1,1127 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ODataMultipartMixedBatchOutputContextApiTests.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text.RegularExpressions;
+using System.Threading.Tasks;
+using Microsoft.OData.Edm;
+using Xunit;
+
+namespace Microsoft.OData.Tests
+{
+    public class ODataMultipartMixedBatchOutputContextApiTests
+    {
+        private const string ServiceUri = "http://tempuri.org";
+        private readonly MemoryStream asyncStream;
+        private readonly MemoryStream syncStream;
+        private readonly ODataMessageWriterSettings writerSettings;
+        private const string batchBoundary = "batch_aed653ab";
+        private const string batchResponseBoundary = "batchresponse_aed653ab";
+        // Regex used to replace the random batch boundary before equivalence check
+        private const string batchGuidRegex = @"batch[a-z]*_[0-9a-f]{8}-[0-9a-f]{4}-[1-5][0-9a-f]{3}-[89ab][0-9a-f]{3}-[0-9a-f]{12}";
+
+        private EdmModel model;
+        private EdmEntityType orderEntityType;
+        private EdmEntityType customerEntityType;
+        private EdmEntitySet orderEntitySet;
+        private EdmEntitySet customerEntitySet;
+
+        public ODataMultipartMixedBatchOutputContextApiTests()
+        {
+            this.InitializeEdmModel();
+            this.asyncStream = new MemoryStream();
+            this.syncStream = new MemoryStream();
+            this.writerSettings = new ODataMessageWriterSettings
+            {
+                Version = ODataVersion.V4,
+                EnableMessageStreamDisposal = false,
+                BaseUri = new Uri(ServiceUri)
+            };
+
+            this.writerSettings.SetServiceDocumentUri(new Uri(ServiceUri));
+            this.writerSettings.SetContentType(ODataFormat.Batch);
+        }
+
+        [Fact]
+        public async Task WriteMultipartMixedBatchRequest_APIsShouldYieldSameResult()
+        {
+            var orderResource = CreateOrderResource();
+
+            IODataRequestMessage asyncRequestMessage = new InMemoryMessage { Stream = this.asyncStream };
+            using (var messageWriter = new ODataMessageWriter(asyncRequestMessage, this.writerSettings))
+            {
+                var multipartMixedBatchWriter = await messageWriter.CreateODataBatchWriterAsync();
+                await multipartMixedBatchWriter.WriteStartBatchAsync();
+
+                var operationRequestMessage = await multipartMixedBatchWriter.CreateOperationRequestMessageAsync(
+                    "POST", new Uri($"{ServiceUri}/Orders"), "1");
+
+                using (var nestedMessageWriter = new ODataMessageWriter(operationRequestMessage))
+                {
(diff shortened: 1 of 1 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMultipartMixedBatchOutputContextCongruenceTests.cs
```diff
@@ -0,0 +1,935 @@
...
+                return Regex.Replace(result, batchGuidRegex, batchResponseBoundary);
+            });
+
+            Assert.Equal(asyncResult, syncResult);
+        }
+
+        #region Exception Cases
+
+        [Fact]
+        public async Task WriteMultipartMixedBatchRequestWithChangeset_APIsYieldSameResultForContentIdIsNull()
```
KenitoInc, line 725: Am confused,
What are we testing here that is different from this test `WriteMultipartMixedBatchRequest_APIsYieldSameResultForContentIdNull`

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMultipartMixedBatchOutputContextCongruenceTests.cs
```diff
@@ -0,0 +1,935 @@
...
+                    multipartMixedBatchWriter.WriteEndBatch();
+                }
+
+                this.syncStream.Position = 0;
+                var result = new StreamReader(this.syncStream).ReadToEnd();
+
+                return Regex.Replace(result, batchGuidRegex, batchBoundary);
+            });
+
+            Assert.Equal(asyncResult, syncResult);
```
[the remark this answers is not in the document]
gathogojr: @KenitoInc That is correct. Actually these comparison tests are based on existing unit tests added in the previous pull requests

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMultipartMixedBatchOutputContextCongruenceTests.cs
```diff
@@ -0,0 +1,935 @@
...
+        {
+            var orderResource = CreateOrderResource();
+
+            IODataRequestMessage asyncRequestMessage = new InMemoryMessage { Stream = this.asyncStream };
+            using (var messageWriter = new ODataMessageWriter(asyncRequestMessage, this.writerSettings))
+            {
+                var multipartMixedBatchWriter = await messageWriter.CreateODataBatchWriterAsync();
+                await multipartMixedBatchWriter.WriteStartBatchAsync();
+
+                var operationRequestMessage = await multipartMixedBatchWriter.CreateOperationRequestMessageAsync(
```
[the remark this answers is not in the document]
gathogojr: A random guid is generated

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMultipartMixedBatchOutputContextCongruenceTests.cs
gathogojr: Since the generated contentId is a random Guid, the value in the async result will be different from the value in the sync result so we use Regex to find and replace it with a know value before doing the comparison. It's the same thing we're doing with the batch boundary since it's also randomly generated

gathogojr: The different is that we're not writing a changeset in `WriteMultipartMixedBatchRequest_APIsYieldSameResultForContentIdNull`
Which is why the methods are named:
- WriteMultipartMixedBatchRequest_APIsYieldSameResultForContentIdNull
- WriteMultipartMixedBatchRequestWithChangeset_APIsYieldSameResultForContentIdIsNull

KenitoInc: So with the changeset, we throw an exception?

gathogojr: Yes. And the reason for that is, we don't write Content ID into the payload if there's no changeset which means that if it's not supplied we can let it slide. The `dependsOnIds` are only relevant when changesets are in play. Here's how the payload looks when there's no changeset vs when there's a changeset

**When there's no changeset - whether Content ID is supplied or not:**
```
--batch_732c7424-5d10-4b39-b46c-f850bcdc628d
Content-Type: application/http
Content-Transfer-Encoding: binary

POST http://tempuri.org/Orders HTTP/1.1
OData-Version: 4.0
Content-Type: application/json;odata.metadata=minimal;odata.streaming=true;IEEE754Compatible=false;charset=utf-8

{"Id":1,"Amount":13}
--batch_732c7424-5d10-4b39-b46c-f850bcdc628d
Content-Type: application/http
Content-Transfer-Encoding: binary

POST http://tempuri.org/Customers HTTP/1.1
OData-Version: 4.0
Content-Type: application/json;odata.metadata=minimal;odata.streaming=true;IEEE754Compatible=false;charset=utf-8

{"Id":1,"Name":"Customer 1"}
--batch_732c7424-5d10-4b39-b46c-f850bcdc628d--
```

**Where there's a changeset - Content ID is mandatory:**
```
--batch_e20b091a-23d7-4752-9b7c-ce55044cc36d
Content-Type: multipart/mixed; boundary=changeset_ec3a8d4f

--changeset_ec3a8d4f
Content-Type: application/http
Content-Transfer-Encoding: binary
Content-ID: 1

POST http://tempuri.org/Orders HTTP/1.1
OData-Version: 4.0
Content-Type: application/json;odata.metadata=minimal;odata.streaming=true;IEEE754Compatible=false;charset=utf-8

{"Id":1,"Amount":13}
--changeset_ec3a8d4f
Content-Type: application/http
Content-Transfer-Encoding: binary
Content-ID: 2

POST http://tempuri.org/Customers HTTP/1.1
OData-Version: 4.0
Content-Type: application/json;odata.metadata=minimal;odata.streaming=true;IEEE754Compatible=false;charset=utf-8

{"Id":1,"Name":"Customer 1"}
--changeset_ec3a8d4f--
--batch_e20b091a-23d7-4752-9b7c-ce55044cc36d--
```

KenitoInc: Got it. Maybe you can add a comment in the tests to explain this

KenitoInc (approved):
LGTM

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMultipartMixedBatchOutputContextCongruenceTests.cs
gathogojr: Will do

habbes (approved):

marabooy (approved):
Something that would have been great to see would be instead of comparing the two outputs ..comparing the outputs with a human readable string so they can be read during review. 

Assert.Equals(expected, asyncOutput);
Assert.Equals(expected, syncOutput);

PR #2144, opened 2021-07-29
Skip string allocation for keys of PropertyCache

joaocpaiva:
### Issues

2.5% of all allocations in AGS come from concatenating strings to compose keys used to lookup the PropertyCache. This also represents 13.5% of all string allocations for AGS, and it is the second largest hotspot when it comes to allocations according to CPR.

This code is present in the hot path while writing the OData response payload back to the client.

image

### Description

Changing the type of the key from string to a struct that is composed by the relevant properties: depth, name, fullTypeName. Thus avoiding the string allocation to compose the key. Furthermore, struct is a value type and therefore it is only stored in the stack and does not require any heap allocation overhead.

- [X] *Tests already cover this logic*

### DIFF
--- a/src/Microsoft.OData.Core/PropertyCache.cs
+++ b/src/Microsoft.OData.Core/PropertyCache.cs
@@ -8,7 +8,6 @@ using System;
 using System.Collections.Generic;
 using Microsoft.OData.Edm;
 
-
 namespace Microsoft.OData
 {
     /// <summary>
--- a/src/Microsoft.OData.Core/PropertyCache.cs
+++ b/src/Microsoft.OData.Core/PropertyCache.cs
@@ -4,8 +4,10 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
 using Microsoft.OData.Edm;
+
 
 namespace Microsoft.OData
 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/PropertyCache.cs
```diff
@@ -4,25 +4,80 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using System;
 using System.Collections.Generic;
 using Microsoft.OData.Edm;
 
+
```
KenitoInc, line 11: Remove this extra line

KenitoInc (approved):

Sreejithpin (approved):

src/Microsoft.OData.Core/PropertyCache.cs
```diff
@@ -14,15 +15,68 @@ namespace Microsoft.OData
...
+                }
+
+                return false;
+            }
+
+            public bool Equals(PropertyKey other)
+            {
+                return
+                    this.Depth == other.Depth &&
+                    this.Name == other.Name &&
```
chrisspre, line 66: Should this use String.Equals(this.Name, other.Name, StringComparison.InvariantCulture) ?

src/Microsoft.OData.Core/PropertyCache.cs
```diff
@@ -14,15 +15,68 @@ namespace Microsoft.OData
...
+            }
+
+            public static bool operator !=(PropertyKey lhs, PropertyKey rhs)
+            {
+                return !lhs.Equals(rhs);
+            }
+
+            public override int GetHashCode()
+            {
+                return this.Depth.GetHashCode() ^
```
chrisspre, line 47: Not an expert but saw comments that XOR is not the best mixing function. 
Newer SDKs provide System.Hash.Combine methods that use multiplication and addition.

https://source.dot.net/#System.Reflection.Metadata/System/Reflection/Internal/Utilities/Hash.cs,10
https://source.dot.net/#System.Private.CoreLib/HashCode.cs,76
https://docs.microsoft.com/en-us/dotnet/api/system.hashcode.add?view=net-5.0

joaocpaiva: Unfortunately we cannot use that API, need a common solution for all target runtimes. Honestly I have used this hash code pattern and never found any issues. However I would not be surprised if this is not efficient for very large collections, where collisions become more likely.

joaocpaiva: If there is a better hash code implementation that is common to all runtimes, let me know and will change.

src/Microsoft.OData.Core/PropertyCache.cs
joaocpaiva: I tried to preserve the old behavior which is Ordinal. But let me know if this should be Invariant. Though I rather have Ordinal for performance :)

gathogojr (approved):
LG

Issue #2140, opened 2021-07-26
ArgumentNullException when $filter has empty string.


paggarwa08:
The OData Library has a bug: they throw an unhandled ArgumentNullException when an in $filter involves an empty string on the right-hand-side.

This bubbles up as an unhandled exception in service as well, damaging availability of service.

### Assemblies affected

It is reproducible in 7.8.3 version and lower version as well.

### Reproduce steps

Example: /beta/groups?$filter=displayName in ('')

### Expected result

If it has empty  filter then it should  not so any filtering and ignore the empty string.

### Actual result

ODL is throwing Argument null exception.

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

habbes:
Out of curiosity, why should it ignore the empty string? Isn't an empty string a valid string value? What if `displayName` for a particular entity is empty shouldn't `displayName in ('')` return true in that scenario?

PR #2145, opened 2021-07-29
Allow filter in empty string or guid

KenitoInc:
### Issues

*This pull request fixes issue #2140.*

### Description

Support
```
$filter=SSN in ('')   // Edm.String
$filter=SSN in ( ' ' )   // Edm.String
$filter=MyGuid in ( '' )  // Edm.Guid
$filter=MyGuid in ( '  ' ) // Edm.Guid
```

- [ ] *Test cases added*

### Additional work necessary


marabooy:
Maybe this could also work for string like items e.g. dates, date time offsets and such?

cheng93:
I think it should throw if the Edm is _not_ a string after all

'' is not a valid guid
'' is not a valid date
'' is a valid string

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -2375,6 +2375,36 @@ public void FilterWithInOperationWithEmptyCollection(string filterClause)
...
+        [InlineData("SSN in ( ' ' )")]     // Edm.String
+        [InlineData("SSN in ( \" \" )")]     // Edm.String
+        public void FilterWithInOperationWithEmptyString(string filterClause)
+        {
+            FilterClause filter = ParseFilter(filterClause, HardCodedTestModel.TestModel, HardCodedTestModel.GetPersonType());
+
+            var inNode = Assert.IsType<InNode>(filter.Expression);
+
+            CollectionConstantNode collectionNode = Assert.IsType<CollectionConstantNode>(inNode.Right);
+            Assert.Equal(0, collectionNode.Collection.Count);
```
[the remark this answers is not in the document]
habbes: I assume if `SSN` is an empty string then `SSN in ('')` should return true. But if `SSN` is a non-empty string then `SSN in ('')` should return false.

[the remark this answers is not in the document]
habbes: Is `(' ')` an empty string or is it a string with a single whitespace character?

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -331,10 +352,22 @@ private static string NormalizeGuidCollectionItems(string bracketLiteralText)
...
             if (normalizedText.Length == 0)
             {
                 return "[]";
             }
+
+            string normalizedTextWithoutSpaces = normalizedText.Replace(" ", string.Empty);
+
+            // If we have '' or "" or empty strings in the brackets e.g (' ')
+            // We throw an exception since empty strings are not valid guid values.
+            if (normalizedTextWithoutSpaces == "''" || normalizedTextWithoutSpaces == "\"\"")
```
habbes, line 366: Is there a general way for checking for invalid values? Why do we have special treatment for empty strings? I assume there are other non-empty string values which are not valid GUID or dates. Why don't we do the validation at the same place?

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -155,6 +155,15 @@ private static string NormalizeStringCollectionItems(string literalText)
...
             // remove the '[' and ']'
             string normalizedText = literalText.Substring(1, literalText.Length - 2).Trim();
+
+            string normalizedTextWithoutWhitespaces = normalizedText.Replace(" ", string.Empty);
+
+            // If we have '' or "" or empty string we return empty brackets.
+            if (normalizedTextWithoutWhitespaces == "''" || normalizedTextWithoutWhitespaces == "\"\""  || normalizedTextWithoutWhitespaces.Length == 0)
+            {
+                return "[]";
+            }
```
habbes, line 165: Why do we return empty brackets here? Isn't an empty string a valid string value? I assume a collection with a single empty string value is still a non-empty collection. Why are we removing the value from the collection?

habbes (changes requested):

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
chrisspre: I agree with Clement that  and empty string is a string and 
a collection with only an empty string has the same number of elements as a collection with the string "a"

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -331,10 +352,22 @@ private static string NormalizeGuidCollectionItems(string bracketLiteralText)
...
+            // If we have empty brackets ()
             if (normalizedText.Length == 0)
             {
                 return "[]";
             }
+
+            string normalizedTextWithoutSpaces = normalizedText.Replace(" ", string.Empty);
+
+            // If we have '' or "" or empty strings in the brackets e.g (' ')
+            // We throw an exception since empty strings are not valid guid values.
```
chrisspre, line 365: Just a type. This method normalized collections of DateTime, not Guid.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
chrisspre: +1, lines 362 to 370 should be factored in a helper method since this seems to be useful for other NormalizeXYZcollectionItems.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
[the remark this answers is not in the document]
chrisspre: to be precise, it is a collection with one element. The element is a string with a single whitespace. 

see https://github.com/oasis-tcs/odata-abnf/blob/f16f43979a94d7c1e7ff566c51da9e2acf884502/abnf/odata-abnf-construction-rules.txt#L668
listExpr  = OPEN BWS primitiveLiteral BWS *( COMMA BWS primitiveLiteral BWS ) CLOSE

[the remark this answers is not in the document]
KenitoInc: I have fixed the code to handle whitespaces as collection items.
However an empty string throws exception when creating a ConstantNode https://github.com/OData/odata.net/blob/6003304b3bbb58c91d1bd6f6ac9f43d1e50deb82/src/Microsoft.OData.Core/UriParser/SemanticAst/ConstantNode.cs#L53

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -2375,6 +2375,36 @@ public void FilterWithInOperationWithEmptyCollection(string filterClause)
...
+        [InlineData("MyGuid in ( '' )")]  // Edm.Guid
+        [InlineData("MyGuid in ( '  ' )")]  // Edm.Guid
+        [InlineData("MyGuid in ( \" \" )")]  // Edm.Guid
+        [InlineData("Birthdate in ( '' )")]  // Edm.DateTimeOffset
+        [InlineData("Birthdate in ( \" \" )")]  // Edm.DateTimeOffset
+        [InlineData("Birthdate in ('')")]  // Edm.DateTimeOffset
+        [InlineData("MyDate in ( '' )")]  // Edm.Date
+        [InlineData("MyDate in ( \" \" )")]  // Edm.Date
+        [InlineData("MyDate in ('')")]  // Edm.Date
+        public void FilterWithInOperationWithEmptyQuotesThrows(string filterClause)
```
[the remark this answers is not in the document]
KenitoInc: An exception is thrown. I have added tests to show that

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
KenitoInc: @habbes Yes there is a general way to check for invalid values. I have added tests that confirms that

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
KenitoInc: Fixed

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
KenitoInc: I also agree. Unfortunately with can't create a ConstantNode with an empty string. However we can create a ConstantNode with single/multiple whitespaces.
https://github.com/OData/odata.net/blob/6003304b3bbb58c91d1bd6f6ac9f43d1e50deb82/src/Microsoft.OData.Core/UriParser/SemanticAst/ConstantNode.cs#L53

mikepizzo: This is broken.  It should be valid to have an empty string in the collection.

What would break if we made ConstantNode support string.Empty?

name in ('') should be equivalent to name eq '' -- how do we parse something like name eq '' if we can't create an empty string in a constant expression?

KenitoInc:
@mikepizzo @habbes @chrisspre 
I have fixed the issues raised.
The major issue was the validation issue in the `ConstantNode`. It seems there have been alot of missing tests hence some of these things fell through the cracks.
I have tried to add a couple of tests to validate the change.

src/Microsoft.OData.Core/UriParser/SemanticAst/ConstantNode.cs
```diff
@@ -50,7 +50,9 @@ public ConstantNode(object constantValue, string literalText)
         /// <exception cref="System.ArgumentNullException">Throws if the input literalText is null.</exception>
         public ConstantNode(object constantValue, string literalText, IEdmTypeReference typeReference)
         {
-            ExceptionUtils.CheckArgumentStringNotNullOrEmpty(literalText, "literalText");
+            // Commented the validation line below since we should be able to create a 
+            // ConstantNode whose LiteralText is an empty string '' or String.Empty
+            //ExceptionUtils.CheckArgumentStringNotNullOrEmpty(literalText, "literalText");
```
gathogojr, line 55: @KenitoInc Should we not retain the check for null? In addition the method documentation string has `/// <exception cref...`

gathogojr: Plus if we decide to abandon the null/empty string check altogether, I don't think it's necessary to retain the commented lines

KenitoInc: My concern with doing a null check is because `LiteralText` is still allowed to be `null`. Look at the `<summary>` below.
https://github.com/OData/odata.net/blob/871d97aec1fab62dc590d4f4f10ba2bab8901670/src/Microsoft.OData.Core/UriParser/SemanticAst/ConstantNode.cs#L81-L84
Any thoughts @mikepizzo @habbes ?

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
chrisspre: To be very clear, the constructor `ConstantNode(object constantValue, string literalText)` ensures that literalText is not null or empty.
But `new ConstantNode("", "\"\"")` should be valid and seems to be implemented that way.

src/Microsoft.OData.Core/UriParser/SemanticAst/ConstantNode.cs
mikepizzo: I'm okay removing the null check given that we have an overload that omits the literalText and already handle the resulting null case.

As Christof points out, the current behavior to create an empty string constant node is to pass in the literalText "/"/"".  Is the correct fix for an empty string in the In operator to modify the ConstantNode to allow an empty string, or to pass in "/"/"" as the literal text for an empty string from the InBinder?

KenitoInc: I agree with Christof, The correct way is to pass an empty string. I have updated the fix and retained the Null check

chrisspre (approved):

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -277,6 +284,20 @@ private static int ProcessSingleQuotedStringItem(int start, string input, String
                 {
                     if (k + 1 >= length || input[k + 1] != '\'')
                     {
+                        // If prev and next are both single quotes, then it's an empty string.
+                        if (input[k - 1] == '\'')
+                        {
+                            if(k > 2 && input[k - 2] == '\'')
+                            {
+                                // Ignore we have 3 single quotes e.g 'xyz'''
+                                // It means we need to escape the te double quotes to return the result "xyz'"
```
gathogojr, line 293:
```suggestion
                                // It means we need to escape the the double quotes to return the result "xyz'"
```

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -277,6 +284,20 @@ private static int ProcessSingleQuotedStringItem(int start, string input, String
...
+                        {
+                            if(k > 2 && input[k - 2] == '\'')
+                            {
+                                // Ignore we have 3 single quotes e.g 'xyz'''
+                                // It means we need to escape the te double quotes to return the result "xyz'"
+                                continue;
+                            }
+                            // We append \"\" so as to return "\"\"" instead of "".
+                            // This is to avoid passing an empty string to the ConstantNode.
+                            //sb.Append("''");
```
gathogojr, line 298:
```suggestion
```

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -277,6 +284,20 @@ private static int ProcessSingleQuotedStringItem(int start, string input, String
...
+                            sb.Append("\\\"\\\"");
+                        }
                         // match with single qutoe ('), stop it.
```
gathogojr, line 301:
```suggestion
                        // match with single quote ('), stop it.
```

gathogojr (approved):
A few comments otherwise, LGTM

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -232,6 +232,13 @@ private static int ProcessDoubleQuotedStringItem(int start, string input, String
                 char next = input[k];
                 if (next == '"')
                 {
+                    // If prev and next are both double quotes, then it's an empty string.
+                    if (input[k - 1] == '"')
+                    {
+                        // We append \"\" so as to return "\"\"" instead of "".
+                        // This is to avoid passing an empty string to the ConstantNode.
```
habbes, line 239: Curious to know, why can't we update the `ConstantNode` to make it allow empty strings?

habbes: Oh I see, `\"\"` is how empty strings are represented in the `ConstantNode`?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -2375,6 +2385,118 @@ public void FilterWithInOperationWithEmptyCollection(string filterClause)
...
+        [InlineData("SSN in (\"\")")]     // Edm.String
+        [InlineData("SSN in ( \"\" )")]     // Edm.String
+        public void FilterWithInOperationWithEmptyString(string filterClause)
+        {
+            FilterClause filter = ParseFilter(filterClause, HardCodedTestModel.TestModel, HardCodedTestModel.GetPersonType());
+
+            var inNode = Assert.IsType<InNode>(filter.Expression);
+
+            CollectionConstantNode collectionNode = Assert.IsType<CollectionConstantNode>(inNode.Right);
+            Assert.Equal(1, collectionNode.Collection.Count);
```
habbes, line 2400: We don't need to assert that the element of the collection is actually an empty string? It's possible the implementation may contain bugs and we appended something else to the collection :)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -2375,6 +2385,118 @@ public void FilterWithInOperationWithEmptyCollection(string filterClause)
...
+        public void FilterWithInOperationWithWhitespace(string filterClause)
+        {
+            FilterClause filter = ParseFilter(filterClause, HardCodedTestModel.TestModel, HardCodedTestModel.GetPersonType());
+
+            var inNode = Assert.IsType<InNode>(filter.Expression);
+
+            CollectionConstantNode collectionNode = Assert.IsType<CollectionConstantNode>(inNode.Right);
+
+            // A single whitespace or multiple whitespaces are valid literals
+            Assert.Equal(1, collectionNode.Collection.Count);
```
habbes, line 2417: An assertion that checks the expected contents of the collection would inspire more confidence :)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -2375,6 +2385,118 @@ public void FilterWithInOperationWithEmptyCollection(string filterClause)
...
+        public void FilterWithInOperationDateWithEmptyQuotesThrows(string filterClause, string quotedString)
+        {
+            Action parse = () => ParseFilter(filterClause, HardCodedTestModel.TestModel, HardCodedTestModel.GetPersonType());
+            parse.Throws<ODataException>(Strings.ReaderValidationUtils_CannotConvertPrimitiveValue(quotedString, "Edm.Date"));
+        }
+
+        [Theory]
+        [InlineData("('D01663CF-EB21-4A0E-88E0-361C10ACE7FD', '','492CF54A-84C9-490C-A7A4-B5010FAD8104')")]
+        [InlineData("('D01663CF-EB21-4A0E-88E0-361C10ACE7FD', \"\",'492CF54A-84C9-490C-A7A4-B5010FAD8104')")]
+        public void FilterWithInOperationWithQuotedGuidCollectionWithInvalidValues(string guidsCollection)
```
habbes, line 2470:
```suggestion
        public void FilterWithInOperationWithQuotedGuidCollectionWithInvalidValuesThrows(string guidsCollection)

```

habbes (approved):
Looks good. I would suggest adding assertions to verify the contents of the collections to make the tests more reliable. Unless that's already covered by other tests.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
KenitoInc: @habbes Yes, we use `""` instead of `String.Empty`

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
KenitoInc: #Resolved

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
KenitoInc: #Resolved

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
KenitoInc: #Resolved

PR #2154, opened 2021-08-06
Skip lambda allocation in FindTransientAnnotation

joaocpaiva:
### Issues

Total allocation of this method is at 3.09%: 
- 1% for delegate
- 1.61% for enumerator
- 0.48% for closure.

image

### Description

This change replaces the usage of Linq in this method, with manual implementation, avoiding lambda allocation. Every time we invoke Any, it allocates a lambda. This is normally meaningless, when not present in the hot path. Since this is present in AGS hot path, it is costing 1% of allocations just in this lambda. This is a lot.

Also removes the usage of enumerator, replacing with O(1) random access for loop. The ArrayVersioningList is better for search scenarios because we can save the enumerator allocation all together, and use O(1) random access. The LinkedVersioningList would only be useful if this data structure is changing very often at runtime, which I don't think is the case? Main scenario will be search at runtime? So I am proposing to replace with ArrayVersioningList. Furthermore, when adding a new element, it was swapping to ArrayVersioningList if there were 5 or more elements, hence I don't really see the benefit of not using ArrayVersioningList for all cases, particular if the search scenario is the main path.

Not using Linq.Any(), is not only good for memory but also for CPU as we can see in this benchmark:

image

- [X] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Edm/VersioningList.cs
+++ b/src/Microsoft.OData.Edm/VersioningList.cs
@@ -14,7 +14,7 @@ namespace Microsoft.OData.Edm
     /// "Mutating" operations return a new list (which, for efficiency, may share some of the state of the old one).
     /// </summary>
     /// <typeparam name="TElement">Element type of the list.</typeparam>
-    internal abstract class VersioningList<TElement> : IEnumerable<TElement>
+    internal abstract class VersioningList<TElement> : IReadOnlyList<TElement>, IEnumerable<TElement>
     {
         public abstract int Count { get; }
 
@@ -68,7 +68,7 @@ namespace Microsoft.OData.Edm
 
             public override VersioningList<TElement> Add(TElement value)
             {
-                return new LinkedVersioningList(this, value);
+                return new ArrayVersioningList(this, value);
             }
 
             public override IEnumerator<TElement> GetEnumerator()
@@ -113,140 +113,6 @@ namespace Microsoft.OData.Edm
             }
         }
 
-        internal sealed class LinkedVersioningList : VersioningList<TElement>
-        {
-            private readonly VersioningList<TElement> preceding;
-            private readonly TElement last;
-
-            public LinkedVersioningList(VersioningList<TElement> preceding, TElement last)
-            {
-                this.preceding = preceding;
-                this.last = last;
-            }
-
-            public override int Count
-            {
-                get { return this.preceding.Count + 1; }
-            }
-
-            public VersioningList<TElement> Preceding
-            {
-                get { return this.preceding; }
-            }
-
-            public TElement Last
-            {
-                get { return this.last; }
-            }
-
-            private int Depth
-            {
-                get
-                {
-                    int depth = 0;
-                    LinkedVersioningList layer = this;
-                    while (layer != null)
-                    {
-                        depth++;
-                        layer = layer.Preceding as LinkedVersioningList;
-                    }
-
-                    return depth;
-                }
-            }
-
-            public override VersioningList<TElement> Add(TElement value)
-            {
-                if (this.Depth < 5)
-                {
-                    return new LinkedVersioningList(this, value);
-                }
-
-                return new ArrayVersioningList(this, value);
-            }
-
-            public override IEnumerator<TElement> GetEnumerator()
-            {
-                return new LinkedListEnumerator(this);
-            }
-
-            protected override TElement IndexedElement(int index)
-            {
-                if (index == this.Count - 1)
-                {
-                    return this.last;
-                }
-
-                return this.preceding.IndexedElement(index);
-            }
-
-            protected override VersioningList<TElement> RemoveIndexedElement(int index)
-            {
-                if (index == this.Count - 1)
-                {
-                    return this.preceding;
-                }
-
-                return new LinkedVersioningList(this.preceding.RemoveIndexedElement(index), this.last);
-            }
-        }
-
-        internal sealed class LinkedListEnumerator : IEnumerator<TElement>
-        {
-            private readonly LinkedVersioningList list;
-            private IEnumerator<TElement> preceding;
-            private bool complete;
-
-            public LinkedListEnumerator(LinkedVersioningList list)
-            {
-                this.list = list;
-                this.preceding = list.Preceding.GetEnumerator();
-            }
-
-            public TElement Current
-            {
-                get
-                {
-                    if (this.complete)
-                    {
-                        return this.list.Last;
-                    }
-
-                    return this.preceding.Current;
-                }
-            }
-
-            object System.Collections.IEnumerator.Current
-            {
-                get { return this.Current; }
-            }
-
-            public void Dispose()
-            {
-            }
-
-            public bool MoveNext()
-            {
-                if (this.complete)
-                {
-                    return false;
-                }
-
-                if (!this.preceding.MoveNext())
-                {
-                    this.complete = true;
-                }
-
-                return true;
-            }
-
-            public void Reset()
-            {
-                this.preceding.Reset();
-                this.complete = false;
-            }
-        }
-
         internal sealed class ArrayVersioningList : VersioningList<TElement>
         {
             private readonly TElement[] elements;
@@ -254,13 +120,12 @@ namespace Microsoft.OData.Edm
             public ArrayVersioningList(VersioningList<TElement> preceding, TElement last)
             {
                 this.elements = new TElement[preceding.Count + 1];
-                int index = 0;
-                foreach (TElement element in preceding)
+                for (int i = 0; i < preceding.Count; i++)
                 {
-                    this.elements[index++] = element;
+                    this.elements[i] = preceding[i];
                 }
 
-                this.elements[index] = last;
+                this.elements[preceding.Count] = last;
             }
 
             private ArrayVersioningList(TElement[] elements)
@@ -280,7 +145,7 @@ namespace Microsoft.OData.Edm
 
             public override VersioningList<TElement> Add(TElement value)
             {
-                return new LinkedVersioningList(this, value);
+                return new ArrayVersioningList(this, value);
             }
 
             public override IEnumerator<TElement> GetEnumerator()
--- a/src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmDirectValueAnnotationsManager.cs
+++ b/src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmDirectValueAnnotationsManager.cs
@@ -26,7 +26,7 @@ namespace Microsoft.OData.Edm.Vocabularies
         /// <summary>
         /// Keeps track of transient annotations on elements.
         /// </summary>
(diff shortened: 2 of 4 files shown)


src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmDirectValueAnnotationsManager.cs
```diff
@@ -173,9 +173,13 @@ private static void SetAnnotation(IEnumerable<IEdmDirectValueAnnotation> immutab
             bool needTombstone = false;
             if (immutableAnnotations != null)
             {
-                if (immutableAnnotations.Any(existingAnnotation => existingAnnotation.NamespaceUri == namespaceName && existingAnnotation.Name == localName))
+                foreach (var existingAnnotation in immutableAnnotations)
```
chrisspre, line 176: I don't immediately see how this code is leading to fewer allocations.  Can you please explain.

joaocpaiva: Every time we invoke Any, it allocates a lambda. This is normally meaningless, when not present in the hot path. Since this is present in AGS hot path, it is costing 1% of allocations just in this lambda ...This is a lot.

Not using Linq.Any(), is not only good for memory but also for CPU as you can see in this benchmark:

image

joaocpaiva: I strive for readability and less code verbosity, as much as possible. But when it comes to choose between readability versus performance and COGs, I think we all agree with the choice.

IMHO, I don't think this code hurts readability that much, however it is more verbose I agree. But that is a non-goal when addressing performance issues of code that runs at scale.

In fact, we should encourage developers to stop using Linq for code that runs at scale - not the other way around.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmDirectValueAnnotationsManager.cs
```diff
@@ -245,8 +249,14 @@ private static IEdmDirectValueAnnotation FindTransientAnnotation(object transien
                 else
                 {
                     VersioningList<IEdmDirectValueAnnotation> annotationsList = (VersioningList<IEdmDirectValueAnnotation>)transientAnnotations;
-                    return annotationsList.FirstOrDefault(
-                        existingAnnotation => existingAnnotation.NamespaceUri == namespaceName && existingAnnotation.Name == localName);
+
```
chrisspre, line 252: It hurts me a bit to see that it makes such a change and sacrificing readability. But it is worth the savings.
Can we please add a comment about the foreach stating that for the purpose of optimization, this is not written using Enumerable.Any since this allocates a closure every time.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmDirectValueAnnotationsManager.cs
chrisspre: Thanks JP. And I fully support your change. Want to learn more about the tool you used to analyze this.

But just for the record (and not reverting on my support for this)), no in the absolut way you state "choose between readability versus performance" I do actually disagree. It is comparing the easily measurable cost of performance to the much harder to measure and long term changing cost of maintenance. If we would (could) measure this, the winner is not that obvious anymore.

chrisspre (approved):

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmDirectValueAnnotationsManager.cs
```diff
@@ -277,7 +287,7 @@ private static void RemoveTransientAnnotation(ref object transientAnnotations, s
                             annotationsList = annotationsList.RemoveAt(index);
                             if (annotationsList.Count == 1)
```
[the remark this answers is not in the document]
joaocpaiva: Indeed, this is overhead that can be removed and simplified.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmDirectValueAnnotationsManager.cs
joaocpaiva: I don't think adding a comment is really warranted IMHO, otherwise we would need to add such comments everywhere? :) I mean, it boils down to simple coding best practices that should be followed in the code base, like avoiding Linq in the hot path. I thought many of these best practices were already being adopted, as I have reviewed multiple PRs in the past that were specifically to fix that in OData? Granted, ideally they should be written down somewhere, not sure that is the case.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmDirectValueAnnotationsManager.cs
```diff
@@ -230,96 +219,60 @@ private static void SetAnnotation(IEnumerable<IEdmDirectValueAnnotation> immutab
...
         }
 
-        private static IEdmDirectValueAnnotation FindTransientAnnotation(object transientAnnotations, string namespaceName, string localName)
+        private static IEdmDirectValueAnnotation FindTransientAnnotation(VersioningList<IEdmDirectValueAnnotation> transientAnnotations, string namespaceName, string localName)
         {
             if (transientAnnotations != null)
             {
-                IEdmDirectValueAnnotation singleAnnotation = transientAnnotations as IEdmDirectValueAnnotation;
-                if (singleAnnotation != null)
+                VersioningList<IEdmDirectValueAnnotation> annotationsList = transientAnnotations;
```
gathogojr, line 226: Do we need to declare `annotationsList` yet we can loop over `transientAnnotations`?

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmDirectValueAnnotationsManager.cs
```diff
@@ -230,96 +219,60 @@ private static void SetAnnotation(IEnumerable<IEdmDirectValueAnnotation> immutab
...
         }
 
-        private static void RemoveTransientAnnotation(ref object transientAnnotations, string namespaceName, string localName)
+        private static void RemoveTransientAnnotation(ref VersioningList<IEdmDirectValueAnnotation> transientAnnotations, string namespaceName, string localName)
         {
             if (transientAnnotations != null)
             {
-                IEdmDirectValueAnnotation singleAnnotation = transientAnnotations as IEdmDirectValueAnnotation;
-                if (singleAnnotation != null)
+                VersioningList<IEdmDirectValueAnnotation> annotationsList = transientAnnotations;
```
gathogojr, line 244: Refer to https://github.com/OData/odata.net/pull/2154/files#r684926784

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmDirectValueAnnotationsManager.cs
```diff
@@ -230,96 +219,60 @@ private static void SetAnnotation(IEnumerable<IEdmDirectValueAnnotation> immutab
...
-                {
-                    yield return singleAnnotation;
-                }
-
-                yield break;
-            }
-
-            VersioningList<IEdmDirectValueAnnotation> annotationsList = (VersioningList<IEdmDirectValueAnnotation>)transientAnnotations;
-            foreach (IEdmDirectValueAnnotation existingAnnotation in annotationsList)
+            VersioningList<IEdmDirectValueAnnotation> annotationsList = transientAnnotations;
```
gathogojr, line 264: Refer to https://github.com/OData/odata.net/pull/2154/files#r684926784

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmDirectValueAnnotationsManager.cs
```diff
@@ -197,26 +201,11 @@ private static void SetAnnotation(IEnumerable<IEdmDirectValueAnnotation> immutab
...
-                else
-                {
-                    transientAnnotations = VersioningList<IEdmDirectValueAnnotation>.Create().Add(singleAnnotation).Add(newAnnotation);
-                }
-
-                return;
-            }
-
-            VersioningList<IEdmDirectValueAnnotation> annotationsList = (VersioningList<IEdmDirectValueAnnotation>)transientAnnotations;
+            VersioningList<IEdmDirectValueAnnotation> annotationsList = transientAnnotations;
```
gathogojr, line 208: Refer to https://github.com/OData/odata.net/pull/2154/files#r684926784

marabooy (approved):

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmDirectValueAnnotationsManager.cs
KenitoInc: We discourage use of `var`. You can use the type.

KenitoInc (changes requested):

gathogojr (approved):
LG

KenitoInc (approved):

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmDirectValueAnnotationsManager.cs
```diff
@@ -197,129 +201,77 @@ private static void SetAnnotation(IEnumerable<IEdmDirectValueAnnotation> immutab
...
         }
 
-        private static IEdmDirectValueAnnotation FindTransientAnnotation(object transientAnnotations, string namespaceName, string localName)
+        private static IEdmDirectValueAnnotation FindTransientAnnotation(VersioningList<IEdmDirectValueAnnotation> transientAnnotations, string namespaceName, string localName)
         {
             if (transientAnnotations != null)
             {
-                IEdmDirectValueAnnotation singleAnnotation = transientAnnotations as IEdmDirectValueAnnotation;
-                if (singleAnnotation != null)
+                // This method runs very hot in AGS, it is important to optimize it as much as possible.
```
mikepizzo, line 225: >This method runs very hot in AGS [](http://example.com/codeflow?start=19&length=32)

We should remove an product-specific references from comments.  Maybe replace with something like "This method runs very hot in schemas where direct value annotations are used."

src/Microsoft.OData.Edm/VersioningList.cs
```diff
@@ -68,7 +68,7 @@ public override int Count
 
             public override VersioningList<TElement> Add(TElement value)
             {
-                return new LinkedVersioningList(this, value);
+                return new ArrayVersioningList(this, value);
```
habbes, line 71: If I understand the implementation, this results in a new array being allocated, and elements copied from the previous array to the new one. Doesn't this lead to a larger memory footprint and more GC over time? Is the rationale that we're willing to sacrifice memory in order to get faster search times?

PR #2155, opened 2021-08-06
Fix AggregateException thrown by OData writers async API

gathogojr:
### Issues

Fix `AggregateException` thrown by OData writers async API

### Description

As indicated in [this TODO](https://github.com/OData/odata.net/blob/d49b8b9d9a77816b2cf6664be6493ae71906fc0b/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightWriterTests.cs#L1703), there was a scenario causing an `AggregateException` to be thrown instead of the expected `ODataException`. This PR fixes that by avoiding the use of `TaskUtils.FollowSuccessWithTask` that was responsible for that behaviour.

In the process added a bunch of tests to verify behaviour in scenarios where exceptions are thrown.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
+++ b/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
@@ -190,8 +190,9 @@ namespace Microsoft.OData
         {
             this.VerifyCanWriteEndBatch(false);
             await this.WriteEndBatchImplementationAsync()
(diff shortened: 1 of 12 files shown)


marabooy (approved):

KenitoInc (approved):
LGTM

habbes (approved):

Issue #2147, opened 2021-08-02
Performance - Unnecessary Memory Allocation 


Sreejithpin:
From Feng Yuan 
Found a funny allocation stack in RestAppPool:
 
 
 
Source code:
 
        internal static bool IsQualifiedName(string name)
        {
            string[] nameTokens = name.Split('.');
            if (nameTokens.Length < 2)
            {
                return false;
            }
 
            foreach (string token in nameTokens)
            {
                if (IsNullOrWhiteSpaceInternal(token))
                {
                    return false;
                }
            }
 
            return true;
        }
 
        internal static bool IsNullOrWhiteSpaceInternal(String value)
        {
            return value == null || value.ToCharArray().All(Char.IsWhiteSpace);
        }
 
This can be changed to:
 
       internal static bool IsQualifiedName(string name)
        {
            bool allWhitespace = true;
 
            bool hasDot = false;
 
            foreach (char ch in name)
            {
                if (ch == '.')
                {
                    if (allWhitespace)
                    {
                        return false;
                    }
 
                    hasDot = true;
                    allWhitespace = true;
                }
                else if (!char.IsWhiteSpace(ch))
                {
                    allWhitespace = false;
                }
            }
 
            return hasDot;
        }
 
This version is allocation-free and should be 10 faster.

### Assemblies affected

 OData .Net lib 7.x

### Reproduce steps

*The simplest set of steps to reproduce the issue. If possible, reference a commit that demonstrates the issue.*

### Expected result

*What would happen if there wasn't a bug.*

### Actual result
Less performance

### Additional detail

*Optional, details of the root cause if known. Delete this section if you have no additional details to add.*

Sreejithpin:
From: Mike Holcomb <mikeho@microsoft.com> 

The two tests don’t return the same results for the same data.

The original will return false on “.com”, and the improved version will return true on “.com”.

I think this could be fixed with the addition of a variable to track if non-whitespace were before the “.”.

PR #2156, opened 2021-08-11
Fix string allocation performance issue for IsQualifiedName

Sreejithpin:
### Issues

This pull request fixes issue #2147 .

### Description

Fixing string allocation and perf isue

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/EdmUtil.cs
+++ b/src/Microsoft.OData.Edm/EdmUtil.cs
@@ -294,21 +294,28 @@ namespace Microsoft.OData.Edm
         // This is testing if the name can be parsed and serialized, not if it is valid.
         internal static bool IsQualifiedName(string name)
         {
-            string[] nameTokens = name.Split('.');
-            if (nameTokens.Length < 2)
-            {
-                return false;
-            }
+            bool isAllWhiteSpace = true;
+            bool hasDot = false;
 
-            foreach (string token in nameTokens)
+            foreach (char ch in name)
             {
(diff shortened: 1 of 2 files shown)


mikepizzo (approved):
:shipit:

src/Microsoft.OData.Edm/EdmUtil.cs
```diff
@@ -294,21 +294,28 @@ internal static String JoinInternal<T>(String separator, IEnumerable<T> values)
...
-            string[] nameTokens = name.Split('.');
-            if (nameTokens.Length < 2)
-            {
-                return false;
-            }
+            bool isAllWhiteSpace = true;
+            bool hasDot = false;
 
-            foreach (string token in nameTokens)
+            foreach (char ch in name)
```
chrisspre, line 300: There might be a chance to bail out earlier if a segment is all whitespace
```chsharp
        public static bool IsQualifiedName3(string name)
        {
            bool foundDot = false;
            bool previousSegmentAllWhitespace = true;
            foreach (char ch in name)
            {
                if (ch == '.')
                {
                    foundDot = true;
                    if (previousSegmentAllWhitespace)
                    {
                        return false;
                    }
                    previousSegmentAllWhitespace = true;
                }
                else if (!char.IsWhiteSpace(ch))
                {
                    previousSegmentAllWhitespace = false;
                }
            }

            return foundDot & !previousSegmentAllWhitespace;
        }
```

chrisspre (approved):

Issue #2060, opened 2021-04-21
Add method to retrieve all NavigationProperyBindings.


chrisspre:
For deep insert scenarios and potentially others it is important for a service to inspect all navigation property bindings.
There is currently no convenient method that allows to do that.

Suggestion is to add a method to IEdmEntityContainer (probably an extension method) that effectively returns all NavigationProperyBindings

```csharp
public static IEnumerable<(IEdmEntityContainerElement, IEdmNavigationPropertyBinding)> FindBindings(this IEdmEntityContainer container);
```

There might be better representations of the return type. For example a collection of key/value pairs with the binding paths (an IEdmPathExpression?)  as the key and the IEdmNavigationPropertyBinding as the value. 

### Assemblies affected

"Microsoft.OData.Edm" Version="7.8.3"

add.*

xuzhg:
@chrisspre IEdmNavigationSource interface has the property that retrieves all NPB for that navigation source. See https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Edm/Schema/Interfaces/IEdmNavigationSource.cs#L50

If you want to return all NPB in an entity set container, you can enumerate the navigation source (entityset & singleton) and returns the NPB for each navigation source.

chrisspre:
Correct. This is possible to implement . But it is not obvious code. 

One needs to iterate over all Container elements 
```
foreach(var element in model.EntityContainer.Elements)
```
And then these elements need to be tested if they are EntitySet or Singleton.
There is as far as I know there is no interface that provides NavigationPropertyBindings  property for these two to code woudl look like:
```
                switch (element)
                {
                    case IEdmSingleton singleton:
                         // do something with singleton.NavigationPropertyBindings  ;
                    case IEdmEntitySet entitySet:
                         // do something with entitySet.NavigationPropertyBindings 
                    default:
                        continue;
                }
```

I would like to see this coming with the framework, in a productized. maintained, tested way .

KenitoInc:
Resolved by #2157

PR #2157, opened 2021-08-12
method to get all navigation property bindings

chrisspre:
Add an officially supported method to get all navigation property bindings.

Retrieving all navigation property bindings requires quite some knowledge of the edm model implementation and casting.
This method encapsulated this logic. 

### Issues

addresses #2060

### DIFF
--- a/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
+++ b/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
@@ -1646,6 +1646,39 @@ namespace Microsoft.OData.Edm
             return container.AllElements().OfType<IEdmOperationImport>();
         }
 
+        /// <summary>
+        /// Returns NavigationPropertyBindings with their corresponding container elements (EntitySet or Singelton)
+        /// </summary>
+        /// <param name="container">Reference to the calling object.</param>
+        /// <returns>collection of pairs of container element and NavigationPropertyBindings.</returns>
+        public static IEnumerable<(IEdmEntityContainerElement, IEdmNavigationPropertyBinding)> GetNavigationPropertyBindings(this IEdmEntityContainer container)
+        {
+            EdmUtil.CheckArgumentNull(container, "container");
+
+            var elements = container.AllElements();
+            foreach(var element in elements)
+            {
+                switch (element)
+                {
+                    case IEdmEntitySet entitySet:
+                        foreach(var binding in entitySet.NavigationPropertyBindings)
+                        {
+                            yield return (entitySet, binding);
+                        }
+                        break;
+                    case IEdmSingleton singleton:
+                        foreach (var binding in singleton.NavigationPropertyBindings)
+                        {
+                            yield return (singleton, binding);
+                        }
+                        break;
+                    default:
+                        break;
+                }
+            }
+        }
+
+
         #endregion
 
         #region IEdmTypeReference
--- a/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
+++ b/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
@@ -1651,7 +1651,7 @@ namespace Microsoft.OData.Edm
         /// </summary>
         /// <param name="container">Reference to the calling object.</param>
         /// <returns>collection of pairs of container element and NavigationPropertyBindings.</returns>
-        public static IEnumerable<(IEdmEntityContainerElement, IEdmNavigationPropertyBinding)> GetNavigationPropertyBindings(this IEdmEntityContainer container)
+        public static IEnumerable<Tuple<IEdmEntityContainerElement, IEdmNavigationPropertyBinding>> GetNavigationPropertyBindings(this IEdmEntityContainer container)
         {
             EdmUtil.CheckArgumentNull(container, "container");
 
@@ -1663,13 +1663,13 @@ namespace Microsoft.OData.Edm
                     case IEdmEntitySet entitySet:
                         foreach(var binding in entitySet.NavigationPropertyBindings)
                         {
-                            yield return (entitySet, binding);
+                            yield return Tuple.Create<IEdmEntityContainerElement, IEdmNavigationPropertyBinding>(entitySet, binding);
                         }
                         break;
                     case IEdmSingleton singleton:
                         foreach (var binding in singleton.NavigationPropertyBindings)
                         {
-                            yield return (singleton, binding);
+                            yield return Tuple.Create< IEdmEntityContainerElement, IEdmNavigationPropertyBinding>(singleton, binding);
                         }
                         break;
                     default:
--- a/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
+++ b/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
@@ -1647,10 +1647,10 @@ namespace Microsoft.OData.Edm
         }
 
         /// <summary>
-        /// Returns NavigationPropertyBindings with their corresponding container elements (EntitySet or Singelton)
+        /// Returns navigation property bindings with their corresponding container elements (entity set or singleton)
         /// </summary>
         /// <param name="container">Reference to the calling object.</param>
-        /// <returns>collection of pairs of container element and NavigationPropertyBindings.</returns>
+        /// <returns>Collection of pairs of container element and navigation property bindings.</returns>
         public static IEnumerable<Tuple<IEdmEntityContainerElement, IEdmNavigationPropertyBinding>> GetNavigationPropertyBindings(this IEdmEntityContainer container)
         {
             EdmUtil.CheckArgumentNull(container, "container");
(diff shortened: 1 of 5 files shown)


mikepizzo:
You'll need to update the publicapi baseline for the new method.

mikepizzo:
Is the common pattern to get all nav prop bindings for the entire container, or to get the nav prop bindings for a particular entityset/singleton?

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1545,6 +1545,39 @@ public static IEnumerable<IEdmOperationImport> OperationImports(this IEdmEntityC
...
+        /// <summary>
+        /// Returns NavigationPropertyBindings with their corresponding container elements (EntitySet or Singelton)
+        /// </summary>
+        /// <param name="container">Reference to the calling object.</param>
+        /// <returns>collection of pairs of container element and NavigationPropertyBindings.</returns>
+        public static IEnumerable<(IEdmEntityContainerElement, IEdmNavigationPropertyBinding)> GetNavigationPropertyBindings(this IEdmEntityContainer container)
+        {
+            EdmUtil.CheckArgumentNull(container, "container");
+
+            var elements = container.AllElements();
```
mikepizzo, line 1557: >var [](http://example.com/codeflow?start=12&length=3)

We avoid the use of "var" in the runtime code.

chrisspre:
> Is the common pattern to get all nav prop bindings for the entire container, or to get the nav prop bindings for a particular entityset/singleton?

In my experience, yes. For example, when Graph needs to decide if and how deep insert is possible, they are building a sort of map where to "find" an entity of a given type. This came up in a bug a few month back and I believe it would be good to encapsulate and support a way to do this without understanding the full interface hierarchy of EDM. 

More info in the associated issue#2060

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1545,6 +1545,39 @@ public static IEnumerable<IEdmOperationImport> OperationImports(this IEdmEntityC
             return container.AllElements().OfType<IEdmOperationImport>();
         }
 
+        /// <summary>
+        /// Returns NavigationPropertyBindings with their corresponding container elements (EntitySet or Singelton)
+        /// </summary>
+        /// <param name="container">Reference to the calling object.</param>
+        /// <returns>collection of pairs of container element and NavigationPropertyBindings.</returns>
+        public static IEnumerable<(IEdmEntityContainerElement, IEdmNavigationPropertyBinding)> GetNavigationPropertyBindings(this IEdmEntityContainer container)
```
corranrogue9, line 1553: Will the caller be interested in the specific kind of `IEdmEntityContainerElement` returned for each element? We are already casting inside this method, it would be a shame for the caller to have to do so as well if we can avoid it

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1545,6 +1545,39 @@ public static IEnumerable<IEdmOperationImport> OperationImports(this IEdmEntityC
...
+                switch (element)
+                {
+                    case IEdmEntitySet entitySet:
+                        foreach(var binding in entitySet.NavigationPropertyBindings)
+                        {
+                            yield return (entitySet, binding);
+                        }
+                        break;
+                    case IEdmSingleton singleton:
+                        foreach (var binding in singleton.NavigationPropertyBindings)
```
corranrogue9, line 1569: It's a more fundamental change, but I wonder if it would make sense for `IEdmEntitySet` and `IEdmSingleton` to derive from a common interface that has `NavigationPropertyBindings` on it. Then this whole method is sort of obsolete (or at least very derivative):
```
return container.AllElements().OfType<IEdmWithNavPropertyBindings>().SelectMany(element => element.NavigationPropertyBindings.Select(binding => (element, binding)));
```

corranrogue9 (dismissed):

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
chrisspre: it is true that a user would need to cast the first element of the pair if they are interested in more than what IEdmEntityContainerElement provides (which is the name, container, kind and a few extension methods).
But I doubt that this is common since this API is specifically about all bindings, independent of the type of "bindable" container element.

gathogojr, line 1553:
```suggestion
        public static IEnumerable<Tuple<IEdmEntityContainerElement, IEdmNavigationPropertyBinding>> GetNavigationPropertyBindings(this IEdmEntityContainer container)
```
Concise syntax not supported in C# 7.0 - that's why the tests are failing

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1545,6 +1545,39 @@ public static IEnumerable<IEdmOperationImport> OperationImports(this IEdmEntityC
...
+
+            var elements = container.AllElements();
+            foreach(var element in elements)
+            {
+                switch (element)
+                {
+                    case IEdmEntitySet entitySet:
+                        foreach(var binding in entitySet.NavigationPropertyBindings)
+                        {
+                            yield return (entitySet, binding);
```
gathogojr, line 1565:
```suggestion
                            yield return Tuple.Create(entitySet, binding);
```

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1545,6 +1545,39 @@ public static IEnumerable<IEdmOperationImport> OperationImports(this IEdmEntityC
...
+                        }
+                        break;
+                    case IEdmSingleton singleton:
+                        foreach (var binding in singleton.NavigationPropertyBindings)
+                        {
+                            yield return (singleton, binding);
```
gathogojr, line 1571:
```suggestion
                            yield return Tuple.Create(singleton, binding);
```

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/Semantics/CsdlSemanticsEntityContainerTests.cs
```diff
@@ -70,5 +70,28 @@ public void EnsureActionImportActionPropertyIsUnresolvedAction()
...
+            var singleton1 = new CsdlSingleton("Singleton", "unknown", new[] {
+                new CsdlNavigationPropertyBinding("foo", "bar", testLocation)
+            }, testLocation);
+            var csdlEntityContainer = CsdlBuilder.EntityContainer("Container", entitySets: new [] { entitySet1 }, singletons: new[] { singleton1 } );
+            var schema = CsdlBuilder.Schema("FQ.NS", csdlEntityContainers: new CsdlEntityContainer[] { csdlEntityContainer });
+            var csdlModel = new CsdlModel();
+            csdlModel.AddSchema(schema);
+            
+            // act
+            var container = (IEdmEntityContainer)csdlEntityContainer;
```
gathogojr, line 90: You cannot cast a `CsdlEntityContainer` into an `IEdmEntityContainer`...

gathogojr: Test fixed

KenitoInc (dismissed):

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/Semantics/CsdlSemanticsEntityContainerTests.cs
```diff
@@ -70,5 +70,35 @@ public void EnsureActionImportActionPropertyIsUnresolvedAction()
...
+                    new CsdlSemanticsModel(
+                        csdlModel,
+                        new EdmDirectValueAnnotationsManager(),
+                        Enumerable.Empty<IEdmModel>()),
+                    schema),
+                csdlEntityContainer);
+            var bindings = container.GetNavigationPropertyBindings().ToList();
+
+            // assert
+            Assert.Equal(2, bindings.Count);
```
habbes, line 101: I think we should also assert the actual bindings are what's expected?

KenitoInc (approved):

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/Semantics/CsdlSemanticsEntityContainerTests.cs
gathogojr: @habbes Added requested asserts

KenitoInc (approved):

habbes (approved):

gathogojr (approved):

PR #2158, opened 2021-08-12
Avoid action allocations calling InterceptException

joaocpaiva:
### Issues
2% of allocations in AGS are result of the InterceptException pattern.

image

### Description

Every time InterceptException() is invoked, allocation is done to create an Action and capture state from the enclosing context. 

One way to fix such unnecessary allocations is to have the compiler optimizing the underlying code, however it requires using anonymous functions that don't capture state from the enclosing context, so the compiler optimizes the code to avoid creating allocations on every call to this method.

[Compiler generated code, capturing enclosing context](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLRIMYHsEBNXIwJwYzIA+AAgEwCMAsAFCUAMABJbQHQBKcAdjACWAWwhcAwlhEAHIQBsICAMpKAbkIwoA3EyaUAzB2psAsgE8J8qMmRsA3kzbO2TlzIRC1sCBwAsbACigkrK2DIQABRCgmz8EADuyjA+AJRuzo6MLjlsMAAWQshcAJIhCFoyMIEAHpXCWPyRkalsALwAfHmFxQBqUPJCuD4AKghQ/MhCDU3xSSkwEKmputkuAL56a84eXj7+bP2Dw4tjE1Mz0bFC6dsOGWybdw8A2qYQBVi4JbLyke+fb6/ADyVSEjWKAEEAObQpC2LwQMqDfgxaGpAC6D123kWBzKiwqECqtXq4KanBMUFI5NuOSyuRcRHMD3prMZbGplxW7OcTw5bAwsAw+V59zuAoKCCwCVWAv5Gy26yAA=)

[Compiler generated code, without enclosing context](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLRIMYHsEBNXIwJwYzIA+AAgEwCMAsAFCUAMABJbQHQBKcAdjACWAWwhcAwlhEAHIQBsICAMpKAbkIwoA3EyaUAzB2psAsgE8J8qMmRsA3kzbO2TlzIRC1sCBwAsbACigkrK2DIQABRCgmz8EADuyjA+AJRuzo6MLjlsMAAWQshcAJIhCFoyMIEAHpXCWPyRkcgQ8mAANHGJyT4AClAIUCKpbAC8AHxsre1cAGpQ8kK4PgAqQ/zIQg1N8UkpMBADQyNde72HqbrZLgC+ejfOHl4+/mwLSyuH61Cb20KNaKxITpR4ODJse5giHPbyHN5lQ4VCBVWr1AH8AA8qwAgggAOYsCaRTjUTEWKw2ZBdXEEolsKCkDE0vGEhl00E5LK5FxEcwQrkCnkMpmAgpFLqDQlXIXOKHCtgYWAYfKy8FghUFBBYBLXBXyu4PW5AA=)

In the second example, the compiler is able to optimize the code by caching the action delegate, so subsequent calls don't result in unnecessary allocations.

- [X] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
+++ b/src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
@@ -499,7 +499,7 @@ namespace Microsoft.OData
         /// <param name="newState">The writer state to transition into.</param>
         protected void SetState(BatchWriterState newState)
         {
-            this.InterceptException(() => this.ValidateTransition(newState));
+            this.InterceptException((thisParam, newStateParam) => thisParam.ValidateTransition(newStateParam), newState);
 
             this.state = newState;
         }
@@ -669,11 +669,43 @@ namespace Microsoft.OData
         /// state ExceptionThrown and then re-throw the exception.
         /// </summary>
         /// <param name="action">The action to execute.</param>
-        private void InterceptException(Action action)
+        /// <remarks>
+        /// Make sure to only use anonymous functions that don't capture state from the enclosing context, 
+        /// so the compiler optimizes the code to avoid creating allocations on every call to this method.
+        /// </remarks>
+        private void InterceptException(Action<ODataBatchWriter> action)
         {
             try
             {
-                action();
+                action(this);
+            }
+            catch
+            {
+                if (!IsErrorState(this.state))
+                {
+                    this.SetState(BatchWriterState.Error);
+                }
+
+                throw;
+            }
+        }
+
+        /// <summary>
+        /// Catch any exception thrown by the action passed in; in the exception case move the writer into
+        /// state ExceptionThrown and then rethrow the exception.
+        /// </summary>
+        /// <typeparam name="TArg0">The action argument type.</typeparam>
+        /// <param name="action">The action to execute.</param>
+        /// <param name="arg0">The argument value provided to the action.</param>
+        /// <remarks>
+        /// Make sure to only use anonymous functions that don't capture state from the enclosing context, 
+        /// so the compiler optimizes the code to avoid creating allocations on every call to this method.
+        /// </remarks>
+        private void InterceptException<TArg0>(Action<ODataBatchWriter, TArg0> action, TArg0 arg0)
+        {
+            try
+            {
+                action(this, arg0);
             }
             catch
             {
@@ -716,11 +748,11 @@ namespace Microsoft.OData
         {
             if (!this.isInChangeset)
             {
-                this.InterceptException(this.IncreaseBatchSize);
+                this.InterceptException((thisParam) => thisParam.IncreaseBatchSize());
             }
             else
             {
-                this.InterceptException(this.IncreaseChangeSetSize);
+                this.InterceptException((thisParam) => thisParam.IncreaseChangeSetSize());
             }
 
             // Add a potential Content-ID header to the URL resolver so that it will be available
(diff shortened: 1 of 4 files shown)


src/Microsoft.OData.Core/Batch/ODataBatchWriter.cs
```diff
@@ -499,7 +499,7 @@ public virtual Task OnInStreamErrorAsync()
         /// <param name="newState">The writer state to transition into.</param>
         protected void SetState(BatchWriterState newState)
         {
-            this.InterceptException(() => this.ValidateTransition(newState));
+            this.InterceptException((thisParam, newStateParam) => thisParam.ValidateTransition(newStateParam), newState);
```
chrisspre, line 502: Since this is about allocating Lambdas, and this Lambda has no free variables, can this also be created statically

```
this.InterceptException(ValidateTransition), newState);
```

```
private static ValidateTransition(??? thisParam, ??? newStateParam) => thisParam.ValidateTransition(newStateParam);
```

src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
```diff
@@ -508,20 +511,52 @@ private void StartPayloadInStartState()
...
             if (current.State == CollectionWriterState.Start)
             {
-                this.InterceptException(this.StartPayload);
+                this.InterceptException((thisParam) => thisParam.StartPayload());
+            }
+        }
+
+        /// <summary>
+        /// Catch any exception thrown by the action passed in; in the exception case move the writer into
+        /// state ExceptionThrown and then re-throw the exception.
```
gathogojr, line 520:
```suggestion
        /// state Error and then re-throw the exception.
```

KenitoInc (approved):

src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
```diff
@@ -508,20 +511,52 @@ private void StartPayloadInStartState()
...
-        private void InterceptException(Action action)
+        /// <remarks>
+        /// Make sure to only use anonymous functions that don't capture state from the enclosing context, 
+        /// so the compiler optimizes the code to avoid creating allocations on every call to this method.
+        /// </remarks>
+        private void InterceptException(Action<ODataCollectionWriterCore> action)
+        {
+            try
+            {
+                action(this);
```
habbes, line 531: Maybe it would be a good idea to mention that this is done as an optimization to avoid delegate allocations?

src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
```diff
@@ -508,20 +511,52 @@ private void StartPayloadInStartState()
...
+        /// <remarks>
+        /// Make sure to only use anonymous functions that don't capture state from the enclosing context, 
+        /// so the compiler optimizes the code to avoid creating allocations on every call to this method.
+        /// </remarks>
+        private void InterceptException<TArg0>(Action<ODataCollectionWriterCore, TArg0> action, TArg0 arg0)
         {
             try
             {
-                action();
+                action(this, arg0);
```
habbes, line 559: Might be a good idea to indicate that this pattern is used as an optimization? Or?

habbes (approved):

PR #2159, opened 2021-08-12
update namespaces to use aliases if aliases for the namespaces are set

ElizabethOkerio:
### Issues

OData spec says:

5.1 Alias
A schema MAY specify an alias which MUST be a simple identifier.

If a schema specifies an alias, the alias MUST be used instead of the namespace within qualified names throughout the document to identify model elements of that schema. A mixed use of namespace-qualified names and alias-qualified names is not allowed.

https://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html#sec_Alias

### Description

*Currently the issue is with using aliases for referenced models if they are set instead of using fully qualified names when writing the csdl.*

- [x] *Ensure aliases are document global and do not apply only to the schema defining them*
- [x] *Add support for namespace aliases in referenced models*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Csdl/CsdlXmlWriter.cs
+++ b/src/Microsoft.OData.Edm/Csdl/CsdlXmlWriter.cs
@@ -5,6 +5,7 @@
 //---------------------------------------------------------------------
 
 using System;
+using System.Collections.Generic;
 using System.Diagnostics;
 using System.Xml;
 using Microsoft.OData.Edm.Csdl.Serialization;
@@ -96,8 +97,31 @@ namespace Microsoft.OData.Edm.Csdl
 
         private void WriteReferenceElements()
         {
-            EdmModelReferenceElementsXmlVisitor visitor = new EdmModelReferenceElementsXmlVisitor(this.model, this.writer, this.edmxVersion);
-            visitor.VisitEdmReferences(this.model);
+            EdmModelReferenceElementsXmlVisitor visitor;
+            IEnumerable<IEdmReference> references = model.GetEdmReferences();
+            if (references != null)
+            {
+                foreach (IEdmReference reference in references)
+                {
+                    //loop through the includes and set the namespace alias 
+                    if (reference.Includes != null)
+                    {
+                        foreach (IEdmInclude include in reference.Includes)
+                        {
+                            if (include.Alias != null)
+                            {
+                                model.SetNamespaceAlias(include.Namespace, include.Alias);
+                            }
+                        }
+                    }
+                }
+
+                foreach (IEdmReference edmReference in references)
+                {
+                    visitor = new EdmModelReferenceElementsXmlVisitor(this.model, this.writer, this.edmxVersion);
+                    visitor.VisitEdmReferences(this.model, edmReference);
+                }
+            }
         }
 
         private void WriteDataServicesElement()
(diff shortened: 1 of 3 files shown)


src/Microsoft.OData.Edm/Csdl/CsdlXmlWriter.cs
```diff
@@ -96,8 +97,31 @@ private void WriteConceptualModelsElement()
 
         private void WriteReferenceElements()
         {
-            EdmModelReferenceElementsXmlVisitor visitor = new EdmModelReferenceElementsXmlVisitor(this.model, this.writer, this.edmxVersion);
-            visitor.VisitEdmReferences(this.model);
+            EdmModelReferenceElementsXmlVisitor visitor;
+            IEnumerable<IEdmReference> references = model.GetEdmReferences();
+            if (model != null && references != null)
```
corranrogue9, line 102: `model` won't be `null` here if `model.GetEdmReferences` returned

src/Microsoft.OData.Edm/Csdl/CsdlXmlWriter.cs
```diff
@@ -96,8 +97,31 @@ private void WriteConceptualModelsElement()
...
+                            {
+                                model.SetNamespaceAlias(include.Namespace, include.Alias);
+                            }
+                        }
+                    }
+                }
+
+                foreach (IEdmReference tmp in references)
+                {
+                    visitor = new EdmModelReferenceElementsXmlVisitor(this.model, this.writer, this.edmxVersion);
```
corranrogue9, line 121: Is there a reason to have two loops here? I think that the body of this loop can be incorporated into the body of the first loop, but I could be missing something

corranrogue9 (approved):

xuzhg (approved):

corranrogue9 (approved):

src/Microsoft.OData.Edm/Csdl/CsdlXmlWriter.cs
ElizabethOkerio: @corranrogue9  the purpose of the two loops is to first have all the namespace-alias mapping for all the references before we start writing the references. 

For example 
`xml
<?xml version="1.0" encoding="utf-16"?>
<Edmx Version="4.0" xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx">
	<Reference Uri="https://example.com/Org.OData.Authorization.V1.xml">
		<Include Namespace="Org.OData.Authorization.V1" Alias="Auth">
			<Include Namespace="Org.OData.Authorization.V1" Alias="Auth">
				<Annotation Term="Org.OData.Core.V1.LongDescription" String="Include Description." />
			</Include>
			<IncludeAnnotations TermNamespace="org.example.validation" />
			<IncludeAnnotations TermNamespace="org.example.display" Qualifier="Tablet" />
		</Include>
		<Reference Uri="https://example.com/Org.OData.Core.V1.xml">
			<Include Namespace="Org.OData.Core.V1" Alias="Core" />
			<IncludeAnnotations TermNamespace="org.example.hcm" TargetNamespace="com.example.Sales" />
			<IncludeAnnotations TermNamespace="org.example.hcm" Qualifier="Tablet" TargetNamespace="com.example.Person" />
			<Annotation Term="Core.LongDescription" String="EdmReference Description." />
		</Reference>
		<DataServices />
	</Reference>
</Edmx>
if we have such  a csdl, The <code>Org.OData.Core.V1 </code annotation is defined in the second reference but is used within the first reference. In that case, a look-up of the alias for that namespace won't be found.

PR #2160, opened 2021-08-16
Expand fully qualified type

### LINKED ISSUE #1870: ExpandBinder throws on $expand ending in fully qualified type
When using a URL with an $expand that ends in a fully qualified type name, similar to example 40 from [OData Version 4.01. Part 2: URL Conventions](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html) - i.e. http://host/service/Orders?$expand=Customer/Model.VipCustomer - the ExpandBinder throws an ODataException.

As the conventions state: "For to-many relationships only Model.VipCustomer instances would be inlined" for this example. So it should be valid.

### Assemblies affected

Microsoft.OData.Core 7.7.1 and earlier

### Reproduce steps

Add this test to ExpandBinderTests.cs:

```
        [Fact]
        public void BindingOnTreeWithWithTypeTokenDoesNotThrow()
        {
            // Arrange: $expand=MyPeople/Fully.Qualified.Namespace.Employee
            NonSystemToken innerSegment = new NonSystemToken("Fully.Qualified.Namespace.Employee", null, null);
            NonSystemToken navProp = new NonSystemToken("MyPeople", null, innerSegment);
            ExpandToken expandToken = new ExpandToken(new ExpandTermToken[] { new ExpandTermToken(navProp) });

            // Act & Assert
            var binderForDog = new SelectExpandBinder(this.V4configuration, new ODataPathInfo(HardCodedTestModel.GetDogType(), null), null);
            var result = binderForDog.Bind(expandToken, null);
        }

```

### Expected result

The test would run successfully (even though it doesn't assert anything useful).

### Actual result

The test throws this exception:

```
  Message: 
    Microsoft.OData.ODataException : Found a path traversing multiple navigation properties. Please rephrase the query such that each expand path contains only type segments and navigation properties.
  Stack Trace: 
    SelectExpandBinder.GenerateExpandItem(ExpandTermToken tokenIn) line 362
    SelectEnumerableIterator`2.MoveNext()
    WhereEnumerableIterator`1.MoveNext()
    List`1.InsertRange(Int32 index, IEnumerable`1 collection)
    List`1.AddRange(IEnumerable`1 collection)
    SelectExpandBinder.Bind(ExpandToken expandToken, SelectToken selectToken) line 120
    ExpandBinderTests.BindingOnTreeWithWithTypeTokenDoesNotThrow() line 45

```

KenitoInc:
### Issues

*This pull request fixes #1870 .*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Microsoft.OData.Core.txt
+++ b/src/Microsoft.OData.Core/Microsoft.OData.Core.txt
@@ -846,7 +846,7 @@ PathParser_TypeCastOnlyAllowedInDerivedTypeConstraint=Type cast segment '{0}' on
 
 ODataResourceSet_MustNotContainBothNextPageLinkAndDeltaLink=A resource set may contain a next page link, a delta link or neither, but must not contain both.
 
-ODataExpandPath_OnlyLastSegmentMustBeNavigationProperty=The last segment, and only the last segment, must be a navigation property in $expand.
+ODataExpandPath_OnlyLastSegmentMustBeNavigationPropertyOrTypeSegment=The last segment, and only the last segment, must be a navigation property or type segment in $expand.
 ODataExpandPath_InvalidExpandPathSegment=Found a segment of type '{0} in an expand path, but only NavigationProperty, Property and Type segments are allowed.
 
 ODataSelectPath_CannotOnlyHaveTypeSegment=TypeSegment cannot be the only segment in a $select.
--- a/src/Microsoft.OData.Core/Microsoft.OData.Core.txt
+++ b/src/Microsoft.OData.Core/Microsoft.OData.Core.txt
@@ -846,7 +846,8 @@ PathParser_TypeCastOnlyAllowedInDerivedTypeConstraint=Type cast segment '{0}' on
 
 ODataResourceSet_MustNotContainBothNextPageLinkAndDeltaLink=A resource set may contain a next page link, a delta link or neither, but must not contain both.
 
-ODataExpandPath_OnlyLastSegmentMustBeNavigationPropertyOrTypeSegment=The last segment, and only the last segment, must be a navigation property or type segment in $expand.
+ODataExpandPath_OnlyLastSegmentCanBeNavigationProperty =The last segment, and only the last segment, can be a navigation property in $expand.
+ODataExpandPath_LastSegmentMustBeNavigationPropertyOrTypeSegment =The last segment must be a navigation property or type segment in $expand.
 ODataExpandPath_InvalidExpandPathSegment=Found a segment of type '{0} in an expand path, but only NavigationProperty, Property and Type segments are allowed.
 
 ODataSelectPath_CannotOnlyHaveTypeSegment=TypeSegment cannot be the only segment in a $select.
--- a/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
@@ -372,17 +372,15 @@ namespace Microsoft.OData.UriParser
             bool isRef = false;
             bool isCount = false;
             bool hasDerivedType = false;
-            PathSegmentToken derivedTypeToken = null;
-            IEdmEntityType derivedType = null;
+            IEdmType derivedType = null;
+            TypeSegment derivedTypeSegment = null;
 
-            // Handle $expand=Customer/VipCustomer
-            // The deriveTypeToken is VipCustomer
+            // Handle $expand=Customer/Fully.Qualified.Namespace.VipCustomer
+            // The deriveTypeToken is Fully.Qualified.Namespace.VipCustomer
             if (firstNonTypeToken.NextToken != null && firstNonTypeToken.NextToken.IsNamespaceOrContainerQualified())
             {
                 hasDerivedType = true;
-                derivedTypeToken = firstNonTypeToken.NextToken;
-                derivedType = UriEdmHelpers.FindTypeFromModel(this.Model, derivedTypeToken.Identifier, this.configuration.Resolver) as IEdmEntityType;
-                firstNonTypeToken = derivedTypeToken;
+                derivedType = UriEdmHelpers.FindTypeFromModel(this.Model, firstNonTypeToken.NextToken.Identifier, this.configuration.Resolver);
             }
 
             // ensure that we're always dealing with proper V4 syntax
@@ -391,14 +389,17 @@ namespace Microsoft.OData.UriParser
                 throw new ODataException(ODataErrorStrings.ExpandItemBinder_TraversingMultipleNavPropsInTheSamePath);
             }
 
-            if (firstNonTypeToken.NextToken != null)
+            if ((firstNonTypeToken.NextToken != null && !hasDerivedType) || 
+                (firstNonTypeToken.NextToken != null && firstNonTypeToken.NextToken.NextToken != null && hasDerivedType))
             {
+                PathSegmentToken nextToken = hasDerivedType ? firstNonTypeToken.NextToken.NextToken : firstNonTypeToken.NextToken;
+
                 // lastly... make sure that, since we're on a NavProp, that the next token isn't null.
-                if (firstNonTypeToken.NextToken.Identifier == UriQueryConstants.RefSegment)
+                if (nextToken.Identifier == UriQueryConstants.RefSegment)
                 {
                     isRef = true;
                 }
-                else if (firstNonTypeToken.NextToken.Identifier == UriQueryConstants.CountSegment)
+                else if (nextToken.Identifier == UriQueryConstants.CountSegment)
                 {
                     isCount = true;
                 }
@@ -411,37 +412,6 @@ namespace Microsoft.OData.UriParser
             // Add the segments in select and expand to parsed segments
             List<ODataPathSegment> parsedPath = new List<ODataPathSegment>(this.parsedSegments);
             parsedPath.AddRange(pathSoFar);
-
-            // Replace the currentNavProp with a Navigation property with the derived type.
-            if (derivedType != null)
-            {
-                EdmMultiplicity edmMultiplicity;
-
-                if (currentNavProp.Type.IsCollection())
-                {
-                    edmMultiplicity = EdmMultiplicity.Many;
-                }
-                else if (currentNavProp.Type.IsNullable)
-                {
-                    edmMultiplicity = EdmMultiplicity.ZeroOrOne;
-                }
-                else
-                {
-                    edmMultiplicity = EdmMultiplicity.One;
-                }
-
-                EdmNavigationPropertyInfo edmNavigationPropertyInfo = new EdmNavigationPropertyInfo()
-                {
-                    Name = currentNavProp.Name,
-                    Target = derivedType,
-                    TargetMultiplicity = edmMultiplicity,
-                    ContainsTarget = currentNavProp.ContainsTarget,
-                    OnDelete = currentNavProp.OnDelete
-                };
-
-                currentNavProp = EdmNavigationProperty.CreateNavigationProperty(currentNavProp.DeclaringType, edmNavigationPropertyInfo);
-            }
-
             IEdmNavigationSource targetNavigationSource = null;
             if (this.NavigationSource != null)
             {
@@ -452,6 +422,14 @@ namespace Microsoft.OData.UriParser
             NavigationPropertySegment navSegment = new NavigationPropertySegment(currentNavProp, targetNavigationSource);
             pathSoFar.Add(navSegment);
             parsedPath.Add(navSegment); // Add the navigation property segment to parsed segments for future usage.
+
+            if (hasDerivedType)
+            {
+                derivedTypeSegment = new TypeSegment(derivedType, targetNavigationSource);
+                pathSoFar.Add(derivedTypeSegment);
+                parsedPath.Add(derivedTypeSegment);
+            }
+
             ODataExpandPath pathToNavProp = new ODataExpandPath(pathSoFar);
 
             // $apply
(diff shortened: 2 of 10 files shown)


src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -369,15 +369,28 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
...
 
+            bool isRef = false;
+            bool isCount = false;
+            bool hasDerivedType = false;
+            PathSegmentToken derivedTypeToken = null;
+            IEdmEntityType derivedType = null;
+
+            // Handle $expand=Customer/VipCustomer
+            // The deriveTypeToken is VipCustomer
+            if (firstNonTypeToken.NextToken != null && firstNonTypeToken.NextToken.IsNamespaceOrContainerQualified())
```
xuzhg, line 380: >firstNonTypeToken.NextToken.IsNamespaceOrContainerQualified() [](http://example.com/codeflow?start=55&length=61)

IsNamespaceOrContainerQualified or 
Not IsNamespaceOrContainerQualified

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -399,6 +412,36 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
             List<ODataPathSegment> parsedPath = new List<ODataPathSegment>(this.parsedSegments);
             parsedPath.AddRange(pathSoFar);
 
+            // Replace the currentNavProp with a Navigation property with the derived type.
+            if (derivedType != null)
```
xuzhg, line 416: >if (derivedType != null) [](http://example.com/codeflow?start=12&length=24)

It doesn't make sense to me.

Why don't we create the type case segment?

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
KenitoInc: @xuzhg I have updated the comment

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
KenitoInc: I have update the code to create a type segment

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -369,23 +369,40 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
...
+            // Handle $expand=Customer/Fully.Qualified.Namespace.VipCustomer
+            // The deriveTypeToken is Fully.Qualified.Namespace.VipCustomer
+            if (firstNonTypeToken.NextToken != null && firstNonTypeToken.NextToken.IsNamespaceOrContainerQualified())
+            {
+                hasDerivedType = true;
+                derivedType = UriEdmHelpers.FindTypeFromModel(this.Model, firstNonTypeToken.NextToken.Identifier, this.configuration.Resolver);
+
+                // In this example: $expand=Customer/Fully.Qualified.Namespace.VipCustomer
+                // We validate that the derived type Fully.Qualified.Namespace.VipCustomer is related to Navigation property Customer.
+                UriEdmHelpers.CheckRelatedTo(currentNavProp.ToEntityType(), derivedType);
```
gathogojr, line 386: Does `CheckRelatedTo` method throw an exception if derived type is not assigned to the navigation property type?

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -369,23 +369,40 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
                 currentNavProp = ParseComplexTypesBeforeNavigation(currentComplexProp, ref firstNonTypeToken, pathSoFar);
             }
 
+            bool isRef = false;
+            bool isCount = false;
+            bool hasDerivedType = false;
```
gathogojr, line 374: `hasTypeSegment` better maybe?

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -369,23 +369,40 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
...
+                derivedType = UriEdmHelpers.FindTypeFromModel(this.Model, firstNonTypeToken.NextToken.Identifier, this.configuration.Resolver);
+
+                // In this example: $expand=Customer/Fully.Qualified.Namespace.VipCustomer
+                // We validate that the derived type Fully.Qualified.Namespace.VipCustomer is related to Navigation property Customer.
+                UriEdmHelpers.CheckRelatedTo(currentNavProp.ToEntityType(), derivedType);
+            }
+
             // ensure that we're always dealing with proper V4 syntax
-            if (firstNonTypeToken.NextToken != null && firstNonTypeToken.NextToken.NextToken != null)
+            if (firstNonTypeToken.NextToken != null && firstNonTypeToken.NextToken.NextToken != null && !hasDerivedType)
```
gathogojr, line 390:
```suggestion
            if (firstNonTypeToken?.NextToken?.NextToken != null && !hasDerivedType)
```

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -369,23 +369,40 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
...
             {
                 throw new ODataException(ODataErrorStrings.ExpandItemBinder_TraversingMultipleNavPropsInTheSamePath);
             }
 
-            bool isRef = false;
-            bool isCount = false;
-
-            if (firstNonTypeToken.NextToken != null)
+            if ((firstNonTypeToken.NextToken != null && !hasDerivedType) || 
+                (firstNonTypeToken.NextToken != null && firstNonTypeToken.NextToken.NextToken != null && hasDerivedType))
```
gathogojr, line 396:
```suggestion
                (firstNonTypeToken?.NextToken?.NextToken != null && hasDerivedType))
```

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataExpandPath.cs
```diff
@@ -60,23 +60,19 @@ private void ValidatePath()
             {
                 if (segment is TypeSegment)
```
gathogojr, line 61: Instead of retaining this empty block, would changing the `else` block at the bottom as follows be better:
```csharp
                else if (!(segment is TypeSegment))
                {
                    throw new ODataException(ODataErrorStrings.ODataExpandPath_InvalidExpandPathSegment(segment.GetType().Name));
                }
```

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
KenitoInc: Yes it does;
https://github.com/OData/odata.net/blob/13e697b106a445050d55e6fb7995c9e91805ebcc/src/Microsoft.OData.Core/UriParser/UriEdmHelpers.cs#L51-L59

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
KenitoInc: Used `hasDerivedTypeSegment` since there are other instances where we have TypeSegments before the first non-type segment e.g `Fully.Qualified.Namespace.Employee/Friends`

gathogojr (approved):
LGTM

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -369,23 +369,40 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
...
+            bool isCount = false;
+            bool hasDerivedTypeSegment = false;
+            IEdmType derivedType = null;
+
+            // Handle $expand=Customer/Fully.Qualified.Namespace.VipCustomer
+            // The deriveTypeToken is Fully.Qualified.Namespace.VipCustomer
+            if (firstNonTypeToken.NextToken != null && firstNonTypeToken.NextToken.IsNamespaceOrContainerQualified())
+            {
+                hasDerivedTypeSegment = true;
+                derivedType = UriEdmHelpers.FindTypeFromModel(this.Model, firstNonTypeToken.NextToken.Identifier, this.configuration.Resolver);
```
mikepizzo, line 382: >derivedType = UriEdmHelpers.FindTypeFromModel [](http://example.com/codeflow?start=16&length=45)

do you need any special handling for FindTypeFromModel not finding the type? In that case, I believe derivedType would return null, and it looks like CheckRelatedTo may throw a null ref exception as IsRelatedTo will return false, but the exception that we try to generate on line 57 of EdmUriHelpers will try to call FullTypeName() on the null value.

Minimally, we should fix EdmUriHelpers.CheckRelatedTo to handle the null childType and add tests for adding a cast segment with a type that is not found.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/SelectExpandFunctionalTests.cs
```diff
@@ -1667,6 +1667,96 @@ public void ExpandWithNavigationPropCountWithSearchOptionWorks()
             Assert.NotNull(expandedCountSelectItem.SearchOption);
         }
 
+        // $expand=navProp/fully.qualified.type/$count
```
mikepizzo, line 1670: should also have tests for /$expand=navProp/fully.qualified.type/$ref

mikepizzo, line 1670: Can we add tests for fully.qualified.type being an alias?

mikepizzo, line 1670: Please add a (negative) test case for fully.qualified.undefinedType and verify meaningful exception (i.e., saying that the type fully.qualified.undefinedType was not found).

mikepizzo (changes requested):
:clock1:

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
KenitoInc: When `FindTypeFromModel` doesn't find the type it returns a null. I have added code to throw an exception when the `derivedType` is `null`.

Regarding `EdmUriHelpers.CheckRelatedTo`, I have handled the null `childType` similar to how a null `parentType` is handled.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/SelectExpandFunctionalTests.cs
KenitoInc: Done

KenitoInc: Done

KenitoInc: @mikepizzo I am not sure what you mean. Kindly clarify.
cc @gathogojr

KenitoInc: From the spec, parameter aliases are only used for `$filter` or `$orderby` https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_ParameterAliases

mikepizzo: Sorry; i meant using a (namespace alias)[http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_Alias] to qualify the type name, rather than the full namespace name.

For example:

<Schema Namespace="fully.qualified" Alias="alias">
...
</Schema>

$expand=navProp/alias.type

KenitoInc: @mikepizzo I have added the tests

src/Microsoft.OData.Core/Microsoft.OData.Core.txt
```diff
@@ -846,7 +846,7 @@ PathParser_TypeCastOnlyAllowedInDerivedTypeConstraint=Type cast segment '{0}' on
 
 ODataResourceSet_MustNotContainBothNextPageLinkAndDeltaLink=A resource set may contain a next page link, a delta link or neither, but must not contain both.
 
-ODataExpandPath_OnlyLastSegmentMustBeNavigationProperty=The last segment, and only the last segment, must be a navigation property in $expand.
+ODataExpandPath_OnlyLastSegmentMustBeNavigationPropertyOrTypeSegment=The last segment, and only the last segment, must be a navigation property or type segment in $expand.
```
habbes, line 849: Does this imply that it's not allowed to have type segments before the last segment?

KenitoInc: Good catch. 
We can have type segments before the last segments.
The error string should be 
`ODataExpandPath_LastSegmentMustBeNavigationPropertyOrTypeSegment` = `The last segment must be a navigation property or type segment in $expand.`

habbes (approved):

mikepizzo (approved):
:shipit:

PR #2164, opened 2021-08-23
key alias support

ElizabethOkerio:
### Issues

### Description

Key Alias support in ODL

OData spec says:

$Key
The value of $Key is an array with one item per key property.

Key properties without a key alias are represented as strings containing the property name.

Key properties with a key alias are represented as objects with one member whose name is the key alias and whose value is a string containing the path to the property.

The second one is not supported in ODL.

```C SHARP
"Category": {
  "$Kind": "EntityType",
  "$Key": [
    {
      "EntityInfoID": "Info/ID"
    }
  ],
  "Info": {
    "$Type": "self.EntityInfo"
  },
  "Name": {
    "$Nullable": true
  }
},
"EntityInfo": {
  "$Kind": "ComplexType",
  "ID": {
    "$Type": "Edm.Int32"
  },
  "Created": {
    "$Precision": 0
  }
}
```

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/tools/CustomMSBuild/Versioning.props
+++ b/tools/CustomMSBuild/Versioning.props
@@ -4,7 +4,7 @@
   <PropertyGroup>
     <VersionMajor Condition="'$(VersionMajor)' == ''">7</VersionMajor>
     <VersionMinor Condition="'$(VersionMinor)' == ''">9</VersionMinor>
-    <VersionBuildNumber Condition="'$(VersionBuildNumber)' == ''">1</VersionBuildNumber>
+    <VersionBuildNumber Condition="'$(VersionBuildNumber)' == ''">2</VersionBuildNumber>
     <VersionRelease Condition="'$(VersionRelease)' == ''"></VersionRelease>
   </PropertyGroup>
 


src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlPropertyReference.cs
```diff
@@ -12,16 +12,23 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Ast
...
     {
         private readonly string propertyName;
+        private readonly string alias;
 
-        public CsdlPropertyReference(string propertyName, CsdlLocation location)
+        public CsdlPropertyReference(string propertyName, string alias, CsdlLocation location)
             : base(location)
         {
             this.propertyName = propertyName;
+            this.alias = alias;
```
marabooy, line 21: nit: Maybe we could switch this to auto properties. :)

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -139,5 +161,30 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
 
             return null;
         }
+
+        private IEnumerable<KeyValuePair<IEdmStructuralProperty, string>> ComputeDeclaredKeyAlias()
+        {
+            if (this.entity.Key != null)
```
marabooy, line 167: nit: Invert if to reduce nesting.

marabooy (dismissed):

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -139,5 +161,30 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+                foreach (CsdlPropertyReference keyProperty in this.entity.Key.Properties)
+                {
+                    string propertyName = "";
+                    if (!String.IsNullOrEmpty(keyProperty.Alias))
+                    {
+                        string[] keyProperties = keyProperty.PropertyName.Split('/');
+                        propertyName = keyProperties[1];
+
+                        IEdmStructuralProperty structuralProperty = this.FindProperty(propertyName) as IEdmStructuralProperty;
+                        if (structuralProperty != null)
```
odero, line 179:
```suggestion
                        if (this.FindProperty(propertyName) is IEdmStructuralProperty structuralProperty)

```
nit: Simplifying

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
```diff
@@ -157,6 +157,14 @@ internal override void WritePropertyRefElement(IEdmStructuralProperty property)
             this.WriteEndElement();
         }
 
+        internal override void WritePropertyRefElement(string keyname, string keyAlias)
```
odero, line 160:
```suggestion
        internal override void WritePropertyRefElement(string keyName, string keyAlias)

```
nit: on naming consistency

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +123,18 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
                 List<IEdmStructuralProperty> key = new List<IEdmStructuralProperty>();
                 foreach (CsdlPropertyReference keyProperty in this.entity.Key.Properties)
                 {
-                    IEdmStructuralProperty structuralProperty = this.FindProperty(keyProperty.PropertyName) as IEdmStructuralProperty;
+                    string propertyName = "";
+                    if (!String.IsNullOrEmpty(keyProperty.Alias))
+                    {
+                        string[] keyProperties = keyProperty.PropertyName.Split('/');
+                        propertyName = keyProperties[1];
```
habbes, line 130: Does this mean we only support one level of nesting for property aliases?
According to the [spec](https://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html#sec_Key):

    A key property MUST be a non-nullable primitive property of the entity type itself, including non-nullable primitive properties of non-nullable single-valued complex properties, recursively.

I believe this means we an alias can refer to the property of a property of a property etc. so we can have multiple `/` in the property name.

habbes: And if this is an alias, what's the rationale behind treating the last segment as the property name? Why not take the `keyProperty.Alias` as the property name? Just curious.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +123,18 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+                    {
+                        string[] keyProperties = keyProperty.PropertyName.Split('/');
+                        propertyName = keyProperties[1];
+                    }
+                    else 
+                    {
+                        propertyName = keyProperty.PropertyName;
+                    }
+
+                    IEdmStructuralProperty structuralProperty = this.FindProperty(propertyName) as IEdmStructuralProperty;
```
habbes, line 137: If the property is an alias, will it be found by `this.FindProperty()` or will it return null?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -122,14 +144,14 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
                         // If keyProperty is a duplicate, it will come back as non-structural from FindProperty, but it still might be structural
                         // inside the DeclaredProperties, so try it. If it is not in the DeclaredProperties or it is not structural there,
                         // then fall back to unresolved.
-                        structuralProperty = this.DeclaredProperties.FirstOrDefault(p => p.Name == keyProperty.PropertyName) as IEdmStructuralProperty;
+                        structuralProperty = this.DeclaredProperties.FirstOrDefault(p => p.Name == propertyName) as IEdmStructuralProperty;
```
habbes, line 147: Will this find aliases or return null?

src/Microsoft.OData.Edm/Schema/CoreModel/EdmCoreModelEntityType.cs
```diff
@@ -122,6 +123,15 @@ public IEnumerable<IEdmStructuralProperty> DeclaredKey
             get { return Enumerable.Empty<IEdmStructuralProperty>(); }
         }
 
+        /// <summary>
+        /// Gets the key aliases for the structural properties of the entity type that make up the entity key.
+        /// The Edm.EntityType is always without any declared key aliases.
```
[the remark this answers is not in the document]
ElizabethOkerio: It is the same as saying it does not have any declared key aliases by default but they can be added.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -139,5 +161,30 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
 
             return null;
         }
+
+        private IEnumerable<KeyValuePair<IEdmStructuralProperty, string>> ComputeDeclaredKeyAlias()
```
habbes, line 165: What does this method do?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -139,5 +161,30 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+            if (this.entity.Key != null)
+            {
+                List<KeyValuePair<IEdmStructuralProperty, string>> keyAlias = new List<KeyValuePair<IEdmStructuralProperty, string>>();
+                foreach (CsdlPropertyReference keyProperty in this.entity.Key.Properties)
+                {
+                    string propertyName = "";
+                    if (!String.IsNullOrEmpty(keyProperty.Alias))
+                    {
+                        string[] keyProperties = keyProperty.PropertyName.Split('/');
+                        propertyName = keyProperties[1];
```
habbes, line 176: What if `propertyName` has multiple slashes?

Also, what if it does not have a slash at all, e.g. using an alias to an immediate property of the entity? This would throw an `IndexOutOfBoundException`. Or is there somewhere we're validating the the property name must include `/` if the `keyProperty.Alias` is non-empty?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -139,5 +161,30 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+
+                        IEdmStructuralProperty structuralProperty = this.FindProperty(propertyName) as IEdmStructuralProperty;
```
habbes, line 178: If `propertyName` is the part of a reference to a property of a complex property or related entity, then I assume that means that the reference property is not declared on this entity. If that's the case, how would `this.FindProperty()` find that property?

Furthermore, if you strip out the slashes, then it's possible you might end up with a different property altogether.

For example, let's say we have the following entity:
```xml
<EntitType Name="Foo">
  <Key>
     <PropertyRef Alias="ID" Name="Complex/Bar">
  </Key>
  <Property Name="Bar" Type="Edm.Int32" Nullable="False" />
  <Property Name="Complex" Type="NS.Complex" Nullable="False" />
</EntityType>
<ComplexType Name="Complex">
   <Property Name="Bar" Type="Edm.Int32" Nullable="False" />
</ComplexType>
```
If the property name is "Complex/Bar" then you know that it uniquely refers to the `Bar` property of the `Complex` property of the entity. But if you split the property name and just remain with `Bar` then you might end up with the wrong `Bar` property cause now it's ambiguous. It's also possible that the entity has multiple complex properties with a property called `Bar`.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
```diff
@@ -145,7 +145,7 @@ protected override void ProcessEntityType(IEdmEntityType element)
             this.BeginElement(element, this.schemaWriter.WriteEntityTypeElementHeader);
             if (element.DeclaredKey != null && element.DeclaredKey.Any())
             {
-                this.VisitEntityTypeDeclaredKey(element.DeclaredKey);
+                this.VisitEntityTypeDeclaredKey(element.DeclaredKey, element.DeclaredKeyAlias);
```
habbes, line 148: Does this mean that you have separate properties for the key properties that are defined on the entity (structural properties) and the key properties that are aliases to properties on complex or related types?

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
```diff
@@ -145,7 +145,7 @@ protected override void ProcessEntityType(IEdmEntityType element)
             this.BeginElement(element, this.schemaWriter.WriteEntityTypeElementHeader);
             if (element.DeclaredKey != null && element.DeclaredKey.Any())
```
habbes, line 146: What if `element.DeclaredKey` is null but `element.DeclareKeyAlias` is not null (i.e. the entity type key is only composed of aliases)` ? Shouldn't you still visit the key in that case?

habbes (changes requested):

src/Microsoft.OData.Edm/Schema/BadEntityType.cs
```diff
@@ -34,5 +35,7 @@ public bool HasStream
         {
             get { return false; }
         }
+
+        public IEnumerable<KeyValuePair<IEdmStructuralProperty, string>> DeclaredKeyAlias => throw new NotImplementedException();
```
habbes, line 39: Would a dictionary mapping aliases to their properties (or property name) not be a better way to access aliases?

src/Microsoft.OData.Edm/Schema/CoreModel/EdmCoreModelEntityType.cs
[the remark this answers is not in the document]
habbes: "always without" and "they can be added" seem contradictory.

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
```diff
@@ -84,6 +87,14 @@ public virtual IEnumerable<IEdmStructuralProperty> DeclaredKey
             get { return this.declaredKey; }
         }
 
+        /// <summary>
+        /// Gets the aliases for the declared key properties of the entity type.
+        /// </summary>
+        public virtual IEnumerable<KeyValuePair<IEdmStructuralProperty, string>> DeclaredKeyAlias
```
habbes, line 93: I'm curious to know what the structural property that's returned from this field look like, what would be its `DeclaringType` be set as? In practice the property will be declared on different complex type or entity type other than the one that `this` refers to. So how do we capture that fact here? How can we tell which property of the entity contains the property that's being referenced by the alias?

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderJsonTests.cs
```diff
@@ -1858,6 +1858,41 @@ public void ParsingPropertyWithUntypedTypeAsPropertyTypeInJsonWorks()
...
+  ""NS1"": {
+    ""Product"": {
+      ""$Kind"": ""EntityType"",
+      ""$Key"": [
+        {
+          ""ExampleObjectName"": ""NS1.ExampleComplexObject/ExampleObject""
+        }
+      ],
+      ""ExampleObject"": {
+        ""$Type"": ""NS1.ExampleComplexObject""
```
habbes, line 1876: The `NS1.ExampleComplexObject` type does not appear to be defined in this CSDL schema.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderJsonTests.cs
```diff
@@ -1858,6 +1858,41 @@ public void ParsingPropertyWithUntypedTypeAsPropertyTypeInJsonWorks()
...
+      ""$Kind"": ""EntityContainer"",
+      ""Products"": {
+        ""$Collection"": true,
+        ""$Type"": ""NS1.Product""
+      }
+    }
+  }
+}"
+;
+            IEdmModel model = Parse(csdl);
```
habbes, line 1893: Could you add assertions to verify that the model has been successfully read? Could you demonstrate the key aliases being retrieved and the key properties being inspected?

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
```diff
@@ -1699,5 +1699,32 @@ public void ImportSchemaElementsOnlyFromReferencedNamespacesl()
...
+        {
+            var csdl = @"<?xml version=""1.0"" encoding=""utf-8""?>
+            <edmx:Edmx Version=""4.0"" xmlns:edmx=""http://docs.oasis-open.org/odata/ns/edmx"">
+              <edmx:DataServices>
+                <Schema Namespace=""NS1"" xmlns=""http://docs.oasis-open.org/odata/ns/edm"">
+                  <EntityType Name=""Product"">
+                    <Key>
+                      <PropertyRef Name=""NS1.ExampleComplexObject/ExampleObject"" Alias=""ExampleObjectName""/>
+                    </Key>
+                    <Property Name=""ExampleObject"" Type=""NS1.ExampleComplexObject"" Nullable=""false""/>
```
habbes, line 1714: The `NS1.ExampleComplexObject` does not appear to be defined in this schema.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
```diff
@@ -1699,5 +1699,32 @@ public void ImportSchemaElementsOnlyFromReferencedNamespacesl()
...
+                  </EntityType>
+                  <EntityContainer Name=""Container"">
+                    <EntitySet Name=""Products"" EntityType=""NS1.Product""/>
+                  </EntityContainer>
+                </Schema>
+              </edmx:DataServices>
+            </edmx:Edmx>
+            ";
+            var model = CsdlReader.Parse(XElement.Parse(csdl).CreateReader());
+            var productType = model.FindType("NS1.Product");
```
habbes, line 1726: Could you add assertions to verify that the model works as expected, that we can retrieve the key properties and aliases?

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
```diff
@@ -2460,6 +2460,77 @@ private void WriteAndVerifyXml(IEdmModel model, string expected, CsdlTarget targ
...
+        public void VerifyKeyAliasForComplexPropertiesIsSupported()
+        {
+            // Arrange
+            var model = new EdmModel();
+            var entity = new EdmEntityType("NS1", "Product");
+            var complexType = new EdmComplexType("NS1", "ExampleComplexObject");
+            complexType.AddStructuralProperty("Name", EdmCoreModel.Instance.GetString(false));
+            var entityId = entity.AddStructuralProperty("ExampleObject", new EdmComplexTypeReference(complexType, isNullable: false));
+            entity.AddKeys(entityId);
+            entity.AddKeyAlias(new KeyValuePair<IEdmStructuralProperty, string>(entityId, "ExampleObjectName"));
```
habbes, line 2473: This `entityId` is a reference to the complex property `ExampleObject`. However, according to the spec, the alias must reference a property of a complex property. It must not be a reference to the complex property itself ([the spec](https://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html#sec_Key) says that a key property must primitive).

habbes: So for example, the alias may reference the `Name` of the `ExampleObject` property, but it may not be a reference to `ExampleObject` itself, which is what seems to be happening here.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
```diff
@@ -2460,6 +2460,77 @@ private void WriteAndVerifyXml(IEdmModel model, string expected, CsdlTarget targ
...
+            entityContainer.AddElement(set1);
+
+            // Act & Assert for XML
+            WriteAndVerifyXml(model, "<?xml version=\"1.0\" encoding=\"utf-16\"?>" +
+                "<edmx:Edmx Version=\"4.0\" xmlns:edmx=\"http://docs.oasis-open.org/odata/ns/edmx\">" +
+                  "<edmx:DataServices>" +
+                    "<Schema Namespace=\"NS1\" xmlns=\"http://docs.oasis-open.org/odata/ns/edm\">" +
+                      "<EntityType Name=\"Product\">" +
+                        "<Key>" +
+                          "<PropertyRef Name=\"NS1.ExampleComplexObject/ExampleObject\" Alias=\"ExampleObjectName\" />" +
```
habbes, line 2490: I think the name of the property ref in this case should be `<NameOfComplexProperty>/<NameOfReferencedProperty>` not ~`<TypeOfProperty>/<NameOfProperty>`~. So a valid example would have been:
```xml
<PropertyRef Name="ExampleObject/Name" Alias="ExampleObjectName" />
```

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
```diff
@@ -2460,6 +2460,77 @@ private void WriteAndVerifyXml(IEdmModel model, string expected, CsdlTarget targ
...
         }
 
+        [Fact]
+        public void VerifyKeyAliasForComplexPropertiesIsSupported()
+        {
+            // Arrange
+            var model = new EdmModel();
+            var entity = new EdmEntityType("NS1", "Product");
+            var complexType = new EdmComplexType("NS1", "ExampleComplexObject");
+            complexType.AddStructuralProperty("Name", EdmCoreModel.Instance.GetString(false));
```
habbes, line 2470: Could you also add the complex type to the model?

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
```diff
@@ -2460,6 +2460,77 @@ private void WriteAndVerifyXml(IEdmModel model, string expected, CsdlTarget targ
...
+                      "</EntityType>" +
+                      "<EntityContainer Name=\"Container\">" +
+                        "<EntitySet Name=\"Products\" EntityType=\"NS1.Product\" />" +
+                      "</EntityContainer>" +
+                    "</Schema>" +
+                  "</edmx:DataServices>" +
+                "</edmx:Edmx>");
+
+            // Act & Assert for JSON
+            WriteAndVerifyJson(model, @"{
```
habbes, line 2504: Same comments as with xml version.

habbes (changes requested):
I think there's a point of confusion. According to this PR the alias seems to reference a complex property. But according to [the spec](https://docs.oasis-open.org/odata/odata-csdl-json/v4.01/odata-csdl-json-v4.01.html#sec_Key) (or at least how I understand it), the alias should reference a property of a complex property (or a related entity). Furthermore, I believe multiple levels of nesting should be supported (i.e. a property of a complex property of a complex property, etc.), but in this PR only one level is supported. I think there should also be more test verifying these scenarios.

Also one thing that's not clear is how we intend the public API for accessing key properties to work now that aliases are being supported and we have cases where the properties referenced in the key are not structural properties of the entity. I think you should emphasize that in the description since that's a non-negligible design decision that we'll not be able to change until another major release. Is this PR meant for 7.x or the next major release?

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
ElizabethOkerio: yes. seen this is an error

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
ElizabethOkerio: <code> ExampleObject </code> is the name of the property in the declaring Type. and the type of <code> ExampleObject </code> is the complex type. <code> ExampleComplexObject </code>.

ElizabethOkerio:
```xml
<EntityType Name="Category">
  <Key>
    <PropertyRef Name="Info/ID" Alias="EntityInfoID" />
  </Key>
  <Property Name="Info" Type="Sales.EntityInfo" Nullable="false" />
  <Property Name="Name" Type="Edm.String" />
</EntityType>

 

<ComplexType Name="EntityInfo">

  <Property Name="ID" Type="Edm.Int32" Nullable="false" />
  <Property Name="Created" Type="Edm.DateTimeOffset" />
</ComplexType>
```

ElizabethOkerio: aah I see what you mean.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
ElizabethOkerio: Key properties are structural properties within an entity type (in which they are declared) that have been declared as key properties. The DeclaredKeyAliases has a mapping between these KeyProperties and their aliases. I don't think we can have a situation where we have aliases but no key properties but the reverse is possible.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
ElizabethOkerio: What is happening here is: we first visit the key then using the key name we check if that key has an alias or not.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
```diff
@@ -680,18 +680,34 @@ private void ProcessFacets(IEdmTypeReference element, bool inlineType)
...
             this.schemaWriter.WriteArrayEndElement();
         }
 
-        private void VisitPropertyRefs(IEnumerable<IEdmStructuralProperty> properties)
+        private void VisitPropertyRefs(IEnumerable<IEdmStructuralProperty> properties, IEnumerable<KeyValuePair<IEdmStructuralProperty, string>> keyAliases)
         {
             foreach (IEdmStructuralProperty property in properties)
             {
-                this.schemaWriter.WritePropertyRefElement(property);
+                if (keyAliases != null)
```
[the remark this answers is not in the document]
ElizabethOkerio: We are not looping through the key aliases, we are looping through the declared key properties. there could be more than one key property. For each key we check if it has an alias in the DeclaredKeyAlias, if it does then we write its PropertyRef with an alias.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
ElizabethOkerio: This method reads DeclaredKeyAliases from the <code>CsdlPropertyReference <code> object.

src/Microsoft.OData.Edm/Csdl/Parsing/SchemaJsonParser.cs
```diff
@@ -655,17 +655,24 @@ internal static CsdlKey ParseCsdlKey(string name, JsonElement keyArray, JsonPars
...
+                        pName = keyName.ParseAsString(p);
+                        pValue = keyAlias;
+                    });
+
+                    return new CsdlPropertyReference(pName, pValue, context.Location());
                 }
 
                 string propertyName = v.ParseAsString(context);
-                return new CsdlPropertyReference(propertyName, context.Location());
+                return new CsdlPropertyReference(propertyName, null, context.Location());
```
corranrogue9, line 675: nit: it might be good to have two constructors so that we don't have to pass `null` to get the "correct" overload

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +112,39 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
                 foreach (CsdlPropertyReference keyProperty in this.entity.Key.Properties)
                 {
-                    IEdmStructuralProperty structuralProperty = this.FindProperty(keyProperty.PropertyName) as IEdmStructuralProperty;
+                    IEdmStructuralProperty structuralProperty = null;
+                    if (!string.IsNullOrEmpty(keyProperty.Alias))
+                    {
+                        string[] keyProperties = keyProperty.PropertyName.Split('/');
+                        for (int i = 0; i < keyProperties.Length; i++)
+                        {
+                            if (i == 0)
```
corranrogue9, line 121: Please unroll the loop so that we don't have to do this check for every subsequent iteration

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +112,39 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+                            {
+                                structuralProperty = this.FindProperty(keyProperties[i]) as IEdmStructuralProperty;
+                                i++;
+                            }
+                            else
+                            {
+                                structuralProperty = this.FindProperty(keyProperties[i]) as IEdmStructuralProperty;
+                                i++;
```
corranrogue9, line 129: It seems that we are actually counting by 2. It might be good to just have the `for` loop use `i += 2` instead so that the loop body is more clear

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +112,39 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+                            }
+
+                            if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Complex)
+                            {
+                                structuralProperty = structuralProperty.Type.AsComplex().FindProperty(keyProperties[i]) as IEdmStructuralProperty;
```
corranrogue9, line 134: I don't know if this is getting validated somewhere else (maybe in `CsdlPropertyReference`?) but it would be really bad if we went passed the end of `keyProperties` here. Maybe we should check the length is even before we start the loop

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
corranrogue9: Actually, the body of these two `if` statements look the same to me. Why do we need this check here?

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
```diff
@@ -691,7 +691,15 @@ private void VisitPropertyRefs(IEnumerable<IEdmStructuralProperty> properties)
...
             {
-                this.schemaWriter.WritePropertyRefElement(property);
+                var prop = property as IStructuredKeyAlias;
+                if (string.IsNullOrEmpty(prop?.KeyAlias))
+                {
+                    this.schemaWriter.WritePropertyRefElement(property);
+                }
+                else 
+                {
+                    this.schemaWriter.WritePropertyRefElement(prop.Path, prop.KeyAlias);
```
corranrogue9, line 701: Maybe this is asserted by the caller somewhere, but can't `prop` still potentially be `null` at this point?

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
```diff
@@ -1699,5 +1699,39 @@ public void ImportSchemaElementsOnlyFromReferencedNamespacesl()
...
+                  </ComplexType>
+                  <EntityContainer Name=""Container"">
+                    <EntitySet Name=""Products"" EntityType=""NS1.Product""/>
+                  </EntityContainer>
+                </Schema>
+              </edmx:DataServices>
+            </edmx:Edmx>
+            ";
+
+            var model = CsdlReader.Parse(XElement.Parse(csdl).CreateReader());
```
corranrogue9, line 1729: `CreateReader` returns an object that implements `IDisposable`. Please wrap it in a `using` block.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
ElizabethOkerio: am checking if the prop is null in the if--statement.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
```diff
@@ -2460,6 +2460,84 @@ private void WriteAndVerifyXml(IEdmModel model, string expected, CsdlTarget targ
...
+        public void VerifyKeyAliasForComplexPropertiesIsSupported()
+        {
+            // Arrange
+            var model = new EdmModel();
+            var entity = new EdmEntityType("NS1", "Product");
+            var complexType = new EdmComplexType("NS1", "ExampleComplexObject");
+            entity.AddStructuralProperty("ExampleObject", new EdmComplexTypeReference(complexType, isNullable: false));
+            entity.AddStructuralProperty("Name", EdmCoreModel.Instance.GetString(false));
+            entity.AddStructuralProperty("UpdatedTime", EdmCoreModel.Instance.GetDate(false));
+            var keyProperty = complexType.AddStructuralPropertyAlias("Name", EdmCoreModel.Instance.GetString(false), "ExampleObjectName", new List<string> { "ExampleObject", "Name" });
```
habbes, line 2473: I think it's redundant to have both `Name` and `{ ExampleObject, Name }` as parameters since you can retrieve the property name from the latter.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +112,30 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+                    if (!string.IsNullOrEmpty(keyProperty.Alias))
+                    {
+                        string[] keyProperties = keyProperty.PropertyName.Split('/');
+                        for (int i = 0; i < keyProperties.Length; i++)
+                        {
+                            structuralProperty = this.FindProperty(keyProperties[i]) as IEdmStructuralProperty;
+
+                            if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Complex)
+                            {
+                                structuralProperty = structuralProperty.Type.AsComplex().FindProperty(keyProperties[i]) as IEdmStructuralProperty;
```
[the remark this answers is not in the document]
habbes: After looking at the tests, I think I know the tests are passing despite this apparent bug:
- In your tests you have an entity type `Product` which has a key property that aliases `ExampleObject/Name`
- `Product` also has a `Name` property
- When this method is called, the key property name is split into `[ExampleObject, Name]`
- In the first iteration (`i = 0`), `structuralProperty = this.FindProperty(keyProperties[i])` sets the `structureProperty` to `ExampleObject` initially.
- But then in the same iteration you call `ExampleObject.Type...FindProperty(keyProerties[i])` which attempts to find `ExampleObject` but returns null cause the property doesn't exist
- In the second iteration `i = 1`, you replace `structuralProperty = this.FindProperty(keyProperties[i])`, in this case the property path segment is `Name`. Since `Name` is also a property of `Product`, `Product.Name` is returned (not ExampleObject.Name), since it's neither a complex or entity type, the if statements are skipped over, avoiding the `NullReferenceException`
- So this actual returns the wrong property for the key, but since the test doesn't assert the type, declaring type, etc. of the returned property, it doesn't detect this bug.

habbes (changes requested):
The overall design looks okay to me, however there seems to be potential bugs in the code and gaps in the tests which cause the bugs to go undetected by the tests.

I suggest improving the tests before fixing the bugs:
- When reading a CSDL including key aliases, make sure to assert all relevant properties of the key properties, not just the name and the alias (e.g. `Type`, `DeclaringType`, `PropertyKind`, etc.)
- Include tests that include property refs with more than 2 levels of nesting
- Include tests for key aliases that reference properties of a related entity/navigation property, not just complex type
- Include tests where the key is composed of multiple properties, a mixture of actual properties and property refs.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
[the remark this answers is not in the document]
ElizabethOkerio: > I initially assumed this loop is trying to walk down the property path to find the actual property being referenced. So let's assume `PropertyName` is initially `Address/City/Town`, and `keyProperties` become `[Address, City, Town]`.
> 
> When `i` = 0:
> 
> * `structuralProperty = this.FindProperty(keyProperties[i])` will set the `structuralProperty` to `Address`
> * then if `structuralProperty` is a complex type, then wouldn't `structuralProperty...FindProperty(keyProperties[i])` also try to find the `Address` property in `structuralProperty`? Isn't that an error? Shouldn't we be trying to find the property `City` instead which is `keyProperties[i + 1]`?. What does `Address.FindProperty(Address)` actually return?
> 
> If trying to find a property that doesn't exist returns `null`, then next iteration of the loop will throw a `NullReferenceException`.
> 
> Is this method covered in tests? I assume it should cause tests to fail? Or I am misunderstanding this code.

Yes I agree with this.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
```diff
@@ -691,7 +691,15 @@ private void VisitPropertyRefs(IEnumerable<IEdmStructuralProperty> properties)
         {
             foreach (IEdmStructuralProperty property in properties)
             {
-                this.schemaWriter.WritePropertyRefElement(property);
+                var prop = property as IEdmStructuralPropertyAlias;
+                if (string.IsNullOrEmpty(prop?.KeyAlias))
```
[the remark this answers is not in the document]
ElizabethOkerio: I thin your suggestion still supports adding a new override of WritePropertyRefElement that takes 2 properties. 

```c sharp 

var prop = property as IEdmStructuralPropertyAlias;
            if (prop == null)
            {
                this.WriteRequiredAttribute(CsdlConstants.Attribute_Name, property.Name, EdmValueWriter.StringAsXml);
            }
            else 
            {
                this.WriteRequiredAttribute(CsdlConstants.Attribute_Name, string.Join("/", prop.Path), EdmValueWriter.StringAsXml);
                this.WriteRequiredAttribute(CsdlConstants.Attribute_Alias, prop.KeyAlias, EdmValueWriter.StringAsXml);
            }
```

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaWriter.cs
```diff
@@ -71,6 +71,7 @@ internal EdmModelCsdlSchemaWriter(IEdmModel model, Version edmVersion)
         internal abstract void WriteDeclaredKeyPropertiesElementHeader();
 
         internal abstract void WritePropertyRefElement(IEdmStructuralProperty property);
```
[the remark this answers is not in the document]
ElizabethOkerio: resolved

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
```diff
@@ -0,0 +1,59 @@
...
+// <copyright file="EdmStructuralPropertyAlias.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using System.Collections.Generic;
+
+namespace Microsoft.OData.Edm
+{
+    public class EdmStructuralPropertyAlias : EdmProperty, IEdmStructuralPropertyAlias
```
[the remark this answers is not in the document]
ElizabethOkerio: @habbes  So if we put it as PropertyRef it's like we are tying it to the <PropertyRef> element which is in most cases tied to the key property.

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <param name="path">The path to the key property.</param>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", MessageId = "string", Justification = "defaultValue might be confused for an IEdmValue.")]
+        public EdmStructuralPropertyAlias(IEdmStructuredType declaringType, string name, IEdmTypeReference type, string alias, IEnumerable<string> path)
+            : base(declaringType, name, type)
+        {
+            this.alias = alias;
+            this.path = path;
+        }
+
+        public string KeyAlias
```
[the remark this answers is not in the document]
ElizabethOkerio: @habbes Alias is a reserved keyword. Maybe call it PropertyAlias

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
```diff
@@ -0,0 +1,59 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EdmStructuralPropertyAlias"/> class.
+        /// </summary>
+        /// <param name="declaringType">The type that declares this property.</param>
+        /// <param name="name">Name of the property.</param>
+        /// <param name="type">The type of the property.</param>
+        /// <param name="alias">The alias of the key property.</param>
+        /// <param name="path">The path to the key property.</param>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", MessageId = "string", Justification = "defaultValue might be confused for an IEdmValue.")]
+        public EdmStructuralPropertyAlias(IEdmStructuredType declaringType, string name, IEdmTypeReference type, string alias, IEnumerable<string> path)
```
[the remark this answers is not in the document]
ElizabethOkerio: The type that owns the <PropertyRef> can be gotten from the path. It can be gotten from the first path segment.

src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
```diff
@@ -163,6 +163,19 @@ public EdmNavigationProperty AddUnidirectionalNavigation(EdmNavigationPropertyIn
             return property;
         }
 
+        /// <summary>
+        /// Creates and adds a structural property alias to this type.
+        /// </summary>
+        /// <param name="declaring">Name of the property.</param>
+        /// <param name="type">Type of the property.</param>
+        /// <returns>Created structural property.</returns>
+        public EdmStructuralPropertyAlias AddStructuralPropertyAlias(string name, IEdmTypeReference type, string alias, IEnumerable<string> path)
```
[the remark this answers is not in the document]
ElizabethOkerio: The reason am passing the name here is to have some uniformity with the existing methods for adding structural properties.

```
AddStructuralProperty(string name, IEdmTypeReference type)
```

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderJsonTests.cs
ElizabethOkerio: added it.

src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
```diff
@@ -163,6 +163,19 @@ public EdmNavigationProperty AddUnidirectionalNavigation(EdmNavigationPropertyIn
...
 
+        /// <summary>
+        /// Creates and adds a structural property alias to this type.
+        /// </summary>
+        /// <param name="declaring">Name of the property.</param>
+        /// <param name="type">Type of the property.</param>
+        /// <returns>Created structural property.</returns>
+        public EdmStructuralPropertyAlias AddStructuralPropertyAlias(string name, IEdmTypeReference type, string alias, IEnumerable<string> path)
+        {
+            EdmStructuralPropertyAlias property = new EdmStructuralPropertyAlias(this, name, type, alias, path);
```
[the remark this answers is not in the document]
ElizabethOkerio: this is this case is the complex type where the key property is defined. It is the containing type of the property name. 
```
**complexType**.AddStructuralPropertyAlias(EdmCoreModel.Instance.GetString(false), "ExampleObjectName", new List<string> { "ExampleObject", "Name" });
```

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +112,36 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+                                {
+                                    structuralProperty = structuralProperty.Type.AsComplex().FindProperty(keyPropertySegments[i]) as IEdmStructuralProperty;
+                                }
+                                else if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Entity)
+                                {
+                                    structuralProperty = structuralProperty.Type.AsEntity().FindProperty(keyPropertySegments[i]) as IEdmStructuralProperty;
+                                }
+                                else 
+                                {
+                                    throw new ArgumentNullException();
```
KenitoInc, line 134: We should throw an ODataException?

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaJsonWriter.cs
```diff
@@ -208,8 +208,18 @@ internal override void WritePropertyRefElement(IEdmStructuralProperty property)
         {
             // Key properties without a key alias are represented as strings containing the property name.
             // Key properties with a key alias are represented as objects with one member whose name is the key alias and whose value is a string containing the path to the property.
-            // TODO: It seems the second one is not supported.
-            this.jsonWriter.WriteStringValue(property.Name);
+            var prop = property as IEdmStructuralPropertyAlias;
```
KenitoInc, line 211:
```suggestion
            var prop = property as IEdmStructuralPropertyAlias;

```

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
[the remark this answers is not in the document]
KenitoInc: I am happy with the name `EdmStructuralPropertyAlias`

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
```diff
@@ -157,6 +158,37 @@ public void AddKeys(IEnumerable<IEdmStructuralProperty> keyProperties)
...
+                        {
+                            structuralProperty = structuralProperty.Type.AsComplex().FindProperty(keyPropertySegments[i]) as IEdmStructuralProperty;
+                        }
+                        else if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Entity)
+                        {
+                            structuralProperty = structuralProperty.Type.AsEntity().FindProperty(keyPropertySegments[i]) as IEdmStructuralProperty;
+                        }
+                        else
+                        {
+                            throw new ArgumentNullException();
```
KenitoInc, line 179: Throw an `ODataException`

PR #2168, opened 2021-08-25
Configure remote profiles for running benchmarks

### LINKED ISSUE #2167: Run benchmarks on dedicated machines
We want to run our performance benchmarks on dedicated machines to get more stable results that we can more reliably compare against each other. We can do this by configuring new profiles in the `benchmarks.yml` file. Currently we only have a `local` profile for running the benchmark suite on the local machine.

habbes:
### Issues

Fix #2167 

### Description

Added two profiles for running benchmarks: `lab-windows` and `lab-linux`. These are remote machines dedicated for running on-demand benchmarks with `Microsoft.Crank`.

They are used by the ASP.NET team for running their benchmarks, see: https://github.com/aspnet/Benchmarks/tree/main/scenarios and https://github.com/aspnet/Benchmarks/blob/main/scenarios/aspnet.profiles.yml

Here's how you'd use them to run benchmarks:

```
crank --config benchmarks.yml --scenario Components --profile lab-linux
```

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/README.md
+++ b/README.md
@@ -184,6 +184,32 @@ The easiest way to run the perf benchmarks is to use the [Microsoft.Crank](https
     crank --config benchmarks.yml --scenario UriParser --profile local
     ```
 
+#### Run benchmarks on remote dedicated agents
+
+The `local` profile is provided for testing purposes, but it's not ideal for running benchmarks.
+For more stable results and results that we can more reliably compare, the following profiles are also
+available and should be preferred whenever possible:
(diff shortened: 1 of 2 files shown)


habbes:
Here's a sample run of the benchmarks:

image

marabooy (approved):

KenitoInc (approved):

PR #2173, opened 2021-08-30
EdmLib Vocabulary Annotation Default Values

lisicase:
### Description

This pull request adds support for using default values in vocabulary annotations in EdmLib.

### Checklist

- [x] *Add default values*
- [x] *Serialization*
- [x] *Deserialization*
- [x] *Reserialization*
- [x] *Add test cases*

### Additional work necessary

- Currently only supports default values of type String -> Support all kinds of value types
- Currently introduces breaking interface change -> Turn this into an extension

### DIFF
--- a/src/Microsoft.OData.Edm/Csdl/CsdlReader.cs
+++ b/src/Microsoft.OData.Edm/Csdl/CsdlReader.cs
@@ -341,6 +341,9 @@ namespace Microsoft.OData.Edm.Csdl
                         model.SetEdmVersion(edmVersion);
                     }
 
+                    // Note: Considered manually replacing expressions after adding the terms to the model
+                    // so I could initialize them with default values. (Would need to replace because
+                    // the annotation.Value property is readonly.) However, this is somewhat expensive/complex.
                     /*IEnumerable<IEdmVocabularyAnnotation> annotations = model.VocabularyAnnotations;
                     for (int i = 0; i < annotations.Count(); i++)
                     {
--- a/src/Microsoft.OData.Edm/Csdl/CsdlReader.cs
+++ b/src/Microsoft.OData.Edm/Csdl/CsdlReader.cs
@@ -340,6 +340,18 @@ namespace Microsoft.OData.Edm.Csdl
                     {
                         model.SetEdmVersion(edmVersion);
                     }
+
+                    /*IEnumerable<IEdmVocabularyAnnotation> annotations = model.VocabularyAnnotations;
+                    for (int i = 0; i < annotations.Count(); i++)
+                    {
+                        IEdmVocabularyAnnotation currAnnotation = annotations.ElementAt(i);
+                        if (currAnnotation.UsesDefault)
+                        {
+                            IEdmVocabularyAnnotation replacementValue = new EdmExpression(...);
+                            IEdmVocabularyAnnotation replacementAnnotation = new IEdmExpression(...);
+                            annotations.SetElement(i, replacementAnnotation);
+                        }
+                    }*/
                 }
                 else
                 {
--- a/src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlAnnotation.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlAnnotation.cs
@@ -14,6 +14,7 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Ast
         private readonly CsdlExpressionBase expression;
         private readonly string qualifier;
         private readonly string term;
+        private readonly bool usesDefault;
 
         public CsdlAnnotation(string term, string qualifier, CsdlExpressionBase expression, CsdlLocation location)
             : base(location)
@@ -21,6 +22,7 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Ast
             this.expression = expression;
             this.qualifier = qualifier;
             this.term = term;
+            this.usesDefault = expression == null;
         }
 
         public CsdlExpressionBase Expression
@@ -36,6 +38,11 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Ast
         public string Term
         {
             get { return this.term; }
+        }
+
+        public bool UsesDefault
+        {
+            get { return this.usesDefault; }
         }
     }
 }
--- a/src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlAnnotation.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlAnnotation.cs
@@ -14,7 +14,7 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Ast
         private readonly CsdlExpressionBase expression;
         private readonly string qualifier;
         private readonly string term;
-        private readonly bool usesDefault;
+        private readonly bool usesDefault; // Note: I may take off this property again if it's not needed for the implementation
 
         public CsdlAnnotation(string term, string qualifier, CsdlExpressionBase expression, CsdlLocation location)
             : base(location)
@@ -22,7 +22,7 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Ast
             this.expression = expression;
             this.qualifier = qualifier;
             this.term = term;
-            this.usesDefault = expression == null;
+            this.usesDefault = expression == null; // if the expression is missing, we want the term's default value
         }
 
         public CsdlExpressionBase Expression
--- a/src/Microsoft.OData.Edm/Csdl/Parsing/CsdlDocumentParserBase.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Parsing/CsdlDocumentParserBase.cs
@@ -364,6 +364,16 @@ namespace Microsoft.OData.Edm.Csdl.Parsing
             string qualifier = Optional(CsdlConstants.Attribute_Qualifier);
             CsdlExpressionBase expression = this.ParseAnnotationExpression(element, childValues);
 
+
+            // if expression is null, set it to term's default value
+            if (expression == null)
+            {
+                //IEdmTerm defaultTerm = element.FindTerm("NS.MyDefaultTerm");
+                //Assert.Equal("This is a test", defaultTerm.DefaultValue);
+                XmlAttributeInfo termComponent = element.Attributes[CsdlConstants.Attribute_Term]; // my attempt to start getting to term's default value
+                //termComponent. // doesn't have UsesDefault
+            }
+
             return new CsdlAnnotation(term, qualifier, expression, element.Location);
         }
 
--- a/src/Microsoft.OData.Edm/Csdl/Parsing/CsdlDocumentParserBase.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Parsing/CsdlDocumentParserBase.cs
@@ -365,14 +365,13 @@ namespace Microsoft.OData.Edm.Csdl.Parsing
             CsdlExpressionBase expression = this.ParseAnnotationExpression(element, childValues);
 
 
-            // if expression is null, set it to term's default value
-            if (expression == null)
+            // Note: Attempted to set the expression to the term's default value,
+            // but there wasn't a good way to access that value -- just the term's name.
+            /* (expression == null)
             {
-                //IEdmTerm defaultTerm = element.FindTerm("NS.MyDefaultTerm");
-                //Assert.Equal("This is a test", defaultTerm.DefaultValue);
-                XmlAttributeInfo termComponent = element.Attributes[CsdlConstants.Attribute_Term]; // my attempt to start getting to term's default value
-                //termComponent. // doesn't have UsesDefault
-            }
+                XmlAttributeInfo termComponent = element.Attributes[CsdlConstants.Attribute_Term];
+                //termComponent. // doesn't have DefaultValue
+            }*/
 
             return new CsdlAnnotation(term, qualifier, expression, element.Location);
         }
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
@@ -26,6 +26,7 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
         private readonly CsdlSemanticsAnnotations annotationsContext;
         private readonly Cache<CsdlSemanticsVocabularyAnnotation, IEdmExpression> valueCache = new Cache<CsdlSemanticsVocabularyAnnotation, IEdmExpression>();
         private static readonly Func<CsdlSemanticsVocabularyAnnotation, IEdmExpression> ComputeValueFunc = (me) => me.ComputeValue();
+        private readonly bool usesDefault;
 
         private readonly Cache<CsdlSemanticsVocabularyAnnotation, IEdmTerm> termCache = new Cache<CsdlSemanticsVocabularyAnnotation, IEdmTerm>();
         private static readonly Func<CsdlSemanticsVocabularyAnnotation, IEdmTerm> ComputeTermFunc = (me) => me.ComputeTerm();
@@ -43,6 +44,7 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
             this.qualifier = qualifier ?? annotation.Qualifier;
             this.targetContext = targetContext;
             this.annotationsContext = annotationsContext;
+            this.usesDefault = false;
         }
 
         public CsdlSemanticsSchema Schema { get; }
@@ -55,6 +57,15 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
         public string Qualifier
         {
             get { return this.qualifier; }
+        }
+
+        /// <summary>
+        /// Gets whether the annotation uses a default value
+        /// (not defined with a provided value).
+        /// </summary>
+        public bool UsesDefault
+        {
+            get { return this.usesDefault; }
         }
 
         public override CsdlSemanticsModel Model
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
@@ -26,7 +26,6 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
         private readonly CsdlSemanticsAnnotations annotationsContext;
         private readonly Cache<CsdlSemanticsVocabularyAnnotation, IEdmExpression> valueCache = new Cache<CsdlSemanticsVocabularyAnnotation, IEdmExpression>();
         private static readonly Func<CsdlSemanticsVocabularyAnnotation, IEdmExpression> ComputeValueFunc = (me) => me.ComputeValue();
-        private readonly bool usesDefault;
 
         private readonly Cache<CsdlSemanticsVocabularyAnnotation, IEdmTerm> termCache = new Cache<CsdlSemanticsVocabularyAnnotation, IEdmTerm>();
         private static readonly Func<CsdlSemanticsVocabularyAnnotation, IEdmTerm> ComputeTermFunc = (me) => me.ComputeTerm();
@@ -44,7 +43,6 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
             this.qualifier = qualifier ?? annotation.Qualifier;
             this.targetContext = targetContext;
             this.annotationsContext = annotationsContext;
-            this.usesDefault = false;
         }
 
         public CsdlSemanticsSchema Schema { get; }
@@ -65,7 +63,7 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
         /// </summary>
         public bool UsesDefault
         {
-            get { return this.usesDefault; }
+            get { return this.Annotation.UsesDefault; }
         }
 
         public override CsdlSemanticsModel Model
@@ -138,14 +136,29 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
         private IEdmExpression ComputeValue()
         {
             IEdmTypeReference termType = Term is UnresolvedVocabularyTerm ? null : Term.Type;
-            CsdlExpressionBase adjustedExpression = AdjustStringConstantUsingTermType((this.Annotation).Expression, termType);
+            /*CsdlExpressionBase adjustedExpression;
+            if (!this.Annotation.UsesDefault)
+            {
+                adjustedExpression = AdjustStringConstantUsingTermType((this.Annotation).Expression, termType);
+            } else
+            {
+                adjustedExpression = AdjustStringConstantUsingTermType(this.Term.DefaultValue, termType);
+            }*/
 
+            /*if (!this.Annotation.UsesDefault)
+            {
+                var defaultExpression = new CsdlSemanticsStringConstantExpression(Term.DefaultValue, this.Schema);
+                return CsdlSemanticsModel.WrapExpression(defaultExpression, TargetBindingContext, this.Schema);
+            }*/
+
+            CsdlExpressionBase adjustedExpression = AdjustStringConstantUsingTermType((this.Annotation).Expression, termType);
             return CsdlSemanticsModel.WrapExpression(adjustedExpression, TargetBindingContext, this.Schema);
         }
 
         private static CsdlExpressionBase AdjustStringConstantUsingTermType(CsdlExpressionBase expression, IEdmTypeReference termType)
         {
             if (expression == null || termType == null)
+            // if (expression == null || (expression != null && termType == null))
             {
                 return expression;
             }
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
@@ -136,15 +136,7 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
         private IEdmExpression ComputeValue()
         {
             IEdmTypeReference termType = Term is UnresolvedVocabularyTerm ? null : Term.Type;
-            /*CsdlExpressionBase adjustedExpression;
-            if (!this.Annotation.UsesDefault)
-            {
-                adjustedExpression = AdjustStringConstantUsingTermType((this.Annotation).Expression, termType);
-            } else
-            {
-                adjustedExpression = AdjustStringConstantUsingTermType(this.Term.DefaultValue, termType);
-            }*/
-
+            // Note: Attempted to create a wrapped version for the term's default value, but could not access said value.
             /*if (!this.Annotation.UsesDefault)
             {
                 var defaultExpression = new CsdlSemanticsStringConstantExpression(Term.DefaultValue, this.Schema);
@@ -157,9 +149,7 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
 
         private static CsdlExpressionBase AdjustStringConstantUsingTermType(CsdlExpressionBase expression, IEdmTypeReference termType)
         {
-            if (expression == null || termType == null)
-            // if (expression == null || (expression != null && termType == null))
-            {
+            if (expression == null || termType == null)            {
                 return expression;
             }
 
--- a/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
@@ -504,7 +504,10 @@ namespace Microsoft.OData.Edm.Csdl.Serialization
             this.WriteOptionalAttribute(CsdlConstants.Attribute_Qualifier, annotation.Qualifier, EdmValueWriter.StringAsXml);
             if (isInline)
             {
-                if (!annotation.UsesDefault) // interface doesn't have UsesDefault -> I added it
+                if (!annotation.UsesDefault)
+                // Note: If I recall correctly, casting didn't work here (it caused problems in other tests),
+                // so I added UsesDefault to the interface. This is a breaking change, so it'll need to become
+                // an interface extension or similar in the future.
                 {
                     this.WriteInlineExpression(annotation.Value);
                 }                
--- a/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
@@ -504,7 +504,10 @@ namespace Microsoft.OData.Edm.Csdl.Serialization
             this.WriteOptionalAttribute(CsdlConstants.Attribute_Qualifier, annotation.Qualifier, EdmValueWriter.StringAsXml);
             if (isInline)
             {
-                this.WriteInlineExpression(annotation.Value);
+                if (!annotation.UsesDefault) // interface doesn't have UsesDefault -> I added it
+                {
+                    this.WriteInlineExpression(annotation.Value);
+                }                
             }
         }
 
--- a/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
@@ -344,7 +344,16 @@ namespace Microsoft.OData.Edm.Csdl.Serialization
 
         protected override void ProcessAnnotation(IEdmVocabularyAnnotation annotation)
         {
-            bool isInline = IsInlineExpression(annotation.Value);
+            bool isInline;
+            // Note: This currently needs to be separated out because CsdlSemanticsVocabularyAnnotation
+            // doesn't return the default value from Value at the moment.
+            if (!annotation.UsesDefault)
+            {
+                isInline = IsInlineExpression(annotation.Value);
+            } else
+            {
+                isInline = IsInlineExpression(new EdmStringConstant(annotation.Term.DefaultValue));
+            }
             this.BeginElement(annotation, t => this.schemaWriter.WriteVocabularyAnnotationElementHeader(t, isInline));
             if (!isInline)
             {
--- a/src/Microsoft.OData.Edm/EdmModelVisitor.cs
+++ b/src/Microsoft.OData.Edm/EdmModelVisitor.cs
@@ -642,7 +642,13 @@ namespace Microsoft.OData.Edm
         protected virtual void ProcessAnnotation(IEdmVocabularyAnnotation annotation)
         {
             this.ProcessVocabularyAnnotation(annotation);
-            this.VisitExpression(annotation.Value);
+            if (!annotation.UsesDefault)
+            {
+                this.VisitExpression(annotation.Value);
+            } else
+            {
+                this.VisitExpression(new EdmStringConstant(annotation.Term.DefaultValue));
+            }
         }
 
         protected virtual void ProcessPropertyValueBinding(IEdmPropertyValueBinding binding)
--- a/src/Microsoft.OData.Edm/Validation/ValidationRules.cs
+++ b/src/Microsoft.OData.Edm/Validation/ValidationRules.cs
@@ -2485,15 +2485,34 @@ namespace Microsoft.OData.Edm.Validation
                 (context, annotation) =>
                 {
                     IEnumerable<EdmError> errors;
-                    if (!annotation.Value.TryCast(annotation.Term.Type, out errors))
+                    if (!annotation.UsesDefault)
                     {
-                        foreach (EdmError error in errors)
+                        if (!annotation.Value.TryCast(annotation.Term.Type, out errors))
                         {
-                            if (error.ErrorCode != EdmErrorCode.RecordExpressionMissingRequiredProperty)
+                            foreach (EdmError error in errors)
                             {
-                                context.AddError(error);
+                                if (error.ErrorCode != EdmErrorCode.RecordExpressionMissingRequiredProperty)
+                                {
+                                    context.AddError(error);
+                                }
                             }
                         }
+                    }
+                });
+
+        /// <summary>
+        /// Validates that if a vocabulary annotation doesn't have a value, it should declare a type with a default value.
+        /// </summary>
+        public static readonly ValidationRule<IEdmVocabularyAnnotation> VocabularyAnnotationUseDefaultTermMustHaveDefaultValue =
+            new ValidationRule<IEdmVocabularyAnnotation>(
+                (context, annotation) =>
+                {
+                    if (annotation.UsesDefault && annotation.Term.DefaultValue == null)
+                    {
+                        context.AddError(
+                            annotation.Location(),
+                            EdmErrorCode.ExpressionMissingValueOrTermWithDefaultValue,
+                            "Annotation expressions must specify a value or use a term with a specified default value.");
                     }
                 });
 
--- a/src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
+++ b/src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
@@ -52,17 +52,17 @@ namespace Microsoft.OData.Edm.Vocabularies
         }
 
         /// <summary>
-        /// NEW: Initializes a new instance of the EdmVocabularyAnnotation class
+        /// Initializes a new instance of the EdmVocabularyAnnotation class
         /// to the default value.
         /// </summary>
         /// <param name=“target”>Element the annotation applies to.</param>
         /// <param name=“term”>Term bound by the annotation</param>
         public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term)
-            //: this(target, term, null, new EdmStringConstant(term.DefaultValue))
         {
             // Check arguments
             if (term.DefaultValue == null) // throw error if no default value
             {
+                // Note: I'm not sure which is the better way of throwing the exception:
                 //throw new ArgumentNullException("value");
                 EdmUtil.CheckArgumentNull(new EdmStringConstant(null), "value");
             }
@@ -77,31 +77,35 @@ namespace Microsoft.OData.Edm.Vocabularies
             this.target = target;
             this.term = term;
             this.qualifier = null;
-            //this.value = BuildDefaultValue(term);
-            this.value = new EdmStringConstant(term.DefaultValue);
+            //this.value = BuildDefaultValue(term); // Note: Eventually this method will parse the default value
+            this.value = new EdmStringConstant(term.DefaultValue); // Note: temporarily only supporting Strings
             this.usesDefault = true;
         }
 
+        /// <summary>
+        /// Parses a <paramref name="defaultValue"/> into an IEdmExpression value of the correct type.
+        /// </summary>
+        /// <param name=“typeReference”>The type of value.</param>
+        /// <param name=“defaultValue”>Original default value.</param>
+        /// <returns>An IEdmExpression of type <paramref name="typeReference".</paramref></returns>
         /*private static IEdmExpression BuildDefaultValue(IEdmTypeReference typeReference, string defaultValue)
         {
             //string defaultValue = term.DefaultValue;
             EdmTypeKind termTypeKind = typeReference.TypeKind();
 
+            // Create constants for the corresponding types
             switch(termTypeKind)
             {
                 case EdmTypeKind.Primitive:
                     return BuildPrimitiveValue(typeReference.AsPrimitive(), defaultValue);
-
                 case EdmTypeKind.Enum:
-
-                    // EdmEnumMemberExpression
-                    // do the enum express
-                    return new EdmEnumMemberExpression(....);
-
+                    // TODO: return new EdmEnumMemberExpression(....);
                 case EdmTypeKind.Complex:
+                    // TODO
                 case EdmTypeKind.Entity:
+                    // Example:
                     // {
-                    //   "City":"Remodn",
+                    //   "City":"Redmond",
                     //   "Street": "156TH, AVE..."
                     // }
                     var properties = ParseObject(defaultValue);
@@ -114,16 +118,14 @@ namespace Microsoft.OData.Edm.Vocabularies
 
                     // build the properties expression one by one
                     return new EdmRecordExpression(....);
-
                 case EdmTypeKind.Collection:
                     IEdmCollectionTypeReference collectionType = (IEdmCollectionTypeReference)typeReference;
                     IEdmTypeReference elementType = collectionType.ElementType();
-                    // Be sure is it in the scope?
-                    // If ti's in the scope
+                    // If it's in the scope:
                     // term.DefaultValue could be a JSON string
                     // [5, 6, 9]
-                    // [{"city":"red,[]mond", },{}]
-                    // Have to parse the JSON array into Items
+                    // [{"city":["redmond","seattle"]},{}]
+                    // Have to recursively parse the JSON array into items
 
                     var items = ParseCollection(defaultValue);
                     IList<IEdmExpression> itemExpressions = new List<IEdmExpression>();
@@ -135,40 +137,18 @@ namespace Microsoft.OData.Edm.Vocabularies
 
                     return new EdmCollectionExpression(elementType, itemExpressions);
 
-                case EdmTypeKind.TypeDefinition:
-                    ....
-
-
-            }*/
-
-            //return new EdmNavigationPropertyPathExpression(defaultValue);
-
-            //IEdmExpression defaultValueExp = new EdmStringConstant(defaultValue);
-
-            /*if (object.Equals(termTypeKind, IEdmExpression.BinaryConstant))
-            {
-                return new EdmBinaryConstant(System.Binary.Parse(defaultValue));
+                //case EdmTypeKind.TypeDefinition:
+                //    ....
+                // Note: Need to have cases for the following types:
+                // DateTimeOffsetConstant, DecimalConstant, FloatingConstant, GuidConstant, IntegerConstant,
+                // StringConstant, DurationConstant, Null, Record, Collection, Path, If, Cast, IsType,
+                // FunctionApplication, LabeledExpressionReference, Labeled, PropertyPath, NavigationPropertyPath,
+                // DateConstant, TimeOfDayConstant, EnumMember, AnnotationPath
             }
-            else if (object.Equals(termTypeKind, IEdmExpression.BooleanConstant))
-            {
-                return new EdmBooleanConstant(System.Boolean.Parse(defaultValue));
-            }
-            // ...
-            else
-            {
-                return new EdmStringConstant(defaultValue);
-            }
-
-
-            // DateTimeOffsetConstant, DecimalConstant, FloatingConstant, GuidConstant, IntegerConstant,
-            // StringConstant, DurationConstant, Null, Record, Collection, Path, If, Cast, IsType,
-            // FunctionApplication, LabeledExpressionReference, Labeled, PropertyPath, NavigationPropertyPath,
-            // DateConstant, TimeOfDayConstant, EnumMember, AnnotationPath
-
-            //return defaultValueExp;
         }*/
 
-        /* static IDictionary<string, string> ParseObject(string defaultValue)
+        // Parse an object from a string
+        /*private static IDictionary<string, string> ParseObject(string defaultValue)
         {
             // parse the JSON array into items
             // "[5, 6, 8]"  => return "5", "6", "8"
@@ -176,6 +156,7 @@ namespace Microsoft.OData.Edm.Vocabularies
             return null;
         }
 
+        // Parse a collection from a string
         private static IEnumerable<string> ParseCollection(string defaultValue)
         {
             // parse the JSON array into items
@@ -184,6 +165,7 @@ namespace Microsoft.OData.Edm.Vocabularies
             return null;
         }*/
 
+        // Return an IEdmExpression for a primitive value from a string
         /*private static IEdmExpression BuildPrimitiveValue(IEdmPrimitiveTypeReference reference, string defaultValue)
         {
             switch (reference.PrimitiveKind())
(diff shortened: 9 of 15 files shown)


src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlAnnotation.cs
```diff
@@ -14,13 +14,15 @@ internal class CsdlAnnotation : CsdlElement
...
         public CsdlAnnotation(string term, string qualifier, CsdlExpressionBase expression, CsdlLocation location)
             : base(location)
         {
             this.expression = expression;
             this.qualifier = qualifier;
             this.term = term;
+            this.usesDefault = expression == null; // if the expression is missing, we want the term's default value
         }
 
         public CsdlExpressionBase Expression
```
mikepizzo, line 28: >public CsdlExpressionBase Expression [](http://example.com/codeflow?start=8&length=36)

should add a comment that this will be null if UsesDefault == true. Which means, for this internal class, we may not need the UsesDefault property -- we can just check if Expression == null.  Since it's internal we can update any logic that accesses Expression to understand that null == usesdefault.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
```diff
@@ -127,15 +136,20 @@ protected IEdmTerm ComputeTerm()
         private IEdmExpression ComputeValue()
         {
             IEdmTypeReference termType = Term is UnresolvedVocabularyTerm ? null : Term.Type;
-            CsdlExpressionBase adjustedExpression = AdjustStringConstantUsingTermType((this.Annotation).Expression, termType);
+            // Note: Attempted to create a wrapped version for the term's default value, but could not access said value.
+            /*if (!this.Annotation.UsesDefault)
+            {
+                var defaultExpression = new CsdlSemanticsStringConstantExpression(Term.DefaultValue, this.Schema);
+                return CsdlSemanticsModel.WrapExpression(defaultExpression, TargetBindingContext, this.Schema);
+            }*/
```
mikepizzo, line 144: These seems the appropriate place to put the logic to compute the default value. Was the issue that the term was not found, or that Term.DefaultValue had not been set?

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
```diff
@@ -344,7 +344,16 @@ protected override void ProcessFunctionImport(IEdmFunctionImport functionImport)
 
         protected override void ProcessAnnotation(IEdmVocabularyAnnotation annotation)
         {
-            bool isInline = IsInlineExpression(annotation.Value);
+            bool isInline;
+            // Note: This currently needs to be separated out because CsdlSemanticsVocabularyAnnotation
+            // doesn't return the default value from Value at the moment.
```
mikepizzo, line 349: Right. This should go away if we can fix the code that returns the default value in annotation.Value.

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
```diff
@@ -2485,18 +2485,37 @@ public static class ValidationRules
                 (context, annotation) =>
                 {
                     IEnumerable<EdmError> errors;
-                    if (!annotation.Value.TryCast(annotation.Term.Type, out errors))
+                    if (!annotation.UsesDefault)
                     {
```
mikepizzo, line 2489: should not need this once we fix it so that annotation.Value always returns a value (either explicit or default)

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
```diff
@@ -2485,18 +2485,37 @@ public static class ValidationRules
...
                             }
                         }
                     }
                 });
 
+        /// <summary>
+        /// Validates that if a vocabulary annotation doesn't have a value, it should declare a type with a default value.
+        /// </summary>
+        public static readonly ValidationRule<IEdmVocabularyAnnotation> VocabularyAnnotationUseDefaultTermMustHaveDefaultValue =
+            new ValidationRule<IEdmVocabularyAnnotation>(
```
mikepizzo, line 2507: this rule will have to be updated to use our new interface that contains UsesDefault; otherwise rule looks great.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -44,8 +48,213 @@ public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term,
...
+            this.target = target;
+            this.term = term;
+            this.qualifier = null;
+            //this.value = BuildDefaultValue(term); // Note: Eventually this method will parse the default value
+            this.value = new EdmStringConstant(term.DefaultValue); // Note: temporarily only supporting Strings
+            this.usesDefault = true;
+        }
+
+        /// <summary>
+        /// Parses a <paramref name="defaultValue"/> into an IEdmExpression value of the correct type.
```
mikepizzo, line 86: We might want to move this logic (once it works) to a general helper class so that it can be called from multiple locations (i.e., EdmVocabularyAnnotation as well as CsdlSemanticsVocabularyAnnotation)

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -44,8 +48,213 @@ public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term,
...
+            return null;
+        }*/
+
+        // Return an IEdmExpression for a primitive value from a string
+        /*private static IEdmExpression BuildPrimitiveValue(IEdmPrimitiveTypeReference reference, string defaultValue)
+        {
+            switch (reference.PrimitiveKind())
+            {
+                case EdmPrimitiveTypeKind.Binary:
+                    byte[] binary = new byte[0];
```
mikepizzo, line 174: > = new byte[0] [](http://example.com/codeflow?start=33&length=14)

shouldn't have to initialize -- can just do byte[] binary;

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -44,8 +48,213 @@ public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term,
...
+        /// <param name=“target”>Element the annotation applies to.</param>
+        /// <param name=“term”>Term bound by the annotation</param>
+        public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term)
+        {
+            // Check arguments
+            if (term.DefaultValue == null) // throw error if no default value
+            {
+                // Note: I'm not sure which is the better way of throwing the exception:
+                //throw new ArgumentNullException("value");
+                EdmUtil.CheckArgumentNull(new EdmStringConstant(null), "value");
```
mikepizzo, line 67: CheckArgumentNull is used for parameters passed into the public method.  So, in this case, better to throw an OData exception than cal CheckArgumentNull.

src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmVocabularyAnnotation.cs
```diff
@@ -30,5 +30,11 @@ public interface IEdmVocabularyAnnotation : IEdmElement
         /// Gets the expression producing the value of the annotation.
         /// </summary>
         IEdmExpression Value { get; }
+
+        /// <summary>
+        /// Gets whether the annotation uses a default value
+        /// (not defined with a provided value).
+        /// </summary>
+        bool UsesDefault { get; }
```
mikepizzo, line 38: as per separate comment, changing a public interface is a breaking change (it breaks any apps that may have implemented the interface, since they won't have implemented this property) so we have to define  new interface that derives from IEdmVocabularyAnnotation and adds the UsesDefault method.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
lisicase: I had thought that I hadn't been able to access the default value, just the term, but it seems to be fine now. It's now working so that a CSDL annotation's value is the default value when applicable!

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
lisicase: Done! ✅

src/Microsoft.OData.Edm/EdmModelVisitor.cs
```diff
@@ -642,7 +642,13 @@ protected virtual void ProcessImmediateValueAnnotation(IEdmDirectValueAnnotation
...
         {
             this.ProcessVocabularyAnnotation(annotation);
-            this.VisitExpression(annotation.Value);
+            if (!annotation.UsesDefault)
+            {
+                this.VisitExpression(annotation.Value);
+            } else
+            {
+                this.VisitExpression(new EdmStringConstant(annotation.Term.DefaultValue));
+            }
```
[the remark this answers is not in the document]
lisicase: Done ✅

src/Microsoft.OData.Edm/Csdl/Parsing/CsdlDocumentParserBase.cs
```diff
@@ -364,6 +364,15 @@ private CsdlAnnotation OnAnnotationElement(XmlElementInfo element, XmlElementVal
...
             CsdlExpressionBase expression = this.ParseAnnotationExpression(element, childValues);
 
+
+            // Note: Attempted to set the expression to the term's default value,
+            // but there wasn't a good way to access that value -- just the term's name.
+            /* (expression == null)
+            {
+                XmlAttributeInfo termComponent = element.Attributes[CsdlConstants.Attribute_Term];
+                //termComponent. // doesn't have DefaultValue
+            }*/
```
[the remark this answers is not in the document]
lisicase: Yep, as in a response to a different comment! ✅

src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlAnnotation.cs
lisicase: That makes sense to reduce unnecessary fields! Done ✅

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
```diff
@@ -504,7 +504,13 @@ internal override void WriteVocabularyAnnotationElementHeader(IEdmVocabularyAnno
             this.WriteOptionalAttribute(CsdlConstants.Attribute_Qualifier, annotation.Qualifier, EdmValueWriter.StringAsXml);
             if (isInline)
             {
-                this.WriteInlineExpression(annotation.Value);
+                if (!annotation.UsesDefault)
+                // Note: If I recall correctly, casting didn't work here (it caused problems in other tests),
+                // so I added UsesDefault to the interface. This is a breaking change, so it'll need to become
+                // an interface extension or similar in the future.
```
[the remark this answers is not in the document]
lisicase: I've now implemented the interface! ✅

src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmVocabularyAnnotation.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
lisicase: Should I use the new interface in the header (as in ValidationRule<IEdmVocabularyAnnotationWithDefault>) or cast in the body?

mikepizzo: hmmm... you should be able to use the new interface in the validation rule as long as the interface is declared public.

lisicase: Okay, sounds good, thanks! ✅

src/Microsoft.OData.Edm/Microsoft.OData.Edm.csproj
```diff
@@ -95,6 +95,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.Data.OData" Version="5.8.4" />
```
xuzhg, line 98: >Include="Microsoft.Data.OData" Version="5.8.4" /> [](http://example.com/codeflow?start=22&length=49)

Why do we add dependency for OData v3 lib? It doesn't make sense.

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
```diff
@@ -2497,6 +2497,22 @@ public static class ValidationRules
...
+        public static readonly ValidationRule<IEdmVocabularyAnnotationWithDefault> VocabularyAnnotationUseDefaultTermMustHaveDefaultValue =
+            new ValidationRule<IEdmVocabularyAnnotationWithDefault>(
+                (context, annotation) =>
+                {
+                    if (annotation.UsesDefault && annotation.Term.DefaultValue == null)
+                    {
+                        context.AddError(
+                            annotation.Location(),
+                            EdmErrorCode.ExpressionMissingValueOrTermWithDefaultValue,
+                            "Annotation expressions must specify a value or use a term with a specified default value.");
```
xuzhg, line 2512: >Annotation expressions must specify a value or use a term with a specified default value. [](http://example.com/codeflow?start=29&length=89)

Move the error message into the .txt file.

src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmVocabularyAnnotationWithDefault.cs
```diff
@@ -0,0 +1,20 @@
...
+    /// <summary>
+    /// Represents an EDM vocabulary annotation that may use a term's default value as its value.
+    /// </summary>
+    public interface IEdmVocabularyAnnotationWithDefault : IEdmVocabularyAnnotation
+    {
+        /// <summary>
+        /// Gets whether the annotation uses a default value
+        /// (not defined with a provided value).
+        /// </summary>
+        bool UsesDefault { get; }
```
xuzhg, line 18: >bool UsesDefault { get; } [](http://example.com/codeflow?start=8&length=25)

I still save my opinion that we could not need this interface. And Only use the Value to identify whether it's using default value or not.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -77,5 +289,13 @@ public IEdmExpression Value
         {
             get { return this.value; }
         }
+
+        /// <summary>
+        /// Gets whether the annotation uses a default value
+        /// (not defined with a provided value).
+        /// </summary>
+        public bool UsesDefault {
```
xuzhg, line 297: >public bool UsesDefault { [](http://example.com/codeflow?start=8&length=25)

Since it's only "Getter". So, I don't think we need such interface.

lisicase: I think the main reason behind this was to allow us to support null values in the future.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -44,8 +49,215 @@ public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term,
...
+            if (term.DefaultValue == null) // throw error if no default value
+            {
+                // Note: why is this string not okay but the one in LiteralFormatter.cs is?
+                // Should I create the following in Microsoft.OData.Core.cs?
+                // internal const string EdmVocabularyAnnotations_DidNotFindDefaultValue = "EdmVocabularyAnnotations_DidNotFindDefaultValue";
+                //throw new ODataException("Type name should not be null or empty when serializing an Enum value for URI key.");
+                EdmUtil.CheckArgumentNull(new EdmStringConstant(null), "value");
+            }
+            EdmUtil.CheckArgumentNull(target, "target");
+            EdmUtil.CheckArgumentNull(term, "term");
```
xuzhg, line 73: >EdmUtil.CheckArgumentNull(term, "term"); [](http://example.com/codeflow?start=12&length=40)

EdmUtil.CheckArgumentNull(term, "term"); should go before 

if (term.DefaultValue == null)
{...}

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -44,8 +49,215 @@ public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term,
...
+
+            // Check arguments
+            EdmUtil.CheckArgumentNull(target, "target");
+            EdmUtil.CheckArgumentNull(term, "term");
```
xuzhg, line 77: remove duplicated.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
xuzhg: Yes. We can supply two set of constructor.

1) The constructor without "IEdmExpression", it means to use the default value from Term.

```C#
public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term)
public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term, string qualifier)
{
      useDefault = true;
}
```

2) The constructor with "IEdmExpression", it means to use the value from customer.
     We can also change the verification logic in these constructor to accept "null" value. The only difference is "useDefault".
     In the above, the useDefault = true
     in this case, the useDefault = false.     

```C#
public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term, IEdmExpression value)
public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term, string qualifier, IEdmExpression value)
{
      useDefault = false;
}
```

We can public "UseDefault" property, but, it seems there's no needed to add it into the interface, or create a new interface. That's my opinion.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
lisicase: Ah, I think I see what you mean -- rather than add a getter method for UsesDefault (and therefore a new interface), simply make the property public.

src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmVocabularyAnnotationWithDefault.cs
mikepizzo: For backward compatibility, we need the Value property to always return a non-null value (i.e., the value of the annotation, whether it is explicitly specified or taken from the default).  Thus, in order to round-trip, we need the ability to separately track whether the value was explicit or taken from the default.

The setting of this is in the constructor, so the property can be read only. Same is true for the value, term, target etc. -- all of these are specified when we create the annotation and are immutable once created.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
mikepizzo: Adding a public property on EdmVocabularyAnnotation works for cases where *we* create the type representing the annotation (i.e., we could attempt to cast to EdmVocabularyAnnotation when we serialize, in order to read the UsesDefault property), but that doesn't allow other model implementations (CsdlVocabularyAnnotation or other custom implementations of IEdmVocabularyAnnottaion) to have a way to say that the value uses the default. The reason we define interfaces, along side the Edm* types, is so that we can support developers providing their own custom implementations.

In general, all of our deserialization should go through common, public interfaces, rather than rely on private implementation.

So, I would keep the new interface and implement it on EdmVocabularyAnnotation and CsdlVocabularyAnnotation, and check for it in the serializer.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
lisicase: Good catch! Done ✅

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Microsoft.OData.Edm.csproj
lisicase: I'm not sure how this line got added. Should I just remove it?

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Csdl/CsdlReader.cs
```diff
@@ -340,6 +340,21 @@ private bool TryParse(IEnumerable<IEdmModel> referencedModels, bool includeDefau
...
+                    for (int i = 0; i < annotations.Count(); i++)
+                    {
+                        IEdmVocabularyAnnotation currAnnotation = annotations.ElementAt(i);
+                        if (currAnnotation.UsesDefault)
+                        {
+                            IEdmVocabularyAnnotation replacementValue = new EdmExpression(...);
+                            IEdmVocabularyAnnotation replacementAnnotation = new IEdmExpression(...);
+                            annotations.SetElement(i, replacementAnnotation);
+                        }
+                    }*/
```
mikepizzo, line 357: Should remove this since it's not part of the final solution.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
```diff
@@ -504,7 +504,13 @@ internal override void WriteVocabularyAnnotationElementHeader(IEdmVocabularyAnno
             this.WriteOptionalAttribute(CsdlConstants.Attribute_Qualifier, annotation.Qualifier, EdmValueWriter.StringAsXml);
             if (isInline)
             {
-                this.WriteInlineExpression(annotation.Value);
+                if (annotation is IEdmVocabularyAnnotationWithDefault annotationWithDefault)
```
mikepizzo, line 507: >if (annotation is IEdmVocabularyAnnotationWithDefault annotationWithDefault) [](http://example.com/codeflow?start=16&length=76)

doesn't this result in the annotation value not being written if the annotation is not IEdmVocabularyAnnotationWithDefault? (i.e., it doesn't use default but is an older implementation so it doesn't implement the new interface)

src/Microsoft.OData.Edm/Csdl/CsdlReader.cs
lisicase: Good catch, thanks! ✅

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -44,8 +47,45 @@ public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term,
...
+        /// <summary>
+        /// NEW: Initializes a new instance of the EdmVocabularyAnnotation class 
+        /// to the default value.
+        /// </summary>
+        /// <param name=“target”>Element the annotation applies to.</param>
+        /// <param name=“term”>Term bound by the annotation</param>
+        /// <param name=“qualifier”>Qualifier used to distinguish bindings.</param>
+        public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term,
+        string qualifier)
+        { }*/
```
mikepizzo, line 87: so what is the solution if someone wants to build an EdmVocabularyAnnotation with a qualifier that uses a default value?

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
lisicase: Very good point! The tests were fine, so I missed that, but that would be a breaking change -- I've now updated it, thanks! ✅

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
```diff
@@ -2214,6 +2214,323 @@ public void CanWriteAnnotationPathExpression()
...
+            annotation = new EdmVocabularyAnnotation(complex, term2);
+            Assert.True(annotation.UsesDefault);
+            Assert.Equal(annotation.Term.DefaultValue, ((EdmStringConstant)annotation.Value).Value);
+            annotation.SetSerializationLocation(model, EdmVocabularyAnnotationSerializationLocation.Inline);
+            model.SetVocabularyAnnotation(annotation);
+
+            annotation = new EdmVocabularyAnnotation(complex, term3);
+            Assert.True(annotation.UsesDefault);
+            Assert.True(bool.Parse(annotation.Term.DefaultValue));
+            annotation.SetSerializationLocation(model, EdmVocabularyAnnotationSerializationLocation.Inline);
```
mikepizzo, line 2246: >        annotation.SetSerializationLocation(model, EdmVocabularyAnnotationSerializationLocation.Inline); [](http://example.com/codeflow?start=4&length=104)

add a variant of the test that doesn't set location to inline. currently I think that test will fail (i.e., it will write the default value in xml).

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -44,8 +47,45 @@ public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term,
...
+            {
+                throw new InvalidOperationException(Strings.EdmVocabularyAnnotations_DidNotFindDefaultValue(term.Name));
+            }
+
+            // Initialize
+            this.target = target;
+            this.term = term;
+            this.qualifier = null;
+            this.value = EdmTermExtensions.BuildDefaultEdmExpression(term.Type, term.DefaultValue);
+            this.usesDefault = true;
```
mikepizzo, line 74: this constructor copies most of the logic from the previous constructor. not a big deal, since the logic is neither complex nor long, but consider centralizing constructor logic, either by having one constructor call another with default values (as is done on line 29) or by having a common "initialize" method called by constructors.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
lisicase: Ah yeah, I must have missed this from my draft -- I had been wondering if someone had any advice there, but I have an idea I will submit.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
lisicase: Okay, thanks! I had noticed that as well, but since I can't call another constructor in the middle of the constructor body, I hadn't done that -- but it makes sense to just create a helper method!

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
```diff
@@ -0,0 +1,446 @@
...
+using System.IO;
+using System.Text;
+using Microsoft.OData.Edm.Csdl.Parsing.Ast;
+
+namespace Microsoft.OData.Edm.Vocabularies
+{
+    internal static class EdmTermExtensions
+    {
+        /// <summary>
+        /// Parses a <paramref name="defaultValue"/> into an IEdmExpression value of the correct EDM type.
```
mikepizzo, line 19: >Parses a <paramref name="defaultValue"/> into an IEdmExpression value of the correct EDM type. [](http://example.com/codeflow?start=12&length=94)

More generically, this method Parses a *string value* into an IEdmExpression value of the correct EDM type. takes a string value and generates an IEdmExpression.  As you are currently using it the string represents a default value, but perhaps it could be used other places to translate some other type of string to an expreession.  Perhaps name it more generically; i.e., just BuildEdmExpression(IEdmTypeReference typeReference, string value).

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
```diff
@@ -0,0 +1,446 @@
...
+                    // return BuildEdmCollectionExp(typeReference, defaultValue);
+                case EdmTypeKind.Path:
+                    // return BuildEdmPathExp(typeReference.AsPath(), defaultValue);
+                case EdmTypeKind.TypeDefinition:
+                    // return BuildEdmTypeDefinitionExp(typeReference.AsTypeDefinition(), defaultValue);
+                    throw Error.NotImplemented();
+                case EdmTypeKind.EntityReference:
+                case EdmTypeKind.Untyped:
+                default:
+                    throw Error.NotSupported();
```
mikepizzo, line 48: >throw Error.NotSupported(); [](http://example.com/codeflow?start=20&length=27)

do you want it to throw an exception or return null?

note that, by spec, reading an annotation value should not throw an exception, so perhaps we should just return null here (and in line 44) and make sure we handle the null case in the caller.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
```diff
@@ -0,0 +1,446 @@
...
+        {
+            EdmTypeKind termTypeKind = typeReference.TypeKind();
+
+            // Create expressions/constants for the corresponding types
+            switch (termTypeKind) // TODO: Create test cases for all of these
+            {
+                case EdmTypeKind.Primitive:
+                    return BuildEdmPrimitiveValueExp(typeReference.AsPrimitive(), defaultValue);
+                case EdmTypeKind.Enum:
+                    // return BuildEdmEnumExp(typeReference.AsEnum(), defaultValue)
```
mikepizzo, line 34: >BuildEdmEnumExp [](http://example.com/codeflow?start=30&length=15)

you have an implementation for this method; does it work? can you uncomment this line?

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
lisicase: I think I called it "BuildDefaultEdmExpression" to distinguish between that and "BuildDefaultCsdlExpression" since I need a different number of parameters for the different constant expression types they return.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
```diff
@@ -0,0 +1,446 @@
...
+                case EdmPrimitiveTypeKind.GeometryCollection:
+                case EdmPrimitiveTypeKind.GeometryMultiPolygon:
+                case EdmPrimitiveTypeKind.GeometryMultiLineString:
+                case EdmPrimitiveTypeKind.GeometryMultiPoint:
+                case EdmPrimitiveTypeKind.PrimitiveType:
+                case EdmPrimitiveTypeKind.None:
+                default:
+                    break;
+            }
+            throw Error.NotSupported();
```
mikepizzo, line 343: >throw Error.NotSupported(); [](http://example.com/codeflow?start=12&length=27)

consider returning null rather than throwing

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
lisicase: I see, gotcha! To double check, it would be reasonable to throw an exception in the caller in the case that the expression is null, but not when reading the annotation value itself?

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
```diff
@@ -0,0 +1,446 @@
...
+                    // return BuildEdmEnumExp(typeReference.AsEnum(), defaultValue)
+                case EdmTypeKind.Complex:
+                case EdmTypeKind.Entity:
+                    // return BuildEdmStructuredTypeExp(typeReference.AsStructured(), defaultValue);
+                case EdmTypeKind.Collection:
+                    // return BuildEdmCollectionExp(typeReference, defaultValue);
+                case EdmTypeKind.Path:
+                    // return BuildEdmPathExp(typeReference.AsPath(), defaultValue);
+                case EdmTypeKind.TypeDefinition:
+                    // return BuildEdmTypeDefinitionExp(typeReference.AsTypeDefinition(), defaultValue);
```
mikepizzo, line 43: you have implementations for some of these. Which work and which don't? Do enum and typedefinition work?

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
```diff
@@ -0,0 +1,446 @@
...
+            throw Error.NotSupported();
+        }
+
+        /// <summary>
+        /// Parses a <paramref name="defaultValue"/> into an IEdmExpression value of the correct CSDL type.
+        /// </summary>
+        /// <param name=“typeReference”>The type of value.</param>
+        /// <param name=“defaultValue”>Original default value.</param>
+        /// <returns>An IEdmExpression of type <paramref name="typeReference".</paramref></returns>
+        public static CsdlExpressionBase BuildDefaultCsdlExpression(IEdmTypeReference typeReference, string defaultValue, CsdlLocation location)
```
mikepizzo, line 352: >BuildDefaultCsdlExpression [](http://example.com/codeflow?start=41&length=26)

We should consider whether we can consolidate these code paths; maybe in ODL 8.0. 

Do EdmVocabularyAnnotation and CsdlVocabularyAnnotation really care that their expressions are EdmExpression/CsdlExpression rather than IEdmExpression?

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
lisicase: I hadn't tested the non-primitive types thoroughly, so that's why it's commented at the moment. I can see if I can test it once I've implemented the other feedback!

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
mikepizzo: yes; programmatically creating it's fine to throw an exception, and it's fine for model validation to fail, but we shouldn't error when reading an invalid annotation.

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
mikepizzo: maybe BuildEdmExpression and BuildCsdlExpression?

lisicase: Ah, I see what you're getting at!

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -44,8 +47,67 @@ public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term,
...
             this.qualifier = qualifier;
             this.value = value;
+            this.usesDefault = false;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the EdmVocabularyAnnotation class
+        /// to the default value.
+        /// </summary>
+        /// <param name=“target”>Element the annotation applies to.</param>
```
xuzhg, line 57: >�target� [](http://example.com/codeflow?start=24&length=8)

should use " directly, don't use other lanaguage """.

lisicase: Done ✅

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -44,8 +47,67 @@ public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term,
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the EdmVocabularyAnnotation class
+        /// to the default value.
+        /// </summary>
+        /// <param name="target">Element the annotation applies to.</param>
+        /// <param name="term">Term bound by the annotation</param>
+        public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term)
+            :this(target, term, "")
```
xuzhg, line 60: >"" [](http://example.com/codeflow?start=32&length=2)

null, not empty string

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -44,8 +47,67 @@ public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term,
...
+            // Initialize
+            this.target = target;
+            this.term = term;
+            if (qualifier.Length > 0)
+            {
+                this.qualifier = qualifier;
+            } else
+            {
+                this.qualifier = null;
+            }
```
xuzhg, line 91: this.qualifier = qualifier;   is enough

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
```diff
@@ -0,0 +1,456 @@
...
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using Microsoft.OData.Edm.Csdl;
+using System.Collections.Generic;
+using System.Linq;
+using System.IO;
+using System.Text;
+using Microsoft.OData.Edm.Csdl.Parsing.Ast;
```
xuzhg, line 12: put all "system.*" ahead and sorting

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
lisicase: Null was my first thought as well, but then the constructor call is ambiguous with the other three-constructor method and I therefore get an error. I opted for "" instead because it's string-specific, so the computer can tell which constructor I'm intending. If you have a different suggestion, that would definitely help!

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
```diff
@@ -2497,6 +2497,22 @@ public static class ValidationRules
...
                 });
 
+        /// <summary>
+        /// Validates that if a vocabulary annotation doesn't have a value, it should declare a type with a default value.
+        /// </summary>
+        public static readonly ValidationRule<IEdmVocabularyAnnotationWithDefault> VocabularyAnnotationUseDefaultTermMustHaveDefaultValue =
+            new ValidationRule<IEdmVocabularyAnnotationWithDefault>(
+                (context, annotation) =>
+                {
+                    if (annotation.UsesDefault && annotation.Term.DefaultValue == null)
```
xuzhg, line 2507: >UsesDefault [](http://example.com/codeflow?start=35&length=11)

The logic here doesn't make sense to me. 

if (annotation.UsesDefault == true), it means annotation.Term has the default value. Otherwise, you can not create annotation. (because it will throw in 
EdmVocabularyAnnotation constructor).

So, it seems it's a conflict. Since we have the verification in the EdmVocabularyAnnotation constructor, we can consider remove this rule?  @mikepizzo

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
```diff
@@ -504,7 +504,13 @@ internal override void WriteVocabularyAnnotationElementHeader(IEdmVocabularyAnno
             this.WriteOptionalAttribute(CsdlConstants.Attribute_Qualifier, annotation.Qualifier, EdmValueWriter.StringAsXml);
             if (isInline)
```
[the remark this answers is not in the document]
xuzhg: So far, OData.Edm doesn't write second pattern.

src/Microsoft.OData.Edm/Microsoft.OData.Edm.csproj
```diff
@@ -99,6 +99,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Newtonsoft.Json" Version="13.0.1" />
```
xuzhg: ><PackageReference Include="Newtonsoft.Json" Version="13.0.1" /> [](http://example.com/codeflow?start=4&length=63)

remove this one

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
```diff
@@ -11,13 +11,14 @@
 using Microsoft.OData.Edm.Csdl.Parsing.Ast;
 using Microsoft.OData.Edm.Validation;
 using Microsoft.OData.Edm.Vocabularies;
+using static Microsoft.OData.Edm.Vocabularies.EdmTermExtensions;
```
xuzhg, line 14: remove this line

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
```diff
@@ -127,15 +137,27 @@ protected IEdmTerm ComputeTerm()
...
             return CsdlSemanticsModel.WrapExpression(adjustedExpression, TargetBindingContext, this.Schema);
         }
 
         private static CsdlExpressionBase AdjustStringConstantUsingTermType(CsdlExpressionBase expression, IEdmTypeReference termType)
         {
-            if (expression == null || termType == null)
-            {
+            if (expression == null || termType == null)            {
                 return expression;
             }
```
xuzhg, line 162: revert this change

test/FunctionalTests/Microsoft.OData.Edm.Tests/Vocabularies/EdmTermExtensionsTests.cs
```diff
@@ -0,0 +1,37 @@
...
+        {
+            string defaultValue = "[{\"city\":[\"redmond\",\"s{eattle\"]},{\"state\":[\"wa\",\"ca\"]},\"test\"]";
+            IEnumerable<string> results = EdmTermExtensions.ParseCollection(defaultValue);
+            var resultsArray = results.ToArray();
+            Assert.Equal(3, resultsArray.Length);
+            Assert.Equal("{\"city\":[\"redmond\",\"s{eattle\"]}", resultsArray[0]);
+            Assert.Equal("{\"state\":[\"wa\",\"ca\"]}", resultsArray[1]);
+            Assert.Equal("\"test\"", resultsArray[2]);
+
+            // TODO: add additional cases - empty array, empty string, null, whitespace, special chars, etc.
```
xuzhg, line 34: should add test case to test the mail methods in this class. 
For example: 

BuildDefaultEdmExpression

src/Microsoft.OData.Edm/Microsoft.OData.Edm.csproj
lisicase: Done ✅

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
mikepizzo: Type checks and casts are expensive.  This results in one of each. Better to try a type cast and check for null.  Not quite a concise, but better performant.  Something like:

IEdmVocabularyAnnotationWithDefault annotationWithDefault = annotation as IEdmVocabularyAnnotationWithDefault;
if(annotationWithDefault == null || annotationWithDefault.UsesDefault)
{
...
}

Or, if you want to be a little more concise:

if(!(annotation is IEdmVocabularyAnnotationWithDefault annotationWithDefault) || annotationWithDefault.UsesDefault))
{
...
}

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
mikepizzo: Yes;  for *our* implementation of IEdmVocabularyAnnotationWithDefault, we will fail before creating the annotation if UsesDefault==true and the term has no default value.  I was okay with adding the rule because implementation of IEdmVocabularyAnnotationWithDefault may not have that logic, and I thought it would be a good exercise for Lisi to author a model validation rule.  However, I agree that, now that we have validation logic in EdmVocabularyAnnotation, we can remove the rule and save performance of running an extra rule.

lisicase: Removed ✅

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
lisicase: Based on a discussion I had with Sam, the distinction is important internally. CsdlExpressionBase is an intermediate wrapper -- the final wrapper is the CsdlSemantics*Expression, which implements IEdmExpression -- but it sounds like it's still needed. However, to simplify things for clients and as encouraged by Sam, I've now created a different public method that does the more external work of returning an IEdmExpression (still creating a CsdlExpressionBase internally as above).

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
lisicase: Done ✅

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmTermExtensions.cs
lisicase: [Deprioritizing for this PR]

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
mikepizzo: can you do
         :this(target, term, (string)null)
?

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -44,8 +47,67 @@ public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term,
...
 
+        /*private void initialize(IEdmVocabularyAnnotatable target, IEdmTerm term, string qualifier, IEdmExpression value)
+        {
+            this.target = target;
+            this.term = term;
+            this.qualifier = qualifier;
+            this.value = value;
+            this.usesDefault = true;
+        }*/
+
```
mikepizzo: you can remove these lines

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
```diff
@@ -127,8 +136,21 @@ protected IEdmTerm ComputeTerm()
...
             IEdmTypeReference termType = Term is UnresolvedVocabularyTerm ? null : Term.Type;
-            CsdlExpressionBase adjustedExpression = AdjustStringConstantUsingTermType((this.Annotation).Expression, termType);
 
+            if (this.Annotation.Expression == null && termType != null)
+            {
+                CsdlExpressionBase defaultExpBase = EdmTermExtensions.BuildCsdlExpression(termType, Term.DefaultValue, (CsdlLocation) this.Annotation.Location);
+
+                if (defaultExpBase == null)
+                {
+                    throw Error.NotSupported();
```
mikepizzo: Can we throw a better/more descriptive exception here?

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
lisicase: I think this might be from a previous version -- it's now removed

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
mikepizzo: Wait a second; when we read XML CSDL, we'll create a CsdlSemanticsVocabularyAnnotation, and I don't think that currently validates that either a value is specified or that the term has a default value -- nor should it validate, since we don't want to throw exceptions for invalid annotations when reading CSDL.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
lisicase: Done ✅

test/FunctionalTests/Microsoft.OData.Edm.Tests/Vocabularies/EdmTermExtensionsTests.cs
lisicase: I finished testing the first constructor in the class. TODO: Test BuildEdmExpression(CsdlSemanticsVocabularyAnnotation annotation)

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -4,17 +4,20 @@
...
     /// Represents an EDM annotation with an immediate value.
     /// </summary>
-    public class EdmVocabularyAnnotation : EdmElement, IEdmVocabularyAnnotation
+    public class EdmVocabularyAnnotation : EdmElement, IEdmVocabularyAnnotationWithDefault
     {
         private readonly IEdmVocabularyAnnotatable target;
         private readonly IEdmTerm term;
         private readonly string qualifier;
         private readonly IEdmExpression value;
+        private readonly bool usesDefault;
```
corranrogue9, line 20: I don't think we need a field here since it is `readonly` and `UsesDefault` only has a getter

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
```diff
@@ -412,6 +412,225 @@ public void ReadNavigationPropertyPartnerTypeHierarchyTest()
...
+                         "<Annotation Term=\"NS.MyAnnotationPathTerm\" AnnotationPath=\"abc/efg\" />" +
+                         "<Annotation Term=\"NS.MyDefaultTerm\" />" + // does not specify value
+                       "</ComplexType>" +
+                       "<Term Name=\"MyAnnotationPathTerm\" Type=\"Edm.AnnotationPath\" Nullable=\"false\" />" +
+                       "<Term Name=\"MyDefaultTerm\" Type=\"Edm.String\" DefaultValue=\"This is a test\" AppliesTo=\"Property Term\" Nullable=\"false\" />" +
+                     "</Schema>" +
+                   "</edmx:DataServices>" +
+                 "</edmx:Edmx>";
+            // Parse into CSDL
+            var model = CsdlReader.Parse(XElement.Parse(csdl).CreateReader());
```
corranrogue9, line 433: `CreateReader` returns an object that implements `IDisposable`. It should be wrapped in a `using` block

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
```diff
@@ -436,6 +436,23 @@ internal override void WriteCollectionTypeElementHeader(IEdmCollectionType colle
...
+            }
+
+            CsdlSemanticsVocabularyAnnotation csdlAnnotation = annotation as CsdlSemanticsVocabularyAnnotation;
+            if (csdlAnnotation != null)
+            {
+                return csdlAnnotation.UseDefault;
+            }
+
+            return false;
+        }
```
mikepizzo, line 454: I don't love this logic, because it special cases the EdmVocabularyAnnotation and CsdlSemanticsVocabularyAnnotation, meaning that a custom implementation of IEdmVocabularyAnnotation can't implement UseDefault; however, I am okay putting it here as long as we fix it by adding usesDefault to IEdmVocabularyAnnotation in 8.0.  @xuzhg -- can you open an issue, tagged 8.0/breaking change opportunity, to address this there?

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -4,6 +4,8 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using System;
+
```
mikepizzo, line 8: Is this needed?

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -4,6 +4,8 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using System;
```
mikepizzo, line 7: Do we need this added using?

mikepizzo (approved):
:shipit:

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
xuzhg: Done at: https://github.com/OData/odata.net/issues/2228

PR #2178, opened 2021-09-02
Clone the reader setting when creating reader setting

xuzhg:
### Issues

### Description

* Web API OData typically re-set the BaseUri in the ReaderSetting.
* When retrieves the ReaderSetting from DI , in the same request, since it's scoped, the only singleton reader setting returns.
* So, It's better to clone the reader setting any time to use the reader setting.
* That's a fix for BingAds.

- [ ] *Test cases added*

### Additional work necessary


mikepizzo (approved):
:shipit:

Issue #2165, opened 2021-08-24
ODL should be able to read untyped nested complex types with ReadUntypedAsString=false


xuzhg:
For the following 
| request payload  | ReadUntypedAsString=true | ReadUntypedAsString=false 
-- | -- | --
{     "name": "Peter",     "homeaddress": {           "Street": "street0",           "City": "city0"      }   }  | ODataProperty         (ODataUntypedValue)    ODataProperty         (ODataUntypedValue) |  ODataProperty         (ODataPrimitiveValue)    ODataNestedResourceInfo

When 'ReadUntypedAsString=false', ODL throws exception at:

![image](https://user-images.githubusercontent.com/9426627/130699898-72b10e85-10e7-484a-8fa7-e54b231b2b1e.png)

PR #2179, opened 2021-09-03
Fix the issue when read nested resource untyped in the request

xuzhg:
### Issues

*This pull request fixes https://github.com/OData/odata.net/issues/2165

### Description

* If we have a nested resource whose property name is wrong case or its undeclared property,
* When we put that property in to the request body, ODL can't read it as Untyped even we set the ReadUntypedAsString=false.
* Since we understand it's undeclared property, can we directly read it as normal untyped value.
* This is a simple fix for this issue

- [x ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
@@ -425,15 +425,12 @@ namespace Microsoft.OData.JsonLight
                 ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, isCollection, propertyName, payloadTypeReference);
                 if (isCollection)
                 {
-                    readerNestedResourceInfo = this.ReadingResponse
-                        ? ReadExpandedResourceSetNestedResourceInfo(resourceState, null, payloadTypeReference.ToStructuredType(), propertyName, /*isDeltaResourceSet*/ false)
-                        : ReadEntityReferenceLinksForCollectionNavigationLinkInRequest(resourceState, null, propertyName, /*isExpanded*/ true);
+                    readerNestedResourceInfo =
+                        ReadExpandedResourceSetNestedResourceInfo(resourceState, null, payloadTypeReference.ToStructuredType(), propertyName, /*isDeltaResourceSet*/ false);
                 }
                 else
                 {
-                    readerNestedResourceInfo = this.ReadingResponse
-                        ? ReadExpandedResourceNestedResourceInfo(resourceState, null, propertyName, payloadTypeReference.ToStructuredType(), this.MessageReaderSettings)
-                        : ReadEntityReferenceLinkForSingletonNavigationLinkInRequest(resourceState, null, propertyName, /*isExpanded*/ true);
+                    readerNestedResourceInfo = ReadExpandedResourceNestedResourceInfo(resourceState, null, propertyName, payloadTypeReference.ToStructuredType(), this.MessageReaderSettings);
                 }
 
                 resourceState.PropertyAndAnnotationCollector.ValidatePropertyUniquenessOnNestedResourceInfoStart(readerNestedResourceInfo.NestedResourceInfo);
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightEntryAndFeedDeserializerUndeclaredTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightEntryAndFeedDeserializerUndeclaredTests.cs
@@ -87,7 +87,7 @@ namespace Microsoft.Test.OData.TDD.Tests.Reader.JsonLight
             message.SetHeader("Content-Type", "application/json");
 
             using (var msgReader = readRequest ?
-                new ODataMessageReader((IODataResponseMessage)message, readerSettings, this.serverModel) :
+                new ODataMessageReader((IODataRequestMessage)message, readerSettings, this.serverModel) :
                 new ODataMessageReader((IODataResponseMessage)message, readerSettings, this.serverModel))
             {
                 var reader = msgReader.CreateODataResourceReader(entitySet, entityType);
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightEntryAndFeedDeserializerUndeclaredTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightEntryAndFeedDeserializerUndeclaredTests.cs
@@ -80,12 +80,15 @@ namespace Microsoft.Test.OData.TDD.Tests.Reader.JsonLight
         }
         // ----------- end of edm for entry reader ----------- 
 
-        private void ReadEntryPayload(string payload, EdmEntitySet entitySet, EdmEntityType entityType, Action<ODataReader> action, bool readUntypedAsValue = false)
+        private void ReadEntryPayload(string payload, EdmEntitySet entitySet, EdmEntityType entityType, Action<ODataReader> action, bool readUntypedAsValue = false, bool readRequest = false)
         {
             ODataMessageReaderSettings readerSettings = readUntypedAsValue ? UntypedAsValueReaderSettings : UntypedAsStringReaderSettings;
             var message = new InMemoryMessage() { Stream = new MemoryStream(Encoding.UTF8.GetBytes(payload)) };
             message.SetHeader("Content-Type", "application/json");
-            using (var msgReader = new ODataMessageReader((IODataResponseMessage)message, readerSettings, this.serverModel))
+
+            using (var msgReader = readRequest ?
+                new ODataMessageReader((IODataResponseMessage)message, readerSettings, this.serverModel) :
+                new ODataMessageReader((IODataResponseMessage)message, readerSettings, this.serverModel))
             {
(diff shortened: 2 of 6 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/Microsoft.OData.Core.Tests.csproj
```diff
@@ -9,6 +9,7 @@
     <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
     <GenerateAssemblyConfigurationAttribute>false</GenerateAssemblyConfigurationAttribute>
     <RestorePackages>true</RestorePackages>
+    <GenerateTargetFrameworkAttribute>false</GenerateTargetFrameworkAttribute>
```
xuzhg, line 12: can ignore this change. It seems I have to add this configuration at my side to make the VS happen.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightEntryAndFeedDeserializerUndeclaredTests.cs
```diff
@@ -80,12 +80,15 @@ public ODataJsonLightEntryAndFeedDeserializerUndeclaredTests()
...
+        private void ReadEntryPayload(string payload, EdmEntitySet entitySet, EdmEntityType entityType, Action<ODataReader> action, bool readUntypedAsValue = false, bool readRequest = false)
         {
             ODataMessageReaderSettings readerSettings = readUntypedAsValue ? UntypedAsValueReaderSettings : UntypedAsStringReaderSettings;
             var message = new InMemoryMessage() { Stream = new MemoryStream(Encoding.UTF8.GetBytes(payload)) };
             message.SetHeader("Content-Type", "application/json");
-            using (var msgReader = new ODataMessageReader((IODataResponseMessage)message, readerSettings, this.serverModel))
+
+            using (var msgReader = readRequest ?
+                new ODataMessageReader((IODataResponseMessage)message, readerSettings, this.serverModel) :
+                new ODataMessageReader((IODataResponseMessage)message, readerSettings, this.serverModel))
```
gathogojr, line 91: What's the difference between the consequent and the alternative expressions here?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -425,15 +425,12 @@ protected ODataJsonLightReaderNestedResourceInfo InnerReadUndeclaredProperty(IOD
                 ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, isCollection, propertyName, payloadTypeReference);
                 if (isCollection)
                 {
-                    readerNestedResourceInfo = this.ReadingResponse
-                        ? ReadExpandedResourceSetNestedResourceInfo(resourceState, null, payloadTypeReference.ToStructuredType(), propertyName, /*isDeltaResourceSet*/ false)
-                        : ReadEntityReferenceLinksForCollectionNavigationLinkInRequest(resourceState, null, propertyName, /*isExpanded*/ true);
+                    readerNestedResourceInfo =
+                        ReadExpandedResourceSetNestedResourceInfo(resourceState, null, payloadTypeReference.ToStructuredType(), propertyName, /*isDeltaResourceSet*/ false);
```
gathogojr, line 429: For my own understanding, how does removing the alternative expression fix the issue?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightEntryAndFeedDeserializerUndeclaredTests.cs
mikepizzo: +1 -- should line 90 cast message to IODataRequestMessage?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -425,15 +425,12 @@ protected ODataJsonLightReaderNestedResourceInfo InnerReadUndeclaredProperty(IOD
...
-                    readerNestedResourceInfo = this.ReadingResponse
-                        ? ReadExpandedResourceSetNestedResourceInfo(resourceState, null, payloadTypeReference.ToStructuredType(), propertyName, /*isDeltaResourceSet*/ false)
-                        : ReadEntityReferenceLinksForCollectionNavigationLinkInRequest(resourceState, null, propertyName, /*isExpanded*/ true);
+                    readerNestedResourceInfo =
+                        ReadExpandedResourceSetNestedResourceInfo(resourceState, null, payloadTypeReference.ToStructuredType(), propertyName, /*isDeltaResourceSet*/ false);
                 }
                 else
                 {
-                    readerNestedResourceInfo = this.ReadingResponse
-                        ? ReadExpandedResourceNestedResourceInfo(resourceState, null, propertyName, payloadTypeReference.ToStructuredType(), this.MessageReaderSettings)
```
mikepizzo, line 435: Are "ReadExpandedResourceNestedResourceInfo"
 and "ReadEntityReferenceLinksForCollectionNavigationLinkInRequest" still used, or can those methods be deleted?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -425,15 +425,12 @@ protected ODataJsonLightReaderNestedResourceInfo InnerReadUndeclaredProperty(IOD
                 ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, isCollection, propertyName, payloadTypeReference);
                 if (isCollection)
                 {
-                    readerNestedResourceInfo = this.ReadingResponse
-                        ? ReadExpandedResourceSetNestedResourceInfo(resourceState, null, payloadTypeReference.ToStructuredType(), propertyName, /*isDeltaResourceSet*/ false)
```
mikepizzo, line 429: I assume that ReadExpandedResourceSetNestedResourceInfo (and ReadExpandedResourceNestedResourceInfo) handle the case of the payload just containing links, and that the developer experience behaves the same for this case?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
mikepizzo: @xuzhg, please correct my understanding if wrong, but just to make sure that I understand the change --

The old code assumed that only links would ever be sent in a request payload, so it didn't support reading nested content.

In fact, it should be valid to have nested content in a request payload as well as a response payload, so removing the special case code for request payload handles nested content (whether links or actual content) for both cases.

Right?

mikepizzo (commented):
Please see minor comments.  Assuming the behavior doesn't change for reading nested links in a request payload, I'm fine with the change.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
KenitoInc: @mikepizzo I think `ReadExpandedResourceNestedResourceInfo` is still being used

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightEntryAndFeedDeserializerUndeclaredTests.cs
xuzhg: Yes.  Thanks.

xuzhg:
No, it doesn't change. The change is only for the Undeclared. 

---
In reply to: [748268333](https://github.com/OData/odata.net/pull/2179#issuecomment-748268333) [](http://example.com/codeflow?ancestors=748268333,748268333)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
xuzhg: The change is only in 'InnerReadUndeclaredProperty', We are in the method because It's "Undeclared" property.
For normal property, 'ReadEntityReferenceLinksForCollectionNavigationLinkInRequest' and other methods are still in use.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
xuzhg: Yes. 
For "normal" (aka declared) property, the process never goes into this method.

For "undeclared" property, since it's undeclared, we don't know the type. Since, we don't know the type, why shall we treat it as entity/entity reference if it's a request payload.

So, I think in the "undeclared" property reading, we should read it as nested resource info no matter it's request or response.

xuzhg:
@mikepizzo @gathogojr are the comments ok for you? Shall I go to merge it? 

BingAds is waiting the hot release.

mikepizzo (approved):
:shipit:

mikepizzo:
@xuzhg -- LGTM I am okay to merge.

---
In reply to: [918666404](https://github.com/OData/odata.net/pull/2179#issuecomment-918666404) [](http://example.com/codeflow?ancestors=918666404,918666404)

PR #2187, opened 2021-09-10
Implement interfaces in support of asynchronous reading of OData message payloads

gathogojr:
### Issues

*This pull request is in partial fulfilment of #2019 .*

### Description

Implement interfaces in support of asynchronous reading of OData message payloads

### DIFF
--- a/src/Microsoft.OData.Core/ODataBinaryStreamReader.cs
+++ b/src/Microsoft.OData.Core/ODataBinaryStreamReader.cs
@@ -50,7 +50,7 @@ namespace Microsoft.OData
             Debug.Assert(reader != null, "reader cannot be null");
 
             this.reader = reader;
-            chars = new char[charLength];
+            this.chars = new char[this.charLength];
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.OData.Core/ODataBinaryStreamReader.cs
```diff
@@ -50,7 +50,7 @@ internal ODataBinaryStreamReader(Func<char[], int, int, int> reader)
             Debug.Assert(reader != null, "reader cannot be null");
 
             this.reader = reader;
-            chars = new char[charLength];
+            this.chars = new char[this.charLength];
```
gathogojr, line 53: Note: Changes to this file are primarily intended to make it easy to tell apart class and local variables

odero (approved):

habbes (approved):

KenitoInc (commented):
Is the goal of introducing the interfaces, to allow injecting custom implementation?

ElizabethOkerio (approved):

gathogojr:
> Is the goal of introducing the interfaces, to allow injecting custom implementation?

@KenitoInc The primary reason for introducing the interfaces is to prevent a breaking change. If it wasn't for that, I'd have made the changes in the existing `IJsonReader`, `IJsonStreamReader` and `IJsonReaderFactory` interfaces. In the next major release, I'll merge these new interfaces into the existing interfaces. The interfaces though do make it possible for custom implementations to be injected

KenitoInc (approved):

PR #2188, opened 2021-09-10
Avoid creating unused objects when calling ODataUri.Clone()

### LINKED ISSUE #2163: Performance - expensive allocations due to ODataUri.Clone()
ODataUri.Clone() is costing 1.72% in allocations.

image

Do we need to clone **ODataUri** every time we enter the scope in **ODataWriterCore**? I suppose this is needed for mutability, since the code is modifying it? Though I wonder if there is an opportunity to change/refactor the code to avoid or reduce the need of this expensive allocation?

### Assemblies affected

Microsoft.OData.Core 7.8.3

### Reproduce steps

Exercise Microsoft.OData.ODataWriterCore.EnterScope() method

### Expected result

Less number of allocations.

### Actual result

Excessive number of allocations.

@mikepizzo @Sreejithpin @chrisspre

habbes:
### Issues

*This pull request (partially) addresses #2163 .* 

#### The fix

`ODataUri.Clone()` creates a new instance of `ODataUri` using `new ODataUri()` then shallow copies the properties (references) from the "old" `ODataUri` to the new one. However, the default `ODataUri` constructor creates a new instance of `Dictionary<string, string>` and a new instance of `ParameterAliasValueAccessor`. The  `ODataUri.ParameterAliasValueaAccessor` property is immediately replaced by `ODataUri.Clone()` with the reference from the old uri. So the instances created in the default constructor are a waste of memory and unnecessary allocation/GC overhead.

I fixed this issue by creating the `ParameterAliasValueAccessor` instance lazily in the property getter. I thought this to be the safest solution, below are other solutions that I considered:
- Removing the new instances from the constructor without modifying the property getter
  -  I cannot guarantee that this would not lead to null reference exceptions in cases where the default constructor is used outside of `Clone()`, and I did not have time to investigate all those usage scenarios
- Using the overloaded constructor that accepts more arguments (including `ParameterAliasValueAccessor`) that are then directly passed to properties. This constructor seems close to what `ODataUri.Clone` is doing, except that when setting the `CustomQueryOptions` property: `this.CustomQueryOptions = new ReadOnlyCollection<QueryNode>(customQueryOptions.ToList());`, where as in `ODataUri.Clone()` the property reference is simply copied without creating any new object. So this would not be a suitable substitute.
- Creating a new constructor that does not create new instances. This felt unnecessary.

#### Expected impact
While this change does not reduce the number of `ODataUri` instances allocated (which is what the original issue pointed to in the screenshot), it does significantly reduce the number of `Dictionary` instances created as a result of `ODataUri.Clone()`.

This CPR screenshot shows that `Dictionary<string, string>`  created from `ODataUri.Clone()` account for and 1.3% (0.69 + 0.61) of allocated size:
image

The following screenshot shows `Dictionary<string, SingleValueNode>`  from `ODataUri.Clone()` account for  0.65%.
AGS SingleValueNode dict allocation size

The following screenshot shows `ParameterAliasValueAccessor` from `ODataUri.Clone()` accounts for 0.23%

ParameterAliasValueAccessor alloc size

So in total, there's up to 2.18% that we can potentially shed off. This is based on the assumptions that the sizes indicated in the screenshots do not include sizes of referenced objects (otherwise `ParameterAliasValueAccessor` would have a larger inclusive size than the dictionaries it contains).

I've done some local profiling before and after the fix to get a better estimate of how much we can expect to reduce. The profiling were based on running a simple service that uses that OData writer to write a response of 5000 entities (base on [this experments project](https://github.com/habbes/experiments/tree/master/ODataWriterVsSystemTextJson)) and collecting allocation data using the .NET Object Allocation Tracker tool that's part of the Visual Studio Performance Profiler toolset.

The following estimates may not hold true in production given the difference in usage patterns and data, but it may be a good idea to compare the results once we measure them in production with the estimates here later on. I think that will help us make better estimates over time.

##### Impact on number of allocations
The first 2 screenshots below show the total allocations from `ODataUri.Clone()` from my local profile before and after this change. You can see total allocations dropped from 175k to 35k, i.e. about 140k drop in allocations, which is about 80%. `ODataUri.Clone()` accounts in total for 4.13% of allocations (inclusive), 80% drop would should **shed off about 3.3% of allocations in AGS** assuming the experiment data is a reflection of what happens in production.

ODataUri.Clone() Function Allocations Before

ODataUri.Clone() Function Allocations After

ODataUri.Clone() allocations on CPR

##### Impact on allocation size

Impact on allocation size is trickier to estimate cause the .NET allocation tracker in VS doesn't display the inclusive allocation size for a given type. I'm also not sure how the inclusive size is computed in CPR (on the stack tab, both inclusive and exclusive columns have the same value), on the Frames tab, they have different values, but this tab doesn't show the breakdown of the types allocated by a function.

That said, I thin we can realistically estimate that at least all `Dictionary<string, string>`, `Dictionary<string, SingleValueNode>` and `ParameterAliasValueAccessor` allocated from `ODataUri.Clone()` will be eliminated, and I've already calculated that above to be **2.18% of total allocated bytes**.

Below are some screenshots from my local profile showing the allocated size of `Dictionary` instances before and after the change.

Dictionaries alloc size local before

Dictionaries alloc size local after

##### Impact on collection

On my local profile, there were 3 fewer GC collections after the change. But I'm not sure if this is reliable.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Uri/ODataUri.cs
+++ b/src/Microsoft.OData.Core/Uri/ODataUri.cs
@@ -30,13 +30,13 @@ namespace Microsoft.OData
         /// </summary>
         private Uri serviceRoot;
 
+        private ParameterAliasValueAccessor parameterAliasValueAccessor;
+
         /// <summary>
         /// Initializes a new instance of the <see cref="ODataUri"/> class.
         /// </summary>
         public ODataUri()
         {
-            IDictionary<string, string> dictionary = new Dictionary<string, string>(StringComparer.Ordinal);
-            this.ParameterAliasValueAccessor = new ParameterAliasValueAccessor(dictionary);
         }
 
         /// <summary>
@@ -214,7 +214,23 @@ namespace Microsoft.OData
         /// <summary>
         /// Gets or sets the ParameterAliasValueAccessor.
         /// </summary>
-        internal ParameterAliasValueAccessor ParameterAliasValueAccessor { get; set; }
+        internal ParameterAliasValueAccessor ParameterAliasValueAccessor
+        {
+            get
+            {
+                if (parameterAliasValueAccessor == null)
+                {
+                    IDictionary<string, string> dictionary = new Dictionary<string, string>(StringComparer.Ordinal);
+                    parameterAliasValueAccessor = new ParameterAliasValueAccessor(dictionary);
+                }
+
+                return parameterAliasValueAccessor;
+            }
+            set
+            {
+                parameterAliasValueAccessor = value;
+            }
+        }
 
         /// <summary>
         /// Return a copy of current ODataUri.


joaocpaiva (approved):
Fix loos good to me. Lazy initialization in the getter minimizes the changes and given this class is not meant to be used concurrently it works just fine. It also improves any scenarios where ODataUri is created, but ParameterAliasValueAccessor is never used.

src/Microsoft.OData.Core/Uri/ODataUri.cs
```diff
@@ -30,13 +30,13 @@ public sealed class ODataUri
         /// </summary>
         private Uri serviceRoot;
 
+        private ParameterAliasValueAccessor parameterAliasValueAccessor;
```
joaocpaiva, line 33: Line 130, ParameterAliasNodes has a null check on the property. That null check does not seem required? Or if it is, it should be replaced with the field, instead of the property?

Sreejithpin (commented):
Moving initialization from constructor to getter looks fine . Can you see how much % improvement we got . Let me know if you want me to run a local benchmark test

chrisspre:
> Moving initialization from constructor to getter looks fine . Can you see how much % improvement we got . Let me know if you want me to run a local benchmark test

If we do this kind of work more often I don't think it is good to wait for deployment and measurement via Graph. I would recommend writing benchmark tests (one sdk should be sufficient) and compare the before and after. graph measurements are a moving target and it takes a long time to collect that information.

KenitoInc (approved):

src/Microsoft.OData.Core/Uri/ODataUri.cs
```diff
@@ -214,7 +214,23 @@ public Uri ServiceRoot
         /// <summary>
         /// Gets or sets the ParameterAliasValueAccessor.
         /// </summary>
-        internal ParameterAliasValueAccessor ParameterAliasValueAccessor { get; set; }
+        internal ParameterAliasValueAccessor ParameterAliasValueAccessor
+        {
+            get
+            {
+                if (parameterAliasValueAccessor == null)
```
g2mula, line 221: This is a change of behaviour - could be for the better, but still a change :-) - and there probably should be a test to expect this current behaviour...
- Previously: value is set in ctor, you could then be able to `set` a null to that value and a subsequent `get` would give null.
- Currently: even if you ever `set` a null value, you would never `get` it
It could be that current behaviour is fixing unwanted null gets but yeah, probably a test to document this...

habbes: Hmm, good point

chrisspre: Would it be possible to change the setter from internal to private ?

joaocpaiva: Even the getter, does not seem it needs to be public as of now.

src/Microsoft.OData.Core/Uri/ODataUri.cs
joaocpaiva: This is for the ParameterAliasNodes getter - it has a null check for ParameterAliasValueAccessor which is redundant in the current iteration, since ParameterAliasValueAccessor is never null. Hence should check against the **field** instead of the **property**, or perhaps remove the null check.

src/Microsoft.OData.Core/Uri/ODataUri.cs
mikepizzo: Yes; the setter is currently internal and only called when initialized to a non-null value in the constructor, so we should be able to assume it is never null.  We can change the setter (and getter) to private and change the null check on 128-133 with an non-null assert.

mikepizzo (approved):
:shipit:

Issue #2192, opened 2021-09-14
Memory Usage - Lambda closures in hot paths cause excessive allocations


habbes:
We have more cases of closures causing excessive allocation in AGS. These are cases where we use a lambda expression that references a variable from the enclosing scope. This causes the compiler to create a new object (`DisplayClass`) that will store the referenced variables so that the lambda can still access them when it's called. In addition, the generated `Func` or `Action` is not cached, so a new `Func` or `Action` is created each time.

We can optimize these lambdas refactoring them such that all the variables  that they need be passed as arguments. In the case of LINQ methods like `IEnumerable.Any()` we could consider using plain for loops.

There's also another subtle source of "surprise" allocations when passing a static method as an argument to method expecting a func (or action).

Consider the following scenario:
```c#
using System;
public class C {
    public void M() {
        Helpers.CallFilter(Helpers.DoStuff);
    }
}

public static class Helpers {
    public static bool DoStuff() { return true; }
   
    public static void CallFilter(Func<bool> filter)
    {
        filter();
    }
}
```
Passing `Helpers.DoStuff` directly to `Helpers.CallFilter` causes a `new Func<bool>(Helpers.DoStuff)` to be created each time [as you can see here](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLRIMYHsEBNXIwJwYzIA+AAgEwCMAsAFCUDMABDWwMJsDeTbQR3aUALGwCyACgCUfAUMUAJCABsADhATIAdFyirVAMQCWqmFqkqNW3QBEsAZXhgwMgNwLBAXya/GTKwctABsHNRs1pra8oyKXsLBYcBYWKpsDs5wrrJ8HADsbERwEO5s/kIJQZShHOL6hqbmlpQArAA8KWkAfGxgZhYIMgn8cYpC/c0Isp5jPn5AA=)

But when you wrap the static method into a lambda as follows:
```c#
using System;
public class C {
    public void M() {
        Helpers.CallFilter(() => Helpers.DoStuff());
    }
}

public static class Helpers {
    public static bool DoStuff() { return true; }
   
    public static void CallFilter(Func<bool> filter)
    {
        filter();
    }
}
```
[...]

PR #2193, opened 2021-09-14
Remove delegate allocation in ODataWriterCore.CheckForNestedResourceInfoWithContent()

joaocpaiva:
### Issues

This addresses one of the issues pointed out in [2192](https://github.com/OData/odata.net/issues/2192).

### Description

These is an issue inadvertently missed in this [PR](https://github.com/OData/odata.net/pull/2158).

### Additional work necessary

Going forward I think we should consider using static code analyzers to detect these issues at compile time, in OData critical code path/assemblies. That will avoid missing such issues, inadvertently. This is already done in AGS.

@habbes


odero (approved):

KenitoInc (approved):

KenitoInc:
Thanks @joaocpaiva for the PR

Issue #2190, opened 2021-09-13
Performance - Excessive Enumerator objects created in hot paths


habbes:
A lot of `Enumerator` instances are getting allocated. This is due to using `foreach`  and LINQ against `IEnumerable`s on hot paths.

In cases where the collection provides a cheap way to get the count and has cheap random access (e.g. `List<T>` and arrays) we could consider using simple `for` loops and accessing elements by index directly. It would probably make the code less readable/elegant, but the savings would add up to > 1% allocation savings in AGS.

Here are some of the "hot" loops we could consider optimizing:
- `JsonLightInstanceAnnotationWriter.WriteInstanceAnnotations(IEnumerable<ODataInstanceAnnotation>, InstanceAnnotationWriteTracker, bool, string)` has a `foreach` loop that's used on `List<T>` instances
- `ODataJsonLightResourceSerializer.WriteResourceEndMetadataProperties(IODataJsonLightWriterResourceState, IDuplicatePropertyNameChecker)`: this function calls `actions.Any()` and `functions.Any()`, calling `Any()` causes the `List<T>.GetEnumerator()` to be called. I don't completely understand this because from the `Enumerable.Any` [source code](https://source.dot.net/#System.Linq/System/Linq/AnyAll.cs,11), it's not supposed to call `GetEnumerator` if the `IEnumerable<T>` also implements `ICollection<T>`, which `List<T>` does, in such cases it's supposed to check `Count != 0`.
- `Edm.ExtensionMethods.FindVocabularyAnnotations(IEdmModel, IEdmVocabularyAnnotatable)` has a `foreach` loop.
- `ODataWriterCore.AppendEntitySetKeySegment(ODataPath, bool)` has a `foreach (IEdmStructuralProperty property in currentEntityType.Key())` loop
- `ODataPath.ctor(IEnumerable<ODataPathSegments)` (the constructor) passes the segments `IEnumerable` to the `List` constructor, the `List` constructor iterates over the collection to add the elements to the list. Not sure if this can be optimized easily.
- `.ODataJsonLightValueSerializer.WriteCollectionValue(ODataCollectionValue, IEdmTypeReference, IEdmTypeReference, bool, bool, bool)` has a `foreach (object item in items)` loop
- Use of `scopeStack.First()` in `ODataWriterCore` (not sure if this can be easily optimized)
- `ODataResourceBase.VerifyProperties(IEnumerable<ODataProperty>)`
[...]

PR #2197, opened 2021-09-22
Reduce memory allocations from closures and lambdas

### LINKED ISSUE #2192: Memory Usage - Lambda closures in hot paths cause excessive allocations

habbes:
### Issues

*This pull request fixes #2192 .*

### Description

Optimizes the hot spots specified in the issue description and also adds new benchmarks to the performance tests solution.

- [x] Avoid lambda capturing variable from outer scope in `ODataWriterCore.CheckForNestedResourceInfoWithContentAsync`
- [x] Pass lambda instead of extension method as func argument in `ExtensionMethods.FindType`
- [x] Replace `list.FirstOrDefault` with for loop to avoid closure allocations in `EdmExtensions.ReplaceAlias`
- [x] Replace `DerivedTypeConstraints.Any()` with `foreach` loop in `WriterValidationUtils.ValidatePropertyDerivedTypeConstraints`. Using `foreach` does cause an `IEnumerator` to be allocated on the heap, but so does using `Any()` in this case (since the underlying `IEnumerable` is not a collection with a cheap count). There's a [separate issue](https://github.com/OData/odata.net/issues/2190) that aims to reduce `IEnumerator` allocations, and that would involve checking whether this collection can be converted to a `List<T>`. I will do that investigation when dealing with that issue. 
- [x] Replace `derivedTypeConstraints.Any()` with `foreach` loop in `WriterValidationUtils.ValidateDerivedTypeConstraints`.  (Same remarks as the item above)
- [x] Replace `this.children.Any()` with `this.children.TryGetValue` in `SelectedPropertiesNode.GetSelectedPropertiesForNavigationProperty`. Also replace the `this.GetMatchinTypeSegments().Select()` with a method `GetSelectedPropertieForTypeSegmentsNavigationProperty` that encapsulates that does the same thing as the `Select()` but without lambdas.

I've also added a new benchmarks project. The purpose of these benchmarks is to compare performance between the JsonSerializer and ODataWriter (both sync and async). These are based on the experiments that I've been running locally to uncover issues with writing. The pre-existing benchmarks did not cover async code paths and do not give us insight into how we're doing compared System.Text.Json or some other "standard" serialization library.

#### Benchmarks

On my local profile, the number of total allocations from `Microsoft.OData.Core` assembly went down from 1,475,609 to 1,385,600 (About 6% reduction).

The benchmarks below were run before and after the set of changes. They indicate a reduction of about 3.2-3.4% in allocated memory for the synchronous ODataWriter, and about 1.4% reduction for the async ODataWriter.

There also seems to be consistent reduction in run time, but I'm not sure to which extent this reduction is significant.

These benchmarks have been added to the repo and you can run them using (not sure if this was the best name of the benchmarks):
```
crank --config benchmarks.yml --scenario SerializerBaselines --profile lab-windows
```

**Before**
image

**After**
image

Here are the pre-existing `ODataWriter` benchmarks:
```
crank --config benchmarks.yml --scenario Writer --profile lab-windows
```
**Before**
image

**After**
image

- [x] *Test cases added*

### Additional work necessary

cc @joaocpaiva


src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -294,10 +294,19 @@ internal SelectedPropertiesNode GetSelectedPropertiesForNavigationProperty(IEdmS
             }
 
             // $select=Orders will include the entire subtree when there are no same expanded entity.
-            if (this.selectedProperties != null && this.selectedProperties.Contains(navigationPropertyName) &&
-                (this.children == null || !this.children.Any(n => n.Key.Equals(navigationPropertyName, StringComparison.Ordinal) && n.Value.isExpandedNavigationProperty)))
```
joaocpaiva, line 298: This is the problem of using Linq lightly and excessively 🤦‍♂️

Any() is an inefficient O(n) operation when this can be done using O(1) since **children** is a dictionary. Not to mention the lambda + closure allocation, though that is a smaller issue than the O(n) cost.

        public static bool Any<TSource>(this IEnumerable<TSource> source)
        {
            if (source == null)
            {
                throw Error.ArgumentNull("source");
            }

            using (IEnumerator<TSource> enumerator = source.GetEnumerator())
            {
                if (enumerator.MoveNext())
                {
                    return true;
                }
            }

            return false;
        }

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
```diff
@@ -312,8 +321,7 @@ internal SelectedPropertiesNode GetSelectedPropertiesForNavigationProperty(IEdmS
 
                 // try to find a child with a type segment before it that matches the current type.
                 // Note: the result of this aggregation will be either empty or a found child node.
-                return this.GetMatchingTypeSegments(structuredType)
-                    .Select(typeSegmentChild => typeSegmentChild.GetSelectedPropertiesForNavigationProperty(structuredType, navigationPropertyName))
+                return GetSelectePropertiesForTypeSegmentsNavigationProperties(this.GetMatchingTypeSegments(structuredType), structuredType, navigationPropertyName)
                     .Aggregate(child, CombineNodes);
```
joaocpaiva, line 325: Use lambda to save the allocation?

 (l, r) => CombineNodes(l, r)

test/PerformanceTests/SerializationBaselineTests/SerializationBaselineTests.csproj
```diff
@@ -0,0 +1,17 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
```
joaocpaiva, line 5: Any specific reason to not use net5, the latest release? AGS is also on net5 FWIW.

KenitoInc (approved):

ElizabethOkerio (approved):

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
habbes: Not sure why `Any()` was used in this case given that it's searching by key and `this.children` is dictionary.

src/Microsoft.OData.Core/SelectedPropertiesNode.cs
habbes: Done

test/PerformanceTests/SerializationBaselineTests/SerializationBaselineTests.csproj
habbes: Added .net5

PR #2199, opened 2021-09-28
Implement asynchronous support in ODataTextStreamReader class

gathogojr:
### Issues

*This pull request is in partial fulfilment of issue #2019.*

### Description

Implement asynchronous support in **`ODataTextStreamReader`** class

- [ ] *Test cases added*

### Additional work necessary


test/FunctionalTests/Microsoft.OData.Core.Tests/ODataTextStreamReaderTests.cs
```diff
@@ -0,0 +1,71 @@
...
+    /// </summary>
+    public class ODataTextStreamReaderTests
+    {
+        private MemoryStream stream;
+        private const int maxLength = 14;
+
+        public ODataTextStreamReaderTests()
+        {
+            this.stream = new MemoryStream(Encoding.UTF8.GetBytes("The quick brown fox jumps over the lazy dog."));
+            this.stream.Position = 16;
```
KenitoInc, line 26: Why are we setting the Position as 16?
I thought to read a stream, the position need to be at 0.

KenitoInc (approved):

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataTextStreamReaderTests.cs
gathogojr: @KenitoInc For the fun of it. I just wanted to start to read from a random position instead of the typical 0

corranrogue9: It might be good to have a test for both

marabooy: Maybe randomize it from 0 to length :)

Issue #2037, opened 2021-03-26
Provide a convenient way to specify a HttpClient instance


JechoJekov:
At the moment using the client with `TestServer` requires implementation a custom `DataServiceClientRequestMessage` class as described here: https://docs.microsoft.com/en-us/odata/client/using-httpclient

A convenient way to provide an `HttpClient` instance would significantly simplify using the OData client for testing.

### Assemblies affected

Microsoft.OData.Client v7.8.3

ElizabethOkerio:
@JechoJekov, Using <code>  Microsoft.OData.Client v7.8.3 </code> You can set OData Client to use HttpClient like below: 
<pre> 
DefaultContainer context = new DefaultContainer(new Uri("https://localhost:44307/odata/"));
context.HttpRequestTransportMode = HttpRequestTransportMode.HttpClientRequestMessage; 
</pre> 

There is no direct way in which you can create an instance of HttpClient and use it with OData Client. We are looking into this though to see how we can support it. We are also welcoming contributions on this. 

Let me know if this is helpful to you.

chura-sk:
@JechoJekov check my comment "ugly hack" in my issue, it can be applied to achieve your goal as well, meanwhile

gathogojr:
Fixed by https://github.com/OData/odata.net/pull/2431 and https://github.com/OData/odata.net/pull/2883 and

PR #2207, opened 2021-09-30
support injecting httpclient to odata client

### LINKED ISSUE #2184: No way to reuse HTTP connections (keep-alive), since .NET Core
*Reuse of HTTP connections is desired. The current code supports HttpWebRequest with keep-alive=true, but WebRequest doesn't behave since .NET Core accordingly (supposed to be deprecated in .NetCore). The other method - HttpClient - it is new'ed in the HttpClientRequestMessage constructor*

### Assemblies affected

*Microsoft.OData.Client*

### Reproduce steps

*In .NET Core app, just fire any OData request in a loop. Both using HttpWebRequest and HttpClient method will keep spawning new TCP/IP connections (see e.g. in wireshark)*

### Expected result

*TCP/IP connections are reused, according to keep-alive pattern*

### Actual result

*Separate connection is opened for each request*

### Additional detail

*Either make the HttpClient static, or give an option to enforce a pre-created HttpClient to the whole Container class.
Consider relationship to https://github.com/OData/odata.net/issues/2037*

ElizabethOkerio:
### Issues

*This pull request fixes #2184, #2037*

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceClientRequestMessageArgs.cs
+++ b/src/Microsoft.OData.Client/DataServiceClientRequestMessageArgs.cs
@@ -10,6 +10,7 @@ namespace Microsoft.OData.Client
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
+    using System.Net.Http;
 
     /// <summary>
     /// Arguments for creating an instance of DataServiceClientRequestMessage.
@@ -47,6 +48,35 @@ namespace Microsoft.OData.Client
         }
 
         /// <summary>
+        /// Initializes a new instance of the <see cref="DataServiceClientRequestMessageArgs"/> class.
+        /// </summary>
+        /// <param name="method">Method of the request.</param>
+        /// <param name="requestUri">The Request Uri.</param>
+        /// <param name="useDefaultCredentials">True if the default credentials need to be sent with the request. Otherwise false.</param>
+        /// <param name="usePostTunneling">True if the request message must use POST verb for the request and pass the actual verb in X-HTTP-Method header, otherwise false.</param>
+        /// <param name="headers">The set of headers for the request.</param>
+        /// <param name="client">The HttpClient Instance for the request</param>
+        public DataServiceClientRequestMessageArgs(string method, Uri requestUri, bool useDefaultCredentials, bool usePostTunneling, IDictionary<string, string> headers, HttpClient client)
+        {
+            Debug.Assert(method != null, "method cannot be null");
+            Debug.Assert(requestUri != null, "requestUri cannot be null");
+            Debug.Assert(headers != null, "headers cannot be null");
+
+            this.Headers = headers;
+            this.Method = method;
+            this.RequestUri = requestUri;
+            this.UsePostTunneling = usePostTunneling;
+            this.UseDefaultCredentials = useDefaultCredentials;
+            this.HttpClient = client;
+
+            this.actualMethod = this.Method;
+            if (this.UsePostTunneling && this.Headers.ContainsKey(XmlConstants.HttpXMethod))
+            {
+                this.actualMethod = XmlConstants.HttpMethodPost;
+            }
+        }
+
+        /// <summary>
         /// Gets the method.
         /// </summary>
         public string Method { get; private set; }
@@ -75,6 +105,14 @@ namespace Microsoft.OData.Client
             {
                 return this.actualMethod;
             }
+        }
+
+        /// <summary>
+        /// Gets the httpclient instance for requests
+        /// </summary>
+        public HttpClient HttpClient
+        {
+            get;
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -749,7 +749,6 @@ namespace Microsoft.OData.Client
                     HttpClientHandler handler = new HttpClientHandler();
                     this.httpClient = new HttpClient(handler, disposeHandler: true);
                     shouldDisposeClient = true;
-                   // this.httpClient.BaseAddress = null;
                     return this.httpClient;
                 }
             }
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -22,6 +22,7 @@ namespace Microsoft.OData.Client
     using System.Linq;
     using System.Linq.Expressions;
     using System.Net;
+    using System.Net.Http;
     using System.Reflection;
     using System.Threading;
     using System.Threading.Tasks;
@@ -70,7 +71,7 @@ namespace Microsoft.OData.Client
     /// The <see cref="Microsoft.OData.Client.DataServiceContext" /> represents the runtime context of the data service.
     /// </summary>
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Maintainability", "CA1506", Justification = "Central class of the API, likely to have many cross-references")]
-    public class DataServiceContext
+    public class DataServiceContext : IDisposable
     {
         /// <summary>Same version as <see cref="maxProtocolVersion"/> but stored as instance of <see cref="Version"/> for easy comparisons.</summary>
         internal Version MaxProtocolVersionAsVersion;
@@ -168,6 +169,13 @@ namespace Microsoft.OData.Client
         /// <summary>Whether a Where clause that compares only the key property, will generate a $filter query option.</summary>
         private bool keyComparisonGeneratesFilterQuery;
 
+        /// <summary>
+        /// The HttpClient instance to use for requests
+        /// </summary>
+        private HttpClient httpClient;
+        private bool shouldDisposeClient;
+        private bool disposed;
+
         /// <summary>A factory class to use in selecting the the request message transport mode implementation </summary>
         private IDataServiceRequestMessageFactory requestMessageFactory = new DataServiceRequestMessageFactory();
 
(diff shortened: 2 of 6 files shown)


src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -15,6 +15,9 @@ namespace Microsoft.OData.Client
     using System.Net.Http;
     using System.Net.Http.Headers;
     using System.Threading.Tasks;
+#if NETSTANDARD2_0
```
KenitoInc, line 18: What will happen when we target `NETSTANDARD2_1`?

KenitoInc (approved):

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -671,6 +677,22 @@ public virtual EntityTracker EntityTracker
...
+        {
+            get
+            {
+                return this.httpClient;
+            }
+
+            set
+            {
+                this.httpClient = null;
+            }
```
odero, line 693: Why not just have a get only prop instead?

odero (changes requested):

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -4092,5 +4114,20 @@ internal static ClientEdmModel GetModel(ODataProtocolVersion maxProtocolVersion)
...
             }
         }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing)
```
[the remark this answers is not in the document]
marabooy: https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#the-disposebool-method-overload

marabooy (changes requested):

src/Microsoft.OData.Client/DataServiceContext.cs
ElizabethOkerio: will people be able to pass their own instance if I make it only a get prop?

ElizabethOkerio: ok got it

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -71,16 +74,39 @@ public class HttpClientRequestMessage : DataServiceClientRequestMessage, ISendin
...
         {
             _messageStream = new MemoryStream();
-            _handler = new HttpClientHandler();
-            _client = new HttpClient(_handler, disposeHandler: true);
+            if (client == null)
+            {
+#if NETSTANDARD2_0
+                var serviceProvider = new ServiceCollection().AddHttpClient().BuildServiceProvider();
+                var httpClientFactory = serviceProvider.GetService<IHttpClientFactory>();
+                _client = httpClientFactory.CreateClient();
```
habbes, line 99: Since this service provider is created anew for each request, does that also mean we're creating a new client factory for each request? Is this the intended usage of a http client factory?

habbes (changes requested):
I'd assume that there would be changes to the `DataServiceContext`, to allow people to pass the `HttpClient` (or `IHttpClientFactory`) to the DataServiceContext and then the data service context would pass the client down to the `HttpRequestMessage` since users don't generally create `HttpRequestMessage` directly.

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -71,16 +74,39 @@ public class HttpClientRequestMessage : DataServiceClientRequestMessage, ISendin
...
+        /// Constructor for HttpClientRequestMessage that takes a HttpClient instance.
+        /// Initializes the <see cref="HttpClientRequestMessage"/> class.
         /// The args.ActualMethod is the actual method. In post tunneling situations method will be POST instead of the specified verb method.
         /// The args.method is the specified verb method
+        /// The httpClient is an instance of HttpClient that will be used in making requests.
         /// </summary>
-        /// </summary>
-        public HttpClientRequestMessage(DataServiceClientRequestMessageArgs args) 
+        public HttpClientRequestMessage(DataServiceClientRequestMessageArgs args, HttpClient client)
             : base(args.ActualMethod)
```
corranrogue9, line 91: I know this isn't actually your change, but if `args` is `null` this will throw a `NullReferenceException`

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -71,16 +74,39 @@ public class HttpClientRequestMessage : DataServiceClientRequestMessage, ISendin
...
-            _handler = new HttpClientHandler();
-            _client = new HttpClient(_handler, disposeHandler: true);
+            if (client == null)
+            {
+#if NETSTANDARD2_0
+                var serviceProvider = new ServiceCollection().AddHttpClient().BuildServiceProvider();
+                var httpClientFactory = serviceProvider.GetService<IHttpClientFactory>();
+                _client = httpClientFactory.CreateClient();
+#else
+                _handler = new HttpClientHandler();
```
corranrogue9, line 101: `_handler` never gets set for 2.0 which means we will have a `NullReferenceException` thrown by `Credentials`. I think it should something like this instead:
```
_handler = new HttpClientHandler();
#if NETSTANDARD2_0
...
#else
...
#endif
```

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
ElizabethOkerio: there is already a null check for args before a call to this class is made.

corranrogue9: I must be missing something, sorry. `HttpClientRequestMessage` is a `public` class, and this is a `public` constructor, so we don't control all callers, right? So it doesn't really matter if some callers check `args` for `null`; if any other callers pass `null`, we will throw `NullReferenceException`. As a minimal repro, consider this code:
```
public class ThrowBase
{
  public ThrowBase(string foo)
  {
  }
}

public class ThrowDerivedArgs
{
  public string Foo { get; set; }
}

public class ThrowDerived : ThrowBase
{
  public ThrowDerived(ThrowDerivedArgs args)
    : base(args.Foo)
  {
    if (args == null)
    {
      throw new ArgumentNullException(nameof(args));
    }
  }
}
```

`new ThrowDerived(null);` will throw `NullReferenceException` instead of `ArgumentNullException`

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -71,16 +71,24 @@ public class HttpClientRequestMessage : DataServiceClientRequestMessage, ISendin
...
         /// </summary>
-        /// </summary>
-        public HttpClientRequestMessage(DataServiceClientRequestMessageArgs args) 
+        public HttpClientRequestMessage(DataServiceClientRequestMessageArgs args, HttpClient client)
             : base(args.ActualMethod)
         {
             _messageStream = new MemoryStream();
-            _handler = new HttpClientHandler();
-            _client = new HttpClient(_handler, disposeHandler: true);
+            _client = client;
```
corranrogue9, line 91: Shouldn't the behavior of the parameter with no client provided be preserved? Current callers will now have a `null` client. Is that ok?

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -71,16 +74,39 @@ public class HttpClientRequestMessage : DataServiceClientRequestMessage, ISendin
...
+                var httpClientFactory = serviceProvider.GetService<IHttpClientFactory>();
+                _client = httpClientFactory.CreateClient();
+#else
+                _handler = new HttpClientHandler();
+                _client = new HttpClient(_handler, disposeHandler: true);
+#endif
+            }
+            else
+            {
+                _client = client;
```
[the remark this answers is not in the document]
corranrogue9: It is at least needed for the `Credentials` property, maybe others...

src/Microsoft.OData.Client/DataServiceClientRequestMessageArgs.cs
```diff
@@ -46,6 +47,35 @@ public DataServiceClientRequestMessageArgs(string method, Uri requestUri, bool u
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DataServiceClientRequestMessageArgs"/> class.
+        /// </summary>
+        /// <param name="method">Method of the request.</param>
+        /// <param name="requestUri">The Request Uri.</param>
+        /// <param name="useDefaultCredentials">True if the default credentials need to be sent with the request. Otherwise false.</param>
+        /// <param name="usePostTunneling">True if the request message must use POST verb for the request and pass the actual verb in X-HTTP-Method header, otherwise false.</param>
+        /// <param name="headers">The set of headers for the request.</param>
+        /// <param name="client">The HttpClient Instance for the request</param>
+        public DataServiceClientRequestMessageArgs(string method, Uri requestUri, bool useDefaultCredentials, bool usePostTunneling, IDictionary<string, string> headers, HttpClient client)
```
gathogojr, line 59: Can this constructor call the other existing constructor (or vice versa) instead of duplicating the initialization statements

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -71,25 +72,43 @@ public class HttpClientRequestMessage : DataServiceClientRequestMessage, ISendin
...
+                {
+                    _handler = new HttpClientHandler();
+                    _client = new HttpClient(_handler, disposeHandler: true);
+                }
+                else
+                {
+#if (NETCOREAPP || NETSTANDARD)
+                    _handler = (HttpClientHandler)(_client.GetType().BaseType.GetField("_handler", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(_client) as HttpMessageHandler);
+#else
+                    _handler = _client.GetType().BaseType.GetField("handler", BindingFlags.NonPublic | BindingFlags.Instance).GetValue(_client) as HttpClientHandler;
```
gathogojr, line 93: How come it's `.GetField("handler", ...)` on this line while 2 lines above it's `.GetField("_handler", ...)`?

ElizabethOkerio: This depends on the HttpClient implementation on .NETCORE AND .NET. In .netcore the field has the underscore while in .net it doesn't.

inferno31:
Is this pr dead?

KenitoInc:
> Is this pr dead?

This will be released in the next breaking change release.

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
habbes: Is there another alternative that does not tie us to private implementation details like this? This seems very brittle.

habbes: If we really need the `_handler` instance, maybe we should inject `HttpClientHandler` instead of `HttpClient` and then use the injected `HttpClientHandler` to create a new `HttpClient` each time. We should avoid relying on private fields.

habbes: Sinc we're using the `_handler` to make some configuration possible like the `Credentials` property, maybe we should just throw an UnsupportedException or InvalidOperationException if the developer tries to access that property when they have injected a client. The message could mention that in such a scenario, those settings should be done on the handler directly before injecting the `HttpClient` to the DataServicContext.

habbes (changes requested):
If we use an injected `HttpClient`, there are high chances that this client is shared across multiple requests. Our current implementation makes changes to the `HttpClient` which are not thread-safe.

For example, the `HttpClientRequestMessage.Timeout` is intended to set the timeout of that particular request. But it's implemented by modifying `HttpClient.Timeout` which sets the timeout of all requests handled by that client. That's fine when we use the client only for one request, but not when the same client handles multiple request. This property is also not thread-safe, so it's possible that multiple threads handling different requests can try to modify the client's timeout property concurrently leading to race conditions.

Another example is the `HttpClientRequestMessage.Abort` method which currently calls `_client.CancelPendingRequests()`. This method cancels all the pending requests of the client. So it's also something we should change.

These particular examples can probably be solved using cancellation tokens. But in general, we should look out for places where we're modifying the http client inside the `HttpClienRequestMessage` and avoid such changes.

habbes:
This PR has been superseded by #2431

PR #2209, opened 2021-09-30
Provide option for alternative form of delete link request Uri

### LINKED ISSUE #2201: Tests failing while using DeleteLink of Microsoft.OData.Client
I see that a request like https://graph.windows.net/adfabdev3.onmicrosoft.com/devices/c09433ca-bf6d-42e5-8ec4-387c464d98f9/registeredUsers/fa9e1472-dca3-45c2-b9f2-89fc41784448/$ref?api-version=1.6-internal works fine with OData 4.0. So, we want to understand why DeleteLink generates a request as EntitySet/<key>/NavigationPropertyLink?$id=entitySet/<key> instead of EntitySet/<key>/NavigationPropertyLink/<key> for Delete request.  With $id in the query arguments, I see that the OData parser parses this request as TypeSegment/KeySegment/NavigationPropertyLinkSegment and the $id being part of query arguments gets ignored while parsing. 

The problem is Microsoft.OData.Client has DeleteLink method which creates a request with $id in query options and our service throws an exception because $id gets ignored. Previously,  System.Data.Services.Client generated a request like https://graph.windows.net/adfabdev3.onmicrosoft.com/devices/c09433ca-bf6d-42e5-8ec4-387c464d98f9/registeredUsers/fa9e1472-dca3-45c2-b9f2-89fc41784448/$ref but when we switched to support OData 4.0 client proxy and using Microsoft.OData.Client, it is breaking the tests.

### Assemblies affected

Microsoft.OData.Client

gathogojr:
### Issues

*This pull request fixes #2201.*

### Description

Provide option for form of delete link request Uri.

According to the OData spec, a delete link request should have a Uri like:
```
{ServiceUri}/{EntitySet}/{Key}/{NavigationProperty}/$ref?$id={ServiceUri}/{RelatedEntitySet}/{RelatedKey}
```
But in V3 and V4.01, the following alternative form is supported:
```
{ServiceUri}/{EntitySet}/{Key}/{NavigationProperty}/{RelatedKey}/$ref
```

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Client/BaseSaveResult.cs
+++ b/src/Microsoft.OData.Client/BaseSaveResult.cs
@@ -1093,7 +1093,7 @@ namespace Microsoft.OData.Client
         /// <param name="binding">The binding.</param>
         /// <param name="targetResource">The target's entity descriptor.</param>
         /// <returns>The original link uri or one with the target entity key appended.</returns>
-        private static Uri AppendTargetEntityKeyIfNeeded(Uri linkUri, LinkDescriptor binding, EntityDescriptor targetResource)
+        private Uri AppendTargetEntityKeyIfNeeded(Uri linkUri, LinkDescriptor binding, EntityDescriptor targetResource)
         {
             // To delete from a collection, we need to append the key.
             // For example: if the navigation property name is "Purchases" and the resource type is Order with key '1', then this method will generate 'baseuri/Purchases(1)'
@@ -1104,8 +1104,26 @@ namespace Microsoft.OData.Client
 
             Debug.Assert(targetResource != null, "targetResource != null");
             StringBuilder builder = new StringBuilder();
-            builder.Append(UriUtil.UriToString(linkUri));
-            builder.Append(UriHelper.QUESTIONMARK + XmlConstants.HttpQueryStringId + UriHelper.EQUALSSIGN + targetResource.Identity);
+            string uriString = UriUtil.UriToString(linkUri);
+
+            if (this.RequestInfo.Context.DeleteLinkUriOption == DeleteLinkUriOption.RelatedKeyAsSegment)
+            {
+                // Related key segment should appear before /$ref
+                int indexOfLinkSegment = uriString.IndexOf(XmlConstants.UriLinkSegment, StringComparison.Ordinal);
+                builder.Append(indexOfLinkSegment > 0 ? uriString.Substring(0, indexOfLinkSegment - 1) : uriString);
+                this.RequestInfo.Context.UrlKeyDelimiter.AppendKeyExpression(targetResource.EdmValue, builder);
+                if (indexOfLinkSegment > 0)
+                {
+                    builder.Append('/');
+                    builder.Append(XmlConstants.UriLinkSegment);
+                }
+            }
+            else
+            {
+                builder.Append(uriString);
+                builder.Append(UriHelper.QUESTIONMARK + XmlConstants.HttpQueryStringId + UriHelper.EQUALSSIGN + targetResource.Identity);
+            }
+
             return UriUtil.CreateUri(builder.ToString(), UriKind.RelativeOrAbsolute);
         }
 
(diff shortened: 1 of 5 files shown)


marabooy (approved):

src/Microsoft.OData.Client/BaseSaveResult.cs
```diff
@@ -1104,8 +1104,26 @@ private static Uri AppendTargetEntityKeyIfNeeded(Uri linkUri, LinkDescriptor bin
...
             StringBuilder builder = new StringBuilder();
-            builder.Append(UriUtil.UriToString(linkUri));
-            builder.Append(UriHelper.QUESTIONMARK + XmlConstants.HttpQueryStringId + UriHelper.EQUALSSIGN + targetResource.Identity);
+            string uriString = UriUtil.UriToString(linkUri);
+
+            if (this.RequestInfo.Context.DeleteLinkUriOption == DeleteLinkUriOption.RelatedKeyAsSegment)
+            {
+                // Related key segment should appear before /$ref
+                int indexOfLinkSegment = uriString.IndexOf(XmlConstants.UriLinkSegment, StringComparison.Ordinal);
+                builder.Append(indexOfLinkSegment > 0 ? uriString.Substring(0, indexOfLinkSegment - 1) : uriString);
```
habbes, line 1113: why not `>= 0` since 0 can be a valid index?

habbes: Maybe we can use `Span` and slice the string to avoid allocation with `Substring`

src/Microsoft.OData.Client/DeleteLinkUriOption.cs
```diff
@@ -0,0 +1,26 @@
...
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Client
+{
+    /// <summary>
+    /// Used to specify the form of Uri to be used for a delete link request.
+    /// </summary>
+    public enum DeleteLinkUriOption
```
habbes, line 12: Out of curiosity, does this issue with different URI options only affect DELETE requests?

src/Microsoft.OData.Client/DeleteLinkUriOption.cs
```diff
@@ -0,0 +1,26 @@
...
+    {
+        /// <summary>
+        /// Pass the id of the related entity as a query param, i.e.,
+        /// {ServiceUri}/{EntitySet}/{Key}/{NavigationProperty}/$ref?$id={ServiceUri}/{RelatedEntitySet}/{RelatedKey}
+        /// </summary>
+        DollarIdQueryParam = 0,
```
habbes, line 18: I'd suggest the name to be `IdQueryParam`. In case we have situtations where the `$` is optional.

habbes (approved):

src/Microsoft.OData.Client/DeleteLinkUriOption.cs
KenitoInc: `$` sign is only optional in system query options e.g select, top, expand etc

KenitoInc (approved):

xuzhg (approved):
:shipit:

mikepizzo (approved):
Please create an issue to consider changing the default in the next breaking change release.

src/Microsoft.OData.Client/BaseSaveResult.cs
gathogojr: Using `>= 0` would suggest that `uriString` starts with `$ref` which is not practical in this case. So while we can make that change, it's really not a practical scenario

gathogojr: @habbes `Span` (and `ReadOnlySpan`) are not available in NET45 but we could introduce the `System.Memory` package to take care of that. Maybe to keep the change small we could live with that single extra allocation since this code is not on a hot path

PR #2210, opened 2021-10-01
Cache vocabulary annotations if model is immutable

### LINKED ISSUE #2189: Consider caching FindVocabularyAnnotations

### Assemblies affected

### Reproduce steps

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

habbes:
### Issues

*This pull request fixes #2189*

### Description

I added a cache for a model's vocabulary annotations if the model is immutable.

We do not expect the model to change at runtime is most scenarios, but we can't rule out the possibility that it can change. E.g. an `EdmModel` created manually can still be change from the user's code. So I added a flag to indicate whether the model is immutable, and I check the presence flag via the `model.IsImmutable()` extension method in the `model.FindVocabularyAnnotations(element)` method to decide whether to use the cache.

Since we don't have a guarantee that models are immutable in existing services, I thought it would be a breaking change to consider models immutable by default, for this reason, customers would have to manually call `model.MarkAsImmutable()` to indicate that the model is immutable. I find this to be inconvenient since it should be set in probably most of the cases, but I didn't want to risk introducing a breaking change. But we should communicate this to concerned workloads so that they can take advantage of these optimizations (and others that we can make on in the future). I'd like to hear your thoughts about this decision.

I have used the `SetAnnotationValue` to store the flag that indicates to immutable, but also to store the cache associated with a a particular model. I'm not sure if this is the ideal or most efficient way to store this data, but I've found that `SetAnnotationValue` has been used in similar ways, for example to store the edm version, to store the `ClrTypeCache` in WebApi, etc. Let me know if this isn't the intended use. PS: I've tried looking at the implementation of SetAnnotationValue in `EdmDirectValueAnnotationManager` and I don't fully understand what the purpose is.

I'm caching the `FindVocabularyAnnotation(this IEdmModel, IEdmVocabularyAnnotatable)` overload, which seemed to account for the majority of allocations, it's responsible for calling `FindVocabularyAnnotations` for the target element and its base types and from across different models and concatenate them.

But this method is also called by `FindVocabularyAnnotation<T>(IEdmModel, IEdmVocabularyAnnotatble, IEdmTerm, string)` which is also responsible for `OfTypeIterator`. I'm investigating whether caching at this level would be more effective or whether it would unnecessarily lead to more cache entries and cache misses.

### Benchmarks

####Before####

image

####After####

The third column states whether or not the model is immutable. You can see that for the same benchmarks, making the model immutable now leads to about a 5-6% reduction in allocated memory:

image

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1131,6 +1150,56 @@ public static void SetPrimitiveValueConverter(this IEdmModel model, IEdmTypeDefi
...
+        /// of the caller to ensure that the underlying EDM model is not modified after calling
+        /// this method.
+        /// </summary>
+        /// <param name="model"></param>
+        public static void MarkAsImmutable(this IEdmModel model)
+        {
+            EdmUtil.CheckArgumentNull(model, "model");
+
+            // it doesn't matter what value we store, so long as it's not null
+            model.SetAnnotationValue(model, EdmConstants.InternalUri, CsdlConstants.IsImmutable, new object());
```
habbes, line 1165: There are some static objects in `CacheHelper` like `CacheHelper.Unknown`, `CacheHelper.BoxedBool(true/false)` which I initially thought would be good candidates to use here to avoid allocating a new object. But I'm not sure if they were intendent for that purpose. Also not sure if they would make a significant difference since this object is only allocated once for the entire lifetime of the model.

joaocpaiva:
> Since we don't have a guarantee that models are immutable in existing services, I thought it would be a breaking change to consider models immutable by default, for this reason, customers would have to manually call `model.MarkAsImmutable()` to indicate that the model is immutable. I find this to be inconvenient since it should be set in probably most of the cases, but I didn't want to risk introducing a breaking change. But we should communicate this to concerned workloads so that they can take advantage of these optimizations (and others that we can make on in the future). I'd like to hear your thoughts about this decision.

Your approach makes total sense to me @habbes, and avoids introducing a breaking change to be safe.

habbes:
I also did consider caching the `GetDerivedTypeConstraints` which is the one responsible for the majority of the calls to `FindVocabularyAnnotations` based on profiler call stacks. `GetDerivedTypeConstraints` also accounts for a few allocations on top of those by `FindVocabularyAnnotations`, namely `OfTypeIterator`s and enumerators to some extent. I've eliminated the `OfTypeIterator` (which was showing up in CPR), but I haven't eliminated the enumerators from `foreach`. I think that could be addressed [here](https://github.com/OData/odata.net/issues/2190) if feasible.

test/PerformanceTests/SerializationBaselineTests/FileSerializationBenchmarks.cs
```diff
@@ -1,90 +0,0 @@
-﻿using BenchmarkDotNet.Attributes;
-using System.Collections.Generic;
-using System.IO;
-
-namespace SerializationBaselineTests
-{
-    [MemoryDiagnoser]
-    public class FileSerializationBenchmarks
```
habbes, line 8: I combined these benchmarks and the in-memory serialization benchmarks into a single class and instead used the benchmark categories feature to distinguish them

chrisspre:
Ideally the model should be a immutable data structure but that is most definitely a breaking change.

For the next major version we should look into truly treating this a cache (invalidate when updates are made, re-calculate at the first request). I don't think the on/off switch is a maintainable API but necessary for now.

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -293,9 +312,10 @@ public static IEnumerable<IEdmVocabularyAnnotation> FindVocabularyAnnotations(th
 
             List<T> result = null;
 
-            foreach (T annotation in model.FindVocabularyAnnotations(element).OfType<T>())
+            // this loop runs in a hot, we avoid using OfType<T>() to avoid the extra allocations of OfTypeIterator
```
gathogojr, line 315:
```suggestion
            // this loop runs in a hot path, we avoid using OfType<T>() to avoid the extra allocations of OfTypeIterator
```

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -347,12 +367,16 @@ public static IEnumerable<IEdmVocabularyAnnotation> FindVocabularyAnnotations(th
 
             if (EdmUtil.TryGetNamespaceNameFromQualifiedName(termName, out namespaceName, out name))
             {
-                foreach (T annotation in model.FindVocabularyAnnotations(element).OfType<T>())
+                // this loop runs in a hot, we avoid using OfType<T>() to avoid the extra allocations of OfTypeIterator
```
gathogojr, line 370:
```suggestion
                // this loop runs in a hot path, we avoid using OfType<T>() to avoid the extra allocations of OfTypeIterator

```

gathogojr (approved):
LGTM

marabooy (approved):
Looks great.

KenitoInc (approved):

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1131,6 +1156,56 @@ public static void SetPrimitiveValueConverter(this IEdmModel model, IEdmTypeDefi
             model.SetPrimitiveValueConverter(typeDefinition.Definition, converter);
         }
 
+        /// <summary>
+        /// Marks the model as immutable. This may enable optimizations
+        /// that assume the model does not get modified. It is the responsibility
+        /// of the caller to ensure that the underlying EDM model is not modified after calling
+        /// this method.
+        /// </summary>
+        /// <param name="model"></param>
```
joaocpaiva, line 1165: <param name="model">The model involved</param>

src/Microsoft.OData.Edm/VocabularyAnnotationCache.cs
```diff
@@ -0,0 +1,41 @@
...
+            new ConcurrentDictionary<IEdmVocabularyAnnotatable, IEnumerable<IEdmVocabularyAnnotation>>();
+
+        /// <summary>
+        /// Adds vocabulary annotations for the specified <paramref name="element"/> to the cache.
+        /// </summary>
+        /// <param name="element">The element for which to cache annotations.</param>
+        /// <param name="annotations">The vocabulary annotations to cache</param>
+        public void AddVocabularyAnnotations(IEdmVocabularyAnnotatable element, IEnumerable<IEdmVocabularyAnnotation> annotations)
+        {
+            annotationsCache[element] = annotations;
```
joaocpaiva, line 26: annotationsCache.TryAdd() is more efficient than using the assignment operator, because if the value already exists it will proceed, and not acquire internal locks just to assign the same annotations again.

habbes: Thanks. That's a good point.

PR #2212, opened 2021-10-08
Implement asynchronous support in ODataBinaryStreamReader

gathogojr:
### Issues

*This pull request is in partial fulfilment of issue #2019.*

### Description

Implement asynchronous support in **`ODataBinaryStreamReader`** class

- [ ] *Test cases added*

### Additional work necessary


test/FunctionalTests/Microsoft.OData.Core.Tests/ODataBinaryStreamReaderTests.cs
```diff
@@ -0,0 +1,71 @@
...
+            int bytesRead = await this.stream.ReadAsync(buffer, offset, maxLength);
+            buffer.CopyTo(chars, 0);
+
+            return bytesRead;
+        }
+
+        //[Fact]
+        //public void ODataTextStreamReaderShouldRead()
+        //{
+        //    var reader = new ODataBinaryStreamReader(new Func<char[], int, int, int>(this.ReadChars));
```
corranrogue9, line 50: I know that these are commented out, but `reader` in both methods should be wrapped in a `using` block

PR #2216, opened 2021-10-11
#2132: Relative URIs for @oData.context are not supported

### LINKED ISSUE #2132: Relative URIs for @oData.context are not supported
The lib expects an absolute URI in the @oData.context field of an entity. If the server uses relative URIs, parsing of the response fails.

### Assemblies affected

OData .Net lib 7.9.0 and earlier

### Reproduce steps

Use an oData service which uses relative URIs in the @oData.context field (for example SAP oData V4 service; mine is password protected so I can't share it).

### Expected result

The entities should be returned.

### Actual result

Parsing of response fails with an InvalidOperationException: `The response payload is a not a valid response payload. Please make sure that the top level element is a valid Atom or JSON element or belongs to 'http://docs.oasis-open.org/odata/ns/data' namespace.`

### Additional detail

An ODataException is thrown at ODataJsonLightContextUriParser.cs line 86:

```
Uri contextUri;
if (!Uri.TryCreate(contextUriFromPayload, UriKind.Absolute, out contextUri))
{
        throw new ODataException(ODataErrorStrings.ODataJsonLightContextUriParser_TopLevelContextUrlShouldBeAbsolute(contextUriFromPayload));
}
```

The ODataException is eaten at ODataJsonLightPayloadKindDetectionDeserializer on line 57 and an InvalidOperationException is thrown at ODataMaterializer on line 359.

karimabd84:
### Issues

*This pull request fixes #2132 .*
The issue was created from my college https://github.com/chris472

### Description

*The lib expects an absolute URI in the @oData.context field of an entity. If the server uses relative URIs, parsing of the response fails.*

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
@@ -18,6 +18,7 @@ namespace Microsoft.OData.JsonLight
     using Microsoft.OData;
     using Microsoft.OData.Evaluation;
     using Microsoft.OData.Json;
+    using System.IO;
     #endregion Namespaces
 
     /// <summary>
@@ -239,6 +240,15 @@ namespace Microsoft.OData.JsonLight
                 propertyAndAnnotationCollector,
                 isReadingNestedPayload,
                 allowEmptyPayload);
+
+            //https://github.com/OData/odata.net/issues/2132
+            //Relative URIs for @oData.context are not supported #2132
+            //Concate the absoluteUri with the contextUri, when the contextUri is arelativeUri. 
+            Uri contextUri;
+            if (!Uri.TryCreate(contextUriAnnotationValue, UriKind.Absolute, out contextUri))
+            {
+                contextUriAnnotationValue = Path.Combine(this.AbsoluteUri, contextUriAnnotationValue);
+            }
 
             ODataJsonLightContextUriParseResult parseResult = null;
 
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
@@ -11,6 +11,7 @@ namespace Microsoft.OData.JsonLight
     using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
     using System.Globalization;
+    using System.IO;
     using System.Linq;
     using System.Text;
     using System.Threading.Tasks;
@@ -18,7 +19,6 @@ namespace Microsoft.OData.JsonLight
     using Microsoft.OData;
     using Microsoft.OData.Evaluation;
     using Microsoft.OData.Json;
-    using System.IO;
     #endregion Namespaces
 
     /// <summary>
@@ -243,7 +243,9 @@ namespace Microsoft.OData.JsonLight
 
             //Concate the absoluteUri with the contextUri, when the contextUri is a relativeUri. 
             Uri contextUri;
-            if (!Uri.TryCreate(contextUriAnnotationValue, UriKind.Absolute, out contextUri))
+            if (!string.IsNullOrEmpty(this.AbsoluteUri) && 
+                !string.IsNullOrEmpty(contextUriAnnotationValue) &&                 
+                !Uri.TryCreate(contextUriAnnotationValue, UriKind.Absolute, out contextUri))
             {
                 contextUriAnnotationValue = Path.Combine(this.AbsoluteUri, contextUriAnnotationValue);
             }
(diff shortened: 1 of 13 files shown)


src/Microsoft.OData.Core/ODataDeserializer.cs
```diff
@@ -68,6 +68,17 @@ internal IEdmModel Model
...
         }
 
+        /// <summary>
+        /// Get the absolute uri
+        /// </summary>
+        internal string AbsoluteUri
+        {
+            get
+            {
+                return this.inputContext?.MessageReaderSettings?.BaseUri?.AbsoluteUri;
```
corranrogue9, line 78: Given the other properties, I think it is safe to assume `inputContext` is not `null`

corranrogue9 (approved):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -240,6 +241,17 @@ internal static bool TryParsePropertyAnnotation(string propertyAnnotationName, o
                 isReadingNestedPayload,
                 allowEmptyPayload);
 
+            //https://github.com/OData/odata.net/issues/2132
```
KenitoInc, line 244: The first 2 comments are not necessary

KenitoInc (changes requested):
Add tests

tywunon:
Any Idea when this is going to be released?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -240,6 +241,17 @@ internal static bool TryParsePropertyAnnotation(string propertyAnnotationName, o
                 isReadingNestedPayload,
                 allowEmptyPayload);
 
+            //https://github.com/OData/odata.net/issues/2132
+            //Relative URIs for @oData.context are not supported #2132
+            //Concate the absoluteUri with the contextUri, when the contextUri is arelativeUri.
```
tywunon, line 246: Typo? arelativeUri -> a relativeUri

tywunon:
Hey Guys,

any update on this?

I am waiting for this, because a client is not able to switch to absolute paths, and I would like to use this Library for my project. Without this fix, i cannot use it. Can you please review this and release it?

Thanks.

KenitoInc:
> Hey Guys,
> 
> any update on this?
> 
> I am waiting for this, because a client is not able to switch to absolute paths, and I would like to use this Library for my project. Without this fix, i cannot use it. Can you please review this and release it?
> 
> Thanks.

I am waiting for @karimabd84 to add tests

karimabd84:
@KenitoInc Which project can use to add my tests?

KenitoInc:
> @KenitoInc Which project can use to add my tests?

OData.sln has tests

tywunon:
Hey, sorry for bugging you all about this.
But if this can't get through by Monday, I would have to start looking for alternatives.
If you can tell me this will not be the case anyway, I could start looking.

Thanks, and sorry.

KenitoInc:
> Hey, sorry for bugging you all about this. But if this can't get through by Monday, I would have to start looking for alternatives. If you can tell me this will not be the case anyway, I could start looking.
> 
> Thanks, and sorry.

I would to merge and release this PR, unfortunately I can't do that without tests. cc @karimabd84 
Let me see if we can get a resource to do that

tywunon:
Sorry for pushing this. I would just make my life way easier.

tywunon:
> 

And I totally get why you just can't merge this.

src/Microsoft.OData.Core/ODataDeserializer.cs
xuzhg: inputContext is from te ctor and there's a null check but using Debug.Assert.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -240,6 +241,20 @@ internal static bool TryParsePropertyAnnotation(string propertyAnnotationName, o
...
 
+            //Concatenate the metadata uri with the contextUri, when the contextUri is a relativeUri. 
+            Uri contextUri;
+            if (!string.IsNullOrEmpty(this.BaseUri) && 
+                !string.IsNullOrEmpty(contextUriAnnotationValue) &&                 
+                !Uri.TryCreate(contextUriAnnotationValue, UriKind.Absolute, out contextUri))
+            {
+                // If the Base uri string is http://odata.org/test
+                // The MetadataDocumentUri will be http://odata.org/test/$metadata
+                // If the contextUriAnnotation valie is Customers(1)/Name
```
xuzhg, line 252: typo

src/Microsoft.OData.Core/ODataDeserializer.cs
```diff
@@ -68,6 +68,17 @@ internal IEdmModel Model
             }
         }
 
+        /// <summary>
+        /// Get the base uri.
+        /// </summary>
+        internal string BaseUri
```
xuzhg, line 74: >string [](http://example.com/codeflow?start=17&length=6)

why just return "Uri" type?

xuzhg (approved):
:shipit:

KenitoInc (approved):

PR #2217, opened 2021-10-12
Implement asynchronous support in JsonReader

gathogojr:
### Issues

*This pull request is a partial fulfilment of #2019 .*

### Description

Implement asynchronous support in **`JsonReader`**

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ExceptionUtils.cs
+++ b/src/Microsoft.OData.Core/ExceptionUtils.cs
@@ -34,6 +34,7 @@ namespace Microsoft.OData
         /// true if the specified exception can be caught and handled;
         /// false otherwise.
         /// </returns>
+        [DebuggerStepThrough]
         internal static bool IsCatchableExceptionType(Exception e)
         {
             Debug.Assert(e != null, "Unexpected null exception!");
@@ -51,6 +52,7 @@ namespace Microsoft.OData
         /// <param name="value">Argument whose value needs to be checked.</param>
         /// <param name="parameterName">Name of the argument, used for exception message.</param>
         /// <returns>The value</returns>
+        [DebuggerStepThrough]
         internal static T CheckArgumentNotNull<T>([ValidatedNotNull] T value, string parameterName) where T : class
         {
             Debug.Assert(!string.IsNullOrEmpty(parameterName), "!string.IsNullOrEmpty(parameterName)");
@@ -70,6 +72,7 @@ namespace Microsoft.OData
         /// </summary>
         /// <param name="value">Argument whose value needs to be checked.</param>
         /// <param name="parameterName">Name of the argument, used for exception message.</param>
+        [DebuggerStepThrough]
         internal static void CheckArgumentStringNotEmpty(string value, string parameterName)
         {
             Debug.Assert(!string.IsNullOrEmpty(parameterName), "!string.IsNullOrEmpty(parameterName)");
@@ -87,6 +90,7 @@ namespace Microsoft.OData
         /// </summary>
         /// <param name="value">Argument whose value needs to be checked.</param>
         /// <param name="parameterName">Name of the argument, used for exception message.</param>
+        [DebuggerStepThrough]
         internal static void CheckArgumentStringNotNullOrEmpty([ValidatedNotNull] string value, string parameterName)
         {
             Debug.Assert(!string.IsNullOrEmpty(parameterName), "!string.IsNullOrEmpty(parameterName)");
@@ -104,6 +108,7 @@ namespace Microsoft.OData
         /// </summary>
         /// <param name="value">Argument whose value needs to be checked.</param>
         /// <param name="parameterName">Name of the argument, used for exception message.</param>
+        [DebuggerStepThrough]
         internal static void CheckIntegerNotNegative(int value, string parameterName)
         {
             Debug.Assert(!string.IsNullOrEmpty(parameterName), "!string.IsNullOrEmpty(parameterName)");
@@ -121,6 +126,7 @@ namespace Microsoft.OData
         /// </summary>
         /// <param name="value">Argument whose value needs to be checked.</param>
         /// <param name="parameterName">Name of the argument, used for exception message.</param>
+        [DebuggerStepThrough]
         internal static void CheckIntegerPositive(int value, string parameterName)
         {
             Debug.Assert(!string.IsNullOrEmpty(parameterName), "!string.IsNullOrEmpty(parameterName)");
@@ -138,6 +144,7 @@ namespace Microsoft.OData
         /// </summary>
         /// <param name="value">Argument whose value needs to be checked.</param>
         /// <param name="parameterName">Name of the argument, used for exception message.</param>
+        [DebuggerStepThrough]
         internal static void CheckLongPositive(long value, string parameterName)
         {
             Debug.Assert(!string.IsNullOrEmpty(parameterName), "!string.IsNullOrEmpty(parameterName)");
@@ -156,6 +163,7 @@ namespace Microsoft.OData
         /// <param name="value">Argument whose value needs to be checked.</param>
         /// <param name="parameterName">Name of the argument, used for exception message.</param>
         /// <typeparam name="T">Type of the collection. It does not matter.</typeparam>
+        [DebuggerStepThrough]
         internal static void CheckArgumentCollectionNotNullOrEmpty<T>(ICollection<T> value, string parameterName)
         {
             Debug.Assert(!string.IsNullOrEmpty(parameterName), "!string.IsNullOrEmpty(parameterName)");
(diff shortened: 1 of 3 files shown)


KenitoInc (approved):

src/Microsoft.OData.Core/ExceptionUtils.cs
```diff
@@ -34,6 +34,7 @@ internal static class ExceptionUtils
         /// true if the specified exception can be caught and handled;
         /// false otherwise.
         /// </returns>
+        [DebuggerStepThrough]
```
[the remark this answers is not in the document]
gathogojr: @KenitoInc I'd like to apply these attribute to methods that one almost never wants to step into when debugging, `CheckArgumentNotNull` for example. It improves the debugging experience

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonReaderTests.cs
```diff
@@ -120,6 +111,651 @@ public void ShouldUseArrayPoolIfSet()
...
+
+            Assert.Equal("Other", await reader.GetValueAsync());
+            await reader.ReadAsync(); // Read property value
+            Assert.Equal("", await reader.GetValueAsync());
+        }
+
+        [Fact]
+        public async Task ReadNullStringValue()
+        {
+            IJsonStreamReaderAsync reader = new JsonReader(new StringReader("{\"Text\":null,\"Other\":\"\"}"), isIeee754Compatible: false);
```
xuzhg, line 354: >JsonReader [](http://example.com/codeflow?start=48&length=10)

do we dispose JsonReader since it's disposable?

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonReaderTests.cs
```diff
@@ -120,6 +111,651 @@ public void ShouldUseArrayPoolIfSet()
...
+                Assert.Equal(pangram, new string(chars, 0, charsRead));
+            }
+
+            Assert.Equal("Other", await reader.GetValueAsync());
+            await reader.ReadAsync(); // Read property value
+            Assert.Equal("", await reader.GetValueAsync());
+        }
+
+        [Fact]
+        public async Task ReadNullStringValue()
```
xuzhg, line 352: It seems we can put "non-null" and "null" test in to one [Theory] test?

xuzhg (approved):
:shipit:

src/Microsoft.OData.Core/Json/JsonReader.cs
```diff
@@ -505,6 +505,298 @@ public TextReader CreateTextReader()
...
+
+                Debug.Assert(
+                    this.tokenStartIndex < this.storedCharacterCount && !IsWhitespaceCharacter(this.characterBuffer[this.tokenStartIndex]),
+                    "The SkipWhitespacesAsync didn't correctly skip all whitespace characters from the input.");
+            }
+
+            switch (currentScope.Type)
+            {
+                case ScopeType.Root:
+                    if (commaFound)
```
corranrogue9, line 631: I almost feel like this `switch` would be easier to understand if there was one for `commaFound == true` and one for `commaFound == false`. Probably not worth changing now, but maybe something to consider.

src/Microsoft.OData.Core/Json/JsonReader.cs
```diff
@@ -1261,19 +1544,682 @@ private bool ReadInput()
...
+                case 't':
+                case 'f':
+                    this.nodeValue = await this.ParseBooleanPrimitiveValueAsync()
+                        .ConfigureAwait(false);
+                    break;
+
+                default:
+                    // COMPAT 47: JSON number can start with dot.
+                    // The JSON spec doesn't allow numbers to start with ., but WCF DS does. We will follow the WCF DS behavior for compatibility.
+                    if (Char.IsDigit(currentCharacter) || (currentCharacter == '-') || (currentCharacter == '.'))
```
corranrogue9, line 1626: Super minor potential performance change, let `'-'` and `'.'` be cases in the `switch` statement

src/Microsoft.OData.Core/Json/JsonReader.cs
```diff
@@ -1261,19 +1544,682 @@ private bool ReadInput()
...
+                    this.scopes.Peek().ValueCount++;
+
+                    // move to next node
+                    await this.ReadAsync().ConfigureAwait(false);
+                    return charsRead;
+                }
+
+                if (character == this.streamOpeningQuoteCharacter)
+                {
+                    this.balancedQuoteCount++;
```
corranrogue9, line 2036: I think that this is just trying to undo the decrement from line 2014. If that's the case, it would probably make more sense to just have line 2014 read `... && this.balancedQuoteCount - 1 < 1`

src/Microsoft.OData.Core/Json/JsonReader.cs
```diff
@@ -1316,25 +2262,35 @@ private bool ReadInput()
...
+        }
+
+        /// <summary>
+        /// Determines if a given character is allowed in a JSON property name.
+        /// </summary>
+        /// <param name="character">The character to check.</param>
+        /// <returns>true if the character is allowed in a JSON property name; otherwise false.</returns>
+        private bool IsCharacterAllowedInPropertyName(char character)
+        {
+            return character == '_' ||
```
corranrogue9, line 2290: I'd be interested to see a benchmark comparing this with an equivalent `switch` statement

corranrogue9 (approved):

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonReaderTests.cs
gathogojr: In the [Theory] test, it's `\"null\"` (as a string) so we're just verify that it's read as a string. In this test, its just `null` and we're verifying that's its read as `null`. I hope I understood your question correctly

src/Microsoft.OData.Core/Json/JsonReader.cs
gathogojr: @corranrogue9 The `switch` equivalent would look something like this:
```csharp
switch (character)
{
    case '_':
    case '$':
        return true;
    default:
        if (Char.IsLetterOrDigit(character) || (this.allowAnnotations && (character == '.' || character == '@')))
            return true;
        else
            return false;
}
```
I'm not sure I like this alternative better. However, one micro-optimization that is easy to apply is making `Char.IsLetterOrDigit(character)` the first condition to check given that letters and digits are the most common characters in a property name.

src/Microsoft.OData.Core/Json/JsonReader.cs
gathogojr: @corranrogue9 Note that on line 2014 there's the condition that has to be satisfied before the decrement happens is `character == GetClosingQuoteCharacter(this.streamOpeningQuoteCharacter)` which is a little different from the condition for incrementing `character == this.streamOpeningQuoteCharacter`

src/Microsoft.OData.Core/Json/JsonReader.cs
corranrogue9: I would definitely like to see that benchmark if you do it. I think `switch` is supposed to be O(~1), so I think running any code that has branching logic before the `switch` would make it overall slower. 

As for this feedback, you can take it or leave it, just a suggestion

gathogojr: @corranrogue9 I grabbed the csdl for a particular workload that receives relatively high traffic and set up 2 benchmarks - the first benchmark calling a variant of `IsCharacterAllowedInPropertyName` method that uses a `switch` statement and the second benchmark calling a variant using the code in my PR. The idea being to verify that the characters that make up the properties in the different entity and complex types are allowed. Here are the observed benchmark results after 5 separate runs:
```
|         Method |     Mean |    Error |   StdDev |
|--------------- |---------:|---------:|---------:|
| NotUsingSwitch | 77.49 us | 1.244 us | 1.103 us |
|    UsingSwitch | 76.78 us | 1.060 us | 1.041 us |

|         Method |     Mean |    Error |   StdDev |
|--------------- |---------:|---------:|---------:|
| NotUsingSwitch | 59.80 us | 0.191 us | 0.159 us |
|    UsingSwitch | 79.53 us | 1.561 us | 1.603 us |

|         Method |     Mean |    Error |   StdDev |
|--------------- |---------:|---------:|---------:|
| NotUsingSwitch | 77.14 us | 1.537 us | 3.106 us |
|    UsingSwitch | 80.51 us | 1.577 us | 3.112 us |

|         Method |     Mean |    Error |   StdDev |
|--------------- |---------:|---------:|---------:|
| NotUsingSwitch | 95.47 us | 1.397 us | 1.239 us |
|    UsingSwitch | 93.38 us | 1.785 us | 2.321 us |

|         Method |     Mean |    Error |   StdDev |
|--------------- |---------:|---------:|---------:|
| NotUsingSwitch | 61.46 us | 0.993 us | 0.829 us |
|    UsingSwitch | 80.51 us | 1.394 us | 3.029 us |
```

PR #2223, opened 2021-10-19
Performance fix for FindAcrossModels

### LINKED ISSUE #2186: Performance - FilterOperationCandidatesBasedOnParameterList takes >4% of CPU
In AGS **FilterOperationCandidatesBasedOnParameterList** is invoked in hot path, during parser, and it is currently taking more than 4% of CPU.

This stack has been improved in the past, but I think there is still room for more. The cost comes from [UnqualifiedODataUriResolver](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs) 
 **FindAcrossModels()**, invoked by ResolveUnboundOperations and ResolveBoundOperations.

There are a few issues in the code (the top two being the most critical):
1. A lot of Linq enumerable iterations and type casting because of  **refModel.SchemaElements.OfType<T>**
2. O(n) search because of .Where(e => string.Equals(qualifiedName, e.Name, caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal)), making **Equals** show very hot in the stack.
3. Lambda and closure allocations.
4. The .Concat resulting in more allocations.

The main EDM models of AGS (beta and v1.0) are immutable. It seems like it would make sense to cache bound and unbound operations for each EDM model, using relevant inputs for the key (i.e. identifier, bindingType, etc) to make this a very fast O(1) lookup? Even if, cache miss goes through the inefficient path, cache hits will become very fast and address all the issues above. The CPU cost should become barely visible, if at all. Hit ratio should be very high.

image

image

### Assemblies affected

Microsoft.OData.Core.dll 7.9.1

### Reproduce steps

Exercise Microsoft.OData.UriParser.ODataUriParser!ParsePath() repeatedly for the same model.

### Expected result

CPU utilization should be close to 0.

### Actual result

CPU utilization is very high in FilterOperationCandidatesBasedOnParameterList.

@habbes, @syprieur

Sreejithpin:
### Issues

*This pull request fixes #2186 *

### Description

Performance improvement for finding operations inmodel

attached benchmark results

image

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
@@ -178,18 +178,24 @@ namespace Microsoft.OData.UriParser
             }
 
             // Only filter if there is more than one and its needed.
-            if (possibleFunctions.Count() > 1)
+            if(possibleFunctions.HasAny())
             {
-                possibleFunctions = possibleFunctions.FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType(entityType);
-            }
+                int functionsCount = possibleFunctions.Count();
 
-            // If more than one overload matches, try to select based on optional parameters
-            if (possibleFunctions.Count() > 1 && parameterNames.Count > 0)
-            {
-                possibleFunctions = possibleFunctions.FilterOverloadsBasedOnParameterCount(parameterNames.Count);
-            }
+                if(functionsCount > 1)
+                {
+                    possibleFunctions = possibleFunctions.FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType(entityType);
 
-            if (!possibleFunctions.HasAny())
+                    int parameterNamesCount = parameterNames.Count;
+
+                    // If more than one overload matches, try to select based on optional parameters
+                    if (parameterNamesCount > 0)
+                    {
+                        possibleFunctions = possibleFunctions.FilterOverloadsBasedOnParameterCount(parameterNamesCount);
+                    }
+                }
+            }
+            else
             {
                 segment = null;
                 return false;
--- a/src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
@@ -178,24 +178,18 @@ namespace Microsoft.OData.UriParser
             }
 
             // Only filter if there is more than one and its needed.
-            if(possibleFunctions.HasAny())
+            if (possibleFunctions.Count() > 1)
             {
-                int functionsCount = possibleFunctions.Count();
+                possibleFunctions = possibleFunctions.FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType(entityType);
 
-                if(functionsCount > 1)
+                // If more than one overload matches, try to select based on optional parameters
+                if (parameterNames.Count > 0 && possibleFunctions.Count() > 1)
                 {
-                    possibleFunctions = possibleFunctions.FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType(entityType);
-
-                    int parameterNamesCount = parameterNames.Count;
-
-                    // If more than one overload matches, try to select based on optional parameters
-                    if (parameterNamesCount > 0)
-                    {
-                        possibleFunctions = possibleFunctions.FilterOverloadsBasedOnParameterCount(parameterNamesCount);
-                    }
+                    possibleFunctions = possibleFunctions.FilterOverloadsBasedOnParameterCount(parameterNames.Count);
                 }
             }
-            else
+
+            if (!possibleFunctions.HasAny())
             {
                 segment = null;
                 return false;
--- a/src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
@@ -50,26 +50,24 @@ namespace Microsoft.OData.UriParser
             return FindOperations(model, identifier, this.EnableCaseInsensitive, true, bindingType);
         }
 
-        private IEnumerable<IEdmOperation> FindOperations(IEdmModel model, string qualifiedName, bool caseInsensitive, bool isBound = false, IEdmType bindingType = null)
+        private static IEnumerable<IEdmOperation> FindOperations(IEdmModel model, string qualifiedName, bool caseInsensitive, bool isBound = false, IEdmType bindingType = null)
         {
-            List<IEdmOperation> results = new List<IEdmOperation>();
+            IList<IEdmOperation> results = new List<IEdmOperation>();
 
             StringComparison strComparison = caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
 
-            results.AddRange(GetOperationsForModel(model, qualifiedName, isBound, bindingType, strComparison));
+            GetOperationsForModel(model, qualifiedName, results, isBound, bindingType, strComparison);
 
             foreach (IEdmModel reference in model.ReferencedModels)
             {
-                results.AddRange(GetOperationsForModel(reference, qualifiedName, isBound, bindingType, strComparison));
+                GetOperationsForModel(reference, qualifiedName, results, isBound, bindingType, strComparison);
             }
 
             return results;
         }
 
-        private static IList<IEdmOperation> GetOperationsForModel(IEdmModel model, string qualifiedName, bool isBound, IEdmType bindingType, StringComparison strComparison)
+        private static void GetOperationsForModel(IEdmModel model, string qualifiedName, IList<IEdmOperation> results, bool isBound, IEdmType bindingType, StringComparison strComparison)
         {
-            IList<IEdmOperation> results = new List<IEdmOperation>();
-
             foreach (IEdmSchemaElement schemaElement in model.SchemaElements)
             {
                 if (schemaElement is IEdmOperation operation)
@@ -81,8 +79,6 @@ namespace Microsoft.OData.UriParser
                     }
                 }
             }
-
-            return results;
         }
 
     }
--- a/src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
@@ -30,7 +30,7 @@ namespace Microsoft.OData.UriParser
                 return base.ResolveUnboundOperations(model, identifier);
             }
 
-            return FindOperations(model, identifier, this.EnableCaseInsensitive);
+            return GetUnBoundOperationsForModel(model, identifier, this.EnableCaseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal);
         }
 
         /// <summary>
@@ -47,21 +47,7 @@ namespace Microsoft.OData.UriParser
                 return base.ResolveBoundOperations(model, identifier, bindingType);
             }
           
-            return FindOperations(model, identifier, this.EnableCaseInsensitive, true, bindingType);
-        }
-
-        private static IEnumerable<IEdmOperation> FindOperations(IEdmModel model, string qualifiedName, bool caseInsensitive, bool isBound = false, IEdmType bindingType = null)
-        {
-            StringComparison strComparison = caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
-
-            if (isBound)
-            {
-                return GetBoundOperationsForModel(model, qualifiedName, bindingType, strComparison);
-            }
-            else
-            {
-                return GetUnBoundOperationsForModel(model, qualifiedName, strComparison);
-            }
+            return GetBoundOperationsForModel(model, identifier, bindingType, this.EnableCaseInsensitive? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal);
         }
 
         private static IList<IEdmOperation> GetUnBoundOperationsForModel(IEdmModel model, string qualifiedName, StringComparison strComparison)
--- a/src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
@@ -7,6 +7,7 @@
 namespace Microsoft.OData.UriParser
 {
     using System;
+    using System.Collections.Concurrent;
     using System.Collections.Generic;
     using System.Linq;
     using Microsoft.OData.Edm;
@@ -16,7 +17,9 @@ namespace Microsoft.OData.UriParser
     /// </summary>
     public class UnqualifiedODataUriResolver : ODataUriResolver
     {
-         /// <summary>
+        private static readonly ConcurrentDictionary<IEdmModel, ConcurrentDictionary<string, IEnumerable<IEdmSchemaElement>>> schemaElementCache = new ConcurrentDictionary<IEdmModel, ConcurrentDictionary<string, IEnumerable<IEdmSchemaElement>>>();
+
+        /// <summary>
         /// Resolve unbound operations based on name.
         /// </summary>
         /// <param name="model">The model to be used.</param>
@@ -53,21 +56,49 @@ namespace Microsoft.OData.UriParser
                     && operation.Parameters.Any()
                     && operation.HasEquivalentBindingType(bindingType));
         }
-
+               
         private static IEnumerable<T> FindAcrossModels<T>(IEdmModel model, String qualifiedName, bool caseInsensitive) where T : IEdmSchemaElement
         {
-            Func<IEdmModel, IEnumerable<T>> finder = (refModel) =>
-                refModel.SchemaElements.OfType<T>()
-                .Where(e => string.Equals(qualifiedName, e.Name, caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal));
+            IList<T> results = new List<T>();
 
-            IEnumerable<T> results = finder(model);
+            ConcurrentDictionary<string, IEnumerable<IEdmSchemaElement>> nameDict;
+            IEnumerable<IEdmSchemaElement> nameResults;
+
+            if (schemaElementCache.TryGetValue(model, out nameDict) )
+            {
+                if (nameDict.TryGetValue(qualifiedName, out nameResults))
+                {
+                    return nameResults as IList<T>;
+                }                
+            }
+            else
+            {
+                schemaElementCache.TryAdd(model, new ConcurrentDictionary<string, IEnumerable<IEdmSchemaElement>>());
+            }
+            
+            StringComparison strComparison = caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
+
+            GetSchemaElements(model, qualifiedName, results, strComparison);
 
             foreach (IEdmModel reference in model.ReferencedModels)
             {
-                results.Concat(finder(reference));
+                GetSchemaElements(reference, qualifiedName, results, strComparison);
             }
 
+            schemaElementCache[model].TryAdd(qualifiedName, results as IEnumerable<IEdmSchemaElement>);
+
             return results;
+        }
+
+        private static void GetSchemaElements<T>(IEdmModel model, string qualifiedName, IList<T> results, StringComparison strComparison) where T : IEdmSchemaElement
+        {
+            foreach (IEdmSchemaElement edmSchemaElement in model.SchemaElements)
+            {
+                if (string.Equals(qualifiedName, edmSchemaElement.Name, strComparison) && edmSchemaElement is T schemaElement)
+                {
+                    results.Add(schemaElement);
+                }
+            }
         }
     }
 }
--- a/src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
@@ -17,8 +17,6 @@ namespace Microsoft.OData.UriParser
     /// </summary>
     public class UnqualifiedODataUriResolver : ODataUriResolver
     {
-        private static readonly ConcurrentDictionary<IEdmModel, ConcurrentDictionary<string, IList<IEdmOperation>>> operationCache = new ConcurrentDictionary<IEdmModel, ConcurrentDictionary<string, IList<IEdmOperation>>>();
-
         /// <summary>
         /// Resolve unbound operations based on name.
         /// </summary>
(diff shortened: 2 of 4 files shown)


src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
```diff
@@ -47,27 +47,43 @@ public override IEnumerable<IEdmOperation> ResolveBoundOperations(IEdmModel mode
...
         {
-            Func<IEdmModel, IEnumerable<T>> finder = (refModel) =>
-                refModel.SchemaElements.OfType<T>()
-                .Where(e => string.Equals(qualifiedName, e.Name, caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal));
+            List<IEdmOperation> results = new List<IEdmOperation>();
+
+            StringComparison strComparison = caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal;
 
-            IEnumerable<T> results = finder(model);
+            results.AddRange(GetOperationsForModel(model, qualifiedName, isBound, bindingType, strComparison));
```
joaocpaiva, line 59: AddRange is doing extra allocation to copy the list returned from GetOperationsForModel. Better to pass the List of results to GetOperationsForModel and append directly, so we avoid AddRange extra allocations.

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
```diff
@@ -47,27 +47,43 @@ public override IEnumerable<IEdmOperation> ResolveBoundOperations(IEdmModel mode
...
 
             foreach (IEdmModel reference in model.ReferencedModels)
             {
-                results.Concat(finder(reference));
+                results.AddRange(GetOperationsForModel(reference, qualifiedName, isBound, bindingType, strComparison));
```
joaocpaiva, line 63: Same here.

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
```diff
@@ -47,27 +49,55 @@ public override IEnumerable<IEdmOperation> ResolveBoundOperations(IEdmModel mode
...
             return results;
         }
+
+        private static void GetOperationsForModel(IEdmModel model, string qualifiedName, bool isBound, IEdmType bindingType, IList<IEdmOperation> results, StringComparison strComparison)
+        {
+            foreach (IEdmSchemaElement schemaElement in model.SchemaElements)
+            {
+                if (schemaElement is IEdmOperation operation)
+                {
+                    if (string.Equals(qualifiedName, operation.Name, strComparison) && ((!isBound && !operation.IsBound) ||
```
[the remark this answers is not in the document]
joaocpaiva: Agreed. I think it makes sense to defer the string comparison in this evaluation. And it does show very hot in the traces.

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
```diff
@@ -47,27 +49,55 @@ public override IEnumerable<IEdmOperation> ResolveBoundOperations(IEdmModel mode
...
+                foreach (IEdmModel reference in model.ReferencedModels)
+                {
+                    GetOperationsForModel(reference, qualifiedName, isBound, bindingType, results, strComparison);
+                }
+
+                nameDict.TryAdd(qualifiedName, results);
+                operationCache[model] = nameDict;
+            }
+           
             return results;
```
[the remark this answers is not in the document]
joaocpaiva: We could cache ReadOnlyCollection or something similar. We should not create a new instance every time though, as that would defeat the purpose of caching.

joaocpaiva (approved):

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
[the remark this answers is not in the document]
corranrogue9: Well, there would still be some performance benefits even if we create a new list each time, but point taken. If you're referring to `System.Collections.ObjectModel.ReadOnlyCollection`, I am not really a fan for the same reason because it has the issue where someone who wants to be able to modify the underlying items can do so through the `Items` property. But, effectively that is the approach I am suggestion: an `IReadOnlyList` implementation that does not offer a way to get at the items that we are storing internally.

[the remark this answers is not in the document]
joaocpaiva: Items is protected though, but it could be invoked by a caller using reflection for example. I am with you, I meant something that is IReadOnlyList and throws an exception if caller tries to mutate.

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
```diff
@@ -47,27 +47,39 @@ public override IEnumerable<IEdmOperation> ResolveBoundOperations(IEdmModel mode
...
+        private static void GetOperationsForModel(IEdmModel model, string qualifiedName, IList<IEdmOperation> results, bool isBound, IEdmType bindingType, StringComparison strComparison)
+        {
+            foreach (IEdmSchemaElement schemaElement in model.SchemaElements)
+            {
+                if (schemaElement is IEdmOperation operation)
+                {
+                    if (((!isBound && !operation.IsBound) || (isBound && operation.IsBound && operation.Parameters.Any() && operation.HasEquivalentBindingType(bindingType))) &&
+                        string.Equals(qualifiedName, operation.Name, strComparison))
+                    {
+                        results.Add(operation);
```
mikepizzo, line 78: if the result is an IEnumerable<IEdmOperation>, why are we building a list at all, rather than just yielding here?

the list obviously works better for caching, but it's unclear at this point where the right place would be to cache.

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
```diff
@@ -47,27 +47,39 @@ public override IEnumerable<IEdmOperation> ResolveBoundOperations(IEdmModel mode
...
             return results;
         }
+
+        private static void GetOperationsForModel(IEdmModel model, string qualifiedName, IList<IEdmOperation> results, bool isBound, IEdmType bindingType, StringComparison strComparison)
+        {
+            foreach (IEdmSchemaElement schemaElement in model.SchemaElements)
+            {
+                if (schemaElement is IEdmOperation operation)
+                {
+                    if (((!isBound && !operation.IsBound) || (isBound && operation.IsBound && operation.Parameters.Any() && operation.HasEquivalentBindingType(bindingType))) &&
```
mikepizzo, line 75: >(!isBound && !operation.IsBound) [](http://example.com/codeflow?start=25&length=32)

This is currently only called from Resolve*Bound*Operations, so isBound is always true.  Unless we decide to use this method from other places where isBound may be false, we can simplify the logic by removing this check (and removing the isBound parameter from GetOperationsForModel).

mikepizzo, line 75: >operation.IsBound [](http://example.com/codeflow?start=73&length=17)

this is the very first check in HasEquivalentBindingType, so don't really need it here.

mikepizzo, line 75: >operation.Parameters.Any() [](http://example.com/codeflow?start=94&length=26)

the second thing checked in HasEquivalentBindingType is whether FirstOrDefault returns null, so this pre-check is probably also not helping anything.

mikepizzo, line 75: >HasEquivalentBindingType(bindingType) [](http://example.com/codeflow?start=134&length=37)

HasEquivalentBindingType is actually pretty expensive, so probably makes sense to check it after the name comparison.

mikepizzo (changes requested):
:clock1:

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
```diff
@@ -47,27 +47,39 @@ public override IEnumerable<IEdmOperation> ResolveBoundOperations(IEdmModel mode
...
             foreach (IEdmModel reference in model.ReferencedModels)
             {
-                results.Concat(finder(reference));
+                GetOperationsForModel(reference, qualifiedName, results, isBound, bindingType, strComparison);
             }
 
             return results;
         }
+
+        private static void GetOperationsForModel(IEdmModel model, string qualifiedName, IList<IEdmOperation> results, bool isBound, IEdmType bindingType, StringComparison strComparison)
```
mikepizzo, line 69: >qualifiedName [](http://example.com/codeflow?start=74&length=13)

this is not actually the qualified name -- it's the unqualified name (hence this logic, rather than invoking the logic in the base class in line 47)

mikepizzo, line 69: >GetOperationsForModel [](http://example.com/codeflow?start=28&length=21)

IEdmModel has a FindDeclaredBoundOperations that is already cached in the base model implementation.  Rather than loop through all schema elements (which, in a service like MS Graph, is huge), why not just call FindDeclaredBoundOperations and iterate through that list, yielding operations with the right name?

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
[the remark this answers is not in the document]
corranrogue9: Thanks for pointing out the `protected` thing, that was a fun 20 minute rabbit hole. I'd forgotten the intricacies of when you can access `private` vs `protected` from within and without the declaring type :)

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
corranrogue9: I would say there was merit in the original PR, but after the recent changes I agree that this should not be `List` anymore and just use `yield return`

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
joaocpaiva: Ohh I wonder if that expensive O(n) search (loop through all schema elements) will explain most of the insane cost we are seeing from this callstack in AGS ...

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -5,6 +5,7 @@
 //---------------------------------------------------------------------
 
 using System;
+using System.Collections.Concurrent;
```
mikepizzo, line 8: You don't need this reference, do you?

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
```diff
@@ -46,28 +46,50 @@ public override IEnumerable<IEdmOperation> ResolveBoundOperations(IEdmModel mode
...
+                    {
+                        results.Add(operation);
+                    }
+                }
             }
 
             return results;
         }
+
+        private static IList<IEdmOperation> GetBoundOperationsForModel(IEdmModel model, string qualifiedName, IEdmType bindingType, StringComparison strComparison)
```
mikepizzo, line 79: >qualifiedName [](http://example.com/codeflow?start=95&length=13)

please rename to "operationName" to avoid confusion, since this is not the qualified name.

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
```diff
@@ -47,27 +49,55 @@ public override IEnumerable<IEdmOperation> ResolveBoundOperations(IEdmModel mode
...
+            ConcurrentDictionary<string, IList<IEdmOperation>> nameDict;
 
-            IEnumerable<T> results = finder(model);
-
-            foreach (IEdmModel reference in model.ReferencedModels)
+            if (!operationCache.TryGetValue(model, out nameDict))
             {
-                results.Concat(finder(reference));
+                nameDict = new ConcurrentDictionary<string, IList<IEdmOperation>>();
+                operationCache.TryAdd(model, nameDict);
```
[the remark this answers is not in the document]
mikepizzo: As cache code has been removed, this comment is no longer relevant for this PR.

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
```diff
@@ -47,27 +49,55 @@ public override IEnumerable<IEdmOperation> ResolveBoundOperations(IEdmModel mode
...
+
+                GetOperationsForModel(model, qualifiedName, isBound, bindingType, results, strComparison);
+
+                foreach (IEdmModel reference in model.ReferencedModels)
+                {
+                    GetOperationsForModel(reference, qualifiedName, isBound, bindingType, results, strComparison);
+                }
+
+                nameDict.TryAdd(qualifiedName, results);
+                operationCache[model] = nameDict;
```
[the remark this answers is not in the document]
mikepizzo: As cache code has been removed, this comment is no longer relevant for this PR.

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
mikepizzo: yes; findboundoperations should be hugely faster if we don't have to loop through all schema elements.

It's unfortunate that we don't have a FindDeclaredUnboundOperations in IEdmModel. We should add one for V8.x. In the meantime, we could consider adding something to the EdmModelBase and attempt to cast the IEdmModel to EdmModelBase for the common case.

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
```diff
@@ -46,28 +46,50 @@ public override IEnumerable<IEdmOperation> ResolveBoundOperations(IEdmModel mode
...
+
+        private static IList<IEdmOperation> GetBoundOperationsForModel(IEdmModel model, string qualifiedName, IEdmType bindingType, StringComparison strComparison)
+        {
+            IList<IEdmOperation> results = new List<IEdmOperation>();
+
+            foreach (IEdmOperation operation in model.FindDeclaredBoundOperationsAcrossModels(bindingType))
+            {
+                if (string.Equals(qualifiedName, operation.Name, strComparison))
+                {
+                    results.Add(operation);
```
mikepizzo, line 87: >results.Add(operation); [](http://example.com/codeflow?start=20&length=23)

rather than build a list, just yield

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
```diff
@@ -46,28 +46,50 @@ public override IEnumerable<IEdmOperation> ResolveBoundOperations(IEdmModel mode
...
+            return GetBoundOperationsForModel(model, identifier, bindingType, this.EnableCaseInsensitive? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal);
         }
 
-        private static IEnumerable<T> FindAcrossModels<T>(IEdmModel model, String qualifiedName, bool caseInsensitive) where T : IEdmSchemaElement
+        private static IList<IEdmOperation> GetUnBoundOperationsForModel(IEdmModel model, string qualifiedName, StringComparison strComparison)
         {
-            Func<IEdmModel, IEnumerable<T>> finder = (refModel) =>
-                refModel.SchemaElements.OfType<T>()
-                .Where(e => string.Equals(qualifiedName, e.Name, caseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal));
+            IList<IEdmOperation> results = new List<IEdmOperation>();
```
mikepizzo, line 55: >IList<IEdmOperation> results = new List<IEdmOperation>(); [](http://example.com/codeflow?start=12&length=57)

rather than return a list, make the signature IEnumerable<IEdmOperation> and just yield on lines 61 and 71.

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -100,6 +98,31 @@ public static IEnumerable<IEdmOperation> FindBoundOperations(this IEdmModel mode
             return FindAcrossModels(model, bindingType, findBoundOperations, mergeFunctions);  // search built-in EdmCoreModel and CoreVocabularyModel.
         }
 
+        /// Searches for bound operations based on the binding type, across models, returns an empty enumerable if no operation exists.
+        /// </summary>
+        /// <param name="model">The model to search.</param>
+        /// <param name="bindingType">Type of the binding.</param>
+        /// <returns>A set of operations that share the binding type or empty enumerable if no such operation exists.</returns>
+        public static IEnumerable<IEdmOperation> FindDeclaredBoundOperationsAcrossModels(this IEdmModel model, IEdmType bindingType)
```
mikepizzo, line 106: >FindDeclaredBoundOperationsAcrossModels [](http://example.com/codeflow?start=49&length=39)

For consistency, should be named "FindBoundOperationsAcrossModels"

mikepizzo: actually, most of these find across models, so perhaps just "findBoundOperations".  But, "FindBoundOperations" already exists -- what's the difference between the existing findboundoperations and this one?  How would I know when to use one versus the other?

mikepizzo: In particular, if this is just a more efficient FindBoundOperations, we should just change FindBoundOperations to have this improved implementation and not have two extension methods that do the same thing.

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
```diff
@@ -46,28 +46,42 @@ public override IEnumerable<IEdmOperation> ResolveBoundOperations(IEdmModel mode
...
+                }
+            }
 
-            foreach (IEdmModel reference in model.ReferencedModels)
+            foreach(IEdmModel reference in model.ReferencedModels)
             {
-                results.Concat(finder(reference));
+                foreach (IEdmSchemaElement schemaElement in reference.SchemaElements)
+                {
+                    if (schemaElement is IEdmOperation operation && !operation.IsBound && string.Equals(operationName, operation.Name, strComparison))
```
corranrogue9, line 67: I'm assuming the two `if` statements are intended to be an identical predicate for all of the schema elements available. If that's the case, it might make more sense to combine them into one:

```
return model
  .SchemaElements
  .Concat(model.ReferencedModels.SelectMany(reference => reference.SchemaElements))
  .Where(schemaElement => schemaElement is IEdmOperation operation && !operation.IsBound && string.Equals(operationName, operation.Name, strComparison));
```

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -100,6 +98,31 @@ public static IEnumerable<IEdmOperation> FindBoundOperations(this IEdmModel mode
...
+        /// Searches for bound operations based on the binding type, across models, returns an empty enumerable if no operation exists.
+        /// </summary>
+        /// <param name="model">The model to search.</param>
+        /// <param name="bindingType">Type of the binding.</param>
+        /// <returns>A set of operations that share the binding type or empty enumerable if no such operation exists.</returns>
+        public static IEnumerable<IEdmOperation> FindDeclaredBoundOperationsAcrossModels(this IEdmModel model, IEdmType bindingType)
+        {
+            IList<IEdmOperation> bindableOperations = new List<IEdmOperation>();
+
+            foreach (IEdmOperation operation in model.FindDeclaredBoundOperations(bindingType))
```
corranrogue9, line 110: Please check preconditions similar to other methods (this will throw `NullReferenceException` if `model` is `null`)

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -100,6 +98,31 @@ public static IEnumerable<IEdmOperation> FindBoundOperations(this IEdmModel mode
...
+
+            foreach (IEdmModel reference in model.ReferencedModels)
+            {
+                foreach (IEdmOperation operation in reference.FindDeclaredBoundOperations(bindingType))
+                {
+                    bindableOperations.Add(operation);
+                }
+            }
+
+            return bindableOperations;
```
corranrogue9, line 123: This might be a good candidate for `yield return` as well

mikepizzo: +1

src/Microsoft.OData.Core/UriParser/Resolver/UnqualifiedODataUriResolver.cs
mikepizzo: Without knowing how they were optimized, I would be a bit concerned about the overhead of the concat and SelectMany.  General guidance as been to avoid LINQ in perf sensitive scenarios.

src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
```diff
@@ -178,18 +178,24 @@ internal static bool TryBindAsOperation(PathSegmentToken pathToken, IEdmModel mo
...
-            }
+                if(functionsCount > 1)
+                {
+                    possibleFunctions = possibleFunctions.FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType(entityType);
+
+                    int parameterNamesCount = parameterNames.Count;
 
-            if (!possibleFunctions.HasAny())
+                    // If more than one overload matches, try to select based on optional parameters
+                    if (parameterNamesCount > 0)
```
mikepizzo, line 192: I think we may have inadvertently changed the logic here; previously, this was only called if there was more than one possible function *after* calling FilterBoundOperationsWIthSameTypeHierarchyToTypeCLosestToBindingType() and parameterNames.Count > 0.  Now we call this even if there is only one match after calling FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType().  To preserve the old logic, we have to confirm that possibleFunctions is still >1 after callingFilterBoundOperationsWIthSameTypeHierarchyToTypeClosestToBindingType().

Also, previous logic checked to see if there were any functions *after* filtering based on parameter count, where-as the new logic assumes there will be at least one function with the exact parameter count.

But, we should still only check this if the initial count > 1, and we should check parameterNames.Count before possibleFunctions.Count().  And, we probably only have to do the last Any() check if we tried to filter by parameter count.

The more I look at this, the more I think it needs to be entirely rewritten.  In the meantime, I think the following code is a slight improvement and preserves the existing logic:

            // Only filter if there is more than one and its needed.
            if (possibleFunctions.Count() > 1)
            {
                possibleFunctions = possibleFunctions.FilterBoundOperationsWithSameTypeHierarchyToTypeClosestToBindingType(entityType);

                // If there is still more than one overload that matches, try to select based on optional parameters
                if (parameterNames.Count > 0 && possibleFunctions.Count() > 1 )
                {
                    possibleFunctions = possibleFunctions.FilterOverloadsBasedOnParameterCount(parameterNames.Count);
                    if (!possibleFunctions.HasAny())
                    {
                        segment = null;
                        return false;
                    }
                }
            }
            else
            {
                segment = null;
                return false;
            }

src/Microsoft.OData.Core/UriParser/Binders/SelectPathSegmentTokenBinder.cs
```diff
@@ -178,18 +178,24 @@ internal static bool TryBindAsOperation(PathSegmentToken pathToken, IEdmModel mo
             }
 
             // Only filter if there is more than one and its needed.
-            if (possibleFunctions.Count() > 1)
+            if(possibleFunctions.HasAny())
```
mikepizzo, line 181: Not sure whether the cost of doing a HasAny() and then calling Count() if there are any is worth the potential savings in shortcutting the next set of logic if there aren't any.

This whole thing should be rewritten to avoid the multiple iterations. For now, I'd skip doing the initial HasAny().

joaocpaiva: Honestly, calling Count() to start with is better from a performance standpoint:
1. If IEnumerable is a ICollection Count() will cast and invoke the property **Count** on the collection which is relatively cheap.
2. Even if IEnumerable is not a ICollection but the collection is empty the underlying O(n) algorithm has the same cost as invoking HasAny() for an empty collection. If the collection is not empty, then we need to call Count() anyway.

mikepizzo (approved):
:shipit:

PR #2224, opened 2021-10-19
DataServiceEntityAttribute and DataServiceKeyAttribute / maintenance-5.x branch

tapika:
### Description

It should be possible to use DataServiceEntityAttribute and DataServiceKeyAttribute attribute types on interfaces as well, not
only classes.

### DIFF
--- a/WCFDataService/Client/System/Data/Services/Client/Common/DataServiceKeyAttribute.cs
+++ b/WCFDataService/Client/System/Data/Services/Client/Common/DataServiceKeyAttribute.cs
@@ -28,7 +28,7 @@ namespace System.Data.Services.Common
 
     /// <summary>Denotes the key property or properties of an entity. </summary>
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1019:DefineAccessorsForAttributeArguments", Justification = "Accessors are available for processed input.")]
-    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Interface, AllowMultiple = false)]
     public sealed class DataServiceKeyAttribute : System.Attribute
     {
         /// <summary>Name of the properties that form the key.</summary>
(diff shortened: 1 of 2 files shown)


WCFDataService/Client/System/Data/Services/Client/Common/DataServiceKeyAttribute.cs
```diff
@@ -28,7 +28,7 @@ namespace System.Data.Services.Common
 
     /// <summary>Denotes the key property or properties of an entity. </summary>
     [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Design", "CA1019:DefineAccessorsForAttributeArguments", Justification = "Accessors are available for processed input.")]
-    [AttributeUsage(AttributeTargets.Class, AllowMultiple = false)]
+    [AttributeUsage(AttributeTargets.Class | AttributeTargets.Interface, AllowMultiple = false)]
```
corranrogue9, line 31: Is there a test we can add to ensure that the attribute is affecting interfaces correctly?

tapika: I havent found any tests in older branch - do they even exists ?

xuzhg:
ODL v5.x buildable source codes are hosted on ADO. It's better to port it to ADO repo. 
However, ODL v5.x is in maintenance status until we have a hot/security fix to release.

tapika:
> ODL v5.x buildable source codes are hosted on ADO. It's better to port it to ADO repo. However, ODL v5.x is in maintenance status until we have a hot/security fix to release.

Can you post a link where is it ? Tried to google, but could not find.

Ive also integrated source codes directly in here: 
https://github.com/tapika/swupd

Also tried to upgrade it to net6.0.

xuzhg:
It's internal and I don't think it can be accessed from outside.

habbes:
@tapika As @xuzhg has said, no new versions of 5.x will be released except for security fixes. For that reason, we will not merge this PR.

If this change is critical for you, I would suggest either:
- Migrate to the latest version of the library (if that's feasible)
- Or build your fork of the branch that contains the changes (if that's feasible)

PR #2231, opened 2021-10-28
remove single quotes from the type in a cast function

### LINKED ISSUE #1366: Url creating wrong for OData cast function
Hi, just detected an error:

When using the code generated by the OData V4 Client code generator 7.5.1 to filter a entity set, the OData $filter query "cast" function is 
A: added to the url
B: added wrongly to the url.

### Assemblies affected

OData V4 Client code generator 7.5.1

### Reproduce steps

TechService.Container tc = GetTC(); // TechService.Container is a subclass of  global::Microsoft.OData.Client.DataServiceContext generated by the OData V4 Client code generator.
var list = from entityset in tc.ESAllPrim where entityset.PropertyInt16 == ((short)32767) select entityset ;

### Expected result

For A: i would think no cast is required
For B: if cast is added the output should be as follows
"$filter=cast(PropertyInt16,Edm.Int32) eq 32767

(no quotes around Edm.Int32)

### Actual result

For B: the output is as
"$filter=cast(PropertyInt16,'Edm.Int32') eq 32767

(with quotes around Edm.Int32)

but according to the OData specification the type identifier should not be quoted.
ABNF:
castExpr = 'cast' OPEN BWS [ commonExpr BWS COMMA BWS ] qualifiedTypeName BWS CLOSE
qualifiedTypeName = singleQualifiedTypeName                  
                  / 'Collection' OPEN singleQualifiedTypeName CLOSE
singleQualifiedTypeName = qualifiedEntityTypeName 
                        / qualifiedComplexTypeName
                        / qualifiedTypeDefinitionName
                        / qualifiedEnumTypeName
                        / primitiveTypeName 
primitiveTypeName = 'Edm.' ( 'Binary'
                           / 'Boolean'
                           / 'Byte'
                           / 'Date' 
                           / 'DateTimeOffset'
                           / 'Decimal'
                           / 'Double'

thanks and kind regards,
Sven

ElizabethOkerio:
### Issues

*This pull request fixes #1366 .*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
+++ b/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
@@ -610,10 +610,7 @@ namespace Microsoft.OData.Client
                                 this.Visit(u.Operand);
                                 this.builder.Append(UriHelper.COMMA);
                             }
-
-                            this.builder.Append(UriHelper.QUOTE);
                             this.builder.Append(UriHelper.GetTypeNameForUri(u.Type, this.context));
-                            this.builder.Append(UriHelper.QUOTE);
                             this.builder.Append(UriHelper.RIGHTPAREN);
                         }
                     }
--- a/test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientQueryTests.cs
+++ b/test/EndToEndTests/Tests/Client/Build.Desktop/ClientTests/ClientQueryTests.cs
@@ -109,10 +109,10 @@ namespace Microsoft.Test.OData.Tests.Client
         [Fact]
         public void PrimitiveTypeInRequestUrlTest()
         {
-            const string stringOfCast = "cast(PersonId,'Edm.Byte')";
+            const string stringOfCast = "cast(PersonId,Edm.Byte)";
             var context = this.CreateContext().Context;
 
-            //GET http://jinfutanodata01:9090/AstoriaDefault635157546921762475/Person()?$filter=cast(cast(PersonId,'Edm.Byte'),'Edm.Int32')%20gt%200 HTTP/1.1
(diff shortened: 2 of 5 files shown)


src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
```diff
@@ -611,10 +611,18 @@ internal override Expression VisitUnary(UnaryExpression u)
                                 this.builder.Append(UriHelper.COMMA);
                             }
 
-                            this.builder.Append(UriHelper.QUOTE);
-                            this.builder.Append(UriHelper.GetTypeNameForUri(u.Type, this.context));
-                            this.builder.Append(UriHelper.QUOTE);
-                            this.builder.Append(UriHelper.RIGHTPAREN);
+                            if (this.builder.ToString().Contains(UriHelper.CAST))
```
corranrogue9, line 614: From line 606, it seems like we can't get here without `this.builder` containing "cast". If that is correct, then it seems like we are also missing the negative test that takes the `else` branch

corranrogue9: I'm not sure what you mean. To get to this line, 606 had to be called, which adds "cast" to the builder. Nothing that I can see between lines 606 and 614 will remove "cast" from the builder, so this `if` statement will always evaluate to `true` and we never take the else branch.

ElizabethOkerio: yes I think the if--else is not necessary here.

corranrogue9 (approved):

gathogojr (approved):
:shipit:

PR #2232, opened 2021-10-28
Enable property case insensitive reading

xuzhg:
### Issues

### Description

* This PR enables to read the property with the different cases as its property type.

```json
{
   "name": "ABC"
}
```

1) If customer enable "reading property case-insensitive", ODL reads the property value as string value.
2) if customer doesn't enable the setting, ODL reads the property value as untyped value.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
+++ b/src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
@@ -46,6 +46,7 @@ namespace Microsoft.OData
             this.EnableCharactersCheck = false;
             this.Version = odataVersion;
             this.LibraryCompatibility = ODataLibraryCompatibility.Latest;
+            this.EnablePropertyNameCaseInsensitive = false; // for back-compatible
 
             Validator = new ReaderValidator(this);
             if (odataVersion < ODataVersion.V401)
@@ -116,6 +117,11 @@ namespace Microsoft.OData
                 this.baseUri = UriUtils.EnsureTaillingSlash(value);
             }
         }
+
+        /// <summary>
+        /// Gets or sets a boolean value indicating whether or not supports property name case insensitive in reading.
+        /// </summary>
+        public bool EnablePropertyNameCaseInsensitive { get; set; }
 
         /// <summary>
         /// Gets or sets custom type resolver used by the Client.
@@ -288,6 +294,7 @@ namespace Microsoft.OData
             this.Version = other.Version;
             this.ReadAsStreamFunc = other.ReadAsStreamFunc;
             this.ArrayPool = other.ArrayPool;
+            this.EnablePropertyNameCaseInsensitive = other.EnablePropertyNameCaseInsensitive;
         }
     }
 }
--- a/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
+++ b/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
@@ -1906,6 +1906,62 @@ namespace Microsoft.OData.Edm
         /// <summary>
         /// Finds a property from the definition of this reference.
         /// </summary>
+        /// <param name="structuredType">Reference to the calling object.</param>
+        /// <param name="propertyName">Name of the property to find.</param>
+        /// <param name="caseInsensitive">Property name case-sensitive or not.</param>
+        /// <returns>The requested property if it exists. Otherwise, null.</returns>
+        public static IEdmProperty FindProperty(this IEdmStructuredTypeReference structuredType, string propertyName, bool caseInsensitive)
+        {
+            EdmUtil.CheckArgumentNull(structuredType, "structuredType");
+            return structuredType.StructuredDefinition().FindProperty(propertyName, caseInsensitive);
+        }
+
+        /// <summary>
+        /// Finds a property from the definition of this reference.
+        /// </summary>
+        /// <param name="structuredType">Reference to the calling object.</param>
+        /// <param name="propertyName">Name of the property to find.</param>
+        /// <param name="caseInsensitive">Property name case-sensitive or not.</param>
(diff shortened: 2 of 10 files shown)


src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1903,6 +1903,62 @@ public static IEnumerable<IEdmStructuralProperty> StructuralProperties(this IEdm
...
         }
 
+        /// <summary>
+        /// Finds a property from the definition of this reference.
+        /// </summary>
+        /// <param name="structuredType">Reference to the calling object.</param>
+        /// <param name="propertyName">Name of the property to find.</param>
+        /// <param name="caseInsensitive">Property name case-sensitive or not.</param>
+        /// <returns>The requested property if it exists. Otherwise, null.</returns>
+        public static IEdmProperty FindProperty(this IEdmStructuredTypeReference structuredType, string propertyName, bool caseInsensitive)
```
corranrogue9, line 1913: Should the existing `FindProperty` overloads delegate to these new ones?

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1903,6 +1903,62 @@ public static IEnumerable<IEdmStructuralProperty> StructuralProperties(this IEdm
...
+            IEdmProperty edmProperty = structuredType.FindProperty(propertyName);
+            if (edmProperty != null || !caseInsensitive)
+            {
+                return edmProperty;
+            }
+
+            // So, we can't find edm property whose name is exactly same as "propertyName"
+            // Seach again using case-insensitive
+            IEdmProperty[] foundProperties = structuredType.Properties()
+                .Where(p => p.Name.Equals(propertyName, StringComparison.OrdinalIgnoreCase)).ToArray();
```
corranrogue9, line 1948: Is there any situation where `p.Name` is `null`? It might be safer to use `string.Equals(p.Name, propertyName, ...)` just in case

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1903,6 +1903,62 @@ public static IEnumerable<IEdmStructuralProperty> StructuralProperties(this IEdm
...
+            if (foundProperties.Length < 1)
+            {
+                return null;
+            }
+            else if (foundProperties.Length == 1)
+            {
+                return foundProperties[0];
+            }
+
+            throw new InvalidOperationException(Edm.Strings.MultipleMatchingPropertiesFound(propertyName, structuredType.FullTypeName()));
```
corranrogue9, line 1959: Is it truly an exceptional case here? Or is it possible that we allocate somewhat large arrays but only need to determine if there are 0, 1, or 2 elements?

corranrogue9 (approved):

src/Microsoft.OData.Core/ODataMessageReaderSettings.cs
```diff
@@ -46,6 +46,7 @@ public ODataMessageReaderSettings(ODataVersion odataVersion)
             this.EnableCharactersCheck = false;
             this.Version = odataVersion;
             this.LibraryCompatibility = ODataLibraryCompatibility.Latest;
+            this.EnablePropertyNameCaseInsensitive = false; // for back-compatible
```
gathogojr, line 49: @xuzhg Will we make case-insensitive comparison the default in 8.x major release? If yes, maybe you could create an issue to track that and tag it appropriately

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1903,6 +1903,62 @@ public static IEnumerable<IEdmStructuralProperty> StructuralProperties(this IEdm
...
+            // But for "Title", any property name, such as "tiTle", "Title", "title", etc returns #3 property.
+            IEdmProperty edmProperty = structuredType.FindProperty(propertyName);
+            if (edmProperty != null || !caseInsensitive)
+            {
+                return edmProperty;
+            }
+
+            // So, we can't find edm property whose name is exactly same as "propertyName"
+            // Seach again using case-insensitive
+            IEdmProperty[] foundProperties = structuredType.Properties()
```
gathogojr, line 1947: Maybe just like `FindProperty` we could introduce a `FindProperties(string, bool)` and move some of the logic there. Also makes it reusable

gathogojr (approved):
Left a few comments; otherwise :shipit:

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -1903,6 +1903,62 @@ public static IEnumerable<IEdmStructuralProperty> StructuralProperties(this IEdm
...
+            if (foundProperties.Length < 1)
+            {
+                return null;
+            }
+            else if (foundProperties.Length == 1)
+            {
+                return foundProperties[0];
+            }
+
+            throw new InvalidOperationException(Edm.Strings.MultipleMatchingPropertiesFound(propertyName, structuredType.FullTypeName()));
```
mikepizzo, line 1959: Rather than do ToArray, then get length and initial value, just loop through and error if you hit a second match:


foreach (IEdmProperty property in structuredType.Properties)
{
   if (String.Equals(property.Name, propertyName, StringComparison.OrdinalIgnoreCase)
   {
       if (edmProperty != null)
       {
           throw new InvalidOperationException(Edm.Strings.MultipleMatchingPropertiesFound(propertyName, structuredType.FullTypeName()));
       }

        edmProperty = property;
   }
}

return property;

PR #2237, opened 2021-11-03
optimized some counting so that we don't enumerate multiple times

corranrogue9:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsVocabularyAnnotation.cs
```diff
@@ -577,5 +575,31 @@ private IEnumerable<IEdmOperation> FindParameterizedOperationFromList(IEnumerabl
...
+            using (var enumerator = source.GetEnumerator())
+            {
+                int length;
+                for (length = 0; length < 2 && enumerator.MoveNext(); ++length)
+                {
+                }
+
+                if (length == 1)
+                {
+                    value = enumerator.Current;
```
Sreejithpin, line 594: If length ==1, enumerator.Current will be default/null, right? If length is 1 , we are trying to return the first funtion, right.

PR #2247, opened 2021-11-10
Fix memory leak

### LINKED ISSUE #2234: Memory leaks in DataServiceContext.FromAsync methods
Closure capturing in cancellationToken

Now code in **DataServiceContext.FromAsync** methods looks like:
`IAsyncResult asyncResult = beginMethod(callback, state);`        
`cancellationToken.Register(() => this.CancelRequest(asyncResult));`
`cancellationToken.ThrowIfCancellationRequested();`
`return asyncResult;`

How it should looks:
`IAsyncResult asyncResult = beginMethod(callback, state);`        
`var registration = cancellationToken.Register(() => this.CancelRequest(asyncResult));`

and, of course:

`registration.Dispose();`

after endMethod

KenitoInc:
### Issues

*This pull request fixes #2234 .*

### Description

*Current problem*
`CancellationTokenRegistration` is a callback delegate that has been registered with a `CancellationToken`. https://docs.microsoft.com/en-us/dotnet/api/system.threading.cancellationtokenregistration?view=net-5.0
We are not calling `CancellationTokenRegistration.Dispose()` to release all resources used by the current instance of the `CancellationTokenRegistration`.
In this case, what happens is that the callback is not unregistered.

It's also important to note, the cancellation callback is registered with the CancellationTokenSource (see https://referencesource.microsoft.com/#mscorlib/system/threading/CancellationToken.cs,31a5196df8e4df6d), not with CancellationToken. So, if CancellationTokenRegistration.Dispose() is not correctly scoped, the registration will remain active for the lifetime of the parent CancellationTokenSource object. This may lead to an unexpected callback when the scope of the async operation is over.

*Solution*
When a callback is registered to a CancellationToken (`CancellationToken.Register(Action)`), the returned object is a `CancellationTokenRegistration`. This is a light struct type that is IDiposable, and disposing this registration object causes the callback to be deregistered. A guarantee is made that after the Dispose() method has returned, the registered callback is neither running nor will subsequently commence. Source https://devblogs.microsoft.com/pfxteam/net-4-cancellation-framework/

The `using` pattern is a convenient way to make sure CancellationTokenRegistration.Unregister() is called automatically.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -2784,8 +2784,11 @@ namespace Microsoft.OData.Client
                 (callback, state) =>
                 {
                     IAsyncResult asyncResult = beginMethod(callback, state);
-                    cancellationToken.Register(() => this.CancelRequest(asyncResult));
-                    cancellationToken.ThrowIfCancellationRequested();
+                    using (CancellationTokenRegistration cancellationTokenRegistration = cancellationToken.Register(() => this.CancelRequest(asyncResult)))
+                    {
+                        cancellationToken.ThrowIfCancellationRequested();
+                    }
+
                     return asyncResult;
                 },
                 endMethod,
@@ -2814,8 +2817,11 @@ namespace Microsoft.OData.Client
                 (arg1, callback, state) =>
                 {
                     IAsyncResult asyncResult = beginMethod(arg1, callback, state);
-                    cancellationToken.Register(() => this.CancelRequest(asyncResult));
-                    cancellationToken.ThrowIfCancellationRequested();
+                    using (CancellationTokenRegistration cancellationTokenRegistration = cancellationToken.Register(() => this.CancelRequest(asyncResult)))
+                    {
+                        cancellationToken.ThrowIfCancellationRequested();
+                    }
+
                     return asyncResult;
                 },
                 endMethod,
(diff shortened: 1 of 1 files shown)


gathogojr (approved):
:shipit:

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2784,8 +2784,11 @@ public virtual bool TryGetUri(object entity, out Uri identity)
                 (callback, state) =>
                 {
                     IAsyncResult asyncResult = beginMethod(callback, state);
-                    cancellationToken.Register(() => this.CancelRequest(asyncResult));
-                    cancellationToken.ThrowIfCancellationRequested();
+                    using (CancellationTokenRegistration cancellationTokenRegistration = cancellationToken.Register(() => this.CancelRequest(asyncResult)))
```
corranrogue9, line 2787: I'm only just learning about `CancellationTokenRegistration`s, but I think this is not the desired behavior for us. We want the callback to be executed when the token is cancelled, but by disposing the registration, I think that callback is unregistered. It appears that disposing the `CancellationTokenSource` disposes all associated registrations, so I think the memory leak would be correct (without changing existing functionality) if we would call `Dispose` on all appropriate `CancellationTokenSource`s. Am I grossly misunderstanding this?

xuzhg (approved):

src/Microsoft.OData.Client/DataServiceContext.cs
KenitoInc: In this case, our goal is to ensure `CancellationTokenRegistration.Unregister` is called, that's why it's important to scope the disposable `CancellationTokenRegistration` with `using`.

In these internal `FromAsync` methods,, we are passing a `CancellationToken` not a `CancellationTokenSource`.

marabooy (approved):

odero (approved):

KenitoInc:
@EvgenyGrishnov work on a PR, we will review

PR #2250, opened 2021-11-16
Support writing nested delta resource sets with non-delta writer.

### LINKED ISSUE #2252: Cannot write a nested delta payload to a top-level resource
A valid response to PATCH against a single resource with a collection is a response containing a nested delta resource set. However, ODL doesn't allow you to write a top-level resource with a nested DeltaResourceSet.

### Assemblies affected

OData.Net 7.x

### Reproduce steps

1. Create a resource writer
2. Call WriteStart(resource)
3. Call WriteStart(nestedResourceInfo)
4. Call WriteStart(deltaResourceSet)

### Expected result

Successfully write the resource with the nested resource set as an nested delta collection.

### Actual result

An error, since you can't write a delta resource set if you start with a resource writer.

### Additional detail

Today that is not possible because you can only write a delta response if you declare in your top-level resource set writer that you are writing a delta payload.

There is no longer any reason to declare up-front whether a writer is writing a delta request/response versus a non-delta request/response. A user should be able to call WriteStartDeltaResourceSet and write a delta resource set at the top level or nested within a resource.

mikepizzo:
### Issues

*This pull request fixes #2252 *

### Description
A valid response to PATCH against a single resource with a collection is a response containing a nested delta resource set.

Today that is not possible because you can only write a delta response if you declare in your top-level resource set writer that you are writing a delta payload.

There is no longer any reason to declare up-front whether a writer is writing a delta request/response versus a non-delta request/response. A user should be able to call WriteStartDeltaResourceSet and write a delta resource set at the top level or nested within a resource.

This PR enables writing a delta resource set within a non-delta resource set writer. It retains the checks to verify that a deleted resource is only supported within a delta resource set, and an added/deleted link within a top-level delta resource set. Everything else should be common between a resourceSet and deltaResourceSet, at the top level or nested within a resource.

- [x] *Test cases added*

### Additional work necessary

*In 8.0, remove the writingDelta flag from creating resource set writer*


src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -3264,8 +3269,6 @@ await thisParam.StartResourceSetAsync(resourceSetParam)
         private async Task VerifyCanWriteStartDeltaResourceSetAsync(bool synchronousCall, ODataDeltaResourceSet deltaResourceSet)
         {
             ExceptionUtils.CheckArgumentNotNull(deltaResourceSet, "deltaResourceSet");
-
-            this.VerifyWritingDelta();
```
Sreejithpin, line 3268: Curious why was this removed

xuzhg (approved):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
```diff
@@ -811,7 +810,7 @@ protected override void StartDeletedResource(ODataDeletedResource resource)
             {
                 // Writing a deleted resource within an entity set
                 DeltaResourceSetScope deltaResourceSetScope = this.ParentScope as DeltaResourceSetScope;
-                Debug.Assert(deltaResourceSetScope != null, "Writing child of delta set and parent scope is not DeltaResourceSetScope");
+                Debug.Assert(this.ParentScope.State == WriterState.Start || deltaResourceSetScope != null, "Writing child of delta set and parent scope is not DeltaResourceSetScope");
```
xuzhg, line 813: shall we also update the error sentence to refresh the changes?

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -33,7 +33,7 @@ internal abstract class ODataWriterCore : ODataWriter, IODataOutputInStreamError
         private readonly bool writingResourceSet;
 
         /// <summary>True if the writer was created for writing a delta response; false otherwise.</summary>
-        private readonly bool writingDelta;
+        private readonly Stack<bool> writingDelta = new Stack<bool>();
```
xuzhg, line 36: Why cannot we put the status into the Scope?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
```diff
@@ -2252,6 +2252,81 @@ public void WriteNestedDeltasIn41(bool isResponse)
...
+            {
+                messageWriter = new ODataMessageWriter((IODataResponseMessage)message, settings, model);
+            }
+            else
+            {
+                messageWriter = new ODataMessageWriter((IODataRequestMessage)message, settings, model);
+            }
+
+            var writer = messageWriter.CreateODataResourceWriter(this.GetCustomers(), this.GetCustomerType());
+//            ODataJsonLightWriter writer = new ODataJsonLightWriter(GetV401OutputContext(isResponse), this.GetCustomers(), this.GetCustomerType(), false, false, true);
```
xuzhg, line 2308: remove unused?

xuzhg: or why don't use this line code?

src/Microsoft.OData.Core/ODataWriterCore.cs
mikepizzo: Previously, this checked to see whether the writer was created with isDelta=true.

Such a check was unnecessary, and was legacy from when we had different writers under the covers for delta versus non delta collections.

It should be valid to start writing a delta resource set regardless of how the writer was created. In fact, calling writeStartDeltaResourceSet is what sets the writingDelta to true. The only thing writingDelta should validate is whether we are currently in the context of writing a delta collection (in which case we can write a deletedResource or added/deleted link) or we are not (in which case we can't).

This is particularly important for enabling writing a nested delta set within a single resource (for example, for a patch request/response) since there is no way to specify writingDelta when creating a reader/writer for a single resource.

Sreejithpin (approved):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
mikepizzo: I think the error text is still valid (and even more correct than previously).  Previously the assert only checked that the parent scope was a delta resource set scope, not that what was being written was being written as the child of a delta set.  The new assert validates that either the parent is scope is a deltaResourceSetScope or we're not attempting to write a child of a delta set (i.e., if the current state is start then we're writing at the top level, not as a child of a delta set.)

src/Microsoft.OData.Core/ODataWriterCore.cs
mikepizzo: We probably could, I was just minimizing the change for this PR, with the knowledge that this is a good candidate for clean-up in 8.0.

xuzhg (approved):

PR #2253, opened 2021-11-18
Cache DuplicateNameChecker in the OutputContext

### LINKED ISSUE #2180: Performance - DuplicatePropertyNameCheck costing 1.5% of allocations
This object is costing 1.5% of allocations in AGS. This relative cost will increase once AGS starts using OData 7.9.1 because there are other heavy allocations fixed in that version.

Most of the allocation cost comes from initializing and resizing the internal **propertyState** dictionary. 

Would it be possible to reuse instances (ideally a single instance) of **DuplicatePropertyNameChecker** throughout the lifetime of each response being serialized? Right now multiple instances are created per response. If we were able to reuse the same instance, throughout the response, it will significantly reduce the number of allocations these allocations.

An object pool would be another option, but better than a pool would be creating create a single instance of **DuplicatePropertyNameChecker** for each response, if that was viable/possible.

image

image

### Assemblies affected

Microsoft.OData.Core 7.9.1

### Reproduce steps

Write a response using **ODataMessageWriter**

### Expected result

Allocation cost of DuplicatePropertyNameChecker should be close to 0% at scale.

### Actual result

Allocation cost of DuplicatePropertyNameChecker is roughly 1.5% at scale.

@habbes

KenitoInc:
### Issues

*This pull request fixes #2180 .*

### Description

When serializing a response, we are calling `[Serializer].CreateDuplicatePropertyNameChecker()` multiple times to create an instance of the `DuplicatePropertyNameChecker` each time. This causes lots of allocations from initializing and resizing the internal `propertyState` dictionary.

In this PR, we create an Instance of the `DuplicatePropertyNameChecker` when we create the `OutputContext`. This allows us to re-use the `DuplicatePropertyNameChecker` throughout the response serialization process.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/DuplicatePropertyNameChecker.cs
+++ b/src/Microsoft.OData.Core/DuplicatePropertyNameChecker.cs
@@ -120,7 +120,10 @@ namespace Microsoft.OData
         /// </summary>
         public void Reset()
         {
-            propertyState.Clear();
+            if (propertyState.Count > 0)
+            {
+                propertyState.Clear();
+            }
         }
     }
 
--- a/src/Microsoft.OData.Core/Json/JsonLightInstanceAnnotationWriter.cs
+++ b/src/Microsoft.OData.Core/Json/JsonLightInstanceAnnotationWriter.cs
@@ -206,6 +206,10 @@ namespace Microsoft.OData
                     expectedType,
                     treatLikeOpenProperty,
                     this.valueSerializer.JsonLightOutputContext.DuplicatePropertyNameChecker);
+
+                // Reset the PropertyState Dictionary in the DuplicatePropertyNameChecker
+                this.valueSerializer.JsonLightOutputContext.DuplicatePropertyNameChecker.Reset();
+
                 return;
             }
 
--- a/src/Microsoft.OData.Core/Json/JsonLightInstanceAnnotationWriter.cs
+++ b/src/Microsoft.OData.Core/Json/JsonLightInstanceAnnotationWriter.cs
@@ -205,7 +205,7 @@ namespace Microsoft.OData
                 this.valueSerializer.WriteResourceValue(resourceValue,
                     expectedType,
                     treatLikeOpenProperty,
-                    this.valueSerializer.CreateDuplicatePropertyNameChecker());
+                    this.valueSerializer.JsonLightOutputContext.DuplicatePropertyNameChecker);
                 return;
             }
 
@@ -394,7 +394,7 @@ namespace Microsoft.OData
                 await this.valueSerializer.WriteResourceValueAsync(resourceValue,
                     expectedType,
                     treatLikeOpenProperty,
-                    this.valueSerializer.CreateDuplicatePropertyNameChecker()).ConfigureAwait(false);
+                    this.valueSerializer.JsonLightOutputContext.DuplicatePropertyNameChecker).ConfigureAwait(false);
                 return;
             }
 
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
@@ -82,7 +82,7 @@ namespace Microsoft.OData.JsonLight
                         property,
                         null /*owningType*/,
                         true /* isTopLevel */,
-                        this.CreateDuplicatePropertyNameChecker(),
+                        this.JsonLightOutputContext.DuplicatePropertyNameChecker,
                         null /* metadataBuilder */);
                     this.JsonLightValueSerializer.AssertRecursionDepthIsZero();
 
@@ -280,7 +280,7 @@ namespace Microsoft.OData.JsonLight
                         property,
                         null /*owningType*/,
                         true /* isTopLevel */,
-                        this.CreateDuplicatePropertyNameChecker(),
+                        this.JsonLightOutputContext.DuplicatePropertyNameChecker,
                         null /* metadataBuilder */).ConfigureAwait(false);
                     this.JsonLightValueSerializer.AssertRecursionDepthIsZero();
 
@@ -649,7 +649,7 @@ namespace Microsoft.OData.JsonLight
                 resourceValue,
                 this.currentPropertyInfo.MetadataType.TypeReference,
                 isOpenPropertyType,
-                this.CreateDuplicatePropertyNameChecker());
+                this.JsonLightOutputContext.DuplicatePropertyNameChecker);
         }
 
         /// <summary>
@@ -1074,7 +1074,7 @@ namespace Microsoft.OData.JsonLight
                 resourceValue,
                 this.currentPropertyInfo.MetadataType.TypeReference,
                 isOpenPropertyType,
-                this.CreateDuplicatePropertyNameChecker()).ConfigureAwait(false);
+                this.JsonLightOutputContext.DuplicatePropertyNameChecker).ConfigureAwait(false);
         }
 
         /// <summary>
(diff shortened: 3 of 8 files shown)


habbes:
Seems some tests are failing due to duplicate names? Maybe they are situations where you don't want to share the same property checker? My assumption is that this may due to nested properties (navigation or complex): maybe the same property name exists on the parent entity as well as in a nested object, so a different property checker should be used.

Maybe we can make the property checker hiearchical, instead of creating a new checker every time we enter a scope, or having a single checker for the entire response, we can create a new checker per scope but cache it so that the next time we enter the scope for the same nested property, we re-use the same checker?

@joaocpaiva had also given an object pool as another suggestion.

joaocpaiva:
> Seems some tests are failing due to duplicate names? Maybe they are situations where you don't want to share the same property checker? My assumption is that this may due to nested properties (navigation or complex): maybe the same property name exists on the parent entity as well as in a nested object, so a different property checker should be used.
> 
> Maybe we can make the property checker hiearchical, instead of creating a new checker every time we enter a scope, or having a single checker for the entire response, we can create a new checker per scope but cache it so that the next time we enter the scope for the same nested property, we re-use the same checker?
> 
> @joaocpaiva had also given an object pool as another suggestion.

Seems like Kennedy was just missing a Reset(). Obviously, to reuse a collection it is important to call Clear() every time we enter a new scope that needs it with an empty state. Reusing single collection per operation, should go a long way in reducing allocations for this stack.

KenitoInc:
> > Seems some tests are failing due to duplicate names? Maybe they are situations where you don't want to share the same property checker? My assumption is that this may due to nested properties (navigation or complex): maybe the same property name exists on the parent entity as well as in a nested object, so a different property checker should be used.
> > Maybe we can make the property checker hiearchical, instead of creating a new checker every time we enter a scope, or having a single checker for the entire response, we can create a new checker per scope but cache it so that the next time we enter the scope for the same nested property, we re-use the same checker?
> > @joaocpaiva had also given an object pool as another suggestion.
> 
> Seems like Kennedy was just missing a Reset(). Obviously, to reuse a collection it is important to call Clear() every time we enter a new scope that needs it with an empty state. Reusing single collection per operation, should go a long way in reducing allocations for this stack.

I updated the code and called `Reset()` to fix the issue

src/Microsoft.OData.Core/Json/JsonLightInstanceAnnotationWriter.cs
```diff
@@ -394,7 +398,7 @@ await this.WriteInstanceAnnotationNameAsync(propertyName, annotationName)
                 await this.valueSerializer.WriteResourceValueAsync(resourceValue,
                     expectedType,
                     treatLikeOpenProperty,
-                    this.valueSerializer.CreateDuplicatePropertyNameChecker()).ConfigureAwait(false);
+                    this.valueSerializer.JsonLightOutputContext.DuplicatePropertyNameChecker).ConfigureAwait(false);
```
xuzhg, line 401: At a different level, it could be a "concurrency" problem?

I mean when we write a payload with multiple "levels", for example, Top Resource,  Child resource, GrandChild resource...

The PropertyNameChecker is reused in each "level"?  If yes, it's a big problem if Top Resource has the same property name with GrandChild resource?

habbes:
@joaocpaiva @KenitoInc I still think using a shared `DuplicateNameChecker` when writing a resource with nested resources might be problematic even if `Reset()` is used. I think it would only be safe to share one instance per nesting level.

Consider this example:
```json
{
   "Id": 1,
   "Foo": "Bar",
   "Nested": {
       "Fizz": "Buzz",
       "Foo": "Bar"
    },
   "Id": 2
}
```

If we use the same duplicate checker without resetting, then it will incorrectly flag the top-level `Foo` property and the `Nested.Foo` property as duplicates, even though they're not duplicates.

However, if we call `Reset()` on the instance before entering the `Nested` property and/or after leaving the `Nested` property, then the dictionary will be empty and it will not detect that the second `Id` field as being a duplicate and potentially serializing invalid json.

Since the tests are all passing, maybe we need more tests to cover these scenarios, or maybe my assumptions about how the duplicate checker is used is false.

joaocpaiva:
> @joaocpaiva @KenitoInc I still think using a shared `DuplicateNameChecker` when writing a resource with nested resources might be problematic even if `Reset()` is used. I think it would only be safe to share one instance per nesting level.
> 
> Consider this example:
> 
> ```json
> {
>    "Id": 1,
>    "Foo": "Bar",
>    "Nested": {
>        "Fizz": "Buzz",
>        "Foo": "Bar"
>     },
>    "Id": 2
> }
> ```
> 
> If we use the same duplicate checker without resetting, then it will incorrectly flag the top-level `Foo` property and the `Nested.Foo` property as duplicates, even though they're not duplicates.
> 
> However, if we call `Reset()` on the instance before entering the `Nested` property and/or after leaving the `Nested` property, then the dictionary will be empty and it will not detect that the second `Id` field as being a duplicate and potentially serializing invalid json.
> 
> Since the tests are all passing, maybe we need more tests to cover these scenarios, or maybe my assumptions about how the duplicate checker is used is false.

Makes sense @habbes @KenitoInc. We should make sure there is a test for that use case. Yet another alternative would be to check all top level properties, before processing the nested properties, so we could reset is safely at the end of every level?

mikepizzo:
Checking top level properties before writing nested properties isn't really an option, as the writer supports streaming the values so the service doesn't have to keep the entire response object in memory.

---
In reply to: [978052709](https://github.com/OData/odata.net/pull/2253#issuecomment-978052709) [](http://example.com/codeflow?ancestors=978052709,978052709)

habbes:
@joaocpaiva @mikepizzo if checking all top-level properties before nested properties is not an option, maybe we could still make significant gains from creating only one duplicate checker per nesting level if we have a response with a lot of entities with nested properties.

For example, assuming we're writing a response with 10 entities like:
```js
[
{
    "Prop1": "value",
    "Nested1": {
         "N1Prop": "value"
     },
     "Prop2": "value",
     "Nested2": {
         "N2Prop": "value",
         "N2Nested": { ... }
      }
}
...
]

```

In the current implementation, there'll be 1 checker for the collection + (1 for `Nested1` property + 1 for `Nested2` + 1 for `N2Nested`) * 10 = 31 instances.

But we can safely created once instance of `Nested1` and reuse it for `Nested2` (after resetting) since they're at the same level and therefore do not overlap. And we can create another instance for `N2Nested`. We can cache these 3 instances and reuse them for all entities in the response. So in total we'll have 4 instances (including the one for the top-level collection) instead of 31, scaling only based on the depth of the response, rather than the length * number of nested properties in the response.

Also after processing an entity in the response, each dictionary would have grown to fit the largest number of properties at its nesting level, which means (I think) the dictionaries will probably not be resized after the first entity is written.

I assume this is the same improvement we'd get if we used an object pool?

mikepizzo (changes requested):
Current code seems to change semantics of duplicate name checking.  Perhaps there's a better way to improve perf without caching/resetting.

src/Microsoft.OData.Core/DuplicatePropertyNameChecker.cs
```diff
@@ -120,7 +120,10 @@ public void ValidatePropertyOpenForAssociationLink(string propertyName)
         /// </summary>
         public void Reset()
         {
-            propertyState.Clear();
+            if (propertyState.Count > 0)
```
corranrogue9, line 123: I'm not sure if there was another reason for adding this `if` statement, but `Dictionary<TKey, TValue>.Clear` already does an identical count check as an optimization

corranrogue9:
Is there any way to write a test to ensure that the caching is working correctly?

KenitoInc:
Created a new PR #2328 with a different implementation

PR #2254, opened 2021-11-18
Fix issues with performance tests

habbes:
### Issues

### Description

Some performance tests were failing or not returning results. In most cases the issue was related to incorrect usage of `[IterationSetup]`. This attribute is meant to be used on a method that should execute before each invocation of a benchmark method (it's actually not recommended to use unless for benchmarks that run for 100ms+ per invocation because it can spoil the results). However the issue was because I had multiple methods with `[IterationSetup]` and `[IterationSetup(Target=...)]` expecting each to be executed for the same benchmark method. But only one gets executed, and so some benchmarks were executing with some values not being properly initialized.

In some of these cases, I realized that the setup code didn't need to run for each invocation of a benchmark method, it would be sufficient for it run once per method (e.g. the same model is reused in all invocations of a test). In such cases I moved the setup code in the `[GlobalSetup]` method which is meant for that.

For more info about setup and clean up code in BenchmarkDotNet see: https://benchmarkdotnet.org/articles/features/setup-and-cleanup.html

Some tests still make use of `IterationSetup`, especially when writing to a stream or file and I want to reset the stream or file each time the method is invoked.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/PerformanceTests/SerializationBaselineTests/SerializationBenchmarks.cs
+++ b/test/PerformanceTests/SerializationBaselineTests/SerializationBenchmarks.cs
@@ -18,13 +18,10 @@ namespace SerializationBaselineTests
         IExperimentWriter odataSyncCharPoolWriter;
         IEnumerable<Customer> data;
 
-        [Params(1000, 5000, 10000)]
-        public int dataSize;
+        public int dataSize = 5000;
 
         [Params(true, false)]
         public bool isModelImmutable;
-
-        IEdmModel model;
 
         string tempFile;
         Stream fileStream;
@@ -33,9 +30,9 @@ namespace SerializationBaselineTests
         public void PrepareDataset()
         {
             data = DataSet.GetCustomers(dataSize);
+            InitWriters();
         }
 
-        [IterationSetup]
         public void InitWriters()
         {
             var model = DataModel.GetEdmModel();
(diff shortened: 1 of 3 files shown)


test/PerformanceTests/SerializationBaselineTests/SerializationBenchmarks.cs
```diff
@@ -18,24 +18,21 @@ public class SerializationBenchmarks
         IExperimentWriter odataSyncCharPoolWriter;
         IEnumerable<Customer> data;
 
-        [Params(1000, 5000, 10000)]
-        public int dataSize;
+        public int dataSize = 5000;
```
habbes, line 21: I set this to a fixed size instead of having different data sizes because the tests are linear. Increasing the data size increases the run time linearly which did not seem to provide any additional insight to justify the additional time spent running the benchmarks.

marabooy: Could this be reduced to 2500 and still give the same depth?

marabooy (approved):
Looks good but do add method and parameter documentation to improve the code.

odero (approved):

gathogojr (approved):

KenitoInc (approved):

PR #2255, opened 2021-11-18
Implement asynchronous support in ODataJsonReaderCoreUtils

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataJsonReaderCoreUtils`**

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/ODataJsonReaderCoreUtils.cs
+++ b/src/Microsoft.OData.Core/Json/ODataJsonReaderCoreUtils.cs
@@ -15,6 +15,7 @@ namespace Microsoft.OData.Json
     using Microsoft.OData.Edm;
     using Microsoft.OData.Metadata;
     using ODataErrorStrings = Microsoft.OData.Strings;
+    using System.Threading.Tasks;
 
     #endregion Namespaces
 
@@ -56,19 +57,19 @@ namespace Microsoft.OData.Json
                 return null;
             }
 
-            Microsoft.Spatial.ISpatial spatialValue = null;
+            ISpatial spatialValue = null;
             if (insideJsonObjectValue || jsonReader.NodeType == JsonNodeType.StartObject)
             {
                 IDictionary<string, object> jsonObject = ReadObjectValue(jsonReader, insideJsonObjectValue, inputContext, recursionDepth);
-                Microsoft.Spatial.GeoJsonObjectFormatter jsonObjectFormatter =
-                    Microsoft.Spatial.SpatialImplementation.CurrentImplementation.CreateGeoJsonObjectFormatter();
+                GeoJsonObjectFormatter jsonObjectFormatter =
+                    SpatialImplementation.CurrentImplementation.CreateGeoJsonObjectFormatter();
                 if (expectedValueTypeReference.IsGeography())
                 {
-                    spatialValue = jsonObjectFormatter.Read<Microsoft.Spatial.Geography>(jsonObject);
+                    spatialValue = jsonObjectFormatter.Read<Geography>(jsonObject);
                 }
                 else
                 {
-                    spatialValue = jsonObjectFormatter.Read<Microsoft.Spatial.Geometry>(jsonObject);
+                    spatialValue = jsonObjectFormatter.Read<Geometry>(jsonObject);
                 }
             }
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
```diff
@@ -391,8 +392,453 @@ internal static JsonNodeType ReadNext(this IJsonReader jsonReader)
...
+                    .ConfigureAwait(false);
+
+                return resourceValue;
+            }
+            else if (jsonReader.NodeType == JsonNodeType.StartArray)
+            {
+                await jsonReader.ReadStartArrayAsync()
+                    .ConfigureAwait(false);
+                ODataCollectionValue collectionValue = new ODataCollectionValue();
+                IList<object> properties = new List<object>();
```
odero, line 792: Using IList seems unnecessary
```suggestion
                List<object> properties = new List<object>();

```

src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
```diff
@@ -459,5 +905,45 @@ private static void ValidateNodeType(this IJsonReader jsonReader, JsonNodeType e
...
+        /// </summary>
+        /// <param name="jsonReader">The <see cref="JsonReader"/> to read from.</param>
+        /// <param name="expectedNodeType">The expected <see cref="JsonNodeType"/> of the read node.</param>
+        private static async Task ReadNextAsync(this IJsonReaderAsync jsonReader, JsonNodeType expectedNodeType)
+        {
+            Debug.Assert(jsonReader != null, "jsonReader != null");
+            Debug.Assert(expectedNodeType != JsonNodeType.None, "expectedNodeType != JsonNodeType.None");
+
+            jsonReader.ValidateNodeType(expectedNodeType);
+            await jsonReader.ReadAsync().ConfigureAwait(false);
```
odero, line 920:
```suggestion
        private static Task ReadNextAsync(this IJsonReaderAsync jsonReader, JsonNodeType expectedNodeType)
        {
            Debug.Assert(jsonReader != null, "jsonReader != null");
            Debug.Assert(expectedNodeType != JsonNodeType.None, "expectedNodeType != JsonNodeType.None");

            jsonReader.ValidateNodeType(expectedNodeType);
            return jsonReader.ReadAsync().ConfigureAwait(false);

```

src/Microsoft.OData.Core/Json/ODataJsonReaderCoreUtils.cs
```diff
@@ -220,5 +324,125 @@ private static IEnumerable<object> ReadArrayValue(IJsonReader jsonReader, ODataI
...
+        /// <returns>a list of json objects.</returns>
+        private static async Task<IEnumerable<object>> ReadArrayValueAsync(IJsonReaderAsync jsonReader, ODataInputContext inputContext, int recursionDepth)
+        {
+            Debug.Assert(jsonReader != null, "jsonReader != null");
+            Debug.Assert(jsonReader.NodeType == JsonNodeType.StartArray, "jsonReader.NodeType == JsonNodeType.StartArray");
+            Debug.Assert(inputContext != null, "inputContext != null");
+
+            ValidationUtils.IncreaseAndValidateRecursionDepth(ref recursionDepth, inputContext.MessageReaderSettings.MessageQuotas.MaxNestingDepth);
+
+            List<object> array = new List<object>();
```
odero, line 416: nit: calling a non-array item an array

```suggestion
            List<object> items= new List<object>();

```

src/Microsoft.OData.Core/Json/ODataJsonReaderCoreUtils.cs
```diff
@@ -121,6 +122,109 @@ internal static class ODataJsonReaderCoreUtils
...
+                && await TryReadNullValueAsync(jsonReader, inputContext, expectedValueTypeReference, validateNullValue, propertyName)
+                .ConfigureAwait(false))
+            {
+                return null;
+            }
+
+            ISpatial spatialValue = null;
+            if (insideJsonObjectValue || jsonReader.NodeType == JsonNodeType.StartObject)
+            {
+                IDictionary<string, object> jsonObject = await ReadObjectValueAsync(jsonReader, insideJsonObjectValue, inputContext, recursionDepth)
```
odero, line 163:
```suggestion
                Dictionary<string, object> jsonObject = await ReadObjectValueAsync(jsonReader, insideJsonObjectValue, inputContext, recursionDepth)

```

src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
gathogojr: @odero Eliding may not be a good idea in this method since `ValidateNodeType` does throw an exception when the node type is not the expected one. The state machine will allow us to stash the exception in the returned task

gathogojr:
@odero For the comments relevant to `JsonReaderExtensions` class, I'll consider them in the dependent PR #2248.

src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
```diff
@@ -391,8 +392,453 @@ internal static JsonNodeType ReadNext(this IJsonReader jsonReader)
...
+
+        /// <summary>
+        /// Asynchronously reads the next node from the <paramref name="jsonReader"/> and verifies that it is a PrimitiveValue node of type string.
+        /// </summary>
+        /// <param name="jsonReader">The <see cref="JsonReader"/> to read from.</param>
+        /// <param name="propertyName">The name of the property for which to read the string; used in error messages only.</param>
+        /// <returns>A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains the string value read from the reader;
+        /// throws an exception if no string value could be read.</returns>
+        internal static async Task<string> ReadStringValueAsync(this IJsonReaderAsync jsonReader, string propertyName)
```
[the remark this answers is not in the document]
gathogojr: @odero I considered it but there are also people who frown on this:
```csharp
if (!string.IsNullOrEmpty(propertyName))
{
    throw CreateException(Strings.JsonReaderExtensions_CannotReadPropertyValueAsString(value, propertyName));
}
else
{
    throw CreateException(Strings.JsonReaderExtensions_CannotReadValueAsString(value));
}
```
In your view, this is a lesser evil, no?

src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
```diff
@@ -391,8 +392,453 @@ internal static JsonNodeType ReadNext(this IJsonReader jsonReader)
...
+        /// throws an exception if no string value could be read.</returns>
+        internal static async Task<string> ReadStringValueAsync(this IJsonReaderAsync jsonReader)
+        {
+            Debug.Assert(jsonReader != null, "jsonReader != null");
+
+            object value = await jsonReader.ReadPrimitiveValueAsync()
+                .ConfigureAwait(false);
+
+            string stringValue = value as string;
+            if (value == null || stringValue != null)
```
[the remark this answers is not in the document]
gathogojr: @odero There can be. The scenario below
```json
{"Prop":null}
```

src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
```diff
@@ -391,8 +392,453 @@ internal static JsonNodeType ReadNext(this IJsonReader jsonReader)
...
+            if (value == null || doubleValue != null)
+            {
+                return doubleValue;
+            }
+
+            int? intValue = value as int?;
+            if (intValue != null)
+            {
+                return (double)intValue;
+            }
```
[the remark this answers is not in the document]
gathogojr: @odero Apparently not as this code snippet would demonstrate:
```csharp
int fourint = 4;
object fourobj = fourint;
double? fourdbl = fourobj as double?; // fourdbl = null
```

src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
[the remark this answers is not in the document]
odero: Hmm interesting. So in this case what would `stringValue` be?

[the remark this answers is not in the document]
gathogojr: `stringValue` would be `null`. When reading, if what follows a property and the separator starts with `n`, then we try to read it as `null`; otherwise we throw an exception. For example we'd through an exception if we came across this: `{"Prop":nil}`. A non-null string however is expected to start with quotes, e.g., `{"Prop":"foo"}`. If we removed the `value == null` condition, then for `{"Prop":null}`, an exception would be thrown 4 lines below

src/Microsoft.OData.Core/Json/JsonReaderExtensions.cs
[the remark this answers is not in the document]
odero: Looks like lesser evil to me :)

odero (approved):

src/Microsoft.OData.Core/Json/ODataJsonReaderCoreUtils.cs
gathogojr: @odero This won't work because the `ReadObjectValueAsync` method returns `IDictionary<string, object>` and that can't be implicitly converted to `Dictionary<string, object>`.

marabooy: Since the method is private can't we make the change without any issues? IDictionary may lead to a bit of boxing IIRC.

src/Microsoft.OData.Core/Json/ODataJsonReaderCoreUtils.cs
```diff
@@ -220,5 +324,125 @@ private static IEnumerable<object> ReadArrayValue(IJsonReader jsonReader, ODataI
...
+        /// <summary>
+        /// Asynchronously reads the json array from the reader.
+        /// </summary>
+        /// <param name="jsonReader">JsonReader instance.</param>
+        /// <param name="inputContext">The input context with all the settings.</param>
+        /// <param name="recursionDepth">The recursion depth to start with.</param>
+        /// <returns>A task that represents the asynchronous read operation.
+		/// The value of the TResult parameter contains a lit of JSON objects.</returns>
+        /// <returns>a list of json objects.</returns>
+        private static async Task<IEnumerable<object>> ReadArrayValueAsync(IJsonReaderAsync jsonReader, ODataInputContext inputContext, int recursionDepth)
```
marabooy, line 408: Maybe consider using concrete types for the private method? As I have noticed some of the perf changes are about removing IEnumerable interface from the hot path which in most cases may result in some speed and less allocations.

marabooy (approved):

src/Microsoft.OData.Core/Json/ODataJsonReaderCoreUtils.cs
gathogojr: @marabooy Point

PR #2256, opened 2021-11-22
Refactor alternate key resolver to avoid relying on exceptions

### LINKED ISSUE #2221: First chance odata exception resolving alternate keys
When a model contains alternate keys, you have to deal with a first chance odata exception for resolving each. Consider a model with many hundreds of entities each having at least one alternate key. Then it is annoying to debug with catching first chance exceptions on.

### Assemblies affected

OData .Net lib 8.x* (Microsoft.OData.Core)

### Reproduce steps

Debug the test `AlternateKeyShouldWork` with 'break when \<All Common Language Exceptions not in this list\> are thrown' exception setting is activated in Visual Studio. https://github.com/OData/odata.net/blob/64bed784177b9202b73c4510d2461bebf5ad03b7/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs#L383

### Expected result

No first chance exception by checking chance of success upfront.

### Actual result

A first chance `ODataException` is caught for each alternate key in the model.

### Additional detail

It is inappropriate that `UriParser/Resolver/AlternateKeysODataUriResolver.ResolveKeys` relies so heavily on trying to resolve the primary key first and then trying to resolve the alternate key by catching prior exception. It would be more smooth to eliminate certain failures by checking the number and name of keys first.

```diff
public override IEnumerable<KeyValuePair<string, object>> ResolveKeys(IEdmEntityType type, IDictionary<string, string> namedValues, Func<IEdmTypeReference, string, object> convertFunc)
{
    IEnumerable<KeyValuePair<string, object>> convertedPairs;
-   try
+   string[] keyNames = type.DeclaredKey.Select(keyProperty => keyProperty.Name).ToArray();
+   if (namedValues.Keys.Count == keyNames.Length)
    {
+       if (namedValues.Keys.All(keyProperty => keyNames.Contains(keyProperty, StringComparer.OrdinalIgnoreCase)))
+       {
            convertedPairs = base.ResolveKeys(type, namedValues, convertFunc);
+       }
    }
-   catch (ODataException)
+   else
    {
        if (!TryResolveAlternateKeys(type, namedValues, convertFunc, out convertedPairs))
        {
-           throw;
+           throw new ODataException(...);
        }
 -  }

    return convertedPairs;
}

```

habbes:
### Issues

*This pull request fixes #2221 .*

### Description

The `AlternateKeyODataUriResolver.ResolveKeys` first calls the `base.ResolveKeys()` to attempt to resolve the keys, and if it fails (exception thrown), it tries to resolve alternate keys. Relying on the exception being thrown adds "noise" when debugging due to first chance exceptions causing the debugger to break (as mentioned in the issue). We are also paying the price of try-catch when we don't need to.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Parameterized.Microsoft.OData.Core.cs
+++ b/src/Microsoft.OData.Core/Parameterized.Microsoft.OData.Core.cs
@@ -5215,9 +5215,12 @@ namespace Microsoft.OData {
         /// <summary>
         /// A string like "Only $filter and $search query options are allowed within $count."
         /// </summary>
-        internal static string UriQueryExpressionParser_IllegalQueryOptioninDollarCount()
+        internal static string UriQueryExpressionParser_IllegalQueryOptioninDollarCount
         {
-            return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.UriQueryExpressionParser_IllegalQueryOptioninDollarCount);
+            get
+            {
+                return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.UriQueryExpressionParser_IllegalQueryOptioninDollarCount);
+            }
         }
 
         /// <summary>
@@ -5506,9 +5509,12 @@ namespace Microsoft.OData {
         /// <summary>
         /// A string like "The next token in a CountSegmentNode must be a collection."
         /// </summary>
-        internal static string MetadataBinder_CountSegmentNextTokenNotCollectionValue()
+        internal static string MetadataBinder_CountSegmentNextTokenNotCollectionValue
         {
-            return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.MetadataBinder_CountSegmentNextTokenNotCollectionValue);
+            get
+            {
+                return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.MetadataBinder_CountSegmentNextTokenNotCollectionValue);
+            }
         }
 
         /// <summary>
@@ -6215,9 +6221,12 @@ namespace Microsoft.OData {
         /// <summary>
         /// A string like "Empty parenthesis not allowed."
         /// </summary>
-        internal static string UriParser_EmptyParenthesis()
+        internal static string UriParser_EmptyParenthesis
         {
-            return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.UriParser_EmptyParenthesis);
+            get
+            {
+                return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.UriParser_EmptyParenthesis);
+            }
         }
 
         /// <summary>
@@ -6823,6 +6832,14 @@ namespace Microsoft.OData {
         }
 
         /// <summary>
+        /// A string like "The number or names of keys specified in the URI do not match the declared or alternate key properties for the resource '{0}'."
+        /// </summary>
+        internal static string BadRequest_KeyOrAlternateKeyMismatch(object p0)
+        {
+            return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.BadRequest_KeyOrAlternateKeyMismatch, p0);
+        }
+
+        /// <summary>
         /// A string like "Segments with multiple key values must specify them in 'name=value' form."
         /// </summary>
         internal static string RequestUriProcessor_KeysMustBeNamed
@@ -6953,13 +6970,13 @@ namespace Microsoft.OData {
         }
 
         /// <summary>
-        /// A string like "No property is allowed after $ref segment."
+        /// A string like "$count is not allowed in $select option."
         /// </summary>
-        internal static string ExpressionToken_NoPropAllowedAfterRef
+        internal static string ExpressionToken_DollarCountNotAllowedInSelect
         {
             get
             {
-                return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.ExpressionToken_NoPropAllowedAfterRef);
+                return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.ExpressionToken_DollarCountNotAllowedInSelect);
             }
         }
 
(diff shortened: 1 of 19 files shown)


src/Microsoft.OData.Core/UriParser/Resolver/AlternateKeysODataUriResolver.cs
```diff
@@ -39,16 +39,21 @@ public AlternateKeysODataUriResolver(IEdmModel model)
         /// <returns>The resolved key list.</returns>
         public override IEnumerable<KeyValuePair<string, object>> ResolveKeys(IEdmEntityType type, IDictionary<string, string> namedValues, Func<IEdmTypeReference, string, object> convertFunc)
         {
-            IEnumerable<KeyValuePair<string, object>> convertedPairs;
-            try
+            IEnumerable<KeyValuePair<string, object>> convertedPairs = null;
+            string[] keyNames = type.DeclaredKey.Select(keyProperty => keyProperty.Name).ToArray();
```
xuzhg, line 43: Only the declared keys, or keys from base type also?

habbes: I've replaced with `type.Key()`

habbes:
In retrospect, I am not sure if this implementation is better than the precious one. With this change, if the `base.ResolveKeys()` method throws an exception, it will not be caught by the alternate key resolver even if it should be able to successfully handle that scenario, which might cause the method to fail when it should succeed. I don't have an actual scenario in mind where this would happen, the only one I can think of is if we have two properties which only differ by their case, one used in the declared key and one used as an alternate key.

I think the safer thing to do would be to have an implementation of the base `ResolveKeys()` that instead of throwing an exception on error, it returns false. It seems that I'm already doing a bit of that partially by checking the names and count of the  keys in the alternate key resolver before calling the base one. I'm not sure whether this duplication of effort is worth avoid catching the exception.

habbes:
I have opted to add a `TryResolveKeys` method that returns true if key resolution was successful and false otherwise.
This is now the basis for the `ResolveKeys` in the base `ODataUriResolver` class, which throws an exception when the method returns false. I've created a generic, but somewhat descriptive, error message to compensate for that fact that I'm now throwing the same error message regardless of the cause of failure.

I reused the `TryResolveKeys` in the `AlternateKeyODataUriResolver` to avoid relying on the base exception (which is what the PR is try to fix) but also to avoid enumerating the keys redundantly in the alternate key resolver.

marabooy (approved):

src/Microsoft.OData.Core/Parameterized.Microsoft.OData.Core.cs
```diff
@@ -5215,9 +5215,12 @@ internal static string UriQueryExpressionParser_CloseParenOrOperatorExpected(obj
         /// <summary>
         /// A string like "Only $filter and $search query options are allowed within $count."
         /// </summary>
-        internal static string UriQueryExpressionParser_IllegalQueryOptioninDollarCount()
+        internal static string UriQueryExpressionParser_IllegalQueryOptioninDollarCount
```
habbes, line 5218: These changes were auto-generated by running the `Microsoft.OData.Core.tt` and `Parameterized.Microsot.OData.Core.tt` T4 templates. I think the files might have been previously updated manually, leading to some inconsistencies between the template files and the code files, which explains why they have been modified after re-running the T4 files.

src/Microsoft.OData.Core/UriParser/Resolver/AlternateKeysODataUriResolver.cs
```diff
@@ -39,20 +39,17 @@ public AlternateKeysODataUriResolver(IEdmModel model)
...
             }
-            catch (ODataException)
+
+            if (!TryResolveAlternateKeys(type, namedValues, convertFunc, out IEnumerable<KeyValuePair<string, object>> alternateConvertedPairs))
             {
-                if (!TryResolveAlternateKeys(type, namedValues, convertFunc, out convertedPairs))
-                {
-                    throw;
-                }
+                throw ExceptionUtil.CreateBadRequestError(Strings.BadRequest_KeyOrAlternateKeyMismatch(type.FullName()));
```
gathogojr, line 49: Why is `BadRequest_KeyOrAlternateKeyMismatch` not `BadRequest_AlternateKeyMismatch`?

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -378,19 +378,34 @@ public virtual IEnumerable<IEdmOperationImport> ResolveOperationImports(IEdmMode
...
+            resolvedKeys = null;
             var convertedPairs = new Dictionary<string, object>(StringComparer.Ordinal);
 
-            // Throw an error if key size from url doesn't match that from model.
-            // Other derived ODataUriResolver intended for alternative key resolution, such as the built in AlternateKeysODataUriResolver,
-            // should override this ResolveKeys method.
             IEnumerable<IEdmStructuralProperty> keys = type.Key();
-            if (keys.Count() != namedValues.Count)
-            {
-                throw ExceptionUtil.CreateBadRequestError(Strings.BadRequest_KeyCountMismatch(type.FullName()));
```
gathogojr, line 389: Can we conclude that the exception with the `BadRequest_KeyCountMismatch` message will no longer be raised?

src/Microsoft.OData.Core/UriParser/Resolver/AlternateKeysODataUriResolver.cs
habbes: Because at this point, we were not able to much neither the declared key nor the alternate key.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
habbes: Yeah, you're right. I can remove it.

habbes: Actually it's still used. There's another `ResolveKeys()` overload that deals with positional keys (as opposed to named keys) that still uses this error message.

gathogojr (approved):

src/Microsoft.OData.Core/UriParser/Resolver/AlternateKeysODataUriResolver.cs
corranrogue9: Is this because the previous code was incorrect, or is this because the new code is unable to differentiate? I don't see why, in the new code, the `Try` variants of the methods don't have an additional `out` parameter which is the exception that would have otherwise been throw, that way we get to keep the richness of our current errors.

PR #2264, opened 2021-11-25
Implement asynchronous support in BufferingJsonReader

gathogojr:
### Issues

### Description

Implement asynchronous support in **`BufferingJsonReader`**

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
+++ b/src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
@@ -6,12 +6,13 @@
 
 namespace Microsoft.OData.Json
 {
-    using System;
     #region Namespaces
+
+    using System;
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.IO;
-    using System.Text;
+    using System.Threading.Tasks;
     using Microsoft.OData.JsonLight;
 
     #endregion Namespaces
@@ -19,7 +20,7 @@ namespace Microsoft.OData.Json
     /// <summary>
     /// Reader for the JSON format (http://www.json.org) that supports look-ahead.
     /// </summary>
-    internal class BufferingJsonReader : IJsonStreamReader
+    internal class BufferingJsonReader : IJsonStreamReader, IJsonStreamReaderAsync
     {
         /// <summary>The (possibly empty) list of buffered nodes.</summary>
         /// <remarks>This is a circular linked list where this field points to the first item of the list.</remarks>
@@ -34,6 +35,11 @@ namespace Microsoft.OData.Json
         /// The inner JSON reader.
         /// </summary>
         private readonly IJsonReader innerReader;
+
+        /// <summary>
+        /// The inner asynchronous JSON reader.
+        /// </summary>
+        private readonly IJsonReaderAsync asyncInnerReader;
 
         /// <summary>
         /// The maximum number of recursive internalexception objects to allow when reading in-stream errors.
@@ -74,6 +80,12 @@ namespace Microsoft.OData.Json
             Debug.Assert(innerReader != null, "innerReader != null");
 
             this.innerReader = innerReader;
+            // IJsonReaderAsync inherits from IJsonReader in the current case
+            if (innerReader is IJsonReaderAsync asyncInnerReader)
+            {
+                this.asyncInnerReader = asyncInnerReader;
+            }
+
             this.inStreamErrorPropertyName = inStreamErrorPropertyName;
             this.maxInnerErrorDepth = maxInnerErrorDepth;
             this.bufferedNodesHead = null;
@@ -103,6 +115,7 @@ namespace Microsoft.OData.Json
                     return this.bufferedNodesHead.NodeType;
                 }
 
+                // This should work okay even in async scenarios since IJsonReaderAsync inherits from IJsonReader?
                 return this.innerReader.NodeType;
             }
         }
@@ -141,6 +154,7 @@ namespace Microsoft.OData.Json
         {
             get
             {
+                // This should work okay even in async scenarios since IJsonReaderAsync inherits from IJsonReader?
                 return this.innerReader.IsIeee754Compatible;
             }
         }
@@ -233,6 +247,142 @@ namespace Microsoft.OData.Json
 
             // read the next node and check whether it is an in-stream error
             return this.ReadInternal();
+        }
+
+        /// <summary>
+        /// The value of the last reported node.
+        /// </summary>
+        /// <remarks>
+        /// Depending on whether buffering is on or off this will return the node value of the last
+        /// buffered read or the node value of the last unbuffered read.
+        /// </remarks>
+        /// <returns>A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains the node value of the last buffered read if buffering is on or off;
+        /// otherwise the node value of the last unbuffered read.</returns>
+        public Task<object> GetValueAsync()
+        {
+            this.AssertAsynchronous();
+
+            if (this.bufferedNodesHead != null)
+            {
+                if (this.isBuffering)
+                {
+                    Debug.Assert(this.currentBufferedNode != null, "this.currentBufferedNode != null");
+                    return Task.FromResult(this.currentBufferedNode.Value);
+                }
+
+                // In non-buffering mode if we have buffered nodes satisfy the request from the first node there
+                return Task.FromResult(this.bufferedNodesHead.Value);
+            }
+
+            return this.asyncInnerReader.GetValueAsync();
+        }
+
+        /// <summary>
+        /// Asynchronously reads the next node from the input.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains true if a new node was found, or false if end of input was reached.</returns>
+        public Task<bool> ReadAsync()
+        {
+            Debug.Assert(!this.parsingInStreamError, "!this.parsingInStreamError");
+            this.AssertAsynchronous();
+
+            // Read the next node and check whether it is an in-stream error
+            return this.ReadInternalAsync();
+        }
+
+        /// <summary>
+        /// Asynchronously checks whether or not the current value can be streamed.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous operation.
+        /// true if the current value can be streamed; otherwise false.</returns>
+        public virtual async Task<bool> CanStreamAsync()
+        {
+            this.AssertAsynchronous();
+
+            if (!this.isBuffering && this.asyncInnerReader is IJsonStreamReaderAsync asyncStreamReader)
+            {
+                return await asyncStreamReader.CanStreamAsync()
+                    .ConfigureAwait(false);
+            }
+
+            object value = await this.GetValueAsync()
+                .ConfigureAwait(false);
+            return value is string || value == null || this.NodeType == JsonNodeType.StartArray || this.NodeType == JsonNodeType.StartObject;
+        }
+
+        /// <summary>
+        /// Asynchronously creates a stream for reading a stream value.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains a <see cref="Stream"/> used to read a stream value.</returns>
+        public virtual async Task<Stream> CreateReadStreamAsync()
+        {
+            this.AssertAsynchronous();
+
+            IJsonStreamReaderAsync asyncStreamReader = this.asyncInnerReader as IJsonStreamReaderAsync;
+            if (!this.isBuffering && asyncStreamReader != null)
+            {
+                return await asyncStreamReader.CreateReadStreamAsync()
+                    .ConfigureAwait(false);
+            }
+
+            Stream result;
+
+            object value = await this.GetValueAsync()
+                .ConfigureAwait(false);
+
+            if (value == null)
+            {
+                result = new MemoryStream(new byte[0]);
+            }
+            else
+            {
+                result = new MemoryStream(Convert.FromBase64String((string)value));
+            }
+
+            await this.asyncInnerReader.ReadAsync()
+                .ConfigureAwait(false);
+
+            return result;
+        }
+
+        /// <summary>
+        /// Asynchronously creates a TextReader for reading a text value.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains a <see cref="TextReader"/> for reading the text value.</returns>
+        public virtual async Task<TextReader> CreateTextReaderAsync()
+        {
+            this.AssertAsynchronous();
+
+            IJsonStreamReaderAsync asyncStreamReader = this.asyncInnerReader as IJsonStreamReaderAsync;
+            if (!this.isBuffering && asyncStreamReader != null)
+            {
+                return await asyncStreamReader.CreateTextReaderAsync()
+                    .ConfigureAwait(false);
+            }
+
+            TextReader result;
+
+            object value = await this.GetValueAsync()
+                .ConfigureAwait(false);
+
+            if (value == null)
+            {
+                result = new StringReader("");
+            }
+            else
+            {
+                result = new StringReader((string)value);
+            }
+
+
+            await this.asyncInnerReader.ReadAsync()
+                .ConfigureAwait(false);
+
+            return result;
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -235,6 +249,111 @@ public bool Read()
...
         }
 
+        /// <summary>
+        /// The value of the last reported node.
+        /// </summary>
+        /// <remarks>
+        /// Depending on whether buffering is on or off this will return the node type of the last
+        /// buffered read or the node type of the last unbuffered read.
+        /// </remarks>
+        // TODO: Document <returns/>
```
habbes, line 259: Just a reminder not to forget the TODO, just in case.

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -235,6 +249,111 @@ public bool Read()
...
+        public Task<object> GetValueAsync()
+        {
+            if (this.bufferedNodesHead != null)
+            {
+                if (this.isBuffering)
+                {
+                    Debug.Assert(this.currentBufferedNode != null, "this.currentBufferedNode != null");
+                    return Task.FromResult(this.currentBufferedNode.Value);
```
habbes, line 267: Out of curiosity, you opted not to use ValueTask in this case because this is a public method or because the synchronous code-path is less frequent than the async one?

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -341,13 +489,77 @@ internal bool StartBufferingAndTryToReadInStreamErrorPropertyValue(out ODataErro
...
+        {
+            Debug.Assert(!this.isBuffering, "Buffering is already turned on. Must not call StartBuffering again.");
+            this.AssertAsynchronous();
+
+            if (this.bufferedNodesHead == null)
+            {
+                // capture the current state of the reader as the first item in the buffer (if there are none)
+                object value = await this.asyncInnerReader.GetValueAsync()
+                    .ConfigureAwait(false);
+                this.bufferedNodesHead = new BufferedNode(this.asyncInnerReader.NodeType, value);
```
habbes, line 506: I was curious about the relationship between `IJsonReader.NodeType` and `IJsonReader.Value` and whether accessing the value (or `GetValueAsync` could modify the node type). Based on my perusal of the code, it seems `NodeType` is only updated by `IJsonReader.Read()` (and ReadAsync). And `GetValueAsync()` does not actually end up calling `ReadAsync` (though it calls the inner text writer's `ReadAsync`). In short I don't think this is a problem, but I thought maybe you could confirm since you're more familiar with the code.

If it happens that accessing the value has a side effect on `NodeType`, then the order in which you access them matters (otherwise it doesn't). Because in this async method you're accessing the value by calling `GetValueAsync()` first, then the node type. But in the sync version, you're accessing `NodeType` first, then `Value`.

src/Microsoft.OData.Core/TaskUtils.cs
```diff
@@ -55,6 +65,38 @@ internal static Task CompletedTask
             }
         }
 
+        /// <summary>
+        /// Returns a completed task with the result set to true.
+        /// </summary>
+        internal static Task<bool> TrueResultTask
```
habbes, line 71: Seems like the built-in `Task.FromResult` already caches completed tasks for boolean values (see https://source.dot.net/#System.Private.CoreLib/Task.cs,5198). Maybe this might be redundant?

src/Microsoft.OData.Core/Json/JsonReader.cs
```diff
@@ -553,6 +553,23 @@ public virtual async Task<object> GetValueAsync()
             return this.nodeValue;
         }
 
+        /// <summary>
+        /// Asynchronously checks whether or not the current value can be streamed.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous operation.
+        /// true if the current value can be streamed; otherwise false.</returns>
+        public Task<bool> CanStreamAsync()
```
habbes, line 561: Why does this method need to be async if doesn't do any async work?

src/Microsoft.OData.Core/Json/IJsonStreamReaderAsync.cs
```diff
@@ -28,5 +28,12 @@ public interface IJsonStreamReaderAsync : IJsonReaderAsync
         /// <returns>A task that represents the asynchronous operation.
         /// The value of the TResult parameter contains a <see cref="TextReader"/> for reading the string value.</returns>
         Task<TextReader> CreateTextReaderAsync();
+
+        /// <summary>
+        /// Asynchronously checks whether or not the current value can be streamed.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous operation.
+        /// true if the current value can be streamed; otherwise false.</returns>
+        Task<bool> CanStreamAsync();
```
habbes, line 37: Is this necessary?

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -341,13 +489,77 @@ internal bool StartBufferingAndTryToReadInStreamErrorPropertyValue(out ODataErro
...
+        /// A method to asynchronously detect whether the current property value represents an in-stream error.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains a tuple comprising of:
+        /// 1). A value of true if the current value is an in-stream error value; otherwise false.
+        /// 2). An <see cref="ODataError"/> instance that was read from the payload.</returns>
+        internal async Task<Tuple<bool, ODataError>> StartBufferingAndTryToReadInStreamErrorPropertyValueAsync()
+        {
+            this.AssertNotBuffering();
+            this.AssertAsynchronous();
```
habbes, line 537: Why is this assertion here and not in other async methods? Is it necessary?

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -474,6 +681,150 @@ protected virtual void ProcessObjectValue()
...
+                Debug.Assert(this.currentBufferedNode != null, "this.currentBufferedNode != null");
+
+                if (this.currentBufferedNode.Next != this.bufferedNodesHead)
+                {
+                    this.currentBufferedNode = this.currentBufferedNode.Next;
+                    result = true;
+                }
+                else
+                {
+                    if (this.parsingInStreamError)
```
habbes, line 720:
```suggestion
                else if (this.parsingInStreamError)
                {

```

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -474,6 +681,150 @@ protected virtual void ProcessObjectValue()
...
+                        // read the next node from the input stream and check
+                        // whether it is an in-stream error
+                        result = await this.ReadNextAndCheckForInStreamErrorAsync()
+                            .ConfigureAwait(false);
+                    }
+                }
+            }
+            else
+            {
+                if (this.bufferedNodesHead == null)
```
habbes, line 742:
```suggestion
            else if (this.bufferedNodesHead == null)
            {

```

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -474,6 +681,150 @@ protected virtual void ProcessObjectValue()
...
+            // Only check for in-stream errors if the buffering reader is told to do so
+            if (!this.DisableInStreamErrorDetection)
+            {
+                // Move to the first property of the potential error object (or the end-object node if no properties exist)
+                await this.ReadInternalAsync()
+                    .ConfigureAwait(false);
+
+                // We only consider this to be an in-stream error if the object has a single 'error' property
+                bool errorPropertyFound = false;
+                Tuple<bool, ODataError> readInStreamErrorPropertyResult = null;
```
habbes, line 790:
```suggestion
               (bool hasError, ODataError error) readInStreamErrorPropertyResult = null;
```
Naming the fields of the tuple makes it easier to refer to them, and avoids using `Item1`, `Item2`, et

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -474,6 +681,150 @@ protected virtual void ProcessObjectValue()
...
+                    errorPropertyFound = true;
+
+                    // Position the reader over the property value
+                    await this.ReadInternalAsync()
+                        .ConfigureAwait(false);
+
+                    readInStreamErrorPropertyResult = await this.TryReadInStreamErrorPropertyValueAsync()
+                        .ConfigureAwait(false);
+
+                    if (!readInStreamErrorPropertyResult.Item1)
```
habbes, line 812:
```suggestion
                    if (!readInStreamErrorPropertyResult.hasError)
```

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -474,6 +681,150 @@ protected virtual void ProcessObjectValue()
...
+                        // This means we thought we saw an in-stream error, but then
+                        // we didn't see an intelligible error object, so we give up an reading the in-stream error
+                        // and return. We will fail later in some other way. This payload is totally messed up.
+                        return;
+                    }
+                }
+
+                if (errorPropertyFound)
+                {
+                    throw new ODataErrorException(readInStreamErrorPropertyResult?.Item2);
```
habbes, line 823:
```suggestion
                    throw new ODataErrorException(readInStreamErrorPropertyResult?.error);

```

habbes: Can `readInStreamErrorPropertyResult` be null at this point? And if so, what would be the outcome of passing `null` to `ODataErrorException`?

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -999,6 +1350,589 @@ private void RemoveFirstNodeInBuffer()
...
+
+        /// <summary>
+        /// Try to read an error structure from the stream. Return null if no error structure can be read.
+        /// </summary>
+        /// <param name="error">An <see cref="ODataError"/> instance that was read from the reader or null if none could be read.</param>
+        /// <returns>A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains a tuple comprising of:
+        /// 1). A value of true if an <see cref="ODataError"/> instance that was read; otherwise false.
+        /// 2). An <see cref="ODataError"/> instance that was read from the reader or null if none could be read.</returns>
+        private async Task<Tuple<bool, ODataError>> TryReadInStreamErrorPropertyValueAsync()
```
habbes, line 1420:
```suggestion
        private async Task<(bool, ODataError)> TryReadInStreamErrorPropertyValueAsync()

```

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -999,6 +1350,589 @@ private void RemoveFirstNodeInBuffer()
...
+        {
+            Debug.Assert(this.parsingInStreamError, "this.parsingInStreamError");
+            this.AssertBuffering();
+            this.AssertAsynchronous();
+
+            // We expect a start-object node here
+            if (this.currentBufferedNode.NodeType != JsonNodeType.StartObject)
+            {
+                return Tuple.Create(false, (ODataError)null);
```
habbes, line 1429:
```suggestion
                return (false, (ODataError)null);

```
You could also name the fields:
`return (hasError: false, error: (ODataError)null);`

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -999,6 +1350,589 @@ private void RemoveFirstNodeInBuffer()
...
+
+                    case JsonConstants.ODataErrorDetailsName:
+                        if (!ODataJsonLightReaderUtils.ErrorPropertyNotFound(
+                                ref propertiesFoundBitmask,
+                                ODataJsonLightReaderUtils.ErrorPropertyBitMask.Details))
+                        {
+                            return Tuple.Create(false, error);
+                        }
+
+                        Tuple<bool, List<ODataErrorDetail>> readErrorDetailsPropertyResult = await this.TryReadErrorDetailsPropertyValueAsync()
```
habbes, line 1509: Consider naming the tuple fields for better readibility.

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -999,6 +1350,589 @@ private void RemoveFirstNodeInBuffer()
...
+
+                        error.Message = readErrorStringPropertyResult.Item2;
+                        break;
+
+                    case JsonConstants.ODataErrorTargetName:
+                        if (!ODataJsonLightReaderUtils.ErrorPropertyNotFound(
+                            ref propertiesFoundBitmask,
+                            ODataJsonLightReaderUtils.ErrorPropertyBitMask.Target))
+                        {
+                            return Tuple.Create(false, error);
```
habbes, line 1488:
```suggestion
                            return (false, error);

```

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -999,6 +1350,589 @@ private void RemoveFirstNodeInBuffer()
...
+            await this.ReadInternalAsync()
+                .ConfigureAwait(false);
+
+            ODataError error = new ODataError();
+
+            // We expect one of the supported properties for the value (or end-object)
+            ODataJsonLightReaderUtils.ErrorPropertyBitMask propertiesFoundBitmask = ODataJsonLightReaderUtils.ErrorPropertyBitMask.None;
+            while (this.currentBufferedNode.NodeType == JsonNodeType.Property)
+            {
+                Tuple<bool, string> readErrorStringPropertyResult;
```
habbes, line 1442: Maybe this variable should be defined where it's used instead of at the top of the switch case? So we don't have to keep it in mind until we need it (when reading the code).

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -999,6 +1350,589 @@ private void RemoveFirstNodeInBuffer()
...
+                var propertyName = (string)this.currentBufferedNode.Value;
+
+                switch (propertyName)
+                {
+                    case JsonConstants.ODataErrorCodeName:
+                        if (!ODataJsonLightReaderUtils.ErrorPropertyNotFound(
+                                ref propertiesFoundBitmask,
+                                ODataJsonLightReaderUtils.ErrorPropertyBitMask.Code))
+                        {
+                            return Tuple.Create(false, detail);
```
habbes, line 1646:
```suggestion
                            return (false, detail);

```

habbes (changes requested):

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -235,6 +249,111 @@ public bool Read()
...
+                    .ConfigureAwait(false);
+            }
+
+            Stream result;
+
+            object value = this.GetValueAsync().ConfigureAwait(false);
+
+            if (value == null)
+            {
+                result = new MemoryStream(new byte[0]);
```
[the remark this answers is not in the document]
gathogojr: @habbes `Array.Empty<byte>()` is not available in **net45** and **netstandard1.1**.  I don't know if it'd be worthwhile to do this:
```csharp
#if NETSTANDARD2_0
                result = new MemoryStream(Array.Empty<byte>());
#else
                result = new MemoryStream(new byte[0]);
#endif
```

src/Microsoft.OData.Core/Json/JsonReader.cs
gathogojr: @habbes In `JsonReader` it just returns `true` or `false` but in the `BufferingJsonReader` it actually does call an asynchronous method - `GetValueAsync`. I had actually not considered having an asynchronous equivalent of `CanStream` method initially but I realized I couldn't avoid it when implementing asynchronous support in `BufferingJsonReader`

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
gathogojr: I believe I added the assertion statement in other async methods as well but I'll verify

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
gathogojr: @habbes Named tuples are not support in net45 and netstandard1.1. We could use the `#if NETSTANDARD2_0` directives but it'd be too messy. I'd rather we adopt named tuples after we abandon support for the 2 standards I referenced

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
gathogojr: @habbes No. It can't be null so I'll modify the logic to take that into account. However, passing a `null` value as the argument doesn't cause a problem

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
gathogojr: See my comment here https://github.com/OData/odata.net/pull/2264#discussion_r759368356

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
gathogojr: See my comment here https://github.com/OData/odata.net/pull/2264#discussion_r759368356

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
gathogojr: See my comment here https://github.com/OData/odata.net/pull/2264#discussion_r759368356

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
gathogojr: See my comment here https://github.com/OData/odata.net/pull/2264#discussion_r759368356

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
gathogojr: See my comment here https://github.com/OData/odata.net/pull/2264#discussion_r759368356

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
gathogojr: See my comment here https://github.com/OData/odata.net/pull/2264#discussion_r759368356

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -235,6 +249,140 @@ public bool Read()
...
+        /// <summary>
+        /// Asynchronously checks whether or not the current value can be streamed.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous operation.
+        /// true if the current value can be streamed; otherwise false.</returns>
+        public virtual async Task<bool> CanStreamAsync()
+        {
+            this.AssertAsynchronous();
+
+            if (!this.isBuffering && this.asyncInnerReader is IJsonStreamReaderAsync asyncStreamReader)
```
corranrogue9, line 304: I see this check for `IJsonStreamReaderAsync ` in a few different places. I wonder if it makes sense to have a field where this is stored so that the casting only has to happen once. I also wonder if that field should be set in the constructor. I can see there being different use cases where it makes sense to do it on the fly versus doing it in the constructor. Any thoughts?

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -235,6 +249,140 @@ public bool Read()
...
+            {
+                return await asyncStreamReader.CanStreamAsync()
+                    .ConfigureAwait(false);
+            }
+
+            object value = await this.GetValueAsync()
+                .ConfigureAwait(false);
+            return value is string || value == null || this.NodeType == JsonNodeType.StartArray || this.NodeType == JsonNodeType.StartObject;
```
corranrogue9, line 312: I don't think the subsequent conditions are predicated on this first one for any reason, so I'm wondering if it makes sense to do the cast last due to short circuit evaluation

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
gathogojr: @habbes `IJsonReader.Value` (and `IJsonReaderAsync.GetValueAsync()`) only read the value. They do not advance the reader. For this same reason, you can call them multiple times with the same result. All the same I have rewritten the logic to mirror the sync version.

habbes (approved):

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
gathogojr: @corranrogue9 Does a cast actually happen here given that `value` is already boxed and `string` is a reference type? In addition, supposing `StartArray` and `StartObject` occur rarely, do we not end up always doing the "cheap" comparisons as well as evaluating `value is string` expression?

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -999,6 +1316,571 @@ private void RemoveFirstNodeInBuffer()
...
+                            ODataJsonLightReaderUtils.ErrorPropertyBitMask.Message))
+                        {
+                            return Tuple.Create(false, error);
+                        }
+
+                        readErrorStringPropertyResult = await this.TryReadErrorStringPropertyValueAsync()
+                            .ConfigureAwait(false);
+                        if (!readErrorStringPropertyResult.Item1)
+                        {
+                            return Tuple.Create(false, error);
```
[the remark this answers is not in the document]
gathogojr: @marabooy To explain the logic, the only time this method returns the second component of the tuple as `null` is before entering this while loop. Within this loop, we return the `error` object with as many properties we have read before coming across some unintelligible property. For example, if successfully read `code`, `message`, `target`, and `innererror` properties and the come across `details` and we're unhappy, we don't return a tuple with `false` as first component and `null` as second component. Instead, the second component will contain the `error` object with the properties we had read successfully populated.

marabooy (approved):

KenitoInc (approved):

PR #2267, opened 2021-12-02
Implement asynchronous support in ODataJsonLightDeserializer

gathogojr:
### Issues

### Description

This PR implements support for the following 3 classes:
- `ODataJsonLightDeserializer`
- `ODataJsonLightPropertyAndValueDeserializer`
- `ODataJsonLightResourceDeserializer`

It would have been desirable to have separate PRs for the above 3 classes but there exists a dependency that was not that easy to navigate around without risking a breaking change.
Using the synchronous API to illustrate, the `ODataJsonLightDeserializer` expects the `ReadPropertyCustomAnnotationValue` `Func` delegate property to be initialized from the derived classes (`ODataJsonLightPropertyAndValueDeserializer` or `ODataJsonLightResourceDeserializer`). Implementation of asynchronous support `ODataJsonLightDeserializer` involved introducing an asynchronous `Func` delegate `ReadPropertyCustomAnnotationValueAsync`. Merging changes to `ODataJsonLightDeserializer` without relevant changes to the derived classes would have caused a breaking changes if someone invoked the asynchronous API that was previously implemented as an asynchronous wrapper on top of the synchronous methods. In particular, if the `ProcessPropertyAsync` method defined in `ODataJsonLightDeserializer` class was to be invoked before the `Func` delegate is initialized, particular code paths would break.

`ODataJsonLightDeserializer` is an abstract class that classes that handle the actual deserialization implement
`ODataJsonLightPropertyAndValueDeserializer` implements the logic for deserializing properties and their corresponding values
`ODataJsonLightResourceDeserializer` implements the logic for deserializing an OData resource

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/IODataJsonOperationsDeserializerContext.cs
+++ b/src/Microsoft.OData.Core/Json/IODataJsonOperationsDeserializerContext.cs
@@ -21,6 +21,11 @@ namespace Microsoft.OData.Json
         IJsonReader JsonReader { get; }
 
         /// <summary>
+        /// The asynchronous JSON reader to read the operations value from.
+        /// </summary>
+        IJsonReaderAsync AsynchronousJsonReader { get; }
+
+        /// <summary>
         /// Given a URI from the payload, this method will try to make it absolute, or fail otherwise.
         /// </summary>
         /// <param name="uriFromPayload">The URI string from the payload to process.</param>
--- a/src/Microsoft.OData.Core/JsonLight/JsonLightConstants.cs
+++ b/src/Microsoft.OData.Core/JsonLight/JsonLightConstants.cs
@@ -116,5 +116,20 @@ namespace Microsoft.OData.JsonLight
 
         /// <summary>The simplified Removed property name.</summary>
         internal const string SimplifiedODataRemovedPropertyName = "@removed";
+
+        /// <summary>The prefixed OData Context property name.</summary>
+        internal const string PrefixedODataContextPropertyName = "@odata.context";
+
+        /// <summary>The prefixed OData Type property name.</summary>
+        internal const string PrefixedODataTypePropertyName = "@odata.type";
+
+        /// <summary>The prefixed OData Id property name.</summary>
+        internal const string PrefixedODataIdPropertyName = "@odata.id";
+
+        /// <summary>The prefixed OData Null property name.</summary>
+        internal const string PrefixedODataNullPropertyName = "@odata.null";
+
+        /// <summary>The prefixed OData Removed property name.</summary>
+        internal const string PrefixedODataRemovedPropertyName = "@odata.removed";
     }
 }
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
@@ -11,12 +11,10 @@ namespace Microsoft.OData.JsonLight
     using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
     using System.Globalization;
-    using System.IO;
     using System.Linq;
-    using System.Text;
     using System.Threading.Tasks;
-    using Microsoft.OData.Edm;
     using Microsoft.OData;
+    using Microsoft.OData.Edm;
     using Microsoft.OData.Evaluation;
     using Microsoft.OData.Json;
     #endregion Namespaces
@@ -171,7 +169,30 @@ namespace Microsoft.OData.JsonLight
         /// <summary>
         /// Function called to read property custom annotation value.
         /// </summary>
+        /// <Remarks>
+        /// Function takes:
+        /// * The duplicate property names checker
+        /// * The annotation name
+        /// Function returns:
+        /// * The read property annotation value
+        /// </Remarks>
         protected Func<PropertyAndAnnotationCollector, string, object> ReadPropertyCustomAnnotationValue
+        {
+            get;
+            set;
+        }
+
+        /// <summary>
+        /// Delegate called to read property custom annotation value asynchronously.
+        /// </summary>
+        /// <Remarks>
+        /// Function takes:
+        /// * The duplicate property names checker
+        /// * The annotation name
+        /// Function returns:
+        /// * A task that represents the asynchronous read operation. The value of the TResult parameter contains the read property annotation value
+        /// </Remarks>
+        protected Func<PropertyAndAnnotationCollector, string, Task<object>> ReadPropertyCustomAnnotationValueAsync
         {
             get;
             set;
@@ -279,53 +300,52 @@ namespace Microsoft.OData.JsonLight
 
 
         /// <summary>
-        /// Read the start of the top-level data wrapper in JSON responses.
+        /// Asynchronously read the start of the top-level data wrapper in JSON responses.
         /// </summary>
         /// <param name="payloadKind">The kind of payload we are reading; this guides the parsing of the context URI.</param>
         /// <param name="propertyAndAnnotationCollector">The duplicate property names checker.</param>
         /// <param name="isReadingNestedPayload">true if we are deserializing a nested payload, e.g. a resource, a resource set or a collection within a parameters payload.</param>
         /// <param name="allowEmptyPayload">true if we allow a completely empty payload; otherwise false.</param>
-        /// <returns>The parsed context URI.</returns>
+        /// <returns>A task that represents the asynchronous read operation.</returns>
         /// <remarks>
         /// Pre-Condition:  JsonNodeType.None:      assumes that the JSON reader has not been used yet when not reading a nested payload.
         /// Post-Condition: The reader is positioned on the first property of the payload after having read (or skipped) the context URI property.
         ///                 Or the reader is positioned on an end-object node if there are no properties (other than the context URI which is required in responses and optional in requests).
         /// </remarks>
-        internal Task ReadPayloadStartAsync(
+        internal async Task ReadPayloadStartAsync(
             ODataPayloadKind payloadKind,
             PropertyAndAnnotationCollector propertyAndAnnotationCollector,
             bool isReadingNestedPayload,
             bool allowEmptyPayload)
         {
             this.JsonReader.AssertNotBuffering();
-            Debug.Assert(isReadingNestedPayload || this.JsonReader.NodeType == JsonNodeType.None, "Pre-Condition: JSON reader must not have been used yet when not reading a nested payload.");
+            Debug.Assert(
+                isReadingNestedPayload || this.JsonReader.NodeType == JsonNodeType.None,
+                "Pre-Condition: JSON reader must not have been used yet when not reading a nested payload.");
 
-            return TaskUtils.GetTaskForSynchronousOperation(() =>
-                {
-                    string contextUriAnnotationValue = this.ReadPayloadStartImplementation(
+            string contextUriAnnotationValue = await this.ReadPayloadStartImplementationAsync(
                         payloadKind,
                         propertyAndAnnotationCollector,
                         isReadingNestedPayload,
-                        allowEmptyPayload);
+                        allowEmptyPayload).ConfigureAwait(false);
 
-                    // The context URI is only recognized in non-error response top-level payloads.
-                    // If the payload is nested (for example when we read URL literals) we don't recognize the context URI.
-                    // Top-level error payloads don't need and use the context URI.
-                    if (!isReadingNestedPayload && payloadKind != ODataPayloadKind.Error && contextUriAnnotationValue != null)
-                    {
-                        this.contextUriParseResult = ODataJsonLightContextUriParser.Parse(
-                            this.Model,
-                            contextUriAnnotationValue,
-                            payloadKind,
-                            this.MessageReaderSettings.ClientCustomTypeResolver,
-                            this.JsonLightInputContext.ReadingResponse || payloadKind == ODataPayloadKind.Delta,
-                            this.JsonLightInputContext.MessageReaderSettings.ThrowIfTypeConflictsWithMetadata);
-                    }
+            // The context URI is only recognized in non-error response top-level payloads.
+            // If the payload is nested (for example when we read URL literals) we don't recognize the context URI.
+            // Top-level error payloads don't need and use the context URI.
+            if (!isReadingNestedPayload && payloadKind != ODataPayloadKind.Error && contextUriAnnotationValue != null)
+            {
+                this.contextUriParseResult = ODataJsonLightContextUriParser.Parse(
+                    this.Model,
+                    contextUriAnnotationValue,
+                    payloadKind,
+                    this.MessageReaderSettings.ClientCustomTypeResolver,
+                    this.JsonLightInputContext.ReadingResponse || payloadKind == ODataPayloadKind.Delta,
+                    this.JsonLightInputContext.MessageReaderSettings.ThrowIfTypeConflictsWithMetadata);
+            }
 
 #if DEBUG
-                    this.contextUriParseResultReady = true;
+            this.contextUriParseResultReady = true;
 #endif
-                });
         }
 
         /// <summary>
@@ -544,7 +564,7 @@ namespace Microsoft.OData.JsonLight
 
             // Must make sure the input odata.context has a '@' prefix
             string propertyName = this.JsonReader.GetPropertyName();
-            if (string.CompareOrdinal(JsonLightConstants.ODataPropertyAnnotationSeparatorChar + ODataAnnotationNames.ODataContext, propertyName) != 0
+            if (!string.Equals(JsonLightConstants.PrefixedODataContextPropertyName, propertyName, StringComparison.Ordinal)
                 && !this.CompareSimplifiedODataAnnotation(JsonLightConstants.SimplifiedODataContextPropertyName, propertyName))
             {
                 if (!failOnMissingContextUriAnnotation || payloadKind == ODataPayloadKind.Unsupported)
@@ -567,6 +587,203 @@ namespace Microsoft.OData.JsonLight
         }
 
         /// <summary>
+        /// Asynchronously reads and validates a string value from the json reader.
+        /// </summary>
+        /// <param name="annotationName">The name of the annotation being read (used for error reporting).</param>
+        /// <returns>
+        /// A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains the string that was read.
+        /// </returns>
+        internal async Task<string> ReadAndValidateAnnotationStringValueAsync(string annotationName)
+        {
+            string valueRead = await this.JsonReader.ReadStringValueAsync(annotationName)
+                .ConfigureAwait(false);
+            ODataJsonLightReaderUtils.ValidateAnnotationValue(valueRead, annotationName);
+            
+            return valueRead;
+        }
+
+        /// <summary>
+        /// Asynchronously reads a string value from the json reader.
+        /// </summary>
+        /// <param name="annotationName">The name of the annotation being read (used for error reporting).</param>
+        /// <returns>
+        /// A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains the string that was read.
+        /// </returns>
+        internal Task<string> ReadAnnotationStringValueAsync(string annotationName)
+        {
+            return this.JsonReader.ReadStringValueAsync(annotationName);
+        }
+
+        /// <summary>
+        /// Asynchronously reads a string value from the json reader and processes it as a Uri.
+        /// </summary>
+        /// <param name="annotationName">The name of the annotation being read (used for error reporting).</param>
+        /// <returns>
+        /// A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains the Uri that was read.
+        /// </returns>
+        internal async Task<Uri> ReadAnnotationStringValueAsUriAsync(string annotationName)
+        {
+            string stringValue = await this.JsonReader.ReadStringValueAsync(annotationName)
+                .ConfigureAwait(false);
+
+            if (stringValue == null)
+            {
+                return null;
+            }
+
+            return this.ReadingResponse ? this.ProcessUriFromPayload(stringValue) : new Uri(stringValue, UriKind.RelativeOrAbsolute);
+        }
+
+        /// <summary>
+        /// Asynchronously reads and validates a string value from the json reader and processes it as a Uri.
+        /// </summary>
+        /// <param name="annotationName">The name of the annotation being read (used for error reporting).</param>
+        /// <returns>
+        /// A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains the Uri that was read.
+        /// </returns>
+        internal async Task<Uri> ReadAndValidateAnnotationStringValueAsUriAsync(string annotationName)
+        {
+            string stringValue = await this.ReadAndValidateAnnotationStringValueAsync(annotationName)
+                .ConfigureAwait(false);
+            return this.ReadingResponse ? this.ProcessUriFromPayload(stringValue) : new Uri(stringValue, UriKind.RelativeOrAbsolute);
+        }
+
+        /// <summary>
+        /// Asynchronously reads and validates a value from the json reader and processes it as a long.
+        /// The input value could be string or number
+        /// </summary>
+        /// <param name="annotationName">The name of the annotation being read.</param>
+        /// <returns>
+        /// A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains the <see cref="long"/> value that was read.
+        /// </returns>
+        internal async Task<long> ReadAndValidateAnnotationAsLongForIeee754CompatibleAsync(string annotationName)
+        {
+            object value = await this.JsonReader.ReadPrimitiveValueAsync()
+                .ConfigureAwait(false);
+
+            ODataJsonLightReaderUtils.ValidateAnnotationValue(value, annotationName);
+
+            if ((value is string) ^ this.JsonReader.IsIeee754Compatible)
+            {
+                throw new ODataException(Strings.ODataJsonReaderUtils_ConflictBetweenInputFormatAndParameter(Metadata.EdmConstants.EdmInt64TypeName));
+            }
+
+            return (long)ODataJsonLightReaderUtils.ConvertValue(
+                    value,
+                    EdmCoreModel.Instance.GetInt64(false),
+                    this.MessageReaderSettings,
+                    validateNullValue: true,
+                    propertyName: annotationName,
+                    converter: this.JsonLightInputContext.PayloadValueConverter);
+        }
+
+        /// <summary>
+        /// Asynchronously parses JSON object property starting with the current position of the JSON reader.
+        /// </summary>
+        /// <param name="propertyAndAnnotationCollector">The duplicate property names checker to use, it will also store the property annotations found.</param>
+        /// <param name="readPropertyAnnotationValueDelegate">Delegate to read property annotation value.</param>
+        /// <param name="handlePropertyDelegate">Delegate to handle the result of parsing property.</param>
+        /// <returns>A task that represents the asynchronous read operation.</returns>
+        internal async Task ProcessPropertyAsync(
+            PropertyAndAnnotationCollector propertyAndAnnotationCollector,
+            Func<string, Task<object>> readPropertyAnnotationValueDelegate,
+            Func<PropertyParsingResult, string, Task> handlePropertyDelegate)
+        {
+            Debug.Assert(propertyAndAnnotationCollector != null, $"{nameof(propertyAndAnnotationCollector)} != null");
+            Debug.Assert(readPropertyAnnotationValueDelegate != null, $"{nameof(readPropertyAnnotationValueDelegate)} != null");
+            Debug.Assert(handlePropertyDelegate != null, $"{nameof(handlePropertyDelegate)} != null");
+            this.AssertJsonCondition(JsonNodeType.Property);
+
+            Tuple<PropertyParsingResult, string> parsePropertyResult = await this.ParsePropertyAsync(
+                propertyAndAnnotationCollector,
+                readPropertyAnnotationValueDelegate).ConfigureAwait(false);
+            PropertyParsingResult propertyParsingResult = parsePropertyResult.Item1;
+            string propertyName = parsePropertyResult.Item2;
+
+            while (propertyParsingResult == PropertyParsingResult.CustomInstanceAnnotation && this.ShouldSkipCustomInstanceAnnotation(propertyName))
+            {
+                // Read over the property name
+                await this.JsonReader.ReadAsync()
+                    .ConfigureAwait(false);
+
+                // Skip over the instance annotation value
+                await this.JsonReader.SkipValueAsync()
+                    .ConfigureAwait(false);
+                parsePropertyResult = await this.ParsePropertyAsync(
+                    propertyAndAnnotationCollector,
+                    readPropertyAnnotationValueDelegate).ConfigureAwait(false);
+                propertyParsingResult = parsePropertyResult.Item1;
+                propertyName = parsePropertyResult.Item2;
+            }
+
+            await handlePropertyDelegate(propertyParsingResult, propertyName)
+                .ConfigureAwait(false);
+            if (propertyParsingResult != PropertyParsingResult.EndOfObject
+                && propertyParsingResult != PropertyParsingResult.CustomInstanceAnnotation)
+            {
+                propertyAndAnnotationCollector.MarkPropertyAsProcessed(propertyName);
+            }
+        }
+
+        /// <summary>
+        /// Asynchronously reads the odata.context annotation.
+        /// </summary>
+        /// <param name="payloadKind">The payload kind for which to read the context URI.</param>
+        /// <param name="propertyAndAnnotationCollector">The duplicate property names checker.</param>
+        /// <param name="failOnMissingContextUriAnnotation">true if the method should fail if the context URI annotation is missing, false if that can be ignored.</param>
+        /// <returns>
+        /// A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains the value of the context URI annotation.
+        /// </returns>
+        internal async Task<string> ReadContextUriAnnotationAsync(
+            ODataPayloadKind payloadKind,
+            PropertyAndAnnotationCollector propertyAndAnnotationCollector,
+            bool failOnMissingContextUriAnnotation)
+        {
+            if (this.JsonReader.NodeType != JsonNodeType.Property)
+            {
+                if (!failOnMissingContextUriAnnotation || payloadKind == ODataPayloadKind.Unsupported)
+                {
+                    // Do not fail during payload kind detection
+                    return null;
+                }
+
+                throw new ODataException(Strings.ODataJsonLightDeserializer_ContextLinkNotFoundAsFirstProperty);
+            }
+
+            // Must make sure the input odata.context has a '@' prefix
+            string propertyName = await this.JsonReader.GetPropertyNameAsync()
+                .ConfigureAwait(false);
+            if (!string.Equals(JsonLightConstants.PrefixedODataContextPropertyName, propertyName, StringComparison.Ordinal)
+                && !this.CompareSimplifiedODataAnnotation(JsonLightConstants.SimplifiedODataContextPropertyName, propertyName))
+            {
+                if (!failOnMissingContextUriAnnotation || payloadKind == ODataPayloadKind.Unsupported)
+                {
+                    // Do not fail during payload kind detection
+                    return null;
+                }
+
+                throw new ODataException(Strings.ODataJsonLightDeserializer_ContextLinkNotFoundAsFirstProperty);
+            }
+
+            if (propertyAndAnnotationCollector != null)
+            {
+                propertyAndAnnotationCollector.MarkPropertyAsProcessed(propertyName);
+            }
+
+            // Read over the property name
+            await this.JsonReader.ReadNextAsync()
+                .ConfigureAwait(false);
+            return await this.JsonReader.ReadStringValueAsync()
+                .ConfigureAwait(false);
+        }
+
+        /// <summary>
         /// Compares the JSON property name with the simplified OData annotation property name.
         /// </summary>
         /// <param name="simplifiedPropertyName">The simplified OData annotation property name.</param>
@@ -578,7 +795,7 @@ namespace Microsoft.OData.JsonLight
             Debug.Assert(simplifiedPropertyName.IndexOf('.') == -1, "simplifiedPropertyName must not be namespace-qualified.");
 
             return this.JsonLightInputContext.OptionalODataPrefix &&
-                   string.CompareOrdinal(simplifiedPropertyName, propertyName) == 0;
+                   string.Equals(simplifiedPropertyName, propertyName, StringComparison.Ordinal);
         }
 
         /// <summary>
@@ -624,7 +841,7 @@ namespace Microsoft.OData.JsonLight
         /// <returns>true is the instance annotation, false is not</returns>
         private static bool IsInstanceAnnotation(string annotationName)
         {
-            if (!String.IsNullOrEmpty(annotationName) && annotationName[0] == JsonLightConstants.ODataPropertyAnnotationSeparatorChar)
+            if (!string.IsNullOrEmpty(annotationName) && annotationName[0] == JsonLightConstants.ODataPropertyAnnotationSeparatorChar)
             {
                 return true;
             }
@@ -642,9 +859,9 @@ namespace Microsoft.OData.JsonLight
         /// Note that when we add new odata annotations that cannot be skipped, we would bump the protocol version.
         /// </remarks>
         /// <param name="annotationName">The annotation name in question.</param>
-        /// <param name="annotationValue">Outputs the .</param>
+        /// <param name="annotationValue">The annotation value that was read.</param>
         /// <returns>Returns true if the annotation name and value is skipped; returns false otherwise.</returns>
-        private bool SkippedOverUnknownODataAnnotation(string annotationName, out object annotationValue)
+        private bool SkipOverUnknownODataAnnotation(string annotationName, out object annotationValue)
         {
             Debug.Assert(!string.IsNullOrEmpty(annotationName), "!string.IsNullOrEmpty(annotationName)");
             this.AssertJsonCondition(JsonNodeType.Property);
@@ -720,7 +937,7 @@ namespace Microsoft.OData.JsonLight
                 bool isPropertyAnnotation = TryParsePropertyAnnotation(nameFromReader, out propertyNameFromReader, out annotationNameFromReader);
 
                 // reading a nested delta resource set
-                if (isPropertyAnnotation && String.CompareOrdinal(this.CompleteSimplifiedODataAnnotation(annotationNameFromReader), ODataAnnotationNames.ODataDelta) == 0)
+                if (isPropertyAnnotation && string.Equals(this.CompleteSimplifiedODataAnnotation(annotationNameFromReader), ODataAnnotationNames.ODataDelta, StringComparison.Ordinal))
                 {
                     // Read over the property name.
                     this.JsonReader.Read();
@@ -737,7 +954,7 @@ namespace Microsoft.OData.JsonLight
 
                 // If parsedPropertyName is set and is different from the property name the reader is currently on,
                 // we have parsed a property annotation for a different property than the one at the current position.
-                if (parsedPropertyName != null && string.CompareOrdinal(parsedPropertyName, propertyNameFromReader) != 0)
+                if (parsedPropertyName != null && !string.Equals(parsedPropertyName, propertyNameFromReader, StringComparison.Ordinal))
                 {
                     if (ODataJsonLightReaderUtils.IsAnnotationProperty(parsedPropertyName))
                     {
@@ -752,10 +969,10 @@ namespace Microsoft.OData.JsonLight
                 {
                     annotationNameFromReader = this.CompleteSimplifiedODataAnnotation(annotationNameFromReader);
 
-                    // If this is a unknown odata annotation targeting a property, we skip over it. See remark on the method SkippedOverUnknownODataAnnotation() for detailed explanation.
+                    // If this is a unknown odata annotation targeting a property, we skip over it. See remark on the method SkipOverUnknownODataAnnotation() for detailed explanation.
                     // Note that we don't skip over unknown odata annotations targeting another annotation. We don't allow annotations (except odata.type) targeting other annotations,
                     // so this.ProcessPropertyAnnotation() will test and fail for that case.
-                    if (!ODataJsonLightReaderUtils.IsAnnotationProperty(propertyNameFromReader) && this.SkippedOverUnknownODataAnnotation(annotationNameFromReader, out annotationValue))
+                    if (!ODataJsonLightReaderUtils.IsAnnotationProperty(propertyNameFromReader) && this.SkipOverUnknownODataAnnotation(annotationNameFromReader, out annotationValue))
                     {
                         propertyAndAnnotationCollector.AddODataPropertyAnnotation(propertyNameFromReader, annotationNameFromReader, annotationValue);
                         continue;
@@ -769,8 +986,8 @@ namespace Microsoft.OData.JsonLight
                     continue;
                 }
 
-                // If this is a unknown odata annotation, skip over it. See remark on the method SkippedOverUnknownODataAnnotation() for detailed explanation.
-                if (isInstanceAnnotation && this.SkippedOverUnknownODataAnnotation(propertyNameFromReader, out annotationValue))
+                // If this is a unknown odata annotation, skip over it. See remark on the method SkipOverUnknownODataAnnotation() for detailed explanation.
+                if (isInstanceAnnotation && this.SkipOverUnknownODataAnnotation(propertyNameFromReader, out annotationValue))
                 {
                     // collect 'odata.<unknown>' annotation:
                     // here we know the original property name contains no '@', but '.' dot
@@ -834,7 +1051,7 @@ namespace Microsoft.OData.JsonLight
         private void ProcessPropertyAnnotation(string annotatedPropertyName, string annotationName, PropertyAndAnnotationCollector propertyAndAnnotationCollector, Func<string, object> readPropertyAnnotationValue)
         {
             // We don't currently support annotation targeting an instance annotation except for the @odata.type property annotation.
-            if (ODataJsonLightReaderUtils.IsAnnotationProperty(annotatedPropertyName) && string.CompareOrdinal(annotationName, ODataAnnotationNames.ODataType) != 0)
+            if (ODataJsonLightReaderUtils.IsAnnotationProperty(annotatedPropertyName) && !string.Equals(annotationName, ODataAnnotationNames.ODataType, StringComparison.Ordinal))
             {
                 throw new ODataException(Strings.ODataJsonLightDeserializer_OnlyODataTypeAnnotationCanTargetInstanceAnnotation(annotationName, annotatedPropertyName, ODataAnnotationNames.ODataType));
             }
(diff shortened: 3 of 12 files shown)


src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -908,6 +1099,369 @@ private void ReadODataOrCustomInstanceAnnotationValue(string annotatedPropertyNa
...
+
+        /// <summary>
+        /// Asynchronously reads "odata." or custom instance annotation's value.
+        /// </summary>
+        /// <param name="annotationName">The annotation name.</param>
+        /// <returns>The annotation value.</returns>
+        /// <returns>A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains the annotation value.</returns>
+#if NETSTANDARD2_0
+        private async ValueTask<object> ReadODataOrCustomInstanceAnnotationValueAsync(string annotationName)
```
habbes, line 1146: Out of curiosity, why does this return a `ValueTask`, all code paths seem to be async? Or do async paths also return `ValueTask`s?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -908,6 +1099,369 @@ private void ReadODataOrCustomInstanceAnnotationValue(string annotatedPropertyNa
...
+        {
+            // We don't currently support annotation targeting an instance annotation except for the @odata.type property annotation.
+            if (ODataJsonLightReaderUtils.IsAnnotationProperty(annotatedPropertyName)
+                && string.CompareOrdinal(annotationName, ODataAnnotationNames.ODataType) != 0)
+            {
+                return TaskUtils.GetFaultedTask<ODataException>(
+                    new ODataException(Strings.ODataJsonLightDeserializer_OnlyODataTypeAnnotationCanTargetInstanceAnnotation(
+                        annotationName,
+                        annotatedPropertyName,
+                        ODataAnnotationNames.ODataType)));
```
habbes, line 1353: Maybe you've explained this before, but why don't we use built-ins like `Task.FromException`?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -908,6 +1099,369 @@ private void ReadODataOrCustomInstanceAnnotationValue(string annotatedPropertyNa
...
+                    || (this is ODataJsonLightErrorDeserializer && this.MessageReaderSettings.ShouldIncludeAnnotation == null))
+                {
+                    propertyAndAnnotationCollector.CheckIfPropertyOpenForAnnotations(annotatedPropertyName, annotationName);
+                    await this.JsonReader.SkipValueAsync()
+                        .ConfigureAwait(false);
+                }
+                else
+                {
+                    Debug.Assert(ReadPropertyCustomAnnotationValueAsync != null, "ReadPropertyCustomAnnotationValueAsync != null");
+                    var propertyCustomAnnotationValue = await ReadPropertyCustomAnnotationValueAsync(propertyAndAnnotationCollector, annotationName)
```
habbes, line 1399: Maybe use explicit type here instead of `var`?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeserializerTests.cs
```diff
@@ -1144,25 +1145,385 @@ public void ParsingExpectedComplexPropertyActualNotShouldThrow()
...
+            bool isResponse = true,
+            bool isAsync = false,
+            bool isIeee754Compatible = false)
         {
             var messageInfo = new ODataMessageInfo
             {
                 IsResponse = isResponse,
-                MediaType = JsonLightUtils.JsonLightStreamingMediaType,
-                IsAsync = false,
+                MediaType = CreateMediaType(isIeee754Compatible),
```
habbes, line 1500: Why change the media type?

habbes (approved):
Looks good, but had some minor questions and comments.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -327,6 +322,103 @@ internal object ReadODataOrCustomInstanceAnnotationValue(string annotationName,
...
+            // If this term is defined in the model, look up its type. If the term is not in the model, this will be null.
+            IEdmTypeReference expectedTypeFromTerm = MetadataUtils.LookupTypeOfTerm(annotationName, this.Model);
+            return this.ReadNonEntityValueImplementationAsync(
+                odataType,
+                expectedTypeFromTerm,
+                null, /*propertyAndAnnotationCollector*/
+                null, /*collectionValidator*/
+                false, /*validateNullValue*/
+                false, /*isTopLevelPropertyValue*/
+                false, /*insideResourceValue Always pass false here to try read @odata.type annotation in custom instance annotations*/
```
g2mula:
```suggestion
                propertyAndAnnotationCollector: null,
                collectionValidator: null,
                validateNullValue: false,
                isTopLevelPropertyValue: false,
                insideResourceValue: false, // Always pass false here to try read @odata.type annotation in custom instance annotations
```
Nit: Prefer named parameters to /* parameterName */
Probably a solutionwide discussion required for this.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,122 +1157,404 @@ protected EdmTypeKind GetNonEntityValueKind()
...
+                    if (typeReference.IsNonEntityCollectionType())
+                    {
+                        ReaderValidationUtils.ValidateNullValue(typeReference, true,
+                                                  true, propertyName, false);
+                    }
+                    else
+                    {
+                        throw new ODataException(ODataErrorStrings.ODataJsonLightResourceDeserializer_CannotReadCollectionNestedResource(jsonReaderNodeType, propertyName));
+                    }
+
```
g2mula:
```suggestion
```
Nit: extra whitespace

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,122 +1157,404 @@ protected EdmTypeKind GetNonEntityValueKind()
...
+                || this.CompareSimplifiedODataAnnotation(JsonLightConstants.SimplifiedODataTypePropertyName, propertyName))
+            {
+                // Read over the property name
+                this.JsonReader.ReadNext();
+                payloadTypeName = this.ReadODataTypeAnnotationValue();
+                result = true;
+            }
+
+            this.AssertJsonCondition(JsonNodeType.Property, JsonNodeType.EndObject);
+            return result;
```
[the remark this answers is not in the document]
g2mula: Probably meant to also remove the extra whitespace after the return.

g2mula (commented):
This PR is really large, it could benefit from a high level summary in the PR description and what reviewers might wish to focus on...

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -566,6 +574,203 @@ internal void AssertJsonCondition(params JsonNodeType[] allowedNodeTypes)
...
+                    return null;
+                }
+
+                throw new ODataException(Strings.ODataJsonLightDeserializer_ContextLinkNotFoundAsFirstProperty);
+            }
+
+            // Must make sure the input odata.context has a '@' prefix
+            string propertyName = await this.JsonReader.GetPropertyNameAsync()
+                .ConfigureAwait(false);
+            if (string.CompareOrdinal(JsonLightConstants.ODataPropertyAnnotationSeparatorChar + ODataAnnotationNames.ODataContext, propertyName) != 0
```
KenitoInc, line 750: I don't understand the logic on 750-760

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -923,6 +1128,403 @@ private void ReadODataOrCustomInstanceAnnotationValue(string annotatedPropertyNa
                 this.AssertJsonCondition(JsonNodeType.Property, JsonNodeType.EndObject);
             }
 
+            return null;
+        }
```
KenitoInc, line 1132:
```suggestion
        }

```

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -1061,136 +1062,2239 @@ internal ODataJsonLightReaderNestedInfo ReadPropertyWithoutValue(IODataJsonLight
...
+            Debug.Assert(resourceState != null, "resourceState != null");
+            this.AssertJsonCondition(JsonNodeType.Property, JsonNodeType.EndObject);
+
+            // If the current node is the odata.type property - read it.
+            if (this.JsonReader.NodeType == JsonNodeType.Property)
+            {
+                string propertyName = await this.JsonReader.GetPropertyNameAsync()
+                    .ConfigureAwait(false);
+                // TODO: const to avoid repeated concatenation - JsonLightConstants.ODataPropertyAnnotationSeparatorChar + ODataAnnotationNames.ODataType?
+                if (string.CompareOrdinal(JsonLightConstants.ODataPropertyAnnotationSeparatorChar + ODataAnnotationNames.ODataType, propertyName) == 0
```
marabooy: nit:
```suggestion
                if (string.CompareOrdinal(string.Concat(JsonLightConstants.ODataPropertyAnnotationSeparatorChar, ODataAnnotationNames.ODataType), propertyName) == 0

```

marabooy: I am guessing the compile will do this as well behind the scenes.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -1061,136 +1062,2239 @@ internal ODataJsonLightReaderNestedInfo ReadPropertyWithoutValue(IODataJsonLight
...
+                    {
+                        // If this is not called for reading ResourceSetStart and we already scanned ahead and processed all resource set properties, we already checked for duplicate property names.
+                        // Use an empty duplicate property name checker since this.ParsePropertyAsync() read through the same property annotation of instance annotations again.
+                        propertyAndAnnotationCollector = new PropertyAndAnnotationCollector(false);
+                    }
+
+                    await this.ProcessPropertyAsync(
+                        propertyAndAnnotationCollector,
+                        this.ReadTypePropertyAnnotationValueAsync,
+                        async (propertyParseResult, propertyName) =>
```
marabooy: To reduce indentation we could move this to another function within the class instead of the lambda. This would reduce the captured arguments e.g. resourceSet  in 1629 which would reduce the number of allocations.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -1061,136 +1062,2239 @@ internal ODataJsonLightReaderNestedInfo ReadPropertyWithoutValue(IODataJsonLight
...
+                        .ConfigureAwait(false);
+                    break;
+
+                case ODataAnnotationNames.ODataDeltaLink:
+                    resourceSet.DeltaLink = await this.ReadAndValidateAnnotationStringValueAsUriAsync(ODataAnnotationNames.ODataDeltaLink)
+                        .ConfigureAwait(false);
+                    break;
+                case ODataAnnotationNames.ODataType:
+
+                    // TODO: skip the odata.type;
```
marabooy: Todo, todo, ..... :)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -1061,136 +1062,2239 @@ internal ODataJsonLightReaderNestedInfo ReadPropertyWithoutValue(IODataJsonLight
...
+        internal async Task<ODataJsonLightReaderNestedInfo> ReadPropertyWithoutValueAsync(IODataJsonLightReaderResourceState resourceState, string propertyName)
+        {
+            Debug.Assert(resourceState != null, "resourceState != null");
+            Debug.Assert(!string.IsNullOrEmpty(propertyName), "!string.IsNullOrEmpty(propertyName)");
+            this.AssertJsonCondition(JsonNodeType.Property, JsonNodeType.EndObject);
+
+            ODataJsonLightReaderNestedInfo readerNestedInfo = null;
+            IEdmStructuredType resourceType = resourceState.ResourceType;
+            IEdmProperty edmProperty = this.ReaderValidator.ValidatePropertyDefined(propertyName, resourceType);
+            if (edmProperty != null && !edmProperty.Type.IsUntyped())
```
marabooy: nit: If we inverted the conditions and returned from each block we could reduce the indentation :). not a blocker.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,148 +1163,439 @@ protected EdmTypeKind GetNonEntityValueKind()
...
-                string payloadTypeName = null;
-                if (this.ReadingResourceProperty(propertyAndAnnotationCollector, expectedPropertyTypeReference, out payloadTypeName))
+                // Complex property or collection of complex property.
+                bool isCollection = payloadTypeReference.IsCollection();
+                await ValidateExpandedNestedResourceInfoPropertyValueAsync(
+                    this.JsonReader.NodeType,
+                    () => this.JsonReader.GetValueAsync(),
+                    isCollection,
+                    propertyName,
+                    payloadTypeReference).ConfigureAwait(false);
```
KenitoInc:
```suggestion
                    payloadTypeReference).ConfigureAwait(false);

```

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -1061,136 +1062,2239 @@ internal ODataJsonLightReaderNestedInfo ReadPropertyWithoutValue(IODataJsonLight
...
+                            ? ReadExpandedResourceNestedResourceInfo(resourceState, navigationProperty, propertyName, navigationProperty.Type.ToStructuredType(), this.MessageReaderSettings)
+                            : ReadEntityReferenceLinkForSingletonNavigationLinkInRequest(resourceState, navigationProperty, propertyName, /*isExpanded*/ true);
+                    }
+
+                    resourceState.PropertyAndAnnotationCollector.ValidatePropertyUniquenessOnNestedResourceInfoStart(readerNestedResourceInfo.NestedResourceInfo);
+                    readerNestedInfo = readerNestedResourceInfo;
+                }
+                else
+                {
+                    var derivedTypeConstraints = this.JsonLightInputContext.Model.GetDerivedTypeConstraints(edmProperty);
```
marabooy: Use the type instead of var for consistency :)

KenitoInc (commented):
A few comments!
The PR looks good

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -1061,136 +1062,2239 @@ internal ODataJsonLightReaderNestedInfo ReadPropertyWithoutValue(IODataJsonLight
...
+
+            if (readerContext.JsonReader.NodeType != JsonNodeType.StartObject)
+            {
+                throw new ODataException(ODataErrorStrings.ODataJsonOperationsDeserializerUtils_OperationsPropertyMustHaveObjectValue(metadataReferencePropertyName, readerContext.JsonReader.NodeType));
+            }
+
+            // read over the start-object node of the metadata object for the operations
+            readerContext.JsonReader.ReadStartObject();
+
+            var operation = this.CreateODataOperationAndAddToEntry(readerContext, metadataReferencePropertyName);
```
marabooy: Explicit type

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -923,6 +1128,403 @@ private void ReadODataOrCustomInstanceAnnotationValue(string annotatedPropertyNa
...
+
+                    await this.ProcessPropertyAnnotationAsync(
+                        propertyNameFromReader,
+                        annotationNameFromReader,
+                        propertyAndAnnotationCollector,
+                        readPropertyAnnotationValueDelegate).ConfigureAwait(false);
+                    continue;
+                }
+
+                // If this is a unknown odata annotation, skip over it. See remark on the method SkippedOverUnknownODataAnnotation() for detailed explanation.
```
habbes, line 1306:
```suggestion
                // If this is a unknown odata annotation, skip over it. See remark on the method SkipOverUnknownODataAnnotationAsync() for detailed explanation.

```

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
habbes: Gentle reminder just in case you missed this question.

habbes (commented):
Here's a preliminary review of the `ODataJsonLightDeserializer`, will go through the other components later.

Minor comments here and there. Looks good otherwise.

I'm wondering whether it would be better to take note of the places where you think `ValueTask` could be beneficial, but stick to `Task<T>` for now and then later resort to `ValueTask` after you can measure the impact. Now it's not 100% clear why you decided to use it in some places but not others.

habbes (commented):
Left some comments.

src/Microsoft.OData.Core/Json/IODataJsonOperationsDeserializerContext.cs
```diff
@@ -20,6 +20,11 @@ internal interface IODataJsonOperationsDeserializerContext
         /// </summary>
         IJsonReader JsonReader { get; }
 
+        /// <summary>
+        /// The asynchronous JSON reader to read the operations value from.
+        /// </summary>
+        IJsonReaderAsync AsynchronousJsonReader { get; }
```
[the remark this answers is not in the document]
gathogojr: @ElizabethOkerio It's used heavily in `ODataJsonLightResourceDeserializer`. For example [here](https://github.com/OData/odata.net/pull/2267/files#diff-ad2dd50a642df648d41ba83b23b775dbd0192e089fd4980d473664da2cc89ecaR3910)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
gathogojr: @habbes Because `Task.FromException` is not available in **net45** and **netstandard1.1**

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeserializerTests.cs
gathogojr: For IEEE754Compatible `long` and `decimal` values to be returned by the service, the request header needs to include `IEEE754Compatible=true` in the `Content-Type` header; otherwise `IEEE754Compatible=false` (default) is included. By changing the media type, we're able to test both scenarios

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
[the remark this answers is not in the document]
gathogojr: @habbes Because we do this kind of comparison quite a lot, I'm thinking of introducing constants that comprise of the prefixed OData annotation names, i.e. `ODataAnnotationNames.PrefixedODataContext` with the value as `@odata.context`. That will altogether deal with the unnecessary allocation.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -923,6 +1128,403 @@ private void ReadODataOrCustomInstanceAnnotationValue(string annotatedPropertyNa
...
+                        .ConfigureAwait(false);
+                    parsedPropertyName = propertyNameFromReader;
+                    return Tuple.Create(PropertyParsingResult.NestedDeltaResourceSet, parsedPropertyName);
+                }
+
+                bool isInstanceAnnotation = false;
+                if (!isPropertyAnnotation)
+                {
+                    isInstanceAnnotation = IsInstanceAnnotation(nameFromReader);
+                    propertyNameFromReader = isInstanceAnnotation ? this.CompleteSimplifiedODataAnnotation(nameFromReader.Substring(1)) : nameFromReader;
```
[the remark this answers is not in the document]
gathogojr: @habbes We only call `Substring` once inside this function so I think it's better to leave such an optimization for another time

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,122 +1157,404 @@ protected EdmTypeKind GetNonEntityValueKind()
...
+        /// Creates an instance of an ODataResourceSet that represents a nested collection
+        /// </summary>
+        /// <param name="resourceState">The current resource state</param>
+        /// <param name="propertyName">The name of the collection property being read</param>
+        /// <returns>An ODataResourceSet with properties set from any instance annotations</returns>
+        private static ODataResourceSet CreateCollectionResourceSet(IODataJsonLightReaderResourceState resourceState, string propertyName)
+        {
+            ODataResourceSet collectionResourceSet = new ODataResourceSet();
+
+            foreach (var propertyAnnotation
```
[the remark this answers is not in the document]
gathogojr: @KenitoInc What in particular (or how) do you think we can optimize? The `GetODataPropertyAnnotations` method currently returns an `IDictionary<string, object>` and the usage is quite widespread. `PropertyAndAnnotationCollector` class is internal and we could change the return type to `Dictionary<string, object>` but it's a non-trivial exercise that would only be justifiable after analyzing the performance benefit to be derived and in a separate PR.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -923,6 +1116,396 @@ private void ReadODataOrCustomInstanceAnnotationValue(string annotatedPropertyNa
...
+                }
+
+                // If parsedPropertyName is set and is different from the property name the reader is currently on,
+                // we have parsed a property annotation for a different property than the one at the current position.
+                if (parsedPropertyName != null && string.CompareOrdinal(parsedPropertyName, propertyNameFromReader) != 0)
```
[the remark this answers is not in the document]
gathogojr: @g2mula Apparently, `string.CompareOrdinal`, or using an overload of `string.Compare` which allows the comparison to be provided, and specifying an ordinal comparison, e.g. `string.Compare(s1, s2, StringComparison.Ordinal)` are said to be the fastest. It's quite likely that one is a syntactic sugar of the other. I don't find `string.CompareOrdinal(s1, s2)`, any less straightforward than `string.Compare(s1, s2, StringComparison.Ordinal)` and I like it better for being shorter.

A longer explanation about the two methods can be found [here](https://stackoverflow.com/a/1452054)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -551,6 +559,189 @@ internal void AssertJsonCondition(params JsonNodeType[] allowedNodeTypes)
...
+        {
+            object value = await this.JsonReader.ReadPrimitiveValueAsync()
+                .ConfigureAwait(false);
+
+            ODataJsonLightReaderUtils.ValidateAnnotationValue(value, annotationName);
+
+            if ((value is string) ^ this.JsonReader.IsIeee754Compatible)
+            {
+                throw new ODataException(Strings.ODataJsonReaderUtils_ConflictBetweenInputFormatAndParameter(Metadata.EdmConstants.EdmInt64TypeName));
+            }
```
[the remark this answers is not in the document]
gathogojr: @habbes The following decision should help summarize what outcome the code block is intended to achieve

| value is string | IsIeee754Compatible | (value is string) ^ IsIeee754Compatible | Outcome |
|-----------------|---------------------|-----------------------------------------|---------|
| true | true | false | No exception thrown - No conflict: `value` is a string and `IsIeee754Compatible` setting is enabled |
| true | false | true | Exception thrown - Conflict: `value` is a string yet `IsIeee754Compatible` setting is not enabled |
| false | true | true | Exception thrown - Conflict: `value` is not a string yet `IsIeee754Compatible` setting is enabled |
| false | false | false | No exception thrown - No conflict: `value` is not a string and `IsIeee754Compatible` setting is not enabled |

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
gathogojr: @marabooy Fixed by introducing constants comprising of the concatenated string

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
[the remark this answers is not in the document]
habbes: Just made a quick check, it doesn't seem like one is a "syntactic sugar" of the other (by that I assume you meant that one is wrapper of the other?). But they both seem to do the same thing and call the same helper method. `string.Compare(s1, s2, StringComparison.Ordinal)` has a few extra checks to validate the comparison type and a switch statement for the different comparison types.
- string.Compare(..., StringComparion.Ordinal): https://source.dot.net/#System.Private.CoreLib/String.Comparison.cs,248
- string.CompareOrdinal(...): https://source.dot.net/#System.Private.CoreLib/String.Comparison.cs,431

Don't know what the worth of this information is, was just curious :)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -327,6 +322,109 @@ internal object ReadODataOrCustomInstanceAnnotationValue(string annotationName,
...
+        {
+            Debug.Assert(propertyAndAnnotationCollector != null, "propertyAndAnnotationCollector != null");
+            Debug.Assert(!string.IsNullOrEmpty(name), "!string.IsNullOrEmpty(name)");
+
+            object propertyAnnotation;
+            string odataType = null;
+            if (propertyAndAnnotationCollector.GetODataPropertyAnnotations(name)
+                .TryGetValue(ODataAnnotationNames.ODataType, out propertyAnnotation))
+            {
+                odataType = ReaderUtils.AddEdmPrefixOfTypeName(ReaderUtils.RemovePrefixOfTypeName((string)propertyAnnotation));
```
habbes: Seems like an opportunity to revisit in the future to cut down string allocations.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -422,7 +520,7 @@ protected ODataJsonLightReaderNestedResourceInfo InnerReadUndeclaredProperty(IOD
 
                 // Complex property or collection of complex property.
                 bool isCollection = payloadTypeReference.IsCollection();
-                ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, isCollection, propertyName, payloadTypeReference);
+                ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader.NodeType, () => this.JsonReader.Value, isCollection, propertyName, payloadTypeReference);
```
habbes: Could you explain the reasoning behind these changes? Why is it better to pass the node type and value delegate separately instead of passing the `JsonReader` itself?

Also, note that `() => this.JsonReader.Value` will cause a new Func to be allocated  each time the method  is called as [demonstrated here](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLRIMYHsEBNXIwJwYzIA+AAgEwCMAsAFCUDMABDWwMJsDeTbQR3YAlaLggI2AWQCeYqBIQBuAULWDWbYFiwAbNgtwAKAJR8NQwVmAArCKTYA3KHrgQ2AXjYwAFgEtkADoANVd/XFgIYzMvAD4fAOC5I0lQ13dTVUYrK0oAdmcMjwBCbwA7OD09bKsAX0tLLRt7RzC9CKjjSgBWAB4WhxgElzcIABEIPQgAcyjzfhzczULR90npuZhorMsGxn2mLU5UqUW89kG24r42GYgYZTZkB6f9uqA=)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,154 +1163,1458 @@ protected EdmTypeKind GetNonEntityValueKind()
...
+            }
+            else
+            {
+                string payloadTypeName = null;
+                Tuple<bool, string> readingResourcePropertyResult = await this.ReadingResourcePropertyAsync(
+                    propertyAndAnnotationCollector,
+                    expectedPropertyTypeReference).ConfigureAwait(false);
+                payloadTypeName = readingResourcePropertyResult.Item2;
+
+                if (readingResourcePropertyResult.Item1)
```
habbes: Maybe you could store the `.Item1` value in a descriptive variable and execute the if statement against the variable like you have done other places for better readability.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,154 +1163,1458 @@ protected EdmTypeKind GetNonEntityValueKind()
...
+                    // Read through all top-level properties, ignore the ones with reserved names (i.e., reserved
+                    // characters in their name) and throw if we find none or more than one properties without reserved name.
+                    while (this.JsonReader.NodeType == JsonNodeType.Property)
+                    {
+                        await this.ProcessPropertyAsync(
+                            propertyAndAnnotationCollector,
+                            readTopLevelPropertyAnnotationValueDelegate,
+                            async (propertyParsingResult, propertyName) =>
+                            {
+                                if (this.JsonReader.NodeType == JsonNodeType.Property)
```
habbes: Opportunity to cut down lambda allocations in the future by passing the `JsonReader` as an argument to the lambda.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1599,16 +3033,17 @@ private object ReadEnumValue(bool insideJsonObjectValue, IEdmEnumTypeReference e
             List<ODataProperty> properties = new List<ODataProperty>();
             while (this.JsonReader.NodeType == JsonNodeType.Property)
             {
-                this.ReadPropertyCustomAnnotationValue = this.ReadCustomInstanceAnnotationValue;
-                this.ProcessProperty(
+                this.ReadPropertyCustomAnnotationValueAsync = this.ReadCustomInstanceAnnotationValueAsync;
```
habbes: Maybe this is not an issue, but does this value get modified inside the while loop?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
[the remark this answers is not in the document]
g2mula: This was about the use of `Compare` over `Equals`  where one only uses Compare against 0 
[Ref caution](https://docs.microsoft.com/en-us/dotnet/api/system.string.compare?view=netstandard-2.0#:~:text=or%20follow%20it.-,Caution,determine%20whether%20two%20strings%20are%20equivalent%2C%20call%20the%20Equals%20method.,-One%20or%20both)
> The [Compare(String, Int32, String, Int32, Int32, CultureInfo, CompareOptions)](https://docs.microsoft.com/en-us/dotnet/api/system.string.compare?view=netstandard-2.0#system-string-compare(system-string-system-int32-system-string-system-int32-system-int32-system-globalization-cultureinfo-system-globalization-compareoptions)) method is designed primarily for use in sorting or alphabetizing operations. It should not be used when the primary purpose of the method call is to determine whether two substrings are equivalent (that is, when the purpose of the method call is to test for a return value of zero). To determine whether two strings are equivalent, call the [Equals](https://docs.microsoft.com/en-us/dotnet/api/system.string.equals?view=netstandard-2.0) method.

src/Microsoft.OData.Core/JsonLight/JsonLightConstants.cs
```diff
@@ -116,5 +116,20 @@ internal static class JsonLightConstants
...
+        /// <summary>The prefixed OData Type property name.</summary>
+        internal const string PrefixedODataTypePropertyName = "@odata.type";
+
+        /// <summary>The prefixed OData property name.</summary>
+        internal const string PrefixedODataIdPropertyName = "@odata.id";
+
+        /// <summary>The prefixed OData property name.</summary>
+        internal const string PrefixedODataNullPropertyName = "@odata.null";
+
+        /// <summary>The prefixed OData property name.</summary>
```
KenitoInc, line 132:
```suggestion
        /// <summary>The prefixed OData removed property name.</summary>
```

KenitoInc (approved):
Minor comments.
Otherwise LGTM

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -1333,13 +3463,14 @@ private ODataJsonLightReaderNestedInfo TryReadAsStream(IODataJsonLightReaderReso
                     isCollection &&
                     propertyType.Definition.AsElementType().IsUntyped()) ||
                 (propertyType == null
-                    && (isCollection || this.JsonReader.CanStream())
+                    && (isCollection || await this.JsonReader.CanStreamAsync().ConfigureAwait(false))
                     && readAsStream != null
                     && readAsStream(null, isCollection, propertyName, property)))
```
habbes: Should this delegate also be synchronous in the async function (I'm not sure what it does, but it starts with `read` so just wanted to double-check that it doesn't any I/O operation that should be async).

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -1552,7 +3685,12 @@ private ODataJsonLightReaderNestedInfo InnerReadUndeclaredProperty(IODataJsonLig
             if (payloadTypeOrItemType != null)
             {
                 // Complex property or collection of complex property.
-                ValidateExpandedNestedResourceInfoPropertyValue(this.JsonReader, isCollection, propertyName, payloadTypeReference);
+                await ValidateExpandedNestedResourceInfoPropertyValueAsync(
+                    this.JsonReader.NodeType,
+                    () => this.JsonReader.GetValueAsync(),
```
habbes: Similar questions as before regarding the use of a delegate here instead of just passing the json reader directly.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
gathogojr: @habbes No. The delegate doesn't ready from the stream. It's a `Func` property declared in `ODataMessageReaderSettings` class that someone can use to direct the reader to read particular properties as a stream.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -1033,164 +1032,2265 @@ internal ODataJsonLightReaderNestedInfo ReadPropertyWithoutValue(IODataJsonLight
...
+                    this.ReadEntryPropertyAnnotationValueAsync,
+                    async (propertyParsingResult, propertyName) =>
+                    {
+                        switch (propertyParsingResult)
+                        {
+                            case PropertyParsingResult.ODataInstanceAnnotation:
+                            case PropertyParsingResult.CustomInstanceAnnotation:
+                                await this.ReadOverPropertyNameAsync()
+                                    .ConfigureAwait(false);
+                                object value = await this.ReadODataOrCustomInstanceAnnotationValueAsync(resourceState, propertyParsingResult, propertyName)
```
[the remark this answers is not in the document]
gathogojr: If you look at Line 1478, the statement does not end with a semi colon. `.ConfigureAwait(false)` is on the line that follows

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
gathogojr: It doesn't get modified within the while loop. While I could consider moving the line outside the loop, I'm a little antsy since this logic is adapted from the sync equivalent and maybe the idea behind stashing it within the while loop was to ensure it's not assigned if there are no properties. That doesn't seem like a very compelling reason though. We'll discuss it and decide which way to go

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,154 +1163,1458 @@ protected EdmTypeKind GetNonEntityValueKind()
...
+        private object ReadNonEntityValueImplementation(
+            string payloadTypeName,
+            IEdmTypeReference expectedTypeReference,
+            PropertyAndAnnotationCollector propertyAndAnnotationCollector,
+            CollectionWithoutExpectedTypeValidator collectionValidator,
+            bool validateNullValue,
+            bool isTopLevelPropertyValue,
+            bool insideResourceValue,
+            string propertyName,
+            bool? isDynamicProperty = null)
```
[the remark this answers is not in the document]
gathogojr: This parameter is rarely passed. Having it as a nullable bool allows for the default to be `null` (as opposed to `false`) thus giving room for the caller to be explicit.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,154 +1163,1458 @@ protected EdmTypeKind GetNonEntityValueKind()
...
+            // the responses written by 6.x version of this library.
+            if (this.IsTopLevel6xNullValue())
+            {
+                // NOTE: when reading a null value we will never ask the type resolver (if present) to resolve the
+                //       type; we always fall back to the expected type.
+                this.ReaderValidator.ValidateNullValue(
+                    expectedPropertyTypeReference,
+                    /*validateNullValue*/ true,
+                    /*propertyName*/ null,
+                    null);
```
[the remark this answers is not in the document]
gathogojr: This parameter is rarely passed. Having it as a nullable bool allows for the default to be `null` (as opposed to `false`) thus giving room for the caller to be explicit.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -923,6 +1128,403 @@ private void ReadODataOrCustomInstanceAnnotationValue(string annotatedPropertyNa
...
+        /// 3). The first component contains ODataInstanceAnnotation if an odata instance annotation was found, while the second component contains the name of the annotation.
+        ///                             The reader is positioned on the value of the annotation.
+        /// 4). The first component contains CustomInstanceAnnotation if a custom instance annotation was found, while the second component contains name of the annotation.
+        ///                             The reader is positioned on the value of the annotation.
+        /// 5). The first component contains MetadataReferenceProperty if a property which is a reference into the metadata was found, while the second component contains the name of the property.
+        ///                             The reader is positioned on the value of the property.
+        /// 6). The first component contains EndOfObject if end of the object scope was reached and no properties are to be reported, while the second component contains null.
+        ///                             This can only happen if there's a property annotation which is ignored (for example custom one) at the end of the object.
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -923,6 +1128,403 @@ private void ReadODataOrCustomInstanceAnnotationValue(string annotatedPropertyNa
...
+                {
+                    // Read over the property name.
+                    await this.JsonReader.ReadAsync()
+                        .ConfigureAwait(false);
+                    parsedPropertyName = propertyNameFromReader;
+                    return Tuple.Create(PropertyParsingResult.NestedDeltaResourceSet, parsedPropertyName);
+                }
+
+                bool isInstanceAnnotation = false;
+                if (!isPropertyAnnotation)
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,154 +1163,1458 @@ protected EdmTypeKind GetNonEntityValueKind()
...
 
         /// <summary>
-        /// Creates an instance of an ODataResourceSet that represents a nested collection
+        /// Try to asynchronously read or peek the odata.type annotation.
         /// </summary>
-        /// <param name="resourceState">The current resource state</param>
-        /// <param name="propertyName">The name of the collection property being read</param>
-        /// <returns>An ODataResourceSet with properties set from any instance annotations</returns>
-        private static ODataResourceSet CreateCollectionResourceSet(IODataJsonLightReaderResourceState resourceState, string propertyName)
+        /// <param name="propertyAndAnnotationCollector">The current level's PropertyAndAnnotationCollector.</param>
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,154 +1163,1458 @@ protected EdmTypeKind GetNonEntityValueKind()
...
-            Debug.Assert(propertyAndAnnotationCollector != null, "propertyAndAnnotationCollector != null");
-
-            expectedPropertyTypeReference = this.UpdateExpectedTypeBasedOnContextUri(expectedPropertyTypeReference);
-
-            object propertyValue = missingPropertyValue;
-            var customInstanceAnnotations = new Collection<ODataInstanceAnnotation>();
+            object propertyValue = null;
+            payloadTypeReference = ResolveUntypedType(
+                this.JsonReader.NodeType,
+                this.JsonReader.Value,
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/JsonLightConstants.cs
```diff
@@ -116,5 +116,20 @@ internal static class JsonLightConstants
...
         /// <summary>The simplified Removed property name.</summary>
         internal const string SimplifiedODataRemovedPropertyName = "@removed";
+
+        /// <summary>The prefixed OData property name.</summary>
+        internal const string PrefixedODataContextPropertyName = "@odata.context";
+
+        /// <summary>The prefixed OData Type property name.</summary>
+        internal const string PrefixedODataTypePropertyName = "@odata.type";
+
+        /// <summary>The prefixed OData property name.</summary>
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/JsonLightConstants.cs
```diff
@@ -116,5 +116,20 @@ internal static class JsonLightConstants
...
+        internal const string PrefixedODataIdPropertyName = "@odata.id";
+
+        /// <summary>The prefixed OData property name.</summary>
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/JsonLightConstants.cs
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -1033,164 +1032,2265 @@ internal ODataJsonLightReaderNestedInfo ReadPropertyWithoutValue(IODataJsonLight
...
+        {
+            Debug.Assert(resourceState != null, "resourceState != null");
+            Debug.Assert(resourceState.ResourceType != null && this.Model.IsUserModel(), "A non-null resource type and non-null model are required.");
+            Debug.Assert(
+                this.JsonReader.NodeType == JsonNodeType.Property || this.JsonReader.NodeType == JsonNodeType.EndObject,
+                "Pre-Condition: JsonNodeType.Property or JsonNodeType.EndObject");
+            this.JsonReader.AssertNotBuffering();
+
+            ODataJsonLightReaderNestedInfo readerNestedResourceInfo = null;
+            Debug.Assert(resourceState.ResourceType != null, "In JSON we must always have an structured type when reading resource.");
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -1033,164 +1032,2265 @@ internal ODataJsonLightReaderNestedInfo ReadPropertyWithoutValue(IODataJsonLight
...
+        /// </returns>
+        /// <remarks>
+        /// Pre-Condition:  JsonNodeType.Property          The first property after the odata.context in the resource object.
+        ///                 JsonNodeType.EndObject         End of the resource object.
+        /// Post-Condition: JsonNodeType.Property          The property after the odata.type (if there was any), or the property on which the method was called.
+        ///                 JsonNodeType.EndObject         End of the resource object.
+        ///
+        /// This method Creates an ODataDeltaDeletedEntry and fills in the Id and Reason properties, if specified in the payload.
+        /// </remarks>
+        internal async Task<ODataDeletedResource> ReadDeletedResourceAsync()
```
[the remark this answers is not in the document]
gathogojr: @habbes I'll have an offline discussion with you on this. The number of files changed is substantial already. I'd like to assess the blast radius first

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,154 +1163,1458 @@ protected EdmTypeKind GetNonEntityValueKind()
...
+                }
+            }
+
+            expectedTypeReference = expectedTypeReference != null && expectedTypeReference.IsUntyped() ? null : expectedTypeReference;
+            ODataTypeAnnotation typeAnnotation;
+            EdmTypeKind targetTypeKind;
+            IEdmTypeReference targetTypeReference = this.ReaderValidator.ResolvePayloadTypeNameAndComputeTargetType(
+                EdmTypeKind.None,
+                /*expectStructuredType*/ null,
+                /*defaultPrimitivePayloadType*/ null,
```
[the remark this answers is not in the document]
gathogojr: The diff was messed up so here you were referring to code in a sync method. I'd prefer to leave existing code in sync methods unchanged

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
gathogojr: Fixed by removing the delegate

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,154 +1163,1458 @@ protected EdmTypeKind GetNonEntityValueKind()
...
+            // An expanded link with null resource requires a primitive null node here;
+            if (jsonReaderNodeType == JsonNodeType.StartArray)
+            {
+                if (isCollection == false)
+                {
+                    throw new ODataException(ODataErrorStrings.ODataJsonLightResourceDeserializer_CannotReadSingletonNestedResource(jsonReaderNodeType, propertyName));
+                }
+            }
+            else if (
+                (jsonReaderNodeType == JsonNodeType.PrimitiveValue && (await readNodeValueDelegate().ConfigureAwait(false)) == null)
```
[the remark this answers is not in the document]
gathogojr: Fixed by removing the delegate

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,154 +1163,1458 @@ protected EdmTypeKind GetNonEntityValueKind()
...
-                    null);
+                ODataJsonLightReaderNestedResourceInfo readerNestedResourceInfo = null;
 
-                // We don't allow properties or non-custom annotations in the null payload.
-                this.ValidateNoPropertyInNullPayload(propertyAndAnnotationCollector);
+                // Complex property or collection of complex property.
+                bool isCollection = payloadTypeReference.IsCollection();
+                await ValidateExpandedNestedResourceInfoPropertyValueAsync(
+                    this.JsonReader.NodeType,
+                    () => this.JsonReader.GetValueAsync(),
```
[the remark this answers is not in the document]
gathogojr: Fixed by removing the delegate

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,154 +1163,1458 @@ protected EdmTypeKind GetNonEntityValueKind()
...
-                || this.CompareSimplifiedODataAnnotation(JsonLightConstants.SimplifiedODataTypePropertyName, propertyName))
+            PropertyAndAnnotationCollector propertyAndAnnotationCollector = resourceState.PropertyAndAnnotationCollector;
+            bool insideResourceValue = false;
+            string outerPayloadTypeName = ValidateDataPropertyTypeNameAnnotation(propertyAndAnnotationCollector, propertyName);
+            string payloadTypeName = await this.TryReadOrPeekPayloadTypeAsync(propertyAndAnnotationCollector, propertyName, insideResourceValue)
+                .ConfigureAwait(false);
+            EdmTypeKind payloadTypeKind;
+            IEdmType payloadType = ReaderValidationUtils.ResolvePayloadTypeName(
+                this.Model,
+                null, // expectedTypeReference
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
gathogojr: Fixed by removing the delegate

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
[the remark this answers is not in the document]
habbes: Is the behaviour different between when passing `null` compared to when passing `false`? If the behaviour is not different, both semantically and practically, why not default to `false`?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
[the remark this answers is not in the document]
habbes: Ah good question.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
gathogojr: For now removed the use of ValueTask with the plan being to going in after and applying that where actual value can be derived

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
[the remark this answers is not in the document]
gathogojr: Fixed by replacing string CompareOrdinal with Equals

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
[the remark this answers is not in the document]
gathogojr: Not a 100% sure about that but I think its safer to follow similar logic as the sync equivalent for now

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -177,6 +177,15 @@ internal ODataJsonLightInputContext JsonLightInputContext
             set;
         }
 
+        /// <summary>
+        /// Delegate called to read property custom annotation value asynchronously.
+        /// </summary>
+        protected Func<PropertyAndAnnotationCollector, string, Task<object>> ReadPropertyCustomAnnotationValueAsync
```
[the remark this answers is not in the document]
gathogojr: Addressed by adding a `<Remarks>` to the `Func` property documentation what the delegate takes and what it returns

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
gathogojr: To be handled in a separate PR for fixing issues surrounding the extra allocations occasioned by capturing of variables

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,154 +1163,1458 @@ protected EdmTypeKind GetNonEntityValueKind()
...
+            {
+                object itemValue = this.ReadNonEntityValueImplementation(
+                    /*payloadTypeName*/ null,
+                    itemType,
+                    propertyAndAnnotationCollector,
+                    collectionValidator,
+                    /*validateNullValue*/ true,
+                    /*isTopLevelPropertyValue*/ false,
+                    /*insideResourceValue*/ false,
+                    /*propertyName*/ null);
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1059,154 +1163,1458 @@ protected EdmTypeKind GetNonEntityValueKind()
...
+                                            // Now read the property value
+                                            propertyValue = this.ReadNonEntityValue(
+                                                    payloadTypeName,
+                                                    expectedPropertyTypeReference,
+                                                    /*propertyAndAnnotationCollector*/ null,
+                                                    /*collectionValidator*/ null,
+                                                    /*validateNullValue*/ true,
+                                                    /*isTopLevelPropertyValue*/ true,
+                                                    /*insideResourceValue*/ false,
+                                                    /*propertyName*/ propertyName);
```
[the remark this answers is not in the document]
gathogojr: Fixed

PR #2279, opened 2021-12-16
Default to using entityType for incomplete serializationInfo.

### LINKED ISSUE #2278: Odatawriter unable to find key property of nested resource entity when key is present
When  ODataWriterCore.WriteStart(ODataNestedResourceInfo nestedResourceInfo) is called with an entity, odata is unable to find the key property and throws error - "Microsoft.OData.ODataException: An ODataResource of type 'NS.S2C' is found without key properties. When writing without a user model, each resource must contain at least one property whose 'ODataProperty.SerializationInfo.PropertyKind' set to 'ODataPropertyKind.Key'. When writing with a user model, the entity type 'NS.S2C' defined in the model must define at least one key property."

### Assemblies affected

7.9.4, 7.9.1, 7.8.3

### Reproduce steps

Sample schema that reproduces the issue - 
```
 <?xml version="1.0" encoding="utf-8"?><Schema Namespace="NS" xmlns="http://docs.oasis-open.org/odata/ns/edm">
  <EntityType Name="S1C">
    <Key>
      <PropertyRef Name="id" />
    </Key>
    <Property Name="id" Type="Edm.String" Nullable="false" />
    <NavigationProperty Name="S2CList" Type="Collection(NS.S2C)" ContainsTarget="true">
    </NavigationProperty>
  </EntityType>
  <EntityType Name="S2C" >
    <Key>
      <PropertyRef Name="id" />
    </Key>
    <Property Name="id" Type="Edm.String" Nullable="false" />
    <NavigationProperty Name="parentList" Type="NS.S1C" ContainsTarget="true" />
  </EntityType>
  <Action Name="move" IsBound="true" EntitySetPath="bindingParameter">
    <Parameter Name="bindingParameter" Type="NS.S2C" />
    <ReturnType Type="NS.S2C" />
  </Action>
  <EntityContainer Name="GraphService">
    <Singleton Name="theS1C" Type="NS.S1C" />
  </EntityContainer>
</Schema>

```
Sample url called - 
URL -- ` "theS1C/S2CList/id/move"  `

Odata callstack - 
```
 ODataResourceMetadataContext.ValidateEntityTypeHasKeyProperties(KeyValuePair`2[] keyProperties, String actualEntityTypeName)
    ODataResourceMetadataContext.GetKeyProperties(ODataResourceBase resource, ODataResourceSerializationInfo serializationInfo, IEdmEntityType actualEntityType)
    ODataWriterCore.AppendEntitySetKeySegment(ODataPath odataPath, Boolean throwIfFail)
    ODataWriterCore.EnterScope(WriterState newState, ODataItem item)
    ODataWriterCore.WriteStartNestedResourceInfoImplementation(ODataNestedResourceInfo nestedResourceInfo)
    ODataWriterCore.WriteStart(ODataNestedResourceInfo nestedResourceInfo)
```
sample payload - 
```
{
    "parentList": {
        "id": "list-id"
    },
    "@odata.type": "#NS.S2C",
    "id": "task-immutable-id",
}
```

### Expected result

Expected ODL to successfully serialize the given nestedresourceinfo

### Actual result

*What is actually happening.*
ODL throws error - 
`"Microsoft.OData.ODataException: An ODataResource of type 'NS.S2C' is found without key properties. When writing without a user model, each resource must contain at least one property whose 'ODataProperty.SerializationInfo.PropertyKind' set to 'ODataPropertyKind.Key'. When writing with a user model, the entity type 'NS.S2C' defined in the model must define at least one key property."`
### Additional detail

mikepizzo:
### Issues

*This pull request fixes #2278*

### Description

ODL expects that, if you set SerializationInfo on the resource then you are taking responsibility for providing all of the information necessary for serialization.  

When ODL tries to get key properties in order to append the key segment, it checks to see if the serializationInfo for the resource is null -- if so, it uses the model to determine key properties and all is good. On the other hand, if serializationInfo on the resource is non-null then it looks for properties whose serializationInfo identify them as key properties.  Finding none, it errors.

This change makes ODL more robust by falling back to using the entityType to identify key properties if the resource has serializationInfo but none of the properties have serializationInfo that identifies them as key properties.

- [x] *Test cases added*

### Additional work necessary

*None*

### DIFF
--- a/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
@@ -174,9 +174,10 @@ namespace Microsoft.OData.Evaluation
             IEdmEntityType actualEntityType,
             bool requiresId)
         {
+            Debug.Assert(resource != null, "GetKeyProperties called for a null resource.");
 
             KeyValuePair<string, object>[] keyProperties = null;
-            string actualEntityTypeName = resource.TypeName ?? actualEntityType?.FullName();
+            string actualEntityTypeName = string.IsNullOrEmpty(resource.TypeName) ? actualEntityType?.FullName() : resource.TypeName;
 
             // if we have serializationInfo, try that first
             if (serializationInfo != null)
@@ -288,7 +289,8 @@ namespace Microsoft.OData.Evaluation
 
             for (int keyProperty = 0; keyProperty < keyProperties.Length; keyProperty++)
             {
-                if (keyProperties[keyProperty].Value == null || (keyProperties[keyProperty].Value is ODataValue && !(keyProperties[keyProperty].Value is ODataEnumValue)))
+                object keyValue = keyProperties[keyProperty].Value;
+                if (keyValue == null || (keyValue is ODataValue && !(keyValue is ODataEnumValue)))
                 {
                     if (throwOnError)
                     {
--- a/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
@@ -174,45 +174,37 @@ namespace Microsoft.OData.Evaluation
             IEdmEntityType actualEntityType,
             bool requiresId)
         {
+
             KeyValuePair<string, object>[] keyProperties = null;
-            string actualEntityTypeName = resource.TypeName ?? actualEntityType.FullName();
+            string actualEntityTypeName = resource.TypeName ?? actualEntityType?.FullName();
 
             // if we have serializationInfo, try that first
             if (serializationInfo != null)
             {
-                if (String.IsNullOrEmpty(actualEntityTypeName))
-                {
-                    throw new ODataException(Strings.ODataResourceTypeContext_ODataResourceTypeNameMissing);
-                }
-
                 keyProperties = ODataResourceMetadataContextWithoutModel.GetPropertiesBySerializationInfoPropertyKind(resource, ODataPropertyKind.Key, actualEntityTypeName);
             }
 
             // if we didn't get any keys from serializationInfo, try using entity type
             if ((keyProperties == null || keyProperties.Length == 0) && actualEntityType != null)
             {
-                keyProperties = GetPropertyValues(actualEntityType.Key(), resource, actualEntityType, /*isKeyProperty*/ true, requiresId).ToArray();
+                keyProperties = GetPropertyValues(actualEntityType.Key(), resource, actualEntityType, requiresId).ToArray();
             }
 
-            if (requiresId)
-            {
-                ValidateEntityTypeHasKeyProperties(keyProperties, actualEntityTypeName);
-            }
-            else if (keyProperties == null)
-            {
-                keyProperties = Enumerable.Empty<KeyValuePair<string, object>>().ToArray();
+            if (!ValidateEntityTypeHasKeyProperties(keyProperties, actualEntityTypeName, requiresId))
+            { 
+                return Enumerable.Empty<KeyValuePair<string, object>>().ToArray();
             }
 
             return keyProperties;
         }
 
-        private static IEnumerable<KeyValuePair<string, object>> GetPropertyValues(IEnumerable<IEdmStructuralProperty> properties, ODataResourceBase resource, IEdmEntityType actualEntityType, bool isKeyProperty, bool isRequired)
+        private static IEnumerable<KeyValuePair<string, object>> GetPropertyValues(IEnumerable<IEdmStructuralProperty> properties, ODataResourceBase resource, IEdmEntityType actualEntityType, bool isRequired)
         {
             string actualEntityTypeName = actualEntityType.FullName();
             object primitiveValue;
             foreach (IEdmStructuralProperty property in properties)
             {
-                if (TryGetPrimitiveOrEnumPropertyValue(resource, property.Name, actualEntityTypeName, isKeyProperty, isRequired, out primitiveValue))
+                if (TryGetPrimitiveOrEnumPropertyValue(resource, property.Name, actualEntityTypeName, isRequired, out primitiveValue))
                 {
                     yield return new KeyValuePair<string, object>(property.Name, primitiveValue);
                 }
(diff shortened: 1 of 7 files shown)


xuzhg (approved):

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -166,41 +166,45 @@ public IODataResourceTypeContext TypeContext
...
             ODataResourceBase resource,
             ODataResourceSerializationInfo serializationInfo,
-            IEdmEntityType actualEntityType)
+            IEdmEntityType actualEntityType,
+            bool requiresId)
         {
+
             KeyValuePair<string, object>[] keyProperties = null;
-            string actualEntityTypeName = null;
+            string actualEntityTypeName = resource.TypeName ?? actualEntityType?.FullName();
```
corranrogue9, line 179: Do you know how important the old check for empty string is? We are missing that now, maybe this should be `string.IsNullOrEmpty(resource.TypeName) ? actualEntityType?.FullName() : resource.TypeName`

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -267,13 +270,38 @@ private static object GetPrimitiveOrEnumPropertyValue(string entityTypeName, ODa
...
+                }
+                else
+                {
+                    return false;
+                }
+            }
+
+            for (int keyProperty = 0; keyProperty < keyProperties.Length; keyProperty++)
+            {
+                if (keyProperties[keyProperty].Value == null || (keyProperties[keyProperty].Value is ODataValue && !(keyProperties[keyProperty].Value is ODataEnumValue)))
```
corranrogue9, line 291: It appears that the compiler is not smart enough to optimize the multiple indexing operations (probably because nothing guarantees that there are no side effects):

[Multiple indexes](https://sharplab.io/#v2:C4LglgNgPgAgTARgLACgYGYAE9MGFMDeqmJ2WMALJgLIAUAysAE5gB2A5gNoC6mAhk3YBnAJSFipSQDMA9k0y02wTGEwBeTAAYA3CswAefoKEA6ADIBTDsAAWugNT2wIiZJJEUbryqkKBwzjBeAEINVgBXCAgXT29JDzi4mAQATlp/IUDuEW1XRIBfPMlC2JIS/KA===)
[Storing the indexed value](https://sharplab.io/#v2:C4LglgNgPgAgTARgLACgYGYAE9MGFMDeqmJ2WMALJgLIAUAysAE5gB2A5gNoC6mAhk3YBnAJSFipSQDMA9k0y02wTGEwBeTAAYA3CswAefoKEA6ADIBTDsAAWugNT2wIiZJJEUbr5gBuAo+zqAUKcYNzart4kYFIKAoEAhBqsAK4QEC6eUaQe2d4wCACctPGiEVl5AL6RktUVmHWVQA=)

In the first example, the IL indicates that the indexing is performed a second time for the `WriteLine` call. The JITer might be able to optimize this, but there's really no way to know for sure. It would probably make the most sense to store `keyProperties[keyProperty]` in a variable scoped to the loop so that we only index the one time, instead of 4 times.

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -483,16 +508,12 @@ internal ODataResourceMetadataContextWithModel(ODataResourceBase resource, IODat
...
-                                ValidateEntityTypeHasKeyProperties(this.keyProperties, this.ActualResourceTypeName);
-                            }
+                            this.keyProperties = GetPropertyValues(entityType.Key(), resource, entityType, this.requiresId).ToArray();
                         }
-                        else
+
+                        if (!ValidateEntityTypeHasKeyProperties(this.keyProperties, this.ActualResourceTypeName, this.requiresId))
                         {
-                            this.keyProperties = Enumerable.Empty<KeyValuePair<string, object>>().ToArray();
+                            return Enumerable.Empty<KeyValuePair<string, object>>().ToArray();
```
corranrogue9, line 516: Probably not worth it as part of this change, but this will allocate a new array every time we return, even though all 0-length arrays are the same (we actually get warnings from the compiler for this). We could avoid unnecessary allocations and garbage collection with this small class:

```
public static class Array
{
  public static T[] Empty<T>()
  {
    return Internal<T>.Empty;
  }
        
  private static class Internal<T>
  {
    public static T[] Empty = new T[0];
  }
}
```

New versions of .NET already have these singletons, but for back-compat we could write this one ourselves.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2426,20 +2426,18 @@ private void StartNestedStreamInfo(ODataJsonLightReaderStreamInfo readerStreamIn
         /// <returns>True if successfully append key segment.</returns>
         private bool TryAppendEntitySetKeySegment(ref ODataPath odataPath)
         {
-            try
+            if (EdmExtensionMethods.HasKey(this.CurrentScope.NavigationSource, this.CurrentScope.ResourceType as IEdmStructuredType))
             {
-                if (EdmExtensionMethods.HasKey(this.CurrentScope.NavigationSource, this.CurrentScope.ResourceType as IEdmStructuredType))
+                IEdmEntityType currentEntityType = this.CurrentScope.ResourceType as IEdmEntityType;
+                ODataResourceBase resource = this.CurrentScope.Item as ODataResourceBase;
+                KeyValuePair<string, object>[] keys = ODataResourceMetadataContext.GetKeyProperties(resource, null, currentEntityType, false);
```
corranrogue9, line 2433: `GetKeyProperties` seems to make the assumption that `resource` isn't `null`. I know that this isn't really related to your change, but not validating that `currentEntityType` is actually a `IEdmEntityType` means we could get a `NullReferenceException` here.

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
```diff
@@ -512,6 +512,146 @@ static AutoComputePayloadMetadataInJsonIntegrationTests()
...
+                ContainsTarget = true
+            });
+
+            var container = new EdmEntityContainer("NS", "Container");
+            var entitySet = container.AddEntitySet("EntitySet", entityType);
+
+            model.AddElements(new IEdmSchemaElement[] { entityType, nestedEntityType, container });
+
+            // setup writer
+            var stream = new MemoryStream();
```
corranrogue9, line 554: `MemoryStream` implements `IDisposable`. Please wrap it in a `using` block

corranrogue9 (approved):

xuzhg (approved):

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
mikepizzo: It is unlikely this will be an empty string, but I like the robustness of your suggestion and will adopt it.

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
mikepizzo: Good catch -- thanks!

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
mikepizzo: Yes; I tried to use Array.Empty<T>(), but discovered it wasn't supported in .net45, so left it as it was. I agree we should go through and optimize everywhere we return empty arrays.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
mikepizzo: This is our code and should never be called with a null resource. I added an assert to GetKeyProperties.

PR #2299, opened 2022-01-20
fix filter using in operator with open types

### LINKED ISSUE #2270: ArgumentNullException for $filter IN operator on dynamic property
Use of the `in` operator in a `$filter` on a **dynamic property** results in an unhandled ODL exception (`ArgumentNullException`).

### Assemblies affected

All up to 7.9.4.

### Reproduce steps

``` cs
// Setup basic model
var model = new EdmModel();
var user = model.AddEntityType("graph", "user", baseType: null, isAbstract: false, isOpen: true, hasStream: false);
var container = model.AddEntityContainer("graph", "service");
container.AddEntitySet("users", user);

// Setup URI parser
var parser = new ODataUriParser(model, new System.Uri("/users?$filter=openPropertyFoo in ('bar')", System.UriKind.Relative));
parser.ParsePath();

// This throws an ArgumentNullException:
var filter = parser.ParseFilter();
```

### Expected result

Either:
 - a valid FilterClause.
Or:
 - ODL Exception calling out that $filter on dynamic properties is not supported

### Actual result

Unhandled exception thrown by ODL:

```
System.ArgumentNullException: Value cannot be null. (Parameter 'elementType');
   at Microsoft.OData.Edm.EdmUtil.CheckArgumentNull[T](T value, String parameterName)
   at Microsoft.OData.Edm.EdmCollectionType..ctor(IEdmTypeReference elementType)
   at Microsoft.OData.UriParser.InBinder.BindInOperator(InToken inToken, BindingState state)
   at Microsoft.OData.UriParser.MetadataBinder.Bind(QueryToken token)
   at Microsoft.OData.UriParser.FilterBinder.BindFilter(QueryToken filter)
   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseFilterImplementation(String filter, ODataUriParserConfiguration configuration, ODataPathInfo odataPathInfo)
   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseFilter()
   at Microsoft.OData.UriParser.ODataUriParser.ParseFilter()   
```

ElizabethOkerio:
### Issues

*This pull request fixes #2270.*

### Description

For open types, the <code>TypeReference</code> of the left Node in an InBind Operation is always null. Have therefore put a null check to check for the left Node's TypeReference.  If it is null, then an <code> UnTyped Typerefence </code> is used. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
+++ b/src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
@@ -625,6 +625,7 @@ namespace Microsoft.OData
         {
             ODataMessageReaderSettings settings = new ODataMessageReaderSettings();
             settings.Validations &= ~ValidationKinds.ThrowOnUndeclaredPropertyForNonOpenType;
+            settings.ReadUntypedAsString = false;
 
             using (StringReader reader = new StringReader(value))
             {
--- a/src/Microsoft.OData.Core/UriParser/SemanticAst/InNode.cs
+++ b/src/Microsoft.OData.Core/UriParser/SemanticAst/InNode.cs
@@ -50,11 +50,14 @@ namespace Microsoft.OData.UriParser
             this.left = left;
             this.right = right;
 
-            if (!this.left.GetEdmTypeReference().IsAssignableFrom(this.right.ItemType) &&
-                !this.right.ItemType.IsAssignableFrom(this.left.GetEdmTypeReference()))
+            if (!right.GetEdmTypeReference().IsUntyped())
             {
-                throw new ArgumentException(ODataErrorStrings.Nodes_InNode_CollectionItemTypeMustBeSameAsSingleItemType(
-                    this.right.ItemType.FullName(), this.left.GetEdmTypeReference().FullName()));
+                if (!this.left.GetEdmTypeReference().IsAssignableFrom(this.right.ItemType) &&
+                    !this.right.ItemType.IsAssignableFrom(this.left.GetEdmTypeReference()))
+                {
+                    throw new ArgumentException(ODataErrorStrings.Nodes_InNode_CollectionItemTypeMustBeSameAsSingleItemType(
+                        this.right.ItemType.FullName(), this.left.GetEdmTypeReference().FullName()));
+                }
             }
         }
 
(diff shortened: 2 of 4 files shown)


xuzhg (approved):

src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
```diff
@@ -625,6 +625,7 @@ private static object ConvertFromResourceOrCollectionValue(string value, IEdmMod
         {
             ODataMessageReaderSettings settings = new ODataMessageReaderSettings();
             settings.Validations &= ~ValidationKinds.ThrowOnUndeclaredPropertyForNonOpenType;
+            settings.ReadUntypedAsString = false;
```
xuzhg, line 628: We should retrieve it from the message reader setting. Should we? or add a configuration in the Uri query parsing?

KenitoInc: +1

src/Microsoft.OData.Core/UriParser/SemanticAst/InNode.cs
```diff
@@ -50,11 +50,14 @@ public InNode(SingleValueNode left, CollectionNode right)
             this.left = left;
             this.right = right;
 
-            if (!this.left.GetEdmTypeReference().IsAssignableFrom(this.right.ItemType) &&
-                !this.right.ItemType.IsAssignableFrom(this.left.GetEdmTypeReference()))
+            if (!right.GetEdmTypeReference().IsUntyped())
```
KenitoInc, line 53: What should happen if right node is untyped?

src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
mikepizzo: This setting controls how we read untyped values -- if our logic is to read untyped values as structured, not as string (which it should be) then I think it's correct to statically set to false.

src/Microsoft.OData.Core/UriParser/SemanticAst/InNode.cs
mikepizzo: if right is untyped, then we don't have to validate the assignability, which is all the enclosed logic does.

mikepizzo (approved):
:shipit:

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -2818,7 +2818,19 @@ public void FilterWithInOperationWithDateTimeOffsetCollectionContainingSingleNul
             Assert.Equal(new object[] {null},
                 Assert.IsType<CollectionConstantNode>(inNode.Right).Collection.Select(x => x.Value));
         }
-#endregion
+
+        [Fact]
+        public void FilterWithInOperationWithOpenTypes()
```
mikepizzo, line 2823: >FilterWithInOperationWithOpenTypes [](http://example.com/codeflow?start=20&length=34)

Do we have any end-to-end tests that shows this actually generates the appropriate LINQ expression?

PR #2300, opened 2022-01-24
Fix cast in expand

KenitoInc:
### Issues

*This pull request fixes https://github.com/OData/AspNetCoreOData/issues/445*

### Description

In #2160 we added support for the `TypeSegment` after `NavigationPropertySegment`. 
Example: `~/Orders?$expand=Customer/Model.VipCustomer`
This means that we will expand only the Customers that are an instance of Model.VipCustomer.
Note: `Model.VipCustomer` is a derived type of the `Customer`.

Something that was not captured in that PR is scenarios where we have query options affecting properties that are only in the derived type but not in the base type.

Examples: 
`~/Orders?$expand=Customer/Model.VipCustomer($filter=VipCustomerID eq '12345')`
`~/Orders?$expand=Customer/Model.VipCustomer($expand=VipCustomerOrders')`
`~/Orders?$expand=Customer/Model.VipCustomer($orderby=VipEmail)`

This PR fixes that by ensuring when we are doing semantic binding, we can bind the properties in the derived type.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
@@ -350,7 +350,16 @@ namespace Microsoft.OData.UriParser
             // If we are Binding the expanded property Trips, currentLevelEntityType should be the type of the VipCustomer
             if (this.parsedSegments.Count > 0 && this.parsedSegments.Last() is TypeSegment)
             {
-                currentLevelEntityType = this.parsedSegments.Last().EdmType as IEdmStructuredType;
+                IEdmType typeSegmentType = this.parsedSegments.Last().EdmType;
+
+                if (typeSegmentType.IsStructuredCollectionType())
+                {
+                    currentLevelEntityType = (typeSegmentType as IEdmCollectionType).ElementType.Definition as IEdmStructuredType;
+                }
+                else
+                {
+                    currentLevelEntityType = typeSegmentType as IEdmStructuredType;
+                }
             }
 
             List<ODataPathSegment> pathSoFar = new List<ODataPathSegment>();
@@ -681,7 +690,16 @@ namespace Microsoft.OData.UriParser
             // If we are Binding the selected property VipCustomerName, currentLevelEntityType should be the type of the VipCustomer
             if (this.parsedSegments.Count > 0 && this.parsedSegments.Last() is TypeSegment)
             {
-                currentLevelType = this.parsedSegments.Last().EdmType as IEdmStructuredType;
+                IEdmType typeSegmentType = this.parsedSegments.Last().EdmType;
+
+                if (typeSegmentType.IsStructuredCollectionType())
+                {
+                    currentLevelType = (typeSegmentType as IEdmCollectionType).ElementType.Definition as IEdmStructuredType;
+                }
+                else
+                {
+                    currentLevelType = typeSegmentType as IEdmStructuredType;
+                }
             }
 
             // first, walk through all type segments in a row, converting them from tokens into segments.
--- a/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
@@ -346,8 +346,8 @@ namespace Microsoft.OData.UriParser
 
             // If we have a previous level binding with the last segment as a Type segment,
             // The currentLevelEntityType should be the Type of the TypeSegment
-            // E.g Books?$expand=Authors/Ns.PrimaryAuthor($expand=Awards)
-            // If we are Binding the expanded property Awards, currentLevelEntityType should be the type of the PrimaryAuthor
+            // E.g /Orders?$expand=Customer/NS.VipCustomer($expand=Trips)
+            // If we are Binding the expanded property Trips, currentLevelEntityType should be the type of the VipCustomer
             if (this.parsedSegments.Count > 0 && this.parsedSegments.Last() is TypeSegment)
             {
                 currentLevelEntityType = this.parsedSegments.Last().EdmType as IEdmStructuredType;
@@ -497,7 +497,7 @@ namespace Microsoft.OData.UriParser
         /// <param name="applyToken">The apply tokens to visit.</param>
         /// <param name="resourcePathNavigationSource">The navigation source at the resource path.</param>
         /// <param name="targetNavigationSource">The target navigation source at the current level.</param>
-        /// <param name="elementType">The Edm element type.</param>
+        /// <param name="elementType">The target element type if different from the type of the target navigation source.</param>
         /// <returns>The null or the built apply clause.</returns>
         private ApplyClause BindApply(IEnumerable<QueryToken> applyToken, IEdmNavigationSource resourcePathNavigationSource, IEdmNavigationSource targetNavigationSource, IEdmTypeReference elementType = null)
         {
@@ -517,7 +517,7 @@ namespace Microsoft.OData.UriParser
         /// <param name="computeToken">The compute token to visit.</param>
         /// <param name="resourcePathNavigationSource">The navigation source at the resource path.</param>
         /// <param name="targetNavigationSource">The target navigation source at the current level.</param>
-        /// <param name="elementType">The target element type.</param>
+        /// <param name="elementType">The target element type if different from the type of the target navigation source.</param>
         /// <returns>The null or the built compute clause.</returns>
         private ComputeClause BindCompute(ComputeToken computeToken, IEdmNavigationSource resourcePathNavigationSource, IEdmNavigationSource targetNavigationSource, IEdmTypeReference elementType = null)
         {
@@ -537,7 +537,7 @@ namespace Microsoft.OData.UriParser
         /// <param name="filterToken">The filter token to visit.</param>
         /// <param name="resourcePathNavigationSource">The navigation source at the resource path.</param>
         /// <param name="targetNavigationSource">The target navigation source at the current level.</param>
-        /// <param name="elementType">The Edm element type.</param>
+        /// <param name="elementType">The target element type if different from the type of the target navigation source.</param>
         /// <param name="generatedProperties">The generated properties.</param>
         /// <param name="collapsed">The collapsed boolean value.</param>
         /// <returns>The null or the built filter clause.</returns>
@@ -560,7 +560,7 @@ namespace Microsoft.OData.UriParser
         /// <param name="orderByToken">The orderby token to visit.</param>
         /// <param name="resourcePathNavigationSource">The navigation source at the resource path.</param>
         /// <param name="targetNavigationSource">The target navigation source at the current level.</param>
-        /// <param name="elementType">The Edm element type.</param>
+        /// <param name="elementType">The target element type if different from the type of the target navigation source.</param>
         /// <param name="generatedProperties">The generated properties.</param>
         /// <param name="collapsed">The collapsed boolean value.</param>
         /// <returns>The null or the built filter clause.</returns>
@@ -584,7 +584,7 @@ namespace Microsoft.OData.UriParser
         /// <param name="searchToken">The search token to visit.</param>
         /// <param name="resourcePathNavigationSource">The navigation source at the resource path.</param>
         /// <param name="targetNavigationSource">The target navigation source at the current level.</param>
-        /// <param name="elementType">The Edm element type.</param>
+        /// <param name="elementType">The target element type if different from the type of the target navigation source.</param>
         /// <returns>The null or the built search clause.</returns>
         private SearchClause BindSearch(QueryToken searchToken, IEdmNavigationSource resourcePathNavigationSource, IEdmNavigationSource targetNavigationSource, IEdmTypeReference elementType)
         {
@@ -606,7 +606,7 @@ namespace Microsoft.OData.UriParser
         /// <param name="segments">The parsed segments to visit.</param>
         /// <param name="resourcePathNavigationSource">The navigation source at the resource path.</param>
         /// <param name="targetNavigationSource">The target navigation source at the current level.</param>
-        /// <param name="elementType">The Edm element type.</param>
+        /// <param name="elementType">The target element type if different from the type of the target navigation source.</param>
         /// <param name="generatedProperties">The generated properties.</param>
         /// <param name="collapsed">The collapsed boolean value.</param>
         /// <returns>The null or the built select and expand clause.</returns>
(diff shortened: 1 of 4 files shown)


habbes (commented):
Looks good to me, but I don't know enough of the inner working to ensure that the tests are sufficient.

Could you add some tests that verify that the sample queries you made actually return the expected results? And also tests to verify the validation (e.g. if the nav prop's type and the type segment's type are not related, the expected exception is thrown?

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -441,6 +441,8 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
 
             ODataExpandPath pathToNavProp = new ODataExpandPath(pathSoFar);
 
+            IEdmTypeReference elementTypeReference = hasDerivedTypeSegment ? derivedType.ToTypeReference() : null;
```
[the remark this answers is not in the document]
KenitoInc: It's null since we can get it from the expanded property in these lines
https://github.com/OData/odata.net/blob/145252226e60b4c9735f1c643fe3e01af039eab0/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs#L924-L929

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -451,13 +453,13 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
             bool collapsed = applyOption?.Transformations.Any(t => t.Kind == TransformationNodeKind.Aggregate || t.Kind == TransformationNodeKind.GroupBy) ?? false;
 
             // $filter
-            FilterClause filterOption = BindFilter(tokenIn.FilterOption, this.ResourcePathNavigationSource, targetNavigationSource, null, generatedProperties, collapsed);
+            FilterClause filterOption = BindFilter(tokenIn.FilterOption, this.ResourcePathNavigationSource, targetNavigationSource, elementTypeReference, generatedProperties, collapsed);
```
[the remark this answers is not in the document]
KenitoInc: #1 - Same as the answer above. When null, we can get it from the expanded property (`targetNavigationSource`)
#2 - Will look into `BindCompute`

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -659,8 +661,27 @@ private List<ODataPathSegment> ProcessSelectTokenPath(PathSegmentToken tokenIn)
...
+            }
+
+            // When we have a derived type, the current edmType is different from the edmType of the RangeVariable type reference.
+            // e.g /Orders?$expand=Customer/Model.VipCustomer($select=VipCustomerName)
+            // We need to use the edmType of the RangeVariable type reference so that we can get the VipCustomerName property
+            // from the Model.VipCustomer.
+            if (implicitRangeVariableType != null && implicitRangeVariableType != this.edmType)
+            {
+                // Validate that the derived type and the base type are related.
+                UriEdmHelpers.CheckRelatedTo(this.state.ImplicitRangeVariable.TypeReference.Definition, this.edmType);
```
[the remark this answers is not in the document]
KenitoInc: The order of the arguments doesn't matter

gathogojr (approved):
LGTM

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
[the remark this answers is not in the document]
gathogojr: @KenitoInc Did you look into `BindCompute`?

src/Microsoft.OData.Core/UriParser/ODataPathInfo.cs
```diff
@@ -35,7 +35,10 @@ public ODataPathInfo(ODataPath odataPath)
                 }
 
                 this.targetNavigationSource = lastSegment.TargetEdmNavigationSource;
-                this.targetEdmType = lastSegment.TargetEdmType;
+
+                // If the last segment is a TypeSegment, we should assign the targetEdmType as lastSegment.EdmType
+                // instead of lastSegment.TargetEdmType which references the Base Type EdmType.
```
mikepizzo, line 40: Shouldn't the TargetEdmType of a TypeSegment be the EdmType?

Do we have other places in code where we make the same assumption (that the targetEdmType of a Type Segment is the actual EdmType, not the base type?

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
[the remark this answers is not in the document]
mikepizzo: I see that we are getting the type in SelectExpandBinder, but what is the contract with BindFilter, BindOrderby, etc? These private functions take an elementType which gets passed eventually to CreateBindingState where the logic is to fall back to the navigationSource, if it's not otherwise specified. But it's not clear in the various private methods that the elementType is nullable, and future code could make invalid assumptions about whether it may be null.  For maintainability, maybe at least clarify in the description of the elementType parameter (i.e., line 529) "The target element type if different from the type of the target navigation source."

mikepizzo:
>             IEdmType targetElementType = lastSegment.TargetEdmType;

Do we need similar logic here for taking a cast segment into account for $select?

---
Refers to: src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs:287 in aecf2af. [](commit_id = aecf2af6836a10a5b2dda5f0bc0aeedc6de3cc8d, deletion_comment = False)

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -441,6 +441,8 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
 
             ODataExpandPath pathToNavProp = new ODataExpandPath(pathSoFar);
 
+            IEdmTypeReference elementTypeReference = hasDerivedTypeSegment ? derivedType.ToTypeReference() : null;
+
             // $apply
             ApplyClause applyOption = BindApply(tokenIn.ApplyOptions, this.ResourcePathNavigationSource, targetNavigationSource);
```
mikepizzo, line 447: >BindApply [](http://example.com/codeflow?start=38&length=9)

should we be passing elementTypeReference to BindApply?

mikepizzo:
>             ComputeClause computeOption = BindCompute(tokenIn.ComputeOption, this.ResourcePathNavigationSource, targetNavigationSource);

Should we be passing elementTypeReference to BindCompute?

---
Refers to: src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs:450 in aecf2af. [](commit_id = aecf2af6836a10a5b2dda5f0bc0aeedc6de3cc8d, deletion_comment = False)

mikepizzo (changes requested):
:clock1:

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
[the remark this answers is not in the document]
habbes: Not a blocker, but is there a test that confirms that by any chance?

src/Microsoft.OData.Core/UriParser/ODataPathInfo.cs
KenitoInc: Reverted this change since it was solving one use case `$expand=Prop1($expand=Prop2`.
Handling this in the `GenerateExpandItem` method in the `SelectExpandBinder`

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
KenitoInc: Yes - I have fixed the code.

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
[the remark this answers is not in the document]
KenitoInc: @mikepizzo I have updated the description.

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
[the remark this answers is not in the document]
KenitoInc: I have refactored that logic. We don't need to run that validation here since it's validated in the previous level binding i.e 
In this example
`Orders?$expand=Customer/NS.VipCustomer($select=VipCustomerName)`
We are binding the `$select`. However validation has already happened in the previous level `$expand=Customer/NS.VipCustomer`
https://github.com/OData/odata.net/blob/a510d92aba9ea220ae6de67e2ef659cd1d971b02/src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs#L392

Test are here https://github.com/OData/odata.net/blob/a510d92aba9ea220ae6de67e2ef659cd1d971b02/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/SelectExpandFunctionalTests.cs#L1802

KenitoInc:
> > ```
> >         ComputeClause computeOption = BindCompute(tokenIn.ComputeOption, this.ResourcePathNavigationSource, targetNavigationSource);
> > ```
> 
> Should we be passing elementTypeReference to BindCompute?
> 
> Refers to: src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs:450 in [aecf2af](https://github.com/OData/odata.net/commit/aecf2af6836a10a5b2dda5f0bc0aeedc6de3cc8d). [](commit_id = [aecf2af](https://github.com/OData/odata.net/commit/aecf2af6836a10a5b2dda5f0bc0aeedc6de3cc8d), deletion_comment = False)

I have fixed this

habbes (approved):
The code looks good. I think adding some e2e tests that verify that the expected data is returned from a request would be great.

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -343,6 +343,27 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
             PathSegmentToken currentToken = tokenIn.PathToNavigationProp;
 
             IEdmStructuredType currentLevelEntityType = this.EdmType;
+
+            // If we have a previous level binding with the last segment as a Type segment,
+            // The currentLevelEntityType should be the Type of the TypeSegment
+            // E.g /Orders?$expand=Customer/NS.VipCustomer($expand=Trips)
+            // If we are Binding the expanded property Trips, currentLevelEntityType should be the type of the VipCustomer
+            if (this.parsedSegments.Count > 0 && this.parsedSegments.Last() is TypeSegment)
```
corranrogue9, line 351: Instead of calling `Last`, we could call `this.parsedSegments[this.parsedSegments.Count - 1]` to save a cast that LINQ will have to do

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -343,6 +343,27 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
...
 
             IEdmStructuredType currentLevelEntityType = this.EdmType;
+
+            // If we have a previous level binding with the last segment as a Type segment,
+            // The currentLevelEntityType should be the Type of the TypeSegment
+            // E.g /Orders?$expand=Customer/NS.VipCustomer($expand=Trips)
+            // If we are Binding the expanded property Trips, currentLevelEntityType should be the type of the VipCustomer
+            if (this.parsedSegments.Count > 0 && this.parsedSegments.Last() is TypeSegment)
+            {
+                IEdmType typeSegmentType = this.parsedSegments.Last().EdmType;
```
corranrogue9, line 353: Whatever you decide for the above comment, it would be best to store the result of the last element in a local variable so that we don't have to deference or cast twice

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -343,6 +343,27 @@ private SelectItem GenerateExpandItem(ExpandTermToken tokenIn)
...
+
+                IEdmCollectionType collection = typeSegmentType as IEdmCollectionType;
+
+                if (collection != null)
+                {
+                    currentLevelEntityType = (typeSegmentType as IEdmCollectionType).ElementType.Definition as IEdmStructuredType;
```
corranrogue9, line 359: I'm not sure why we need to call `as` twice, couldn't this just read `collection.ElementType.Definition as IEdmStructureType`?

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -524,7 +548,7 @@ private ComputeClause BindCompute(ComputeToken computeToken, IEdmNavigationSourc
         /// <param name="filterToken">The filter token to visit.</param>
         /// <param name="resourcePathNavigationSource">The navigation source at the resource path.</param>
         /// <param name="targetNavigationSource">The target navigation source at the current level.</param>
-        /// <param name="elementType">The Edm element type.</param>
+        /// <param name="elementType">The target element type if different from the type of the target navigation source.</param>
```
corranrogue9, line 551: I think it would be really good for this updated documentation to specify that `null` indicates that the element type is the same as the target navigation source type

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
```diff
@@ -659,8 +683,29 @@ private List<ODataPathSegment> ProcessSelectTokenPath(PathSegmentToken tokenIn)
...
 
             List<ODataPathSegment> pathSoFar = new List<ODataPathSegment>();
+
             IEdmStructuredType currentLevelType = this.edmType;
 
+            // If we have a previous level binding with the last segment as a Type segment,
+            // The currentLevelEntityType should be the Type of the TypeSegment
+            // E.g /Orders?$expand=Customer/NS.VipCustomer($select=VipCustomerName)
+            // If we are Binding the selected property VipCustomerName, currentLevelEntityType should be the type of the VipCustomer
+            if (this.parsedSegments.Count > 0 && this.parsedSegments.Last() is TypeSegment)
```
corranrogue9, line 693: I have all of the same micro-optimization feedback here, though I do wonder if we can factor this code into a common method so it doesn't need to be duplicated

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs
KenitoInc: #Resolved

corranrogue9 (approved):

mikepizzo:
>             IEdmType targetElementType = lastSegment.TargetEdmType;

Is it as simple as changing this to lastSegment.EdmType?

---
In reply to: [1030205204](https://github.com/OData/odata.net/pull/2300#issuecomment-1030205204) [](http://example.com/codeflow?ancestors=1030205204,1030205204)

---
Refers to: src/Microsoft.OData.Core/UriParser/Binders/SelectExpandBinder.cs:287 in 8d62b20. [](commit_id = 8d62b207e37440bd15a1ef353caf1e50b3f8a092, deletion_comment = False)

mikepizzo (approved):
:shipit:

Issue #2295, opened 2022-01-12
Not all tests projects run on the build machine


corranrogue9:
Not all tests projects currently run on the build machine. Because of this, some tests are failing locally. We should fix the failing tests and enable those tests on the build machine. [This branch](https://github.com/OData/odata.net/commits/corranrogue9/onboarding3) can be used to fix all but these 3 of the test projects locally:
- microsoft.odata.service.design.t4.unittests
- microsoft.odata.service.tddunittests
- microsoft.test.odata.tests.client

habbes:
Some tests also trigger a dialog box on failure:
![image](https://user-images.githubusercontent.com/8460169/153821700-d9942a6f-c536-434e-8b7f-e21fe849da55.png)

Not sure why this is the case, and not sure how this behaviour in an environment with no GUI. But we should probably use the normal failure reporting.

corranrogue9:
This is what happens when a `Debug.Assert` condition is not met. I will be addressing these as well

corranrogue9:
Make sure to add a follow-up process for nightly builds, assuming we don't add the debug configurations to CI

corranrogue9:
Make sure rolling builds are also running the correct tests

corranrogue9:
There are 7 remaining items:

1. Automatically alert when a nightly build has failed
2. Address the test failures in microsoft.odata.service.design.t4.unittests. Either remove the project, or fix the tests and add the project to `OData.Pipeline.sln`
3. Address the test failures in microsoft.odata.service.tddunittests. Either remove the project, or fix the tests and add the project to `OData.Pipeline.sln`
4. Address the test failures in microsoft.test.odata.tests.client. Either remove the project, or fix the tests and add the project to `OData.Pipeline.sln`.  Some of the tests might be addressed with this branch: `corranrogue9/clienttests`
5. Address the build failures in datawebrules. Either remove the project, or fix the build and add the project to `OData.Pipeline.sln`.
6. Address the build failures in servicewrapperapp. Either remove the project, or fix the build and add the project to `OData.Pipeline.sln`.
7. Update `buildandtest.yml` to use `OData.E2E.sln` and remove `OData.Pipeline.sln`

PR #2315, opened 2022-02-14
Fix e2e tests build error

habbes:
### Issues

### Description

A [recent commit](https://github.com/OData/odata.net/commit/7dcad74478debcfe54e93c63f47b7cb57f2d2e67) made the `RemoveRule` method of the `InstanceAnnotationDictWeakKeyComparer` class static. This broke the e2e tests that were still referencing the method as if it were an instance method. This PR fixes that issue and gets the tests building again.

PS: Some tests are still failing locally, and are not running on CI (which is likely why this build error was not caught when the PR that introduced the issue was merged). There are other open issues to tackle those problems:
- https://github.com/OData/odata.net/issues/2295

- [ ] *Test cases added*


KenitoInc (approved):

gathogojr (approved):
:shipit:

PR #2316, opened 2022-02-14
Implement asynchronous support in ReorderingJsonReader

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ReorderingJsonReader`**.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ReorderingJsonReader.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ReorderingJsonReader.cs
@@ -11,6 +11,7 @@ namespace Microsoft.OData.JsonLight
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.IO;
+    using System.Threading.Tasks;
     using Microsoft.OData.Json;
     #endregion Namespaces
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.OData.Core/JsonLight/ReorderingJsonReader.cs
```diff
@@ -82,6 +83,85 @@ public override bool CanStream()
...
+            Stream result;
+            object value = null;
+
+            try
+            {
+                value = await this.GetValueAsync()
+                    .ConfigureAwait(false);
+
+                if (value == null)
+                {
```
marabooy, line 106: suggestion: Maybe use the Stream.Null to avoid any allocations if we don't reuse this. 
https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.null?view=net-6.0

marabooy: Ps you may need to evaluate if this would be potentially breaking.

marabooy (approved):

KenitoInc (approved):

src/Microsoft.OData.Core/JsonLight/ReorderingJsonReader.cs
gathogojr: I'm inclined towards not making use of `Stream.Null` at this time to allow myself to understand what effect and impact that might have. I'd also not want the logic in async method to differ too much from the sync method. I'll see about doing a PR that addresses from the sync and async implementations.

PR #2324, opened 2022-02-18
Implement asynchronous support in ODataNotificationReader

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataNotificationReader`**.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataNotificationReader.cs
+++ b/src/Microsoft.OData.Core/ODataNotificationReader.cs
@@ -20,15 +20,24 @@ namespace Microsoft.OData
     internal sealed class ODataNotificationReader : TextReader
 #endif
     {
-        private TextReader textReader;
-        private IODataStreamListener listener;
-        private bool disposed = false;
+        private readonly TextReader textReader;
+        private readonly IODataStreamListener listener;
         private readonly bool synchronous;
+        private bool disposed = false;
 
+        /// <summary>
+        /// Initializes a new instance of the <see cref="ODataNotificationReader"/> class.
+        /// </summary>
+        /// <param name="textReader">The wrapped text reader.</param>
+        /// <param name="listener">Listener to notify when the text reader is being disposed.</param>
+        /// <param name="synchronous">true if execution context is synchronous; otherwise false.</param>
+        /// <remarks>
+        /// When an instance of this class is disposed, it in turn disposes the wrapped text reader.
+        /// </remarks>
         internal ODataNotificationReader(TextReader textReader, IODataStreamListener listener, bool synchronous = true)
         {
             Debug.Assert(textReader != null, "Creating a notification reader for a null textReader.");
-            Debug.Assert(listener != null, "Creating a notification reader with a null textReader.");
+            Debug.Assert(listener != null, "Creating a notification reader with a null listener.");
 
             this.textReader = textReader;
             this.listener = listener;
@@ -126,33 +135,26 @@ namespace Microsoft.OData
                 // Tell the listener that the stream is being disposed.
                 if (synchronous)
                 {
-                    this.listener?.StreamDisposed();
+                    this.listener.StreamDisposed();
                 }
                 else
                 {
-                    this.listener?.StreamDisposedAsync().Wait();
+                    this.listener.StreamDisposedAsync().Wait();
                 }
-
-                this.listener = null;
-                // NOTE: Do not dispose the text reader since this instance does not own it.
-                this.textReader = null;
             }
 
             this.disposed = true;
+            this.textReader?.Dispose();
             base.Dispose(disposing);
         }
 
 #if NETSTANDARD2_0
         public async ValueTask DisposeAsync()
         {
-            if (!this.disposed && this.listener != null)
+            if (!this.disposed)
             {
                 await this.listener.StreamDisposedAsync()
                     .ConfigureAwait(false);
-
-                this.listener = null;
-                // NOTE: Do not dispose the text reader since this instance does not own it.
-                this.textReader = null;
             }
 
             // Dispose unmanaged resources
--- a/src/Microsoft.OData.Core/ODataNotificationReader.cs
+++ b/src/Microsoft.OData.Core/ODataNotificationReader.cs
@@ -6,6 +6,7 @@
 
 namespace Microsoft.OData
 {
+    using System;
     using System.Diagnostics;
     using System.IO;
     using System.Threading.Tasks;
@@ -13,18 +14,25 @@ namespace Microsoft.OData
     /// <summary>
     /// Wrapper for TextReader to listen for dispose.
     /// </summary>
+#if NETSTANDARD2_0
+    internal sealed class ODataNotificationReader : TextReader, IAsyncDisposable
+#else
     internal sealed class ODataNotificationReader : TextReader
+#endif
     {
-        private readonly TextReader textReader;
+        private TextReader textReader;
         private IODataStreamListener listener;
+        private bool disposed = false;
+        private readonly bool synchronous;
 
-        internal ODataNotificationReader(TextReader textReader, IODataStreamListener listener)
+        internal ODataNotificationReader(TextReader textReader, IODataStreamListener listener, bool synchronous = true)
         {
             Debug.Assert(textReader != null, "Creating a notification reader for a null textReader.");
             Debug.Assert(listener != null, "Creating a notification reader with a null textReader.");
 
             this.textReader = textReader;
             this.listener = listener;
+            this.synchronous = synchronous;
         }
 
         /// <inheritdoc/>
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/ODataNotificationReader.cs
```diff
@@ -113,18 +121,44 @@ public override Task<string> ReadToEndAsync()
...
+                if (synchronous)
                 {
-                    // Tell the listener that the stream is being disposed.
-                    this.listener.StreamDisposed();
-                    this.listener = null;
+                    this.listener?.StreamDisposed();
                 }
+                else
+                {
+                    this.listener?.StreamDisposedAsync().Wait();
```
habbes, line 133: Out of curiosity, is it better to call a synchronous method or to `Wait()` on an async method?

src/Microsoft.OData.Core/ODataNotificationReader.cs
```diff
@@ -113,18 +121,44 @@ public override Task<string> ReadToEndAsync()
...
+                }
+
+                this.listener = null;
+                // NOTE: Do not dispose the text reader since this instance does not own it.
+                this.textReader = null;
```
habbes, line 138: Does this mean the original implementation had a bug? Since it disposed the `textReader`?

src/Microsoft.OData.Core/ODataNotificationReader.cs
```diff
@@ -113,18 +121,44 @@ public override Task<string> ReadToEndAsync()
...
             }
 
-            this.textReader.Dispose();
+            this.disposed = true;
             base.Dispose(disposing);
```
habbes, line 142: Should we still call `base.Dispose(disposing)` if `this.disposed` is `true`?

habbes (commented):
I have some concerns regarding the ownership of the wrapped `TextReader`. You mention in the comments that it should not be disposed by the `ODataNotificationReader` since it's not the owner. However, in the original version, the `ODataNotificationReader` did dispose of the `TextReader`.  Can you confirm that your changes are the expected behaviour and the previous implementation was a bug?

I have found [the following scenario](https://github.com/OData/odata.net/blob/511895d1362cd56ac911f48c96fb345049800d0e/src/Microsoft.OData.Core/ODataReaderCore.cs#L440) where an `ODataNotificationReader` is created:

```c#
public override sealed TextReader CreateTextReader()
{
    if (this.State == ODataReaderState.Stream)
    {
        /* ... omitted for brevity */
        return new ODataNotificationReader(this.InterceptException(this.CreateTextReaderImplementation), this);
    }
    else
    {
        throw new ODataException(Strings.ODataReaderCore_CreateTextReaderCalledInInvalidState);
    }
}
```

Here, the a `TextReader` is created inline and passed directly to the `ODataNotificationReader` constructor. There's no other reference left for the `TextReader`. So if the `ODataNotificationReader` does not dispose it, then `TextReader` will not be disposed as there's no other which has a reference to it. And the calling method will only get back a reference of the `ODataNotificationReader` since that's the return value, and will have no idea that it's wrapping another `TextReader`.

If `ODataNotificationReader` is not the owner, then that logic  should be refactored such to ensure `TextReader` gets properly disposed. If `ODataNotificationReader` is the owner (and the above snippet suggests that it is), then it should dispose of `TextReader`. In either case, I think the doc comments of `ODataNotificationReader` should be updated to mention whether or not it assumes the ownership responsibility of disposing `TextReader`.

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationReaderTests.cs
```diff
@@ -0,0 +1,161 @@
...
+            // 2nd call to Dispose
+            notificationReader.Dispose();
+
+            var result = ReadStreamContents();
+
+            // StreamDisposed/StreamDisposeAsync was written only once
+            Assert.Equal(expected, result);
+        }
+
+#if NETCOREAPP3_1
```
habbes, line 67: Why not `NETSTANDARD2_0` since that's what you used on the class definition?

src/Microsoft.OData.Core/ODataNotificationReader.cs
```diff
@@ -113,18 +121,44 @@ public override Task<string> ReadToEndAsync()
...
+#if NETSTANDARD2_0
+        public async ValueTask DisposeAsync()
+        {
+            if (!this.disposed && this.listener != null)
+            {
+                await this.listener.StreamDisposedAsync()
+                    .ConfigureAwait(false);
+
+                this.listener = null;
+                // NOTE: Do not dispose the text reader since this instance does not own it.
```
corranrogue9, line 154: I'm a little nervous about this change of behavior for not disposing `textReader`. I agree with the change on principal (the `ODataNotificationReader` does not own the `textReader`) but anyone currently using the `ODataNotificationReader` is already going to be expecting it to dispose the `TextReader`. Perhaps we should add a flag in the constructor so that current consumers get the existing behavior, but future consumers are able to maintain control of their `TextReader`?

src/Microsoft.OData.Core/ODataNotificationReader.cs
```diff
@@ -113,18 +121,44 @@ public override Task<string> ReadToEndAsync()
         /// <param name="disposing">True if called from Dispose; false if called from the finalizer.</param>
         protected override void Dispose(bool disposing)
         {
-            if (disposing)
+            if (!this.disposed && disposing)
```
corranrogue9, line 124: Thank you for adding this! I'm actually going to post next week in the maintainability channel more about disposing. However, this does *now* follow the MSDN guidelines for implementing dispose, I love it!

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationReaderTests.cs
```diff
@@ -0,0 +1,161 @@
...
+{
+    public class ODataNotificationReaderTests
+    {
+        private MemoryStream stream;
+        private TextReader reader;
+        private IODataStreamListener streamListener;
+
+        public ODataNotificationReaderTests()
+        {
+            this.stream = new MemoryStream();
```
corranrogue9, line 22: This stream and the reader should both really be disposed after each test. I'm not familiar enough with XUnit to know how to do that. Is there a way to have cleanup code called after each test is run?

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationReaderTests.cs
```diff
@@ -0,0 +1,161 @@
...
+            var result = await this.ReadStreamContentsAsync();
+
+            Assert.Equal("StreamDisposedAsync", result);
+        }
+#endif
+
+        private string ReadStreamContents()
+        {
+            this.stream.Position = 0;
+            return new StreamReader(this.stream).ReadToEnd();
```
corranrogue9, line 120: Please make sure to dispose this intermediate object before the reference to it is lost

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationReaderTests.cs
```diff
@@ -0,0 +1,161 @@
...
+        }
+
+        private Task<string> ReadStreamContentsAsync()
+        {
+            this.stream.Position = 0;
+            return new StreamReader(this.stream).ReadToEndAsync();
```
corranrogue9, line 126: Same here

corranrogue9 (changes requested):

src/Microsoft.OData.Core/ODataNotificationReader.cs
gathogojr: @corranrogue9 The `ODataNotificationReader` is internal and the callers are internal. Instead of taking the longer path of introducing a flag, I'll verify and ensure that the callers dispose the `textReader` object since they own it

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationReaderTests.cs
gathogojr: @corranrogue9 To quote from another source... `MemoryStream` does not have any unmanaged resources to dispose, so you don’t technically have to dispose of it. The effect of not disposing a `MemoryStream` is roughly the same thing as dropping a reference to a `byte[]` -- the GC will clean both up the same way.
Specifically for tests, I wouldn't sweat too much over that

src/Microsoft.OData.Core/ODataNotificationReader.cs
gathogojr: @habbes Given that the `Dispose` method is defined in a framework class and advertised as idempotent, it should be okay to call `base.Dispose(disposing)` any number of times.

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationReaderTests.cs
gathogojr: Because `NETSTANDARD2_0` is not a target framework in the test project (Microsoft.OData.Core.Tests) but it is in Microsoft.OData.Core project.
`await using` is only supported in `NETCOREAPP3_1` or higher

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationReaderTests.cs
gathogojr: Fixed by having the test class implement `IDisposable` interface and implementing the `Dispose` method that Xunit relies on for cleanup after each test is ran

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNotificationReaderTests.cs
gathogojr: Fixed by having the test class implement `IDisposable` interface and implementing the `Dispose` method that Xunit relies on for cleanup after each test is ran

corranrogue9 (approved):

habbes (approved):

src/Microsoft.OData.Core/ODataNotificationReader.cs
habbes: I'd argue in this case (based on the usage scenarios) that the caller "transferred" ownership of the `textReader` to this class.

gathogojr: @habbes True. That's the reasoning I went with in the end. I added a remark to the constructor indicating that when this class is disposed, it in turns disposes the wrapped text reader

src/Microsoft.OData.Core/ODataNotificationReader.cs
```diff
@@ -113,18 +121,44 @@ public override Task<string> ReadToEndAsync()
...
+        public async ValueTask DisposeAsync()
+        {
+            if (!this.disposed && this.listener != null)
+            {
+                await this.listener.StreamDisposedAsync()
+                    .ConfigureAwait(false);
+
+                this.listener = null;
+                // NOTE: Do not dispose the text reader since this instance does not own it.
+                this.textReader = null;
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/ODataNotificationReader.cs
gathogojr: See the discussion preceding [this](https://github.com/OData/odata.net/pull/2324#discussion_r814506311) comment

src/Microsoft.OData.Core/ODataNotificationReader.cs
gathogojr: @habbes I have created a work item to do an investigation here and determine the better way

PR #2328, opened 2022-02-25
Add DuplicatePropertyNameChecker to ObjectPool

### LINKED ISSUE #2180: Performance - DuplicatePropertyNameCheck costing 1.5% of allocations

KenitoInc:
### Issues

*This pull request fixes #2180 .*

### Description
We were creating a new instance of the `DuplicatePropertyNameChecker` when serializing every `ODataResource`. This resulted in too many allocations causing pressure on the GC.

This PR adds an ObjectPool which stores instances of `DuplicatePropertyNameChecker`. Everytime we are serializing an `ODataResource`, we request a `DuplicatePropertyNameChecker` from the ObjectPool. Once we are done serializing the `ODataResource`, we release the`DuplicatePropertyNameChecker` to the ObjectPool.

### Performance profiler comparison results
We wrote a sample service that writes this response 5000 times
```json
{
  "@odata.context": "http://www.example.com/$metadata#Customers",
  "value": [
    {
      "Id": 1,
      "Name": "Cust1",
      "Emails": [
        "emailA$1",
        "emailB1 "
      ],
      "HomeAddress": {
        "City": "City1",
        "Street": "Street1"
      },
      "Addresses": [
        {
          "Name": "CityA1",
          "City": "CityA1",
          "Street": "StreetA1",
          "PostalAddress": {
            "Name": "PostalAddressNameA1",
            "PostalBox": "PostalBoxA1",
            "PostalCode": "PostalCodeA1",
            "PostalAddressLocation": {
              "Town": "TownA1",
              "County": "CountyA1"
            }
          }
        },
        {
          "Name": "CityB1",
          "City": "CityB1",
          "Street": "StreetB1",
          "PostalAddress": {
            "Name": "PostalAddressNameB1",
            "PostalBox": "PostalBoxB1",
            "PostalCode": "PostalCodeB1",
            "PostalAddressLocation": {
              "Town": "TownB1",
              "County": "CountyB1"
            }
          }
        },
        {
          "Name": "CityC1",
          "City": "CityC1",
          "Street": "StreetC1",
          "PostalAddress": {
            "Name": "PostalAddressNameC1",
            "PostalBox": "PostalBoxC1",
            "PostalCode": "PostalCodeC1",
            "PostalAddressLocation": {
              "Town": "TownC1",
              "County": "CountyC1"
            }
          }
        },
        {
          "Name": "CityD1",
          "City": "CityD1",
          "Street": "StreetD1",
          "PostalAddress": {
            "Name": "PostalAddressNameD1",
            "PostalBox": "PostalBoxD1",
            "PostalCode": "PostalCodeD1",
            "PostalAddressLocation": {
              "Town": "TownD1",
              "County": "CountyD1"
            }
          }
        },
        {
          "Name": "CityE1",
          "City": "CityE1",
          "Street": "StreetE1",
          "PostalAddress": {
            "Name": "PostalAddressNameE1",
            "PostalBox": "PostalBoxE1",
            "PostalCode": "PostalCodeE1",
            "PostalAddressLocation": {
              "Town": "TownE1",
              "County": "CountyE1"
            }
          }
        }
      ]
    }
  ]
}
```
#### Before the fix
We had 85000 allocations of the `DuplicatePropertyNameChecker`
Master-Allocations

Total CPU usage was 0.52% + 0.07% = 0.59%
Master-CPU
Master-CPU2

#### After the fix
Allocations reduced to 4
Branch-Allocations

Total CPU usage dropped to 0.07% + 0.07% = 0.14%
Branch-CPU
Branch-CPU2

#### After the fix with Array
Allocations reduced to 4
Branch-AllocationsArray

CPU reduced to 0.22%
Branch-CPUArray2

#### Results Comparer
Allocations
ResultsComparer-Allocations

#### Before the fix
Writer-Master

#### After the fix
Writer-Branch

#### After the fix with Array
Writer-BranchArray3

- [ ] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/WriterValidator.cs
```diff
@@ -35,9 +54,17 @@ internal WriterValidator(ODataMessageWriterSettings settings)
...
-                   ? (IDuplicatePropertyNameChecker)new DuplicatePropertyNameChecker()
-                   : (IDuplicatePropertyNameChecker)new NullDuplicatePropertyNameChecker();
+            return this.duplicatePropertyNameCheckerObjectPool.GetObject();
+        }
+
+        /// <summary>
+        /// Release a DuplicatePropertyNameChecker instance.
+        /// </summary>
+        /// <returns>The created instance.</returns>
+        public void ReleaseDuplicatePropertyNameChecker(IDuplicatePropertyNameChecker duplicatePropertyNameChecker)
```
odero, line 64: Curious if it would make sense to use `IDisposable` pattern for this

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightValueSerializerTests.cs
```diff
@@ -216,6 +216,47 @@ public void WritingCollectionResourceValueWithPropertiesShouldWrite()
             Assert.Equal(@"[{""City"":""MyCity1""},{""City"":""MyCity2""}]", result);
         }
 
+        [Fact]
+        public void WritingResourceValueAndNestedResourceValueWithSimilarPropertiesShouldWrite()
+        {
+            var complexType = new EdmComplexType("NS", "Address");
```
odero, line 222: `var` vs explicit types in this file?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightValueSerializerTests.cs
```diff
@@ -216,6 +216,47 @@ public void WritingCollectionResourceValueWithPropertiesShouldWrite()
...
+                            }
+                        }
+                    }
+                };
+
+                var entityTypeRef = new EdmEntityTypeReference(entityType, false);
+                serializer.WriteResourceValue(resourceValue, entityTypeRef, false, serializer.CreateDuplicatePropertyNameChecker());
+            });
+
+            Assert.Equal(@"{""Name"":""MyName"",""Location"":{""Name"":""AddressName"",""City"":""MyCity""}}", result);
```
odero, line 257: Could you include an example test for the failing case?

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,49 @@
...
+{
+    internal class ObjectPool<T>
+    {
+        private readonly List<T> _objectsInUse;
+        private readonly List<T> _objectsAvailable;
+        private readonly Func<T> _objectGenerator;
+
+        public ObjectPool(Func<T> objectGenerator)
+        {
+            _objectGenerator = objectGenerator ?? throw new ArgumentNullException(nameof(objectGenerator));
```
odero, line 20: nit: `_field` pattern is not the convention used in this repo. Seems out of place

odero (changes requested):

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,49 @@
...
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.OData
+{
+    internal class ObjectPool<T>
```
corranrogue9, line 12: Can you please add xmldoc to the members of this class?

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,49 @@
...
+        {
+            _objectGenerator = objectGenerator ?? throw new ArgumentNullException(nameof(objectGenerator));
+            _objectsInUse = new List<T>();
+            _objectsAvailable = new List<T>();
+        }
+        public T GetObject()
+        {
+            if (_objectsAvailable.Count != 0)
+            {
+                var obj = _objectsAvailable[0];
```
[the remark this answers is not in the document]
corranrogue9: I believe because we remove the `0` element two lines down, so `0` is always the "next" available. It might make more sense to use a stack for this in that case, though I'm a bit skeptical of the .net stack implementation (no real reason, I just don't have enough experience with it to say that it's good). 

It might be best to just store the index of the next available, and increment/decrement from that. `List<T>.RemoveAt` is pretty expensive. I would guess you would have even more performance improvement using an index field instead.

corranrogue9:
Seeing the allocations reduced to 4 seems to confirm a suspicion I had while reading the code. I don't think the new object pool is being called from multiple threads, is it? If not, then pretty much every use of the object pool says "give me an unused object", does some work, and the gives the object back. Are you sure that we aren't just creating a single `DuplicatePropertyNameChecker` and then wrapping it up in an object pool that constantly moves it from one list to the other?

src/Microsoft.OData.Core/WriterValidator.cs
```diff
@@ -35,9 +54,17 @@ internal WriterValidator(ODataMessageWriterSettings settings)
...
+        {
+            duplicatePropertyNameChecker.Reset();
```
habbes, line 66: Since we're releasing the duplicate name checker, we are not sure it's going to be used again. Maybe it might be more economical to reset it when it's being requested from the pool. Here I'm also making the assumption that resetting a newly created duplicate name checker has negligible cost.

src/Microsoft.OData.Core/WriterValidator.cs
habbes: Making the `DuplicatePropertyNameChecker` disposable or the `ObjectPool` disposable?

habbes (changes requested):
I think the object pool has unnecessary costs.

src/Microsoft.OData.Core/ObjectPool.cs
habbes: Since you've made a generic object pool, I think you should also mention in the xmldoc that it's not thread-safe and not intended to be shared across concurrent requests just in case someone thinks of re-using it in different scenario.

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,49 @@
...
+    {
+        private readonly List<T> _objectsInUse;
+        private readonly List<T> _objectsAvailable;
+        private readonly Func<T> _objectGenerator;
+
+        public ObjectPool(Func<T> objectGenerator)
+        {
+            _objectGenerator = objectGenerator ?? throw new ArgumentNullException(nameof(objectGenerator));
+            _objectsInUse = new List<T>();
+            _objectsAvailable = new List<T>();
```
habbes, line 22: I think you can delay creating the array until the first it object is requested (or returned). Not sure if that will make a difference in practice though since most responses are objects.

src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
```diff
@@ -261,11 +261,11 @@ internal static string ConvertToResourceLiteral(ODataResourceValue resourceValue
                     model,
                     messageWriterSettings,
                     textWriter,
-                    (serializer) => serializer.WriteResourceValue(
+                    (serializer, duplicatePropertyNameChecker) => serializer.WriteResourceValue(
                         resourceValue, /* resourceValue */
                         null, /* metadataTypeReference */
                         true, /* isOpenPropertyType */
-                        serializer.CreateDuplicatePropertyNameChecker()));
+                        duplicatePropertyNameChecker));
```
habbes, line 268: Consider using named parameters here.

src/Microsoft.OData.Core/WriterValidator.cs
```diff
@@ -20,13 +20,32 @@ internal class WriterValidator : IWriterValidator
...
         /// <summary>
         /// Creates a WriterValidator instance and binds it to settings.
         /// </summary>
         /// <param name="settings">The ODataMessageWriterSettings instance to bind to.</param>
         internal WriterValidator(ODataMessageWriterSettings settings)
         {
             this.settings = settings;
+            this.duplicatePropertyNameCheckerObjectPool = settings.ThrowOnDuplicatePropertyNames ?
+                new ObjectPool<IDuplicatePropertyNameChecker>(() => new DuplicatePropertyNameChecker()) :
+                new ObjectPool<IDuplicatePropertyNameChecker>(() => new NullDuplicatePropertyNameChecker());
```
habbes, line 37: Since the `NullDuplicatePropertyNameChecker` holds no state, maybe it can be cached or created as a singleton and reused to avoid creating a new instance each time.

src/Microsoft.OData.Core/IWriterValidator.cs
```diff
@@ -20,6 +20,12 @@ internal interface IWriterValidator
         /// <returns>The created instance.</returns>
         IDuplicatePropertyNameChecker CreateDuplicatePropertyNameChecker();
 
+        /// <summary>
+        /// Release a DuplicatePropertyNameChecker instance.
+        /// </summary>
+        /// <returns>The created instance.</returns>
+        void ReleaseDuplicatePropertyNameChecker(IDuplicatePropertyNameChecker duplicatePropertyNameChecker);
```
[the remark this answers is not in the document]
KenitoInc: #Resolved

src/Microsoft.OData.Core/ODataSerializer.cs
```diff
@@ -88,5 +88,14 @@ internal IDuplicatePropertyNameChecker CreateDuplicatePropertyNameChecker()
         {
             return MessageWriterSettings.Validator.CreateDuplicatePropertyNameChecker();
         }
+
+        /// <summary>
+        /// Releases a new instance of a duplicate property names checker.
+        /// </summary>
+        /// <returns>The newly created instance of duplicate property names checker.</returns>
+        internal void ReleaseDuplicatePropertyNameChecker(IDuplicatePropertyNameChecker duplicatePropertyNameChecker)
```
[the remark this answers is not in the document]
KenitoInc: #Resolved

src/Microsoft.OData.Core/WriterValidator.cs
```diff
@@ -35,9 +54,17 @@ internal WriterValidator(ODataMessageWriterSettings settings)
...
-            return settings.ThrowOnDuplicatePropertyNames
-                   ? (IDuplicatePropertyNameChecker)new DuplicatePropertyNameChecker()
-                   : (IDuplicatePropertyNameChecker)new NullDuplicatePropertyNameChecker();
+            return this.duplicatePropertyNameCheckerObjectPool.GetObject();
+        }
+
+        /// <summary>
+        /// Release a DuplicatePropertyNameChecker instance.
+        /// </summary>
+        /// <returns>The created instance.</returns>
```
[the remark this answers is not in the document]
KenitoInc: #Resolved

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightValueSerializerTests.cs
KenitoInc: We can use `var` in tests.
The file uses `var` alot.

src/Microsoft.OData.Core/ObjectPool.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/WriterValidator.cs
odero: @habbes `DuplicateNameChecker` seeing as it's always calling "create" -> "do stuff" -> "release"

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,49 @@
...
+        }
+        public T GetObject()
+        {
+            if (_objectsAvailable.Count != 0)
+            {
+                var obj = _objectsAvailable[0];
+                _objectsInUse.Add(obj);
+                _objectsAvailable.RemoveAt(0);
```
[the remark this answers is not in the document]
odero: You mention elsewhere that `RemoveAt(0)` is constant time??

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            T item = firstItem;
+
+            if (item != null)
+            {
+                firstItem = default(T);
+                return item;
+            }
+
+            for (int i = 0; i < items.Length; i++)
```
habbes, line 48: I think there's a potential micro-optimization here. You could treat this is a stack and keep track of the last index where an object was found. After retrieving an object, you decrement that index, it's likely that the decremented index will contain an object. If you want to return an object, the index after the decrement index will likely be empty. But for a small fix-sized array, this is probably not worth it.

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// <param name="obj">The object to add to the pool.</param>
+        public void Return(T obj)
+        {
+            if (firstItem == null)
+            {
+                firstItem = obj;
+            }
+            else
+            {
+                for (var i = 0; i < items.Length; ++i)
```
habbes, line 74:
```suggestion
                for (int i = 0; i < items.Length; ++i)

```

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,91 @@
...
+        /// Return an object to the pool.
+        /// </summary>
+        /// <param name="obj">The object to add to the pool.</param>
+        public void Return(T obj)
+        {
+            if (firstItem == null)
+            {
+                firstItem = obj;
+            }
+            else
```
habbes, line 72: Suggestion: `return` after the statement in the if block and avoid the `else`.

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,91 @@
...
+                    if (items[i].Element == null)
+                    {
+                        items[i].Element = obj;
+                        break;
+                    }
+                }
+            }
+        }
+
+        // PERF: the struct wrapper avoids array-covariance-checks from the runtime when assigning to elements of the array.
```
habbes, line 85: suggestion: consider adding a link to the PR and/or issue discussion thread where this was suggested so that people can find additional context.

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,91 @@
...
+                }
+            }
+
+            return objectGenerator();
+        }
+
+        /// <summary>
+        /// Return an object to the pool.
+        /// </summary>
+        /// <param name="obj">The object to add to the pool.</param>
```
habbes, line 65:
```suggestion
        /// <param name="obj">The object to return to the pool.</param>

```

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,91 @@
...
+                    items[i].Element = default(T);
+                    return item;
+                }
+            }
+
+            return objectGenerator();
+        }
+
+        /// <summary>
+        /// Return an object to the pool.
```
habbes, line 63: I think you should mention that the call should ensure that the object is in a "default" state where it can be reused before being returned to the pool (e.g. `duplicateNameChecker. Clear()`). Alternatively, you could a reset callback function to constructor arguments and then the pool can call that function before retrieving an object from the pool. In this case, the pool would bear the responsibility for resetting the objects. e.g.:

```c#
class ObjectPool<T> where T : class
{
     // ...
    Action<T> resetObject;
   
    Get<T>()
    {
        if (firstItem != null) {
               item = firstItem;
               resetObject?.Invoke(item);
               firstItem = null;
               return item;
         }

        // ... same thing for item picked from the array
    }
}
```

habbes (changes requested):

src/Microsoft.OData.Core/IWriterValidator.cs
```diff
@@ -15,10 +15,15 @@ namespace Microsoft.OData
     internal interface IWriterValidator
     {
         /// <summary>
-        /// Creates a DuplicatePropertyNameChecker instance.
+        /// Get a DuplicatePropertyNameChecker instance from the object pool.
         /// </summary>
         /// <returns>The created instance.</returns>
```
ElizabethOkerio, line 20: nit: The DuplicatePropertyNameChecker instance. If it is just getting it and not creating it, i don't see why you are using the word 'created'

src/Microsoft.OData.Core/ObjectPool.cs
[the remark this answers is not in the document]
KenitoInc: Changed implementation to use a fixed size array.

src/Microsoft.OData.Core/ObjectPool.cs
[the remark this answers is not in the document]
KenitoInc: #removed

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,91 @@
...
+                item = items[i].Element;
+
+                if (item != null)
+                {
+                    items[i].Element = default(T);
+                    return item;
+                }
+            }
+
+            return objectGenerator();
```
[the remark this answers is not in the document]
KenitoInc: I switched to 8

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,91 @@
...
+        public T Get()
+        {
+            T item = firstItem;
+
+            if (item != null)
+            {
+                firstItem = default(T);
+                return item;
+            }
+
```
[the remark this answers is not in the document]
KenitoInc: #Resolved

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightValueSerializerTests.cs
KenitoInc: There are several existing tests https://github.com/OData/odata.net/blob/76c691bd822fa9eaec9a54863ff50991717ce482/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeserializerTests.cs#L701
https://github.com/OData/odata.net/blob/76c691bd822fa9eaec9a54863ff50991717ce482/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Streaming/ODataJsonLightStreamWritingTests.cs#L590

src/Microsoft.OData.Core/WriterValidator.cs
KenitoInc: Decided to keep the implementation as it is. We don't need to tightly couple the object pool and the `DuplicatePropertyNameChecker`.
Similar concept with the `ArrayPool` where we call `Rent` and `Return` directly.

KenitoInc:
> Seeing the allocations reduced to 4 seems to confirm a suspicion I had while reading the code. I don't think the new object pool is being called from multiple threads, is it? If not, then pretty much every use of the object pool says "give me an unused object", does some work, and the gives the object back. Are you sure that we aren't just creating a single `DuplicatePropertyNameChecker` and then wrapping it up in an object pool that constantly moves it from one list to the other?

The goal of this object pool is not to call it from multiple threads, but to call it everytime we are writing an `ODataResourceValue` and return it when we are done writing it.

habbes:
@KenitoInc consider using the [ResultsComparer](https://github.com/OData/odata.net/tree/master/tools/perf/ResultsComparer) to generate a report where it's easier to see the improvements and regressions. Also an opportunity to provide (much needed) feedback on the tool :)

habbes:
Do you have some profiler data as well? I think that would help explain whether the changes in perf are due to the expected reasons from your change, and could also help understand why there's a drastic drop in CPU time when `isModelImmutable` is `false` compared to when it's `true`.

src/Microsoft.OData.Core/WriterValidator.cs
xuzhg: Typically, we should call 'Reset' in the GetDuplicatePropertyNameChecker()'

src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
```diff
@@ -577,7 +579,18 @@ private static void WriteJsonLightLiteral(IEdmModel model, ODataMessageWriterSet
                 new ODataJsonLightOutputContext(textWriter, messageInfo, messageWriterSettings))
             {
                 ODataJsonLightValueSerializer jsonLightValueSerializer = new ODataJsonLightValueSerializer(jsonOutputContext);
-                writeValue(jsonLightValueSerializer);
+
+                if (ignoreDuplicatePropertyNameChecker)
```
xuzhg, line 583: >ignoreDuplicatePropertyNameChecker [](http://example.com/codeflow?start=20&length=34)

why don't use the setting in ODataMessageWriterSettings and avoid adding this parameter?

src/Microsoft.OData.Core/WriterValidator.cs
```diff
@@ -20,24 +20,45 @@ internal class WriterValidator : IWriterValidator
...
         /// <summary>
         /// Creates a WriterValidator instance and binds it to settings.
         /// </summary>
         /// <param name="settings">The ODataMessageWriterSettings instance to bind to.</param>
         internal WriterValidator(ODataMessageWriterSettings settings)
         {
             this.settings = settings;
+            this.duplicatePropertyNameCheckerObjectPool = settings.ThrowOnDuplicatePropertyNames ?
+                new ObjectPool<IDuplicatePropertyNameChecker>(() => new DuplicatePropertyNameChecker()) :
+                new ObjectPool<IDuplicatePropertyNameChecker>(() => new NullDuplicatePropertyNameChecker());
```
xuzhg, line 37: >NullDuplicatePropertyNameChecker [](http://example.com/codeflow?start=72&length=32)

shall we pool the NullDuplicatePropertyNameChecker

It seems we can use 'ONE' null duplicatePropertyNameChecker?

src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
```diff
@@ -106,7 +106,7 @@ protected IDuplicatePropertyNameChecker DuplicatePropertyNameChecker
             {
                 return duplicatePropertyNameChecker
                        ?? (duplicatePropertyNameChecker
-                           = outputContext.MessageWriterSettings.Validator.CreateDuplicatePropertyNameChecker());
+                           = outputContext.MessageWriterSettings.Validator.GetDuplicatePropertyNameChecker());
```
xuzhg, line 109: shall we return it?

src/Microsoft.OData.Core/ODataParameterWriterCore.cs
```diff
@@ -84,7 +84,7 @@ protected IDuplicatePropertyNameChecker DuplicatePropertyNameChecker
                 return this.duplicatePropertyNameChecker ??
                        (this.duplicatePropertyNameChecker =
                            outputContext.MessageWriterSettings.Validator
-                           .CreateDuplicatePropertyNameChecker());
+                           .GetDuplicatePropertyNameChecker());
```
xuzhg, line 87: shall we return it?

src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs
KenitoInc: I have changed the naming. This boolean was supposed to distinguish between `Resource` and `Collection` values. I have renamed to `isResourceValue`

src/Microsoft.OData.Core/WriterValidator.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -4133,7 +4133,7 @@ internal ResourceBaseScope(WriterState state, ODataResourceBase resource, ODataR
 
                 if (resource != null)
                 {
-                    duplicatePropertyNameChecker = writerSettings.Validator.CreateDuplicatePropertyNameChecker();
+                    duplicatePropertyNameChecker = writerSettings.Validator.GetDuplicatePropertyNameChecker();
```
corranrogue9, line 4136: Is this what gets returned in the `ODataJsonLightWriter`?

src/Microsoft.OData.Core/ObjectPool.cs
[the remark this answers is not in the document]
corranrogue9: I don't know what was here before, but there's really no need to have an instance of `ObjectPool` with a `null` value for `items`. I don't see us ever set it back to `null`, so we could mark it `readonly` if we initialize it in the constructor. Also, doing this means that we will check the conditional every time `Get` is called, which seems suboptimal.

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// </remarks>
+    internal class ObjectPool<T> where T : class
+    {
+        private readonly Func<T> objectGenerator;
+        private ObjectWrapper[] items;
+        private T firstItem;
+
+        // The pool size is equal to the level of nesting in the response.
+        // 32 is an arbitrary figure. Could be adjusted appropriately.
+        private const int POOL_SIZE = 8;
```
corranrogue9, line 26: Given how generic this class already is, it might make sense for this to be configurable through the constructor instead.

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// </summary>
+    /// <typeparam name="T">The type of objects to pool.</typeparam>
+    /// <remarks>This object pool is not thread-safe and not intended to be shared across concurrent requests.
+    /// Consider using the DefaultObjectPool in 8.x once we drop support for netstandard 1.1
+    /// </remarks>
+    internal class ObjectPool<T> where T : class
+    {
+        private readonly Func<T> objectGenerator;
+        private ObjectWrapper[] items;
+        private T firstItem;
```
corranrogue9, line 22: I'm not sure what benefit we get out of keeping track of the first item instead of keeping track of the index of the first item (maybe we are trying to remove one deference as an optimization?). Doing this without some other field means that we *must* use `null` to keep track of what slots in `items` are available, which means that `T` *must* be constrained to reference types. We also can't tell the difference between a pool that has loaned all of its objects, or a pool that hasn't generated any objects. This means that we won't notice when we have introduced a 9th level of nesting in a response, and we will start hitting perf issues at that point because the `Get` method will see `items` as only containing `null` values, so it will call `objectGenerator`, and the `Return` method will see `items` as being completely full and drop the returned object on the floor.

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,100 @@
...
+
+            for (int i = 0; i < items.Length; ++i)
+            {
+                if (items[i].Element == null)
+                {
+                    items[i].Element = obj;
+                    break;
+                }
+            }
+        }
```
corranrogue9, line 91: It's strange that if we get to the end without being able to return the object, nothing happens. The pool is fixed size, but there is nothing to the caller that indicates this, and if they *do* hit the limit, they will never notice because then we will just start generated new instances that leave scope and get garbage collected, removing the whole point of using the pool.

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// </summary>
+        /// <returns>An object of type <typeparamref name="T"/> from the pool.</returns>
+        /// <remarks>It's the responsibility of the caller to clean up the object before using it.</remarks>
+        public T Get()
+        {
+            T item = firstItem;
+
+            if (item != null)
+            {
+                firstItem = default(T);
```
corranrogue9, line 48: Is this some kind of optimization for when we are in the same nesting level so that we don't have to iterate over the loop many times?

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,100 @@
...
+namespace Microsoft.OData
+{
+    /// <summary>
+    /// Manages a pool of reusable objects. An object retrieved from the pool should be returned when it's no longer being used so that it can be later recycled.
+    /// </summary>
+    /// <typeparam name="T">The type of objects to pool.</typeparam>
+    /// <remarks>This object pool is not thread-safe and not intended to be shared across concurrent requests.
+    /// Consider using the DefaultObjectPool in 8.x once we drop support for netstandard 1.1
+    /// </remarks>
+    internal class ObjectPool<T> where T : class
```
corranrogue9, line 18: I might be trivializing this, but why isn't this the implementation:
```
public sealed class ObjectPool<T>
{
  private readonly Func<T> factory;
  private readonly T[] items;
  private readonly int next;
  private readonly numberOfObjectsCreated;

  public ObjectPool(Func<T> factory, int size)
  {
    this.factory = factory;
    this.items = new T[size];
    this.next = 0;
    this.numberOfObjectsCreated = 0;
  }

  public T Get()
  {
    // do we need to create new items?
    if (this.next == this.numberOfObjectsCreated)
    {
      // are we out of slots for new items?
      if (this.numberOfObjectsCreated > this.items.Length)
      {
        throw new Exception("no more available");
      }

      this.items[this.next] = this.factory();
      this.numberOfObjectsCreated++;
    }

    return this.items[this.next++];
  }

  public void Return()
  {
    this.next--;
  }
}
```

src/Microsoft.OData.Core/Json/JsonLightInstanceAnnotationWriter.cs
```diff
@@ -201,11 +201,13 @@ internal void WriteInstanceAnnotationsForError(ICollection<ODataInstanceAnnotati
...
             if (resourceValue != null)
             {
+                IDuplicatePropertyNameChecker duplicatePropertyNameChecker = this.valueSerializer.GetDuplicatePropertyNameChecker();
                 this.WriteInstanceAnnotationName(propertyName, name);
                 this.valueSerializer.WriteResourceValue(resourceValue,
                     expectedType,
                     treatLikeOpenProperty,
-                    this.valueSerializer.CreateDuplicatePropertyNameChecker());
+                    duplicatePropertyNameChecker);
+                this.valueSerializer.ReturnDuplicatePropertyNameChecker(duplicatePropertyNameChecker);
```
corranrogue9, line 210: This really makes me think we should do the `IDisposable` thing that someone else was talking about. Don't we want to return the checker even if `WriteResourceValue` throws an exception?

xuzhg (approved):

src/Microsoft.OData.Core/WriterValidator.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/Json/JsonLightInstanceAnnotationWriter.cs
KenitoInc: 1. `IDisposable` will make the `ObjectPool` be tightly coupled with the `DuplicatePropertyNameChecker`, and this is something we are trying to avoid.

src/Microsoft.OData.Core/ObjectPool.cs
KenitoInc: Yes

src/Microsoft.OData.Core/ObjectPool.cs
```diff
@@ -0,0 +1,100 @@
...
+                {
+                    items[i].Element = obj;
+                    break;
+                }
+            }
+        }
+
+        // PERF: the struct wrapper avoids array-covariance-checks from the runtime when assigning to elements of the array.
+        // See comment in this PR https://github.com/dotnet/extensions/pull/314#discussion_r169390645
+        private struct ObjectWrapper
```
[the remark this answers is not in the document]
KenitoInc: By avoiding runtime checks, we reduce CPU time.

src/Microsoft.OData.Core/ODataWriterCore.cs
KenitoInc: Yes, in the `EndResource(ODataResource resource)` method.

src/Microsoft.OData.Core/ODataCollectionWriterCore.cs
KenitoInc: Done

src/Microsoft.OData.Core/ODataParameterWriterCore.cs
KenitoInc: Done

src/Microsoft.OData.Core/ObjectPool.cs
habbes: I'm not sure if it's a good idea to throw an exception when we're out of slots. I think in that case we should either "resize" the array (which is maybe more costly than we want to pay) or create a new object that's not stored in the pool. If there's more demand than what can fit in the pool, then the excess is not pooled, but doesn't throw either. My 2 cents.

src/Microsoft.OData.Core/Json/JsonLightInstanceAnnotationWriter.cs
habbes: That's a good point about returning objects even when there are exceptions.

src/Microsoft.OData.Core/ObjectPool.cs
habbes: I think avoiding array allocations on responses with 1 level of nesting could be beneficial.  I don't think it's a problem to constrain `T` to reference types since we are trying to reduce heap allocations. If we were storing value types in the pool, new copies would be created when we call `Get()` anyway.

I think you make a good point about the hitting the 9th level of nesting. Though I don't understand what you meant by "`Return` method will see `items` as being completely full and drop the returned object on the floor". If `Return` sees `items` as full, it will replace one of the `null` slots with the items being returned.

What would you propose `Get()` do when hitting the max pool size? I feel like throwing an exception is too extreme.

src/Microsoft.OData.Core/ObjectPool.cs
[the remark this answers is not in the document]
habbes: Good point about calling `if (items == null)` on every `Get()` call seeming suboptimal. My suggestion was based on eliminating an array allocation (and deallocation) when it's not necessary. I would assume it would be much easier for JIT or CPU to optimize `if (items == null)` away via branch prediction or some other techniques if `items == null` evaluates to the same value most of the times. So in both cases there's trade-off to be made. But this lazy initialization pattern seems common in BCL source code, e.g.:
- https://source.dot.net/#System.Text.Json/System/Text/Json/Serialization/WriteStack.cs,95
- https://source.dot.net/#System.Text.Json/System/Text/Json/BitStack.cs,61
- https://source.dot.net/#System.Private.CoreLib/Dictionary.cs,506

Can this be evaluated using benchmarks?

src/Microsoft.OData.Core/ObjectPool.cs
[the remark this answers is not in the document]
habbes: Is there a noticeable CPU difference in profiler or benchmarks with and without `ObjectWrapper`?

src/Microsoft.OData.Core/ObjectPool.cs
habbes: I think when we hit the limit we have a number of options:
- throw an exception
- grow the pool
- do nothing

- I personally don't thin throwing an exception is a good idea. The object pool is just an optimization, an implementation detail. It shouldn't impose a limit on how much nesting the writer should support (the nesting limit is defined in the writer settings I believe, and the default is much higher, maybe 100).
- Doing nothing means the array does not need to be resized. It means for sufficiently shallow responses (we assume that's the majority of responses) the fixed size is good enough. And for more deeply nested responses, the performance would not be worse than the current implementation as far as allocations are concerned.
- Growing the pool means allocating a new array and copying over items, like `Array.Resize`. Maybe this in acceptable compromise for deeply-nested responses since the resizing is infrequent. What do you think about this? And if we allow resizing, would it make sense to start with a smaller array size (say 4 instead of 8)?

src/Microsoft.OData.Core/WriterValidator.cs
KenitoInc: Fixed

src/Microsoft.OData.Core/ObjectPool.cs
KenitoInc: Fixed

src/Microsoft.OData.Core/ObjectPool.cs
KenitoInc: I have modified the code:
- When an object cannot be returned to the array when the array is full, we dispose it
- When we are done with the object pool, it's disposed

src/Microsoft.OData.Core/ObjectPool.cs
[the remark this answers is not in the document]
corranrogue9: I think some sort of numbers is really necessary to justify introducing this complexity.

src/Microsoft.OData.Core/ObjectPool.cs
corranrogue9: I'm not sure why we ended up implementing `IDisposable` here, but I think this will ultimately be resolved by understanding what we really want the implementation to do (see the other open comment).

src/Microsoft.OData.Core/ObjectPool.cs
corranrogue9: I've consolidated my different feedback threads into this one because I feel like something is getting lost. 

1. The majority of my feedback revolves around the fact that we are not just using indices to track the next items. Doing so removes so much of the code, and seems to make the total number of operations consistent for each scenario. Particularly, notice how the implementation above is O(1), while the current implementation is O(n). There is a lot of extra code to try to make it mostly perform as O(1), but I don't think we should be optimizing an O(n) algorithm when an O(1) option is available. 
2. By being fixed size, we lose out on the optimization if the nesting level ever changes. I'm not actually proposing that we just throw an exception if the nesting level changes, and that the customer ends up perplexed. I'm saying that a fixed size `ObjectPool` should have in its contract that it throws when it runs out of space. We should catch this exception somewhere else (probably in the `ODataSerializer`). If we don't throw an exception and we increase the nesting level, we will never notice that we are losing out on this optimization. I think we should either use a dynamically sized object pool, or we should throw an exception that we surface in `Debug` configurations and don't surface in `Release` configurations. 

Ultimately, I think we should be using the .NET object pool implementation and we don't use an object pool for versions that don't have it. This should incentivize users to move to newer versions of .NET in order to receive these performance benefits.

src/Microsoft.OData.Core/Json/JsonLightInstanceAnnotationWriter.cs
corranrogue9: There's no coupling with `DuplicatePropertyNameChecker`. `ObjectPool<T>` could have a contract like:
```
public sealed class ObjectPool<T>
{
  public Lease<T> Get()
  {
    ...
  }

  private void Return(T value)
  {
    ...
  }

  public sealed class Lease<T> : IDisposable
  {
    public Lease(T value, ObjectPool<T> pool)
    {
      ...
    }

    public T Value { get; }
    public void Dispose()
    {
      this.pool.Return(this.Value);
    }
  }
}
```

Whether we actually use `IDisposable` or not, we should probably be wrapping the `ReturnDuplicatePropertyNameChecker` calls in a `finally` block (the equivalent of `using` on the proposed `Lease` type).

KenitoInc: Why do we need to return the object to the pool when we have an error? The writing process is already terminated. When we start writing a new response, a new ObjectPool will be created. The existing object pool will be cleaned up.

src/Microsoft.OData.Core/ObjectPool.cs
KenitoInc: @corranrogue9 
1. Your solution still doesn't the resolve the issue where we have a fixed size object pool. We shouldn't be throwing an exception.
2. My solution is flexible when the nesting is larger than the array size. We will still spin more objects. What will happen is when returning the objects to the pool, the extra objects will be disposed. The solution is modelled around the .Net object pool.
3. In the ObjectPool<T> docstring I have indicated we should switch to the DefaultObjectPool in 8.x once we drop support for `netstandard 1.1`. Having 2 types object pools in ODL now only adds extra complexity.

Conclusion: This PR was intended on to reduce heap allocation. At the same time not regress cpu time. Further optimizations can be done later.

corranrogue9: > Your solution still doesn't the resolve the issue where we have a fixed size object pool

It does. It throws an exception.

> We shouldn't be throwing an exception

Yes we should. We need to notice when we have made an unrelated change that causes us to go beyond the current nesting level and therefore lose out on this optimization. *Please address this concern that I have*: if we don't throw an exception, how do we, as a team, know that we have inadvertently increased the nesting level?

> My solution is flexible when the nesting is larger than the array size. We will still spin more objects. 

What is the benefit of doing this? From what I see, it only serves to hide potential bugs. 

> The solution is modelled around the .Net object pool.

The .NET implementation has an extra requirement of being thread-safe, meaning that `Interlocked.Exchange` is *imperative* to its functioning. Because of this, reference types are needed and keeping a reference to the next item is essential. Without that requirement, we can improve performance by using a different algorithm rather than by optimizing the .NET algorithm.

> In the ObjectPool docstring I have indicated we should switch to the DefaultObjectPool in 8.x once we drop support for netstandard 1.1

Please create an issue for this so that the work can be planned for and assigned

> Having 2 types object pools in ODL now only adds extra complexity

I am not proposing that we have 2. I am proposing that we don't ship this optimization for older versions of .NET. This will incentivize customers to use newer versions of .NET. Alternatively, we could simply take the .NET implementation and place it directly into our repo, referencing where it came from. Then, we can ship down-level, but have complete consistency across versions.

src/Microsoft.OData.Core/Json/JsonLightInstanceAnnotationWriter.cs
habbes: @corranrogue9 I like the `Lease<T>` idea, but I also think it introduces some implicit coupling. Now methods that return `DuplicatePropertyNameChecker` will have to change their signature to return `Lease<DuplicatePropertyNameChecker>`:
```c#
using Lease<DuplicatePropertyNameChecker> leasedChecker = ...GetDuplicateNameChecker();
DuplicateNameChecker duplicateNameChecker = leastedChecker.Value;
```

 Maybe this indirection is better than calling `Return()` manually. But there would probably still be cases where you can't use `using` and have to call `Dispose()` manually.

But I think Ken also makes a good point. We're not holding an external resource. The risk of not returning an object is relatively low, it will still be eventually garbage collected. And as Ken mentions, the lifetime of the pool is the request, so if a request fails, the pool and objects not yet returned will be eventually cleaned up.

What if we code that tries to verify correctness of the object pool at debug or test time. Maybe debug assertions that keep track of how many objects were created and returned and verified before the request is disposed. And test cases could also verify this. I think this avoids the need for `Lease<T>` or throwing exceptions when the nesting level of the request exceeds the size of the object pool.

src/Microsoft.OData.Core/ObjectPool.cs
habbes: I don't think it's a good idea to throw an exception, because that's implying that having responses with depths that exceed the object pool size is an error. The size of the object pool shouldn't impose a limit on the depth of the response (there's a separate setting for that).

If the goal of the exception is for sanity checks to ensure we are not exceeding the object pool size due a bug in our code rather than due to the depth of the response, maybe we could have debug messages or assertions when the object pool size is exceeded, when the number of returned objects is not equal the number of requested objects, etc.

Or did you mean that we conditionally throw exceptions based on `#if DEBUG` or something like that?

I prefer the simpler implementation provided by @corranrogue9, but I still think it's valuable to delay allocating the array until the second element is requested, that way we avoid unnecessary allocations for non-nested responses. Unless we can demonstrate that the cost of having the `if` statement that's always checked is worse.

Also, since the point of this PR is to reduce allocations (and GC), I think it's okay to only care about reference types.

src/Microsoft.OData.Core/Json/JsonLightInstanceAnnotationWriter.cs
habbes: The other concern I have with `Lease<T>` approach is that it will add an extra allocation each an object is requested from the pool. So you'll end up having as many `Lease<T>` as nested objects in the response, which is the exact thing this PR is trying to avoid.

src/Microsoft.OData.Core/ObjectPool.cs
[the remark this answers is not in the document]
KenitoInc: Removed the ObjectPool and used the `DefaultObjectPool` from dotnet extensions.

src/Microsoft.OData.Core/Json/JsonLightInstanceAnnotationWriter.cs
corranrogue9: I'm not too attached to the `Lease` idea, a note on the allocations is that a `Lease` *could* be a `struct` I think? I'm going to resolve though

corranrogue9 (dismissed):

src/Microsoft.OData.Core/WriterValidator.cs
```diff
@@ -29,15 +45,50 @@ internal WriterValidator(ODataMessageWriterSettings settings)
...
+#else
+                duplicatePropertyNameChecker = new DuplicatePropertyNameChecker();
+#endif
+            }
+            else
+            {
+                if (this.nullDuplicatePropertyNameChecker == null)
+                {
+                    // NullDuplicatePropertyNameChecker does nothing, so we can use a single instance during the writing process.
+                    this.nullDuplicatePropertyNameChecker = new NullDuplicatePropertyNameChecker();
```
habbes, line 73: Maybe we could even re-use a single instance across all requests. A singleton basically.

src/Microsoft.OData.Core/WriterValidator.cs
```diff
@@ -29,15 +45,50 @@ internal WriterValidator(ODataMessageWriterSettings settings)
...
+
+            return duplicatePropertyNameChecker;
+        }
+
+        /// <inheritdoc/>
+        public void ReturnDuplicatePropertyNameChecker(IDuplicatePropertyNameChecker duplicatePropertyNameChecker)
+        {
+#if NETSTANDARD2_0_OR_GREATER
+            // We only return the DuplicatePropertyNameChecker to the object pool and ignore the NullDuplicatePropertyNameChecker.
+            if (settings.ThrowOnDuplicatePropertyNames)
```
habbes, line 87: Curious, does this setting guarantee that we're dealing with `DuplicatePropertyNameChecker` and not `NullDuplicatePropertyNameChecker`?

src/Microsoft.OData.Core/DuplicatePropertyNameChecker.cs
```diff
@@ -15,12 +15,13 @@ namespace Microsoft.OData
     ///   2) No duplicate "@odata.associationLink" on a property.
     ///   3) "@odata.associationLink"s are put on navigation properties.
     /// </summary>
-    internal class DuplicatePropertyNameChecker : IDuplicatePropertyNameChecker
+    internal class DuplicatePropertyNameChecker : IDuplicatePropertyNameChecker, IDisposable
```
habbes, line 18: I haven't seen the `Dispose()` method or a `using` statement being used anywhere for objects of this type. How is the `IDisposable` applied for this type?

When getting the item from the pool, you're calling `duplicateNameChecker.Reset()` manually, which is the only thing that dispose seems to be doing. Not sure where else dispose is being applied.

src/Microsoft.OData.Core/DuplicatePropertyNameChecker.cs
```diff
@@ -122,6 +123,25 @@ public void Reset()
...
             propertyState.Clear();
         }
+
+        protected virtual void Dispose(bool disposing)
+        {
+            if (!disposedValue)
+            {
+                if (disposing)
+                {
+                    Reset();
```
habbes, line 133: Is it necessary to reset the duplicate checker if it's not going to be used again?

src/Microsoft.OData.Core/WriterValidator.cs
KenitoInc: Here we are creating it once for every writing cycle

src/Microsoft.OData.Core/DuplicatePropertyNameChecker.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/DuplicatePropertyNameChecker.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/WriterValidator.cs
KenitoInc: #Resolved

src/Microsoft.OData.Core/DuplicatePropertyNameChecker.cs
```diff
@@ -15,12 +15,13 @@ namespace Microsoft.OData
...
     ///   3) "@odata.associationLink"s are put on navigation properties.
     /// </summary>
-    internal class DuplicatePropertyNameChecker : IDuplicatePropertyNameChecker
+    internal class DuplicatePropertyNameChecker : IDuplicatePropertyNameChecker, IDisposable
     {
         /// <summary>
         /// Caches property processing state.
         /// </summary>
         private IDictionary<string, State> propertyState = new Dictionary<string, State>();
+        private bool disposedValue;
```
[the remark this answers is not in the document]
KenitoInc: #Resolved

habbes (approved):

odero (approved):

PR #2335, opened 2022-03-03
Implement asynchronous support in ODataJsonLightReader

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataJsonLightReader`**.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
@@ -4,22 +4,24 @@
 // </copyright>
 //---------------------------------------------------------------------
 
-using System;
-using System.IO;
-using System.Linq;
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.Diagnostics.CodeAnalysis;
-using System.Threading.Tasks;
-using Microsoft.OData.Evaluation;
-using Microsoft.OData.Json;
-using Microsoft.OData.Metadata;
-using Microsoft.OData.UriParser;
-using Microsoft.OData.Edm;
-using ODataErrorStrings = Microsoft.OData.Strings;
-
 namespace Microsoft.OData.JsonLight
 {
+    #region Namespaces
+
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.IO;
+    using System.Threading.Tasks;
+    using Microsoft.OData.Edm;
+    using Microsoft.OData.Evaluation;
+    using Microsoft.OData.Json;
+    using Microsoft.OData.Metadata;
+    using Microsoft.OData.UriParser;
+    using ODataErrorStrings = Microsoft.OData.Strings;
+
+    #endregion Namespaces
+
     /// <summary>
     /// OData reader for the JsonLight format.
     /// </summary>
@@ -170,7 +172,7 @@ namespace Microsoft.OData.JsonLight
             Scope currentScope = this.CurrentScope;
             if (this.jsonLightInputContext.Model.IsUserModel())
             {
-                var derivedTypeConstraints = this.jsonLightInputContext.Model.GetDerivedTypeConstraints(currentScope.NavigationSource);
+                IEnumerable<string> derivedTypeConstraints = this.jsonLightInputContext.Model.GetDerivedTypeConstraints(currentScope.NavigationSource);
                 if (derivedTypeConstraints != null)
                 {
                     currentScope.DerivedTypeValidator = new DerivedTypeValidator(currentScope.ResourceType, derivedTypeConstraints, "navigation source", currentScope.NavigationSource.Name);
@@ -190,7 +192,7 @@ namespace Microsoft.OData.JsonLight
         ///                 when reading a resource:  the first node of the first nested resource info value, null for a null expanded link or an end object
         ///                                         node if there are no navigation links.
         /// </remarks>
-        protected override Task<bool> ReadAtStartImplementationAsync()
+        protected override async Task<bool> ReadAtStartImplementationAsync()
         {
             Debug.Assert(this.State == ODataReaderState.Start, "this.State == ODataReaderState.Start");
             Debug.Assert(
@@ -205,14 +207,16 @@ namespace Microsoft.OData.JsonLight
             ODataPayloadKind payloadKind = this.ReadingResourceSet ?
                 this.ReadingDelta ? ODataPayloadKind.Delta : ODataPayloadKind.ResourceSet : ODataPayloadKind.Resource;
 
-            return this.jsonLightResourceDeserializer.ReadPayloadStartAsync(
+            await this.jsonLightResourceDeserializer.ReadPayloadStartAsync(
                 payloadKind,
                 propertyAndAnnotationCollector,
                 this.IsReadingNestedPayload,
-                /*allowEmptyPayload*/false)
-                .FollowOnSuccessWith(t => ResolveScopeInfoFromContextUrl())
-                .FollowOnSuccessWith(t =>
-                    this.ReadAtStartImplementationSynchronously(propertyAndAnnotationCollector));
+                allowEmptyPayload: false).ConfigureAwait(false);
+
+            ResolveScopeInfoFromContextUrl();
+
+            return await this.ReadAtStartInternalImplementationAsync(propertyAndAnnotationCollector)
+                .ConfigureAwait(false);
         }
 
         #endregion ReadAtStartImplementation
@@ -245,7 +249,7 @@ namespace Microsoft.OData.JsonLight
         /// </remarks>
         protected override Task<bool> ReadAtResourceSetStartImplementationAsync()
         {
-            return TaskUtils.GetTaskForSynchronousOperation<bool>(this.ReadAtResourceSetStartImplementationSynchronously);
+            return this.ReadAtResourceSetStartInternalImplementationAsync();
         }
 
         /// <summary>
@@ -281,7 +285,7 @@ namespace Microsoft.OData.JsonLight
         /// </remarks>
         protected override Task<bool> ReadAtResourceSetEndImplementationAsync()
         {
-            return TaskUtils.GetTaskForSynchronousOperation<bool>(this.ReadAtResourceSetEndImplementationSynchronously);
+            return this.ReadAtResourceSetEndInternalImplementationAsync();
         }
 
         #endregion ResourceSet
@@ -326,7 +330,7 @@ namespace Microsoft.OData.JsonLight
         /// </remarks>
         protected override Task<bool> ReadAtResourceStartImplementationAsync()
         {
-            return TaskUtils.GetTaskForSynchronousOperation<bool>(this.ReadAtResourceStartImplementationSynchronously);
+            return TaskUtils.GetTaskForSynchronousOperation<bool>(this.ReadAtResourceStartImplementation);
         }
 
         /// <summary>
@@ -354,7 +358,7 @@ namespace Microsoft.OData.JsonLight
         /// </remarks>
         protected override Task<bool> ReadAtResourceEndImplementationAsync()
         {
-            return TaskUtils.GetTaskForSynchronousOperation<bool>(this.ReadAtResourceEndImplementationSynchronously);
+            return this.ReadAtResourceEndInternalImplementationAsync();
         }
 
         #endregion Resource
@@ -381,9 +385,22 @@ namespace Microsoft.OData.JsonLight
         /// Pre-Condition:  JsonNodeType.PrimitiveValue         end of primitive value in a collection
         /// Post-Condition: The reader is positioned on the first node after the primitive value
         /// </remarks>
-        protected override Task<bool> ReadAtPrimitiveImplementationAsync()
+        protected override async Task<bool> ReadAtPrimitiveImplementationAsync()
         {
-            return TaskUtils.GetTaskForSynchronousOperation<bool>(this.ReadAtPrimitiveSynchronously);
+            Debug.Assert(
+                this.jsonLightResourceDeserializer.JsonReader.NodeType == JsonNodeType.PrimitiveValue,
+                "Pre-Condition: JsonNodeType.PrimitiveValue (null or untyped)");
+
+            this.PopScope(ODataReaderState.Primitive);
+
+            // Read over the end object node (or null value) and position the reader on the next node in the input.
+            // This should never hit the end of the input.
+            await this.jsonLightResourceDeserializer.JsonReader.ReadAsync()
+                .ConfigureAwait(false);
+            await this.ReadNextResourceSetItemAsync()
+                .ConfigureAwait(false);
+
+            return true;
         }
 
         #endregion Primitive
@@ -397,15 +414,6 @@ namespace Microsoft.OData.JsonLight
         protected override bool ReadAtNestedPropertyInfoImplementation()
         {
             return this.ReadAtNestedPropertyInfoSynchronously();
-        }
-
-        /// <summary>
-        /// Implementation of the reader logic when in state 'PropertyInfo'.
-        /// </summary>
-        /// <returns>A task which returns true if more items can be read from the reader; otherwise false.</returns>
-        protected override Task<bool> ReadAtNestedPropertyInfoImplementationAsync()
-        {
-            return TaskUtils.GetTaskForSynchronousOperation<bool>(this.ReadAtNestedPropertyInfoSynchronously);
         }
 
         #endregion
@@ -425,9 +433,28 @@ namespace Microsoft.OData.JsonLight
         /// Implementation of the reader logic when in state 'Stream'.
         /// </summary>
         /// <returns>A task which returns true if more items can be read from the reader; otherwise false.</returns>
-        protected override Task<bool> ReadAtStreamImplementationAsync()
+        protected override async Task<bool> ReadAtStreamImplementationAsync()
         {
-            return TaskUtils.GetTaskForSynchronousOperation<bool>(this.ReadAtStreamSynchronously);
+            this.PopScope(ODataReaderState.Stream);
+
+            if (this.State == ODataReaderState.ResourceSetStart ||
+                this.State == ODataReaderState.DeltaResourceSetStart)
+            {
+                // We are reading a stream within a collection
+                await this.ReadNextResourceSetItemAsync()
+                    .ConfigureAwait(false);
+
+                return true;
+            }
+
+            if (this.State == ODataReaderState.NestedProperty)
+            {
+                this.PopScope(ODataReaderState.NestedProperty);
+            }
+
+            // We are reading a stream value
+            return await this.ReadNextNestedInfoAsync()
+                .ConfigureAwait(false);
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -946,6 +946,33 @@ private T InterceptException<T>(Func<T> action)
...
+        /// </returns>
+        private async Task<T> InterceptExceptionAsync<T>(Func<ODataReaderCore, Task<T>> action)
+        {
+            try
+            {
+                return await action(this).ConfigureAwait(false);
+            }
+            catch (Exception e)
+            {
+                if (ExceptionUtils.IsCatchableExceptionType(e))
```
corranrogue9, line 967: I'm sure we don't care very much about optimizing the failure cases, but catching all exceptions to only then figure out if we can handle them makes exceptions that aren't expected more expensive

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightWriterTests.cs
```diff
@@ -50,7 +50,7 @@ private void InitializeEdmModel()
             this.model = new EdmModel();
 
             this.customerTypeEnumType = new EdmEnumType("NS", "CustomerType");
-            this.addressComplexType = new EdmComplexType("NS", "Address", /*baseType*/ null, /*isAbstract*/ true, /*isOpen*/ true);
+            this.addressComplexType = new EdmComplexType("NS", "Address", baseType: null, isAbstract: true, isOpen: true);
```
corranrogue9, line 53: Thanks for this!

corranrogue9:
I don't see anything bad that stands out in the change, but unfortunately I don't feel comfortable approving because I don't really have context. Hopefully someone else can approve.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2568,7 +2624,1548 @@ private void ResolveScopeInfoFromContextUrl()
...
+        /// A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains true if more items can be read from the reader; otherwise false.
+        /// </returns>
+        /// <remarks>
+        /// Pre-Condition:  JsonNodeType.None:      assumes that the JSON reader has not been used yet when not reading a nested payload.
+        /// Post-Condition: when reading a resource set:    the reader is positioned on the first item in the resource set or the end array node of an empty resource set
+        ///                 when reading a resource:  the first node of the first nested resource info value, null for a null expanded link or an end object
+        ///                                         node if there are no navigation links.
+        /// </remarks>
+        private async Task<bool> ReadAtStartImplementationAsynchronously(PropertyAndAnnotationCollector propertyAndAnnotationCollector)
```
xuzhg, line 2641: just use "Async" as suffix?

src/Microsoft.OData.Core/ODataReaderCoreAsync.cs
```diff
@@ -175,95 +177,206 @@ protected virtual Task<bool> ReadAtDeltaDeletedLinkImplementationAsync()
...
+
+        /// <summary>
+        /// Asynchronously creates a <see cref="Stream"/> for reading a stream property
+        /// when reader in state <see cref="ODataReaderState.Stream"/>.
+        /// </summary>
+        /// <returns>
+        /// A task that represents the asynchronous operation.
+        /// The value of the TResult parameter contains a <see cref="Stream"/> for reading the stream property.
+        /// </returns>
+        protected virtual Task<Stream> CreateReadStreamImplementationAsync()
```
xuzhg, line 295: make all as abstract methods?

don't we add 'async' identifier?

xuzhg (approved):

src/Microsoft.OData.Core/ODataReaderCore.cs
```diff
@@ -396,7 +396,7 @@ public override sealed bool Read()
         public override sealed Task<bool> ReadAsync()
         {
             this.VerifyCanRead(false);
-            return this.ReadAsynchronously().FollowOnFaultWith(t => this.EnterScope(new Scope(ODataReaderState.Exception, null, null)));
+            return this.InterceptExceptionAsync(thisParam => thisParam.ReadAsynchronously());
```
habbes, line 399: Is this and the old code functionally equivalent? How come you don't `EnterScope` here?

habbes: I've seen i in the `InterceptException` definition below

src/Microsoft.OData.Core/ODataReaderCoreAsync.cs
```diff
@@ -175,95 +177,206 @@ protected virtual Task<bool> ReadAtDeltaDeletedLinkImplementationAsync()
...
+        /// Asynchronously creates a <see cref="TextReader"/> for reading a string property
+        /// when reader in state <see cref="ODataReaderState.Stream"/>.
+        /// </summary>
+        /// <returns>
+        /// A task that represents the asynchronous operation.
+        /// The value of the TResult parameter contains a <see cref="TextReader"/> for reading the string property..
+        /// </returns>
+        protected virtual Task<TextReader> CreateTextReaderImplementationAsync()
+        {
+            throw new NotImplementedException();
```
habbes, line 310: Follow up to @xuzhg's [question](https://github.com/OData/odata.net/pull/2335/files#r821216779), will this throw the exception even when you call it synchronously without `await`, is that the intended behaviour?

xuzhg (approved):

src/Microsoft.OData.Core/ODataReaderCore.cs
gathogojr: @habbes That's correct. The `InterceptExceptionAsync` method will take care of that

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
gathogojr: @xuzhg @KenitoInc @habbes So, there's a reason why the `private` methods are suffixed with `Asynchronously` instead of `Async` - _and this goes even for the existing synchronous methods_. It's because the `protected` methods are already using the `Async` suffix. Where you see a `private` `ReadAtStartImplementationAsynchronously` method, there's a `protected` `ReadAtStartImplementationAsync` method. I just followed the pattern I found to avoid introducing confusion.

src/Microsoft.OData.Core/ODataReaderCoreAsync.cs
gathogojr: @KenitoInc For most of those methods you see declared as `public virtual` (especially in public classes or interfaces), its because they were introduced midway and it done that way to prevent a breaking change. The safest time to make those methods `abstract` is in the next major release.

gathogojr: @xuzhg We don't need to add `async` identifier for those `protected virtual` methods. The only purpose the `throw new NotImplementedException();` stub serves is to not force all classes that inherit from this class implement those new methods if there's no need to. `async` identifier or no, it'd be a bad day if any of the derived classes was to ship before a proper implementation for those methods has been supplied and any of those methods got called.

gathogojr: About making the methods `abstract`, in as much as this class is `internal`, I'm inclined to follow what appears to be the established practice going by the existing code - new methods in `abstract` classes are introduced as `virtual`.

src/Microsoft.OData.Core/ODataReaderCoreAsync.cs
gathogojr: @habbes See my response here https://github.com/OData/odata.net/pull/2335#discussion_r825012910

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2568,7 +2624,1548 @@ private void ResolveScopeInfoFromContextUrl()
...
+        ///                 JsonNodeType.PrimitiveValue (null)  Expanded null resource of the nested resource info to read next.
+        ///                 JsonNodeType.Property               The next property after a deferred link or entity reference link
+        ///                 JsonNodeType.EndObject              If no (more) properties exist in the resource's content
+        /// Post-Condition: JsonNodeType.StartObject            Start of the expanded resource of the nested resource info to read next.
+        ///                 JsonNodeType.StartArray             Start of the expanded resource set of the nested resource info to read next.
+        ///                 JsonNodeType.PrimitiveValue (null)  Expanded null resource of the nested resource info to read next.
+        ///                 JsonNodeType.Property               The next property after a deferred link or entity reference link
+        ///                 JsonNodeType.EndObject              If no (more) properties exist in the resource's content
+        /// </remarks>
+        private async Task<bool> ReadAtDeletedResourceStartImplementationAsynchronously()
```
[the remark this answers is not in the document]
gathogojr: @KenitoInc Here's the reason these methods in particular are suffixed with `Asynchronously` - https://github.com/OData/odata.net/pull/2335#discussion_r824999140

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightReaderTests.cs
```diff
@@ -0,0 +1,2039 @@
+﻿//---------------------------------------------------------------------
```
habbes, line 1: Do you have tests for when JsonReader does not support streaming?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -474,6 +474,66 @@ protected override TextReader CreateTextReaderImplementation()
...
+            if (streamReader != null)
+            {
+                stream = await streamReader.CreateReadStreamAsync()
+                    .ConfigureAwait(false);
+            }
+            else
+            {
+                // JsonReader doesn't support streaming; read as a string and convert
+                // Skip over property or start array
+                await this.jsonLightInputContext.JsonReader.ReadAsync()
```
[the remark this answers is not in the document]
gathogojr: @habbes If you can recall our session on OData JSON writers, you can substitute the inbuilt `JsonReader` by injecting your own custom reader. The custom reader may not necessarily inherit from `IJsonStreamReader` and `IJsonStreamReaderAsync`. It wouldn't be a simple effort but I can actually implement a custom reader and use that to test this code block. In any case I was planning to do that when making changes to `ODataJsonLightInputContext`.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2568,7 +2629,1554 @@ private void ResolveScopeInfoFromContextUrl()
...
+        /// <summary>
+        /// Asynchronously verifies that the current item is an <see cref="ODataNestedResourceInfo"/> instance,
+        /// sets the cardinality of the link (IsCollection property) and moves the reader
+        /// into state 'NestedResourceInfoEnd'.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous read operation.</returns>
+        /// <param name="isCollection">A flag indicating whether the link represents a collection or not.</param>
+        private Task ReadExpandedNestedResourceInfoEndAsync(bool isCollection)
+        {
+            return TaskUtils.GetTaskForSynchronousOperation(() => this.ReadExpandedNestedResourceInfoEnd(isCollection));
```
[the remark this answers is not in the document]
gathogojr: It so happens that the async equivalent, `ReadExpandedNestedResourceInfoEnd` doesn't actually make any synchronous reads. The reason I had created an equivalent async method is out of the fear that in the future someone might make changes to the synchronous method such it makes synchronous reads. I think for now I'll just drop the asynchronous equivalent

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -474,6 +474,66 @@ protected override TextReader CreateTextReaderImplementation()
...
+        /// Asynchronously creates <see cref="Stream"/> for reading an inline stream property.
+        /// </summary>
+        /// <returns>
+        /// A task that represents the asynchronous operation.
+        /// The value of the TResult parameter contains a <see cref="Stream"/> for reading the stream property.
+        /// </returns>
+        protected override async Task<Stream> CreateReadStreamImplementationAsync()
+        {
+            Stream stream;
+            IJsonStreamReaderAsync streamReader = this.jsonLightInputContext.JsonReader as IJsonStreamReaderAsync;
```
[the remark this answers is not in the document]
gathogojr: @habbes Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2568,7 +2624,1548 @@ private void ResolveScopeInfoFromContextUrl()
...
+
+                return true;
+            }
+
+            // read the next item in an untyped collection
+            await this.ReadNextResourceSetItemAsync()
+                .ConfigureAwait(false);
+
+            return true;
+        }
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2568,7 +2624,1548 @@ private void ResolveScopeInfoFromContextUrl()
...
+        {
+            ODataResourceBase currentResource = this.Item as ODataResourceBase;
+            if (currentResource != null && !this.IsReadingNestedPayload)
+            {
+                this.CurrentResourceState.ResourceTypeFromMetadata = this.ParentScope.ResourceType as IEdmStructuredType;
+                ODataResourceMetadataBuilder builder =
+                    this.jsonLightResourceDeserializer.MetadataContext.GetResourceMetadataBuilderForReader(
+                        this.CurrentResourceState,
+                        this.jsonLightInputContext.ODataSimplifiedOptions.EnableReadingKeyAsSegment,
+                        this.ReadingDelta);
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2568,7 +2624,1548 @@ private void ResolveScopeInfoFromContextUrl()
...
+
+            this.PopScope(this.State == ODataReaderState.ResourceEnd ? ODataReaderState.ResourceEnd : ODataReaderState.DeletedResourceEnd);
+
+            // Read over the end object node (or null value) and position the reader on the next node in the input.
+            // This can hit the end of the input.
+            await this.jsonLightResourceDeserializer.JsonReader.ReadAsync()
+                .ConfigureAwait(false);
+
+            // Analyze the next Json token to determine whether it is start object (next resource), end array (resource set end) or eof (top-level resource end)
+            bool result = true;
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2568,7 +2624,1548 @@ private void ResolveScopeInfoFromContextUrl()
...
+        ///                 JsonNoteType.Primitive (null)   expanded null resource nested resource info to read next
+        ///                 JsonNoteType.Property           property after deferred link or entity reference link
+        ///                 JsonNodeType.EndObject          end of the parent resource
+        /// </remarks>
+        private async Task<bool> ReadAtNestedPropertyInfoAsynchronously()
+        {
+            ODataPropertyInfo propertyInfo = this.CurrentScope.Item as ODataPropertyInfo;
+            Debug.Assert(propertyInfo != null, "Reading Nested Property Without an ODataPropertyInfo");
+
+            ODataStreamPropertyInfo streamPropertyInfo = propertyInfo as ODataStreamPropertyInfo;
```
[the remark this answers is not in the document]
gathogojr: Fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2568,7 +2629,1554 @@ private void ResolveScopeInfoFromContextUrl()
...
+                // Next nested resource info on the resource
+                this.StartNestedResourceInfo(readerNestedResourceInfo);
+            }
+            else
+            {
+                ODataJsonLightReaderNestedPropertyInfo readerNestedStreamInfo = nestedInfo as ODataJsonLightReaderNestedPropertyInfo;
+                Debug.Assert(readerNestedStreamInfo != null, "NestedInfo is not a resource, stream, string");
+                if (readerNestedStreamInfo != null)
+                {
+                    await this.StartNestedPropertyInfoAsync(readerNestedStreamInfo)
```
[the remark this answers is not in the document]
gathogojr: Fixed following out offline discussion

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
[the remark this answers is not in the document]
gathogojr: Fixed following our offline discussion

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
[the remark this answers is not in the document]
gathogojr: @habbes Like we observed, this could easily be dead code. We'll revisit this at a later date to either remove the block altogether or to make changes to make the code reachable

src/Microsoft.OData.Core/ODataReaderCore.cs
gathogojr: @corranrogue9 What we do in `IsCatchableExceptionType` method is to check that we're not dealing with an `OutOfMemoryException`. We do a little optimization in that method by storing the `OutOfMemoryException` type in a static field

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2400,8 +2472,6 @@ private void StartNestedResourceInfo(ODataJsonLightReaderNestedResourceInfo read
         private void StartNestedPropertyInfo(ODataJsonLightReaderNestedPropertyInfo readerNestedPropertyInfo)
         {
             Debug.Assert(readerNestedPropertyInfo != null, "readerNestedResourceInfo != null");
-            Debug.Assert(this.jsonLightResourceDeserializer.JsonReader.CanStream() || this.CurrentScope is JsonLightResourceSetScope,
```
gathogojr, line 2403: Sacrificed this debug assert statement. Otherwise I'd need to support 2 variants; the sync variant that calls `CanStream` method and the async variant that calls `CanStreamAsync`. That would mean adding an equivalent `StartNestedPropertyInfoAsync` method on account of a debug assert statement!

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -2413,8 +2483,6 @@ private void StartNestedPropertyInfo(ODataJsonLightReaderNestedPropertyInfo read
         private void StartNestedStreamInfo(ODataJsonLightReaderStreamInfo readerStreamInfo)
         {
             Debug.Assert(readerStreamInfo != null, "readerNestedResourceInfo != null");
-            Debug.Assert(this.jsonLightResourceDeserializer.JsonReader.CanStream() || this.CurrentScope is JsonLightResourceSetScope,
```
gathogojr, line 2416: Explanation for this change provided [here](https://github.com/OData/odata.net/pull/2335/files#r826681344)

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -399,15 +416,6 @@ protected override bool ReadAtNestedPropertyInfoImplementation()
             return this.ReadAtNestedPropertyInfoSynchronously();
         }
 
-        /// <summary>
```
gathogojr, line 402: No need for this since the implementation in the base class mirrors this, i.e.,
```csharp
/// <summary>
/// Implementation of the reader logic when in state 'PropertyInfo'.
/// </summary>
/// <returns>A task which returns true if more items can be read from the reader; otherwise false.</returns>
protected virtual Task<bool> ReadAtNestedPropertyInfoImplementationAsync()
{
    return TaskUtils.GetTaskForSynchronousOperation<bool>(this.ReadAtNestedPropertyInfoImplementation);
}
```

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -695,27 +788,6 @@ protected override bool ReadAtDeletedResourceStartImplementation()
             return this.ReadAtDeletedResourceStartImplementationSynchronously();
         }
 
-        /// <summary>
```
gathogojr, line 698: No need for this since the implementation in the base class mirror this, i.e.,
```csharp
protected virtual Task<bool> ReadAtDeletedResourceStartImplementationAsync()
{
    return TaskUtils.GetTaskForSynchronousOperation<bool>(this.ReadAtDeletedResourceStartImplementation);
}
```

KenitoInc (approved):

habbes (approved):

PR #2341, opened 2022-03-08
Validate key not defined multiple times in type hierarchy

### LINKED ISSUE #2340: Edm Validation Rule EntityTypeInvalidKeyKeyDefinedInBaseClass only checks direct ancestor
EntityTypeInvalidKeyKeyDefinedInBaseClass is intended to make sure that there is no key defined in a type if it is already defined in a base type. However, the rule only checks the direct base type, not more distant ancestors, so if a base type has a key, and interim type doesn't define a key, but a type derived from the interim type does define a key, the rule doesn't catch it.

### Assemblies affected

*Which assemblies and versions are known to be affected Microsoft.OData.Edm.

### Reproduce steps

Define a simple schema with a base type that defines a key, an interim type, and a derived type that defines a key:

~~~xml
<edmx:Edmx xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx" Version="4.0">
	<edmx:DataServices>
		<Schema xmlns="http://docs.oasis-open.org/odata/ns/edm" Namespace="Self">
			<EntityType Name="BaseType">
				<Key>
					<PropertyRef Name="id"/>
				</Key>
			</EntityType>
			<EntityType Name="InterimType" BaseType="Self.BaseType"></EntityType>
			<EntityType Name="DerivedType" BaseType="Self.InterimType">
				<Key>
					<PropertyRef Name="id2"/>
				</Key>
				<Property Name="id2" Type="Edm.Int32" Nullable="false"/>
			</EntityType>
		</Schema>
	</edmx:DataServices>
</edmx:Edmx>
~~~

Load and validate the schema:

~~~c#
            IEdmModel model;
            IEnumerable<EdmError> errors;
            XmlReader reader = XmlReader.Create(new FileStream("DuplicateKeys.xml", FileMode.Open));
            if (!CsdlReader.TryParse(reader, out model, out errors))
            {
                writeErrors(errors);
                return;
            }
            Console.WriteLine("Loaded Model");
            if (!model.Validate(out errors))
            {
                writeErrors(errors);
                return;
            }
~~~

### Expected result

An error saying that the key could not be redefined.

### Actual result

No error returned.

### Additional detail

EntityTypeInvalidKeyKeyDefinedInBaseClass only checks direct basetype for key, not nested ancestors:

~~~c#
        public static readonly ValidationRule<IEdmEntityType> EntityTypeInvalidKeyKeyDefinedInBaseClass =
            new ValidationRule<IEdmEntityType>(
                (context, entityType) =>
                {
                    if (entityType.BaseType != null &&
                        entityType.DeclaredKey != null &&
                        entityType.BaseType.TypeKind == EdmTypeKind.Entity &&
                        entityType.BaseEntityType().DeclaredKey != null)
                    {
                        context.AddError(
                        entityType.Location(),
                        EdmErrorCode.InvalidKey,
                        Strings.EdmModel_Validator_Semantic_InvalidKeyKeyDefinedInBaseClass(entityType.Name, entityType.BaseEntityType().Name));
                    }
                });
~~~

mikepizzo:
### Issues

*This pull request fixes #2340.*

### Description

Exiting rule for validating that a key is not redefined only checks immediate base type. Fix makes sure that key is only defined once in the hierarchy.

For backward compatibility, this is added as a new rule, not part of the default ruleset, in order to ensure we don't break apps whose current schema validates but violate this rule.  The existing rule is marked as deprecated, and in 8.x. we should remove the old rule and make this rule part of the default ruleset.

Also marked an existing rule as obsolete because it is no longer valid due to a change in protocol. This was already removed from the default ruleset, but I added a comment explaining why and that it should also be removed in 8.x.

- [ x ] *Test cases added*
- [ x ] *Build and test with one-click build and test script passed*

### Additional work necessary

In 8.x, remove the deprecated rules and update the ruleset accordingly.

### DIFF
--- a/src/Microsoft.OData.Edm/Validation/ValidationRules.cs
+++ b/src/Microsoft.OData.Edm/Validation/ValidationRules.cs
@@ -924,28 +924,62 @@ namespace Microsoft.OData.Edm.Validation
                     }
                 });
 
+        // Remove this in 8.x. It is superseded by EntityTypeInvalidKeyKeyDefinedInAncestor
         /// <summary>
         /// Validates that a key is not defined if there is already a key in the base type.
         /// </summary>
+        [Obsolete("Please use EntityTypeInvalidKeyKeyDefinedInAncestor")]
         public static readonly ValidationRule<IEdmEntityType> EntityTypeInvalidKeyKeyDefinedInBaseClass =
+            new ValidationRule<IEdmEntityType>(
+               (context, entityType) =>
+               {
+                   if (entityType.BaseType != null &&
+                       entityType.DeclaredKey != null &&
+                       entityType.BaseType.TypeKind == EdmTypeKind.Entity &&
+                       entityType.BaseEntityType().DeclaredKey != null)
+                   {
+                       context.AddError(
+                       entityType.Location(),
+                       EdmErrorCode.InvalidKey,
+                       Strings.EdmModel_Validator_Semantic_InvalidKeyKeyDefinedInBaseClass(entityType.Name, entityType.BaseEntityType().Name));
+                   }
+               });
+
+        /// <summary>
+        /// Validates that a key is not more than once in a type hierarchy
+        /// </summary>
+        public static readonly ValidationRule<IEdmEntityType> EntityTypeInvalidKeyKeyDefinedInAncestor =
             new ValidationRule<IEdmEntityType>(
                 (context, entityType) =>
                 {
-                    if (entityType.BaseType != null &&
-                        entityType.DeclaredKey != null &&
-                        entityType.BaseType.TypeKind == EdmTypeKind.Entity &&
-                        entityType.BaseEntityType().DeclaredKey != null)
+                    bool foundKey = entityType.DeclaredKey != null;
+                    IEdmEntityType baseType = entityType;
+                    while ((baseType = baseType.BaseEntityType()) != null)
                     {
-                        context.AddError(
-                        entityType.Location(),
-                        EdmErrorCode.InvalidKey,
-                        Strings.EdmModel_Validator_Semantic_InvalidKeyKeyDefinedInBaseClass(entityType.Name, entityType.BaseEntityType().Name));
+                        if (baseType.DeclaredKey != null)
+                        {
+                            if(foundKey)
+                            {
+                                context.AddError(
+                                entityType.Location(),
+                                EdmErrorCode.InvalidKey,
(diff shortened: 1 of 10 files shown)


src/Microsoft.OData.Edm/Validation/ValidationRules.cs
```diff
@@ -924,28 +924,62 @@ public static class ValidationRules
...
             new ValidationRule<IEdmEntityType>(
                 (context, entityType) =>
                 {
-                    if (entityType.BaseType != null &&
-                        entityType.DeclaredKey != null &&
-                        entityType.BaseType.TypeKind == EdmTypeKind.Entity &&
-                        entityType.BaseEntityType().DeclaredKey != null)
+                    bool foundKey = entityType.DeclaredKey != null;
+                    IEdmEntityType baseType = entityType;
+                    while ((baseType = baseType.BaseEntityType()) != null)
```
corranrogue9, line 957: Will we have already checked for cycles in the inheritance tree before using this function?

corranrogue9 (approved):

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
mikepizzo: That's a really good question --  turns out that yes, we do have code in Validate() to protect against circular references in type hierarchies.

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
```diff
@@ -924,28 +924,62 @@ public static class ValidationRules
...
+                       entityType.Location(),
+                       EdmErrorCode.InvalidKey,
+                       Strings.EdmModel_Validator_Semantic_InvalidKeyKeyDefinedInBaseClass(entityType.Name, entityType.BaseEntityType().Name));
+                   }
+               });
+
+        /// <summary>
+        /// Validates that a key is not more than once in a type hierarchy
+        /// </summary>
+        public static readonly ValidationRule<IEdmEntityType> EntityTypeInvalidKeyKeyDefinedInAncestor =
```
xuzhg, line 951: Why double "key" in the naming?

xuzhg (approved):

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
mikepizzo: I followed the convention of the previous test. I believe it's intended to be parsed with a space between the two keys, as in: "InvalidKey, KeyDefinedInAncestor"

PR #2346, opened 2022-03-09
Set default scale to null

habbes:
### Issues

This PR partially addresses this issue: https://github.com/OData/WebApi/issues/2384

https://github.com/OData/WebApi/pull/2420

### Description

At the moment EdmLib sets the default scale for decimal types to 0. This is problematic because a scale of 0 technically means there shouldn't be any digits after the decimal point. The following PR adds support for scale in WebApi: https://github.com/OData/WebApi/issues/2384 it does so by rounding values that have more decimal digits than the scale to fit the scale. If 0 is the default scale, that update will cause all decimal values where scale is not explicitly defined to be truncated.

I'm not sure about this fix because it could be a breaking change. But introducing the WebApi PR without this fix would potentially be a bigger breaking change. Also, it seems that the scale was set to 0 intentionally, given that the default `Precision` is `null` in the codebase.

- [x] *Test cases added*

### Additional work necessary


habbes:
Actually, the [spec](http://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_Scale) says that if scale is not set, it should default to 0.

mikepizzo:
Although the xml spec says that scale not set should be 0, the newer JSON CSDL spec says not set scale should be variable. This was intentional, as variable reflects the more common behavior, including what we currently implement in our libraries.  In the OASIS TC we are debating how to relax this wording and reflect the realities and expected behavior based on current implementations.

So, we should accept this change and not default scale to zero if not specified.

habbes:
@mikepizzo  thanks for the clarifications.

On the last part:

    However, when we write out CSDL, we should not add the Scale="variable" if it wasn't specified when we read in the CSDL.

Should I interpret to also mean that if we read a CSDL where Scale is explicitly set to `variable`, when serialize the parsed the CSDL we should explicitly write `Scale="variable"` instead of omitting it as the default?

habbes:
> @mikepizzo thanks for the clarifications.
> 
> On the last part:
> 
> ```
> However, when we write out CSDL, we should not add the Scale="variable" if it wasn't specified when we read in the CSDL.
> ```
> 
> Should I interpret to also mean that if we read a CSDL where Scale is explicitly set to `variable`, when serialize the parsed the CSDL we should explicitly write `Scale="variable"` instead of omitting it as the default?
> 
> If this is the case, my initial thought would be to add a flag to `IEdmDecimalTypeReference` to determine whether a variable scale should be explicitly written. When reading a csdl with an explicit `Scale="variable"`, this will be set to true. In all other cases it will be false. When writing out an xml csdl document, if this flag is set to true and the scale is `null`, then `Scale="variable"` will be written out explicitly. This approach will mean adding a new property to a public interface, which is potentially also a breaking change, and refactoring some code with (potentially over-specific) `if` checks.
> 
> I am hesitant to modify public interfaces for this change. Any thoughts?

After implementing the approach above, I am reluctant to push that commit because not only does it change the public API, but it also touches so many files and methods for what is a very specific and focused feature that may or may not be widely used.

I have another alternative, a setting on the CSDL writer to enable writing default attributes. When this opt-in feature is enabled, the writer will write all attributes that have a default value instead of omitting them. This means `Scale="variable"` will also be written. I haven't tried this yet but I assume it will be cleaner and less brittle. The downside is that if the user is reading a CSDL with `Scale="variable"`, they will have to manually enable that setting on the writer so that the serialized CSDL also includes `Scale="variable"`. The other downside is that this will affect attributes other than scale or default scale attributes which were not set to `variable` explicitly. I could make the setting more granular and just target scale attributes, or allow the user to decide which attributes it should apply to. Maybe that would adding unnecessary complexity.

Any thoughts?

habbes:
> After implementing the approach above, I am reluctant to push that commit because not only does it change the public API, but it also touches so many files and methods for what is a very specific and focused feature that may or may not be widely used.

I have implemented a less disruptive alternative: a `ShouldWriteDefaultScale` has been added to the `CsdlSemanticsDecimalTypeReference` specific implementation which is internal and not to the public `IEdmDecimalTypeReference`. When reading the CSDL, a flag is returned indicating whether the `Scale` attribute was specifically set to `variable`. If this is true, then `ShouldWriteDefaultScale` is set to true. When writing the XML CSDL, the writer will check whether the `IEdmDecimalTypeReference` is a `CsdlSemanticsDecimalTypeReference` and whether the `ShouldWriteDefaultScale` to true, if the condition passes, then it will explicitly write the scale even when it's the default value.

I'm not very fond of the runtime type check and the condition that will now be evaluated every time you write a decimal property, even though the condition will probably be false most of the time. Hopefully it's optimized away by the processor. Also not fond of the very specific flag that has to be passed as an `out` variable to a bunch of methods until it reaches `CsdlSemanticsDecimalTypeReference`. But this is the cleanest, least disruptive implementation I could think of.

If it's not a requirement that we should write `Scale="variable"`, then I'd be happy to undo all these extra steps. cc @mikepizzo

mikepizzo:
> > After implementing the approach above, I am reluctant to push that commit because not only does it change the public API, but it also touches so many files and methods for what is a very specific and focused feature that may or may not be widely used.
> 
> I have implemented a less disruptive alternative: a `ShouldWriteDefaultScale` has been added to the `CsdlSemanticsDecimalTypeReference` specific implementation which is internal and not to the public `IEdmDecimalTypeReference`. When reading the CSDL, a flag is returned indicating whether the `Scale` attribute was specifically set to `variable`. If this is true, then `ShouldWriteDefaultScale` is set to true. When writing the XML CSDL, the writer will check whether the `IEdmDecimalTypeReference` is a `CsdlSemanticsDecimalTypeReference` and whether the `ShouldWriteDefaultScale` to true, if the condition passes, then it will explicitly write the scale even when it's the default value.
> 
> I'm not very fond of the runtime type check and the condition that will now be evaluated every time you write a decimal property, even though the condition will probably be false most of the time. Hopefully it's optimized away by the processor. Also not fond of the very specific flag that has to be passed as an `out` variable to a bunch of methods until it reaches `CsdlSemanticsDecimalTypeReference`. But this is the cleanest, least disruptive implementation I could think of.
> 
> If it's not a requirement that we should write `Scale="variable"`, then I'd be happy to undo all these extra steps. cc @mikepizzo

To preserve backward compatibility for 7.x, I think this is the right approach. Perhaps we can revisit in 8.x.

src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlDecimalTypeReference.cs
```diff
@@ -11,11 +11,29 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Ast
...
         }
+
+        /// <summary>
+        /// Specifies whether the default scale should be written out explicitly
+        /// when the model is being serialized into a CSDL (XML) document.
+        /// </summary>
+        /// <remarks>
+        /// Usually the attribute is not written when it contains the default value.
+        /// This flag is only here for backwards compatibility:
+        /// Up to v7.10.0, the default scale was 0. This is consistent with the XML CSDL spec (but not JSON spec),
```
gathogojr, line 29: `(but not JSON spec)` - Not sure I understand how writing a default scale of 0 was inconsistent with JSON specification...

habbes: In CSDL JSON the default scale is null, not 0.

gathogojr: argh, so you meant CSDL JSON, not JSON...

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaJsonWriter.cs
```diff
@@ -463,7 +463,7 @@ internal override void WriteDecimalTypeAttributes(IEdmDecimalTypeReference refer
             // The value of $Scale is a number or a string with one of the symbolic values floating or variable.
             // TODO: the symbolic values floating or variable is not supported now.
             // Absence of $Scale means variable.
-            this.jsonWriter.WriteOptionalProperty("$Scale", reference.Scale, CsdlConstants.Default_Scale);
+            this.jsonWriter.WriteOptionalProperty("$Scale", reference.Scale);
```
KenitoInc, line 466: Why this change?
Have we stopped writing `scale="variable"`?

KenitoInc: Why don't you introduce the condition you introduced in the xml writer?

KenitoInc (commented):
A few comments...
Note: Also create an issue to cleanup this in 8.x

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsDecimalTypeReference.cs
```diff
@@ -27,5 +27,22 @@ public CsdlSemanticsDecimalTypeReference(CsdlSemanticsSchema schema, CsdlDecimal
...
+        /// </summary>
+        /// <remarks>
+        /// Usually the attribute is not written when it contains the default value.
+        /// This flag is only here for backwards compatibility:
+        /// Up to v7.10.0, the default scale was 0. This is consistent with the XML CSDL spec (but not JSON spec),
+        /// see https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_Scale
+        /// However, later versions changed the default scale from 0 to null:
+        /// see PR: https://github.com/OData/odata.net/pull/2346
+        /// When we parse an XML CSDL document that contains an explicit Scale="variable" into an IEdmModel,
+        /// we want to ensure that serializing the IEdmModel back into CSDL will also explicitly write
```
gathogojr, line 43:
```suggestion
        /// we want to ensure that serializing the Edm model back into CSDL will also explicitly write
```

src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlDecimalTypeReference.cs
```diff
@@ -11,11 +11,29 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Ast
...
+        /// see https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_Scale
+        /// However, later versions changed the default scale from 0 to null:
+        /// see PR: https://github.com/OData/odata.net/pull/2346
+        /// When we parse an XML CSDL document that contains an explicit Scale="variable" into an IEdmModel,
+        /// we want to ensure that serializing the IEdmModel back into CSDL will also explicitly write
```
gathogojr, line 34:
```suggestion
        /// we want to ensure that serializing the Edm model back into CSDL will also explicitly write
```

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaJsonWriter.cs
habbes: Hmm those are good questions.

The reason for this change was that `this.WriteOptionalProperty` overload that was used before accepted a default value as the last argument. It is implemented as follows: 
```c#
public static void WriteOptionalProperty(this Utf8JsonWriter writer, string name, int? value, int defaultValue)
{
    EdmUtil.CheckArgumentNull(writer, nameof(writer));

    if (value != null && value.Value != defaultValue)
    {
        writer.WritePropertyName(name);
        writer.WriteNumberValue(value.Value);
    }
}
```
Now, the `CsdlConstants.Default_Scale` is no longer an `int` but a nullable `int?`, and so can't be passed to this overload. But since the default scale is `null`, it's not necessary to use this overload, we can use the following one:
```c#
public static void WriteOptionalProperty(this Utf8JsonWriter writer, string name, int? value)
{
    EdmUtil.CheckArgumentNull(writer, nameof(writer));

    if (value != null)
    {
        writer.WritePropertyName(name);
        writer.WriteNumberValue(value.Value);
    }
}
```

Since the default scale is `null`, the behaviour of both methods is the same because they both write nothing if the input `value` is `null` regardless of what the default value is. So this change does not affect the existing behaviour.

habbes: To answer the other question, why I didn't introduce the condition I introduced in the XML writer (i.e. to check whether `ShouldWriteDefaultScale` is set to true) is because prior to this PR, the default behaviour when writing scale was different between XML and JSON CSDL. In XML 0 was the default scale, and having a null scale would end up writing it to the XML. In JSON, null was the default scale and it was not written out to the CSDL document. So setting the default scale to null would not have any breaking changes for JSON CSDLs, it only affects XML CSDL.

I'm pushing new commits which change the XML logic to always write the scale value regardless of default values, this would have even fewer chances of breaking changes.

habbes:
Following an offline discussion with @mikepizzo the conclusion was to always write the scale attribute to XML CSDL regardless of the default. This is still valid with respect to the standard, which says that implementations *may* omit the scale attribute when it's default, but are not required to omit it.

It also avoids breaking expectations of existing customers who expect `Scale="Variable"` to be written out when the scale is null, and whose automated tests assert these expectations.

KenitoInc (approved):
Am happy with this change.
Approved!

corranrogue9 (approved):

PR #2349, opened 2022-03-10
added script to increment version number and update PublicAPI.Shipped…

corranrogue9:
….txt files when doing a release

### Issues

N/A

### Description

Added a script that increments the version number according to semantic versioning and merges the `PublicAPI.Unshipped.txt` changes into the associated `PublicAPI.Shipped.txt` file. 

- [ ] *Test cases added*

### Additional work necessary

There's plenty more work needed to automated releases. It is not currently planned.

### DIFF
--- a/tools/IncrementVersion.ps1
+++ b/tools/IncrementVersion.ps1
@@ -1,0 +1,58 @@
+$versionPath = "$PSScriptRoot\CustomMSBuild\Versioning.props"
+
+$lastReleaseCommit = git log -n 1 --pretty=format:%H -- $versionPath
+Write-Host "The last release's commit was $lastReleaseCommit"
+
+$changedFiles = git diff --name-only HEAD $lastReleaseCommit
+
+$apiChanges = $false
+ForEach ($line in $($changedFiles -split "`r`n"))
+{
+  $fileName = [System.IO.Path]::GetFileName($line)
+  if ($fileName -eq "PublicAPI.Shipped.txt")
+  {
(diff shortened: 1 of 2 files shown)


tools/IncrementVersion.ps1
```diff
@@ -0,0 +1,58 @@
+$versionPath = "$PSScriptRoot\CustomMSBuild\Versioning.props"
+
+$lastReleaseCommit = git log -n 1 --pretty=format:%H -- $versionPath
```
xuzhg, line 3: It looks good so far, and we should make sure there's only version changing in this file.

tools/IncrementVersion.ps1
```diff
@@ -0,0 +1,58 @@
...
+Write-Host "The last release's commit was $lastReleaseCommit"
+
+$changedFiles = git diff --name-only HEAD $lastReleaseCommit
+
+$apiChanges = $false
+ForEach ($line in $($changedFiles -split "`r`n"))
+{
+  $fileName = [System.IO.Path]::GetFileName($line)
```
xuzhg, line 11: why can't compare the content between Shipped and unShipped?

xuzhg (approved):

tools/IncrementVersion.ps1
corranrogue9: I don't want this script to need to know all of the projects that we release in order to do the task of incrementing version numbers. I suppose that it could navigate the directory structure to find the files with these names, but generally speaking there are far fewer files changed in our commits between releases than there are files in our repo, so I'd rather use this.

PR #2351, opened 2022-03-15
Implement asynchronous support in ODataJsonLightDeltaReader

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataJsonLightDeltaReader`**.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeltaReader.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeltaReader.cs
@@ -8,11 +8,9 @@ namespace Microsoft.OData.JsonLight
 {
     #region Namespaces
 
-    using System;
     using System.Diagnostics;
     using System.Threading.Tasks;
     using Microsoft.OData.Edm;
-    using Microsoft.OData.Metadata;
 
     #endregion Namespaces
 
@@ -44,7 +42,14 @@ namespace Microsoft.OData.JsonLight
             IEdmNavigationSource navigationSource,
             IEdmEntityType expectedEntityType)
         {
-            this.underlyingReader = new ODataJsonLightReader(jsonLightInputContext, navigationSource, expectedEntityType, /*readingResourceSet*/ true, /*readingParameter*/ false, /*readingDelta*/ true, /*listener*/ null);
+            this.underlyingReader = new ODataJsonLightReader(
+                jsonLightInputContext,
+                navigationSource,
+                expectedEntityType,
+                readingResourceSet: true,
+                readingParameter: false,
+                readingDelta: true,
+                listener: null);
         }
 
         #endregion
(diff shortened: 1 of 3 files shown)


gathogojr:
**NOTE:** Changes to _test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightReaderTests.cs_ consist of improvements to make the logic terse and remove superfluous `async` and `await` usages

marabooy (approved):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeltaReader.cs
```diff
@@ -166,36 +171,32 @@ public override bool Read()
...
+                .ConfigureAwait(false);
+
+            if (this.underlyingReader.State == ODataReaderState.DeletedResourceStart)
             {
-                if (this.underlyingReader.State == ODataReaderState.DeletedResourceStart)
+                while ((result = await this.underlyingReader.ReadAsync().ConfigureAwait(false)) && this.underlyingReader.State != ODataReaderState.DeletedResourceEnd)
                 {
-                    this.SkipToDeletedResourceEnd();
+                    // Skip to end of a deleted resource
+                    this.SetNestedLevel();
```
habbes, line 188: Why do we call `this.SetNestedLevel()` here? In the existing version, the `SkipToDeletedResourceEnd()` method does not call `this.SetNestedLevel()`

gathogojr: @habbes The previous implementation was unwieldy/buggy starting with the void async method (`private async void SkipToDeletedResourceEnd()`) that made it not awaitable - HIGHLY discouraged. If you look at the revised implementation it mirrors the synchronous implementation and I added tests to verify behaviour unlike before.

gathogojr: Here's how the synchronous `Read` is implemented:
```csharp
        public override bool Read()
        {
            // Translate state transitions to legacy states
            bool response = this.underlyingReader.Read();
            if (this.underlyingReader.State == ODataReaderState.DeletedResourceStart)
            {
                while ((response = this.underlyingReader.Read()) && this.underlyingReader.State != ODataReaderState.DeletedResourceEnd)
                {
                    // skip to end of a deleted resource
                    this.SetNestedLevel();
                }
            }
            this.SetNestedLevel();
            return response;
        }
```

habbes: @gathogojr does that mean the existing `ReadAsync()` implementation has a bug? Since it seems to be different the `Read()`.

habbes (dismissed):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeltaReader.cs
gathogojr: @habbes That would be my bet. And it would have gone unnoticed due to reason of async API being rarely used. If you look at the test file, the commented tests I got rid of were for the async API

gathogojr: > What is the `SkipToDeletedResourceEnd()` method supposed to do?

It was a poorly designed asynchronous method (`void` return type) that would call the `ReadAsync` method of the underlying reader (the `ODataJsonLightReader` in this case) and itself recursively until it gets to the end of the deleted resource.

habbes (approved):

KenitoInc (approved):

PR #2354, opened 2022-03-17
Expose more information about path segments and selected items

ificator:
### Issues

Improve `ODataPath` and `SelectExpandClause` so more information is exposed through the abstractions.

### Description

This small PR contains a few changes:
1. Allow specific segments to be retrieved from `ODataPath` to avoid unnecessary enumeration.
2. Expose a struct enumerator on `ODataPath` to avoid boxing. We could have directly exposed `List<ODataPathSegment>.Enumerator` but the fact it's using a `List<T>` is an implementation detail.
3. Return `IList<SelectItem>` instead of an `IEnumerable<SelectItem>` so the count can be efficiently obtained, and indexing can be performed for random access.

- [x] *Test cases added*


ificator:
So it turns out when I ran all the unit tests in Visual Studio 2022 Preview the `PublicApiTests` did not run. I cannot make them run either (0 Passed, 0 Failed, 0 Skipped)... is there a trick to this?

src/Microsoft.OData.Edm/Schema/EdmContainedEntitySet.cs
```diff
@@ -12,7 +12,7 @@ namespace Microsoft.OData.Edm
     /// <summary>
     /// Represents an EDM contained entity set.
     /// </summary>
-    internal class EdmContainedEntitySet : EdmEntitySetBase, IEdmContainedEntitySet
+    public class EdmContainedEntitySet : EdmEntitySetBase, IEdmContainedEntitySet
```
xuzhg, line 15: That's weird for a long time to make it internal. I do think the author wants to change it later so make it internal. So, Any reason to public it?

xuzhg:
Others look good to me. :shipit:

src/Microsoft.OData.Edm/Schema/EdmContainedEntitySet.cs
ificator: A lot of stuff used to be unnecessarily `internal`, so I wouldn't be surprised if this was missed when things were cleaned up. I cannot see a reason for it to be internal, as it makes it impossible to create a `NavigationPropertySegment` that represents a contained navigation without reimplementing `EdmContainedEntitySet`.

In my code I ended up using a compiled lambda to create the instance I needed, but I'd much rather just use `new` :D

ificator:
@corranrogue9 / @gathogojr any feedback on this PR?

ificator:
@corranrogue9 / @gathogojr  ping?

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -94,9 +107,18 @@ internal IList<ODataPathSegment> Segments
...
+            return new Enumerator(this.segments);
+        }
+
+        /// <summary>
+        /// Get the segments enumerator
+        /// </summary>
+        /// <returns>The segments enumerator</returns>
+        IEnumerator<ODataPathSegment> IEnumerable<ODataPathSegment>.GetEnumerator()
+        {
+            return this.GetEnumerator();
```
corranrogue9, line 121: I think this implementation should remain as `return this.segments.GetEnumerator()`. We will have to do boxing either way, but calling `this.GetEnumerator()` will add an extra level of indirection for each call to the `IEnumerator`

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -171,5 +193,38 @@ internal void AddRange(ODataPath oDataPath)
...
+            }
+
+            public bool MoveNext()
+            {
+                return this.enumerator.MoveNext();
+            }
+
+            public void Reset()
+            {
+                // Unfortunately "Reset" is implemented explicitly and so we need to box IEnumerable<ODataPathSegment>, call "Reset",
```
corranrogue9, line 222: Is there a reason that we don't want to explicitly implement this as well? Honestly, I think I would rather just throw `NotSupportedException`

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -82,6 +82,19 @@ public int Count
             get { return this.segments.Count; }
         }
 
+        /// <summary>
+        /// Get the segment in the path at the specified index.
+        /// </summary>
+        /// <param name="i">The index.</param>
+        /// <returns>The segment corresponding to the index.</returns>
+        public ODataPathSegment this[int i]
```
corranrogue9, line 90: I'd like to understand what we are actually trying to get out of this. At this point, it seems like `ODataPath` started out not trying to leak implementation details, but now we might as well have it implement `IReadOnlyList<ODataPathSegment>`

src/Microsoft.OData.Edm/Schema/EdmContainedEntitySet.cs
corranrogue9: Which constructor would you end up using if you didn't use the lambda?

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
ificator: I'm happy to go either way - I opted for this because that way the same enumeration logic is used for all `GetEnumerator` variations, but since it's a simple enumerator that really isn't a strong argument.

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
ificator: That opens a potential regression, because `Reset()` would have worked prior to my refactoring.

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
ificator: I just needed a way to access the segments other than `First` and `Last` without enumerating through. Implementing `IReadOnlyList<ODataPathSegment>` would be a good idea - let me do that and we'll see how it looks.

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -94,9 +107,18 @@ internal IList<ODataPathSegment> Segments
         /// Get the segments enumerator
         /// </summary>
         /// <returns>The segments enumerator</returns>
-        public IEnumerator<ODataPathSegment> GetEnumerator()
+        public Enumerator GetEnumerator()
```
ificator, line 110: Another option is to simply expose `List<T>.Enumerator` but is that exposing implementation details?

I was going to say we have some default enumerators for wrapping standard data structures with struct-based enumerators:
```
public struct ODataListEnumerator<T> : IEnumerator<T>
```
```
public struct ODataArrayEnumerator<T> : IEnumerator<T>
```
But that would likewise expose implementation details... perhaps I'm over thinking, and as a result over complicating, this.

src/Microsoft.OData.Edm/Schema/EdmContainedEntitySet.cs
ificator: I'm using this one:

```
public EdmContainedEntitySet(IEdmNavigationSource parentNavigationSource, IEdmNavigationProperty navigationProperty)
```

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
ificator: I'm reverting my enumeration change - the indexer gets me what I need perf wise

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
ificator: I'm reverting my enumeration change - the indexer gets me what I need perf wise

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
ificator: I'm reverting my enumeration change - the indexer gets me what I need perf wise

corranrogue9: > That opens a potential regression, because Reset() would have worked prior to my refactoring.

Great point, sorry about that

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
corranrogue9: > is that exposing implementation details?

Yeah, while reviewing I kept trying to figure out how to do this. I ended up with like 3 nested generics and still wasn't any closer to a solution.

src/Microsoft.OData.Core/UriParser/SemanticAst/SelectExpandClause.cs
```diff
@@ -41,29 +41,28 @@ public SelectExpandClause(IEnumerable<SelectItem> selectedItems, bool allSelecte
...
         /// <summary>
-        /// Gets a flag indicating that everything at this level has been selected.
+        /// Gets the selected properties and operations.
         /// </summary>
-        /// <remarks>
-        /// If true, then all structural properties, bound actions and functions, and all navigations in the SelectedItems list have been selected.
-        /// </remarks>
-        public bool AllSelected
+        /// <remarks>This list includes expanded navigations properties, which may have additional nested selections and expansions.</remarks>
+        public IList<SelectItem> SelectedItems
```
corranrogue9, line 61: I would prefer `IReadOnlyCollection` or `IReadOnlyList`, whichever works best for your case.

src/Microsoft.OData.Edm/Schema/EdmContainedEntitySet.cs
corranrogue9: Thanks, I haven't take the time to fully evaluate this, but I would want to verify anything quality-related before taking this `public`. Just at a glance, for that constructor a `null` value for `navigationProperty` will result in a `NulReferenceException`, and it really should be an `ArgumentNullException` with the name of the parameter. Would you be willing to use your existing code for now and I will file an issue and make any quality updates that are necessary?

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -171,5 +161,22 @@ internal void AddRange(ODataPath oDataPath)
...
+        /// <exception cref="System.ArgumentNullException">Throws if the input other is null.</exception>
+        internal bool Equals(ODataPath other)
+        {
+            ExceptionUtils.CheckArgumentNotNull(other, "other");
+            if (this.segments.Count != other.segments.Count)
+            {
+                return false;
+            }
+
+            return !this.segments.Where((t, i) => !t.Equals(other.segments[i])).Any();
```
habbes, line 179: Since this is now an indexable list, maybe using a plain for loop would be efficient, skipping the boxing and iterator overhead. But I don't know whether this is on a hot path.

corranrogue9:
```suggestion
            for (int i = 0; i < this.segments.Count; ++i)
            {
                if (!this.segments[i].Equals(other.segments[i]))
                {
                    return false;
                }
            }
            
            return true;
```

corranrogue9:
```suggestion
            return !this.segments.SequenceEquals(other.segments);
```

ificator: I did a quick benchmark and the manual evaluation is best. Surprisingly, `SequenceEquals` is worst:
```
|        Method |      Mean |     Error |    StdDev | Allocated |
|-------------- |----------:|----------:|----------:|----------:|
|        Manual |  3.607 us | 0.0025 us | 0.0021 us |         - |
| SequenceEqual | 30.819 us | 0.2881 us | 0.2554 us |      48 B |
|   WhereNotAny | 18.368 us | 0.1094 us | 0.0913 us |      68 B |
```

src/Microsoft.OData.Edm/Schema/EdmContainedEntitySet.cs
ificator: Scope it down a bit and validated the input parameters for nullity - hopefully this will satisfy both of us :)

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -73,23 +77,26 @@ public ODataPathSegment LastSegment
                 return this.segments.LastOrDefault();
             }
         }
-
         /// <summary>
-        /// Get the number of segments in this path.
+        /// Get the List of ODataPathSegments.
         /// </summary>
-        public int Count
+        internal IList<ODataPathSegment> Segments
```
KenitoInc, line 83: Can we have a `ReadOnlyCollection` similar to this https://github.com/OData/WebApi/blob/7a3372ac031968996141de9f77ff0ac93333e642/src/Microsoft.AspNet.OData.Shared/Routing/ODataPath.cs#L94

PR #2356, opened 2022-03-18
Support 'case' conditional function

### LINKED ISSUE #2309: CASE not working in $compute and $filter
CASE is not working at all in $compute and $filter clauses

### Assemblies affected

OData .Net lib 8.0.7

### Reproduce steps

Call OData V4 service built using NET 6 and try a $compute with CASE clause 

### Expected result

Tha additional field

### Actual result

Exception in parser

xuzhg:
### Issues

*This pull request fixes https://github.com/OData/odata.net/issues/2309.*

### Description

Support 'case' conditional function parsing.

Here's the information for the "case" function:

5.1.1.12 [Conditional Functions](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_ConditionalFunctions)
[5.1.1.12.1 ](https://github.com/OData/odata.net/issues/2309)[case](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_case)
The case function has the following signature:

expression case(Edm.Boolean:expression, ..., Edm.Boolean:expression)

Each parameter is a pair of expressions separated by a colon (:), where the first expression – the condition – MUST be a Boolean expression, and the second expression – the result – may evaluate to any type.

The case function evaluates the condition in each pair, starting with the leftmost pair, and stops as soon as a condition evaluates to true. It then returns the value of the result of this pair. It returns null if none of the conditions in any pair evaluates to true. Clients can specify a last pair whose condition is true to get a non-null “default/else/otherwise” result.

Clients SHOULD ensure that the results in all pairs are compatible. If all results are of the same type, the type of the case expression is of that type. If all results are of numeric type, then the type of the case expression is a numeric type capable of representing any of these expressions according to standard type promotion rules.

Services MAY support case expressions containing parameters of incompatible types, in which case the case expression is treated as Edm.Untyped and its value has the type of the parameter expression selected by the case statement.

Example 97: compute signum(X)

$compute=case(X gt 0 :1,X lt 0 :-1,true :0) as SignumX

## [Discussion]

**We should ask whitespace before the ':', otherwise we have the following problem:**, (However, it sounds ridiculous)
That's because ':' is a valid character both for separate or char in time value.

1)  `case(Id eq 2:1)` => should 'case(Id eq 2 :1)`

2) `case(CreatedTime eq 10:2:1)`  ==> We can't distinguish it from TimeOfDay value

      But, this does:  `case(CreatedTime eq 10:2 :1)`

3)  `case(CreatedTime eq 10:2:1:2:3)`
      It can be:   ` 10:2:1 :  2:3`  or `10:2 : 1:2:3`

- [x] *Test cases added*

### Additional work necessary


sliekens:
Can this be picked up again? 🙏 I want to start using `$compute` as introduced in https://devblogs.microsoft.com/odata/compute-and-search-in-asp-net-core-odata-8/

but `$compute` without `case()` is very, very limited

xuzhg:
> Can this be picked up again? 🙏 I want to start using `$compute` as introduced in https://devblogs.microsoft.com/odata/compute-and-search-in-asp-net-core-odata-8/
> 
> but `$compute` without `case()` is very, very limited

Would you please share a little bit more detail about your usage/sceneries and yourself to us so we can priority it carefully? Thanks to:  saxu@microsoft.com

sliekens:
Right now I only have one use case which is null coalescing:

case(field1 ne null:field1, true:field2)

Which can be read as:

coalesce(field1, field2)

But I expect to find more ways to use this in the future.

sliekens:
Also see this related issue https://github.com/OData/AspNetCoreOData/issues/684

xuzhg:
It's supported in 9.x, close this draft.

PR #2359, opened 2022-03-21
Implement asynchronous support in ODataJsonLightBatchReader

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataJsonLightBatchReader`**.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Batch/ODataBatchReader.cs
+++ b/src/Microsoft.OData.Core/Batch/ODataBatchReader.cs
@@ -4,15 +4,19 @@
 // </copyright>
 //---------------------------------------------------------------------
 
-using System;
-using System.Collections.Generic;
-using System.Diagnostics;
-using System.Diagnostics.CodeAnalysis;
-using System.IO;
-using System.Threading.Tasks;
-
 namespace Microsoft.OData
 {
+    #region Namespaces
+
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics;
+    using System.Diagnostics.CodeAnalysis;
+    using System.IO;
+    using System.Threading.Tasks;
+
+    #endregion Namespaces
+
     /// <summary>
     /// Abstract class for reading OData batch messages; also verifies the proper sequence of read calls on the reader.
     /// </summary>
@@ -135,7 +139,7 @@ namespace Microsoft.OData
         public bool Read()
         {
             this.VerifyCanRead(true);
-            return this.InterceptException((Func<bool>)this.ReadSynchronously);
+            return this.InterceptException((thisParam) => thisParam.ReadSynchronously());
         }
 
         /// <summary>Asynchronously reads the next part from the batch message payload.</summary>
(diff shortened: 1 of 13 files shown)


marabooy (dismissed):

src/Microsoft.OData.Core/Batch/ODataBatchReader.cs
```diff
@@ -374,6 +371,82 @@ protected virtual string GetCurrentGroupIdImplementation()
...
+        }
+
+        /// <summary>
+        /// Asynchronous implementation of the reader logic when in state 'ChangesetStart'.
+        /// </summary>
+        /// <returns>
+        /// A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains the batch reader state after the read.
+        /// </returns>
+        protected virtual Task<ODataBatchReaderState> ReadAtChangesetStartImplementationAsync()
```
KenitoInc, line 433: Why are we wrapping a sync operation in an async method?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchBodyContentReaderStream.cs
```diff
@@ -364,5 +452,133 @@ private static void WriteCurrentJsonObject(IJsonReader reader, IJsonWriter jsonW
...
+        private static async Task WriteCurrentJsonObjectAsync(IJsonReaderAsync reader, IJsonWriterAsync jsonWriter)
+        {
+            Stack<JsonNodeType> nodeTypes = new Stack<JsonNodeType>();
+
+            do
+            {
+                switch (reader.NodeType)
+                {
+                    case JsonNodeType.PrimitiveValue:
+                        object primitiveValue;
```
KenitoInc, line 519:
```suggestion
                        object primitiveValue;

```

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchPayloadItemPropertiesCache.cs
```diff
@@ -198,7 +241,7 @@ private void ScanJsonProperties()
 
                         case PropertyNameDependsOn:
                             {
-                                IList<string> dependsOnIds = new List<string>();
+                                List<string> dependsOnIds = new List<string>();
```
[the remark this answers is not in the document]
gathogojr: @KenitoInc Because of the investigation previously done by @habbes...
`IList<T>` will use more memory than `List<T>`, because `List<T>` will use `public Enumerator GetEnumerator(){}` which returns an unboxed struct.
Whenever the interface versions of `GetEnumerator()` are called, the struct is boxed and returned as an interface type (`IEnumerator`). This causes an object to be allocated on the heap.
Plus there's actually no real value in using `IList` in the local variable here

Because

src/Microsoft.OData.Core/Microsoft.OData.Core.txt
```diff
@@ -203,6 +203,11 @@ ODataBatchReaderStream_UnexpectedEndOfInput=Encountered an unexpected end of inp
 
 ODataBatchReaderStreamBuffer_BoundaryLineSecurityLimitReached=Too many white spaces after a boundary delimiter and before the terminating line resource set. For security reasons, the total number of characters for a boundary including white spaces must not exceed {0}.
 
+ODataJsonLightBatchPayloadItemPropertiesCache_UnknownPropertyForMessageInBatch=Unknown property name '{0}' for message in batch.
+
```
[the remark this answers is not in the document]
gathogojr: @KenitoInc As far as I know, the _Microsoft.OData.Core.txt_ file is populated **manually** while the _Microsoft.OData.Core.cs_ is generated by running the _Microsoft.OData.Core.tt_ as a custom (TT) tool.
In addition, it does appear to me that we consistently try to leave a line space between error messages used on specific classes. See screen grab below
image

src/Microsoft.OData.Core/Batch/ODataBatchReader.cs
gathogojr: @KenitoInc We're doing this to avoid a breaking change. Several classes inherit from `ODataBatchReader`. Since all those classes at the moment implement the sync API, the async wrapper over the async is a better way than throwing a `NotImplementedException`. Since the async methods are declared as virtual, they are overridden in the derived classes and this stubbed implementation never gets executed

KenitoInc (approved):
LGTM

marabooy (approved):

PR #2363, opened 2022-03-24
Avoid zero-length array allocations

gathogojr:
### Description

This pull request addresses issue with zero-length array allocations

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
+++ b/src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
@@ -201,8 +201,8 @@ namespace Microsoft.OData.Json
                 return streamReader.CreateReadStream();
             }
 
-            Stream result = new MemoryStream(this.Value == null ? new byte[0] :
-                           Convert.FromBase64String((string)this.Value));
+            Stream result = this.Value == null ? Stream.Null :
+                new MemoryStream(Convert.FromBase64String((string)this.Value));
             this.innerReader.Read();
             return result;
         }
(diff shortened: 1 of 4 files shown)


marabooy (approved):

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -338,7 +338,7 @@ public virtual async Task<Stream> CreateReadStreamAsync()
 
             if (value == null)
             {
-                result = new MemoryStream(new byte[0]);
+                result = Stream.Null;
```
ElizabethOkerio, line 341: Is there a difference between using 
``` 
result = null
```
and 

```
result = Stream.Null

```

gathogojr: @ElizabethOkerio Yes. `result = null` is equivalent of `string[] array = null` while `result = Stream.Null` is the equivalent of `string[] array = Array.Empty<string>();`

`Stream.Null` is a stream with no backing store, which is not the same as null.

https://docs.microsoft.com/en-us/dotnet/api/system.io.stream.null?view=net-6.0

gathogojr: We're doing this to avoid unnecessary allocations. When you do `new byte[0]`, an allocation is done. Think of `Stream.Null` as a **reusable** zero-length array that helps avoid repeated zero-length allocations

ElizabethOkerio (approved):

PR #2364, opened 2022-03-25
Implement asynchronous support in ODataRawInputContext

gathogojr:
### Issues

### Description

Implement asynchronous support in **`ODataRawInputContext`**.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataAsynchronousReader.cs
+++ b/src/Microsoft.OData.Core/ODataAsynchronousReader.cs
@@ -31,6 +31,8 @@ namespace Microsoft.OData
         /// </summary>
         private readonly IServiceProvider container;
 
+        private byte[] byteBuffer = new byte[1];
+
         /// <summary>
         /// Constructor.
         /// </summary>
@@ -69,7 +71,7 @@ namespace Microsoft.OData
         {
             this.VerifyCanCreateResponseMessage(false);
 
-            return TaskUtils.GetTaskForSynchronousOperation(() => this.CreateResponseMessageImplementation());
+            return this.CreateResponseMessageImplementationAsync();
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.OData.Core/ODataAsynchronousReader.cs
```diff
@@ -31,6 +31,8 @@ public sealed class ODataAsynchronousReader
         /// </summary>
         private readonly IServiceProvider container;
 
+        private byte[] byteBuffer = new byte[1];
```
KenitoInc, line 34: Why is the buffer size 1?

KenitoInc (commented):
Fix public api

src/Microsoft.OData.Core/ODataAsynchronousReader.cs
gathogojr: @KenitoInc @ElizabethOkerio When reading the first line plus the headers lines of the raw input payload, we perform the reading one character at a time to ensure their validity
```
HTTP/1.1 200 OK
Content-Type: application/json
OData-Version: 4.0
```
For the synchronous API, we make use of the framework's `ReadByte` method defined in the `Stream` class. For the asynchronous API, we achieve the same using the `ReadAsync` method, reading a single byte at a time. That's what we're using this single-byte array for. We are declaring and initializing it here to avoid unnecessary allocations
```csharp
private async Task<int> ReadByteAsync()
{
    if (await this.rawInputContext.Stream.ReadAsync(byteBuffer, 0, 1).ConfigureAwait(false) != 0)
    {
        return byteBuffer[0];
    }

    return -1;
}
```

gathogojr:
> Fix public api

Undid the change that had caused a change to the public API

src/Microsoft.OData.Core/ODataAsynchronousReader.cs
marabooy: @KenitoInc this is also so that it can be used in closures, not sure if that is correct @gathogojr

gathogojr: I'm using it here https://github.com/OData/odata.net/pull/2364/files#diff-7ed6791834f0962c4865ca7047d3fc29b976468cfcc949afbef91b87f6662613R455

src/Microsoft.OData.Core/ODataRawInputContext.cs
```diff
@@ -251,5 +228,146 @@ private object ReadRawValue(IEdmPrimitiveTypeReference expectedPrimitiveTypeRefe
...
+
+            do
+            {
+                data = new byte[BufferSize];
+                result = await this.stream.ReadAsync(data, 0, data.Length)
+                    .ConfigureAwait(false);
+                numberOfBytesRead += result;
+                byteData.Add(data);
+            }
+            while (result == data.Length);
```
marabooy, line 286: suggestion: Maybe a small improvement would be for seekable streams check `CanSeek` Then read to the stream end as an small optimization for seekable streams.

KenitoInc (dismissed):
LGTM

src/Microsoft.OData.Core/ODataAsynchronousReader.cs
```diff
@@ -302,5 +304,160 @@ private int ReadByte()
...
+        /// Asynchronously reads a byte from the underlying stream.
+        /// </summary>
+        /// <returns>
+        /// A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains the byte read from the stream.
+        /// </returns>
+        private async Task<int> ReadByteAsync()
+        {
+            // Stream does not support ReadByteAsync - Improvise
+            if (await this.rawInputContext.Stream.ReadAsync(byteBuffer, 0, 1).ConfigureAwait(false) != 0)
```
habbes, line 455: If we're reading into a byte buffer directly, maybe it would make sense to return the read byte as `byte`. Or this could return a boolean then the caller reads `byteBuffer[0]` if the return was successful. That way we avoid `byte`->`int` or `byte`->`char` comparisons/conversions. But you could verify that improves things before applying the suggestion.

src/Microsoft.OData.Core/ODataRawInputContext.cs
```diff
@@ -115,7 +117,7 @@ internal override object ReadValue(IEdmPrimitiveTypeReference expectedPrimitiveT
         internal override Task<object> ReadValueAsync(IEdmPrimitiveTypeReference expectedPrimitiveTypeReference)
         {
             // Note that the reading is actually synchronous since we buffer the entire input when getting the stream from the message.
-            return TaskUtils.GetTaskForSynchronousOperation(() => this.ReadValueImplementation(expectedPrimitiveTypeReference));
+            return this.ReadValueImplementationAsync(expectedPrimitiveTypeReference);
```
habbes, line 120: If this is actually synchronous, was it necessary to create an async implementation?

src/Microsoft.OData.Core/ODataRawInputContext.cs
```diff
@@ -251,5 +228,146 @@ private object ReadRawValue(IEdmPrimitiveTypeReference expectedPrimitiveTypeRefe
...
+        /// <summary>
+        /// Initializes the <see cref="TextReader"/> class variable.
+        /// </summary>
+        private void InitializeTextReader()
+        {
+            this.textReader = this.Encoding == null ? new StreamReader(this.stream) : new StreamReader(this.stream, this.Encoding);
+        }
+
+        /// <summary>
+        /// Checks whether the value can be read as binary give its expected primitive type.
```
habbes, line 325:
```suggestion
        /// Checks whether the value can be read as binary given its expected primitive type.

```

src/Microsoft.OData.Core/ODataRawInputContext.cs
habbes: Do you know at which point this buffering of the entire input takes place?

src/Microsoft.OData.Core/ODataRawInputContext.cs
habbes: According to one of the comments above, all the data has already been buffered and the reading is probably in memory. If that's the case, it's like to be a seekable stream.

habbes:
Just seen this [HttpParser](https://github.com/dotnet/aspnetcore/blob/main/src/Servers/Kestrel/Core/src/Internal/Http/HttpParser.cs) implementation for Kestrel. Maybe we can leverage it in future, or at least use it as a reference.

src/Microsoft.OData.Core/ODataRawInputContext.cs
gathogojr: @habbes See my comment [here](https://github.com/OData/odata.net/pull/2364/files#r844296036)

gathogojr: @marabooy I implemented your suggestion in both `ReadBinaryValue` and `ReadBinaryValueAsync`

src/Microsoft.OData.Core/ODataAsynchronousReader.cs
```diff
@@ -302,5 +304,160 @@ private int ReadByte()
...
+            {
+                string headerName, headerValue;
+                ValidateHeaderLine(headerLine, out headerName, out headerValue);
+
+                if (headers.ContainsKey(headerName))
+                {
+                    throw new ODataException(Strings.ODataAsyncReader_DuplicateHeaderFound(headerName));
+                }
+
+                headers.Add(headerName, headerValue);
```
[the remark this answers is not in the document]
gathogojr: @habbes `TryAdd` was introduced in .NET Core 2.0 and .NET Standard 2.1. In particular, it does not exist on .NET Framework (any version) and it does exist on .NET 5. It's however supported by `ConcurrentDictionary`.

src/Microsoft.OData.Core/ODataAsynchronousReader.cs
gathogojr: @habbes Our offline discussion about the reasons we do it refers

src/Microsoft.OData.Core/ODataAsynchronousReader.cs
```diff
@@ -302,5 +304,160 @@ private int ReadByte()
...
+        }
+
+        /// <summary>
+        /// Asynchronously reads the headers of an async response.
+        /// </summary>
+        /// <returns>
+        /// A task that represents the asynchronous read operation.
+        /// The value of the TResult parameter contains a dictionary of header names to header values; never null.
+        /// </returns>
+        private async Task<Dictionary<string, string>> ReadHeadersAsync()
```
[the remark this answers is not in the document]
gathogojr: @habbes Our offline discussion about the reasons we do it refers

src/Microsoft.OData.Core/ODataAsynchronousReader.cs
```diff
@@ -302,5 +304,160 @@ private int ReadByte()
...
+
+                    if (ch != '\n')
+                    {
+                        throw new ODataException(Strings.ODataAsyncReader_InvalidNewLineEncountered('\r'));
+                    }
+
+                    return lineBuilder.ToString();
+                }
+
+                lineBuilder.Append((char)ch);
```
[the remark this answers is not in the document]
gathogojr: @habbes Our offline discussion about the reasons we do it refers

src/Microsoft.OData.Core/ODataRawInputContext.cs
gathogojr: @habbes @marabooy Added tests for reading from non-seekable stream

marabooy (approved):

KenitoInc (approved):

PR #2376, opened 2022-04-13
updated Microsoft.OData.Core.Tests and Microsoft.OData.Core reference…

### LINKED ISSUE #2296: Some projects in the OData.E2E solution always rebuild
`Commander.Common`, `Microsoft.OData.Core`, `Microsoft.Data.Web.Design.T4`, `Microsoft.OData.Service`, and `AstoriaTestFramework` always rebuild when the OData.E2E solution is built, even if no changes have been made.

corranrogue9:
…s to Microsoft.Bcl.AsyncInterfaces to not conflict with other packages

### Issues

*This pull request fixes #2296.*

### Description

*The `Microsoft.OData.Core.Tests project does not support incremental builds, and therefore always rebuilds even if no changes have occurred. This is due to a conflict in the version of `Microsoft.Bcl.AsyncInterfaces`. It is currently using version 5.0.0, while other projects in the repo reference a nuget packages which depends on version 1.0.0.*

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
+++ b/src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
@@ -96,4 +96,10 @@
       <Version>6.0.3</Version>
     </PackageReference>
   </ItemGroup>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces">
+      <Version>5.0.0</Version>
+    </PackageReference>
+  </ItemGroup>
 </Project>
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
```diff
@@ -92,9 +92,6 @@
   </ItemGroup>
 
   <ItemGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
-    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces">
```
xuzhg, line 95: is it a breaking changes to downgrade the dependency version? 

Why can't we upgrade the dependency version in the test project?

PR #2377, opened 2022-04-19
Fix csproj

KenitoInc:
### Issues

When I rebased my fork, I got this diff.
Probably caused by this commit https://github.com/OData/odata.net/commit/bae11a51b2e0be642e2e42351774f14508becf55

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
+++ b/src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
@@ -77,20 +77,6 @@
     </Content>
   </ItemGroup>
 
-  <ItemGroup>
-    <Compile Update="Microsoft.OData.Core.cs">
-      <DesignTime>True</DesignTime>
-      <AutoGen>True</AutoGen>
-      <DependentUpon>Microsoft.OData.Core.tt</DependentUpon>
-    </Compile>
-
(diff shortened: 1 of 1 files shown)


src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
```diff
@@ -77,20 +77,6 @@
     </Content>
   </ItemGroup>
 
-  <ItemGroup>
```
xuzhg, line 80: why do we remove them? They are needed for the resource generation

KenitoInc: Ping @corranrogue9

corranrogue9: I recall having an intermediate commit in my PR where I had accidentally removed this section, but these changes are not in the master branch, and I agree with Sam that they should stay.

KenitoInc: @corranrogue9  This [commit](https://github.com/OData/odata.net/commit/bae11a51b2e0be642e2e42351774f14508becf55) is in the master branch and it did some changes to csproj. Kindly have a look

PR #2407, opened 2022-05-12
Encode special chars in array elements in Contains expression

### LINKED ISSUE #2405: LinQ syntax: Array.Contain() doesn't encode & character
When using Microsoft.OData.Client , with LinQ syntax, the generated RequestUri doesn't encode "&" character, thus when server receive the request, it can't parse query and throw exception "Request Found an unbalanced bracket expression"

### Assemblies affected

Microsoft.OData.Client 7.11.0 and earlier versions

### Reproduce steps

```
using Microsoft.OData.Client;
using System;
using System.Linq;
namespace OData
{
    internal class Program
    {
        static void Main(string[] args)
        {
            var arr = new string[]
            {
                "a & b"
            };
            var ctx = new DataServiceContext(new Uri("http://localhost"));

            var queryWithEqualityFilter = ctx.CreateQuery<Model>(nameof(Model)).Where(x => x.Name == arr[0]) as DataServiceRequest;
            Console.WriteLine(queryWithEqualityFilter.RequestUri); 
            //WORK FINE: Output: http://localhost/Model?$filter=Name eq 'a %26 b'

            var queryWithContainFilter = ctx.CreateQuery<Model>(nameof(Model)).Where(x => arr.Contains(x.Name)) as DataServiceRequest;
            Console.WriteLine(queryWithContainFilter.RequestUri); 
            //ISSUE HAPPEN: Output: http://localhost/Model?$filter=Name in ('a & b')
        }
    }
    public class Model
    {
    }
}
```

### Expected result

The last line should be similar to equality filter, it should encode the & character, such as http://localhost/Model?$filter=Name in ('a %26 b')

### Actual result

It doesn't encode & character: http://localhost/Model?$filter=Name in ('a & b')

gathogojr:
### Issues

*This pull request fixes #2405.*

### Description

The `ConvertToUriLiteral` method used previous [does not do any escaping](https://github.com/OData/odata.net/blob/4e263ca4b94d01f2d105d8282c62ec272cfcd1bb/src/Microsoft.OData.Core/Uri/ODataUriUtils.cs#L93) so it's not the best fit for formatting literals.
In addition, `ConvertToUriLiteral` handles `ODataNullValue`, `ODataResourceValue`, `ODataCollectionValue`, `ODataEnumValue`, `ODataResource`, `ODataEntityReferenceLink`, `ODataEntityReferenceLinks`, `ODataResource` collection and primitive values. The only candidates for the "in" operator are primitive values so using this method is an overkill.

The change made in this PR is to use same method used to format values in other `Where` predicates

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
+++ b/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
@@ -526,7 +526,7 @@ namespace Microsoft.OData.Client
             else if (m != null && ReflectionUtil.IsSequenceMethod(m.Method, SequenceMethod.Contains))
             {
                 StringBuilder listExpr = new StringBuilder();
-                ODataVersion version = CommonUtil.ConvertToODataVersion(this.uriVersion);
+
                 foreach (object item in (IEnumerable)c.Value)
                 {
                     if (listExpr.Length != 0)
@@ -534,7 +534,24 @@ namespace Microsoft.OData.Client
                         listExpr.Append(UriHelper.COMMA);
                     }
 
-                    string uriLiteral = ODataUriUtils.ConvertToUriLiteral(item, version);
+                    string uriLiteral;
+
+                    try
+                    {
+                        uriLiteral = LiteralFormatter.ForConstants.Format(item);
+                    }
+                    catch (InvalidOperationException)
+                    {
+                        if (this.cantTranslateExpression)
+                        {
+                            // There's already a problem in the parents.
+                            // Return.A up the stack will throw a better exception
+                            return c;
+                        }
+
+                        throw new NotSupportedException(Strings.ALinq_CouldNotConvert(item));
+                    }
+
                     listExpr.Append(uriLiteral);
                 }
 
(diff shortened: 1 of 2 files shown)


xuzhg (approved):

src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
```diff
@@ -526,15 +526,32 @@ internal override Expression VisitConstant(ConstantExpression c)
...
+                    catch (InvalidOperationException)
+                    {
+                        if (this.cantTranslateExpression)
+                        {
+                            // There's already a problem in the parents.
+                            // Return.A up the stack will throw a better exception
+                            return c;
+                        }
+
+                        throw new NotSupportedException(Strings.ALinq_CouldNotConvert(item));
```
corranrogue9, line 552: Will an instance of this expression writer be expected to handle future calls, even in the event of a previous exception? It might make sense to set `cantTranslateExpression` so that the instance remains internally consistent.

corranrogue9 (approved):

src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
gathogojr: @corranrogue9 The approach adopted here mirrors the approach taken when escaping special characters in a typical filter expression https://github.com/OData/odata.net/pull/2407/files#diff-7d25f0d0e6eeabce6281ebb5949c72093911c347a158342109816424376c5d73R584

Issue #1817, opened 2020-06-23
Support OData key alias


xuzhg:
OData spec says:

## $Key

The value of $Key is an array with one item per key property.

**Key properties without a key alias are represented as strings containing the property name.**

**Key properties with a key alias are represented as objects with one member whose name is the key alias and whose value is a string containing the path to the property.**

The second one is not supported in ODL.

Here's an example for Key alias:

```JSON
"Category": {

  "$Kind": "EntityType",

  "$Key": [

    {

      "EntityInfoID": "Info/ID"

    }

  ],

  "Info": {

    "$Type": "self.EntityInfo"

  },

  "Name": {

    "$Nullable": true

  }

},

"EntityInfo": {

  "$Kind": "ComplexType",

  "ID": {

    "$Type": "Edm.Int32"

  },

  "Created": {

    "$Type": "Edm.DateTimeOffset",

    "$Precision": 0

  }

}
```

gathogojr:
Fixed by https://github.com/OData/odata.net/pull/3362

PR #2411, opened 2022-05-19
Key Alias support

ElizabethOkerio:
### Issues

This pull request fixes https://github.com/OData/odata.net/issues/1817

Key Alias support in ODL

```C SHARP
"Category": {
  "$Kind": "EntityType",
  "$Key": [
    {
      "EntityInfoID": "Info/ID"
    }
  ],
  "Info": {
    "$Type": "self.EntityInfo"
  },
  "Name": {
    "$Nullable": true
  }
},
"EntityInfo": {
  "$Kind": "ComplexType",
  "ID": {
    "$Type": "Edm.Int32"
  },
  "Created": {
    "$Precision": 0
  }
}
```
<code>EntityInfoID</code> is the alias and <code>Info/ID</code> is the path to the key.

Changes made in this PR:
**New interfaces and Classes;**
1. **IEdmStructuralPropertyAlias : IEdmStructuralProperty** - this interface has two properties: - <code>IEnumerable<string> Path { get; } </code> and <code> string PropertyAlias { get; } </code>
2. **EdmStructuralPropertyAlias : EdmProperty, IEdmStructuralPropertyAlias**

**EdmStructuredType**
Added a new method 
public EdmStructuralPropertyAlias AddStructuralPropertyAlias(IEdmTypeReference type, string alias, IEnumerable<string> path) - This is the method that users will use to add key aliases. 
1. type is the type of the complex property 
2. alias is the key alias to use
4. path is a list of the path segments to the key property. 

**EdmEntityType** 
In the <code>AddKeys(IEnumerable<IEdmStructuralProperty> keyProperties) </code> method, when looping through each property, we cast it into an <code>IEdmStructuralPropertyAlias</code> and check if the key has an alias or not. If it does,  we loop through the keyProperties till we get to the last path segment which is the key  and create an <code>EdmStructuralPropertyAlias </code> and add it to the declaredKeys IEnumerable. 
### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlPropertyReference.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlPropertyReference.cs
@@ -12,6 +12,7 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Ast
     internal class CsdlPropertyReference : CsdlElement
     {
         private readonly string propertyName;
+        private readonly string propertyAlias;
 
         public CsdlPropertyReference(string propertyName, CsdlLocation location)
             : base(location)
@@ -19,9 +20,21 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Ast
             this.propertyName = propertyName;
         }
 
+        public CsdlPropertyReference(string propertyName, string propertyAlias, CsdlLocation location)
+            : base(location)
+        {
+            this.propertyName = propertyName;
+            this.propertyAlias = propertyAlias;
+        }
+
         public string PropertyName
         {
             get { return this.propertyName; }
+        }
+
+        public string PropertyAlias
+        {
+            get { return this.propertyAlias; }
         }
     }
 }
--- a/src/Microsoft.OData.Edm/Csdl/Parsing/SchemaJsonParser.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Parsing/SchemaJsonParser.cs
@@ -655,13 +655,20 @@ namespace Microsoft.OData.Edm.Csdl.Parsing
             {
                 if (v.ValueKind == JsonValueKind.Object)
                 {
-                    // TODO: ODL doesn't support the key referencing a property of a complex type as below.
                     // "$Key": [
                     //    {
                     //      "EntityInfoID": "Info/ID"
                     //    }
                     //  ]
-                    p.ReportError(EdmErrorCode.InvalidKeyValue, "It doesn't support the key object.");
+                    string pName = "";
+                    string pValue = "";
+                    v.ParseAsObject(p, (keyAlias, keyName) =>
+                    {
+                        pName = keyName.ParseAsString(p);
+                        pValue = keyAlias;
+                    });
+
+                    return new CsdlPropertyReference(pName, pValue, context.Location());
                 }
 
                 string propertyName = v.ParseAsString(context);
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
@@ -107,37 +107,57 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
 
         private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
         {
-            if (this.entity.Key != null)
+            if (this.entity.Key == null)
             {
-                List<IEdmStructuralProperty> key = new List<IEdmStructuralProperty>();
-                foreach (CsdlPropertyReference keyProperty in this.entity.Key.Properties)
+                return null;
+            }
+
+            List<IEdmStructuralProperty> key = new List<IEdmStructuralProperty>();
+            foreach (CsdlPropertyReference keyProperty in this.entity.Key.Properties)
+            {
+                IEdmStructuralProperty structuralProperty = null;
+                if (!string.IsNullOrEmpty(keyProperty?.PropertyAlias))
                 {
-                    IEdmStructuralProperty structuralProperty = this.FindProperty(keyProperty.PropertyName) as IEdmStructuralProperty;
+                    string[] keyPropertySegments = keyProperty.PropertyName.Split('/');
+                    if (keyPropertySegments.Length > 0)
+                    {
+                        structuralProperty = this.FindProperty(keyPropertySegments[0]) as IEdmStructuralProperty;
+                        if (structuralProperty == null)
+                        {
+                            return null;
+                        }
+
+                        structuralProperty = PropertyAliasHelpers.GetKeyProperty(structuralProperty, keyPropertySegments, keyProperty.PropertyAlias);
+                    }
+
+                }
+                else
+                {
+                    structuralProperty = this.FindProperty(keyProperty.PropertyName) as IEdmStructuralProperty;
+                }
+
+                if (structuralProperty != null)
+                {
+                    key.Add(structuralProperty);
+                }
+                else
+                {
+                    // If keyProperty is a duplicate, it will come back as non-structural from FindProperty, but it still might be structural
+                    // inside the DeclaredProperties, so try it. If it is not in the DeclaredProperties or it is not structural there,
+                    // then fall back to unresolved.
+                    structuralProperty = this.DeclaredProperties.FirstOrDefault(p => p.Name == keyProperty.PropertyName) as IEdmStructuralProperty;
                     if (structuralProperty != null)
                     {
                         key.Add(structuralProperty);
                     }
                     else
                     {
-                        // If keyProperty is a duplicate, it will come back as non-structural from FindProperty, but it still might be structural
-                        // inside the DeclaredProperties, so try it. If it is not in the DeclaredProperties or it is not structural there,
-                        // then fall back to unresolved.
-                        structuralProperty = this.DeclaredProperties.FirstOrDefault(p => p.Name == keyProperty.PropertyName) as IEdmStructuralProperty;
-                        if (structuralProperty != null)
-                        {
-                            key.Add(structuralProperty);
-                        }
-                        else
-                        {
-                            key.Add(new UnresolvedProperty(this, keyProperty.PropertyName, this.Location));
-                        }
+                        key.Add(new UnresolvedProperty(this, keyProperty.PropertyName, this.Location));
                     }
                 }
-
-                return key;
             }
 
-            return null;
+            return key;
         }
     }
 }
--- a/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaJsonWriter.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaJsonWriter.cs
@@ -208,8 +208,18 @@ namespace Microsoft.OData.Edm.Csdl.Serialization
         {
             // Key properties without a key alias are represented as strings containing the property name.
             // Key properties with a key alias are represented as objects with one member whose name is the key alias and whose value is a string containing the path to the property.
-            // TODO: It seems the second one is not supported.
-            this.jsonWriter.WriteStringValue(property.Name);
+            IEdmStructuralPropertyAlias prop = property as IEdmStructuralPropertyAlias;
+            if (prop == null)
+            {
+                this.jsonWriter.WriteStringValue(property.Name);
+            }
+            else
+            {
+                this.jsonWriter.WriteStartObject();
+                this.jsonWriter.WritePropertyName(prop.PropertyAlias);
+                this.jsonWriter.WriteStringValue(string.Join("/", prop.Path));
+                this.jsonWriter.WriteEndObject();
+            }
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaJsonWriter.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaJsonWriter.cs
@@ -208,17 +208,16 @@ namespace Microsoft.OData.Edm.Csdl.Serialization
         {
             // Key properties without a key alias are represented as strings containing the property name.
             // Key properties with a key alias are represented as objects with one member whose name is the key alias and whose value is a string containing the path to the property.
-            IEdmStructuralPropertyAlias prop = property as IEdmStructuralPropertyAlias;
-            if (prop == null)
-            {
-                this.jsonWriter.WriteStringValue(property.Name);
-            }
-            else
+            if (property is IEdmStructuralPropertyAlias propertyAlias)
             {
                 this.jsonWriter.WriteStartObject();
-                this.jsonWriter.WritePropertyName(prop.PropertyAlias);
-                this.jsonWriter.WriteStringValue(string.Join("/", prop.Path));
+                this.jsonWriter.WritePropertyName(propertyAlias.PropertyAlias);
+                this.jsonWriter.WriteStringValue(string.Join("/", propertyAlias.Path));
                 this.jsonWriter.WriteEndObject();
+            }
+            else 
+            {
+                this.jsonWriter.WriteStringValue(property.Name);
             }
         }
 
--- a/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
@@ -153,7 +153,17 @@ namespace Microsoft.OData.Edm.Csdl.Serialization
         internal override void WritePropertyRefElement(IEdmStructuralProperty property)
         {
             this.xmlWriter.WriteStartElement(CsdlConstants.Element_PropertyRef);
-            this.WriteRequiredAttribute(CsdlConstants.Attribute_Name, property.Name, EdmValueWriter.StringAsXml);
+            IEdmStructuralPropertyAlias prop = property as IEdmStructuralPropertyAlias;
+            if (prop == null)
+            {
+                this.WriteRequiredAttribute(CsdlConstants.Attribute_Name, property.Name, EdmValueWriter.StringAsXml);
+            }
+            else
+            {
+                this.WriteRequiredAttribute(CsdlConstants.Attribute_Name, string.Join("/", prop.Path), EdmValueWriter.StringAsXml);
+                this.WriteRequiredAttribute(CsdlConstants.Attribute_Alias, prop.PropertyAlias, EdmValueWriter.StringAsXml);
+            }
+
             this.WriteEndElement();
         }
 
--- a/src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
+++ b/src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
@@ -1,0 +1,75 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="EdmStructuralPropertyAlias.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.OData.Edm
+{
+    /// <summary>
+    /// Represents an EDM structural (i.e. non-navigation) property with an alias.
+    /// </summary>
+    public class EdmStructuralPropertyAlias : EdmProperty, IEdmStructuralPropertyAlias
+    {
+        private string propertyAlias;
+        private IEnumerable<string> path;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EdmStructuralPropertyAlias"/> class.
+        /// </summary>
+        /// <param name="declaringType">The type that owns this property.</param>
+        /// <param name="name">Name of the property.</param>
+        /// <param name="type">The type of the property.</param>
+        /// <param name="alias">The property alias.</param>
+        /// <param name="path">The path to the referenced property.</param>
+        public EdmStructuralPropertyAlias(IEdmStructuredType declaringType, string name, IEdmTypeReference type, string propertyAlias, IEnumerable<string> path)
+            : base(declaringType, name, type)
+        {
+            if (propertyAlias == null)
+            {
+                throw new ArgumentNullException(nameof(propertyAlias));
+            }
+
+            if (path == null)
+            {
+                throw new ArgumentNullException(nameof(path));
+            }
+
+            this.propertyAlias = propertyAlias;
+            this.path = path;
+        }
+
+        /// <inheritdoc/>
+        public string PropertyAlias
+        {
+            get { return this.propertyAlias; }
+
+        }
+
+        /// <inheritdoc/>
+        public IEnumerable<string> Path
+        {
+            get { return this.path; }
+        }
+
+        /// <summary>
+        /// Gets the default value of this property.
+        /// </summary>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", Justification = "defaultValue might be confused for an IEdmValue.")]
+        public string DefaultValueString
+        {
+            get;
+        }
+
+        /// <summary>
+        /// Gets the kind of this property.
+        /// </summary>
+        public override EdmPropertyKind PropertyKind
+        {
+            get { return EdmPropertyKind.Structural; }
+        }
+    }
+}
--- a/src/Microsoft.OData.Edm/PropertyAliasHelpers.cs
+++ b/src/Microsoft.OData.Edm/PropertyAliasHelpers.cs
@@ -1,0 +1,42 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="PropertyAliasHelpers.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using System;
+using System.Linq;
+
+namespace Microsoft.OData.Edm
+{
+    internal static class PropertyAliasHelpers
+    {
+        internal static IEdmStructuralProperty GetKeyProperty(IEdmStructuralProperty property, string[] keyPropertySegments, string propertyAlias)
+        {
+            IEdmStructuralProperty structuralProperty = property;
+            for (int i = 1; i < keyPropertySegments.Length; i++)
+            {
+                if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Complex)
+                {
+                    structuralProperty = structuralProperty.Type.AsComplex().FindProperty(keyPropertySegments[i]) as IEdmStructuralProperty;
+                }
+                else if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Entity)
+                {
+                    structuralProperty = structuralProperty.Type.AsEntity().FindProperty(keyPropertySegments[i]) as IEdmStructuralProperty;
+                }
+                else
+                {
+                    throw new InvalidOperationException(Strings.Bad_UnresolvedType(structuralProperty));
+                }
+            }
+
+            if (structuralProperty != null)
+            {
+                structuralProperty = new EdmStructuralPropertyAlias(structuralProperty.DeclaringType, structuralProperty.Name, structuralProperty.Type, propertyAlias, keyPropertySegments);
+            }
+
+            return structuralProperty;
+
+        }
+    }
+}
--- a/src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
+++ b/src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
@@ -7,6 +7,7 @@
 using System;
 using System.Collections.Generic;
 using System.Diagnostics;
+using System.Linq;
 
 namespace Microsoft.OData.Edm
 {
@@ -157,7 +158,28 @@ namespace Microsoft.OData.Edm
                     this.declaredKey = new List<IEdmStructuralProperty>();
                 }
 
-                this.declaredKey.Add(property);
+                IEdmStructuralPropertyAlias structuralPropertyAlias = property as IEdmStructuralPropertyAlias;
+                if (!string.IsNullOrEmpty(structuralPropertyAlias?.PropertyAlias))
+                {
+                    string[] keyPropertySegments = structuralPropertyAlias.Path.ToArray();
+                    IEdmStructuralProperty structuralProperty = property.DeclaringType.FindProperty(keyPropertySegments[0]) as IEdmStructuralProperty;
+                    if (structuralProperty != null)
+                    {
+                        IEdmStructuralProperty prop = PropertyAliasHelpers.GetKeyProperty(structuralProperty, keyPropertySegments, structuralPropertyAlias.PropertyAlias);
+                        if (prop != null)
+                        {
+                            this.declaredKey.Add(prop);
+                        }
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException(Strings.Bad_UnresolvedType(structuralProperty));
+                    }
+                }
+                else 
+                {
+                    this.declaredKey.Add(property);
+                }
             }
         }
 
--- a/src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
+++ b/src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
@@ -6,6 +6,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Linq;
 
 namespace Microsoft.OData.Edm
 {
@@ -159,6 +160,25 @@ namespace Microsoft.OData.Edm
 
             EdmNavigationProperty property = EdmNavigationProperty.CreateNavigationProperty(this, propertyInfo);
 
+            this.AddProperty(property);
+            return property;
+        }
+
+        /// <summary>
+        /// Creates and adds a structural property alias to this type.
+        /// </summary>
+        /// <param name="type">Type of the property.</param>
+        /// <param name="alias">The property alias.</param>
+        /// <param name="path">The key path segments</param>
+        /// <returns>Created structural property.</returns>
+        public EdmStructuralPropertyAlias AddStructuralPropertyAlias(IEdmTypeReference type, string alias, IEnumerable<string> path)
+        {
+            if (path == null)
+            {
+                throw new ArgumentNullException(nameof(path));
+            }
+            string propertyName = path.FirstOrDefault();
+            EdmStructuralPropertyAlias property = new EdmStructuralPropertyAlias(this, propertyName, type, alias, path);
             this.AddProperty(property);
             return property;
         }
--- a/test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
@@ -2891,5 +2891,114 @@ namespace Microsoft.OData.Edm.Tests.Csdl
   }
 }");
         }
+
+        [Fact]
+        public void VerifyKeyAliasForNavigationPropertiesIsSupported()
+        {
+            // Arrange
+            var model = new EdmModel();
+            var entity = new EdmEntityType("NS1", "Product");
+            var orderEntity = new EdmEntityType("NS1", "Order");
+            var key = orderEntity.AddStructuralProperty("Id", EdmCoreModel.Instance.GetInt32(false));
+            orderEntity.AddKeys(key);
+            orderEntity.AddStructuralProperty("Name", EdmCoreModel.Instance.GetString(false));
+            var complexType = new EdmComplexType("NS1", "ExampleComplexObject");
+            complexType.AddStructuralProperty("Name", EdmCoreModel.Instance.GetString(false));
+            var keyProperty = entity.AddStructuralPropertyAlias(new EdmEntityTypeReference(orderEntity, isNullable: false), "OrderId", new List<string> { "Order", "Id" });
+            entity.AddKeys(keyProperty);
+            entity.AddStructuralProperty("Name", EdmCoreModel.Instance.GetString(false));
+            entity.AddStructuralProperty("UpdatedTime", EdmCoreModel.Instance.GetDate(false));
+
+            model.AddElement(entity);
+            model.AddElement(orderEntity);
+            model.AddElement(complexType);
+
+            var entityContainer = new EdmEntityContainer("NS1", "Container");
+            model.AddElement(entityContainer);
+            EdmEntitySet set1 = new EdmEntitySet(entityContainer, "Products", entity);
+            EdmEntitySet set2 = new EdmEntitySet(entityContainer, "Orders", orderEntity);
+            entityContainer.AddElement(set1);
+            entityContainer.AddElement(set2);
+
+            // Act & Assert for XML
+            WriteAndVerifyXml(model, "<?xml version=\"1.0\" encoding=\"utf-16\"?>" +
+                "<edmx:Edmx Version=\"4.0\" xmlns:edmx=\"http://docs.oasis-open.org/odata/ns/edmx\">" +
+                  "<edmx:DataServices>" +
+                    "<Schema Namespace=\"NS1\" xmlns=\"http://docs.oasis-open.org/odata/ns/edm\">" +
+                      "<EntityType Name=\"Product\">" +
+                        "<Key>" +
+                          "<PropertyRef Name=\"Order/Id\" Alias=\"OrderId\" />" +
+                        "</Key>" +
+                        "<Property Name=\"Order\" Type=\"NS1.Order\" Nullable=\"false\" />" +
+                        "<Property Name=\"Name\" Type=\"Edm.String\" Nullable=\"false\" />" +
+                        "<Property Name=\"UpdatedTime\" Type=\"Edm.Date\" Nullable=\"false\" />" +
+                      "</EntityType>" +
+                      "<EntityType Name=\"Order\">" +
+                        "<Key>" +
+                          "<PropertyRef Name=\"Id\" />" +
+                        "</Key>" +
+                        "<Property Name=\"Id\" Type=\"Edm.Int32\" Nullable=\"false\" />" +
+                        "<Property Name=\"Name\" Type=\"Edm.String\" Nullable=\"false\" />" +
+                      "</EntityType>" +
+                      "<ComplexType Name=\"ExampleComplexObject\">" +
+                        "<Property Name=\"Name\" Type=\"Edm.String\" Nullable=\"false\" />" +
+                      "</ComplexType>" +
+                      "<EntityContainer Name=\"Container\">" +
+                        "<EntitySet Name=\"Products\" EntityType=\"NS1.Product\" />" +
+                        "<EntitySet Name=\"Orders\" EntityType=\"NS1.Order\" />" +
+                      "</EntityContainer>" +
+                    "</Schema>" +
+                  "</edmx:DataServices>" +
+                "</edmx:Edmx>");
+
+            // Act & Assert for JSON
+            WriteAndVerifyJson(model, @"{
+  ""$Version"": ""4.0"",
+  ""$EntityContainer"": ""NS1.Container"",
+  ""NS1"": {
+    ""Product"": {
+      ""$Kind"": ""EntityType"",
+      ""$Key"": [
+        {
+          ""OrderId"": ""Order/Id""
+        }
+      ],
+      ""Order"": {
+        ""$Type"": ""NS1.Order""
+      },
+      ""Name"": {},
+      ""UpdatedTime"": {
+        ""$Type"": ""Edm.Date""
+      }
+    },
+    ""Order"": {
+      ""$Kind"": ""EntityType"",
+      ""$Key"": [
+        ""Id""
+      ],
+      ""Id"": {
+        ""$Type"": ""Edm.Int32""
+      },
+      ""Name"": {}
+    },
+    ""ExampleComplexObject"": {
+      ""$Kind"": ""ComplexType"",
+      ""Name"": {}
+    },
+    ""Container"": {
+      ""$Kind"": ""EntityContainer"",
+      ""Products"": {
+        ""$Collection"": true,
+        ""$Type"": ""NS1.Product""
+      },
+      ""Orders"": {
+        ""$Collection"": true,
+        ""$Type"": ""NS1.Order""
+      }
+    }
+  }
+}");
+        }
+
     }
 }
--- a/test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
@@ -2812,5 +2812,84 @@ namespace Microsoft.OData.Edm.Tests.Csdl
   }
 }");
         }
+
+        [Fact]
+        public void VerifyKeyAliasForComplexPropertiesIsSupported()
+        {
+            // Arrange
+            var model = new EdmModel();
+            var entity = new EdmEntityType("NS1", "Product");
+            var complexType = new EdmComplexType("NS1", "ExampleComplexObject");
+            complexType.AddStructuralProperty("Name", EdmCoreModel.Instance.GetString(false));
+            var keyProperty = entity.AddStructuralPropertyAlias(new EdmComplexTypeReference(complexType, isNullable: false), "ExampleObjectName", new List<string> { "ExampleObject", "Name" });
+            entity.AddKeys(keyProperty);
+            entity.AddStructuralProperty("Name", EdmCoreModel.Instance.GetString(false));
+            entity.AddStructuralProperty("UpdatedTime", EdmCoreModel.Instance.GetDate(false));
+
+            model.AddElement(entity);
+            model.AddElement(complexType);
+
+            var entityContainer = new EdmEntityContainer("NS1", "Container");
+            model.AddElement(entityContainer);
+            EdmEntitySet set1 = new EdmEntitySet(entityContainer, "Products", entity);
+            entityContainer.AddElement(set1);
+
+            // Act & Assert for XML
+            WriteAndVerifyXml(model, "<?xml version=\"1.0\" encoding=\"utf-16\"?>" +
+                "<edmx:Edmx Version=\"4.0\" xmlns:edmx=\"http://docs.oasis-open.org/odata/ns/edmx\">" +
+                  "<edmx:DataServices>" +
+                    "<Schema Namespace=\"NS1\" xmlns=\"http://docs.oasis-open.org/odata/ns/edm\">" +
+                      "<EntityType Name=\"Product\">" +
+                        "<Key>" +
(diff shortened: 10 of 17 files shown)


src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +112,36 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+                                {
+                                    structuralProperty = structuralProperty.Type.AsComplex().FindProperty(keyPropertySegments[i]) as IEdmStructuralProperty;
+                                }
+                                else if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Entity)
+                                {
+                                    structuralProperty = structuralProperty.Type.AsEntity().FindProperty(keyPropertySegments[i]) as IEdmStructuralProperty;
+                                }
+                                else
+                                {
+                                    throw new ArgumentNullException();
```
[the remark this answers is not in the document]
ElizabethOkerio: Yes. Also it is not an entity type.

src/Microsoft.OData.Edm/Csdl/Parsing/SchemaJsonParser.cs
```diff
@@ -655,13 +655,20 @@ internal static CsdlKey ParseCsdlKey(string name, JsonElement keyArray, JsonPars
...
-                    p.ReportError(EdmErrorCode.InvalidKeyValue, "It doesn't support the key object.");
+                    string pName = "";
+                    string pValue = "";
+                    v.ParseAsObject(p, (keyAlias, keyName) =>
+                    {
+                        pName = keyName.ParseAsString(p);
+                        pValue = keyAlias;
+                    });
+
+                    return new CsdlPropertyReference(pName, pValue, context.Location());
```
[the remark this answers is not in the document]
ElizabethOkerio: yes

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +112,36 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+                    {
+                        string[] keyPropertySegments = keyProperty.PropertyName.Split('/');
+                        if (keyPropertySegments.Length > 0)
+                        {
+                            structuralProperty = this.FindProperty(keyPropertySegments[0]) as IEdmStructuralProperty;
+                            for (int i = 1; i < keyPropertySegments.Length; i++)
+                            {
+                                if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Complex)
+                                {
+                                    structuralProperty = structuralProperty.Type.AsComplex().FindProperty(keyPropertySegments[i]) as IEdmStructuralProperty;
```
habbes, line 126: `structuralProperty` could be set to null here as well.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +112,36 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+                                }
+                                else if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Entity)
+                                {
+                                    structuralProperty = structuralProperty.Type.AsEntity().FindProperty(keyPropertySegments[i]) as IEdmStructuralProperty;
```
habbes, line 130: `structuralProperty` could be set to `null` here as well.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +112,36 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+                                }
+                                else
+                                {
+                                    key.Add(new UnresolvedProperty(this, keyProperty.PropertyName, this.Location));
+                                }
+                            }
+                            structuralProperty = new EdmStructuralPropertyAlias(structuralProperty.DeclaringType, structuralProperty.Name, structuralProperty.Type, keyProperty.Alias, keyPropertySegments);
```
habbes, line 137: It's possible that the last segment is a complex or entity type. I think it has to be a primitive for this to be valid. Is this something we want to validate, or is it intentional to let it through?

In the current version we are also not checking whether the final `structureProperty` is a primitive type. Maybe it's by design that we don't check. But just thought this should be mentioned clearly to eliminate any doubt or confusion.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
```diff
@@ -153,7 +153,17 @@ internal override void WriteDeclaredKeyPropertiesElementHeader()
         internal override void WritePropertyRefElement(IEdmStructuralProperty property)
         {
             this.xmlWriter.WriteStartElement(CsdlConstants.Element_PropertyRef);
-            this.WriteRequiredAttribute(CsdlConstants.Attribute_Name, property.Name, EdmValueWriter.StringAsXml);
+            var prop = property as IEdmStructuralPropertyAlias;
```
habbes, line 156:
```suggestion
            IEdmStructuralPropertyAlias propertyAlias = property as IEdmStructuralPropertyAlias;

```

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
```diff
@@ -0,0 +1,62 @@
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EdmStructuralPropertyAlias"/> class.
+        /// </summary>
+        /// <param name="declaringType">The type that owns this property.</param>
+        /// <param name="name">Name of the property.</param>
+        /// <param name="type">The type of the property.</param>
+        /// <param name="alias">The property alias.</param>
+        /// <param name="path">The path to the referenced property.</param>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", MessageId = "string", Justification = "defaultValue might be confused for an IEdmValue.")]
```
habbes, line 27: What does this suppression message apply to? Is it meant for any of the constructor arguments? None of them contains the word `string` and none of them is called `defaultValue`

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
```diff
@@ -0,0 +1,62 @@
...
+            this.alias = alias;
+            this.path = path;
+        }
+
+        public string PropertyAlias
+        {
+            get { return this.alias; }
+
+        }
+        public IEnumerable<string> Path
```
habbes, line 40: nitpick: new line after `}`

```suggestion

        public IEnumerable<string> Path

```

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// <param name="path">The path to the referenced property.</param>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", MessageId = "string", Justification = "defaultValue might be confused for an IEdmValue.")]
+        public EdmStructuralPropertyAlias(IEdmStructuredType declaringType, string name, IEdmTypeReference type, string alias, IEnumerable<string> path)
+            : base(declaringType, name, type)
+        {
+            this.alias = alias;
+            this.path = path;
+        }
+
+        public string PropertyAlias
```
habbes, line 35: Add doc comments to public members.  Or `/// <inherit />`?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +112,36 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
                 List<IEdmStructuralProperty> key = new List<IEdmStructuralProperty>();
                 foreach (CsdlPropertyReference keyProperty in this.entity.Key.Properties)
                 {
-                    IEdmStructuralProperty structuralProperty = this.FindProperty(keyProperty.PropertyName) as IEdmStructuralProperty;
+                    IEdmStructuralProperty structuralProperty = null;
+                    if (!string.IsNullOrEmpty(keyProperty?.Alias))
```
habbes, line 116: suggestion: Since this block is kind of complex and easy to get lost in with all the `if` statements (and the edge cases that are still missing [see comments below]), It might be a good idea to extract it into a separate method.

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
```diff
@@ -157,6 +158,37 @@ public void AddKeys(IEnumerable<IEdmStructuralProperty> keyProperties)
...
 
+                //The Key property is the last segment in the path segment. 
+                IEdmStructuralPropertyAlias structuralPropertyAlias = property as IEdmStructuralPropertyAlias;
+                if (!string.IsNullOrEmpty(structuralPropertyAlias?.PropertyAlias))
+                {
+                    string[] keyPropertySegments = structuralPropertyAlias.Path.ToArray();
+                    IEdmStructuralProperty structuralProperty = this.FindProperty(keyPropertySegments[0]) as IEdmStructuralProperty;
+                    for (int i = 1; i < keyPropertySegments.Length; i++)
+                    {
+                        if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Complex)
```
habbes, line 169: Potential `NullReferenceException` here because `structuralProperty` is potentially `null` here.

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
```diff
@@ -157,6 +158,37 @@ public void AddKeys(IEnumerable<IEdmStructuralProperty> keyProperties)
...
+                        }
+                        else
+                        {
+                            throw new InvalidOperationException(Strings.Bad_UnresolvedType(structuralProperty));
+                        }
+                    }
+
+                    if (structuralProperty != null)
+                    {
+                        structuralProperty = new EdmStructuralPropertyAlias(structuralProperty.DeclaringType, structuralProperty.Name, structuralProperty.Type, structuralPropertyAlias.PropertyAlias, keyPropertySegments);
```
habbes, line 185: Why create a new `EdmStructuralPropertyAlias` if the `property` is already an `IEdmStructuralPropertyAlias`

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
```diff
@@ -157,6 +158,37 @@ public void AddKeys(IEnumerable<IEdmStructuralProperty> keyProperties)
...
+                    }
+
+                    this.declaredKey.Add(structuralProperty);
+                    return;
+                }
+
```
habbes, line 191: question: It seems as though if the property is an `IEdmStructuralPropertyAlias` but the `PropertyAlias` is null or empty, we process it as if it was not a property alias. Is this intentional?

Since you're already validating that the property path can be resolved, I think it would make sense to throw an exception if the alias is empty.

src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
```diff
@@ -163,6 +164,20 @@ public EdmNavigationProperty AddUnidirectionalNavigation(EdmNavigationPropertyIn
             return property;
         }
 
+        /// <summary>
+        /// Creates and adds a structural property alias to this type.
+        /// </summary>
+        /// <param name="declaring">Name of the property.</param>
+        /// <param name="type">Type of the property.</param>
```
habbes, line 171: These don't match the parameters of the method.

src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlPropertyReference.cs
```diff
@@ -12,16 +12,29 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Ast
     internal class CsdlPropertyReference : CsdlElement
     {
         private readonly string propertyName;
+        private readonly string alias;
```
gathogojr, line 15:
```suggestion
        private readonly string propertyAlias;
```

src/Microsoft.OData.Edm/Csdl/Parsing/Ast/CsdlPropertyReference.cs
```diff
@@ -12,16 +12,29 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Ast
...
         private readonly string propertyName;
+        private readonly string alias;
 
         public CsdlPropertyReference(string propertyName, CsdlLocation location)
             : base(location)
         {
             this.propertyName = propertyName;
         }
 
+        public CsdlPropertyReference(string propertyName, string alias, CsdlLocation location)
```
gathogojr, line 23:
```suggestion
        public CsdlPropertyReference(string propertyName, string propertyAlias, CsdlLocation location)
```

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +112,36 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
                 List<IEdmStructuralProperty> key = new List<IEdmStructuralProperty>();
                 foreach (CsdlPropertyReference keyProperty in this.entity.Key.Properties)
                 {
-                    IEdmStructuralProperty structuralProperty = this.FindProperty(keyProperty.PropertyName) as IEdmStructuralProperty;
+                    IEdmStructuralProperty structuralProperty = null;
+                    if (!string.IsNullOrEmpty(keyProperty?.Alias))
+                    {
+                        string[] keyPropertySegments = keyProperty.PropertyName.Split('/');
```
gathogojr, line 118: Curious, is it the `PropertyName` property that would have the forward slashes (`/`) or the `Alias`?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
gathogojr, line 137:
```suggestion
                            }
                            
                            structuralProperty = new EdmStructuralPropertyAlias(structuralProperty.DeclaringType, structuralProperty.Name, structuralProperty.Type, keyProperty.Alias, keyPropertySegments);
```

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +112,36 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+                        if (keyPropertySegments.Length > 0)
+                        {
+                            structuralProperty = this.FindProperty(keyPropertySegments[0]) as IEdmStructuralProperty;
+                            for (int i = 1; i < keyPropertySegments.Length; i++)
+                            {
+                                if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Complex)
+                                {
+                                    structuralProperty = structuralProperty.Type.AsComplex().FindProperty(keyPropertySegments[i]) as IEdmStructuralProperty;
+                                }
+                                else if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Entity)
```
gathogojr, line 128: I thought the segments in the key property can only reference properties declared using primitive or complex type...
Would this imply that a segment can reference a navigational property?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
gathogojr, line 137: Curious, do we perform any validation to verify that the referenced property is not a collection property? For example, if a segment references a collection property `FavoriteColors` of strings that may hold values `["Black","White"]` would that property still be valid to have in the key alias?

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
```diff
@@ -0,0 +1,62 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.OData.Edm
+{
+    /// <summary>
+    /// Represents an EDM structural (i.e. non-navigation) property with an alias.
+    /// </summary>
+    public class EdmStructuralPropertyAlias : EdmProperty, IEdmStructuralPropertyAlias
+    {
+        private string alias;
```
gathogojr, line 16:
```suggestion
        private string propertyAlias;
```

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
```diff
@@ -157,6 +158,37 @@ public void AddKeys(IEnumerable<IEdmStructuralProperty> keyProperties)
...
+                        {
+                            structuralProperty = structuralProperty.Type.AsComplex().FindProperty(keyPropertySegments[i]) as IEdmStructuralProperty;
+                        }
+                        else if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Entity)
+                        {
+                            structuralProperty = structuralProperty.Type.AsEntity().FindProperty(keyPropertySegments[i]) as IEdmStructuralProperty;
+                        }
+                        else
+                        {
+                            throw new InvalidOperationException(Strings.Bad_UnresolvedType(structuralProperty));
```
gathogojr, line 179: How come in _src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs_ you don't throw an exception here?

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
```diff
@@ -157,6 +158,37 @@ public void AddKeys(IEnumerable<IEdmStructuralProperty> keyProperties)
                     this.declaredKey = new List<IEdmStructuralProperty>();
                 }
 
+                //The Key property is the last segment in the path segment. 
+                IEdmStructuralPropertyAlias structuralPropertyAlias = property as IEdmStructuralPropertyAlias;
+                if (!string.IsNullOrEmpty(structuralPropertyAlias?.PropertyAlias))
```
gathogojr, line 163: Can the logic in this block be placed in a location where it can be shared given that it looks similar to that in _src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs_?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
ElizabethOkerio: correct.

src/Microsoft.OData.Edm/Csdl/Parsing/SchemaJsonParser.cs
corranrogue9, line 671: Is there any other scenario where we reach this branch? The `ReportError` message doesn't seem specific to this specific feature (even if the comment is specific to the feature). Maybe we should be more a little bit more defensive, something like:

```suggestion
                    bool error = false;
                    string pName = null;
                    string pValue = null;
                    v.ParseAsObject(p, (keyAlias, keyName) =>
                    {
                        if (pName != null || pValue != null)
                        {
                            error = true;
                            return;
                        }
                        
                        pName = keyName.ParseAsString(p);
                        pValue = keyAlias;
                    });
                    if (error || pName == null || pValue == null)
                    {
                        p.ReportError(EdmErrorCode.InvalidKeyValue, "It doesn't support the key object.");
                    }
                    else
                    {
                        return new CsdlPropertyReference(pName, pValue, context.Location());
                    }
```

Alternatively, we could do some `Debug.Assert`s if we want to be a little more flexible, but that might actually result in bigger issues for customers if something goes wrong.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -112,7 +112,36 @@ private IEnumerable<IEdmStructuralProperty> ComputeDeclaredKey()
...
+                                    key.Add(new UnresolvedProperty(this, keyProperty.PropertyName, this.Location));
+                                }
+                            }
+                            structuralProperty = new EdmStructuralPropertyAlias(structuralProperty.DeclaringType, structuralProperty.Name, structuralProperty.Type, keyProperty.Alias, keyPropertySegments);
+                        }
+                    }
+                    else
+                    {
+                        structuralProperty = this.FindProperty(keyProperty.PropertyName) as IEdmStructuralProperty;
+                    }
```
corranrogue9, line 143: The nested `if` above means that we won't always reach this branch even if we should fallback to the existing behavior. 

```suggestion
                    structuralProperty = structuralProperty ?? this.FindProperty(keyProperty.PropertyName) as IEdmStructuralProperty;
```

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
```diff
@@ -0,0 +1,62 @@
...
+        public IEnumerable<string> Path
+        {
+            get { return this.path; }
+        }
+
+        /// <summary>
+        /// Gets the default value of this property.
+        /// </summary>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", Justification = "defaultValue might be confused for an IEdmValue.")]
+        public string DefaultValueString
```
corranrogue9, line 49: Is this being used anywhere?

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
corranrogue9: Actually, I think it makes the most sense to go one step further even: the `EdmStructuralPropertyAlias` constructor should do this work, otherwise there are far too many validations that need to happen in order to ensure the `EdmStructuralPropertyAlias` is not malformed.

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
```diff
@@ -0,0 +1,62 @@
...
+        /// <summary>
+        /// Initializes a new instance of the <see cref="EdmStructuralPropertyAlias"/> class.
+        /// </summary>
+        /// <param name="declaringType">The type that owns this property.</param>
+        /// <param name="name">Name of the property.</param>
+        /// <param name="type">The type of the property.</param>
+        /// <param name="alias">The property alias.</param>
+        /// <param name="path">The path to the referenced property.</param>
+        [System.Diagnostics.CodeAnalysis.SuppressMessage("Microsoft.Naming", "CA1720:IdentifiersShouldNotContainTypeNames", MessageId = "string", Justification = "defaultValue might be confused for an IEdmValue.")]
+        public EdmStructuralPropertyAlias(IEdmStructuredType declaringType, string name, IEdmTypeReference type, string alias, IEnumerable<string> path)
```
corranrogue9, line 28: With all of this being `public`, we should probably validate everything. For example, don't we need to check the consistency of `declaringType`, `name`, and `type`? And shouldn't `name` always be the last element of `path`? And shouldn't `path` always have at least two elements?

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
corranrogue9: See my comment on the constructor, there just so much validation that would need to happen

src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
```diff
@@ -163,6 +164,20 @@ public EdmNavigationProperty AddUnidirectionalNavigation(EdmNavigationPropertyIn
...
+        /// <summary>
+        /// Creates and adds a structural property alias to this type.
+        /// </summary>
+        /// <param name="declaring">Name of the property.</param>
+        /// <param name="type">Type of the property.</param>
+        /// <returns>Created structural property.</returns>
+        public EdmStructuralPropertyAlias AddStructuralPropertyAlias(IEdmTypeReference type, string alias, IEnumerable<string> path)
+        {
+            string propertyName = path.FirstOrDefault();
+            EdmStructuralPropertyAlias property = new EdmStructuralPropertyAlias(this, propertyName, type, alias, path);
```
corranrogue9, line 176: We should probably confirm that `this` type does indeed have properties that follow `path` and result in a property with name `propertyName` and of type `type`.

src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
```diff
@@ -163,6 +164,20 @@ public EdmNavigationProperty AddUnidirectionalNavigation(EdmNavigationPropertyIn
...
 
+        /// <summary>
+        /// Creates and adds a structural property alias to this type.
+        /// </summary>
+        /// <param name="declaring">Name of the property.</param>
+        /// <param name="type">Type of the property.</param>
+        /// <returns>Created structural property.</returns>
+        public EdmStructuralPropertyAlias AddStructuralPropertyAlias(IEdmTypeReference type, string alias, IEnumerable<string> path)
+        {
+            string propertyName = path.FirstOrDefault();
```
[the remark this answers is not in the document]
ElizabethOkerio: This is the complex property that is being added to the entity type and not the key property.  The key property is what is the last element in the path. This method was specifically added to support adding a complex property which is a link to a property with an alias.

src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
ElizabethOkerio: Am not sure I understand this. So, here we are adding a property - `propertyName` of type `type` to `this` type

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
ElizabethOkerio: Most of the validation are done when the `base` gets called. Like the `declaringType` and `type`. The `name` is the first element of the path. This is the `complex property` in the entity type that acts like a `base` for the path to the `type` that has the `primitive type` which will be used as key for the entity type.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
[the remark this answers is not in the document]
ElizabethOkerio: Because we are the ones creating this string and not the client - I think we should take care of all this things at the point of creating the string when writing to the csdl.

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
ElizabethOkerio: This is a property of the `IEdmStructuralProperty`. It is not being used anywhere though.

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
ElizabethOkerio: The `EdmStructuralPropertyAlias` is not the `property` with an alias in this case.

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
ElizabethOkerio: The `property` is not the key property - This is the complex property - that is why we need to loop through the path segments to get to the last path segment which will be the desired key property and create an `EdmStructuralPropertyAlias` for it.

src/Microsoft.OData.Edm/Csdl/Parsing/SchemaJsonParser.cs
ElizabethOkerio: so before we were returning the error because the feature was not supported. Now that the feature is being supported, there is no need to return the error.

corranrogue9: I guess I'm just making sure that the *only* time we return that error is because the feature was not supported. Is that right?

src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
corranrogue9: I'm sure most of my comments around this has been nonsensical regarding what an alias means, but I think 
my points stands. Hopefully this example will help make sure I'm understanding correctly. There is a type `buzz` with a property `foo/bar/fizz` that is of type `Edm.Decimal`. The caller wishes to add an alias to `buzz` called `frob` that points to `foo/bar/fizz`. 

In this case, doesn't `frob` *have* to be of type `Edm.Decimal`? If that's the case, then what I'm suggesting is that we need to walk the `path` parameter in this method to ensure that:
1. Each segment of the path is in fact valid for `this`
2. The type of the property pointed to by `path` must match `type`

ElizabethOkerio: The `key` for the type `buzz` is not in `buzz`. The key is in another type let's `fizz`. But the type `buzz` has a complex property `bar` whose type is `fizz` and `fizz` has a property `Id` whose type is `Edm.Int`. This is a 2-segment path. we could have a longer segment than that but that should be the flow. So the key for `buzz` will be represented as `bar/Id`.  The alias is an identifier for this path `bar/Id`. Instead of using this path, we use an alias which does not have the slashes and that is easy to work with. The Alias will always be a string.

ElizabethOkerio: so the method `AddStructuralPropertyAlias` is used to add `bar` to `buzz` but because this property `bar` acts a link to the actual `key` property for type `buzz`, that's why we provide more information to it like the alias and the key path. The path is later processed to get to the last path segment to get the real `key` property for the type.

src/Microsoft.OData.Edm/Csdl/Parsing/SchemaJsonParser.cs
habbes: Since the feature *is* supported, I think that error message would be misleading. If we want to validate against null key or value, I think the error message should be adjusted accordingly.

src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
corranrogue9:
```
var model = new EdmModel();
var entity = new EdmEntityType("NS1", "Product");
var complexType = new EdmComplexType("NS1", "ExampleComplexObject");
complexType.AddStructuralProperty("Name", EdmCoreModel.Instance.GetString(false));
var keyProperty = entity.AddStructuralPropertyAlias(new EdmComplexTypeReference(complexType, isNullable: false), "ExampleObjectName", new List<string> { "ExampleObject", "Name" });
entity.AddKeys(keyProperty);
entity.AddStructuralProperty("Name", EdmCoreModel.Instance.GetString(false));
```

In this test that you wrote, for this line: `var keyProperty = entity.AddStructuralPropertyAlias(new EdmComplexTypeReference(complexType, isNullable: false), "ExampleObjectName", new List<string> { "ExampleObject", "Name" });`, doesn't this *imply* that the `Product` entity has a property named `ExampleObject` of type `ExampleComplexObject` and that `ExampleComplexObject` has a property named `Name`? There are two things that are weird to me:
1. Can't I also call `var keyProperty = entity.AddStructuralPropertyAlias(new EdmComplexTypeReference(complexType, isNullable: false), "ExampleObjectName", new List<string> { "ExampleObject",  "*Garrett*" });`, even though that would be incorrect (because there is no property named "*Garrett*" on `ExampleComplexObject`)?
2. Why is it that the `ExampleObject` property of type `ExampleComplexObject` is added implicitly to `Product` when this method is called? Shouldn't I need to call `entity.AddStructuralProperty("ExampleObject", complexType)` first?

ElizabethOkerio: @corranrogue9 
1. For the first case, Yes the user can pass   this ` "*Garrett*" ` that does not exist in `"ExampleObjectName"` but this will lead to model validation errors and an exception will be thrown. 
2. For the second case, If you want to support key alias in your model, then you use the `entity.AddStructuralPropertyAlias(...)` method  instead of using `entity.AddStructuralProperty(..)`

corranrogue9: Number 2 makes sense now, thanks for explaining that. 

For number 1, my point is, when we internally instantiate `EdmStructuralPropertyAlias`, we only do so through the use of `PropertyAliasHelpers.GetKeyProperty`, which guarantees the correctness of the alias in the model. If we moved the logic in that method to the constructor of `EdmStructuralPropertyAlias`, we would guarantee the same correctness, but for *non-internal* callers. It was *also* allow us to have access to both the original path and the split path segments (since they could be computed by the constructor), so we wouldn't need to call `string.Join` on the path in other parts of the code.

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
habbes: @ElizabethOkerio what was the conclusion to this thread?

habbes: I'm confused a bit. If the property here is already an `IEdmStructuralPropertyAlias` why do you need to possibly duplicate `EdmStructuralPropertyAlias` via `PropertyAliasHelpers.GetKeyProperty()` instead of adding the property directly to the declared key?

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
```diff
@@ -2810,6 +2810,85 @@ public void ShouldSubstituteFullyQualifiedNamespaceWithAliasIfAliasIsSet()
       }
     }
   }
+}");
+        }
+
+        [Fact]
+        public void VerifyKeyAliasForComplexPropertiesIsSupported()
```
habbes, line 2817: What about aliases that go through entity navigation properties?

src/Microsoft.OData.Edm/Csdl/Parsing/SchemaJsonParser.cs
ElizabethOkerio: the key cannot be null here.

src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
corranrogue9: For performance, open/close principle, and single-responsibility reasons, even if the model validation catches it, the logic should move to the constructor.

habbes: @corranrogue9 I had @xuzhg mention that by design EDM does not validate the model unless the user explicitly chooses to do so. But maybe the context was different.

src/Microsoft.OData.Edm/PropertyAliasHelpers.cs
```diff
@@ -0,0 +1,42 @@
...
+//---------------------------------------------------------------------
+
+using System;
+using System.Linq;
+
+namespace Microsoft.OData.Edm
+{
+    internal static class PropertyAliasHelpers
+    {
+        internal static IEdmStructuralProperty GetKeyProperty(IEdmStructuralProperty property, string[] keyPropertySegments, string propertyAlias)
```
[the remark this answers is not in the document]
ElizabethOkerio: This method loops through the path segments to get to the last segment which is the key property.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
ElizabethOkerio: done

src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
ElizabethOkerio: This logic is used in computing the declared key. I don't know whether it can move to the constructor.

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
ElizabethOkerio: @habbes The `IEdmStructuralPropertyAlias` is for the first segment but this method  `PropertyAliasHelpers.GetKeyProperty()`  returns `IEdmStructuralPropertyAlias` for the last segment..

src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
ElizabethOkerio: @corranrogue9 So when we are adding the `complex property` to the model, we call `AddStructuralPropertyAlias`. Within this method we call `EdmStructuralPropertyAlias`. We can move the logic for checking the path and everything to the constructor. This will check for validity of the path. When we call `AddKeys` method for adding the key to the model, if we say that we pass in the last segment on the path and not do any validation because we already did it in the constructor, I think this will open up some loopholes where someone can just pass in something that is not even part of the path. That's why when we are adding the keys to the entity type's `DeclaredKey`, we still have to go through the path and make sure that whatever we are adding to the `DeclaredKey` is in the path and is the last segment on the path.

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
ElizabethOkerio: @habbes  am not sure I understand this. So we are the ones who create the `EdmStructuralPropertyAlias ` once we've gotten to the last segment on the path.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
ElizabethOkerio: We check the type of referenced property, if it's not of type complex or entity, then we throw an exception.

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
ElizabethOkerio: Am not sure there is any fast way of getting the last segment.. When we call the `AddKeys` method, we pass in the first segment property and the path. So to get to the last segment we have to traverse the whole path. If we decide on passing the last segment property to the `AddKeys` method, that will mean we are assuming the path is valid and no path traversal is needed.

src/Microsoft.OData.Edm/Csdl/Parsing/SchemaJsonParser.cs
```diff
@@ -655,13 +655,20 @@ internal static CsdlKey ParseCsdlKey(string name, JsonElement keyArray, JsonPars
             {
                 if (v.ValueKind == JsonValueKind.Object)
                 {
-                    // TODO: ODL doesn't support the key referencing a property of a complex type as below.
                     // "$Key": [
                     //    {
                     //      "EntityInfoID": "Info/ID"
                     //    }
                     //  ]
```
gathogojr, line 662: Should have removed these lines as you removed the TODO, no?

src/Microsoft.OData.Edm/Csdl/Parsing/SchemaJsonParser.cs
```diff
@@ -655,13 +655,20 @@ internal static CsdlKey ParseCsdlKey(string name, JsonElement keyArray, JsonPars
...
                 if (v.ValueKind == JsonValueKind.Object)
                 {
-                    // TODO: ODL doesn't support the key referencing a property of a complex type as below.
                     // "$Key": [
                     //    {
                     //      "EntityInfoID": "Info/ID"
                     //    }
                     //  ]
-                    p.ReportError(EdmErrorCode.InvalidKeyValue, "It doesn't support the key object.");
+                    string pName = "";
```
gathogojr, line 663: Is there a variable by the name `propertyName` in scope such that we cannot use that in place of `pName`?

src/Microsoft.OData.Edm/Csdl/Parsing/SchemaJsonParser.cs
```diff
@@ -655,13 +655,20 @@ internal static CsdlKey ParseCsdlKey(string name, JsonElement keyArray, JsonPars
...
+                    string pValue = "";
```
gathogojr, line 664: Is there a variable by the name `propertyValue` in scope such that we cannot use that in place of `pValue`?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -107,37 +107,57 @@ private IEdmEntityType ComputeBaseType()
...
+                        structuralProperty = this.FindProperty(keyPropertySegments[0]) as IEdmStructuralProperty;
+                        if (structuralProperty == null)
+                        {
+                            return null;
+                        }
+
+                        structuralProperty = PropertyAliasHelpers.GetKeyProperty(structuralProperty, keyPropertySegments, keyProperty.PropertyAlias);
+                    }
+
+                }
```
gathogojr, line 133:
```suggestion
                    }
                }
```

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -107,37 +107,57 @@ private IEdmEntityType ComputeBaseType()
...
+                    string[] keyPropertySegments = keyProperty.PropertyName.Split('/');
+                    if (keyPropertySegments.Length > 0)
+                    {
+                        structuralProperty = this.FindProperty(keyPropertySegments[0]) as IEdmStructuralProperty;
+                        if (structuralProperty == null)
+                        {
+                            return null;
+                        }
+
+                        structuralProperty = PropertyAliasHelpers.GetKeyProperty(structuralProperty, keyPropertySegments, keyProperty.PropertyAlias);
```
gathogojr, line 130: Given that `structuralProperty` was assigned a value 5 or 6 line above here, could you add a comment above here to give someone an idea what happens when you call `PropertyAliasHelpers.GetKeyProperty` and what new value `structuralProperty` assumes

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
```diff
@@ -153,7 +153,17 @@ internal override void WriteDeclaredKeyPropertiesElementHeader()
...
+            IEdmStructuralPropertyAlias prop = property as IEdmStructuralPropertyAlias;
+            if (prop == null)
+            {
+                this.WriteRequiredAttribute(CsdlConstants.Attribute_Name, property.Name, EdmValueWriter.StringAsXml);
+            }
+            else
+            {
+                this.WriteRequiredAttribute(CsdlConstants.Attribute_Name, string.Join("/", prop.Path), EdmValueWriter.StringAsXml);
+                this.WriteRequiredAttribute(CsdlConstants.Attribute_Alias, prop.PropertyAlias, EdmValueWriter.StringAsXml);
+            }
```
gathogojr, line 165:
```suggestion
            if (property is IEdmStructuralPropertyAlias prop)
            {
                this.WriteRequiredAttribute(CsdlConstants.Attribute_Name, string.Join("/", prop.Path), EdmValueWriter.StringAsXml);
                this.WriteRequiredAttribute(CsdlConstants.Attribute_Alias, prop.PropertyAlias, EdmValueWriter.StringAsXml);
            }
            else
            {
                this.WriteRequiredAttribute(CsdlConstants.Attribute_Name, property.Name, EdmValueWriter.StringAsXml);
            }
```

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaJsonWriter.cs
```diff
@@ -208,8 +208,17 @@ internal override void WritePropertyRefElement(IEdmStructuralProperty property)
         {
             // Key properties without a key alias are represented as strings containing the property name.
             // Key properties with a key alias are represented as objects with one member whose name is the key alias and whose value is a string containing the path to the property.
-            // TODO: It seems the second one is not supported.
-            this.jsonWriter.WriteStringValue(property.Name);
+            if (property is IEdmStructuralPropertyAlias propertyAlias)
+            {
+                this.jsonWriter.WriteStartObject();
+                this.jsonWriter.WritePropertyName(propertyAlias.PropertyAlias);
+                this.jsonWriter.WriteStringValue(string.Join("/", propertyAlias.Path));
```
gathogojr, line 215: This `join` operation is done in quite a few places. Could we have taken care of it by including a `readonly` property to the `IEdmStructuralPropertyAlias` interface and it's derivations that returns the concatenated thing? We could even apply an optimization there by doing the concatenation only once when the getter is first called and saving the result in a private field

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.OData.Edm
+{
+    /// <summary>
+    /// Represents an EDM structural (i.e. non-navigation) property with an alias.
+    /// </summary>
+    public class EdmStructuralPropertyAlias : EdmProperty, IEdmStructuralPropertyAlias
+    {
+        private string propertyAlias;
```
gathogojr, line 17: Make `propertyAlias` readonly

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
```diff
@@ -0,0 +1,75 @@
...
+        private IEnumerable<string> path;
```
gathogojr, line 18: Make `path` readonly

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// </summary>
+        /// <param name="declaringType">The type that owns this property.</param>
+        /// <param name="name">Name of the property.</param>
+        /// <param name="type">The type of the property.</param>
+        /// <param name="alias">The property alias.</param>
+        /// <param name="path">The path to the referenced property.</param>
+        public EdmStructuralPropertyAlias(IEdmStructuredType declaringType, string name, IEdmTypeReference type, string propertyAlias, IEnumerable<string> path)
+            : base(declaringType, name, type)
+        {
+            if (propertyAlias == null)
```
gathogojr, line 31: Is an empty string valid as property alias?

src/Microsoft.OData.Edm/EdmStructuralPropertyAlias.cs
```diff
@@ -0,0 +1,75 @@
...
+            {
+                throw new ArgumentNullException(nameof(propertyAlias));
+            }
+
+            if (path == null)
```
gathogojr, line 36: Is an empty collection valid as path?

src/Microsoft.OData.Edm/PropertyAliasHelpers.cs
gathogojr, line 14: Document this method so it's clear what it does

gathogojr: Is this method called from multiple places? If no, would it have been better for it to reside in the class from where it's called seeing it's the only method in this class and file...

src/Microsoft.OData.Edm/PropertyAliasHelpers.cs
```diff
@@ -0,0 +1,42 @@
...
+        {
+            IEdmStructuralProperty structuralProperty = property;
+            for (int i = 1; i < keyPropertySegments.Length; i++)
+            {
+                if (structuralProperty.Type.Definition.TypeKind == EdmTypeKind.Complex)
```
gathogojr, line 19: Could have used a switch...

src/Microsoft.OData.Edm/Schema/EdmEntityType.cs
```diff
@@ -157,7 +158,28 @@ public void AddKeys(IEnumerable<IEdmStructuralProperty> keyProperties)
...
-                this.declaredKey.Add(property);
+                IEdmStructuralPropertyAlias structuralPropertyAlias = property as IEdmStructuralPropertyAlias;
+                if (!string.IsNullOrEmpty(structuralPropertyAlias?.PropertyAlias))
+                {
+                    string[] keyPropertySegments = structuralPropertyAlias.Path.ToArray();
+                    IEdmStructuralProperty structuralProperty = property.DeclaringType.FindProperty(keyPropertySegments[0]) as IEdmStructuralProperty;
+                    if (structuralProperty != null)
+                    {
+                        IEdmStructuralProperty prop = PropertyAliasHelpers.GetKeyProperty(structuralProperty, keyPropertySegments, structuralPropertyAlias.PropertyAlias);
+                        if (prop != null)
```
gathogojr, line 169: There's nothing we should do if `prop` is null?

src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
```diff
@@ -163,6 +164,25 @@ public EdmNavigationProperty AddUnidirectionalNavigation(EdmNavigationPropertyIn
...
+        /// <param name="alias">The property alias.</param>
+        /// <param name="path">The key path segments</param>
+        /// <returns>Created structural property.</returns>
+        public EdmStructuralPropertyAlias AddStructuralPropertyAlias(IEdmTypeReference type, string alias, IEnumerable<string> path)
+        {
+            if (path == null)
+            {
+                throw new ArgumentNullException(nameof(path));
+            }
+            string propertyName = path.FirstOrDefault();
```
gathogojr, line 180:
```suggestion
            }

            string propertyName = path.FirstOrDefault();
```

src/Microsoft.OData.Edm/Schema/EdmStructuredType.cs
```diff
@@ -163,6 +164,25 @@ public EdmNavigationProperty AddUnidirectionalNavigation(EdmNavigationPropertyIn
...
+        /// <summary>
+        /// Creates and adds a structural property alias to this type.
+        /// </summary>
+        /// <param name="type">Type of the property.</param>
+        /// <param name="alias">The property alias.</param>
+        /// <param name="path">The key path segments</param>
+        /// <returns>Created structural property.</returns>
+        public EdmStructuralPropertyAlias AddStructuralPropertyAlias(IEdmTypeReference type, string alias, IEnumerable<string> path)
+        {
+            if (path == null)
```
gathogojr, line 176: Is an empty collection valid as path?

xuzhg:
@ElizabethOkerio Let's kick it off again:

here's my thought:

Basically, we can add a new interface:
```C#
    public interface IEdmPropertyRef: IEdmNamedElement
    {
        IEdmProperty Property { get; set; }

        string Path { get; set; }  // or IEdmPathExpression

        string Alias { get; set; }
    }

```
So, change:


```diff
public class EdmEntityType
{
-       private List<IEdmStructuralProperty> declaredKey;
+       private List<IEdmPropertyRef> declaredKey;

+       public void AddKeys(params IEdmPropertyRef[] keyProperties)
+       {}

+       public void AddKeys(params IEdmPropertyRef[] keyProperties)
+       {}
}
```

Change the existing AddKeys, and add others.

gathogojr:
Closed by https://github.com/OData/odata.net/pull/3362

PR #2418, opened 2022-05-30
Expose http client handler

### LINKED ISSUE #2417: Expose HttpClientHandler of the underlying request
Customer would like to override the server certificate validation logic. Since the PR to allow injecting a custom `HttpClient` is on hold, we can consider exposing the `HttpClientHandler` and other components of the underlying request to the public API.

With `HttpWebRequestMessage`, the underlying `HttpWebRequest` is accessible via a public property, making it possible to override some behaviour, including certificate validation.

With `HttpClientRequestMessage`, the underlying request, client and handler are private or internal.

### Assemblies affected

Microsoft.OData.Client 7.11.0

habbes:
### Issues

*This pull request fixes #2417 *

Expose the underyling request components of a `HttpClientRequestMessage` to provide more opportunities for customers to override or customize request behaviour (e.g. certificate validation).

### Description

- Makes `HttpClientRequestMessage.HttpRequestMessage` public
- Add public property `HttpClientRequestMessage.HttpClientHandler`

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
+++ b/src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
@@ -207,6 +207,17 @@ namespace Microsoft.OData.Client
 #endif
 
         /// <summary>
+        /// Returns the underlying <see cref="HttpClientHandler"/>.
+        /// </summary>
+        public HttpClientHandler HttpClientHandler
+        {
+            get
+            {
+                return _handler;
+            }
+        }
+
+        /// <summary>
         /// Returns the value of the header with the given name.
         /// </summary>
         /// <param name="headerName">Name of the header.</param>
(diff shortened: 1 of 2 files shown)


xuzhg (approved):

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -206,6 +206,17 @@ public override bool SendChunked
         }
 #endif
 
+        /// <summary>
+        /// Returns the underlying <see cref="HttpClientHandler"/>.
+        /// </summary>
+        public HttpClientHandler HttpClientHandler
```
corranrogue9, line 212: Having conversation offline for privacy, but adding it here so we don't lose track: these types implement `IDisposable`. We probably should not be exposing them.

corranrogue9 (changes requested):
We really need to address this `IDisposable` issue

habbes:
Closing this PR since it has been superseded by: https://github.com/OData/odata.net/pull/2418

PR #2429, opened 2022-06-08
Exposed the HttpClientHandler certificate validation callback in Http…

### LINKED ISSUE #2417: Expose HttpClientHandler of the underlying request

corranrogue9:
…ClientRequestMessage

### Issues

*This pull request fixes #2417.*

### DIFF
--- a/src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
+++ b/src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
@@ -14,8 +14,10 @@ namespace Microsoft.OData.Client
     using System.Net;
     using System.Net.Http;
     using System.Net.Http.Headers;
+#if NETSTANDARD2_0_OR_GREATER
     using System.Net.Security;
     using System.Security.Cryptography.X509Certificates;
+#endif
     using System.Threading.Tasks;
     using Microsoft.OData;
 
--- a/src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
+++ b/src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
@@ -14,6 +14,8 @@ namespace Microsoft.OData.Client
     using System.Net;
     using System.Net.Http;
     using System.Net.Http.Headers;
+    using System.Net.Security;
+    using System.Security.Cryptography.X509Certificates;
     using System.Threading.Tasks;
     using Microsoft.OData;
 
@@ -88,6 +90,24 @@ namespace Microsoft.OData.Client
                 this.SetHeader(keyValue.Key, keyValue.Value);
             }
         }
+
+#if NETSTANDARD2_0_OR_GREATER
+        /// <summary>
+        /// Gets the underlying <see cref="HttpClientHandler.ServerCertificateCustomValidationCallback"/> so that callers can customize their certificate valiation
+        /// during OData requests
+        /// </summary>
+        public Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool> ServerCertificateCustomValidationCallback
+        {
+            get
+            {
+                return this._handler.ServerCertificateCustomValidationCallback;
+            }
+            set
+            {
+                this._handler.ServerCertificateCustomValidationCallback = value;
+            }
+        }
+#endif
 
         /// <summary>
         /// Returns the collection of request headers.
--- a/src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
+++ b/src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
@@ -95,7 +95,7 @@ namespace Microsoft.OData.Client
 
 #if NETSTANDARD2_0_OR_GREATER
         /// <summary>
-        /// Gets the underlying <see cref="HttpClientHandler.ServerCertificateCustomValidationCallback"/> so that callers can customize their certificate valiation
+        /// Gets or sets the underlying <see cref="HttpClientHandler.ServerCertificateCustomValidationCallback"/> so that callers can customize their certificate valiation
         /// during OData requests
         /// </summary>
         public Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool> ServerCertificateCustomValidationCallback
--- a/test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
@@ -27,12 +27,16 @@ namespace Microsoft.OData.Client.Tests.Serialization
             var httpClientRequestMessage = new HttpClientRequestMessage(
                 new DataServiceClientRequestMessageArgs("GET", new Uri("http://localhost"), false, false, new Dictionary<string, string>()));
 
+            // set a predicate that is always true
             httpClientRequestMessage.ServerCertificateCustomValidationCallback = (_, __, ___, ____) => true;
             Assert.True(httpClientRequestMessage.ServerCertificateCustomValidationCallback(null, null, null, SslPolicyErrors.None));
 
+            // set a predicate that is always false
             httpClientRequestMessage.ServerCertificateCustomValidationCallback = (_, __, ___, ____) => false;
             Assert.False(httpClientRequestMessage.ServerCertificateCustomValidationCallback(null, null, null, SslPolicyErrors.None));
 
+    
+            // aggregate an always true predicate with an always false predicate
             httpClientRequestMessage.ServerCertificateCustomValidationCallback = (_, __, ___, ____) => true;
             Assert.True(httpClientRequestMessage.ServerCertificateCustomValidationCallback(null, null, null, SslPolicyErrors.None));
             httpClientRequestMessage.ServerCertificateCustomValidationCallback += (_, __, ___, ____) => false;
--- a/test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
 using System.Net.Http;
+using System.Net.Security;
 using System.Threading.Tasks;
 using Xunit;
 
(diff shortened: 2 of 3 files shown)


src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -89,6 +91,24 @@ public HttpClientRequestMessage(DataServiceClientRequestMessageArgs args)
             }
         }
 
+#if NETSTANDARD2_0_OR_GREATER
+        /// <summary>
+        /// Gets the underlying <see cref="HttpClientHandler.ServerCertificateCustomValidationCallback"/> so that callers can customize their certificate valiation
```
xuzhg, line 96: Gets / sets...,

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -89,6 +91,24 @@ public HttpClientRequestMessage(DataServiceClientRequestMessageArgs args)
             }
         }
 
+#if NETSTANDARD2_0_OR_GREATER
+        /// <summary>
+        /// Gets the underlying <see cref="HttpClientHandler.ServerCertificateCustomValidationCallback"/> so that callers can customize their certificate valiation
+        /// during OData requests
+        /// </summary>
+        public Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool> ServerCertificateCustomValidationCallback
```
xuzhg, line 99: Can we wrap those input/required parameters together as a CertificateValidationContext?
Then, it could be more clear to use.
And We can add more into the context later and without breaking anything?

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -89,6 +91,24 @@ public HttpClientRequestMessage(DataServiceClientRequestMessageArgs args)
...
+        /// </summary>
+        public Func<HttpRequestMessage, X509Certificate2, X509Chain, SslPolicyErrors, bool> ServerCertificateCustomValidationCallback
+        {
+            get
+            {
+                return this._handler.ServerCertificateCustomValidationCallback;
+            }
+            set
+            {
+                this._handler.ServerCertificateCustomValidationCallback = value;
```
xuzhg, line 107: shall we check the input 'null'?

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
```diff
@@ -18,5 +19,25 @@ public void UnwrapAggregateException()
...
         }
+
+#if NETCOREAPP2_0_OR_GREATER
+        [Fact]
+        public void SetCertificateValidationCallback()
+        {
+            var httpClientRequestMessage = new HttpClientRequestMessage(
+                new DataServiceClientRequestMessageArgs("GET", new Uri("http://localhost"), false, false, new Dictionary<string, string>()));
+
+            httpClientRequestMessage.ServerCertificateCustomValidationCallback = (_, __, ___, ____) => true;
```
xuzhg, line 30: any difference between line 30 - 31 and line 36- 37 ?

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
mikepizzo: I assume null is valid to clear the callback. Regardless, by passing it down we can simply reflect any validation that the HttpClientHandler may already be imposing, rather than masking lower level behavior/validation.

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
mikepizzo: +1

mikepizzo (dismissed):
:shipit:

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -14,6 +14,8 @@ namespace Microsoft.OData.Client
     using System.Net;
     using System.Net.Http;
     using System.Net.Http.Headers;
+    using System.Net.Security;
+    using System.Security.Cryptography.X509Certificates;
```
mikepizzo, line 18: should these also be #if NETSTANDARD2_0_OR_GREATER?

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
mikepizzo: We are just surfacing the property of the underlying handler; seems like it would be confusing to change how the parameters were specified, and would require them to have a different func than they use for custom certificate validation in other places.

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
corranrogue9:
```suggestion
        /// Gets or sets the underlying <see cref="HttpClientHandler.ServerCertificateCustomValidationCallback"/> so that callers can customize their certificate valiation
```

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
corranrogue9: Agreed with Mike here. @xuzhg, maybe this can be part of our "long-term 7.x" discussion, but I think here we should maintain the contract set out by `HttpClientHandler`. Is that alright for you?

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
corranrogue9: Not really, 36-37 were supposed to be part of the larger 36-39 portion. So, 30-31 is "check always true", 33-34 is "check always false", and 36-39 are "check an aggregation of always true and always false". 36 is really just resetting what we did in 33. I've added some comments to clarify.

mikepizzo (approved):
:shipit:

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
xuzhg: Emm... alright.

xuzhg (dismissed):

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
```diff
@@ -18,5 +19,29 @@ public void UnwrapAggregateException()
...
+                new DataServiceClientRequestMessageArgs("GET", new Uri("http://localhost"), false, false, new Dictionary<string, string>()));
+
+            // set a predicate that is always true
+            httpClientRequestMessage.ServerCertificateCustomValidationCallback = (_, __, ___, ____) => true;
+            Assert.True(httpClientRequestMessage.ServerCertificateCustomValidationCallback(null, null, null, SslPolicyErrors.None));
+
+            // set a predicate that is always false
+            httpClientRequestMessage.ServerCertificateCustomValidationCallback = (_, __, ___, ____) => false;
+            Assert.False(httpClientRequestMessage.ServerCertificateCustomValidationCallback(null, null, null, SslPolicyErrors.None));
+
```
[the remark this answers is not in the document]
corranrogue9:
```suggestion

```

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
```diff
@@ -18,5 +19,28 @@ public void UnwrapAggregateException()
...
+            // set a predicate that is always false
+            httpClientRequestMessage.ServerCertificateCustomValidationCallback = (_, __, ___, ____) => false;
+            Assert.False(httpClientRequestMessage.ServerCertificateCustomValidationCallback(null, null, null, SslPolicyErrors.None));
+    
+            // aggregate an always true predicate with an always false predicate
+            httpClientRequestMessage.ServerCertificateCustomValidationCallback = (_, __, ___, ____) => true;
+            Assert.True(httpClientRequestMessage.ServerCertificateCustomValidationCallback(null, null, null, SslPolicyErrors.None));
+            httpClientRequestMessage.ServerCertificateCustomValidationCallback += (_, __, ___, ____) => false;
+            Assert.False(httpClientRequestMessage.ServerCertificateCustomValidationCallback(null, null, null, SslPolicyErrors.None));
```
habbes, line 42: Could you clarify the intent and rationale behind these two test cases?

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
```diff
@@ -18,5 +19,28 @@ public void UnwrapAggregateException()
...
+#if NETCOREAPP2_0_OR_GREATER
+        [Fact]
+        public void SetCertificateValidationCallback()
+        {
+            var httpClientRequestMessage = new HttpClientRequestMessage(
+                new DataServiceClientRequestMessageArgs("GET", new Uri("http://localhost"), false, false, new Dictionary<string, string>()));
+
+            // set a predicate that is always true
+            httpClientRequestMessage.ServerCertificateCustomValidationCallback = (_, __, ___, ____) => true;
+            Assert.True(httpClientRequestMessage.ServerCertificateCustomValidationCallback(null, null, null, SslPolicyErrors.None));
```
habbes, line 32: Is there a way to verify that these properties are actually set on the underlying HttpClientHandler?

habbes: Alternatively, to verify that the callback is called when a request is made for which that callback is applicable?

habbes:
@corranrogue9 @mikepizzo if customers switch to using the HttpClient transport mode, I fear they might experience a significant increase in latency per request. This seems to be due to the fact that our current implementation creates a new `HttpClient` and `HttpClientHandler` per request, which is inefficient. The screenshot below shows the difference in latencies when making 10 sequential requests on .NET 6 using OData Client with HttpWebRequest and HttpClient transport modes respectivley:

image

HttpWebRequest mode is slow for the first request (because of setting DataServiceContext and other details internal to the .NET's `HttpWebRequest` implementation) but then averages around 600-700ms for subsequent requests.

HttpClient mode mode is also slow for the first request but only averages around 1800-2000ms for subsequent requests. This is because for each request it has to setup a new HttpClient, probably a new connection, handshake, etc.

If the customer happens to be on .NET Core 3.1 however, they won't notice a latency regression. This is because .NET Core 3.1 implementation of `HttpWebRequest` is based on creating a new `HttpClient` and handler per request, which suffers the same performance issues. This was fixed in .NET 5.0+ by caching and re-using the `HttpClient` after the first request.

The following screenshot shows the results of running the same experiments with the plain `HttpWebRequest` (without OData). You can see that's it's much slower on .NET Core 3.1 compared to .NET 6.0 and .NET Framework. This shows what we stand to gain by re-using HttpClient:

image

PR #2431, opened 2022-06-14
Allow user to provide HttpClient instance in OData Client

habbes:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceClientRequestMessageArgs.cs
+++ b/src/Microsoft.OData.Client/DataServiceClientRequestMessageArgs.cs
@@ -28,6 +28,20 @@ namespace Microsoft.OData.Client
         /// <param name="usePostTunneling">True if the request message must use POST verb for the request and pass the actual verb in X-HTTP-Method header, otherwise false.</param>
         /// <param name="headers">The set of headers for the request.</param>
         public DataServiceClientRequestMessageArgs(string method, Uri requestUri, bool useDefaultCredentials, bool usePostTunneling, IDictionary<string, string> headers)
+            : this(method, requestUri, useDefaultCredentials, usePostTunneling, headers, null)
+        {
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DataServiceClientRequestMessageArgs"/> class.
+        /// </summary>
+        /// <param name="method">Method of the request.</param>
+        /// <param name="requestUri">The Request Uri.</param>
+        /// <param name="useDefaultCredentials">True if the default credentials need to be sent with the request. Otherwise false.</param>
+        /// <param name="usePostTunneling">True if the request message must use POST verb for the request and pass the actual verb in X-HTTP-Method header, otherwise false.</param>
+        /// <param name="headers">The set of headers for the request.</param>
+        /// <param name="httpClientFactory">Function that returns the <see cref="HttpClient"/> that should be used to send the request message.</param>
+        public DataServiceClientRequestMessageArgs(string method, Uri requestUri, bool useDefaultCredentials, bool usePostTunneling, IDictionary<string, string> headers, IHttpClientProvider httpClientFactory)
         {
             Debug.Assert(method != null, "method cannot be null");
             Debug.Assert(requestUri != null, "requestUri cannot be null");
@@ -38,6 +52,7 @@ namespace Microsoft.OData.Client
             this.RequestUri = requestUri;
             this.UsePostTunneling = usePostTunneling;
             this.UseDefaultCredentials = useDefaultCredentials;
+            this.HttpClientFactory = httpClientFactory;
 
             this.actualMethod = this.Method;
             if (this.UsePostTunneling && this.Headers.ContainsKey(XmlConstants.HttpXMethod))
@@ -81,5 +96,10 @@ namespace Microsoft.OData.Client
         /// Gets a System.Boolean value that controls whether default credentials are sent with requests.
         /// </summary>
         public bool UseDefaultCredentials { get; private set; }
+
+        /// <summary>
+        /// Returns the HttpClient that should be used to send the request message.
+        /// </summary>
+        public IHttpClientProvider HttpClientFactory { get; private set; }
     }
 }
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -22,6 +22,7 @@ namespace Microsoft.OData.Client
     using System.Linq;
     using System.Linq.Expressions;
     using System.Net;
+    using System.Net.Http;
     using System.Reflection;
     using System.Threading;
     using System.Threading.Tasks;
@@ -271,6 +272,31 @@ namespace Microsoft.OData.Client
         /// that will be combined with System.Windows.Browser.HtmlPage.Document.DocumentUri.
         /// </remarks>
         internal DataServiceContext(Uri serviceRoot, ODataProtocolVersion maxProtocolVersion, ClientEdmModel model)
+            : this(serviceRoot, maxProtocolVersion, model, httpClientProvider: null)
+        {
+        }
+
+        /// <summary>
+        /// Instantiates a new context with the specified <paramref name="serviceRoot"/> Uri.
+        /// The library expects the Uri to point to the root of a data service,
+        /// but does not issue a request to validate it does indeed identify the root of a service.
+        /// If the Uri does not identify the root of the service, the behavior of the client library is undefined.
+        /// </summary>
+        /// <param name="serviceRoot">
+        /// An absolute, well formed http or https URI without a query or fragment which identifies the root of a data service.
+        /// A Uri provided with a trailing slash is equivalent to one without such a trailing character
+        /// </param>
+        /// <param name="maxProtocolVersion">max protocol version that the client understands.</param>
+        /// <param name="model">The client edm model to use. Provided for testability.</param>
+        /// <param name="httpClientProvider">The provider used to get an instance of <see cref="HttpClient"/> to use when making a request
+        /// under the <see cref="HttpRequestTransportMode.HttpClient"/>. If no <see cref="HttpClientProvider"/> is specified, a new <see cref="HttpClient"/>
+        /// will created for each request.</param>
+        /// <exception cref="ArgumentOutOfRangeException">If the <paramref name="maxProtocolVersion"/> is not a valid value.</exception>
+        /// <remarks>
+        /// With Silverlight, the <paramref name="serviceRoot"/> can be a relative Uri
+        /// that will be combined with System.Windows.Browser.HtmlPage.Document.DocumentUri.
+        /// </remarks>
+        internal DataServiceContext(Uri serviceRoot, ODataProtocolVersion maxProtocolVersion, ClientEdmModel model, IHttpClientProvider httpClientProvider)
         {
             Debug.Assert(model != null, "model != null");
             this.model = model;
@@ -289,6 +315,7 @@ namespace Microsoft.OData.Client
             this.UsePostTunneling = false;
             this.keyComparisonGeneratesFilterQuery = false;
             this.deleteLinkUriOption = DeleteLinkUriOption.IdQueryParam;
+            this.HttpClientProvider = httpClientProvider;
         }
 
         #endregion
@@ -717,13 +744,21 @@ namespace Microsoft.OData.Client
             set { this.httpRequestTransportMode = value; }
         }
 
-        /// <summary>Gets or sets the HttpRequest mode to use in making Http Requests.</summary>
-        /// <returns>TransportModeFactory.</returns>
+        /// <summary>
+        /// Gets or sets the provider used to get the HttpClient to use for a particular request message
+        /// under the <see cref="HttpRequestTransportMode.HttpClient"/>.
+        /// If this is not set, a new <see cref="HttpClient"/> instance will be created for each request.
+        /// </summary>
+        public IHttpClientProvider HttpClientProvider { get; set; }
+
+        /// <summary>Gets or sets the <see cref="IDataServiceRequestMessageFactory"/> used to build request messages.</summary>
+        /// <returns>RequestMessageFactory</returns>
         internal IDataServiceRequestMessageFactory RequestMessageFactory
         {
             get { return this.requestMessageFactory; }
             set { this.requestMessageFactory = value; }
         }
+
 
         /// <summary>
         /// Gets or sets a System.Boolean value that controls whether default credentials are sent with requests.
--- a/src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
+++ b/src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
@@ -14,6 +14,7 @@ namespace Microsoft.OData.Client
     using System.Net;
     using System.Net.Http;
     using System.Net.Http.Headers;
+    using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.OData;
 
@@ -50,7 +51,9 @@ namespace Microsoft.OData.Client
         private readonly HttpRequestMessage _requestMessage;
         private readonly HttpClient _client;
         private readonly HttpClientHandler _handler;
+        private readonly IHttpClientProvider _clientProvider;
         private readonly MemoryStream _messageStream;
+        private TimeSpan _timeout;
 
         /// <summary>
         /// This will be used to cache content headers to be retrieved later. 
@@ -63,6 +66,9 @@ namespace Microsoft.OData.Client
 
         /// <summary>True if SendingRequest2Event is currently invoked, otherwise false.</summary>
         private bool inSendingRequest2Event;
+        private bool shouldDisposeClient = false;
+        private bool _disposed = false;
+        private CancellationTokenSource _cancellationSource;
 
         /// <summary>
         /// Constructor for HttpClientRequestMessage.
@@ -70,13 +76,19 @@ namespace Microsoft.OData.Client
         /// The args.ActualMethod is the actual method. In post tunneling situations method will be POST instead of the specified verb method.
         /// The args.method is the specified verb method
         /// </summary>
-        /// </summary>
         public HttpClientRequestMessage(DataServiceClientRequestMessageArgs args) 
             : base(args.ActualMethod)
         {
             _messageStream = new MemoryStream();
-            _handler = new HttpClientHandler();
-            _client = new HttpClient(_handler, disposeHandler: true);
+
+            _clientProvider = args.HttpClientFactory;
+            if (_clientProvider == null)
+            {
+                _handler = new HttpClientHandler();
+                _client = new HttpClient(_handler, disposeHandler: true);
+                shouldDisposeClient = true;
+            }
+            
             _contentHeaderValueCache = new Dictionary<string, string>();
             _effectiveHttpMethod = args.Method;
             _requestUrl = args.RequestUri;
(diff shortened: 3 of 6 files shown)


src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -253,6 +274,25 @@ public DataServiceContext(Uri serviceRoot, ODataProtocolVersion maxProtocolVersi
...
+        /// A Uri provided with a trailing slash is equivalent to one without such a trailing character.
+        /// With Silverlight, the <paramref name="serviceRoot"/> can be a relative Uri
+        /// that will be combined with System.Windows.Browser.HtmlPage.Document.DocumentUri.
+        /// </remarks>
+        public DataServiceContext(Uri serviceRoot, ODataProtocolVersion maxProtocolVersion, IHttpClientHandlerProvider httpClientHandlerProvider)
+            : this(serviceRoot, maxProtocolVersion, ClientEdmModelCache.GetModel(maxProtocolVersion), httpClientHandlerProvider: httpClientHandlerProvider)
+        {
+        }
+
         /// <summary>
```
mikepizzo, line 296: Given that the httpclienthandlerprovider is settable on the DataServiceContext, is there any reason we need to add constructor overloads for it?

src/Microsoft.OData.Client/IHttpClientHandlerProvider.cs
```diff
@@ -0,0 +1,29 @@
...
+{
+    /// <summary>
+    /// Provides the <see cref="HttpClientHandler"/> instance(s) used by OData Client
+    /// to execute requests.
+    /// </summary>
+    public interface IHttpClientHandlerProvider
+    {
+        /// <summary>
+        /// Returns an <see cref="HttpClientHandler"/> instance that OData Client should
+        /// used to eecute a request. This method is called per request.
```
mikepizzo, line 19: >used to eecute [](http://example.com/codeflow?start=12&length=14)

"used to eecute"=>"used to execute"

src/Microsoft.OData.Client/IHttpClientHandlerProvider.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// </summary>
+        /// <returns>The <see cref="HttpClientHandler"/> instanced that will be used to execute the request.</returns>
```
mikepizzo, line 21: >instanced [](http://example.com/codeflow?start=57&length=9)

"instanced"=>"instance"

src/Microsoft.OData.Client/DataServiceClientRequestMessageArgs.cs
```diff
@@ -81,5 +96,10 @@ public string ActualMethod
         /// Gets a System.Boolean value that controls whether default credentials are sent with requests.
         /// </summary>
         public bool UseDefaultCredentials { get; private set; }
+
+        /// <summary>
+        /// Returns the <see cref="HttpClientHandler"/> that should be used to send the request message.
```
mikepizzo, line 101: >Returns [](http://example.com/codeflow?start=12&length=7)

Maybe change "Returns" to "Gets" for consistency.

mikepizzo:
A few minor comments, but otherwise looks good to me. Main question is whether we need to add constructor overloads for DataServiceContext, or can just call the setter for the IHttpClientHandlerProvider.

mikepizzo (changes requested):
:clock1:

src/Microsoft.OData.Client/DataServiceContext.cs
habbes: Not it's not necessary. Just that I started with the constructor argument then added the setter after the fact.

Since the `HttpRequestTransportMode` is set via property and not constructor, it does make sense to keep this on the property only.

src/Microsoft.OData.Client/DataServiceClientRequestMessageArgs.cs
```diff
@@ -28,6 +28,20 @@ public class DataServiceClientRequestMessageArgs
...
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="DataServiceClientRequestMessageArgs"/> class.
+        /// </summary>
+        /// <param name="method">Method of the request.</param>
+        /// <param name="requestUri">The Request Uri.</param>
+        /// <param name="useDefaultCredentials">True if the default credentials need to be sent with the request. Otherwise false.</param>
+        /// <param name="usePostTunneling">True if the request message must use POST verb for the request and pass the actual verb in X-HTTP-Method header, otherwise false.</param>
+        /// <param name="headers">The set of headers for the request.</param>
+        /// <param name="httpClientHandlerProvider">Returns the <see cref="HttpClientHandler"/> that should be used to send the request message.</param>
```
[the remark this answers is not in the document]
habbes: I meant to say: The httpClientHandlerProvider returns the httpClientHandler...

Let me rephrase it.

mikepizzo (approved):
:shipit:

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -50,6 +52,7 @@ public class HttpClientRequestMessage : DataServiceClientRequestMessage, ISendin
         private readonly HttpRequestMessage _requestMessage;
         private readonly HttpClient _client;
         private readonly HttpClientHandler _handler;
+        private readonly IHttpClientHandlerProvider _clientHandlerProvider;
```
corranrogue9, line 55:
```suggestion
```

I don't think we need a member for this, it seems we only use it in the constructor

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -64,19 +67,31 @@ public class HttpClientRequestMessage : DataServiceClientRequestMessage, ISendin
...
+            if (_clientHandlerProvider == null)
+            {
+                _handler = new HttpClientHandler();
+                _client = new HttpClient(_handler, disposeHandler: true);
+            }
+            else
+            {
+                _handler = _clientHandlerProvider.GetHttpClientHandler();
+                _client = new HttpClient(_handler, disposeHandler: false);
+            }
```
corranrogue9, line 93:
```suggestion
            var clientHandlerProvider = args.HttpClientHandlerProvider;
            if (clientHandlerProvider == null)
            {
                this._handler = new HttpClientHandler();
                this._client = new HttpClient(this._handler, disposeHandler: true);
            }
            else
            {
                try
                {
                    this._handler = clientHandlerProvider.GetHttpClientHandler();
                    this._client = new HttpClient(this._handler, disposeHandler: false);
                }
                catch
                {
                    this._messageStream.Dispose();
                    throw;
                }
            }
```

The `HttpClient` constructor will throw in case of a `null` handler, plus we likely don't own the provider code, so it could throw anything really.

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -514,11 +529,23 @@ public void Dispose()
...
+
+            _disposed = true;
+
+            if (disposing)
+            {
+                HttpResponseMessage response = _httpResponseMessage;
+                _httpResponseMessage = null;
+                if (response != null)
+                {
+                    ((IDisposable)response).Dispose();
```
corranrogue9, line 545: I have a feeling the cast here gets JITed away anyway, but I'm curious where this comes from. It's in the original code too. I've only seem something like this with an explicit interface implementation or something. Do you know why we are doing this cast?

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -514,11 +529,23 @@ public void Dispose()
         /// <param name="disposing">If 'true' this method is called from user code; if 'false' it is called by the runtime.</param>
         protected virtual void Dispose(bool disposing)
         {
-            HttpResponseMessage response = _httpResponseMessage;
-            _httpResponseMessage = null;
-            if (response != null)
+            if (_disposed)
```
corranrogue9, line 532:
```suggestion
            if (this._disposed)
```

src/Microsoft.OData.Client/IHttpClientHandlerProvider.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+        /// <summary>
+        /// Returns an <see cref="HttpClientHandler"/> instance that OData Client should
+        /// used to execute a request. This method is called per request.
+        /// </summary>
+        /// <returns>The <see cref="HttpClientHandler"/> instance that will be used to execute the request.</returns>
+        /// <remarks>
+        /// It's possible that the returned <see cref="HttpClientHandler"/> may be mutated inside a request hook
+        /// e.g. via <see cref="HttpClientRequestMessage.Credentials"/>.
+        /// OData Client does not dispose the <see cref="HttpClientHandler"/> instance.
```
corranrogue9, line 25: The lack of disposing is the entire point of the providing, right? It seems like this should be noted in the interface summary and the method summary rather than a remark, since it's critical to both the performance improvement and the memory management that the implementer understands this.

src/Microsoft.OData.Client/IHttpClientHandlerProvider.cs
```diff
@@ -0,0 +1,29 @@
...
+        /// </remarks>
+        HttpClientHandler GetHttpClientHandler();
```
corranrogue9, line 27: Did we want to provide the context or args for this? It seems like it will be difficult for the implementer to know when to re-use a `HttpClientHandler` instance or when to provide a new one. (this is truly a discussion question, I think I don't know enough about `HttpClientHandler` and `HttpClient` to know the answer).

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
```diff
@@ -18,5 +28,185 @@ public void UnwrapAggregateException()
...
+                    usePostTunneling: false,
+                    new Dictionary<string, string>(),
+                    httpClientHandlerProvider);
+
+                using (var request = new HttpClientRequestMessage(args))
+                {
+                    // Act
+                    IODataResponseMessage response = request.GetResponse();
+                    var stream = response.GetStream();
+                    string contents = "";
```
corranrogue9, line 55:
```suggestion
                    string contents = "";
                    using (var stream = response.GetStream())
```

Better to be explicit about this stuff I think...

corranrogue9, line 55:
```suggestion
                    string contents = "";
                    using (IODataResponseMessage response = request.GetResponse())
                    using (var stream = response.GetStream())
```

Better to be explicit about this stuff I think...

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
```diff
@@ -18,5 +28,185 @@ public void UnwrapAggregateException()
...
+                var newResponse = await client.GetAsync("http://foo");
+
+                // Assert
+                // We can still make request from the client after HttpClientRequestMessage is disposed
+                var contents = await newResponse.Content.ReadAsStringAsync();
+                Assert.Equal(expectedResponse, contents);
+                Assert.Equal(2, handler.Requests.Count);
+                Assert.Equal("GET http://localhost/", handler.Requests[0]);
+                Assert.Equal("GET http://foo/", handler.Requests[1]);
+                Assert.False(handler.Disposed);
```
corranrogue9, line 107:
```suggestion
                using (var request = new HttpClientRequestMessage(args))
                {
                    // Act
                    IODataResponseMessage response = request.GetResponse();
                    var stream = response.GetStream();
                    stream.Dispose();
                    request.Dispose();

                    // Re-use handler with a different HttpClient after initial request is disposed
                    using (var client = new HttpClient(handler, disposeHandler: false))
                    using (var newResponse = await client.GetAsync("http://foo"))
                    {
                        // Assert
                        // We can still make request from the client after HttpClientRequestMessage is disposed
                        var contents = await newResponse.Content.ReadAsStringAsync();
                        Assert.Equal(expectedResponse, contents);
                        Assert.Equal(2, handler.Requests.Count);
                        Assert.Equal("GET http://localhost/", handler.Requests[0]);
                        Assert.Equal("GET http://foo/", handler.Requests[1]);
                        Assert.False(handler.Disposed);
                    }
                }
```

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockHttpClientHandler.cs
```diff
@@ -0,0 +1,88 @@
...
+        /// </summary>
+        /// <param name="expectedResponse">The contents to return in each response.</param>
+        public MockHttpClientHandler(string expectedResponse)
+        {
+            _requestHandler = (_) =>
+            {
+                var response = new HttpResponseMessage(HttpStatusCode.OK);
+                response.Content = new StringContent(expectedResponse);
+                return response;
+            };
```
corranrogue9, line 39:
```suggestion
        : this((_) =>
            {
                var response = new HttpResponseMessage(HttpStatusCode.OK);
                response.Content = new StringContent(expectedResponse);
                return response;
            })
        {
```

I'm not sure if this actually compiles, or that it's even readable, but I think there's something to be said about the explicit nature of having a primary constructor.

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockHttpClientHandlerProvider.cs
```diff
@@ -0,0 +1,40 @@
...
+    /// </summary>
+    internal sealed class MockHttpClientHandlerProvider: IHttpClientHandlerProvider
+    {
+        private readonly HttpClientHandler _handler;
+
+        /// <summary>
+        /// Creates a new instance of <see cref="MockHttpClientHandlerProvider"/>.
+        /// </summary>
+        /// <param name="client">The <see cref="HttpClientHandler"/> to return.</param>
+        public MockHttpClientHandlerProvider(HttpClientHandler handler, bool requestClientDisposal = false)
```
corranrogue9, line 23:
```suggestion
        public MockHttpClientHandlerProvider(HttpClientHandler handler)
```

I think that was probably just left over from something else? I'm not seeing it used anywhere

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockHttpClientHandlerProvider.cs
```diff
@@ -0,0 +1,40 @@
...
+
+        /// <summary>
+        /// Number of times <see cref="GetHttpClientHandler()"/> was
+        /// called.
+        /// </summary>
+        public int NumCalls { get; private set; } = 0;
+
+        public HttpClientHandler GetHttpClientHandler()
+        {
+            NumCalls++;
```
corranrogue9, line 36: I suppose we aren't concerned about thread-safety in the tests where this is used, but consider `Interlocked.Increment`

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockUnresponsiveHttpClientHandler.cs
```diff
@@ -0,0 +1,71 @@
...
+        /// Callback that is triggered when a request has started.
+        /// This can be used to coordinate with a task on a separate
+        /// thread that should run only after a request has started.
+        /// For example, we may want to test aborting a request after
+        /// the request has started.
+        /// </summary>
+        public Action OnRequestStarted { get; set; }
+
+        public HttpResponseMessage Send(HttpRequestMessage request, CancellationToken cancellationToken)
+        {
```
corranrogue9, line 35: I think this is a case where the non-async method should call the async one. I think this will do what you want:

```
protected override Task<HttpResponseMessage> SendAsync(HttpRequestMessage request, CancellationToken cancellationToken)
{
    await Task.Delay(TimeSpan.FromMilliseconds(MaxWaitTimeMillis), cancellationToken).ConfigureAwait(false);
    throw new Exception(
        $"{nameof(MockUnresponsiveHttpClientHandler)} has exceeded maximum {MaxWaitTimeMillis}ms wait time in request."
        + "Review your logic to ensure the request is cancelled in time.");
}

public HttpResponseMessage Send(HttpRequestMessage request, CancellationToken cancellationToken)
{
    return SendAsync(request, cancellationToken).GetAwaiter().GetResult();
}
```

src/Microsoft.OData.Client/IHttpClientHandlerProvider.cs
```diff
@@ -0,0 +1,29 @@
...
+// </copyright>
+//---------------------------------------------------------------------
+
+using System.Net.Http;
+
+namespace Microsoft.OData.Client
+{
+    /// <summary>
+    /// Provides the <see cref="HttpClientHandler"/> instance(s) used by OData Client
+    /// to execute requests.
```
ElizabethOkerio, line 13: is the `HttpcClienthHandler` used to execute requests? can the documentation be improved to mirror what the `HttpClienthHandler` does?

habbes: Technically, the `HttpClientHandler` does execute the requests. The `HttpClient`' `SendAsync` method, calls the `HttpClientHandler`'s `SendAsync` method.

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockHttpClientHandlerProvider.cs
habbes: Yeah, this is left over from a previous implementation. Nice catch.

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
habbes: Here I wanted to make sure the request is disposed *before* creating the new client. If I create a `using` block, maybe it makes sense to close before the line that creates a new client and to do away with the explicit `request.Dispose()` and `stream.Dispose()` calls.

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
habbes: `IODataResponseMessage` is not `IDisposable` though...

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockHttpClientHandler.cs
habbes: It does compile.

src/Microsoft.OData.Client/IHttpClientHandlerProvider.cs
habbes: I had passed the http request message at the beginning, and there was consideration to pass context with more info. But I removed it after Mike's suggestion to keep things simple for 7.x

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
```diff
@@ -18,5 +28,185 @@ public void UnwrapAggregateException()
...
+                    "GET",
+                    new Uri("http://localhost"),
+                    useDefaultCredentials: false,
+                    usePostTunneling: false,
+                    new Dictionary<string, string>(),
+                    httpClientProvider);
+
+                using (var request = new HttpClientRequestMessage(args))
+                {
+                    request.Timeout = 1;
```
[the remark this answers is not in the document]
habbes: Could you elaborate on what race condition would occur and how a sleep would prevent that?

ElizabethOkerio (approved):

PR #2436, opened 2022-06-17
delta serialization response fix

ElizabethOkerio:
### Issues

### Description

A response for a deep update of a `nested delta resource set`  is truncated if the `OData-Version` is not specified as `v4.01` in the request. This PR removes the check for the `OData-Version`.  This means that the `OData-Version` need not be passed in the request and the response should  be serialized properly. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataWriterCore.cs
+++ b/src/Microsoft.OData.Core/ODataWriterCore.cs
@@ -2978,7 +2978,7 @@ namespace Microsoft.OData
 
                     break;
                 case WriterState.NestedResourceInfoWithContent:
-                    if (newState != WriterState.ResourceSet && newState != WriterState.Resource && newState != WriterState.Primitive && (this.Version < ODataVersion.V401 || (newState != WriterState.DeltaResourceSet && newState != WriterState.DeletedResource)))
+                    if (newState != WriterState.ResourceSet && newState != WriterState.Resource && newState != WriterState.Primitive && newState != WriterState.DeltaResourceSet && newState != WriterState.DeletedResource)
                     {
                         throw new ODataException(Strings.ODataWriterCore_InvalidTransitionFromExpandedLink(this.State.ToString(), newState.ToString()));
                     }
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
@@ -2097,7 +2097,7 @@ namespace Microsoft.OData.Tests.JsonLight
         [InlineData(/*isResponse*/true)]
         [InlineData(/*isResponse*/false)]
         [Theory]
-        public void WriteNestedDeltaResourceSetIn40ShouldFail(bool isResponse)
+        public void WriteNestedDeltaResourceSetIn40ShouldNotFail(bool isResponse)
         {
             this.TestInit(this.GetModel());
 
@@ -2114,7 +2114,7 @@ namespace Microsoft.OData.Tests.JsonLight
                 writer.WriteStart(new ODataDeltaResourceSet());
             };
 
-            writeAction.Throws<ODataException>(Strings.ODataWriterCore_InvalidTransitionFromExpandedLink("NestedResourceInfoWithContent", "DeltaResourceSet"));
+            writeAction.DoesNotThrow();
         }
 
         [InlineData(/*isResponse*/true)]
@@ -2137,7 +2137,7 @@ namespace Microsoft.OData.Tests.JsonLight
                 writer.WriteDeltaDeletedEntry(new ODataDeltaDeletedEntry("Products/1", DeltaDeletedEntryReason.Deleted));
             };
 
-            writeAction.Throws<ODataException>(Strings.ODataWriterCore_InvalidTransitionFromExpandedLink("NestedResourceInfoWithContent", "DeletedResource"));
+            writeAction.Throws<ODataException>(Strings.ODataWriterCore_NestedContentNotAllowedIn40DeletedEntry);
         }
 
         [InlineData(/*isResponse*/true)]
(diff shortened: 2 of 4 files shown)


src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -2942,7 +2942,7 @@ private void ValidateTransition(WriterState newState)
                             throw new ODataException(Strings.ODataWriterCore_InvalidTransitionFromResourceSet(this.State.ToString(), newState.ToString()));
                         }
 
-                        if (this.State == WriterState.DeletedResource && this.Version < ODataVersion.V401 && newState == WriterState.NestedResourceInfo)
+                        if (this.State == WriterState.DeletedResource && newState == WriterState.NestedResourceInfo)
```
KenitoInc, line 2945: Won't this change be a breaking change?
Will this fix also affect requests, not only responses?

ElizabethOkerio: it only affects responses not requests

xuzhg (dismissed):

corranrogue9 (changes requested):
Please add tests for this change

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
```diff
@@ -2119,7 +2119,7 @@ public void WriteNestedDeltaResourceSetIn40ShouldFail(bool isResponse)
 
         [InlineData(/*isResponse*/true)]
         [InlineData(/*isResponse*/false)]
-        [Theory]
+        [Theory(Skip ="Should not throw exception")]
```
KenitoInc, line 2122: Looking at the number of tests you are allowing to pass in this PR confirms my fears that we should not be doing this fix.
From the spec, `Deep insert` was introduced in `v4` and Deep update in `v4.01`. See https://docs.oasis-open.org/odata/new-in-odata/v4.01/cn04/new-in-odata-v4.01-cn04.html#sec_NewDeepUpdate

Therefore we should leave the `OData-Version` validation as it is and ensure properly document that `Deep update` only works in `v4.01`.

What do you think? @mikepizzo @xuzhg

KenitoInc (changes requested):

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
gathogojr, line 2122: If test is no longer relevant (or factual), the right thing to do should be to replace it with a relevant test(s) - e.g. a test to verify we can write a nested deleted resource when passed version is 4.0. **I don't think nested deleted resource is supported though in the 4.0 spec though**

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightOutputContextApiTests.cs
```diff
@@ -3064,7 +3064,7 @@ public async Task WriteNestedDeltaResourceSet_APIsThrowException()
             Assert.Equal(expectedMessage, syncException.Message);
         }
 
-        [Theory]
+        [Theory(Skip = "Supported and should not throw exception")]
```
gathogojr, line 3067: See my comment here https://github.com/OData/odata.net/pull/2436/files#r908402778

gathogojr:
@ElizabethOkerio I have an issue with the proposed change because it'll cause a nested deleted resource to be written even for V4 requests... The right fix could be to retrieve the OData version from the request header and use that to determine whether to write a nested deleted resource or not...

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
ElizabethOkerio: I was trying out to skip them to see whether there are other tests that are failing.

src/Microsoft.OData.Core/ODataWriterCore.cs
mikepizzo: I believe this turns an error state into a success state for ODataVersion.V401, which is generally considered safe.

mikepizzo:
+1

Change looks fine to me, please add tests.

---
In reply to: [1014238840](https://github.com/OData/odata.net/pull/2436#issuecomment-1014238840) [](http://example.com/codeflow?ancestors=1014238840,1014238840)

mikepizzo:
+1

---
In reply to: [1014238840](https://github.com/OData/odata.net/pull/2436#issuecomment-1014238840) [](http://example.com/codeflow?ancestors=1014238840,1014238840)

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
```diff
@@ -2137,7 +2137,7 @@ public void WriteNestedDeletedResourceIn40ShouldFail(bool isResponse)
                 writer.WriteDeltaDeletedEntry(new ODataDeltaDeletedEntry("Products/1", DeltaDeletedEntryReason.Deleted));
             };
 
-            writeAction.Throws<ODataException>(Strings.ODataWriterCore_InvalidTransitionFromExpandedLink("NestedResourceInfoWithContent", "DeletedResource"));
+            writeAction.Throws<ODataException>(Strings.ODataWriterCore_NestedContentNotAllowedIn40DeletedEntry);
```
KenitoInc, line 2140: Why are we still throwing this exception?
Is the behaviour similar to v4.01?

ElizabethOkerio: this will be in a different pr to fix delta deleted resources in v4.0. This PR just solves nested delta resource sets.

PR #2437, opened 2022-06-20
Updated ODataUriParser to allow select and expand property names that…

### LINKED ISSUE #2410: Property named "true" in $select throws exception
Creating a field called "true" then using it in a `$select=true` throws an error.

### Assemblies affected

OData .Net lib 7.10

### Reproduce steps

Create a model with an entity property called "true" then try to use it in a $select

### Expected result

The $select works and does not throw an error.

### Actual result
OData parses the "true" as a boolean and fails because it is not an identifier.
```
ODataException : An identifier was expected at position ..
Stack:
   at Microsoft.OData.UriParser.ExpressionToken.GetIdentifier()
   at Microsoft.OData.UriParser.SelectExpandTermParser.ParseSegment(PathSegmentToken previousSegment, Boolean allowRef)
   at Microsoft.OData.UriParser.SelectExpandTermParser.ParseTerm(Boolean allowRef)
   at Microsoft.OData.UriParser.SelectExpandParser.ParseSingleSelectTerm()
   at Microsoft.OData.UriParser.SelectExpandParser.ParseCommaSeparatedSelectList(Func`2 ctor, Func`1 termParsingFunc)
   at Microsoft.OData.UriParser.SelectExpandSyntacticParser.Parse(String selectClause, String expandClause, IEdmStructuredType parentStructuredType, ODataUriParserConfiguration configuration, ExpandToken& expandTree, SelectToken& selectTree)
   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseSelectAndExpandImplementation(String select, String expand, ODataUriParserConfiguration configuration, ODataPathInfo odataPathInfo)
   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseSelectAndExpand()
```

corranrogue9:
… conflict with reserved keywords

### Issues

*This pull request fixes #2410.*

### Description

The `SelectExpandParser` used by the `ODataUriParser` takes a two-pass approach to parsing the select and expand options. The first pass is intended to be syntactic only, while the second pass applies semantics. For code re-use, The syntactic pass leverages the `ExpressionLexer`. However, `ExpressionLexer` applies some semantics while parsing when a perceived "identifier" token is reached. I have added a flag in `ExpressionLexer` to optionally disable these semantics, and updated the `SelectExpandParser` to disable this behavior for its lexer. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
+++ b/src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
@@ -83,6 +83,11 @@ namespace Microsoft.OData.UriParser
         /// <summary>Whether the lexer is being used to parse function parameters. If true, will allow/recognize parameter aliases and typed nulls.</summary>
         private readonly bool parsingFunctionParameters;
 
+        /// <summary>
+        /// Whether or not to interpret the semantics of a token that appears to be an identifier (i.e. consider if it is a keyword instead of an identifier)
+        /// </summary>
+        private readonly bool applySemanticsToIdentifiers;
+
         /// <summary>Lexer ignores whitespace</summary>
         private bool ignoreWhitespace;
 
@@ -108,6 +113,19 @@ namespace Microsoft.OData.UriParser
         /// <param name="useSemicolonDelimiter">If true, the lexer will tokenize based on semicolons as well.</param>
         /// <param name="parsingFunctionParameters">Whether the lexer is being used to parse function parameters. If true, will allow/recognize parameter aliases and typed nulls.</param>
         internal ExpressionLexer(string expression, bool moveToFirstToken, bool useSemicolonDelimiter, bool parsingFunctionParameters)
+            : this(expression, moveToFirstToken, useSemicolonDelimiter, parsingFunctionParameters, true)
+        {
+        }
+
+        /// <summary>Initializes a new <see cref="ExpressionLexer"/>.</summary>
+        /// <param name="expression">Expression to parse.</param>
+        /// <param name="moveToFirstToken">If true, this constructor will call NextToken() to move to the first token.</param>
+        /// <param name="useSemicolonDelimiter">If true, the lexer will tokenize based on semicolons as well.</param>
+        /// <param name="parsingFunctionParameters">Whether the lexer is being used to parse function parameters. If true, will allow/recognize parameter aliases and typed nulls.</param>
+        /// <param name="applySemanticsToIdentifiers">
+        /// Whether or not to interpret the semantics of a token that appears to be an identifier (i.e. consider if it is a keyword instead of an identifier)
+        /// </param>
+        internal ExpressionLexer(string expression, bool moveToFirstToken, bool useSemicolonDelimiter, bool parsingFunctionParameters, bool applySemanticsToIdentifiers)
         {
             Debug.Assert(expression != null, "expression != null");
 
@@ -116,6 +134,7 @@ namespace Microsoft.OData.UriParser
             this.TextLen = this.Text.Length;
             this.useSemicolonDelimiter = useSemicolonDelimiter;
             this.parsingFunctionParameters = parsingFunctionParameters;
+            this.applySemanticsToIdentifiers = applySemanticsToIdentifiers;
             this.parsingDoubleQuotedString = false;
 
             this.SetTextPos(0);
@@ -810,7 +829,7 @@ namespace Microsoft.OData.UriParser
 
                 this.HandleQuotedValues();
             }
-            else
+            else if (this.applySemanticsToIdentifiers)
             {
                 // Handle keywords.
                 // Get built in type literal prefix.
--- a/src/Microsoft.OData.Core/UriParser/Parsers/SelectExpandParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/SelectExpandParser.cs
@@ -81,7 +81,7 @@ namespace Microsoft.OData.UriParser
 
             // Sets up our lexer. We don't turn useSemicolonDelimiter on since the parsing code for expand options,
             // which is the only thing that needs it, is in a different class that uses it's own lexer.
-            this.lexer = clauseToParse != null ? new ExpressionLexer(clauseToParse, false /*moveToFirstToken*/, false /*useSemicolonDelimiter*/) : null;
+            this.lexer = clauseToParse != null ? new ExpressionLexer(clauseToParse, false, false, false, false) : null;
 
             this.enableCaseInsensitiveBuiltinIdentifier = enableCaseInsensitiveBuiltinIdentifier;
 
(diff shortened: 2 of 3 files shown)


src/Microsoft.OData.Core/UriParser/Parsers/SelectExpandParser.cs
```diff
@@ -81,7 +81,7 @@ internal sealed class SelectExpandParser
 
             // Sets up our lexer. We don't turn useSemicolonDelimiter on since the parsing code for expand options,
             // which is the only thing that needs it, is in a different class that uses it's own lexer.
-            this.lexer = clauseToParse != null ? new ExpressionLexer(clauseToParse, false /*moveToFirstToken*/, false /*useSemicolonDelimiter*/) : null;
+            this.lexer = clauseToParse != null ? new ExpressionLexer(clauseToParse, false, false, false, false) : null;
```
xuzhg, line 84: >false [](http://example.com/codeflow?start=105&length=5)

so, only in select and expand, it's false, in others it's true?

We don't allow: $filter=true eq true, if one of the 'true' is for the property?

corranrogue9: There's a lot going on here. For `$filter=true eq true`, that's the same as just saying `$filter=true` or not providing the filter at all. We interpret the "true" literal as a boolean expression, which as far as I can tell is in accordance with the standard. If there is a property named "true", the above is still accurate, but the client can disambiguate that they are referring to the property by using `$it`, as in `$filter=$it/true eq true`. The reality is that we also have a bug in that case as well. I talked with Mike yesterday and we decided that we would address the `$filter` issue separately, since it implies a larger change (specifically, `ExpressionLexer` applies semantics, but many query options require on a syntax-only pass first; this syntax-only pass using `ExpressionLexer` has issues across all of the query option implementations).

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
```diff
@@ -26,6 +26,410 @@ public class ODataUriParserTests
...
+            foo.AddStructuralProperty("true", EdmCoreModel.Instance.GetString(true));
+            model.AddElement(foo);
+            var container = new EdmEntityContainer("foobar", "FooService");
+            container.AddEntitySet("foos", foo);
+            model.AddElement(container);
+
+            var parser = new ODataUriParser(model, new Uri("/foos?$select=null", UriKind.Relative));
+            var odataException = Assert.Throws<ODataException>(() => parser.ParseUri());
+
+            Assert.Equal("Could not find a property named 'null' on type 'foobar.foo'.", odataException.Message);
```
mikepizzo, line 182: Should have another test where the type is defined as open -- in which case, no error should be returned.

src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
```diff
@@ -108,6 +113,19 @@ internal ExpressionLexer(string expression, bool moveToFirstToken, bool useSemic
         /// <param name="useSemicolonDelimiter">If true, the lexer will tokenize based on semicolons as well.</param>
         /// <param name="parsingFunctionParameters">Whether the lexer is being used to parse function parameters. If true, will allow/recognize parameter aliases and typed nulls.</param>
         internal ExpressionLexer(string expression, bool moveToFirstToken, bool useSemicolonDelimiter, bool parsingFunctionParameters)
```
mikepizzo, line 115: This is internal -- how many places do we call this existing method from? Would it be easier to add the parameter to the existing overload and explicitly pass "true", rather than define a new overload?

src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
```diff
@@ -108,6 +113,19 @@ internal ExpressionLexer(string expression, bool moveToFirstToken, bool useSemic
         /// <param name="useSemicolonDelimiter">If true, the lexer will tokenize based on semicolons as well.</param>
         /// <param name="parsingFunctionParameters">Whether the lexer is being used to parse function parameters. If true, will allow/recognize parameter aliases and typed nulls.</param>
         internal ExpressionLexer(string expression, bool moveToFirstToken, bool useSemicolonDelimiter, bool parsingFunctionParameters)
+            : this(expression, moveToFirstToken, useSemicolonDelimiter, parsingFunctionParameters, true)
```
habbes, line 116: For better legibility:
```suggestion
            : this(expression, moveToFirstToken, useSemicolonDelimiter, parsingFunctionParameters, applySemanticsToIdentifiers: true)

```

src/Microsoft.OData.Core/UriParser/Parsers/SelectExpandParser.cs
habbes, line 84: Hard to tell what all these `false` params are for, suggestion to use named parameters in such cases:
```suggestion
            this.lexer = clauseToParse != null ? new ExpressionLexer(clauseToParse, moveToFirstToken: false, useSemicolonDelimiter: false, parsingFunctionParameters: false, applySemanticsToIdentifiers: false) : null;

```

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/ODataUriParserTests.cs
```diff
@@ -26,6 +26,410 @@ public class ODataUriParserTests
...
+            var selectedItems = odataUri.SelectAndExpand.SelectedItems.ToArray();
+            Assert.Single(selectedItems);
+            var expandedNavigationSelectItem = selectedItems[0] as ExpandedNavigationSelectItem;
+            Assert.Equal("null", expandedNavigationSelectItem.NavigationSource.Name);
+            Assert.Equal("foobar.bar", expandedNavigationSelectItem.NavigationSource.EntityType().FullTypeName());
+            Assert.Empty(expandedNavigationSelectItem.SelectAndExpand.SelectedItems);
+        }
+
+        /// <summary>
+        /// Expands a property named "null" when such a property doesn't exist on the entity
```
habbes, line 361: I think comment should mention that this results in an exception.

habbes (commented):
Looks good to me. Left minor suggestions to improve legibility.

src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs
```diff
@@ -83,6 +83,11 @@ internal class ExpressionLexer
         /// <summary>Whether the lexer is being used to parse function parameters. If true, will allow/recognize parameter aliases and typed nulls.</summary>
         private readonly bool parsingFunctionParameters;
 
+        /// <summary>
+        /// Whether or not to interpret the semantics of a token that appears to be an identifier (i.e. consider if it is a keyword instead of an identifier)
+        /// </summary>
+        private readonly bool applySemanticsToIdentifiers;
```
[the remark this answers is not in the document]
xuzhg: how can customer config this?

xuzhg (approved):

xuzhg:
Others look good to me

PR #2460, opened 2022-07-19
Implement async API for ODataUtf8JsonWriter

habbes:
### Issues

*This pull request fixes #2421 

### Description

Add `IJsonWriterAsync` implementation to `ODataUtf8JsonWriter`.

### Checklist (Uncheck if it is not completed)

- [x] *Test cases added*
- [x] *Build and test with one-click build and test script passed*

### Additional work necessary

*If documentation update is needed, please add "Docs Needed" label to the issue and provide details about the required document change in the issue.*

PR #2416, opened 2022-05-24
Create ODataUtf8JsonWriter

habbes:
### Issues

*This pull request fixes #2415 *

### Description

This PR introduces a new implementation of `IJsonWriter` that is based on `Utf8JsonWriter`.

This is an opt-in feature with no breaking changes. Customers who do not opt-in to use this writer should not be impacted. To configure the `ODataMessageWriter` to use the new JSON writer, the user would have to inject the appropriate JSON writer factory in dependency container. Since the existing `IJsonWriterFactory` expects a `TextWriter` input, I resorted to creating a new interface `IStreamBasedJsonWriterFactory` that accepts a `Stream` as input and returns an `IJsonWriter`. I also created a default implementation `DefaultStreamBasedJsonWriterFactory` that returns an instance of the new `ODataUtf8JsonWriter`.

Summary of the new types:

- `ODataUtf8JsonWriter`:  the new JSON writer implementation (*internal*)
- `IStreamBasedJsonWriterFactory`: interface for factory that creates an `IJsonWriter` based on a `Stream` rather than a `TextWriter`
- `DefaultStreamBasedJsonWriterFactory`: default implementation of the aforementioned interface

To use this new writer, here's how the user would configure `ODataMessageWriter`:

```c#
IContainerBuilder containerBuilder = // createContainerBuilder...
containerBuilder.AddDefaultODataSerices();
containerBuilder.AddService<IStreamBasedJsonWriterFactory>(
  ServiceLifetime.Singleton, sp => new DefaultStreamBasedJsonWriterFactory());

IODataRequestMessage request = ...;
request.Container = containerBuilder.BuildContainer();

var writer = new ODataMessageWriter(request, settings, model);
```

#### Remarks

This is the first PR in a series of PRs that will complete the implementation of `ODataUtf8JsonWriter`. This PR focuses on implementing the core writer methods and ensures that they behave the same as the existing `JsonWriter`. There are a number of gaps, limitations and optimizations that will be addressed in follow-up PRs, for example:
- This feature only support .NET Core 3.1 and later because `Utf8JsonWriter` is only available in .NET Core. It's not available in .NET Standard.
[...]

PR #2461, opened 2022-07-19
Add CreateAsynchronousJsonWriter to IStreamBasedJsonWriterFactoryAsync

habbes:
### Issues

Related to #2421 

### Description

This PR adds the `CreateAsynchronousJsonWriter` method to the `IStreamBasedJsonWriter` interface that was introduced as part of the `ODataUtf8JsonWriter` work (https://github.com/OData/odata.net/pull/2416).

This method will allow the same interface to also be used to create `IJsonWriterAsync`. This is a change that is being back-ported from the currently pending PR that's implementing the async API for `ODataUtf8JsonWriter` (https://github.com/OData/odata.net/pull/2460).

The goal is to merge this PR before the next release so that when we release the async API for `ODataUtf8JsonWriter`, it will not be a breaking change.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/DefaultStreamBasedJsonWriterFactory.cs
+++ b/src/Microsoft.OData.Core/Json/DefaultStreamBasedJsonWriterFactory.cs
@@ -50,6 +50,11 @@ namespace Microsoft.OData.Json
 
             return new ODataUtf8JsonWriter(stream, isIeee754Compatible, encoding, encoder: this.encoder);
         }
+
+        public IJsonWriterAsync CreateAsynchronousJsonWriter(Stream stream, bool isIeee754Compatible, Encoding encoding)
+        {
+            throw new NotImplementedException();
+        }
     }
 }
 #endif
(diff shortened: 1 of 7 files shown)


src/Microsoft.OData.Core/Json/DefaultStreamBasedJsonWriterFactory.cs
```diff
@@ -50,6 +50,11 @@ public IJsonWriter CreateJsonWriter(Stream stream, bool isIeee754Compatible, Enc
 
             return new ODataUtf8JsonWriter(stream, isIeee754Compatible, encoding, encoder: this.encoder);
         }
+
+        public IJsonWriterAsync CreateAsynchronousJsonWriter(Stream stream, bool isIeee754Compatible, Encoding encoding)
+        {
+            throw new NotImplementedException();
```
KenitoInc, line 56: Why are we throwing `NotImplementedException`?

gathogojr: @KenitoInc The explanation is provided in the description

KenitoInc: Aah, got it! I just looked at the pending PR.

KenitoInc (approved):
LGTM

gathogojr (approved):

PR #2462, opened 2022-07-20
Fix DisposeAsync implementation for .NET Core 3.1 and higher

gathogojr:
### Description

Fixes `DisposeAsync` implementation by overriding base .NET Core 3.1 and higher

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataNotificationStream.cs
+++ b/src/Microsoft.OData.Core/ODataNotificationStream.cs
@@ -15,7 +15,7 @@ namespace Microsoft.OData
     /// <summary>
     /// Wrapper to listen for dispose on a <see cref="Stream"/>.
     /// </summary>
-#if NETSTANDARD2_0 || NETCOREAPP3_1_OR_GREATER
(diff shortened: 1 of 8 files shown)


src/Microsoft.OData.Core/ODataNotificationStream.cs
```diff
@@ -227,8 +227,29 @@ protected override void Dispose(bool disposing)
...
+            return DisposeAsyncCore();
+        }
+#endif
+
+#if NETSTANDARD2_0 || NETCOREAPP3_1_OR_GREATER
+        /// <summary>
+        /// Asynchronously releases all resources used by the <see cref="ODataNotificationStream"/> object.
+        /// </summary>
+        /// <returns>A task that represents the asynchronous dispose operation.</returns>
+        public async ValueTask DisposeAsyncCore()
```
habbes, line 252: Should this class be made public?

KenitoInc (dismissed):

KenitoInc (approved):

PR #2466, opened 2022-07-22
Implement ODataUtf8JsonWriter async API

### LINKED ISSUE #2421: Implement IJsonWriterAsync to ODataUtf8JsonWriter
Currently `ODataUtf8JsonWriter` only implements the synchronous `IJsonWriter` interface. We should async support and implement `IJsonWriterAsync` as well.

### Assemblies affected

### Reproduce steps

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

habbes:
### Issues

*This pull request fixes #2421 

### Description

This PR adds an implementation of the `IJsonWriterAsync` API to `ODataUtf8JsonWriter`. At the basic level, it adds `Write***Async` methods to the `ODataUtf8JsonWriter` class.

I also had to make some changes to the `ODataJsonLightOutputContext`. This class was tightly coupled to the `JsonWriter` implementation. @gathogojr explained to me the rationale behind this: When writing spatial data we use a class that's part of Microsoft.Spatial library, that library expects an `IJsonWriter` (not the async version) as input. In the async methods we have an `IJsonWriterAsync`. So now we had to make sure that this `IJsonWriterAsync` is actually a JSON writer that implements both `IJsonWriterAsync` and `IJsonWriter`. We had to ensure that it's the same instance of the writer that does all of the writing, and not one writer for the async methods and another writer for the synchronous methods since the writer needs to keep track of internal state about the current scope being written.

Here are the changes I made: instead of relying on `JsonWriter`. I added some checks to ensure that when the user supplies an `IJsonWriter` implementation, that it also implements `IJsonWriterAsync` and vice-versa, otherwise an exception to be thrown.

In 8.x these two interfaces will be unified and that will do away with all these workarounds aimed to prevent potential breaking changes.

#### Perf

According to the following benchmark results, using `ODataMessageWriter` async API with `ODataUtf8JsonWriter` is up to 2x faster and uses significantly less memory.

Benchmakrs

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/IStreamBasedJsonWriterFactory.cs
+++ b/src/Microsoft.OData.Core/Json/IStreamBasedJsonWriterFactory.cs
@@ -11,7 +11,7 @@ using System.Text;
 namespace Microsoft.OData.Json
 {
     /// <summary>
-    /// The interface of factories that create <see cref="IJsonWriter"/> instances
+    /// The interface of factories that create <see cref="IJsonWriterAsync"/> and <see cref="IJsonWriter"/> instances
     /// that write directly to a <see cref="Stream"/> rather than a <see cref="TextWriter"/>.
     /// </summary>
     [CLSCompliant(false)]
--- a/src/Microsoft.OData.Core/Json/IStreamBasedJsonWriterFactory.cs
+++ b/src/Microsoft.OData.Core/Json/IStreamBasedJsonWriterFactory.cs
@@ -11,7 +11,7 @@ using System.Text;
 namespace Microsoft.OData.Json
 {
     /// <summary>
-    /// The interface of factories that create <see cref="IJsonWriterAsync"/> and <see cref="IJsonWriter"/> instances
+    /// The interface of factories that create <see cref="IJsonWriter"/> and <see cref="IJsonWriterAsync"/> instances
     /// that write directly to a <see cref="Stream"/> rather than a <see cref="TextWriter"/>.
     /// </summary>
     [CLSCompliant(false)]
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
@@ -25,7 +25,7 @@ namespace Microsoft.OData.Json
     {
         private const int DefaultBufferSize = 16 * 1024;
         private readonly float bufferFlushThreshold;
-        private readonly static ReadOnlyMemory<byte> Parentheses = new byte[] { (byte)'(', (byte)')' };
+        private readonly static ReadOnlyMemory<byte> parentheses = new byte[] { (byte)'(', (byte)')' };
 
         private readonly Stream outputStream;
         private readonly Stream writeStream;
@@ -319,7 +319,7 @@ namespace Microsoft.OData.Json
         public async Task StartPaddingFunctionScopeAsync()
         {
             await this.FlushAsync().ConfigureAwait(false);
-            await this.writeStream.WriteAsync(Parentheses[..1]).ConfigureAwait(false);
+            await this.writeStream.WriteAsync(parentheses[..1]).ConfigureAwait(false);
         }
 
         public async Task WritePaddingFunctionNameAsync(string functionName)
@@ -331,7 +331,7 @@ namespace Microsoft.OData.Json
         public async Task EndPaddingFunctionScopeAsync()
         {
             await this.FlushAsync().ConfigureAwait(false);
-            await this.writeStream.WriteAsync(Parentheses[1..2]).ConfigureAwait(false);
+            await this.writeStream.WriteAsync(parentheses[1..2]).ConfigureAwait(false);
         }
 
         public async Task StartObjectScopeAsync()
@@ -364,7 +364,6 @@ namespace Microsoft.OData.Json
             await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
         }
 
-        
         public async Task WriteValueAsync(bool value)
         {
             this.writer.WriteBooleanValue(value);
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
@@ -13,17 +13,19 @@ using System.Text;
 using System.Text.Json;
 using System.Text.Encodings.Web;
 using Microsoft.OData.Edm;
+using System.Threading.Tasks;
 
 namespace Microsoft.OData.Json
 {
     /// <summary>
-    /// Implementation of <see cref="IJsonWriter"/> that is based on
+    /// Implementation of <see cref="IJsonWriter"/> and <see cref="IJsonWriterAsync"/>that is based on
     /// <see cref="Utf8JsonWriter"/>.
     /// </summary>
-    internal sealed class ODataUtf8JsonWriter : IJsonWriter, IDisposable
+    internal sealed class ODataUtf8JsonWriter : IJsonWriter, IDisposable, IJsonWriterAsync, IAsyncDisposable
     {
         private const int DefaultBufferSize = 16 * 1024;
         private readonly float bufferFlushThreshold;
+        private readonly static ReadOnlyMemory<byte> Parantheses = new byte[] { (byte)'(', (byte)')' };
 
         private readonly Stream outputStream;
         private readonly Stream writeStream;
@@ -84,7 +86,7 @@ namespace Microsoft.OData.Json
             }
         }
 
-        public void FlushIfBufferThresholdReached()
+        private void FlushIfBufferThresholdReached()
         {
             if (this.writer.BytesPending >= this.bufferFlushThreshold)
             {
@@ -311,6 +313,246 @@ namespace Microsoft.OData.Json
             this.Dispose(disposing: true);
             GC.SuppressFinalize(this);
         }
+
+        #region "Asynchronous API"
+
+        public async Task StartPaddingFunctionScopeAsync()
+        {
+            await this.FlushAsync().ConfigureAwait(false);
+            // writes '('
+            await this.writeStream.WriteAsync(Parantheses[..1]).ConfigureAwait(false);
+        }
+
+        public async Task WritePaddingFunctionNameAsync(string functionName)
+        {
+            await this.FlushAsync().ConfigureAwait(false);
+            await this.writeStream.WriteAsync(Encoding.UTF8.GetBytes(functionName)).ConfigureAwait(false);
+        }
+
+        public async Task EndPaddingFunctionScopeAsync()
+        {
+            await this.FlushAsync().ConfigureAwait(false);
+            // writes ')'
+            await this.writeStream.WriteAsync(Parantheses[1..2]).ConfigureAwait(false);
+        }
+
+        public async Task StartObjectScopeAsync()
+        {
+            this.writer.WriteStartObject();
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task EndObjectScopeAsync()
+        {
+            this.writer.WriteEndObject();
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task StartArrayScopeAsync()
+        {
+            this.writer.WriteStartArray();
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task EndArrayScopeAsync()
+        {
+            this.writer.WriteEndArray();
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteNameAsync(string name)
+        {
+            this.writer.WritePropertyName(name);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        
+        public async Task WriteValueAsync(bool value)
+        {
+            this.writer.WriteBooleanValue(value);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(int value)
+        {
+            this.writer.WriteNumberValue(value);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(float value)
+        {
+            this.writer.WriteNumberValue(value);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(short value)
+        {
+            this.writer.WriteNumberValue(value);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(long value)
+        {
+            if (this.isIeee754Compatible)
+            {
+                this.writer.WriteStringValue(value.ToString(CultureInfo.InvariantCulture));
+            }
+            else
+            {
+                this.writer.WriteNumberValue(value);
+            }
+
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(double value)
+        {
+            this.writer.WriteNumberValue(value);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(Guid value)
+        {
+            this.writer.WriteStringValue(value);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(decimal value)
+        {
+            if (this.isIeee754Compatible)
+            {
+                this.writer.WriteStringValue(value.ToString(CultureInfo.InvariantCulture));
+            }
+            else
+            {
+                this.writer.WriteNumberValue(value);
+            }
+
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(DateTimeOffset value)
+        {
+            this.writer.WriteStringValue(value);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(TimeSpan value)
+        {
+            string stringValue = EdmValueWriter.DurationAsXml(value);
+            this.writer.WriteStringValue(stringValue);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(Date value)
+        {
+            this.writer.WriteStringValue(value.ToString());
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(TimeOfDay value)
+        {
+            this.writer.WriteStringValue(value.ToString());
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(byte value)
+        {
+            this.writer.WriteNumberValue(value);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(sbyte value)
+        {
+            this.writer.WriteNumberValue(value);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(string value)
+        {
+            if (value == null)
+            {
+                this.writer.WriteNullValue();
+            }
+            else
+            {
+                this.writer.WriteStringValue(value.ToString());
+            }
+
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(byte[] value)
+        {
+            if (value == null)
+            {
+                this.writer.WriteNullValue();
+            }
+            else
+            {
+                this.writer.WriteBase64StringValue(value);
+            }
+
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteRawValueAsync(string rawValue)
+        {
+            if (rawValue == null)
+            {
+                return;
+            }
+
+            // Consider using Utf8JsonWriter.WriteRawValue() in .NET 6+
+            // see: https://github.com/OData/odata.net/issues/2420
+
+            await this.FlushAsync().ConfigureAwait(false); // ensure we don't write to the stream while there are still pending data in the buffer
+            await this.writeStream.WriteAsync(Encoding.UTF8.GetBytes(rawValue)).ConfigureAwait(false);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task FlushAsync()
+        {
+            if (this.writer.BytesPending > 0)
+            {
+                await this.writer.FlushAsync().ConfigureAwait(false);
+            }
+        }
+
+        public async ValueTask DisposeAsync()
+        {
+            if (this.writeStream != null)
+            {
+                await this.writeStream.FlushAsync().ConfigureAwait(false);
+            }
+
+            if (this.writer != null)
+            {
+                await this.writer.DisposeAsync().ConfigureAwait(false);
+            }
+
+            if (this.outputStream != this.writeStream && this.writeStream != null)
+            {
+                await this.writeStream.DisposeAsync().ConfigureAwait(false);
+            }
+
+            if (!this.leaveStreamOpen && this.outputStream != null)
+            {
+                await this.outputStream.DisposeAsync().ConfigureAwait(false);
+            }
+
+            this.Dispose(false);
+        }
+
+        private async ValueTask FlushIfBufferThresholdReachedAsync()
+        {
+            if (this.writer.BytesPending >= this.bufferFlushThreshold)
+            {
+                await this.FlushAsync().ConfigureAwait(false);
+            }
+        }
+
+        #endregion
     }
 }
 #endif
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
@@ -319,8 +319,7 @@ namespace Microsoft.OData.Json
         public async Task StartPaddingFunctionScopeAsync()
         {
             await this.FlushAsync().ConfigureAwait(false);
-            // writes '('
-            await this.writeStream.WriteAsync(Parantheses[..1]).ConfigureAwait(false);
+            await this.writeStream.WriteAsync(stackalloc byte[] { '(' }.ConfigureAwait(false);
         }
 
         public async Task WritePaddingFunctionNameAsync(string functionName)
@@ -332,8 +331,7 @@ namespace Microsoft.OData.Json
         public async Task EndPaddingFunctionScopeAsync()
         {
             await this.FlushAsync().ConfigureAwait(false);
-            // writes ')'
-            await this.writeStream.WriteAsync(Parantheses[1..2]).ConfigureAwait(false);
+            await this.writeStream.WriteAsync(stackalloc byte[] {  ')' }]).ConfigureAwait(false);
         }
 
         public async Task StartObjectScopeAsync()
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
@@ -25,7 +25,7 @@ namespace Microsoft.OData.Json
     {
         private const int DefaultBufferSize = 16 * 1024;
         private readonly float bufferFlushThreshold;
-        private readonly static ReadOnlyMemory<byte> Parantheses = new byte[] { (byte)'(', (byte)')' };
+        private readonly static ReadOnlyMemory<byte> Parentheses = new byte[] { (byte)'(', (byte)')' };
 
         private readonly Stream outputStream;
         private readonly Stream writeStream;
@@ -319,7 +319,7 @@ namespace Microsoft.OData.Json
         public async Task StartPaddingFunctionScopeAsync()
         {
             await this.FlushAsync().ConfigureAwait(false);
-            await this.writeStream.WriteAsync(stackalloc byte[] { '(' }.ConfigureAwait(false);
+            await this.writeStream.WriteAsync(Parentheses[..1]).ConfigureAwait(false);
         }
 
         public async Task WritePaddingFunctionNameAsync(string functionName)
@@ -331,7 +331,7 @@ namespace Microsoft.OData.Json
         public async Task EndPaddingFunctionScopeAsync()
         {
             await this.FlushAsync().ConfigureAwait(false);
-            await this.writeStream.WriteAsync(stackalloc byte[] {  ')' }]).ConfigureAwait(false);
+            await this.writeStream.WriteAsync(Parentheses[1..2]).ConfigureAwait(false);
         }
 
         public async Task StartObjectScopeAsync()
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
@@ -77,7 +77,7 @@ namespace Microsoft.OData.JsonLight
         /// Used to guarantee that the synchronous and asynchronous writers share the same scope(s) specifically
         /// because the synchronous writer is used to write spatial data in both synchronous and asynchronous scenarios.
         /// </summary>
-        private JsonWriter concreteJsonWriter;
+        //private JsonWriter concreteJsonWriter;
 
         /// <summary>
         /// Constructor.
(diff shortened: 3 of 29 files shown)


src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -84,7 +86,7 @@ public void Flush()
             }
         }
 
-        public void FlushIfBufferThresholdReached()
```
habbes, line 87: Set this to `private` cause there was no reason for it to be public. It's not a breaking change because the class is internal.

src/Microsoft.OData.Core/Json/IStreamBasedJsonWriterFactory.cs
```diff
@@ -11,7 +11,7 @@
 namespace Microsoft.OData.Json
 {
     /// <summary>
-    /// The interface of factories that create <see cref="IJsonWriter"/> instances
+    /// The interface of factories that create <see cref="IJsonWriterAsync"/> and <see cref="IJsonWriter"/> instances
```
gathogojr, line 14:
```suggestion
    /// The interface of factories that create <see cref="IJsonWriter"/> and <see cref="IJsonWriterAsync"/> instances
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -13,17 +13,19 @@
...
-    /// Implementation of <see cref="IJsonWriter"/> that is based on
+    /// Implementation of <see cref="IJsonWriter"/> and <see cref="IJsonWriterAsync"/>that is based on
     /// <see cref="Utf8JsonWriter"/>.
     /// </summary>
-    internal sealed class ODataUtf8JsonWriter : IJsonWriter, IDisposable
+    internal sealed class ODataUtf8JsonWriter : IJsonWriter, IDisposable, IJsonWriterAsync, IAsyncDisposable
     {
         private const int DefaultBufferSize = 16 * 1024;
         private readonly float bufferFlushThreshold;
+        private readonly static ReadOnlyMemory<byte> Parentheses = new byte[] { (byte)'(', (byte)')' };
```
gathogojr, line 28:
```suggestion
        private readonly static ReadOnlyMemory<byte> parentheses = new byte[] { (byte)'(', (byte)')' };
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -311,6 +313,244 @@ public void Dispose()
...
+        }
+
+        public async Task WriteNameAsync(string name)
+        {
+            this.writer.WritePropertyName(name);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        
+        public async Task WriteValueAsync(bool value)
```
gathogojr, line 368:
```suggestion
        }
        
        public async Task WriteValueAsync(bool value)
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -311,6 +313,244 @@ public void Dispose()
...
+
+        public async Task FlushAsync()
+        {
+            if (this.writer.BytesPending > 0)
+            {
+                await this.writer.FlushAsync().ConfigureAwait(false);
+            }
+        }
+
+        public async ValueTask DisposeAsync()
```
gathogojr, line 520: Shouldn't we be overriding `DisposeAsync` here?

src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
```diff
@@ -103,4 +103,8 @@
       <Version>6.0.3</Version>
     </PackageReference>
   </ItemGroup>
+
```
gathogojr, line 106: How did this change come about?

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -302,6 +446,88 @@ public void WhenInjectingStreamBasedJsonWriterFactory_ThrowException_IfFactoryRe
...
+            settings.SetContentType(ODataFormat.Json);
+            EdmModel model = new EdmModel();
+            using ODataMessageWriter writer = new ODataMessageWriter((IODataRequestMessage)request, settings, model);
+            await writer.WritePropertyAsync(new ODataProperty()
+            {
+                Name = "Name",
+                Value = "This is a test ия"
+            });
+
+            request.GetStream().Position = 0;
```
gathogojr, line 476:
```suggestion
            stream.Position = 0;
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -302,6 +446,88 @@ public void WhenInjectingStreamBasedJsonWriterFactory_ThrowException_IfFactoryRe
...
+            using StreamReader reader = new StreamReader(request.GetStream());
```
gathogojr, line 477:
```suggestion
            using StreamReader reader = new StreamReader(stream);
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -302,6 +446,88 @@ public void WhenInjectingStreamBasedJsonWriterFactory_ThrowException_IfFactoryRe
...
+
+        [Theory]
+        [InlineData("utf-8")]
+        [InlineData("utf-16")]
+        [InlineData("utf-16BE")]
+        [InlineData("utf-32")]
+        public async Task WhenInjectingStreamBasedJsonWriterFactoryAsync_CreatesWriterUsingConfiguredEncoding(string encodingCharset)
+        {
+            // Arrange
+            ODataMessageWriterSettings settings = new ODataMessageWriterSettings();
```
gathogojr, line 490: You could move some of the logic to a helper method and reuse it in this test and the other one above

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -302,6 +446,88 @@ public void WhenInjectingStreamBasedJsonWriterFactory_ThrowException_IfFactoryRe
...
+
+            // Act
+            await writer.WritePropertyAsync(new ODataProperty()
+            {
+                Name = "Name",
+                Value = "This is a test ия"
+            });
+
+            request.GetStream().Position = 0;
+            using StreamReader reader = new StreamReader(request.GetStream(), Encoding.GetEncoding(encodingCharset));
```
gathogojr, line 520:
```suggestion
            using StreamReader reader = new StreamReader(stream, Encoding.GetEncoding(encodingCharset));
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -223,6 +288,85 @@ public void SupportsStreamBasedJsonWriter()
...
+
+            // Act
+            Action writePropertyAction = () => writer.WriteProperty(new ODataProperty()
+            {
+                Name = "Name",
+                Value = "This is a test ия"
+            });
+
+            writePropertyAction.DoesNotThrow();
+            request.GetStream().Position = 0;
```
gathogojr, line 324:
```suggestion
            stream.Position = 0;
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -223,6 +288,85 @@ public void SupportsStreamBasedJsonWriter()
...
+            using StreamReader reader = new StreamReader(request.GetStream(), Encoding.UTF8);
```
gathogojr, line 325:
```suggestion
            using StreamReader reader = new StreamReader(stream, Encoding.UTF8);
```

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataUtf8JsonWriterAsyncTests.cs
```diff
@@ -0,0 +1,454 @@
...
+            this.writer = new ODataUtf8JsonWriter(this.stream, false, Encoding.UTF8, encoder: JavaScriptEncoder.UnsafeRelaxedJsonEscaping);
+            await this.writer.WritePrimitiveValueAsync(input);
+
+            Assert.Equal(expected, await this.ReadStreamAsync());
+        }
+
+        #endregion Custom JavaScriptEncoder
+
+
+        private Task<string> ReadStreamAsync()
```
gathogojr, line 437:
```suggestion
        #endregion Custom JavaScriptEncoder

        private Task<string> ReadStreamAsync()
```

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/MockSyncOnlyJsonWriter.cs
```diff
@@ -0,0 +1,152 @@
...
+    /// <summary>
+    /// A mock implementation of <see cref="IJsonWriter"/>
+    /// that does not implement the async interface <see cref="IJsonWriterAsync"/>
+    /// use for testing purposes.
+    /// 
+    /// Note: most methods are not actually implemented.
+    /// </summary>
+    internal class MockSyncOnlyJsonWriter : IJsonWriter
+    {
+        public void EndArrayScope()
```
gathogojr, line 22: nit: You could keep the class shorter by using the shorter form, e.g., `public void EndArrayScope() => new NotImplementedException();`

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/MockJsonWriter.cs
```diff
@@ -154,5 +155,143 @@ public void Flush()
         {
             throw new NotImplementedException();
         }
+
+        public Task StartPaddingFunctionScopeAsync()
```
gathogojr, line 159: nit: You could keep the class shorter by using the shorter form, e.g., public void StartPaddingFunctionScopeAsync() => new NotImplementedException();

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/MockAsyncOnlyJsonWriter.cs
```diff
@@ -0,0 +1,151 @@
...
+namespace Microsoft.OData.Tests.Json
+{
+    /// <summary>
+    /// A mock implementation of <see cref="IJsonWriterAsync"/> that
+    /// does not implement the synchronous interface <see cref="IJsonWriter"/>.
+    /// This is used for testing purposes. Most of the methods have not been implemented.
+    /// </summary>
+    internal class MockAsyncOnlyJsonWriter : IJsonWriterAsync
+    {
+        public Task EndArrayScopeAsync()
```
gathogojr, line 21: nit: You could keep the class shorter by using the shorter form, e.g., public void EndArrayScopeAsync() => new NotImplementedException();

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/MockJsonWriterFactoryAsync.cs
```diff
@@ -0,0 +1,25 @@
...
+{
+    internal sealed class MockJsonWriterFactoryAsync : IJsonWriterFactoryAsync
+    {
+        private IJsonWriterAsync jsonWriter;
+
+        public MockJsonWriterFactoryAsync(IJsonWriterAsync jsonWriter)
+        {
+            this.jsonWriter = jsonWriter;
+        }
+        public IJsonWriterAsync CreateAsynchronousJsonWriter(TextWriter textWriter, bool isIeee754Compatible)
```
gathogojr, line 20:
```suggestion
        }

        public IJsonWriterAsync CreateAsynchronousJsonWriter(TextWriter textWriter, bool isIeee754Compatible)
```

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -953,32 +945,73 @@ protected override void Dispose(bool disposing)
             IJsonWriterAsync asynchronousJsonWriter;
             if (container == null)
             {
-                if (this.concreteJsonWriter != null)
-                {
-                    return this.concreteJsonWriter;
-                }
+                // we don't create a default JsonWriter here because that's already handled by the CreateJsonWriter() method.
+                return null;
```
gathogojr, line 949: Is `null` the right thing to return? Maybe `return this.jsonWriter`?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -953,32 +945,73 @@ protected override void Dispose(bool disposing)
...
+            {
+                this.asynchronousJsonWriter = jsonWriterAsync;
             }
 
-            this.concreteJsonWriter = asynchronousJsonWriter as JsonWriter;
-            if (this.concreteJsonWriter != null && writerSettings.ArrayPool != null)
+            if (this.asynchronousJsonWriter != null && this.jsonWriter == null)
             {
-                this.concreteJsonWriter.ArrayPool = writerSettings.ArrayPool;
+                if (this.asynchronousJsonWriter is IJsonWriter jsonWriter)
```
gathogojr, line 1001: Naming this local variable `jsonWriter` creates room for confusion

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -953,32 +945,73 @@ protected override void Dispose(bool disposing)
...
+                    this.jsonWriter = jsonWriter;
+                }
+                else
+                {
+                    throw new ODataException(Strings.ODataMessageWriter_IJsonWriterAsync_Must_Implement_IJsonWriter);
+                }
             }
 
-            return asynchronousJsonWriter;
+            if (this.asynchronousJsonWriter != null && !this.asynchronousJsonWriter.Equals(this.jsonWriter))
```
gathogojr, line 1011: I assume/hope this is reference equality

gathogojr, line 1011:
```suggestion
            if (this.asynchronousJsonWriter != null && !object.ReferenceEquals(this.asynchronousJsonWriter, this.jsonWriter))
```

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -953,32 +945,73 @@ protected override void Dispose(bool disposing)
...
-                }
+                throw new ODataException(Strings.ODataMessageWriter_StreamBasedJsonWriterFactory_ReturnedNull(encoding.WebName, isIeee754Compatible));
+            }
 
-                asynchronousJsonWriter = asynchronousJsonWriterFactory.CreateAsynchronousJsonWriter(textWriter, isIeee754Compatible);
-                Debug.Assert(asynchronousJsonWriter != null, "asynchronousJsonWriter != null");
+            return jsonWriter;
+        }
+
+        private void VerifyJsonWriter()
```
gathogojr, line 980: Is `VerifyJsonWriter` a good fit for this method given that it does a lot more than verify?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -953,32 +945,73 @@ protected override void Dispose(bool disposing)
...
             }
 
-            this.concreteJsonWriter = asynchronousJsonWriter as JsonWriter;
-            if (this.concreteJsonWriter != null && writerSettings.ArrayPool != null)
+            if (this.asynchronousJsonWriter != null && this.jsonWriter == null)
             {
-                this.concreteJsonWriter.ArrayPool = writerSettings.ArrayPool;
+                if (this.asynchronousJsonWriter is IJsonWriter jsonWriter)
+                {
+                    this.jsonWriter = jsonWriter;
```
gathogojr, line 1003: How come we don't return after this line?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
habbes: That check is currently done in the `VerifyJsonWriter()`. I guess I could do that here as well, but I reasoned that it would easier to handle matching the different properties in one place, and then keep these `Create**JsonWriter` as simple `static` helper methods. I realize I forgot to make the make static here.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
habbes: No it's not. Not sure how to name it to reflect what it does, that could be a sign that it does too much 🥲

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -953,32 +945,73 @@ protected override void Dispose(bool disposing)
...
+            // for writing spatial data is the same instance as the IJsonWriterAsync used for writing
+            // everything else in order to guarantee that the writer state is correctly maintained
+            // throughout the writing process (e.g. keeping track of the writer's current scope).
+            // When an IJsonWriterAsync is provided, it must also implement IJsonWriter so that the same instance
+            // can be reused for spatial data. If we somehow end up with 2 separate instances, then we fail early with an exception.
+            // Merging IJsonWriter and IJsonWriterAsync interface in a major release will simplify this.
+
+            if (this.jsonWriter != null && this.asynchronousJsonWriter == null && this.jsonWriter is IJsonWriterAsync jsonWriterAsync)
+            {
+                this.asynchronousJsonWriter = jsonWriterAsync;
```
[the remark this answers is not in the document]
habbes: Good point. It should return here.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
habbes: I've renamed it to `syncJsonWriter`.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
habbes: Good point. We should return here.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
habbes: That's a good point. In most cases it should be reference equality. But we can't guarantee that. If the class happens to override `Equals` then this might not work. I have switched to the more reliable `Object.ReferenceEquals()` method.

habbes: Ooh I see you had already suggested it.

src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
habbes: I think it comes after running the strings generator. Let me remove it.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/MockAsyncOnlyJsonWriter.cs
habbes: Thanks for the tip. I'm looking at all the methods I'd have to do this for and wondering whether I should 🥲

habbes: Done!

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
habbes: I've moved code to set up the container in a helper method.

habbes: I've also extended the existing helper methods `WriteAndGetPayload/Async` to handle the test cases I added. It did involve a bit of a parameter bloat.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -13,17 +13,19 @@
 using System.Text.Json;
 using System.Text.Encodings.Web;
 using Microsoft.OData.Edm;
+using System.Threading.Tasks;
 
 namespace Microsoft.OData.Json
 {
     /// <summary>
-    /// Implementation of <see cref="IJsonWriter"/> that is based on
+    /// Implementation of <see cref="IJsonWriter"/> and <see cref="IJsonWriterAsync"/>that is based on
```
ElizabethOkerio, line 21: nit: is this statement complete?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -311,6 +313,243 @@ public void Dispose()
...
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(bool value)
+        {
+            this.writer.WriteBooleanValue(value);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        public async Task WriteValueAsync(int value)
```
ElizabethOkerio, line 373: so these methods are just wrappers of the synchronous methods?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -953,32 +945,73 @@ protected override void Dispose(bool disposing)
...
+                return;
+            }
+
+            if (this.asynchronousJsonWriter != null && this.jsonWriter == null)
             {
-                IJsonWriterFactoryAsync asynchronousJsonWriterFactory = container.GetRequiredService<IJsonWriterFactoryAsync>();
-                if (asynchronousJsonWriterFactory is DefaultJsonWriterFactory && this.concreteJsonWriter != null)
+                if (this.asynchronousJsonWriter is IJsonWriter syncJsonWriter)
```
ElizabethOkerio, line 1002: why didn't you combine the two statements here like the `if` above:

`if (this.asynchronousJsonWriter != null && this.jsonWriter == null && this.asynchronousJsonWriter is IJsonWriter syncJsonWriter)`

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
ElizabethOkerio: adding a comment of what it does can also help

src/Microsoft.OData.Core/ODataMessageWriter.cs
```diff
@@ -1309,19 +1305,17 @@ private Task<TResult> WriteToOutputAsync<TResult>(ODataPayloadKind payloadKind,
...
-                        this.GetOrCreateMessageInfo(streamTask.Result, true),
-                        this.settings))
-                .FollowOnSuccessWithTask(
-                    createOutputContextTask =>
-                    {
-                        this.outputContext = createOutputContextTask.Result;
-                        return writeFunc(this.outputContext);
-                    });
+            return WriteToOutputInnerAsync();
+            async Task<TResult> WriteToOutputInnerAsync()
```
ElizabethOkerio, line 1310: wh is this method caalled `WriteToOutputInnerAsync` and not `WriteToOutputContextAsync`

gathogojr: @ElizabethOkerio The method it belongs to is named `WriteToOutputAsync` so the inner method just adopted that name with `Inner` just before `Async`

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
habbes: Renamed to `EnsureJsonWritersAreTheSame`

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: Yes it's complete. The statement continues on the next line, i.e.:
```
... is based on
<see cref="Utf8JsonWriter">.
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: They wrap the synchronous methods of the inner `Utf8JsonWriter` instance. The `Utf8JsonWriter` exposes only synchronous Write methods. This is because it exclusively writes to memory, it does not write directly to stream, so it never performs any I/O operations via its Write methods that need to be async. That's why after each `Write***Value` I call the `FlushIfBufferThresholdReachedAsync()` method which will write the buffered data to the stream if the buffer has reached a certain threshold. Because this method potentially writes to a stream, it is async.

gathogojr (dismissed):

ElizabethOkerio (dismissed):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
ElizabethOkerio: Can it be documented so they know that they can use Async Json Writer if they do not call the `AddDefaultODataServices()`

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -498,10 +743,21 @@ private async Task<string> WriteAndGetPayloadAsync(IEdmModel edmModel, string co
             }
 
             message.Stream.Seek(0, SeekOrigin.Begin);
-            using (StreamReader reader = new StreamReader(message.Stream))
+            using (StreamReader reader = new StreamReader(message.Stream, encoding: encoding ?? Encoding.UTF8))
             {
-                return reader.ReadToEnd();
+                string contents = await reader.ReadToEndAsync();
```
KenitoInc, line 748:
```suggestion
                string contents = await reader.ReadToEndAsync();

```

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonLight/SingletonBatchRoundtripJsonLightTests.cs
```diff
@@ -995,6 +997,30 @@ public void BatchJsonLightTestUsingJson()
             BatchJsonLightTestUsingBatchFormat(BatchFormat.ApplicationJson, 0);
         }
 
+#if NETCOREAPP3_1_OR_GREATER
+        [Fact]
+        public void BatchJsonLightTest_WithODataUtf8JsonWriter()
```
KenitoInc, line 1002:
```suggestion
        public void MultipartMixedBatchTest_WithODataUtf8JsonWriter()
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonLight/SingletonBatchRoundtripJsonLightTests.cs
```diff
@@ -995,6 +997,30 @@ public void BatchJsonLightTestUsingJson()
...
+#if NETCOREAPP3_1_OR_GREATER
+        [Fact]
+        public void BatchJsonLightTest_WithODataUtf8JsonWriter()
+        {
+            Action<IContainerBuilder> configure = (IContainerBuilder builder) =>
+            {
+                builder.AddService<IStreamBasedJsonWriterFactory>(ServiceLifetime.Singleton, _ => DefaultStreamBasedJsonWriterFactory.Default);
+            };
+
+            BatchJsonLightTestUsingBatchFormat(BatchFormat.MultipartMIME, 0, configure);
```
KenitoInc, line 1009: This name is vague;
Batch is either `Multipart mixed` or `json batch`

src/Microsoft.OData.Core/Json/IStreamBasedJsonWriterFactory.cs
```diff
@@ -11,7 +11,7 @@
 namespace Microsoft.OData.Json
 {
     /// <summary>
-    /// The interface of factories that create <see cref="IJsonWriter"/> instances
+    /// The interface of factories that create <see cref="IJsonWriter"/> and <see cref="IJsonWriterAsync"/> instances
```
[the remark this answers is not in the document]
habbes: There's a full stop, the statement ends on the following line. Here's the full summary:

```c#
/// <summary>
/// The interface of factories that create <see cref="IJsonWriter"/> and <see cref="IJsonWriterAsync"/> instances
/// that write directly to a <see cref="Stream"/> rather than a <see cref="TextWriter"/>.
/// </summary>
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonLight/SingletonBatchRoundtripJsonLightTests.cs
habbes: I was using the same name from exists tests. The convention in these test cases is that when the test case uses multipart, it's not mentioned in the name, and when it uses JSON, it uses the `UsingJson` suffix. I can change the existing tests to use `UsingMultipartMixed` suffix.

ElizabethOkerio (approved):

KenitoInc (dismissed):

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonReaderWriterInjectionTests.cs
```diff
@@ -420,99 +408,126 @@ public void WriteName(string name)
...
-                throw new NotImplementedException();
+                this.textWriter.Flush();
             }
 
-            public void WriteValue(string value)
+            public Task StartPaddingFunctionScopeAsync() => throw new NotImplementedException();
+
+            public Task EndPaddingFunctionScopeAsync() => throw new NotImplementedException();
+
+            public async Task StartObjectScopeAsync()
```
gathogojr, line 462: Does this need to be async or can the body change to `return this.textWriter.WriteAsync('<');`

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonReaderWriterInjectionTests.cs
```diff
@@ -420,99 +408,126 @@ public void WriteName(string name)
...
             {
-                this.textWriter.Write("\"{0}\",", value);
+                await this.textWriter.WriteAsync('<');
             }
 
-            public void WriteValue(byte[] value)
+            public async Task EndObjectScopeAsync()
             {
-                throw new NotImplementedException();
+                await this.textWriter.WriteAsync('>');
```
gathogojr, line 469: Does this need to be async or can the body change to `return this.textWriter.WriteAsync('>');`

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonReaderWriterInjectionTests.cs
```diff
@@ -420,99 +408,126 @@ public void WriteName(string name)
...
             }
 
-            public void WriteValue(Date value)
+            public Task StartArrayScopeAsync() => throw new NotImplementedException();
+
+            public Task EndArrayScopeAsync() => throw new NotImplementedException();
+
+            public async Task WriteNameAsync(string name)
```
gathogojr, line 476: Does this need to be async? It's almost impossible that the code in the `if` block would throw so the last line in the function could change to `return this.textWriter.WriteAsync(string.Format("\"{0}\":", name));`

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonReaderWriterInjectionTests.cs
```diff
@@ -420,99 +408,126 @@ public void WriteName(string name)
...
+
+                await this.textWriter.WriteAsync(string.Format("\"{0}\":", name));
             }
 
-            public void WriteValue(TimeOfDay value)
+            public Task WritePaddingFunctionNameAsync(string functionName) => throw new NotImplementedException();
+
+            public Task WriteValueAsync(bool value) => throw new NotImplementedException();
+
+            public async Task WriteValueAsync(int value)
```
gathogojr, line 490: Does this need to be async or can the body change to `return this.textWriter.WriteAsync(string.Format("{0},", value));`?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
habbes: Yes, will mention that when updating the docs.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonReaderWriterInjectionTests.cs
habbes: It would through if `name` is `null`.

gathogojr (approved):
:shipit:

ElizabethOkerio (approved):

PR #2491, opened 2022-08-20
allow configuring alternate key vocabulary and address performance co…

### LINKED ISSUE #2490: Make alternate key vocabularies configurable

### Assemblies affected

### Reproduce steps

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

corranrogue9:
…ncerns

### Issues

*This pull request fixes #2490.*

### Description

Updated the `AlternateKeyODataUriResolver` to be able to configure which alternate key vocabularies are used and removed the use of a lambda which presented a memory leak concern. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
+++ b/src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
@@ -2156,13 +2156,36 @@ namespace Microsoft.OData.Edm
         /// <returns>Alternate Keys of this type.</returns>
         public static IEnumerable<IDictionary<string, IEdmProperty>> GetAlternateKeysAnnotation(this IEdmModel model, IEdmEntityType type)
         {
+            return GetAlternateKeysAnnotation(model, type, new[] { AlternateKeysVocabularyModel.AlternateKeysTerm, CoreVocabularyModel.AlternateKeysTerm });
+        }
+
+        /// <summary>
+        /// Gets the declared alternate keys of the most defined entity with a declared key present.
+        /// </summary>
+        /// <param name="model">The model to be used.</param>
+        /// <param name="type">Reference to the calling object.</param>
+        /// <param name="alternateKeyTerms">The <see cref="IEdmTerm"/>s that are used within <paramref name="model"/> to represent alternate key annotations</param>
+        /// <returns>Alternate Keys of this type.</returns>
+        /// <exception cref="ArgumentNullException">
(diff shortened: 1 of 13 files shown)


src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -3462,24 +3485,31 @@ internal static bool TryGetStaticEntitySet(this IEdmPathExpression pathExpressio
...
+        private static IEnumerable<IDictionary<string, IEdmProperty>> GetDeclaredAlternateKeysForType(
+            IEdmEntityType type,
+            IEdmModel model, 
+            IEnumerable<IEdmTerm> alternateKeyTerms)
         {
-            IEdmVocabularyAnnotation annotationValue = model.FindVocabularyAnnotations<IEdmVocabularyAnnotation>(type, AlternateKeysVocabularyModel.AlternateKeysTerm).FirstOrDefault();
-            IEdmVocabularyAnnotation coreAnnotationValue = model.FindVocabularyAnnotations<IEdmVocabularyAnnotation>(type, CoreVocabularyModel.AlternateKeysTerm).FirstOrDefault();
-
-            if (annotationValue != null || coreAnnotationValue != null)
+            var any = false;
```
mikepizzo, line 3498: >var [](http://example.com/codeflow?start=12&length=3)

We have traditionally avoided use of "var" in production code (we do use it in test code).

src/Microsoft.OData.Edm/ExtensionMethods/ExtensionMethods.cs
```diff
@@ -3462,24 +3485,31 @@ internal static bool TryGetStaticEntitySet(this IEdmPathExpression pathExpressio
...
 
-                Action<IEdmVocabularyAnnotation> retrieveAnnotationAction = ann =>
+                var annotationValue = model.FindVocabularyAnnotations<IEdmVocabularyAnnotation>(type, alternateKeyTerm).FirstOrDefault();
+                if (annotationValue != null)
                 {
-                    if (ann == null)
-                    {
-                        return;
-                    }
+                    any = true;
```
mikepizzo, line 3510: >any = true; [](http://example.com/codeflow?start=20&length=11)

should we wait to set this until we have a value (i.e., versus the collectionexpression being empty)?

mikepizzo (changes requested):
:clock1:

PR #2497, opened 2022-08-30
Pass correct output stream to IStreamBasedJsonWriterFactory

habbes:
### Issues

*This pull request fixes https://github.com/OData/AspNetCoreOData/issues/673

### Description

The `ODataJsonLightOutputContext` was passing the wrong stream to the `IStreamBasedJsonWriterFactory` when writing an async message. Normally, it [creates a `BufferedStream` that wraps the message stream](https://github.com/habbes/odata.net/blob/master/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs#L108), then uses the buffered stream as the output stream for the async writer. But it was passing the message stream directly to the `IStreamBasedJsonWriterFactory`, meaning that the `ODataUtf8JsonWriter` was bypassing the `BufferedStream` and flushing its output to the internal message stream.

The issue arises because in WebAPI, the `DefaultODataBatchHandler` [copies contents from sub-request responses to the message writer's stream](https://github.com/habbes/odata.net/blob/odataurislim-class-experiment/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs#L117), which in this case happens to be `BufferedStream`, but it uses the JSON writer to write the rest of the metadata in the batch response. This means that we have two writers who are not writing to the same stream. The result is that the bodies of the sub-requests get written out of order because the JSON writer gets flushed before the `BufferedStream`.

The existing batch test cases did not use the async writer. They also did not copy content directly to the stream, and instead used `ODataMessageWriter` instances to write the sub responses. For these reasons, the existing test cases would not detect this bug. I have added test cases to reproduce the issue and to validate more scenarios related to writing batch responses.

Another hidden bug was that the `ODataUtf8JsonWriter` was disposing the output stream when it was closed. This is an error because the writer does not own the stream in this case. The reason it was undetected is because the message stream passed by `ODataMessageWriter` is set to ignore disposals by default. However, `BufferedStream` throws an exception when used after disposal. I have fixed this issue by instructing `ODataUtf8JsonWriter` to leave the stream open when disposed.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchWriterTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchWriterTests.cs
@@ -418,7 +418,7 @@ namespace Microsoft.OData.Core.Tests.JsonLight
                     await jsonLightBatchWriter.WriteStartBatchAsync();
 
                     var customerOperationMessage = await jsonLightBatchWriter.CreateOperationResponseMessageAsync("1");
-                    customerOperationMessage.SetHeader("odata-version", "4.0");
+                    customerOperationMessage.SetHeader(ODataConstants.ODataVersionHeader, "4.0");
                     customerOperationMessage.SetHeader("Content-Type", "application/json");
 
                     // Copies operation's response body to the main batch response's stream
@@ -486,7 +486,7 @@ namespace Microsoft.OData.Core.Tests.JsonLight
                     }
 
                     var orderOperationMessage = await jsonLightBatchWriter.CreateOperationResponseMessageAsync("2");
-                    orderOperationMessage.SetHeader("odata-version", "4.0");
+                    orderOperationMessage.SetHeader(ODataConstants.ODataVersionHeader, "4.0");
                     orderOperationMessage.SetHeader("Content-Type", "application/json");
 
                     using (var orderOperationStream = CreateOrderResponseStream(orderId: "1"))
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchWriterTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchWriterTests.cs
@@ -421,6 +421,8 @@ namespace Microsoft.OData.Core.Tests.JsonLight
                     customerOperationMessage.SetHeader("odata-version", "4.0");
                     customerOperationMessage.SetHeader("Content-Type", "application/json");
 
+                    // Copies operation's response body to the main batch response's stream
+                    // This is similar to how the DefaultODataBatchHandler in WebApi writes batch responses.  
                     using (var customerResponseStream = CreateCustomerResponseStream(customerId: "1"))
                     using (var stream = await customerOperationMessage.GetStreamAsync())
                     {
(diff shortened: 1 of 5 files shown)


KenitoInc (dismissed):

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonLight/SingletonBatchRoundtripJsonLightTests.cs
```diff
@@ -1256,6 +1309,81 @@ private void BatchJsonLightTestUsingBatchFormat(BatchFormat batchFormat, int idx
...
+                case 1:
+                    requestPayload = CreateBatchRequestWithChangesetFirstAndQueryLast(
+                        GetContentTypeHeader(batchFormat));
+                    break;
+
+                default:
+                    throw new ArgumentException("Unknown batch request index value", "idx");
+            }
+
+            VerifyPayload(requestPayload, batchFormat, true /*for request*/, idx);
```
gathogojr, line 1331: Maybe use `<parameterName>: true` instead of `true /*for request*/`

Including other statements below

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchWriterTests.cs
```diff
@@ -405,6 +409,112 @@ public async Task WriteBatchResponseWithChangesetAsync()
...
+                builder.AddService<IStreamBasedJsonWriterFactory>(ServiceLifetime.Singleton, _ =>DefaultStreamBasedJsonWriterFactory.Default);
+            };
+
+            var result = await SetupJsonLightBatchWriterAndRunTestAsync(
+                async (jsonLightBatchWriter) =>
+                {
+                    await jsonLightBatchWriter.WriteStartBatchAsync();
+
+                    var customerOperationMessage = await jsonLightBatchWriter.CreateOperationResponseMessageAsync("1");
+                    customerOperationMessage.SetHeader("odata-version", "4.0");
```
gathogojr, line 476: Not sure these `SetHeader` lines are necessary but if you choose to retain them, maybe change `odata-version` to `OData-Version` since that's what's defined in https://github.com/OData/odata.net/blob/056e8fdb4cdc1ae21a872fdbd5847c527a079b3e/src/Microsoft.OData.Core/ODataConstants.cs#L55

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchWriterTests.cs
```diff
@@ -405,6 +409,112 @@ public async Task WriteBatchResponseWithChangesetAsync()
...
+        [Fact]
+        public async Task WriteBatchResponseAsync_WithStreamCopy()
+        {
+            var result = await SetupJsonLightBatchWriterAndRunTestAsync(
+                async (jsonLightBatchWriter) =>
+                {
+                    await jsonLightBatchWriter.WriteStartBatchAsync();
+
+                    var customerOperationMessage = await jsonLightBatchWriter.CreateOperationResponseMessageAsync("1");
+                    customerOperationMessage.SetHeader("odata-version", "4.0");
```
gathogojr, line 421: Not sure these `SetHeader` lines are necessary but if you choose to retain them, maybe change `odata-version` to `OData-Version` since that's what's defined in https://github.com/OData/odata.net/blob/056e8fdb4cdc1ae21a872fdbd5847c527a079b3e/src/Microsoft.OData.Core/ODataConstants.cs#L55

habbes:
I rewrote the batch tests based on comments from @gathogojr to make them simpler and more targeted to the issues being addressed by this PR.

ElizabethOkerio (approved):

PR #2500, opened 2022-08-31
ODataUriSlim struct

### LINKED ISSUE #2163: Performance - expensive allocations due to ODataUri.Clone()

habbes:
### Issues

*This pull request fixes #2163 

### Description

This PR introduces a new internal struct called `ODataUriSlim` to be used by `ODataWriterCore` instead of cloning `ODataUri` each time it enters a new scope. Now the scope stores a `ODataUriSlim` struct instead of a reference to the cloned `ODataUri`.

Making the type a struct avoids allocating instances to the heap. I have tried to keep the struct small by only storing properties that can be modified from one writer scope to another, namely the `SelectAndExpand` and `Path` properties. The other properties are never changed by the writer, and therefore do not need to be cloned.

To avoid excessive copies of the struct, I passed it by reference to all methods using the `in` keyword. Some copies were inevitable. For example, a copy will be made when assigning the struct value to a property of a class instance. Also, storing the struct in a class makes instance of that class larger because the memory of struct is embedded in the class, instead of a reference to a different memory location. Another case of inevitable copy is when retrieving the value from a property. You cannot pass property value by reference. For example, the following snippet will not compile:
```c#
SomeMethodThatAcceptsAStructByReference(in someObject.ODataUri); // someObject.ODataUri is a struct
```
In such cases I ended doing something like the following:
```c#
ODataUriSlim odataUri = someObject.ODataUri
SomeMethodThatAcceptsAStructByReference(in odataUri);
```

For ease of use, I have added a reference to the `ODataUri` to the `ODataUriSlim` and made the properties of `ODataUri` accessible through `ODataUriSlim`. Technically, this makes the ODataUriSlim slightly bigger by adding a reference, but this might prevent bugs. If these properties were not accessible through `ODataUriSlim`, then I would have had to pass `ODataUri` and `ODataUriSlim` to some methods which need to access the other properties. Using both `ODataUri` and `ODataUriSlim` can lead to bugs because someone can accidentally access the `ODataUri.Path` property, which would be different from the `ODataUriSlim.Path` property.

I have had to refactor a few other methods and classes that the writer passes the `ODataUri` to. In some cases I have changed the input or property type to `ODataUriSlim?` from `ODataUri`. In other cases I have created new method overloads that use `ODataUriSlim` as input instead of `ODataUri`. Luckily, there were not many places where I needed to do this. I avoided adding both `ODataUri` and `ODataUriSlim` properties to the same class because that would introduce a lot more complexity and maintainability issues than is necessary. I also avoided changing classes and methods beyond what the writer needs, because that's beyond the scope of this PR.

I also experimented with making `ODataUriSlim` a class instead of a struct (see: https://github.com/habbes/odata.net/blob/odataurislim-class-experiment/src/Microsoft.OData.Core/Uri/ODataUriSlim.cs). This would still lead to heap allocations, but the objects allocated would be smaller in size. This avoids the inevitable copies of the struct. We can also avoid code duplication by creating a common interface between `ODataUriSlim` and `ODataUri`, say `IODataUri` which we use as the argument for methods that can accept either type. We could have done the same with the struct, but passing a struct to a method that accepts an interface would lead to boxing, which implies heap allocations, which would have defeated my purpose for using a struct. `ODataUriSlim` class experiment ended up allocating slightly more memory than `ODataUriSlim` struct, but the benchmark results were pretty close.

### Benchmark comparisons

According to the [SerializationComparisonTests](https://github.com/OData/odata.net/tree/master/test/PerformanceTests/SerializationComparisonsTests) benchmarks in our repo, use of `ODataUriSlim` struct has resulted in an 11% reduction in allocated memory when using `ODataMessageWriter` synchronous API.  When using the async API, the reduction was 7% when using the default `JsonWriter` and 10% when using `ODataUtf8JsonWriter` (more impact because it has a much smaller memory footprint compared to `JsonWriter` async). I did not observe significant improvement in speed, the figures in the time column fluctuated between different runs of the benchmarks, in some cases they were almost the same.

#### Baseline (before the PR)

**[SerializationComparisons](https://github.com/OData/odata.net/tree/master/test/PerformanceTests/SerializationComparisonsTests)** benchmarks

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.20348
Intel Xeon E-2336 CPU 2.90GHz, 1 CPU, 12 logical and 6 physical cores
.NET SDK=        5.0.404 [C:\Program Files\dotnet\sdk]
  [Host] : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT

Toolchain=InProcessEmitToolchain  InvocationCount=1  UnrollFactor=1

----------------- |----------------------------------------------------- |-----------:|----------:|----------:|-----------:|----------:|-----------:|
 WriteToFileAsync |                                   ODataMessageWriter | 294.143 ms | 0.7563 ms | 0.7075 ms | 41000.0000 |         - | 251,958 KB |
 WriteToFileAsync |                             ODataMessageWriter-Async | 897.045 ms | 2.4466 ms | 2.1689 ms | 66000.0000 | 1000.0000 | 403,831 KB |
 WriteToFileAsync |                    ODataMessageWriter-Utf8JsonWriter | 276.190 ms | 0.5155 ms | 0.4822 ms | 40000.0000 | 1000.0000 | 245,533 KB |
 WriteToFileAsync |              ODataMessageWriter-Utf8JsonWriter-Async | 466.277 ms | 0.7230 ms | 0.6409 ms | 44000.0000 | 1000.0000 | 270,701 KB |

**[ODataWriterFeedTests](https://github.com/OData/odata.net/tree/master/test/PerformanceTests/ComponentTests)** benchmarks

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.20348
Intel Xeon E-2336 CPU 2.90GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=        5.0.404 [C:\Program Files\dotnet\sdk]

Toolchain=InProcessEmitToolchain  InvocationCount=1  UnrollFactor=1

|                                             Method | isModelImmutable |      Mean |    Error |   StdDev |      Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------------------------------------------------- |----------------- |----------:|---------:|---------:|-----------:|----------:|------:|----------:|
|                                          WriteFeed |             True |  83.66 ms | 0.533 ms | 0.498 ms |  9000.0000 |         - |     - |  57.68 MB |
|                            WriteFeedIncludeSpatial |             True | 263.05 ms | 0.353 ms | 0.313 ms | 51000.0000 | 2000.0000 |     - | 308.71 MB |
|                            WriteFeedWithExpansions |             True |  91.39 ms | 0.062 ms | 0.058 ms | 10000.0000 |         - |     - |  63.11 MB |
|              WriteFeedIncludeSpatialWithExpansions |             True | 109.92 ms | 0.073 ms | 0.068 ms | 14000.0000 |         - |     - |  88.23 MB |
|                             WriteFeed_NoValidation |             True |  73.14 ms | 0.073 ms | 0.064 ms |  8000.0000 |         - |     - |  53.47 MB |
|               WriteFeedIncludeSpatial_NoValidation |             True | 244.18 ms | 0.290 ms | 0.271 ms | 50000.0000 | 1000.0000 |     - | 304.04 MB |
|               WriteFeedWithExpansions_NoValidation |             True |  75.56 ms | 0.060 ms | 0.053 ms |  8000.0000 |         - |     - |  52.61 MB |
| WriteFeedIncludeSpatialWithExpansions_NoValidation |             True |  94.08 ms | 0.292 ms | 0.273 ms | 12000.0000 |         - |     - |  77.68 MB |

#### With `ODataUriSlim` struct

**[SerializationComparisons](https://github.com/OData/odata.net/tree/master/test/PerformanceTests/SerializationComparisonsTests)** benchmarks

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.20348
Intel Xeon E-2336 CPU 2.90GHz, 1 CPU, 12 logical and 6 physical cores
.NET SDK=        5.0.404 [C:\Program Files\dotnet\sdk]
  [Host] : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT

Toolchain=InProcessEmitToolchain  InvocationCount=1  UnrollFactor=1

|           Method |                                           WriterName |       Mean |     Error |    StdDev |     Median |      Gen 0 |     Gen 1 |  Allocated |
----------------- |----------------------------------------------------- |-----------:|----------:|----------:|-----------:|-----------:|----------:|-----------:|
 WriteToFileAsync |                                   ODataMessageWriter | 277.459 ms | 0.6373 ms | 0.5961 ms | 277.459 ms | 36000.0000 |         - | 224,025 KB |
 WriteToFileAsync |                             ODataMessageWriter-Async | 877.478 ms | 4.2997 ms | 3.8116 ms | 877.849 ms | 61000.0000 | 1000.0000 | 375,895 KB |
 WriteToFileAsync |                    ODataMessageWriter-Utf8JsonWriter | 255.648 ms | 0.2981 ms | 0.2788 ms | 255.745 ms | 35000.0000 | 1000.0000 | 217,600 KB |
 WriteToFileAsync |              ODataMessageWriter-Utf8JsonWriter-Async | 470.307 ms | 0.6589 ms | 0.6163 ms | 470.346 ms | 39000.0000 | 1000.0000 | 242,768 KB |

**[ODataWriterFeedTests](https://github.com/OData/odata.net/tree/master/test/PerformanceTests/ComponentTests)** benchmarks

 BenchmarkDotNet=v0.12.1, OS=Windows 10.0.20348
Intel Xeon E-2336 CPU 2.90GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=        5.0.404 [C:\Program Files\dotnet\sdk]
  [Host] : .NET Core 3.1.28 (CoreCLR 4.700.22.36202, CoreFX 4.700.22.36301), X64 RyuJIT

Toolchain=InProcessEmitToolchain  InvocationCount=1  UnrollFactor=1

|                                             Method | isModelImmutable |      Mean |    Error |   StdDev |    Median |      Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------------------------------------------------- |----------------- |----------:|---------:|---------:|----------:|-----------:|----------:|------:|----------:|
|                                          WriteFeed |             True |  83.77 ms | 0.142 ms | 0.119 ms |  83.77 ms |  9000.0000 |         - |     - |  55.61 MB |
|                            WriteFeedIncludeSpatial |             True | 266.62 ms | 0.362 ms | 0.339 ms | 266.68 ms | 51000.0000 | 2000.0000 |     - | 306.34 MB |
|                            WriteFeedWithExpansions |             True |  92.23 ms | 0.109 ms | 0.102 ms |  92.19 ms | 10000.0000 |         - |     - |  60.11 MB |
|              WriteFeedIncludeSpatialWithExpansions |             True | 110.49 ms | 0.078 ms | 0.073 ms | 110.50 ms | 14000.0000 |         - |     - |   85.2 MB |
|                             WriteFeed_NoValidation |             True |  73.12 ms | 0.315 ms | 0.280 ms |  73.13 ms |  8000.0000 |         - |     - |   51.4 MB |
|               WriteFeedIncludeSpatial_NoValidation |             True | 246.96 ms | 0.387 ms | 0.362 ms | 246.92 ms | 50000.0000 | 1000.0000 |     - | 301.67 MB |
|               WriteFeedWithExpansions_NoValidation |             True |  75.21 ms | 0.312 ms | 0.292 ms |  75.01 ms |  8000.0000 |         - |     - |  49.61 MB |
| WriteFeedIncludeSpatialWithExpansions_NoValidation |             True |  93.38 ms | 0.108 ms | 0.090 ms |  93.36 ms | 12000.0000 |         - |     - |  74.65 MB |

#### With `ODataUriSlim` class

**SerializationComparisons** benchmarks

 BenchmarkDotNet=v0.13.1, OS=Windows 10.0.20348
Intel Xeon E-2336 CPU 2.90GHz, 1 CPU, 12 logical and 6 physical cores
.NET SDK=        5.0.404 [C:\Program Files\dotnet\sdk]
  [Host] : .NET 6.0.8 (6.0.822.36306), X64 RyuJIT

Toolchain=InProcessEmitToolchain  InvocationCount=1  UnrollFactor=1

----------------- |----------------------------------------------------- |-----------:|----------:|----------:|-----------:|----------:|-----------:|
 WriteToFileAsync |                                   ODataMessageWriter | 284.516 ms | 0.7395 ms | 0.6918 ms | 37000.0000 |         - | 227,321 KB |
 WriteToFileAsync |                             ODataMessageWriter-Async | 890.175 ms | 7.5653 ms | 6.7064 ms | 62000.0000 | 1000.0000 | 379,179 KB |
 WriteToFileAsync |                    ODataMessageWriter-Utf8JsonWriter | 265.688 ms | 0.4744 ms | 0.4438 ms | 36000.0000 | 1000.0000 | 220,905 KB |
 WriteToFileAsync |              ODataMessageWriter-Utf8JsonWriter-Async | 479.593 ms | 0.2380 ms | 0.1987 ms | 40000.0000 | 1000.0000 | 246,073 KB |

**ODataWriterFeedTests** benchmarks

 BenchmarkDotNet=v0.12.1, OS=Windows 10.0.20348
Intel Xeon E-2336 CPU 2.90GHz, 1 CPU, 12 logical and 6 physical cores
.NET Core SDK=        5.0.404 [C:\Program Files\dotnet\sdk]
  [Host] : .NET Core 3.1.28 (CoreCLR 4.700.22.36202, CoreFX 4.700.22.36301), X64 RyuJIT

Toolchain=InProcessEmitToolchain  InvocationCount=1  UnrollFactor=1

|                                             Method | isModelImmutable |      Mean |    Error |   StdDev |      Gen 0 |     Gen 1 | Gen 2 | Allocated |
|--------------------------------------------------- |----------------- |----------:|---------:|---------:|-----------:|----------:|------:|----------:|
|                                          WriteFeed |             True |  84.19 ms | 0.425 ms | 0.398 ms |  9000.0000 |         - |     - |  55.89 MB |
|                            WriteFeedIncludeSpatial |             True | 264.23 ms | 0.259 ms | 0.229 ms | 51000.0000 | 2000.0000 |     - | 306.65 MB |
|                            WriteFeedWithExpansions |             True |  92.59 ms | 0.116 ms | 0.097 ms | 10000.0000 |         - |     - |  60.51 MB |
|              WriteFeedIncludeSpatialWithExpansions |             True | 110.50 ms | 0.105 ms | 0.093 ms | 14000.0000 |         - |     - |   85.6 MB |
|                             WriteFeed_NoValidation |             True |  72.41 ms | 0.192 ms | 0.180 ms |  8000.0000 |         - |     - |  51.68 MB |
|               WriteFeedIncludeSpatial_NoValidation |             True | 243.16 ms | 0.200 ms | 0.178 ms | 50000.0000 | 1000.0000 |     - | 301.98 MB |
|               WriteFeedWithExpansions_NoValidation |             True |  75.64 ms | 0.075 ms | 0.067 ms |  8000.0000 |         - |     - |  50.01 MB |
| WriteFeedIncludeSpatialWithExpansions_NoValidation |             True |  93.07 ms | 0.129 ms | 0.107 ms | 12000.0000 |         - |     - |  75.05 MB |

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Evaluation/ODataMetadataContext.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataMetadataContext.cs
@@ -36,7 +36,7 @@ namespace Microsoft.OData.Evaluation
         /// <summary>
         /// Gets the OData uri.
         /// </summary>
-        ODataUri ODataUri { get; }
+        ODataUriSlim ODataUri { get; }
 
         /// <summary>
         /// Gets an entity metadata builder for the given resource.
@@ -95,7 +95,7 @@ namespace Microsoft.OData.Evaluation
         /// <summary>
         /// The OData Uri.
         /// </summary>
-        private readonly ODataUri odataUri;
+        private readonly ODataUriSlim odataUri;
 
         /// <summary>
         /// The service base Uri.
@@ -115,8 +115,8 @@ namespace Microsoft.OData.Evaluation
         /// <param name="metadataDocumentUri">The metadata document uri.</param>
         /// <param name="odataUri">The request Uri.</param>
         /// <remarks>This overload should only be used by the writer.</remarks>
-        public ODataMetadataContext(bool isResponse, IEdmModel model, Uri metadataDocumentUri, ODataUri odataUri)
-            : this(isResponse, /*OperationsBoundToEntityTypeMustBeContainerQualified*/ null, EdmTypeWriterResolver.Instance, model, metadataDocumentUri, odataUri)
+        public ODataMetadataContext(bool isResponse, IEdmModel model, Uri metadataDocumentUri, in ODataUriSlim odataUri)
+            : this(isResponse, /*OperationsBoundToEntityTypeMustBeContainerQualified*/ null, EdmTypeWriterResolver.Instance, model, metadataDocumentUri, in odataUri)
         {
         }
 
@@ -136,7 +136,7 @@ namespace Microsoft.OData.Evaluation
             EdmTypeResolver edmTypeResolver,
             IEdmModel model,
             Uri metadataDocumentUri,
-            ODataUri odataUri)
+            in ODataUriSlim odataUri)
         {
             Debug.Assert(edmTypeResolver != null, "edmTypeResolver != null");
             Debug.Assert(model != null, "model != null");
@@ -166,9 +166,9 @@ namespace Microsoft.OData.Evaluation
             EdmTypeResolver edmTypeResolver,
             IEdmModel model,
             Uri metadataDocumentUri,
-            ODataUri odataUri,
+            in ODataUriSlim odataUri,
             JsonLightMetadataLevel metadataLevel)
-            : this(isResponse, operationsBoundToEntityTypeMustBeContainerQualified, edmTypeResolver, model, metadataDocumentUri, odataUri)
+            : this(isResponse, operationsBoundToEntityTypeMustBeContainerQualified, edmTypeResolver, model, metadataDocumentUri, in odataUri)
         {
             Debug.Assert(metadataLevel != null, "MetadataLevel != null");
 
@@ -215,7 +215,7 @@ namespace Microsoft.OData.Evaluation
         /// <summary>
         /// Gets the OData uri.
         /// </summary>
-        public ODataUri ODataUri
+        public ODataUriSlim ODataUri
         {
             get
             {
--- a/src/Microsoft.OData.Core/Evaluation/ODataMetadataContext.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataMetadataContext.cs
@@ -36,7 +36,7 @@ namespace Microsoft.OData.Evaluation
         /// <summary>
         /// Gets the OData uri.
         /// </summary>
-        ODataUriSlim ODataUri { get; }
+        ODataUriSlim? ODataUri { get; }
 
         /// <summary>
         /// Gets an entity metadata builder for the given resource.
@@ -95,7 +95,7 @@ namespace Microsoft.OData.Evaluation
         /// <summary>
         /// The OData Uri.
         /// </summary>
-        private readonly ODataUriSlim odataUri;
+        private readonly ODataUriSlim? odataUri;
 
         /// <summary>
         /// The service base Uri.
@@ -115,7 +115,7 @@ namespace Microsoft.OData.Evaluation
         /// <param name="metadataDocumentUri">The metadata document uri.</param>
         /// <param name="odataUri">The request Uri.</param>
         /// <remarks>This overload should only be used by the writer.</remarks>
-        public ODataMetadataContext(bool isResponse, IEdmModel model, Uri metadataDocumentUri, in ODataUriSlim odataUri)
+        public ODataMetadataContext(bool isResponse, IEdmModel model, Uri metadataDocumentUri, in ODataUriSlim? odataUri)
             : this(isResponse, /*OperationsBoundToEntityTypeMustBeContainerQualified*/ null, EdmTypeWriterResolver.Instance, model, metadataDocumentUri, in odataUri)
         {
         }
@@ -136,7 +136,7 @@ namespace Microsoft.OData.Evaluation
             EdmTypeResolver edmTypeResolver,
             IEdmModel model,
             Uri metadataDocumentUri,
-            in ODataUriSlim odataUri)
+            in ODataUriSlim? odataUri)
         {
             Debug.Assert(edmTypeResolver != null, "edmTypeResolver != null");
             Debug.Assert(model != null, "model != null");
@@ -166,7 +166,7 @@ namespace Microsoft.OData.Evaluation
             EdmTypeResolver edmTypeResolver,
             IEdmModel model,
             Uri metadataDocumentUri,
-            in ODataUriSlim odataUri,
+            in ODataUriSlim? odataUri,
             JsonLightMetadataLevel metadataLevel)
             : this(isResponse, operationsBoundToEntityTypeMustBeContainerQualified, edmTypeResolver, model, metadataDocumentUri, in odataUri)
         {
@@ -215,7 +215,7 @@ namespace Microsoft.OData.Evaluation
         /// <summary>
         /// Gets the OData uri.
         /// </summary>
-        public ODataUriSlim ODataUri
+        public ODataUriSlim? ODataUri
         {
             get
             {
--- a/src/Microsoft.OData.Core/ODataWriterCore.cs
+++ b/src/Microsoft.OData.Core/ODataWriterCore.cs
@@ -87,7 +87,7 @@ namespace Microsoft.OData
                 itemType: resourceType,
                 skipWriting: false,
                 selectedProperties: outputContext.MessageWriterSettings.SelectedProperties,
-                odataUri: in odataUri,
+                odataUri: odataUri,
                 enableDelta: true));
             this.CurrentScope.DerivedTypeConstraints = this.outputContext.Model.GetDerivedTypeConstraints(navigationSource)?.ToList();
         }
@@ -1250,7 +1250,7 @@ namespace Microsoft.OData
             in ODataUriSlim odataUri)
         {
             Debug.Assert(this.CurrentScope != null, "Creating a nested resource info scope with a null parent scope.");
-            return new NestedResourceInfoScope(writerState, navLink, navigationSource, itemType, skipWriting, selectedProperties, in odataUri, this.CurrentScope);
+            return new NestedResourceInfoScope(writerState, navLink, navigationSource, itemType, skipWriting, selectedProperties, odataUri, this.CurrentScope);
         }
 
         /// <summary>
@@ -2801,7 +2801,7 @@ namespace Microsoft.OData
                 navigationSource = this.CurrentScope.NavigationSource ?? odataUri.Path.TargetNavigationSource();
             }
 
-            this.PushScope(newState, item, navigationSource, itemType, skipWriting, selectedProperties, in odataUri, derivedTypeConstraints);
+            this.PushScope(newState, item, navigationSource, itemType, skipWriting, selectedProperties, odataUri, derivedTypeConstraints);
 
             this.NotifyListener(newState);
         }
@@ -2850,8 +2850,15 @@ namespace Microsoft.OData
             {
                 Scope startScope = this.scopeStack.Pop();
                 Debug.Assert(startScope.State == WriterState.Start, "startScope.State == WriterState.Start");
-                ODataUriSlim odataUri = startScope.ODataUri;
-                this.PushScope(WriterState.Completed, /*item*/null, startScope.NavigationSource, startScope.ResourceType, /*skipWriting*/false, startScope.SelectedProperties, in odataUri, null);
+                this.PushScope(
+                    state: WriterState.Completed,
+                    item: null,
+                    navigationSource: startScope.NavigationSource,
+                    itemType: startScope.ResourceType,
+                    skipWriting: false,
+                    selectedProperties: startScope.SelectedProperties,
+                    odataUri: startScope.ODataUri,
+                    derivedTypeConstraints: null);
                 this.InterceptException((thisParam) => thisParam.EndPayload());
                 this.NotifyListener(WriterState.Completed);
             }
@@ -3073,17 +3080,17 @@ namespace Microsoft.OData
             switch (state)
             {
                 case WriterState.Resource:
-                    scope = this.CreateResourceScope((ODataResource)item, navigationSource, resourceType, skipWriting, selectedProperties, in odataUri, isUndeclaredResourceOrResourceSet);
+                    scope = this.CreateResourceScope((ODataResource)item, navigationSource, resourceType, skipWriting, selectedProperties, odataUri, isUndeclaredResourceOrResourceSet);
                     break;
                 case WriterState.DeletedResource:
-                    scope = this.CreateDeletedResourceScope((ODataDeletedResource)item, navigationSource, (IEdmEntityType)itemType, skipWriting, selectedProperties, in odataUri, isUndeclaredResourceOrResourceSet);
+                    scope = this.CreateDeletedResourceScope((ODataDeletedResource)item, navigationSource, (IEdmEntityType)itemType, skipWriting, selectedProperties, odataUri, isUndeclaredResourceOrResourceSet);
                     break;
                 case WriterState.DeltaLink:
                 case WriterState.DeltaDeletedLink:
-                    scope = this.CreateDeltaLinkScope((ODataDeltaLinkBase)item, navigationSource, (IEdmEntityType)itemType, selectedProperties, in odataUri);
+                    scope = this.CreateDeltaLinkScope((ODataDeltaLinkBase)item, navigationSource, (IEdmEntityType)itemType, selectedProperties, odataUri);
                     break;
                 case WriterState.ResourceSet:
-                    scope = this.CreateResourceSetScope((ODataResourceSet)item, navigationSource, itemType, skipWriting, selectedProperties, in odataUri, isUndeclaredResourceOrResourceSet);
+                    scope = this.CreateResourceSetScope((ODataResourceSet)item, navigationSource, itemType, skipWriting, selectedProperties, odataUri, isUndeclaredResourceOrResourceSet);
                     if (this.outputContext.Model.IsUserModel())
                     {
                         Debug.Assert(scope is ResourceSetBaseScope, "Create a scope for a resource set that is not a ResourceSetBaseScope");
@@ -3092,7 +3099,7 @@ namespace Microsoft.OData
 
                     break;
                 case WriterState.DeltaResourceSet:
-                    scope = this.CreateDeltaResourceSetScope((ODataDeltaResourceSet)item, navigationSource, resourceType, skipWriting, selectedProperties, in odataUri, isUndeclaredResourceOrResourceSet);
+                    scope = this.CreateDeltaResourceSetScope((ODataDeltaResourceSet)item, navigationSource, resourceType, skipWriting, selectedProperties, odataUri, isUndeclaredResourceOrResourceSet);
                     if (this.outputContext.Model.IsUserModel())
                     {
                         Debug.Assert(scope is ResourceSetBaseScope, "Create a scope for a delta resource set that is not a ResourceSetBaseScope");
@@ -3101,21 +3108,21 @@ namespace Microsoft.OData
 
                     break;
                 case WriterState.Property:
-                    scope = this.CreatePropertyInfoScope((ODataPropertyInfo)item, navigationSource, resourceType, selectedProperties, in odataUri);
+                    scope = this.CreatePropertyInfoScope((ODataPropertyInfo)item, navigationSource, resourceType, selectedProperties, odataUri);
                     break;
                 case WriterState.NestedResourceInfo:            // fall through
                 case WriterState.NestedResourceInfoWithContent:
-                    scope = this.CreateNestedResourceInfoScope(state, (ODataNestedResourceInfo)item, navigationSource, itemType, skipWriting, selectedProperties, in odataUri);
+                    scope = this.CreateNestedResourceInfoScope(state, (ODataNestedResourceInfo)item, navigationSource, itemType, skipWriting, selectedProperties, odataUri);
                     break;
                 case WriterState.Start:
-                    scope = new Scope(state, item, navigationSource, itemType, skipWriting, selectedProperties, in odataUri, /*enableDelta*/ true);
+                    scope = new Scope(state, item, navigationSource, itemType, skipWriting, selectedProperties, odataUri, /*enableDelta*/ true);
                     break;
                 case WriterState.Primitive:                 // fall through
                 case WriterState.Stream:                    // fall through
                 case WriterState.String:                    // fall through
                 case WriterState.Completed:                 // fall through
                 case WriterState.Error:
-                    scope = new Scope(state, item, navigationSource, itemType, skipWriting, selectedProperties, in odataUri, /*enableDelta*/ false);
+                    scope = new Scope(state, item, navigationSource, itemType, skipWriting, selectedProperties, odataUri, /*enableDelta*/ false);
                     break;
                 default:
                     string errorMessage = Strings.General_InternalError(InternalErrorCodes.ODataWriterCore_Scope_Create_UnreachableCodePath);
@@ -3655,7 +3662,6 @@ namespace Microsoft.OData
             {
                 Scope startScope = this.scopeStack.Pop();
                 Debug.Assert(startScope.State == WriterState.Start, "startScope.State == WriterState.Start");
-                ODataUriSlim odataUri = startScope.ODataUri;
                 this.PushScope(
                     WriterState.Completed,
                     /*item*/ null,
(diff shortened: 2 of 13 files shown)


src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
```diff
@@ -0,0 +1,139 @@
...
+
+namespace Microsoft.OData
+{
+    /// <summary>
+    /// A lightweight version of <see cref="ODataUri"/> that is
+    /// used by the writer to efficiently create a copy of the URI when entering
+    /// each scope. This is used to avoid cloning <see cref="ODataUri"/>
+    /// excessively.
+    /// </summary>
+    internal struct ODataUriSlim
```
xuzhg, line 20: can we 'implicitly' convert between ODataUri and ODatUriSlim, back and forth?

xuzhg (approved):

src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
habbes: We could add implicit conversion from `ODataUri` to `ODataUriSlim` such that someone is able to do assign an `ODataUriSlim` variable an `ODataUri` value:
```c#
ODataUri odataUri = ...;
ODataUriSlim odataUriSlim = odataUri;
```
or pass an `ODataUri` to a method expecting an `ODataUriSlim`:

```c#
public void Foo(in ODataUriSlim odataUriSlim)
{
 ....
}
// ..

ODataUri odataUri ...;
Foo(odataUri)
```

But I don't think we should add conversion from `ODataUriSlim` to `ODataUri`. Since the `ODataUri` instance referenced in `ODataUriSlim` might have different values for some of the properties, exposing it could lead to inconsistencies or unwanted mutations to the `ODataUri`. To avoid that, we may need to clone the `ODataUri`, which beats the purpose of `ODataUriSlim`.

In the current code, I don't see a lot of use-cases for `ODataUri`-`ODataUriSlim` implicit conversions. I also think keeping things explicit (like in the current PR) makes it less likely that someone will use the wrong type accidentally.

src/Microsoft.OData.Core/Evaluation/ODataMetadataContext.cs
```diff
@@ -95,7 +95,7 @@ internal sealed class ODataMetadataContext : IODataMetadataContext
         /// <summary>
         /// The OData Uri.
         /// </summary>
-        private readonly ODataUri odataUri;
+        private readonly ODataUriSlim? odataUri;
```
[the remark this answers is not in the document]
habbes: Because in the previous code, an `ODataMetadataContext` can be created with a `null` `ODataUri`. This happens in [ODataJsonLightDeserializer(https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs#L126)]. The deserializer passes the `this.ODataUri` property to the `ODataMetadataContext` constructor. The `ODataJsonLightDeserializer.ODataUri` property can be null as seen [here](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs#L106).

So for this PR, I needed a way to handle the situation where `IODataMetadataContext.ODataUri` can be null. Since `ODataUriSlim` is a struct, it can't be null, so that's why I decided to use the nullable version `ODataUriSlim?` for that. `Nullable<T>` is a struct, so we don't incur extra heap allocations when making the struct nullable.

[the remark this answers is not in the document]
gathogojr: @KenitoInc I'm guessing because when `ODataUri` class was used in the field declaration, it could accommodate nulls out of the box. It's therefore necessary to declare the field as nullable when using `ODataUriSlim` struct (value type)

src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// </summary>
+        public ODataPath Path { get; set; }
+
+        /// <summary>
+        /// Gets or sets any $select or $expand option for this uri.
+        /// </summary>
+        public SelectExpandClause SelectAndExpand { get; set; }
+
+        /// <summary>
+        /// Gets or sets the request Uri.
```
gathogojr, line 60:
```suggestion
        /// Gets the request Uri.
```
Justification: Property is read only

src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// </summary>
+        public Uri RequestUri => this.odataUri.RequestUri;
+
+        /// <summary>
+        /// Gets or sets the service root Uri.
```
gathogojr, line 65:
```suggestion
        /// Gets the service root Uri.
```
Justification: Property is read only

src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// </summary>
+        public Uri ServiceRoot => this.odataUri.ServiceRoot;
+
+        /// <summary>
+        /// Get the parameter alias nodes info.
```
gathogojr, line 70:
```suggestion
        /// Gets the parameter alias nodes info.
```

src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// </summary>
+        public IDictionary<string, SingleValueNode> ParameterAliasNodes => this.odataUri.ParameterAliasNodes;
+
+        /// <summary>
+        /// Gets or sets any custom query options for this uri.
```
gathogojr, line 75:
```suggestion
        /// Gets any custom query options for this uri.
```
Justification: Property is read only

src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// </summary>
+        public IEnumerable<QueryNode> CustomQueryOptions => this.odataUri.CustomQueryOptions;
+
+        /// <summary>
+        /// Gets or sets any $filter option for this uri.
+        /// </summary>
+        public FilterClause Filter => this.odataUri.Filter;
+
+        /// <summary>
+        /// Gets or sets any $orderby option for this uri.
```
gathogojr, line 85:
```suggestion
        /// Gets any $orderby option for this uri.
```

src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// </summary>
+        public OrderByClause OrderBy => this.odataUri.OrderBy;
+
+        /// <summary>
+        /// Gets or sets any $search option for this uri.
```
gathogojr, line 90:
```suggestion
        /// Gets any $search option for this uri.
```

src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
```diff
@@ -0,0 +1,139 @@
...
+        /// </summary>
+        public SearchClause Search => this.odataUri.Search;
+
+        /// <summary>
+        /// Gets or sets any $apply option for this uri.
```
gathogojr, line 95:
```suggestion
        /// Gets any $apply option for this uri.
```
Justification: Property is read only

src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
```diff
@@ -0,0 +1,139 @@
...
+
+        /// <summary>
+        /// Gets or sets any $deltatoken option for this uri.
+        /// </summary>
+        public string DeltaToken => this.odataUri.DeltaToken;
+
+        /// <summary>
+        /// Get or sets the MetadataDocumentUri, which is always ServiceRoot + $metadata
+        /// </summary>
+        internal Uri MetadataDocumentUri => this.odataUri.MetadataDocumentUri;
```
gathogojr, line 137: Remove `or sets ` part from the property doc

src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
```diff
@@ -0,0 +1,139 @@
...
+        }
+
+        /// <summary>
+        /// Creates an instance of <see cref="ODataUriSlim"/> by
+        /// cloning the provided <see cref="ODataUriSlim"/>.
+        /// </summary>
+        /// <param name="odataUriSlim">The URI to copy.</param>
+        public ODataUriSlim(ODataUriSlim odataUriSlim)
+        {
+            this.odataUri = odataUriSlim.odataUri;
```
gathogojr, line 44: In quite a few places, the `ODataUriSlim` property is declared as nullable. We have any guarantees that this constructor can't be called with a null value for `odataUriSlim`?

src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
```diff
@@ -0,0 +1,139 @@
...
+        private readonly ODataUri odataUri;
+
+        /// <summary>
+        /// Creates an instance of <see cref="ODataUriSlim"/>
+        /// by copying the provided <see cref="ODataUri"/> instance.
+        /// </summary>
+        /// <param name="odataUri">The URI to copy.</param>
+        public ODataUriSlim(ODataUri odataUri)
+        {
+            Debug.Assert(odataUri != null, "odataUri != null");
```
gathogojr, line 31: Maybe throw an `ArgumentNullException`?

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -70,17 +70,17 @@ internal abstract class ODataWriterCore : ODataWriter, IODataOutputInStreamError
                 resourceType = this.outputContext.EdmTypeResolver.GetElementType(navigationSource);
             }
 
-            ODataUri odataUri = outputContext.MessageWriterSettings.ODataUri.Clone();
+            ODataUriSlim odataUri = new ODataUriSlim(outputContext.MessageWriterSettings.ODataUri);
```
gathogojr, line 73: What happens when `outputContext.MessageWriterSettings.ODataUri` is null?

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -2601,8 +2601,9 @@ private void EnterScope(WriterState newState, ODataItem item)
                                 if (serializationInfo.NavigationSourceName != null)
                                 {
                                     ODataUriParser uriParser = new ODataUriParser(model, new Uri(serializationInfo.NavigationSourceName, UriKind.Relative), this.outputContext.Container);
-                                    odataUri = uriParser.ParseUri();
-                                    navigationSource = odataUri.Path.NavigationSource();
+                                    ODataUri odataUri = uriParser.ParseUri();
+                                    odataUriSlim = new ODataUriSlim(odataUri);
+                                    navigationSource = odataUriSlim.Path.NavigationSource();
```
gathogojr, line 2606:
```suggestion
                                    ODataUri odataUri = new ODataUriSlim(uriParser.ParseUri());
                                    navigationSource = odataUri.Path.NavigationSource();
```
I'd also be curious whether `uriParser.ParseUri()` can return a null...

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -70,17 +70,17 @@ internal abstract class ODataWriterCore : ODataWriter, IODataOutputInStreamError
...
-            if (!writingResourceSet && odataUri != null && odataUri.Path != null)
+            if (!writingResourceSet && odataUri.Path != null)
             {
                 odataUri.Path = odataUri.Path.TrimEndingKeySegment();
             }
 
             this.listener = listener;
 
-            this.scopeStack.Push(new Scope(WriterState.Start, /*item*/null, navigationSource, resourceType, /*skipWriting*/false, outputContext.MessageWriterSettings.SelectedProperties, odataUri, /*enableDelta*/ true));
+            this.scopeStack.Push(new Scope(WriterState.Start, /*item*/null, navigationSource, resourceType, /*skipWriting*/false, outputContext.MessageWriterSettings.SelectedProperties, in odataUri, /*enableDelta*/ true));
```
gathogojr, line 83: Because you "touched" this line (and others that share the same pattern), I'd have suggested that you replace `/* ... */` with named parameters. However, feel free to disregard this since with the version of C# we're building the project against on ADO you'd have to apply named parameters for every parameter following the first named parameter.

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -2561,10 +2561,10 @@ private void EnterScope(WriterState newState, ODataItem item)
             IEdmNavigationSource navigationSource = null;
             IEdmType itemType = null;
             SelectedPropertiesNode selectedProperties = currentScope.SelectedProperties;
-            ODataUri odataUri = currentScope.ODataUri.Clone();
-            if (odataUri.Path == null)
+            ODataUriSlim odataUriSlim = new ODataUriSlim(currentScope.ODataUri);
```
gathogojr, line 2564:
```suggestion
            ODataUriSlim odataUri = new ODataUriSlim(currentScope.ODataUri);
```
I wish you could retain the variable name as `odataUri`. It'd even makes your diff smaller

src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
habbes: Yes, it's guaranteed by the compiler. You cannot assign `null` to `ODataUriSlim` variable. Similarly, you cannot assign a variable of type `ODataUriSlim?` (nullable) to a variable of type `ODataUriSlim` (not nullable). You can only go from `ODataUriSlim` to `ODataUriSlim?`, but not the other way around. Otherwise, the code will not compile.

src/Microsoft.OData.Core/ODataWriterCore.cs
habbes: `ODataMessageWriterSettings.ODataUri` getter initializes a new instance of `ODataUri` if none was set. So, it can't return null. The old code also did not check for null at this site.

Also, in case it were null here, it's probably better to just fail than try to proceed with a null URL since most of `ODataWriterCore` expect an `ODataUri` to be present.

src/Microsoft.OData.Core/ODataWriterCore.cs
habbes: `uriParse.ParseUri` does not return null. It creates a new instance of `ODataUri` directly and returns it.

src/Microsoft.OData.Core/ODataWriterCore.cs
habbes: I thought using `odataUriSlim` would make the changes and intentions more clear, and reduce confusion if we have both `ODataUri` and `ODataUriSlim` instances in the same scope. But I'm open to changing if you feel like it's better.

src/Microsoft.OData.Core/ODataWriterCore.cs
habbes: I've opted to name all parameters

src/Microsoft.OData.Core/Uri/ODataUriSlim.cs
habbes: I'm a bit conflicted since this add null check for something that is called often but should never be null unless we have an internal bug (not a user-input error).

gathogojr (approved):
:shipit:

KenitoInc (approved):

PR #2505, opened 2022-09-06
ignore changes to test files in PR Quantifier

mikepizzo:
### Description

The PR Quantifier currently complains about PRs over 200 lines of code, but with test changes even minor PRs exceed that boundary.

Adding a prquantifier.yaml file to limit the quantifier to look at code in the source directory.

### Additional work necessary

If this works on this repo, we should add similar logic to the other repos as well.


prquantifier.yaml
```diff
@@ -0,0 +1,3058 @@
...
+  Value: 1000
+  Label: Extra Large
+  Color: Red
+  Formula: Sum
+LanguageOptions:
+  IgnoreSpaces: true
+  IgnoreComments: true
+  IgnoreCodeBlockSeparator: true
+DynamicBehaviour: false
+IgnoreRenamed: true
```
corranrogue9, line 49: Renaming files in c# usually results in renaming classes, which although only tedious, I think we might want to reflect in the PR size
```suggestion
IgnoreRenamed: false
```

prquantifier.yaml
```diff
@@ -0,0 +1,3058 @@
...
+IgnoreCopied: true
```
corranrogue9, line 50: I'm not exactly sure what this means, but if someone copy/pastes something and then only changes 5% of it, the whole thing needs to be confirmed for correctness, so we probably want to include this in the size as well
```suggestion
IgnoreCopied: false
```

xuzhg (approved):

prquantifier.yaml
```diff
@@ -0,0 +1,3058 @@
...
+AdditionPercentile:
```
KenitoInc, line 51: What do these figures represent?

PR #2506, opened 2022-09-09
Replace ConditionalWeakTable with simpler non-static Dictionary-based cache

### LINKED ISSUE #2503: OData Client reportedly cause excessive memory usage after .NET 6 upgrade
A customer reported that after upgrading to .NET 6  from .NET Core 3.1 and updating OData Client to 7.10.0, their performance and load testing reveal excessive memory usage and they have attributed it to OData client. Here are parts of the report:

Performance testing has revealed a memory issue: high private bytes being consumed in the App Service during load testing. You can see that before the migration to .netcore6 and the update to OData, the memory was stable; after, we see the same application using massive amounts of memory. Analysis of memory dumps taken during the testing points to OData as the culprit as identified in the analysis below.

-Overall memory consumption is high:
``` 
--- Usage Summary ---------------- RgnCount ----------- Total Size -------- %ofBusy %ofTotal
Free                                    799     7de6`e6827000 ( 125.902 TB)           98.36%
Other                                   210      200`002e0000 (   2.000 TB)  95.33%    1.56%
<unknown>                              3253       18`7dbe4000 (  97.965 GB)   4.56%    0.07%
Stack                                   577        0`75990000 (   1.837 GB)   0.09%    0.00%
Image                                  3545        0`15b00000 ( 347.000 MB)   0.02%    0.00%
Heap                                    803        0`102f4000 ( 258.953 MB)   0.01%    0.00%
TEB                                     192        0`00180000 (   1.500 MB)   0.00%    0.00%
PEB                                       1        0`00001000 (   4.000 kB)   0.00%    0.00%
--- Type Summary (for busy) ------ RgnCount ----------- Total Size -------- %ofBusy %ofTotal
MEM_MAPPED                              228      200`009a7000 (   2.000 TB)  95.33%    1.56%
MEM_PRIVATE                            4806       19`03092000 ( 100.047 GB)   4.66%    0.08%
MEM_IMAGE                              3545        0`15b00000 ( 347.000 MB)   0.02%    0.00%
--- State Summary ---------------- RgnCount ----------- Total Size -------- %ofBusy %ofTotal
MEM_FREE                                801     7de6`e6ab7000 ( 125.902 TB)           98.36%
MEM_RESERVE                            1204      217`4a0ee000 (   2.091 TB)  99.66%    1.63%
MEM_COMMIT                             7375        1`cf44b000 (   7.239 GB)   0.34%    0.01%
``` 
 
97.965GB of  virtualalloc, out of which, GC is only 6GB
GC Committed Heap Size:    Size: 0x16b85f000 (6,098,907,136) bytes.
 
-Machine-wide CPU is very high:
```
0:000> !threadpool
logStart: 31
logSize: 200
CPU utilization: 100% Worker Thread: Total: 83 Running: 42 Idle: 41 MaxLimit: 32767 MinLimit: 300
Work Request in Queue: 0
--------------------------------------
Number of Timers: 4
--------------------------------------
 
-Most busy threads in the user-mode:
0:000> !runaway
 User Mode Time
  Thread       Time
   38:d0       0 days 0:37:18.765 <=GC
   41:2e80     0 days 0:23:07.671 <=GC
   40:198c     0 days 0:15:45.140 <=GC
   39:2560     0 days 0:15:40.515 <=GC
   58:158c     0 days 0:09:32.250 <=GC
   61:25cc     0 days 0:05:28.218 <=GC
   60:bb4      0 days 0:03:31.187 <=GC
   59:ea0      0 days 0:03:30.031 <=GC
```
-To find out why long-running threads are busy with Garbage Collection, we will need to look
at the managed heap to find out which objects are accumulating faster causing GC to run frequently. As GC runs frequently, which is an expensive cpu-intensive operation, will shoot-up CPU consumption.
 
-Here is the most offender list:
 
 
As seen clearly, Microsoft.OData.* objects related to Entity Framework are flooding the managed heap.
 
Here is a partial output of the gcroot
```
0:000> !gcroot 000002001c910348
Thread 1a54:
    00000040FC51E950 00007FFF32E6EB24 System.Runtime.CompilerServices.ConditionalWeakTable`2+Container[[System.__Canon, System.Private.CoreLib],[System.__Canon, System.Private.CoreLib]].FindEntry(System.__Canon, System.Object ByRef) [/_/src/libraries/System.Private.CoreLib/src/System/Runtime/CompilerServices/ConditionalWeakTable.cs @ 525]
        rsi:
            ->  000001FC1B5354B8 System.Runtime.CompilerServices.ConditionalWeakTable`2+Container[[Microsoft.OData.ODataAnnotatable, Microsoft.OData.Core],[Microsoft.OData.Client.Materialization.ODataItemExtensions+MaterializerPropertyValue, Microsoft.OData.Client]]
            ->  000002001C910348 System.Runtime.CompilerServices.ConditionalWeakTable`2+Entry[[Microsoft.OData.ODataAnnotatable, Microsoft.OData.Core],[Microsoft.OData.Client.Materialization.ODataItemExtensions+MaterializerPropertyValue, Microsoft.OData.Client]][]
```
 
-We also see reachable queue is not cleared. You are explicitly coding destructor/Finalize(), instead Developers should be using Dispose pattern implementing IDisposable interface. If there are no unmanaged objects (that you opened-up explicitly in your code) to clean, then DO NOT implement destructor/Finalize().If there are unmanaged objects to clean, then use the IDisposable (Dispose pattern). It will let the framework clean-up your managed objects while executing your custom code for cleaning/closing any unmanaged object. 

### Assemblies affected

### Reproduce steps

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

habbes:
### Issues

*This pull request addresses #2503 

When OData Client is reading items from the reader, it stores some metadata as "annotations" in a cache using the `SetAnnotation()` extension method. It retrieves this metadata during materialization when converting the OData items into client CLR types. For the cache, it uses the custom `InternalDictionary` static class that is based on [`ConditionalWeakTable`](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.compilerservices.conditionalweaktable-2).

The `ConditionalWeakTable` has the following properties relevant to understand the issue and the solution implemented in this PR:
- Entries are held in the table by weak references (`DependentHandle` to be precise) and are not kept alive simply by being in the `ConditionalWeakTable`. If an entry's key is not referenced from outside the table, it will be eventually garbage collected even if the table is still alive.
- The table uses reference equality to identify keys, and this behavior cannot be overridden (e.g. via custom `GetHashCode`, `Equals` or comparer implementations)
- The `ConditionalWeakTable` is thread-safe. It uses locks to synchronize writes to the table
- The `ConditionalWeakTable` contains an internal `Container` class that actually holds the entries array. This class defines a destructor but is not IDisposable. When the array gets full a new `Container` instance with more capacity will be created and the old one will be eventually garbage collector. The destructor also needs to acquire the lock to clean up property and avoid data corruption.
- The `ConditionalWeakTable` does not shrink in size. It can recycle the space of old entries that have expired or have been removed, over item, it either remains the size or grows larger (by creating containers with larger capacities).

OData Client create static `ConditionalWeakTable<ODataAnnotatable, T>`, a different instance static instance is created for each type `T` that is used. `MaterializerEntry` is used as the value type for `ODataResource` keys, `IEnumerable<ODataResource>` as the value type for `ODataResourceSet` keys, etc.

The fact that they are static means they never get garbage collected and they are shared by all requests. This why it's important the tables are thread-safe. But this also causes efficiency and performance issues: when we have a lot of concurrent requests, we will have concurrent writes to the same `ConditionalWeakTable`, which result in:
- lock contention issues as different threads try to acquire locks in order to add entries
- more concurrent requests, means more cache entries will in-use at the same time, meaning their space cannot be recycled. This increases the chances of the cache the `ConditionalWeakTable` being resized
- Resizing the table not only means we allocate more memory, but the old internal container will get garbage collected, which means it will end up in the finalization queue and make GC more expensive
- Because the Container's destructor needs to acquire a lock, it might be a victim (or cause) of lock contention and clog the finalization process

This explains (at least in part) why the customer reported so much memory use and GC activity (up to 11-12GB heap size). This is not sufficient to explain the discrepancy between heap allocations observed in .NET Core 3.1 (~1-3GB) and .NET 6.0.

### Description

This PR removes the `ConditionalWeakTable`-based static cache and replaces it with a new `MaterializerCache`. The new cache is based on a simple `Dictionary<ODataAnnotatable, object>` and is not static. Instead, it's created for each request and only remains alive for the duration of the request. This has the following benefits over the previous implementation:
- No need for thread synchronization since each request has its own dictionary
- Dictionary size does not grow indefinitely. Since they're short-lived their memory is reclaimed once garbage collected.

The customers tested this implementation in a nightly build and reported that memory use dropped from 11-12GB to about 4.5-6GB (~50% improvement) and remains stable around that level. This is however still higher than what it was on .NET Core 3.1, so there's still more investigations to be carried out.

When implementing this PR I had to make the following considerations:
- Ensure the `MaterializerCache` is functionally equivalent to old cache
- At what point in the request/response pipeline should the cache be created
- How to pass the cache a long wherever its needed

#### How `MaterializerCache` is functionally equivalent to the old cache implementation

The old cache was actually a handful of static caches. For example, in a given request, an `<ODataResource, MaterializerEntry>` pair would be stored in a different `ConditionalWeakTable` from an `<ODataProperty, MaterializedPropertyValue>` pair. So, the `ConditionalWeakTable<ODataAnnotatable, MaterializedPropertyValue>` actually contains `ODataProperty` and values from different requests. But the important thing that makes this work is each odata item only has a single value. An `ODataProperty` instance is only once once as key and exists in only one dictionary, you won't find an `ODataProperty` as key in the `ConditionalWeakTable<ODataAnnotatble, MaterializerEntry>` dictionary. Otherwise, that would lead to incorrect behaviour.

Since each instance is only used as key once, then it's safe to have keys of different types in the same `Dictionary<ODataAnnotatable, object>` dictionary. We know that when we fetch the value for an `ODataResource` it will be a `MaterializerEntry` value, and when fetch an `ODataProperty` it will return a `MaterializedPropertyValue` value since this how OData Client is already using the cache. If these assumptions do not hold, then a different design would be necessary. I have also ensured that the dictionary uses a reference-equality comparer.

#### Where to create the `MaterializerCache`

I noticed that the `ODataMaterializer` and similar classes accept or create an `ODataMaterializerContext` instance. Initially I created the `MaterializerCache` inside the context class and exposed it as a property. This worked fine for `ExecuteAsync()` requests but did not work well for `SaveChangesAsync` requests. The latter calls `HandleOperationResponse` at least twice, leading two different materializer contexts, but expects to find data in the cache that was stored in a previouse context. This caused tests to fail. For this reason, I moved the `MaterializerCache` to `BaseAsyncResult`. For context, `QueryResult` (which extends `BaseAsyncResult`) is created inside `Execute()` or `GetValue()` methods of a `DataServiceRequest` object. So, the lifetime is limited to those methods.  I don't know if there's a way to give it short lifespan.

#### How to pass `MaterializerCache` around

Since the materializer cache is no longer static, I needed to refactor methods that access it to accept is an argument. For most methods, instead of passing the materializer cache directly as method, I passed the materializer context instead. This appeared to more convenient and more robust to changes in the future since we can add stuff to the context instead of adding more granular parameters. It also avoids the inconsistency of having to figure when to pass materializer context and when to pass something else. But on the other hand, it has coupled some methods to the materializer context even though they only need to access the cache. And once, in some cases it has forced me to create a new instance of materializer context where it did not previous exist so that I can pass it around. If people believe that this is not a good approach, I'm open to refactoring it to reduce places where I use the materializer context as parameter (it still makes sense to keep the materializer cache as a property of the materializer context in classes that already had dependency on the materializer context).

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Client/Materialization/EnumValueMaterializationPolicy.cs
```diff
@@ -40,9 +40,9 @@ public object MaterializeEnumTypeProperty(Type valueType, ODataProperty property
             object materializedValue = null;
             ODataEnumValue value = property.Value as ODataEnumValue;
             this.MaterializeODataEnumValue(valueType, value.TypeName, value.Value, () => "TODO: Is this reachable?", out materializedValue);
-            if (!property.HasMaterializedValue())
+            if (!property.HasMaterializedValue(this.context))
```
gathogojr, line 43: I wish the field was named `materializationContext` just like in other materialization policy classes. To disambiguate between `DataServiceContext` and `ODataMaterializationContext`

src/Microsoft.OData.Client/Materialization/MaterializerAnnotationsCache.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// The cache is designed to store one entry type per key. For example, if the key is
+    /// an <see cref="ODataResource"/>, its value should be a <see cref="MaterializerEntry"/>. It also
+    /// expects the value for any given key will not change.
+    /// So you should call `cache.SetAnnotation(odataResource, materializerEntry)` add the cache entry
+    /// and `cache.GetAnnotations<MaterializerEntry>(odataResource)` to retrieve the entry.
+    /// 
+    /// If the requirements change such that we need to allow updating the value of a key to a different type,
+    /// then the cache should be redesigned.
+    /// </remarks>
+    internal sealed class MaterializerAnnotationsCache
```
gathogojr, line 30: I'd prefer `MaterializerCache` or `ODataMaterializerCache`

src/Microsoft.OData.Client/Materialization/MaterializerAnnotationsCache.cs
```diff
@@ -0,0 +1,66 @@
...
+//---------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Microsoft.OData.Client.Materialization
+{
+    /// <summary>
+    /// A cache used to store temporary materialization metadata
+    /// for deseriliazed <see cref="ODataItem"/>s during materialization of response payloads
```
gathogojr, line 14:
```suggestion
    /// for deserialiazed <see cref="ODataItem"/>s during materialization of response payloads

```

src/Microsoft.OData.Client/Materialization/MaterializerAnnotationsCache.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// into client CLR objects. Keys are identified using reference equality.
+    /// </summary>
+    /// <remarks>
+    /// Instances of the cache are not thread-safe. Each instance is expected to be used within the
+    /// scope of a single request.
+    /// 
+    /// The cache is designed to store one entry type per key. For example, if the key is
+    /// an <see cref="ODataResource"/>, its value should be a <see cref="MaterializerEntry"/>. It also
+    /// expects the value for any given key will not change.
+    /// So you should call `cache.SetAnnotation(odataResource, materializerEntry)` add the cache entry
```
gathogojr, line 24:
```suggestion
    /// So you should call `cache.SetAnnotation(odataResource, materializerEntry)` to add a cache entry
```

src/Microsoft.OData.Client/Materialization/MaterializerAnnotationsCache.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// and `cache.GetAnnotations<MaterializerEntry>(odataResource)` to retrieve the entry.
+    ///
```
gathogojr, line 26:
```suggestion
```

src/Microsoft.OData.Client/Materialization/MaterializerAnnotationsCache.cs
```diff
@@ -0,0 +1,66 @@
...
+    /// If the requirements change such that we need to allow updating the value of a key to a different type,
+    /// then the cache should be redesigned.
+    /// </remarks>
+    internal sealed class MaterializerAnnotationsCache
+    {
+        private readonly Dictionary<ODataAnnotatable, object> cache = new Dictionary<ODataAnnotatable, object>(ReferenceEqualityComparer<ODataAnnotatable>.Instance);
+
+        /// <summary>
+        /// Adds an entry to the cache with the <paramref name="annotatable"/>
+        /// set as key,
```
gathogojr, line 36:
```suggestion
        /// set as key.
```

src/Microsoft.OData.Client/Materialization/MaterializerAnnotationsCache.cs
```diff
@@ -0,0 +1,66 @@
...
+            this.cache.Add(annotatable, value);
+        }
+
+        /// <summary>
+        /// Retrieves the value associated with the specified <paramref name="annotatable"/>.
+        /// Returns null if the cache does not contain an entry with the specifiedy key.
+        /// </summary>
+        /// <typeparam name="T">The expected type of the value.</typeparam>
+        /// <param name="annotatable"></param>
+        /// <returns>The value associated with the specified <paramref name="annotatable"/> if the entry exists, or null otherwise.</returns>
```
gathogojr, line 53:
```suggestion
        /// <returns>The value associated with the specified <paramref name="annotatable"/> if the entry exists; otherwise null.</returns>
```

src/Microsoft.OData.Client/Materialization/MaterializerAnnotationsCache.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <param name="value">The value to associate with the key.</param>
+        public void SetAnnotation<T>(ODataAnnotatable annotatable, T value) where T : class
+        {
+            Debug.Assert(annotatable != null, "annotatable != null");
+            this.cache.Add(annotatable, value);
+        }
+
+        /// <summary>
+        /// Retrieves the value associated with the specified <paramref name="annotatable"/>.
+        /// Returns null if the cache does not contain an entry with the specifiedy key.
```
gathogojr, line 49:
```suggestion
        /// Returns null if the cache does not contain an entry with the specified key.
```

src/Microsoft.OData.Client/DataServiceRequest.cs
```diff
@@ -77,7 +80,8 @@ internal ODataPayloadKind PayloadKind
                 return MaterializeAtom.EmptyResults;
             }
 
-            return new MaterializeAtom(responseInfo, queryComponents, plan, message, expectedPayloadKind);
+
```
ElizabethOkerio, line 83: nit: remove whitespace

src/Microsoft.OData.Client/BaseSaveResult.cs
```diff
@@ -19,6 +19,7 @@ namespace Microsoft.OData.Client
     using System.Text;
     using System.Threading;
     using Microsoft.OData;
+    using Microsoft.OData.Client.Materialization;
```
ElizabethOkerio, line 22: remove unused import. 
```suggestion
    using Microsoft.OData.Client.Materialization;
remove unused import . 
```

src/Microsoft.OData.Client/Materialization/MaterializerAnnotationsCache.cs
ElizabethOkerio:
```suggestion
    /// for **deseriliazed** <see cref="ODataItem"/>s during materialization of response payloads
    
        /// for **deserialized** <see cref="ODataItem"/>s during materialization of response payloads

```

src/Microsoft.OData.Client/Materialization/MaterializerContextExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// <param name="context">The materializer context.</param>
+        /// <param name="annotatable">The item to annotate.</param>
+        /// <param name="value">The annotation value.</param>
+        public static void SetAnnotation<T>(this IODataMaterializerContext context, ODataAnnotatable annotatable, T value) where T : class
+        {
+            context.AnnotationsCache.SetAnnotation(annotatable, value);
+        }
+
+        /// <summary>
+        /// Retrieves the annotation value associated with te specified <paramref name="annotatable"/>.
```
gathogojr, line 25:
```suggestion
        /// Retrieves the value associated with the specified <paramref name="annotatable"/>.

```

src/Microsoft.OData.Client/Materialization/MaterializerContextExtensions.cs
```diff
@@ -0,0 +1,36 @@
...
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData.Client.Materialization
+{
+    internal static class MaterializerContextExtensions
+    {
+        /// <summary>
+        /// Associates the specified annotation <paramref name="value"/> with the specified
```
gathogojr, line 12:
```suggestion
        /// Associates the specified <paramref name="value"/> with the specified

```

src/Microsoft.OData.Client/Materialization/MaterializerAnnotationsCache.cs
ElizabethOkerio, line 14:
```suggestion
    /// for deserialized <see cref="ODataItem"/>s during materialization of response payloads

```

src/Microsoft.OData.Client/Materialization/MaterializerAnnotationsCache.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// </summary>
+        /// <typeparam name="T">The expected type of the value.</typeparam>
+        /// <param name="annotatable"></param>
```
ElizabethOkerio, line 52: add doc string for this param

src/Microsoft.OData.Client/Materialization/MaterializerContextExtensions.cs
ElizabethOkerio, line 25:
```suggestion
        /// Retrieves the annotation value associated with the specified <paramref name="annotatable"/>.

```

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
```diff
@@ -534,11 +534,11 @@ private object GetOrCreateCollectionProperty(object instance, ClientPropertyAnno
             ClientEdmModel edmModel = this.MaterializerContext.Model;
             ClientTypeAnnotation collectionType = edmModel.GetClientTypeAnnotation(edmModel.GetOrCreateEdmType(property.ResourceSetItemType));
 
-            IEnumerable<ODataResource> entries = MaterializerFeed.GetFeed(feed).Entries;
+            IEnumerable<ODataResource> entries = MaterializerFeed.GetFeed(feed, this.MaterializerContext).Entries;
```
gathogojr, line 537: Is `MaterializerFeed.GetFeed(feed, this.MaterializerContext)` equivalent to `this.MaterializerContext.GetAnnotatable<MaterializerFeed>(feed)`?

src/Microsoft.OData.Client/Materialization/EntityTrackingAdapter.cs
```diff
@@ -27,9 +27,10 @@ internal class EntityTrackingAdapter
         /// <param name="mergeOption">The merge option.</param>
         /// <param name="model">The model.</param>
         /// <param name="context">The context.</param>
-        internal EntityTrackingAdapter(EntityTrackerBase entityTracker, MergeOption mergeOption, ClientEdmModel model, DataServiceContext context)
+        /// <param name="materializerContext">The materialization context used to retrieve materialization-related information.</param>
```
ElizabethOkerio, line 30:
```suggestion
        /// <param name="materializerContext">The materialization context used to retrieve materialization related information.</param>

```

src/Microsoft.OData.Client/MaterializeFromAtom.cs
```diff
@@ -110,8 +113,8 @@ internal MaterializeAtom(ResponseInfo responseInfo, IEnumerable<ODataResource> e
             Type implementationType;
             Type materializerType = GetTypeForMaterializer(this.expectingPrimitiveValue, this.elementType, responseInfo.Model, out implementationType);
             QueryComponents qc = new QueryComponents(null, Util.ODataVersionEmpty, elementType, null, null);
-            ODataMaterializerContext context = new ODataMaterializerContext(responseInfo);
-            EntityTrackingAdapter entityTrackingAdapter = new EntityTrackingAdapter(responseInfo.EntityTracker, responseInfo.MergeOption, responseInfo.Model, responseInfo.Context);
+            ODataMaterializerContext context = new ODataMaterializerContext(responseInfo, annotationsCache);
+            EntityTrackingAdapter entityTrackingAdapter = new EntityTrackingAdapter(responseInfo.EntityTracker, responseInfo.MergeOption, responseInfo.Model, responseInfo.Context, context);
```
gathogojr, line 117: We have `responseInfo.Context` where `Context` is `DataServiceContext`. Let's help the next person looking at this code by using `materializerContext` instead of `context`

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/Materialization/EntryMaterializationPolicyForComplexResourceTests.cs
```diff
@@ -82,7 +82,8 @@ public void ApplyStringValueForCollectionPropertyShouldError()
         [Fact]
         public void MaterializeDerivedComplexForBaseComplexTypeProperty()
         {
-            TestMaterializerContext context = new TestMaterializerContext();
+            var annotationsCache = new MaterializerAnnotationsCache();
+            TestMaterializerContext context = new TestMaterializerContext(annotationsCache);
```
gathogojr, line 86:
```suggestion
            TestMaterializerContext materializerContext = new TestMaterializerContext(annotationsCache);
```

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/Materialization/EntryMaterializationPolicyForComplexResourceTests.cs
```diff
@@ -193,25 +194,25 @@ public void NullValueShouldBeAppliedToSubComplexValueProperty()
 
         private void ApplyInnerProperty(ODataResource innerResource, ComplexTypeWithChildComplexType parentInstance, TestMaterializerContext context = null)
```
gathogojr, line 195:
```suggestion
        private void ApplyInnerProperty(ODataResource innerResource, ComplexTypeWithChildComplexType parentInstance, TestMaterializerContext materializerContext = null)
```

corranrogue9 (approved):
I'm approving due to the urgent nature of this change, and because I find the design changes to be a significant improvement over what we currently have. I have not reviewed in detail.

mikepizzo:
I haven't had time to do a line-by-line review, but I agree with the intent and design of this solution.  Thanks, Clement, for the research and for putting a solution together so quickly, and thanks everyone who has jumped on to provide feedback and help turn this around.

mikepizzo (dismissed):
:shipit:

src/Microsoft.OData.Client/Materialization/EnumValueMaterializationPolicy.cs
habbes: Yeah, I do agree that does it does lead to some confusion, especially since in some places `context` refers to `DataServiceContext`. I can rename the field. I avoided out of fear of adding more "noise" to the PR.

src/Microsoft.OData.Client/Materialization/MaterializerAnnotationsCache.cs
habbes: I don't have an objection to the name, but I feel like this cache is used for a very specific and limited use-case, as opposed to a general purpose cache.

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
habbes: Yes, it is. The method is kind of superfluous come to think of it. But I think it can be refactored later to avoid unnecessarily passing the context around too much.

src/Microsoft.OData.Client/Materialization/MaterializerAnnotationsCache.cs
habbes: I think you meant:
```suggestion
    /// for deserialized <see cref="ODataItem"/>s during materialization of response payloads

```

src/Microsoft.OData.Client/Materialization/MaterializerAnnotationsCache.cs
habbes: I have renamed it to `MaterializerCache` and renamed the parameters accordingly.

gathogojr (approved):
:shipit:

ElizabethOkerio (approved):

gathogojr (approved):
:shipit:

KenitoInc (approved):

src/Microsoft.OData.Client/ProjectionPlanCompiler.cs
```diff
@@ -59,12 +64,14 @@ internal class ProjectionPlanCompiler : ALinqExpressionVisitor
         /// Initializes a new <see cref="ProjectionPlanCompiler"/> instance.
         /// </summary>
         /// <param name="normalizerRewrites">Rewrites introduces by normalizer.</param>
-        private ProjectionPlanCompiler(Dictionary<Expression, Expression> normalizerRewrites)
+        /// <param name="materializerContext">The materializer context.</param>
+        private ProjectionPlanCompiler(Dictionary<Expression, Expression> normalizerRewrites, IODataMaterializerContext materializerContext)
         {
```
corranrogue9, line 69: `Debug.Assert(materializerContext != null, "materializerContext != null");`

src/Microsoft.OData.Client/QueryResult.cs
```diff
@@ -707,7 +707,8 @@ private MaterializeAtom CreateMaterializer(ProjectionPlan plan, ODataPayloadKind
                 plan,
                 this.ContentType,
                 responseMessageWrapper,
-                payloadKind);
+                payloadKind,
+                this.materializerCache);
```
corranrogue9, line 711: `base.materializerCache`

src/Microsoft.OData.Client/SaveResult.cs
```diff
@@ -360,7 +360,7 @@ protected override MaterializeAtom GetMaterializer(EntityDescriptor entityDescri
         {
             Debug.Assert(this.cachedResponse.Exception == null && this.cachedResponse.MaterializerEntry != null, "this.cachedResponse.Exception == null && this.cachedResponse.Entry != null");
             ODataResource entry = this.cachedResponse.MaterializerEntry == null ? null : this.cachedResponse.MaterializerEntry.Entry;
-            return new MaterializeAtom(responseInfo, new[] { entry }, entityDescriptor.Entity.GetType(), this.cachedResponse.MaterializerEntry.Format);
+            return new MaterializeAtom(responseInfo, new[] { entry }, entityDescriptor.Entity.GetType(), this.cachedResponse.MaterializerEntry.Format, this.materializerCache);
```
corranrogue9, line 363: `base.materializerCache`

src/Microsoft.OData.Client/SaveResult.cs
```diff
@@ -865,7 +865,8 @@ private void HandleOperationResponseData(IODataResponseMessage responseMsg, Stre
                             responseMsg.StatusCode,
                             () => responseStream);
 
-                        entry = ODataReaderEntityMaterializer.ParseSingleEntityPayload(responseMessageWrapper, responseInfo, entityDescriptor.Entity.GetType());
+                        ODataMaterializerContext materializerContext = new ODataMaterializerContext(responseInfo, this.materializerCache);
```
corranrogue9, line 868: `base.materializerCache`

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/Materialization/CollectionValueMaterializationPolicyTests.cs
```diff
@@ -141,7 +141,7 @@ public void DataServicCollectionOfTAsCollectionTypeShouldFailForPrimitiveOrCompl
         [Fact]
         public void CreateCollectionInstanceShouldFailOnTypeWithNoParametersLessConstructors()
         {
-            var testContext = new TestMaterializerContext();
+            var testContext = new TestMaterializerContext(new MaterializerCache());
```
corranrogue9, line 144: Would it be easier to update the default constructor for `TestMaterializerContext` so that it uses a new `MaterializerCache`?

habbes: Yes it would be easier. I did think of that, but only after writing the tests and realizing just how many changes I made as a result. I can refactor it again.

src/Microsoft.OData.Client/BaseAsyncResult.cs
```diff
@@ -26,6 +27,11 @@ internal abstract class BaseAsyncResult : IAsyncResult
         /// <summary>wrapped request</summary>
         protected PerRequest perRequest;
 
+        /// <summary>
+        /// Cache used to store temporary metadata used for materialization of OData items.
+        /// </summary>
+        protected MaterializerCache materializerCache = new MaterializerCache();
```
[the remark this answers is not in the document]
habbes: I think it's a good idea, but since this is an internal class, I think it would be better to instantiate the cache until there's a need to make it configurable. This way, we think of it as an implementation detail of the `BaseAsyncResult` that the caller does not need to know about. If we pass it on the constructor, the caller will have to take responsibility for passing down the cache, and we might make the argument that it should be configurable at that level, all the way up to the DataServiceContext. But the buck has to stop somewhere since this isn't a user input. I think `BaseAsyncResult` is good place because it's what glues the request together, and this cache is scoped to a request.

PR #2519, opened 2022-10-02
Add matchesPattern support

### LINKED ISSUE #1901: An unknown function with name 'matchesPattern' was found
I'm using Microsoft.AspNet.OData.

I'm trying to use $filter=matchesPattern(Property, '%pattern%') but I'm receiving Exception: Microsoft.OData.ODataException: 'An unknown function with name 'matchesPattern' was found. This may also be a function import or a key lookup on a navigation property, which is not allowed.'

Is 'matchesPattern' function not supported?

It is a [specification ](http://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_matchesPattern)of OData v4 protocol

Arithmomaniac:
### Issues

This pull request fixes #1901.

### Description

Adds support for the `matchesPattern` function introduced in [OData v4](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_matchesPattern). In addition to being added as a supported function to the raw parser, it is mapped in .NET to `Regex.IsMatch(input, pattern, RegexOptions.ECMAScript)`.

The basic methodology was to look for any C# file in the repository that handled _all_ of the string predicates (`startswith`, `endswith`, etc.) and added `matchesPattern` to that set. There was at least one legacy provider with no support for e.g. `contains`; that provider was unmodified.

As part of this work, the case-insensitivity logic had to be changed to account for `matchesPattern` being the first OData function whose canonical name is mixed-case.

### Checklist

- [X] *Test cases added*
    - Almost of the tests passed locally
    - Some unit tests failed intermittently from the build script but worked in VS (different tests each time)
    - Astoria complained about `Unable to cast COM object of type 'AstoriaUnitTests.Tests.JScriptEngine'...`)


test/EndToEndTests/Tests/Client/Build.Desktop/LegacyReferences/TrippinServiceReference.cs
```diff
@@ -490,6 +490,7 @@ private abstract class GeneratedEdmModel
         <Annotation Term=""Org.OData.Capabilities.V1.FilterFunctions"">
           <Collection>
             <String>contains</String>
+			<String>matchesPattern</String>
```
mikepizzo, line 493: nit: looks like extra whitespace before the string in some editors (maybe a tab instead of space characters)

mikepizzo (dismissed):
:shipit:

PR #2521, opened 2022-10-05
shouldn't require id to write a deleted resource

### LINKED ISSUE #2482: Shouldn't require id in a bulk operation response for a failed create
Today, ODataDeletedResource (used to represent a deleted resource in a delta response) requires an id. 

However, ODataDeletedResource is also used in a delta response to a bulk operation to represent a successfully deleted entity (which is fine) or for a failed create/insert. In the case of a failed create/insert, the id should not be required if the server generates the key. 

### Assemblies affected

ODL 7.x

### Reproduce steps

PATCH to an entity collection to add new instances with server generated keys.  i.e., 

PATCH /users

```json
{
"value": [
     {
            "@contentId": 1,
            "firstName":"ted",
            "lastName":"lasso"
      }
    ]
}
```
If the insert fails, the response should look like:

```json
{
    "value": [
        {
            "@removed":{"reason":"deleted"},
            "@Core.ContentId": "1",
            "@Core.ModificationException": {
                "failedOperation": "insert",
                "responseCode": 409
            }
        }
    ]
}
```

### Expected result

You could write the above valid payload without an error.

### Actual result

When attempting to write an ODataDeletedResource with no id, you receive the following OData exception:

 "No Id or key properties were found. A resource in a delta response requires an ID or key properties be specified." 

Note: use of "@Core.ContentId" correlates entities in the request with entities in the response. Perhaps, at least in the absence of @Core.ContentId, for server-generated keys, we should return all of the properties of the entity that failed to be inserted.

ElizabethOkerio:
### Issues

*This pull request fixes #2482 .*

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
@@ -1431,17 +1431,29 @@ namespace Microsoft.OData.Tests.JsonLight
                 );
         }
 
-        [InlineData(/*isResponse*/true)]
-        [InlineData(/*isResponse*/false)]
-        [Theory]
-        public void WriteDeletedEntryWithoutKeyOrIdShouldFail(bool isResponse)
+        [Fact]
+        public void WriteDeletedEntryWithoutKeyOrIdShouldNotFailWhenWritingAResponse()
+        {
+            this.TestInit(this.GetModel());
+            Action writeAction = () =>
+            {
+                ODataJsonLightWriter writer = new ODataJsonLightWriter(GetV401OutputContext(true), this.GetCustomers(), this.GetCustomerType(), true, false, true);
+                writer.WriteStart(true ? feed : requestFeed);
+                writer.WriteStart(new ODataDeletedResource());
+            };
+
+            writeAction.DoesNotThrow();
+        }
+
+        [Fact]
+        public void WriteDeletedEntryWithoutKeyOrIdShouldFailWhenWritingARequest()
         {
             this.TestInit(this.GetModel());
 
             Action writeAction = () =>
             {
-                ODataJsonLightWriter writer = new ODataJsonLightWriter(GetV401OutputContext(isResponse), this.GetCustomers(), this.GetCustomerType(), true, false, true);
-                writer.WriteStart(isResponse ? feed : requestFeed);
+                ODataJsonLightWriter writer = new ODataJsonLightWriter(GetV401OutputContext(false), this.GetCustomers(), this.GetCustomerType(), true, false, true);
+                writer.WriteStart(false ? feed : requestFeed);
                 writer.WriteStart(new ODataDeletedResource());
             };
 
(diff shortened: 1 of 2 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
```diff
@@ -1434,7 +1434,7 @@ public void WriteContentIn41DeletedEntry(bool isResponse)
         [InlineData(/*isResponse*/true)]
         [InlineData(/*isResponse*/false)]
         [Theory]
-        public void WriteDeletedEntryWithoutKeyOrIdShouldFail(bool isResponse)
+        public void WriteDeletedEntryWithoutKeyOrIdShouldNotFail(bool isResponse)
```
KenitoInc, line 1437: Won't this be a breaking change?

From the issue, have you considered this?
`May be able to leverage transient entities (i.e., requiring setting IsTransient=true when writing the resource).`

xuzhg (dismissed):

gathogojr (approved):
LGTM

KenitoInc (approved):
LGTM

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
```diff
@@ -1431,17 +1431,29 @@ public void WriteContentIn41DeletedEntry(bool isResponse)
...
+        public void WriteDeletedEntryWithoutKeyOrIdShouldFailWhenWritingARequest()
         {
             this.TestInit(this.GetModel());
 
             Action writeAction = () =>
             {
-                ODataJsonLightWriter writer = new ODataJsonLightWriter(GetV401OutputContext(isResponse), this.GetCustomers(), this.GetCustomerType(), true, false, true);
-                writer.WriteStart(isResponse ? feed : requestFeed);
+                ODataJsonLightWriter writer = new ODataJsonLightWriter(GetV401OutputContext(false), this.GetCustomers(), this.GetCustomerType(), true, false, true);
+                writer.WriteStart(false ? feed : requestFeed);
```
odero, line 1456: @ElizabethOkerio wont this always be false?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
```diff
@@ -1431,17 +1431,29 @@ public void WriteContentIn41DeletedEntry(bool isResponse)
...
-        [Theory]
-        public void WriteDeletedEntryWithoutKeyOrIdShouldFail(bool isResponse)
+        [Fact]
+        public void WriteDeletedEntryWithoutKeyOrIdShouldNotFailWhenWritingAResponse()
+        {
+            this.TestInit(this.GetModel());
+            Action writeAction = () =>
+            {
+                ODataJsonLightWriter writer = new ODataJsonLightWriter(GetV401OutputContext(true), this.GetCustomers(), this.GetCustomerType(), true, false, true);
+                writer.WriteStart(true ? feed : requestFeed);
```
odero, line 1441: Wont this always be true?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightDeltaWriterTests.cs
ElizabethOkerio: yes. I think what needed to be done here was:  `writer.WriteStart(requestFeed)` or writer.WriteStart(feed).

PR #2523, opened 2022-10-12
Fix issue where CsdlWriter TryWriteCsdl does not trigger flushing of XmlWriter buffer

### LINKED ISSUE #2522: CsdlWriter.TryWriteCsdl does not flush xml writer
CsdlWriter.TryWriteCsdl does not flush xml writer. This may be by design but it caused some confusion recently and want to verify the expected functionality. If it is by design then perhaps the xml comment should be updated to reflect this.

### Assemblies affected

Microsoft.OData.Edm.dll 7.12.3

### Reproduce steps

```csharp
        var sb = new StringBuilder();
        using var xw = XmlWriter.Create(sb, new XmlWriterSettings
        {
            Encoding = Encoding.UTF8,
            Indent = true,
            OmitXmlDeclaration = true,
            ConformanceLevel = ConformanceLevel.Fragment
        });
        
        var success = CsdlWriter.TryWriteCsdl(model, xw, CsdlTarget.OData, out var errors);

        if (!success)
        {
            return null;
        }

        xw.Flush(); //this is required and seems like it really should not be

        return sb.ToString();
```

### Expected result

The TryWriteCsdl method would flush the XmlWriter

### Actual result

TryWriteCsdl is not flushing the XmlWriter

gathogojr:
### Issues

*This pull request fixes #2522.*

### Description

Ensure `CsdlWriter` `TryWriteCsdl` method triggers flushing of `XmlWriter` buffer before exiting

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
@@ -9,6 +9,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.IO;
 using System.Linq;
+using System.Text;
 #if NETCOREAPP3_1
 using System.Text.Encodings.Web;
 using System.Text.Json;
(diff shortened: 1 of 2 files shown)


KenitoInc (approved):

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
```diff
@@ -2812,5 +2813,46 @@ public void ShouldSubstituteFullyQualifiedNamespaceWithAliasIfAliasIsSet()
...
+            customerEntityType.AddKeys(key);
+            customerEntityType.AddStructuralProperty("Name", EdmPrimitiveTypeKind.String);
+            model.AddElement(customerEntityType);
+
+            var builder = new StringBuilder();
+            using (var writer = XmlWriter.Create(builder, new XmlWriterSettings { Encoding = Encoding.UTF8 }))
+            {
+                if (!CsdlWriter.TryWriteCsdl(model, writer, csdlTarget, out var errors))
+                {
+                    Assert.True(false, "Serialization was unsuccessful");
```
habbes, line 2835: Wish there was a clearer way to "assert" failure :)

habbes (approved):

habbes:
I've approved the PR because I doesn't hurt to flush in this case, and might improve user experience, but I am not convincing that what it's "fixing" is an actual issue (see my comments on the issue: https://github.com/OData/odata.net/issues/2522#issuecomment-1278473556)

xuzhg (approved):

PR #2527, opened 2022-10-14
Fix bugs causing separator to be omitted after ODataUtf8JsonWriter.WriteRawValue 

### LINKED ISSUE #2525: ODataUtf8JsonWriter does not emit item separator after untyped values
The `ODataUtf8JsonWriter` does not write the `,` separator after the writing an untyped value, which leads to invalid JSON when an untyped value appears in the middle of an object or array.

### Assemblies affected

Microsoft.OData.Core 7.12.3

### Reproduce steps

Open the SerializedComparisonsTests solution in the [`test/PerformanceTests/SerializationComparisonTests`](https://github.com/OData/odata.net/tree/master/test/PerformanceTests/SerializationComparisonsTests) directory. Then edit the model in [`Lib/DataModel.cs`](https://github.com/OData/odata.net/blob/master/test/PerformanceTests/SerializationComparisonsTests/Lib/DataModel.cs) file to make the `Customer` entity type an open type:

```diff
- EdmEntityType customer = new EdmEntityType("NS", "Customer");
+ EdmEntityType customer = new EdmEntityType("NS", "Customer", baseType: null, isAbstract: false, isOpen: true);
```

Then add some untyped values to the customer `ODataResource` in [`ODataMessageWriterPayloadWriter`](https://github.com/OData/odata.net/blob/master/test/PerformanceTests/SerializationComparisonsTests/Lib/ODataMessageWriterPayloadWriter.cs#L63) class:

```diff
var customerResouce = new ODataResource
{
    new ODataProperty
    {
        Name = "Id",
        Value = customer.Id
    },
    new ODataProperty { Name = "Name", Value = customer.Name },
    new ODataProperty
    {
        Name = "Emails",
        Value = new ODataCollectionValue
        {
            Items = customer.Emails,
            TypeName = "Collection(Edm.String)"
        }
+    },
+   new ODataProperty
+    {
+       Name = "DynamicArrayProp",
+        Value = new ODataUntypedValue
+       {
+            RawValue = "[10, 20, 30, 40, 50]"
+       }
+    },
+   new ODataProperty
+    {
+       Name = "DynamicObjectProp",
+        Value = new ODataUntypedValue
+       {
+            RawValue = "{\"foo\":\"bar\"}"
+       }
+    },
}
```

Then run the `TestServer` project and execute the following requests:

`GET http://localhost:7120/customers/ODataMessageWriter` (will use the default JsonWriter)

`GET http://localhost:7120/customers/ODataMessageWriter-Utf8JsonWriter` (will use `ODataUtf8JsonWriter`)

### Expected result

Both requests should return valid JSON responses, which should look fairly the same apart from potential escaping differences, something like:

image

### Actual result

The request using `ODataUtf8JsonWriter` returns invalid JSON (I've added a new line where the commas are missing for better legibility):

```
{"@odata.context":"https://services.odata.org/V4/OData/OData.svc/$metadata#Customers","value":[{"Id":1,"Name":"Cust1 \uD800\uDC05 \u00E4","Emails":["emailA@mailer.com1","emailB@mailer.com1"],"DynamicArrayProp":[10, 20, 30, 40, 50]"
DynamicObjectProp":{"foo":"bar"}"
CollectionOfUntypedProp@odata.type":"#Collection(Untyped)","CollectionOfUntypedProp":[{"foo":"bar"}
[1, 2, 3, 4]
"test"
"what"],
```

### Additional detail

This is due to the fact that in the current implementation of `ODataUtf8JsonWriter.WriteRawValue(string)` bypasses the internal `Utf8JsonWriter` and writes to the stream directly. Since the internal `Utf8JsonWriter` is not aware that a value was written, it does not add a separator before writing the next item.

habbes:
### Issues

*This pull request fixes #2525 

### Description

This PR fixes a bug in `ODataUtf8JsonWriter.WriteRawValue` that leads to invalid JSON being written. One common use case for `WriteRawValue` is to write the value of `ODataUntypedValue` objects, which could be anything.

`ODataUtf8JsonWriter.WriteRawValue` was implemented by writing the input directly to the output stream, bypassing the internal `Utf8JsonWriter`. This is because `Utf8JsonWriter.WriteRawValue` method does not exist in .NET Core 3.1. We intend to use `Utf8JsonWriter` in ODL v8 when we drop support for older frameworks.

`Utf8JsonWriter` keeps track of whether or not it's writing the first item in an object or an array. It knows that it needs to place a separator (comma) between the end of one key-value pair and the beginning of another, it also knows that it needs to place separators between array elements.

Here's an example of a normal flow:

- `ODataUtf8JsonWriter.WriteName("SomeKey")` -> calls `Utf8JsonWriter.WritePropertyName("SomeKey")`
- `ODataUtf8JsonWriter.WriteValue("SomeVal")` -> calls `Utf8JsonWriter.WriteStringValue("SomeVal")`. `Utf8JsonWriter` sets a flag signaling that the next property write should place a comma before the property name
- `ODataUtf8JsonWriter.WriteName("AnotherKey")` -> calls `Utf8JsonWriter.WritePropertyName("AnotherKey")`. `Utf8JsonWriter` writes a comma before the property name
output excerpt: `"SomeKey":"SomeVal","AnotherKey":`

When you bypass the `Utf8JsonWriter` and write directly to the output stream, the `Utf8JsonWriter` is not aware of this write. So, the `Utf8JsonWriter` will behave as if that raw value was never written. This may lead to missing or repeated separators, both of which lead to invalid JSON being written to the output.

Here's an example of an invalid JSON from using the raw value:

- `ODataUtf8JsonWriter.WriteName("SomeKey")` -> calls `Utf8JsonWriter.WritePropertyName("SomeKey")`
- `ODataUtf8JsonWriter.WriteRawValue("\"SomeRawVal\"")` -> calls `writeStream.Write("\"SomeRawVal\"")`. `Utf8JsonWriter is not aware of this and does not set a flag to write a separator before the next property`.
- `ODataUtf8JsonWriter.WriteName("AnotherKey") -> calls `Utf8JsonWriter.WritePropertyName("AnotherKey"). 
output excerpt: `"SomeKey":"SomeVal""AnotherKey":` (comma missing between `"SomeRawVal"` and `"AnotherKey"`)

Similarly, `Utf8JsonWriter` automatically places a separator before each element of an array except for the first element. However, when raw values are involved, the following bugs surface:
- The separator is not placed before raw values, because `Ut8fJsonWriter` is not aware of them
- When a raw value is the first item of an array, then followed by a non-raw value (written by `Utf8JsonWriter`), not comma will be placed before this non-raw value because `Utf8JsonWriter` mistakenly think it's the first item of the array.

To fix these issues, this PR writes a separator manually in the following scenarios:
- Inside an object, before writing a property name that follows a raw value
- Inside an array, before writing a raw value, unless the raw value is the first element in the array
- Inside an array, before writing a value that follows a sequence of one or more consecutive raw values that start the array e.g. `["raw1", "raw2", "raw3", "raw3", "non-raw value"]` (from the perspective of `Utf8JsonWriter`, `"non-raw value"` is the first item in the array, and so it won't automatically place a comma before it).

This PR achieves this by adding a couple of variables to keep track of when we've written raw values as well as a stack to keep track of whether we're in an array or object. The stack is based on [`BitStack`](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/BitStack.cs) implemented I "borrowed" from .NET runtime (which is used in the `Utf8JsonWriter` implementation). I slightly adapted because it used some syntax not supported in the C# version used by ODL and also because its `Pop()` method does not return the most recently pushed item on the stack. Not sure whether this is a bug or not. But I also pulled [the tests](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/tests/System.Text.Json.Tests/BitStackTests.cs) and adjusted them accordingly. The `BitStack` is an efficient stack implementation optimized for stacks where there are only two types of values (represented as true and false). It stores the values in a sequence of integers, where a value in the stack is stored in a bit in one of the ints.

The first iteration of this PR caused a performance regression when the calls to `ODataUtf8JsonWriter.WriteRawValue()` are made. This was mainly due to frequent flushing (I flushed the writer before writing raw values to the stream, to ensure correct order of writes). I fixed the perf issue by writing to the same `ArrayBufferWriter<byte>` that is passed to `Utf8JsonWriter`, this means there's no need to flush before the buffer is full. This improved performance. I also updated the benchmarks to include to optionally include raw values.

- [x] *Test cases added*

### Additional work necessary


habbes:
I've added to raw values to the benchmarks and noticed a considerable perf degradation. After I added raw values to the written payload, `ODataMessageWriter-ODataUtf8JsonWriter` seemed to get slower than `ODataMessageWriter`. Memory allocations have also increased. The memory increased memory allocation in `NoOpJsonWriter-Direct` (from 18KB to 5K`) is very baffling. I do expect things to be slighly slower and allocate slightly more memory cause we're writing more data, but not by the observed extent. And I'd still expected `ODataUtf8JsonWriter`-based writer to be faster than their `JsonWriter` counterparts.

Benchmarks with PR changes, but before writing raw values:

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.20348
Intel Xeon E-2336 CPU 2.90GHz, 1 CPU, 12 logical and 6 physical cores
.NET SDK=        5.0.404 [C:\Program Files\dotnet\sdk]
  [Host] : .NET 6.0.10 (6.0.1022.47605), X64 RyuJIT

Toolchain=InProcessEmitToolchain  InvocationCount=1  UnrollFactor=1

|           Method |                                           WriterName |       Mean |     Error |    StdDev |     Median |      Gen 0 |     Gen 1 |  Allocated |
----------------- |----------------------------------------------------- |-----------:|----------:|----------:|-----------:|-----------:|----------:|-----------:|
| WriteToFileAsync |                                       JsonSerializer |  33.444 ms | 0.6622 ms | 0.8132 ms |  32.950 ms |          - |         - |      12 KB |
 WriteToFileAsync |                                    NoOpWriter-Direct |   1.317 ms | 0.0133 ms | 0.0118 ms |   1.313 ms |          - |         - |       8 KB |
 WriteToFileAsync |                               ODataJsonWriter-Direct |  35.745 ms | 0.3827 ms | 0.3579 ms |  35.850 ms |  1000.0000 |         - |   6,675 KB |
 WriteToFileAsync |                         ODataJsonWriter-Direct-Async | 179.549 ms | 0.5134 ms | 0.4551 ms | 179.462 ms |  8000.0000 |         - |  48,550 KB |
 WriteToFileAsync |                                   ODataMessageWriter | 280.770 ms | 0.7588 ms | 0.7098 ms | 280.512 ms | 36000.0000 |         - | 224,024 KB |
 WriteToFileAsync |                             ODataMessageWriter-Async | 834.186 ms | 2.3173 ms | 2.0542 ms | 833.958 ms | 54000.0000 | 1000.0000 | 334,997 KB |
 WriteToFileAsync |                              ODataMessageWriter-NoOp | 226.985 ms | 1.1005 ms | 1.0294 ms | 227.011 ms | 35000.0000 |         - | 217,365 KB |
 WriteToFileAsync |                        ODataMessageWriter-NoOp-Async | 335.748 ms | 0.7169 ms | 0.6705 ms | 335.673 ms | 35000.0000 |         - | 217,368 KB |
 WriteToFileAsync |                      ODataMessageWriter-NoValidation | 247.049 ms | 0.8430 ms | 0.7885 ms | 246.715 ms | 32000.0000 |         - | 200,583 KB |
 WriteToFileAsync |                ODataMessageWriter-NoValidation-Async | 794.218 ms | 1.2409 ms | 1.1607 ms | 794.490 ms | 51000.0000 | 1000.0000 | 311,358 KB |
 WriteToFileAsync |                             ODataMessageWriter-Utf16 | 292.286 ms | 0.6426 ms | 0.6011 ms | 292.312 ms | 36000.0000 |         - | 224,018 KB |
 WriteToFileAsync |                    ODataMessageWriter-Utf8JsonWriter | 268.767 ms | 0.5205 ms | 0.4614 ms | 268.758 ms | 35000.0000 | 1000.0000 | 217,599 KB |
 WriteToFileAsync |              ODataMessageWriter-Utf8JsonWriter-Async | 438.272 ms | 0.4929 ms | 0.4611 ms | 438.149 ms | 35000.0000 | 1000.0000 | 219,011 KB |
 WriteToFileAsync |       ODataMessageWriter-Utf8JsonWriter-NoValidation | 232.884 ms | 0.3154 ms | 0.2950 ms | 232.763 ms | 31000.0000 | 1000.0000 | 194,160 KB |
 WriteToFileAsync | ODataMessageWriter-Utf8JsonWriter-NoValidation-Async | 392.491 ms | 0.5989 ms | 0.5309 ms | 392.433 ms | 31000.0000 | 1000.0000 | 195,572 KB |
 WriteToFileAsync |              ODataMessageWriter-Utf8JsonWriter-Utf16 | 272.817 ms | 0.5228 ms | 0.4890 ms | 272.629 ms | 35000.0000 | 1000.0000 | 217,600 KB |
 WriteToFileAsync |                           ODataUtf8JsonWriter-Direct |  21.425 ms | 0.0402 ms | 0.0376 ms |  21.438 ms |          - |         - |     241 KB |
 WriteToFileAsync |                     ODataUtf8JsonWriter-Direct-Async |  43.708 ms | 0.3825 ms | 0.3391 ms |  43.687 ms |          - |         - |     520 KB |
 WriteToFileAsync |         Utf8JsonWriter-Direct-ArrayPool-NoValidation |  18.465 ms | 0.0574 ms | 0.0537 ms |  18.482 ms |          - |         - |      29 KB |


**Benchmarks results when including raw values in payload**:

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.20348
Intel Xeon E-2336 CPU 2.90GHz, 1 CPU, 12 logical and 6 physical cores
.NET SDK=        5.0.404 [C:\Program Files\dotnet\sdk]
  [Host] : .NET 6.0.10 (6.0.1022.47605), X64 RyuJIT

Toolchain=InProcessEmitToolchain  InvocationCount=1  UnrollFactor=1

|           Method |                                           WriterName |         Mean |     Error |    StdDev |      Gen 0 |     Gen 1 |  Allocated |
----------------- |----------------------------------------------------- |-------------:|----------:|----------:|-----------:|----------:|-----------:|
 WriteToFileAsync |                                       JsonSerializer |    39.725 ms | 0.2850 ms | 0.2526 ms |          - |         - |      16 KB |
 WriteToFileAsync |                                    NoOpWriter-Direct |     3.808 ms | 0.0745 ms | 0.0697 ms |          - |         - |   5,003 KB |
 WriteToFileAsync |                               ODataJsonWriter-Direct |    43.317 ms | 0.8210 ms | 0.7680 ms |  1000.0000 |         - |  11,675 KB |
 WriteToFileAsync |                         ODataJsonWriter-Direct-Async |   205.117 ms | 0.5807 ms | 0.4849 ms |  8000.0000 |         - |  53,798 KB |
 WriteToFileAsync |                                   ODataMessageWriter |   298.201 ms | 0.3155 ms | 0.2797 ms | 39000.0000 |         - | 239,573 KB |
 WriteToFileAsync |                             ODataMessageWriter-Async |   866.389 ms | 1.6864 ms | 1.4082 ms | 57000.0000 | 1000.0000 | 350,650 KB |
 WriteToFileAsync |                              ODataMessageWriter-NoOp |   231.295 ms | 0.4075 ms | 0.3812 ms | 38000.0000 |         - | 232,914 KB |
 WriteToFileAsync |                        ODataMessageWriter-NoOp-Async |   354.659 ms | 0.9602 ms | 0.8982 ms | 38000.0000 |         - | 232,917 KB |
 WriteToFileAsync |                      ODataMessageWriter-NoValidation |   264.587 ms | 0.6677 ms | 0.6246 ms | 35000.0000 |         - | 216,132 KB |
 WriteToFileAsync |                ODataMessageWriter-NoValidation-Async |   812.586 ms | 1.9369 ms | 1.8118 ms | 53000.0000 | 1000.0000 | 327,209 KB |
 WriteToFileAsync |                             ODataMessageWriter-Utf16 |   310.894 ms | 0.5446 ms | 0.4828 ms | 39000.0000 |         - | 239,567 KB |
 WriteToFileAsync |                    ODataMessageWriter-Utf8JsonWriter |   422.922 ms | 0.8506 ms | 0.7957 ms | 38000.0000 |         - | 236,455 KB |
 WriteToFileAsync |              ODataMessageWriter-Utf8JsonWriter-Async | 1,295.575 ms | 2.9140 ms | 2.7258 ms | 62000.0000 | 1000.0000 | 378,681 KB |
 WriteToFileAsync |       ODataMessageWriter-Utf8JsonWriter-NoValidation |   383.425 ms | 0.8009 ms | 0.7492 ms | 34000.0000 |         - | 213,016 KB |
 WriteToFileAsync | ODataMessageWriter-Utf8JsonWriter-NoValidation-Async | 1,202.503 ms | 5.7807 ms | 5.4073 ms | 58000.0000 | 1000.0000 | 355,241 KB |
 WriteToFileAsync |              ODataMessageWriter-Utf8JsonWriter-Utf16 |   470.336 ms | 0.5858 ms | 0.4892 ms | 38000.0000 |         - | 238,800 KB |
 WriteToFileAsync |                           ODataUtf8JsonWriter-Direct |   117.280 ms | 0.4300 ms | 0.3812 ms |  1000.0000 |         - |   8,547 KB |
 WriteToFileAsync |                     ODataUtf8JsonWriter-Direct-Async |   353.166 ms | 3.5165 ms | 3.2893 ms |  8000.0000 |         - |  51,325 KB |
 WriteToFileAsync |         Utf8JsonWriter-Direct-ArrayPool-NoValidation |    32.796 ms | 0.0554 ms | 0.0491 ms |          - |         - |   5,034 KB |

habbes:
Based on some high-level profiler analysis, one of the possible explanations of the performance regression could be the fact that when we're writing raw values value, we have to flush the contents of the `Utf8JsonWriter` buffer to the stream to ensure correct order of writes. However, this doesn't sound like a good explanation for the memory regression on `NoOpWriter-Direct`

src/Microsoft.OData.Core/Json/BitStack.cs
```diff
@@ -0,0 +1,176 @@
...
+namespace Microsoft.OData.Json
+{
+    /// <summary>
+    /// An efficient stack optimized for storing data that has two possible kinds of value, represented
+    /// as true and false.
+    /// </summary>
+    /// <remarks>
+    /// This has been adapted from the .NET runtime's internal BitStack which is used by Utf8JsonWriter
+    /// https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/BitStack.cs
+    /// This has been slightly modified because the original Pop() method did not return the last value pushed.
```
ElizabethOkerio, line 17: Just curious. What does the original Pop() method return?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -35,6 +36,48 @@ internal sealed class ODataUtf8JsonWriter : IJsonWriter, IDisposable, IJsonWrite
...
         private bool disposed;
+        /// The Utf8JsonWriter internally keeps track of when to write the item separtor ','
+        /// between key-value pairs in an object and between items in an array
+        /// However, we bypass the Utf8JsonWriter in our implementation of WriteRawValue
+        /// and write directly to the destination stream. This means that we have to manually
+        /// write an item separator when writing raw values. And since the Utf8JsonWriter is not
+        /// aware of this write, we have to manually keep track of state to synchronise with the Utf8JsonWriter
+        /// to ensure we don't write extra separators or skip a separator where it's needed, both of
+        /// which would result in invalid JSON.
+        ///
```
ElizabethOkerio, line 47:
```suggestion

```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -35,6 +36,48 @@ internal sealed class ODataUtf8JsonWriter : IJsonWriter, IDisposable, IJsonWrite
...
+        /// 
+        /// The following variables are used to keep track of state that allows us to figure out where to manually place a separator
+        /// This will not be necessary when we implement WriteRawValue using Utf8JsonWriter.WriteRawValue()
+        /// see: https://github.com/OData/odata.net/issues/2420
+
+        /// <summary>
+        /// Whether we should manually write the item separator (,) before
+        /// the next object or array entry.
+        /// </summary>
+        private bool shouldWriteSeparator = false;
```
ElizabethOkerio, line 63:
```suggestion
        private bool shouldWriteSeparator = false;


```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -35,6 +36,48 @@ internal sealed class ODataUtf8JsonWriter : IJsonWriter, IDisposable, IJsonWrite
...
+        /// <summary>
+        /// Stack used to keep track of scope transitions i.e,
+        /// whether we're currently in an Object or an Array.
+        /// </summary>
+        private BitStack bitStack;
```
ElizabethOkerio, line 68:
```suggestion
        private BitStack bitStack;

```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -35,6 +36,48 @@ internal sealed class ODataUtf8JsonWriter : IJsonWriter, IDisposable, IJsonWrite
...
+        /// This will not be necessary when we implement WriteRawValue using Utf8JsonWriter.WriteRawValue()
+        /// see: https://github.com/OData/odata.net/issues/2420
+
+        /// <summary>
+        /// Whether we should manually write the item separator (,) before
+        /// the next object or array entry.
+        /// </summary>
+        private bool shouldWriteSeparator = false;
+        /// <summary>
+        /// Stack used to keep track of scope transitions i.e,
```
ElizabethOkerio, line 65:
```suggestion
        /// <summary>
        
        /// Stack used to keep track of scope transitions i.e,

```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -35,6 +36,48 @@ internal sealed class ODataUtf8JsonWriter : IJsonWriter, IDisposable, IJsonWrite
...
+        /// The following variables are used to keep track of state that allows us to figure out where to manually place a separator
+        /// This will not be necessary when we implement WriteRawValue using Utf8JsonWriter.WriteRawValue()
+        /// see: https://github.com/OData/odata.net/issues/2420
+
+        /// <summary>
+        /// Whether we should manually write the item separator (,) before
+        /// the next object or array entry.
+        /// </summary>
+        private bool shouldWriteSeparator = false;
+        /// <summary>
```
ElizabethOkerio, line 64:
```suggestion
        private bool shouldWriteSeparator = false;
        
        /// <summary>
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -274,12 +339,167 @@ public void WriteRawValue(string rawValue)
...
             // Consider using Utf8JsonWriter.WriteRawValue() in .NET 6+
             // see: https://github.com/OData/odata.net/issues/2420
-
-            this.Flush(); // ensure we don't write to the stream while there are still pending data in the buffer
             this.writeStream.Write(Encoding.UTF8.GetBytes(rawValue));
-            this.FlushIfBufferThresholdReached();
+
+            // since we bypass the Utf8JsonWriter, we need to signal to other
+            // Write methods that a separator should be written first
+            if ((this.isWritingFirstElementInArray || this.isWritingConsecutiveRawValuesAtStartOfArray) || !this.IsInArray())
```
ElizabethOkerio, line 357: what if we writing the RawValue at the end of array or at the middle of array?

ElizabethOkerio: can we have a case like this? [`NonRawValue`, `RawValue`, `NonRawValue`, `RawValue`]

ElizabethOkerio: I also haven't seen the case of an object scenario. Like where is `this.shouldWriteSeparator = true;` set for object scenario?

src/Microsoft.OData.Core/Json/BitStack.cs
habbes: It seems to return the value of the item before the most recent item, or something like that.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -274,12 +339,167 @@ public void WriteRawValue(string rawValue)
...
+        }
+
+        /// <summary>
+        /// Mark that the writer has transitioned to write an array.
+        /// </summary>
+        private void EnterArrayScope()
+        {
+            this.isWritingConsecutiveRawValuesAtStartOfArray = false;
+            this.isWritingFirstElementInArray = true;
+            this.bitStack.PushFalse();
```
ElizabethOkerio, line 447: So for objectscope we push true to the bitstack then for array scope we push false?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: Good questions. All those scenarios are supported, and the payload in the test cases include such examples.

In the middle of the array, raw values are not such a big deal. I just need to make sure I place a comma before the raw value, but I don't need to worry about what comes next. This is because inside array the `Utf8JsonWriter` will automatically add a comma before each value, so we don't need to add one manually.

At the end of the array, we don't need to add separators either. So what I'm doing at the end of arrays in the `ExitScope` method, I just reset the state of the tracking variables.

I handle objects differently from arrays because in objects I just need to make sure I place the comma before property names that a follow a raw value. So you'll notice inside `WriterName` I check whether we should add a separator by checking the `this.shouldWriterSeparator` variable. And I set that variable to true inside the `WriteRawValue` method if we're *not* inside an array (i.e. if we're inside an object).

The reason I have special treatment for raw values at the start of an array is because it was the tricky case: I have to ensure that I don't add a separator before the raw value (which is a different treatment from raw values in the middle of an array), I also have to make sure that I have a separator after the raw values of the start of the array before the first non-raw value because `Utf8JsonWriter` will not add it automatically since it mistakenly thinks it's at the beginning of the array.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: Yes, that's exactly what it means.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -274,12 +339,167 @@ public void WriteRawValue(string rawValue)
...
+            // method, so we pop then push back.
+            isInObject = this.bitStack.Pop();
+            if (isInObject)
+            {
+                this.bitStack.PushTrue();
+            }
+            else
+            {
+                this.bitStack.PushFalse();
+            }
```
ElizabethOkerio, line 500: is this bit here necessary? From the code just above it, we check the `CurrentDepth` if it is `0` we automatically return false which means the writer is in the `ArrayScope`  otherwise we return true for `ObjectScope`. In this code segment we are just popping and pushing which is not used in determining the current scope.

habbes: The return value is not the value that was peeked, it is an indicator of whether or not we were able to peek. When the `CurrentDepth` is `0`, the stack is empty, and therefore we can't peek so we return false, otherwise we return true. The value that was peeked is placed in the `isInObject` `out` variable. The reason we pop then push the value back is because `BitStack` doesn't have a `Peek()` method, so we simulate one by popping the pushing back the value we popped.

habbes:
I made some changes to the PR to address the performance issues. Instead of a passing the `writeStream` to `Utf8JsonWriter`, I pass an `ArrayBufferWriter<byte>` buffer. Now when we need to write raw values directly, we write them to the buffer writer, this means we don't need to flush to ensure correct order of writes. We only write to the stream when we flush.

**Here are the new perf figures when the payload includes raw values**

BenchmarkDotNet=v0.13.1, OS=Windows 10.0.20348
Intel Xeon E-2336 CPU 2.90GHz, 1 CPU, 12 logical and 6 physical cores
.NET SDK=5.0.404 [C:\Program Files\dotnet\sdk]
  [Host] : .NET 6.0.11 (6.0.1122.52304), X64 RyuJIT

Toolchain=InProcessEmitToolchain  InvocationCount=1  UnrollFactor=1


|             Method |                                           WriterName |       Mean |     Error |    StdDev |      Gen 0 |     Gen 1 |     Gen 2 |  Allocated |
|------------------- |----------------------------------------------------- |-----------:|----------:|----------:|-----------:|----------:|----------:|-----------:|
 WriteWithRawValues |                                    NoOpWriter-Direct |   3.918 ms | 0.0755 ms | 0.0776 ms |          - |         - |         - |   5,003 KB |
 WriteWithRawValues |                               ODataJsonWriter-Direct |  44.062 ms | 0.8742 ms | 0.8978 ms |  1000.0000 |         - |         - |  11,675 KB |
 WriteWithRawValues |                         ODataJsonWriter-Direct-Async | 206.520 ms | 1.0503 ms | 0.9825 ms |  8000.0000 |         - |         - |  53,797 KB |
 WriteWithRawValues |                                   ODataMessageWriter | 299.875 ms | 0.6858 ms | 0.6415 ms | 39000.0000 |         - |         - | 239,573 KB |
 WriteWithRawValues |                             ODataMessageWriter-Async | 894.690 ms | 1.8934 ms | 1.6784 ms | 57000.0000 | 1000.0000 |         - | 350,650 KB |
 WriteWithRawValues |                    ODataMessageWriter-Utf8JsonWriter | 288.920 ms | 0.6315 ms | 0.5907 ms | 38000.0000 | 1000.0000 |         - | 236,511 KB |
 WriteWithRawValues |              ODataMessageWriter-Utf8JsonWriter-Async | 487.874 ms | 0.4348 ms | 0.4067 ms | 40000.0000 | 3000.0000 | 2000.0000 | 256,846 KB |
 WriteWithRawValues |                           ODataUtf8JsonWriter-Direct |  32.389 ms | 0.0979 ms | 0.0916 ms |  1000.0000 |         - |         - |   8,604 KB |
 WriteWithRawValues |                     ODataUtf8JsonWriter-Direct-Async |  60.699 ms | 0.2026 ms | 0.1896 ms |  3000.0000 | 2000.0000 | 2000.0000 |  28,935 KB |


**New benchmark figures without raw values**

`ODataUtf8JsonWriter` still performing better than `JsonWriter`. There's also been a slight performance improvement in `ODataUtf8JsonWriter`'s async API.

----------------- |----------------------------------------------------- |-----------:|----------:|----------:|-----------:|----------:|-----------:|
 WriteToFileAsync |                                    NoOpWriter-Direct |   1.383 ms | 0.0237 ms | 0.0210 ms |          - |         - |       4 KB |
 WriteToFileAsync |                               ODataJsonWriter-Direct |  35.790 ms | 0.1317 ms | 0.1100 ms |  1000.0000 |         - |   6,675 KB |
 WriteToFileAsync |                         ODataJsonWriter-Direct-Async | 178.943 ms | 0.6467 ms | 0.5733 ms |  8000.0000 |         - |  48,537 KB |
 WriteToFileAsync |                                   ODataMessageWriter | 271.745 ms | 0.3669 ms | 0.3432 ms | 36000.0000 |         - | 224,024 KB |
 WriteToFileAsync |                             ODataMessageWriter-Async | 811.026 ms | 2.0030 ms | 1.7756 ms | 54000.0000 | 1000.0000 | 335,004 KB |
 WriteToFileAsync |                    ODataMessageWriter-Utf8JsonWriter | 260.598 ms | 0.4387 ms | 0.4104 ms | 35000.0000 | 1000.0000 | 217,443 KB |
 WriteToFileAsync |                           ODataUtf8JsonWriter-Direct |  21.663 ms | 0.0379 ms | 0.0317 ms |          - |         - |      84 KB |
 WriteToFileAsync |                     ODataUtf8JsonWriter-Direct-Async |  43.895 ms | 0.2531 ms | 0.2114 ms |          - |         - |     294 KB |

habbes:
I've added tests to the flush-if-threshold-reached logic because there was a gap in the tests.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -26,15 +27,59 @@ internal sealed class ODataUtf8JsonWriter : IJsonWriter, IDisposable, IJsonWrite
...
+        /// the next object or array entry.
+        /// </summary>
+        private bool shouldWriteSeparator = false;
+        /// <summary>
+        /// Stack used to keep track of scope transitions i.e,
+        /// whether we're currently in an Object or an Array.
+        /// </summary>
+        private BitStack bitStack;
+        /// <summary>
+        /// Whether we're about the first element in array. This helps
```
gathogojr, line 72:
```suggestion
        /// Whether we're about to write the first element in array. This helps
```

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonWriterAsyncBaseTests.cs
```diff
@@ -0,0 +1,243 @@
...
+                await jsonWriter.WriteValueAsync((short)1124);
+                await jsonWriter.WriteNameAsync("ByteProp");
+                await jsonWriter.WriteValueAsync((byte)10);
+                await jsonWriter.WriteNameAsync("LongProp");
+                await jsonWriter.WriteValueAsync(234234L);
+                await jsonWriter.WriteNameAsync("SignedByteProp");
+                await jsonWriter.WriteValueAsync((sbyte)-10);
+                await jsonWriter.WriteNameAsync("GuidProp");
+                await jsonWriter.WriteValueAsync(new Guid("00000012-0000-0000-0000-012345678900"));
+                await jsonWriter.WriteNameAsync("ArrayPropWithEveryValueRaw");
```
gathogojr, line 76: `ArrayPropWithEveryValueRaw`? It'd seem to me like the values in the array alternate between non-raw and raw...

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonWriterAsyncBaseTests.cs
```diff
@@ -0,0 +1,243 @@
...
+                await jsonWriter.EndArrayScopeAsync();
+                await jsonWriter.StartArrayScopeAsync();
+                await jsonWriter.StartArrayScopeAsync();
+                await jsonWriter.WriteRawValueAsync(@"""rawAtArrayStartBeforeTimeSpan""");
+                await jsonWriter.WriteValueAsync(new TimeSpan(12, 42, 30));
+                await jsonWriter.EndArrayScopeAsync();
+                await jsonWriter.WriteRawValueAsync(@"""rawAtArrayStartBeforeObject""");
+                await jsonWriter.StartObjectScopeAsync();
+                await jsonWriter.EndObjectScopeAsync();
+                await jsonWriter.EndArrayScopeAsync();
```
gathogojr, line 178: Did you really intend this:
```
[[\"rawAtArrayStartBeforeTimeSpan\",\"PT12H42M30S\"],\"rawAtArrayStartBeforeObject\",{}]
```
or this:
```
[\"rawAtArrayStartBeforeTimeSpan\",\"PT12H42M30S\"],[\"rawAtArrayStartBeforeObject\",{}]
```
If you intended the latter, _and my guts tells me that's what you intended_, I'd expect the code here to look like:
```suggestion
                await jsonWriter.StartArrayScopeAsync();
                await jsonWriter.WriteRawValueAsync(@"""rawAtArrayStartBeforeTimeSpan""");
                await jsonWriter.WriteValueAsync(new TimeSpan(12, 42, 30));
                await jsonWriter.EndArrayScopeAsync();
                await jsonWriter.StartArrayScopeAsync();
                await jsonWriter.WriteRawValueAsync(@"""rawAtArrayStartBeforeObject""");
                await jsonWriter.StartObjectScopeAsync();
                await jsonWriter.EndObjectScopeAsync();
                await jsonWriter.EndArrayScopeAsync();
```

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonWriterBaseTests.cs
```diff
@@ -0,0 +1,242 @@
...
+                jsonWriter.EndArrayScope();
+                jsonWriter.StartArrayScope();
+                jsonWriter.StartArrayScope();
+                jsonWriter.WriteRawValue(@"""rawAtArrayStartBeforeTimeSpan""");
+                jsonWriter.WriteValue(new TimeSpan(12, 42, 30));
+                jsonWriter.EndArrayScope();
+                jsonWriter.WriteRawValue(@"""rawAtArrayStartBeforeObject""");
+                jsonWriter.StartObjectScope();
+                jsonWriter.EndObjectScope();
+                jsonWriter.EndArrayScope();
```
gathogojr, line 177: Did you really intend this:
```
[[\"rawAtArrayStartBeforeTimeSpan\",\"PT12H42M30S\"],\"rawAtArrayStartBeforeObject\",{}]
```
or this:
```
[\"rawAtArrayStartBeforeTimeSpan\",\"PT12H42M30S\"],[\"rawAtArrayStartBeforeObject\",{}]
```
If you intended the latter, _and my guts tells me that's what you intended_, I'd expect the code here to look like:
```suggestion
                jsonWriter.StartArrayScope();
                jsonWriter.WriteRawValue(@"""rawAtArrayStartBeforeTimeSpan""");
                jsonWriter.WriteValue(new TimeSpan(12, 42, 30));
                jsonWriter.EndArrayScope();
                jsonWriter.StartArrayScope();
                jsonWriter.WriteRawValue(@"""rawAtArrayStartBeforeObject""");
                jsonWriter.StartObjectScope();
                jsonWriter.EndObjectScope();
                jsonWriter.EndArrayScope();
```

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonWriterBaseTests.cs
```diff
@@ -0,0 +1,242 @@
...
+                jsonWriter.WriteValue((short)1124);
+                jsonWriter.WriteName("ByteProp");
+                jsonWriter.WriteValue((byte)10);
+                jsonWriter.WriteName("LongProp");
+                jsonWriter.WriteValue(234234L);
+                jsonWriter.WriteName("SignedByteProp");
+                jsonWriter.WriteValue((sbyte)-10);
+                jsonWriter.WriteName("GuidProp");
+                jsonWriter.WriteValue(new Guid("00000012-0000-0000-0000-012345678900"));
+                jsonWriter.WriteName("ArrayPropWithEveryValueRaw");
```
gathogojr, line 75: Is it only raw values that are written after this line?

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataUtf8JsonWriterAsyncTests.cs
```diff
@@ -433,19 +433,131 @@ public async Task AllowsCustomJavaScriptEncoder()
...
+
+            Assert.Equal(@"[""well"",1,""well"",""well"",""Hello, World""]", await this.ReadStreamAsync(jsonWriter, stream, Encoding.UTF8));
+        }
+
+        [Fact]
+        public async Task FlushingWriterAsyncBeforeBufferIsFullShouldBeOkay()
+        {
+            using var stream = new MemoryStream();
+            using var jsonWriter = new ODataUtf8JsonWriter(stream, true, Encoding.UTF8, bufferSize: 10, leaveStreamOpen: true);
+            jsonWriter.StartArrayScope();
```
gathogojr, line 497:
```suggestion
            await jsonWriter.StartArrayScopeAsync();
```

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataUtf8JsonWriterAsyncTests.cs
```diff
@@ -433,19 +433,131 @@ public async Task AllowsCustomJavaScriptEncoder()
...
+            Assert.Equal(9, stream.Length); // not yet flushed
+
+            await jsonWriter.WriteValueAsync("well");
+            Assert.Equal(23, stream.Length); // data flushed. Stream contents: ["well",1,"well","well"
+
+            await jsonWriter.WriteValueAsync("Hello, World"); // write data larger than buffer size
+            Assert.Equal(38, stream.Length); // stream contents: ["well",1,"well","well","Hello, World"
+
+            jsonWriter.EndArrayScope();
+            jsonWriter.Flush();
```
gathogojr, line 459:
```suggestion
            await jsonWriter.EndArrayScopeAsync();
            await jsonWriter.FlushAsync();
```

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataUtf8JsonWriterAsyncTests.cs
```diff
@@ -433,19 +433,131 @@ public async Task AllowsCustomJavaScriptEncoder()
...
+            Assert.Equal(9, stream.Length); // not yet flushed
+
+            await jsonWriter.WriteRawValueAsync(@"""well""");
+            Assert.Equal(23, stream.Length); // data flushed. Stream contents: ["well",1,"well","well"
+
+            await jsonWriter.WriteRawValueAsync(@"""Hello, World"""); // write data larger than buffer size
+            Assert.Equal(38, stream.Length); // stream contents: ["well",1,"well","well","Hello, World"
+
+            jsonWriter.EndArrayScope();
+            jsonWriter.Flush();
```
gathogojr, line 487:
```suggestion
            await jsonWriter.EndArrayScopeAsync();
            await jsonWriter.FlushAsync();
```

test/PerformanceTests/SerializationComparisonsTests/Lib/ODataMessageWriterAsyncPayloadWriter.cs
```diff
@@ -92,14 +94,30 @@ public async Task WritePayloadAsync(IEnumerable<Customer> payload, Stream stream
                 // start write homeAddress
                 await writer.WriteStartAsync(homeAddressInfo);
 
-                var homeAddressResource = new ODataResource
+                ODataResource homeAddressResource;
+                if (includeRawValues)
```
gathogojr, line 98: I don't know if there was a specific reason why in _test/PerformanceTests/SerializationComparisonsTests/Lib/DataModel.cs_ you added `Misc` property in between `City` and `Street` but if you were to add it to the bottom, I believe you'd be able to avoid duplicate code in this file **plus other files with similar duplicated code**:
```csharp
var homeAddressPropertes = new List<ODataProperty>
{
    new ODataProperty { Name = "City", Value = customer.HomeAddress.City },
    new ODataProperty { Name = "Street", Value = customer.HomeAddress.Street }
};

if (includeRawValues)
{
    homeAddressProperties.Add(new ODataProperty { Name = "Misc", Value = new ODataUntypedValue() { RawValue = $"\"{customer.HomeAddress.Misc}\"" } })
}

var homeAddressResource = new ODataResource
{
    Properties = homeAddressProperties
}
```
Even with `Misc` sandwiched between `City` and `Street`, you should be able to do this:
```csharp
var homeAddressPropertes = new List<ODataProperty>
{
    new ODataProperty { Name = "City", Value = customer.HomeAddress.City }
};

if (includeRawValues)
{
    homeAddressProperties.Add(new ODataProperty { Name = "Misc", Value = new ODataUntypedValue() { RawValue = $"\"{customer.HomeAddress.Misc}\"" } })
}

homeAddressProperties.Add(
    new ODataProperty { Name = "Street", Value = customer.HomeAddress.Street });

var homeAddressResource = new ODataResource
{
    Properties = homeAddressProperties
}
```

This comment applies to both sync and async sections of similar code in the following files:
- _test/PerformanceTests/SerializationComparisonsTests/Lib/ODataMessageWriterPayloadWriter.cs_
- _test/PerformanceTests/SerializationComparisonsTests/Lib/ODataMessageWriterAsyncPayloadWriter.cs_

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -80,90 +126,121 @@ public ODataUtf8JsonWriter(Stream outputStream, bool isIeee754Compatible, Encodi
...
-            {
-                this.writer.Flush();
-            }
+            this.CommitWriterContentsToBuffer();
+            this.writeStream.Write(this.bufferWriter.WrittenMemory.Span);
+            this.bufferWriter.Clear();
+            this.writeStream.Flush();
         }
 
         private void FlushIfBufferThresholdReached()
```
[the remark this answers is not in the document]
habbes: Let me try it out. I usually leave that to the compiler to decide, but in the BCL there's a lot of explicit `MethodIml(MethodImplOptions.AggressiveInlining)`. I'm also curious to see the result, and also whether the compiler is already inlining this.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -80,90 +126,121 @@ public ODataUtf8JsonWriter(Stream outputStream, bool isIeee754Compatible, Encodi
...
                 this.Flush();
             }
         }
 
+        /// <summary>
+        /// Commits the contents of the <see cref="Utf8JsonWriter"/> to the
+        /// bufferWriter. This should be called before writing directly
+        /// to the bufferWriter.
+        /// </summary>
+        private void CommitWriterContentsToBuffer()
```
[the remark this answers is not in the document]
habbes: Let me try that out and share the results.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -500,20 +746,46 @@ public async Task WriteRawValueAsync(string rawValue)
...
+            {
+                this.shouldWriteSeparator = true;
+            }
+
+            if (this.isWritingFirstElementInArray || this.isWritingConsecutiveRawValuesAtStartOfArray)
+            {
+                this.isWritingConsecutiveRawValuesAtStartOfArray = true;
+            }
+
+            this.isWritingFirstElementInArray = false;
```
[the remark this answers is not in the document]
habbes: Alright.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonWriterAsyncBaseTests.cs
habbes: Hmm. Good point. I think the naming does reflect what was intended.

habbes: Or did the test not reflect the what the naming intended 🤔

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonWriterAsyncBaseTests.cs
habbes: I think I did intentend the latter, i.e. to have the raw value at the start of the array before the object.

test/PerformanceTests/SerializationComparisonsTests/Lib/ODataMessageWriterAsyncPayloadWriter.cs
habbes: Thanks. There's no reason to have it in between. Add the property first, then added the conditional statement later in order to be able to run tests without raw values.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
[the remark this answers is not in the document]
habbes: Moved the common logic to `WriteRawValueCore` method.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonWriterAsyncBaseTests.cs
habbes: I think the name should have been `ArrayProperWithEveryOtherValueRaw`

test/PerformanceTests/SerializationComparisonsTests/Lib/ODataMessageWriterAsyncPayloadWriter.cs
habbes: By the way, the reason I had the if statement outside of the loop was to avoid calling the if statement in each iteration since the value of `includeRawValues` doesn't change inside the loop. I thought having the if statement inside the loop means the benchmark test would pay the penalty of the if statement even when the test run doesn't include raw values. Maybe this is something that the system can optimize on its own via branch prediction. I didn't compare the benchmarks between having the if statement in the loop and outside to validate whether it makes a difference.

habbes: I have modified the `foreach (address in customer.Addresses)` code to move the if statement inside the loop and noticed the performance slow down by 10s on the `ODataMessageWriter` sync and async tests. However, this could be a fluke. But I've just realized your question was based on `customer.HomeAddress` and not `customer.Addresses`.

habbes: I think the perf degradation was noise, I've ran it again and the results were closer to the original run.

habbes: That said, there's still probably a minor cost, initially I was using arrays to store the properties, now I'd be using `List` and `List.Add()`

habbes: There's an increase in memory allocations by about 3MB and this does not appear to be a fluke (it's consistent across multiple runs). But it seems too big a difference to be explained by changing from arrays to lists.

habbes: I think I should keep the initial code at the moment since the memory allocations has reduced when I reverted it to what it was.

gathogojr: @habbes Curious, why does `includeRawValues` have to be an option? Why can't you _just_ update the existing benchmark to include raw values?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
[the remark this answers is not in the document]
habbes: I've added the inlining attribute, but I can't tell from the benchmarks whether it leads to an improvement. The absolute mean time of `ODataUtf8JsonWriter-Direct` benchmark has stayed the same (about 22ms). The `ODataMessageWriter-Utf8JsonWriter` mean duration seems to have gone down to <254ms. But this happens sometimes, so it's not necessarily due to this change. The relative gap between that and the baseline `ODataMessageWriter` has remain the same (about 6-7% difference). Though it went up as high as 9% difference in one of the runs.

test/PerformanceTests/SerializationComparisonsTests/Lib/ODataMessageWriterAsyncPayloadWriter.cs
habbes: The reason I made it an option is that it made the benchmarks slower (partly because we write more data as result of additional property and partly because we have to manually write a semicolon keep track of related flags), which makes it harder to compare to older benchmarks that did not include raw values. I also wanted to ensure that changes made to address raw values did not affect performance when not using raw values. My assumption is that raw values are an edge case and so people who do not use raw values should not incur a noticeable performance penalty because we added code that handles raw values. So, I wanted to be able to see the perf with and without raw values.

habbes: Another reason the raw value benchmarks are slower is because I'm doing the string interpolation inside the benchmark and not when the benchmark data is generated.

gathogojr: https://github.com/OData/odata.net/pull/2527#discussion_r1035540039
"_people who do not use raw values should not incur a noticeable performance penalty because we added code that handles raw values_" - By this statement I assume you're only referring to performance penalty in respect to running the benchmarks, not penalty in using the actual library..?

src/Microsoft.OData.Core/Json/BitStack.cs
```diff
@@ -0,0 +1,179 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="ODataUtf8JsonWriterAsyncTests.cs" company="Microsoft">
```
gathogojr, line 2:
```suggestion
// <copyright file="BitStack.cs" company="Microsoft">

```

test/PerformanceTests/SerializationComparisonsTests/Lib/ODataMessageWriterAsyncPayloadWriter.cs
habbes: Performance penalty with respect to the benchmarks. But for me to know that performance for scenarios without raw values was not affected after the change, I needed a way to run benchmarks without writing raw values.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -26,15 +28,59 @@ internal sealed class ODataUtf8JsonWriter : IJsonWriter, IDisposable, IJsonWrite
...
+        /// 
+        /// In particular, we have to ensure we write a separator in the following scenarios:
+        /// - inside a JSON object, before writing a property name that's preceded by a raw value
+        /// - inside an array, before writing an item that's preceded by one or more consecutive raw values at the start of the array.
+        /// We only need to do this for the first non-raw value after a sequence of raw values. That's because the Utf8JsonWriter
+        /// (not aware of the raw values) will think that this is the first item in the array and not include a separator before it.
+        /// - inside an array, before writing a raw value that is not at the start of the array
+        /// 
+        /// The following variables are used to keep track of state that allows us to figure out where to manually place a separator
+        /// This will not be necessary when we implement WriteRawValue using Utf8JsonWriter.WriteRawValue()
```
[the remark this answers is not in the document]
habbes: We already have an issue for this, it's mentioned in the code comments :
```
/// see: https://github.com/OData/odata.net/issues/2420
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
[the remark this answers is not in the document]
habbes: See: https://github.com/OData/odata.net/pull/2527/files#r1035532741

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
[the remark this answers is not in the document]
gathogojr: @KenitoInc Line below 👇

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -268,18 +358,159 @@ public void WriteValue(byte[] value)
...
+        /// </returns>
+        private bool TryPeekScope(out bool isInObject)
+        {
+            isInObject = false;
+            if (this.bitStack.CurrentDepth == 0)
+            {
+                return false;
+            }
+
+            // BitStack doesn't implement a Peek()
```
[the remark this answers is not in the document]
habbes: It would probably be more efficient to have a specific `Peek()` but it would increase the complexity of the code, especially if the depth > 64 (this is rare in practice, but we'd still need to implement the logic to handle it). Since `Peek()` and `Pop()` are relatively cheap and are aggressively inlined, I don't think we'll save that much in implementing a `Peek()` method. That's why I felt like it wasn't worth the effort. That said, I haven't actually measured the cost savings of implementing a `Peek()` (because doing so would require me to implement `Peek()` in the first place).

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -268,18 +358,159 @@ public void WriteValue(byte[] value)
...
+            this.bufferWriter.Write(Encoding.UTF8.GetBytes(rawValue));
+
+            // since we bypass the Utf8JsonWriter, we need to signal to other
+            // Write methods that a separator should be written first
+            if ((this.isWritingFirstElementInArray || this.isWritingConsecutiveRawValuesAtStartOfArray) || !this.IsInArray())
+            {
+                this.shouldWriteSeparator = true;
+            }
+
+            if (this.isWritingFirstElementInArray || this.isWritingConsecutiveRawValuesAtStartOfArray)
```
gathogojr: Shouldn't this be just:
```suggestion
            if (this.isWritingFirstElementInArray)
```
Because otherwise we're setting `this.isWritingConsecutiveRawValuesAtStartOfArray` to `true` when it already has a value of `true`...

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -257,29 +350,170 @@ public void WriteValue(byte[] value)
...
+            this.bufferWriter.Write(Encoding.UTF8.GetBytes(rawValue));
+
+            // since we bypass the Utf8JsonWriter, we need to signal to other
+            // Write methods that a separator should be written first
+            if ((this.isWritingAtStartOfArray || this.isWritingConsecutiveRawValuesAtStartOfArray) || !this.IsInArray())
+            {
+                this.shouldWriteSeparator = true;
+            }
+
+            if (this.isWritingAtStartOfArray || this.isWritingConsecutiveRawValuesAtStartOfArray)
```
gathogojr, line 404:
```suggestion
            if (this.isWritingAtStartOfArray)
```
Otherwise we're setting `this.isWritingConsecutiveRawValuesAtStartOfArray` to `true` when it already has a value of `true`...

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -257,29 +348,170 @@ public void WriteValue(byte[] value)
...
+            this.isWritingAtStartOfArray = false;
+        }
+
+        /// <summary>
+        /// Manually writes an item separator ','
+        /// to the output if necessary, bypassing the Utf8JsonWriter.
+        /// Should be used before property names in an object or values
+        /// in an array.
+        /// This method should not be called by <see cref="WriteRawValue(string)"/>.
+        /// </summary>
```
gathogojr, line 418:
```suggestion
        /// </summary>
        /// <remarks>
        /// This method should not be called by <see cref="WriteRawValue(string)"/>.
        /// </remarks>
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -257,29 +348,170 @@ public void WriteValue(byte[] value)
...
+        }
+
+        /// <summary>
+        /// Manually writes an item separator ',' to the output, bypassing the Utf8JsonWriter.
+        /// </summary>
+        private void WriteItemSeparator()
+        {
+            this.CommitUtf8JsonWriterContentsToBuffer();
+            Span<byte> buf = this.bufferWriter.GetSpan(1);
+            buf[0] = (byte)',';
```
gathogojr, line 439: Another place where you could use `parantheses` field?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -257,29 +348,170 @@ public void WriteValue(byte[] value)
...
+            this.isWritingAtStartOfArray = true;
+            this.bitStack.PushFalse();
+        }
+
+        /// <summary>
+        /// Exit the current object or array scope.
+        /// </summary>
+        private void ExitScope()
+        {
+            this.isWritingAtStartOfArray = false;
```
gathogojr, line 468:
```suggestion
            this.isWritingAtStartOfArray = false;
            this.isWritingConsecutiveRawValuesAtStartOfArray = false;
```
For our piece of mind...

gathogojr: Or maybe that could introduce a bug... Please double-check

gathogojr (approved):
:shipit:
Great work!

gathogojr (approved):
:shipit:

xuzhg (approved):

Issue #2512, opened 2022-09-26
Validate that a recursive complex-typed property must be nullable


mikepizzo:
Marking a recursive complex-typed property as non-nullable would lead to an infinite recursion.  For example:
```xml
⁠<ComplexType Name="myType>
    <Property Name="nested" Type="self.myType" Nullable="false">
</ComplexType>
```
or

```xml
⁠<ComplexType Name="myType>
    <Property Name="nested" Type="self.derivedType" Nullable="false">
</ComplexType>

⁠<ComplexType Name="derivedType" BaseType="self.myType">
    <Property Name="foo" Type="Edm.String">
</ComplexType>
```

We should validate against the above definitions (similar to our existing rule "ValidationRules.NavigationPropertyWithRecursiveContainmentTargetMustBeOptional"

### Assemblies affected

EdmLib.dll

### Reproduce steps

Create and validate the above model.

### Expected result

A validation error saying that a recursive complex-typed property must be nullable.

### Actual result

No error.

PR #2548, opened 2022-11-10
Validate that a recursive complex-typed property must be nullable

lisicase:
### Issues

This pull request addresses #2512.

### Description

Validates that if a complex-typed property is recursive, it must also be nullable.

- [x] *Test cases added*


src/Microsoft.OData.Edm/Validation/EdmErrorCode.cs
```diff
@@ -1383,6 +1383,11 @@ public enum EdmErrorCode
         /// </summary>
         AnnotationApplyToNotAllowedAnnotatable = 400,
 
+        /// <summary>
+        /// A recursive complex-typed property must be optional.
+        /// </summary>
+        RecursiveComplexTypedPropertyMustBeOptional = 401,
```
corranrogue9, line 1389: This will be a breaking change, you should probably move it to the end of the value list. You don't have to assign it a value, but if you do, make sure it is consistent with the other values at the end that aren't assigned values

sodiqjonovich (approved):

src/Microsoft.OData.Edm/Validation/EdmErrorCode.cs
lisicase: ✅

lisicase:
Replaced by [PR#2558](https://github.com/OData/odata.net/pull/2558)

PR #2529, opened 2022-10-18
Fix compilation on Linux

AntiGuideAkquinet:
### Issues

*This pull request fixes #2406.*

### Description

Renaming the PublicApi directory to PublicAPI fixed compilation on Linux devices. This influences CI/CD pipelines and developers on Linux machines in general.

### Checklist (Uncheck if it is not completed)

- [ ] *Test cases added*
- [ ] *Build and test with one-click build and test script passed*

PR #2549, opened 2022-11-11
Fix compilation on Linux (duplicate)

### LINKED ISSUE #2406: incorrect case for PublicAPI folder
The last three letters of the path odata.net/src/Microsoft.OData.Edm/PublicApi should be upper case so it compiles on Linux without generating a bunch of RS0016 errors

### Assemblies affected

Current version of Microsoft.OData.Edm

### Reproduce steps

Pull source onto a Linux machine and try to compile Microsoft.OData.Edm project

### Expected result

It would compile

### Actual result

It doesn't compile and yields a whole slew of RS0016 errors as it can't find the shipped and unshipped API files

habbes:
Renaming the PublicApi directory to PublicAPI fixed compilation on Linux devices. This influences CI/CD pipelines and developers on Linux machines in general.

### Issues

*This pull request fixes #2406 .*

### Description
This is a duplicate of the PR #2529 which was failing to run the CI build pipeline.

- [ ] *Test cases added*


ElizabethOkerio (approved):

gathogojr (approved):
:shipit:

PR #2552, opened 2022-11-11
issue #2098: fix the issue ODataException: To compute an entity's metadata, its key and concurrency-token property values must be provided

### LINKED ISSUE #2098: ODataException: To compute an entity's metadata, its key and concurrency-token property values must be provided
Entities having auto expand property is throwing an exception "ODataException: To compute an entity's metadata, its key and concurrency-token property values must be provided" when using $select=prop1. 

### Assemblies affected

  Microsoft.OData.Core 7.7.3.0

### Reproduce steps
This is an issue where entities have auto expand property enabled. So, even if we are requesting another property like displayName, it throws an exception below while writing the response using writeNestedResourceInfo(nestedResourceInfo).

GET  ~/authorizationPolicy/authorizationPolicy?api-version=AGSbetaB-internal&$select=displayName

If id is also selected, we don't see this exception then.
Example : GET  ~/authorizationPolicy/authorizationPolicy?api-version=AGSbetaB-internal&$select=id, displayName

### Expected result

Retrieve the displayName of the object requested.

### Actual result
Microsoft.ODataException: "The entity instance value of type 'Microsoft.DirectoryServices.AuthorizationPolicy' doesn't have a value for property 'id'. To compute an entity's metadata, its key and concurrency-token property values must be provided."

xuzhg:
…adata, its key and concurrency-token property values must be provided

### Issues

*This pull request fixes #2098.*

### Description

*Basically, to generate a link for a contained resource we need the id of the containing resource. If the id value wasn’t included in a $select, we don’t have the id so we can’t create the link and throw an exception. We used to catch and silently ignore this exception, which resulted in creating an invalid URL, which was bad.  However, now we throw an exception in cases we didn’t before, which is breaking existing apps.

We could fix this at the webapi level by always making sure the key properties are included (similar to the workaround someone provided), but by the time it gets to ODL it’s too late if the key property isn’t available.  The best we can do is probably to not generate the link if we are missing this information, although even that is non-ideal.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataWriterCore.cs
+++ b/src/Microsoft.OData.Core/ODataWriterCore.cs
@@ -2826,19 +2826,11 @@ namespace Microsoft.OData
         private ODataPath AppendEntitySetKeySegment(ODataPath odataPath, bool throwIfFail)
         {
             ODataPath path = odataPath;
-            
-            if (EdmExtensionMethods.HasKey(this.CurrentScope.NavigationSource, this.CurrentScope.ResourceType))
+
+            KeyValuePair<string, object>[] keys = GetKeyProperties(throwIfFail);
+            if (keys == null || keys.Length == 0)
             {
                 IEdmEntityType currentEntityType = this.CurrentScope.ResourceType as IEdmEntityType;
-                ODataResourceBase resource = this.CurrentScope.Item as ODataResourceBase;
-                Debug.Assert(resource != null,
-                    "If the current state is Resource the current item must be an ODataResource as well (and not null either).");
-
-                ODataResourceSerializationInfo serializationInfo = this.GetResourceSerializationInfo(resource);
-
-                KeyValuePair<string, object>[] keys = ODataResourceMetadataContext.GetKeyProperties(resource,
-                        serializationInfo, currentEntityType, throwIfFail);
-
                 path = path.AddKeySegment(keys, currentEntityType, this.CurrentScope.NavigationSource);
             }
 
@@ -2847,10 +2839,23 @@ namespace Microsoft.OData
 
         private bool TryBuildKeySegment(out KeySegment keySegment)
         {
-            keySegment = null;
+            KeyValuePair<string, object>[] keys = GetKeyProperties(false /*throwIfFail*/);
+            if (keys == null || keys.Length == 0)
+            {
+                keySegment = null;
+                return false;
+            }
+
+            IEdmEntityType currentEntityType = this.CurrentScope.ResourceType as IEdmEntityType;
+            keySegment = new KeySegment(keys, currentEntityType, this.CurrentScope.NavigationSource);
+            return true;
+        }
+
+        private KeyValuePair<string, object>[] GetKeyProperties(bool throwIfFail)
+        {
             if (!EdmExtensionMethods.HasKey(this.CurrentScope.NavigationSource, this.CurrentScope.ResourceType))
             {
-                return false;
+                return null;
             }
 
             IEdmEntityType currentEntityType = this.CurrentScope.ResourceType as IEdmEntityType;
@@ -2860,15 +2865,7 @@ namespace Microsoft.OData
 
             ODataResourceSerializationInfo serializationInfo = this.GetResourceSerializationInfo(resource);
 
-            KeyValuePair<string, object>[] keys = ODataResourceMetadataContext.GetKeyProperties(resource, serializationInfo, currentEntityType, false /*throwIfFail*/);
-            if (keys == null || keys.Length == 0)
-            {
-                return false;
-            }
-
-            keySegment = new KeySegment(keys, currentEntityType, this.CurrentScope.NavigationSource);
-
-            return true;
+            return ODataResourceMetadataContext.GetKeyProperties(resource, serializationInfo, currentEntityType, throwIfFail);
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Core/ODataWriterCore.cs
+++ b/src/Microsoft.OData.Core/ODataWriterCore.cs
@@ -2760,27 +2760,41 @@ namespace Microsoft.OData
                                         throw new ODataException(Strings.ODataWriterCore_PathInODataUriMustBeSetWhenWritingContainedElement);
                                     }
 
-                                    // If there's only non-key properties selected in the request URI, then We don't have the key values to calcuate the key segment.
-                                    // So far, let's simply not to generate the link if we are missing this key information. Although even that is non-ideal.
-                                    // We should find a better solution for that. One way is to expand ODataResource to include the 'Key' properties if that's not presented in $select.
-                                    KeySegment keySegment;
-                                    if (TryBuildKeySegment(out keySegment))
+                                    ODataPath newPath = null;
+                                    if (!EdmExtensionMethods.HasKey(this.CurrentScope.NavigationSource, this.CurrentScope.ResourceType))
                                     {
-                                        odataPath.AddKeySegment(keySegment);
+                                        // if there's no key, for example in a complex property, just use the existing odata path.
+                                        newPath = odataPath;
+                                    }
+                                    else
+                                    {
+                                        // If there's only non-key properties selected in the request URI, then We don't have the key values to calcuate the key segment.
+                                        // So far, let's simply not to generate the link if we are missing this key information. Although even that is non-ideal.
+                                        // We should find a better solution for that. One way is to expand ODataResource to include the 'Key' properties if that's not presented in $select.
+                                        KeySegment keySegment;
+                                        if (TryBuildKeySegment(out keySegment))
+                                        {
+                                            newPath = odataPath.AddKeySegment(keySegment);
+                                        }
+                                        else
+                                        {
+                                            newPath = null;
+                                        }
+                                    }
 
+                                    if (newPath != null)
+                                    {
                                         if (typeCastFromExpand != null)
                                         {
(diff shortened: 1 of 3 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
```diff
@@ -512,6 +516,21 @@ static AutoComputePayloadMetadataInJsonIntegrationTests()
...
+            StringWriter stringWriter = new StringWriter(stringBuilder);
+
+            using (var writer = XmlWriter.Create(stringWriter))
+            {
+                IEnumerable<EdmError> errors;
+                CsdlWriter.TryWriteCsdl(Model, writer, CsdlTarget.OData, out errors);
+            }
+
+            string modelAsString = stringBuilder.ToString();
+        }
```
mikepizzo, line 532: Is this test intended to be part of this PR?

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
```diff
@@ -1356,6 +1375,172 @@ private static IEdmModel GetModelWithAbstractType(out EdmEntitySet entitySet)
...
+        [InlineData("full")]
+        public void WritingContainedNavigationPropertyWithoutProvideKeyPropertyOnParentResourceWorks(string metadataLevel)
+        {
+            MemoryStream outputStream = new MemoryStream();
+            var container = ContainerBuilderHelper.BuildContainer(null);
+            IODataResponseMessage message = new InMemoryMessage() { Stream = outputStream, Container = container };
+
+            message.SetHeader("Content-Type", metadataLevel == "full" ? "application/json;odata.metadata=full" : "application/json");
+            ODataMessageWriterSettings settings = new ODataMessageWriterSettings();
+            var result = new ODataQueryOptionParser(Model, EntityType, EntitySet,
```
mikepizzo, line 1389: >result [](http://example.com/codeflow?start=16&length=6)

maybe a better variable name -- "queryOptions" or "selectExpand"?

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
```diff
@@ -1356,6 +1375,172 @@ private static IEdmModel GetModelWithAbstractType(out EdmEntitySet entitySet)
...
+        [InlineData("minimal")]
+        [InlineData("full")]
+        public void WritingContainedNavigationPropertyWithoutProvideKeyPropertyOnParentResourceWorks(string metadataLevel)
+        {
+            MemoryStream outputStream = new MemoryStream();
+            var container = ContainerBuilderHelper.BuildContainer(null);
+            IODataResponseMessage message = new InMemoryMessage() { Stream = outputStream, Container = container };
+
+            message.SetHeader("Content-Type", metadataLevel == "full" ? "application/json;odata.metadata=full" : "application/json");
+            ODataMessageWriterSettings settings = new ODataMessageWriterSettings();
```
mikepizzo, line 1388: >      ODataMessageWriterSettings settings = new ODataMessageWriterSettings(); [](http://example.com/codeflow?start=6&length=77)

for better readability, move this down to immediately before it's used in 1401?

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
```diff
@@ -1356,6 +1375,172 @@ private static IEdmModel GetModelWithAbstractType(out EdmEntitySet entitySet)
...
+                    TypeName = "Namespace.EntityType",
+                    Properties = new[]
+                    {
+                        new ODataProperty { Name = "Name", Value = "SampleName" }
+                    }
+                };
+
+                if (metadataLevel == "full")
+                {
+                    resource.Id = new Uri("http://example.com/EntitySet(1)");
```
mikepizzo, line 1419: what happens if metadata=full and we don't set this? Do we get the exception, or just omit? Is there any change to the old behavior for full metadata?

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
```diff
@@ -1356,6 +1375,172 @@ private static IEdmModel GetModelWithAbstractType(out EdmEntitySet entitySet)
...
+                  "\"@odata.mediaEditLink\":\"EntitySet(1)/$value\"," +
+                  "\"Name\":\"SampleName\"," +
+                  "\"ContainedNavProp@odata.associationLink\":\"http://example.com/EntitySet(1)/ContainedNavProp/$ref\"," +
+                  "\"ContainedNavProp@odata.navigationLink\":\"http://example.com/EntitySet(1)/ContainedNavProp\"," +
+                  "\"ContainedNavProp\":[]}", output);
+            }
+            else
+            {
+                Assert.Equal("{\"@odata.context\":\"http://example.com/$metadata#EntitySet(Name,ContainedNavProp())/$entity\"," +
+                  "\"Name\":\"SampleName\"," +
```
mikepizzo, line 1454: this is the scenario that (prior to this PR) threw the exception, right? 

What would the behavior be if we *did* include the key properties?

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -2834,6 +2845,32 @@ private ODataPath AppendEntitySetKeySegment(ODataPath odataPath, bool throwIfFai
             return path;
         }
 
+        private bool TryBuildKeySegment(out KeySegment keySegment)
```
mikepizzo, line 2848: >TryBuildKeySegment [](http://example.com/codeflow?start=21&length=18)

This seems to be the same logic as in AppendKeySegment for getting the key segment, except that AppendKeySegment returns the path with the key segment appended.  Can AppendKeySegment call this method to get the key segment in order to avoid duplicating the code in two places?

mikepizzo (changes requested):
:clock1:

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
xuzhg: No, That's for my helper. I will remove it.

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
xuzhg: It throws an exception. It's not for contained navigation property. If we only serialize a top level resource without "id" in full metadata, it also throws the same exception.

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
xuzhg: #Resolved.

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
xuzhg: #Resolved.

src/Microsoft.OData.Core/ODataWriterCore.cs
xuzhg: Another reference (only use) for 'AppendEntitySetKeySegment' is for a nested complex property. I think it could be related to the nested navigation property. If it's ok, we 'd need to do the same thing for complex property? Any thougths?

By the way, i have to create another private method to include the duplicate codes.

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
xuzhg: I added a new test case to cover this.

xuzhg:

mikepizzo (dismissed):
:shipit:

mikepizzo (dismissed):
:shipit:

PR #2558, opened 2022-11-16
Validate that a recursive complex-typed property must be nullable

lisicase:
### Issues

This pull request addresses #2512.

### Description

Validates that if a complex-typed property is recursive, it must also be nullable.

- [x] *Test cases added*

### Additional work necessary

Adding this validation rule to the base rule set is a breaking change that should be added in a future release.


src/Microsoft.OData.Edm/Microsoft.OData.Edm.txt
```diff
@@ -103,6 +103,7 @@ EdmModel_Validator_Semantic_InaccessibleType=The named type '{0}' could not be f
 EdmModel_Validator_Semantic_AmbiguousType=The named type '{0}' is ambiguous from the model being validated.
 EdmModel_Validator_Semantic_InvalidNavigationPropertyType=The type of the navigation property '{0}' is invalid. The navigation target type must be an entity type or a collection of entity type. The navigation target entity type must match the declaring type of the partner property.
 EdmModel_Validator_Semantic_NavigationPropertyWithRecursiveContainmentTargetMustBeOptional=The target multiplicity of the navigation property '{0}' is invalid. If a navigation property has 'ContainsTarget' set to true and declaring entity type of the property is the same or inherits from the target entity type, then the property represents a recursive containment and it must have an optional target represented by a collection or a nullable entity type.
+EdmModel_Validator_Semantic_RecursiveComplexTypedPropertyMustBeOptional=The type of the property '{0}' is invalid. If a complex typed property is of the same type (or base type) as its declaring type, then the property represents a recursive containment and the given property must be optional.
```
xuzhg, line 106: The nullability of the property type is invalid. ....

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
```diff
@@ -1192,6 +1192,27 @@ public static class ValidationRules
...
+        /// declaring type of the complex type or a derived type, then the complex type is nullable.
+        /// </summary>
+        public static readonly ValidationRule<IEdmStructuralProperty> RecursiveComplexTypedPropertyMustBeOptional =
+            new ValidationRule<IEdmStructuralProperty>(
+                (context, property) =>
+                {
+                    IEdmTypeReference propType = property.Type;
+                    IEdmStructuredType declaredType = property.DeclaringType;
+
+                    if (propType.Definition == declaredType && !propType.IsNullable)
```
xuzhg, line 1207: do we care about the inheritance? 
For example:

A is base type, B is derived type.
Scenario:

1. A has A type property
2. A has B type proprety
3. B has A type property
4. B has B type property (same as 1)

xuzhg (dismissed):

src/Microsoft.OData.Edm/Microsoft.OData.Edm.txt
lisicase: ✅

test/FunctionalTests/Microsoft.OData.Edm.Tests/Validation/ValidationRulesTests.cs
```diff
@@ -795,6 +795,110 @@ public void SameFunctionsShouldInWithOneInReferencedModelShouldRaiseError()
...
+                ValidationRules.RecursiveComplexTypedPropertyMustBeOptional,
+                aProp,
+                EdmErrorCode.RecursiveComplexTypedPropertyMustBeOptional,
+                Strings.EdmModel_Validator_Semantic_RecursiveComplexTypedPropertyMustBeOptional("aProp"));
+        }
+
+        [Fact]
+        public void ComplexTypedPropertyWithSameSecondInheritedBaseTypeAsDeclaringTypeShouldError()
+        {
+            EdmComplexType a = new EdmComplexType("ns", "a");
```
xuzhg, line 866: It is probably ok using "a", "b", "c". 
but, if we can use more meaningful naming, it could be better.

xuzhg (dismissed):

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
lisicase: Expanded scope to include inheritance ✅

test/FunctionalTests/Microsoft.OData.Edm.Tests/Validation/ValidationRulesTests.cs
lisicase: ✅

xuzhg (dismissed):

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
```diff
@@ -1192,6 +1192,37 @@ public static class ValidationRules
                     }
                 });
 
+        // Add this to rule set as a breaking change for a future release.
+        /// <summary>
+        /// Validates that if a complex-typed property has a type or base type that is the same as
+        /// the complex type, then the complex type is nullable.
+        /// </summary>
+        public static readonly ValidationRule<IEdmStructuralProperty> RecursiveComplexTypedPropertyMustBeOptional =
```
corranrogue9, line 1200:
```suggestion
        public static readonly ValidationRule<IEdmStructuralProperty> RecursiveComplexTypePropertyMustBeOptional =

```

Not sure if this was a typo or not

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
```diff
@@ -1192,6 +1192,37 @@ public static class ValidationRules
                     }
                 });
 
+        // Add this to rule set as a breaking change for a future release.
+        /// <summary>
+        /// Validates that if a complex-typed property has a type or base type that is the same as
```
corranrogue9, line 1197: Is this something required by the standard, or is this something we are adding to help customers since this is likely not what they want?

test/FunctionalTests/Microsoft.OData.Edm.Tests/Validation/ValidationRuleSetTests.cs
```diff
@@ -25,7 +25,8 @@ public void NewValidationRulesShouldBeInTheRuleSetExceptBaselinedExceptionRules(
                 f.Name != "NavigationPropertyEntityMustNotIndirectlyContainItself" &&
                 f.Name != "EntityTypeKeyMissingOnEntityType" &&
                 f.Name != "VocabularyAnnotationTargetAllowedApplyToElement" &&
-                f.Name != "EntityTypeInvalidKeyKeyDefinedInAncestor")
+                f.Name != "EntityTypeInvalidKeyKeyDefinedInAncestor" && 
+                f.Name != "RecursiveComplexTypedPropertyMustBeOptional")
```
corranrogue9, line 29:
```suggestion
                f.Name != nameof("RecursiveComplexTypedPropertyMustBeOptional"))

```

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
lisicase: I borrowed this wording from the attached issue -- although I suppose that could have been a typo

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
lisicase: This rule is meant to be a feature to help customers identify issues with recursive structures earlier on in designing their models. It would be a breaking change to add it to the rule set now, but I believe it was discussed that we're planning to add it in the future.

src/Microsoft.OData.Edm/Validation/ValidationRules.cs
```diff
@@ -1192,6 +1192,37 @@ public static class ValidationRules
...
+
+                    if (!currentPropTypeRef.IsNullable)
+                    {
+                        while (currentPropType != null && currentPropType != declaringType)
+                        {
+                            IEdmStructuredType baseType = ((IEdmStructuredType)currentPropType).BaseType;
+                            currentPropType = baseType;
+                        }
+
+                        if (currentPropType == declaringType)
```
corranrogue9, line 1216: Shouldn't the error be inside the `while` loop? Something like:
```
for (; currentPropType != null; currentPropType = ((IEdmStructuredType)currentPropType).BaseType)
{
  if (currentPropType == declaringType)
  {
    context.AddError(...);
  }
}
```
This way, the caller gets two errors for something like this:
```
<ComplexType Name="foo">
  <Property Name="prop1" Type="foo" />
  <Property Name="prop2" Type="foo" />
</ComplexType>
```

Because both `prop1` nor `prop2` should be nullable

lisicase: That's a good point! I'll add a test to capture this case.

corranrogue9: Ok, I think I see why the `if` statement is outside of the `while` loop. Can you please add a test case for the "two property" case anyway?

corranrogue9: Also, take this model:
```
<ComplexType Name="foo">
  <Property Name="prop1" Type="bar" />
</ComplexType>
<ComplexType Name="bar">
  <Property Name="prop2" Type="fizz" />
  <Property Name="prop3" Type="foo" />
</ComplexType>
<ComplexType Name="fizz">
  <Property Name="prop4" Type="foo" />
</ComplexType>
```
Can you add a test to confirm that there is an error for both `prop3` and `prop4`?

lisicase: Actually, I believe this case falls under the umbrella of the general case. Because the validation rule is for a given property, rather than a given complex type, it will evaluate each of those properties individually. The `currentPropType` is cycling through the inheritance tree of types for that given property (rather than each of the properties in a given complex type).

corranrogue9: If I'm reading the issue correctly, it sounds like we want to make sure we capture all of these infinite recursions. Are you saying there's already validation for some of them? And are there associated tests for that validation (that match my suggestions above)?

lisicase: Added test for multiple invalid properties for completeness ✅

test/FunctionalTests/Microsoft.OData.Edm.Tests/Validation/ValidationRulesTests.cs
```diff
@@ -795,6 +795,135 @@ public void SameFunctionsShouldInWithOneInReferencedModelShouldRaiseError()
...
+            IEdmTypeReference complexTypeRef = complexType.GetTypeReference(false);
+            complexType.AddStructuralProperty("nested1", complexTypeRef);
+            complexType.AddStructuralProperty("nested2", complexTypeRef);
+
+            var model = new EdmModel();
+            model.AddElement(complexType);
+            IEnumerable<EdmError> errors;
+            List<ValidationRule> rules = new List<ValidationRule>{ValidationRules.RecursiveComplexTypedPropertyMustBeOptional};
+
+            model.Validate(new ValidationRuleSet(new List<ValidationRule>(rules)), out errors);
```
corranrogue9, line 913:
```suggestion
            model.Validate(new ValidationRuleSet(rules), out errors);

```

test/FunctionalTests/Microsoft.OData.Edm.Tests/Validation/ValidationRulesTests.cs
```diff
@@ -795,6 +795,135 @@ public void SameFunctionsShouldInWithOneInReferencedModelShouldRaiseError()
...
+        {
+            EdmComplexType complexType = new EdmComplexType("ns", "myType");
+            IEdmTypeReference complexTypeRef = complexType.GetTypeReference(false);
+            complexType.AddStructuralProperty("nested1", complexTypeRef);
+            complexType.AddStructuralProperty("nested2", complexTypeRef);
+
+            var model = new EdmModel();
+            model.AddElement(complexType);
+            IEnumerable<EdmError> errors;
+            List<ValidationRule> rules = new List<ValidationRule>{ValidationRules.RecursiveComplexTypedPropertyMustBeOptional};
```
corranrogue9, line 911:
```suggestion
            List<ValidationRule> rules = new List<ValidationRule>
            {
                ValidationRules.RecursiveComplexTypedPropertyMustBeOptional
            };
```
nit: this syntax is usually written with some newlines, if you're interested in that formatting

corranrogue9 (dismissed):

corranrogue9 (approved):

PR #2569, opened 2022-12-02
Fix performance regression in DataServiceContext DefaultResolveType method

### LINKED ISSUE #2514: Performance degradation in ClientTypeUtil.TryResolveType
After updating ODataLib from 7.8.2 to 7.12.2, we are observing 2x CPU usage in our benchmark tests. Our test sends a high load of concurrent requests to a service that serves get and post endpoints, using a generated odata client. Strangely, this only reproduced for get endpoints.

### Assemblies affected

ODL 7.8.2-7.12.2

### Reproduce steps

Sample controller code

For every outgoing call, a new generated client is created via the `IODataClientFactory`.
```
        public async Task<IActionResult> Get(Guid key)
        {
            var client = this.clientFactory.CreateClient<MyServiceClient>("MyService");

            var entity = await client.SimpleEntities.ByKey(key).GetValueAsync();

            return Ok(entity);
        }

        public async Task<IActionResult> Post([FromBody]SimpleFEEntity newEntity)
        {
            var client = this.clientFactory.CreateClient<MyServiceClient>("MyService");

            SimpleEntity simpleEntity = new SimpleEntity() { Key = newEntity.Key, Description = newEntity.Description };
            client.AddToSimpleEntities(simpleEntity);
            await client.SaveChangesAsync();

            return Created(newEntity);
        }
```

Generated client

We use the templates from ODataConnectedService + some custom code and our client is most similar to https://github.com/OData/ODataConnectedService/blob/master/test/ODataConnectedService.Tests/CodeGenReferences/NamespacePrefixWithDoubleNamespaces.cs

This is the stack trace from profiling, where the ClientTypeUtil.TryResolveType accounted for over 99% inclusive CPU of the process
```
coreclr.dll!UTSemReadWrite::LockRead
coreclr.dll!CMDSemReadWrite::LockRead
coreclr.dll!MDInternalRW::GetParentToken
coreclr.dll!MetaDataImport::GetParentToken
System.Reflection.CustomAttribute!FilterCustomAttributeRecord
System.Reflection.CustomAttribute!AddCustomAttributes
System.Reflection.CustomAttribute!GetCustomAttributes
Microsoft.OData.Client.Metadata.ClientTypeUtil!TryResolveType
Microsoft.OData.Client.DataServiceContext!DefaultResolveType
MyNamespace.MyServiceClient!ResolveTypeFromName
Microsoft.OData.Client.TypeResolver!ResolveTypeFromName
Microsoft.OData.Client.TypeResolver!ResolveTypeForMaterialization
Microsoft.OData.Client.Materialization.EntryValueMaterializationPolicy!Materialize
Microsoft.OData.Client.Materialization.ODataEntityMaterializerInvoker!DirectMaterializePlan
Microsoft.OData.Client.Materialization.ODataEntityMaterializer!ReadImplementation
Microsoft.OData.Client.MaterializeAtom!MoveNextInternal
Microsoft.OData.Client.MaterializeAtom!MoveNext
System.Linq.Enumerable+<CastIterator>d__63`1[System.__Canon]!MoveNext
System.Linq.Enumerable+SelectEnumerableIterator`2[System.__Canon,System.__Canon]!MoveNext
System.Linq.Enumerable+WhereEnumerableIterator`1[System.__Canon]!MoveNext
Microsoft.AspNet.OData.Formatter.Serialization.ODataResourceSetSerializer+<WriteResourceSetAsync>d__7!MoveNext
ODataResourceSetSerializer
Serialization
Formatter
OData
AspNet
Microsoft
```

There was a related change mentioned here https://github.com/OData/odata.net/pull/2130#issuecomment-910325405, which I assume was released some time between the two versions.

How come this was not an issue with ODL 7.8.2 while creating a new client for every request? And is there any pattern in the client code that we could change to avoid the cache miss case?

### Expected result

No performance regression.

### Actual result

High CPU from benchmark testing.

### LINKED ISSUE #2557: Performance degradation in OData.Client materialization from version 7.9.1 to 7.9.2
We experience a pretty big performance degradation when upgrading the Microsoft.OData.Client package. We have narrowed it down to the change from version 7.9.1 (which is fast) to 7.9.2 (which is slow). The newest version we have tested is version 7.12.5, which is also slow.

### Assemblies affected

Microsoft.OData.Client

### Reproduce steps

We have boiled it down to the following testcode:
``` csharp
    [DataTestMethod]
    [TestCategory("Manual Only")]
    [DataRow(100, 3, false, DisplayName = "100_3_NoMat")]
    [DataRow(100, 3, true, DisplayName = "100_3_Mat")]
    public async Task PerformanceTest(int batchSize, int runs, bool materialize)
    {
        TestContext.WriteLine(
            $"{DateTime.Now} Job fetch Performance. Batchsize:{batchSize}, Runs:{runs}");
        var timer = new Stopwatch();
        for (int i = 1; i <= runs; i++)
        {
            TestContext.WriteLine($"¤¤¤ Run {i} ¤¤¤");
            timer.Restart();
            for (var jobId = 700000; jobId < 700000 + batchSize; jobId++)
            {
                var Container = ClientFactory.CreateClient<Container>(ODataUri);
                var jobs = await ((DataServiceQuery<D2bJob>)Container.Jobs.Where(j => j.JobId >= jobId).Take(2))
                                                .ExecuteAsync();
                if (materialize) jobs.ToArray();
            }
            timer.Stop();
            TestContext.WriteLine($"GetJobAsync: {timer.ElapsedMilliseconds} ms");
        }
    }
```
When this unittest is run on version 7.9.2 we get output like the following:
### Version 7.9.2:
```
TestContext Messages:
16-11-2022 14:03:22 Job fetch Performance. Batchsize:100, Runs:3, Materialize:False
¤¤¤ Run 1 ¤¤¤
GetJobs: 550 ms
¤¤¤ Run 2 ¤¤¤
GetJobs: 550 ms
¤¤¤ Run 3 ¤¤¤
GetJobs: 554 ms
```
And
```
TestContext Messages:
16-11-2022 14:03:24 Job fetch Performance. Batchsize:100, Runs:3, Materialize:True
¤¤¤ Run 1 ¤¤¤
GetJobs: 2621 ms
¤¤¤ Run 2 ¤¤¤
GetJobs: 2277 ms
¤¤¤ Run 3 ¤¤¤
GetJobs: 2233 ms
```
### Version 7.9.1:
```
TestContext Messages:
16-11-2022 14:08:59 Job fetch Performance. Batchsize:100, Runs:3, Materialize:False
¤¤¤ Run 1 ¤¤¤
GetJobs: 575 ms
¤¤¤ Run 2 ¤¤¤
GetJobs: 601 ms
¤¤¤ Run 3 ¤¤¤
GetJobs: 548 ms
```
And
```
TestContext Messages:
16-11-2022 14:09:01 Job fetch Performance. Batchsize:100, Runs:3, Materialize:True
¤¤¤ Run 1 ¤¤¤
GetJobs: 1145 ms
¤¤¤ Run 2 ¤¤¤
GetJobs: 826 ms
¤¤¤ Run 3 ¤¤¤
GetJobs: 745 ms
```
It seems that the cost of materialization have gone up around 300% in version 7.9.2 and newer.

Adtitionally we can see that materialization is much cheaper if we reuse the container between calls. This is apparantly due to a cache being reset when a new client is created.
Is it possible for this cache to reside in the ClientFactory? Is there something else we can do to avoid this performance degradation?

gathogojr:
### Issues

*This pull request fixes #2557, fixes #2514*

### Description

Fix performance regression in `DataServiceContext` `DefaultResolveType` method.
- By searching for the materialization type from the assembly containing the data service context instance **first**
- By searching for the materialization type from the executing assembly **next** - where it's different from the assembly in the first step
- By searching for the materialization type from the other loaded assemblies while skipping framework and OData libraries where the type wouldn't be expected to be

Performance improvements stats

**Before this fix:**
| Run |      Method |     Mean |     Error |    StdDev |
|-----:|------------ |---------:|----------:|----------:|
| 1 | ResolveType | 9.559 ms | 0.2664 ms | 0.7771 ms |
| 2 | ResolveType | 10.82 ms | 0.187 ms | 0.511 ms |
| 3 | ResolveType | 10.15 ms | 0.218 ms | 0.644 ms |

**After this fix:**
| Run |      Method |     Mean |     Error |    StdDev |
|-----:|------------ |---------:|----------:|----------:|
| 1 | ResolveType | 3.173 ms | 0.0407 ms | 0.1153 ms |
| 2 | ResolveType | 4.023 ms | 0.0813 ms | 0.2254 ms |
| 3 | ResolveType | 3.161 ms | 0.0625 ms | 0.1772 ms |

**Before the change that introduced the performance regression:**
| Run |      Method |     Mean |     Error |    StdDev |
|-----|------------ |---------:|----------:|----------:|
| 1 | ResolveType | 4.901 ms | 0.3239 ms | 0.9396 ms |
| 2 | ResolveType | 3.148 ms | 0.0417 ms | 0.1195 ms |
| 3 | ResolveType | 5.302 ms | 0.2001 ms | 0.5511 ms |
| 4 | ResolveType | 3.965 ms | 0.2389 ms | 0.6855 ms |
| 5 | ResolveType | 3.495 ms | 0.1477 ms | 0.4354 ms |

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
+++ b/src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
@@ -454,7 +454,7 @@ namespace Microsoft.OData.Client.Metadata
         /// <returns>The server defined type name.</returns>
         internal static string GetServerDefinedTypeName(Type type)
         {
-            ODataTypeInfo typeInfo = GetODataTypeInfo(type);           
+            ODataTypeInfo typeInfo = GetODataTypeInfo(type);
 
             return typeInfo.ServerDefinedTypeName;
         }
@@ -478,7 +478,7 @@ namespace Microsoft.OData.Client.Metadata
         internal static string GetClientFieldName(Type t, string serverDefinedName)
         {
             ODataTypeInfo typeInfo = GetODataTypeInfo(t);
-                        
+
             List<string> serverDefinedNames = serverDefinedName.Split(',').Select(name => name.Trim()).ToList();
             List<string> clientMemberNames = new List<string>();
             foreach (var serverSideName in serverDefinedNames)
@@ -618,7 +618,7 @@ namespace Microsoft.OData.Client.Metadata
             if (containerProperty == null)
             {
                 Type propertyType = propertyInfo.PropertyType;
-                
+
                 // Handle Dictionary<,> , SortedDictionary<,> , ConcurrentDictionary<,> , etc - must also have parameterless constructor
                 if (!propertyType.IsInterface() && !propertyType.IsAbstract() && propertyType.GetInstanceConstructor(true, new Type[0]) != null)
                 {
@@ -631,7 +631,7 @@ namespace Microsoft.OData.Client.Metadata
                     containerProperty = (IDictionary<string, object>)Util.ActivatorCreateInstance(dictionaryType);
                 }
                 else
-                { 
+                {
                     // Not easy to figure out the implementing type
                     return false;
                 }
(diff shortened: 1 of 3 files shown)


habbes (commented):
The code looks good to me. I would request some clarifications on the performance measurements:
- Are the figures shared measuring the expected/best case (i.e., the type is in the `targetAssembly`)?
- If the provided figures are for the expected case, could you share figures for the worst case if possible (the type cannot be resolved)
- Could you run the tests on a version of the client before the changes that introduced the regression were made, just so we can have an idea how close it is to initial expectations, if not the same (or better), at least for the common case (the type does exist in the target or entry assembly).

habbes (approved):

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
```diff
@@ -688,59 +688,152 @@ private static bool IsOverride(Type type, PropertyInfo propertyInfo)
...
         /// <param name="fullNamespace">Namespace of the type.</param>
         /// <param name="languageDependentNamespace">Namespace that the resolved type is expected to be.
         /// Usually same as <paramref name="fullNamespace"/> but can be different
         /// where namespace for client types does not match namespace in service types.</param>
         /// <param name="matchedType">The resolved type.</param>
-        /// <returns>true if type was successfully resolved; otherwise false.</returns>
-        internal static bool TryResolveType(string typeName, string fullNamespace, string languageDependentNamespace, out Type matchedType)
-        {
+        /// <returns>true if a type with the specified name is successfully resolved; otherwise false.</returns>
+        internal static bool TryResolveType(
```
xuzhg, line 701: can we cache the 'Resolving' ? 

I mean once we resolved the type, we should cache it for next use, right?

xuzhg (approved):

src/Microsoft.OData.Client/Metadata/ClientTypeUtil.cs
gathogojr: @xuzhg The cache is in the `DataServiceContext` class.

gathogojr:
> The code looks good to me. I would request some clarifications on the performance measurements:
> 
> * Are the figures shared measuring the expected/best case (i.e., the type is in the `targetAssembly`)?
> * If the provided figures are for the expected case, could you share figures for the worst case if possible (the type cannot be resolved)
> * Could you run the tests on a version of the client before the changes that introduced the regression were made, just so we can have an idea how close it is to initial expectations, if not the same (or better), at least for the common case (the type does exist in the target or entry assembly).

@habbes Failure to find a type to use for materialization when deserializing a payload results into an exception. However, I was able to create a simple benchmark to measure the time the `DefaultResolveType` method takes if a type is not found. The `DefaultResolveType` method is protected that's why I did it this way.
```csharp
public class DataServiceContextBenchmark : DataServiceContext
{
    [Benchmark]
    [IterationCount(targetIterationCount: 100)]
    public void BenchmarkResolveType()
    {
        for (var i = 0; i < 250; i++)
        {
            this.DefaultResolveType("NonExistentType", "NS.Models", "NS.Models");
        }
    }
}
```

Stats:

**Before this fix:**
| Run |               Method |     Mean |     Error |   StdDev |
|-----|--------------------- |---------:|----------:|---------:|
| 1 | BenchmarkResolveType | 4.559 us | 0.3813 us | 1.124 us |
| 2 | BenchmarkResolveType | 2.599 us | 0.0847 us | 0.2319 us |
| 3 | BenchmarkResolveType | 5.856 us | 0.7412 us | 2.174 us |
| 4 | BenchmarkResolveType | 4.339 us | 0.1353 us || 0.3882 us |
| 5 | BenchmarkResolveType | 3.541 us | 0.2066 us | 0.5828 us |

**After this fix:**
| Run |               Method |     Mean |     Error |    StdDev |
|-----|--------------------- |---------:|----------:|----------:|
| 1 | BenchmarkResolveType | 2.912 us | 0.0930 us | 0.2653 us |
| 2 | BenchmarkResolveType | 3.260 us | 0.1559 us | 0.4572 us |
| 3 | BenchmarkResolveType | 2.262 us | 0.0394 us | 0.1136 us |
| 4 | BenchmarkResolveType | 4.269 us | 0.3640 us | 1.068 us |
| 5 | BenchmarkResolveType | 2.603 us | 0.1108 us | 0.3160 us |

The results are a bit inconsistent across runs but it'd be right to say their is an observable improvement after the fix

habbes:
Thanks for adding the additional benchmark data. Looks good to me!

PR #2571, opened 2022-12-02
fix: ExpressionWriter only prepend slash to method call when we have …

### LINKED ISSUE #2570: ExpressionWriter generates wrong expression for custom static method call
ALinq ExpressionWriter generates wrong expression for custom static method call.

### Assemblies affected

Microsoft.OData.Client 7.x*

### Reproduce steps

My OData service implements a CustomUriFunction "me" that takes a string and returns a boolean. Like this:
            CustomUriFunctions.AddCustomUriFunction(
                "me",
                new FunctionSignatureWithReturnType(
                    TypeToReference(typeof(bool)), TypeToReference(typeof(string))));

It is expected to be used like this...
/OData/Files?$filter=me(Officer_Value)

Or with qualified name...
/OData/Files?$filter=OData.me(Officer_Value)

My File type has a property "Officer_Value" of type string. The "me" function is static and unbound and is transformed server side to filter the Files set to only return Files that has an assigned Officer whose initials matched the user doing the request.

I am trying to get Microsoft.OData.Client to construct those requests from a Linq query like...

from f in ctx.Files where OData.me(f.Officer_Value) select f

The client side implementation is like...
    public partial class OData
    {
        public static bool me(string property)
        {
            return true;
        }
    }

It almost works, but the generated Uri has a wrong leading slash in the filter.

### Expected result

The generated Uri should be:

/OData/Files?$filter=WorkZone.me(property=$it/Officer_Value)

### Actual result

It almost works, but the generated Uri has a wrong leading slash in the filter:

/OData/Files?$filter=/WorkZone.me(property=$it/Officer_Value)

### Additional detail

uffelauesen:
…a parent #2570

### Issues

*This pull request fixes #2570 

### Description

Only prepend a slash to method name if it is non static (that is it is having a parent).

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
+++ b/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
@@ -413,7 +413,10 @@ namespace Microsoft.OData.Client
 
                     if (m.Method.Name != "GetValue" && m.Method.Name != "GetValueAsync")
                     {
-                        this.builder.Append(UriHelper.FORWARDSLASH);
+                        if (this.parent != null)
+                        {
+                            this.builder.Append(UriHelper.FORWARDSLASH);
+                        }
 
                         // writing functions in query options
                         writingFunctionsInQuery = true;


uffelauesen:
Messy commits. I Will create a new clean PR.

PR #2573, opened 2022-12-05
Ported DataServiceContext.AutoNullPropagation from V3 ODL

### LINKED ISSUE #669: DataServiceClient - Option to have automatic null propagation in projection.
_Null propagation checks on implicit expands in the projection part of a query is something that makes your link query less readable. It is also often forgotten and you could end up in bugs being found late or in production. I propose a mechanism for letting the projection plan compiler handle nulls automatically. Could be a property (AutoNullPropagation) you set on DataServiceContext._  
### Assemblies affected

_V3 Data Services Client 5.7.0_  
_V4 OData Client x.x.x_  
### Reproduce steps

_The following query would fail if either of the implicit expanded properties have a null returned from the server._
`var q = from r in ctx.Records
        select new Record
        {
            File = new File
            {
                Officer = new EmployeeDomain
                {
                    Name = new Contact
                    {
                        ID = r.File.Officer.Name.ID
                    }
                }
            }
        };` 

_The correct query would be:_
`var q = from r in ctx.Records
        select new Record
        {
            File = r.File == null ? null : new File
            {
                Officer = r.File.Officer == null ? null : new EmployeeDomain
                {
                    Name = r.File.Officer.Name == null ? null : new Contact
                    {
                        ID = r.File.Officer.Name.ID
                    }
                }
            }
        };`
_It is tedious to write the null-propagation checks. Also - it cannot be done on deeper levels than the 1st due to issue #379._
### Expected result

_It would be very handy to be able to just set dataServiceContext.AutoNullPropagation = true;. And then you don't have to include the null checks at all._  
### Actual result

_Actually both shown queries will fail depending on the data sent from the server even if the second one should succeed #379 ._

uffelauesen:
### Issues

*This pull request fixes #669.*

### Description

*Added boolean property AutoNullPropagation to DataServiceContext.*

- [ ] *Test cases added*

### Additional work necessary

*"Docs Needed" - Document the added AutoNullPropagation property. Default is false (that is unchanged bahavior). Setting this property to true will treat all projected expands as having implicit null propagation checks. See the old issue #669*


src/Microsoft.OData.Client/Materialization/ODataEntityMaterializer.cs
```diff
@@ -386,9 +386,13 @@ internal static ProjectionPlan CreatePlanForShallowMaterialization(Type lastSegm
         /// <param name="entry">Entry to get sub-entry from.</param>
         /// <param name="name">Name of sub-entry.</param>
         /// <param name="materializerContext">The materializer context.</param>
-        /// <returns>The sub-entry (never null).</returns>
+        /// <returns>The sub-entry (or null if materializerContext.AutoNullPropagation is true).</returns>
```
xuzhg, line 389: Is it a breaking change?

xuzhg (dismissed):

src/Microsoft.OData.Client/Materialization/ODataEntityMaterializer.cs
uffelauesen: Default value of AutoNullPropagation is false - meaning it is not a breaking change.

KenitoInc:
@uffelauesen 
The build is failing. Kindly fix the errors

KenitoInc:
@uffelauesen 
We still have a build failure issue `test\FunctionalTests\Tests\DataServices\UnitTests\Client.TDD.Tests\Tests\Materialization\TestMaterializerContext.cs(18,46): Error CS0535: 'TestMaterializerContext' does not implement interface member 'IODataMaterializerContext.AutoNullPropagation'`

KenitoInc:
@uffelauesen One more public api test failure.
Kindly run the build locally and resolve all issues

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -476,6 +476,13 @@ public virtual bool ApplyingChanges
             internal set { this.applyingChanges = value; }
         }
 
+
+        /// <summary>Gets or sets whether query projection will handle null propagation automatically. If set to true null propagation checks can be omitted from queries.</summary>
```
KenitoInc, line 480:
```suggestion
        /// <summary>
        /// Gets or sets whether query projection will handle null propagation automatically. If set to true null propagation checks can be omitted from queries.
        ///</summary>
```

KenitoInc: Is this statement correct?
`If set to true null propagation checks can be omitted from queries.`

uffelauesen: That is correct.

KenitoInc (commented):
LGTM

KenitoInc (approved):

habbes (approved):

Issue #2433, opened 2022-06-14
Public API changes don't fail build when applied to a subset of target frameworks


corranrogue9:
The [`corranrogue9/publicapi`](https://github.com/OData/odata.net/commits/corranrogue9/publicapi) branch reproduces the issue. There is a new property that only applies to .NET standard 2.0. When building locally, the build fails. When building on the build server, the build succeeds.

These 4 need to be true:
1. The build fails on the build server
2. The build fails locally
3. The public api tests fail locally
4. The public api tests fail on the build server

PR #2575, opened 2022-12-07
Fix PublicAPI build checks

corranrogue9:
Currently, we don't get build errors when the public API surface has changed but the corresponding `PublicAPI.Unshipped.txt` has not been updated. This is because we get warnings that there are duplicate entries in `PublicAPI.Unshipped.txt` and `PublicAPI.Shipped.txt`. Because the analyzer doesn't know what to do when it encounters those warnings, it doesn't go on to generate the errors that there are missing entries in the text file (effectively disabling the public API build checks). 

This change upgrades those duplicate entry warnings to errors, and it also upgrades the warning that an entry exists in the text file but doesn't correspond to a public API. These two will prevent us from accidentally "disabling" the public API build checks in the future.

This change also addresses the issues in the existing `PublicAPI.*.txt` files, removing the duplicate entries from the `Unshipped` variants.

It also adds support in `Core` and `Client` for multiple target frameworks in preparation for #2433.


src/Microsoft.OData.Client/Microsoft.OData.Client.csproj
```diff
@@ -12,8 +12,13 @@
     <DefineConstants>$(DefineConstants);ASSEMBLY_ATTRIBUTE_ON_NETSTANDARD_20</DefineConstants>
   </PropertyGroup>
   
-  <Import Project="..\Build.props" /> 
-
+  <Import Project="..\Build.props" />
+  
+  <PropertyGroup Condition="'$(TreatWarningsAsErrors)' == 'false' or '$(TreatWarningsAsErrors)' == ''">
```
chrisspre, line 17: Does this not belong in the ../Build.props?

chrisspre (approved):

src/Microsoft.OData.Client/Microsoft.OData.Client.csproj
corranrogue9: We use `Build.props` in our test projects as well, which we don't ship publicly. I can try to consolidate the csproj elements used by our shipped projects into a `.props` file though if you think that makes sense.

xuzhg (approved):

mikepizzo (approved):
:shipit:

lisicase (approved):

PR #2585, opened 2022-12-22
Prevent multiple cast segments

### LINKED ISSUE #2444: OData lib duplicates type cast segment in editLink URI
When derived entity is accessed through parent type entityset and no editLink is provided the default etitLink is generated based on entity ID

### Assemblies affected

OData .Net lib 7.x

### Reproduce steps

Create entities like this
```xml
<EntityType Name="Entity6" Abstract="true">
      <Key>
        <PropertyRef Name="Id"/>
    </Key>
    <Property Name="Id" Type="Edm.String" Nullable="false"/>
</EntityType>
<EntityType Name="Entity7" BaseType="referencedomain.Entity6">
    <Property Name="Extended_attr" Type="Edm.String"/>
</EntityType>

<EntitySet Name="Entity6s" EntityType="Entity6" />
```
### Expected result

"@odata.type": "#referencedomain.Entity7",
"@odata.id": "http://localhost.fiddler:1981/ReferenceDomain.svc/Entity6s('0')/referencedomain.Entity7",
"@odata.editLink": "Entity6s('0')/referencedomain.Entity7",
"Id": "0",

### Actual result

"@odata.type": "#referencedomain.Entity7",
"@odata.id": "http://localhost.fiddler:1981/ReferenceDomain.svc/Entity6s('0')/referencedomain.Entity7",
"@odata.editLink": "Entity6s('0')/referencedomain.Entity7/referencedomain.Entity7",
"Id": "0",

### Additional detail

By the default ID URI has typecast segment and ODataConventionalEntityMetadataBuilder.ComputeEditLink() adds another one.

dnperfors:
### Issues

*This pull request fixes #2444 *

### Description

When a resource is a derived type, by default the id link doesn't contains a cast. The AspNetCoreOData library does add this cast, but then the edit link will contain a double cast. This library prevents this.

- [X] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
@@ -406,7 +406,7 @@ namespace Microsoft.OData.Evaluation
             Uri uri = this.ResourceMetadataContext.Resource.HasNonComputedId ? this.ResourceMetadataContext.Resource.NonComputedId : this.ComputedId;
 
             Debug.Assert(this.ResourceMetadataContext != null && this.ResourceMetadataContext.TypeContext != null, "this.resourceMetadataContext != null && this.resourceMetadataContext.TypeContext != null");
-            if (this.ResourceMetadataContext.ActualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName)
+            if (this.ResourceMetadataContext.ActualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName && uri.Segments.Last() != this.ResourceMetadataContext.ActualResourceTypeName)
             {
                 uri = this.UriBuilder.AppendTypeSegment(uri, this.ResourceMetadataContext.ActualResourceTypeName);
             }
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/Evaluation/ODataConventionalEntityMetadataBuilderTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/Evaluation/ODataConventionalEntityMetadataBuilderTests.cs
@@ -247,9 +247,10 @@ namespace Microsoft.OData.Tests.Evaluation
         }
 
         [Fact]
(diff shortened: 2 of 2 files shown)


KenitoInc:
Related #2592

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -406,7 +406,7 @@ private Uri ComputeEditLink()
             Uri uri = this.ResourceMetadataContext.Resource.HasNonComputedId ? this.ResourceMetadataContext.Resource.NonComputedId : this.ComputedId;
 
             Debug.Assert(this.ResourceMetadataContext != null && this.ResourceMetadataContext.TypeContext != null, "this.resourceMetadataContext != null && this.resourceMetadataContext.TypeContext != null");
-            if (this.ResourceMetadataContext.ActualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName)
+            if (this.ResourceMetadataContext.ActualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName && uri.Segments.Last() != this.ResourceMetadataContext.ActualResourceTypeName)
```
xuzhg, line 409: Does OData spec say "~/odata/customers/NS.VipCustomer/NS.SubVipCustomer not valid?

I think there's no such limitation.

xuzhg (approved):

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
dnperfors: It might not be invalid, but it is weird and unnecessary and it gives an implementation burden when implementing the endpoints for it. (With the net core library)

mikepizzo: It is valid, but not generally useful to put multiple *different* type cast segments -- we shouldn't prevent that.  There is no reason to append the same type segment multiple times, and we shouldn't generate such a URL (which is what this PR fixes).

mikepizzo (approved):
:shipit:

gathogojr:
Issue fixed by https://github.com/OData/AspNetCoreOData/pull/831

PR #2592, opened 2023-01-11
Fix wrong links generated for derived type with navigation property when odata.metadata=full

### LINKED ISSUE #797: Conflicting Microsoft.Data.Services and Microsoft.Data.Services.Client 5.8.2
Microsoft.Data.Services.Client v5.8.2 is not available if the project is using Microsoft.Data.Services in the same project due to inconsistent versions.

### Assemblies affected
Microsoft.Data.Services.Client v5.8.2
Microsoft.Data.Services v5.8.1

### Reproduce steps
Reference both Microsoft.Data.Services.Client v5.8.1 and Microsoft.Data.Services v.5.8.1 in the same project. Try to upgrade Microsoft.Data.Services.Client to 5.8.2. As Microsoft.Data.Services is not available in v5.8.2, this fails.

### Expected result
Upgrading should be possible without errors.

### Actual result
Unable to resolve dependencies. 'Microsoft.Data.Services.Client 5.8.2' is not compatible with 'Microsoft.Data.Services 5.8.1 constraint: Microsoft.Data.Services.Client (= 5.8.1)'. As Microsoft.Data.Services uses Microsoft.Data.Services.Client 5.8.1, this fails.

gathogojr:
### Issues

*This pull request fixes https://github.com/OData/AspNetCoreOData/issues/797.*

### Description

Fix wrong links generated for derived type with navigation property when `odata.metadata=full`.

There are scenarios where the resource id already contains the cast segment but `ComputeEditLink` method defined in `ODataConventionalEntityMetadataBuilder` does not detected that so it ends up adding an extra cast segment.

When that happens, you end up with the response payload looking as follows:
```json
{
    "@odata.context": "http://localhost:5219/odata/$metadata#Customers/ODataAlternateKeySample.Models.GoldCustomer/$entity",
    "@odata.type": "#ODataAlternateKeySample.Models.GoldCustomer",
    "@odata.id": "http://localhost:5219/odata/Customers(1)/ODataAlternateKeySample.Models.GoldCustomer",
    "@odata.editLink": "Customers(1)/ODataAlternateKeySample.Models.GoldCustomer/ODataAlternateKeySample.Models.GoldCustomer",
    "Id": 1,
    "Name": "Tom",
    "CountryOrRegion": null,
    "Passport": null,
    "SSN": "SSN-1-101",
    "Titles@odata.type": "#Collection(String)",
    "Titles": [ "abc", null, "efg" ],
    "Contact@odata.associationLink": "http://localhost:5219/odata/Customers(1)/ODataAlternateKeySample.Models.GoldCustomer/ODataAlternateKeySample.Models.GoldCustomer/Contact/$ref",
    "Contact@odata.navigationLink": "http://localhost:5219/odata/Customers(1)/ODataAlternateKeySample.Models.GoldCustomer/ODataAlternateKeySample.Models.GoldCustomer/Contact"
}
```
This pull request fixes the generation of `EditLink`, `AssociationLink` and `NagivationLink` such that the response payload looks as follows:
```json
{
    "@odata.context": "http://localhost:5219/odata/$metadata#Customers/ODataAlternateKeySample.Models.GoldCustomer/$entity",
    "@odata.type": "#ODataAlternateKeySample.Models.GoldCustomer",
    "@odata.id": "http://localhost:5219/odata/Customers(1)/ODataAlternateKeySample.Models.GoldCustomer",
    "@odata.editLink": "Customers(1)/ODataAlternateKeySample.Models.GoldCustomer",
    "Id": 1,
    "Name": "Tom",
    "CountryOrRegion": null,
    "Passport": null,
    "SSN": "SSN-1-101",
    "Titles@odata.type": "#Collection(String)",
    "Titles": [ "abc", null, "efg" ],
    "Contact@odata.associationLink": "http://localhost:5219/odata/Customers(1)/ODataAlternateKeySample.Models.GoldCustomer/Contact/$ref",
    "Contact@odata.navigationLink": "http://localhost:5219/odata/Customers(1)/ODataAlternateKeySample.Models.GoldCustomer/Contact"
}
```

#### Explanation for the refactor of code in test classes
The code in the following test classes has been significantly refactored:
- `ODataConventionalEntityMetadataBuilderTests`
- `ODataConventionalUriBuilderTests`
- `ODataNavigationLinkTests`
- `TestModel`

Subsequent to the introduction of `ODataUriParser` and `ODataPath` classes to determine whether the resource id already contains the cast segment, tests in some classes started failing. An investigation revealed that in most scenarios, the test logic was impractical.

In the `ODataConventionalEntityMetadataBuilderTests` test class for example, the tests for multi-property key were impractical since the entity type that the tests were based upon wasn't defined with a multi-property key. A refactor was therefore necessary since the `ODataUriParser` validates the supplied keys against the Edm model. The test class setup code had a lot of logical inconsistencies.

In the `ODataNavigationLinkTests` test class, the refactor involved setting the delegates that are invoked to determine the model and the base Uri.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
@@ -15,6 +15,7 @@ namespace Microsoft.OData.Evaluation
     using System.Text;
     using Microsoft.OData.Edm;
     using Microsoft.OData.Edm.Vocabularies.V1;
+    using Microsoft.OData.UriParser;
     #endregion
 
     /// <summary>
@@ -405,10 +406,34 @@ namespace Microsoft.OData.Evaluation
         {
             Uri uri = this.ResourceMetadataContext.Resource.HasNonComputedId ? this.ResourceMetadataContext.Resource.NonComputedId : this.ComputedId;
 
-            Debug.Assert(this.ResourceMetadataContext != null && this.ResourceMetadataContext.TypeContext != null, "this.resourceMetadataContext != null && this.resourceMetadataContext.TypeContext != null");
-            if (this.ResourceMetadataContext.ActualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName)
+            Debug.Assert(this.ResourceMetadataContext != null && this.ResourceMetadataContext.TypeContext != null,
+                "this.resourceMetadataContext != null && this.resourceMetadataContext.TypeContext != null");
+
+            string actualResourceTypeName= this.ResourceMetadataContext.ActualResourceTypeName;
+
+            if (actualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName)
             {
-                uri = this.UriBuilder.AppendTypeSegment(uri, this.ResourceMetadataContext.ActualResourceTypeName);
+                IEdmStructuredType actualResourceType = this.ResourceMetadataContext.ActualResourceType;
+
+                // In some scenarios, the resource id will already contain the cast segment. For example,
+                // when full metadata is requested and a derived entity containing one or more navigation properties
+                // is being serialized, and the association and navigation links need to be serialized as well, the cast
+                // segment will already be contained in the resource id.
+                ODataUriParser odataUriParser = new ODataUriParser(this.MetadataContext.Model, this.MetadataContext.ServiceBaseUri, uri);
+                ODataPath odataPath = odataUriParser.ParsePath();
+
+                // Accomodate two scenarios:
+                // 1. ~/Customers(1)/NS.EnterpriseCustomer
+                // 2. ~/Customers/NS.EnterpriseCustomer(1) - key expression on type segment
+                if (!((odataPath.LastSegment is TypeSegment typeSegment1 
+                    && typeSegment1.TargetEdmType == actualResourceType)
+                    || (odataPath.LastSegment is KeySegment 
+                    && odataPath.Segments.Count >= 2
+                    && odataPath.Segments[odataPath.Segments.Count - 2] is TypeSegment typeSegment2
+                    && typeSegment2.TargetEdmType == actualResourceType)))
+                {
+                    uri = this.UriBuilder.AppendTypeSegment(uri, actualResourceTypeName);
+                }
             }
 
             return uri;
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/Evaluation/ODataConventionalEntityMetadataBuilderTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/Evaluation/ODataConventionalEntityMetadataBuilderTests.cs
@@ -21,9 +21,15 @@ namespace Microsoft.OData.Tests.Evaluation
 
         private readonly ODataConventionalUriBuilder uriBuilder = new ODataConventionalUriBuilder(DefaultBaseUri,
             ODataUrlKeyDelimiter.Parentheses);
-        private readonly TestMetadataContext metadataContext = new TestMetadataContext { GetMetadataDocumentUriFunc = () => MetadataDocumentUri, GetModelFunc = () => TestModel.Model, OperationsBoundToStructuredTypeMustBeContainerQualifiedFunc = type => false };
+        private readonly TestMetadataContext metadataContext = new TestMetadataContext
+        {
+            GetServiceBaseUriFunc = () => DefaultBaseUri,
+            GetMetadataDocumentUriFunc = () => MetadataDocumentUri,
+            GetModelFunc = () => TestModel.Model,
+            OperationsBoundToStructuredTypeMustBeContainerQualifiedFunc = type => false
+        };
         private ODataResource productEntry;
-        private Dictionary<string, object> sinlgeKeyCollection;
+        private Dictionary<string, object> singleKeyCollection;
         private Dictionary<string, object> multiKeysCollection;
         private ODataConventionalEntityMetadataBuilder productConventionalEntityMetadataBuilder;
         private ODataResource derivedMultiKeyMultiEtagMleEntry;
@@ -32,20 +38,21 @@ namespace Microsoft.OData.Tests.Evaluation
         private ODataConventionalEntityMetadataBuilder containedCollectionProductConventionalEntityMetadataBuilder;
         private ODataResource containedProductEntry;
         private ODataConventionalEntityMetadataBuilder containedProductConventionalEntityMetadataBuilder;
-        private Dictionary<string, object> containedSinlgeKeyCollection;
-        private Dictionary<string, object> containedMultiKeysCollection;
+        private Dictionary<string, object> containedSingleKeyCollection;
 
         private const string EntitySetName = "Products";
         private const string EntityTypeName = "TestModel.Product";
-        private const string DerivedEntityTypeName = "TestModel.DerivedProduct";
-        private const string DerivedMleEntityTypeName = "TestModel.DerivedMleProduct";
+
+        private const string MultiKeyEntitySetName = "MultipleKeySet";
+        private const string MultiKeyEntityTypeName = "TestModel.MultipleKeyType";
+        private const string DerivedMleMultiKeyEntityTypeName = "TestModel.DerivedMleMultiKeyType";
 
         public ODataConventionalEntityMetadataBuilderTests()
         {
             #region Product Entry
 
             this.productEntry = new ODataResource();
-            this.sinlgeKeyCollection = new Dictionary<string, object>() { { "Id", 42 } };
+            this.singleKeyCollection = new Dictionary<string, object>() { { "Id", 42 } };
             this.multiKeysCollection = new Dictionary<string, object>() { { "KeyA", "keya" }, { "KeyB", 1 } };
 
             TestFeedAndEntryTypeContext productTypeContext = new TestFeedAndEntryTypeContext
@@ -63,7 +70,7 @@ namespace Microsoft.OData.Tests.Evaluation
                 TypeContext = productTypeContext,
                 Resource = this.productEntry,
                 ETagProperties = new[] { new KeyValuePair<string, object>("Name", "Value") },
-                KeyProperties = this.sinlgeKeyCollection,
+                KeyProperties = this.singleKeyCollection,
                 ActualResourceTypeName = EntityTypeName,
                 SelectedBindableOperations = new IEdmOperation[0],
                 SelectedNavigationProperties = new IEdmNavigationProperty[0],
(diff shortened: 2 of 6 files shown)


src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -408,7 +408,11 @@ private Uri ComputeEditLink()
             Debug.Assert(this.ResourceMetadataContext != null && this.ResourceMetadataContext.TypeContext != null, "this.resourceMetadataContext != null && this.resourceMetadataContext.TypeContext != null");
             if (this.ResourceMetadataContext.ActualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName)
             {
-                uri = this.UriBuilder.AppendTypeSegment(uri, this.ResourceMetadataContext.ActualResourceTypeName);
+                // In some scenarios, the resource Id will already contain the cast segment
+                if (!uri.ToString().EndsWith(string.Concat('/', this.ResourceMetadataContext.ActualResourceTypeName), StringComparison.Ordinal))
```
gathogojr, line 412: I considered using the following logic as an alternative but was concerned about increase in allocations:
```csharp
var odataUriParser = new OData.UriParser.ODataUriParser(this.MetadataContext.Model, this.MetadataContext.ServiceBaseUri, uri);
var odataPath = odataUriParser.ParsePath();
if (!(odataPath.LastSegment is OData.UriParser.TypeSegment typeSegment && typeSegment.TargetEdmType == this.ResourceMetadataContext.ActualResourceType))
{
	uri = this.UriBuilder.AppendTypeSegment(uri, this.ResourceMetadataContext.ActualResourceTypeName);
}
```

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -408,7 +408,11 @@ private Uri ComputeEditLink()
             Debug.Assert(this.ResourceMetadataContext != null && this.ResourceMetadataContext.TypeContext != null, "this.resourceMetadataContext != null && this.resourceMetadataContext.TypeContext != null");
             if (this.ResourceMetadataContext.ActualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName)
             {
-                uri = this.UriBuilder.AppendTypeSegment(uri, this.ResourceMetadataContext.ActualResourceTypeName);
+                // In some scenarios, the resource Id will already contain the cast segment
+                if (!uri.ToString().EndsWith(string.Concat('/', this.ResourceMetadataContext.ActualResourceTypeName), StringComparison.Ordinal))
+                {
+                    uri = this.UriBuilder.AppendTypeSegment(uri, this.ResourceMetadataContext.ActualResourceTypeName);
+                }
```
gathogojr, line 415: I considered using the following logic as an alternative but was concerned about increase in allocations:
```csharp
var odataUriParser = new OData.UriParser.ODataUriParser(this.MetadataContext.Model, this.MetadataContext.ServiceBaseUri, uri);
var odataPath = odataUriParser.ParsePath();
if (!(odataPath.LastSegment is OData.UriParser.TypeSegment typeSegment && typeSegment.TargetEdmType == this.ResourceMetadataContext.ActualResourceType))
{
	uri = this.UriBuilder.AppendTypeSegment(uri, this.ResourceMetadataContext.ActualResourceTypeName);
}
```

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -408,7 +408,11 @@ private Uri ComputeEditLink()
             Debug.Assert(this.ResourceMetadataContext != null && this.ResourceMetadataContext.TypeContext != null, "this.resourceMetadataContext != null && this.resourceMetadataContext.TypeContext != null");
             if (this.ResourceMetadataContext.ActualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName)
             {
-                uri = this.UriBuilder.AppendTypeSegment(uri, this.ResourceMetadataContext.ActualResourceTypeName);
+                // In some scenarios, the resource Id will already contain the cast segment
```
xuzhg, line 411: what scenarios? a little bit details?

xuzhg (dismissed):

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
habbes, line 412: `string.Concat` does not have an overload that accepts a `char` as the first argument, so this method call will probably invoke an overload that takes `object` as an argument, which means the `/` char will be boxed and cause an allocation.

I suspect it would be more efficient to call `string.Concat("/", ...)` instead.

Here's an illustration of the IL generated for both variants: https://sharplab.io/#v2:C4LglgNgNAJiDUAfAAgJgIwFgBQyDMABGgQMIEDeOB1Rhy6ADKQPYB2AxgIbAkAWnAJwAU9JgGcAlBSo1ZyAOwEAysAFhWAcwB0JNl2BCA5AHpDUApIDcM6gF8bBB/iKMWHbirWaRrydOyycoqiOnrcQgBExhHmVg722LZAA

habbes: Also, it may be possible to match the string ending without performing a concatenation, but maybe the code complexity this would introduce is not worth it for this edge case (and it's not guaranteed that it would perform better). e.g.,: `if (foo.EndsWith(bar) && foo[foo.length - bar.length - 1] == '/' ...)` (there's probably an off-by-one error in there, but you get the idea).

habbes: This logic seems to hinge on the fact `ActualResourceTypeName` is a fully-qualified name, and in this context only a type segment would be a fully qualified name. Is that correct?

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
gathogojr: @xuzhg Added an example of a scenario

habbes (dismissed):

uffelauesen:
“No type-cast segment is added to the canonical URL, even if the entity is an instance of a type derived from the declared entity type of its entity set.”
Taken from the spec: http://docs.oasis-open.org/odata/odata/v4.0/errata03/os/complete/part2-url-conventions/odata-v4.0-errata03-os-part2-url-conventions-complete.html#_Toc453752341
To me this equals to odata.id not having a type cast segment as well. 
Think about it - the Canonical URL already uniquely identifies the entity, a type cast segment does nothing good in identifying the entity. Services are required to and do return the correct derived object with all its additions included, no need to have the server perform a type filter on top of an id/key look up. The issue reported is in AspNetCoreOData not in OData repos. 

Thanks for looking into it. 
uffe

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -405,10 +405,24 @@ private Uri ComputeEditLink()
...
+            string actualResourceTypeName= this.ResourceMetadataContext.ActualResourceTypeName;
+
+            if (actualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName)
             {
-                uri = this.UriBuilder.AppendTypeSegment(uri, this.ResourceMetadataContext.ActualResourceTypeName);
+                // In some scenarios, the resource id will already contain the cast segment. For example,
+                // when full metadata is requested and a derived entity containing one or more navigation properties
+                // is being serialized, and the association and navigation links need to be serialized as well, the cast
+                // segment will already be contained in the resource id.
+                string uriToString = uri.ToString();
```
mikepizzo, line 419: >ToString() [](http://example.com/codeflow?start=41&length=10)

I wonder if we should trim any trailing forward-slash.

mikepizzo (dismissed):
:shipit:

ElizabethOkerio (dismissed):

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -405,10 +405,24 @@ private Uri ComputeEditLink()
...
-            if (this.ResourceMetadataContext.ActualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName)
+            Debug.Assert(this.ResourceMetadataContext != null && this.ResourceMetadataContext.TypeContext != null,
+                "this.resourceMetadataContext != null && this.resourceMetadataContext.TypeContext != null");
+
+            string actualResourceTypeName= this.ResourceMetadataContext.ActualResourceTypeName;
+
+            if (actualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName)
             {
-                uri = this.UriBuilder.AppendTypeSegment(uri, this.ResourceMetadataContext.ActualResourceTypeName);
+                // In some scenarios, the resource id will already contain the cast segment. For example,
```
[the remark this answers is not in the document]
gathogojr: @KenitoInc I made use of the `ODataUriParser` and `ODataPath` classes to take care of this scenario. Please check the updated logic

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
gathogojr: Eventually adopted this approach since it made it easy to cater for the scenario raised [here](https://github.com/OData/odata.net/pull/2592/files#r1080945648)

test/FunctionalTests/Microsoft.OData.Core.Tests/Evaluation/ODataConventionalEntityMetadataBuilderTests.cs
```diff
@@ -21,9 +21,15 @@ public class ODataConventionalEntityMetadataBuilderTests
 
         private readonly ODataConventionalUriBuilder uriBuilder = new ODataConventionalUriBuilder(DefaultBaseUri,
             ODataUrlKeyDelimiter.Parentheses);
-        private readonly TestMetadataContext metadataContext = new TestMetadataContext { GetMetadataDocumentUriFunc = () => MetadataDocumentUri, GetModelFunc = () => TestModel.Model, OperationsBoundToStructuredTypeMustBeContainerQualifiedFunc = type => false };
+        private readonly TestMetadataContext metadataContext = new TestMetadataContext
```
gathogojr, line 24: **Note:** The explanation for the refactor of the code in this test class in provided in the PR description

test/FunctionalTests/Microsoft.OData.Core.Tests/Evaluation/TestModel.cs
```diff
@@ -19,8 +19,10 @@ internal class TestModel
         public IEdmEntityType ProductType { get; set; }
         public IEdmEntityType DerivedProductType { get; set; }
         public IEdmEntityType MultipleKeyType { get; set; }
+        public IEdmEntityType DerivedMleMultiKeyType { get; set; }
```
gathogojr, line 22: **Note:** The explanation for the refactor of the code in this test class in provided in the PR description

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataNavigationLinkTests.cs
```diff
@@ -21,6 +22,21 @@ public class ODataNavigationLinkTests
 
         public ODataNavigationLinkTests()
         {
+            var model = new EdmModel();
```
gathogojr, line 25: **Note:** The explanation for the refactor of the code in this test class in provided in the PR description

KenitoInc (approved):
LGTM

gathogojr:
Replaced by https://github.com/OData/AspNetCoreOData/pull/831

PR #2595, opened 2023-01-13
Allow @odata.count to be written after enumeration #2594

### LINKED ISSUE #2594: Allow @odata.count to be written after enumeration.
In ODataJsonLightWriter.cs @odata.count annotations are written before enumeration of the resource set in StartResourceSetAsync or StartDeltaResourceSetAsync. This forces us to get the count calculated before we fetch data. In the implementation I have, this will require issuing a "select count(*) from..." query on my DB before the actual data query is issued "select a, b from...". In cases where all items in a resource set is returned in the payload we could avoid the "select count(*) from..." query as we already know the total number items.

@odata.nextLink annotation already behaves this way. It is either written early (before enumeration), if it set early in the resource set or after enumeration if it set later.

As far as I can read in the spec on odata.org, there is no mention of whether or not @odata.count should come early in the payload, so...

For performance reasons, please allow @odata.count annotations to be written in EndResourceSetAsync if not already written in StartResourceSetAsync or StartDeltaResourceSetAsync, like it is already possible with @odata.nextLink.

### Assemblies affected

*OData .Net lib 7.x*

uffelauesen:
### Issues

*This pull request fixes #2594.*

### Description

*Allow writing @odata.count after enumeration in ODataJsonLightWriter.*

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs
@@ -1668,6 +1668,10 @@ namespace Microsoft.OData.JsonLight
 
                 if (this.writingResponse)
                 {
+                    // Write the count if it's available.
+                    await this.WriteResourceSetCountAsync(resourceSet.Count, null)
+                        .ConfigureAwait(false);
+
                     // Write the next link if it's available.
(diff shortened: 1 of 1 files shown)


ElizabethOkerio:
@uffelauesen add tests.

Issue #2596, opened 2023-01-17
ODataBinaryStreamValue's Stream is never disposed


syprieur:
`ODataBinaryStreamValue` takes ownership of a `Stream` on construction, but this `Stream` is never closed by ODL after it's been serialized.

### Assemblies affected

All.

### Reproduce steps

Set an `ODataResource.Properties` containing a property which value is an `ODataBinaryStreamValue`.
Note how after the `ODataWriter` is done writing the `ODataResource` (or even done being used entirely), the original `Stream` passed to the `ODataBinaryStreamValue` has never had its `Dispose` method called.

### Expected result

As an `ODataBinaryStreamValue` takes ownership of its `Stream` parameter on construction, the `Stream` should be disposed once it's done being serialized.

### Actual result

Dispose is never called.

### Additional detail

For backward compatibility, disposal should be optional (and maybe default to `leaveOpen`).

mikepizzo:
We didn't make the assumption that passing the stream to the constructor of the ODataBinaryStreamValue implied that we were taking over the lifetime of the stream, but in the common case I suspect that will be true so I like the added leaveOpen property.

For backward compatibility, we probably need to keep the default false (as per the PR), but in 8.x we should consider changing the default to true to be more consistent with, i.e., StreamWriter.

gathogojr:
Fixed by https://github.com/OData/odata.net/pull/2597

PR #2459, opened 2022-07-15
Fix synchronous operations disallowed error when using Kestrel web server

gathogojr:
### Issues

*This pull request fixes https://github.com/OData/AspNetCoreOData/issues/624.*

### Description

Fix synchronous operations disallowed error when using Kestrel web server.

`AllowSynchronousIO` option is disabled by default from .NET Core 3.0.0-preview3 in (**Kestrel**, **HttpSys**, **IIS In-Process**, **TestServer**) web servers because these APIs are source of thread starvation and application hangs.
A call to `Dispose` method of `StreamWriter` from `RawValueWriter` was ultimately causing the `Flush` method of `Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpResponseStream` to be triggered.
As a result, a _**System.InvalidOperationException: Synchronous operations are disallowed. Call ReadAsync or set AllowSynchronousIO to true instead**_ is thrown when using Kestrel web server.
What this change does is ensure that `DisposeAsync` method is trigged instead of Dispose from `RawValueWriter` class that is used when writing batch payloads.

#### Stack Trace

> Connection id "0HMJ5BJNH8K3F", Request id "0HMJ5BJNH8K3F:00000001": An unhandled exception was thrown by the application.
> System.AggregateException: One or more errors occurred. (Synchronous operations are disallowed. Call WriteAsync or set AllowSynchronousIO to true instead.)
>  ---> System.InvalidOperationException: Synchronous operations are disallowed. Call WriteAsync or set AllowSynchronousIO to true instead.
>    at Microsoft.AspNetCore.Server.Kestrel.Core.Internal.Http.HttpResponseStream.Flush()
>    at Microsoft.OData.MessageStreamWrapper.MessageStreamWrappingStream.Flush() in C:\odata.net\src\Microsoft.OData.Core\MessageStreamWrapper.cs:line 161
>    at System.IO.BufferedStream.Flush()
>    at Microsoft.OData.MessageStreamWrapper.MessageStreamWrappingStream.Flush() in C:\odata.net\src\Microsoft.OData.Core\MessageStreamWrapper.cs:line 161
>    at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
>    at System.IO.StreamWriter.Dispose(Boolean disposing)
>    at System.IO.TextWriter.Dispose()
[...]

PR #2597, opened 2023-01-17
Capability to close ODataBinaryStreamValue stream after serialization

syprieur:
### Issues

#2596 

### Description

Add a capability for an `ODataBinaryStreamValue` to have its `Stream` disposed by ODL.
- This behavior is optional and is disabled by default for backward compatibility.
- Implemented via an internal property (`bool LeaveOpen`) rather than making `ODataBinaryStreamValue` an `IDisposable` in order to avoid a breaking change to the public API.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
@@ -733,13 +733,21 @@ namespace Microsoft.OData.JsonLight
                 Stream stream = await streamWriter.StartStreamValueScopeAsync().ConfigureAwait(false);
                 await streamValue.Stream.CopyToAsync(stream).ConfigureAwait(false);
                 await stream.FlushAsync().ConfigureAwait(false);
+#if NETCOREAPP3_1_OR_GREATER
+                await stream.DisposeAsync();
+#else
                 stream.Dispose();
+#endif
                 await streamWriter.EndStreamValueScopeAsync().ConfigureAwait(false);
             }
 
             if (!streamValue.LeaveOpen)
             {
+#if NETCOREAPP3_1_OR_GREATER
+                await streamValue.Stream.DisposeAsync();
+#else
                 streamValue.Stream.Dispose();
+#endif
             }
         }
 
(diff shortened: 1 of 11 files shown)


syprieur:
@microsoft-github-policy-service agree [company="Microsoft"]

syprieur:

src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
```diff
@@ -732,6 +736,11 @@ public virtual async Task WriteStreamValueAsync(ODataBinaryStreamValue streamVal
                 stream.Dispose();
                 await streamWriter.EndStreamValueScopeAsync().ConfigureAwait(false);
             }
+
+            if (!streamValue.LeaveOpen)
+            {
+                streamValue.Stream.Dispose();
```
habbes, line 742: Would `Stream.DisposeAsync()` be more appropriate here in case the disposal performs any I/O? (not sure why [line 736](https://github.com/OData/odata.net/pull/2597/files#diff-2cce78e3afeedcc8e1b6c32359a94bdbc867447c0b09b336e2d3c7bcfa6b727eR736) calls `stream.Dispose()`].

habbes: Aah, good point. Yes, I suggest adding conditional compilation block in that case.

We have currently have code like this in certain areas:

```c#
#if NETCOREAPP3_1_OR_GREATER

#else 

#endif
```
(e.g.: https://github.com/OData/odata.net/pull/2459/files#diff-57ee33c26e730ea9ad1e185c8617b2c707ec2b62e1880df463aab29e668df703R93)

habbes: We did run into an issue in some edge cases where synchronous disposal lead to synchronous I/O which caused exceptions in AspNetCore because Kestrel disables synchronous I/O by default.

syprieur: Change made. Sadly, I didn't find an efficient way to prevent code duplication (couldn't extract block to its own method due to lack of `Task.CompletedTask` in older runtimes).

habbes (approved):

ElizabethOkerio (approved):

habbes:
Comment to trigger build.

gathogojr (approved):
LGTM

habbes (approved):

gathogojr (approved):
:shipit:

PR #2600, opened 2023-01-18
Flush out release process documentation

lisicase:
Adding more detail to the release instructions for easier onboarding.

### DIFF
--- a/docs/release.md
+++ b/docs/release.md
@@ -1,23 +1,54 @@
-1. Increment the version number referenced during build in [`Versioning.props`](tools/CustomMSBuild/Versioning.props) by using [semantic versioning](https://semver.org/). Also copy the contents of any `PublicAPI.Unshipped.txt` file that has changed since the last release into its corresponding `PublicAPI.Shipped.txt` file. Both of these steps can usually be accomplished by running the `tools\IncrementVersion.cmd` script. Create a pull request with these changes and merge it.
-2. Kick off a new [nightly build](https://identitydivision.visualstudio.com/OData/_build?definitionId=1104) to generate the nuget packages that will be published for this release:
-3. Generate release notes in the [change log](https://github.com/MicrosoftDocs/OData-docs/blob/main/Odata-docs/changelog/odatalib-7x.md) to be published on release of the new version. This is done by referencing the PRs that have been merged into `master` since the last version increment. In github, each commit should have a link to the PR that was merged to generate that commit, so you can use github history to generate the changelog. Create a pull request with these changes in the docs repo and merge it. 
-4. Download the build artifacts from the nightly build to your local machine
-5. Mark the nightly build to be retained indefinitely
-6. Download NuGet Package Explorer from the windows store and use it to verify the URLs of each package for license, release notes, and package information
-7. Log into to your nuget.org account and link it to your AAD account. Ask someone from the team to add you as an administrator. You will need to accept this request once it is sent to you
-8. Create a new API key for this release and limit the lifetime so that it won't be re-used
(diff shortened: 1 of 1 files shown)


docs/release.md
```diff
@@ -1,23 +1,54 @@
...
+    2. Ask someone from the team to add you as an administrator on NuGet. You will need to accept this request once it is sent to you via email.
+* NuGet Package Explorer
+    1. Download NuGet Package Explorer from the Windows store.
+* NuGet Command Line Interface
+    1. Install according to [these instructions](https://learn.microsoft.com/en-us/nuget/reference/nuget-exe-cli-reference).
+
+## Directions
+
+### Initiate the Release
+1. Increment the version number and update the Public API files; both of these steps can usually be accomplished by running the `tools\IncrementVersion.cmd` script. Then create a pull request with these changes and merge it.
```
corranrogue9, line 15: I'd prefer we not say "usually" here. If the script doesn't work, there's a bug in the script, and whoever discovers the bug should probably fix it, or file a bug. Both of those cases are inherent to all of the tools. We aren't saying that "usually" you can verify the nuget package signatures by running `nuget verify`, but if there's a bug in nuget, then nuget won't work in that case (the same way the script here won't work).

docs/release.md
```diff
@@ -1,23 +1,54 @@
...
+2. Kick off a new [nightly build](https://identitydivision.visualstudio.com/OData/_build?definitionId=1104) by running the pipeline to generate the NuGet packages that will be published for this release.
+3. Generate release notes in the [change log](https://github.com/MicrosoftDocs/OData-docs/blob/main/Odata-docs/changelog/odatalib-7x.md) to be published on release of the new version. Then create a pull request with these changes in the [Docs](https://github.com/MicrosoftDocs/OData-docs/) repo and merge it.
+    * Generate new release notes by referencing the PRs that have been merged into `master` since the last version increment. In GitHub, each commit should have a link to the PR that was merged to generate that commit, so you can use GitHub history to generate the changelog.
+
+### Create the NuGet Packages
+4. Download the build artifacts from the nightly build to your local machine.
+    * Access the artifacts by navigating to the "published" items in the "Scheduled" section of the summary page for the new build.
```
corranrogue9, line 24: Totally fine with this being in the document, but I remember while we went through it that I didn't write this down the first time because the UI for the artifacts changes *while I was writing the doc* and I didn't want to write instructions that were going to constantly be obsolete.

docs/release.md
```diff
@@ -1,23 +1,54 @@
...
-9. Verify the signatures of the packages downloaded from the build artifacts by running the following command: `nuget verify -Signature Microsoft.Spatial.7.4.4.nupkg`. Do the same for Microsoft.OData.Edm, Microsoft.OData.Core, and Microsoft.OData.Client
-10. Upload all 4 nupkg files using the nuget.org UI in the following order:
+# Release Procedure
+
+## Prerequisites
+* Administrator account on NuGet
+    1. Log into your nuget.org account and link it to your AAD account by signing in with your Microsoft work account.
+    2. Ask someone from the team to add you as an administrator on NuGet. You will need to accept this request once it is sent to you via email.
+* NuGet Package Explorer
+    1. Download NuGet Package Explorer from the Windows store.
```
corranrogue9, line 8: nit: if you have time, a URL would be nice.

corranrogue9 (dismissed):

docs/release.md
lisicase: Ah I can see that. I might keep this for now since I needed help finding the artifacts, but I'm open to leaving it out as well.

docs/release.md
lisicase: ✅

corranrogue9 (approved):

PR #2602, opened 2023-01-18
fixes #2601: AddDefaultODataServices should pass OData version to low layer

### LINKED ISSUE #2601: AddDefaultODataServices should pass OData version to lower layer

### Assemblies affected

### Reproduce steps

```C#
public static IContainerBuilder AddDefaultODataServices(this IContainerBuilder builder, ODataVersion odataVersion)
```

has the ODataVersion argument.

this ODataVersion is passed to ODataSimplifiedOptions, but it should also be passed to others.

for example: ODataMessageReaderSettings

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

xuzhg:
…er layer

### Issues

*This pull request fixes #2601 .*

### Description

* Pass OData version to lower layer in AddDefaultODataServices

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ContainerBuilderExtensions.cs
+++ b/src/Microsoft.OData.Core/ContainerBuilderExtensions.cs
@@ -136,7 +136,7 @@ namespace Microsoft.OData
             builder.AddService<IJsonWriterFactoryAsync>(ServiceLifetime.Singleton, sp => new DefaultJsonWriterFactory());
             builder.AddService(ServiceLifetime.Singleton, sp => ODataMediaTypeResolver.GetMediaTypeResolver(null));
             builder.AddService<ODataMessageInfo>(ServiceLifetime.Scoped);
-            builder.AddServicePrototype(new ODataMessageReaderSettings());
+            builder.AddServicePrototype(new ODataMessageReaderSettings(odataVersion));
             builder.AddService(ServiceLifetime.Scoped, sp => sp.GetServicePrototype<ODataMessageReaderSettings>().Clone());
             builder.AddServicePrototype(new ODataMessageWriterSettings());
             builder.AddService(ServiceLifetime.Scoped, sp => sp.GetServicePrototype<ODataMessageWriterSettings>().Clone());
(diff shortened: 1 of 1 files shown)


gathogojr (dismissed):

src/Microsoft.OData.Core/ContainerBuilderExtensions.cs
```diff
@@ -136,7 +136,7 @@ public static IContainerBuilder AddDefaultODataServices(this IContainerBuilder b
             builder.AddService<IJsonWriterFactoryAsync>(ServiceLifetime.Singleton, sp => new DefaultJsonWriterFactory());
             builder.AddService(ServiceLifetime.Singleton, sp => ODataMediaTypeResolver.GetMediaTypeResolver(null));
             builder.AddService<ODataMessageInfo>(ServiceLifetime.Scoped);
-            builder.AddServicePrototype(new ODataMessageReaderSettings());
+            builder.AddServicePrototype(new ODataMessageReaderSettings(odataVersion));
             builder.AddService(ServiceLifetime.Scoped, sp => sp.GetServicePrototype<ODataMessageReaderSettings>().Clone());
             builder.AddServicePrototype(new ODataMessageWriterSettings());
```
ElizabethOkerio, line 141: should it be passed to the writer settings too?

xuzhg: ODataMessageWriterSettings doesn't have a constructor to accept ODataVersion.

ElizabethOkerio: How will the version be passed to the writer settings?

ElizabethOkerio (dismissed):

src/Microsoft.OData.Core/ContainerBuilderExtensions.cs
ElizabethOkerio: or why does the reader settings have a constructor that accepts an ODataVersion but the writer settings doesn't.

KenitoInc (dismissed):

src/Microsoft.OData.Core/ContainerBuilderExtensions.cs
xuzhg: I have the same concern. Maybe @mikepizzo  has some input.

gathogojr, line 141: @xuzhg Can we do it this way for `ODataMessageWriterSettings`?
```suggestion
            builder.AddServicePrototype(new ODataMessageWriterSettings { Version = odataVersion });
```

PR #2605, opened 2023-01-24
Capability to close ODataBinaryStreamValue stream after serialization

habbes:
*Note*: This is a duplicate of PR https://github.com/OData/odata.net/pull/2597. Recreated here because the build was not running on the other PR.

### Issues

#2596 

### Description

Add a capability for an `ODataBinaryStreamValue` to have its `Stream` disposed by ODL.
- This behavior is optional and is disabled by default for backward compatibility.
- Implemented via an internal property (`bool LeaveOpen`) rather than making `ODataBinaryStreamValue` an `IDisposable` in order to avoid a breaking change to the public API.

- [x] *Test cases added*


habbes:
Closing this because the original PR #2597 was merged

PR #2610, opened 2023-01-26
Create cache to speed up case-insensitive look up of schema elements

### LINKED ISSUE #2534: Microsoft.OData.UriParser.ODataUriResolver!FindSchemaElements running hot on CPU
This callstack is taking up to 6% on CPU, in many regions.

We already did meaningful improvements to this code path in OData but it is still too expensive. There seems to be a lot of room to reduce CPU from this callstack. Are we doing O(n) searches in FindAcrossModels that can be cached and become O(1) operations?

This callstack has huge potential to cut down CPU and reflect to COGs (i.e., running less number of VMs for the same traffic).

image

### Assemblies affected

Microsoft.OData.Core lib 7.12.2

### Reproduce steps

A variety of requests that all go through Microsoft.OData.UriParser.ODataUriParser!ParsePath

### Expected result

Using much less CPU to parse requests

### Actual result

Using up to 6% of CPU to parse requests

cc: @habbes

habbes:
### Issues

*This pull request fixes #2534 

### Description

I created a cache called `NormalizedSchemaElementsCache` which caches all the schema elements of an `IEdmModel` using normalized uppercase qualified names as keys. I use this cache to look up elements in the `ODataUriResolver` if the case-insensitive mode is enabled. This saves us from having to do a full model scan each time a name that doesn't match a declared name is used.

For this cache to be effective, it has to be built up front, i.e., populate the cache with all the model elements at once, then use it as a read-only cache from that point forward. Because we populate the cache only when it's created, then it's important that no new entries are added/removed/modified in the model. That's why I only use the cache if the model is marked as immutable.

Populating the cache up front means that if the element is not in the cache, we can conclude that it doesn't exist without having to perform any additional searches. If we did not populate the cache up front, we would not be able to make this assumption: it could be that the element is in the model but has not been queried before, or the element does not exist. If the model does not exist, we'd have to pick between two undesirable options:
- keep track of keys that don't exist. This could lead to an infinite number of entries, and we would eventually run out of memory and crash.
- don't keep track of keys that don't exist, but we'd have to do a full model search because we can't distinguish between a key that doesn't exist and an existing key that hasn't been visited before.

I considered introducing a new method that the customer can call to activate/build the cache, but I decided against that because I didn't want to add new methods to the public API. With the current implementation, the cache is triggered automatically if the customer is using `model.MarkAsImmutable()`. But this design choice came with a cost: If two or more threads concurrently find out that the cache has not yet been created, they will all build the cache, but only the last one to attach it to the model will come into effect. This doesn't affect the correctness of the cache since the caches should be identical (the model is immutable). But it is wasteful. But it only affects the first N concurrent requests, after that it's no longer an issue. If this is a major concern, I can refactor it. Or we can revisit it if it turns out to be a problem in the real world.

**Benchmark results**

The first entry does not use the cache. The second entry uses the cache.

 |                                    Method |      Mean |    Error |   StdDev |      Gen 0 |     Gen 1 | Gen 2 | Allocated |
------------------------------------------- |----------:|---------:|---------:|-----------:|----------:|------:|----------:|
|     ParsePathCaseInsensitiveWithMutableModel | 148.36 ms | 0.296 ms | 0.277 ms | 13500.0000 |         - |     - |  81.33 MB |
| ParsePathCaseInsensitiveWithImmutableModel |  99.38 ms | 0.086 ms | 0.081 ms | 11333.3333 |         - |     - |  67.1 MB |

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
@@ -77,7 +77,7 @@ namespace Microsoft.OData.Edm
                 }
             }
 
-            return results as IReadOnlyList<T>;
+            return results;
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
@@ -32,7 +32,7 @@ namespace Microsoft.OData.Edm
             this.cache = cache;
         }
 
-        public IReadOnlyList<IEdmSchemaElement> FindElement(string qualifiedName)
+        public List<IEdmSchemaElement> FindElement(string qualifiedName)
         {
             if (cache.TryGetValue(qualifiedName.ToUpperInvariant(), out List<IEdmSchemaElement> results))
             {
--- a/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
@@ -16,7 +16,7 @@ namespace Microsoft.OData.Edm
     /// it doesn't exist in the model. For this reason, it's important that
     /// the model be immutable.
     /// </summary>
-    internal class CaseInsensitiveSchemaElementsCache
+    internal sealed class CaseInsensitiveSchemaElementsCache
     {
         private static readonly List<IEdmSchemaElement> emptyList = new List<IEdmSchemaElement>();
         // This cache is meant to be populate up front and remain read-only
@@ -68,7 +68,7 @@ namespace Microsoft.OData.Edm
         {
             IReadOnlyList<IEdmSchemaElement> elements = FindElements(qualifiedName);
 
-            IList<T> results = new List<T>();
+            List<T> results = new List<T>();
             for (int i = 0; i < elements.Count; i++)
             {
                 if (elements[i] is T element)
--- a/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
@@ -4,9 +4,9 @@
 // </copyright>
 //---------------------------------------------------------------------
 
-using Microsoft.OData.Edm.Vocabularies;
 using System;
 using System.Collections.Generic;
+using Microsoft.OData.Edm.Vocabularies;
 
 namespace Microsoft.OData.Edm
 {
@@ -19,11 +19,9 @@ namespace Microsoft.OData.Edm
     /// </summary>
     internal sealed class CaseInsensitiveSchemaElementsCache
     {
-        private static readonly List<IEdmSchemaElement> emptyList = new List<IEdmSchemaElement>();
-        // This cache is meant to be populate up front and remain read-only
-        // after that. Therefore, it doesn't need
-        // to be a concurrent dictionary.
-        private readonly Dictionary<string, List<IEdmSchemaElement>> cache = new Dictionary<string, List<IEdmSchemaElement>>(StringComparer.OrdinalIgnoreCase);
+        // We create different caches for different types of schema elements because all current usage request schema elements
+        // of specific types. If we were to use a single dictionary <string, ISchemaElement> we would need
+        // to do additional work (and allocations) during lookups to filter the results to the susbset that matches the request type.
         private readonly Dictionary<string, List<IEdmSchemaType>> schemaTypesCache = new Dictionary<string, List<IEdmSchemaType>>(StringComparer.OrdinalIgnoreCase);
         private readonly Dictionary<string, List<IEdmOperation>> operationsCache = new Dictionary<string, List<IEdmOperation>>(StringComparer.OrdinalIgnoreCase);
         private readonly Dictionary<string, List<IEdmTerm>> termsCache = new Dictionary<string, List<IEdmTerm>>(StringComparer.OrdinalIgnoreCase);
@@ -35,106 +33,57 @@ namespace Microsoft.OData.Edm
         /// <param name="model">The model whose schema elements to cache. This model should be immutable. See <see cref="ExtensionMethods.MarkAsImmutable(IEdmModel)"/>.</param>
         public CaseInsensitiveSchemaElementsCache(IEdmModel model)
         {
-            //Dictionary<string, List<IEdmSchemaElement>> cache = new Dictionary<string, List<IEdmSchemaElement>>(StringComparer.OrdinalIgnoreCase);
-
             PopulateSchemaElements(model);
 
             foreach (IEdmModel referencedModel in model.ReferencedModels)
             {
                 PopulateSchemaElements(referencedModel);
             }
-
-            //this.cache = cache;
         }
 
         /// <summary>
-        /// Find all schema elements that match the specified <paramref name="qualifiedName"/>.
+        /// Find all schema types that match the <paramref name="qualifiedName"/>.
         /// </summary>
         /// <param name="qualifiedName">The case-insensitive fully qualified name to match.</param>
-        /// <returns>The list of elements that matched the <paramref name="qualifiedName"/>.</returns>
-        public IReadOnlyList<IEdmSchemaElement> FindElements(string qualifiedName)
+        /// <returns>A list of matching schema types, or null or no schema type matches.</returns>
+        public List<IEdmSchemaType> FindSchemaTypes(string qualifiedName)
         {
-            if (cache.TryGetValue(qualifiedName, out List<IEdmSchemaElement> results))
+            if (schemaTypesCache.TryGetValue(qualifiedName, out List<IEdmSchemaType> results))
             {
                 return results;
             }
 
-            return emptyList;
+            return null;
         }
 
         /// <summary>
-        /// Find all schema elements of type <typeparamref name="T"/> that match the specified <paramref name="qualifiedName"/>.
+        /// Find all operations that match the <paramref name="qualifiedName"/>.
         /// </summary>
-        /// <typeparam name="T">The type of schema elements to match.</typeparam>
         /// <param name="qualifiedName">The case-insensitive fully qualified name to match.</param>
-        /// <returns>The list of elements of type <typeparamref name="T"/> that matched the <paramref name="qualifiedName"/>.</returns>
-        public IReadOnlyList<T> FindElementsOfType<T>(string qualifiedName) where T : IEdmSchemaElement
+        /// <returns>A list of matching schema types, or null if no operation matches.</returns>
+        public List<IEdmOperation> FindOperations(string qualifiedName)
         {
-            IReadOnlyList<IEdmSchemaElement> elements = FindElements(qualifiedName);
-
-            List<T> results = new List<T>();
-            for (int i = 0; i < elements.Count; i++)
+            if (operationsCache.TryGetValue(qualifiedName, out var results))
             {
-                if (elements[i] is T element)
-                {
-                    results.Add(element);
-                }
+                return results;
             }
 
-            return results;
+            return null;
         }
 
         /// <summary>
-        /// Find a unique element of type <typeparamref name="T"/> that matches the <paramref name="qualifiedName"/>.
-        /// And exception is thrown if duplicates are found.
+        /// Find all vocabulary terms that match the <paramref name="qualifiedName"/>.
         /// </summary>
-        /// <typeparam name="T">The type of element to match.</typeparam>
         /// <param name="qualifiedName">The case-insensitive fully qualified name to match.</param>
-        /// <param name="duplicateErrorFunc">A function that generates an error message if a duplicate is found.</param>
-        /// <returns>The element that was found, or `null` if no match was found.</returns>
-        /// <exception cref="ODataException">Thrown if duplicate matches were found.</exception>
-        public T FindSingleOfType<T>(string qualifiedName, Func<object, string> duplicateErrorFunc) where T : IEdmSchemaElement
+        /// <returns>A list of matching terms, or null if no operation matches.</returns>
+        public List<IEdmTerm> FindTerms(string qualifiedName)
         {
-            IReadOnlyList<IEdmSchemaElement> elements = FindElements(qualifiedName);
-
-            if (elements.Count == 0)
+            if (termsCache.TryGetValue(qualifiedName, out var results))
             {
-                return default;
+                return results;
             }
 
-            T match = default;
-            for (int i = 0; i < elements.Count; i++)
-            {
-                if (elements[i] is T element)
-                {
-                    if (match == null)
-                    {
-                        match = element;
-                    }
-                    else
-                    {
-                        throw new ODataException(duplicateErrorFunc(qualifiedName));
-                    }
-                }
-            }
-
-            return match;
-        }
-        
-        private static void PopulateSchemaElements(IEdmModel model, Dictionary<string, List<IEdmSchemaElement>> cache)
-        {
-            foreach (IEdmSchemaElement element in model.SchemaElements)
-            {
-                string normalizedKey = element.FullName();
-                List<IEdmSchemaElement> results;
-                if (!cache.TryGetValue(normalizedKey, out results))
-                {
-                    results = new List<IEdmSchemaElement>();
-                    cache[normalizedKey] = results;
-                }
-
-                results.Add(element);
-            }
+            return null;
         }
 
         private void PopulateSchemaElements(IEdmModel model)
@@ -153,10 +102,6 @@ namespace Microsoft.OData.Edm
                 {
                     AddElementToCache(term, termsCache);
                 }
-                else
-                {
-                    AddElementToCache(element, cache);
-                }
             }
         }
 
@@ -172,35 +117,5 @@ namespace Microsoft.OData.Edm
 
             results.Add(element);
         }
-
-        public List<IEdmSchemaType> FindSchemaTypes(string qualifiedName)
-        {
-            if (schemaTypesCache.TryGetValue(qualifiedName, out var results))
-            {
-                return results;
-            }
-
-            return null;
-        }
-
-        public List<IEdmOperation> FindOperations(string qualifiedName)
-        {
-            if (operationsCache.TryGetValue(qualifiedName, out var results))
-            {
-                return results;
-            }
-
-            return null;
-        }
-        public List<IEdmTerm> FindTerms(string qualifiedName)
-        {
-            if (termsCache.TryGetValue(qualifiedName, out var results))
-            {
-                return results;
-            }
-
-            return null;
-        }
-
     }
 }
--- a/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
@@ -121,7 +121,7 @@ namespace Microsoft.OData.Edm
         {
             foreach (IEdmSchemaElement element in model.SchemaElements)
             {
-                string normalizedKey = element.FullName().ToUpperInvariant();
+                string normalizedKey = element.FullName();
                 List<IEdmSchemaElement> results;
                 if (!cache.TryGetValue(normalizedKey, out results))
                 {
--- a/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
@@ -31,7 +31,7 @@ namespace Microsoft.OData.Edm
         /// <param name="model">The model whose schema elements to cache. This model should be immutable. See <see cref="ExtensionMethods.MarkAsImmutable(IEdmModel)"/>.</param>
         public CaseInsensitiveSchemaElementsCache(IEdmModel model)
         {
-            Dictionary<string, List<IEdmSchemaElement>> cache = new Dictionary<string, List<IEdmSchemaElement>>();
+            Dictionary<string, List<IEdmSchemaElement>> cache = new Dictionary<string, List<IEdmSchemaElement>>(StringComparer.OrdinalIgnoreCase);
 
             PopulateSchemaElements(model, cache);
 
@@ -50,7 +50,7 @@ namespace Microsoft.OData.Edm
         /// <returns>The list of elements that matched the <paramref name="qualifiedName"/>.</returns>
         public IReadOnlyList<IEdmSchemaElement> FindElements(string qualifiedName)
         {
-            if (cache.TryGetValue(qualifiedName.ToUpperInvariant(), out List<IEdmSchemaElement> results))
+            if (cache.TryGetValue(qualifiedName, out List<IEdmSchemaElement> results))
             {
                 return results;
             }
--- a/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
@@ -4,6 +4,7 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using Microsoft.OData.Edm.Vocabularies;
 using System;
 using System.Collections.Generic;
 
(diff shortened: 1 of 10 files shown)


src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
```diff
@@ -0,0 +1,136 @@
...
+        }
+
+        /// <summary>
+        /// Find all schema elements that match the specified <paramref name="qualifiedName"/>.
+        /// </summary>
+        /// <param name="qualifiedName">The case-insensitive fully qualified name to match.</param>
+        /// <returns>The list of elements that matched the <paramref name="qualifiedName"/>.</returns>
+        public IReadOnlyList<IEdmSchemaElement> FindElements(string qualifiedName)
+        {
+            if (cache.TryGetValue(qualifiedName.ToUpperInvariant(), out List<IEdmSchemaElement> results))
```
syprieur, line 53: Avoid these allocations for each look-up.
Make `cache` use StringComparison.OrdinalIgnoreCase comparison instead.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -530,5 +539,29 @@ internal static ODataUriResolver GetUriResolver(IServiceProvider container)
...
+            if (cache == null)
+            {
+                // There's a chance 2 or more threads can reach here concurrently
+                // for the first N parallel requests, and each will build the cache.
+                // While that is wasteful, it doesn't affect the behaviour of the cache since the model is immutable.
+                // The last cache to be attached to the model will "win" and be used for all subsequent requests.
+                // We can avoid this waste by providing a method that user can call manually to build
+                // the cache before any request is made. But I did not want to add a new method to the public API.
+                // We revisit this if it turns out to be a problem in practice.
+                cache = new CaseInsensitiveSchemaElementsCache(model);
```
syprieur, line 560: Why not do this when the model is marked as immutable?

habbes: I could do that. But that means I'd need to define the cache in the `Microsoft.OData.Edm` and make it public so that it can be accessed from the `Microsoft.OData.Core` lib where the `ODataUriResolver` resides. But maybe that's not as harmful.

src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
habbes: Aah, yeah. Thanks.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
habbes: I'm considering moving this to code to take place when the model is marked as immutable. One downside of that approach compared to the current implementation is that the current implementation only creates the cache if there are actually case-insensitive look ups.

syprieur: That is true.
However, OData services _not in control of **all** clients calling them_ are bound to receive bad requests (or different casing) fairly often, which would trigger creation of the cache.

For Microsoft Graph. this will happen within seconds of start. I do not know enough other OData services to assess whether we're a special case or the norm.

habbes: @syprieur makes sense. I think if case-insensitive is enabled, the cache should be built even before the first request. But my concern is around building the cache even when case-insensitivity is not enabled. It's beneficial to mark a model as immutable even when you don't support case-insensitive URI resolution. When case-insensitive mode is not enabled, this code path is never activated and therefore the cache is not used (I may have to double-check this).

I'm contemplating introducing a method that triggers the building of the cache, that the user has to call. But I think this is not the ideal developer experience, and it introduces a new method to the public API.

syprieur: Right, my bad! Case-sensitivity is an ODataUriParser setting (via ODataUriResolver), it's not tied to an EdmModel.
My suggestion did not make sense, please ignore this thread.

habbes: Apparently WebAPI 8 enable case-insensitivity by default. So, it's reasonable to assume most customers have this enabled. This makes me lean more towards your proposal to have it in `MarkAsImmutable()`.

But before that, I'd like to understand whether the current state of the PR would be a significant concern in AGS. The cache is created on the first case-insensitive lookup. However, multiple threads can reach this point concurrently and each try to build the cache independently. In the end, only one will end up committed to the model and the others should be discarded in the next GC. So, if this issue occurs in practice, it will probably occur within the first few seconds of the service running and then disappear (assuming case-insensitive requests are common). Is this one-time transient cost acceptable in AGS?

src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
```diff
@@ -0,0 +1,136 @@
...
+            }
+
+            return match;
+        }
+        
+        private static void PopulateSchemaElements(IEdmModel model, Dictionary<string, List<IEdmSchemaElement>> cache)
+        {
+            foreach (IEdmSchemaElement element in model.SchemaElements)
+            {
+                string normalizedKey = element.FullName().ToUpperInvariant();
```
syprieur, line 124: Remove the `ToUpperInvariant` allocations: the cache dictionary now uses `OrdinalIgnoreCase` comparison so it's not needed.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -504,8 +507,14 @@ internal static ODataUriResolver GetUriResolver(IServiceProvider container)
             return container.GetRequiredService<ODataUriResolver>();
         }
 
-        private static IList<T> FindAcrossModels<T>(IEdmModel model, String qualifiedName, bool caseInsensitive) where T : IEdmSchemaElement
+        private static IReadOnlyList<T> FindAcrossModels<T>(IEdmModel model, String qualifiedName, bool caseInsensitive) where T : IEdmSchemaElement
```
syprieur, line 510: Beyond the scope of this PR: the private methods `FindAcrossModels` and `FindSchemaElements` don't seem to **need** to return an IList: all their callers pretty much use the return value as an IEnumerable.
As a result, the List allocation could probably be avoided (including in the new CacheInsensitiveSchemaElementsCache.FindElementsOfType).

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
syprieur: It's fine - if we see latency impact we can always make a warm-up junk request after marking the model as immutable.

src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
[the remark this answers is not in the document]
joaocpaiva: Sorry it seems like I forgot how to use github - so just sent my comments now that I had done last week, doh!

src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
habbes: Thanks for the catch. Must have forgotten this.

src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
```diff
@@ -0,0 +1,136 @@
...
+            return emptyList;
+        }
+
+        /// <summary>
+        /// Find all schema elements of type <typeparamref name="T"/> that match the specified <paramref name="qualifiedName"/>.
+        /// </summary>
+        /// <typeparam name="T">The type of schema elements to match.</typeparam>
+        /// <param name="qualifiedName">The case-insensitive fully qualified name to match.</param>
+        /// <returns>The list of elements of type <typeparamref name="T"/> that matched the <paramref name="qualifiedName"/>.</returns>
+        public IReadOnlyList<T> FindElementsOfType<T>(string qualifiedName) where T : IEdmSchemaElement
```
joaocpaiva, line 67: From the code it seems these O(n) searches (**FindElementsOfType**, **FindSingleOfType**) will still happen in AGS, even with the immutable cache? They can't be made O(1)? Or it is not worth it?

habbes: In this case O(n) would be relative to the number of elements that match the same name, rather than the total number of elements in the schema. If the AGS schema doesn't have schema elements with similar names, then this will be close to O(1) since each list would only have a single entry.  We can probably achieve O(1) if we store elements of different types in different collections, for example one dictionary for schema types, one dictionary for operations, etc. so we don't have to filter by type

src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
```diff
@@ -0,0 +1,136 @@
...
+        {
+            IReadOnlyList<IEdmSchemaElement> elements = FindElements(qualifiedName);
+
+            List<T> results = new List<T>();
```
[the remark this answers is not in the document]
habbes: I think in most cases it would not be larger. I assume for most schemas duplicate names will only exist for function overloads. 

Maybe I could also first check whether all the elements are of the same type, and if they are, then just return the original list without allocating a new one. This might require 2 loops, but if the list size is small, it may be worth it.

src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
joaocpaiva: Sounds good. And if these lists are expected to be small (e.g., less than ~4 elements) for most cases, List search can actually beat Dictionary.

@syprieur this should be small lists in AGS scenario?

src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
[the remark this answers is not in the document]
joaocpaiva: I would not add 2 loops I think. 

Either keep the code as is or provide a capacity hint which in this case seems it should be 'elements.Count'. As long as the hint is not expected to be drastically different from the actual number of elements added in the end, providing a good capacity hint is meant to reduce internal allocations overhead to expand the list on insertions.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
habbes: There are methods which index into the results, e.g. `ResolveBoundOperations`. But they do the index lookup in a plan for loop, so this could be achieved with a `foreach` on an `IEnumerable`. But I assume a for loop was chosen to avoid the cost of `foreach` against an `IEnumerable`.

src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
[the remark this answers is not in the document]
habbes: Actually, I can't return an `IReadOnlyList<IEdmSchemaElements>` cast an `IReadOnlyList<T>` where `T` is a subtype, cause of covariance rules. So, if I wanted to completely eliminate this `List<T>` allocation, I would need cache the elements in different collections based on type. Maybe a cache of <string, IEdmSchemaType>, a cache of <string, IEdmOperation>, etc. Do you think that would be worthwhile?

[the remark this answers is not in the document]
habbes: So, if I wanted to completely eliminate this `List<T>` allocation, I would need cache the elements in different collections based on type. Maybe a cache of <string, IEdmSchemaType>, a cache of <string, IEdmOperation>, etc. Do you think that would be worthwhile?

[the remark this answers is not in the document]
habbes: The cost here is that we'd have multiple dictionaries and maybe that could have a larger overhead in some cases.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -514,7 +523,7 @@ internal static ODataUriResolver GetUriResolver(IServiceProvider container)
                 FindSchemaElements<T>(reference, qualifiedName, caseInsensitive, ref results);
             }
 
-            return results;
+            return results as IReadOnlyList<T>;
```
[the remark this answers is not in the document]
habbes: Yeah. And since we're keeping this API internal, we don't expect the caller to make any changes. I can refactor to remove the casts.

src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
habbes: For smaller models, with no name collisions, I think the extra dictionaries might cost more overhead. However, the concern is with the `List<T>` allocations in `FindElementsOfType<T>`. I try to optimize the case where only one item is needed with `FindSingleOfType<T>` to be allocation-free in the common case. But if `FindElementsOfType<T>` is still on the hot path, then it may be worthwhile eliminating the `new List<T>`.

habbes: I've created an experiment based on multiple caches (1 cache for schema types, 1 cache for operations, 1 cache for vocabulary terms and 1 cache for every other schema element). So instead of calling `FindElementsOfType<IEdmOperation>`, I'd call `FindOperations` which looks up the cache that stores `IEdmOperation` instances. So, I don't need to create a new `List<T>` each time. On my benchmark (which may or may not not be representative of AGS) memory allocations went slightly down (from 67.41MB to 67.1MB).

habbes:
@joaocpaiva @syprieur I've refactored the code to populate different dictionaries for different types, one for schema types, one for operations and one for vocabulary terms. This is informed by scenarios that called `FindAcrossModels`, in all cases it was looking for elements of a specific types. So instead of making the cache generic, it only caters for these 3 types now. The benefit is that we no longer have to filter the results by type or create a new `List<T>` during cache look ups. For a small model with no name collisions, it's possible there overhead of the dictionaries could be higher, but for cases like AGS, separate dictionaries would probably be more beneficial. In local benchmarks there was slight decrease in memory allocations from 67.41MB to 67.1MB.

I found that the `ResolveBoundOperations` and `ResolveUnboundOperations` methods still allocate lists. It may be possible to do away with some of these by further splitting the operations cache into two, one for bound and one for unbound. Then for unbound we'd also have to group elements by binding type (maybe using a composite key for the dictionary). But it could get a bit complex here since the binding type could have sub types. So, I decided not to handle this complexity for now (unless you think I should).

src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
syprieur: Yes, these Lists will be small for AGS: we enforce FQN unicity on our models (with 1-2 exceptions for legacy APIs), so as Clément stated, we'll only have more than one element for operations overloads.

joaocpaiva (approved):

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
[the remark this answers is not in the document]
joaocpaiva: Nah it is ok, small nitpick. I was just calling out that the object is not really immutable from the caller side, since we are using a simple cast.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -167,7 +164,16 @@ public virtual IEdmTerm ResolveTerm(IEdmModel model, string termName)
                 return term;
             }
 
-            IList<IEdmTerm> results = FindAcrossModels<IEdmTerm>(model, termName, /*caseInsensitive*/ true);
+            IReadOnlyList<IEdmTerm> results;
```
ElizabethOkerio, line 167: this logic is repeated in several places.

habbes: Which block exactly do you think should be turned into a method, is it the following:
```c#
IReadOnlyList<IEdmTerm> results;
if (model.IsImmutable())
{
    CaseInsensitiveSchemaElementsCache cache = GetCaseInsensitiveSchemaElementsCache(model);
    results = cache.FindTerms(termName);
}
else
{
    results = FindAcrossModels<IEdmTerm>(model, termName, /*caseInsensitive*/ true);
}
```

If that's the one, then the reason I didn't refactor it into a method is that doing so would probably introduce unnecessary complexity and overhead. The main issue is that I'm calling a different cache method depending on the type, (i.e. `cache.FindTerms`, `cache.FindOperations`, `cache.FindSchemaTypes`). So to generalize this block of code to support these different cases I would need to pass a delegate to the method and make sure not to capture the `termName` variable. e.g. 

```c#
void CaseInsensitiveFind<T>(IEdmModel model,  string key, Func<Cache, key, List<T> cacheLookupFunc)  where T : IEdmSchemaElement
{
   if (model.IsImmutable())
   {
       CaseInsensitiveSchemaElementsCache cache = GetCaseInsensitiveSchemaElementsCache(model);
       return cacheLookupFunc(cache, key);
   }

   return FindAcrossModels<T>(model, key, /*caseInsensitive*/ true);
}
```
This is not that complex to be honest but given that this block is fairly simple and straightforward, and used in only 4 places, I didn't feel it was worthwhile to refactor in this way.

But it's not a strong opinion. Given this, do you think it's worth going ahead and refactoring it?

joaocpaiva: If we refactor and add a delegate, make sure it is [static](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-9.0/static-anonymous-functions) to not introduce unnecessary delegate or closure allocations on every call.

ElizabethOkerio (dismissed):

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
habbes: I did a refactor by creating static method and making sure the lambda does not capture the outer scope. Benchmark results remain the same.

src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
```diff
@@ -0,0 +1,121 @@
...
+        private readonly Dictionary<string, List<IEdmOperation>> operationsCache = new Dictionary<string, List<IEdmOperation>>(StringComparer.OrdinalIgnoreCase);
+        private readonly Dictionary<string, List<IEdmTerm>> termsCache = new Dictionary<string, List<IEdmTerm>>(StringComparer.OrdinalIgnoreCase);
+
+        /// <summary>
+        /// Builds a case-insensitive cache of schema elements from
+        /// the specified <paramref name="model"/>.
+        /// </summary>
+        /// <param name="model">The model whose schema elements to cache. This model should be immutable. See <see cref="ExtensionMethods.MarkAsImmutable(IEdmModel)"/>.</param>
+        public CaseInsensitiveSchemaElementsCache(IEdmModel model)
+        {
```
gathogojr, line 35: @habbes Should we do a null check for the `model` here? In addition, would this be a good place to check if the model is immutable or would that contribute to unnecessary coupling of things that address separate concerns?

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -504,7 +503,22 @@ internal static ODataUriResolver GetUriResolver(IServiceProvider container)
             return container.GetRequiredService<ODataUriResolver>();
         }
 
-        private static IList<T> FindAcrossModels<T>(IEdmModel model, String qualifiedName, bool caseInsensitive) where T : IEdmSchemaElement
+        private static IReadOnlyList<T> FindElementsCaseInsensitive<T>(
```
gathogojr, line 506:
```suggestion
        private static IReadOnlyList<T> FindElements<T>(
```
Or if you must be explicit about it being a case-insensitive search, `FindElementsIgnoreCase` might be a better fit

habbes: I wasn't sure what call it cause there are other `Find*` methods: `FindAcrossModels` and `FindSchemaElements`. Maybe I should call the new method `FindSchemaElements` and rename the existing `FindSchemaElements` to `FindInModel` or `FindSchemaElementsInModel`.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -530,5 +544,29 @@ internal static ODataUriResolver GetUriResolver(IServiceProvider container)
                 }
             }
         }
+
+        private static CaseInsensitiveSchemaElementsCache GetCaseInsensitiveSchemaElementsCache(IEdmModel model)
+        {
+            if (model == null)
```
gathogojr, line 550: I think a `Debug.Assert` should suffice here since this is a private method. The null check should be in public/internal methods or constructors

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/CaseInsensitiveSchemaElementsCacheTests.cs
```diff
@@ -0,0 +1,166 @@
...
+            model.AddTerm("NS.Models", "SomeTerm", EdmPrimitiveTypeKind.String);
+            model.AddTerm("NS.Models", "OtherTerm", EdmPrimitiveTypeKind.String);
+
+            var cache = new CaseInsensitiveSchemaElementsCache(model);
+
+            var matches = cache.FindTerms("ns.models.some");
+            Assert.Null(matches);
+        }
+    }
+}
```
gathogojr, line 166: Add a test for a scenario where the model passed to the `CaseInsensitiveSchemaElementsCache` constructor is null

src/Microsoft.OData.Core/UriParser/Resolver/CaseInsensitiveSchemaElementsCache.cs
habbes: I already check whether the model is immutable in the caller, before calling in the constructor. So I reasoned it would be of use to call it again in the constructor because this is an internal class, it would add error-handling complexity and I didn't want to call `IsImmutable()` twice in the same code path because there's some cost associated with looking up the model's annotations (which is where we store whether or not it's immutable)

habbes: I think we could do a `Debug.Assert`. This is an internal class that shouldn't be called if the model is null. If it does get called with null, then something else must have gone wrong either in the ODataUriResolver or higher up the chain.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
gathogojr: Don't forget the method is private...

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -530,5 +545,26 @@ internal static ODataUriResolver GetUriResolver(IServiceProvider container)
                 }
             }
         }
+
+        private static NormalizedSchemaElementsCache GetCaseInsensitiveSchemaElementsCache(IEdmModel model)
```
gathogojr, line 549:
```suggestion
        private static NormalizedSchemaElementsCache GetNormalizedSchemaElementsCache(IEdmModel model)

```

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -504,20 +504,35 @@ internal static ODataUriResolver GetUriResolver(IServiceProvider container)
...
+        {
+            if (model.IsImmutable())
+            {
+                NormalizedSchemaElementsCache cache = GetCaseInsensitiveSchemaElementsCache(model);
+                return cacheLookupFunc(cache, identifier);
+            }
+
+            return FindAcrossModels<T>(model, identifier, caseInsensitive: true);
+
+        }
```
gathogojr, line 520:
```suggestion
            return FindAcrossModels<T>(model, identifier, caseInsensitive: true);
        }
```

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
habbes: I renamed the `FindElementsCaseInsensitive` to `FindSchemaElements` and `FindSchemaElements` to `FindSchemaElementsInModel`

gathogojr (approved):

joaocpaiva (approved):

syprieur:
[ODataUriResolver.ResolveNavigationSource](https://github.com/OData/odata.net/pull/2610/files#diff-c162edb909ae9d17f81555ce76d0b886ded378567ae37b987f5863ec786d9088R101): still doing a linear search over **all** schema elements by design?

Even if the number of container elements is low (~120 for MSGraph beta at this moment), it's still a lot of Linq typecasts over all other schema elements.

PR #2618, opened 2023-02-17
support reading relative context urls with bulk updates

ElizabethOkerio:
### Issues

This PR adds support for reading payloads with relative context URLs.  Currently we support `@context` with absolute URLS:  
EXA of a payload with a relative context URL that will be supported by these changes:
```json
{
    "@context": "#$delta",
    "value": [
        {
            "displayName": "Big cat",
            "percentageWeight": 100
        }
    ]
}
```

```json
{
    "@context": "Products",
    "value": [
        {
            "displayName": "Big cat",
            "percentageWeight": 100
        }
    ]
}
```
```json
{
    "@context": "$metadata#Products",
    "value": [
        {
            "displayName": "Big cat",
            "percentageWeight": 100
        }
    ]
}
```

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
@@ -71,19 +71,55 @@ namespace Microsoft.OData.JsonLight
             ODataPayloadKind payloadKind,
             Func<IEdmType, string, IEdmType> clientCustomTypeResolver,
             bool needParseFragment,
-            bool throwIfMetadataConflict = true)
+            bool throwIfMetadataConflict = true,
+            Uri baseUri = null,
+            IEdmNavigationSource navigationSource = null)
         {
             if (contextUriFromPayload == null)
             {
                 throw new ODataException(ODataErrorStrings.ODataJsonLightContextUriParser_NullMetadataDocumentUri);
             }
 
-            // Create an absolute URI from the payload string
-            // TODO: Support relative context uri and resolving other relative uris
+            // Create a context URI from the payload string.
             Uri contextUri;
             if (!Uri.TryCreate(contextUriFromPayload, UriKind.Absolute, out contextUri))
             {
-                throw new ODataException(ODataErrorStrings.ODataJsonLightContextUriParser_TopLevelContextUrlShouldBeAbsolute(contextUriFromPayload));
+                if (baseUri == null)
+                {
+                    throw new ODataException(ODataErrorStrings.ODataJsonLightContextUriParser_InvalidContextUrl(contextUriFromPayload));
+                }
+                else
+                {
+                    ODataUri oDataUri = new ODataUri() { ServiceRoot = baseUri };
+
+                    // This caters for this format: #$delta.
+                    if (contextUriFromPayload == ODataConstants.HashDeltaResourceSet)
+                    {
+                        if (string.IsNullOrEmpty(navigationSource?.Name))
+                        {
+                            throw new ODataException(ODataErrorStrings.ODataJsonLightContextUriParser_InvalidContextUrl(contextUriFromPayload));
+                        }
+                        else
+                        {
+                            contextUriFromPayload = oDataUri.MetadataDocumentUri.ToString() + 
+                                ODataConstants.ContextUriFragmentIndicator +
+                                navigationSource.Name +
+                                ODataConstants.UriSegmentSeparator +
+                                ODataConstants.DeltaResourceSet;
+                        }      
+                    }
+                    else
+                    {
+                        contextUriFromPayload = contextUriFromPayload.StartsWith(ODataConstants.UriMetadataSegmentHash, StringComparison.Ordinal)
+                        ? baseUri + contextUriFromPayload
+                        : oDataUri.MetadataDocumentUri.ToString() + ODataConstants.ContextUriFragmentIndicator + contextUriFromPayload;
+                    }
+
+                    if (!Uri.TryCreate(baseUri, contextUriFromPayload, out contextUri))
+                    {
+                        throw new ODataException(ODataErrorStrings.ODataJsonLightContextUriParser_InvalidContextUrl(contextUriFromPayload));
+                    }
+                }         
             }
 
             ODataJsonLightContextUriParser parser = new ODataJsonLightContextUriParser(model, contextUri);
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
@@ -258,7 +258,8 @@ namespace Microsoft.OData.JsonLight
             ODataPayloadKind payloadKind,
             PropertyAndAnnotationCollector propertyAndAnnotationCollector,
             bool isReadingNestedPayload,
-            bool allowEmptyPayload)
+            bool allowEmptyPayload,
+            IEdmNavigationSource navigationSource = null)
         {
             this.JsonReader.AssertNotBuffering();
             Debug.Assert(isReadingNestedPayload || this.JsonReader.NodeType == JsonNodeType.None, "Pre-Condition: JSON reader must not have been used yet when not reading a nested payload.");
@@ -298,7 +299,9 @@ namespace Microsoft.OData.JsonLight
                     payloadKind,
                     this.MessageReaderSettings.ClientCustomTypeResolver,
                     this.JsonLightInputContext.ReadingResponse || payloadKind == ODataPayloadKind.Delta,
-                    this.JsonLightInputContext.MessageReaderSettings.ThrowIfTypeConflictsWithMetadata);
+                    this.JsonLightInputContext.MessageReaderSettings.ThrowIfTypeConflictsWithMetadata,
+                    this.BaseUri, 
+                    navigationSource);
             }
 
             this.contextUriParseResult = parseResult;
@@ -325,7 +328,8 @@ namespace Microsoft.OData.JsonLight
             ODataPayloadKind payloadKind,
             PropertyAndAnnotationCollector propertyAndAnnotationCollector,
             bool isReadingNestedPayload,
-            bool allowEmptyPayload)
+            bool allowEmptyPayload, 
+            IEdmNavigationSource navigationSource = null)
         {
             this.JsonReader.AssertNotBuffering();
             Debug.Assert(
@@ -349,7 +353,9 @@ namespace Microsoft.OData.JsonLight
                     payloadKind,
                     this.MessageReaderSettings.ClientCustomTypeResolver,
                     this.JsonLightInputContext.ReadingResponse || payloadKind == ODataPayloadKind.Delta,
-                    this.JsonLightInputContext.MessageReaderSettings.ThrowIfTypeConflictsWithMetadata);
+                    this.JsonLightInputContext.MessageReaderSettings.ThrowIfTypeConflictsWithMetadata,
+                    this.BaseUri, 
+                    navigationSource);
             }
 
 #if DEBUG
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightContextUriParserTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightContextUriParserTests.cs
@@ -31,13 +31,29 @@ namespace Microsoft.OData.Tests.JsonLight
             return model;
         }
 
-        // TODO: Support relative context uri and resolving other relative uris
         [Fact]
-        public void ParseRelativeContextUrlShouldThrowException()
+        public void ParseRelativeContextUrlShouldNotThrowException()
         {
-            string relativeUrl = "$metadata#R";
-            Action parseContextUri = () => ODataJsonLightContextUriParser.Parse(new EdmModel(), relativeUrl, ODataPayloadKind.Unsupported, null, true);
-            parseContextUri.Throws<ODataException>(ErrorStrings.ODataJsonLightContextUriParser_TopLevelContextUrlShouldBeAbsolute(relativeUrl));
+            string relativeUrl = "$metadata#People";
+            var parsedContextUrl = ODataJsonLightContextUriParser.Parse(this.GetModel(), relativeUrl, ODataPayloadKind.Unsupported, null, true, true, new Uri("https://www.example.com/api/"));
+            Assert.Equal(new Uri("https://www.example.com/api/$metadata#People"), parsedContextUrl.ContextUri);
+        }
+
+        [Fact]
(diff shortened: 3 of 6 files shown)


xuzhg (approved):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
```diff
@@ -71,19 +71,44 @@ private ODataJsonLightContextUriParser(IEdmModel model, Uri contextUriFromPayloa
             ODataPayloadKind payloadKind,
             Func<IEdmType, string, IEdmType> clientCustomTypeResolver,
             bool needParseFragment,
-            bool throwIfMetadataConflict = true)
+            bool throwIfMetadataConflict = true,
+            Uri baseUri = null,
+            string navigationSource = null)
```
gathogojr, line 76: Why not make this `IEdmNavigationSource navigationSource` instead of `string navigationSource`. You'd avoid doing `this.CurrentNavigationSource?.Name` is so many place, plus I believe it's the better option

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
```diff
@@ -71,19 +71,44 @@ private ODataJsonLightContextUriParser(IEdmModel model, Uri contextUriFromPayloa
...
+            // Create a context URI from the payload string.
             Uri contextUri;
             if (!Uri.TryCreate(contextUriFromPayload, UriKind.Absolute, out contextUri))
             {
-                throw new ODataException(ODataErrorStrings.ODataJsonLightContextUriParser_TopLevelContextUrlShouldBeAbsolute(contextUriFromPayload));
+                if (baseUri != null)
+                {
+                    ODataUri oDataUri = new ODataUri() { ServiceRoot = baseUri };
+
+                    // Graph requires this format for delta request body context.
```
gathogojr, line 91: Does it mean this code caters to MS Graph's unique requirement? Can it be generalized such that it will work for different categories of customers?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
```diff
@@ -71,19 +71,44 @@ private ODataJsonLightContextUriParser(IEdmModel model, Uri contextUriFromPayloa
...
+                    if (contextUriFromPayload.Equals("#$delta", StringComparison.OrdinalIgnoreCase))
+                    {
+                        contextUriFromPayload = oDataUri.MetadataDocumentUri.ToString() + ODataConstants.ContextUriFragmentIndicator + navigationSource + ODataConstants.UriSegmentSeparator + "$delta";
```
gathogojr, line 94: What happens here if `navigationSource` is null?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
```diff
@@ -71,19 +71,44 @@ private ODataJsonLightContextUriParser(IEdmModel model, Uri contextUriFromPayloa
...
             Uri contextUri;
             if (!Uri.TryCreate(contextUriFromPayload, UriKind.Absolute, out contextUri))
             {
-                throw new ODataException(ODataErrorStrings.ODataJsonLightContextUriParser_TopLevelContextUrlShouldBeAbsolute(contextUriFromPayload));
+                if (baseUri != null)
+                {
+                    ODataUri oDataUri = new ODataUri() { ServiceRoot = baseUri };
+
+                    // Graph requires this format for delta request body context.
+                    if (contextUriFromPayload.Equals("#$delta", StringComparison.OrdinalIgnoreCase))
```
gathogojr, line 92: Don't we have a constant for `$delta`? Can we use it here?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
gathogojr, line 94: Don't we have a constant for `$delta`? Shouldn't we introduce it if it doesn't exist and use it here?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
```diff
@@ -71,19 +71,44 @@ private ODataJsonLightContextUriParser(IEdmModel model, Uri contextUriFromPayloa
...
+                    {
+                        contextUriFromPayload = oDataUri.MetadataDocumentUri.ToString() + ODataConstants.ContextUriFragmentIndicator + navigationSource + ODataConstants.UriSegmentSeparator + "$delta";
+                    }
+                    else
+                    {
+                        contextUriFromPayload = contextUriFromPayload.StartsWith("$metadata#", StringComparison.OrdinalIgnoreCase)
```
gathogojr, line 98: Do we have a constant for `$metadata`?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -258,7 +258,8 @@ internal static bool TryParsePropertyAnnotation(string propertyAnnotationName, o
             ODataPayloadKind payloadKind,
             PropertyAndAnnotationCollector propertyAndAnnotationCollector,
             bool isReadingNestedPayload,
-            bool allowEmptyPayload)
+            bool allowEmptyPayload,
+            string navigationSource = null)
```
gathogojr, line 262: Why not make this `IEdmNavigationSource navigationSource` instead of `string navigationSource`.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightDeserializer.cs
```diff
@@ -325,7 +327,8 @@ internal static bool TryParsePropertyAnnotation(string propertyAnnotationName, o
             ODataPayloadKind payloadKind,
             PropertyAndAnnotationCollector propertyAndAnnotationCollector,
             bool isReadingNestedPayload,
-            bool allowEmptyPayload)
+            bool allowEmptyPayload, 
+            string navigationSource = null)
```
gathogojr, line 331: Why not make this `IEdmNavigationSource navigationSource` instead of `string navigationSource`.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightContextUriParserTests.cs
```diff
@@ -31,21 +31,28 @@ private EdmModel GetModel()
...
+            parseContextUri.DoesNotThrow();
         }
 
 
 #if NETCOREAPP2_0
         [Fact]
         public void ParseContextUrlWithEscapedSpecailMeaningCharactersShouldSucceed()
         {
-            string urlWithUnescapedSpecialMeaningCharacters = "https://www.example.com/api/$metadata#people('i%3A0%23.f%7Cmembership%7Cexample%40example.org')/Dogs";
+            string urlWithUnescapedSpecialMeaningCharacters = "$metadata#people('i%3A0%23.f%7Cmembership%7Cexample%40example.org')/Dogs";
```
gathogojr, line 55: Why change this to test the relative URL scenario? Instead you should parameterize this test such that it covers both the scenarios where the URL is absolute and relative...

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
ElizabethOkerio: Or did you mean not including that comment?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
ElizabethOkerio: the contextUriFromPayload will be `#$delta`..This means we'll have to do some concatenations before doing the check.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
ElizabethOkerio: fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
ElizabethOkerio: an exception should be thrown

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
ElizabethOkerio: I believe this will include a lot of concatenations before doing the check

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightContextUriParserTests.cs
ElizabethOkerio: I don't remember making this change. Let me fix it.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
ElizabethOkerio: fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
ElizabethOkerio: fixed

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
```diff
@@ -71,19 +71,51 @@ private ODataJsonLightContextUriParser(IEdmModel model, Uri contextUriFromPayloa
...
+                    if (contextUriFromPayload.Equals(ODataConstants.HashDeltaResourceSet, StringComparison.OrdinalIgnoreCase))
+                    {
+                        if (string.IsNullOrEmpty(navigationSource?.Name))
+                        {
+                            throw new ODataException(ODataErrorStrings.ODataJsonLightContextUriParser_InvalidContextUrl(contextUriFromPayload));
+                        }
+                        else
+                        {
+                            contextUriFromPayload = oDataUri.MetadataDocumentUri.ToString() + ODataConstants.ContextUriFragmentIndicator + navigationSource.Name + ODataConstants.UriSegmentSeparator + ODataConstants.DeltaResourceSet;
+                        }
```
habbes, line 101: The `else` keyword isn't necessary because the method execution will stop if the exception in the if statement is thrown. Consider removing the unnecessary indentation:
```suggestion
                        if (string.IsNullOrEmpty(navigationSource?.Name))
                        {
                            throw new ODataException(ODataErrorStrings.ODataJsonLightContextUriParser_InvalidContextUrl(contextUriFromPayload));
                        }
                        
                        contextUriFromPayload = oDataUri.MetadataDocumentUri.ToString() + ODataConstants.ContextUriFragmentIndicator + navigationSource.Name + ODataConstants.UriSegmentSeparator + ODataConstants.DeltaResourceSet;

```

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
```diff
@@ -71,19 +71,51 @@ private ODataJsonLightContextUriParser(IEdmModel model, Uri contextUriFromPayloa
...
             Uri contextUri;
             if (!Uri.TryCreate(contextUriFromPayload, UriKind.Absolute, out contextUri))
             {
-                throw new ODataException(ODataErrorStrings.ODataJsonLightContextUriParser_TopLevelContextUrlShouldBeAbsolute(contextUriFromPayload));
+                if (baseUri != null)
+                {
+                    ODataUri oDataUri = new ODataUri() { ServiceRoot = baseUri };
+
+                    // This caters for this format: #$delta.
+                    if (contextUriFromPayload.Equals(ODataConstants.HashDeltaResourceSet, StringComparison.OrdinalIgnoreCase))
```
habbes, line 92: Consider checking for the error case first to avoid unnecessary indentation:

Instead of

```c#
if (baseUri != null)
{
    // do stuff
}
else
{
    throw ...
}
```
Consider:
```c#
if (baseUri == null)
{
    throw ...;
}

// do stuff
```

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightContextUriParserTests.cs
```diff
@@ -31,13 +31,21 @@ private EdmModel GetModel()
...
+            parseContextUri.DoesNotThrow();
+        }
+
+        [Fact]
+        public void ParseRelativeContextUrlWithOnlyDeltaSegmentShouldNotThrowException()
+        {
+            string relativeUrl = "#$delta";
+            IEdmNavigationSource navigationSource = this.GetModel().FindDeclaredEntitySet("People") as IEdmNavigationSource;
+            Action parseContextUri = () => ODataJsonLightContextUriParser.Parse(this.GetModel(), relativeUrl, ODataPayloadKind.Unsupported, null, true, true, new Uri("https://www.example.com/api"), navigationSource);
+            parseContextUri.DoesNotThrow();
```
habbes, line 48: Add assertions to verify that full context URI was generated as expected.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightContextUriParserTests.cs
```diff
@@ -31,13 +31,21 @@ private EdmModel GetModel()
             return model;
         }
 
-        // TODO: Support relative context uri and resolving other relative uris
         [Fact]
-        public void ParseRelativeContextUrlShouldThrowException()
+        public void ParseRelativeContextUrlShouldNotThrowException()
```
habbes, line 35: In the PR description you mention that relative context URLs of the form: `@context: "Products"` should be supported, but there are no tests to verify this scenario.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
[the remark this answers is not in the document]
ElizabethOkerio: Leaving this one out. It should not be case-insensitive.

xuzhg (approved):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs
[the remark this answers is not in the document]
ElizabethOkerio: I'm not sure this is possible. Like we do away with the baseURL?

KenitoInc (approved):

PR #2619, opened 2023-02-17
Add support for writing JsonElement values in ODataWriter

habbes:
### Issues

This PR allows `ODataWriter` to write `JsonElement` values directly. This is useful in cases where then input payload is already in OData JSON format. In such cases you may want to adjust some of the input but keep most of it intact. The problem is that if you parse JSON input, then convert it to `ODataProperty/ODataValue` items, then write them out using `ODataMessageWriter`, you'd end up allocating JSON content multiple times, which can be costly.

Here's what the scenario that this feature enables:

- Assume workload response as a stream of size X bytes. 
- Stream is parsed to `JsonDocument` in `System.Text.Json` lib. 
- This new object is an index on top of memory stream. So extra X bytes are not allocated. 
- From this `JsonDocument`, `ODataEntry`/`ODataProperty` is created and sent to `ODataWriter`. 

### Description

This PR implements the following changes:
- Introduces new `ODataJsonElementValue` subclass of `ODataValue` that wraps a `JsonElement`
- Introduces a new `WriteValue(JsonElement` overload in `IJsonWriter` (as well as the async version)
  - The `ODataUtf8JsonWriter` implementation uses `JsonElement.WriteTo(Utf8JsonWriter)` method to write the payload effeiciently
  - The `JsonWriter` implementation reads the `JsonElement` and writes using the existing `WriteValue()` methods based on the data type. This is less efficient and results in double allocation, so this feature should ideally only be used with the `ODataUtf8JsonWriter`
  - Updated `ODataJsonLightPropertySerializer` to properly handle `ODataJsonElementValue`.

Here's how the feature looks like in action:

```c#
string jsonString = @"{""Name"": ""John Doe"", ""Address"": { ""Foo"": ""Bar"" } }";
var jsonDoc = JsonDocument.Parse(jsonString);

List<ODataProperty> properties = new List<ODataProperty>();

foreach (var property in jsonDoc.RootElement.EnumerateObject())
{
    properties.Add(new ODataProperty()
    {
        Name = property.Name,
        Value = new ODataJsonElementValue(property.Value)
    });
}

var resource = new ODataResource { Properties = properties };

messageWriter.WriteStart(resource);
messageWriter.WriteEnd();
messageWriter.Flush();
```

It's important that we set the property value as

```c#
new ODataProperty { ..., Value = new ODataJsonElementValue(property.Value); 
```

instead of

```c#
new ODataProperty { Value = property.Value };
```

While the two will result have the same functional result, the latter will try to convert the `JsonElement` into an `ODataValue` and cause the `JsonElement` struct to be boxed in the process.

Note that at the moment support for `ODataJsonElementValue` is limited to only property values, this PR doesn't support using `JsonElement` as a top-level resource.

#### Benchmarks

BenchmarkDotNet=v0.12.1, OS=Windows 10.0.19045
.NET Core SDK=7.0.102
  [Host]     : .NET Core 3.1.32 (CoreCLR 4.700.22.55902, CoreFX 4.700.22.56512), X64 RyuJIT
  Job-CXDUTP : .NET Core 3.1.32 (CoreCLR 4.700.22.55902, CoreFX 4.700.22.56512), X64 RyuJIT

InvocationCount=1  UnrollFactor=1

|                                                 Method | enableValidation |      Mean |     Error |    StdDev |    Median |     Gen 0 | Gen 1 | Gen 2 | Allocated |
|------------------------------------------------------- |----------------- |----------:|----------:|----------:|----------:|----------:|------:|------:|----------:|
|     WriteParsedJsonWithJsonElementValues_DefaultWriter |            False |  8.555 ms | 0.1786 ms | 0.4704 ms |  8.498 ms |         - |     - |     - |   3.07 MB |
|    WriteParsedJsonWithJsonElementValues_Utf8JsonWriter |            False |  6.628 ms | 0.2292 ms | 0.6077 ms |  6.535 ms |         - |     - |     - |   1.94 MB |
|  WriteParsedJsonWithoutJsonElementValues_DefaultWriter |            False | 30.176 ms | 1.2152 ms | 3.3060 ms | 28.855 ms | 2000.0000 |     - |     - |    9.6 MB |
| WriteParsedJsonWithoutJsonElementValues_Utf8JsonWriter |            False | 32.233 ms | 2.7839 ms | 8.1647 ms | 28.296 ms | 1000.0000 |     - |     - |   7.73 MB |
|     WriteParsedJsonWithJsonElementValues_DefaultWriter |             True |  9.749 ms | 0.4113 ms | 1.1259 ms |  9.341 ms |         - |     - |     - |   3.67 MB |
|    WriteParsedJsonWithJsonElementValues_Utf8JsonWriter |             True |  7.689 ms | 0.4516 ms | 1.2053 ms |  7.279 ms |         - |     - |     - |   2.54 MB |
|  WriteParsedJsonWithoutJsonElementValues_DefaultWriter |             True | 36.798 ms | 2.2392 ms | 6.5673 ms | 32.988 ms | 2000.0000 |     - |     - |  10.87 MB |
| WriteParsedJsonWithoutJsonElementValues_Utf8JsonWriter |             True | 37.780 ms | 2.6307 ms | 7.2895 ms | 34.741 ms | 2000.0000 |     - |     - |   8.99 MB |

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/IJsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/IJsonWriter.cs
@@ -8,6 +8,9 @@ namespace Microsoft.OData.Json
 {
     using System;
     using Microsoft.OData.Edm;
+#if NETCOREAPP3_1_OR_GREATER
+    using System.Text.Json;
+#endif
 
     /// <summary>
     /// Interface for a class that can write arbitrary JSON.
@@ -152,6 +155,15 @@ namespace Microsoft.OData.Json
         /// </summary>
         /// <param name="value">TimeOfDay value to be written.</param>
         void WriteValue(TimeOfDay value);
+
+#if NETCOREAPP3_1_OR_GREATER
+
+        /// <summary>
+        /// Write a <see cref="JsonElement"/> value.
+        /// </summary>
+        /// <param name="value">The <see cref="JsonElement"/> value to be written.</param>
+        void WriteValue(JsonElement value);
+#endif
 
         /// <summary>
         /// Write a raw value.
--- a/src/Microsoft.OData.Core/Json/JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriter.cs
@@ -13,10 +13,11 @@ namespace Microsoft.OData.Json
     using System.Diagnostics.CodeAnalysis;
     using System.Globalization;
     using System.IO;
-    using System.Text;
-    using System.Threading.Tasks;
     using Microsoft.OData.Buffers;
     using Microsoft.OData.Edm;
+#if NETCOREAPP3_1_OR_GREATER
+    using System.Text.Json;
+#endif
     #endregion Namespaces
 
     /// <summary>
@@ -407,6 +408,13 @@ namespace Microsoft.OData.Json
             this.WriteValueSeparator();
             JsonValueUtils.WriteValue(this.writer, value, this.wrappedBuffer, this.ArrayPool);
         }
+
+#if NETCOREAPP3_1_OR_GREATER
+        public void WriteValue(JsonElement value)
+        {
+            throw new NotImplementedException();
+        }
+#endif
 
         /// <summary>
         /// Write a raw value.
--- a/src/Microsoft.OData.Core/Json/JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriter.cs
@@ -412,7 +412,128 @@ namespace Microsoft.OData.Json
 #if NETCOREAPP3_1_OR_GREATER
         public void WriteValue(JsonElement value)
         {
-            throw new NotImplementedException();
+            switch (value.ValueKind)
+            {
+                case JsonValueKind.Null:
+                    this.WriteValue((string)null);
+                    break;
+                case JsonValueKind.False:
+                    this.WriteValue(false);
+                    break;
+                case JsonValueKind.True:
+                    this.WriteValue(true);
+                    break;
+                case JsonValueKind.String:
+                    this.WriteValue(value.GetString());
+                    break;
+                case JsonValueKind.Array:
+                    this.WriteJsonElementArray(value);
+                    break;
+                case JsonValueKind.Object:
+                    this.WriteJsonElementObject(value);
+                    break;
+                case JsonValueKind.Number:
+                    this.WriteJsonElementNumber(value);
+                    break;
+            }
+        }
+
+        private void WriteJsonElementArray(JsonElement value)
+        {
+            Debug.Assert(value.ValueKind == JsonValueKind.Array);
+
+            this.StartArrayScope();
+            foreach (JsonElement item in value.EnumerateArray())
+            {
+                this.WriteValue(item);
+            }
+
+            this.EndArrayScope();
+        }
+
+        private void WriteJsonElementObject(JsonElement value)
+        {
+            Debug.Assert(value.ValueKind == JsonValueKind.Object);
+
+            this.StartObjectScope();
+            foreach (JsonProperty property in value.EnumerateObject())
+            {
+                this.WriteName(property.Name);
+                this.WriteValue(property.Value);
+            }
+
+            this.EndObjectScope();
+        }
+
+        private void WriteJsonElementNumber(JsonElement value)
+        {
+            Debug.Assert(value.ValueKind == JsonValueKind.Number);
+
+            if (value.TryGetByte(out byte byteValue))
+            {
+                this.WriteValue(byteValue);
+                return;
+            }
+
+            if (value.TryGetDecimal(out decimal decimalValue))
+            {
+                this.WriteValue(decimalValue);
+                return;
+            }
+
+            if (value.TryGetDouble(out double doubleValue))
+            {
+                this.WriteValue(doubleValue);
+                return;
+            }
+
+            if (value.TryGetInt16(out short shortValue))
+            {
+                this.WriteValue(shortValue);
+                return;
+            }
+
+            if (value.TryGetInt32(out int intValue))
+            {
+                this.WriteValue(intValue);
+                return;
+            }
+
+            if (value.TryGetInt64(out long longValue))
+            {
+                this.WriteValue(longValue);
+                return;
+            }
+
+            if (value.TryGetSByte(out sbyte sbyteValue))
+            {
+                this.WriteValue(sbyteValue);
+                return;
+            }
+
+            if (value.TryGetSingle(out float floatValue))
+            {
+                this.WriteValue(floatValue);
+                return;
+            }
+
+            if (value.TryGetUInt16(out ushort ushortValue))
+            {
+                this.WriteValue(ushortValue);
+                return;
+            }
+
+            if (value.TryGetUInt32(out uint uintValue))
+            {
+                this.WriteValue(uintValue);
+                return;
+            }
+
+            if (value.TryGetUInt64(out ulong ulongValue))
+            {
+                this.WriteValue((decimal)ulongValue);
+                return;
+            }
         }
 #endif
 
--- a/src/Microsoft.OData.Core/Json/JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriter.cs
@@ -414,6 +414,9 @@ namespace Microsoft.OData.Json
         {
             switch (value.ValueKind)
             {
+                case JsonValueKind.String:
+                    this.WriteValue(value.GetString());
+                    break;
                 case JsonValueKind.Null:
                     this.WriteValue((string)null);
                     break;
@@ -423,8 +426,8 @@ namespace Microsoft.OData.Json
                 case JsonValueKind.True:
                     this.WriteValue(true);
                     break;
-                case JsonValueKind.String:
-                    this.WriteValue(value.GetString());
+                case JsonValueKind.Number:
+                    this.WriteJsonElementNumber(value);
                     break;
                 case JsonValueKind.Array:
                     this.WriteJsonElementArray(value);
@@ -432,9 +435,7 @@ namespace Microsoft.OData.Json
                 case JsonValueKind.Object:
                     this.WriteJsonElementObject(value);
                     break;
-                case JsonValueKind.Number:
-                    this.WriteJsonElementNumber(value);
-                    break;
+                
             }
         }
 
@@ -469,6 +470,18 @@ namespace Microsoft.OData.Json
         {
             Debug.Assert(value.ValueKind == JsonValueKind.Number);
 
+            if (value.TryGetInt32(out int intValue))
+            {
+                this.WriteValue(intValue);
+                return;
+            }
+
+            if (value.TryGetDouble(out double doubleValue))
+            {
+                this.WriteValue(doubleValue);
+                return;
+            }
+
             if (value.TryGetByte(out byte byteValue))
             {
                 this.WriteValue(byteValue);
@@ -481,21 +494,9 @@ namespace Microsoft.OData.Json
                 return;
             }
 
-            if (value.TryGetDouble(out double doubleValue))
-            {
-                this.WriteValue(doubleValue);
-                return;
-            }
-
             if (value.TryGetInt16(out short shortValue))
             {
                 this.WriteValue(shortValue);
-                return;
-            }
-
-            if (value.TryGetInt32(out int intValue))
-            {
-                this.WriteValue(intValue);
                 return;
             }
 
--- a/src/Microsoft.OData.Core/Json/JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriter.cs
@@ -435,7 +435,9 @@ namespace Microsoft.OData.Json
                 case JsonValueKind.Object:
                     this.WriteJsonElementObject(value);
                     break;
-                
+                default:
+                    // We have already exhausted all known JSON types. Treat this case undefined behaviour and do nothing.
+                    break;
             }
         }
 
--- a/src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
+++ b/src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
@@ -249,31 +249,29 @@ namespace Microsoft.OData.Json
         }
 
 #if NETCOREAPP3_1_OR_GREATER
-        public async Task WriteValueAsync(JsonElement value)
+        public Task WriteValueAsync(JsonElement value)
         {
             switch (value.ValueKind)
             {
                 case JsonValueKind.Null:
-                    await this.WriteValueAsync((string)null).ConfigureAwait(false);
-                    break;
+                    return this.WriteValueAsync((string)null);
                 case JsonValueKind.False:
-                    await this.WriteValueAsync(false).ConfigureAwait(false);
-                    break;
+                    return this.WriteValueAsync(false);
                 case JsonValueKind.True:
-                    await this.WriteValueAsync(true).ConfigureAwait(false);
-                    break;
+                    return this.WriteValueAsync(true);
                 case JsonValueKind.String:
-                    await this.WriteValueAsync(value.GetString()).ConfigureAwait(false);
-                    break;
+                    return this.WriteValueAsync(value.GetString());
                 case JsonValueKind.Array:
-                    await this.WriteJsonElementArrayAsync(value).ConfigureAwait(false);
-                    break;
+                    return this.WriteJsonElementArrayAsync(value);
                 case JsonValueKind.Object:
-                    await this.WriteJsonElementObjectAsync(value).ConfigureAwait(false);
-                    break;
+                    return this.WriteJsonElementObjectAsync(value);
                 case JsonValueKind.Number:
-                    await this.WriteJsonElementNumberAsync(value).ConfigureAwait(false);
-                    break;
+                    return this.WriteJsonElementNumberAsync(value);
+                default:
+                    // we've exhausted all known JSON types, if we get
+                    // to this point, then it's undefined behavior
+                    Debug.Fail($"Unexpected JSON ValueKind {value.ValueKind}");
+                    return Task.CompletedTask;
             }
         }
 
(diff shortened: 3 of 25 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/Json/MockJsonWriter.cs
```diff
@@ -161,5 +168,12 @@ public Task WriteRawValueAsync(string rawValue)
         }
 
         public Task FlushAsync() => throw new NotImplementedException();
+
```
xuzhg, line 171: same as above: => throw new NotImplementedException();

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/MockAsyncOnlyJsonWriter.cs
```diff
@@ -69,5 +72,12 @@ internal class MockAsyncOnlyJsonWriter : IJsonWriterAsync
         public Task WriteValueAsync(Date value) => throw new NotImplementedException();
 
         public Task WriteValueAsync(TimeOfDay value) => throw new NotImplementedException();
+
+#if NETCOREAPP3_1_OR_GREATER
+        public Task WriteValueAsync(JsonElement value)
+        {
```
xuzhg, line 78: Same as above: => throw new NotImplementedException();

src/Microsoft.OData.Core/Json/IJsonWriter.cs
```diff
@@ -153,6 +156,15 @@ public interface IJsonWriter
         /// <param name="value">TimeOfDay value to be written.</param>
         void WriteValue(TimeOfDay value);
 
+#if NETCOREAPP3_1_OR_GREATER
+
+        /// <summary>
+        /// Write a <see cref="JsonElement"/> value.
+        /// </summary>
+        /// <param name="value">The <see cref="JsonElement"/> value to be written.</param>
+        void WriteValue(JsonElement value);
```
xuzhg, line 165: If a property value is a JsonElement, what type is it?

Edm.JsonElement?

System.Text.Json.JsonElement?

src/Microsoft.OData.Core/Json/JsonWriter.cs
```diff
@@ -408,6 +409,134 @@ public void WriteValue(byte[] value)
...
+                    break;
+                case JsonValueKind.Array:
+                    this.WriteJsonElementArray(value);
+                    break;
+                case JsonValueKind.Object:
+                    this.WriteJsonElementObject(value);
+                    break;
+                case JsonValueKind.Number:
+                    this.WriteJsonElementNumber(value);
+                    break;
```
xuzhg, line 437: don't have a default? How about 'Undefined'?

xuzhg (approved):

xuzhg:
The codes look good to me.

habbes:
Tests are failing cause compilation fails in netcoreapp2.1 because the mock JSON writers in the tests do not implement the new method `IJsonWriter.WriteValue(JsonElement)` (and the async equivalent). However, these new methods are only available in netcoreapp3.1 and greater (using conditional compilation), so it's puzzling why it's causing issues in netcoreapp2.1.

image

src/Microsoft.OData.Core/Json/IJsonWriter.cs
habbes: It's `System.Text.Json.JsonElement`. The `JsonElement` type isn't mapped to any specific EDM type.

src/Microsoft.OData.Core/Json/JsonWriter.cs
habbes: We have enumerated all known JSON types, `JsonValueKind` does actually have a `JsonValueKind.Undefined` variant which means "no value". But it's not clear to me what its use case is when parsing JSON because anything that does not map to the known JSON types (null, bool, number, string, array and object) would have resulted in an exception during parsing.

So in this case I think we have handled all known cases. I should add a `default` case, but I don't think it should write anything to the output. It should either:
- do nothing and continue (and treat this edge case as undefined behaviour), or
- throw an exception

What do you think @xuzhg ?

src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
```diff
@@ -245,6 +248,134 @@ public async Task WriteValueAsync(byte[] value)
             await this.writer.WriteValueAsync(value, this.wrappedBuffer, this.ArrayPool).ConfigureAwait(false);
         }
 
+#if NETCOREAPP3_1_OR_GREATER
+        public async Task WriteValueAsync(JsonElement value)
```
gathogojr, line 252: Remove `async` and use `return ...` in the respective case blocks.

src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
```diff
@@ -245,6 +248,134 @@ public async Task WriteValueAsync(byte[] value)
...
+            foreach (JsonProperty property in value.EnumerateObject())
+            {
+                await this.WriteNameAsync(property.Name).ConfigureAwait(false);
+                await this.WriteValueAsync(property.Value).ConfigureAwait(false);
+            }
+
+            await this.EndObjectScopeAsync().ConfigureAwait(false);
+        }
+
+        private async Task WriteJsonElementNumberAsync(JsonElement value)
```
gathogojr, line 307: Elide async/await...

src/Microsoft.OData.Core/Value/ODataJsonElementValue.cs
```diff
@@ -0,0 +1,38 @@
...
+//---------------------------------------------------------------------
+
+#if NETCOREAPP3_1_OR_GREATER
+using System.Text.Json;
+
+namespace Microsoft.OData
+{
+    /// <summary>
+    /// Represents a <see cref="JsonElement"/> value.
+    /// Useful for writingparsed  JSON inputs directly without reading them first, which would result in double allocations.
```
gathogojr, line 14:
```suggestion
    /// Useful for writing parsed JSON inputs directly without reading them first, which would result in double allocations.
```

test/FunctionalTests/Microsoft.OData.Core.Tests/Microsoft.OData.Core.Tests.csproj
```diff
@@ -19,6 +19,10 @@
     <OutputPath>..\..\..\bin\AnyCPU\Release\Test\</OutputPath>
     <DefineConstants>TRACE</DefineConstants>
   </PropertyGroup>
+
+  <PropertyGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+    <DefineConstants>$(DefineConstants);NETCOREAPP3_1_OR_GREATER;NETCOREAPP;NETCOREAPP3_1</DefineConstants>
```
habbes, line 24: Why was this necessary?

Well, in the library and test code bases, I used `NETCOREAPP3_1_OR_GREATER` to conditionally compile code only for .NET Core 3.1+. However, it turns out this constant is not defined by default when you're building the project using .NET Core 3.1 SDK, it's only available by default with .NET 5 SDK and greater. So that's why I manually defined this constant. the `NETCOREAPP3_1` and `NETCOREAPP` symbols are defined by default in .NET Core 3.1, so I didn't need add them here. But I just added them for completeness because, as it turns out, I had already done the same thing in the library `Microsoft.OData.Core.csproj` project file.

The tests were failing to build in the CI pipeline because it's using .NET Core 3.1 SDK to build the tests, so the code under the `NETCOREAPP3_1_OR_GREATER` symbols was not getting compiled. Not it works. This has also made me realize that none of the tests under this symbol (including the `ODataUtf8JsonWriter` tests) ever ran in our build pipeline. We just never noticed. 

Check out [this guide](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives) to see which compiler symbols are auto-defined in the different framework versions.

ElizabethOkerio: This is weird that there are some tests that might not be running on the pipeline and we just don't realize.

src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
```diff
@@ -245,6 +248,124 @@ public async Task WriteValueAsync(byte[] value)
             await this.writer.WriteValueAsync(value, this.wrappedBuffer, this.ArrayPool).ConfigureAwait(false);
         }
 
+#if NETCOREAPP3_1_OR_GREATER
+        public Task WriteValueAsync(JsonElement value)
+        {
+            switch (value.ValueKind)
+            {
+                case JsonValueKind.Null:
+                    return this.WriteValueAsync((string)null);
```
ElizabethOkerio, line 257: will this be converted to an empty string? `string(null)` why can't you pass directly? or the string equivalent of what `string(null)` gives?

src/Microsoft.OData.Core/Value/ODataJsonElementValue.cs
```diff
@@ -0,0 +1,39 @@
...
+    /// </summary>
+    public sealed class ODataJsonElementValue : ODataValue
+    {
+        /// <summary>
+        /// Creates an intsance of <see cref="ODataJsonElementValue"/> based
+        /// on the specified <see cref="JsonElement"/> value.
+        /// </summary>
+        /// <param name="value">The <see cref="JsonElement"/> to wrap.</param>
+        /// <remarks>
+        /// The caller is responsible for ensuring the contents of the <see cref="JsonElement"/> are valid
```
ElizabethOkerio, line 24: what happens if the caller does not ensure the validity of contents? Will ODataWriter just write them?

habbes: Yes, it will write whatever is in the `JsonElement` without checking it.

src/Microsoft.OData.Core/Json/JsonWriter.cs
```diff
@@ -408,6 +409,137 @@ public void WriteValue(byte[] value)
             JsonValueUtils.WriteValue(this.writer, value, this.wrappedBuffer, this.ArrayPool);
         }
 
+#if NETCOREAPP3_1_OR_GREATER
+        public void WriteValue(JsonElement value)
+        {
+            switch (value.ValueKind)
+            {
+                case JsonValueKind.String:
```
gathogojr, line 417: Would a Guid be handled as a string?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightPropertySerializerTests.cs
```diff
@@ -577,15 +579,55 @@ public void WritingPropertyShouldWriteCollectionResourceInstanceAnnotation()
...
+            };
+
+            Action<IContainerBuilder> configureServices = (IContainerBuilder builder) =>
+            {
+                builder.AddService<IStreamBasedJsonWriterFactory>(ServiceLifetime.Singleton, _ => DefaultStreamBasedJsonWriterFactory.Default);
+            };
+
+            var result = SerializeProperty(null, property, configureServices);
+            Assert.Equal("{\"JsonProp\":{\"foo\":\"bar\"}}", result);
+        }
```
ElizabethOkerio, line 618: How come the input string is not similar to the output string?

ElizabethOkerio: Like to access the json object you have to do `result.JsonProp`? is the `result.JsonProp` what gets returned?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
```diff
@@ -17,8 +17,11 @@ namespace Microsoft.OData.JsonLight
     using Microsoft.OData.Json;
     using Microsoft.OData.Metadata;
     using ODataErrorStrings = Microsoft.OData.Strings;
+#if NETCOREAPP3_0_OR_GREATER
```
gathogojr, line 20: Are the changes in this file needed?

src/Microsoft.OData.Core/Value/ODataJsonElementValue.cs
```diff
@@ -0,0 +1,39 @@
...
+namespace Microsoft.OData
+{
+    /// <summary>
+    /// Represents a <see cref="JsonElement"/> value.
+    /// Useful for writing parsed  JSON inputs directly without reading them first, which would result in double allocations.
+    /// </summary>
+    public sealed class ODataJsonElementValue : ODataValue
+    {
+        /// <summary>
+        /// Creates an intsance of <see cref="ODataJsonElementValue"/> based
```
gathogojr, line 19:
```suggestion
        /// Creates an instance of <see cref="ODataJsonElementValue"/> based
```

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightPropertySerializerTests.cs
```diff
@@ -577,15 +579,55 @@ public void WritingPropertyShouldWriteCollectionResourceInstanceAnnotation()
                 result);
         }
 
+#if NETCOREAPP3_1_OR_GREATER
+        [Fact]
+        public void WritingJsonElementProperties_ShouldSerializeJsonInput()
+        {
+            string jsonInputString = "{\"foo\":\"bar\"}";
+            var jsonDocument = System.Text.Json.JsonDocument.Parse(jsonInputString);
```
gathogojr, line 587: Why not import `System.Text.Json` namespace?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightPropertySerializerTests.cs
```diff
@@ -577,15 +579,55 @@ public void WritingPropertyShouldWriteCollectionResourceInstanceAnnotation()
...
+
+            Action<IContainerBuilder> configureServices = (IContainerBuilder builder) =>
+            {
+                builder.AddService<IStreamBasedJsonWriterFactory>(ServiceLifetime.Singleton, _ => DefaultStreamBasedJsonWriterFactory.Default);
+            };
+
+            var result = SerializeProperty(null, property, configureServices);
+            Assert.Equal("{\"JsonProp\":{\"foo\":\"bar\"}}", result);
+        }
+#endif
```
gathogojr, line 619: In the `ODataValueUtils`, there's this block:
```csharp
if (objectToConvert is JsonElement jsonElement)
{
    return new ODataJsonElementValue(jsonElement);
}
```
Do you have a test/tests for that scenario?

gathogojr (dismissed):

src/Microsoft.OData.Core/Json/JsonWriterAsync.cs
habbes: I can't pass `WriteValueAsync(null)` because the compiler won't be sure whether to call `WriteValueAsync(string)` or `WriteValueAsync(byte[])`. Since both `string` and `byte[]` can accept `null` as a value, I need to be explicit about wich overload I am targeting otherwise the code won't compile.

src/Microsoft.OData.Core/Json/JsonWriter.cs
habbes: Yes. Actually, JSON has no native support for Guids and dates, so those are all represented as strings. So, in such cases, we'll just write those back as strings. It will be up to the client on the other side, armed with a copy of the model, to figure out the correct type.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
habbes: This is not needed, nice catch.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightPropertySerializerTests.cs
habbes: Initially I thought I'd only use it in this single statement and didn't feel the need to add an import. But I think I should have. Let me add the using statement.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightPropertySerializerTests.cs
habbes: It's because the `SerializerProperty` helper methods looks like this:

```c#
jsonLightOutputContext.JsonWriter.StartObjectScope();
serializer.WriteProperties(
    owningType,
    new[] { odataProperty },
    /*isComplexValue*/ false,
    new NullDuplicatePropertyNameChecker(),
    null);
jsonLightOutputContext.JsonWriter.EndObjectScope();
```

So, the `JsonElement` is treated as the value of the `ODataProperty` with name `JsonProp` and the property itself is wrapped inside a JSON object (delimited by the `StartObjectScope()` and `EndObjectScope()` calls);

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightPropertySerializerTests.cs
habbes: I had not added tests for this, thanks for the catch. Let me add them.

I've just looked at the description of the `ODataValueUtiles.FromODataValue()` method and have decided to remove the special handling for `JsonElement` from that method. It seems like that method was only intended to unwrap primitive and null values. However, I'll keep the `JsonElement` handling in `ToODataValue()` and add relevant tests.

test/FunctionalTests/Microsoft.OData.Core.Tests/Microsoft.OData.Core.Tests.csproj
habbes: Yes. Though I do not have hard evidence that the tests did not run. Looking at past logs, I don't see detailed break down of all the tests that ran.

ElizabethOkerio (approved):

gathogojr (approved):
:shipit:

PR #2620, opened 2023-02-20
Add HasExplicitODataId to ODataResourceBase

KenitoInc:
### Issues

When deserializing a resource, we set the `Id` property in 2 ways:
- Getting it over the wire from `@odata.id`
- Computing it. e.g if we have a key property with `Id=1` and the resource is of type `Customer` we generate an Id `Customers(1)` using the `MetadataBuilder`.
Also, since the Id can be set publicly, we could do as follows:
```csharp
ODataResource resource = new ODataResource()
{
    Id = new Uri("Customers(1)")
}
```

Considering the above, it's not possible to determine if `resource.Id` was populated from `@odata.id` or not.
In bulk operation handlers, if `resource.Id` was populated from `@odata.id` we should call `TryAddRelatedObject` since we are linking an existing object.

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataResource.cs
+++ b/src/Microsoft.OData.Core/ODataResource.cs
@@ -213,6 +213,14 @@ namespace Microsoft.OData
             set { this.SetInstanceAnnotations(value); }
         }
 
+        /// <summary>Gets or sets the value that shows if the resource has @odata.id or @id.</summary>
+        /// <returns>true if the resource has @odata.id or @id, false otherwise.</returns>
+        public bool HasExplicitODataId
+        {
+            get;
+            internal set;
+        }
+
         /// <summary>
         /// The metadata builder for this OData resource.
         /// </summary>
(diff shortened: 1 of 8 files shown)


xuzhg:
others look good to me. :shipit:

habbes (approved):

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Reader/JsonLight/ODataJsonLightSingletonReaderTests.cs
```diff
@@ -70,10 +70,11 @@ public void ReadBasicSingletonTest()
             Assert.Equal(2, entry.Properties.Count());
             Assert.Equal(10, entry.Properties.Single(p => p.Name == "WebId").Value);
             Assert.Equal("SingletonWeb", entry.Properties.Single(p => p.Name == "Name").Value);
+            Assert.False(entry.HasExplicitODataId);
         }
 
         [Fact]
-        public void ReadSingletonWithIdSpecifiedTest()
+        public void ReadSingletonWithODataIdSpecifiedTest()
```
ElizabethOkerio, line 77: There is no `@odata.id` here.

src/Microsoft.OData.Core/ODataResource.cs
```diff
@@ -213,6 +213,14 @@ public ICollection<ODataInstanceAnnotation> InstanceAnnotations
             set { this.SetInstanceAnnotations(value); }
         }
 
+        /// <summary>Gets or sets the value that shows if the resource has @odata.id or @id.</summary>
+        /// <returns>true if the resource has @odata.id or @id, false otherwise.</returns>
+        public bool HasExplicitODataId
```
mikepizzo, line 218: >HasExplicitODataId [](http://example.com/codeflow?start=20&length=18)

how does this relate to NonComputedId?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Reader/JsonLight/ODataJsonLightSingletonReaderTests.cs
KenitoInc: Check line 81

PR #2623, opened 2023-02-22
Cache container elements in ODataUriResolver model elements cache

### LINKED ISSUE #2534: Microsoft.OData.UriParser.ODataUriResolver!FindSchemaElements running hot on CPU

habbes:
### Issues

*This pull request fixes #2534 .*

### Description

This is a follow up to PR #2610, it adds container elements (operation imports and navigation sources) to the case-insensitive cache used during case-insensitive URI resolution.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
@@ -1,5 +1,5 @@
 ﻿//---------------------------------------------------------------------
-// <copyright file="NormalizedSchemaElementsCache.cs" company="Microsoft">
+// <copyright file="NormalizedModelElementsCache.cs" company="Microsoft">
 //      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
 // </copyright>
 //---------------------------------------------------------------------
@@ -8,18 +8,17 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using Microsoft.OData.Edm.Vocabularies;
-using Microsoft.OData.Metadata;
 
 namespace Microsoft.OData.Edm
 {
     /// <summary>
-    /// Cache used to store schema elements using case-normalized names
+    /// Cache used to store model elements using case-normalized names
     /// to speed up case-insensitive model lookups. The cache is populated
     /// up front so that if an item is not found in the cache, we can assume
     /// it doesn't exist in the model. For this reason, it's important that
     /// the model be immutable.
     /// </summary>
-    internal sealed class NormalizedSchemaElementsCache
+    internal sealed class NormalizedModelElementsCache
     {
         // We create different caches for different types of schema elements because all current usage request schema elements
         // of specific types. If we were to use a single dictionary <string, ISchemaElement> we would need
@@ -35,7 +34,7 @@ namespace Microsoft.OData.Edm
         /// the specified <paramref name="model"/>.
         /// </summary>
         /// <param name="model">The model whose schema elements to cache. This model should be immutable. See <see cref="ExtensionMethods.MarkAsImmutable(IEdmModel)"/>.</param>
-        public NormalizedSchemaElementsCache(IEdmModel model)
+        public NormalizedModelElementsCache(IEdmModel model)
         {
             Debug.Assert(model != null);
 
@@ -94,6 +93,11 @@ namespace Microsoft.OData.Edm
             return null;
         }
 
+        /// <summary>
+        /// Find all navigation sources that match the <paramref name="name"/>.
+        /// </summary>
+        /// <param name="name">The case-insensitive name to match.</param>
+        /// <returns>A list of matching navigation sources, or null if no navigation source matches the name.</returns>
         public List<IEdmNavigationSource> FindNavigationSources(string name)
         {
             if (navigationSourcesCache.TryGetValue(name, out List<IEdmNavigationSource> results))
@@ -104,6 +108,11 @@ namespace Microsoft.OData.Edm
             return null;
         }
 
+        /// <summary>
+        /// Find all operation imports that match the <paramref name="name"/>.
+        /// </summary>
+        /// <param name="name">The case-insensitive name to match.</param>
+        /// <returns>A list of matching operation imports, or null if no operation import matches the name.</returns>
         public List<IEdmOperationImport> FindOperationImports(string name)
         {
             if (operationImportsCache.TryGetValue(name, out List<IEdmOperationImport> results))
@@ -120,15 +129,15 @@ namespace Microsoft.OData.Edm
             {
                 if (element is IEdmSchemaType schemaType)
                 {
-                    AddElementToCache(schemaType, schemaTypesCache);
+                    AddSchemaElementToCache(schemaType, schemaTypesCache);
                 }
                 else if (element is IEdmOperation operation)
                 {
-                    AddElementToCache(operation, operationsCache);
+                    AddSchemaElementToCache(operation, operationsCache);
                 }
                 else if (element is IEdmTerm term)
                 {
-                    AddElementToCache(term, termsCache);
+                    AddSchemaElementToCache(term, termsCache);
                 }
             }
         }
@@ -153,7 +162,7 @@ namespace Microsoft.OData.Edm
             }
         }
 
-        private static void AddElementToCache<T>(T element, Dictionary<string, List<T>> cache) where T : IEdmSchemaElement
+        private static void AddSchemaElementToCache<T>(T element, Dictionary<string, List<T>> cache) where T : IEdmSchemaElement
         {
             List<T> results;
             string normalizedKey = element.FullName();
--- a/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
@@ -93,7 +93,7 @@ namespace Microsoft.OData.UriParser
 
             if (model.IsImmutable())
             {
-                NormalizedModelElementsCache cache = GetNormalizedSchemaElementsCache(model);
+                NormalizedModelElementsCache cache = GetNormalizedModelElementsCache(model);
                 IList<IEdmNavigationSource> cachedResults = cache.FindNavigationSources(identifier);
 
                 if (cachedResults != null)
@@ -313,7 +313,7 @@ namespace Microsoft.OData.UriParser
 
             if (model.IsImmutable())
             {
-                NormalizedModelElementsCache cache = GetNormalizedSchemaElementsCache(model);
+                NormalizedModelElementsCache cache = GetNormalizedModelElementsCache(model);
                 IEnumerable<IEdmOperationImport> cachedResults = cache.FindOperationImports(identifier);
 
                 if (cachedResults != null)
@@ -545,7 +545,7 @@ namespace Microsoft.OData.UriParser
         {
             if (model.IsImmutable())
             {
-                NormalizedModelElementsCache cache = GetNormalizedSchemaElementsCache(model);
+                NormalizedModelElementsCache cache = GetNormalizedModelElementsCache(model);
                 return cacheLookupFunc(cache, identifier);
             }
 
@@ -579,7 +579,7 @@ namespace Microsoft.OData.UriParser
             }
         }
 
-        private static NormalizedModelElementsCache GetNormalizedSchemaElementsCache(IEdmModel model)
+        private static NormalizedModelElementsCache GetNormalizedModelElementsCache(IEdmModel model)
         {
             Debug.Assert(model != null);
 
--- a/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
@@ -93,7 +93,7 @@ namespace Microsoft.OData.UriParser
 
             if (model.IsImmutable())
             {
-                NormalizedSchemaElementsCache cache = GetNormalizedSchemaElementsCache(model);
+                NormalizedModelElementsCache cache = GetNormalizedSchemaElementsCache(model);
                 IList<IEdmNavigationSource> cachedResults = cache.FindNavigationSources(identifier);
 
                 if (cachedResults != null)
@@ -313,7 +313,7 @@ namespace Microsoft.OData.UriParser
 
             if (model.IsImmutable())
             {
-                NormalizedSchemaElementsCache cache = GetNormalizedSchemaElementsCache(model);
+                NormalizedModelElementsCache cache = GetNormalizedSchemaElementsCache(model);
                 IEnumerable<IEdmOperationImport> cachedResults = cache.FindOperationImports(identifier);
 
                 if (cachedResults != null)
@@ -541,11 +541,11 @@ namespace Microsoft.OData.UriParser
         private static IReadOnlyList<T> FindSchemaElements<T>(
             IEdmModel model,
             string identifier,
-            Func<NormalizedSchemaElementsCache, string, List<T>> cacheLookupFunc) where T : IEdmSchemaElement
+            Func<NormalizedModelElementsCache, string, List<T>> cacheLookupFunc) where T : IEdmSchemaElement
         {
             if (model.IsImmutable())
             {
-                NormalizedSchemaElementsCache cache = GetNormalizedSchemaElementsCache(model);
+                NormalizedModelElementsCache cache = GetNormalizedSchemaElementsCache(model);
                 return cacheLookupFunc(cache, identifier);
             }
 
@@ -579,11 +579,11 @@ namespace Microsoft.OData.UriParser
             }
         }
 
-        private static NormalizedSchemaElementsCache GetNormalizedSchemaElementsCache(IEdmModel model)
+        private static NormalizedModelElementsCache GetNormalizedSchemaElementsCache(IEdmModel model)
         {
             Debug.Assert(model != null);
 
-            NormalizedSchemaElementsCache cache = model.GetAnnotationValue<NormalizedSchemaElementsCache>(model);
+            NormalizedModelElementsCache cache = model.GetAnnotationValue<NormalizedModelElementsCache>(model);
             if (cache == null)
             {
                 // There's a chance 2 or more threads can reach here concurrently
@@ -593,7 +593,7 @@ namespace Microsoft.OData.UriParser
                 // We can avoid this waste by providing a method that user can call manually to build
                 // the cache before any request is made. But I did not want to add a new method to the public API.
                 // We revisit this if it turns out to be a problem in practice.
-                cache = new NormalizedSchemaElementsCache(model);
+                cache = new NormalizedModelElementsCache(model);
                 model.SetAnnotationValue(model, cache);
             }
 
--- a/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
@@ -91,6 +91,27 @@ namespace Microsoft.OData.UriParser
                 return navSource;
             }
 
+            if (model.IsImmutable())
+            {
+                NormalizedSchemaElementsCache cache = GetNormalizedSchemaElementsCache(model);
+                IList<IEdmNavigationSource> cachedResults = cache.FindNavigationSources(identifier);
+
+                if (cachedResults != null)
+                {
+                    if (cachedResults.Count == 1)
+                    {
+                        return cachedResults[0];
+                    }
+
+                    if (cachedResults.Count > 1)
+                    {
+                        throw new ODataException(Strings.UriParserMetadata_MultipleMatchingNavigationSourcesFound(identifier));
+                    }
+                }
+
+                return null;
+            }
+
             IEdmEntityContainer container = model.EntityContainer;
             if (container == null)
             {
@@ -288,6 +309,19 @@ namespace Microsoft.OData.UriParser
             if (results.Any() || !EnableCaseInsensitive)
             {
                 return results;
+            }
+
+            if (model.IsImmutable())
+            {
+                NormalizedSchemaElementsCache cache = GetNormalizedSchemaElementsCache(model);
+                IEnumerable<IEdmOperationImport> cachedResults = cache.FindOperationImports(identifier);
+
+                if (cachedResults != null)
+                {
+                    return cachedResults;
+                }
+
+                return Enumerable.Empty<IEdmOperationImport>();
             }
 
             IEdmEntityContainer container = model.EntityContainer;
--- a/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
+++ b/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
@@ -6,7 +6,6 @@
 
 using System;
 using System.Collections.Generic;
-using System.Diagnostics;
 using System.Linq;
 using Microsoft.OData.Edm;
 using Microsoft.OData.Edm.Vocabularies;
@@ -85,6 +84,9 @@ namespace Microsoft.OData.UriParser
         /// <returns>The resolved navigation source.</returns>
         public virtual IEdmNavigationSource ResolveNavigationSource(IEdmModel model, string identifier)
         {
+            ExceptionUtils.CheckArgumentNotNull(model, nameof(model));
+            ExceptionUtils.CheckArgumentNotNull(identifier, nameof(identifier));
+
             IEdmNavigationSource navSource = model.FindDeclaredNavigationSource(identifier);
             if (navSource != null || !EnableCaseInsensitive)
             {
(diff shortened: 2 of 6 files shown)


src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
```diff
@@ -12,30 +12,34 @@
...
+        private readonly Dictionary<string, List<IEdmNavigationSource>> navigationSourcesCache = new Dictionary<string, List<IEdmNavigationSource>>(StringComparer.OrdinalIgnoreCase);
+        private readonly Dictionary<string, List<IEdmOperationImport>> operationImportsCache = new Dictionary<string, List<IEdmOperationImport>>(StringComparer.OrdinalIgnoreCase);
 
         /// <summary>
         /// Builds a case-insensitive cache of schema elements from
         /// the specified <paramref name="model"/>.
         /// </summary>
         /// <param name="model">The model whose schema elements to cache. This model should be immutable. See <see cref="ExtensionMethods.MarkAsImmutable(IEdmModel)"/>.</param>
-        public NormalizedSchemaElementsCache(IEdmModel model)
+        public NormalizedModelElementsCache(IEdmModel model)
```
xuzhg, line 37: Why can't we do the cache in the 'EdmModel'?

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
```diff
@@ -89,26 +93,76 @@ public List<IEdmTerm> FindTerms(string qualifiedName)
...
         }
 
+        /// <summary>
+        /// Find all navigation sources that match the <paramref name="name"/>.
+        /// </summary>
+        /// <param name="name">The case-insensitive name to match.</param>
+        /// <returns>A list of matching navigation sources, or null if no navigation source matches the name.</returns>
+        public List<IEdmNavigationSource> FindNavigationSources(string name)
+        {
+            if (navigationSourcesCache.TryGetValue(name, out List<IEdmNavigationSource> results))
```
xuzhg, line 103: Is the name the simple identifier of the navigation source? not the qualified name?

How about searching the navigation source in the referenced model? with the same name but a different namespace?

xuzhg (approved):

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
habbes: What do you mean by cache in the `EdmModel`? Does the `EdmModel` have a cache?

Also, in this case we're dealing with the `IEdmModel` interface, we can't guarantee whether it's an `EdmModel`, a `CsdlSemanticsModel` or some other implementation.

I've checked around other users in the codebase and in WebAPI, and any time some metadata is added to the model, it's added as an "annotation". In this case this cache will also be an annotation that's bound to the model. It's not a static cache.

Have I answered your question?

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
habbes: Yes, it's the simple identifier of the navigation source. And yes, it's only searching the top-level model entity container.

The reason I implemented it this way is because I wanted to retain the same behaviour as the existing implementation which performs the search as follows:

```c#
IEdmEntityContainer container = model.EntityContainer;
if (container == null)
{
    return null;
}

var result = container.Elements.OfType<IEdmNavigationSource>()
    .Where(source => string.Equals(identifier, source.Name, StringComparison.OrdinalIgnoreCase));
```
see: https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs#L94-L102

As you can see the existing implementation only searches the top-level model's container and it only compares the `source.Name`, not the fully qualified name (including the container name).

If I change the behaviour of the cache to include referenced models or fully qualified identifier, then there would be an inconsistency between the cache and non-cache code paths.

On the other hand, if the expected behaviour is that the URI resolver should search referenced models and/or fully qualified name, then the current implementation is a bug. And I think that should be a separate discussion since that change of behaviour could be observable to the customer.

habbes: The [doc comment](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs#L283) of the `ResolveOperationImports` describes the `identifier` parameter as follows:
```
The qualified name of the operation import which may or may not include the container name.
```

However, the existing implementation does not consider the qualified name in its search:

```c#
IEdmEntityContainer container = model.EntityContainer;
if (container == null)
{
    return Enumerable.Empty<IEdmOperationImport>();
}

return container.Elements.OfType<IEdmOperationImport>()
    .Where(source => string.Equals(identifier, source.Name, StringComparison.OrdinalIgnoreCase));
```

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
```diff
@@ -89,26 +93,76 @@ public List<IEdmTerm> FindTerms(string qualifiedName)
             return null;
         }
 
+        /// <summary>
+        /// Find all navigation sources that match the <paramref name="name"/>.
+        /// </summary>
+        /// <param name="name">The case-insensitive name to match.</param>
+        /// <returns>A list of matching navigation sources, or null if no navigation source matches the name.</returns>
+        public List<IEdmNavigationSource> FindNavigationSources(string name)
```
ElizabethOkerio, line 101: This method returns `List<IEdmNavigationSource>` and also outs `List<IEdmNavigationSource>`. Can we use one or the other?

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
ElizabethOkerio: If we are specifying the navigation source identifier and not searching in referenced models, is there a scenario where whatever that gets returned is a list and not just a single navigation source?

ElizabethOkerio: Like can we have several navigation sources with the same identifier in the same model?

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
habbes: This method returns `List<IEdmNavigationSource>`, it does not "out" it, the signature is:
```c#
List<IEdmNavigationSource FindNavigationSources(string name);
```
and not
```c#
List<IEdmNavigationSource FindNavigationSources(string name, out List<IEdmNavigationSource> ...)
```

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
habbes: Since we're searching a single container, I don't think that's possible. Maybe if we were searching in multiple containers that could be possible, but that would be arguably a new feature, extending the existing behaviour, which is beyond the scope of this PR.

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
ElizabethOkerio: Yea. I thought I saw the out param in the method signature.

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
```diff
@@ -89,26 +93,76 @@ public List<IEdmTerm> FindTerms(string qualifiedName)
...
+        }
+
+        /// <summary>
+        /// Find all operation imports that match the <paramref name="name"/>.
+        /// </summary>
+        /// <param name="name">The case-insensitive name to match.</param>
+        /// <returns>A list of matching operation imports, or null if no operation import matches the name.</returns>
+        public List<IEdmOperationImport> FindOperationImports(string name)
+        {
+            if (operationImportsCache.TryGetValue(name, out List<IEdmOperationImport> results))
```
ElizabethOkerio, line 118: same question here..can we have several operations with the same name in a model or what's the rationale of outing a `List<IEdmOperationImport>`

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
habbes: However, because we are dealing with a case insensitive scenario, we can have multiple navigation sources in the same container whose names only differ by case. That's why we return a list.

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
```diff
@@ -89,26 +93,76 @@ public List<IEdmTerm> FindTerms(string qualifiedName)
...
                 {
-                    AddElementToCache(term, termsCache);
+                    AddSchemaElementToCache(term, termsCache);
+                }
+            }
+        }
+
+        private void PopulateContainerElements(IEdmModel model)
+        {
+            if (model.EntityContainer is null)
```
ElizabethOkerio, line 147: can the model be null here?

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
habbes: In theory, we can have multiple similar names that only differ by case (e.g. `FooBar` and `fooBar`). For operations, we can also have overloads with the same name but different parameters.

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
```diff
@@ -89,26 +93,76 @@ public List<IEdmTerm> FindTerms(string qualifiedName)
...
+            {
+                return results;
+            }
+
+            return null;
+        }
+
         private void PopulateSchemaElements(IEdmModel model)
         {
             foreach (IEdmSchemaElement element in model.SchemaElements)
```
ElizabethOkerio, line 128: can the model be null here or the check is done by the caller?

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
habbes: It's the responsibility of the caller to ensure that the model is not null here. Since this is an internal class, I placed a `Debug.Assert(model != null)` in the constructor.

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
habbes: That is the responsibility of the caller. In this class I only placed a `Debug.Assert(model != null)` in the constructor.

That said, I checked the public methods in the public `ODataUriResolver` class and they're not performing any null-checks on the args.

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
ElizabethOkerio: can we have navigation sources in the same container whose names only differ by case? or we can access a navigation source using different cases?

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
ElizabethOkerio: so here do you check for the params too?

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -91,6 +91,27 @@ public virtual IEdmNavigationSource ResolveNavigationSource(IEdmModel model, str
...
+                if (cachedResults != null)
+                {
+                    if (cachedResults.Count == 1)
+                    {
+                        return cachedResults[0];
+                    }
+
+                    if (cachedResults.Count > 1)
+                    {
+                        throw new ODataException(Strings.UriParserMetadata_MultipleMatchingNavigationSourcesFound(identifier));
```
ElizabethOkerio, line 108: why throw an exception is cachedResults are more than one and you said this is a possibility? or what does this mean?

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -290,6 +311,19 @@ public virtual IEnumerable<IEdmOperationImport> ResolveOperationImports(IEdmMode
...
             }
 
+            if (model.IsImmutable())
+            {
+                NormalizedModelElementsCache cache = GetNormalizedModelElementsCache(model);
+                IEnumerable<IEdmOperationImport> cachedResults = cache.FindOperationImports(identifier);
+
+                if (cachedResults != null)
+                {
+                    return cachedResults;
```
ElizabethOkerio, line 321: why don't you do the checks for multiple operations here like you doing with navigation sources?

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -545,11 +579,11 @@ internal static ODataUriResolver GetUriResolver(IServiceProvider container)
             }
         }
 
-        private static NormalizedSchemaElementsCache GetNormalizedSchemaElementsCache(IEdmModel model)
+        private static NormalizedModelElementsCache GetNormalizedModelElementsCache(IEdmModel model)
         {
             Debug.Assert(model != null);
```
ElizabethOkerio, line 584: why are doing these checks here if you already did them in the constructor?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/NormalizedModelElementsCacheTests.cs
```diff
@@ -157,10 +157,115 @@ public void FindTerms_ReturnsNullIfThereAreNoMatches()
...
+            var cache = new NormalizedModelElementsCache(model);
 
             var matches = cache.FindTerms("ns.models.some");
             Assert.Null(matches);
         }
+
+        [Theory]
+        [InlineData("People")]
+        [InlineData("people")]
+        public void FindNavigationSources_ReturnsNavigationSourcesThatMatcheName(string name)
```
ElizabethOkerio, line 169:
```suggestion
        public void FindNavigationSources_ReturnsNavigationSourcesThatMatchesName(string name)

```
```suggestion
        public void FindNavigationSources_ReturnsNavigationSourcesThatMatchName(string name)

```

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/NormalizedModelElementsCacheTests.cs
```diff
@@ -157,10 +157,115 @@ public void FindTerms_ReturnsNullIfThereAreNoMatches()
...
+            var container = model.AddEntityContainer("NS", "Container");
+            var entitySet1 = container.AddEntitySet("People", person);
+            var entitySet2 = container.AddSingleton("peoPle", person);
+            container.AddEntitySet("Persons", person);
+
+            var cache = new NormalizedModelElementsCache(model);
+
+            var matches = cache.FindNavigationSources(name);
+
+            Assert.Equal(2, matches.Count);
```
ElizabethOkerio, line 184: Why do I think this should be 1 match? EntitySet for person is `Persons` and not people..but you are using people to find the navigation sources? Or there is no need of having that persons in the test

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/NormalizedModelElementsCacheTests.cs
```diff
@@ -157,10 +157,115 @@ public void FindTerms_ReturnsNullIfThereAreNoMatches()
...
+            Assert.Contains(matches, match => match == entitySet1);
+            Assert.Contains(matches, match => match == entitySet2);
+        }
+
+        [Fact]
+        public void FindNavigationSources_ReturnsNullIfTHereAreNoMatches()
```
ElizabethOkerio, line 190:
```suggestion
        public void FindNavigationSources_ReturnsNullIfThereAreNoMatches()

```

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/NormalizedModelElementsCacheTests.cs
```diff
@@ -157,10 +157,115 @@ public void FindTerms_ReturnsNullIfThereAreNoMatches()
...
+        public void FindNavigationSources_ReturnsNavigationSourcesThatMatcheName(string name)
+        {
+            var model = new EdmModel();
+            var person = model.AddEntityType("NS", "People");
+            person.AddKeys(person.AddStructuralProperty("Id", EdmPrimitiveTypeKind.Int32));
+
+            var container = model.AddEntityContainer("NS", "Container");
+            var entitySet1 = container.AddEntitySet("People", person);
+            var entitySet2 = container.AddSingleton("peoPle", person);
+            container.AddEntitySet("Persons", person);
```
ElizabethOkerio, line 178: this line,,i don't think it is necessary here.

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/NormalizedModelElementsCacheTests.cs
```diff
@@ -157,10 +157,115 @@ public void FindTerms_ReturnsNullIfThereAreNoMatches()
...
+
+            var cache = new NormalizedModelElementsCache(model);
+
+            var matches = cache.FindNavigationSources("Products");
+
+            Assert.Null(matches);
+        }
+
+        [Fact]
+        public void FindNavigationSources_ReturnsNull_IfThereIsNotContainer()
```
ElizabethOkerio, line 207:
```suggestion
        public void FindNavigationSources_ReturnsNull_IfThereIsNoContainer()

```

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
habbes: The edm model is case-sensitive by default, so it is possible to have navigation sources which only differ by case because they are different names as far as the `IEdmModel` is concerned. If the `IEdmModel` were case-insensitive, then this cache would not be necessary.

Since this cache is case-insensitive, for the same key we might have multiple matching entries, that's why were store the values in lists.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
habbes: This cache-based code path replicates the existing behaviour of the `ResolveNavigationSource` method, which is to throw an exception if multiple matches are found.

Notice that this method returns a single `IEdmNavigationSource`, not a collection. If there are multiple matches, then it means the URI resolver doesn't know which one to use and throws an exception.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
habbes: Because this cache-based code path aims to replicate the existing behaviour of the `ResolveOperationImports` method, which is to return all the matches that were found.

Unlike the `FindNavigationSource` method, this method returns a collection. Keep in mind that the `IEdmModel` can have multiple operation overloads with the same name but different parameters. So this is handled differently from navigation sources.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
habbes: I could remove it. I think someone else suggested having it here just to be sure. But I don't think it makes a major difference to have it here or not.

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/NormalizedModelElementsCacheTests.cs
habbes: I added this line to ensure that the method returns only the items that match the keyword.

To explain the rationale, let's say I had not added "Persons" to the container and only had "People" and "peoPle". Now let's say that in the implementation of `cache.FindNavigationSources` I had a bug that caused every navigation source in the container to be returned, as opposed to only returning those that match. Such a bug would not be caught by this test, the test would be pass because everything in the container happens to match. So to be able to catch such bugs, I added something that does not match to the container to verify that such items are not included in the result.

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Metadata/NormalizedModelElementsCacheTests.cs
habbes: It's to matches because the test is searching for "People" (and also "people"), and there are two navigation sources that match these keywords, the "People" entity set and the "peoPle" singleton.

The reason for having "Persons" in the test is to verify that navigation sources that do not match the key are not included in the result.

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
habbes: I have added null argument checks to the public methods of ODataUriResolver.

ElizabethOkerio (approved):

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
habbes: I removed it. Added argument null checks to public methods in this class instead.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -85,12 +84,36 @@ public TypeFacetsPromotionRules TypeFacetsPromotionRules
...
             {
                 return navSource;
             }
 
+            if (model.IsImmutable())
+            {
+                NormalizedModelElementsCache cache = GetNormalizedModelElementsCache(model);
+                IList<IEdmNavigationSource> cachedResults = cache.FindNavigationSources(identifier);
+
+                if (cachedResults != null)
```
gathogojr, line 101: This is the reason why it's preferrable to return an empty collection rather than null from a function that has a collection return type.

habbes: While I agree that it's generally preferrable, return an empty collection for the cache would introduce more complexity to the code than this single null-check that's not exposed to the user.

I would have to create a static empty list for each element type that the cache supports:
```c#
static readonly List<IEdmSchemaType> emptySchemaTypesList = new List<IEdmSchemaType>();
static readonly List<IEdmOperation> emptyOperationsList = new List<...>;
static readonly List<IEdmTerm> emptyTermsList = ...;
static readonly List<IEdmNavigationSource> emptyNavigationSourcesList = ...;
static readonly List<IEdmOperationImport> emptyOperationImportsList = ...;

// then in the find methods:
if (navigationSources.TryGetValue(...))
{
      return results;
}

return emptyNavigationSourcesList;
```

The runtime overhead is not that bad since these collections are allocated only once, but it didn't feel worthwhile since the cache is an internal implementation detail used in a very limited scope.

That said, I don't hold this opinion strongly. If you still believe that I should return an empty collection despite the explanation above, I can go ahead and make the change.

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
ElizabethOkerio: how do you check for the params? or when two operations with the same name are returned but with different params how do you know which is which?

habbes: I don't know exactly. That is outside the scope of the cache, and possibly resolved elsewhere. The cache was only concerned in replicating the already existing behaviour in a more efficient manner.

Here's a snippet of the existing implementation (this code path will still be used when the model is not immutable):
```c#
return container.Elements.OfType<IEdmOperationImport>()
                .Where(source => string.Equals(identifier, source.Name, StringComparison.OrdinalIgnoreCase));
```
It would be a bug if the cache returned something different from the non-cached version.

gathogojr (approved):
:shipit:

PR #2625, opened 2023-02-23
Content-Type Header changed to 'odata.metadata=minimal' when doing a SelectQuery

### LINKED ISSUE #2567: Query with $select causes error 500 but works in a browser
I try to query Navision with OData, but when I include a $select, the query return Error 500. When I try the exact same query in a browser, the expected result is returned. I have no clue if this is a problem of the client or the server or what's happening at all.

### Assemblies affected
Microsoft.OData.Client: 7.12.5
Microsoft.OData.Core: 7.12.5
Microsoft.OData.Edm: 7.12.5
Microsoft.OData.Extensions.Client: 1.0.6
Microsoft.OData.Extensions.Client.Abstract: 1.0.6

### Reproduce steps
`var client = new NAV(svcRoot, ODataProtocolVersion.V4); //NAV is generated with "OData Connected Service"
var query3 = client.Abholbescheide.OrderBy(x => x.Customer_No).Take(10).Select(x => new { x.Customer_No, x.EAR_Status_Text });
Log.Debug(query3.ToString()); // prints http://%secret%/Abholbescheide?$orderby=Customer_No&$top=10&$select=Customer_No,EAR_Status_Text

var list3 = query3.ToList();
Log.Information($"Abholbescheide: {list3.Count} Items");`

### Expected result

should return a list with 10 items (only with [Customer_No] and [EAR_Status_Text]

### Actual result

Exception thrown: 'Microsoft.OData.Client.DataServiceQueryException' in Microsoft.OData.Client.dll
An error occurred while processing this request.

[15:00:13 ERR] An error occurred while processing this request.
Microsoft.OData.Client.DataServiceQueryException: An error occurred while processing this request.
 ---> Microsoft.OData.Client.DataServiceClientException: {"error":{"code":"","message":"An error has occurred."}}
 ---> Microsoft.OData.ODataErrorException: An error was read from the payload. See the 'Error' property for more details.
   --- End of inner exception stack trace ---
   at Microsoft.OData.Client.QueryResult.ExecuteQuery()
   at Microsoft.OData.Client.DataServiceRequest.Execute[TElement](DataServiceContext context, QueryComponents queryComponents)
   --- End of inner exception stack trace ---
   at Microsoft.OData.Client.DataServiceRequest.Execute[TElement](DataServiceContext context, QueryComponents queryComponents)
   at Microsoft.OData.Client.DataServiceQuery`1.Execute()
   at Microsoft.OData.Client.DataServiceQuery`1.GetEnumerator()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at %secret%.Program.Main(String[] args) in C:\code\test

### Additional detail

when i take the output of query1.ToString() and paste it to the browser, I get the expected results. 

i captured the traffic with wireshark.

this is the ouput from the browser-query:
`GET %secret%/Abholbescheide?$orderby=Customer_No&$top=10&$select=Customer_No,EAR_Status_Text HTTP/1.1
Host: %secret%
User-Agent: Mozilla/5.0 (Windows NT 10.0; Win64; x64; rv:107.0) Gecko/20100101 Firefox/107.0
Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,*/*;q=0.8
Accept-Language: en-US,de;q=0.7,en;q=0.3
Accept-Encoding: gzip, deflate
DNT: 1
Authorization: %secret%
Connection: keep-alive
Upgrade-Insecure-Requests: 1
Sec-GPC: 1

HTTP/1.1 200 OK
Content-Length: 1854
Content-Type: application/json; odata.metadata=minimal
Server: Microsoft-HTTPAPI/2.0
request-id: %secret%
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true
Access-Control-Expose-Headers: Date
Access-Control-Expose-Headers: Content-Length
Access-Control-Expose-Headers: Server
Access-Control-Expose-Headers: OData-Version
OData-Version: 4.0
Date: Wed, 30 Nov 2022 13:59:37 GMT`

and this is the output from my program:
`GET %secret%/Abholbescheide?$orderby=Customer_No&$top=10&$select=Customer_No,EAR_Status_Text HTTP/1.1
Host: %secret%
OData-Version: 4.0
OData-MaxVersion: 4.0
Accept: application/json;odata.metadata=full
Accept-Charset: UTF-8
User-Agent: Microsoft.OData.Client/7.12.5
Connection: Keep-Alive
Authorization: %secret%

HTTP/1.1 500 Internal Server Error
Content-Length: 56
Content-Type: application/json; odata.metadata=full; charset=utf-8
Server: Microsoft-HTTPAPI/2.0
request-id: %secret%
Access-Control-Allow-Origin: *
Access-Control-Allow-Credentials: true
Access-Control-Expose-Headers: Date
Access-Control-Expose-Headers: Content-Length
Access-Control-Expose-Headers: Server
Access-Control-Expose-Headers: OData-Version
OData-Version: 4.0
Date: Wed, 30 Nov 2022 14:00:31 GMT

{"error":{"code":"","message":"An error has occurred."}}`

koldomon:
### Issues

This pull request fixes #2567.

### Description

When a SelectQuery is running, you want a subset of the entity, so the header should be set to "application/json; odata.metadata=minimal"

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceClientFormat.cs
+++ b/src/Microsoft.OData.Client/DataServiceClientFormat.cs
@@ -362,10 +362,10 @@ namespace Microsoft.OData.Client
         {
             if (hasSelectQueryOption)
             {
-                return MimeApplicationJsonODataLightWithAllMetadata;
+                return MimeApplicationJsonODataLight;
             }
 
-            return MimeApplicationJsonODataLight;
+            return MimeApplicationJsonODataLightWithAllMetadata;
         }
 
         /// <summary>
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
@@ -2149,7 +2149,7 @@ namespace Microsoft.OData.Tests.IntegrationTests.Evaluation
                 "} ] }";
 
             InMemoryMessage message = new InMemoryMessage();
-            message.SetHeader("Content-Type", "application/json;odata.metadata=mini");
+            message.SetHeader("Content-Type", "application/json;odata.metadata=minimal");
             message.Stream = new MemoryStream(Encoding.UTF8.GetBytes(payload));
 
             ODataResource topLevelEntry = null;


koldomon:
> @koldomon please read the following Contributor License Agreement(CLA). If you agree with the CLA, please reply with the following information.
> 
> ```
> @microsoft-github-policy-service agree [company="{your company}"]
> ```
> 
> > Options:
> > 
> > * (default - no company specified) I have sole ownership of intellectual property rights to my Submissions and I am not making Submissions in the course of work for my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > 
> > * (when company given) I am making Submissions in the course of work for my employer (or my employer has intellectual property rights in my Submissions by contract or applicable law). I have permission from my employer to make Submissions and enter into this Agreement on behalf of my employer. By signing below, the defined term “You” includes me and my employer.
> > 
> > ```
> > @microsoft-github-policy-service agree company="Microsoft"
> > ```
> 
> Contributor License Agreement

KenitoInc:
@koldomon Add select query tests

koldomon:
> @koldomon Add select query tests

I'm willing to provide a UnitTest, but tbh I'm not that deep in this library and I'd need some help to fullfil your request. I looked some of the other unit-tests, but can't find something where a request is generated and the header is inspected before sending it out. 

Could you please give me some hints how to create a "select-query" and check the header.

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
```diff
@@ -2149,7 +2149,7 @@ public void ReadingInMinialMetadataModeWithContainedElementShouldBeAbleToGenerat
                 "} ] }";
 
             InMemoryMessage message = new InMemoryMessage();
-            message.SetHeader("Content-Type", "application/json;odata.metadata=mini");
+            message.SetHeader("Content-Type", "application/json;odata.metadata=minimal");
```
habbes, line 2152: This test doesn't not seem to be related to change that you made.

src/Microsoft.OData.Client/DataServiceClientFormat.cs
```diff
@@ -362,10 +362,10 @@ private static string ChooseMediaType(bool hasSelectQueryOption)
         {
             if (hasSelectQueryOption)
             {
-                return MimeApplicationJsonODataLightWithAllMetadata;
+                return MimeApplicationJsonODataLight;
             }
 
-            return MimeApplicationJsonODataLight;
+            return MimeApplicationJsonODataLightWithAllMetadata;
```
habbes, line 368: Could this cause breaking changes?

test/FunctionalTests/Microsoft.OData.Core.Tests/IntegrationTests/Evaluation/AutoComputePayloadMetadataInJsonIntegrationTests.cs
koldomon: I just found this while try to fix my related Issue. In my opinion this has been a typo

src/Microsoft.OData.Client/DataServiceClientFormat.cs
koldomon: I'm not really sure. At least it fixes my issue and didn't break any other tests. I tried to find something in the standard about the requirements for the header, but didn't find something. Since there was only one reproducable case [here](https://github.com/OData/odata.net/issues/2567#issuecomment-1384168044) it's more on the server-side where this can cause problems, and my Navision server is having a problem when the header is set to "metadata.full" on a "select-query".
As I interprete this function, the 2 values just got mixed up. I tried to clarify this question [here](https://github.com/OData/odata.net/issues/2567#issuecomment-1403638600) but didn't get an answer.

habbes: Based on the doc comment:
```c#
 /// <param name="hasSelectQueryOption">
 ///   Whether or not the select query option is present in the request URI.
 ///   If true, indicates that the client should ask the server to include all metadata in a JSON-Light response.
 /// </param>
```
It seems that it was intentional to return all metadata when select query option is enabled, and not just a matter of getting the two values mixed up. 

I don't have the full context of why this implemented this way, I have tried to trace its history using `git blame` but it looks like it may predate the git repo.

If setting metadata to full by default for select queries has issues, then it might be worthwhile to default to minimal metadata.

But I don't think it's a good idea to switch the default to full metadata for all other scenarios like what you are PR is doing. If other scenarios work well so far, I think we should keep using minimal metadata for them.

In conclusion, maybe we should default to minimal metadata for all cases instead of switching (both with and without select)

habbes:
It doesn't appear that the full test suite has run for this PR. Let me see if I can fix that.

habbes:
@koldomon I have triggered the pipeline and there are tests failing:

image

The tests that fail seem to all be in this class: https://github.com/OData/odata.net/blob/master/test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DataServiceClientFormatTests.cs

You should open that `ODataE2E.sln` to work with these tests.

The first one to fail seems expected full metadata when using a select query. This is the one that you should target for your change.

The others fail because they expect minimal metadata but your change returned full metadata. These should not be affected by your change.

But seeing that the existing behaviour was intentional (as confirmed by the doc comments and the tests), and we don't know whether it would cause any unwanted side-effects, and this doesn't seem like a common complaint, I don't think we should change default behaviour to return full metadata when there's select query. I think we should add a configuration flag that allows the user to decide whether to user full metadata or minimal metadata (for example a `DataServiceContext.UseFullMetadataForSelectQueries` that defaults to `true`). I don't think this is an elegant solution (this adds _yet another_ obscure, edge-case specific, configuration setting), but seems to be the safe way out for now. If there are no negative side effects observed when turning this flag on in production, we can consider whether to make it the default behaviour.

koldomon:
> @koldomon I have triggered the pipeline and there are tests failing:
> 
> image
> 
> The tests that fail seem to all be in this class: https://github.com/OData/odata.net/blob/master/test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DataServiceClientFormatTests.cs
> 
> You should open that `ODataE2E.sln` to work with these tests.
> 
> The first one to fail seems expected full metadata when using a select query. This is the one that you should target for your change.
> 
> The others fail because they expect minimal metadata but your change returned full metadata. These should not be affected by your change.
> 
> But seeing that the existing behaviour was intentional (as confirmed by the doc comments and the tests), and we don't know whether it would cause any unwanted side-effects, and this doesn't seem like a common complaint, I don't think we should change default behaviour to return full metadata when there's select query. I think we should add a configuration flag that allows the user to decide whether to user full metadata or minimal metadata (for example a `DataServiceContext.UseFullMetadataForSelectQueries` that defaults to `true`). I don't think this is an elegant solution (this adds _yet another_ obscure, edge-case specific, configuration setting), but seems to be the safe way out for now. If there are no negative side effects observed when turning this flag on in production, we can consider whether to make it the default behaviour.

@habbes I see for myself, that this PR is not a solution for my problem. In fact, your advice for the UnitTest to inspect the header leads me to a solution for my problem. I added your suggested eventhandler in my solution, replace the "Accept" header when I have a $select option and everything works fine. Thanks a lot for your help and the advice.

I also see, that this problem is only a edge-case, but I like your idea to make this configurable. 

Maybe you could find out, if this is a problem in the Navision OData Services? 

I'm expecting you to reject this PR (which makes totally sense), so you could fix the typo in the UnitTest I also provided with this PR.

PR #2632, opened 2023-03-02
Fix some error messages in SaveChangesOptions validation

KenitoInc:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -3562,13 +3562,13 @@ namespace Microsoft.OData.Client
             // UseRelativeUri can only be used in Batch Requests
             if (Util.IsFlagSet(options, SaveChangesOptions.UseRelativeUri) && !Util.IsBatch(options))
             {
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -3562,13 +3562,13 @@ private void ValidateSaveChangesOptions(SaveChangesOptions options)
             // UseRelativeUri can only be used in Batch Requests
             if (Util.IsFlagSet(options, SaveChangesOptions.UseRelativeUri) && !Util.IsBatch(options))
             {
-                throw Error.InvalidOperation(Strings.Context_MustBeUsedWith("SaveChangesOptions.UseRelativeUri", "DataServiceCollection"));
+                throw Error.ArgumentOutOfRange(nameof(options));
```
habbes, line 3565: Why is an `ArgumentOutOfRange` exception more suitable in this case?

habbes (commented):
Could you add more details in the description? It's not clear what problem this PR is solving and why it's the right solution.

PR #2638, opened 2023-03-21
add trace data

ElizabethOkerio:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataWriter.cs
+++ b/src/Microsoft.OData.Core/ODataWriter.cs
@@ -12,7 +12,7 @@ namespace Microsoft.OData
     using System.IO;
     using System.Threading.Tasks;
 
-#endregion Namespaces
+    #endregion Namespaces
 
     /// <summary>
     /// Base class for OData writers.
@@ -44,6 +44,7 @@ namespace Microsoft.OData
             WriteEnd();
             return this;
         }
+
 
         /// <summary> Asynchronously start writing a resource set. </summary>
         /// <returns>A task instance that represents the asynchronous write operation.</returns>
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/ODataWriter.cs
```diff
@@ -170,6 +176,13 @@ public ODataWriter Write(ODataDeltaDeletedLink deltaDeletedLink)
         /// <param name="resource">The resource or item to write.</param>
         public virtual Task WriteStartAsync(ODataResource resource)
         {
+#if NETCOREAPP3_1_OR_GREATER || NETSTANDARD2_0_OR_GREATER
+            using (Activity activity = source.StartActivity("OData Resource"))
+            {
+                activity?.SetTag("Writing ODataResource", resource);
+                activity?.AddEvent(new ActivityEvent("Started writing odata resource"));
+            }
```
habbes, line 184: I think the activity here should be writing, e.g. `Write`, or `MessageWriter.Write`, or `MessageWriter.WriteResource`, etc. or some other convention. "OData Resource" doesn't sound like an activity.

From my understanding, this activity should be disposed when writing this resource finishes, not just when the `WriteStart` method ends (like how you're using `using` here). I don't think that will allow us to see much going on since the activity will be completed immediately. Maybe you can test this in a sample application and find out whether this setup (having a separate activity for start and ends allows you to trace the span of work properly).
I thought you'd have to start an activity in the `WriteStart` then save it somewhere (probably in the writerscope) and dispose of it in the equivalent `WriteEnd` method.

The tags should contain key-value metadata. `Writing ODataResource" doesn't like a property, but an event. Furthermore, I'm not sure if setting the entire `resource` as a tag is a good idea because it may contain PII. Maybe we can consider things like payload kind and "safe" data within the resource object that may be useful for debugging.

Lastly, I don't think it's very useful to add an event here since it doesn't add any information beyond what the activity is already capturing.

habbes: We should probably encapsulate the "logging" inside a method for better maintainability and to avoid having the `#if` macros all over the code.

PR #2639, opened 2023-03-21
Add telemetry to validations

KenitoInc:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/Telemetry/TelemetryProvider.cs
```diff
@@ -0,0 +1,26 @@
...
+
+namespace Microsoft.OData.Core.Telemetry
+{
+    /// <summary>
+    /// Telemetry provider.
+    /// </summary>
+    public static class TelemetryProvider
+    {
+        public static readonly ActivitySource ActivitySourceODataCore = new ActivitySource("Microsoft.OData.Core");
+        public static readonly ActivitySource ActivitySourceEdmLib = new ActivitySource("Microsoft.OData.Edm");
```
habbes, line 23: Since Edm and Core are separate libraries, would it be better to have the EDM activity source defined in the EDM lib rather than the core lib?

Also, why is the class public? Will consumers of the library interact with these activity sources directly?

src/Microsoft.OData.Core/WriterValidationUtils.cs
```diff
@@ -32,11 +38,27 @@ internal static void ValidateMessageWriterSettings(ODataMessageWriterSettings me
 
             if (messageWriterSettings.BaseUri != null && !messageWriterSettings.BaseUri.IsAbsoluteUri)
             {
+#if NETCOREAPP3_1_OR_GREATER || NETSTANDARD2_0_OR_GREATER
```
habbes, line 41: I think we should encapsulate this in a method to avoid the `#if NETCORE...` everywhere.

src/Microsoft.OData.Core/WriterValidationUtils.cs
```diff
@@ -32,11 +38,27 @@ internal static void ValidateMessageWriterSettings(ODataMessageWriterSettings me
 
             if (messageWriterSettings.BaseUri != null && !messageWriterSettings.BaseUri.IsAbsoluteUri)
             {
+#if NETCOREAPP3_1_OR_GREATER || NETSTANDARD2_0_OR_GREATER
+                using (Activity activity = source.StartActivity("Validation Error"))
+                {
+                    activity?.SetTag("Validation Type", "Writer Validation");
+                    activity?.SetTag("Message", Strings.WriterValidationUtils_MessageWriterSettingsBaseUriMustBeNullOrAbsolute(UriUtils.UriToString(messageWriterSettings.BaseUri)));
+                }
```
habbes, line 46: My 2 cents: the validation error should be an event rather than an activity, because we are tracing an operation here but rather logging that something has happened. And these events can be correlated to an activity.

We should probably use constants for the activity, event and tag names. But we can look at that after the hackthon.

PR #2641, opened 2023-03-23
Apply tracing to OData write APIs

gathogojr:
Apply tracing to OData write APIs

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
@@ -143,89 +143,92 @@ namespace Microsoft.OData.JsonLight
             IDuplicatePropertyNameChecker duplicatePropertyNameChecker,
             ODataResourceMetadataBuilder metadataBuilder)
         {
-            this.WritePropertyInfo(property, owningType, isTopLevel, duplicatePropertyNameChecker, metadataBuilder);
-
-            ODataValue value = property.ODataValue;
-
-            // handle ODataUntypedValue
-            ODataUntypedValue untypedValue = value as ODataUntypedValue;
-            if (untypedValue != null)
+            using (Activity activity = TelemetryHelper.StartActivity(TelemetryConstants.WriteProperty)?.AddTag("IsAsync", false))
             {
-                WriteUntypedValue(untypedValue);
-                return;
-            }
+                this.WritePropertyInfo(property, owningType, isTopLevel, duplicatePropertyNameChecker, metadataBuilder);
 
-            ODataStreamReferenceValue streamReferenceValue = value as ODataStreamReferenceValue;
-            if (streamReferenceValue != null && !(this.JsonLightOutputContext.MetadataLevel is JsonNoMetadataLevel))
-            {
-                Debug.Assert(!isTopLevel, "Stream properties are not allowed at the top level.");
-                WriteStreamValue(streamReferenceValue, property.Name, metadataBuilder);
-                return;
-            }
+                ODataValue value = property.ODataValue;
 
-            if (value is ODataNullValue || value == null)
-            {
-                this.WriteNullProperty(property);
-                return;
-            }
-
-            bool isOpenPropertyType = this.IsOpenProperty(property);
-
-            ODataPrimitiveValue primitiveValue = value as ODataPrimitiveValue;
-            if (primitiveValue != null)
-            {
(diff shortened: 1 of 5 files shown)


src/Microsoft.OData.Core/OpenTelemetryContants.cs
```diff
@@ -0,0 +1,25 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="OpenTelemetryConstants.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData
+{
+    internal class OpenTelemetryConstants
```
habbes, line 9: I think we can remove references to "OpenTelemetry" in the core library since we're using `System.Diagnostics*`

```suggestion
    internal class TelemetryConstants
```

src/Microsoft.OData.Core/ODataWriterCore.cs
```diff
@@ -731,6 +807,9 @@ public sealed override void WriteEntityReferenceLink(ODataEntityReferenceLink en
         /// </remarks>
         public sealed override Task WriteEntityReferenceLinkAsync(ODataEntityReferenceLink entityReferenceLink)
         {
+#if NETSTANDARD2_0 || NETCOREAPP3_1_OR_GREATER
+            OpenTelemetryHelper.ODataTemeletrySource.StartActivity(OpenTelemetryConstants.WriteEntityReferenceLink);
+#endif
```
habbes, line 812: nitpick:

I think it may beneficial to create a helper method like `TelemetryHelper.StartActivity` and then inside that method we have the `#if NETCOREAPP3_1_OR_GREATER`. But this might be problematic if we want to use the returned activity object.

src/Microsoft.OData.Core/OpenTelemetryHelper.cs
```diff
@@ -0,0 +1,17 @@
...
+//---------------------------------------------------------------------
+
+namespace Microsoft.OData
+{
+    using System.Diagnostics;
+
+#if NETSTANDARD2_0 || NETCOREAPP3_1_OR_GREATER
+    internal class OpenTelemetryHelper
+    {
+        public static readonly ActivitySource ODataTemeletrySource = new ActivitySource("Microsoft.OData.Core");
```
habbes, line 14:
```suggestion
        public static readonly ActivitySource ODataTelemetrySource = new ActivitySource("Microsoft.OData.Core");

```

habbes (approved):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertySerializer.cs
```diff
@@ -143,89 +143,98 @@ internal void WriteTopLevelProperty(ODataProperty property)
...
             ODataResourceMetadataBuilder metadataBuilder)
         {
-            this.WritePropertyInfo(property, owningType, isTopLevel, duplicatePropertyNameChecker, metadataBuilder);
-
-            ODataValue value = property.ODataValue;
-
-            // handle ODataUntypedValue
-            ODataUntypedValue untypedValue = value as ODataUntypedValue;
-            if (untypedValue != null)
+#if NETSTANDARD2_0 || NETCOREAPP3_1_OR_GREATER
```
ElizabethOkerio, line 146: why didn't we use NETSTANDARD2_0_OR_GREATER

gathogojr: @ElizabethOkerio Because even in the existing code, e.g. _JsonReader.cs_, we use **`NESTANDARD2_0`** which the library targets specifically.

image

src/Microsoft.OData.Core/TelemetryHelper.cs
```diff
@@ -0,0 +1,62 @@
...
+            Activity.Current?.Stop();
+#endif
+        }
+
+        public static void StopAndDisposeActivity()
+        {
+#if NETSTANDARD2_0 || NETCOREAPP3_1_OR_GREATER
+            Activity.Current?.Dispose(); // Stops and disposes activity
+#endif
+        }
```
habbes, line 43: I don't think we need both of this since `Dispose()` _just_ stops the activity. Maybe we can keep one of these methods and just call `Activity.Current?.Dispose`.

ElizabethOkerio (approved):

PR #2643, opened 2023-03-23
Telemetry to log exception events

habbes:
### Issues

### Description

This PR logs exception events that occur in the `ODataWriter` and attaches them to the current trace activity.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Telemetry.cs
+++ b/src/Microsoft.OData.Core/Telemetry.cs
@@ -12,9 +12,7 @@ namespace Microsoft.OData
 {
     internal static class Telemetry
     {
-
-
-#if NETCOREAPP3_1_OR_GREATER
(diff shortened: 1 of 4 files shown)


src/Microsoft.OData.Core/Telemetry.cs
```diff
@@ -0,0 +1,48 @@
...
+using System.Reflection;
+using System.Text;
+using System.Threading.Tasks;
+#endif
+
+namespace Microsoft.OData
+{
+    internal static class Telemetry
+    {
+
```
ElizabethOkerio, line 15:
```suggestion
```

src/Microsoft.OData.Core/Telemetry.cs
```diff
@@ -0,0 +1,48 @@
...
+
```
ElizabethOkerio, line 16:
```suggestion

```

PR #2652, opened 2023-04-10
Fix JsonElementValue support edge cases

### LINKED ISSUE #2650: ODataJsonElementValue causes exceptions for ID fields
When `ODataJsonElementValue` is used as the value an `ODataProperty` that forms part of the key, an exception may be thrown with the message

example:
```
Microsoft.OData.ODataException: The property 'id' on type 'NS.Type is a non-primitive value. All key and etag properties must be of primitive types.

 

  Stack Trace: 
ODataResourceMetadataContext.GetPrimitiveOrEnumPropertyValue(String entityTypeName, ODataProperty property, Boolean validateNotNull)
ODataResourceMetadataContext.TryGetPrimitiveOrEnumPropertyValue(ODataResourceBase resource, String propertyName, String entityTypeName, Boolean isRequired, Object& value)
ODataResourceMetadataContext.GetPropertyValues(IEnumerable`1 properties, ODataResourceBase resource, IEdmEntityType actualEntityType, Boolean isRequired)+MoveNext()
LargeArrayBuilder`1.AddRange(IEnumerable`1 items)
EnumerableHelpers.ToArray[T](IEnumerable`1 source)
Enumerable.ToArray[TSource](IEnumerable`1 source)
ODataResourceMetadataContextWithModel.get_KeyProperties()
ODataConventionalIdMetadataBuilder.ComputeIdForContainment()
ODataConventionalIdMetadataBuilder.ComputeAndCacheId()
ODataConventionalIdMetadataBuilder.get_ComputedId()
<4 more frames...>
<>c.<WriteStartResourceImplementation>b__146_0(ODataWriterCore thisParam, ODataResource resourceParam)
ODataWriterCore.InterceptException[TArg0](Action`2 action, TArg0 arg0)
ODataWriterCore.WriteStartResourceImplementation(ODataResource resource)
ODataWriterCore.WriteStart(ODataResource resource)
```

The error seems to occur when constructing URLs with id segments: e.g. `@odata.id`, context URLs, contained entity sets, etc.

We should make an exception for `ODataJsonElementValue` since when using this type, you implicitly opt out of most validation.

### Assemblies affected

`Microsoft.OData.Core` 7.15.0+

### Reproduce steps

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

habbes:
### Issues

This pull request fixes #2650

There are certain scenarios where the serializer expects a primitive value and throws an error if what it finds is an `ODataJsonElementValue`. This PR adds support or exceptions for `ODataJsonElementValue` in those cases.

### Description

There are certain scenarios where the serializer expects a primitive value and throws an error if what it finds is an `ODataJsonElementValue`. This PR adds support or exceptions for `ODataJsonElementValue` in those cases:

- When generating an `@odata.id` (e.g. when writing a contained entity), the writer validates that the ID is a primitive type. This PR does not throw the exception if the value is an `ODataJsonElementValue`.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
@@ -14,6 +14,7 @@ using System.Threading.Tasks;
 using Microsoft.OData.Tests.Json;
 using System.Text;
 using Microsoft.Test.OData.DependencyInjection;
+using Microsoft.OData.UriParser;
 
 namespace Microsoft.OData.Tests
 {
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
```diff
@@ -330,6 +330,16 @@ public virtual void WriteNullValue()
         {
             Debug.Assert(value != null, "value != null");
 
+#if NETCOREAPP3_1_OR_GREATER
+            if (value is ODataJsonElementValue jsonElementValue)
+            {
+                // We don't perform validation for ODataJsonElementValue.
+                // We assume the content is valid.
```
xuzhg, line 337: Where do you get such assumptions?

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -14,6 +14,10 @@
 using Microsoft.OData.Tests.Json;
 using System.Text;
 using Microsoft.Test.OData.DependencyInjection;
+using Microsoft.OData.UriParser;
+#if NETCOREAPP3_1_OR_GREATER
+using System.Text.Json;
```
xuzhg, line 19: It could be sorting the usings, right?

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -487,6 +491,361 @@ public async Task WhenInjectingStreamBasedJsonWriterFactoryAsync_CreatesWriterUs
...
+                "application/json; charset=utf-8",
+                writer =>
+                {
+                    var resourceWriter = writer.CreateODataResourceWriter(peopleSet);
+                    resourceWriter.WriteStart(resource);
+                    resourceWriter.WriteEnd();
+                },
+                path: new ODataPath(new EntitySetSegment(peopleSet)));
+
+            // Result
```
xuzhg, line 547: // Assert

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -487,6 +491,361 @@ public async Task WhenInjectingStreamBasedJsonWriterFactoryAsync_CreatesWriterUs
...
+            string expected = @"{""@odata.context"":""http://www.example.com/$metadata#People/$entity"",""Id"":1,""Name"":""John"",""Emails"":[""john@mailer.com"",""john@work.com""],""Address"":{""City"":""Nairobi"",""Country"":""Kenya""}}";
+            Assert.Equal(expected, result);
+        }
+
+        [Fact]
+        public async Task WriteEntityWithJsonElementValues_Async()
+        {
+            // Arrange
+            EdmModel model = new EdmModel();
```
xuzhg, line 556: The model creation codes repeats? could be merge them as a method?

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -487,6 +491,361 @@ public async Task WhenInjectingStreamBasedJsonWriterFactoryAsync_CreatesWriterUs
...
+                @"""Emails"":[""john@mailer.com"",""john@work.com""]," +
+                @"""Address"":{""City"":""Nairobi"",""Country"":""Kenya""}" +
+                @"}";
+            var json = JsonDocument.Parse(source);
+            var jsonRoot = json.RootElement;
+            var resource = new ODataResource
+            {
+                Properties = new[]
+                {
+                    new ODataProperty { Name = "Id", Value = new ODataJsonElementValue(jsonRoot.GetProperty("Id")) },
```
xuzhg, line 528: why don't we use the "int" directly?
I mean what's the benefit to use "ODataJsonElementValue" to wrap the "Id", "Name"... (such primitive type property)?

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -487,6 +491,361 @@ public async Task WhenInjectingStreamBasedJsonWriterFactoryAsync_CreatesWriterUs
...
+                "Emails",
+                new EdmCollectionTypeReference(
+                    new EdmCollectionType(EdmCoreModel.Instance.GetString(isNullable: false))));
+            EdmComplexType addressType = model.AddComplexType("ns", "Address");
+            addressType.AddStructuralProperty("City", EdmPrimitiveTypeKind.String);
+            addressType.AddStructuralProperty("Country", EdmPrimitiveTypeKind.String);
+            personType.AddStructuralProperty("Address", new EdmComplexTypeReference(addressType, isNullable: false));
+            IEdmEntitySet peopleSet = model.AddEntityContainer("ns", "Service").AddEntitySet("People", personType);
+
+
```
xuzhg, line 571: remove useless blank line

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -487,6 +491,361 @@ public async Task WhenInjectingStreamBasedJsonWriterFactoryAsync_CreatesWriterUs
...
+            string source = "{" +
+                @"""Id"": 1," +
+                @"""Name"": ""John""," +
+                @"""Emails"":[""john@mailer.com"",""john@work.com""]," +
+                @"""Address"":{""City"":""Nairobi"",""Country"":""Kenya""}" +
+                @"}";
+            var json = JsonDocument.Parse(source);
```
xuzhg, line 578: It seems like we can use a method to accept the json source and return the rootElement?

xuzhg (dismissed):

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
habbes: Frankly, there's no value in wrapping the primitive here with `ODataJsonElementValue`, but this is how I was able to reproduce the bug reported by AGS. I can check back with AGS whether they have a more representative repro.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
habbes: This assumption is part of the design choices behind the `ODataJsonElementValue` and is documented here: https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Value/ODataJsonElementValue.cs#L24

Essentially, the benefit of using `ODataJsonElementValue` is to pass-through `JsonElement` values without having to parse them. Since the `ODataWriter` does not parse nor inspect the contents of the `ODataJsonElementValue`, we do not attempt to validate it.

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
habbes: I'll sort the `using`s above the `#if`. Do you also suggesting sorting the `#if`? I think it makes sense to group the `#if` separately?

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
habbes: Yeah. But I create different model elements based on scenario. I thought of merging them into a shared method, but since this test file is large, I thought things could get lost if I created a helper method that's only used in 3 tests.

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
habbes: Do you mean something like this:
```c#
JsonElement ParseJson(string source) => JsonDocument.Parse(source).RootElement;
```
It seems like what `JsonDocument.Parse` is already doing, does it really need to be a separate method?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
ElizabethOkerio: I think this should be publicly documented. What do you think?

ElizabethOkerio (approved):

gathogojr (approved):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightValueSerializer.cs
habbes: Technically it is publicly documented since doc comments end up in the public reference documented page for that class. That said, we should include an entry in the main docs as well.

PR #2653, opened 2023-04-11
Feature/deep insert request

KenitoInc:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -2341,7 +2341,7 @@ namespace Microsoft.OData.Client
 
         #region Add, Attach, Delete, Detach, Update, TryGetEntity, TryGetUri
 
-        /// <summary>Adds the specified link to the set of objects the <see cref="Microsoft.OData.Client.DataServiceContext" /> is tracking.</summary>
+        /// <summary>Adds the specified link to the set of objects the <see cref="Microsoft.OData.Client.DataServiceContext" /> is tracking. The sourceProperty MUST be a collection navigation property.</summary>
         /// <param name="source">The source object for the new link.</param>
         /// <param name="sourceProperty">The name of the navigation property on the source object that returns the related object.</param>
         /// <param name="target">The object related to the source object by the new link. </param>
@@ -2483,20 +2483,17 @@ namespace Microsoft.OData.Client
             }
         }
 
-        /// <summary>Adds the specified link to the set of objects the <see cref="Microsoft.OData.Client.DataServiceContext" /> is tracking.</summary>
+        /// <summary>Adds the specified link to the set of objects the <see cref="Microsoft.OData.Client.DataServiceContext" /> is tracking. The sourceProperty MUST be a single-value navigation property.</summary>
         /// <param name="source">The source object for the new link.</param>
         /// <param name="sourceProperty">The name of the navigation property on the source object that returns the related object.</param>
-        /// <param name="target">he object related to the source object by the new link.</param>
+        /// <param name="target">The object related to the source object by the new link.</param>
         /// <exception cref="System.ArgumentNullException">When <paramref name="source" />, <paramref name="sourceProperty" /> or <paramref name="target" /> are null.</exception>
         /// <exception cref="System.InvalidOperationException">When the specified link already exists.-or-When the objects supplied as <paramref name="source" /> or <paramref name="target" /> are in the <see cref="Microsoft.OData.Client.EntityStates.Detached" /> or <see cref="Microsoft.OData.Client.EntityStates.Deleted" /> state.-or-When <paramref name="sourceProperty" /> is not a navigation property that defines a reference to a single related object.</exception>
-        /// <remarks>
-        /// The sourceProperty MUST be a single-value navigation property.
-        /// </remarks>
         public virtual void SetRelatedObjectLink(object source, string sourceProperty, object target)
         {
-            Util.CheckArgumentNull(source, "source");
-            Util.CheckArgumentNullAndEmpty(sourceProperty, "sourceProperty");
-            Util.CheckArgumentNull(target, "target");
+            Util.CheckArgumentNull(source, nameof(source));
+            Util.CheckArgumentNullAndEmpty(sourceProperty, nameof(sourceProperty));
+            Util.CheckArgumentNull(target, nameof(target));
 
             // Validate that the source is an entity and is already tracked by the context.
             ValidateEntityType(source, this.Model);
@@ -2510,11 +2507,11 @@ namespace Microsoft.OData.Client
             }
 
             // Validate that the property is valid and exists on the source
-            ClientTypeAnnotation type = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(source.GetType()));
-            Debug.Assert(type.IsEntityType, "should be an entity type");
+            ClientTypeAnnotation sourceType = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(source.GetType()));
+            Debug.Assert(sourceType.IsEntityType, "should be an entity type");
 
             // will throw InvalidOperationException if property doesn't exist
-            ClientPropertyAnnotation property = type.GetProperty(sourceProperty, UndeclaredPropertyBehavior.ThrowException);
+            ClientPropertyAnnotation property = sourceType.GetProperty(sourceProperty, UndeclaredPropertyBehavior.ThrowException);
 
             if (property.IsKnownType || property.IsEntityCollection)
             {
@@ -2523,13 +2520,13 @@ namespace Microsoft.OData.Client
 
             // if (property.IsEntityCollection) then property.PropertyType is the collection elementType
             // either way you can only have a relation ship between keyed objects
-            type = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(property.EntityCollectionItemType ?? property.PropertyType));
-            Debug.Assert(type.IsEntityType, "should be enforced by just adding an object");
+            ClientTypeAnnotation sourcePropertyType = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(property.EntityCollectionItemType ?? property.PropertyType));
+            Debug.Assert(sourcePropertyType.IsEntityType, "should be an entity type.");
 
-            if ((target != null) && !type.ElementType.IsInstanceOfType(target))
+            if (!sourcePropertyType.ElementType.IsInstanceOfType(target))
             {
                 // target is not of the correct type
-                throw Error.Argument(Strings.Context_RelationNotRefOrCollection, "target");
+                throw Error.Argument(Strings.Context_RelationNotRefOrCollection, nameof(target));
             }
 
             LinkDescriptor descriptor = new LinkDescriptor(source, sourceProperty, target, this.model);
@@ -2569,7 +2566,7 @@ namespace Microsoft.OData.Client
             this.EntityTracker.IncrementChange(resource);
         }
 
-        /// <summary>Adds a related object to the context and creates the link that defines the relationship between the two objects in a single request.</summary>
+        /// <summary>Adds a related object to the context and creates the link that defines the relationship between the two objects in a single request. The sourceProperty MUST be a collection navigation property.</summary>
         /// <param name="source">The parent object that is being tracked by the context.</param>
         /// <param name="sourceProperty">The name of the navigation property that returns the related object based on an association between the two entities.</param>
         /// <param name="target">The related object that is being added.</param>
@@ -2628,13 +2625,10 @@ namespace Microsoft.OData.Client
             this.entityTracker.IncrementChange(targetResource);
         }
 
-        /// <summary>Adds a related object to the context and creates the link that defines the relationship between the two objects in a single request.</summary>
+        /// <summary>Adds a related object to the context and creates the link that defines the relationship between the two objects in a single request. The sourceProperty MUST be a single-value navigation property.</summary>
         /// <param name="source">The parent object that is being tracked by the context.</param>
         /// <param name="sourceProperty">The name of the navigation property that returns the related object based on an association between the two entities.</param>
         /// <param name="target">The related object that is being added.</param>
-        /// /// <remarks>
-        /// The sourceProperty MUST be a single-value object, not a collection.
-        /// </remarks>
         public virtual void SetRelatedObject(object source, string sourceProperty, object target)
         {
             Util.CheckArgumentNull(source, "source");
@@ -2671,7 +2665,7 @@ namespace Microsoft.OData.Client
             if (!propertyElementType.ElementType.IsAssignableFrom(childType.ElementType))
             {
                 // target is not of the correct type
-                throw Error.Argument(Strings.Context_RelationNotRefOrCollection, "target");
+                throw Error.Argument(Strings.Context_RelationNotRefOrCollection, nameof(target));
             }
 
             var targetResource = new EntityDescriptor(this.model)
@@ -2971,9 +2965,9 @@ namespace Microsoft.OData.Client
 
             return identity != null;
         }
-        
+
         /// <summary>
-        /// Processes deep insert requests.
+        /// Processes deep insert requests. Creates an object and creates related navigation items or link existing navigation items in a single request. 
         /// </summary>
         /// <typeparam name="T">The type of top-level object to be deep inserted.</typeparam>
         /// <param name="resource">The top-level object of the type to be deep inserted.</param>
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -2483,6 +2483,72 @@ namespace Microsoft.OData.Client
             }
         }
 
+        /// <summary>Adds the specified link to the set of objects the <see cref="Microsoft.OData.Client.DataServiceContext" /> is tracking.</summary>
+        /// <param name="source">The source object for the new link.</param>
+        /// <param name="sourceProperty">The name of the navigation property on the source object that returns the related object.</param>
+        /// <param name="target">he object related to the source object by the new link.</param>
+        /// <exception cref="System.ArgumentNullException">When <paramref name="source" />, <paramref name="sourceProperty" /> or <paramref name="target" /> are null.</exception>
+        /// <exception cref="System.InvalidOperationException">When the specified link already exists.-or-When the objects supplied as <paramref name="source" /> or <paramref name="target" /> are in the <see cref="Microsoft.OData.Client.EntityStates.Detached" /> or <see cref="Microsoft.OData.Client.EntityStates.Deleted" /> state.-or-When <paramref name="sourceProperty" /> is not a navigation property that defines a reference to a single related object.</exception>
+        /// <remarks>
+        /// The sourceProperty MUST be a single-value navigation property.
+        /// </remarks>
+        public virtual void SetRelatedObjectLink(object source, string sourceProperty, object target)
+        {
+            Util.CheckArgumentNull(source, "source");
+            Util.CheckArgumentNullAndEmpty(sourceProperty, "sourceProperty");
+            Util.CheckArgumentNull(target, "target");
+
+            // Validate that the source is an entity and is already tracked by the context.
+            ValidateEntityType(source, this.Model);
+
+            EntityDescriptor sourceResource = this.entityTracker.GetEntityDescriptor(source);
+
+            // Check for deleted source entity
+            if (sourceResource.State == EntityStates.Deleted)
+            {
+                throw Error.InvalidOperation(Strings.Context_SetRelatedObjectLinkSourceDeleted);
+            }
+
+            // Validate that the property is valid and exists on the source
+            ClientTypeAnnotation type = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(source.GetType()));
+            Debug.Assert(type.IsEntityType, "should be an entity type");
+
+            // will throw InvalidOperationException if property doesn't exist
+            ClientPropertyAnnotation property = type.GetProperty(sourceProperty, UndeclaredPropertyBehavior.ThrowException);
+
+            if (property.IsKnownType || property.IsEntityCollection)
+            {
+                throw Error.InvalidOperation(Strings.Context_SetRelatedObjectLinkNonCollectionOnly);
+            }
+
+            // if (property.IsEntityCollection) then property.PropertyType is the collection elementType
+            // either way you can only have a relation ship between keyed objects
+            type = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(property.EntityCollectionItemType ?? property.PropertyType));
+            Debug.Assert(type.IsEntityType, "should be enforced by just adding an object");
+
+            if ((target != null) && !type.ElementType.IsInstanceOfType(target))
+            {
+                // target is not of the correct type
+                throw Error.Argument(Strings.Context_RelationNotRefOrCollection, "target");
+            }
+
+            LinkDescriptor descriptor = new LinkDescriptor(source, sourceProperty, target, this.model);
+            this.entityTracker.AddLink(descriptor);
+            descriptor.State = EntityStates.Added;
+            this.entityTracker.IncrementChange(descriptor);
+
+            /*LinkDescriptor relation = this.entityTracker.DetachReferenceLink(source, sourceProperty, target, MergeOption.NoTracking);
+
+            if (relation == null)
+            {
+                relation = new LinkDescriptor(source, sourceProperty, target, this.model);
+                this.entityTracker.AddLink(relation);
+            }
+
+            relation.State = EntityStates.Added;
+            this.entityTracker.IncrementChange(relation);*/
+        }
+
         #endregion
 
         #region AddObject, AttachTo, DeleteObject, Detach, TryGetEntity, TryGetUri
@@ -2600,6 +2666,7 @@ namespace Microsoft.OData.Client
             // Validate that the property is valid and exists on the source
             ClientTypeAnnotation parentType = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(source.GetType()));
             ClientPropertyAnnotation property = parentType.GetProperty(sourceProperty, UndeclaredPropertyBehavior.ThrowException);
+
             if (property.IsKnownType || property.IsEntityCollection)
             {
                 throw Error.InvalidOperation(Strings.Context_SetRelatedObjectNonCollectionOnly);
@@ -2611,6 +2678,7 @@ namespace Microsoft.OData.Client
 
             // Validate that the property type matches with the target type
             ClientTypeAnnotation propertyElementType = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(property.PropertyType));
+
             if (!propertyElementType.ElementType.IsAssignableFrom(childType.ElementType))
             {
                 // target is not of the correct type
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -2536,17 +2536,6 @@ namespace Microsoft.OData.Client
             this.entityTracker.AddLink(descriptor);
             descriptor.State = EntityStates.Added;
             this.entityTracker.IncrementChange(descriptor);
-
-            /*LinkDescriptor relation = this.entityTracker.DetachReferenceLink(source, sourceProperty, target, MergeOption.NoTracking);
-
-            if (relation == null)
-            {
-                relation = new LinkDescriptor(source, sourceProperty, target, this.model);
-                this.entityTracker.AddLink(relation);
-            }
-
-            relation.State = EntityStates.Added;
-            this.entityTracker.IncrementChange(relation);*/
         }
 
         #endregion
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -2852,6 +2852,22 @@ namespace Microsoft.OData.Client
 
             return identity != null;
         }
+        
+        /// <summary>
+        /// Processes deep insert requests.
+        /// </summary>
+        /// <typeparam name="T">The type of top-level object to be deep inserted.</typeparam>
+        /// <param name="resource">The top-level object of the type to be deep inserted.</param>
+        internal virtual void DeepInsert<T>(T resource)
+        {
+            if (resource == null)
+            {
+                throw Error.ArgumentNull(nameof(resource));
+            }
+
+            DeepInsertSaveResult result = new DeepInsertSaveResult(this, Util.SaveChangesMethodName, SaveChangesOptions.DeepInsert, callback: null, state: null);
+            result.DeepInsertRequest(resource);
+        }
 
         #endregion
 
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -2341,7 +2341,7 @@ namespace Microsoft.OData.Client
 
         #region Add, Attach, Delete, Detach, Update, TryGetEntity, TryGetUri
 
-        /// <summary>Adds the specified link to the set of objects the <see cref="Microsoft.OData.Client.DataServiceContext" /> is tracking. The sourceProperty MUST be a collection navigation property.</summary>
+        /// <summary>Adds the specified link to the set of objects the <see cref="Microsoft.OData.Client.DataServiceContext" /> is tracking. The <paramref source="sourceProperty"/> MUST be a collection navigation property.</summary>
         /// <param name="source">The source object for the new link.</param>
         /// <param name="sourceProperty">The name of the navigation property on the source object that returns the related object.</param>
         /// <param name="target">The object related to the source object by the new link. </param>
@@ -2483,7 +2483,7 @@ namespace Microsoft.OData.Client
             }
         }
 
-        /// <summary>Adds the specified link to the set of objects the <see cref="Microsoft.OData.Client.DataServiceContext" /> is tracking. The sourceProperty MUST be a single-value navigation property.</summary>
+        /// <summary>Adds the specified link to the set of objects the <see cref="Microsoft.OData.Client.DataServiceContext" /> is tracking. The <paramref source="sourceProperty"/> MUST be a single-value navigation property.</summary>
         /// <param name="source">The source object for the new link.</param>
         /// <param name="sourceProperty">The name of the navigation property on the source object that returns the related object.</param>
         /// <param name="target">The object related to the source object by the new link.</param>
@@ -2566,7 +2566,7 @@ namespace Microsoft.OData.Client
             this.EntityTracker.IncrementChange(resource);
         }
 
-        /// <summary>Adds a related object to the context and creates the link that defines the relationship between the two objects in a single request. The sourceProperty MUST be a collection navigation property.</summary>
+        /// <summary>Adds a related object to the context and creates the link that defines the relationship between the two objects in a single request. The <paramref source="sourceProperty"/> MUST be a collection navigation property.</summary>
         /// <param name="source">The parent object that is being tracked by the context.</param>
         /// <param name="sourceProperty">The name of the navigation property that returns the related object based on an association between the two entities.</param>
         /// <param name="target">The related object that is being added.</param>
@@ -2625,7 +2625,7 @@ namespace Microsoft.OData.Client
             this.entityTracker.IncrementChange(targetResource);
         }
 
-        /// <summary>Adds a related object to the context and creates the link that defines the relationship between the two objects in a single request. The sourceProperty MUST be a single-value navigation property.</summary>
+        /// <summary>Adds a related object to the context and creates the link that defines the relationship between the two objects in a single request. The <paramref source="sourceProperty"/> MUST be a single-value navigation property.</summary>
         /// <param name="source">The parent object that is being tracked by the context.</param>
         /// <param name="sourceProperty">The name of the navigation property that returns the related object based on an association between the two entities.</param>
         /// <param name="target">The related object that is being added.</param>
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -2573,6 +2573,68 @@ namespace Microsoft.OData.Client
             this.entityTracker.IncrementChange(targetResource);
         }
 
+        /// <summary>Adds a related object to the context and creates the link that defines the relationship between the two objects in a single request.</summary>
+        /// <param name="source">The parent object that is being tracked by the context.</param>
+        /// <param name="sourceProperty">The name of the navigation property that returns the related object based on an association between the two entities.</param>
+        /// <param name="target">The related object that is being added.</param>
+        /// /// <remarks>
+        /// The sourceProperty MUST be a single-value object, not a collection.
+        /// </remarks>
+        public virtual void SetRelatedObject(object source, string sourceProperty, object target)
+        {
+            Util.CheckArgumentNull(source, "source");
+            Util.CheckArgumentNullAndEmpty(sourceProperty, "sourceProperty");
+            Util.CheckArgumentNull(target, "target");
+
+            // Validate that the source is an entity and is already tracked by the context.
+            ValidateEntityType(source, this.Model);
+
+            EntityDescriptor sourceResource = this.entityTracker.GetEntityDescriptor(source);
+
+            // Check for deleted source entity
+            if (sourceResource.State == EntityStates.Deleted)
+            {
+                throw Error.InvalidOperation(Strings.Context_SetRelatedObjectSourceDeleted);
+            }
+
+            // Validate that the property is valid and exists on the source
+            ClientTypeAnnotation parentType = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(source.GetType()));
+            ClientPropertyAnnotation property = parentType.GetProperty(sourceProperty, UndeclaredPropertyBehavior.ThrowException);
+            if (property.IsKnownType || property.IsEntityCollection)
+            {
+                throw Error.InvalidOperation(Strings.Context_SetRelatedObjectNonCollectionOnly);
+            }
+
+            // Validate that the target is an entity
+            ClientTypeAnnotation childType = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(target.GetType()));
+            ValidateEntityType(target, this.Model);
+
+            // Validate that the property type matches with the target type
+            ClientTypeAnnotation propertyElementType = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(property.PropertyType));
+            if (!propertyElementType.ElementType.IsAssignableFrom(childType.ElementType))
+            {
+                // target is not of the correct type
+                throw Error.Argument(Strings.Context_RelationNotRefOrCollection, "target");
+            }
+
+            var targetResource = new EntityDescriptor(this.model)
+            {
+                Entity = target,
+                State = EntityStates.Added
+            };
+
+            targetResource.SetParentForInsert(sourceResource, sourceProperty);
+
+            this.EntityTracker.AddEntityDescriptor(targetResource);
+
+            // Add the link in the added state.
+            LinkDescriptor end = targetResource.GetRelatedEnd();
+            end.State = EntityStates.Added;
+            this.entityTracker.AddLink(end);
+
+            this.entityTracker.IncrementChange(targetResource);
+        }
+
         /// <summary>Notifies the <see cref="Microsoft.OData.Client.DataServiceContext" /> to start tracking the specified resource and supplies the location of the resource within the specified resource set.</summary>
         /// <param name="entitySetName">The name of the set that contains the resource.</param>
         /// <param name="entity">The resource to be tracked by the <see cref="Microsoft.OData.Client.DataServiceContext" />. The resource is attached in the Unchanged state.</param>
--- a/src/Microsoft.OData.Client/DeepInsertSaveResult.cs
+++ b/src/Microsoft.OData.Client/DeepInsertSaveResult.cs
@@ -10,6 +10,8 @@ using System.Diagnostics;
 using System.Globalization;
 using System.IO;
 using System.Linq;
+using System.Text;
+using Microsoft.OData.Edm;
 
 namespace Microsoft.OData.Client
 {
@@ -124,7 +126,8 @@ namespace Microsoft.OData.Client
                             // We don't support delete and update in deep insert.
                             if (entityDescriptor.State == EntityStates.Deleted || entityDescriptor.State == EntityStates.Modified)
                             {
-                                throw Error.InvalidOperation(Strings.Context_DeepInsertDeletedOrModified);
+                                string entitySetAndKey = GetEntitySetAndKey(entityDescriptor);
+                                throw Error.InvalidOperation(Strings.Context_DeepInsertDeletedOrModified(entitySetAndKey));
                             }
 
                             bulkUpdateGraph.AddRelatedDescriptor(topLevelDescriptor, entityDescriptor);
@@ -135,17 +138,19 @@ namespace Microsoft.OData.Client
                     {
                         if (linkDescriptor.Source.Equals(topLevelDescriptor.Entity) && !bulkUpdateGraph.Contains(linkDescriptor))
                         {
+                            EntityDescriptor targetDescriptor = this.RequestInfo.Context.GetEntityDescriptor(linkDescriptor.Target);
+
                             // We don't support delete and update in deep insert.
                             if (linkDescriptor.State == EntityStates.Deleted || linkDescriptor.State == EntityStates.Modified)
                             {
-                                throw Error.InvalidOperation(Strings.Context_DeepInsertDeletedOrModified);
+                                string entitySetAndKey = GetEntitySetAndKey(targetDescriptor);
+                                throw Error.InvalidOperation(Strings.Context_DeepInsertDeletedOrModified(entitySetAndKey));
                             }
-
-                            EntityDescriptor targetDescriptor = this.RequestInfo.Context.GetEntityDescriptor(linkDescriptor.Target);
 
                             if (targetDescriptor != null && (targetDescriptor.State == EntityStates.Deleted || targetDescriptor.State == EntityStates.Modified))
                             {
-                                throw Error.InvalidOperation(Strings.Context_DeepInsertDeletedOrModified);
+                                string entitySetAndKey = GetEntitySetAndKey(targetDescriptor);
+                                throw Error.InvalidOperation(Strings.Context_DeepInsertDeletedOrModified(entitySetAndKey));
                             }
 
                             bulkUpdateGraph.AddRelatedDescriptor(topLevelDescriptor, linkDescriptor);
@@ -154,6 +159,15 @@ namespace Microsoft.OData.Client
                     }
                 }
             }
+        }
+
+        private string GetEntitySetAndKey(EntityDescriptor entityDescriptor)
+        {
+            var baseUriString = this.RequestInfo.BaseUriResolver.GetBaseUriWithSlash().ToString();
+            IEdmModel model = this.RequestInfo.Format.ServiceModel;
+            string entitySetUri = entityDescriptor.EditLink.ToString().TrimStart(baseUriString.ToCharArray());
+
+            return entitySetUri;
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Client/DeepInsertSaveResult.cs
+++ b/src/Microsoft.OData.Client/DeepInsertSaveResult.cs
@@ -101,13 +101,6 @@ namespace Microsoft.OData.Client
             {
                 EntityDescriptor topLevelDescriptor = this.RequestInfo.EntityTracker.GetEntityDescriptor(obj);
 
-                // If we do not find an EntityDescriptor in the entitytracker
-                // for any of the provided objects then we throw an exception. 
-                if (topLevelDescriptor == null)
-                {
-                    throw Error.InvalidOperation(Strings.Context_EntityNotContained);
-                }
-
                 // If it's a top-level object sent by the user.
                 if (isRootObject)
                 {
@@ -191,12 +184,6 @@ namespace Microsoft.OData.Client
         /// <returns>Returns the instance of ODataRequestMessage containing all the headers and payload for the delta request.</returns>
         private ODataRequestMessageWrapper GenerateDeepInsertRequest()
         {
-            if (this.bulkUpdateGraph.TopLevelDescriptors.Count == 0)
-            {
-                this.SetCompleted();
-                return null;
-            }
- 
             ODataRequestMessageWrapper deepInsertRequestMessage = this.CreateDeepInsertRequest(this.bulkUpdateGraph);
 
             // we need to fire request after the headers have been written, but before we write the payload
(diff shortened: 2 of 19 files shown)


src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,303 @@
...
+                    if (descriptor is EntityDescriptor entityDescriptor)
+                    {
+                        if (entityDescriptor.ParentEntity != null &&
+                            entityDescriptor.ParentEntity.Equals(topLevelDescriptor.Entity) &&
+                            !bulkUpdateGraph.Contains(entityDescriptor))
+                        {
+                            // We don't support delete and update in deep insert.
+                            if (entityDescriptor.State == EntityStates.Deleted || entityDescriptor.State == EntityStates.Modified)
+                            {
+                                throw Error.InvalidOperation(Strings.Context_DeepInsertDeletedOrModified);
```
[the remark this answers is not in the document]
habbes: Actually not sure.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,303 @@
...
+                            }
+
+                            bulkUpdateGraph.AddRelatedDescriptor(topLevelDescriptor, entityDescriptor);
+                            this.BuildDescriptorGraph(descriptors, false, entityDescriptor.Entity);
+                        }
+                        else
+                        {
+                            // We check if the link representation of the descriptor exists in the entity tracker and it is unchanged.
```
habbes, line 164: Why does it have to be unchanged?

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -358,12 +358,162 @@ internal void WriteEntry(EntityDescriptor entityDescriptor, IEnumerable<LinkDesc
...
+                {
+                    relatedEnd = item as LinkDescriptor;
+                }
+
+                navigationLink.Name = relatedEnd.SourceProperty;
+                bool isCollection = clientType.GetProperty(relatedEnd.SourceProperty, UndeclaredPropertyBehavior.ThrowException).IsEntityCollection;
+
+                odataWriter.WriteNestedResourceInfoStart(navigationLink);
+
+                odataWriter.WriteNestedResourceInfoStart(navigationLink, relatedEnd.Source, relatedEnd.Target);
```
habbes, line 443: Is it by design to call `WriteNestedResourceInfoStart` successively?

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertSaveResultTests.cs
```diff
@@ -0,0 +1,316 @@
+﻿//---------------------------------------------------------------------
```
habbes, line 1: Could you add tests that verify that `ReplaceUpdate` and `PostOnlySetProperties` are handled correctly?

habbes: Could you also add a test that deep inserts and entity that has direct or indirect link to an entity that's neither new, modified nor deleted, i.e. an entity that had already been synced with the server before the inserted entity was attached to the tracker

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertSaveResultTests.cs
```diff
@@ -0,0 +1,316 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="BulkUpdateSaveResultTests.cs" company="Microsoft">
```
habbes, line 2: Could we have tests that call `DeepInsert` directly, demonstrating how the user would interact with the feature?

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -358,12 +358,162 @@ internal void WriteEntry(EntityDescriptor entityDescriptor, IEnumerable<LinkDesc
...
         /// <summary>
         /// Writes the feed entry
         /// </summary>
-        /// <param name="descriptors">A list of descriptors to be written</param>
-        /// <param name="linkDescriptors">Descriptor links for some descriptors.</param>
-        /// /// <param name="bulkUpdateGraph">An instance of the <see cref="BulkUpdateGraph"/></param>
+        /// <param name="descriptors">A list of descriptors to be written.</param>
+        /// <param name="bulkUpdateGraph">An instance of the <see cref="BulkUpdateGraph"/>.</param>
+        /// <param name="oDataWriter">An instance of the <see cref="ODataWriterWrapper"/>.</param>
+        internal void WriteResourceSet(IReadOnlyList<Descriptor> descriptors, Dictionary<Descriptor, List<LinkDescriptor>> linkDescriptors, BulkUpdateGraph bulkUpdateGraph, ODataWriterWrapper oDataWriter)
```
gathogojr, line 474: You are missing a doc comment for `linkDescriptors` param.

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -358,12 +358,162 @@ internal void WriteEntry(EntityDescriptor entityDescriptor, IEnumerable<LinkDesc
...
+            ODataResourceSet resourceSet = new ODataResourceSet();
+            oDataWriter.WriteStart(resourceSet);
+
+            for (int i = 0; i < descriptors.Count; ++i)
+            {
+                Descriptor descriptor = descriptors[i];
+
+                if (descriptor is EntityDescriptor entityDescriptor)
+                {
+                    if (entityDescriptor.State == EntityStates.Deleted || entityDescriptor.State == EntityStates.Modified)
```
gathogojr, line 485: When building the graph, didn't we already check for this and throw exceptions where applicable? Do we need to perform the same checks here?

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -358,12 +358,162 @@ internal void WriteEntry(EntityDescriptor entityDescriptor, IEnumerable<LinkDesc
...
+                }
+            }
+
+            oDataWriter.WriteEnd();
+        }
+
+        /// <summary>
+        /// Writes the feed entry
+        /// </summary>
+        /// <param name="descriptors">A list of descriptors to be written.</param>
```
gathogojr, line 515: Ensure this method retains a doc comment for linkDescriptors param like it did before.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,303 @@
...
+        {
+            EntityDescriptor baseDescriptor = bulkUpdateGraph.TopLevelDescriptors[0];
+            Uri requestUri = UriUtil.CreateUri(this.RequestInfo.BaseUriResolver.GetBaseUriWithSlash(), UriUtil.CreateUri(bulkUpdateGraph.EntitySetName, UriKind.Relative));
+            string httpMethod = XmlConstants.HttpMethodPost;
+
+            HeaderCollection headers = new HeaderCollection();
+
+            this.RequestInfo.Context.Format.SetRequestContentTypeForEntry(headers);
+
+            headers.SetHeader("OData-Version", "4.01");
```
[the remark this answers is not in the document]
ElizabethOkerio: Yes. They are already existing.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,303 @@
...
+        /// </summary>
+        /// <typeparam name="T"> The type of the top-level object to be deep inserted.</typeparam>
+        /// <param name="resource"> The top-level object to be deep inserted.</param>
+        internal void DeepInsertRequest<T>(T resource)
+        {
+            BuildDescriptorGraph(this.ChangedEntries, true, resource);
+
+            ODataRequestMessageWrapper deeepInsertRequestMessage = this.GenerateDeepInsertRequest();
+
+            if (deeepInsertRequestMessage == null)
```
ElizabethOkerio, line 92: nit: 
```suggestion
            if (deepInsertRequestMessage == null)

```

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,303 @@
...
+        /// <summary>
+        /// This method processes all the changed descriptors in the entity tracker.
+        /// It loops through all the descriptors and creates relationships between the descriptors if any.
+        /// </summary>
+        /// <typeparam name="T">The type of the top-level resource we want to do a deep insert.</typeparam>
+        /// <param name="descriptors">The list of descriptors in the entity tracker.</param>
+        /// <param name="isRootObject">True if the <paramref name="objects"/> are root or top-level objects, i.e., they were passed directly by the client, otherwise false.</param>
+        /// <param name="objects">Objects of the top-level type we want to perform a deep insert on.</param>
+        /// <exception cref="InvalidOperationException"></exception>
+        internal void BuildDescriptorGraph<T>(IEnumerable<Descriptor> descriptors, bool isRootObject, params T[] objects)
```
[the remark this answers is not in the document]
ElizabethOkerio: I think the `isRootObject` is what guides us in adding top-level descriptors to the BulkUpdateGraph's top-level descriptors. If it is `true` then we know that's the top-level object that was set by the client but if called from within the method then the `isRootObject` is set to `false`

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,303 @@
...
+                foreach (Descriptor descriptor in descriptors)
+                {
+                    if (descriptor is EntityDescriptor entityDescriptor)
+                    {
+                        if (entityDescriptor.ParentEntity != null &&
+                            entityDescriptor.ParentEntity.Equals(topLevelDescriptor.Entity) &&
+                            !bulkUpdateGraph.Contains(entityDescriptor))
+                        {
+                            // We don't support delete and update in deep insert.
+                            if (entityDescriptor.State == EntityStates.Deleted || entityDescriptor.State == EntityStates.Modified)
```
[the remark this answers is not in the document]
ElizabethOkerio: If the existing entity is unchanged it will not be part of the `ChangedDescriptors` to be considered for a deep insert.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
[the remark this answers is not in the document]
ElizabethOkerio: I think this can be documented so that users are aware that a deep insert is only used for create operations and linking of existing resources to newly created resources. If a user has this kind of information then I think supporting the second option of not throwing an exception then the user being required to call `SaveChanges` to save the other operations that were not taken care of by the `DeepInsert.`  My thoughts though.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,303 @@
...
+                            }
+
+                            bulkUpdateGraph.AddRelatedDescriptor(topLevelDescriptor, entityDescriptor);
+                            this.BuildDescriptorGraph(descriptors, false, entityDescriptor.Entity);
+                        }
+                        else
+                        {
+                            // We check if the link representation of the descriptor exists in the entity tracker and it is unchanged. 
+                            LinkDescriptor linkDescriptor = this.RequestInfo.EntityTracker.Links.Where(
+                                a => a.Source == topLevelDescriptor.Entity &&
```
ElizabethOkerio, line 166: I think this check is not necessary in a `DeepInsert`. So in a deep insert I believe users have to call either `AddLink` or `SetLink` in order to link a resource to an existing link. If they don't do that I don't think we need to worry ourselves with whatever they fetched using `$expand`

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,303 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Gets the link descriptor representations for the provided descriptor.
+        /// </summary>
+        /// <param name="descriptor">The descriptor.</param>
+        /// <returns>All the link descriptors to a given key descriptor.</returns>
+        public List<LinkDescriptor> GetLinkDescriptors(Descriptor descriptor)
```
ElizabethOkerio, line 214: I don't think we do not need this information in a `DeepInsert`

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,303 @@
...
+using System.Globalization;
+using System.IO;
+using System.Linq;
+
+namespace Microsoft.OData.Client
+{
+    /// <summary>
+    /// Handles the deep insert requests and responses (both sync and async).
+    /// </summary>
+    internal class DeepInsertSaveResult : BaseSaveResult
```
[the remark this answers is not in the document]
KenitoInc: Future PR

src/Microsoft.OData.Client/Serialization/Serializer.cs
KenitoInc: Yes. Those 2 methods do different things. It's how the `ODataWriterWrapper` was designed.

```csharp
        internal void WriteNestedResourceInfoStart(ODataNestedResourceInfo navigationLink, object source, object target)
        {
            this.requestPipeline.ExecuteOnNestedResourceInfoStartActions(navigationLink, source, target);
        }

        internal void WriteNestedResourceInfoStart(ODataNestedResourceInfo navigationLink)
        {
            this.odataWriter.WriteStart(navigationLink);
        }
```

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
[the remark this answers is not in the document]
KenitoInc: `BuildDescriptorGraph` method is called recursively that's why we have `params T[]`

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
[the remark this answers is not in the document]
KenitoInc: `SerializeEntry_With_DeleteLink_Fails`
`SerializeEntry_With_UpdateObject_Fails`

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
KenitoInc: Got it.
Fixed

KenitoInc:
```csharp
var customer = new Customer
{
    Id = 999,
    Name = "Customer 999",
    Age = 50
};

var customer2 = context.Customers.Expand("Orders").FirstOrDefault();
var order1 = customer2.Orders.FirstOrDefault();

context.AttachTo("Customers", customer);
context.AddLink(customer, "Orders", order1);

context.DeepInsert<Customer>(customer);
```

src/Microsoft.OData.Client/Serialization/Serializer.cs
KenitoInc: refactored to remove `linkDescriptors` parameter

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -358,12 +358,162 @@ internal void WriteEntry(EntityDescriptor entityDescriptor, IEnumerable<LinkDesc
...
+            for (int i = 0; i < descriptors.Count; ++i)
+            {
+                Descriptor descriptor = descriptors[i];
+
+                if (descriptor is EntityDescriptor entityDescriptor)
+                {
+                    if (entityDescriptor.State == EntityStates.Deleted || entityDescriptor.State == EntityStates.Modified)
+                    {
+                        // We don't support Delete or Update in Deep insert
+                        // Throw exception?
```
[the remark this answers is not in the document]
KenitoInc: removed this check

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
KenitoInc: Removed this code. It's not needed

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -687,7 +816,7 @@ internal void WriteDeltaNestedResourceInfo(EntityDescriptor entityDescriptor, Di
                 {
                     string parentProperty = entityObject.ParentProperty;
 
-                    if (string.IsNullOrEmpty(parentProperty) && descriptorLinks.Count > 0)
+                    if (string.IsNullOrEmpty(parentProperty) && descriptorLinks != null && descriptorLinks.Count > 0)
```
gathogojr, line 819:
```suggestion
                    if (string.IsNullOrEmpty(parentProperty) && descriptorLinks?.Count > 0)
```

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -358,12 +358,162 @@ internal void WriteEntry(EntityDescriptor entityDescriptor, IEnumerable<LinkDesc
...
+
+            IEnumerable<ClientPropertyAnnotation> properties;
+            if ((!Util.IsFlagSet(this.options, SaveChangesOptions.ReplaceOnUpdate) &&
+                entityDescriptor.State == EntityStates.Modified &&
+                entityDescriptor.PropertiesToSerialize.Any()) ||
+                (Util.IsFlagSet(this.options, SaveChangesOptions.PostOnlySetProperties) &&
+                entityDescriptor.State == EntityStates.Added))
+            {
+                properties = entityType.PropertiesToSerialize().Where(prop => entityDescriptor.PropertiesToSerialize.Contains(prop.PropertyName));
+            }
```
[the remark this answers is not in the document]
KenitoInc: refactored this code

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,242 @@
...
+                            bulkUpdateGraph.AddRelatedDescriptor(topLevelDescriptor, entityDescriptor);
+                            this.BuildDescriptorGraph(descriptors, false, entityDescriptor.Entity);
+                        }
+                    }
+                    else if (descriptor is LinkDescriptor linkDescriptor)
+                    {
+                        if (linkDescriptor.Source.Equals(topLevelDescriptor.Entity) && !bulkUpdateGraph.Contains(linkDescriptor))
+                        {
+                            // We don't support delete and update in deep insert.
+                            if (linkDescriptor.State == EntityStates.Deleted || linkDescriptor.State == EntityStates.Modified)
```
ElizabethOkerio, line 146: I think we need to add another check for SaveResultProcessed to check if the descriptor has been processed or not..

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertSaveResultTests.cs
KenitoInc: Actually we shouldn't be supporting `ReplaceUpdate`and `PostOnlySetProperties` in deep insert. So I have removed the codes

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertSaveResultTests.cs
KenitoInc: I will add them in the `response` PR to test end to end.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,303 @@
...
+                                this.BuildDescriptorGraph(descriptors, false, entityDescriptor.Entity);
+                            }
+                        }
+                    }
+                    else if (descriptor is LinkDescriptor link)
+                    {
+                        if (link.Source.Equals(topLevelDescriptor.Entity) && !bulkUpdateGraph.Contains(link))
+                        {
+                            // We don't support delete and update in deep insert.
+                            if (link.State == EntityStates.Deleted || link.State == EntityStates.Modified)
```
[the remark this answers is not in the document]
KenitoInc: `SerializeEntry_With_DeleteLink_Fails`

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,303 @@
...
+                            {
+                                throw Error.InvalidOperation(Strings.Context_DeepInsertDeletedOrModified);
+                            }
+
+                            EntityDescriptor targetDescriptor = this.RequestInfo.Context.GetEntityDescriptor(link.Target);
+
+                            if (targetDescriptor != null && (targetDescriptor.State == EntityStates.Deleted || targetDescriptor.State == EntityStates.Modified))
```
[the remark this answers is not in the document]
KenitoInc: Will test this in E2E tests in the next PR

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -358,12 +358,162 @@ internal void WriteEntry(EntityDescriptor entityDescriptor, IEnumerable<LinkDesc
...
+            ClientTypeAnnotation clientType = model.GetClientTypeAnnotation(model.GetOrCreateEdmType(entityDescriptor.Entity.GetType()));
+
+            foreach (KeyValuePair<string, List<Descriptor>> relatedLinks in groupedRelatedLinks)
+            {
+                LinkDescriptor relatedEnd = null;
+                ODataNestedResourceInfo navigationLink = new ODataNestedResourceInfo();
+
+                //these items will all be of the same type ~ we'll be using the
+                //first item in determining the type of the descriptor.
+                Descriptor item = relatedLinks.Value.FirstOrDefault();
```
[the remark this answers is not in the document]
KenitoInc: Yes the list is guaranteed  to be non empty.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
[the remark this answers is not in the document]
habbes: To clarify the question: Let's assume we have an existing entity A in the tracker. This entity is neither new, nor deleted, nor modified. It was there already. Then we create a new entity B and link it to A. And we perform a deep insert on B. What would happen?

@ElizabethOkerio If I understand what you're saying, since `A` is unchanged, it would not be in the `ChangedDescriptors`. So does that mean we'll insert `B` but leave out `A`? So the link between `B` and `A` will not be persisted on the server?

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,303 @@
...
+        /// <exception cref="InvalidOperationException"></exception>
+        internal void BuildDescriptorGraph<T>(IEnumerable<Descriptor> descriptors, bool isRootObject, params T[] objects)
+        {
+            foreach (object obj in objects)
+            {
+                EntityDescriptor topLevelDescriptor = this.RequestInfo.EntityTracker.GetEntityDescriptor(obj);
+
+                // If we do not find an EntityDescriptor in the entitytracker
+                // for any of the provided objects then we throw an exception. 
+                if (topLevelDescriptor == null)
```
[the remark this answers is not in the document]
KenitoInc: Added a test `SerializeEntry_With_No_TopLevelDescriptor_Fails`
However actually that exception is thrown inside the `GetEntityDescriptor(obj);`

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        /// <summary>
+        /// Generate the deep insert request for all changes to save.
+        /// </summary>
+        /// <returns>Returns the instance of ODataRequestMessage containing all the headers and payload for the delta request.</returns>
+        private ODataRequestMessageWrapper GenerateDeepInsertRequest()
+        {
+            if (this.bulkUpdateGraph.TopLevelDescriptors.Count == 0)
+            {
```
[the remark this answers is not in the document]
KenitoInc: Removed this code. Validation has happened earlier.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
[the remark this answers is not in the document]
KenitoInc: If A exists, we don't need to insert it.
We will do a Deep update on A so as to link the new entity B

`PATCH ~/As`
```json
{
    "@odata.context":"As/$delta",
    "value":[
        {
            "@odata.type":"Namespace.B",
            "Id":1,
            "Bs@odata.delta":[
                {
                    "@odata.id":"B(1)"
                }
            ]
        }
    ]
} 
```

In OData client
```csharp
var a= context.As.Where(x => x.Id == 1).FirstOrDefault();

var b= new B{ Id = 1001 };
context.AttachTo("Bs", b);
context.AddLink(a, "Bs", b);

context.BulkUpdate(a);
```

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
KenitoInc: In BaseSaveResult, we initialize `ChangedEntries`, which we use to create the request payload.

```csharp
this.ChangedEntries = context.EntityTracker.Entities.Cast<Descriptor>()
                                      .Union(context.EntityTracker.Links.Cast<Descriptor>())
                                      .Union(context.EntityTracker.Entities.SelectMany(e => e.StreamDescriptors).Cast<Descriptor>())
                                      .Where(o => o.IsModified && o.ChangeOrder != UInt32.MaxValue)
                                      .OrderBy(o => o.ChangeOrder)
                                      .ToList();
```

The `ChangedEntries` contain descriptors whose state `IsModified` == `true` so we ignore the `Unchanged` descriptors.
Once a result is processed, the descriptor state is changed to `Unchanged`. This ensures that the descriptor is not processed again in a subsequent `SaveChanges()` request.

In DeepInsert, we initialize the `BulkUpdateGraph` using the `ChangedEntries`.

```csharp
internal void DeepInsertRequest<T>(T resource)
{
    BuildDescriptorGraph(this.ChangedEntries, true, resource);

    .....

}
```

Example:
If we have delete and update operations which we then call SaveChanges(), those descriptors will not be part of the subsequent DeepInsert request.

KenitoInc: I will add E2E tests in the Deep insert response PR

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertSaveResultTests.cs
```diff
@@ -0,0 +1,370 @@
...
+
+            MemoryStream stream = (MemoryStream)(odataRequestMessageWrapper.CachedRequestStream.Stream);
+            stream.Position = 0;
+
+            string payload = (new StreamReader(stream)).ReadToEnd();
+            payload.Should().Be("{\"ID\":100,\"Name\":\"Bing\"}");
+        }
+
+        [Fact]
+        public void SerializEntry_With4LevelsOfNesting()
```
gathogojr, line 115: Is it 4 or 3 levels of nesting?

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertSaveResultTests.cs
```diff
@@ -0,0 +1,370 @@
...
+        {
+            var person = new Person
+            {
+                ID = 100,
+                Name = "Bing",
+            };
+
+            var cars = new List<Car>
+            {
+                new Car{
```
gathogojr, line 212: nit:
```suggestion
                new Car
                {
```

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -358,12 +358,134 @@ internal void WriteEntry(EntityDescriptor entityDescriptor, IEnumerable<LinkDesc
...
+        /// <param name="entityDescriptor">The entity.</param>
+        /// <param name="bulkUpdateGraph">An instance of the <see cref="BulkUpdateGraph"/>.</param>
+        /// <param name="odataWriter">The ODataWriter used to write the navigation link.</param>
+        internal void WriteNestedResourceInfo(EntityDescriptor entityDescriptor, BulkUpdateGraph bulkUpdateGraph, ODataWriterWrapper odataWriter)
+        {
+            List<Descriptor> relatedDescriptors = bulkUpdateGraph.GetRelatedDescriptors(entityDescriptor);
+            Dictionary<string, List<Descriptor>> groupedRelatedLinks = GroupRelatedLinksByNavigationProperty(relatedDescriptors, null);
+
+            ClientEdmModel model = this.requestInfo.Model;
+            ClientTypeAnnotation clientType = model.GetClientTypeAnnotation(model.GetOrCreateEdmType(entityDescriptor.Entity.GetType()));
```
gathogojr, line 408: The caller of this method does make these very same calls... Why can't you just pass the `ClientTypeAnnotation` into this method?

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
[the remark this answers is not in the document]
ElizabethOkerio: @KenitoInc  what was agreed on here? Are you throwing exceptions?

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertSaveResultTests.cs
```diff
@@ -0,0 +1,370 @@
...
+                                  .OrderBy(o => o.ChangeOrder)
+                                  .ToList();
+
+            var result = new DeepInsertSaveResult(this.context, Util.SaveChangesMethodName, SaveChangesOptions.DeepInsert, null, null);
+            result.BuildDescriptorGraph<T>(changedEntries, true, (T[])objects);
+            this.bulkUpdateGraph = result.BulkUpdateGraph;
+
+            return this.bulkUpdateGraph;
+        }
+
```
gathogojr, line 276: @KenitoInc Can you add a test where the nested navigation properties are single-valued?

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,229 @@
...
+            EntityDescriptor baseDescriptor = bulkUpdateGraph.TopLevelDescriptors[0];
+            Uri requestUri = UriUtil.CreateUri(this.RequestInfo.BaseUriResolver.GetBaseUriWithSlash(), UriUtil.CreateUri(bulkUpdateGraph.EntitySetName, UriKind.Relative));
+            string httpMethod = XmlConstants.HttpMethodPost;
+
+            HeaderCollection headers = new HeaderCollection();
+
+            this.RequestInfo.Context.Format.SetRequestContentTypeForEntry(headers);
+
+            headers.SetHeader(XmlConstants.HttpODataVersion, ODataVersion.V401.ToString());
+            headers.SetHeader(XmlConstants.HttpODataMaxVersion, ODataVersion.V401.ToString());
```
gathogojr, line 174:
```suggestion
            headers.SetHeader(XmlConstants.HttpODataMaxVersion, XmlConstants.ODataVersion401);
```

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,229 @@
...
+                    }
+                }
+            }
+        }
+
+        /// <summary>
+        /// Creates an ODataRequestMessage for a deep insert request.
+        /// </summary>
+        /// <returns>Returns an instance of the ODataRequestMessage for the deep insert request.</returns>
+        private ODataRequestMessageWrapper CreateDeepInsertRequest(BulkUpdateGraph bulkUpdateGraph)
```
gathogojr, line 163: @KenitoInc I don't think any of your tests hit this method. The lines starting with `headers.SetHeader` both throw the following exception:
```csharp
The value 'V401' of the OData-Version HTTP header is invalid. Only '4.0' and '4.01' are supported as values for the OData-Version header.
```
Without pulling your branch and testing out the feature I wouldn't have caught this... If you had tests covering the changes in this PR, you would have caught this.

gathogojr (dismissed):
@KenitoInc Given:
```csharp
public class Order
{
    public int Id { get; set; }
    public Customer Customer { get; set; }
}

public class Customer
{
    public int Id { get; set; }
}
```

How can I do a deep insert for an `Order` object with a nested `Customer`? Also add tests for the scenario.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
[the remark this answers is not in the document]
KenitoInc: 1) The customer should do a deep update. 
2) Bulk update supports updates, inserts, deletes

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -358,12 +358,134 @@ internal void WriteEntry(EntityDescriptor entityDescriptor, IEnumerable<LinkDesc
...
         /// <summary>
         /// Writes the feed entry
         /// </summary>
-        /// <param name="descriptors">A list of descriptors to be written</param>
-        /// <param name="linkDescriptors">Descriptor links for some descriptors.</param>
-        /// /// <param name="bulkUpdateGraph">An instance of the <see cref="BulkUpdateGraph"/></param>
+        /// <param name="descriptors">A list of descriptors to be written.</param>
+        /// <param name="bulkUpdateGraph">An instance of the <see cref="BulkUpdateGraph"/>.</param>
+        /// <param name="oDataWriter">An instance of the <see cref="ODataWriterWrapper"/>.</param>
+        internal void WriteResourceSet(IReadOnlyList<Descriptor> descriptors, BulkUpdateGraph bulkUpdateGraph, ODataWriterWrapper oDataWriter)
```
gathogojr, line 462: Does this method need to be `internal`? Can it be `private` instead?

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -358,12 +358,134 @@ internal void WriteEntry(EntityDescriptor entityDescriptor, IEnumerable<LinkDesc
...
+            {
+                Descriptor descriptor = descriptors[i];
+
+                if (descriptor is EntityDescriptor entityDescriptor)
+                {
+                    this.WriteDeepInsertEntry(entityDescriptor, bulkUpdateGraph, oDataWriter);
+                }
+                else if (descriptor is LinkDescriptor linkDescriptor)
+                {
+                    WriteODataId(linkDescriptor, oDataWriter);
```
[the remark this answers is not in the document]
KenitoInc: `SerializeEntry_With_TwoNavigationLinks`

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -358,12 +358,134 @@ internal void WriteEntry(EntityDescriptor entityDescriptor, IEnumerable<LinkDesc
...
+
+                odataWriter.WriteStart(navigationLink, relatedEnd.Source, relatedEnd.Target);
+
+
+                if (isCollection)
+                {
+                    WriteResourceSet(relatedLinks.Value, bulkUpdateGraph, odataWriter);
+                }
+                else
+                {
```
gathogojr, line 439: Which scenario leads to this code block being executed? Is there a test I can run to get this block executed?

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -358,12 +358,134 @@ internal void WriteEntry(EntityDescriptor entityDescriptor, IEnumerable<LinkDesc
...
+                //these items will all be of the same type ~ we'll be using the
+                //first item in determining the type of the descriptor.
+                Descriptor item = relatedLinks.Value.FirstOrDefault();
+
+                if (item is EntityDescriptor entDescriptor)
+                {
+                    relatedEnd = new LinkDescriptor(entDescriptor.ParentEntity, entDescriptor.ParentProperty, entDescriptor.Entity, this.requestInfo.Model);
+                }
+                else
+                {
```
gathogojr, line 424: Which scenario leads to this code block being executed? Is there a test I can run to get this block executed?

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
KenitoInc: This should be covered in E2E tests

src/Microsoft.OData.Client/Serialization/Serializer.cs
KenitoInc: Run this test
`SerializeEntry_With_TwoNavigationLinks`

src/Microsoft.OData.Client/Serialization/Serializer.cs
KenitoInc: Done

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
gathogojr: @KenitoInc If that is the case, leave it for PR where you'll add the tests

src/Microsoft.OData.Client/Serialization/Serializer.cs
KenitoInc: Single value navigation property

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertSaveResultTests.cs
KenitoInc: `SerializeEntry_With_SingleValueNavigation`
`SerializeEntry_With_SingleValueNavigationLink`

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertSaveResultTests.cs
KenitoInc: Fixed

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -0,0 +1,229 @@
...
+                    if (descriptor is EntityDescriptor entityDescriptor)
+                    {
+                        if (entityDescriptor.ParentEntity != null &&
+                            entityDescriptor.ParentEntity.Equals(topLevelDescriptor.Entity) &&
+                            !bulkUpdateGraph.Contains(entityDescriptor))
+                        {
+                            // We don't support delete and update in deep insert.
+                            if (entityDescriptor.State == EntityStates.Deleted || entityDescriptor.State == EntityStates.Modified)
+                            {
+                                throw Error.InvalidOperation(Strings.Context_DeepInsertDeletedOrModified);
```
habbes, line 127: I propose including the ID of the of offending entity to the error message to make it easier for the user to debug and track down which entity caused the exception.

habbes:
My concern with this feature is that it goes out of the flow of how users normally interact with the library to update objects (`SaveChanges`). The use of `DeepInsert` comes with additional restrictions that requires to keep track of the entities they're linking to and ensure they have not been modified. I think these two points negatively impact the developer experience to the point that I think most users would prefer to stick to `SaveChanges` API if they have the choice (`$batch` requests are more likely to be supported by the server anyway).

This probably beyond the scope of this PR, but I think (as @g2mula had suggested) we should think of a way to unify these two APIs such that the library figures out which API call to make without burdening the user with that detail.

Here's a high-level view of how we could achieve that:
- Instead of exposing different methods for `DeepInsert` or `DeepUpdate`, we add new flag to `SaveChangesOptions` e.g. `PreferDeepInsert` or `AllowDeepInsert` or something along those lines
- The user will add objects and links as they do normally, using existing APIs like `AddObject`, `AddReleatedObject`, `DataServiceCollection`, etc.
- To sync with the server they call `SaveChangesAsync(SaveChangeOptions.PreferDeepInsert)`
- When that flag is set, the client will try to perform a bulk insert (or bulk update depending if enabled) if it deems it possible, otherwise it defaults to making a batch call or multiple requests to the server.
- How does the client determine whether a bulk update is possible? A naive way is to visit all the changed entries, if they the form a single connected graph, then we know they are all interconnected and can be combined in a single bulk call. In that case it finds a possible root object and makes a bulk update with that as the root. It can use states of filtered objects to decide whether to do an insert or update.
- If the objects do not meet condition (one connected graph), then it falls back to what `SaveChangesAsync` normally does
- This algo could be refined to support multiple root objects of a single bulk update call or multiple update calls if we have multiple disjoint graphs.

ElizabethOkerio:
@habbes I think leaving the library to determine the root object might be tricky because we can have several objects that have connections but they are not the specific resource type that the client wants to deep-insert. Also  deep-insert API will be of a particular type for example. This can lead to several deep insert requests being created for endpoints that do not even exist. I see this approach to be error-prone.

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertSaveResultTests.cs
```diff
@@ -0,0 +1,443 @@
...
+            };
+
+            var car1 = new Car { ID = 1001 };
+            var car2 = new Car { ID = 1002 };
+
+            this.context.AttachTo("Persons", person);
+            this.context.AttachTo("Cars", car1);
+            this.context.AttachTo("Cars", car2);
+            this.context.AddLink(person, "Cars", car1);
+            this.context.AddLink(person, "Cars", car2);
```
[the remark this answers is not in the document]
KenitoInc: I have added an `assert` to confirm it works when the entities are `unchanged`

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
KenitoInc: This is a non-trivial fix. Also  ID alone cannot determine a specific entity since entities might share Ids e.g Customer(1), Order(1)

habbes:
> @habbes I think leaving the library to determine the root object might be tricky because we can have several objects that have connections but they are not the specific resource type that the client wants to deep-insert. Also deep-insert API will be of a particular type for example. This can lead to several deep insert requests being created for endpoints that do not even exist. I see this approach to be error-prone.

@ElizabethOkerio 
Currently, when you call `SaveChangesAsync`, you don't select which objects get sent to the server and which ones do not. So the argument you make about there being resources that may be saved other than the ones the customer wants to deep insert can also be made for the way the client currently works even without deep insert. Even now, it's possible for the client to make `POST` and `PATCH` requests for endpoints which do not exist on the server.

In my suggestion (at least for a conceptual first iteration), I propose for deep insert to be used only if the customer enables the flag and all the changed objects are in the same subgraph such that they could be fulfilled by a single deep insert/update. There would still be a challenge to determine what is/are the root objects (esp. if there are cycles). But we would transparently fall back to the a batch or multi-request approach if these conditions are not met.

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2483,6 +2483,61 @@ public virtual void SetLink(object source, string sourceProperty, object target)
...
+        /// <param name="source">The source object for the new link.</param>
+        /// <param name="sourceProperty">The name of the navigation property on the source object that returns the related object.</param>
+        /// <param name="target">he object related to the source object by the new link.</param>
+        /// <exception cref="System.ArgumentNullException">When <paramref name="source" />, <paramref name="sourceProperty" /> or <paramref name="target" /> are null.</exception>
+        /// <exception cref="System.InvalidOperationException">When the specified link already exists.-or-When the objects supplied as <paramref name="source" /> or <paramref name="target" /> are in the <see cref="Microsoft.OData.Client.EntityStates.Detached" /> or <see cref="Microsoft.OData.Client.EntityStates.Deleted" /> state.-or-When <paramref name="sourceProperty" /> is not a navigation property that defines a reference to a single related object.</exception>
+        /// <remarks>
+        /// The sourceProperty MUST be a single-value navigation property.
+        /// </remarks>
+        public virtual void SetRelatedObjectLink(object source, string sourceProperty, object target)
```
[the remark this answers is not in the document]
habbes: Another question about this method, it does fewer validation checks than the `EnsureRelatable` method used by `AddLink`. Is there a reason we don't do the same set of checks? Would it make sense to re-use `EnsureRelatable` and maybe pass a flag to determine whether the check is for a single or collection navigation property?

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2573,6 +2628,70 @@ public virtual void AddRelatedObject(object source, string sourceProperty, objec
...
+            ClientTypeAnnotation childType = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(target.GetType()));
+            ValidateEntityType(target, this.Model);
+
+            // Validate that the property type matches with the target type
+            ClientTypeAnnotation propertyElementType = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(property.PropertyType));
+
+            if (!propertyElementType.ElementType.IsAssignableFrom(childType.ElementType))
+            {
+                // target is not of the correct type
+                throw Error.Argument(Strings.Context_RelationNotRefOrCollection, "target");
```
habbes, line 2674:
```suggestion
                throw Error.Argument(Strings.Context_RelationNotRefOrCollection, nameof(target));

```

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2573,6 +2628,70 @@ public virtual void AddRelatedObject(object source, string sourceProperty, objec
...
         }
 
+        /// <summary>Adds a related object to the context and creates the link that defines the relationship between the two objects in a single request.</summary>
+        /// <param name="source">The parent object that is being tracked by the context.</param>
+        /// <param name="sourceProperty">The name of the navigation property that returns the related object based on an association between the two entities.</param>
+        /// <param name="target">The related object that is being added.</param>
+        /// /// <remarks>
+        /// The sourceProperty MUST be a single-value object, not a collection.
+        /// </remarks>
+        public virtual void SetRelatedObject(object source, string sourceProperty, object target)
```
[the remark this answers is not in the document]
habbes: So the only difference between this and `AddRelatedObject` is that this method only supports single-value navigation properties whereas `AddRelatedObject` only supports collection navigation properties? Could these not have been unified into a single method or would that be a breaking change?

Though I don't think it's a deal-breaker, I think there's a potential to abstract the common parts into a common method since the methods are almost identical, all except 2 lines or so.

That said, the summaries of the methods don't make it more confusing since they are identical. I think we could move the distinction to the summary instead of being in the remarks section of the xml doc, since that restriction is the reason we have two separate methods instead of one.

I think it would be helpful to have names that make the distinction between the two clearer. Is it possible to consider a name that reflects that this method only applies to single value properties? Like `AddSingleRelatedObject`?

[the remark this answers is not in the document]
habbes: On second thought, with the suggested `AddSingleRelatedObject`, one could also think that we are adding a single related object to a collection property. So it may not be sufficient to disambiguate between the two methods.

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2852,6 +2971,22 @@ public virtual bool TryGetUri(object entity, out Uri identity)
 
             return identity != null;
         }
+        
+        /// <summary>
+        /// Processes deep insert requests.
```
habbes, line 2976: This description seems to assume that the user knows what a deep insert request is. I think it should be more descriptive about what the method does without making such assumptions.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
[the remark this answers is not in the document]
habbes: If `BulkUpdate` supports insertions and has fewer restrictions, why should the customer ever call `DeepInsert`?

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
habbes: @KenitoInc When I say ID, I meant including the entity type, just like you have shown: `Customer(1)` or `Order(1)`

habbes: I'm curious about the effort involved in getting the ID here. Since we have the entity descriptor, could we not use the `descriptor.Identity` url property then (if it exists)? Since it's basically an identifier that can be used to narrow down the specific entity that causes the exception.

src/Microsoft.OData.Client/DataServiceContext.cs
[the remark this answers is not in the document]
KenitoInc: 1) The target property is the difference. I created a new method since I did not want to change how the AddLink method works.
2) We don't need to call `EnsureRelatable` since we are doing the checks inside the method.

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2483,6 +2483,61 @@ public virtual void SetLink(object source, string sourceProperty, object target)
...
+            }
+
+            // Validate that the property is valid and exists on the source
+            ClientTypeAnnotation type = this.model.GetClientTypeAnnotation(this.model.GetOrCreateEdmType(source.GetType()));
+            Debug.Assert(type.IsEntityType, "should be an entity type");
+
+            // will throw InvalidOperationException if property doesn't exist
+            ClientPropertyAnnotation property = type.GetProperty(sourceProperty, UndeclaredPropertyBehavior.ThrowException);
+
+            if (property.IsKnownType || property.IsEntityCollection)
```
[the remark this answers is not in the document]
KenitoInc: `IsKnownType` means it's a known primitive type e.g string, byte etc

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2483,6 +2483,61 @@ public virtual void SetLink(object source, string sourceProperty, object target)
...
+            Util.CheckArgumentNullAndEmpty(sourceProperty, "sourceProperty");
+            Util.CheckArgumentNull(target, "target");
+
+            // Validate that the source is an entity and is already tracked by the context.
+            ValidateEntityType(source, this.Model);
+
+            EntityDescriptor sourceResource = this.entityTracker.GetEntityDescriptor(source);
+
+            // Check for deleted source entity
+            if (sourceResource.State == EntityStates.Deleted)
```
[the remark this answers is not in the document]
KenitoInc: Validations for the target should happen elsewhere depending on the request we are making

src/Microsoft.OData.Client/DataServiceContext.cs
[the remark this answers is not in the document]
KenitoInc: I have updated the summary

src/Microsoft.OData.Client/DataServiceContext.cs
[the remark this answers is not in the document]
KenitoInc: I have updated the summary

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
KenitoInc: Fixed

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2573,6 +2625,67 @@ public virtual void AddRelatedObject(object source, string sourceProperty, objec
             this.entityTracker.IncrementChange(targetResource);
         }
 
+        /// <summary>Adds a related object to the context and creates the link that defines the relationship between the two objects in a single request. The sourceProperty MUST be a single-value navigation property.</summary>
```
habbes, line 2628:
```suggestion
        /// <summary>Adds a related object to the context and creates the link that defines the relationship between the two objects in a single request. The <paramref name="sourceProperty"/> MUST be a single-value navigation property.</summary>

```

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2341,7 +2341,7 @@ internal virtual DataServiceResponse EndBulkUpdate(IAsyncResult asyncResult)
 
         #region Add, Attach, Delete, Detach, Update, TryGetEntity, TryGetUri
 
-        /// <summary>Adds the specified link to the set of objects the <see cref="Microsoft.OData.Client.DataServiceContext" /> is tracking.</summary>
+        /// <summary>Adds the specified link to the set of objects the <see cref="Microsoft.OData.Client.DataServiceContext" /> is tracking. The sourceProperty MUST be a collection navigation property.</summary>
```
habbes, line 2344:
```suggestion
        /// <summary>Adds the specified link to the set of objects the <see cref="Microsoft.OData.Client.DataServiceContext" /> is tracking. The <paramref source="sourceProperty"/> MUST be a collection navigation property.</summary>

```

habbes (approved):
Approved since the feature is still internal. Looking forward to seeing the e2e tests.

src/Microsoft.OData.Client/Serialization/Serializer.cs
```diff
@@ -358,12 +358,132 @@ internal void WriteEntry(EntityDescriptor entityDescriptor, IEnumerable<LinkDesc
...
+                navigationLink.Name = relatedEnd.SourceProperty;
+                navigationLink.Url = this.requestInfo.EntityTracker.GetEntityDescriptor(relatedEnd.Target).GetLatestEditLink();
+                bool isCollection = clientType.GetProperty(relatedEnd.SourceProperty, UndeclaredPropertyBehavior.ThrowException).IsEntityCollection;
+
+                odataWriter.WriteStart(navigationLink, relatedEnd.Source, relatedEnd.Target);
+
+
+                if (isCollection)
+                {
+                    WriteResourceSet(relatedLinks.Value, bulkUpdateGraph, odataWriter);
```
ElizabethOkerio, line 434: why don't we pass the related links to write here? since we already computed them in line 404?

ElizabethOkerio: aah I see. we passing them

src/Microsoft.OData.Client/DataServiceContext.cs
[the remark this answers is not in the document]
ElizabethOkerio: Can this method be used with the normal `SaveChanges` and what will be the implications?

src/Microsoft.OData.Client/DataServiceContext.cs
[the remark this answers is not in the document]
ElizabethOkerio: I think this is the one I was  asking. Cam this be used with   normal `SaveChanges` API call? Are there any implications as we don't post to single-valued navigation properties on a normal `POST`?

src/Microsoft.OData.Client/DataServiceContext.cs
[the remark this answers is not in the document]
ElizabethOkerio: I do not understand why we need to do this a single-value navigation property only and not a collection-valued navigation property?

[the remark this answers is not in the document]
KenitoInc: 1) SetLink creates a `PUT` request that's why the state is `Modified`. I didn't want to change the `SetLink` logic.
2) `AddRelatedObject` works for collection navigation property. So this new method works for single-value navigation property

ElizabethOkerio (approved):

leoerlandsson:
Hi,

Is there any info when the Deep Insert functionality will make it into a Release?

Thanks!

Br,
Leo

PR #2664, opened 2023-05-11
fixes #2659: Support declared Edm.Untyped property serialization using ODataPrimitiveValue, etc

### LINKED ISSUE #2659: Support declared Edm.Untyped property serialization using ODataPrimitiveValue

### Assemblies affected

### Reproduce steps

*
Suppose we have the following OData property:

```xml

<Property Name="Data" Type="Edm.Untyped" />
```

In mapping, we have the following C# property:

```C#
public object Data {get;set;}
```

So, if we have a resource instance, the value of `Data` could be anything. 

```C#
Person p = new Person();
p.Data = 42; // or 
p.Data = null; // or
p.Data = true;
```

So, How can we serialize them in the ODL?  

The existing codes need to create 'ODataUntypedValue' to serialize Declared Edm.Untyped property as:

```C#
new ODataProperty {
  Name = "Data",
  Value = new ODataUntypedValue ...
}
```

But, It's better to support to create 'ODataPrimitiveValue ' to serialize Declared Edm.Untyped property as:
```C#
new ODataProperty {
  Name = "Data",
  Value = new ODataPrimitiveValue { Value = 42... }
}
```

xuzhg:
…g ODataPrimitiveValue, etc

### Issues

*This pull request fixes #2659.*

### Description

* Enable use other OData value to serialize declared untyped property

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/TypeNameOracle.cs
+++ b/src/Microsoft.OData.Core/TypeNameOracle.cs
@@ -318,7 +318,10 @@ namespace Microsoft.OData
                 return typeReferenceFromMetadata;
             }
 
-            Debug.Assert(typeReferenceFromValue.TypeKind() == typeReferenceFromMetadata.TypeKind(), "typeReferenceFromValue.TypeKind() == typeReferenceFromMetadata.TypeKind()");
+            if (!typeReferenceFromMetadata.IsUntyped())
+            {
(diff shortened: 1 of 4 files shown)


mikepizzo (approved):
:shipit:

Issue #2656, opened 2023-04-24
OData Batch Requests fails to handle duplicate http attributes


apiaskowski:
Sending OData batch requests with duplicate http headers will cause an System.ArgumentException that indicates that duplicate http headers are not validated correct.

### Assemblies affected

Microsoft.OData.Core 7.15.0

### Reproduce steps

Create a basic MVC OData service and enable batch requests.
Submit any batch request, which includes the same http header twice

### Expected result

If a duplicate http header has been send, the duplicate header(s) should be ignored or overwrite the previous value.

### Actual result

The OData batch requests returns an HTTP 500 error, with the  System.ArgumentException: An item with the same key has already been added. Key: <http header>.
This is caused by the Microsoft.OData.JsonLight.BatchPayloadItemPropertiesCache which doesn't validate, if an http header already has been added to the headers Dictionary.

apiaskowski:
@habbes my PR bugfix is in preparation.

apiaskowski:
Dear @habbes,

I've got to take back my PR, because of legal reasons with the CLA that have to be sorted out by my employer.
Do you mind, if I suggest you a potential fix, which I've implemented & tested locally?

The issue occurs, because of missing checks in the `Microsoft.OData.JsonLight.ODataJsonLightBatchPayloadItemPropertiesCache` class, in specific within the method `ScanJsonProperties` and `ScanJsonPropertiesAsync`.
All instances of `jsonProperties.Add()` and `headers.Add()` should check, if an element already exists in the respective Dictionary.
Depending on the desired handling, either skip duplicate elements or overwrite the element's value (don't forget to read the next Property by the `asynchronousJsonReader`, if an element is skipped).

Examples:

- [`jsonProperties.Add(propertyName, this.jsonReader.ReadStringValue());`](https://github.com/OData/odata.net/blob/399df9282bf85a60e64df2924abe6c4b88cec3c9/src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchPayloadItemPropertiesCache.cs#L230)
- [`headers.Add(headerName, headerValue);`](https://github.com/OData/odata.net/blob/399df9282bf85a60e64df2924abe6c4b88cec3c9/src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchPayloadItemPropertiesCache.cs#L278)

[...]

PR #2669, opened 2023-05-17
Fixed the http attribute handling of Batch Requests

apiaskowski:
Fixed the http attribute handling in `ODataJsonLightBatchPayloadItemPropertiesCache`, by skipping duplicate properties.

### Issues

[*This pull request fixes Issue-2656*](https://github.com/OData/odata.net/issues/2656)

### Description

OData batching does return an HTTP 500 error, if the request includes duplicate HTTP attributes.
This issue is caused by not proper handling/validation of the HTTP attributes, before adding them into a dictionary.

- [x] *Test cases added*

### Additional work necessary

-/-


apiaskowski:
@microsoft-github-policy-service agree company="d.velop AG"

apiaskowski:
Hi, is there a reason or problem with the review request?

gathogojr:
@apiaskowski What if instead we opted for the following approach (as an example)

Replace:
```csharp
jsonProperties.Add(this.jsonReader.ReadStringValue());
```
with
```csharp
jsonProperties[propertyName] = this.jsonReader.ReadStringValue();
```
instead of:
```csharp
string nameOfProperty = this.jsonReader.ReadStringValue();
if (!jsonProperties.ContainsKey(propertyStringValue))
{
    jsonProperties.Add(propertyStringValue);
}
```

Of course this approach means the last one wins but that's no less correct than the first one winning - it's bad enough that the payload has duplicate properties where one was expected and that we're having to pick one over the other... The benefit with the proposed approach is that it's tidier (less lines)

habbes (changes requested):
I think we can have a discussion about whether we should support duplicate fields `headers` object. But I don't think we should support duplicate fields anywhere else in the request, e.g. duplicate `method` fields.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
```diff
@@ -363,6 +363,109 @@ public async Task ReadBatchRequestWithChangesetAndDependsOnIdsAsync()
                 isResponse: false);
         }
 
+        [Fact]
+        public async Task ReadBatchRequestWithDuplicateAttributesAsync()
+        {
+            var payload = "{\"requests\": [{" +
+                "\"id\": \"1\"," +
+                "\"method\": \"GET\"," +
+                "\"method\": \"POST\"," +
```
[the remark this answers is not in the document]
gathogojr: I think the right course of action is to throw an exception whenever a duplicate header is found.

apiaskowski:
> @apiaskowski What if instead we opted for the following approach (as an example)
> 
> Replace:
> 
> ```cs
> jsonProperties.Add(this.jsonReader.ReadStringValue());
> ```
> 
> with
> 
> ```cs
> jsonProperties[propertyName] = this.jsonReader.ReadStringValue();
> ```
> 
> instead of:
> 
> ```cs
> string propertyStringValue = this.jsonReader.ReadStringValue();
> if (!jsonProperties.ContainsKey(propertyStringValue))
> {
>     jsonProperties.Add(propertyStringValue);
> }
> ```
> 
> Of course this approach means the last one wins but that's no less correct than the first one winning - it's bad enough that the payload has duplicate properties where one was expected and that we're having to pick one over the other... The benefit with the proposed approach is that it's tidier (less lines)

@gathogojr we are open to this approach too.
My solutions follows the same approach as implemented for duplicate cookie headers, shown in [PR-2379](https://github.com/OData/WebApi/pull/2379).
But as described in [Issue-2656](https://github.com/OData/odata.net/issues/2656), both approaches are fine.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
```diff
@@ -363,6 +363,109 @@ public async Task ReadBatchRequestWithChangesetAndDependsOnIdsAsync()
...
+        }
+
+        [Fact]
+        public async Task ReadBatchRequestWithDuplicateHeadersAsync()
+        {
+            var payload = "{\"requests\": [{" +
+                "\"id\": \"1\"," +
+                "\"method\": \"POST\"," +
+                "\"url\": \"http://tempuri.org/Customers\"," +
+                "\"headers\": {\"odata-version\":\"4.0\",\"content-type\":\"application/json;odata.metadata=minimal;odata.streaming=true;IEEE754Compatible=false;charset=utf-8\",\"duplicate-header\":\"value1\",\"duplicate-header\":\"value2\"}, " +
```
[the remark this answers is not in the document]
apiaskowski: @habbes I am sorry, but we have no impact on the incoming requests and as shown in the test data, the requests show duplicate headers, which currently result in HTTP 500 responses.
If decided that duplicate headers are not allowed, we'd like to see this restriction being documented in the OData Specifications [Batch Request Headers](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#sec_BatchRequestHeaders) and handled by a specific HTTP 4XX response.
But as a first approach, I tried to implement the solution, as shown in [Issue-2656](https://github.com/OData/odata.net/issues/2656) for handling duplicate cookie headers.

gathogojr:
Hi @mikepizzo. I'd appreciate your opinion on duplicate headers in batch requests- what should we do if we come across duplicate `method` properties (`method`:`GET` and `method`:`POST`)? What does the spec say or what should it say? To me, picking one or the other seems like a wild gamble.

RE: https://github.com/OData/odata.net/pull/2669#discussion_r1226027252

apiaskowski:
> > 
> 
> Having duplicate "method" properties in the json payload should be a 400 error. I can open an issue to add that to the standard.
> 
> [RFC2616 ](https://www.rfc-editor.org/rfc/rfc2616#section-4.2) allows the same header to appear multiple times in a request only if the header value is a coma-separated list of values, and states that the multiple headers is equivalent to a single header with the values appended, in order, separated by comas. So a valid implementation would be, if we see a header that already exists, append a comma + value to the existing header value, and pass that to the service who can then decide if it's valid or throw an exception (presumably the same as if the user made the request outside of a batch).

@mikepizzo how would you like us to proceed. Should we only return an HTTP 400 error (probably "malformed request syntax"), if the header "method" is duplicate and comma separate all other duplicate http attributes?

mikepizzo:
The only parameter for which RFC2616 defines duplicate values is header. id, method, url, and body should all return 400 if they appear more than once. While we *could* support multiple atomicityGroup, dependsOn, and if values, I would just return a 400 for these as well.

gathogojr:
@apiaskowski I triggered the build and some tests are failing

apiaskowski:
> @apiaskowski I triggered the build and some tests are failing

Dear @gathogojr, thank you for the feedback. You were correct, it was a flaw on my side.
I didn't saw that this test failed, because of existing ~140 test failures in the local project.
Please let the build run again (issue fixed).

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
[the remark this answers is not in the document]
mikepizzo: The current code implements a "last entry wins" semantic. This is different than what I was suggesting which was, *if* we wanted to support duplicate headers, that we align with the semantics for multiple occurrences of the same headers defined in RFC2616 for an HTTP request which specifies that the same header can appear only if the value is a comma separated list, in which case the values are appended.

Which is not exactly the same as having multiple JSON entries with the same value (which is not supported for interoperable JSON clients), but was the closest semantics I could come up with for handling duplicate headers.

I brought this up in our OASIS OData Technical Meeting today (see [https://issues.oasis-open.org/browse/ODATA-1581?jql=project%20%3D%20ODATA](https://issues.oasis-open.org/browse/ODATA-1581?jql=project%20%3D%20ODATA)) and we agreed to clarify in the spec that clients should not have duplicate properties in the batch request, nor duplicate headers in the header collection, and that services *should* return a client error for these cases. 

This is a SHOULD, not a MUST, since some services may already have other behavior, but now that the direction has been defined in the standard we should align and return a client error for duplicate headers (as we do for duplicate request properties).

mikepizzo (changes requested):
:clock1:

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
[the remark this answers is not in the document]
apiaskowski: Understood, I'll prepare the necessary changes.
Are explicit test cases necessary, that proof duplicate attributes & headers throw the client error?

apiaskowski:
Dear @mikepizzo, I've adjusted the handling of headers, based on the diskussion. All duplicate json properties, including the headers should now throw an dedicated exception.

src/Microsoft.OData.Core/Microsoft.OData.Core.cs
```diff
@@ -193,7 +193,9 @@ internal sealed class TextRes {
         internal const string ODataBatchReaderStream_MultiByteEncodingsNotSupported = "ODataBatchReaderStream_MultiByteEncodingsNotSupported";
         internal const string ODataBatchReaderStream_UnexpectedEndOfInput = "ODataBatchReaderStream_UnexpectedEndOfInput";
         internal const string ODataBatchReaderStreamBuffer_BoundaryLineSecurityLimitReached = "ODataBatchReaderStreamBuffer_BoundaryLineSecurityLimitReached";
+        internal const string ODataJsonLightBatchPayloadItemPropertiesCache_DuplicatePropertyForMessageInBatch = "ODataJsonLightBatchPayloadItemPropertiesCache_DuplicatePropertyForMessageInBatch";
```
mikepizzo, line 196: >ODataJsonLightBatchPayloadItemPropertiesCache_DuplicatePropertyForMessageInBatch [](http://example.com/codeflow?start=114&length=80)

These error strings need to be added to Microsoft.OData.Core.txt.

src/Microsoft.OData.Core/Parameterized.Microsoft.OData.Core.cs
```diff
@@ -1647,6 +1647,14 @@ internal static string ODataBatchReaderStreamBuffer_BoundaryLineSecurityLimitRea
             return Microsoft.OData.TextRes.GetString(Microsoft.OData.TextRes.ODataBatchReaderStreamBuffer_BoundaryLineSecurityLimitReached, p0);
         }
 
+        /// <summary>
+        /// A string like "Duplicate property name '{0}' for message in batch."
```
mikepizzo, line 1651: >message [](http://example.com/codeflow?start=61&length=7)

suggest: "request" rather than "message"

mikepizzo:
> Dear @mikepizzo, I've adjusted the handling of headers, based on the diskussion. All duplicate json properties, including the headers should now throw an dedicated exception.

Thanks for being so responsive on this @apiaskowski -- changes look good. I can approve once you add the new error strings to Microsoft.Core.OData.txt.

Thanks again!

src/Microsoft.OData.Core/Parameterized.Microsoft.OData.Core.cs
apiaskowski: Changed naming of the exceptions from "message" to "request"

src/Microsoft.OData.Core/Microsoft.OData.Core.cs
apiaskowski: Added both exceptions into the Microsoft.OData.Core.txt

apiaskowski:
> > Dear @mikepizzo, I've adjusted the handling of headers, based on the diskussion. All duplicate json properties, including the headers should now throw an dedicated exception.
> 
> Thanks for being so responsive on this @apiaskowski -- changes look good. I can approve once you add the new error strings to Microsoft.Core.OData.txt.
> 
> Thanks again!

@mikepizzo, I've followed the suggestions and fixed both issues.
Would it be possible to resolve them and approve the PR?

mikepizzo (approved):
:shipit:

mikepizzo:
@habbes -- can you make sure your concerns have been addressed and, if so, approve and merge?  Thanks!

habbes (commented):
@apiaskowski Thanks for the prompt changes. Based on the agreed specification, I think the changes look good. However, I think you should add tests to verify that the exceptions you've added are thrown as expected. Specifically, a test that verifies that the appropriate exception (and error message) as thrown when we have duplicate properties in the request, and a test that verifies that the appropriate exception is thrown when we have duplicate headers.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
```diff
@@ -363,6 +363,40 @@ public async Task ReadBatchRequestWithChangesetAndDependsOnIdsAsync()
...
+                payload,
+                async (jsonLightBatchReader) =>
+                {
+                    try
+                    {
+                        while (await jsonLightBatchReader.ReadAsync())
+                        {
+                            if (jsonLightBatchReader.State == ODataBatchReaderState.Operation)
+                            {
+                                await jsonLightBatchReader.CreateOperationRequestMessageAsync();
```
habbes, line 386: What is the expected value of the parsed `null-header`? I think you should have assertions that verify your intentions, i.e. if the headers you expect in the request message actually exist in the request message with the expected values.

apiaskowski: @habbes, I've specified the purpose of this test case and improved readability.
That test case should verify that an json property with NULL value won't cause an HTTP 500 internal server error.
I hope the improved procedure is easier to understand.
Secondly I've added additional test cases to verify, that duplicate json properties return in a correct odata error, and not HTTP 500 internal server error.

Would you please check them again?

habbes (approved):

ElizabethOkerio (dismissed):

KenitoInc (dismissed):

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
```diff
@@ -363,6 +363,114 @@ public async Task ReadBatchRequestWithChangesetAndDependsOnIdsAsync()
...
+        public async Task ReadBatchRequestWithNullHeadersAsync()
+        {
+            var payload = "{\"requests\": [{" +
+                "\"id\": \"1\"," +
+                "\"method\": \"POST\"," +
+                "\"url\": \"http://tempuri.org/Customers\"," +
+                "\"headers\": {\"odata-version\":\"4.0\",\"content-type\":\"application/json;odata.metadata=minimal;odata.streaming=true;IEEE754Compatible=false;charset=utf-8\",\"null-header\":null}, " +
+                "\"body\": {\"@odata.type\":\"#NS.Customer\",\"Id\":1,\"Name\":\"Customer 1\",\"Type\":\"Retail\"}}]}";
+
+            await SetupJsonLightBatchReaderAndRunTestAsync(
```
gathogojr, line 376: @apiaskowski Not a deal breaker but if we're looking at this from a best practices point of view, the better way to assert on exceptions is this:

```csharp
var exception = await Assert.ThrowsAsync<ODataException>( // Or ArgumentException as the case maybe
    () => SetupJsonLightBatchReaderAndRunTestAsync(
        // ...
    );

Assert.Equal(/** expected error message **/, exception.Message);
```

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
```diff
@@ -363,6 +363,114 @@ public async Task ReadBatchRequestWithChangesetAndDependsOnIdsAsync()
...
+                    }
+                    catch (ODataException ex)
+                    {
+                        // Verify that the correct duplicate header has raised the ODataException
+                        Assert.Equal(Strings.ODataJsonLightBatchPayloadItemPropertiesCache_DuplicateHeaderForRequestInBatch("duplicate-header"), ex.Message);
+                    }
+                },
+                isResponse: false);
+        }
+
```
gathogojr, line 473: Somewhere within this test project we have tests for the sync API. We should add tests for the sync scenarios as well.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchPayloadItemPropertiesCache.cs
```diff
@@ -267,7 +273,13 @@ private void ScanJsonProperties()
                                 while (this.jsonReader.NodeType != JsonNodeType.EndObject)
                                 {
                                     string headerName = this.jsonReader.ReadPropertyName();
-                                    string headerValue = this.jsonReader.ReadPrimitiveValue().ToString();
+                                    string headerValue = this.jsonReader.ReadPrimitiveValue()?.ToString();
```
gathogojr, line 276: I think we handle the `null` value scenario elsewhere. What scenarios is the null coalescing operator handling? Did you add tests for such scenarios?

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchPayloadItemPropertiesCache.cs
```diff
@@ -407,7 +425,13 @@ await this.asynchronousJsonReader.ReadStartObjectAsync()
                             {
                                 string headerName = await this.asynchronousJsonReader.ReadPropertyNameAsync()
                                     .ConfigureAwait(false);
-                                string headerValue = (await this.asynchronousJsonReader.ReadPrimitiveValueAsync().ConfigureAwait(false)).ToString();
+                                string headerValue = (await this.asynchronousJsonReader.ReadPrimitiveValueAsync().ConfigureAwait(false))?.ToString();
```
gathogojr, line 428: See the comment I left here https://github.com/OData/odata.net/pull/2669/files#r1254207121

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchPayloadItemPropertiesCache.cs
apiaskowski: Not handling the `NULL` values, when executing the json property read function (including the header properties), will result in an `ArgumentNullException` and return an HTTP 500 "internal server error" by the application.
I've identified and described this behavior in the underlying issue ticket.
It's easy to reproduce, so it would be nice, if you have a look at it by yourself (just locally revert the changes and execute the corresponding `ReadBatchRequestWithNullHeadersAsync` UnitTest to verify).
https://github.com/OData/odata.net/issues/2656

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
apiaskowski: Sure, just this use-case confirms, that no Exception is thrown and not that an specific Exception is thrown and catched.
I am not sure if this is possible with the `Assert.ThrowsAsync` method, but I'll confirm it.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
apiaskowski: Unfortunately, I've not found any matching sync `ODataBatchReader` UnitTests, that matches to these use-cases.
There are test cases, where the sync methods are called, e.g. `ODataMultipartMixedBatchReaderTests`, but these focus on different topics, so it doesn't make sense to add null and duplicate header verifications to them.
If for these test scanarios dedicated sync method UnitTests are required / desired, it would be a topic for an upcomming feature ticket.

gathogojr: We strive to ensure that the sync and async APIs do not diverge so this ask is important... We have to ensure that the same exceptions that are thrown in the async workflow will be thrown in the sync workflow.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
gathogojr: You'll find tests within this test file that do it the way I described and verify that the specific exception that is thrown and even assert on the exception message...

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
apiaskowski: I've added 3x synchronous unit tests to verify that `null` headers and duplicate properties & headers.
If you wish further improvements to unit test coverage, please feel free to raise an independent feature ticket, since I won't cover other non existing tests, that don't belong to this bugfix.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
apiaskowski: I've added the Assert.ThrowsAsync and Assert.Throws to the test cases of duplicate headers and properties.
Adding it to the null header test case does not work, since the Throws methods will cause the unit test to fail, if no exception is thrown, which the exact opposite to the test case purpose.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
```diff
@@ -363,6 +363,206 @@ public async Task ReadBatchRequestWithChangesetAndDependsOnIdsAsync()
...
+                            {
+                                var operationRequestMessage = jsonLightBatchReader.CreateOperationRequestMessage();
+                                // Verify that the Property "null-header" exists and it's value is set to NULL
+                                var nullHeaderProperty = operationRequestMessage.Headers.FirstOrDefault(p => p.Key == "null-header");
+                                Assert.NotNull(nullHeaderProperty.Key);
+                                Assert.Null(nullHeaderProperty.Value);
+                            }
+                        }
+                    }
+                    catch (NullReferenceException ex)
```
gathogojr, line 394: A scenario where the library ends up throwing a `NullReferenceException` is not acceptable. That's bad practice. If the header is required, we should check if it's null and throw an `ODataException` explaining that a required header is missing.

Meaning that instead of
```csharp
string headerValue = this.jsonReader.ReadPrimitiveValue()?.ToString();
```
we should have something like:
```csharp
string headerValue = this.jsonReader.ReadPrimitiveValue()?.ToString();
if (headerValue == null)
{
    throw new ODataException("###ERROR_MESSAGE###");
}
```
or
```csharp
object headerObjValue = this.jsonReader.ReadPrimitiveValue();
if (headerObjValue == null)
{
    throw new ODataException("###ERROR_MESSAGE###");
}

string headerValue = headerObjValue.ToString()
```

gathogojr (changes requested):
See my comment here https://github.com/OData/odata.net/pull/2669/files#r1264644863

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchReaderTests.cs
apiaskowski: @gathogojr do you understand the purpose of this unit test?

I made this unit test to proof that no NullReferenceException is throw, by fixing the behavior of the JSON property reader.
If you are telling me, that you don't want me to handle JSON properties with `null` values, I suggest you to change the OData Protocol specifications and then handle it in the way you want it to.

I've done a similar on request of Mike for handling duplicate headers, because he insisted to change the OData specifications to not allow any duplicate headers.
But I won't do this a 2nd time here. It is kind of hard to believe, that this OData project is in productive usage by customers, without having proper unit test coverage, without having basic null or duplicate value validation, and every person doing review here, having a different idea on how the application has to operate and asking for the specification to change.

I'll do as following, revert my changes, removing the unit tests for NullReferenceExceptions check and my adjustments of valid handling of JSON properties, that include `null` values and leave it up to you, to decide and handle, how these should be handled. You can do as you like, either throw an OData Exception or handle these values correctly, it is now up to you to decide and do.

gathogojr: @apiaskowski Reverting the changes was unnecessary. Explaining that the particular test verifies that a `NullReferenceException` is NOT thrown was sufficient. That was a misunderstanding on my part and I apologize for that.

apiaskowski: Sure, no issue, I can add the null value validation(s) as part of a separate PR (if and when decided on how to handle these).
I'd like to get these changes here reviewed, done and delivered.
Adding/improving code coverage, changing specifications, etc. we can always do in separate steps and don't need to block bug fixes.

gathogojr (approved):

gathogojr:
Thanks @apiaskowski for your contribution

apiaskowski:
@gathogojr, thank you for the merge.
I'll prepare the followup PR to handle NullReferenceExceptions from empty JSON properties.
Could you please check that one out too and define, if `null` JSON properties should throw an ODataException or should be handled as empty JSON properties?

PR #2675, opened 2023-05-31
Corranrogue9/assert3

corranrogue9:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary


test/CommonAssemblies/Transform/CollectFiles/Microsoft.Web.Publishing.AllFilesInProjectFolder.targets
```diff
@@ -0,0 +1,181 @@
+﻿<!--
```
corranrogue9, line 1: The new files in CommonAssemblies are pulled from the visual studio installation of Microsoft.Web.Publishing.Tasks. Probably we should have a nuget package for this, but I'm adding the files directly to the repo instead for now since we don't have nuget packages for other similar dependencies at the moment.

PR #2681, opened 2023-06-08
updated odata.context computation to remove trailing cast segments th…

### LINKED ISSUE #2680: Error generating @odata.context for requests that end in a cast segment followed up a key segment
The following exception is thrown when writing the `@odata.context` in a response payload when the request URL ends in a cast segment followed by a key segment:
```
Microsoft.OData.ODataException: The Path property 'identityGovernance/permissionsAnalytics/azure/findings/microsoft.graph.inactiveUserFinding' of ODataMessageWriterSetting.ODataUri must end with the navigation property which the contained elements being written belong to.
```

### Assemblies affected

At least 7.16.0 is affected

### Reproduce steps

Request made in the form /.../{containmentcollection}/{castsegment}({id})

### Expected result

An `@odata.context` is returned in the response payload of `/.../{containmentcollection}`

### Actual result

The above exception

corranrogue9:
…at are following by key segments as well as the case where the standard key syntax is used within a cast segment

### Issues

*This pull request fixes #2680.*

### Description

When the request URL ends in a cast segment followed by a key segment, we throw an exception because the `@odata.context` should not end in such segments, but we do not trim them out of the request URL when generating the `@odata.context`. 

To fix this, the `ODataContextUrlInfo` has been updated to call a new trim extension method for `ODataPath` which trims all cast and key segments from the ends of the path, regardless of the order of the segments. Since there were two methods which both had this logic that needed to be updated, those two methods were refactored to have a single method that contains the relevant logic that both methods call into (reducing duplicate code). 

- [ ] *Test cases added*

### Additional work necessary

We likely shouldn't throw an exception in this case, but instead return a "bad" `@odata.context`. We should keep the check as a `Debug.Assert` probably, but not throw in release configurations.

The new trim method does not leverage the `WalkWith` methods on `ODataPath`. `WalkWith` always starts at the beginning, and always requires enumerating the entire path, even when not necessary. We should update and consolidate other trim variants on `ODataPath` to remove the use of these handlers wherever feasible. 

We could also optimize the use of the `Take` LINQ query by having `ODataPath` take in an `IReadOnlyList`, and then having a "slice" extension on it to allow the `ODataPath` to not need to always enumerate all of the segments passed into the constructor.


src/Microsoft.OData.Core/ODataContextUrlInfo.cs
```diff
@@ -294,19 +294,14 @@ private static string ComputeNavigationPath(EdmNavigationSourceKind kind, ODataU
                 return null;
             }
 
-            string navigationPath = null;
-            if (kind == EdmNavigationSourceKind.ContainedEntitySet && odataUri != null && odataUri.Path != null)
+            if (odataUri == null)
```
corranrogue9, line 297: From meeting with @mikepizzo:

```
public void Compute(ODataUri uri)
{
	Internal(uri);
}

public void Compute(Slim slim)
{
	Internal(new ODataUri(slim.ODataUri));
}

private void Internal(ODataUri)
{
}
```

corranrogue9: Create an 8.0 work item for rationalizing the two types

corranrogue9: I created [this](https://github.com/OData/odata.net/issues/2696) work item to track the 8.0 work.

@habbes do you have any thoughts on if `ComputeNavigationPathImpl` should take in an `ODataUri` vs `ODataUriSlim`?

corranrogue9: There are actually 3 references stored in the `ODataUriSlim`, so this is *not* as performant.

corranrogue9:
@KenitoInc @habbes @xuzhg @ElizabethOkerio @gathogojr  @chrisspre @mikepizzo I'd really love to get a review on this

corranrogue9:
@KenitoInc @habbes @xuzhg @ElizabethOkerio @gathogojr @chrisspre @mikepizzo ping...this is going to start blocking an AGS workload soon

src/Microsoft.Spatial/System/Collections/Generic/ReadOnlyCollectionExtensions.cs
```diff
@@ -0,0 +1,42 @@
+﻿namespace System.Collections.Generic
+{
```
xuzhg, line 2: any reason why do we put this extension into the Microsoft.Spatial lib?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightWriterTests.cs
```diff
@@ -738,6 +745,114 @@ public async Task WriteEntityReferenceLinkAsync()
...
+                "}";
+
+            // load the CSDL from the embedded resources
+            var assembly = Assembly.GetExecutingAssembly();
+            var currentMethod = GetCurrentMethodName();
+            var csdlResourceName = assembly.GetManifestResourceNames().Where(name => name.EndsWith($"{currentMethod}.xml")).Single();
+
+            // parse the CSDL
+            IEdmModel model;
+            using (var csdlResourceStream = assembly.GetManifestResourceStream(csdlResourceName))
```
xuzhg, line 803: Loading from manifest Resource looks good. 
But, If it's only used in one test case, is it too heavy? :)

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPathExtensions.cs
```diff
@@ -137,6 +137,22 @@ public static ODataPath TrimEndingTypeSegment(this ODataPath path)
...
 
+        /// <summary>
+        /// Creates a <see cref="ODataPath"/> that is <paramref name="path"/> with the type segments and key segments removed from the end
+        /// </summary>
+        /// <param name="path">The <see cref="ODataPath"/> to trim the ending of</param>
+        /// <returns>A <see cref="ODataPath"/> without type-cast and key segments at the end</returns>
+        /// <exception cref="System.ArgumentNullException">Thrown if <paramref name="path"/> is <see langword="null"/></exception>
+        public static ODataPath TrimEndingTypeAndKeySegments(this ODataPath path)
+        {
+            if (path == null)
```
xuzhg, line 148: shall we call the 'TrimEndingTypeSegment().TrimEndingKeySegment()' in this new method to avoid duplication?

xuzhg (approved):

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPathExtensions.cs
```diff
@@ -137,6 +137,22 @@ public static ODataPath TrimEndingTypeSegment(this ODataPath path)
             return handler.FirstPart;
         }
 
+        /// <summary>
+        /// Creates a <see cref="ODataPath"/> that is <paramref name="path"/> with the type segments and key segments removed from the end
+        /// </summary>
+        /// <param name="path">The <see cref="ODataPath"/> to trim the ending of</param>
+        /// <returns>A <see cref="ODataPath"/> without type-cast and key segments at the end</returns>
+        /// <exception cref="System.ArgumentNullException">Thrown if <paramref name="path"/> is <see langword="null"/></exception>
+        public static ODataPath TrimEndingTypeAndKeySegments(this ODataPath path)
```
KenitoInc, line 146: Do we have unit tests for this?

KenitoInc (commented):
We should have unit tests for the new public extension method.
Otherwise the PR looks good to merge

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPathExtensions.cs
corranrogue9: I think we had good coverage with the JSON writer, but you're totally right, it's better to have direct coverage of the extension, thanks!

src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPathExtensions.cs
corranrogue9: This is actually the bug that the PR is trying to fix. We don't want to trim the type segments and then trim the key segments. We want to trim any type and key segments. With the unit tests in the PR, you can try replacing this method with your comment and see how those tests fail.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightWriterTests.cs
corranrogue9: I think it gives a straightfoward way to have a different model for different tests. I think it would be good to start doing this in other tests as well going forward. I'd like to keep it this way unless you have a strong objection.

src/Microsoft.OData.Core/ODataContextUrlInfo.cs
corranrogue9: I've added additional context to the work item and reverted this unification. Also added an additional test for the "slim" variant now that they are separate code paths again.

src/Microsoft.Spatial/System/Collections/Generic/ReadOnlyCollectionExtensions.cs
corranrogue9: I have moved the extensions to `Microsoft.OData.Edm` to have the same effect but keep `Microsoft.Spatial` unchaged.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightWriterTests.cs
corranrogue9: Because I've started using this for new issues as well to good effect, I'm going to keep it this way. We can refactor in the future if we need to.

corranrogue9:
> We should have unit tests for the new public extension method. Otherwise the PR looks good to merge

Thanks @KenitoInc. I've added the unit test, but there are a number of other changes too if you can take another look.

corranrogue9:
@KenitoInc @habbes @xuzhg @ElizabethOkerio @gathogojr @chrisspre @mikepizzo ping...

corranrogue9:
@KenitoInc @habbes @xuzhg @ElizabethOkerio @gathogojr @chrisspre @mikepizzo can you please take a look when you get a chance?

xuzhg (approved):

PR #2683, opened 2023-06-09
Implement `IAsyncDisposable` for `ODataMessageWriter`

### LINKED ISSUE #2682: `ODataMessageWriter` cannot be disposed asynchronously
Because `ODataMessageWriter` does not implement `IAsyncDisposable`, the synchronous `Dispose` is called when paired with the `using` statement. This can cause unwanted/unexpected synchronous I/O and/or errors in environments where synchronous I/O is not allowed (e.g. when `AllowSynchronousIO` is set to `false` in ASP.NET Core).

### Assemblies affected

`Microsoft.OData.Core` 7.x

### Reproduce steps

```c#
using (var messageWriter = new ODataMessageWriter(...))
{

}
```

### Expected result

We should be able to do this:

```c#
await using var messageWriter = new ODataMessageWriter(...);
```

### Actual result

`ODataMessageWriter` does not implement `IAsyncDisposable` and therefore we can't use `await using`.

habbes:
### Issues

*This pull request fixes #2682 *

### Description

This PR makes `ODataMessageWriter` `IAsyncDisposable` and attempts to remove all instances of synchronous I/O in the code paths of `ODataMessageWriter.DisposeAsync`.

Because `IAsyncDisposable` and `ValueTask` are not available in older frameworks, I've restricted the feature to .NET Core 3.1 and later.

To test that we aren't performing synchronous I/O when disposing asynchronously or other async code paths, I've created test `AsyncStream` that throws an exception when synchronous `Stream` I/O methods are called. This helped detected a number of cases where we performed synchronous writes in async code.

Note, there are cases where we have `stream.Dispose()` inside constructors in a `try-finally` block. I've seen this in `ODataJsonLightOutputContext` and other `OutputContext` constructors. Having the `Dispose()` calls in the constructor is problematic because it opens potential for synchronous I/O in async code paths. We can't make the constructor asynchronous, so in this case a more involved refactoring process is required. For example, we may consider moving the `try-finally` blocks outside the constructors and up to the callers. Since this has not be linked to many issues in production, I have opted to create a separate issue for this: https://github.com/OData/odata.net/issues/2685.

- [x] *Test cases added*


test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightOutputContextTests.cs
```diff
@@ -438,14 +438,23 @@ public async Task WriteBatchAsync()
                     var operationRequestMessage = await batchWriter.CreateOperationRequestMessageAsync(
                         "POST", new Uri($"{ServiceUri}/Orders"), "1");
 
-                    using (var messageWriter = new ODataMessageWriter(operationRequestMessage))
+                    var messageWriter = new ODataMessageWriter(operationRequestMessage);
```
gathogojr, line 441: Could you have refactored this in the following manner:
```csharp
#if NETCOREAPP3_1_OR_GREATER
                    await using (var messageWriter = new ODataMessageWriter(operationRequestMessage))
#else
                    using (var messageWriter = new ODataMessageWriter(operationRequestMessage))
#endif
                    {
                        var resourceWriter = await messageWriter.CreateODataResourceWriterAsync(this.orderEntitySet, this.orderEntityType);
                        var orderResource = CreateOrderResource();
                        await resourceWriter.WriteStartAsync(orderResource);
                        await resourceWriter.WriteEndAsync();
                    }
```

I'd be curious whether if you did it this way `await messageWriter.DisposeAsync();` would automatically get triggered for `NETCOREAPP3_1_OR_GREATER`; otherwise `messageWriter.Dispose()`

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightOutputContextTests.cs
```diff
@@ -749,7 +758,13 @@ public async Task WriteInStreamError_ThrowsExceptionIfInvokedWhenWritingBatchAsy
             Assert.Equal(Strings.ODataBatchWriter_CannotWriteInStreamErrorForBatch, exception.Message);
         }
 
-        #endregion Async Tests
+        [Fact]
+        public async Task DisposeAsync_ShouldDisposeStream_Writer_Asynchronously()
```
gathogojr, line 762: What purpose does this empty test serve?

test/FunctionalTests/Microsoft.OData.Core.Tests/AsyncOnlyStreamWrapper.cs
```diff
@@ -0,0 +1,185 @@
...
+
+namespace Microsoft.OData.Tests
+{
+    /// <summary>
+    /// A test stream wrapper that that throws an
+    /// exception when synchronous I/O methods are called.
+    /// This is meant for testing asynchronous code paths
+    /// where synchronous I/O is not allowed.
+    /// </summary>
+    internal class AsyncOnlyStreamWrapper : Stream
```
gathogojr, line 20:
```suggestion
    internal class AsyncStream : Stream
```
`AsyncOnlyStreamWrapper` is quite a mouthful. That it's a wrapper is covered in the doc string.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightOutputContextTests.cs
habbes: I actually used this pattern in later code, I just hadn't thought of it until I was half-way through the refactoring.

src/Microsoft.OData.Core/Json/JsonWriter.cs
```diff
@@ -17,14 +17,19 @@ namespace Microsoft.OData.Json
...
 
     /// <summary>
     /// Writer for the JSON format. http://www.json.org
     /// </summary>
     [SuppressMessage("Microsoft.Design", "CA1001:TypesThatOwnDisposableFieldsShouldBeDisposable", Justification = "This class does not own the underlying stream/writer and thus should never dispose it.")]
-    internal sealed partial class JsonWriter : IJsonStreamWriter, IJsonStreamWriterAsync, IDisposable
+#if NETCOREAPP3_1_OR_GREATER
+    internal sealed partial class JsonWriter : IJsonStreamWriter, IJsonStreamWriterAsync, IDisposable, IAsyncDisposable
+#else
+    internal sealed partial class JsonWriter :IJsonStreamWriter, IJsonStreamWriterAsync, IDisposable
```
gathogojr, line 31: nit:
```suggestion
    internal sealed partial class JsonWriter : IJsonStreamWriter, IJsonStreamWriterAsync, IDisposable
```

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/NonIndentedTextWriterAsyncTests.cs
```diff
@@ -65,45 +65,46 @@ public async Task WriteSpecialCharactersShouldWork()
...
             this.buffer = new Ref<char[]>();
-            this.stream = new MemoryStream();
+            this.stream = new AsyncOnlyStreamWrapper(new MemoryStream());
             StreamWriter innerWriter = new StreamWriter(this.stream);
-            innerWriter.AutoFlush = true;
             this.writer = new NonIndentedTextWriter(innerWriter);
         }
 
-        private string StreamToString()
+        private async Task<string> StreamToString()
```
gathogojr, line 102: nit:
```suggestion
        private async Task<string> StreamToStringAsync()
```

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightWriterTests.cs
```diff
@@ -2031,9 +2036,15 @@ public async Task WriteEndAsync_ThrowsExceptionForBinaryStreamNotDisposed()
                 container = containerBuilder.BuildContainer();
             }
 
+            Stream stream = this.stream;
```
gathogojr, line 2039: To avoid confusion, how about you name this local stream `messageStream` since it's assigned to the `MessageStream` property...

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -901,9 +902,25 @@ public async Task WriteMetadataDocumentAsync_WorksForJsonCsdl()
...
+            // We should disable synchronous I/O here once CsdlWriter has an async API.
+            // See: https://github.com/OData/odata.net/issues/2684
             string payload = await this.WriteAndGetPayloadAsync(edmModel, "application/json", async omWriter =>
             {
-                await omWriter.WriteMetadataDocumentAsync();
+                try
+                {
+                    await omWriter.WriteMetadataDocumentAsync();
+                }
+                catch (Exception e)
```
gathogojr, line 915: We should catch _specific_ exception here if possible

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -942,7 +959,19 @@ public async Task WriteMetadataDocumentAsync_WorksForXmlCsdl()
             // Act - XML
             string payload = await this.WriteAndGetPayloadAsync(edmModel, "application/xml", async omWriter =>
             {
-                await omWriter.WriteMetadataDocumentAsync();
+                try
+                {
+                    await omWriter.WriteMetadataDocumentAsync();
+                }
+                catch (Exception e)
```
gathogojr, line 966: We should catch _specific_ exception here if possible

src/Microsoft.OData.Core/ODataMetadataOutputContext.cs
```diff
@@ -182,6 +182,49 @@ internal override void WriteMetadataDocument()
...
+                try
+                {
+                    if (this.xmlWriter != null)
+                    {
+                        // In the async case the underlying stream is the async buffered stream, so we have to flush that explicitly.
+                        if (this.asynchronousOutputStream != null)
+                        {
+                            this.xmlWriter.FlushAsync().Wait();
+                            this.asynchronousOutputStream.FlushAsync().Wait();
+                            this.asynchronousOutputStream.Dispose();
```
mikepizzo, line 197: why do we call async version of flush, but sync version of dispose?

src/Microsoft.OData.Core/ODataOutputContext.cs
```diff
@@ -686,6 +690,30 @@ protected virtual void Dispose(bool disposing)
...
+#if NETCOREAPP3_1_OR_GREATER
+        /// <summary>
+        /// IAsyncDisposable.DisposeAsync() implementation to asynchronously cleanup unmanaged resources of the context.
+        /// </summary>
+        /// <returns>A task representing the asynchronous disposal of the context.</returns>
+        public async ValueTask DisposeAsync()
+        {
+            await this.DisposeAsyncCore().ConfigureAwait(false);
+
+            this.Dispose(false);
```
mikepizzo, line 702: Again, is it okay to call synchronous dispose in asynchronous DisposeAsync?

mikepizzo (changes requested):

src/Microsoft.OData.Core/ODataMessageWriter.cs
```diff
@@ -730,6 +734,20 @@ public void Dispose()
...
+#if NETCOREAPP3_1_OR_GREATER
+        /// <summary>
+        /// <see cref="System.IAsyncDisposable.DisposeAsync"/> implementation to asynchronously
+        /// clean up unmanaged resources of the writer.
+        /// </summary>
+        /// <returns>A task representing the asynchronous operation of disposing the writer.</returns>
+        public async ValueTask DisposeAsync()
+        {
+            await this.DisposeAsyncCore().ConfigureAwait(false);
+            this.Dispose(false);
```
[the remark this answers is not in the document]
habbes: When `Dispose(bool disposing)` is called with `false`, it doesn't dispose managed resources, it only frees unmanaged resources. The `Dispose(bool disposing)` method has an `if` check that only performs the (synchronous) resource disposal if `disposing` is set to `true`. This pattern helps ensure that both the sync and async dispose methods behave uniformly as far as releasing unmanaged resources is concerned. This pattern is also mentioned in the official `DisposeAsync` guidelines: https://learn.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-disposeasync#the-disposeasync-method


That said, since in this case we don't actually have references to any unmanaged resources, `Dispose(false)` doesn't actually do anything and we could remove that call if it causes confusion.

src/Microsoft.OData.Core/ODataMetadataOutputContext.cs
```diff
@@ -182,6 +182,49 @@ internal override void WriteMetadataDocument()
...
+            {
+                try
+                {
+                    if (this.xmlWriter != null)
+                    {
+                        // In the async case the underlying stream is the async buffered stream, so we have to flush that explicitly.
+                        if (this.asynchronousOutputStream != null)
+                        {
+                            this.xmlWriter.FlushAsync().Wait();
+                            this.asynchronousOutputStream.FlushAsync().Wait();
```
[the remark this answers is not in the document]
habbes: Yes. `xmlWriter` is set in the constructor and only set to null in the `Dispose` method.

src/Microsoft.OData.Core/ODataMetadataOutputContext.cs
habbes: Good point. This method was implemented way before we had implemented `DisposeAsync()`. But we should actually call `Flush` instead of `FlushAsync()` in this method since `FlushAsync().Wait()` is arguably more costly than just calling the synchronous `Flush()`: https://github.com/davidfowl/AspNetCoreDiagnosticScenarios/blob/master/AsyncGuidance.md#avoid-using-taskresult-and-taskwait

src/Microsoft.OData.Core/ODataOutputContext.cs
habbes: Same case as https://github.com/OData/odata.net/pull/2683/files#r1228957103

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataUtf8JsonWriterAsyncTests.cs
```diff
@@ -43,17 +43,10 @@ public ODataUtf8JsonWriterAsyncTests()
             this.disposed = false;
         }
 
-        public void Dispose()
+        public async ValueTask DisposeAsync()
```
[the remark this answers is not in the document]
habbes: Thanks. I've opted to use `if (this.disposed)` instead of removing it.

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
habbes: Done. Naming suggestions would be cool too, cause I'm not sure about `ODataSynchronousIOException` :)

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
habbes: See: https://github.com/OData/odata.net/pull/2683/files#r1228980286

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
gathogojr: @habbes ASP.NET Core throws a `System.InvalidOperationException`. I think a name like `SynchronousIOException` would do the job

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -1123,18 +1319,35 @@ private static IEdmModel GetEdmModel()
...
 
+                // Dispose stream manually to avoid synchronous I/O
+#if NETCOREAPP3_1_OR_GREATER
+                await message.Stream.DisposeAsync();
+#else
+                message.Stream.Dispose();
+#endif
                 return contents;
             }
+
```
gathogojr, line 1350: nit:
```suggestion
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -942,7 +958,19 @@ public async Task WriteMetadataDocumentAsync_WorksForXmlCsdl()
             // Act - XML
             string payload = await this.WriteAndGetPayloadAsync(edmModel, "application/xml", async omWriter =>
             {
-                await omWriter.WriteMetadataDocumentAsync();
+                try
+                {
+                    await omWriter.WriteMetadataDocumentAsync();
+                }
+                catch (Exception e)
```
gathogojr, line 965:
```suggestion
                catch (ODataSynchronousIOException e)
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -965,6 +993,160 @@ public async Task WriteMetadataDocumentAsync_WorksForXmlCsdl()
...
+                }
+            };
+
+            ODataMessageWriterSettings writerSettings = new ODataMessageWriterSettings();
+            writerSettings.EnableMessageStreamDisposal = true;
+            writerSettings.BaseUri = new Uri("http://www.example.com/");
+            writerSettings.SetServiceDocumentUri(new Uri("http://www.example.com/"));
+            message.SetHeader("Content-Type", "application/json");
+
+            var msgWriter = new ODataMessageWriter((IODataResponseMessageAsync)message, writerSettings, edmModel);
```
gathogojr, line 1035: Any reason you couldn't use `await using (var msgWriter ...`?

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -965,6 +993,160 @@ public async Task WriteMetadataDocumentAsync_WorksForXmlCsdl()
...
+
+            try
+            {
+                await msgWriter.WriteMetadataDocumentAsync();
+            }
+            catch (Exception e)
```
gathogojr, line 1072:
```suggestion
            catch (ODataSynchronousIOException e)
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -965,6 +993,160 @@ public async Task WriteMetadataDocumentAsync_WorksForXmlCsdl()
...
+            writerSettings.SetServiceDocumentUri(new Uri("http://www.example.com/"));
+            message.SetHeader("Content-Type", "application/xml");
+
+            var msgWriter = new ODataMessageWriter((IODataResponseMessageAsync)message, writerSettings, edmModel);
+
+            try
+            {
+                await msgWriter.WriteMetadataDocumentAsync();
+            }
+            catch (Exception e)
```
gathogojr, line 1110:
```suggestion
            catch (ODataSynchronousIOException e)
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -965,6 +993,160 @@ public async Task WriteMetadataDocumentAsync_WorksForXmlCsdl()
...
+                Stream = stream
+            };
+
+            IEdmModel edmModel = GetEdmModel();
+            var settings = new ODataMessageWriterSettings
+            {
+                EnableMessageStreamDisposal = true,
+            };
+
+            var writer = new ODataMessageWriter((IODataResponseMessageAsync)message, settings, edmModel);
```
gathogojr, line 1140: Any reason you couldn't `await using` here?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchWriterTests.cs
```diff
@@ -190,14 +203,18 @@ public async Task WriteBatchRequestWithDependsOnIdsAsync()
...
                         "POST", new Uri($"{ServiceUri}/Orders"), "2", BatchPayloadUriOption.AbsoluteUri, dependsOnIds);
 
+#if NETCOREAPP3_1_OR_GREATER
+                    await using (var messageWriter2 = new ODataMessageWriter(operationRequestMessage2))
+#else
                     using (var messageWriter2 = new ODataMessageWriter(operationRequestMessage2))
-                    {
-                        var jsonLightWriter = await messageWriter2.CreateODataResourceWriterAsync(this.orderEntitySet, this.orderEntityType);
+#endif
+                        {
```
gathogojr, line 211: Nesting looks off to me...

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightOutputContextTests.cs
habbes: My bad. Thought I had removed this.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightOutputContextTests.cs
habbes: I refactored the remaining `try-finally` versions.

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchWriterTests.cs
habbes: Fixed.

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
habbes: Fixed.

src/Microsoft.OData.Core/ODataMetadataOutputContext.cs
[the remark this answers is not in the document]
mikepizzo: :thumbsup:

src/Microsoft.OData.Core/ODataMetadataOutputContext.cs
mikepizzo: :thumbsup:

mikepizzo (dismissed):
:shipit:

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -858,46 +858,103 @@ await this.FlushAsync()
...
+                        {
+                            disposableWriter.Dispose();
+                        }
+
+                        // In the async case the underlying stream is the async buffered stream, so we have to flush that explicitly.
+                        if (this.asynchronousOutputStream != null)
+                        {
+#if NETSTANDARD1_1
+                        this.asynchronousOutputStream.FlushSync();
+                        this.asynchronousOutputStream.Dispose();
```
gathogojr, line 881: nit: The code formatting here is off...

src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs
```diff
@@ -175,22 +175,53 @@ internal static Utf8JsonWriter CreateJsonWriter(Stream stream, ODataMessageWrite
...
+        }
+
+#if NETCOREAPP3_1_OR_GREATER
+        protected override async ValueTask DisposeAsyncCore()
         {
             try
             {
                 if (this.jsonWriter != null)
                 {
+                    await this.jsonWriter.FlushAsync().ConfigureAwait(false);
```
gathogojr, line 216: The logic here is a bit different from the similar one in `Dispose` method. In `Dispose` method, these two statements are in an `else` block corresponding to `if (this.asynchronousOutputStream != null)`

src/Microsoft.OData.Core/ODataMessageWriter.cs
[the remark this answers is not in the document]
habbes: Add the comment.

src/Microsoft.OData.Core/ODataMessageWriter.cs
```diff
@@ -1193,6 +1211,28 @@ private void Dispose(bool disposing)
...
         }
 
+#if NETCOREAPP3_1_OR_GREATER
+        /// <summary>
+        /// Asynchronously performs the actual cleanup work.
+        /// </summary>
+        /// <returns>A task representing the asynchronous disposal of the writer.</returns>
+        private async ValueTask DisposeAsyncCore()
+        {
+            this.isDisposed = true;
```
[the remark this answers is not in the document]
habbes: It's used by the `VerifyNotDisposed` method to throw an `ObjectDisposedException` if the flag is set to `true`.

src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs
habbes: This `Dispose()` method first checks whether `this.asynchronousOutputStream` is not null, if it's not null, it calls `DisposeOutputStreamAsync()` method which asynchronously flushes and disposes the writer, then calls `Wait()` on the async method, otherwise it performs synchronously flush and dispose in the `else` block.. In `DisposeAsyncCore()` I didn't bother with synchronous flushes and disposal of the `JsonWriter`.

mikepizzo (approved):
:shipit:

gathogojr (approved):
:shipit:

src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs
```diff
@@ -175,22 +175,54 @@ internal static Utf8JsonWriter CreateJsonWriter(Stream stream, ODataMessageWrite
...
+        {
+            if (disposing)
+            {
+                try
+                {
+                    if (this.jsonWriter != null)
+                    {
+                        if (this.asynchronousOutputStream != null)
+                        {
+                            DisposeOutputStreamAsync().Wait();
```
joaocpaiva, line 187: It would be better to avoid the **sync over async** .Wait() call here - dispose the stream synchronously for the Dispose() implementation while the DisposeAsync() implementation can do it properly (asynchronously).

joaocpaiva: Basically, I would try make sure of employing two best practices:
1. No use of **sync over async** (i.e., Task.Wait(), Task.Result, Task.GetAwaiter().GetResult()). In the Dispose() implementation just call synchronous Flush().
2. No use of synchronous Flush() in the async dispose implementation (DisposeAsyncCore). Always uses **await FlushAsync();**

PR #2691, opened 2023-06-15
feature/Deep insert response

KenitoInc:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary


KenitoInc:

src/Microsoft.OData.Client/Materialization/FeedAndEntryMaterializerAdapter.cs
```diff
@@ -86,8 +89,12 @@ internal FeedAndEntryMaterializerAdapter(ODataFormat odataFormat, ODataReaderWra
             if (isDeltaPayload)
             {
                 // If we have a delta payload, we need the stack data structure to keep track of the read items.
-                this.materializerStateItems = new Stack<IMaterializerState>();
-            }    
+                this.deltaMaterializerStateItems = new Stack<IMaterializerState>();
+            }
+            else
+            {
```
ElizabethOkerio, line 95: will this be created all the time even when we are not doing a deep insert? Is it possible to check whether we are doing a deep insert or not from this point?

ElizabethOkerio: Is it possible to use the same `stack` for both deep updates and deep inserts instead of having different `stacks` of the same type?

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
```diff
@@ -591,7 +591,7 @@ private void MaterializeResolvedEntry(MaterializerEntry entry, bool includeLinks
 
                 foreach (ODataNestedResourceInfo link in entry.NestedResourceInfos)
                 {
-                    MaterializerNavigationLink linkState = MaterializerNavigationLink.GetLink(link, this.MaterializerContext);
+                    MaterializerNestedEntry linkState = MaterializerNestedEntry.GetNestedEntry(link, this.MaterializerContext);
```
ElizabethOkerio, line 594: Can we get rid of the `MaterializerNavigationLink`? since we've replaced it with the `MaterializerNestedEntry`? Will it require a lot of refactoring?

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
```diff
@@ -611,7 +611,7 @@ private void MaterializeResolvedEntry(MaterializerEntry entry, bool includeLinks
                     }
 
                     // includeLinks is for Navigation property, so we should handle complex property when includeLinks equals false;
-                    if (!includeLinks && (prop.IsEntityCollection || prop.EntityCollectionItemType != null))
+                    if (!includeLinks && prop.EdmProperty is IEdmNavigationProperty)
```
ElizabethOkerio, line 614: I do not understand this change..Can we have a navigation property if `includeLinks` is false?

src/Microsoft.OData.Client/Materialization/MaterializerNestedEntry.cs
```diff
@@ -20,6 +20,12 @@ internal class MaterializerNestedEntry : IMaterializerState
         /// <summary>The <see cref="IMaterializerState"/> entries.</summary>
         private readonly List<IMaterializerState> nestedItems;
 
+        /// <summary>The entry.</summary>
+        private MaterializerEntry entry;
+
+        /// <summary>The feed.</summary>
+        private ODataResourceSet feed;
+
```
ElizabethOkerio, line 28: why do we need to add these  here? Since the nestedItems is of type `List<IMaterializerState>` and the `MaterializerEntry` and the `MaterializerFeed` that represents the `ODataResourceSet` already derive from `IMaterializerState`?

src/Microsoft.OData.Client/Materialization/ODataMaterializerContext.cs
```diff
@@ -15,15 +15,27 @@ namespace Microsoft.OData.Client.Materialization
...
+        private bool includeLinks;
+
         /// <summary>
         /// Initializes a materializer context
         /// </summary>
         /// <param name="responseInfo">Response information used to initialize with the materializer</param>
         /// <param name="materializerCache">The materializer cache.</param>
-        internal ODataMaterializerContext(ResponseInfo responseInfo, MaterializerCache materializerCache)
+        /// <param name="includeLinks">Whether to include navigation properties when materializing an entry.</param>
+        internal ODataMaterializerContext(ResponseInfo responseInfo, MaterializerCache materializerCache, bool includeLinks = true)
```
ElizabethOkerio, line 26: should the default `includeLinks` value here be true or false?

src/Microsoft.OData.Client/MaterializeFromAtom.cs
```diff
@@ -104,7 +104,8 @@ internal class MaterializeAtom : IDisposable, IEnumerable, IEnumerator
         /// <param name="elementType">result type.</param>
         /// <param name="format">The format of the response being materialized from.</param>
         /// <param name="materializerCache">Cache used to store temporary metadata used for materialization of OData items.</param>
-        internal MaterializeAtom(ResponseInfo responseInfo, IEnumerable<ODataResource> entries, Type elementType, ODataFormat format, MaterializerCache materializerCache)
+        /// <param name="includeLinks">Whether to include navigation properties when materializing an entry.</param>
+        internal MaterializeAtom(ResponseInfo responseInfo, IEnumerable<ODataResource> entries, Type elementType, ODataFormat format, MaterializerCache materializerCache, bool includeLinks = true)
         {
```
ElizabethOkerio, line 109: here too? should the default `includeLinks` value be true or false?

src/Microsoft.OData.Client/Materialization/FeedAndEntryMaterializerAdapter.cs
```diff
@@ -510,28 +542,54 @@ private ODataNestedResourceInfo ReadNestedResourceInfo()
             {
                 if (feed != null)
                 {
-                    MaterializerNavigationLink.CreateLink(link, feed, this.materializerContext);
+                    nestedEntry = MaterializerNestedEntry.CreateNestedEntry(
+                        link,
+                        new List<IMaterializerState>(),
+                        this.materializerContext);
+
+                    nestedEntry.Feed = feed;
```
ElizabethOkerio, line 550: why can't we add the `materializerFeed` to the `nestedEntry`'s `nestedItems` here?

src/Microsoft.OData.Client/Materialization/FeedAndEntryMaterializerAdapter.cs
```diff
@@ -389,6 +396,20 @@ private void AddResourceToParent(IMaterializerState materializerState)
             }
         }
 
+        /// <summary>
+        /// Adds an <see cref="IMaterializerState"/> entry to a parent entry.
+        /// </summary>
+        /// <param name="materializerState">The <see cref="IMaterializerState"/> entry being added to the parent item.</param>
+        private void AddResourceToParentEntry(IMaterializerState materializerState)
```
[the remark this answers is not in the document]
KenitoInc: Because we have different stacks. Let me see if we could use on stack

src/Microsoft.OData.Client/Materialization/MaterializerNestedEntry.cs
KenitoInc: For backwards compatibility to match how the `MaterializerNavigationLink` worked.

src/Microsoft.OData.Client/Materialization/ODataMaterializerContext.cs
KenitoInc: The default behaviour is `true`.
For deep insert, we want the behaviour to be `false` since for each level, we don't need to include links since they will be processed in the next level.

src/Microsoft.OData.Client/MaterializeFromAtom.cs
KenitoInc: true. As explained above

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -2339,6 +2339,28 @@ internal virtual DataServiceResponse EndBulkUpdate(IAsyncResult asyncResult)
 
         #endregion
 
+        #region DeepInsert
+
+        /// <summary>
+        /// Processes deep insert requests. Creates an object and creates related navigation items or link existing navigation items in a single request.
```
ElizabethOkerio, line 2345:
```suggestion
        /// Processes deep insert requests. Creates an object and it's related navigation items or link existing navigation items to the created object in a single request. 

```

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -225,19 +268,258 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
...
+            }
+            catch (InvalidOperationException ex)
+            {
+                HeaderCollection headers = new HeaderCollection(this.batchResponseMessage);
+                int statusCode = this.batchResponseMessage == null ? (int)HttpStatusCode.InternalServerError : (int)this.batchResponseMessage.StatusCode;
+                DataServiceResponse dataServiceResponse = new DataServiceResponse(headers, statusCode, Enumerable.Empty<OperationResponse>(), this.IsBatchRequest);
+                throw new DataServiceRequestException(Strings.DataServiceException_GeneralError, ex, dataServiceResponse);
+            }
+
+            HandleDeepInsertResponseInternal(entry, true, entityDescriptor, null);
```
ElizabethOkerio, line 350: We materialize the response when `EndRequest` gets called but I see that's not the case here. I'm not sure whether there is any effect in either approach.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -112,6 +142,7 @@ internal void BuildDescriptorGraph<T>(IEnumerable<Descriptor> descriptors, bool
                         throw Error.InvalidOperation(Strings.Context_DeepInsertOneTopLevelEntity);
                     }
 
+                    topLevelDescriptor.ContentGeneratedForSave = true;
```
ElizabethOkerio, line 145: what does this do?

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
KenitoInc: That's the case here.
`HandleDeepInsertResponseInternal` is called from `HandleDeepInsertResponse` which is called from `HandleResponse` which is called from `EndRequest`.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
KenitoInc: Ensure when we process an entity descriptor, we also process the related links. see https://github.com/OData/odata.net/blob/4414000c0da72b22ff3c4a96f51cbf35f2316bca/src/Microsoft.OData.Client/BaseSaveResult.cs#L836

src/Microsoft.OData.Client/Materialization/FeedAndEntryMaterializerAdapter.cs
KenitoInc: The original `MaterializerNavigationLink` uses feed(ODataResourceSet). This allows backwards compatibility.
Unless we refactor the existing logic, which is beyond the scope of this PR

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
KenitoInc: Fixed.

src/Microsoft.OData.Client/Materialization/FeedAndEntryMaterializerAdapter.cs
```diff
@@ -86,8 +89,12 @@ internal FeedAndEntryMaterializerAdapter(ODataFormat odataFormat, ODataReaderWra
             if (isDeltaPayload)
             {
                 // If we have a delta payload, we need the stack data structure to keep track of the read items.
-                this.materializerStateItems = new Stack<IMaterializerState>();
-            }    
+                this.deltaMaterializerStateItems = new Stack<IMaterializerState>();
```
habbes, line 92: Why this change?

src/Microsoft.OData.Client/Materialization/FeedAndEntryMaterializerAdapter.cs
KenitoInc: We can't use the same stack since in `DeltaSet` requests, we validate that the top-level object is not a `MaterializerEntry`.

src/Microsoft.OData.Client/Materialization/FeedAndEntryMaterializerAdapter.cs
KenitoInc: We added another stack to track `MaterializerEntry` items. So I renamed to have a clear distinction between them.

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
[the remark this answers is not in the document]
KenitoInc: The intention of the original code was to check:
-  If includeLinks is not true, and
-  If the property is not a collection of complex type.

That code didn't make sense since it ignored single-value navigation property.

In-short `includeLinks` should only apply to navigation properties (single-value and collections) and ignore all complex types.

src/Microsoft.OData.Client/BulkUpdateSaveResult.cs
```diff
@@ -389,7 +389,7 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
         /// <returns>An instance of <see cref="MaterializeAtom"/> that can be used to materialize the response.</returns>
         protected override MaterializeAtom GetMaterializer(EntityDescriptor entityDescriptor, ResponseInfo responseInfo)
         {
-            Debug.Assert(this.cachedResponse.Exception == null && this.materializerStateForDescriptor != null, "this.cachedResponse.Exception == null && this.descriptorForMaterializer != null");
```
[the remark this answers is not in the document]
KenitoInc: It was a wrong property name in the `Debug.Assert` message

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
KenitoInc: `ContentGeneratedForSave` means that the descriptor should be processed during `SaveChanges`. It's an existing property that we are re-using.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -225,19 +268,258 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
...
+                HeaderCollection headers = new HeaderCollection(this.batchResponseMessage);
+
+                ResponseInfo responseInfo = this.CreateResponseInfo(entityDescriptor);
+                HttpWebResponseMessage responseMessageWrapper = new HttpWebResponseMessage(
+                    headers,
+                    this.batchResponseMessage.StatusCode,
+                    () => this.ResponseStream);
+
+                ODataMaterializerContext materializerContext = new ODataMaterializerContext(responseInfo, base.MaterializerCache);
+                entry = ODataReaderEntityMaterializer.ParseSingleEntityPayload(responseMessageWrapper, responseInfo, entityDescriptor.Entity.GetType(), materializerContext);
```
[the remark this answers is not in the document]
KenitoInc: `ParseSingleEntityPayload` is an existing method that is used to read a response and create an entry.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -225,19 +268,258 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
...
+            EntityDescriptor entityDescriptor = this.bulkUpdateGraph.TopLevelDescriptors[0];
+            MaterializerEntry entry = default;
+            DataServiceResponse response = new DataServiceResponse(headers, (int)this.batchResponseMessage.StatusCode, operationResponses, this.IsBatchRequest);
+
+            try
+            {
+                // We are setting the MaxProtocolVersion to v4.01 when reading deep insert responses.
+                this.RequestInfo.Context.MaxProtocolVersion = ODataProtocolVersion.V401;
+                Version responseVersion;
+                Exception exception = HandleResponse(
```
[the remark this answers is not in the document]
KenitoInc: Actually we don't need this exception here since we pass `throwOnFailure : true` in `HandleResponse`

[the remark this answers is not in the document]
KenitoInc: #Fixed

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -225,19 +268,258 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
...
+                DataServiceResponse dataServiceResponse = new DataServiceResponse(headers, statusCode, Enumerable.Empty<OperationResponse>(), this.IsBatchRequest);
+                throw new DataServiceRequestException(Strings.DataServiceException_GeneralError, ex, dataServiceResponse);
+            }
+
+            HandleDeepInsertResponseInternal(entry, true, entityDescriptor, null);
+
+            return response;
+        }
+
+        private void HandleDeepInsertResponseInternal(MaterializerEntry entry, bool isTopLevelDescriptor, Descriptor descriptor, OperationResponse parentOperationResponse)
```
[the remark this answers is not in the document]
KenitoInc: It cannot be `null`.
In the implementation of `ParseSingleEntityPayload`, when the entry is null, an exception is thrown https://github.com/OData/odata.net/blob/75fa5a7591b3af0a14e0c2158b34a7c4b9556787/src/Microsoft.OData.Client/Materialization/ODataReaderEntityMaterializer.cs#L155

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -225,19 +268,258 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
...
+        {
+            OperationResponse response = CreateOperationResponse(entry, isTopLevelDescriptor, descriptor, parentOperationResponse);
+
+            List<Descriptor> relatedDescriptors = this.BulkUpdateGraph.GetRelatedDescriptors(descriptor);
+
+            foreach (IMaterializerState nestedItem in entry?.NestedItems)
```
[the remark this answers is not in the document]
KenitoInc: `entry` can never be null. See this comment https://github.com/OData/odata.net/pull/2691#discussion_r1267557231

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -225,19 +268,258 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
...
+            {
+                if (nestedItem is MaterializerNestedEntry materializerNestedEntry)
+                {
+                    foreach (IMaterializerState nestedEntries in materializerNestedEntry?.NestedItems)
+                    {
+                        if (nestedEntries is MaterializerFeed feed)
+                        {
+                            for (int i = 0; i < feed.Items.Count; i++)
+                            {
+                                if (relatedDescriptors.Count == 0 || relatedDescriptors.Count < i + 1)
```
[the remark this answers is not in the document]
KenitoInc: Ensures the `count` of the `descriptors` matches the `feed.Items` count.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -225,19 +268,258 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
...
+                                {
+                                    break;
```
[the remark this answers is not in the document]
KenitoInc: This is just a sanity check in the unlikely event of having nested objects with no matching descriptors.
Client allows us to detach descriptors.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
KenitoInc: The behaviour in this PR is similar to what is in batch operations. I don't think it should be a problem

src/Microsoft.OData.Client/Materialization/FeedAndEntryMaterializerAdapter.cs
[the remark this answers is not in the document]
KenitoInc: We have different stacks since in bulk update we validate that the top level entity should not be a `MaterializerEntry`

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -225,19 +268,259 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
...
+                ODataMaterializerContext materializerContext = new ODataMaterializerContext(responseInfo, base.MaterializerCache);
+                entry = ODataReaderEntityMaterializer.ParseSingleEntityPayload(responseMessageWrapper, responseInfo, entityDescriptor.Entity.GetType(), materializerContext);
+
+                this.headers = headers;
+                this.topLevelEntry = entry;
+            }
+            catch (DataServiceRequestException)
+            {
+                throw;
+            }
```
gathogojr, line 341: Do we need this block? We just catch the `DataServiceRequestException` and then rethrow therefore I don't see what purpose it serves

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -225,19 +268,259 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
...
+            catch (InvalidOperationException ex)
+            {
+                HeaderCollection headers = new HeaderCollection(this.batchResponseMessage);
+                int statusCode = this.batchResponseMessage == null ? (int)HttpStatusCode.InternalServerError : (int)this.batchResponseMessage.StatusCode;
+                DataServiceResponse dataServiceResponse = new DataServiceResponse(headers, statusCode, Enumerable.Empty<OperationResponse>(), this.IsBatchRequest);
+                throw new DataServiceRequestException(Strings.DataServiceException_GeneralError, ex, dataServiceResponse);
+            }
+
+            HandleDeepInsertResponseInternal(entry, true, entityDescriptor, null);
```
gathogojr, line 350: Can you use named parameters here to give someone an idea what `true` and `null` represent...

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -225,19 +268,259 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
...
+            return response;
+        }
+
+        private void HandleDeepInsertResponseInternal(MaterializerEntry entry, bool isTopLevelDescriptor, Descriptor descriptor, OperationResponse parentOperationResponse)
+        {
+            OperationResponse response = this.CreateOperationResponse(entry, isTopLevelDescriptor, descriptor, parentOperationResponse);
+
+            List<Descriptor> relatedDescriptors = this.BulkUpdateGraph.GetRelatedDescriptors(descriptor);
+
+            foreach (IMaterializerState nestedItem in entry?.NestedItems)
```
gathogojr, line 361: Given your explanation that `entry` can never be null, why retain the `?` operator? It's a distraction to reviewers and maintainers of the code...

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -225,19 +268,259 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
...
+                                    break;
+                                }
+
+                                HandleDeepInsertResponseInternal(feed.Items[i] as MaterializerEntry, false, relatedDescriptors[i], response);
+                            }
+                        }
+                        else if (nestedEntries is MaterializerEntry nestedEntry)
+                        {
+                            // In single value navigation property, the related descriptors count must be 1.
+                            if (relatedDescriptors.Count == 0)
```
gathogojr, line 383: Can we find ourselves here and at the same time find that the `relatedDescriptors` collection has 0 items?

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -225,19 +268,259 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
...
+                if (nestedItem is MaterializerNestedEntry materializerNestedEntry)
+                {
+                    foreach (IMaterializerState nestedEntries in materializerNestedEntry.NestedItems)
+                    {
+                        if (nestedEntries is MaterializerFeed feed)
+                        {
+                            for (int i = 0; i < feed.Items.Count; i++)
+                            {
+                                // Ensures the count of the descriptors matches the feed.Items count.
+                                if (relatedDescriptors.Count < i + 1)
```
gathogojr, line 372: Question: In what scenarios would the number of items in the feed be different from the number of items in the `relatedDescriptors` collection? Trying to understand whether we should silent break and move or raise an exception...

src/Microsoft.OData.Client/Materialization/FeedAndEntryMaterializerAdapter.cs
```diff
@@ -42,8 +42,11 @@ internal class FeedAndEntryMaterializerAdapter
         /// <summary>The current entry.</summary>
         private ODataResource currentEntry;
 
-        /// <summary>The stack of read <see cref="IMaterializerState"/> items.</summary>
-        private readonly Stack<IMaterializerState> materializerStateItems;
+        /// <summary>The stack of read <see cref="IMaterializerState"/> items for ODataDeltaResourceSet.</summary>
+        private readonly Stack<IMaterializerState> deltaMaterializerStateItems;
+
+        /// <summary>The stack of read <see cref="IMaterializerState"/> items for ODataResource.</summary>
+        private readonly Stack<IMaterializerState> entryMaterializerStateItems;
```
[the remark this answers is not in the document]
KenitoInc: An `entry` is an `ODataResource`.
The other one has `delta`

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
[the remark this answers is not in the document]
KenitoInc: `MaterializerNestedEntry`  has new properties that `MaterializerNavigationLink` didn't have. We first introduced `MaterializerNestedEntry ` during `bulk update` implementation. I opted to use them same in `deep insert` and get rid of `MaterializerNavigationLink`

src/Microsoft.OData.Client/Materialization/ODataMaterializer.cs
```diff
@@ -352,16 +364,13 @@ protected static ODataMessageReader CreateODataMessageReader(IODataResponseMessa
         {
             ODataMessageReaderSettings settings = responseInfo.ReadHelper.CreateSettings();
 
-            // We use v401 to read delta payloads.
-            if (payloadKind == ODataPayloadKind.Delta)
+            // We use v4.01 to read all payloads. This is backwards compatible with v4.0
```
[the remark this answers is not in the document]
KenitoInc: The code block with the delta check was introduced during bulk update to support reading delta payloads (which only works in v4.01). We need the same for deep insert that's why I removed the check.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -79,6 +108,7 @@ internal void DeepInsertRequest<T>(T resource)
             try
             {
                 this.batchResponseMessage = this.RequestInfo.GetSynchronousResponse(deepInsertRequestMessage, false);
+                this.responseStream = this.batchResponseMessage.GetStream();
```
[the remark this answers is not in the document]
KenitoInc: The naming can be confusing but we are re-using the `protected IODataResponseMessage batchResponseMessage;` in `BaseSaveResult`.

We could introduce a new property but it serves the same purpose

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
KenitoInc: The top level `entry` can never be `null`.
I am not sure about the nested entries

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
KenitoInc: If the descriptors have been detached

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertE2ETests.cs
```diff
@@ -0,0 +1,715 @@
...
+                Name = "Person 100",
+            };
+
+            var car = new Car
+            {
+                ID = 1001,
+                Name = "Car 1001"
+            };
+
+            this.context.AddObject("Persons", person);
```
habbes, line 164: What happens when you insert an object without an ID and the ID is generated by the server? Will the response contain the updated value? Will this be reflected in the entity tracker?

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertE2ETests.cs
habbes, line 1: Do you have tests where the response contains values that were not generated by the client, whether for the root entity or its nested children, or (better yet) both? Example of real-world use cases are auto-generated IDs, creation dates, status fields, etc. Will the objects in the deserialized response objects include these fields? Will they be updated in the entity tracker? If you don't have such tests, could you add some? Especially curious to see whether the entity tracker entries for the inserted objects get updated if the response contain server-generated values.

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
```diff
@@ -225,19 +268,259 @@ protected override ODataRequestMessageWrapper CreateRequestMessage(string method
...
         protected override DataServiceResponse HandleResponse()
         {
-            throw new NotImplementedException();
+            // This will process the responses and throw if failure was detected
+            if (this.ResponseStream != null)
+            {
+                return this.HandleDeepInsertResponse();
+            }
+
+            return new DataServiceResponse(null, (int)WebExceptionStatus.Success, new List<OperationResponse>(0), batchResponse: true);
```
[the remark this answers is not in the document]
KenitoInc: Fixed.
It should be false

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
KenitoInc: If the descriptors have been detached

src/Microsoft.OData.Client/DeepInsertSaveResult.cs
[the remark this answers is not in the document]
KenitoInc: The underlying logic for SaveChanges and DeepInsert is the same.
The descriptors used in the request must be marked as processed when the request is done. We used the `ContentGeneratedForSave = true` to determine what should be processed.

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertE2ETests.cs
```diff
@@ -0,0 +1,715 @@
...
+
+            Assert.Single(response);
+            Assert.Single(response.Single().NestedResponses);
+
+            var changeOperationResponse = response.First() as ChangeOperationResponse;
+            EntityDescriptor entityDescriptor = changeOperationResponse.Descriptor as EntityDescriptor;
+            var returnedPerson = entityDescriptor.Entity as Person;
+            var nestedResponse = changeOperationResponse.NestedResponses[0] as ChangeOperationResponse;
+            EntityDescriptor carDescriptor = nestedResponse.Descriptor as EntityDescriptor;
+            var returnedCar = carDescriptor.Entity as Car;
```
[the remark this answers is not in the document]
KenitoInc: Yes

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertE2ETests.cs
KenitoInc: The response has the returned ID.
I have added the test.

In the implementation, we update the entry entitydescriptor with the materialized object. We then apply the updated descriptor to the entity tracker.
https://github.com/OData/odata.net/blob/34499522b6a3087aad1bb7568ed2ab97f7961a86/src/Microsoft.OData.Client/Materialization/MaterializerEntry.cs#L305
https://github.com/OData/odata.net/blob/34499522b6a3087aad1bb7568ed2ab97f7961a86/src/Microsoft.OData.Client/AtomMaterializerLog.cs#L180

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/Tests/DeepInsertE2ETests.cs
KenitoInc: Added

habbes (approved):

src/Microsoft.OData.Client/Materialization/ODataMaterializerContext.cs
ElizabethOkerio: I think this should be added for deep update also

PR #2701, opened 2023-07-04
Fix Filter and Orderby in Expand applied on bound functions

KenitoInc:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/ODataPathInfo.cs
+++ b/src/Microsoft.OData.Core/UriParser/ODataPathInfo.cs
@@ -5,6 +5,7 @@
 //---------------------------------------------------------------------
 
 using System.Collections.Generic;
+using System.Linq;
 using Microsoft.OData.Edm;
 
 namespace Microsoft.OData.UriParser
@@ -36,12 +37,26 @@ namespace Microsoft.OData.UriParser
 
                 this.targetNavigationSource = lastSegment.TargetEdmNavigationSource;
                 this.targetEdmType = lastSegment.EdmType;
+
                 if (this.targetEdmType != null)
                 {
                     IEdmCollectionType collectionType = this.targetEdmType as IEdmCollectionType;
                     if (collectionType != null)
                     {
                         this.targetEdmType = collectionType.ElementType.Definition;
+                    }
+                }
+
+                // If the last segment is an OperationSegment for a bound operation and the targetNavigationSource is null,
+                // We use the targetNavigationSource for the previous segment.
+                // e.g People/My.Function.GetPeopleWithDogs()
+                if (this.targetNavigationSource == null && lastSegment is OperationSegment operationSegment)
+                {
+                    IEdmOperation operation = operationSegment.Operations.FirstOrDefault();
+
+                    if (operation.IsBound & previous != null)
+                    {
+                        this.targetNavigationSource = previous.TargetEdmNavigationSource;
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/UriParser/ODataPathInfo.cs
```diff
@@ -44,6 +46,19 @@ public ODataPathInfo(ODataPath odataPath)
...
+                // If the last segment is an OperationSegment for a bound operation and the targetNavigationSource is null,
+                // We use the targetNavigationSource for the previous segment.
+                // e.g People/My.Function.GetPeopleWithDogs()
+                if (this.targetNavigationSource == null && lastSegment is OperationSegment operationSegment)
+                {
+                    IEdmOperation operation = operationSegment.Operations.FirstOrDefault();
+
+                    if (operation.IsBound & previous != null)
+                    {
+                        this.targetNavigationSource = previous.TargetEdmNavigationSource;
```
habbes, line 59: Could `previous.TargetEdmNavigationSource` also be null, or that isn't an issue?

habbes: What if the return type of the of the operation does not match the navigation source of the previous segment?
For example, what if we have something like `Customers/NS.GetTopOrders()?$expand=Products($filter=Category eq 'Electronics')`  where `GetTopOrders` returns a collection of `Order` entities from an `Orders` entity set, which is different from `Customers` entity set?

xuzhg: I don't think we should delegate the targetNS to the NS of preview segment.
I'd figure out why does ODL needs the targetNS? If the last operation doesn't have the targetNS, can we use the return type of the operation to do something? I do think we use the TargetNS to calculate the context URL, if we don't have the NS, we should use the type to generate the Context URL. @mikepizzo

xuzhg (changes requested):

src/Microsoft.OData.Core/UriParser/ODataPathInfo.cs
gathogojr, line 59: @KenitoInc Without even getting into whether this is the right fix, `previous.TargetNavigationSource` is null if `previous` is a `KeySegment` - e.g. for the case where the function is bound to an entity. On the other hand, `(previous as KeySegment).NavigationSource` is not null.

src/Microsoft.OData.Core/UriParser/ODataPathInfo.cs
```diff
@@ -44,6 +46,19 @@ public ODataPathInfo(ODataPath odataPath)
                         this.targetEdmType = collectionType.ElementType.Definition;
                     }
                 }
+
+                // If the last segment is an OperationSegment for a bound operation and the targetNavigationSource is null,
+                // We use the targetNavigationSource for the previous segment.
+                // e.g People/My.Function.GetPeopleWithDogs()
+                if (this.targetNavigationSource == null && lastSegment is OperationSegment operationSegment)
```
gathogojr, line 53: This is where you'd check if `previous` is null to avoid doing a lot of work for nothing...

KenitoInc:
I think there was a  clear reason why by design, applying `query options` (`$filter` applied on `$expand`) required an `EntitySetPath`.
The proposed fix in this PR was to assist in scenarios where the customer doesn't specify the `EntitySetPath`.
I have looked at various solutions and I am unable to come up with a fix because of the following reasons:
•	The function only specifies a return type (IEdmType) not the return entity set
•	It not possible to get the correct entityset from an edmtype since we could have multiple entitysets from one edmtype.

It’s from the `EntitySet` that we get the `TargetNavigationSource` to use in creating the `BindingState` during semantic binding by `SelectExpandBinder`, `FilterBinder` etc.

PR #2707, opened 2023-07-18
Remove sync I/O from JsonValueUtils async write methods

### LINKED ISSUE #2704: ODL performs synchronous I/O when writing escaped or binary strings asynchronously

### Assemblies affected

Microsoft.OData.Core 7.17.0

### Reproduce steps

Serializing the following sample payload may cause this issue:

```
string payload = "{\"PropertyName\": \"" + string.Join('\t', System.Linq.Enumerable.Repeat('\n', 95000)) + "\"}";
```

### Expected result

The async writer should not perform synchronous I/O.

### Actual result

The async writer performs synchronous I/O. This may cause exceptions in environments where synchronous I/O is not permitted, or it may block threads causing performance (and potentially security) issues.

### Additional detail

1: From `JsonValueUtils.WriteEscapedJsonStringAsync`

Stacktrace:
```
   at System.IO.BufferedStream.Write(ReadOnlySpan`1 buffer)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.IO.TextWriter.Write(Char[] buffer, Int32 index, Int32 count)
   at Microsoft.OData.Json.JsonValueUtils.<WriteEscapedJsonStringValueAsync>g__WriteEscapedJsonStringValueInnerAsync|60_0(TextWriter innerWriter, String innerInputString, ODataStringEscapeOption innerStringEscapeOption, Ref`1 innerBuffer, ICharArrayPool innerBufferPool, Int32 innerFirstIndex)
   at Microsoft.OData.Json.JsonValueUtils.WriteEscapedJsonStringAsync(TextWriter writer, String inputString, ODataStringEscapeOption stringEscapeOption, Ref`1 buffer, ICharArrayPool bufferPool)
   at Microsoft.OData.Json.JsonWriter.WriteValueAsync(String value)
   at Microsoft.OData.JsonLight.ODataJsonLightValueSerializer.WritePrimitiveValueAsync(Object value, IEdmTypeReference actualTypeReference, IEdmTypeReference expectedTypeReference)
```

 

2: From `JsonValueUtils.WriteBinaryStringAsync`

Stacktrace:
```
   at System.IO.BufferedStream.Write(ReadOnlySpan`1 buffer)
   at System.IO.StreamWriter.Flush(Boolean flushStream, Boolean flushEncoder)
   at System.IO.TextWriter.Write(Char[] buffer, Int32 index, Int32 count)
   at Microsoft.OData.Json.JsonValueUtils.WriteBinaryStringAsync(TextWriter writer, Byte[] value, Ref`1 buffer, ICharArrayPool arrayPool)
   at Microsoft.OData.Json.JsonValueUtils.WriteValueAsync(TextWriter writer, Byte[] value, Ref`1 buffer, ICharArrayPool arrayPool)
   at Microsoft.OData.Json.JsonWriter.WriteValueAsync(Byte[] value)
   at Microsoft.OData.JsonLight.ODataJsonLightValueSerializer.WritePrimitiveValueAsync(Object value, IEdmTypeReference actualTypeReference, IEdmTypeReference expectedTypeReference)
```

Here's a case where the asynchronous helper `JsonValueUtils.WriteEscapeJsonStringValueAsync` calls the synchronous `WriteEscapedStringToBuffer`: https://github.com/OData/odata.net/blob/75fa5a7591b3af0a14e0c2158b34a7c4b9556787/src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs#L441C1-L442C1

habbes:
### Issues

*This pull request fixes #2704 .*

### Description

The following async helper methods from `JsonValueUtils` had cases of synchronous I/O:
- `WriteEscapedJsonStringAsync`
- `WriteEscapedCharArrayAsync`
- `WriteValueAsync(byte[])`

Part of the reason was because the methods called other helper methods that performed synchronous `TextWriter.Write` calls. In such cases, I created true async versions of the affected methods.

The other reason was that the `NonIndentedTextWriter` did not override the async `WriteAsync(char[] buffer, int index, int count)` method. And the default implementation of `TextWriter.WriteAsync(char[], int, int)` calls the synchronous version. So I added an async override to the class.

Something else what pointing out is that because the internal `StreamWriter` is buffered, it would take a bunch of `Write()` calls before it actually flushes to the stream. This means that this issue would go undetected in many cases. To test this, I create a long payload to ensure that the `StreamWriter` actually flushes its content.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
@@ -788,6 +788,25 @@ namespace Microsoft.OData.Json
         }
 
         /// <summary>
+        /// Writes a substring starting at a specified position on the string to the buffer.
+        /// This method is intended for use in async methods, where you cannot pass ref parameters.
+        /// </summary>
+        /// <param name="inputString">Input string value.</param>
+        /// <param name="currentIndex">The index in the string at which the substring begins.</param>
+        /// <param name="buffer">Char buffer to use for streaming data.</param>
+        /// <param name="bufferIndex">Current position in the buffer after the substring has been written.</param>
+        /// <param name="substrLength">The length of the substring to be copied.</param>
+        private static void WriteSubstringToBuffer(string inputString, Ref<int> currentIndex, char[] buffer, Ref<int> bufferIndex, int substrLength)
+        {
+            Debug.Assert(inputString != null, "inputString != null");
+            Debug.Assert(buffer != null, "buffer != null");
+
+            inputString.CopyTo(currentIndex.Value, buffer, 0, substrLength);
+            bufferIndex.Value = substrLength;
+            currentIndex.Value += substrLength;
+        }
+
+        /// <summary>
         /// Writes an escaped string to the buffer.
         /// </summary>
         /// <param name="writer">The text writer to write the output to.</param>
--- a/src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
@@ -498,7 +498,8 @@ namespace Microsoft.OData.Json
         /// <summary>
         /// Writes an escaped string to the buffer.
         /// </summary>
-        /// <param name="writer">The text writer to write the output to.</param>
+        /// <param name="writer">The text writer to write the output.</param>
+
         /// <param name="inputString">Input string value.</param>
         /// <param name="currentIndex">The index in the string at which copying should begin.</param>
         /// <param name="buffer">Char buffer to use for streaming data.</param>
--- a/src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
@@ -499,7 +499,6 @@ namespace Microsoft.OData.Json
         /// Writes an escaped string to the buffer.
         /// </summary>
         /// <param name="writer">The text writer to write the output.</param>
-
         /// <param name="inputString">Input string value.</param>
         /// <param name="currentIndex">The index in the string at which copying should begin.</param>
         /// <param name="buffer">Char buffer to use for streaming data.</param>
--- a/src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
@@ -412,13 +412,13 @@ namespace Microsoft.OData.Json
                 {
                     innerBuffer.Value = BufferUtils.InitializeBufferIfRequired(innerBufferPool, innerBuffer.Value);
                     int bufferLength = innerBuffer.Value.Length;
-                    int bufferIndex = 0;
-                    int currentIndex = 0;
+                    Ref<int> bufferIndex = new Ref<int>(0);
+                    Ref<int> currentIndex = new Ref<int>(0);
 
                     // Let's copy and flush strings up to the first index of the special char
-                    while (currentIndex < innerFirstIndex)
+                    while (currentIndex.Value < innerFirstIndex)
                     {
-                        int substrLength = innerFirstIndex - currentIndex;
+                        int substrLength = innerFirstIndex - currentIndex.Value;
 
                         Debug.Assert(substrLength > 0, "SubStrLength should be greater than 0 always");
 
@@ -427,23 +427,24 @@ namespace Microsoft.OData.Json
                         // Otherwise copy to the buffer and go on from there.
                         if (substrLength >= bufferLength)
                         {
-                            innerInputString.CopyTo(currentIndex, innerBuffer.Value, 0, bufferLength);
+                            innerInputString.CopyTo(currentIndex.Value, innerBuffer.Value, 0, bufferLength);
                             await innerWriter.WriteAsync(innerBuffer.Value, 0, bufferLength).ConfigureAwait(false);
-                            currentIndex += bufferLength;
+                            currentIndex.Value += bufferLength;
                         }
                         else
                         {
-                            WriteSubstringToBuffer(innerInputString, ref currentIndex, innerBuffer.Value, ref bufferIndex, substrLength);
+                            WriteSubstringToBuffer(innerInputString, currentIndex, innerBuffer.Value, bufferIndex, substrLength);
                         }
                     }
 
                     // Write escaped string to buffer
-                    WriteEscapedStringToBuffer(innerWriter, innerInputString, ref currentIndex, innerBuffer.Value, ref bufferIndex, innerStringEscapeOption);
+                    await WriteEscapedStringToBufferAsync(innerWriter, innerInputString, currentIndex, innerBuffer.Value, bufferIndex, innerStringEscapeOption)
+                        .ConfigureAwait(false);
 
                     // write any remaining chars to the writer
-                    if (bufferIndex > 0)
+                    if (bufferIndex.Value > 0)
                     {
-                        await innerWriter.WriteAsync(innerBuffer.Value, 0, bufferIndex).ConfigureAwait(false);
+                        await innerWriter.WriteAsync(innerBuffer.Value, 0, bufferIndex.Value).ConfigureAwait(false);
                     }
                 }
             }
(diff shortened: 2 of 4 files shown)


src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -787,6 +787,25 @@ private static void WriteSubstringToBuffer(string inputString, ref int currentIn
...
+        /// <param name="currentIndex">The index in the string at which the substring begins.</param>
+        /// <param name="buffer">Char buffer to use for streaming data.</param>
+        /// <param name="bufferIndex">Current position in the buffer after the substring has been written.</param>
+        /// <param name="substrLength">The length of the substring to be copied.</param>
+        private static void WriteSubstringToBuffer(string inputString, Ref<int> currentIndex, char[] buffer, Ref<int> bufferIndex, int substrLength)
+        {
+            Debug.Assert(inputString != null, "inputString != null");
+            Debug.Assert(buffer != null, "buffer != null");
+
+            inputString.CopyTo(currentIndex.Value, buffer, 0, substrLength);
```
xuzhg, line 804: Is it safe to copyto the buffer with less size?

habbes: The amount to copy is determine by the `substrLength`. The caller of this method ensures that the `substrLength` is not greater than the buffer length: https://github.com/OData/odata.net/pull/2707/files#diff-c86855fdf1be1c1e0a820b8852ccc1fef825001c9ef74274af40778f7adfb96dR452.

Should we add an additional `Debug.Assert` in this method to check whether this precondition is violated?

Note that this method is a duplicate of the following existing method: https://github.com/OData/odata.net/pull/2707/files#diff-c86855fdf1be1c1e0a820b8852ccc1fef825001c9ef74274af40778f7adfb96dR780. The only change is that I replaced the `ref` parameters with `Ref<int>` so that we can use it in async methods.

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -490,5 +494,135 @@ private static Task WriteQuotedAsync(this TextWriter writer, string text)
             return writer.WriteAsync(
                 string.Concat(JsonConstants.QuoteCharacter, text, JsonConstants.QuoteCharacter));
         }
+
+        /// <summary>
+        /// Writes an escaped string to the buffer.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output to.</param>
```
ElizabethOkerio, line 501: NIT: ```suggestion
        /// <param name="writer">The text writer to write the output.</param>

```

Comment (ElizabethOkerio) [inline src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs:501]:
```suggestion
        /// <param name="writer">The text writer to write the output.</param>

```

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -490,5 +494,135 @@ private static Task WriteQuotedAsync(this TextWriter writer, string text)
...
+            if (escapedString == null)
+            {
+                buffer[bufferIndex] = character;
+                bufferIndex++;
+            }
+            else
+            {
+                // Okay, an unhandled character was detected.
+                // First lets check if we can fit it in the existing buffer, if not,
+                // flush the current buffer and reset. Add the escaped string to the buffer
```
ElizabethOkerio, line 569: how thread safe is the modification of the buffer and bufferIndex? Can this method be called concurrently by multiple tasks?

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -490,5 +494,135 @@ private static Task WriteQuotedAsync(this TextWriter writer, string text)
...
+        /// caller is responsible for writing the final buffer contents to the writer.
+        /// </remarks>
+        private static async Task<int> EscapeAndWriteCharToBufferAsync(TextWriter writer, char character, char[] buffer, int bufferIndex, ODataStringEscapeOption stringEscapeOption)
+        {
+            int bufferLength = buffer.Length;
+            string escapedString = null;
+
+            if (stringEscapeOption == ODataStringEscapeOption.EscapeNonAscii || character <= 0x7F)
+            {
+                escapedString = JsonValueUtils.SpecialCharToEscapedStringMap[character];
```
ElizabethOkerio, line 555: Is it possible that a character is not found in this map? but it meets the 2 conditions above?

habbes: The map is an array of size `char.MaxValue + 1`. Therefore, every possible `char` (unsigned 16-bit integer) is a valid index of that array.

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
habbes: In this case, the bufferIndex is passed by value, so we're only modifying the local copy. The `buffer` is a reference and so technically there's danger if it's modified by multiple threads at the same time, but that doesn't happen. If you follow the flow of execution, each time the `buffer` is passed to an async method chain, we await it before passing it to another method, so we never really have concurrent writes the same buffer.

ElizabethOkerio (dismissed):

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -490,5 +494,136 @@ private static Task WriteQuotedAsync(this TextWriter writer, string text)
             return writer.WriteAsync(
                 string.Concat(JsonConstants.QuoteCharacter, text, JsonConstants.QuoteCharacter));
         }
+
+        /// <summary>
+        /// Writes an escaped string to the buffer.
+        /// </summary>
+        /// <param name="writer">The text writer to write the output.</param>
+
```
xuzhg, line 502: remove unused whiteline

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
```diff
@@ -490,5 +494,136 @@ private static Task WriteQuotedAsync(this TextWriter writer, string text)
...
+        /// <param name="writer">The text writer to write the output.</param>
+
+        /// <param name="inputString">Input string value.</param>
+        /// <param name="currentIndex">The index in the string at which copying should begin.</param>
+        /// <param name="buffer">Char buffer to use for streaming data.</param>
+        /// <param name="bufferIndex">Current position in the buffer after the string has been written.</param>
+        /// <param name="stringEscapeOption">The string escape option.</param>
+        /// <remarks>
+        /// IMPORTANT: After all characters have been written,
+        /// caller is responsible for writing the final buffer contents to the writer.
```
xuzhg, line 510: how can we guarantee it?

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsAsyncTests.cs
```diff
@@ -298,29 +328,30 @@ public async Task WriteEmptyByteShouldWork()
...
             this.buffer = new Ref<char[]>();
-            this.stream = new MemoryStream();
+            this.stream = new AsyncStream(new MemoryStream());
             StreamWriter innerWriter = new StreamWriter(this.stream);
-            innerWriter.AutoFlush = true;
             this.writer = new NonIndentedTextWriter(innerWriter);
         }
 
-        private string StreamToString()
+        private async Task<string> StreamToString()
```
xuzhg, line 349: is it better to add a suffix "Async" for the async method name?

xuzhg (dismissed):

src/Microsoft.OData.Core/Json/JsonValueUtilsAsync.cs
habbes: It's implemented by the `WriteEscapedJsonStringValueAsync` method: https://github.com/OData/odata.net/pull/2707/files#diff-44795ee4fde32bffe0ac82d8b21b2ac93180517dce7d66dfeb7afce151a1fb8bR447

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonValueUtilsAsyncTests.cs
habbes: Sure

ElizabethOkerio (approved):

PR #2729, opened 2023-09-04
Add support for NaN and infinity to ODataUtf8JsonWriter

### LINKED ISSUE #2728: `ODataUtf8JsonWriter` does not support `NaN` and infinity values
`ODataUtf8JsonWriter` throws an `ArgumentException` when writing `double.NaN`, `double.PositiveInfinity`, `double.NegativeInfinity` (and the `float` equivalents as well). These values are supported by the default `JsonWriter` and are handled as follows:

- `double.NaN` -> `"NaN"`
- `double.PositiveInfinity` -> `"INF"`
- `double.PositiveInfinity` -> `"-INF"`

The `ODataUt8fJsonWriter` should provide similar support for these values.

### Assemblies affected

`Microsoft.OData.Core`  7.17.0

### Reproduce steps

Create a payload that has a numeric property with `double.NaN` as the value and serialize the payload using OData configured to use [`DefaultStreamBasedJsonWriter`](https://review.learn.microsoft.com/en-us/odata/odatalib/using-utf8jsonwriter-for-better-performance?branch=pr-en-us-303).

Or add the following test case to the `ODataUtf8JsonWriterTests` class:

```c#
Fact]
public void WritePrimitiveValueDoubleNaN()
{
    this.VerifyWritePrimitiveValue(double.NaN, "\"NaN\"");
}
```

### Expected result

The corresponding string representation of `NaN` should be written for the value, and the test should pass.

### Actual result

An `ArgumentException` error is thrown with the message:

```
.NET number values such as positive and negative infinity cannot be written as valid JSON. To make it work when using 'JsonSerializer', consider specifying 'JsonNumberHandling.AllowNamedFloatingPointLiterals' (see https://docs.microsoft.com/dotnet/api/system.text.json.serialization.jsonnumberhandling).
```

### Additional detail

In .NET 6 and above, `JsonSerializer.Serialize` options has a [`NumberHandling`](https://learn.microsoft.com/en-us/dotnet/api/system.text.json.jsonserializeroptions.numberhandling) property which when set to [`JsonNumberHandling.AllowNamedFloatingPointLiterals`](https://docs.microsoft.com/en-us/dotnet/api/system.text.json.serialization.jsonnumberhandling) will serialize these values as strings. However, we cannot rely on this feature for the following reasons:
- It serializes `double.PositiveInfinity` as `"Infinity"` and `double.NegativeInfinity` as `"-Infinity"` whereas our default `JsonWriter` uses `"INF"` and `"-INF"` respectively
- This option exists for the high-level `JsonSerializerOptions` but does not exist on the lower-level `JsonWriterOptions` which we use with the `Utf8JsonWriter`

habbes:
### Issues

*This pull request fixes #2728 *

### Description

This PR adds support for `double.NaN`, `double.PositiveInfinity` and `double.NegativeInfinity` by serializing them as the strings `"NaN"`, `"INF"` and `"-INF"` respectively, to match the behaviour of the default `JsonWriter`.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
@@ -29,6 +29,9 @@ namespace Microsoft.OData.Json
         private readonly float bufferFlushThreshold;
         private readonly static ReadOnlyMemory<byte> parentheses = new byte[] { (byte)'(', (byte)')' };
         private readonly static ReadOnlyMemory<byte> itemSeparator = new byte[] { (byte)',' };
+        private readonly static ReadOnlyMemory<byte> nanValue = new byte[] { (byte)'N', (byte)'a', (byte)'N' };
+        private readonly static ReadOnlyMemory<byte> positiveInfinityValue = new byte[] { (byte)'I', (byte)'N', (byte)'F' };
+        private readonly static ReadOnlyMemory<byte> negativeInfinityValue = new byte[] { (byte)'-', (byte)'I', (byte)'N', (byte)'F' };
 
         private readonly Stream outputStream;
         private readonly Stream writeStream;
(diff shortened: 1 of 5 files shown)


syprieur (approved):

gathogojr (approved):
:shipit:

ElizabethOkerio (approved):

KenitoInc (approved):

PR #2755, opened 2023-10-02
Use same format as JsonWriter when serializing DateTimeOffset with ODataUtf8JsonWriter

habbes:
### Issues

The default `JsonWriter` and `Utf8JsonWriter`, while both use ISO8601 format to write `DateTimeOffset`, they differ in the specific case when the time zone offset is 0.

`JsonWriter` uses `Z` to indicate the offset: `2022-11-09T09:42:30Z`
`Utf8JsonWriter` uses `+00:00`, i.e.: `2022-11-09T09:42:30+00:00`

This PR changes `ODataUtf8JsonWriter` to make it consistent with `JsonWriter` when emitting 0-offset DateTimeOffset values.

### Description

In the `WriteValue(DateTimeOffset value)` method, we check if the offset is 0. If the offset is 0, then it uses the DateTimeOffset string formatting utility in `JsonValueUtils` to convert it to a string first.

You may be wondering, why not just convert all `DateTimeOffset` values to string and not just those with 0 offsets? Well it turns out that if you convert the value to string, `Utf8JsonWriter` will escape the `+` offset with `\u002B`, i.e. `2022-11-09T09:42:30\u002B00:00`.

It may also be worse for perf.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
@@ -208,7 +208,18 @@ namespace Microsoft.OData.Json
         internal static void WriteValue(TextWriter writer, DateTimeOffset value, ODataJsonDateTimeFormat dateTimeFormat)
         {
             Debug.Assert(writer != null, "writer != null");
+            string textValue = JsonValueUtils.FormatDateTimeOffset(value, dateTimeFormat);
+            JsonValueUtils.WriteQuoted(writer, textValue);
+        }
 
+        /// <summary>
+        /// Formats the DateTimeOffset value as a string based on the specified format.
+        /// </summary>
+        /// <param name="value">DateTimeOffset value to be formatted.</param>
+        /// <param name="dateTimeFormat">The format to use for conversion.</param>
+        /// <returns>The string representation of the DateTimeOffset based on the given format.</returns>
+        internal static string FormatDateTimeOffset(DateTimeOffset value, ODataJsonDateTimeFormat dateTimeFormat)
+        {
             switch (dateTimeFormat)
             {
                 case ODataJsonDateTimeFormat.ISO8601DateTime:
@@ -220,11 +231,8 @@ namespace Microsoft.OData.Json
                         //  quotation-mark
                         //
                         // offset = 4DIGIT
-                        string textValue = XmlConvert.ToString(value);
-                        JsonValueUtils.WriteQuoted(writer, textValue);
+                        return XmlConvert.ToString(value);
                     }
-
-                    break;
 
                 case ODataJsonDateTimeFormat.ODataDateTime:
                     {
@@ -238,12 +246,13 @@ namespace Microsoft.OData.Json
                         //
                         // ticks = *DIGIT
                         // offset = 4DIGIT
-                        string textValue = FormatDateTimeAsJsonTicksString(value);
-                        JsonValueUtils.WriteQuoted(writer, textValue);
+                        return FormatDateTimeAsJsonTicksString(value);
                     }
-
-                    break;
             }
+
+            // NOTE: This statement will not be reached because
+            // we exhausted all the options in the enum
+            return dateTimeFormat.ToString();
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
+++ b/src/Microsoft.OData.Core/Json/JsonValueUtils.cs
@@ -220,30 +220,36 @@ namespace Microsoft.OData.Json
         /// <returns>The string representation of the DateTimeOffset based on the given format.</returns>
         internal static string FormatDateTimeOffset(DateTimeOffset value, ODataJsonDateTimeFormat dateTimeFormat)
         {
-            if (dateTimeFormat == ODataJsonDateTimeFormat.ISO8601DateTime)
+            switch (dateTimeFormat)
             {
-                // Uses the same format as DateTime but with offset:
-                // jsonDateTime= quotation-mark
-                //  YYYY-MM-DDThh:mm:ss.sTZD
-                //  [("+" / "-") offset]
-                //  quotation-mark
-                //
-                // offset = 4DIGIT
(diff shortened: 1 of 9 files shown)


habbes:
Let's not merge this PR yet. Since both date formats are valid, it may be counterproductive to force `Utf8JsonWriter` to use `JsonWriter`'s representation at the expense of performance.

habbes:
A couple of internal teams have expressed that the difference in formatting could be a high risk for breaking changes. We have also seen past issues raised related to different date formats between `JsonSerializer` and ODL's serializer. It is safe to say that for some workloads, the difference in formatting for the 0-offset date does matter even if both formats are valid according to the spec. For that reason, I think we should move forward with this PR.

e-haq:
There is also an issue with DateOnly field that gets serialized differently. It makes sense to also fix that. Ut8JsonWriter returns something like "Foo": { "Year"=10, "Month" = 1, "Day"=1} whereas JsonWriter returns the format "yyyy'-'MM'-'dd". We are unable to use Utf8 writer in a high-volume service due to both these issues.

src/Microsoft.OData.Core/Json/JsonValueUtils.cs
```diff
@@ -238,11 +246,10 @@ internal static void WriteValue(TextWriter writer, DateTimeOffset value, ODataJs
...
                         // ticks = *DIGIT
                         // offset = 4DIGIT
-                        string textValue = FormatDateTimeAsJsonTicksString(value);
-                        JsonValueUtils.WriteQuoted(writer, textValue);
+                        return FormatDateTimeAsJsonTicksString(value);
                     }
-
-                    break;
+                default:
+                    throw new ODataException(Strings.ODataJsonWriter_UnsupportedDateTimeFormat);
```
ElizabethOkerio, line 252: Is this a breaking change?

ElizabethOkerio: Or it's an internal method..

ElizabethOkerio: Could you add a test when this exception gets thrown

habbes: The thing is that code is unreachable because `ODataJsonDateTimeFormat` enum only has two members and both of those enums are accounted for by the `case` statements. In the original code, there was no `default` block. The difference is that in this version I'm returning a value and the compiler requires each code path to have a return statement, so I needed `default` block to in order to compile. And since there's nothing to return in this scenario, I just throw an exception. So if we ever add another member to that enum and not handle it, an exception will be thrown, which I believe is the right thing to do.

ElizabethOkerio (dismissed):

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -308,7 +308,21 @@ public void WriteValue(decimal value)
         public void WriteValue(DateTimeOffset value)
         {
             this.WriteSeparatorIfNecessary();
-            this.utf8JsonWriter.WriteStringValue(value);
+            if (value.Offset == TimeSpan.Zero)
+            {
+                // The default JsonWriter uses the format 2022-11-09T09:42:30Z when the offset is 0
```
gathogojr, line 313: nit:
```suggestion
                // The default JsonWriter uses the format yyyy-MM-ddThh:mm:ssZ when the offset is 0
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -308,7 +308,21 @@ public void WriteValue(decimal value)
         public void WriteValue(DateTimeOffset value)
         {
             this.WriteSeparatorIfNecessary();
-            this.utf8JsonWriter.WriteStringValue(value);
+            if (value.Offset == TimeSpan.Zero)
+            {
+                // The default JsonWriter uses the format 2022-11-09T09:42:30Z when the offset is 0
+                // Utf8JsonWriter uses the format 2022-11-09T09:42:30+00:00
```
gathogojr, line 314: nit:
```suggestion
                // Utf8JsonWriter uses the format yyyy-MM-ddThh:mm:ss+00:00
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -725,7 +739,21 @@ public async Task WriteValueAsync(decimal value)
         public async Task WriteValueAsync(DateTimeOffset value)
         {
             this.WriteSeparatorIfNecessary();
-            this.utf8JsonWriter.WriteStringValue(value);
+            if (value.Offset == TimeSpan.Zero)
+            {
+                // The default JsonWriter uses the format 2022-11-09T09:42:30Z when the offset is 0
+                // Utf8JsonWriter uses the format 2022-11-09T09:42:30+00:00
```
gathogojr, line 745: See https://github.com/OData/odata.net/pull/2755/files#r1395292179

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -725,7 +739,21 @@ public async Task WriteValueAsync(decimal value)
         public async Task WriteValueAsync(DateTimeOffset value)
         {
             this.WriteSeparatorIfNecessary();
-            this.utf8JsonWriter.WriteStringValue(value);
+            if (value.Offset == TimeSpan.Zero)
+            {
+                // The default JsonWriter uses the format 2022-11-09T09:42:30Z when the offset is 0
```
gathogojr, line 744: See https://github.com/OData/odata.net/pull/2755/files#r1395291348

gathogojr (approved):
LGTM

gathogojr (approved):

ElizabethOkerio (approved):

PR #2779, opened 2023-10-26
Fix issue in parsing OData core vocabulary

### LINKED ISSUE #2246: Microsoft.OData.Edm 7.9.4  can't parse standard core vocabulary
reading https://raw.githubusercontent.com/oasis-tcs/odata-vocabularies/main/vocabularies/Org.OData.Core.V1.xml via 

`CsdlReader.TryParse(reader, out var model, out var errors)` 

returns false and an error 
    
`UnexpectedXmlAttribute : The attribute 'MaxLength' was not expected in the given context. : (523, 75)`

Library version Microsoft.OData.Edm"  7.9.4

gathogojr:
### Issues

*This pull request fixes #2246.*

### Description

When trying to parse the OData core vocabulary CSDL (https://raw.githubusercontent.com/oasis-tcs/odata-vocabularies/main/vocabularies/Org.OData.Core.V1.xml) using the `CsdlReader`, an exception with the following message is thrown:
```
UnexpectedXmlAttribute : The attribute 'MaxLength' was not expected in the given context. : (523, 75)
```

The error in parsing the core vocabulary (Org.OData.Core.V1.xml) is caused by `SimpleIdentifier` type definition that has a `MaxLength` property:
```xml
<TypeDefinition Name="SimpleIdentifier" UnderlyingType="Edm.String" MaxLength="128">
    <Annotation Term="Core.Description" String="A [simple identifier](https://docs.oasis-open.org/odata/odata-csdl-xml/v4.01/odata-csdl-xml-v4.01.html#sec_SimpleIdentifier)" />
    <Annotation Term="Validation.Pattern" String="^[\p{L}\p{Nl}_][\p{L}\p{Nl}\p{Nd}\p{Mn}\p{Mc}\p{Pc}\p{Cf}]{0,}$" />
</TypeDefinition>
```
The Edm library currently doesn't support parsing of facets (`MaxLength`, `Unicode`, `Precision`, `Scale` and `SRID`) on type definitions. The library however supports parsing of facets specified where the type definition is used, e.g.:
```xml
<Term Name="SymbolicName" Type="Core.SimpleIdentifier" MaxLength="128" Nullable="false">
	<Annotation Term="Core.Description" String="A symbolic name for a model element" />
</Term>
```

This pull request updates the OData vocabularies since they had diverged from those on [the OASIS TCS repository](https://github.com/oasis-tcs/odata-vocabularies/blob/main/vocabularies).

Existing tests have also been updated to verify the updated vocabularies.
The `MaxLength` facet on the `SimpleIdentifier` has been intentionally stripped so parsing does break pending the redesign of the library to handle parsing of facets on type definition - to be tracked separately.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Edm/Vocabularies/AuthorizationVocabularies.xml
+++ b/src/Microsoft.OData.Edm/Vocabularies/AuthorizationVocabularies.xml
@@ -20,7 +20,7 @@
   - OData Measures Vocabulary. Latest version: https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Measures.V1.xml.
   - OData Capabilities Vocabulary. Latest version: https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Capabilities.V1.xml.
   - OData Validation Vocabulary. Latest version: https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Validation.V1.xml.
-  - OData Aggregation Vocabulary. Latest version: https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Aggretation.V1.xml.
+  - OData Aggregation Vocabulary. Latest version: https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Aggregation.V1.xml.
   - OData Authorization Vocabulary. Latest version: https://oasis-tcs.github.io/odata-vocabularies/vocabularies/Org.OData.Authorization.V1.xml.
 
   Related work:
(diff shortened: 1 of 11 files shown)


test/FunctionalTests/Microsoft.OData.Edm.Tests/Vocabularies/Org.OData.Core.V1.xml
xuzhg, line 1: In the Microsoft.OData.Edm project, there's a xml added for odata.core vocabulary. Should We add this again? Maintainance them two places?

test/FunctionalTests/Microsoft.OData.Edm.Tests/Microsoft.OData.Edm.Tests.csproj
```diff
@@ -28,12 +28,16 @@
     <DelaySign>false</DelaySign>
   </PropertyGroup>
 
-  <Import Project="..\Build.props"/>
+  <Import Project="..\Build.props" />
```
xuzhg, line 31: revert this change

test/FunctionalTests/Microsoft.OData.Edm.Tests/Vocabularies/Org.OData.Core.V1.xml
gathogojr: @xuzhg I'm aware of the file and I thought I could use it but as it turns out, it's not exactly the same. That file doesn't have the `edmx:Edmx` root element and the `CsdlReader` cannot parse it for that reason. The error that Christof reported is observed when parsing the https://raw.githubusercontent.com/oasis-tcs/odata-vocabularies/main/vocabularies/Org.OData.Core.V1.xml using the `CsdlReader` class.
In the Edm library, the `SchemaReader` class is what is used to parse the [existing schema file](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Edm/Vocabularies/CoreVocabularies.xml) since it's root element is `Schema`.

xuzhg: Can we prepend and append the `<edmx:Edmx ...` after reading the xml and use CsdlReader?

xuzhg (approved):

habbes (approved):

src/Microsoft.OData.Edm/Vocabularies/AuthorizationVocabularies.xml
```diff
@@ -115,39 +111,39 @@
         </Property>
         <Property Name="RefreshUrl" Type="Edm.String" Nullable="true">
           <Annotation Term="Core.Description" String="Refresh Url" />
-          <Annotation Term="Core.IsURL" Bool="true"/>
```
xuzhg, line 118: >Bool="true" [](http://example.com/codeflow?start=40&length=11)

I remember I fixed the ODL to accept omitting 'Bool="true" for tag. Can you double confirm that?

xuzhg (approved):
:shipit:

habbes (approved):

src/Microsoft.OData.Edm/Vocabularies/AuthorizationVocabularies.xml
gathogojr: @xuzhg We read the value optionally https://github.com/OData/odata.net/blob/d2108632a71a780998c25f05e536826972c1eb61/src/Microsoft.OData.Edm/Csdl/Parsing/CsdlDocumentParserBase.cs#L617

PR #2787, opened 2023-11-06
Fix delta deleted entry deserialization failure observed when reason appears before id in payload

### LINKED ISSUE #2604: lost id value if put "id" after "reason" for deleted resource in 4.0 format

### Assemblies affected

### Reproduce steps

If I use the following payload, it can read "id" value and "reason" value.

```json
{
 
   "Departs@delta": [
    {
     "@odata.context":"http://localhost:5000/v1/$metadata#Departments/$deletedEntity",
     "id":"Departments(13)",
     "reason":"deleted"
    }
  ]
} 
```

However, if I change the order of "id" and "reason", ODL can't read the "Id" value.

```json
{
 
   "Departs@delta": [
    {
     "@odata.context":"http://localhost:5000/v1/$metadata#Departments/$deletedEntity",
     "reason":"deleted",
     "id":"Departments(13)"
    }
  ]
} 
```

### Expected result

* The property order doesn't matter.

### Actual result

*The property order does matter.

gathogojr:
### Issues

*This pull request fixes #2604.*

### Description

This pull request fixes deserialization failure observed when `reason` property appears before `id` property in a delta deleted entry (4.0) payload, e.g.:
```json
{"reason":"deleted","id":"http://tempuri.org/Customers(1)"}
```
Fixed by modifying the logic in `ODataJsonLightResourceDeserializer`'s `ReadDeletedEntry` and `ReadDeletedEntryAsync` methods to read the properties in whichever order they appear.

- [x] *Test cases added*

### Additional work necessary


xuzhg (approved):
:shipit:

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -225,42 +225,48 @@ internal ODataDeletedResource ReadDeletedEntry()
             Uri id = null;
             DeltaDeletedEntryReason reason = DeltaDeletedEntryReason.Changed;
 
-            // If the current node is the id property - read it.
-            if (this.JsonReader.NodeType == JsonNodeType.Property &&
-                string.Equals(JsonLightConstants.ODataIdPropertyName, this.JsonReader.GetPropertyName(), StringComparison.Ordinal))
+            do
```
xuzhg, line 228: Typically, for property reading, we are using  `while () {}` loop.
I am not fans for 'while()' loop. But, If we can keep it consistent, it could be better.

xuzhg (dismissed):

habbes:
I honestly find it funny that we have such a bug.

habbes (dismissed):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
gathogojr: @xuzhg I replaced the `do {} while` into a `while {}` loop. I figure it especially made sense if the JSON was malformed

xuzhg (approved):

KenitoInc (approved):

PR #2792, opened 2023-11-15
Reduce buffer allocations

### LINKED ISSUE #2790: Significant byte array allocations from ODL
A customer has raised concerns about the huge of amount of byte[] array allocations from ODL. On one internal service, they account for 15% of allocations, in another service they account for 5% of allocations.

These allocations are from buffers used by ODL's writer:
- BufferedStream allocates 84K byte buffers
- TextWriter (used by JsonWriter) also allocates byte arrays
- ODataUtf8JsonWriter also has an internal 16k buffer

### Assemblies affected

Microsoft.OData.Core 7.x

### Reproduce steps

Open the `SerializationComparisonBenchmarks` solution, open the .NET allocations profiler in Visual Studio and run the TestServer in release mode.

Then use the `bombardier` load testing tool to run 100 requests using 20 connections:
`bombardier -l -c 20 -n 101 https://localhost:7120/customers/ODataMessageWriter-Async`

Then analyse the profiler report.

In my case, I got something like the following:

image

- ~10M bytes worth of byte[] arrays, most of them from ODL:
- The `byte[84k]` are from `BufferedStream`
- The `byte[3k]` are from `TextWriter`

And here are the results when using `ODataUtf8JsonWriter`:

`bombardier -l -c 20 -n 101 https://localhost:7120/customers/ODataMessageWriter-Utf8JsonWriter-Async`

image

- ~12M bytes worth of byte[] array allocations
- `byte[84k]` are from `BufferedStream`
- `byte[16k]` are from `ODataUtf8JsonWriter.

### Additional detail

Here are things we can do to reduce allocations:
- reduce `BufferedStream` buffer size (if it doesn't negatively impact latency/perf)
- when using `ODataUtf8JsonWriter`, do not use `BufferedStream`
- when using `ODataUtf8JsonWriter` use array-pooling (`PooledByteBufferWriter`) to create buffers instead of using `ArrayBufferWriter`

habbes:
### Issues

*This pull request fixes #2790 .*

### Description

This PR reduces allocations of buffer used by ODL's internal JSON writer:
- When using `JsonWriter`:
  - made the buffer size of `BufferedStream` configurable using `ODataMessageWriterSettings.BufferSize`
- When using the `ODataUtf8JsonWriter`:
  - we do not wrap the stream in a `BufferedStream` since `ODataUtf8JsonWriter` handles its own buffering
  - replaces the `ArrayBufferWriter` used by `ODataUtf8JsonWriter` with a custom `PooledByteBufferWriter` that rents buffers from the Array pool

The `PooledByteBufferWriter` was adapted from .NET's implementation: https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/Text/Json/PooledByteBufferWriter.cs

**Update**: Instead of hardcoding the default size to a lower value, I made it configurable by adding a `BufferSize` property to `ODataMessageWriterSettings`. However, this setting is only used when creating an asynchronous `JsonWriter`. It is not used when creating an instance of `ODataUtf8JsonWriter`. The reason is because `IStreamBasedJsonWriterFactory.CreateJsonWriter` unfortunately does not expose a parameter for passing the buffer size.

I also made changes to the `SerializationComparisonTests` benchmarks to ensure that the writers are disposed after the payload has been written. This required refactoring the `JsonWriterBenchmarks` such that a new stream is used for each write call. The benchmarks can't re-use the stream cause the stream gets disposed. The changes make the benchmarks arguably cleaner even though each benchmark call is doing a bit more work than just writing the payload. This means that a comparison of benchmark results between this PR and the master branch is no longer an apples to apples comparison.

 When using `JsonWriter`, I had initially reduced the `BufferedStream` buffer size from 84k to 16k but I observed mixed results from the benchmarks, no strong evidence of improvement, and in some cases there was a regression. As a result I retained the original 84k buffer size but made it configurable via `ODataMessageWriterSettings.BufferSize`

Generally, I didn't see a lot of conclusive improvement in latency. But it doesn't seem to make latency worse. If it improves memory, without degrading latency, I think that's a good enough improvement to have.

### Benchmarks

I used the `TestServer` project in `SerializationComparisonsTests` to compare the heap allocations and latency/throughput before and after the changes.

#### Memory allocations

I set the count to 2000 to get the payload size above 400KB, which is a relatively large payload size, but matches a sample payload size of one of the customers.

Command used for testing:
-  Warm up request `bombardier -l -c 1 -n 1 https://localhost:7120/customers/ODataMessageWriter-Async`
- Concurrent requests: `bombardier -l -c 10 -n 50 https://localhost:7120/customers/ODataMessageWriter-Async`

**`JsonWriter` Before**

image

image

**`JsonWriter` After**

image

image

**`Utf8JsonWriter` Before**

image

image

**`Utf8JsonWriter` After**

image

image

#### Request metrics

Command used for testing:

- Sequential request latency: `bombardier -l -c 1 -n 100 https://localhost:7120/customers/ODataMessageWriter-Async?count=2000`
- Concurrent requests: `bombardier -l -c 50 -d 1m https://localhost:7120/customers/ODataMessageWriter-Async?count=2000`
- High load (timeouts expected): `Utf8JsonWriter`: `bombardier -l -c 100 -d 1m https://localhost:7120/customers/ODataMessageWriter-Async?

**`JsonWriter` before**

- Sequential requests:
  Reqs/sec        12.76      18.91      82.50
  Latency       76.91ms    45.08ms   406.04ms
  Latency Distribution
     50%    64.69ms
     75%    66.31ms
     90%    75.04ms
     95%   174.68ms
     99%   220.88ms
  HTTP codes:
    1xx - 0, 2xx - 100, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:     7.43MB/s

- 50 connections concurrent requests:

  Reqs/sec        63.07     145.61    3849.11
  Latency         0.94s    98.78ms      1.67s
  Latency Distribution
     50%      0.94s
     75%      1.00s
     90%      1.06s
     95%      1.11s
     99%      1.33s
  HTTP codes:
    1xx - 0, 2xx - 3193, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    30.32MB/s

- 100 connections

  Reqs/sec        49.50     117.08    1467.94
  Latency         2.28s      1.88s     13.83s
  Latency Distribution
     50%      1.75s
     75%      1.87s
     90%      4.13s
     95%      6.53s
     99%     12.67s
  HTTP codes:
    1xx - 0, 2xx - 2639, 3xx - 0, 4xx - 0, 5xx - 0
    others - 31

**`JsonWriter` after**

- Sequential requests:

  Reqs/sec        12.68      19.03      88.51
  Latency       78.80ms    40.66ms   426.61ms
  Latency Distribution
     50%    68.24ms
     75%    71.52ms
     90%    81.92ms
     95%   140.85ms
     99%   160.52ms
  HTTP codes:
    1xx - 0, 2xx - 100, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:     7.29MB/s

- 50 connections, concurrent requests:

  Reqs/sec        66.39     139.43    3841.72
  Latency         0.90s    84.35ms      1.28s
  Latency Distribution
     50%      0.90s
     75%      0.95s
     90%      1.00s
     95%      1.04s
     99%      1.13s
  HTTP codes:
    1xx - 0, 2xx - 3361, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    31.99MB/s

- 100 connections

  Reqs/sec        53.45     179.59    5419.08
  Latency         2.36s      2.02s     13.55s
  Latency Distribution
     50%      1.75s
     75%      1.89s
     90%      4.22s
     95%      6.94s
     99%     12.67s
  HTTP codes:
    1xx - 0, 2xx - 2546, 3xx - 0, 4xx - 0, 5xx - 0
    others - 34

**`Utf8JsonWriter` before**
- Sequential requests

  Reqs/sec        15.01      19.06      63.33
  Latency       66.21ms    34.34ms   345.19ms
  Latency Distribution
     50%    56.89ms
     75%    59.79ms
     90%    70.13ms
     95%   109.54ms
     99%   159.28ms
  HTTP codes:
    1xx - 0, 2xx - 100, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:     8.89MB/s

- 50 connections

  Reqs/sec        76.40     160.26    3675.45
  Latency      818.42ms    96.34ms      1.23s
  Latency Distribution
     50%   807.87ms
     75%      0.88s
     90%      0.95s
     95%      1.01s
     99%      1.10s
  HTTP codes:
    1xx - 0, 2xx - 3681, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    35.96MB/s

- 100 connections

  Reqs/sec        71.17     131.48    2669.99
  Latency         1.65s   391.15ms     11.73s
  Latency Distribution
     50%      1.61s
     75%      1.70s
     90%      1.82s
     95%      1.89s
     99%      4.12s
  HTTP codes:
    1xx - 0, 2xx - 3663, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    36.03MB/s

**`Utf8JsonWriter` after**

- Sequential requests

  Reqs/sec        17.12      62.12    1135.20
  Latency       70.11ms    36.60ms   347.11ms
  Latency Distribution
     50%    57.99ms
     75%    61.92ms
     90%   107.97ms
     95%   138.66ms
     99%   156.07ms
  HTTP codes:
    1xx - 0, 2xx - 100, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:     8.40MB/s

- 50 connections

  Reqs/sec        75.81     147.23    3844.68
  Latency      792.53ms    80.07ms      1.33s
  Latency Distribution
     50%   790.36ms
     75%   845.36ms
     90%      0.90s
     95%      0.93s
     99%      1.00s
  HTTP codes:
    1xx - 0, 2xx - 3802, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    37.13MB/s

-100 connections

  Reqs/sec        73.47     209.24    5402.75
  Latency         1.71s   544.48ms      9.21s
  Latency Distribution
     50%      1.63s
     75%      1.74s
     90%      1.86s
     95%      1.97s
     99%      4.21s
  HTTP codes:
    1xx - 0, 2xx - 3533, 3xx - 0, 4xx - 0, 5xx - 0
    others - 1
  Errors:
    tls handshake timed out - 1
  Throughput:    35.44MB/s

#### Writer benchmarks

**Before**
|           Method |                                           WriterName |       Mean |      Error |     StdDev |     Median |      Gen 0 |     Gen 1 |  Allocated |
|----------------- |----------------------------------------------------- |-----------:|-----------:|-----------:|-----------:|-----------:|----------:|-----------:|
| WriteToFileAsync |                                   ODataMessageWriter | 380.980 ms |  7.4531 ms |  6.6070 ms | 379.283 ms | 53000.0000 |         - | 224,003 KB |
| WriteToFileAsync |                             ODataMessageWriter-Async | 684.025 ms |  9.7998 ms |  9.1667 ms | 681.923 ms | 63000.0000 | 1000.0000 | 266,767 KB |
| WriteToFileAsync |                    ODataMessageWriter-Utf8JsonWriter | 353.918 ms |  6.9733 ms |  5.8231 ms | 354.351 ms | 51000.0000 | 1000.0000 | 217,429 KB |
| WriteToFileAsync |              ODataMessageWriter-Utf8JsonWriter-Async | 620.257 ms | 12.2123 ms | 15.8794 ms | 613.840 ms | 51000.0000 | 1000.0000 | 218,772 KB |

**After**

|           Method |                                           WriterName |       Mean |      Error |     StdDev |     Median |      Gen 0 |     Gen 1 |  Allocated |
|----------------- |----------------------------------------------------- |-----------:|-----------:|-----------:|-----------:|-----------:|----------:|-----------:|
| WriteToFileAsync |                                   ODataMessageWriter | 399.640 ms |  7.9563 ms |  9.4714 ms | 396.437 ms | 53000.0000 |         - | 224,018 KB |
| WriteToFileAsync |                             ODataMessageWriter-Async | 704.245 ms | 13.5573 ms | 14.5062 ms | 698.330 ms | 63000.0000 | 1000.0000 | 266,773 KB |
| WriteToFileAsync |                    ODataMessageWriter-Utf8JsonWriter | 378.397 ms |  3.2468 ms |  2.8782 ms | 378.249 ms | 51000.0000 |         - | 217,344 KB |
| WriteToFileAsync |              ODataMessageWriter-Utf8JsonWriter-Async | 627.557 ms | 12.4976 ms | 20.8807 ms | 623.245 ms | 51000.0000 |         - | 218,138 KB |

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
+++ b/src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
@@ -143,9 +143,6 @@ namespace Microsoft.OData.Json
 
         public Span<byte> GetSpan(int sizeHint = MinimumBufferSize)
         {
-            // TODO: this condition is here cause without it the sizeHint is getting set to 0
-            // instead of the MinimumBufferSize when the method is called without params.
-            // I don't understand why this is happening, will investigate later
             if (sizeHint == 0)
             {
                 sizeHint = MinimumBufferSize;
--- a/src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
+++ b/src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
@@ -11,6 +11,7 @@ using System.Diagnostics;
 
 namespace Microsoft.OData.Json
 {
+    // This class has been adapated from https://github.com/dotnet/runtime/blob/main/src/libraries/Common/src/System/Text/Json/PooledByteBufferWriter.cs
     /// <summary>
     /// An implementation of <see cref="IBufferWriter{byte}"/> that rents
     /// buffers from an array pool instead of allocating a new array every time.
@@ -21,12 +22,12 @@ namespace Microsoft.OData.Json
         // it can be used as an IBufferWriter and holds a buffer that should eventually be
         // returned to the shared pool. If rentedBuffer is null, then the instance is in a
         // cleared/disposed state and it must re-rent a buffer before it can be used again.
-        private byte[]? _rentedBuffer;
-        private int _index;
+        private byte[] rentedBuffer;
+        private int index;
 
         private const int MinimumBufferSize = 256;
 
-        // Value copied from Array.MaxLength in System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Array.cs.
+        // Value copied from Array.MaxLength in https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Array.cs
         public const int MaximumBufferSize = 0X7FFFFFC7;
 
         private PooledByteBufferWriter()
@@ -42,8 +43,8 @@ namespace Microsoft.OData.Json
         {
             Debug.Assert(initialCapacity > 0);
 
-            _rentedBuffer = ArrayPool<byte>.Shared.Rent(initialCapacity);
-            _index = 0;
+            this.rentedBuffer = ArrayPool<byte>.Shared.Rent(initialCapacity);
+            this.index = 0;
         }
 
         /// <summary>
@@ -54,9 +55,9 @@ namespace Microsoft.OData.Json
         {
             get
             {
-                Debug.Assert(_rentedBuffer != null);
-                Debug.Assert(_index <= _rentedBuffer.Length);
-                return _rentedBuffer.AsMemory(0, _index);
+                Debug.Assert(this.rentedBuffer != null);
+                Debug.Assert(this.index <= this.rentedBuffer.Length);
+                return this.rentedBuffer.AsMemory(0, this.index);
             }
         }
 
@@ -67,8 +68,8 @@ namespace Microsoft.OData.Json
         {
             get
             {
-                Debug.Assert(_rentedBuffer != null);
-                return _index;
+                Debug.Assert(rentedBuffer != null);
+                return this.index;
             }
         }
 
@@ -79,8 +80,8 @@ namespace Microsoft.OData.Json
         {
             get
             {
-                Debug.Assert(_rentedBuffer != null);
-                return _rentedBuffer.Length;
+                Debug.Assert(this.rentedBuffer != null);
+                return this.rentedBuffer.Length;
             }
         }
 
@@ -92,8 +93,8 @@ namespace Microsoft.OData.Json
         {
             get
             {
-                Debug.Assert(_rentedBuffer != null);
-                return _rentedBuffer.Length - _index;
+                Debug.Assert(this.rentedBuffer != null);
+                return this.rentedBuffer.Length - this.index;
             }
         }
 
@@ -103,10 +104,10 @@ namespace Microsoft.OData.Json
         /// <param name="count"></param>
         public void Advance(int count)
         {
-            Debug.Assert(_rentedBuffer != null);
+            Debug.Assert(this.rentedBuffer != null);
             Debug.Assert(count >= 0);
-            Debug.Assert(_index <= _rentedBuffer.Length - count);
-            _index += count;
+            Debug.Assert(this.index <= this.rentedBuffer.Length - count);
+            this.index += count;
         }
 
         /// <summary>
@@ -117,7 +118,7 @@ namespace Microsoft.OData.Json
         public Memory<byte> GetMemory(int sizeHint = MinimumBufferSize)
         {
             CheckAndResizeBuffer(sizeHint);
-            return _rentedBuffer.AsMemory(_index);
+            return this.rentedBuffer.AsMemory(index);
         }
 
         /// <summary>
@@ -133,7 +134,7 @@ namespace Microsoft.OData.Json
             }
 
             CheckAndResizeBuffer(sizeHint);
-            return _rentedBuffer.AsSpan(_index);
+            return this.rentedBuffer.AsSpan(index);
         }
 
         /// <summary>
@@ -149,37 +150,37 @@ namespace Microsoft.OData.Json
         /// </summary>
         public void Dispose()
         {
-            if (_rentedBuffer == null)
+            if (this.rentedBuffer == null)
             {
                 return;
             }
 
             ClearHelper();
-            byte[] toReturn = _rentedBuffer;
-            _rentedBuffer = null;
+            byte[] toReturn = this.rentedBuffer;
+            this.rentedBuffer = null;
             ArrayPool<byte>.Shared.Return(toReturn);
         }
 
         private void ClearHelper()
         {
-            Debug.Assert(_rentedBuffer != null);
-            Debug.Assert(_index <= _rentedBuffer.Length);
+            Debug.Assert(this.rentedBuffer != null);
+            Debug.Assert(this.index <= this.rentedBuffer.Length);
 
-            _rentedBuffer.AsSpan(0, _index).Clear();
-            _index = 0;
+            this.rentedBuffer.AsSpan(0, index).Clear();
+            this.index = 0;
         }
 
         private void CheckAndResizeBuffer(int sizeHint)
         {
-            Debug.Assert(_rentedBuffer != null);
+            Debug.Assert(rentedBuffer != null);
             Debug.Assert(sizeHint > 0);
 
-            int currentLength = _rentedBuffer.Length;
-            int availableSpace = currentLength - _index;
+            int currentLength = this.rentedBuffer.Length;
+            int availableSpace = currentLength - this.index;
 
             // If we've reached ~1GB written, grow to the maximum buffer
             // length to avoid incessant minimal growths causing perf issues.
-            if (_index >= MaximumBufferSize / 2)
+            if (this.index >= MaximumBufferSize / 2)
             {
                 sizeHint = Math.Max(sizeHint, MaximumBufferSize - currentLength);
             }
@@ -199,21 +200,21 @@ namespace Microsoft.OData.Json
                     }
                 }
 
-                byte[] oldBuffer = _rentedBuffer;
+                byte[] oldBuffer = this.rentedBuffer;
 
-                _rentedBuffer = ArrayPool<byte>.Shared.Rent(newSize);
+                this.rentedBuffer = ArrayPool<byte>.Shared.Rent(newSize);
 
-                Debug.Assert(oldBuffer.Length >= _index);
-                Debug.Assert(_rentedBuffer.Length >= _index);
+                Debug.Assert(oldBuffer.Length >= this.index);
+                Debug.Assert(this.rentedBuffer.Length >= this.index);
 
-                Span<byte> oldBufferAsSpan = oldBuffer.AsSpan(0, _index);
-                oldBufferAsSpan.CopyTo(_rentedBuffer);
+                Span<byte> oldBufferAsSpan = oldBuffer.AsSpan(0, this.index);
+                oldBufferAsSpan.CopyTo(this.rentedBuffer);
                 oldBufferAsSpan.Clear();
                 ArrayPool<byte>.Shared.Return(oldBuffer);
             }
 
-            Debug.Assert(_rentedBuffer.Length - _index > 0);
-            Debug.Assert(_rentedBuffer.Length - _index >= sizeHint);
+            Debug.Assert(this.rentedBuffer.Length - this.index > 0);
+            Debug.Assert(this.rentedBuffer.Length - this.index >= sizeHint);
         }
     }
 }
--- a/src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
+++ b/src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
@@ -28,18 +28,14 @@ namespace Microsoft.OData.Json
         private const int MinimumBufferSize = 256;
 
         // Value copied from Array.MaxLength in https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Array.cs
-        public const int MaximumBufferSize = 0X7FFFFFC7;
-
-        private PooledByteBufferWriter()
-        {
-        }
+        public const int MaximumBufferSize = 0X7FFFFFC7; // 2147483591 (approx 2GB)
 
         /// <summary>
         /// Creates an instance of <see cref="PooledByteBufferWriter"/> to which data can be written,
         /// with a specified initial capacity.
         /// </summary>
         /// <param name="initialCapacity"></param>
-        public PooledByteBufferWriter(int initialCapacity) : this()
+        public PooledByteBufferWriter(int initialCapacity)
         {
             Debug.Assert(initialCapacity > 0);
 
@@ -98,10 +94,7 @@ namespace Microsoft.OData.Json
             }
         }
 
-        /// <summary>
         /// <inheritdoc/>
-        /// </summary>
-        /// <param name="count"></param>
         public void Advance(int count)
         {
             Debug.Assert(this.rentedBuffer != null);
@@ -110,22 +103,14 @@ namespace Microsoft.OData.Json
             this.index += count;
         }
 
-        /// <summary>
         /// <inheritdoc/>
-        /// </summary>
-        /// <param name="sizeHint"></param>
-        /// <returns></returns>
         public Memory<byte> GetMemory(int sizeHint = MinimumBufferSize)
         {
             CheckAndResizeBuffer(sizeHint);
             return this.rentedBuffer.AsMemory(index);
         }
 
-        /// <summary>
         /// <inheritdoc/>
-        /// </summary>
-        /// <param name="sizeHint"></param>
-        /// <returns></returns>
         public Span<byte> GetSpan(int sizeHint = MinimumBufferSize)
         {
             if (sizeHint == 0)
@@ -142,7 +127,7 @@ namespace Microsoft.OData.Json
         /// </summary>
         public void Clear()
         {
-            ClearHelper();
+            ClearInternal();
         }
 
         /// <summary>
@@ -155,13 +140,13 @@ namespace Microsoft.OData.Json
                 return;
             }
 
-            ClearHelper();
+            ClearInternal();
             byte[] toReturn = this.rentedBuffer;
             this.rentedBuffer = null;
             ArrayPool<byte>.Shared.Return(toReturn);
         }
 
-        private void ClearHelper()
+        private void ClearInternal()
         {
             Debug.Assert(this.rentedBuffer != null);
             Debug.Assert(this.index <= this.rentedBuffer.Length);
@@ -187,13 +172,23 @@ namespace Microsoft.OData.Json
 
             if (sizeHint > availableSpace)
             {
+                // If the current size buffer of buffer is relatively small,
+                // we'll grow it by the size hint requested by the caller
+                // otherwise we'll attempt to double the buffer size.
+                // This avoids repeatedly growing the buffer by small
+                // chunks.
                 int growBy = Math.Max(sizeHint, currentLength);
 
                 int newSize = currentLength + growBy;
 
                 if ((uint)newSize > MaximumBufferSize)
                 {
+                    // If the new size exceeds by the maximum buffer size,
+                    // then we fall back and attempt to grow it only by
+                    // the size requested by the caller.
                     newSize = currentLength + sizeHint;
+
+                    // If the size still exceeds the max size, then give up and throw an OOM.
                     if ((uint)newSize > MaximumBufferSize)
                     {
                         throw new OutOfMemoryException(Strings.ODataMessageWriter_Buffer_Maximum_Size_Exceeded(newSize));
--- a/src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
+++ b/src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
@@ -1,0 +1,215 @@
+﻿#if NETCOREAPP3_1_OR_GREATER
+using Microsoft.OData;
+using System.Buffers;
+using System.Diagnostics;
+using System.Diagnostics.CodeAnalysis;
+using System.IO;
+using System.Runtime.CompilerServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.Text.Json
+{
+    /// <summary>
+    /// An implementation of <see cref="IBufferWriter{byte}"/> that rents
+    /// buffers from an array pool instead of allocating a new array every time.
+    /// </summary>
+    internal sealed class PooledByteBufferWriter : IBufferWriter<byte>, IDisposable
+    {
+        // This class allows two possible configurations: if rentedBuffer is not null then
+        // it can be used as an IBufferWriter and holds a buffer that should eventually be
+        // returned to the shared pool. If rentedBuffer is null, then the instance is in a
+        // cleared/disposed state and it must re-rent a buffer before it can be used again.
+        private byte[]? _rentedBuffer;
+        private int _index;
+
+        private const int MinimumBufferSize = 256;
+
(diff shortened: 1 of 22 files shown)


habbes:
Based on the results of the benchmarks, I'm reconsidering whether it's a good idea to reduce the buffer size of `BufferedStream` for the default `JsonWriter`.

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
```diff
@@ -0,0 +1,221 @@
...
+            CheckAndResizeBuffer(sizeHint);
+            return this.rentedBuffer.AsSpan(index);
+        }
+
+        /// <summary>
+        /// Clears the data written to the underlying buffer.
+        /// </summary>
+        public void Clear()
+        {
+            ClearHelper();
```
gathogojr, line 145: The `ClearHelper` method is 4 statements long. Why can't those lines be here?

src/Microsoft.OData.Core/ODataConstants.cs
```diff
@@ -218,7 +218,7 @@ internal static class ODataInternalConstants
         /// <summary>
         /// The default size used for buffering output data.
         /// </summary>
-        internal const int DefaultOutputBufferSize = 84000;
+        internal const int DefaultOutputBufferSize = 16 * 1024;
```
gathogojr, line 221: @habbes Do we have any data to suggest that 16,384 (16*1024) is the sweet spot compared to 84,000? For example, I'd be curious if the higher number is more ideal in scenarios where payloads are much bigger...

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        /// <summary>
+        /// <inheritdoc/>
+        /// </summary>
+        /// <param name="sizeHint"></param>
+        /// <returns></returns>
+        public Span<byte> GetSpan(int sizeHint = MinimumBufferSize)
+        {
+            if (sizeHint == 0)
```
gathogojr, line 131: You can add a comment here that the `IBufferWriter` interface' `GetSpan` method declaration sets a default `sizeHint` parameter value of 0 and overriding the default value in the implementation is currently not supported.

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
```diff
@@ -0,0 +1,221 @@
...
+        // it can be used as an IBufferWriter and holds a buffer that should eventually be
+        // returned to the shared pool. If rentedBuffer is null, then the instance is in a
+        // cleared/disposed state and it must re-rent a buffer before it can be used again.
+        private byte[] rentedBuffer;
+        private int index;
+
+        private const int MinimumBufferSize = 256;
+
+        // Value copied from Array.MaxLength in https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/Array.cs
+        public const int MaximumBufferSize = 0X7FFFFFC7;
```
gathogojr, line 31:
```suggestion
        public const int MaximumBufferSize = 0X7FFFFFC7; // 2147483591
```

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
```diff
@@ -0,0 +1,221 @@
...
+
+        private PooledByteBufferWriter()
```
gathogojr, line 33: Why do we need this constructor? A private constructor is used to prevent the automatic generation of a public parameterless constructor but in your case there's already a public constructor that takes a parameter.

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
```diff
@@ -0,0 +1,221 @@
...
+            {
+                Debug.Assert(this.rentedBuffer != null);
+                return this.rentedBuffer.Length - this.index;
+            }
+        }
+
+        /// <summary>
+        /// <inheritdoc/>
+        /// </summary>
+        /// <param name="count"></param>
```
gathogojr, line 104:
```suggestion
        /// <inheritdoc/>
```

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
```diff
@@ -0,0 +1,221 @@
...
+            Debug.Assert(count >= 0);
+            Debug.Assert(this.index <= this.rentedBuffer.Length - count);
+            this.index += count;
+        }
+
+        /// <summary>
+        /// <inheritdoc/>
+        /// </summary>
+        /// <param name="sizeHint"></param>
+        /// <returns></returns>
```
gathogojr, line 117:
```suggestion
        /// <inheritdoc/>
```

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
```diff
@@ -0,0 +1,221 @@
...
+        public Memory<byte> GetMemory(int sizeHint = MinimumBufferSize)
+        {
+            CheckAndResizeBuffer(sizeHint);
+            return this.rentedBuffer.AsMemory(index);
```
gathogojr, line 121: I assume this method returns the segment of the array that hasn't been written to...

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
```diff
@@ -0,0 +1,221 @@
...
+        }
+
+        /// <summary>
+        /// <inheritdoc/>
+        /// </summary>
+        /// <param name="sizeHint"></param>
+        /// <returns></returns>
```
gathogojr, line 128:
```suggestion
        /// <inheritdoc/>
```

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
```diff
@@ -0,0 +1,221 @@
...
+                return;
+            }
+
+            ClearHelper();
+            byte[] toReturn = this.rentedBuffer;
+            this.rentedBuffer = null;
+            ArrayPool<byte>.Shared.Return(toReturn);
+        }
+
+        private void ClearHelper()
```
gathogojr, line 164:
```suggestion
        private void ClearInternal()
```
Or `ClearImplementation()`

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
```diff
@@ -0,0 +1,221 @@
...
+        {
+            Debug.Assert(this.rentedBuffer != null);
+            Debug.Assert(this.index <= this.rentedBuffer.Length);
+
+            this.rentedBuffer.AsSpan(0, index).Clear();
```
gathogojr, line 169: Curious if the following would be more efficient:
```
Array.Clear(this.rentedBuffer, 0, index);
````

File: src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs [hunk 12]
```diff
@@ -0,0 +1,221 @@
...
+            if (sizeHint > availableSpace)
+            {
+                int growBy = Math.Max(sizeHint, currentLength);
+
+                int newSize = currentLength + growBy;
+
+                if ((uint)newSize > MaximumBufferSize)
+                {
+                    newSize = currentLength + sizeHint;
+                    if ((uint)newSize > MaximumBufferSize)
```
gathogojr, line 197: I don't think I understand what is happening here. Maybe you could add some comments like you did in the `if` block before this...

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/PooledByteBufferWriterTests.cs
```diff
@@ -0,0 +1,187 @@
...
+using System.Threading.Tasks;
+using Microsoft.OData.Json;
+using Xunit;
+
+namespace Microsoft.OData.Tests.Json
+{
+    public class PooledByteBufferWriterTests
+    {
+        [Fact]
+        public void ContstructorInitializesBufferToInititalCapacity()
```
gathogojr, line 22:
```suggestion
        public void ConstructorInitializesBufferToInititalCapacity()
```

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/PooledByteBufferWriterTests.cs
```diff
@@ -0,0 +1,187 @@
...
+            Assert.Equal(10, data.Length);
+            for (int i = 0; i < data.Length; i++)
+            {
+                Assert.Equal(i + 1, data[i]);
+            }
+        }
+    }
+}
+
+#endif
```
gathogojr, line 187: One test that you do not have, and might not be safe to have, is one where you hit the 1GB mark and grow the array to `MaximumBufferSize`. I don't know if it'd be possible to use reflection to modify the `MaximumBufferSize` to a much lower value (in a test) to verify that the relevant logic works fine.

gathogojr: https://github.com/OData/odata.net/pull/2792/files#r1396784482
This would also help you write a test to verify the behavior (exception thrown) when the `MaximumBufferSize` limit is breached.

test/PerformanceTests/SerializationComparisonsTests/JsonWriterBenchmarks/Benchmarks.cs
```diff
@@ -40,26 +39,24 @@ public Benchmarks()
...
-            fileStream.Close();
-            fileStream = null;
+            File.Delete(filePath);
         }
 
         [Benchmark]
         [BenchmarkCategory("InMemory")]
         public async Task WriteToMemoryAsync()
         {
+            using var memoryStream = new MemoryStream();
```
gathogojr, line 59: Shouldn't this happen in `IterationSetup` method?

test/PerformanceTests/SerializationComparisonsTests/JsonWriterBenchmarks/Benchmarks.cs
```diff
@@ -68,23 +65,29 @@ public async Task WriteToMemoryAsync()
...
+            await WritePayloadAsync(includeRawValues: true);
+            await WritePayloadAsync(includeRawValues: true);
+            await WritePayloadAsync(includeRawValues: true);
+            await WritePayloadAsync(includeRawValues: true);
+            await WritePayloadAsync(includeRawValues: true);
+        }
+
+        private async Task WritePayloadAsync(bool includeRawValues = false)
+        {
+            using var outputStream = new FileStream(filePath, FileMode.Create, FileAccess.Write, share: FileShare.ReadWrite);
```
gathogojr, line 89: Initializing the file stream from here won't skew the results?

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
habbes: Good point. It's not needed. Must be some artifact of referenced code.

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
habbes: Yes. The next position to write to is tracked by the `index`.

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
habbes: Good question. Both achieve the same thing, filling the segment with 0's. But peeking at the implementation, [`Array.Clear`](https://source.dot.net/#System.Private.CoreLib/src/System/Array.CoreCLR.cs,141) seems to do more work than [`Span<T>.Clear`](https://source.dot.net/#System.Private.CoreLib/src/libraries/System.Private.CoreLib/src/System/Span.cs,285). Both of them call the same `SpanHelpers.ClearWithoutReferences()` method (after checking whether the element type has references or not). But `Array.Clear` appears to do a lot more work before calling that method, including retrieving the method table and checking the array's dimentionality.

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
```diff
@@ -0,0 +1,216 @@
...
+        /// <summary>
+        /// Creates an instance of <see cref="PooledByteBufferWriter"/> to which data can be written,
+        /// with a specified initial capacity.
+        /// </summary>
+        /// <param name="initialCapacity"></param>
+        public PooledByteBufferWriter(int initialCapacity)
+        {
+            Debug.Assert(initialCapacity > 0);
+
+            this.rentedBuffer = ArrayPool<byte>.Shared.Rent(initialCapacity);
```
xuzhg, line 42: Shall we re-use 'https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/Buffers/ICharArrayPool.cs'?

xuzhg (approved):

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
habbes: Alternatively I could set the default value of the parameter to 0 to match the interface, then keep the `if` statement

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
habbes: `ICharArrayPool` is a custom interface and we don't provide a default implementation. Also, `ICharArrayPool` only works for `char` arrays whereas in this case we're renting `byte` arrays. I think we should remove `ICharArrayPool` in the next major version and only rely on the standard .NET `ArrayPool<T>`.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/PooledByteBufferWriterTests.cs
habbes: I have added a test to verify the exception is thrown when the max buffer size is reached. But I didn't add a test to show that the buffer grows to the max size when the max size/2 threshold is reached. I can't use reflection to change the max buffer size because it's a `const`, i.e. set at compile time, and not a `readonly static` field. So I would actually have to allocate a 1GB+ array to test this and I'm not sure that's safe to do in a test and whether it'll run successfully on the CI environment,

test/PerformanceTests/SerializationComparisonsTests/JsonWriterBenchmarks/Benchmarks.cs
habbes: It may skew the results, but since it's execute for every writer, I think it still makes for a fair comparison. And the relative runtimes and allocations of the writers in the results did not change (as compared to each other), so I think the results are not less reliable than they used to be. But now the tests are cleaner (the previous version of the tests did not delete the temporary file)

test/PerformanceTests/SerializationComparisonsTests/JsonWriterBenchmarks/Benchmarks.cs
habbes: Yeah, it could also happen in `IterationSetup`. I thought doing it here woud help clean up the code. Also the cost of creating and disposing a memory stream should be relatively small since it doesn't involve any actual I/O. But the reason I moved it out of the `IterationSetup` is because the docs warn against using the `IterationSetup` for "microbenchmarks": https://benchmarkdotnet.org/articles/features/setup-and-cleanup.html

src/Microsoft.OData.Core/Json/PooledByteBufferWriter.cs
habbes: I added comments.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -125,6 +118,22 @@ internal sealed class ODataJsonLightOutputContext : ODataOutputContext
...
+                    if (this.Synchronous)
+                    {
+                        outputStream = this.messageOutputStream;
+                    }
+                    else
+                    {
+#if NETSTANDARD1_1
+                        this.asynchronousOutputStream = new AsyncBufferedStream(this.messageOutputStream);
+#else
+                    this.asynchronousOutputStream = new BufferedStream(this.messageOutputStream, this.MessageWriterSettings.BufferSize);
```
gathogojr, line 131: Fix indentation

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
```diff
@@ -143,6 +144,12 @@ public Uri BaseUri
         /// </summary>
         public ICharArrayPool ArrayPool { get; set; }
 
+        /// <summary>
+        /// Gets/sets the size of the buffer used to buffer writes to the output
+        /// stream. Note that this is a hint and may override it when deemed fit.
+        /// </summary>
+        public int BufferSize { get; set; }
```
gathogojr, line 151:
```suggestion
        /// <summary>
        /// Gets or sets the size of the buffer used to buffer writes to the output
        /// stream. Note that this is a hint and may be disregarded where deemed fit.
        /// </summary>
        public int BufferSize { get; set; }
```

src/Microsoft.OData.Core/ODataConstants.cs
habbes: I restored the default buffer size to 84,000 but added a `BufferSize` property to the `ODataMessageWriterSettings` to configure this value.

gathogojr (approved):
LGTM

ElizabethOkerio (approved):

PR #2811, opened 2023-11-29
Add index-based access to ODataPath

### LINKED ISSUE #2804: Perf: Allow index-based random access to segments in `ODataPath`
`ODataPath` implements `IEnumerable<ODataPathSegment>` and also exposes the following properties: `Count`, `FirstSegment`, `LastSegment`. In addition, it exposes a visitor pattern via the `HandleWith` method and `PathHandler` instances. This provides a standard pattern for creating many path-based operations. However, I think we can make some path-based operations more efficient, especially those in hot paths, by making it possible to index segments directly.

One such example is the `GetNavigationSource(ODataPath)` extension method in AspNetCoreOData:
https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Routing/ODataPathExtensions.cs#L73

image

The cost of this method involves an allocation of `ODataPathNavigationSourceHandler`, and calling the `WalkWith` method which is arguably overkill for what could be achieved with a plan loop starting from the end.

There are other path segment operations that we could optimize in a similar fashion where the handler pattern may result in additional cost worth saving.

While we're at it, we ought also to refactor the `ODataPath.LastSegment` and `ODataPath.FirstSegment` properties to use index-based lookup instead of Linq's `LastOrDefault()`:

image

### Assemblies affected

`Microsoft.OData.Core` 7.x

Also indirectly affects `Microsoft.AspNetCore.OData` 8.x

### Reproduce steps

Run an OData service that serializes a payload through a CPU profiler and look at the metrics. For this example, I used the [BenchmarkServer](https://github.com/OData/AspNetCoreOData/tree/main/sample/BenchmarkServer) in AspNetCoreOData.

habbes:
### Issues

*This pull request fixes #2804*

### Description

This PR adds an indexer to `ODataPath` to make random-access of segments in the path possible. This will help us simplify and optimize operations where we rely on Linq, `ToList()` or `PathHandler`s in hot paths.

With this PR, `ODataPath` would statisfy all the requirements of an `IReadOnlyList`, in that case, would it be harmful to explicitly implement that interface (even if in a future release)?

- [x] *Test cases added*

### Additional work necessary

Once (if) this PR is merged, we can go ahead and make improvements to some path operations both in ODL and AspNetCoreOData (e.g. `path.NavigationSource()` extension method in AspNetCoreOData)

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
+++ b/src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
@@ -77,7 +77,7 @@ namespace Microsoft.OData.UriParser
         {
             get
             {
-                return this.segments.Count > 0 ? this.segments[0] : default;
+                return this.segments.Count > 0 ? this.segments[0] : null;
             }
         }
 
@@ -88,7 +88,7 @@ namespace Microsoft.OData.UriParser
         {
             get
             {
-                return this.segments.Count > 0 ? this.segments[this.segments.Count - 1] : default;
+                return this.segments.Count > 0 ? this.segments[this.segments.Count - 1] : null;
             }
         }
 
--- a/src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
+++ b/src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
@@ -93,6 +93,12 @@ namespace Microsoft.OData.UriParser
         }
 
         /// <summary>
+        /// Get the segment at the specified index.
+        /// </summary>
+        /// <param name="index">The index of the segment to retrieve</param>
+        public ODataPathSegment this[int index] => this.segments[index];
+
+        /// <summary>
         /// Get the number of segments in this path.
         /// </summary>
         public int Count
--- a/src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
+++ b/src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
@@ -77,7 +77,7 @@ namespace Microsoft.OData.UriParser
         {
             get
             {
-                return this.segments.FirstOrDefault();
+                return this.segments.Count > 0 ? this.segments[0] : default;
             }
         }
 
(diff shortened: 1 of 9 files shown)


src/Microsoft.OData.Core/UriParser/SemanticAst/ODataPath.cs
```diff
@@ -77,7 +77,7 @@ public ODataPathSegment FirstSegment
         {
             get
             {
-                return this.segments.FirstOrDefault();
```
xuzhg, line 80: >FirstOrDefault [](http://example.com/codeflow?start=37&length=14)

From the .NET Source code, if it's IList<T>, so, it returns the first one in the List directly?

private static TSource? TryGetFirst<TSource>(this IEnumerable<TSource> source, out bool found)
        {
            if (source == null)
            {
                ThrowHelper.ThrowArgumentNullException(ExceptionArgument.source);
            }
 
            if (source is IPartition<TSource> partition)
            {
                return partition.TryGetFirst(out found);
            }
 
            if (source is IList<TSource> list)
            {
                if (list.Count > 0)
                {
                    found = true;
                    return list[0];
                }
            }
            else
            {
                using (IEnumerator<TSource> e = source.GetEnumerator())
                {
                    if (e.MoveNext())
                    {
                        found = true;
                        return e.Current;
                    }
                }
            }
 
            found = false;
            return default;
        }


The above has a bad perf?

habbes: The .NET does that optimization for `FirstOrDefault` and `LastOrDefault`, however if I we already know this it's a list, then we can do it directly and avoid the extra layers of indirection that `FirstOrDefault` does, for example we avoid the casts (`source is IPartition<TSource> and `source is IList<TSource>`) and the extra method call. From the attached issue, you can see some CPU usage from `Enumerable.TryGetLast` that I think is worth optimizing if it's a low-hanging fruit.

xuzhg (approved):

gathogojr (approved):

PR #2812, opened 2023-11-30
Add flag to skip property verification in `ODataResource`

### LINKED ISSUE #2803: Perf: Can we skip properties verification in `ODataResourceBase`?
[ODataResourceBase.Properties setter](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/ODataResource.cs#L193) looks as follows:

```c#
            set
            {
                VerifyProperties(value);
                this.properties = value;
            }
```

That `VerifyProperties` methods goes through all properties and items in collection properties to ensure there are no nested `ODataResource` values, i.e. the properties in a `ODataResource` should either be primitive values or collections of primitive values.

During serialization, this verification is relatively expensive since it's called for every resource and nested resource:
image

I propose that we add a flag that allows the caller to skip the verification. This would be beneficial to callers like AspNetCoreOData where we're sure that properties are generated correctly, e.g.:

```
var resource = new ODataResource()
{
   SkipPropertyVerification = true,
   Properties = someProperties
}
```

### Assemblies affected

Microsoft.OData.Core 7.x

### Reproduce steps

Run a service that serializes ODataResources and look at CPU profiles. For this example I used the `BenchmarkServer` in AspNetCoreOData.

### Expected result

`ODataResourceBase.VerifyProperties` should have 0 or negligeable CPU time.

### Actual result

`ODataResourceBase.VerifyProperties` takes up 0.61% of CPU (can be higher if more properties are serialized).

habbes:
### Issues

*This pull request fixes #2803*

### Description

This PR adds a `SkipPropertyVerification` property to `ODataResource` which skips property verification. By default, `ODataResource` will verify the property values when the `Properties` property is set. The aim of the of the verification is to ensure that properties `ODataResourceValue`s are not included in the properties. I.e. the property values should either primitive types or collections of primitive types (`ODataResourceValue` is intended for use in annotations).

However, in scenarios like AspNetCoreOData, where the library already creates `ODataResource` that already match this expectation, it's an unnecessary cost to still go ahead and perform this verification. This flag can be used in AspNetCoreOData to avoid paying this cost for every resource that is created.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/ODataResource.cs
+++ b/src/Microsoft.OData.Core/ODataResource.cs
@@ -176,6 +176,15 @@ namespace Microsoft.OData
             get { return this.MetadataBuilder.GetFunctions(); }
         }
 
+        /// <summary>
+        /// Gets or sets whether to skip property verification. When set to false (default behaviour),
+        /// the properties collection will be verified to ensure it doesn't contain invalid values.
+        /// When set to true, it's the caller's responsibility to ensure the property values are valid
+        /// before setting the <see cref="Properties"/> property. This can be a useful optimization
+        /// in hot paths when you're sure property values are valid.
+        /// </summary>
+        public bool SkipPropertyVerification { get; set; }
+
         /// <summary>Gets or sets the resource properties.</summary>
         /// <returns>The resource properties.</returns>
         /// <remarks>
@@ -190,7 +199,11 @@ namespace Microsoft.OData
 
             set
             {
-                VerifyProperties(value);
+                if (!this.SkipPropertyVerification)
+                {
+                    VerifyProperties(value);
+                }
+
                 this.properties = value;
             }
         }
--- a/src/Microsoft.OData.Core/PublicAPI/net45/PublicAPI.Unshipped.txt
+++ b/src/Microsoft.OData.Core/PublicAPI/net45/PublicAPI.Unshipped.txt
@@ -1,0 +1,2 @@
+Microsoft.OData.ODataResourceBase.SkipPropertyVerification.get -> bool
+Microsoft.OData.ODataResourceBase.SkipPropertyVerification.set -> void
--- a/src/Microsoft.OData.Core/PublicAPI/netcoreapp3.1/PublicAPI.Unshipped.txt
+++ b/src/Microsoft.OData.Core/PublicAPI/netcoreapp3.1/PublicAPI.Unshipped.txt
@@ -1,0 +1,2 @@
+Microsoft.OData.ODataResourceBase.SkipPropertyVerification.get -> bool
+Microsoft.OData.ODataResourceBase.SkipPropertyVerification.set -> void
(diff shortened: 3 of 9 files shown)


src/Microsoft.OData.Core/ODataResource.cs
```diff
@@ -176,6 +176,15 @@ public IEnumerable<ODataFunction> Functions
...
         }
 
+        /// <summary>
+        /// Gets or sets whether to skip property verification. When set to false (default behaviour),
+        /// the properties collection will be verified to ensure it doesn't contain invalid values.
+        /// When set to true, it's the caller's responsibility to ensure the property values are valid
+        /// before setting the <see cref="Properties"/> property. This can be a useful optimization
+        /// in hot paths when you're sure property values are valid.
+        /// </summary>
+        public bool SkipPropertyVerification { get; set; }
```
gathogojr, line 186: Wouldn't this be better as a global setting (e.g. `ODataMessageWriterSettings`/`ODataMessageReaderSettings`) instead of a property that has to be set on each `ODataResource` instance?

habbes: Problem is `ODataResource` doesn't have access to the message writer settings.

habbes: Maybe the serializer that creates ODataResources could use the `ODataMessageWritter` settings to decide whether to set the skip verification flag, but in that case it would still have to set it for every resource.

gathogojr: "it would still have to set it for every resource" - Why would it need to be set for every resource?

A different question, by which means did you envision that the `SkipPropertyVerification` would be set?

KenitoInc: I agree with @gathogojr 
Can we use the same setup as the `IDuplicatePropertyNameChecker` where the serializer have access to the `ODataMessageWriteSettings` and this set in the `WriterValidator`

habbes: @gathogojr 

    I don't think it's any of the ODataResource methods that calls methods for verifying or writing the properties. It's the serializer that does so... If

In the existing code, it is actually the `ODataResource` that calls the method for verifying properties, it happens inside the `ODataResourceBase.Properties` setter. That's why I applied the flag to the `ODataResource`.

Each `ODataResource` instance is independent, and the `ODataMessageWriter` operates on a resource-per-resource basis. The `ODataMessageWriter` does not know about the serializer.

@KenitoInc does have access to the the `ODataMessageWriterSettings` but the `ODataResource` does not. Currently the verification happens within the `ODataResource` itself, not the serializer nor the message writer.

We could change things up such that the verification is moved out of the `ODataResource` and moved into the message writer. Or we could remove the verification altogether and let the serializer handle that manually if it deems fit. I'm not sure if this delayed verification would have unintended consequences in those where verification is desired.

gathogojr (approved):

ElizabethOkerio (approved):

Issue #2827, opened 2023-12-11
ODataPathParser fails valid keys with parenthesis


ificator:
Making a request like `/files/file%20(1).txt` (where the first segment is an entityset and the second is a key) results in a failure with a "syntax error" stated as the cause.

### Assemblies affected

Microsoft.OData.Core

### Reproduce steps

Make a request like `GET /files/file%20(1).txt`

### Expected result

The action is invoked with a key of `file (1).txt`

### Actual result

The request fails

### Additional detail

The issue is that `ExtractSegmentIdentifierAndParenthesisExpression` asserts that if the string contains a `(` it MUST end with `)`, but this is not a valid assumption given that the output of this method is used to determine what kind of segment it is with one option being a key.

habbes:
@ificator I tried to reproduce this in an AspNetCoreOData 8 project and didn't get an error, the key was correctly extracted as `"files (1).txt"`. I am looking to see how this is achieved in AspNetCoreOData 8. According to @xuzhg, the library encloses the key segment in single quotes.

Are you using ODL directly? If so, how are you parsing the path?

That said, we'll review your PR.

ificator:
We're using netframework and not netcore - let me confirm a repro with the lastest aspnet libraries.

PR #2828, opened 2023-12-11
Do not fail if parenthesis are not in expected state

ificator:
### Issues
*This pull request fixes https://github.com/OData/odata.net/issues/2827*

### Description

*Modify `ExtractSegmentIdentifierAndParenthesisExpression` so that it doesn't throw when the parenthesis are in an unexpected state for key-as-parameter.*

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
@@ -69,10 +69,12 @@ namespace Microsoft.OData.UriParser
         }
 
         /// <summary>
-        /// Extracts the segment identifier and, if there is a valid parenthesis pair in the segment, the expression in the parenthesis.
-        /// This function does not validate anything and simply provides the raw text of both the identifier and parenthetical expression.
+        /// Extracts the segment identifier and, if there are parenthesis in the segment, the expression in the parenthesis.
+        /// Will throw if identifier is not found or if the parenthesis expression is malformed. This function does not validate
+        /// anything and simply provides the raw text of both the identifier and parenthetical expression.
         /// </summary>
         /// <remarks>Internal only so it can be called from tests. Should not be used outside <see cref="ODataPathParser"/>.</remarks>
+        /// <param name="enableKeyAsSegment">Whether key-as-segment is enabled.</param>
         /// <param name="segmentText">The segment text.</param>
         /// <param name="identifier">The identifier that was found.</param>
         /// <param name="parenthesisExpression">The query portion that was found. Will be null after the call if no query portion was present.</param>
@@ -112,6 +114,21 @@ namespace Microsoft.OData.UriParser
             {
                 identifier = segmentText;
                 parenthesisExpression = null;
+            }
+
+            if (identifier.Length == 0)
+            {
+                if (enableKeyAsSegment)
+                {
+                    // We already know that the segment itself is non-zero and so if the identifier is empty and we've enabled
+                    // key-as-segment we'll treat the entire segment as the identifier.
+                    identifier = segmentText;
+                    parenthesisExpression = null;
+                }
+                else
+                {
+                    throw ExceptionUtil.ResourceNotFoundError(ODataErrorStrings.RequestUriProcessor_EmptySegmentInRequestUrl);
+                }
             }
         }
 
--- a/src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
@@ -76,7 +76,8 @@ namespace Microsoft.OData.UriParser
         /// <param name="segmentText">The segment text.</param>
         /// <param name="identifier">The identifier that was found.</param>
         /// <param name="parenthesisExpression">The query portion that was found. Will be null after the call if no query portion was present.</param>
-        internal static void TryExtractSegmentIdentifierAndParenthesisExpression(
+        internal static void ExtractSegmentIdentifierAndParenthesisExpression(
+            bool enableKeyAsSegment,
             string segmentText,
             out string identifier,
             out string parenthesisExpression)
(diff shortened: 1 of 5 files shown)


src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -74,36 +74,61 @@ internal ODataPathParser(ODataUriParserConfiguration configuration)
...
+            out string identifier,
+            out string parenthesisExpression)
         {
             Debug.Assert(segmentText != null, "segment != null");
 
+            // We allow a single trailing '/', which results in an empty segment.
+            // However System.Uri removes it, so any empty segment we see is a 404 error.
+            if (segmentText.Length == 0)
+            {
+                throw ExceptionUtil.ResourceNotFoundError(ODataErrorStrings.RequestUriProcessor_EmptySegmentInRequestUrl);
```
habbes, line 93: Is this changing existing behaviour? Does this exception constitute a breaking change?

src/Microsoft.OData.Core/UriParser/Parsers/ODataPathParser.cs
```diff
@@ -74,36 +74,61 @@ internal ODataPathParser(ODataUriParserConfiguration configuration)
...
+            bool containsOpenParenthesis = parenthesisStart >= 0;
+            bool endsWithCloseParenthesis = segmentText[segmentText.Length - 1] == ')';
+
+            if (containsOpenParenthesis != endsWithCloseParenthesis && !enableKeyAsSegment)
             {
-                identifier = segmentText;
-                parenthesisExpression = null;
+                // If the parenthesis are not in the expected state fail with a syntax error. We do not do this for key-as-segment
+                // because in that case the segment text is potentially still a valid key value.
+                throw ExceptionUtil.CreateSyntaxError();
```
habbes, line 105: Is this condition consistent with existing behaviour?

In the current behaviour, if the segment does not contain and open parenthesis but ends with closing parenthesis, e.g. `folder)`, the condition would pass and the `identifier` would be set to the `segmentText`.

However, in your change, if the segment does not contain an open parenthesis but ends with closing parenthesis, the condition would fail when `enableKeyAsSegment` is false and a segment like `folder)` would throw an error.

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/ODataPathParserTests.cs
```diff
@@ -36,30 +36,72 @@ public void CallbackReturnsBatchReferenceSegment()
...
             string actualIdentifier;
             string queryPortion;
 
-            Action emptyString = () => ODataPathParser.ExtractSegmentIdentifierAndParenthesisExpression(string.Empty, out actualIdentifier, out queryPortion);
+            // Success cases
+            ExtractSegmentIdentifierAndParenthesisExpression(enableKeyAsSegment: false, "blah", "blah", null);
+            ExtractSegmentIdentifierAndParenthesisExpression(enableKeyAsSegment: false, "multiple words", "multiple words", null);
+            ExtractSegmentIdentifierAndParenthesisExpression(enableKeyAsSegment: false, "multiple(123)", "multiple", "123");
+            ExtractSegmentIdentifierAndParenthesisExpression(enableKeyAsSegment: false, "multiple(123;321)", "multiple", "123;321");
+            ExtractSegmentIdentifierAndParenthesisExpression(enableKeyAsSegment: false, "set()", "set", string.Empty);
```
habbes, line 48: Could you add a test case for a scenario that does not contain an `(` but ends with a `)`, like `blah)`, I expect the result would be `blah)` for the identifier and `null` for the parenthesis expression.

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/ODataPathParserTests.cs
```diff
@@ -36,30 +36,72 @@ public void CallbackReturnsBatchReferenceSegment()
...
+            string queryPortion;
+
+            // Success cases
+            ExtractSegmentIdentifierAndParenthesisExpression(enableKeyAsSegment: true, "blah", "blah", null);
+            ExtractSegmentIdentifierAndParenthesisExpression(enableKeyAsSegment: true, "multiple words", "multiple words", null);
+            ExtractSegmentIdentifierAndParenthesisExpression(enableKeyAsSegment: true, "multiple(123)", "multiple", "123");
+            ExtractSegmentIdentifierAndParenthesisExpression(enableKeyAsSegment: true, "multiple(123;321)", "multiple", "123;321");
+            ExtractSegmentIdentifierAndParenthesisExpression(enableKeyAsSegment: true, "set()", "set", string.Empty);
+            ExtractSegmentIdentifierAndParenthesisExpression(enableKeyAsSegment: true, "()", "()", null);
+            ExtractSegmentIdentifierAndParenthesisExpression(enableKeyAsSegment: true, "foo(", "foo(", null);
```
habbes, line 92: Could you add a test case for a scenario that does not contain an `(` but ends with a `)`, like `blah)`.

Also, is it a good idea to support different syntaxes depending on whether key as segment is enabled?

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Parsers/UriTemplateParserTests.cs
```diff
@@ -388,7 +388,7 @@ public void ErrorParameterTemplateInputShouldThrow()
                 };
 
                 Action action = () => uriParser.ParsePath();
-                action.Throws<ODataException>(errorCase.Error);
+                action.ThrowsAny<ODataException>(errorCase.Error);
```
habbes, line 391: What's the difference between `Throws` and `ThrowsAny`?

habbes (changes requested):
Confirm whether this change throws an exception when a segment that does not contain an opening parenthesis ends with a closing parenthesis, and whether that would be a regression.

WanjohiSammy:
@ificator as there has been no activity or response on this PR for a while, I am going to close it for now. If you would like to continue working on it, please feel free to reopen it.

PR #2830, opened 2023-12-19
Fixes #2829: support built-in function in $orderby.

### LINKED ISSUE #2829: Support tolower() in $orderby option
Copied from: https://github.com/OData/AspNetCoreOData/issues/1126

Describe the bug
I am not able to use a tolower(text_field_in_model) in an $orderby clause.

Reproduce steps
Hit a request against a model that has a search string:

http://localhost/api/students?$orderby=tolower(name)
Data Model
```C#
public class Student
{
    [Key]
    public string Id {get; set;}
    public string Name {get; set;}
}
```
EDM (CSDL) Model
```xml
<EntityType Name="Student">
    <Key>
        <PropertyRef Name="id" />
    </Key>
    <Property Name="id" Type="Edm.String" Nullable="false" />
    <Property Name="name" Type="Edm.String" Nullable="false" />
</EntityType>
```
Request/Response
http://localhost/api/students?$orderby=tolower(name)

Expected behavior
Results should be ordered by name (the property used in the $orderby clause)

Actual behavior
```C#
The query specified in the URI is not valid. Only ordering by properties is supported for non-primitive collections. Expressions are not supported.
```

I just checked the code and it appears only SingleValuePropertyAccessNode is supported, not SingleValueFunctionCallNode.

P.S.: This is a valid OData construct - https://services.odata.org/V4/TripPinService/People?$orderby=tolower(FirstName)%20desc,UserName

xuzhg:
Add test case to verify built-in function in $orderby

### Issues

*This pull request fixes #2829.*

Let's verify built-in function supported in ODL>

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FullUriFunctionalTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FullUriFunctionalTests.cs
@@ -71,6 +71,17 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
         }
 
         [Fact]
+        public void ParseOrderbyWithBuiltInFunction()
+        {
+            ODataUriParser parser = new ODataUriParser(HardCodedTestModel.TestModel, new Uri("http://www.odata.com/OData"), new Uri("http://www.odata.com/OData/People?$orderby=tolower(Name)"));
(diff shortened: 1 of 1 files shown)


habbes:
@xuzhg does this mean we support the feature in ODL but not in AspNetCoreOData?

habbes (approved):

PR #2834, opened 2024-01-04
Fix memory regression in DuplicatePropertyNameChecker

### LINKED ISSUE #2813: Perf regression: DuplicatePropertyNameChecker allocations
The `DuplicatePropertyNameChecker` is used to validate that we don't serialize two properties with the same name in the same resource (this can be turned off with validation settings). Some moons ago we introduced an object pool to reduce the number of `DuplicatePropertyNameChecker`s that are created per request, i.e. instead of creating one per resource, the object pool creates one per depth level.

We applied the object pool optimization only to `NETSTANDARD2_0_OR_GREATER` because were made used of an `ObjectPool` implementation that was not available in earlier frameworks. However, since we added `netcoreapp3.1` as an explicit target framework, the `NETSTANDARD` compiler flag no longer applied to it. This means that in NET Core 3.1 and greater, we're no longer use the object pool, resulting in excessive allocations:

image

### Assemblies affected

`Microsoft.OData.Core` 7.19.0

### Reproduce steps

Run the `TestServer` in `SerializationComparisonsBenchmarks` under an allocations tracker profiler. In my test, 147 requests were sent, each returning 100 resources that have 1 complex property and 1 complex collection property.

### Expected result

Because there are 2 levels of nesting and 147 request, I expected 2 * 147 = 294 instances of the duplicate name checker to be created.

### Actual result

61,216 instances were created.

### Additional detail

The fix is simple. In [`WriterValidator`](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/WriterValidator.cs), replace `#if NETSTANDARD2_0_OR_GREATER` with `#if NETSTANDARD2_0_OR_GREATER || NETCOREAPP3_1_OR_GREATER` and also install the `Microsoft.Extensions.ObjectPool` package for NET Core 3.1

habbes:
### Issues

*This pull request fixes #2813 

### Description

This PR fixes a couple of "low-hanging" excessive memory allocaitons.

#### `DuplicatePropertyNameChecker`

The `#if` flags we used to conditionally compile object-pooling for the `DuplicatePropertyNameChecker` only applied to .NET Standard 2.0. I've extended them to also cover .NET Core 3.1. The regression was probably introduced when we added `.netcoreapp3.1` as an explicit framework target.

I also made `NullDuplicatePropertyChecker` a "global" singleton instead of creating per request/WriterValidator. The impact here is low, but it doesn't hurt. In any case, I'd argue it makes the code cleaner anyway.

#### LINQ `Func<T,bool>` predicate

Found a case of `Func<T,bool>` lurking on a hot path from calling `resource.NonComputedProperties.SingleOrDefault(r => r.Name == propertyName)`. I reduced the allocation by implementing `SingleOrDefault` manually (raw while loop and `enumerator.MoveNext()`).

While this reduced the allocations, I question our use of `SingleOrDefault`. Personally, I think we should reduce or avoid the use of SingleOrDefault in our library. In cases where we need to validate uniqueness, I think it would much better to consider pre-validating the collection before use, or ways to ensure we validate uniqueness only once, or in some cases just consider duplicates undefined behaviour and leave it to the customer to deal with. In this case, we potentially do a full scan of the property list each time we need to fetch the value of a single property. And yet we'll still validate property name uniqueness with the duplicate property name checker inside the writer.

**Update**

Following this comment thread: https://github.com/OData/odata.net/pull/2834#discussion_r1441858961, I made the following change:

I removed the duplicate property name check from the `ODataResourceMetadataContext` because:
- it doesn't break existing tests
- the duplicate check adds extra cost on a hot path

I opted **not** to move the duplicate-check to the `ODataResource.Properties` setter because:
- we already have duplicate property name checking logic in the writer via `DuplicatePropertyNameChecker`
- in WebApi and OData Client, we can guarantee in the serializer that the property names are unique. That's also the best place to guarantee uniqueness since they have access the original property collection
- For customers using ODL directly, and who have disabled writer validation, we can assume they know what they're doing and already ensure that properties are unique.

However, if the last assumption doesn't hold, this could be a breaking change (though I think chances are extremely low). So, I'm open to restoring the original behaviour or moving the duplicate-check to the `ODataResource.Properties` setter if you or someone else believes it necessary to do so.

- [ ] *Test cases added*

### Results

**Before**

image

**After**

image

**Before**
image

**After**

image

image

**Before**
image

**After**
image

**Before**
image

**After**
image

**Before**
image

**After**
image

### DIFF
--- a/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
@@ -229,30 +229,14 @@ namespace Microsoft.OData.Evaluation
             ODataProperty property = null;
             if (resource.NonComputedProperties != null)
             {
-                // since this method is called frequently, we implement SingleOrDefault manually
-                // to avoid allocating predicate closures.
-                using (IEnumerator<ODataProperty> e = resource.NonComputedProperties.GetEnumerator())
+                // We use a manual loop instead of a for-loop to avoid
+                // closure allocations given how frequently this method is called.
+                foreach (ODataProperty item in resource.NonComputedProperties)
                 {
-                    while (e.MoveNext())
+                    if (item.Name == propertyName)
                     {
-                        ODataProperty temp = e.Current;
-                        if (temp.Name == propertyName)
-                        {
-                            // We should re-evaluate whether it's necessary to check for duplicates
-                            // at this point. Doing this duplicate-check here is inefficient given
-                            // that this method maybe called multiple times for the same resource.
-                            // It's implemented as-is for now to retain the behaviour of the previous
-                            // implementation that made use of SingleOrDefault().
-                            while (e.MoveNext())
-                            {
-                                if (e.Current.Name == propertyName)
-                                {
-                                    throw new InvalidOperationException(Strings.DuplicatePropertyNamesNotAllowed(propertyName));
-                                }
-                            }
-
-                            property = temp;
-                        }
+                        property = item;
+                        break;
                     }
                 }
             }
--- a/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
@@ -226,7 +226,37 @@ namespace Microsoft.OData.Evaluation
         {
             Debug.Assert(resource != null, "resource != null");
 
-            ODataProperty property = resource.NonComputedProperties == null ? null : resource.NonComputedProperties.SingleOrDefault(p => p.Name == propertyName);
+            ODataProperty property = null;
+            if (resource.NonComputedProperties != null)
+            {
+                // since this method is called frequently, we implement SingleOrDefault manually
+                // to avoid allocating predicate closures.
+                using (IEnumerator<ODataProperty> e = resource.NonComputedProperties.GetEnumerator())
+                {
+                    while (e.MoveNext())
+                    {
+                        ODataProperty temp = e.Current;
+                        if (temp.Name == propertyName)
+                        {
+                            // We should re-evaluate whether it's necessary to check for duplicates
+                            // at this point. Doing this duplicate-check here is inefficient given
+                            // that this method maybe called multiple times for the same resource.
+                            // It's implemented as-is for now to retain the behaviour of the previous
+                            // implementation that made use of SingleOrDefault().
+                            while (e.MoveNext())
+                            {
+                                if (e.Current.Name == propertyName)
+                                {
+                                    throw new InvalidOperationException(Strings.DuplicatePropertyNamesNotAllowed(propertyName));
+                                }
+                            }
+
+                            property = temp;
+                        }
+                    }
+                }
+            }
+
             if (property == null)
             {
                 if (isRequired)
--- a/src/Microsoft.OData.Core/WriterValidator.cs
+++ b/src/Microsoft.OData.Core/WriterValidator.cs
@@ -31,11 +31,6 @@ namespace Microsoft.OData
 #endif
 
         /// <summary>
-        /// Instances of the NullDuplicatePropertyNameChecker.
-        /// </summary>
-        private IDuplicatePropertyNameChecker nullDuplicatePropertyNameChecker;
-
-        /// <summary>
         /// Creates a WriterValidator instance and binds it to settings.
         /// </summary>
         /// <param name="settings">The ODataMessageWriterSettings instance to bind to.</param>
@@ -66,13 +61,7 @@ namespace Microsoft.OData
             }
             else
             {
-                if (this.nullDuplicatePropertyNameChecker == null)
-                {
-                    // NullDuplicatePropertyNameChecker does nothing, so we can use a single instance during the writing process.
-                    this.nullDuplicatePropertyNameChecker = new NullDuplicatePropertyNameChecker();
-                }
-
-                duplicatePropertyNameChecker = this.nullDuplicatePropertyNameChecker;
+                duplicatePropertyNameChecker = NullDuplicatePropertyNameChecker.Instance;
             }
 
             return duplicatePropertyNameChecker;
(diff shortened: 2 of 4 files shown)


src/Microsoft.OData.Core/WriterValidator.cs
```diff
@@ -6,9 +6,8 @@
 
 namespace Microsoft.OData
 {
-    using System;
     using System.Collections.Generic;
-#if NETSTANDARD2_0_OR_GREATER
+#if NETSTANDARD2_0_OR_GREATER || NETCOREAPP3_1_OR_GREATER
```
odero, line 10: Why doesn't `NETSTANDARD2_0_OR_GREATER` also include `NETCOREAPP_3_1+`?

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -226,7 +226,37 @@ private static bool TryGetPrimitiveOrEnumPropertyValue(ODataResourceBase resourc
         {
             Debug.Assert(resource != null, "resource != null");
 
-            ODataProperty property = resource.NonComputedProperties == null ? null : resource.NonComputedProperties.SingleOrDefault(p => p.Name == propertyName);
+            ODataProperty property = null;
+            if (resource.NonComputedProperties != null)
+            {
+                // since this method is called frequently, we implement SingleOrDefault manually
+                // to avoid allocating predicate closures.
+                using (IEnumerator<ODataProperty> e = resource.NonComputedProperties.GetEnumerator())
```
odero, line 234: Why not do the check at the time of adding to `NonComputedProperties`? And/or use a data structure that can do this validation at point of adding?

habbes: I think validating at the time of adding NonComputedProperties or using a better suited data structure would indeed be better. But I didn't want to risk breaking existing behaviour since I hadn't done a thorough analysis to ensure changing the behaviour here doesn't break any assumptions its callers.

src/Microsoft.OData.Core/WriterValidator.cs
habbes: In the past #if NETSTANDARD2_0_OR_GREATER` used to apply to .NET Core 3.1 and later. But when we added `.netcoreapp3.1` as an explicit target framework in our csproj file, `NETSTANDARD2_0_OR_GREATER` stopped applying to it. I don't know why actually. But since then, we had to manually use `NETCOREAPP3_1_OR_GREATER`.

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
habbes: One challenge with validating the uniqueness properties at the time that they are added to `NonComputedProperties` is that we don't know when the properties are added to the collection. The `NonComputedProperties` refers to the `ODatarResource.Properties` property, which is an `IEnumerable<ODataProperty>`.

We can detect when the properties collection is set in the setter, and do the validation there, but we can't tell if new entries are added to the collection after the property has been set. For example, if we only validate the properties collection in the setter, we would not be able to catch the following violation:

```c#
var properties = new List<ODataProperty>
{
   new ODataProperty { Name = "Foo", Value = "Bar" },
   new ODataProperty { Name = "A", Value = "B" }
}

odataResource.Properties = properties; // duplicate name validation would occur here

properties.Add(new ODataProperty { Name = "Foo", Value = "Baz" }); // this duplicate property name violation would not be caught
```

It's worthwhile to note that we do already perform so other verification in the setter, and we can argue that the verification would also not catch violations that happen after the property has been set.

My only concern is that taking this route for the uniqueness check is effectively changing existing behaviour where a property is guaranteed to be unique at the type we query its value, even if the collection had changed since the `ODataResource.Properties` was set. This is behaviour if I have not yet verified is safe to break.

That said, my opinion is that we should probably leave it to the user to ensure they gave us unique properties and if they violate that we treat that as unsafe/undefined behaviour with unpredictable results.

It's also worth noting that the duplicate name checker will still verify property uniqueness independently of the check we do here, unless validation is disabled (in which case we probably shouldn't be doing that verification here either anyway). So, this check is sort of redundant.

habbes: I've removed to the duplicate check and no unit test has failed. We also have other cases where we (questionably) use SingleOrDefault in our libraries, so maybe the original author did not added with intentions to ensure strong guarantees of uniqueness, but out of habit.

habbes: I have revised the code and removed the duplicate check, the implementation now has `FirstOrDefault()` semantics instead of `SingleOrDefault()`.

I removed the duplicate-check from this method because:
- it doesn't break existing tests
- the duplicate check adds extra cost on a hot path

I opted **not** to move the duplicate-check to the `ODataResource.Properties` setter because:
- we already have a duplicate property name checking logic in the writer via `DuplicatePropertyNameChecker`
- in WebApi and OData Client, we can guarantee in the serializer that the property names are unique. That's also the best place to guarantee uniqueness since they have access the original property collection
- For customers using ODL directly, and who have disabled writer validation, we can assume they know what they're doing and already ensure that properties are unique.

However, if the last assumption doesn't hold, this could be a breaking change (though I think chances are extremely low). So, I'm open to restoring the original behaviour or moving the duplicate-check to the `ODataResource.Properties` setter if you or someone else believes it necessary to do so.

odero: Are the perf results still the same after removing the duplicate check?

odero (approved):

gathogojr (approved):
:shipit:

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
habbes: I had not collected alloc results after removing the duplicate check (I don't expect the result to be much different), but since I'm currently working on an area that calls this method, I've just taken CPU samples before and after applying this change:

You can see the `TryGetPrimitiveOrEnumPropertyValue` went down from 0.96% to 0.52% CPU. In the before graph, we can see that the lion's share was from `TryGetSingle` (which is called by `SingleOrDefault()`)

**Before**

image

**After**

image

PR #2835, opened 2024-01-05
Optimize `ODataPath` operations

### LINKED ISSUE #2814: Perf: `ODataPath` operations are costly
We perform many operations on `ODataPath` instances many times during serialization. For example, we clone the path and append segments to the clone for every resource and nested resource.

image

image

[`AppendEntityKeySegment(ODataPath, bool)`](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/ODataWriterCore.cs#L2840) is one such expensive operation.
- It uses `GetSingle()` to find key properties from the properties list, this means that it scans the entire collection of properties for each key property (I don't think it's a good idea to validate uniqueness at this point)
- constructing the `KeySegment` is not cheap because the keys collection that has just been obtained is copied into a `List<T>`
- `AddKeySegment` is not cheap. Many of our path operations that rely on `PathSegmentHandler` are not cheap.

Some path operations like `AddKeySegment` use a visitor-like pattern based on custom `PathSegmentHandler` and the `ODataPath.WalkWith` method. The `ODataPath.WalkWith` method provides a generic way to implement various operations that requires traversing the path segments without providing direct access to the segments list. However, this design turns out to lead to inefficient implementations that cost a lot of CPU time and memory allocations in hot paths. We end up iterating of the same list of segments multiple times, copying it multiple times in different temporary collections and resizing arrays along the way.

Let's take the `AddKeySegment` for example:
- It uses the `SplitEndingSegmentOfTypeHandler<T>` to visit the path. It splits the path into two queues, using a segment of a specified type as the separator (in this case a TypeSegment)
- The handler uses the two queues to build 2 ODataPath instances, for the parts upto and after the separator
- The key segment is added to the first path then the second path is added to the first path after the key

As you can see, in this case we allocate and populate queues, which could also result in array resizing. We use the queues to create two `ODataPath`s, this will result in two new List instance getting created and the segments from the queues copied into them. Then finally, we call the Add and AddRange. This means we traverse the same segments multiple times in different collections, we create multiple temporary copies of the segments.

We could optimize this and other such methods by creating just a single list that will contain the segments for the new path. We can pre-calculate the size of this list to avoid resizing the array. We can add the segments to this list directly without requiring the path segment handler or queues. Then we can create the final `ODataPath` once the list of segments is complete. I would also propose introducing an ODataPath constructor or factory function that does not copy the input list to be used in cases we create a new list for the sole purpose of creating an ODataPath (i.e. the list will not be used for anything else after creating the ODataPath). I think that's safer than having `Add()` and `AddRange()` methods which introduce the possibility of mutating an existing `ODataPath` and potential bugs.

### Assemblies affected

`Microsoft.OData.Core` 7.x

### Reproduce steps

I ran the `TestServer` project in the `SerializerComparisonsTests` solution under CPU and allocations tracker profilers.

habbes:
### Issues

*This pull request fixes #2814 .*

### Description

This PR does bunch of optimizations in different places to improve the performance of `ODataPath` operations. My focus was on `AppendKeySegment` since it took 7+% CPU on my tests, but the improvements affect other methods as well. Below is summary of the changes made.

- **`ODataPath`**
  - Reduced array copies and unnecessary array traversals (e.g. to check for null segments)
  - created a constructor that takes a `List<ODataPathSegment>` and sets it as the `path.Segments` property without performing a copy of checking for nulls. This is used in internal methods that create a new path based on a previous path by either adding or removing segments to a new segments list. In such cases, there's no reason to check whether we have null segments in the list since the old `ODataPath` was already valid, we also don't need to copy the segments list in the constructor since the `ODataPath` is being created from a newly created temporary collection, we can just use that list as is.
  - Replaced the use of `SplitEndingSegmentOfTypeHandler` for splitting a path with a basic while loop. Removed `SplitEndingSegmentTypeHandler` since it's no longer used.
- **`KeySegment` and retrieval of key values**
  - Like in the previous case, reduced array copies and validations
  - Added an internal constructor that accepts an `IList<KeyValuePair<string, object>>` and stores this list directly in the `ReadOnlyCollection` instead of calling `ToList()` on the `IEnumerable<KeyValuePair<string, object>>`. This constructor also optionally skips checking whether the key segment type matches the navigation source type
  - Reduced use of LINQ. Using a concrete List<T> to create the key value collection instead of `yield return`.  This avoid the use of `ToArray()` which was expensive. As a result I had to change the signature of the keys field from `KeyValuePair<string, object>[]` to `IList<KeyValuePair<string, object>>[]`
- **`as IEdmEntityType`casts**
  - reduced the number of casts happening. Frequent casts add up and we have a complex type of hierarchy. In some cases we would have a method call `Scope.ResourceType as IEdmEntityType` and then call another method which also performs the same cast. In such cases I pass the already cast `IEdmEntityType` as a parameter to the callee so that I can avoid performing the same casts in called methods.
- **`FullName()` Edm extension method**
  - This method is called frequently. If the element implements `IEdmFullNamedElement`, then the underlying `FullName()` method is called, otherwise it constructs the full name from combining `Namespace` and `Name` properties. The method is called most of the time with an element that implements `IEdmFullNamedElement` but it reads the `Name` and `Namespace` properties first. Reading those properties is somewhat expensive. So I optimized this method by moving the check for `IEdmFullNamedElement` to the top, before checking the `Name` and `Namespace` properties.
  - I reduced how often this method is called. There are many cases where the full name is passed as a parameter to methods when it's only used to generate an exception message when an error is thrown, but not used otherwise. So now instead of calling `FullName()` up the call stack, I just pass the entity type down an call `FullName()` only when the exception is thrown.

#### CPU Profiler

CPU dropped from > 7% to < 3% for the `AppendKeySegment` method

**Before**
image

**After**
image

#### Writer benchmarks

We see a latency and memory usage improvement across all Writer configurations. As an example `ODataMessageWriter-Utf8JsonWriter-Async` has an ~8.5% improvement in latency and 24% drop in memory allocations.

```
BenchmarkDotNet=v0.13.1, OS=Windows 10.0.22621
.NET SDK=8.0.100
  [Host]     : .NET 6.0.25 (6.0.2523.51912), X64 RyuJIT
  DefaultJob : .NET 6.0.25 (6.0.2523.51912), X64 RyuJIT
```

**Before**
|           Method |                                           WriterName |       Mean |      Error |     StdDev |     Median |      Gen 0 |     Gen 1 |  Allocated |
|----------------- |----------------------------------------------------- |-----------:|-----------:|-----------:|-----------:|-----------:|----------:|-----------:|
| WriteToFileAsync |                                   ODataMessageWriter | 423.633 ms |  8.4293 ms | 16.2404 ms | 422.386 ms | 53000.0000 |         - | 224,018 KB |
| WriteToFileAsync |                             ODataMessageWriter-Async | 719.716 ms | 12.4249 ms | 11.6222 ms | 721.246 ms | 63000.0000 | 1000.0000 | 266,770 KB |
| WriteToFileAsync |                      ODataMessageWriter-NoValidation | 346.953 ms |  6.0102 ms |  5.3279 ms | 348.297 ms | 47000.0000 |         - | 200,579 KB |
| WriteToFileAsync |                ODataMessageWriter-NoValidation-Async | 657.755 ms | 12.9246 ms | 25.5119 ms | 646.122 ms | 57000.0000 | 1000.0000 | 243,329 KB |
| WriteToFileAsync |                    ODataMessageWriter-Utf8JsonWriter | 392.370 ms |  6.5777 ms |  8.7811 ms | 393.420 ms | 51000.0000 |         - | 217,340 KB |
| WriteToFileAsync |              ODataMessageWriter-Utf8JsonWriter-Async | 598.974 ms |  5.6653 ms |  4.4231 ms | 598.411 ms | 51000.0000 |         - | 218,139 KB |
| WriteToFileAsync |       ODataMessageWriter-Utf8JsonWriter-NoValidation | 331.546 ms |  6.6227 ms |  6.1949 ms | 329.425 ms | 46000.0000 |         - | 193,903 KB |
| WriteToFileAsync | ODataMessageWriter-Utf8JsonWriter-NoValidation-Async | 533.909 ms |  3.3615 ms |  3.1443 ms | 532.770 ms | 46000.0000 |         - | 194,698 KB |

**After**

|           Method |                                           WriterName |       Mean |      Error |     StdDev |     Median |      Gen 0 |     Gen 1 |  Allocated |
|----------------- |----------------------------------------------------- |-----------:|-----------:|-----------:|-----------:|-----------:|----------:|-----------:|
| WriteToFileAsync |                                   ODataMessageWriter | 330.500 ms |  3.0234 ms |  2.6801 ms | 330.384 ms | 40000.0000 |         - | 171,285 KB |
| WriteToFileAsync |                             ODataMessageWriter-Async | 623.021 ms | 11.8429 ms | 12.1618 ms | 619.559 ms | 50000.0000 | 1000.0000 | 214,036 KB |
| WriteToFileAsync |                      ODataMessageWriter-NoValidation | 284.599 ms |  5.1672 ms |  6.5349 ms | 282.055 ms | 40500.0000 |         - | 171,278 KB |
| WriteToFileAsync |                ODataMessageWriter-NoValidation-Async | 575.517 ms |  9.9837 ms | 16.4034 ms | 570.288 ms | 50000.0000 | 1000.0000 | 214,029 KB |
| WriteToFileAsync |                    ODataMessageWriter-Utf8JsonWriter | 314.829 ms |  2.1455 ms |  1.7916 ms | 314.157 ms | 39000.0000 |  500.0000 | 164,606 KB |
| WriteToFileAsync |              ODataMessageWriter-Utf8JsonWriter-Async | 547.834 ms |  4.6274 ms |  4.1021 ms | 548.000 ms | 39000.0000 |         - | 165,419 KB |
| WriteToFileAsync |       ODataMessageWriter-Utf8JsonWriter-NoValidation | 282.200 ms |  5.4841 ms |  6.0956 ms | 280.777 ms | 39000.0000 |         - | 164,609 KB |
| WriteToFileAsync | ODataMessageWriter-Utf8JsonWriter-NoValidation-Async | 500.182 ms |  9.9743 ms | 15.2319 ms | 495.599 ms | 39000.0000 | 1000.0000 | 165,398 KB |

#### Concurrent HTTP requests benchmarks

Using `bombardier` on the `TestServer`:

We see an ~8% latency improvement when using the async writer and about 11% when using the sync writer. The gap between sync and async is still large. There's still a lot of work to do :)

**Before**

(sync)

```
bombardier -l -c 50 -d 30s https://localhost:7120/customers/ODataMessageWriter-Utf8JsonWriter
```

```
  Reqs/sec      1948.04    1871.73   24414.38
  Latency       27.83ms     6.96ms   243.49ms
  Latency Distribution
     50%    26.59ms
     75%    30.85ms
     90%    38.14ms
     95%    45.50ms
     99%    59.13ms
  HTTP codes:
    1xx - 0, 2xx - 53834, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    50.84MB/s
```

(async)

```
bombardier -l -c 50 -d 30s https://localhost:7120/customers/ODataMessageWriter-Utf8JsonWriter-Async
```
```
  Reqs/sec      1467.15    1671.74   25448.70
  Latency       38.49ms     8.59ms   272.24ms
  Latency Distribution
     50%    36.15ms
     75%    46.81ms
     90%    54.53ms
     95%    63.47ms
     99%    80.15ms
  HTTP codes:
    1xx - 0, 2xx - 38960, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    36.76MB/s
```

**After**

(sync)
```
bombardier -l -c 50 -d 30s https://localhost:7120/customers/ODataMessageWriter-Utf8JsonWriter
```
```
bombardier -l -c 50 -d 30s https://localhost:7120/customers/ODataMessageWriter-Utf8JsonWriter
Bombarding https://localhost:7120/customers/ODataMessageWriter-Utf8JsonWriter for 30s using 50 connection(s)
[=================================================================================================================================================================================================================================================================================] 30s
Done!
  Reqs/sec      2044.23    1621.43   30253.21
  Latency       24.56ms     5.42ms   212.18ms
  Latency Distribution
     50%    23.70ms
     75%    31.45ms
     90%    35.39ms
     95%    40.41ms
     99%    52.06ms
  HTTP codes:
    1xx - 0, 2xx - 61100, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    57.64MB/s
```

(async)

```
bombardier -l -c 50 -d 30s https://localhost:7120/customers/ODataMessageWriter-Utf8JsonWriter-Async
```
```
  Reqs/sec      1548.30    2831.05   89738.59
  Latency       35.64ms     6.90ms   217.80ms
  Latency Distribution
     50%    33.69ms
     75%    41.64ms
     90%    48.24ms
     95%    56.33ms
     99%    71.01ms
  HTTP codes:
    1xx - 0, 2xx - 42080, 3xx - 0, 4xx - 0, 5xx - 0
    others - 0
  Throughput:    39.72MB/s
```

- [x] *Test cases added*


src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -169,60 +169,63 @@ public IODataResourceTypeContext TypeContext
         /// <param name="actualEntityType">The edm entity type of the resource</param>
         /// <param name="requiresId">Whether key properties are required to be returned</param>
         /// <returns>Key value pair array</returns>
```
ElizabethOkerio, line 171: nit: Update this doc string.

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -169,60 +169,63 @@ public IODataResourceTypeContext TypeContext
...
             return keyProperties;
         }
 
-        private static IEnumerable<KeyValuePair<string, object>> GetPropertyValues(IEnumerable<IEdmStructuralProperty> properties, ODataResourceBase resource, IEdmEntityType actualEntityType, bool isRequired)
+        private static List<KeyValuePair<string, object>> GetPropertyValues(IEnumerable<IEdmStructuralProperty> properties, ODataResourceBase resource, IEdmStructuredType actualEntityType, bool isRequired)
         {
-            string actualEntityTypeName = actualEntityType.FullName();
             object primitiveValue;
+            // Initialize capacity to 1 because this method is mostly called to get values of key properties
+            // and most entities have a single key property
```
ElizabethOkerio, line 206: What happens in cases where where entities have more than 1 property?

habbes: Then the list will be resized. Here I was just optimizing for the common case.

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -346,14 +351,27 @@ private static bool ValidateEntityTypeHasKeyProperties(KeyValuePair<string, obje
...
         }
 
+
+
+        /// <summary>
+        /// Gets the entity type name of the specified resource.
+        /// </summary>
+        /// <param name="resource"></param>
+        /// <param name="resourceType"></param>
+        /// <returns></returns>
```
ElizabethOkerio, line 369: These doc strings are empty.

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -531,10 +549,10 @@ internal ODataResourceMetadataContextWithModel(ODataResourceBase resource, IODat
                         IEdmEntityType entityType = this.actualResourceType as IEdmEntityType;
                         if (entityType != null)
                         {
-                            this.keyProperties = GetPropertyValues(entityType.Key(), resource, entityType, this.requiresId).ToArray();
+                            this.keyProperties = GetPropertyValues(entityType.Key(), resource, this.actualResourceType, this.requiresId).ToArray();
                         }
 
-                        if (!ValidateEntityTypeHasKeyProperties(this.keyProperties, this.ActualResourceTypeName, this.requiresId))
+                        if (!ValidateEntityTypeHasKeyProperties(this.keyProperties, resource, entityType, this.requiresId))
```
gathogojr, line 555: `entityType` can be null here...

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -168,61 +168,64 @@ public IODataResourceTypeContext TypeContext
...
             // if we have serializationInfo, try that first
             if (serializationInfo != null)
             {
-                keyProperties = ODataResourceMetadataContextWithoutModel.GetPropertiesBySerializationInfoPropertyKind(resource, ODataPropertyKind.Key, actualEntityTypeName);
+                keyProperties = ODataResourceMetadataContextWithoutModel.GetPropertiesBySerializationInfoPropertyKind(resource, ODataPropertyKind.Key, actualEntityType);
             }
 
             // if we didn't get any keys from serializationInfo, try using entity type
-            if ((keyProperties == null || keyProperties.Length == 0) && actualEntityType != null)
+            if ((keyProperties == null || keyProperties.Count == 0) && actualEntityType != null)
```
gathogojr, line 189:
```suggestion
            if ((keyProperties?.Count == 0) && actualEntityType != null)
```

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
habbes: Good catch. I can pass the `actualResourceType` as-is (`IEdmStructuredType`) instead.

habbes: the `ToArray()` is also not needed here anymore.

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -553,10 +571,9 @@ internal ODataResourceMetadataContextWithModel(ODataResourceBase resource, IODat
                 {
                     if (this.etagProperties == null)
                     {
-                        IEdmEntityType actualEntityType = this.actualResourceType as IEdmEntityType;
                         IEnumerable<IEdmStructuralProperty> properties = this.ComputeETagPropertiesFromAnnotation();
                         this.etagProperties = properties.Any()
-                            ? GetPropertyValues(properties, resource, actualEntityType, /*isRequired*/ false)
+                            ? GetPropertyValues(properties, resource, this.actualResourceType, /*isRequired*/ false)
```
gathogojr, line 576:
```suggestion
                            ? GetPropertyValues(properties, resource, this.actualResourceType, isRequired: false)
```

src/Microsoft.OData.Core/UriParser/SemanticAst/KeySegment.cs
```diff
@@ -45,14 +45,30 @@ public sealed class KeySegment : ODataPathSegment
         /// <exception cref="ODataException">Throws if the input entity set is not related to the input type.</exception>
         [SuppressMessage("Microsoft.Design", "CA1006:DoNotNestGenericTypesInMemberSignatures", Justification = "Using key value pair is exactly what we want here.")]
         public KeySegment(IEnumerable<KeyValuePair<string, object>> keys, IEdmEntityType edmType, IEdmNavigationSource navigationSource)
+            : this(keys.ToList(), edmType, navigationSource, verifyNavigationSourceType: true)
         {
-            this.keys = new ReadOnlyCollection<KeyValuePair<string, object>>(keys.ToList());
+        }
+
+        /// <summary>
+        /// Construct a segment that represents a key lookup.
```
gathogojr, line 53:
```suggestion
        /// Constructs a segment that represents a key lookup.
```

habbes: Same term used on the other constructors. But I've updated this as well as the docs of the other constructors in this class.

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -346,14 +351,27 @@ private static bool ValidateEntityTypeHasKeyProperties(KeyValuePair<string, obje
...
             return properties.ToArray();
         }
 
+
+
+        /// <summary>
+        /// Gets the resource type name of the specified resource.
+        /// </summary>
+        /// <param name="resource"></param>
+        /// <param name="resourceType"></param>
```
ElizabethOkerio, line 368: The doc strings are empty

src/Microsoft.OData.Core/UriParser/SemanticAst/KeySegment.cs
```diff
@@ -37,22 +37,38 @@ public sealed class KeySegment : ODataPathSegment
...
+
+        /// <summary>
+        /// Constructs a segment that represents a key lookup.
+        /// </summary>
+        /// <param name="keys">The set of key property names and the values to be used in searching for the given item.</param>
+        /// <param name="edmType">The type of the item this key returns.</param>
+        /// <param name="navigationSource">The navigation source that this key is used to search.</param>
+        /// <param name="verifyNavigationSourceType">Whether to verify that the entity set is related to the input type.</param>
+        /// <remarks>
+        /// The constructor does not copy the input <paramref name="keys"/> should not use the keys collection after calling this constructor.
```
ElizabethOkerio, line 60: Is there anything that prevents one from using the keys collection after calling this constructor or someone just needs to know. I think you can also add a test or sth,

habbes: Some just needs to know. That's why I made it `internal`.

habbes: External code will still call the public constructor that copies the list and verifies the type.

ElizabethOkerio (approved):

src/Microsoft.OData.Core/Evaluation/ODataResourceMetadataContext.cs
```diff
@@ -168,61 +168,64 @@ public IODataResourceTypeContext TypeContext
...
+            if (!ValidateEntityTypeHasKeyProperties(keyProperties, resource, actualEntityType, requiresId))
+            {
                 return Enumerable.Empty<KeyValuePair<string, object>>().ToArray();
             }
 
             return keyProperties;
         }
 
-        private static IEnumerable<KeyValuePair<string, object>> GetPropertyValues(IEnumerable<IEdmStructuralProperty> properties, ODataResourceBase resource, IEdmEntityType actualEntityType, bool isRequired)
+        private static List<KeyValuePair<string, object>> GetPropertyValues(IEnumerable<IEdmStructuralProperty> properties, ODataResourceBase resource, IEdmStructuredType actualEntityType, bool isRequired)
```
KenitoInc, line 202: We need a better name for `actualEntityType` since the type has changed

KenitoInc, line 202: Follow up question
Is the method called multiple times?

habbes: This method is used to retrieve the values of key properties and ETag properties. So, it would be called, for example, whenever we want to build the key segment, which we do for every entity resource in the response (whether top-level or navigation property).

habbes: I've renamed it to `actualResourceType`

ElizabethOkerio (approved):

PR #2842, opened 2024-01-24
exposing `DataServiceActionQuery<T>` properties for testing purposes

### LINKED ISSUE #2841: DataServiceActionQuery can expose more private fields
For creating tests we would benefit if the class `DataServiceActionQuery<T>` would expose its private fields `Context` and `Parameters` like `DataServiceQuery` does too.

Can this be done please, or are there reasons to not do this?

### Assemblies affected

Microsoft.OData.Client Version 7.20.0

corranrogue9:
### Issues

*This pull request fixes #2841.*

### Description

Exposing two properties on `DataServiceActionQuery<T>` to facilitate testing

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceActionQueryOfT.cs
+++ b/src/Microsoft.OData.Client/DataServiceActionQueryOfT.cs
@@ -20,12 +20,12 @@ namespace Microsoft.OData.Client
         /// <summary>
         /// Context associated with this query.
         /// </summary>
-        private readonly DataServiceContext Context;
+        private readonly DataServiceContext context;
 
         /// <summary>
         /// Parameters of this action.
         /// </summary>
-        private readonly BodyOperationParameter[] Parameters;
+        private readonly BodyOperationParameter[] parameters;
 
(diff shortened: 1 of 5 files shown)


src/Microsoft.OData.Client/DataServiceActionQueryOfT.cs
```diff
@@ -35,9 +35,34 @@ public sealed class DataServiceActionQuery<T>
...
+        /// </summary>
+        public IEnumerable<BodyOperationParameter> Parameters
+        {
+            get
+            {
+                foreach (var element in this.parameters)
+                {
+                    yield return element;
+                }
+            }
```
habbes, line 65: I assume this is done so that the user does not modify the array. But it seems excessive to me, especially given that we do not seem modify the list internally. I think by the time they're casting it back into array to try and modify it, they're already assuming any risks that come with that. But if we really want to prevent that, then maybe it's better to hand them a `ReadOnlyCollection` that's computed once so we don't have to generate the enumerable each time the property is accessed.

PR #2846, opened 2024-01-25
Support writing large json base64 and string values

### LINKED ISSUE #2845: Using `ODataUtf8JsonWriter` leads to increased memory usage and latency when writing properties with large string values
We have customers with scenarios that write payloads with really large string values. For example, string property values with 18 million characters. This causes problems when using `ODataUtf8JsonWriter` because `Utf8JsonWriter.WriteStringValue()` will write that entire string to the buffer all at once. Since this exceeds the buffer size, it will allocate a buffer large enough to fit the entire string in memory. This causes issues with large-object heap (LOH) allocations, increase in latency due to increased GC pressure, etc.

This is a known issue of `Utf8JsonWriter`. At the time of this writing there's a [TODO item in the `Utf8JsonWriter` source](https://github.com/dotnet/runtime/blob/main/src/libraries/System.Text.Json/src/System/Text/Json/Writer/Utf8JsonWriter.WriteValues.String.cs#L115)

image

that references the following issue: https://github.com/dotnet/runtime/issues/29293 from 2019. That issue proposes writing large strings in smaller 4KB chunks instead of all at once. That issue seems to have been closed due to inactivity. But there's an other open issue with an active discussion thread and design proposals https://github.com/dotnet/runtime/issues/67337 that seems to have caught some momentum.

Meanwhile, we could solve this in our library by detecting large strings based on some threshold, write to the buffer in chunks and flush it to the stream intermittently before getting to the point where the buffer needs to be resized. Ironically, our default `JsonWriter` gets this for free by using the `TextWriter`.

The drawback of doing this ourselves (at least until `Utf8JsonWriter` supports it), is that we'll have to handle string escaping, and write the separator manually. Luckily, we already dealt with the separator issue when working on raw values: https://github.com/OData/odata.net/pull/2527

### Assemblies affected

`Microsoft.OData.Core` 7

### Reproduce steps

TODO

### Expected result

TODO

### Actual result

TODO

ElizabethOkerio:
### Issues

*This pull request fixes #2845.*

### Description
This pull request introduces functionality to handle the writing of large string and byte array values. It accomplishes this by breaking down oversized strings and bytes into smaller chunks and directly writing them to the buffer.

The process involves first checking if the size of the string or byte array exceeds the buffer's capacity. If it does, the program iterates through the large string, dividing it into chunks of a predefined size. Each chunk is then written to the buffer until the entire string is successfully written.

Furthermore, special characters within the string are properly escaped to ensure accurate representation during the writing process.

Methods added: 
1. WriteStringValueInChunks(ReadOnlySpan<char> value)
2. WriteStringValueInChunksAsync
3. WriteByteValueInChunks(ReadOnlySpan<byte> value)
4. WriteByteValueInChunksAsync(ReadOnlyMemory<byte> value)

LOH allocations before these changes: 
char[] in LOH:
GetImage
byte[] in LOH from growing the writer buffer when calling WriteStringValue  
GetImage (1)

LOH allocations after these changes: 
No byte[] or char[] in the LOH that comes from ODataUtf8JsonWriter 
 
GetImage (3)

Latency Before:
Bombarding https://localhost:7120/customers/ODataMessageWriter-Utf8JsonWriter-Async?count=10&largeFields=true for 30s using 30 connection(s) 

[=================================================================================================================] 30s 

Done! 

  Reqs/sec        23.70      69.93    1492.98 
  Latency         1.71s      0.87s      7.36s 
  Latency Distribution 
     50%      1.60s 
     75%      1.73s 
     90%      1.87s 
     95%      2.29s 
     99%      6.58s 
  HTTP codes: 
    1xx - 0, 2xx - 534, 3xx - 0, 4xx - 0, 5xx - 0 
    others - 2 
  Errors: 
    tls handshake timed out - 2 
  Throughput:   533.38MB/s 

Latency After: 
Bombarding https://localhost:7120/customers/ODataMessageWriter-Utf8JsonWriter-Async?count=10&largeFields=true for 30s using 30 connection(s) 
[=================================================================================================================] 30s 
Done! 
  Reqs/sec        26.07      44.40     429.41 
  Latency         1.41s   578.31ms      5.48s 
  Latency Distribution 
     50%      1.33s 
     75%      1.47s 
     90%      1.64s 
     95%      1.86s 
     99%      4.79s 
  HTTP codes: 
    1xx - 0, 2xx - 648, 3xx - 0, 4xx - 0, 5xx - 0 
    others - 0 
  Throughput:   651.77MB/s 

With this changes, there is an improvement in latency and there are no allocations in the LOH. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
@@ -16,6 +16,8 @@ using System.Threading.Tasks;
 using System.Buffers;
 using System.Runtime.CompilerServices;
 using Microsoft.OData.Edm;
+using System.Text.Unicode;
+using System.Buffers.Text;
 
 namespace Microsoft.OData.Json
 {
@@ -37,11 +39,13 @@ namespace Microsoft.OData.Json
         private readonly Stream writeStream;
         private readonly Utf8JsonWriter utf8JsonWriter;
         private readonly PooledByteBufferWriter bufferWriter;
+        private readonly JavaScriptEncoder encoder;
         private readonly int bufferSize;
         private readonly bool isIeee754Compatible;
         private readonly bool leaveStreamOpen;
         private readonly Encoding outputEncoding;
         private bool disposed;
+        private const int chunkSize = 2048;
         /// The Utf8JsonWriter internally keeps track of when to write the item separtor ','
         /// between key-value pairs in an object and between items in an array
         /// However, we bypass the Utf8JsonWriter in our implementation of WriteRawValue
@@ -86,7 +90,10 @@ namespace Microsoft.OData.Json
         /// an array e.g. ["rawValue1", "rawValue2", "rawValue3"
         /// </summary>
         private bool isWritingConsecutiveRawValuesAtStartOfArray = false;
-        
+        /// <summary>
+        /// Rrepresents a read-only memory block containing the byte representation of the double quote character ("). 
+        /// </summary>
+        private readonly static ReadOnlyMemory<byte> doubleQuote = new byte[] { (byte)'"' };
 
         /// <summary>
         /// Creates an instance of <see cref="ODataUtf8JsonWriter"/>.
@@ -107,6 +114,7 @@ namespace Microsoft.OData.Json
             this.bufferFlushThreshold = 0.9f * this.bufferSize;
             this.leaveStreamOpen = leaveStreamOpen;
             this.outputEncoding = encoding;
+            this.encoder = encoder ?? JavaScriptEncoder.Default;
 
             if (this.outputEncoding.CodePage == Encoding.UTF8.CodePage)
             {
@@ -123,11 +131,12 @@ namespace Microsoft.OData.Json
 
             this.utf8JsonWriter = new Utf8JsonWriter(
                 this.bufferWriter,
-                new JsonWriterOptions {
+                new JsonWriterOptions
+                {
                     // we don't need to perform validation here since the higher-level
                     // writers already perform validation
                     SkipValidation = true,
-                    Encoder = encoder ?? JavaScriptEncoder.Default
+                    Encoder = this.encoder
                 });
         }
 
@@ -256,7 +265,7 @@ namespace Microsoft.OData.Json
             {
                 this.utf8JsonWriter.WriteNumberValue(value);
             }
-            
+
             this.FlushIfBufferThresholdReached();
         }
 
@@ -279,7 +288,7 @@ namespace Microsoft.OData.Json
             {
                 this.utf8JsonWriter.WriteNumberValue(value);
             }
-            
+
             this.FlushIfBufferThresholdReached();
         }
 
@@ -364,32 +373,260 @@ namespace Microsoft.OData.Json
 
         public void WriteValue(string value)
         {
-            this.WriteSeparatorIfNecessary();
             if (value == null)
             {
+                this.WriteSeparatorIfNecessary();
                 this.utf8JsonWriter.WriteNullValue();
+            }
+            else if (value.Length > bufferWriter.FreeCapacity)
+            {
+                WriteStringValueInChunks(value.AsSpan());
             }
             else
             {
+                this.WriteSeparatorIfNecessary();
                 this.utf8JsonWriter.WriteStringValue(value);
             }
 
             this.FlushIfBufferThresholdReached();
         }
 
+        /// <summary>
+        /// Writes a string value into the buffer in chunks, handling escaping if necessary.
+        /// </summary>
+        /// <param name="value">The string value to write.</param>
+        /// <remarks>
+        /// This method writes the provided string value into the buffer in manageable chunks to avoid
+        /// excessive memory allocations and buffer overflows. If the string requires escaping, it is
+        /// processed accordingly to ensure correct JSON formatting.
+        /// </remarks>
+        private void WriteStringValueInChunks(ReadOnlySpan<char> value)
+        {
+            this.CommitUtf8JsonWriterContentsToBuffer();
+
+            WriteItemWithSeparatorIfNeeded();
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+
+            this.Flush();
+
+            int charsNotProcessedFromPreviousChunk = 0;
+
+            for (int i = 0; i < value.Length; i += chunkSize)
+            {
+                int remainingChars = Math.Min(chunkSize, value.Length - i);
+                bool isFinalBlock = (i + remainingChars) == value.Length;
+                ReadOnlySpan<char> chunk = value.Slice(i - charsNotProcessedFromPreviousChunk, charsNotProcessedFromPreviousChunk + remainingChars);
+                int firstIndexToEscape = NeedsEscaping(chunk);
+
+                Debug.Assert(firstIndexToEscape >= -1 && firstIndexToEscape < value.Length);
+
+                if (firstIndexToEscape != -1)
+                {
+                    WriteEscapedStringChunk(chunk, firstIndexToEscape, isFinalBlock, out charsNotProcessedFromPreviousChunk);
+                }
+                else
+                {
+                    WriteStringChunk(chunk, isFinalBlock);
+                }
+
+                // Flush the buffer if needed
+                this.FlushIfBufferThresholdReached();
+            }
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+
+            // since we bypass the Utf8JsonWriter, we need to signal to other
+            // Write methods that a separator should be written first
+            CheckIfSeparatorNeeded();
+
+            CheckIfAtManualValueArrayStart();
+        }
+
+        /// <summary>
+        /// Determines if any characters in the provided string span require escaping according to the specified encoder.
+        /// </summary>
+        /// <param name="value">The string span to analyze for characters needing escaping.</param>
+        /// <returns>
+        /// The index of the first character requiring escaping, or -1 if no characters need escaping or if the string span is empty.
+        /// </returns>
+        private unsafe int NeedsEscaping(ReadOnlySpan<char> value)
+        {
+            // Some implementations of JavaScriptEncoder.FindFirstCharacterToEncode may not accept
+            // null pointers and guard against that. Hence, check up-front to return -1.
+            if (value.IsEmpty)
+            {
+                return -1;
+            }
+
+            fixed (char* ptr = value)
+            {
+                return this.encoder.FindFirstCharacterToEncode(ptr, value.Length);
+            }
+        }
+
+        /// <summary>
+        /// Writes an escaped string chunk into the buffer.
+        /// </summary>
+        /// <param name="sourceChunk">The source chunk of the string to be escaped.</param>
+        /// <param name="firstIndexToEscape">The index of the first character that requires escaping.</param>
+        /// <param name="isFinalBlock">Indicates whether the chunk is the final block of the string.</param>
+        /// <remarks>
+        /// This method escapes characters within the provided string chunk that require escaping
+        /// according to the specified encoder. It allocates memory for the escaped string and handles
+        /// buffer management using either stack allocation or array pooling. Once the string is escaped,
+        /// it is written into the buffer. If the encoding operation fails, an exception is thrown. 
+        /// </remarks>
+        private void WriteEscapedStringChunk(ReadOnlySpan<char> sourceChunk, int firstIndexToEscape, bool isFinalBlock, out int preChunkNotProcessed)
+        {
+            char[] valueArray = null;
+            const int StackallocCharThreshold = 128;
+
+            int maxSize = firstIndexToEscape + 6 * (sourceChunk.Length - firstIndexToEscape);
+
+            Span<char> destination = maxSize <= StackallocCharThreshold
+                ? stackalloc char[maxSize]
+                : valueArray = ArrayPool<char>.Shared.Rent(maxSize);
+
+            preChunkNotProcessed = 0;
+
+            OperationStatus status = this.encoder.Encode(
+                   sourceChunk,
+                   destination,
+                   out int charsConsumed,
+                   out int charsWritten,
+                   isFinalBlock
+               );
+
+            Debug.Assert(status == OperationStatus.Done || status == OperationStatus.NeedMoreData);
+
+            preChunkNotProcessed = sourceChunk.Length - charsConsumed;
+            WriteStringChunk(destination.Slice(0, charsWritten), isFinalBlock);
+
+            if (valueArray != null)
+            {
+                ArrayPool<char>.Shared.Return(valueArray);
+            }
+        }
+
+        /// <summary>
+        /// Writes a chunk of the string into the buffer, converting it to UTF-8 encoding.
+        /// </summary>
+        /// <param name="chunk">The chunk of the string to be written into the buffer.</param>
+        /// <param name="isFinalBlock">Indicates whether the chunk is the final block of the string.</param>
+        /// <remarks>
+        /// This method converts the provided chunk of the string from UTF-16 encoding to UTF-8 encoding
+        /// and writes it into the buffer. It ensures that the buffer has sufficient space for the encoded
+        /// string and notifies the buffer writer of the write operation. This method assumes that the UTF-8
+        /// encoding does not exceed the maximum byte count specified by the UTF-8 encoding for the given
+        /// chunk of the string.
+        /// </remarks>
+        private void WriteStringChunk(ReadOnlySpan<char> chunk, bool isFinalBlock)
+        {
+            int maxUtf8Length = Encoding.UTF8.GetMaxByteCount(chunk.Length);
+
+            Span<byte> output = bufferWriter.GetSpan(maxUtf8Length);
+            OperationStatus status = Utf8.FromUtf16(chunk, output, out int charsRead, out int charsWritten, isFinalBlock);
+            Debug.Assert(status == OperationStatus.Done || status == OperationStatus.NeedMoreData);
+            Debug.Assert(charsRead == chunk.Length);
+
+            // notify the bufferWriter of the write
+            bufferWriter.Advance(charsWritten);
+        }
+
+        /// <summary>
+        /// Determines whether a separator should be written before the next value.
+        /// </summary>
+        /// <remarks>
+        /// This method checks the current writing context to decide whether a separator should be written
+        /// before the next value in the JSON output. It sets the flag indicating whether a separator should
+        /// be written based on conditions such as whether the writer is at the start of an array, writing
+        /// consecutive raw values at the start of an array, or if the writer is not currently in an array.
+        /// </remarks>
+        private void CheckIfSeparatorNeeded()
+        {
+            if (this.isWritingAtStartOfArray || this.isWritingConsecutiveRawValuesAtStartOfArray || !this.IsInArray())
+            {
+                this.shouldWriteSeparator = true;
+            }
+        }
+
         public void WriteValue(byte[] value)
         {
-            this.WriteSeparatorIfNecessary();
             if (value == null)
             {
+                this.WriteSeparatorIfNecessary();
                 this.utf8JsonWriter.WriteNullValue();
+            }
+            else if (value.Length > bufferWriter.Capacity)
+            {
+                WriteByteValueInChunks(value.AsSpan());
             }
             else
             {
+                this.WriteSeparatorIfNecessary();
                 this.utf8JsonWriter.WriteBase64StringValue(value);
             }
-            
+
             this.FlushIfBufferThresholdReached();
+        }
+
+        /// <summary>
+        /// Writes the byte value represented by the provided ReadOnlySpan in chunks using Base64 encoding.
+        /// </summary>
+        /// <param name="value">The ReadOnlySpan containing the byte value to be written.</param>
+        private void WriteByteValueInChunks(ReadOnlySpan<byte> value)
+        {
+            this.CommitUtf8JsonWriterContentsToBuffer();
+
+            WriteItemWithSeparatorIfNeeded();
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+
+            this.Flush();
+
+            int bytesNotWrittenFromPreviousChunk = 0;
+
+            for (int i = 0; i < value.Length; i += chunkSize)
+            {
+                int remainingChars = Math.Min(chunkSize, value.Length - i);
+                bool isFinalBlock = (i + remainingChars) == value.Length;
+
+                ReadOnlySpan<byte> chunk = value.Slice(i - bytesNotWrittenFromPreviousChunk, remainingChars + bytesNotWrittenFromPreviousChunk);
+
+                Base64EncodeAndWriteChunk(chunk, isFinalBlock, out bytesNotWrittenFromPreviousChunk);
+
+                this.FlushIfBufferThresholdReached();
+            }
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+
+            // since we bypass the Utf8JsonWriter, we need to signal to other
+            // Write methods that a separator should be written first
+            CheckIfSeparatorNeeded();
+
+            CheckIfAtManualValueArrayStart();
+        }
+
+        /// <summary>
+        /// Encodes the specified chunk of bytes using Base64 encoding and writes the encoded data to the buffer writer.
+        /// </summary>
+        /// <param name="chunk">The chunk of bytes to be encoded.</param>
+        /// <param name="isFinalBlock">A boolean value indicating whether the chunk is the final block of data.</param>
+        /// <param name="bytesNotConsumed">An out parameter indicating the number of bytes from the previous chunk that were not processed.</param>
+        private void Base64EncodeAndWriteChunk(ReadOnlySpan<byte> chunk, bool isFinalBlock, out int bytesNotConsumed)
+        {
+            int encodingLength = Base64.GetMaxEncodedToUtf8Length(chunk.Length);
+            var output = this.bufferWriter.GetSpan(encodingLength);
+            bytesNotConsumed = 0;
+
+            OperationStatus status = Base64.EncodeToUtf8(chunk, output, out int consumed, out int written, isFinalBlock);
+
+            Debug.Assert(status == OperationStatus.Done || status == OperationStatus.NeedMoreData);
+
+            bytesNotConsumed = chunk.Length - consumed;
+
+            this.bufferWriter.Advance(written);
         }
 
         public void WriteValue(JsonElement value)
@@ -437,12 +674,38 @@ namespace Microsoft.OData.Json
 
             // since we bypass the Utf8JsonWriter, we need to signal to other
             // Write methods that a separator should be written first
-            if (this.isWritingAtStartOfArray || this.isWritingConsecutiveRawValuesAtStartOfArray || !this.IsInArray())
-            {
-                this.shouldWriteSeparator = true;
-            }
+            CheckIfSeparatorNeeded();
 
             if (this.isWritingAtStartOfArray || this.isWritingConsecutiveRawValuesAtStartOfArray)
+            {
+                this.isWritingConsecutiveRawValuesAtStartOfArray = true;
+            }
+
+            this.isWritingAtStartOfArray = false;
+        }
+
+        /// <summary>
+        /// Writes an item with a separator if needed, based on the context of being within an array and not writing at the start of the array.
+        /// </summary>
+        /// <remarks>
+        /// If the method is called within an array and it's not the first item being written, it places a separator before the value being written.
+        /// </remarks>
+        private void WriteItemWithSeparatorIfNeeded()
+        {
+            if (this.IsInArray() && !isWritingAtStartOfArray)
+            {
+                // Place a separator before the raw value if
+                // this is an array, unless this is the first item in the array.
+                this.bufferWriter.Write(itemSeparator.Slice(0, 1).Span);
+            }
+        }
+
+        // <summary>
+        /// Checks if the writer is at the start of a manual value array and updates the state accordingly.
+        /// </summary>
+        private void CheckIfAtManualValueArrayStart()
+        {
+            if(this.isWritingAtStartOfArray || this.isWritingConsecutiveRawValuesAtStartOfArray)
             {
                 this.isWritingConsecutiveRawValuesAtStartOfArray = true;
             }
@@ -797,32 +1060,132 @@ namespace Microsoft.OData.Json
 
         public async Task WriteValueAsync(string value)
         {
-            this.WriteSeparatorIfNecessary();
             if (value == null)
             {
+                this.WriteSeparatorIfNecessary();
                 this.utf8JsonWriter.WriteNullValue();
+            }
+            else if (value.Length > bufferWriter.FreeCapacity)
+            {
+                await WriteStringValueInChunksAsync(value.AsMemory());
             }
             else
             {
+                this.WriteSeparatorIfNecessary();
                 this.utf8JsonWriter.WriteStringValue(value.ToString());
             }
 
             await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
         }
 
+        /// <summary>
+        /// Writes a string value into the buffer in chunks asynchronously, handling escaping if necessary.
+        /// </summary>
+        /// <param name="value">The string value to write.</param>
+        /// <remarks>
+        /// This method writes the provided string value into the buffer in manageable chunks to avoid
+        /// excessive memory allocations and buffer overflows. If the string requires escaping, it is
+        /// processed accordingly to ensure correct JSON formatting.
+        /// </remarks>
+        private async ValueTask WriteStringValueInChunksAsync(ReadOnlyMemory<char> value)
+        {
+            this.CommitUtf8JsonWriterContentsToBuffer();
+
+            WriteItemWithSeparatorIfNeeded();
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+
+            await this.FlushAsync();
+
+            int charsNotProcessedFromPreviousChunk = 0;
+
+            for (int i = 0; i < value.Length; i += chunkSize)
+            {
+                int remainingChars = Math.Min(chunkSize, value.Length - i);
+                bool isFinalBlock = (i + remainingChars) == value.Length;
+                ReadOnlyMemory<char> chunk = value.Slice(i - charsNotProcessedFromPreviousChunk, charsNotProcessedFromPreviousChunk + remainingChars);
+
+                int firstIndexToEscape = NeedsEscaping(chunk.Span);
+
+                Debug.Assert(firstIndexToEscape >= -1 && firstIndexToEscape < value.Length);
+
+                if (firstIndexToEscape != -1)
+                {
+                    WriteEscapedStringChunk(chunk.Span, firstIndexToEscape, isFinalBlock, out charsNotProcessedFromPreviousChunk);
+                }
+                else
+                {
+                    WriteStringChunk(chunk.Span, isFinalBlock);
+                }
+
+                // Flush the buffer if needed
+                await this.FlushIfBufferThresholdReachedAsync();
+            }
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+
+            // since we bypass the Utf8JsonWriter, we need to signal to other
+            // Write methods that a separator should be written first
+            CheckIfSeparatorNeeded();
+
+            CheckIfAtManualValueArrayStart();
+        }
+
         public async Task WriteValueAsync(byte[] value)
         {
-            this.WriteSeparatorIfNecessary();
             if (value == null)
             {
+                this.WriteSeparatorIfNecessary();
                 this.utf8JsonWriter.WriteNullValue();
+            }
+            else if (value.Length > bufferWriter.Capacity)
+            {
+                await WriteByteValueInChunksAsync(value);
             }
             else
             {
+                this.WriteSeparatorIfNecessary();
                 this.utf8JsonWriter.WriteBase64StringValue(value);
             }
 
             await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+        }
+
+        /// <summary>
+        /// Writes the byte value represented by the provided ReadOnlySpan in chunks using Base64 encoding.
+        /// </summary>
+        /// <param name="value">The ReadOnlySpan containing the byte value to be written.</param>
+        private async ValueTask WriteByteValueInChunksAsync(ReadOnlyMemory<byte> value)
+        {
+            this.CommitUtf8JsonWriterContentsToBuffer();
+
+            WriteItemWithSeparatorIfNeeded();
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+
+            await FlushAsync().ConfigureAwait(false);
+
+            int bytesNotWrittenFromPreviousChunk = 0;
+
+            for (int i = 0; i < value.Length; i += chunkSize)
+            {
+                int remainingChars = Math.Min(chunkSize, value.Length - i);
+                bool isFinalBlock = (i + remainingChars) == value.Length;
+
+                ReadOnlyMemory<byte> chunk = value.Slice(i - bytesNotWrittenFromPreviousChunk, remainingChars + bytesNotWrittenFromPreviousChunk);
+
+                Base64EncodeAndWriteChunk(chunk.Span, isFinalBlock, out bytesNotWrittenFromPreviousChunk);
+
+                await FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+            }
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+
+            // since we bypass the Utf8JsonWriter, we need to signal to other
+            // Write methods that a separator should be written first
+            CheckIfSeparatorNeeded();
+
+            CheckIfAtManualValueArrayStart();
         }
 
         public async Task WriteValueAsync(JsonElement value)
@@ -884,7 +1247,7 @@ namespace Microsoft.OData.Json
             }
         }
 
-#endregion
+        #endregion
 
     }
 }
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
@@ -536,6 +536,10 @@ namespace Microsoft.OData.Json
             Span<byte> output = bufferWriter.GetSpan(maxUtf8Length);
             OperationStatus status = Utf8.FromUtf16(chunk, output, out int charsRead, out int charsWritten, isFinalBlock);
             Debug.Assert(status == OperationStatus.Done || status == OperationStatus.NeedMoreData);
+
+            // The charsRead will always be equal to chunk.Length. This is because the characters
+            // that would cause utf-8 encoding to result in partial processing is already 
+            // taken care of by the JavascriptEncoder when escaping special characters. 
             Debug.Assert(charsRead == chunk.Length);
 
             // notify the bufferWriter of the write
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
@@ -378,7 +378,7 @@ namespace Microsoft.OData.Json
                 this.WriteSeparatorIfNecessary();
                 this.utf8JsonWriter.WriteNullValue();
             }
-            else if (value.Length > bufferWriter.FreeCapacity)
+            else if (value.Length > bufferWriter.FreeCapacity || value.Length > chunkSize)
             {
                 WriteStringValueInChunks(value.AsSpan());
             }
@@ -558,7 +558,7 @@ namespace Microsoft.OData.Json
                 this.WriteSeparatorIfNecessary();
                 this.utf8JsonWriter.WriteNullValue();
             }
-            else if (value.Length > bufferWriter.Capacity)
+            else if (value.Length > bufferWriter.FreeCapacity || value.Length > chunkSize)
             {
                 WriteByteValueInChunks(value.AsSpan());
             }
@@ -1065,7 +1065,7 @@ namespace Microsoft.OData.Json
                 this.WriteSeparatorIfNecessary();
                 this.utf8JsonWriter.WriteNullValue();
             }
-            else if (value.Length > bufferWriter.FreeCapacity)
+            else if (value.Length > bufferWriter.FreeCapacity || value.Length > chunkSize)
             {
                 await WriteStringValueInChunksAsync(value.AsMemory());
             }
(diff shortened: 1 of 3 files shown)


src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -107,6 +109,7 @@ public ODataUtf8JsonWriter(Stream outputStream, bool isIeee754Compatible, Encodi
             this.bufferFlushThreshold = 0.9f * this.bufferSize;
             this.leaveStreamOpen = leaveStreamOpen;
             this.outputEncoding = encoding;
+            this.encoder = encoder ?? JavaScriptEncoder.Default;
```
habbes, line 112: Pass `this.encoder` to the `Utf8JsonWriter` constructor below.

ElizabethOkerio: I think it's already being passed.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +384,99 @@ public void WriteValue(string value)
...
+                {
+                    ArrayPool<char>.Shared.Return(destination.ToArray());
+                }
+            }
+        }
+
+        private void WriteStringChunk(ReadOnlySpan<char> chunk)
+        {
+            // Create a buffer to hold UTF-8 encoded bytes
+            byte[] utf8Bytes = new byte[Encoding.UTF8.GetByteCount(chunk)];
```
habbes, line 474: Instead of allocating a new array, we could rent it from the ArrayPool. But since we're going to write these utf8Bytes to the bufferWriter, maybe we should write it to the buffer directly. Here's an example:

```c#
int maxUtf8Length = Encoding.UTF8.GetMaxByteCount(chunk);

// technically this might resize the buffer. But if we use a small chunk size, say 4K, and since we're flushing
// after every chunk, this is unlikely even if we consider the max byte count
Span<byte> output  = bufferWriter.GetSpan(maxUtf8Length);
OperationStatus status = Utf8.FromUtf16(chunk, output, out int charsRead, out int bytesWritten, isFinalBlock: false /* set to true on the last chunk */);
Debug.Assert(status == OperationStatus.Done);
Debug.Assert(charsRead == chunk.Length); // did we read everything in the chunk

// notify the bufferWriter of the write
bufferWriter.Advance(bytesWritten);
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +384,99 @@ public void WriteValue(string value)
...
+
+            // Write the UTF-8 bytes to the buffer writer
+            this.bufferWriter.Write(utf8Bytes);
```
habbes, line 477: Instead of allocating a new array, we could rent it from the ArrayPool. But since we're going to write these utf8Bytes to the bufferWriter, maybe we should write it to the buffer directly. Here's an example:

```c#
int maxUtf8Length = Encoding.UTF8.GetMaxByteCount(chunk);

// technically this might resize the buffer. But if we use a small chunk size, say 4K, and since we're flushing
// after every chunk, this is unlikely even if we consider the max byte count
Span<byte> output  = bufferWriter.GetSpan(maxUtf8Length);
OperationStatus status = Utf8.FromUtf16(chunk, output, out int charsRead, out int bytesWritten, isFinalBlock: false /* set to true on the last chunk */);
Debug.Assert(status == OperationStatus.Done);
Debug.Assert(charsRead == chunk.Length); // did we read everything in the chunk

// notify the bufferWriter of the write
bufferWriter.Advance(bytesWritten);
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +384,99 @@ public void WriteValue(string value)
...
+                {
+                    WriteStringChunk(chunk);
+                }
+
+                // Flush the buffer if needed
+                this.FlushIfBufferThresholdReached();
+            }
+        }
+
+        private int NeedsEscaping(ReadOnlySpan<char> value)
```
habbes, line 410: since this returns an index, I think you should call it `GetIndexOfFirstCharToEscape`

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +384,99 @@ public void WriteValue(string value)
...
+                if (!string.Equals(value[i].ToString(), encoded, StringComparison.Ordinal))
+                {
+                    return i;
+                }
+            }
+
+            return -1;
+        }
+
+        private void WriteEscapedStringChunk(ReadOnlySpan<char> sourceChunk)
```
habbes, line 433: I think you should pass the index of the first character to index here so that the characters before that index can be copied directly without escaping.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +384,99 @@ public void WriteValue(string value)
...
+        {
+            char[] valueArray = null;
+            const int maxSize = 0;
```
habbes, line 436: You can estimate maxSize as follow: `indexOfFirstCharToEscape + 6 * remainingChars`. See: https://source.dot.net/#System.Text.Json/System/Text/Json/Writer/JsonWriterHelper.Escaping.cs,de155df00ca44e94

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +384,99 @@ public void WriteValue(string value)
...
+            }
+
+            return -1;
+        }
+
+        private void WriteEscapedStringChunk(ReadOnlySpan<char> sourceChunk)
+        {
+            char[] valueArray = null;
+            const int maxSize = 0; 
+            const int threshold = 120;
```
habbes, line 437: I think we can use 128. And maybe store this in a `StackallocCharThreshold` constant.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +384,99 @@ public void WriteValue(string value)
...
+
+            try
+            {
+                OperationStatus status = this.encoder.EncodeUtf8(
+                    MemoryMarshal.Cast<char, byte>(sourceChunk),
+                    MemoryMarshal.Cast<char, byte>(destination), // Use the byte span of the destination
+                    out int bytesConsumed,
+                    out int bytesWritten,
+                    isFinalBlock: true
+                );
```
habbes, line 451: [EncodeUtf8](https://learn.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.memorymarshal.cast?view=net-8.0) expects the input to already be utf-8 text. Casting char span to byte span does not seem safe. I'm almost certain if we interpret the utf-16 text as utf-8 data without transcoding, we may end up with corrupted data.

We should probably be using the [`JavaScriptEncoder.Encode`](https://learn.microsoft.com/en-us/dotnet/api/system.text.encodings.web.textencoder.encode?view=net-7.0#system-text-encodings-web-textencoder-encode(system-readonlyspan((system-char))-system-span((system-char))-system-int32@-system-int32@-system-boolean)) method instead which takes a `ReadOnlySpan<char>` input:

```c#
OperatioStatus status = this.encoder.Encode(
  chunk,
 destination,
 out int charsConsumed,
 out int bytesWritten,
 isFinalBlock: false /* this should be set to true on the last chunk of the string, so should be passed as parameter to this method */
);

Debug.Assert(status == OperationStatus.Done);
Debug.Assert(charsConsumed == chunk.Length);
WriteStringChunk(destination.Slice(0, bytesWritten));

// ...
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +384,99 @@ public void WriteValue(string value)
...
+
+                if (status == OperationStatus.Done)
+                {
+                    WriteStringChunk(destination.Slice(0, bytesWritten));
+                }
+                else
+                {
+                    Console.WriteLine($"Encoding failed with status: {status}");
```
habbes, line 459: I'm not sure how we should handle it. STJ source code just uses `Debug.Assert`. Should we do the same or throw an exception?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +384,99 @@ public void WriteValue(string value)
             this.FlushIfBufferThresholdReached();
         }
 
+        private void WriteStringValueInChunks(ReadOnlySpan<char> value)
+        {
+            int chunkSize = bufferWriter.FreeCapacity;
```
habbes, line 389: I wonder if we should a constant chunkSize instead, like 4096.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +384,99 @@ public void WriteValue(string value)
             this.FlushIfBufferThresholdReached();
         }
 
+        private void WriteStringValueInChunks(ReadOnlySpan<char> value)
+        {
+            int chunkSize = bufferWriter.FreeCapacity;
+
```
habbes, line 390: Before writing the string to the output buffer, we should write the double quote. Maybe we should do that before flushing to prepare for the write. We should also commit the contents of the json write to the buffer writer before we start writing to the buffer writer manually:

```c#
this.CommitUtf8JsonWriterContentsToBuffer();
this.bufferWriter.Write(DoubleQuote.Slice(0, 1).Span;
this.Flush();

int chunkSize = LargeStringChunkSize;
// ... write the chunks in the loop

this.bufferWriter.Write(DoubleQuote);

// signal to the writer that we may need to write separator before the next property or value. See how we do it in `WriteRawValueCore`

```

The `DoubleQuote` definition looks like:

```c#
private readonly static ReadOnlyMemory<byte> doubleQuote = new byte[] { (byte)'"' };
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: This is what is currently happening:

```c#
new JsonWriterOptions {
 ...
  Encoder == encoder ?? JavaScriptEncoder.Default
```

And this is what I'm proposing

```c#
new JsonWriterOptions {
 ...
  Encoder == this.encoder
```

While they're technically equivalent, the first one unnecessarily evaluates the `encoder ?? JavaScriptEncoder.Default` again and it is more prone to bugs. For example if we ever change the `this.encoder` to something else, we might forget to also make a similar change to the instance passed to the constructor and introduce bugs.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +387,117 @@ public void WriteValue(string value)
...
+            for (int i = 0; i < value.Length; i += chunkSize)
+            {
+                int remainingChars = Math.Min(chunkSize, value.Length - i);
+                bool isFinalBlock = (i + remainingChars) == value.Length;
+
+                ReadOnlySpan<char> chunk = value.Slice(i, remainingChars);
+
+                int firstIndexToEscape = NeedsEscaping(chunk);
+
+                Debug.Assert(firstIndexToEscape >= -1 && firstIndexToEscape < value.Length);
```
habbes, line 407: Is there any chance te `firstIndexToEscape` can be  -1?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +387,117 @@ public void WriteValue(string value)
...
+            }
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+
+            // since we bypass the Utf8JsonWriter, we need to signal to other
+            // Write methods that a separator should be written first
+            if (this.isWritingAtStartOfArray || this.isWritingConsecutiveRawValuesAtStartOfArray || !this.IsInArray())
+            {
+                this.shouldWriteSeparator = true;
+            }
```
habbes, line 429: I'd suggest putting this logic in a method and reusing it here and in `WriteRawValueCore` (we'll also need to use it when writing large byte arrays).

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +387,117 @@ public void WriteValue(string value)
...
+                {
+                    WriteEscapedStringChunk(chunk, firstIndexToEscape, isFinalBlock);
+                }
+                else
+                {
+                    WriteStringChunk(chunk, isFinalBlock);
+                }
+
+                // Flush the buffer if needed
+                this.FlushIfBufferThresholdReached();
```
habbes, line 419: I think it's a good to call to only flush if buffer is almost full. But I'm wondering whether doing this with a `chunkSize` that evenly splits the buffer size can lead to resizing the buffer more often than is necessary.

Here's an example:

Our default buffer size is 16k. The chunk size 2k. Let's assume the simple scenario where the string is larger than 16k.

Let's assume the first 7 chunks do not require escaping and will each use 2048 bytes in the output buffer. These first 7 chunks will be written without flushing. Meaning the buffer will have 14,336 bytes used. But this is still less than the buffer threshold (0.9 * 16,384 = 14,745.6) so we still won't flush. If the 8th chunk does not require escaping, we'll fill up the buffer completely but we'll flush before the next iteration. But if the 8th chunk requires escaping even one character, it will exceed the buffer capacity and the buffer will need to be resized.

Now, I don't think this is necessarily a major issue since it would only have to be resized once for the entire payload. Since the buffer size is larger, we're less likely to overflow it again because the flush threshold stays the same, i.e. the flush threshold will remain as 0.9 * 16k even if the buffer size is 32k.

But I was just wondering whether using a different chunk size, such as 3k, where after 5 chunks we'll flush before hitting the buffer size. But there may be other drawbacks of choosing a buffer size which is not a power of 2.

But given that this entire PR is essentially an edge case, and the main goal is avoiding excessive, large buffer allocations, I think it's reasonable to keep it simple and not stress so much of such at the moment. But let me know what you think.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +387,117 @@ public void WriteValue(string value)
             this.FlushIfBufferThresholdReached();
         }
 
+        private void WriteStringValueInChunks(ReadOnlySpan<char> value)
+        {
+            this.CommitUtf8JsonWriterContentsToBuffer(); 
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+            this.Flush();
+
+            int chunkSize = 2048;
```
habbes, line 396: I was wondering what a good chunk size.

With 2048 I think we're less likely to resize the output buffer even if chunks need escaping, but we'd be renting temporary escaping buffer more frequently. But we may be renting the same escape buffer multiple times and not impact memory allocations. With 4096 we have fewer chunks, and therefore fewer times we'd have to go through the escape path if some chunks need escaping. 

Since we don't necessarily flush after every chunk, I think 2048 is probably fine.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +387,117 @@ public void WriteValue(string value)
...
+                }
+            }
+        }
+
+        private void WriteStringChunk(ReadOnlySpan<char> chunk, bool isFinalBlock)
+        {
+            int maxUtf8Length = Encoding.UTF8.GetMaxByteCount(chunk.Length);
+
+            // technically this might resize the buffer. But if we use a small chunk size, say 4K, and since we're flushing
+            // after every chunk, this is unlikely even if we consider the max byte count
```
habbes, line 491: I think since we're not buffering after every chunk, and we're using a 2k chunk size, you should update or remove this comment.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +387,117 @@ public void WriteValue(string value)
...
+            Span<byte> output = bufferWriter.GetSpan(maxUtf8Length);
+            OperationStatus status = Utf8.FromUtf16(chunk, output, out int charsRead, out int bytesWritten, isFinalBlock);
+            Debug.Assert(status == OperationStatus.Done);
+            Debug.Assert(charsRead == chunk.Length); // did we read everything in the chunk
```
habbes, line 495: I don't think this comment is necessary. The statement seems self-explanatory.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +387,117 @@ public void WriteValue(string value)
...
+
+        private void WriteStringChunk(ReadOnlySpan<char> chunk, bool isFinalBlock)
+        {
+            int maxUtf8Length = Encoding.UTF8.GetMaxByteCount(chunk.Length);
+
+            // technically this might resize the buffer. But if we use a small chunk size, say 4K, and since we're flushing
+            // after every chunk, this is unlikely even if we consider the max byte count
+            Span<byte> output = bufferWriter.GetSpan(maxUtf8Length);
+            OperationStatus status = Utf8.FromUtf16(chunk, output, out int charsRead, out int bytesWritten, isFinalBlock);
+            Debug.Assert(status == OperationStatus.Done);
```
habbes, line 494: I wonder what's the likelihood of the `OperationStatus` not being `Done` and what the impact of not handling that case is.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -377,6 +387,117 @@ public void WriteValue(string value)
             this.FlushIfBufferThresholdReached();
         }
 
+        private void WriteStringValueInChunks(ReadOnlySpan<char> value)
```
habbes, line 390: It may be a good idea to have doc comments on these private helper methods for people who may not have context on why this is necessary.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataUtf8JsonWriterTests.cs
```diff
@@ -169,6 +169,24 @@ public void WritePrimitiveValueString()
...
+                new string('x', 30000),
+                "\"" + new string('x', 30000) + "\""
+                );
+        }
+
+        [Fact]
+        public void WritePrimitiveValueLargeStringWithSpecialChars()
+        {
+            this.VerifyWritePrimitiveValue(
+                new string('x', 20000) + "!@#$%" + new string('x', 10000),
```
habbes, line 185: Do these chars really need to be escaped? I think you could add chars that need to be escaped like `"` as well as non-ascii characters that need special utf-8 encoding.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
ElizabethOkerio: Can we bench mark some the things you've raised and see what works best then we'll decide on what to do.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
ElizabethOkerio: I think having benchmarks on this will guide this decision.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
ElizabethOkerio: Can we throw an exception here?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -364,34 +372,406 @@ public void WriteValue(sbyte value)
...
+                // Place a separator before the raw value if
+                // this is an array, unless this is the first item in the array.
+                this.bufferWriter.Write(itemSeparator.Slice(0, 1).Span);
+            }
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+
+            int chunkSize = 2048;
+
+            int prevChunkNotProcessed = 0;
```
habbes, line 482:
```suggestion
            int charsNotProcessedFromPreviousChunk = 0;

```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: If `isFinalBlock` can be `false`, then I don't think `status` is guaranteed to be done. Verify whether we can get an `OperationStatus.NeedMoreData` status.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -364,34 +372,406 @@ public void WriteValue(sbyte value)
...
+            // since we bypass the Utf8JsonWriter, we need to signal to other
+            // Write methods that a separator should be written first
+            CheckIfSeparatorNeeded();
+
+            if (this.isWritingAtStartOfArray || this.isWritingConsecutiveRawValuesAtStartOfArray)
+            {
+                this.isWritingConsecutiveRawValuesAtStartOfArray = true;
+            }
+
+            this.isWritingAtStartOfArray = false;
```
habbes, line 705: Consider abstracting in a reusable method.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -364,34 +372,406 @@ public void WriteValue(sbyte value)
...
+        /// <param name="chunk">The chunk of bytes to be encoded.</param>
+        /// <param name="isFinalBlock">A boolean value indicating whether the chunk is the final block of data.</param>
+        /// <param name="prevBytesNotWritten">An out parameter indicating the number of bytes from the previous chunk that were not processed.</param>
+        private void Base64EncodeAndWriteChunk(ReadOnlySpan<byte> chunk, bool isFinalBlock, out int prevBytesNotWritten)
+        {
+            int encodingLength = Base64.GetMaxEncodedToUtf8Length(chunk.Length);
+            var output = this.bufferWriter.GetSpan(encodingLength);
+            prevBytesNotWritten = 0;
+
+            OperationStatus status = Base64.EncodeToUtf8(chunk, output, out int consumed, out int written, isFinalBlock);
```
habbes, line 768: We should assert the that the operation status is `Done` or `NeedMoreData`

habbes (commented):
Looks good to me except for the concern I raised regarding the expected operation status for utf-8 encoding non-final string chunks.

Also would be good to see some benchmark results, both on CPU, latency on concurrent requests and memory allocations.

When running the benchmark `TestServer`, you can use the `&largeFieldValues=true` query option to return results with large fields. In this case, you may consider reducing the count to avoid exploding the response size.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -364,34 +372,406 @@ public void WriteValue(sbyte value)
...
+
+            if (IsInArray() && !isWritingAtStartOfArray)
+            {
+                // Place a separator before the raw value if
+                // this is an array, unless this is the first item in the array.
+                this.bufferWriter.Write(itemSeparator.Slice(0, 1).Span);
+            }
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+            int chunkSize = 2048;
```
xuzhg, line 415: can we config it?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -364,34 +372,406 @@ public void WriteValue(sbyte value)
 
         public void WriteValue(string value)
         {
-            this.WriteSeparatorIfNecessary();
             if (value == null)
             {
+                this.WriteSeparatorIfNecessary();
                 this.utf8JsonWriter.WriteNullValue();
             }
+            else if (value.Length > bufferWriter.FreeCapacity)
```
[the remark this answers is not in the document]
ElizabethOkerio: can we do 3/4?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: I don't think this is suitable to expose to the user because the chunking is an implementation detail. I also think it will be too many config options. Maybe we can delay exposing it unless people expect it. Right now we are already exposing the overall buffer size. Should we also expose the chunkSize for strings and also chunkSize for byte arrays as separate options. I think that would confuse the user.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -364,34 +372,406 @@ public void WriteValue(sbyte value)
...
+            }
+
+            this.isWritingAtStartOfArray = false;
+        }
+
+        /// <summary>
+        /// Writes the byte value represented by the provided ReadOnlySpan in chunks using Base64 encoding.
+        /// </summary>
+        /// <param name="value">The ReadOnlySpan containing the byte value to be written.</param>
+        private async ValueTask WriteByteValueInChunksAsync(ReadOnlyMemory<byte> value)
```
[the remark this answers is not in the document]
ElizabethOkerio: I've moved this to the async methods region.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
ElizabethOkerio: Updated this to include `OperationStatus.NeedMoreData`.

habbes:
@ElizabethOkerio Kindly add a description to the PR

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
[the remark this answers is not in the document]
habbes: Maybe we should place an upper bound, like:

```c#
if (value.Length > bufferWriter.FreeCapacity || value.Length > someThreshold
```
Where that `someThreshold` could be something like 12k if we are okay with some buffer resize but want to avoid LOH. Or it could be something small like 2048 if we want to avoid buffer resize more aggressively.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -364,34 +372,266 @@ public void WriteValue(sbyte value)
...
+        {
+            this.CommitUtf8JsonWriterContentsToBuffer();
+
+            WriteItemWithSeparatorIfNeeded();
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+
+            this.Flush();
+
+            const int chunkSize = 2048;
```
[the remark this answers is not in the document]
habbes: Also should be probably a constant as well. And if you pick an uncommon value like 1,250, you explain why in the comments.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -797,34 +1063,138 @@ public async Task WriteValueAsync(sbyte value)
...
+        {
+            this.CommitUtf8JsonWriterContentsToBuffer();
+
+            WriteItemWithSeparatorIfNeeded();
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
+
+            await FlushAsync().ConfigureAwait(false);
+
+            const int chunkSize = 4096;
```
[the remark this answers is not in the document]
habbes: That said, even if it resizes, it would likely only happen once for the entire payload being written, and size the buffer is rented, the impact may be tolerable. So may not need to stress too much about it. I would suggest running the allocation benchmarks with different chunk sizes and pick one. We can change it later if we get more real world data. Some overhead is to be expected when writing large values.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataUtf8JsonWriterTests.cs
habbes: None of the characters in this string are "special chars" (they won't be escaped). See: https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYAMBYAUBgRj1UwAJUCA6AFQFMAPYSgKQGcB7KAbj19wDcAhmFKtSAXlJRaAd3IFMACgDk9ZTFIKAlKUSkARAEJUqAMQASAKT7dU2fKWr1mzFp4DhokZLacAyrRgAJaCADZBAF6BlAHBYZG0iqxueEIiDAAOtADGwLRwEqTm+gA6+gDerAC+ZfruFACcSSn4BE2sYC2NisUAIkEFtKzZghlBUADmpOzZ2RBgIKSVIoaSmTl5cFX6LUA

So this test does not do what it intends (based on the name at least). Consider adding special chars in there (e.g. double quote, control characters like new lines, non ascii characters).

ElizabethOkerio: Yes.I'll update it.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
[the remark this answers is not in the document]
ElizabethOkerio: Used a chunksize of `2048`

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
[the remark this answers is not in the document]
ElizabethOkerio: Here too. Settled for a chunksize of 2048..

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -364,34 +373,262 @@ public void WriteValue(sbyte value)
...
+        /// excessive memory allocations and buffer overflows. If the string requires escaping, it is
+        /// processed accordingly to ensure correct JSON formatting.
+        /// </remarks>
+        private void WriteStringValueInChunks(ReadOnlySpan<char> value)
+        {
+            this.CommitUtf8JsonWriterContentsToBuffer();
+
+            WriteItemWithSeparatorIfNeeded();
+
+            this.bufferWriter.Write(doubleQuote.Slice(0, 1).Span);
```
xuzhg, line 409: I can't understand why you define readonly ReadOnlyMemory<byte> for double Quote?

xuzhg, line 409: since you define the field without '_', so, please use 'this.' prefix to reference the class field?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -364,34 +373,262 @@ public void WriteValue(sbyte value)
...
+        private unsafe int NeedsEscaping(ReadOnlySpan<char> value)
+        {
+            // Some implementations of JavaScriptEncoder.FindFirstCharacterToEncode may not accept
+            // null pointers and guard against that. Hence, check up-front to return -1.
+            if (value.IsEmpty)
+            {
+                return -1;
+            }
+
+            fixed (char* ptr = value)
```
xuzhg, line 462: fixed or unsafe ?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -364,34 +373,262 @@ public void WriteValue(sbyte value)
...
+        /// according to the specified encoder. It allocates memory for the escaped string and handles
+        /// buffer management using either stack allocation or array pooling. Once the string is escaped,
+        /// it is written into the buffer. If the encoding operation fails, an exception is thrown. 
+        /// </remarks>
+        private void WriteEscapedStringChunk(ReadOnlySpan<char> sourceChunk, int firstIndexToEscape, bool isFinalBlock, out int preChunkNotProcessed)
+        {
+            char[] valueArray = null;
+            const int StackallocCharThreshold = 128;
+
+            int maxSize = firstIndexToEscape + 6 * (sourceChunk.Length - firstIndexToEscape);
```
xuzhg, line 485: magic number '6' here needs comments. or same as line 483 to define a const int

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -364,34 +373,262 @@ public void WriteValue(sbyte value)
...
+
+            Span<char> destination = maxSize <= StackallocCharThreshold
+                ? stackalloc char[maxSize]
+                : valueArray = ArrayPool<char>.Shared.Rent(maxSize);
```
xuzhg, line 489: We have ICharArrayPool (maybe different name), why can't use that?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -364,34 +373,262 @@ public void WriteValue(sbyte value)
...
+        /// before the next value in the JSON output. It sets the flag indicating whether a separator should
+        /// be written based on conditions such as whether the writer is at the start of an array, writing
+        /// consecutive raw values at the start of an array, or if the writer is not currently in an array.
+        /// </remarks>
+        private void CheckIfSeparatorNeeded()
+        {
+            if (this.isWritingAtStartOfArray || this.isWritingConsecutiveRawValuesAtStartOfArray || !this.IsInArray())
+            {
+                this.shouldWriteSeparator = true;
+            }
```
xuzhg, line 551: no else clause to set the 'shouldWriteSeparator' = false; ?

xuzhg: can we find a better way to do it? 
It seems you have to change other methods to 'check shouldWriteSeparator' first, then do other things?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: [`bufferWriter.Write`](https://learn.microsoft.com/en-us/dotnet/api/system.buffers.buffersextensions.write?view=net-8.0#system-buffers-buffersextensions-write-1(system-buffers-ibufferwriter((-0))-system-readonlyspan((-0)))) method expects a `Span<T>` as input. We can create the span either by allocating a `new byte[] { (byte)'"'}`  each time, or we can allocate it once as a constant. In this case it has been created once and cached as a constant to avoid re-allocating each time. Another alternative would have been to allocate it on the stack using `stackallock byte[]{(byte)'"'}`.

But if we store it a readonly static field, we cannot store it as a `Span<T>` because `Span<T>` is a ref struct, so we use a `ReadOnlyMemory<T>` instead.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: This method is already in an `unsafe` block. `fixed` here is used to prevent the garbage collector from moving the object during garbage collection or compaction since we're storing its address directly in the pointer. Unfortunately, the [TextEncoder.FindFirstCharacterToEncode](https://learn.microsoft.com/en-us/dotnet/api/system.text.encodings.web.textencoder.findfirstcharactertoencode) does not provide an overload that accepts arrays, so we are forced to use pointers.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: I also suggest creating a well document constant to store that number.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: Our `ICharArrayPool` is only an interface. We don't have any default implementation. I also don't know any of our customers who have actually implemented this interface. So, it remains mostly unused. I think that interface was created before dotnet added standard array pools. I think it's better to use .NET's standard array pools than to create our own.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: Yes, I also wish we can find a better away. I introduced these separator flags in a previous PR. The challenge here is that we are bypassing `Utf8JsonWriter`. `Utf8JsonWriter` does not know that we have written these values, and so it will not write the separate when we call it next time. The result is that we'll end up with invalid JSON. To avoid that issue, we keep track of whether to write the separator ourselves when we write to the buffer manually. This means we have to coordinate with other methods of this class to write the separator before they write the value. It's not ideal, but it's the workaround we have at the moment. We use extensive tests to ensure we do this correctly.

The best way around this is for `Utf8JsonWriter` to provide these features out of the box so that we don't have to implement them ourselves. .NET 6+ added a native implementation of `WriteRawValue`, so we can use that when we upgrade. However they've not provided an ETA for when they'll address the issue with large values and chunking, but they have an open issue: https://github.com/dotnet/runtime/issues/67337

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -364,34 +373,262 @@ public void WriteValue(sbyte value)
...
+        /// chunk of the string.
+        /// </remarks>
+        private void WriteStringChunk(ReadOnlySpan<char> chunk, bool isFinalBlock)
+        {
+            int maxUtf8Length = Encoding.UTF8.GetMaxByteCount(chunk.Length);
+
+            Span<byte> output = bufferWriter.GetSpan(maxUtf8Length);
+            OperationStatus status = Utf8.FromUtf16(chunk, output, out int charsRead, out int charsWritten, isFinalBlock);
+            Debug.Assert(status == OperationStatus.Done || status == OperationStatus.NeedMoreData);
+            Debug.Assert(charsRead == chunk.Length);
```
habbes, line 531: Is `charsRead` guaranteed to be `chunk.Length`? Is it guaranteed that there can be no partial processing of chunks especially for non-final blocks? We do make this consideration for encoding to base64, not sure why we don't do make the same consideration for string to utf8 encoding.

habbes: To be clear, I'm not sure if that scenario can occur for utf-16 to utf-8 transcoding, but I would request that you confirm first and if it's not an issue for this scenario, then make a comment to explain that.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -437,12 +674,38 @@ private void WriteRawValueCore(string rawValue)
...
+            CheckIfSeparatorNeeded();
+
+            if (this.isWritingAtStartOfArray || this.isWritingConsecutiveRawValuesAtStartOfArray)
             {
-                this.shouldWriteSeparator = true;
+                this.isWritingConsecutiveRawValuesAtStartOfArray = true;
             }
 
-            if (this.isWritingAtStartOfArray || this.isWritingConsecutiveRawValuesAtStartOfArray)
+            this.isWritingAtStartOfArray = false;
```
habbes, line 684: Replace this block with `this.CheckIfManualValueAtArrayStart()`.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -437,12 +674,38 @@ private void WriteRawValueCore(string rawValue)
...
+                // Place a separator before the raw value if
+                // this is an array, unless this is the first item in the array.
+                this.bufferWriter.Write(itemSeparator.Slice(0, 1).Span);
+            }
+        }
+
+        // <summary>
+        /// Checks if the writer is at the start of a manual value array and updates the state accordingly.
+        /// </summary>
+        private void CheckIfAtManualValueArrayStart()
```
habbes, line 706: nit:
```suggestion
        private void CheckIfManualValueAtArrayStart()

```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: This might be useful: https://learn.microsoft.com/en-us/dotnet/standard/base-types/character-encoding-introduction

habbes: I think you should add tests with strings that contain surrogate pairs at chunk boundaries.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataUtf8JsonWriterTests.cs
```diff
@@ -169,6 +169,24 @@ public void WritePrimitiveValueString()
...
+
+        [Fact]
+        public void WritePrimitiveValueLargeStringWithSpecialChars()
+        {
+            this.VerifyWritePrimitiveValue(
+                new string('x', 20000) + "Foo ия" + new string('x', 10000),
+                "\"" + new string('x', 20000) + "Foo \\u0438\\u044F" + new string('x', 10000) + "\""
+                );
+        }
+
```
habbes, line 189: I think you should add tests with strings that contain surrogate pairs at chunk boundaries. E.g. when one half of the surrogate pair is at the end of one chunk, and the other half is at the beginning of the other half.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
ElizabethOkerio: I've defined the field with `_`

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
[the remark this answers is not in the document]
ElizabethOkerio: I set a threshold of 2048.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
ElizabethOkerio: done

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
ElizabethOkerio: I think we take care of this when we have characters to escape in the chunk... I'm not sure whether we need to do that here too..

ElizabethOkerio: Let me add tests with surrogate pairs and see how it behaves.

habbes: I have done some [experiments](https://sharplab.io/#v2:C4LgTgrgdgNAJiA1AHwAICYAMBYAUBgRj1UwAJUCA6AFQFMAPYSgUSgGMB7OASygHMAzpQDqtAEYBuYmQo0GTAFICOUKfhlU6jSgFUo3TnFpriBAJwAKAESAeDcAg+1coAZWv2AALAJQncANwCGYKQcAA7A3CoCpAC8pFC0AO6kSioAyrRg3P4ANtwAXhkA8mERUAJ4AN54pDWshhkxyf4BqWyZYXVcGbpl/gBmtABKtNn+9LRwKVDMAmz+Ibx8eAC+ahSWU+mZOfndW1m5Bdb2VjDBJZGe3nh4APS3AUHKEGBstI22DmqPpM+v77ErIBADcA8ztAwAbO0DAN87EOhQMA+zukIGAbp2EVY1Pcfq56kFYlpgJ0jGA7g9AqRsV1cU0Wm1uB12JSegJ+kMRmMJlMZnMFvxvmSBLRgOF+FFYvEkvjCRl0kLFgILNdcALZSLKABBbLZDgJQb+fi0eV6AxdXX6oQa7KK5XCwTqzXagDC7kC/jYwAy8oA5AAdCAAEQAHJgCJ6zj7/Q6HQAxT2KrEMomNBTNfytdoEhPdB1gWj+d0Wa1yuNkjgQYAhMuNcWkNjOsAAbQAbKQAFS/Uv/ZyuPgeAC66oEqRCeoVfKCOYEEGywEaFKJLEzBY7bwHQ5HnkoqVybwsmDOZk8Z1L5bLR8rvBnnDKEAAthMzzOL6QEpkha4ztwBFHeDkAEJatgAGsQFIPocgFYsggNbkJkaY8KyYLcDFoXczhfbg3ygDdqA4VJgEyfhR1MSwABIrAqCcp2AZYzivSc7zgM4Kjo28Jho59X3dWBSAqdDMOWKxFXWCxoPmCZFTwH4jAEYU81KKtElIMQAE93XrABmVtgjLBCuzcdx+zVQdhygIjcEo6dGh0YA+gDSgozADgb2svoCEbCx4LLTdtxQghD1IaTZPCFQHxrSJWMY7SZz4riPy/H9sn/DggJAsDsggtZzAsMiKINKj2JYhimMKti0M498eJi1wBMVPoOCCCwn24RpMAkUhmoAHg4jCuLa7hEEQTxKmqGphIsWtAk8QKf2CqB624XtvBqUgViAA===) with encoding to try and create a scenario that causes Utf16->utf8 transcoding to result in `NeedMoreData` status, but I think you may be right.

The characters that would cause utf-8 encoding to result in partial processing seem to be already escaped away by the javascriptencoder and the result is written utf8-encoded in full. I have tried different JavaScriptEncoder settings, including allowing specific characters to go unescaped, but the result is the same, the resulting escaped substring is always encoded to utf8 in full.

Maybe you could add a comment to the code to explain why the concerns I raised are not an issue.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -86,7 +99,10 @@ internal sealed class ODataUtf8JsonWriter : IJsonWriter, IDisposable, IJsonWrite
         /// an array e.g. ["rawValue1", "rawValue2", "rawValue3"
         /// </summary>
         private bool isWritingConsecutiveRawValuesAtStartOfArray = false;
-        
+        /// <summary>
+        /// Rrepresents a read-only memory block containing the byte representation of the double quote character ("). 
+        /// </summary>
+        private readonly static ReadOnlyMemory<byte> _doubleQuote = new byte[] { (byte)'"' };
```
habbes, line 105: nit: We use `this.` instead of `_` for fields. Maybe for constants and statics we should be consistent with `PascalCase`
```suggestion
        private readonly static ReadOnlyMemory<byte> doubleQuote = new byte[] { (byte)'"' };

```

habbes (dismissed):

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
ElizabethOkerio: I think since it's a static you can't use `this.` I remove the static qualifier then.

habbes (approved):

PR #2849, opened 2024-01-26
Create test cases and benchmarks for writing large json base64 and string values

habbes:
### Issues

This PR partially addresses #2845

### Description

It adds test cases that ensures writing large byte arrays and string values with to the JSON writer works correctly. Once merge, this will help verify the optimizations introduced by PR #2846 are correct.

The PR also adds benchmarks with large string and `byte[]` values. I also removed benchmark scenarios that are not necessary, like:
- Utf-16 benchmarks (Utf-16 as the output encoding is not common)
- Writing directly with `JsonWriter` or `Utf8JsonWriter` without using `ODataMessageWriter`

Those benchmark scenarios don't add much value, but they add to the run time of the benchmarks and maintenance overhead.

- [x] *Test cases added*


test/PerformanceTests/SerializationComparisonsTests/Lib/DataModel.cs
```diff
@@ -14,6 +19,11 @@ public class Customer
         public int Id { get; set; }
         public string Name { get; set; }
         public List<string> Emails { get; set; }
+        public string Bio { get; set; }
+
+        [JsonConverter(typeof(Base64Converter))]
```
gathogojr, line 24: Curious why we need a JsonConverter here

test/PerformanceTests/SerializationComparisonsTests/JsonWriterBenchmarks/Benchmarks.cs
```diff
@@ -65,29 +71,30 @@ public async Task WriteToMemoryAsync()
...
+        public async Task WriteToFileWithLargeValuesAsync()
+        {
+            await WritePayloadAsync(dataWithLargeValues);
+        }
+
+        private async Task WritePayloadAsync(IEnumerable<Customer> payload, bool includeRawValues = false)
         {
-            using var outputStream = new FileStream(filePath, FileMode.Create, FileAccess.Write, share: FileShare.ReadWrite);
-            await writer.WritePayloadAsync(data, outputStream, includeRawValues);
+            //using var outputStream = new FileStream(filePath, FileMode.Create, FileAccess.Write, share: FileShare.ReadWrite);
```
gathogojr, line 96: Why did you need to comment this statement instead of deleting it?

gathogojr (approved):
Left a few comments; otherwise LGTM

test/PerformanceTests/SerializationComparisonsTests/JsonWriterBenchmarks/Benchmarks.cs
```diff
@@ -65,29 +71,30 @@ public async Task WriteToMemoryAsync()
         public async Task WriteToFileAsync()
         {
             // multiple writes to increase benchmark duration
```
ElizabethOkerio, line 73: This comment needs to be updated also?

test/PerformanceTests/SerializationComparisonsTests/JsonWriterBenchmarks/Benchmarks.cs
```diff
@@ -65,29 +71,30 @@ public async Task WriteToMemoryAsync()
...
-            await WritePayloadAsync();
+            await WritePayloadAsync(data);
+            await WritePayloadAsync(data);
         }
 
         [Benchmark]
         [BenchmarkCategory("RawValues")]
         public async Task WriteWithRawValues()
         {
             // multiple writes to increase benchmark duration
```
ElizabethOkerio, line 82: nit: here to..

test/PerformanceTests/SerializationComparisonsTests/Lib/ODataMessageWriterAsyncPayloadWriter.cs
```diff
@@ -35,6 +35,7 @@ public ODataMessageWriterAsyncPayloadWriter(IEdmModel model, Func<Stream, IOData
         public async Task WritePayloadAsync(IEnumerable<Customer> payload, Stream stream, bool includeRawValues)
         {
             ODataMessageWriterSettings settings = new ODataMessageWriterSettings();
+            settings.EnableMessageStreamDisposal = false;
```
ElizabethOkerio, line 38: why did you add this setting?

ElizabethOkerio (approved):

test/PerformanceTests/SerializationComparisonsTests/JsonWriterBenchmarks/Benchmarks.cs
habbes: Two writes are still multiple. I think the intent behind the comment still stands.

test/PerformanceTests/SerializationComparisonsTests/JsonWriterBenchmarks/Benchmarks.cs
habbes: See: https://github.com/OData/odata.net/pull/2849/files#r1472336559

test/PerformanceTests/SerializationComparisonsTests/JsonWriterBenchmarks/Benchmarks.cs
habbes: Oversight. Thanks.

test/PerformanceTests/SerializationComparisonsTests/Lib/DataModel.cs
habbes: `JsonSerializer` does not seem to serialize `byte[]` fields by default.

test/PerformanceTests/SerializationComparisonsTests/Lib/ODataMessageWriterAsyncPayloadWriter.cs
habbes: When this setting is set to true and dispose `ODataMessageWriter`, it will dispose the output stream as well. I disabled it because I wanted to use the output stream after finishing the write. For example, in the unit tests for these benchmarks I rewind the `MemoryStream` after the writer has completed so that I can read the contents and verify that it wrote the right thing.

ElizabethOkerio (approved):

PR #2851, opened 2024-01-30
Update target frameworks to net8.0 and net48

habbes:
### Description

Sets .net8.0 as the target framework for ODL 8. Since we've not reached a conclusion regarding .NET FX support in ODL 8, I also added .net48 as the target framework.

I removed all other target frameworks from library and test projects. I update the code and tests such that all the tests pass. Most changes were changing conditional compiler directives like `#if NETCOREAPP3_1` and `#if NETSTANDARD` to `#if NETCOREAPP`. `NETCOREAPP` is set for all versions of .NET Core, including 8.x and above.

In some minor cases I had to change test cases to pass. For example, in `HttpClientRequestMessageTests.Abort_CanclesPendingRequest` and `HttpClientRequestMessageTests.Timeout_CancelsPendingRequestAfterTimeout`, the exception thrown in .net48 is `OperationCanceledException` whereas in .net8.0 the exception is `DataServiceTransportException`.

The E2E test projects were all target `.net45`, none was target `.netstandard` or `.netcoreapp`. They also rely on old .NET FX only APIs like `System.Data`. I changed their target to `.net48`. We should plan to port/rewrite them from `.NET 8`. It's not good not to have E2E tests for the primary framework we're targeting.

I removed `FxCopRules` from the e2e solution because it was failing to build and because we had long talked about dropping FxCopRules in favour of roslyn analyzers anyway.

I removed `PublicApiTests` from solutions. They caused some errors and we don't need them anyway since we have `PublicApiAnalyzers`.

I didn't manage to get E2E tests to run in the pipeline. I am skipping them for now by changing the test solution from `OData.Pipeline.sln` (which includes e2e tests) to `OData.sln` which only includes unit tests.  I have created an issue to track fixing e2e tests: https://github.com/OData/odata.net/issues/2852


robertmclaws:
[This official Microsoft documentation](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core) says that .NET 8.0, 7.0, and 6.0 are the current officially-supported versions, with .NET 7.0 support ending in May and .NET 6.0 support ending in November.

The adjustments here are a good start, but the OData team needs to be shipping against the currently-supported versions.

**@habbes It also means the correct required compilation constant is `NET6_0_OR_GREATER`**, not NETCOREAPP

habbes:
@robertmclaws we plan to better align future releases with .NET releases, at least the LTS versions. We don't have fixed ETA for ODL 8 release, but it will probably be out before the end of FY24 (next quarter). That means there's likelihood that it's released after .NET 7 has gone out support. ODL 8 will also have longer lifetime than .NET 6. To avoid the same issue we are in (where we can't drop support for unsupported frameworks like .NET Core 3.1, 2.1, .NET FX 4.5, etc.) because it would be a breaking change, we find it more reasonable not to support .NET 6 given that it's going out of support this year. We're still undecided on .NET FX support.

Furthermore, while .NET 6 will still be supported after ODL 8 is release, it will probably only have <= 6 months of support left. The [official Microsoft support documentation](https://dotnet.microsoft.com/en-us/platform/support/policy/dotnet-core) state that the last 6 months of support are maintenance mode and .NET releases are updated to mitigate security vulnerabilities, only.

That said, ODL 7 will continue to be supported for the time being. And it supports .NET 6, 7 and as well as older versions.

As for the use of `NETCOREAPP` instead of `NET_6_0_OR_GREATER`, since this PR changes the explicit targets to `.net48` and `.net8.0`, the `NETCOREAPP` is sufficient to distinguish between code that only runs on modern .NET and classic .NET FX. Using `.NETCOREAPP` means we don't need to update the directive for future .NET versions. However, if we target multiple .NET versions and some code is only available for one version, we can use something like `.NET_X_0_OR_GREATER` for code that only works on a specific subset of versions. My 2 cents.

At the moment, this PR is just laying down the groundwork for ODL 8 work. But there's some planning still going on. We'll share more details with the community soon to get your feedback on the direction and changes planned for ODL 8. Thanks for sharing your thoughts on this.

src/Microsoft.OData.Client/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,1030 @@
+﻿abstract Microsoft.OData.Client.ALinq.UriParser.PathSegmentToken.Accept(Microsoft.OData.Client.ALinq.UriParser.IPathSegmentTokenVisitor visitor) -> void
```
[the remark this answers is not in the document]
habbes: Because we have not yet shipped.

src/Microsoft.OData.Edm/PublicAPI/net48/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,1929 @@
+abstract Microsoft.OData.Edm.EdmModelBase.DeclaredNamespaces.get -> System.Collections.Generic.IEnumerable<string>
```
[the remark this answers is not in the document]
habbes: Because we have not yet shipped.

src/Microsoft.OData.Edm/PublicAPI/net8.0/PublicAPI.Unshipped.txt
```diff
@@ -0,0 +1,1953 @@
+﻿abstract Microsoft.OData.Edm.EdmModelBase.DeclaredNamespaces.get -> System.Collections.Generic.IEnumerable<string>
```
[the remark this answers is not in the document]
habbes: Because we have not yet shipped.

src/Microsoft.Spatial/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,372 @@
-﻿
\ No newline at end of file
+﻿abstract Microsoft.Spatial.Geography.IsEmpty.get -> bool
```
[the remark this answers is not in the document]
habbes: Because we have not yet shipped.

src/Microsoft.OData.Edm/Microsoft.OData.Edm.csproj
```diff
@@ -86,7 +86,8 @@
   </ItemGroup>
 
   <ItemGroup>
-    <AdditionalFiles Include="PublicAPI\$(TargetFramework)\**" />
+    <AdditionalFiles Include="PublicAPI/$(TargetFramework)/PublicAPI.Shipped.txt" />
+    <AdditionalFiles Include="PublicAPI/$(TargetFramework)/PublicAPI.Unshipped.txt" />
```
ElizabethOkerio, line 90: Are there other files in the `PublicAPI/$(TargetFramework)`.

ElizabethOkerio: If at some point these files get renamed then someone has to update them here... Also, if more files are added to the directory then they have to be added individually?

src/CodeGen/Microsoft.Data.Web.Design.T4.csproj
```diff
@@ -7,7 +7,7 @@
     <ConditionalAPTCA_L2>true</ConditionalAPTCA_L2>
     <DocumentationFile>$(AssemblyName).xml</DocumentationFile>
     <RestorePackages>true</RestorePackages>
-    <TargetFrameworks>net45</TargetFrameworks>
+    <TargetFrameworks>net48</TargetFrameworks>
```
ElizabethOkerio, line 10: Can we get rid of the codegen project in .net 8.0?

habbes: I think we can. But that could be done in a separate PR. The focus of this PR was to change the target frameworks and I only made the changes necessary to get that to work. Other changes are beyond the scope of what this PR was trying to achieve.

src/Microsoft.OData.Edm/Microsoft.OData.Edm.csproj
habbes: These are the files expected by the Public API analyzer. See: https://github.com/dotnet/roslyn-analyzers/blob/main/src/PublicApiAnalyzers/PublicApiAnalyzers.Help.md#conditional-api-differences

ElizabethOkerio: I thought having `<AdditionalFiles Include="PublicAPI\$(TargetFramework)\**" /> ` was sufficient for it to work

sln/OData.E2E.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.30011.22
+# Visual Studio Version 17
+VisualStudioVersion = 17.8.34511.84
```
gathogojr, line 3: Did we need to change VS version?

sln/OData.Pipeline.sln
```diff
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.30011.22
+# Visual Studio Version 17
+VisualStudioVersion = 17.8.34511.84
```
gathogojr, line 3: Did we need to change VS version?

src/Microsoft.OData.Client/Microsoft.OData.Client.csproj
```diff
@@ -1,14 +1,14 @@
...
+    <TargetFrameworks>net8.0;net48</TargetFrameworks>
     <AssemblyName>Microsoft.OData.Client</AssemblyName>
     <RootNamespace>Microsoft.OData.Client</RootNamespace>
     <DocumentationFile>$(AssemblyName).xml</DocumentationFile>
     <DefineConstants>$(DefineConstants);ODATA_CLIENT;SUPPRESS_PORTABLELIB_TARGETFRAMEWORK_ATTRIBUTE;DelaySignKeys</DefineConstants>
    <GenerateAssemblyInfo>false</GenerateAssemblyInfo>
   </PropertyGroup>
 
-  <PropertyGroup Condition="'$(TargetFramework)' == 'netstandard2.0'">
+  <PropertyGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
gathogojr, line 11: Looking at what is nested within this element, I don't think we need this `PropertyGroup` at all

src/Microsoft.OData.Client/PublicAPI/net48/PublicAPI.Shipped.txt
```diff
@@ -0,0 +1,1030 @@
+Microsoft.OData.Client.UriFunctionAttribute
```
gathogojr, line 1: How come your comment here https://github.com/OData/odata.net/pull/2851/files#r1482628586, doesn't apply to this Shipped.txt file?

sln/OData.E2E.sln
habbes: Was changed automatically by Visual Studio. Should I reset it back?

src/Microsoft.OData.Client/Microsoft.OData.Client.csproj
habbes: Okay. I was reluctant to remove things I was not sure about. But if you're sure it's not used I can go ahead and remove it.

src/Microsoft.OData.Client/PublicAPI/net8.0/PublicAPI.Shipped.txt
```diff
@@ -0,0 +1,1030 @@
+Microsoft.OData.Client.UriFunctionAttribute
```
gathogojr, line 1: How come your comment here https://github.com/OData/odata.net/pull/2851/files#r1482628586 does apply to this Shipped.txt file?

src/Microsoft.OData.Client/PublicAPI/net48/PublicAPI.Shipped.txt
habbes: You and @KenitoInc are right. I had not noticed that some are in shipped and some in unshipped. I'll correct this.

src/Microsoft.OData.Client/PublicAPI/net8.0/PublicAPI.Shipped.txt
habbes: See: https://github.com/OData/odata.net/pull/2851#discussion_r1484004611

src/Microsoft.OData.Client/PublicAPI.Unshipped.txt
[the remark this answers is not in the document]
habbes: Thanks for the clarification. I'll fix the issue.

src/Microsoft.OData.Edm/PublicAPI/net48/PublicAPI.Unshipped.txt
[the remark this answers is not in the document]
habbes: https://github.com/OData/odata.net/pull/2851#discussion_r1484007403

src/Microsoft.OData.Edm/PublicAPI/net8.0/PublicAPI.Unshipped.txt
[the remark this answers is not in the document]
habbes: https://github.com/OData/odata.net/pull/2851#discussion_r1484007403

src/Microsoft.Spatial/PublicAPI.Unshipped.txt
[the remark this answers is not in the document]
habbes: https://github.com/OData/odata.net/pull/2851#discussion_r1484007403

src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
```diff
@@ -95,13 +91,14 @@
     </PackageReference>
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
gathogojr, line 94: You should confirm whether a higher version of Microsoft.Extensions.ObjectPool is available for net8.0. Or if it's even required at all

src/Microsoft.OData.Edm/Csdl/CsdlJsonReaderSettings.cs
```diff
@@ -4,7 +4,7 @@
 // </copyright>
 //---------------------------------------------------------------------
 
-#if NETSTANDARD2_0
+#if NETCOREAPP
```
gathogojr, line 7: Confirm it this compiler flag applies to net8.0 even after they drop "Core" from the framework name

habbes: It applies to all .NET Core and .NET 5+ versions. See: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/preprocessor-directives

gathogojr:
I'd like to understand the thing about the Shipped.txt and Unshipped.txt files. In some cases, the Unshipped file is populated while Shipped is empty, in other cases the reverse is true, in yet other cases both are populated

src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
habbes: Since I was focusing on just making the changes necessary to get this to work. I avoided removing packages, to avoid accidental breakages or lengthening the scope of this PR.

test/Common/Microsoft.Test.OData.DependencyInjection/Microsoft.Test.OData.DependencyInjection.csproj
```diff
@@ -12,12 +12,9 @@
 
   <Import Project="..\Build.props" />
 
-  <ItemGroup Condition=" '$(TargetFramework)' == 'net472' or '$(TargetFramework)' == 'netcoreapp2.1' or '$(TargetFramework)' == 'netcoreapp3.1' ">
+  <ItemGroup Condition=" '$(TargetFramework)' == 'net48' or '$(TargetFramework)' == 'net8.0'">
```
gathogojr, line 15: Why do we need this condition yet net48 and net8.0 are the only 2 frameworks we support

test/EndToEndTests/Tests/Client/Build.Desktop/Microsoft.Test.OData.Tests.Client.csproj
```diff
@@ -222,6 +222,10 @@
     </None>
   </ItemGroup>
 
+  <ItemGroup>
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
+  </ItemGroup>
```
gathogojr, line 227: Why this change?

test/FunctionalTests/Microsoft.OData.Core.Tests/Microsoft.OData.Core.Tests.csproj
```diff
@@ -79,19 +71,13 @@
...
   </ItemGroup>
 
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp2.1'">
-    <PackageReference Include="Microsoft.Bcl.AsyncInterfaces">
-      <Version>8.0.0</Version>
-    </PackageReference>
-  </ItemGroup>
-
-  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+  <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
gathogojr, line 74: I don't think we need this ItemGroup because we don't need to add a reference to Microsoft.Bcl.AsyncInterfaces in the case of net8.0. You should check other references to that library across the solution and remove them

test/FunctionalTests/Microsoft.OData.Edm.Tests/Microsoft.OData.Edm.Tests.csproj
```diff
@@ -38,26 +38,22 @@
...
+     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="17.8.0" />
   </ItemGroup>
    
   <ItemGroup>
     <PackageReference Include="xunit" Version="2.4.1" />
     <PackageReference Include="xunit.runner.visualstudio" Version="2.4.1" />
   </ItemGroup>
  
-   <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+   <ItemGroup Condition="'$(TargetFramework)' == 'net8.0'">
```
gathogojr, line 50: Confirm whether we need this condition in the case of net8.0. It's even possible STJ dependency comes with the framework reference

sln/OData.E2E.sln
gathogojr: @habbes I think what you should do is check if after this change you can open the solution using VS 2019. If VS 2019 works with net8.0 and .net framework 4.8, then it'd make more sense to retain it as it was.

src/Microsoft.OData.Client/Microsoft.OData.Client.csproj
gathogojr: @habbes What is nested is `<DefineConstants>$(DefineConstants);ASSEMBLY_ATTRIBUTE_ON_NETSTANDARD_20</DefineConstants>`. Clearly, this is only relevant to NetStandard 2.0 so it wouldn't have any relevance after we remove support for that standard.

src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
gathogojr: @habbes Especially in the case of Microsoft.Bcl.AsyncInterfaces that I flagged elsewhere, my concern is that if it's unneeded and we fail to remove it now, we might not have revisit it again to clean up

src/Microsoft.OData.Edm/Microsoft.OData.Edm.csproj
habbes: But the files have to exist with those names.

test/EndToEndTests/Tests/Client/Build.Desktop/Microsoft.Test.OData.Tests.Client.csproj
habbes: Hmm. I don't know what this is. Must have been added by the IDE while I was changing things. Let me see what happens if I remove it.

src/Microsoft.OData.Client/PublicAPI.Unshipped.txt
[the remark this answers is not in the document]
habbes: I fixed the issue. I moved all the API declarations everything to `Unshipped.txt` files and removed duplicates.

src/Microsoft.OData.Client/Microsoft.OData.Client.csproj
habbes: Removed.

src/Microsoft.OData.Client/PublicAPI/net48/PublicAPI.Shipped.txt
habbes: Fixed. Moved all the API declarations to `Unshipped.text` files.

src/Microsoft.OData.Client/Microsoft.OData.Client.csproj
habbes: Removing this caused compilation errors. There's some code which depends on this flag. Maybe we can revisit this later?

habbes: I could create a work item to perform clean up after this so that this PR doesn't block other work from taking place.

src/Microsoft.OData.Core/Microsoft.OData.Core.csproj
habbes: Removing `Microsoft.Extensions.ObjectPool` reference causes a compilations error. It's not built-in to .NET. It's documentation is not listed as part of .NET but rather in .NET Platform Extensions: https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.objectpool?view=dotnet-plat-ext-8.0

habbes: However, removing the `Bcl.AsyncInterfaces` references seems to cause no issues.

test/FunctionalTests/Microsoft.OData.Core.Tests/Microsoft.OData.Core.Tests.csproj
habbes: Removed.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Microsoft.OData.Edm.Tests.csproj
habbes: Removed.

test/EndToEndTests/Tests/Client/Build.Desktop/Microsoft.Test.OData.Tests.Client.csproj
habbes: Removed.

sln/OData.Pipeline.sln
habbes: I think this version refers to the last Visual Studio version that modified the file. It does not change the minimum Visual Studio version (see line 4). This file references Version 16 because we had never opened this solution in VS 2022 since this is mostly used in the build pipeline. But if you look at the `OData.sln` which we work with during development, the visual studio version is also 17.

sln/OData.E2E.sln
habbes: I think this version refers to the last Visual Studio version that modified the file. It does not change the minimum Visual Studio version (see line 4). This file references Version 16 because we had never opened this solution in VS 2022 since this is mostly used in the build pipeline. But if you look at the OData.sln which we work with during development, the visual studio version is also 17.

Also, Visual Studio 19 is officially unsupported for .NET 6+ (although there may exist some workarounds). Previously, our solutions did not explicitly target .NET 6 even though the produced dlls could run on .NET 6.

gathogojr (approved):
LGTM

ElizabethOkerio (approved):

PR #2858, opened 2024-02-16
ODL 8.x - Deprecate support for JsonPCallback

### LINKED ISSUE #2478: Remove JSONP support
JSONP has potential security issues, and there are performance issues in some cases.

gathogojr:
### Issues

*This pull request fixes #2478.*

### Description

Deprecate support for JsonPCallback

We currently wrap a response will a callback function if the callback function is specified using the `JsonPCallback` property of `ODataMessageWriterSettings` class. Owing to the decision to ensure that in 8.x customers can still write the same payload they did in 7.x, I opted for decorating the `JsonPCallback` property with `Obsolete("This will be dropped in 9.x release")` attribute. No additional value would be derived from introducing a compatibility flag for the customer to toggle since they can express that intention by initializing the `JsonPCallback` property or leaving it uninitialized.

This pull request also marks as obsolete the following methods defined in `IJsonWriter` and `IJsonWriterAsync`:
- `StartPaddingFunctionScope`
- `EndPaddingFunctionScope`
- `WritePaddingFunctionName`
- `StartPaddingFunctionScopeAsync`
- `EndPaddingFunctionScopeAsync`
- `WritePaddingFunctionNameAsync`

Dropped `StartJsonPaddingIfRequiredAsync` and `EndJsonPaddingIfRequiredAsync` because they add little/no value. They are only called from `WritePayloadStartAsync` and `WritePayloadEndAsync`. Their use introduces additional overheads in the form of async state machine and deeper call stack.

Replaced `TaskUtils.Completed` with `Task.CompleteTask` since `Task.Completed` is supported by the two frameworks that 8.x supports.

- [x] *Test cases added*

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataJsonWriterUtilsAsyncTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataJsonWriterUtilsAsyncTests.cs
@@ -21,56 +21,13 @@ namespace Microsoft.OData.Tests.Json
     {
         private StringWriter stringWriter;
         private IJsonWriterAsync jsonWriter;
-        private ODataMessageWriterSettings settings;
         private Func<ICollection<ODataInstanceAnnotation>, Task> writeInstanceAnnotationsDelegate;
 
         public ODataJsonWriterUtilsAsyncTests()
         {
             this.stringWriter = new StringWriter();
             this.jsonWriter = new JsonWriter(this.stringWriter, isIeee754Compatible: true);
-            this.settings = new ODataMessageWriterSettings();
             this.writeInstanceAnnotationsDelegate = async (ICollection<ODataInstanceAnnotation> instanceAnnotations) => await TaskUtils.CompletedTask;
-        }
-
-        [Fact]
-        public async Task StartJsonPaddingIfRequiredAsync_DoesNothingIfNullFunctionName()
-        {
-            settings.JsonPCallback = null;
-            await ODataJsonWriterUtils.StartJsonPaddingIfRequiredAsync(this.jsonWriter, settings);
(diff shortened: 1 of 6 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataJsonWriterUtilsAsyncTests.cs
```diff
@@ -21,58 +21,15 @@ public class ODataJsonWriterUtilsAsyncTests
...
 
         public ODataJsonWriterUtilsAsyncTests()
         {
             this.stringWriter = new StringWriter();
             this.jsonWriter = new JsonWriter(this.stringWriter, isIeee754Compatible: true);
-            this.settings = new ODataMessageWriterSettings();
             this.writeInstanceAnnotationsDelegate = async (ICollection<ODataInstanceAnnotation> instanceAnnotations) => await TaskUtils.CompletedTask;
         }
 
-        [Fact]
```
habbes, line 35: Do we have a test elsewhere that shows that jsonp is still supported? i.e. that if there's jsonp callback set, that the output will be wrapped in a json function?

gathogojr: @habbes Yes we do. After removing the async methods from `ODataJsonWriterUtils`, I confirmed that the code I transferred to `ODataJsonLightSerializer` is still being tested: https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightSerializerTests.cs#L306-L326

habbes (approved):

KenitoInc (approved):

Issue #1960, opened 2020-12-17
Remove ODataClient support for HttpWebRequest in ODL 8.0


mikepizzo:
Issue #1953 added support to ODataClient for building on top of HttpClient, which is a more modern client supported by frameworks like Blazor.

For backwards compatibility we had to continue to support HttpWebRequest, resulting in a dual code path in our library.  We marked support for HttpWebRequest as deprecated.

In ODataClient 8.0 we should drop support for HttpWebRequest all together.

### Assemblies affected

*OData .Net lib 7.x*

mikepizzo:
Assigning to Sreejith to track as part of ODL 8.0.

gathogojr:
Fixed by https://github.com/OData/odata.net/pull/2860

Issue #2380, opened 2022-04-25
Allow odata client to take an http client


corranrogue9:
PR #2207 is addressing the issue, but requires a breaking change

corranrogue9:
This is a subset of the work described in #2485, please track 2485 instead.

PR #2860, opened 2024-02-19
[ODL 8]: Remove `HttpWebRequestMessage`

### LINKED ISSUE #2486: Remove HttpWebRequestMessage
`HttpWebRequestMessage` is based on [`HttpWebRequest`](https://docs.microsoft.com/en-us/dotnet/api/system.net.httpwebrequest#remarks) which is no longer recommended for new development. 

### Assemblies affected

### Reproduce steps

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

habbes:
### Issues

*This pull request fixes #2486 #1960 

This PR removes `HttpWebRequestMessage` from OData Client. That class is based on the legacy [`HttpWebRequest`](https://learn.microsoft.com/en-us/dotnet/api/system.net.httpwebrequest) API. We had marked that option as obsolete and recommend the use of `HttpClient` instead. We want to remove support for `HttpWebRequest` in ODL 8.

### Description

- Removes `HttpWebRequestMessage` class
- Removes `HttpRequestTransportMode` enum (since only `HttpClient` is supported now)
- Removes `DataServiceContext.HttpRequestTranstportMode` property (since only `HttpClient` is supported)
- Removes `HttpWebResponseMessage.HttpWebResponse` property since `HttpWebResponse` is only possible if the request was `HttpWebRequest`
- Removes the `HttpWebResponseMessage(HttpWebResponse)` constructor overload (since `HttpWebResponse` is only possible if the request was `HttpWebRequest`)
- Removes `HttpWebRequestMessage` and references of `HttpWebRequest` and `HttpWebResponse` from E2E tests. A lot of the tests had tight coupling to these APIs, so there are a lot of file changes there. I ran the E2E tests (at least those that can run) locally to verify things still work.

A special note on the last item. I found that this constructor is used by `HttpClientRequestMessage.ConvertToDataServiceTransportException(WebException)` method:

```c#
private static T UnwrapAggregateException<T>(Func<T> action)
{
    try
    {
        return action();
    }
    catch (AggregateException aggregateException)
    {
        if (aggregateException.InnerExceptions.Count == 1)
        {
            throw ConvertToDataServiceTransportException(new WebException(aggregateException.InnerException.ToString()));
        }
        throw;
    }
}


private static DataServiceTransportException ConvertToDataServiceTransportException(WebException webException)
{
    HttpWebResponseMessage errorResponseMessage = null;
    if (webException.Response != null)
    {
        HttpWebResponse httpResponse = (HttpWebResponse)webException.Response;
        errorResponseMessage = new HttpWebResponseMessage(httpResponse);
    }

    return new DataServiceTransportException(errorResponseMessage, webException);
}
```

This is the only usage of this constructor overload. Since `WebException` is a newly created instance in this case (see the `UnwrapAggregateException`, we expect `webException.Response` to be `null`, and so that constructor will not be reached. So, it's safe to remove.


src/Microsoft.OData.Client/Materialization/HttpWebResponseMessage.cs
```diff
@@ -51,7 +50,7 @@ public HttpWebResponseMessage(IDictionary<string, string> headers, int statusCod
         /// Constructor.
         /// </summary>
         /// <param name="httpResponse">HttpWebResponse instance.</param>
-        public HttpWebResponseMessage(HttpWebResponse httpResponse)
+        internal HttpWebResponseMessage(HttpWebResponse httpResponse)
```
ElizabethOkerio, line 53: why the change?

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -8,14 +8,12 @@ namespace Microsoft.OData.Client
 { 
     using System;
     using System.Collections.Generic;
```
ElizabethOkerio, line 10: I think there is more cleaning that needs to be done to this class since there are some properties like `ReadWriteTimeOut` and `Credentials` that were overridden but should be removed..

KenitoInc: Also remove these
https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Client/Serialization/DataServiceClientRequestMessage.cs#L63
https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Client/Serialization/DataServiceClientRequestMessage.cs#L73

habbes: I was planning to work on these in a separate work item: #2380

src/Microsoft.OData.Client/Materialization/HttpWebResponseMessage.cs
habbes: This constructor exposes `HttpWebResponse` and so it's not suitable for public use since we no longer support `HttpWebRequest`. It may be possible to safely remove this constructor entirely. It's only used once internally and I think that callsite is only reached when using `HttpWebRequest`.

habbes: Actually, looking at the PR description again, it appears that I had intended to remove this constructor and had explained the reasoning. Maybe I should go ahead and do that.

gathogojr (approved):
LGTM

ElizabethOkerio (dismissed):

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
ElizabethOkerio: I think it's important to ensure these follow-up work items are tracked and worked on immediately, so they don't get left out.

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -498,14 +496,7 @@ private static T UnwrapAggregateException<T>(Func<T> action)
...
         {
-            HttpWebResponseMessage errorResponseMessage = null;
-            if (webException.Response != null)
-            {
-                HttpWebResponse httpResponse = (HttpWebResponse)webException.Response;
-                errorResponseMessage = new HttpWebResponseMessage(httpResponse);
-            }
-
-            return new DataServiceTransportException(errorResponseMessage, webException);
+            return new DataServiceTransportException(response: null, webException);
```
gathogojr, line 499: Curious how the thrown exception looks like when we pass a null here. We should confirm. Maybe we should add a resource string with an appropriate error message to use here.

test/EndToEndTests/Tests/Client/Build.Desktop/ODataWCFServiceTests/IDReadLinkEditLinkTests.cs
```diff
@@ -39,9 +39,9 @@ public void ClientShouldUseTheEditLinkGotFromPayloadToUpdateTheEntryInMinimalFul
         private void UpdateObject(string mimeType)
         {
             TestClientContext.MergeOption = MergeOption.OverwriteChanges;
-            TestClientContext.SendingRequest2 += (sender, eventArgs) => ((Microsoft.OData.Client.HttpWebRequestMessage)eventArgs.RequestMessage).SetHeader(TestHeader, "EditLink");
+            TestClientContext.SendingRequest2 += (sender, eventArgs) => ((Microsoft.OData.Client.HttpClientRequestMessage)eventArgs.RequestMessage).SetHeader(TestHeader, "EditLink");
```
gathogojr, line 42: Any reason we couldn't import the `Microsoft.OData.Client` namespace instead of typing the whole namespace so many places?

test/EndToEndTests/Tests/Client/Build.Desktop/TransportLayerTests/HttpClientRequestMessage.cs
```diff
@@ -335,14 +335,7 @@ private static T UnwrapAggregateException<T>(Func<T> action)
...
         {
-            HttpWebResponseMessage errorResponseMessage = null;
-            if (webException.Response != null)
-            {
-                var httpResponse = (HttpWebResponse)webException.Response;
-                errorResponseMessage = new HttpWebResponseMessage(httpResponse);
-            }
-
-            return new DataServiceTransportException(errorResponseMessage, webException);
+            return new DataServiceTransportException(response: null, webException);
```
gathogojr, line 338: See my comment here https://github.com/OData/odata.net/pull/2860/files#r1500279823

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/BuildingRequestEventFunctionalTests.cs
```diff
@@ -962,9 +962,13 @@ private string RetrieveLastSegmentOfRequestUri(ReceivingResponseEventArgs arg)
...
             }
-            else
+            else if (arg.Descriptor != null)
             {
-                return (arg.ResponseMessage as HttpWebResponseMessage).Response.ResponseUri.Segments.Last();
+                return arg.Descriptor.State.ToString();
+            }
+            {
+                // this is the top-level batch request
+                return "$batch";
```
gathogojr, line 971: What's with the code block here? It's not even an `else` block and it didn't exist before...

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
gathogojr: I just noticed that the first parameter is of type `IODataResponseMessage`. Is there anything we can pass in place of what was being passed before? The question about the exception appears to the caller when a null is passed is still an important one to answer

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
habbes: The item is tracked and part of ODL 8 plan. I was planning to take it up as soon as this PR is merged.

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
habbes: The code before was passing null if `webException.Response` was null. But `webException.Response` was non-null only when using `HttpWebResponse`. I didn't find a way to reliably get the `HttpResponseMessage` from an exception. From example, with an `OperationCancelledException`, it doesn't have a link to the `HttpResponseMessage`, it's also possible the response message was not even created.

test/EndToEndTests/Tests/Client/Build.Desktop/ODataWCFServiceTests/IDReadLinkEditLinkTests.cs
habbes: Hmm. I don't know why this was done, I was just focused on changing the referenced class. I guess I could import the namespace as well.

test/EndToEndTests/Tests/Client/Build.Desktop/TransportLayerTests/HttpClientRequestMessage.cs
habbes: See my response here: https://github.com/OData/odata.net/pull/2860/files#r1500334711

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/BuildingRequestEventFunctionalTests.cs
```diff
@@ -962,9 +962,14 @@ private string RetrieveLastSegmentOfRequestUri(ReceivingResponseEventArgs arg)
...
             }
+            else if (arg.Descriptor != null)
+            {
+                return arg.Descriptor.State.ToString();
+            }
             else
             {
-                return (arg.ResponseMessage as HttpWebResponseMessage).Response.ResponseUri.Segments.Last();
+                // this is the top-level batch request
+                return "$batch";
```
gathogojr, line 972: Did `return "$batch" exist before though?

gathogojr (approved):
Left one comment, otherwise LGTM.

test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/DataWebClientCSharp/BuildingRequestEventFunctionalTests.cs
habbes: For this test case, this statement is reached when it's a top-level batch request, in which case the `Response.ResponseUri.Segments.Last()` from the old code would return `$batch` as that's the last segment of a batch request URI. But since we don't have access to the `HttpWebResponse` now that we're using `HttpClienRequestMessage`, I've hardcoded `$batch` in this test.

A more robust solution would be to add a `RequestUri` property to the `ReceivingResponseEventArgs` and change the code that triggers the event to pass that information down. But I think that's a basically a new feature that probably would need its own discussion and PR.

PR #2862, opened 2024-02-19
Fix "OriginalNameAttribute not being captured when DataServiceQuery<T> is casted to DataServiceQuery<InterfaceT>"

### LINKED ISSUE #2856: OriginalNameAttribute not being captured when DataServiceQuery<T> is casted to DataServiceQuery<InterfaceT>
I'm getting the error described in the title when trying to resolve an OData query.

### Assemblies affected

Microsoft.OData.Client 7.20.0
NetFramework 4.6.1

### Reproduce steps

```csharp
// Declared model and interface:

[EntitySet("TestEntities")]
public class TestModel : ITestModel
{
	[OriginalName("userId")]
	public int Id { get; set; }

	[OriginalName("user")]
	public string UserName { get; set; }	
}

public interface ITestModel
{
	int Id { get; set; }
	string UserName { get; set; }
}


// Declared context class:

public class MyContext : DataServiceContext
{
	public DataServiceQuery<TestModel> TestEntities;

	public MyContext(Uri serviceRoot) : base(serviceRoot)
	{      
		Format.UseJson();
		TestEntities = CreateQuery<TestModel>("TestEntities");
	}
}


// Main program:

var ctx = new MyContext(uri);

ctx.TestEntities.Where(c => c.Id == 1).FirstOrDefault(); // Works fine

IQueryable<ITestModel> q = ctx.TestEntities;
q.Where(c => c.Id == 1).FirstOrDefault(); // Exception is thrown here because the query is
                                          // being built using `Id` instead of `userId`

```

### Expected result

No exception should be thrown and the query should be built using the object's original type attributtes.

### Actual result

An exception is being thrown showing the following OData service error response:
```
{"error":{"code":"BadRequest","message":"Could not find a property named 'Id' on type 'Microsoft.NAV.testModel'...
```

joecarl:
### Issues

*This pull request fixes #2856.*

### Description

Capture member from original class, so attributes are not lost in case the query generic parameter is casted to an interface.

- [ ] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
+++ b/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
@@ -506,7 +506,8 @@ namespace Microsoft.OData.Client
             }
             else
             {
-                this.builder.Append(ClientTypeUtil.GetServerDefinedName(m.Member));
+                var member = m.Expression.Type.GetMember(m.Member.Name).First();
+                this.builder.Append(ClientTypeUtil.GetServerDefinedName(member));
             }
 
             return m;


joecarl:

habbes:
@joecarl can you add tests?

src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs
```diff
@@ -506,7 +506,8 @@ internal override Expression VisitMemberAccess(MemberExpression m)
             }
             else
             {
-                this.builder.Append(ClientTypeUtil.GetServerDefinedName(m.Member));
+                var member = m.Expression.Type.GetMember(m.Member.Name).First();
```
habbes, line 509: Can you confirm (ideally in tests) that this fix also works with interfaces, abstract classes, different levels of inheritance, etc.

joecarl: Sure I will

joecarl:
I fixed the code in order to pass those failed tests.

Also, I found a third related issue:

If you try to run this code It will throw Exception:

```cs
  [Key("CustomerID")]
  public class BaseCustomer
  {
      [OriginalName("CustomerID")]
      public string Id { get; set; }
  }

  public class OneLevelCustomer2 : BaseCustomer
  {
      public string City { get; set; }
  }  

  // Test:
  IQueryable<OneLevelCustomer2> q = ctx.OneLevelCustomers2; // ctx.OneLevelCustomers2 is a DataServiceQuery<OneLevelCustomer2>
  var r = q.First();

```

```
The current value 'Microsoft.OData.Client.Tests.ALinq.BaseCustomer' type is not compatible with the expected 'Microsoft.OData.Client.Tests.ALinq.OneLevelCustomer2' type.
   at Microsoft.OData.Client.Materialization.EntityTrackingAdapter.TryResolveFromContext(MaterializerEntry entry, Type expectedEntryType) in /workspaces/testnet/odata.net/src/Microsoft.OData.Client/Materialization/EntityTrackingAdapter.cs:line 187
   at Microsoft.OData.Client.Materialization.EntityTrackingAdapter.TryResolveAsExistingEntry(MaterializerEntry entry, Type expectedEntryType) in /workspaces/testnet/odata.net/src/Microsoft.OData.Client/Materialization/EntityTrackingAdapter.cs:line 137
   at Microsoft.OData.Client.Materialization.EntityTrackingAdapter.TryResolveExistingEntity(MaterializerEntry entry, Type expectedEntryType) in /workspaces/testnet/odata.net/src/Microsoft.OData.Client/Materialization/EntityTrackingAdapter.cs:line 107
   at Microsoft.OData.Client.Materialization.EntryValueMaterializationPolicy.Materialize(MaterializerEntry entry, Type expectedEntryType, Boolean includeLinks) in /workspaces/testnet/odata.net/src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs:line 197
   at Microsoft.OData.Client.Materialization.ODataEntityMaterializer.DirectMaterializePlan(ODataEntityMaterializer materializer, MaterializerEntry entry, Type expectedEntryType) in /workspaces/testnet/odata.net/src/Microsoft.OData.Client/Materialization/ODataEntityMaterializer.cs:line 573
   at Microsoft.OData.Client.Materialization.ODataEntityMaterializerInvoker.DirectMaterializePlan(Object materializer, Object entry, Type expectedEntryType) in /workspaces/testnet/odata.net/src/Microsoft.OData.Client/Materialization/ODataEntityMaterializerInvoker.cs:line 175
   at Microsoft.OData.Client.ProjectionPlan.Run(ODataEntityMaterializer materializer, ODataResource entry, Type expectedType) in /workspaces/testnet/odata.net/src/Microsoft.OData.Client/ProjectionPlan.cs:line 78
   at Microsoft.OData.Client.Materialization.ODataEntityMaterializer.ReadImplementation() in /workspaces/testnet/odata.net/src/Microsoft.OData.Client/Materialization/ODataEntityMaterializer.cs:line 927
   at Microsoft.OData.Client.Materialization.ODataMaterializer.Read() in /workspaces/testnet/odata.net/src/Microsoft.OData.Client/Materialization/ODataMaterializer.cs:line 336
   at Microsoft.OData.Client.MaterializeAtom.MoveNextInternal() in /workspaces/testnet/odata.net/src/Microsoft.OData.Client/MaterializeFromAtom.cs:line 349
   at Microsoft.OData.Client.MaterializeAtom.MoveNext() in /workspaces/testnet/odata.net/src/Microsoft.OData.Client/MaterializeFromAtom.cs:line 281
   at System.Linq.Enumerable.CastIterator[TResult](IEnumerable source)+MoveNext()
   at System.Collections.Generic.List`1..ctor(IEnumerable`1 collection)
   at System.Linq.Enumerable.ToList[TSource](IEnumerable`1 source)
   at Microsoft.OData.Client.Tests.ALinq.PreserveTypesAndAttributesTests.OneLevelCase2() in /workspaces/testnet/odata.net/test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs:line 53
   at Program.<Main>$(String[] args) in /workspaces/testnet/tests/Program.cs:line 102
```

This error exists before my PR. I guess it could be easily fixeable too.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
```diff
@@ -0,0 +1,250 @@
...
+        {
+            string uri = "https://services.odata.org/V4/Northwind/Northwind.svc";
+            this.ctx = new TestContext(new Uri(uri));
+        }
+
+        [Fact]
+        public void OneLevelCase1()
+        {
+            IQueryable<ICustomer> q1 = ctx.OneLevelCustomers1;
+            var r1 = q1.Where(c => c.CorporationName != "").OrderBy(c => c.Id).Take(5).ToList();
```
wachugamaina, line 32: Could you better name these variables so that its clear what they represent.

joecarl: Done

joecarl: @wachugamaina are the new names ok for you?

joecarl:
Could I please receive any feedback from the review process? @habbes @wachugamaina @gathogojr @xuzhg Thanks in advance

src/Microsoft.OData.Client/ALinq/ResourceSetExpression.cs
```diff
@@ -104,11 +104,13 @@ internal override QueryableResourceExpression CreateClone()
...
         protected override QueryableResourceExpression CreateCloneWithNewTypes(Type newType, Type newResourceType)
         {
+            bool ignoreNewResourceType = newResourceType.IsAbstract || newResourceType.IsInterface;
+            
             return new ResourceSetExpression(
                 newType,
                 this.source,
                 this.MemberExpression,
-                newResourceType,
+                ignoreNewResourceType ? this.ResourceType : newResourceType,
```
habbes, line 113: What is the impact of this change? Is this something safe to do?

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
```diff
@@ -0,0 +1,250 @@
...
+        {
+            string uri = "https://services.odata.org/V4/Northwind/Northwind.svc";
+            this.ctx = new TestContext(new Uri(uri));
+        }
+
+        [Fact]
+        public void OneLevelCase1()
+        {
+            IQueryable<ICustomer> query1 = ctx.OneLevelCustomers1;
+            var result1 = query1.Where(c => c.CorporationName != "").OrderBy(c => c.Id).Take(5).ToList();
```
habbes, line 32: These tests are making network requests to `https://services.odata.org/V4/Northwind/Northwind.svc`, right? Should we make the test less reliant on the network instead? I think it would be sufficient to test when the URI is generated with the expected property names, without having to make the network requests and return mock responses. Do we have other unit tests doing this @gathogojr  @KenitoInc @ElizabethOkerio?

src/Microsoft.OData.Client/ALinq/ResourceSetExpression.cs
joecarl: This change was made in order to address what I mention here: https://github.com/OData/odata.net/issues/2856#issuecomment-1960616567

I cannot imagine any scenario where this change could be a problem. Also I have been intensively using my modified fork for 3 months, every single day, and this caused 0 issues. So I would say It's safe.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
joecarl: Yes, it relies on external service. I understand what you mean and I think that (for this use case) comparing URIs could be enough. However, other issues, like the one from my other PR (#2863) need to check the response in order to verify that it is working fine.
Could it be an option to add a minimalist local http odata service that we could use for testing purposes?

gathogojr: @habbes @joecarl You can and should mock the response. We do this by intercepting `OnMessageCreating` call and returning a mocked response. You can refer to the following for example:
https://github.com/OData/odata.net/blob/bbe2ac74487fb01fe14d97aeca131d0ca89d2452/test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/DollarApplyTestsBase.cs#L306

joecarl:
@habbes I modified the tests so now they don't depend on the external service (following @gathogojr indications). Please have a look and let me know if there's somethign else I should do. Thanks.

src/Microsoft.OData.Client/ALinq/ResourceSetExpression.cs
ElizabethOkerio: You mean this issue happens when you add OriginalName attribute to the interface properties then use OrderBy?

joecarl: It happens even without the OriginalName attribute. It's just a casting side effect.

Here is a minimalist example to reproduce the issue:

```cs

using Microsoft.OData.Client;


Uri uri = new Uri("https://services.odata.org/V4/Northwind/Northwind.svc");
DataServiceContext ctx = new DataServiceContext(uri);        
DataServiceQuery<CustomerClass> originalQuery = ctx.CreateQuery<CustomerClass>("Customers");

IQueryable<ICustomer> castedQuery = originalQuery;
castedQuery = castedQuery.OrderBy(c => c.CustomerID).Take(1);

// This will throw the mentioned exception at materialization level because it cannot instance an ICustomer
var result = castedQuery.ToList();


// Types declarations:

public interface ICustomer
{
    string CustomerID { get; set; }
}

[Key("CustomerID")]
public class CustomerClass : ICustomer
{
    public string CustomerID { get; set; }
}

```

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
```diff
@@ -0,0 +1,395 @@
...
+
+        private static void AssertEqualRequestUri(IQueryable query1, IQueryable query2)
+        {
+            var dsq1 = query1 as DataServiceQuery;
+            var dsq2 = query2 as DataServiceQuery;
+            Assert.Equal(dsq1.RequestUri, dsq2.RequestUri);
+        }
+
+        [Fact]
+        public void OneLevelCase1()
```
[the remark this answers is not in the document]
habbes: For example, the focus of this test seems to be interfaces.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
```diff
@@ -0,0 +1,395 @@
...
+
+        [OriginalName("ContactName")]
+        public string Name { get; set; }
+
+        [OriginalName("Orders")]
+        public List<Order> OrdersList { get; set; }
+    }
+
+    [Key("OrderID")]
+    public class Order
```
habbes, line 268: Since these model class names are common, it may be good to consider moving them to inner classes or a namespace where they are less likely to cause conflict.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
```diff
@@ -0,0 +1,395 @@
...
+    {
+        public int OrderID { get; set; }
+        public decimal Freight { get; set; }
+    }
+
+
+    public class TestContext
```
habbes, line 275: The name `TestContext` is very generic, but this class is very specific to this test. Already, this will cause conflicts and compilation errors if this other PR https://github.com/OData/odata.net/pull/2863 is merged, because they both define a `TestContext` class in the same namespace. One way to avoid this conflict is to move `TestContext` as an inner class of the test class. But if we do that, it brings me back to the question, why do we need this class in the first place?

The helper methods it provides might as well be moved to the test class then, then we don't need an inner class. We can move them to the bottom of the class so that they don't create distractions or noise when looking at tests.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
```diff
@@ -0,0 +1,395 @@
...
+        public DataServiceQuery<OneLevelCustomer1> OneLevelCustomers1;
+        public DataServiceQuery<TwoLevelCustomer1> TwoLevelCustomers1;
+        public DataServiceQuery<OneLevelCustomer2> OneLevelCustomers2;
+        public DataServiceQuery<TwoLevelCustomer2> TwoLevelCustomers2;
+
+        // We have to create multiple DataServiceContext because the entitySetName is the same for every DataServiceQuery
+        private readonly DataServiceContext _ctx1;
+        private readonly DataServiceContext _ctx2;
+        private readonly DataServiceContext _ctx3;
+        private readonly DataServiceContext _ctx4;
```
habbes, line 286: I don't think it's necessary to have 4 different instances. Since this `TestContext` class is instantiated in the constructor of the test class, it means it's instantiated for each test. Each test seems to be using only one of the `DataServiceContext` instance, but all 4 are created. So, for each test, we create 3 instances that we don't use, which is wasteful. I would suggest refactoring the tests such that we only create one instance. Do reduce boilerplate inside the test methods, this instance can be initialized in the test class constructor. Since we're not using static fields, we shouldn't have to worry about one test method interfering with the state of a different test method.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
```diff
@@ -0,0 +1,395 @@
...
+    /// Tests to check original resource type is used when retrieving attributes like OriginalNameAttribute
+    /// even if the IQueryable<T> object is casted to another type.
+    /// </summary>
+    public class PreserveTypesAndAttributesTests
+    {
+        private readonly TestContext ctx;
+
+        public PreserveTypesAndAttributesTests()
+        {
+            this.ctx = new TestContext();
```
habbes, line 27: Instead of `TestContext` we could consider initializing the `DataServiceContext` here directly.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
```diff
@@ -0,0 +1,395 @@
...
+            {
+                Assert.Equal(result1[i].City, result2[i].City);
+                Assert.Equal(result1[i].CorporationName, result2[i].CorporationName);
+                Assert.Equal(result1[i].Name, result2[i].Name);
+                Assert.Equal(result1[i].Id, result2[i].Id);
+            }
+        }
+
+        [Fact]
+        public void OneLevelCase2()
```
habbes, line 65: See: https://github.com/OData/odata.net/pull/2862/files#r1639169301

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
```diff
@@ -0,0 +1,395 @@
...
+
+            Assert.Equal(result1.Count, result2.Count);
+            for (var i = 0; i < result1.Count; i++)
+            {
+                Assert.Equal(result1[i].Id, result2[i].Id);
+            }
+        }
+
+        [Fact]
+        public void MoreThanOneLevelLevelCase1()
```
habbes, line 88: See: https://github.com/OData/odata.net/pull/2862/files#r1639169301

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
```diff
@@ -0,0 +1,395 @@
...
+            {
+                Assert.Equal(result3[i].Id, result2[i].Id);
+                Assert.Equal(result3[i].CorporationName, result2[i].CorporationName);
+                Assert.Equal(result3[i].City, result2[i].City);
+                Assert.Equal(result3[i].Name, result2[i].Name);
+            }
+        }
+
+        [Fact]
+        public void MoreThanOneLevelLevelCase2()
```
habbes, line 131: See: https://github.com/OData/odata.net/pull/2862/files#r1639169301

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
```diff
@@ -0,0 +1,395 @@
...
+
+            Assert.Equal(result1.Count, result2.Count);
+            for (var i = 0; i < result1.Count; i++)
+            {
+                Assert.Equal(result1[i].City, result2[i].City);
+                Assert.Equal(result1[i].CorporationName, result2[i].CorporationName);
+                Assert.Equal(result1[i].Name, result2[i].Name);
+                Assert.Equal(result1[i].Id, result2[i].Id);
+            }
+        }
```
habbes, line 62: Given the discussion on `DataServiceContext` and `TestContext` down below, I'd suggest a refactor like the following, getting rid of the `TestContext` in favour of referencing `DataServiceContext` directly:

```suggestion
        {
            ctx.CreateQuery<OneLevelCustomer1`> query =  ctx.CreateQuery<OneLevelCustomer1>("Cusomers");
            IQueryable<ICustomer> queryWithInterface= query;
            queryWithInterface = queryWithInterface.Where(c => c.CorporationName != "").OrderBy(c => c.Id).Take(2);

            IQueryable<OneLevelCustomer1> queryWithClass= query.Where(c => c.CorporationName != "").OrderBy(c => c.Id).Take(2);

            AssertEqualRequestUri(queryWithInterface, queryWithClass);

            string resp = "[{\"CustomerID\":\"ALFoKI\",\"CompanyName\":\"Alfreds Futterkiste\",\"ContactName\":\"Maria Anders\",\"City\":\"Berlin\",\"Region\":null},{\"CustomerID\":\"ANATR\",\"CompanyName\":\"Ana Trujillo Emparedados y helados\",\"ContactName\":\"Ana Trujillo\",\"City\":\"M\\u00e9xico D.F.\",\"Region\":null}]";
            ctx.InterceptRequestAndMockResponseValue("Customers", resp);

            var result1 = queryWithInterface.ToList();
            var result2 = queryWithClass.ToList();

            Assert.Equal(result1.Count, result2.Count);
            for (var i = 0; i < result1.Count; i++)
            {
                Assert.Equal(result1[i].City, result2[i].City);
                Assert.Equal(result1[i].CorporationName, result2[i].CorporationName);
                Assert.Equal(result1[i].Name, result2[i].Name);
                Assert.Equal(result1[i].Id, result2[i].Id);
            }
        }
```

habbes: In my opinion, this makes the test clearer, reducing the references that need to be looked up, without adding any more clutter to the original version.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
```diff
@@ -0,0 +1,395 @@
...
...":\"1997-09-05T00:00:00Z\",\"ShippedDate\":\"1997-08-14T00:00:00Z\",\"ShipVia\":1,\"Freight\":43.9000,\"ShipName\":\"Ana Trujillo Emparedados y helados\",\"ShipAddress\":\"Avda. de la Constituci\\u00f3n 2222\",\"ShipCity\":\"M\\u00e9xico D.F.\",\"ShipRegion\":null,\"ShipPostalCode\":\"05021\",\"ShipCountry\":\"Mexico\"},{\"OrderID\":10759,\"CustomerID\":\"ANATR\",\"EmployeeID\":3,\"OrderDate\":\"1997-11-28T00:00:00Z\",\"RequiredDate\":\"1997-12-26T00:00:00Z\",\"ShippedDate\":\"1997-12-12T00:00:00Z\",\"ShipVia\":3,\"Freight\":11.9900,\"ShipName\":\"Ana Trujillo Emparedados y helados\",\"ShipAddress\":\"Avda. de la Constituci\\u00f3n 2222\",\"ShipCity\":\"M\\u00e9xico D.F.\",\"ShipRegion\":null,\"ShipPostalCode\":\"05021\",\"ShipCountry\":\"Mexico\"},{\"OrderID\":10926,\"CustomerID\":\"ANATR\",\"EmployeeID\":4,\"OrderDate\":\"1998-03-04T00:00:00Z\",\"RequiredDate\":\"1998-04-01T00:00:00Z\",\"ShippedDate\":\"1998-03-11T00:00:00Z\",\"ShipVia\":3,\"Freight\":39.9200,\"ShipName\":\"Ana Trujillo Emparedados y helados\",\"ShipAddress\":\"Avda. de la Constituci\\u00f3n 2222\",\"ShipCity\":\"M\\u00e9xico D.F.\",\"ShipRegion\":null,\"ShipPostalCode\":\"05021\",\"ShipCountry\":\"Mexico\"}]}]";
```
gathogojr, line 148: You don't need such a big payload to verify that the code is working as expected. 2 or at most 3 objects in the JSON array are sufficient...

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
```diff
@@ -0,0 +1,395 @@
...
+using Microsoft.OData.Edm;
+using Xunit;
+
+namespace Microsoft.OData.Client.Tests.ALinq
+{
+    /// <summary>
+    /// Tests to check original resource type is used when retrieving attributes like OriginalNameAttribute
+    /// even if the IQueryable<T> object is casted to another type.
+    /// </summary>
+    public class PreserveTypesAndAttributesTests
```
gathogojr, line 21: @joecarl Disabuse yourself from the notion that each test needs it's own instance of `DataServiceContext`. If that was the case and you needed to test 25 scenarios, you'd end up with 25 `DataServiceContext` instances in the `TestContext` class.
I'll reiterate what I said in your #2862 pull request. You don't need the `TestContext` class at all.

Below is a rewrite of the first 2 tests without the `TestContext` class.

The naming for the test method should give someone an idea what scenario is being tested - `OneLevelCase1` and `OneLevelCase2` do not do that...
```csharp
namespace Microsoft.OData.Client.Tests.ALinq
{
    public class PreserveTypesAndAttributesTests // You can make this class implement IDisposable if you feel very strongly about removing the SendingRequest2 delegate...
    {
        private readonly DataServiceContext ctx;
        private readonly EdmModel model;

        public PreserveTypesAndAttributesTests()
        {
            this.model = this.BuildEdmModel();
            this.ctx = new DataServiceContext(new Uri("https://mock.odata.service"));
            this.ctx.Format.UseJson(this.model);
            this.ctx.ResolveName = (type) => $"NS.{type.Name}";
            this.ctx.MergeOption = MergeOption.NoTracking;
        }

        [Fact]
        public void TestMethodNameThatGivesSomeoneABetterIdeaWhatScenarioIsBeingTested1()
        {
            IQueryable<ICustomer> query1 = this.ctx.CreateQuery<OneLevelCustomer1>("Customers")
                .Where(c => c.CorporationName != "").OrderBy(c => c.Id).Take(2);

            IQueryable<OneLevelCustomer1> query2 = this.ctx.CreateQuery<OneLevelCustomer1>("Customers")
                .Where(c => c.CorporationName != "").OrderBy(c => c.Id).Take(2);

            // TODO: Include the `@odata.context` control information property in the payload
            string resp = "[{\"CustomerID\":\"ALFoKI\",\"CompanyName\":\"Alfreds Futterkiste\",\"ContactName\":\"Maria Anders\",\"City\":\"Berlin\",\"Region\":null},{\"CustomerID\":\"ANATR\",\"CompanyName\":\"Ana Trujillo Emparedados y helados\",\"ContactName\":\"Ana Trujillo\",\"City\":\"M\\u00e9xico D.F.\",\"Region\":null}]";

            InterceptRequestAndMockResponseValue("Customers", resp);
            this.ctx.SendingRequest2 += (sender, e) =>
            {
                Assert.EndsWith("<EXPECTED_URL>", e.RequestMessage.Url.OriginalString);
            };

            var result1 = query1.ToList();
            var result2 = query2.ToList();

            Assert.Equal(result1.Count, result2.Count);
            for (var i = 0; i < result1.Count; i++)
            {
                Assert.Equal(result1[i].City, result2[i].City);
                Assert.Equal(result1[i].CorporationName, result2[i].CorporationName);
                Assert.Equal(result1[i].Name, result2[i].Name);
                Assert.Equal(result1[i].Id, result2[i].Id);
            }
        }

        [Fact]
        public void TestMethodNameThatGivesSomeoneABetterIdeaWhatScenarioIsBeingTested2()
        {
            IQueryable<BaseCustomer> query1 = ctx.CreateQuery<OneLevelCustomer2>("Customers");
                .Where(c => c.Id != "").OrderBy(c => c.Id).Select(c => new { c.Id }).Take(2);

            IQueryable<OneLevelCustomer2> query2 = ctx.CreateQuery<OneLevelCustomer2>("Customers")
                .Where(c => c.Id != "").OrderBy(c => c.Id).Select(c => new { c.Id }).Take(2);

            // TODO: Include the `@odata.context` control information property in the payload
            string resp = "[{\"CustomerID\":\"ALFKI\"},{\"CustomerID\":\"ANATR\"}]";

            this.ctx.InterceptRequestAndMockResponseValue("Customers(CustomerID)", resp);
            this.ctx.SendingRequest2 += (sender, e) =>
            {
                Assert.EndsWith("<EXPECTED_URL>", e.RequestMessage.Url.OriginalString);
            };

            var result1 = query1s.ToList();
            var result2 = query2s.ToList();

            Assert.Equal(result1.Count, result2.Count);
            for (var i = 0; i < result1.Count; i++)
            {
                Assert.Equal(result1[i].Id, result2[i].Id);
            }
        }

        // You can put the private helper methods in a #region to make the section optionally collapsible

        // You don't need to pass a DataServiceContext instance since `ctx` object is accessible to this method
        private void InterceptRequestAndMockResponse(string mockResponse)
        {
            // ctx.ResolveName = (type) => $"NS.{type.Name}"; <= DOES NOT BELONG HERE
            // THE REST OF THE CODE ...
        }

       // You don't need the other InterceptRequestAndMockResponse overload. Include the `@odata.context` control information property in the payload (from the test)

        private static EdmModel BuildEdmModel()
        {
            // CODE ..
        }
    }
}
```

gathogojr (dismissed):
See my comments

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/PreserveTypesAndAttributesTests.cs
joecarl: I refactored the whole code following every single indication, even those which from my subjective point of view provide questionable improvement.
Also, fortunately in this case there's no need to add a listener to `SendingRequest2` since we can assert equality between each query's request uri, as I was already doing.

habbes (approved):

WanjohiSammy (approved):

PR #2863, opened 2024-02-19
Add support for "Any" and Predicates method call expressions

### LINKED ISSUE #1996: The method 'Any' is not supported
Hello. I am using Microsoft.OData.Client 7.8.1.

I want create DataServiceContext, CreateQuery and take data. 

Code is:

```c#
var oDataContext = new DataServiceContext(new Uri("https://localhost:5001"));
var dataSet = oDataContext.CreateQuery<Project>("Projects");
var exists = dataSet.Any();
```
But i get next error: "System.NotSupportedException: 'The method 'Any' is not supported.'"

StackTrace:

   at Microsoft.OData.Client.DataServiceQueryProvider.ReturnSingleton[TElement](Expression expression)
   at Microsoft.OData.Client.DataServiceQueryProvider.Execute[TResult](Expression expression)
   at System.Linq.Queryable.Any[TSource](IQueryable`1 source)
   at ASE.MD.MDP2.Product.MDP2Service.Controllers.TestController.GetByOData() in D:\WorkProjects\MDP\2.0\MDP2.0\MDP2Service\Controllers\TestController.cs:line 146
   at Microsoft.Extensions.Internal.ObjectMethodExecutor.Execute(Object target, Object[] parameters)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ActionMethodExecutor.SyncObjectResultExecutor.Execute(IActionResultTypeMapper mapper, ObjectMethodExecutor executor, Object controller, Object[] arguments)
   at Microsoft.AspNetCore.Mvc.Infrastructure.ControllerActionInvoker.<<InvokeActionMethodAsync>g__Logged|12_1>d.MoveNext()

What's wrong i do?

joecarl:
### Issues

*This pull request fixes #1996.*

### Description

This PR implmements `Any` method support. (Internally it fakes a `Count` request and transforms numeric response into boolean value).
It also adds support for the following sequence methods:
`AnyPredicate`, `CountPredicate`, `LongCountPredicate`, `FirstPredicate`, `FirstOrDefaultPredicate`, `SinglePredicate`, `SingleOrDefaultPredicate`

E.g.:
```csharp
var ctx = new DataServiceContext(new Uri(...));
var dataSet = ctx.CreateQuery<Customer>("Customers");
// The following lines will work fine without throwing Exception now:
bool exists1 = dataSet.Any();
bool exists2 = dataSet.Any(c => c.Name.Contains("ab"));
int count1 = dataSet.Count(c => c.Name.Contains("ab"));
long count2 = dataSet.LongCount(c => c.Name.Contains("ab"));
Customer c1 =  dataSet.First(c => c.Name != "John");
Customer c2 =  dataSet.FirstOrDefault(c => c.Name == "John");
Customer c3 =  dataSet.Single(c => c.Id == "234111");
Customer c4 =  dataSet.SingleOrDefault(c => c.Id == "234111");
```

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Client/ALinq/DataServiceQueryProvider.cs
+++ b/src/Microsoft.OData.Client/ALinq/DataServiceQueryProvider.cs
@@ -87,6 +87,23 @@ namespace Microsoft.OData.Client
 
         #endregion
 
+        private static Expression MakeCountExpression(Expression sourceExpression) 
+        {
+            return Expression.Call(
+                typeof(Enumerable), 
+                "Count", 
+                new Type[] { sourceExpression.Type.GetGenericArguments()[0] },
+                sourceExpression
+            );
+        }
+
+        private TElement GetValueForAny<TElement>(MethodCallExpression mce)
+        {
+            var countExpression = MakeCountExpression(mce.Arguments[0]);
+            DataServiceQuery<TElement> query = new DataServiceQuery<TElement>.DataServiceOrderedQuery(countExpression, this);
+            return query.GetValue(Context, ParseQuerySetCount<TElement>);
+        }
+
         /// <summary>Creates and executes a DataServiceQuery based on the passed in expression which results a single value</summary>
         /// <typeparam name="TElement">generic type</typeparam>
         /// <param name="expression">The expression for the new query</param>
@@ -114,6 +131,8 @@ namespace Microsoft.OData.Client
                     case SequenceMethod.LongCount:
                     case SequenceMethod.Count:
                         return ((DataServiceQuery<TElement>)query).GetValue(this.Context, ParseQuerySetCount<TElement>);
+                    case SequenceMethod.Any:
+                        return GetValueForAny<TElement>(mce);
                     case SequenceMethod.SumIntSelector:
                     case SequenceMethod.SumDoubleSelector:
                     case SequenceMethod.SumDecimalSelector:
--- a/src/Microsoft.OData.Client/ALinq/DataServiceQueryProvider.cs
+++ b/src/Microsoft.OData.Client/ALinq/DataServiceQueryProvider.cs
@@ -9,6 +9,7 @@ namespace Microsoft.OData.Client
     #region Namespaces
 
     using System;
+    using System.Collections;
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.IO;
@@ -87,22 +88,6 @@ namespace Microsoft.OData.Client
 
         #endregion
 
-        private static Expression MakeCountExpression(Expression sourceExpression) 
-        {
-            return Expression.Call(
-                typeof(Enumerable), 
-                "Count", 
-                new Type[] { sourceExpression.Type.GetGenericArguments()[0] },
-                sourceExpression
-            );
-        }
-
-        private TElement GetValueForAny<TElement>(MethodCallExpression mce)
-        {
-            var countExpression = MakeCountExpression(mce.Arguments[0]);
-            DataServiceQuery<TElement> query = new DataServiceQuery<TElement>.DataServiceOrderedQuery(countExpression, this);
-            return query.GetValue(Context, ParseQuerySetCount<TElement>);
-        }
 
         /// <summary>Creates and executes a DataServiceQuery based on the passed in expression which results a single value</summary>
         /// <typeparam name="TElement">generic type</typeparam>
@@ -110,7 +95,7 @@ namespace Microsoft.OData.Client
         /// <returns>single valued results</returns>
         internal TElement ReturnSingleton<TElement>(Expression expression)
         {
-            IQueryable<TElement> query = new DataServiceQuery<TElement>.DataServiceOrderedQuery(expression, this);
+            IQueryable<TElement> query = CreateQuery<TElement>(expression);
 
             MethodCallExpression mce = expression as MethodCallExpression;
             Debug.Assert(mce != null, "mce != null");
@@ -122,17 +107,35 @@ namespace Microsoft.OData.Client
                 {
                     case SequenceMethod.Single:
                         return query.AsEnumerable().Single();
+                    case SequenceMethod.SinglePredicate:
+                        query = CreateQuery<TElement>(NestPredicateExpression(mce));
+                        return query.AsEnumerable().Single();
                     case SequenceMethod.SingleOrDefault:
+                        return query.AsEnumerable().SingleOrDefault();
+                    case SequenceMethod.SingleOrDefaultPredicate:
+                        query = CreateQuery<TElement>(NestPredicateExpression(mce));
                         return query.AsEnumerable().SingleOrDefault();
                     case SequenceMethod.First:
                         return query.AsEnumerable().First();
+                    case SequenceMethod.FirstPredicate:
+                        query = CreateQuery<TElement>(NestPredicateExpression(mce));
+                        return query.AsEnumerable().First();
                     case SequenceMethod.FirstOrDefault:
+                        return query.AsEnumerable().FirstOrDefault();
+                    case SequenceMethod.FirstOrDefaultPredicate:
+                        query = CreateQuery<TElement>(NestPredicateExpression(mce));
                         return query.AsEnumerable().FirstOrDefault();
                     case SequenceMethod.LongCount:
                     case SequenceMethod.Count:
                         return ((DataServiceQuery<TElement>)query).GetValue(this.Context, ParseQuerySetCount<TElement>);
+                    case SequenceMethod.LongCountPredicate:
+                    case SequenceMethod.CountPredicate:
+                        query = CreateQuery<TElement>(NestPredicateExpression(mce));
+                        return ((DataServiceQuery<TElement>)query).GetValue(this.Context, ParseQuerySetCount<TElement>);
                     case SequenceMethod.Any:
                         return GetValueForAny<TElement>(mce);
+                    case SequenceMethod.AnyPredicate:
+                        return GetValueForAny<TElement>(NestPredicateExpression(mce));
                     case SequenceMethod.SumIntSelector:
                     case SequenceMethod.SumDoubleSelector:
                     case SequenceMethod.SumDecimalSelector:
@@ -213,6 +216,52 @@ namespace Microsoft.OData.Client
             queryComponents.GroupByKeySelectorMap = applyQueryOptionExpr?.KeySelectorMap;
 
             return queryComponents;
+        }
+
+        /// <summary>
+        /// Transforms the 'any' query into a 'count' request since OData does not have a spcific query for 'any'.
+        /// Then the result is casted to the corresponding return type (boolean).
+        /// </summary>
+        /// <typeparam name="TElement">The return type.</typeparam>
+        /// <param name="mce">The original expression with predicate.</param>
+        /// <returns></returns>
+        private TElement GetValueForAny<TElement>(MethodCallExpression mce)
+        {
+            Expression arg0 = mce.Arguments[0];
+            Expression countExpression = Expression.Call(
+                typeof(Enumerable),
+                "Count",
+                new Type[] { arg0.Type.GetGenericArguments()[0] },
+                arg0
+            );
+            var query = CreateQuery<TElement>(countExpression) as DataServiceQuery<TElement>;
+            return query.GetValue(Context, ParseQuerySetCount<TElement>);
+        }
+
+        /// <summary>
+        /// Transforms the expression type to one of type 'where'.
+        /// Then it wraps this 'where' expression into one of the received type but without a predicate.
+        /// </summary>
+        /// <param name="mce">The original expression with predicate.</param>
+        /// <returns>The wrapped expression.</returns>
+        private static MethodCallExpression NestPredicateExpression(MethodCallExpression mce)
+        {
+            Type resourceType = mce.Arguments[0].Type.GetGenericArguments()[0];
+
+            Expression where = Expression.Call(
+                typeof(Queryable),
+                "Where",
+                new Type[] { resourceType },
+                mce.Arguments[0],
+                mce.Arguments[1]
+            );
+
+            return Expression.Call(
+                typeof(Enumerable),
+                mce.Method.Name,
+                new Type[] { resourceType },
+                where
+            );
         }
 
         /// <summary>
--- a/test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
@@ -71,41 +71,36 @@ namespace Microsoft.OData.Client.Tests.ALinq
         [Fact]
         public void FirstPredicate()
         {
-            // Test behaviour with one match:
             _ctx.InterceptRequestAndAssertUri("/Customers?$filter=ContactName ne 'John'&$top=1");
             _ctx.InterceptRequestAndMockResponseValue("Customers", "[" + TestContext.MockCustomer1 + "]");
             Customer customer = _ctx.Customers.First(c => c.Name != "John");
             Assert.Equal("ALFKI", customer.Id);
             Assert.Equal("Maria Anders", customer.Name);
             Assert.Equal("Berlin", customer.City);
+        }
 
-            // Test behaviour with no match:
-            bool expectedExceptionThrown = false;
-            try
-            {
-                _ctx.InterceptRequestAndAssertUri("/Customers?$filter=ContactName eq 'thisdoesntexist'&$top=1");
-                _ctx.InterceptRequestAndMockResponseValue("Customers", "[]");
-                _ctx.Customers.First(c => c.Name == "thisdoesntexist");
-            }
-            catch (InvalidOperationException)
-            {
-                expectedExceptionThrown = true;
-            }
-            Assert.True(expectedExceptionThrown);
+        [Fact]
+        public void FirstPredicate_ThrowsException_WhenNoMatchExists()
+        {
+            _ctx.InterceptRequestAndAssertUri("/Customers?$filter=ContactName eq 'thisdoesntexist'&$top=1");
+            _ctx.InterceptRequestAndMockResponseValue("Customers", "[]");
+            Assert.Throws<InvalidOperationException>(() => _ctx.Customers.First(c => c.Name == "thisdoesntexist"));
         }
 
         [Fact]
         public void FirstOrDefaultPredicate()
         {
-            // Test behaviour with one match:
             _ctx.InterceptRequestAndAssertUri("/Customers?$filter=ContactName ne 'John'&$top=1");
             _ctx.InterceptRequestAndMockResponseValue("Customers", "[" + TestContext.MockCustomer1 + "]");
             Customer customer = _ctx.Customers.FirstOrDefault(c => c.Name != "John");
             Assert.Equal("ALFKI", customer.Id);
             Assert.Equal("Maria Anders", customer.Name);
             Assert.Equal("Berlin", customer.City);
+        }
 
-            // Test behaviour with no match:
+        [Fact]
+        public void FirstOrDefaultPredicate_ReturnsNull_WhenNoMatchExists()
+        {
             _ctx.InterceptRequestAndAssertUri("/Customers?$filter=ContactName eq 'John'&$top=1");
             _ctx.InterceptRequestAndMockResponseValue("Customers", "[]");
             Assert.Null(_ctx.Customers.FirstOrDefault(c => c.Name == "John"));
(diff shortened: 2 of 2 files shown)


joecarl:
Should I add tests for this? If so, is there any existing file where I could add the tests or should I create a new one?

nyok1912:
Hi, is there any chance I can download a version including this feature? I am migrating a project from EF to OData and this would save me refactoring hundreds of files. Thanks!

xuzhg:
> Should I add tests for this? If so, is there any existing file where I could add the tests or should I create a new one?

You can find the test file in the Microsoft.OData.Client.Tests project and feel free create a new file if no file existed.

xuzhg (approved):

xuzhg:
> Hi, is there any chance I can download a version including this feature? I am migrating a project from EF to OData and this would save me refactoring hundreds of files. Thanks!

Sorry, since this PR is not merged, and we don't have office version released.   If the nightly package is ok for you, we can trigger a nightly package for you to try.

habbes:
> Should I add tests for this? If so, is there any existing file where I could add the tests or should I create a new one?

Yes, kindly add tests. See the comment from Sam

habbes:
Meanwhile, I'm running a nightly build of this PR to generate signed nightly packages that can be installed from the nightly MyGet feed.

habbes (changes requested):
We should add tests for scenarios that this PR introduces support for.

joecarl:
@habbes I added tests. Please let me know if there's something else I should do. Thanks.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -0,0 +1,209 @@
...
+            _ctx = new TestContext();
+        }
+
+        [Fact]
+        public void Any()
+        {
+            _ctx.InterceptRequestAndAssertUri("/Customers/$count");
+            _ctx.InterceptRequestAndMockResponse("91");
+            bool exists1 = _ctx.Customers.Any();
+            Assert.True(exists1);
```
habbes, line 35: nitpick, but I'd suggest more descriptive variable names when possible
```suggestion
            bool hasCustomers = _ctx.Customers.Any();
            Assert.True(hasCustomers);
```

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -0,0 +1,209 @@
...
+    {
+        private readonly TestContext _ctx;
+
+        public SequenceMethodsTests()
+        {
+            _ctx = new TestContext();
+        }
+
+        [Fact]
+        public void Any()
```
habbes, line 30: Would be good to show both `true` and `false` paths in this test case. You could use a `Theory` for that so you don't have to duplicate the code.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        [Fact]
+        public void FirstOrDefaultPredicate()
+        {
+            _ctx.InterceptRequestAndAssertUri("/Customers?$filter=ContactName eq 'John'&$top=1");
+            _ctx.InterceptRequestAndMockResponseValue("Customers", "[]");
+            Customer c2 = _ctx.Customers.FirstOrDefault(c => c.Name == "John");
+            Assert.Null(c2);
+        }
```
habbes, line 88: Please also add a test case for scenario where `FirstOrDefault` returns a non-null value.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
gathogojr, line 35: Even better, use
```csharp
Assert.NotEmpty(_ctx.Customers);
```

Here and in the other tests that you added, e.g.
```csharp
Assert.NotEmpty(_ctx.Customers.Where(c => c.Name.Contains("ab")));
```

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -0,0 +1,209 @@
...
+        public void LongCountPredicate()
+        {
+            _ctx.InterceptRequestAndAssertUri("/Customers/$count?$filter=contains(ContactName,'ab')");
+            _ctx.InterceptRequestAndMockResponse("6");
+            long count2 = _ctx.Customers.LongCount(c => c.Name.Contains("ab"));
+            Assert.Equal(6, count2);
+        }
+
+        [Fact]
+        public void FirstPredicate()
```
habbes, line 71: What does `FirstPredicate` return when there's no match? Could we test that path as well?

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -0,0 +1,209 @@
...
+        public void FirstOrDefaultPredicate()
+        {
+            _ctx.InterceptRequestAndAssertUri("/Customers?$filter=ContactName eq 'John'&$top=1");
+            _ctx.InterceptRequestAndMockResponseValue("Customers", "[]");
+            Customer c2 = _ctx.Customers.FirstOrDefault(c => c.Name == "John");
+            Assert.Null(c2);
+        }
+
+        [Fact]
+        public void SinglePredicate()
```
habbes, line 91: What does `SinglePredicate` return when there's no match or when there are multiple matches? Could we test those paths as well?

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -0,0 +1,209 @@
...
+            Assert.False(exists3);
+        }
+
+        [Fact]
+        public void CountPredicate()
+        {
+            _ctx.InterceptRequestAndAssertUri("/Customers/$count?$filter=contains(ContactName,'ab')");
+            _ctx.InterceptRequestAndMockResponse("6");
+            int count1 = _ctx.Customers.Count(c => c.Name.Contains("ab"));
+            Assert.Equal(6, count1);
```
habbes, line 58:
```suggestion
            int count = _ctx.Customers.Count(c => c.Name.Contains("ab"));
            Assert.Equal(6, count);
```

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -0,0 +1,209 @@
...
+        }
+
+        [Fact]
+        public void LongCountPredicate()
+        {
+            _ctx.InterceptRequestAndAssertUri("/Customers/$count?$filter=contains(ContactName,'ab')");
+            _ctx.InterceptRequestAndMockResponse("6");
+            long count2 = _ctx.Customers.LongCount(c => c.Name.Contains("ab"));
+            Assert.Equal(6, count2);
```
habbes, line 67:
```suggestion
            long count = _ctx.Customers.LongCount(c => c.Name.Contains("ab"));
            Assert.Equal(6, count);
```

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -0,0 +1,209 @@
...
+using System.Text;
+using Microsoft.OData.Edm;
+using Xunit;
+
+namespace Microsoft.OData.Client.Tests.ALinq
+{
+    /// <summary>
+    /// Tests to check sequence methods support.
+    /// </summary>
+    public class SequenceMethodsTests
```
gathogojr, line 20: I think in the tests below, we should verify that OData URL that the expression is translated to. We could for example attach a delegate to the `DataServiceContext` instance `SendingRequest2` event, e.g.:
```csharp
_ctx.SendingRequest2 += (sender, e) =>
{
    Assert.EndsWith("Customers/$count", e.RequestMessage.Url.OriginalString);
};
```
That way we're confident that the expression is being translated into what we expect

joecarl: I'm already doing that, I created the function `InterceptRequestAndAssertUri` which sets the expected uri. I also attached a delegate which verifies the built uri matches the expected one.

joecarl:
@habbes I modified the tests in order to accomplish all your suggestions. Please let me know if there's something else I should do. Thanks.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -0,0 +1,290 @@
...
+            try
+            {
+                _ctx.InterceptRequestAndAssertUri("/Customers?$filter=ContactName eq 'thisdoesntexist'&$top=1");
+                _ctx.InterceptRequestAndMockResponseValue("Customers", "[]");
+                _ctx.Customers.First(c => c.Name == "thisdoesntexist");
+            }
+            catch (InvalidOperationException)
+            {
+                expectedExceptionThrown = true;
+            }
```
habbes, line 93: This could be a separate test method. And we could use `Assert.Throws<InvalidOperationException>` to ensure 

```c#
[Fact]
public void FirstPredicate_ThrowsException_WhenNoMatchExists()
{
     _ctx.InterceptRequestAndAssertUri("/Customers?$filter=ContactName eq 'thisdoesntexist'&$top=1");
      _ctx.InterceptRequestAndMockResponseValue("Customers", "[]");
      Assert.Throws<InvalidOperationException>(()  => _ctx.Customers.First(c => c.Name == "thisdoesntexist"));
}
```

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -0,0 +1,290 @@
...
+            _ctx.InterceptRequestAndMockResponseValue("Customers", "[" + TestContext.MockCustomer1 + "]");
+            Customer customer = _ctx.Customers.FirstOrDefault(c => c.Name != "John");
+            Assert.Equal("ALFKI", customer.Id);
+            Assert.Equal("Maria Anders", customer.Name);
+            Assert.Equal("Berlin", customer.City);
+
+            // Test behaviour with no match:
+            _ctx.InterceptRequestAndAssertUri("/Customers?$filter=ContactName eq 'John'&$top=1");
+            _ctx.InterceptRequestAndMockResponseValue("Customers", "[]");
+            Assert.Null(_ctx.Customers.FirstOrDefault(c => c.Name == "John"));
```
habbes, line 111: Not a blocker, but I think this ought to be a separate test method. Makes debugging easier when tests fail.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -0,0 +1,290 @@
...
+                expectedExceptionThrown = false;
+                _ctx.InterceptRequestAndAssertUri("/Customers?$filter=ContactName ne 'thisdoesntexist'&$top=2");
+                _ctx.InterceptRequestAndMockResponseValue("Customers", "[" + TestContext.MockCustomer1 + "," + TestContext.MockCustomer2 + "]");
+                _ctx.Customers.Single(c => c.Name != "thisdoesntexist");
+            }
+            catch (InvalidOperationException)
+            {
+                expectedExceptionThrown = true;
+            }
+            Assert.True(expectedExceptionThrown);
```
habbes, line 153: Nitpick: Consider refactoring these out into separate methods. Also consider `Assert.Throws` to test exceptions. See: https://github.com/OData/odata.net/pull/2863/files#r1618197588

joecarl:
@habbes I separated and refactored the tests in order to accomplish all your suggestions. Please let me know if there's something else I should do. Thanks.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -0,0 +1,266 @@
...
+            Assert.Equal("Berlin", customer.City);
+        }
+
+        [Fact]
+        public void FirstPredicate_ThrowsException_WhenNoMatchExists()
+        {
+            _ctx.InterceptRequestAndAssertUri("/Customers?$filter=ContactName eq 'thisdoesntexist'&$top=1");
+            _ctx.InterceptRequestAndMockResponseValue("Customers", "[]");
+            Assert.Throws<InvalidOperationException>(() => _ctx.Customers.First(c => c.Name == "thisdoesntexist"));
+        }
```
gathogojr, line 88: @joecarl You should make your tests easier to follow. The following block of code has not place in the constructor:
```csharp
            _ctx.BuildingRequest += (obj, args) =>
            {
                if (_assertRequestUriPath == "") return;
                string expectedUri = _rootUriStr + _assertRequestUriPath;
                string actualUri = args.RequestUri.ToString();
                Assert.Equal(expectedUri, actualUri);
                _assertRequestUriPath = "";
            };
```
Your approach makes is making the logic become more complex than it needs to be. Consider this:
```suggestion
        public void FirstPredicate_ThrowsException_WhenNoMatchExists()
        {
            _ctx.InterceptRequestAndMockResponseValue("Customers", "[]");
            _ctx.BuildingRequest += (obj, args) =>
            {
                Assert.EndsWith("/Customers?$filter=ContactName eq 'thisdoesntexist'&$top=1", args.RequestUri.OriginalString);
            };

            Assert.Throws<InvalidOperationException>(() => _ctx.Customers.First(c => c.Name == "thisdoesntexist"));
        }
```
Clean and simple - The assertions are within the test and the logic is not complicated

joecarl: Asserting with `EndsWith` is not correct, that could assert bad URIs. The whole URI (or at least the whole path and queryString) should be checked.
Also, the event listener should be removed after assertion. (Unless tests run over different instances).
Fixing your code in order to address this issues will add more boilerplate to every single test. Are you sure it is worth it?

joecarl:
@gathogojr I ended up refactoring in a way which I think will align with your expectations. Please have a look and let me know if I should do something else.

joecarl:
@habbes is there something else I should do here? Thanks.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -0,0 +1,336 @@
...
+        [OriginalName("CustomerID")]
+        public string Id { get; set; }
+
+        public string City { get; set; }
+
+        [OriginalName("ContactName")]
+        public string Name { get; set; }
+    }
+
+    public class TestContext
```
gathogojr, line 249: While there are many ways of achieving the same goal, I can't quite understand what value this `TestContext` class brings. In my opinion, it has made the logic more complex that it needs to be - for example having to introduce `Action<string> OnRequestUriBuilt` delegate field to fulfill the requirement to validate the request Uri. It does help to familiarize with how things are done elsewhere in the codebase, otherwise we end up with code that is not easy to follow or maintain. Even where a custom "context" class is needed, the common practice is to subclass `DataServiceContext` instead of composition like you're doing here.

I understand what you are trying to achieve by introducing the `TestContext` but the same could easily be achieved differently, easily, neatly, and with less complexity...
E.g.
```csharp
// From the test class constructor
_ctx = new DataServiceContext();
var model = BuildEdmModel();
_ctx.Format.UseJson(model);
_ctx.ResolveName = (type) => $"NS.{type.Name}";
_ctx.KeyComparisonGeneratesFilterQuery = true;

// ... other code

// From test method
// DO: Call InterceptRequestAndMockResponse
// DO: _ctx.OnBuildingRequest += ...
Customer customer _ctx.CreateQuery<Customer>("Customers").FirstOrDefault(c => c.Name == "John")

// DO: Assert statements
```

**No need to use TestContext anywhere**

Please take a look at this for reference: https://github.com/OData/odata.net/blob/383e29f3cb96eda22e76d19df7bdab11459377b8/test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/DollarApplyTestsBase.cs#L27-L42

joecarl: I wanted to clarify the reasoning behind the implementation of the TestContext class and address some of the concerns raised.

I implemented the TestContext class to separate the logic of the test methods from the connection logic to the data service. This separation aims to bring more cleanliness and organization to the code, contrary to the opinion expressed. The goal is to keep the test methods focused solely on testing, while the mock connection and setup logic is handled elsewhere.

The logic in this file is extremely simple and should not pose a problem for any maintainer. In fact, when comparing the complexity of this file to the overall project (OData), the complexity here is minimal. I find the continuous requests for changes a bit excessive and not entirely justified given the simplicity of the implementation.

Regarding Inheritance, I agree with the point that TestContext could extend DataServiceContext. However, I did not implement it this way initially because I adapted this class from another file where I needed to use four different contexts. I can modify this, but I sincerely believe it is not necessary.

I would appreciate it if we could focus on impactful change requests rather than minor issues. If there are substantial improvements to be made, I am more than willing to address them, but I feel that some of the requests so far have been less about meaningful improvements and more about nitpicking.

Thank you for your understanding.

xuzhg (approved):

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
habbes: Let me contribute my thoughts to this. While there have been a lot of change requests to what is seemingly a simple change, I would argue that those changes have meaningfully improved the completeness and readability of the tests. More edge cases are handled, and tests are easier to read now than in the initial PR.

On this particular issue of the `TestContext`, I think the helper methods like `InterceptRequestAndMockResponse` serve to declutter the test code from boilerplate and noise. However, I don't think having such helper methods defined in the test class, and initial setup code in the test class constructor would be any less readable than having them in a new `TestContext` class. We can easily distinguish private helper methods from test methods with proper organization in the class and test attributes (`[Fact]`, `[Theory]`). For that reason, I think having a separate `TestContext` class that's only used one is unnecessary. It would see the benefit if the helpers were re-used in multiple test classes, but that's not the case here. I would argue that having classes we don't need does impact maintainability because it increases the surface areas of things we're potentially concerned about and things that would cause us to ask "So, what do you really do here?" or wonder whether it's safe to change. The less such things we have, the better.

That said, conceding that my view on this particular change is subjective, I don't think it's worthwhile to hold the change back from going in.

habbes (approved):

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
habbes: @joecarl xUnit instatiates the test class for each test. So, you would not need to write boilerplate per test case. Currently the `TestContext` does not remove the listener either. So, the test code would not have to change significantly if you removed the `TesContext` class.

I do agree with you that testing the entire url is more robust than using `EndsWith`.

joecarl: Thank you very much for your clarification, it really helps me learn. I understand your point of view, and although I believe a separate class can help with better code reuse and scalability in the future, I also understand your perspectives. In my future contributions, I will avoid creating an additional class to encapsulate this kind of functionality.

test/FunctionalTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
habbes: Unfortunately, I have changed my view on this matter. Having gone through this other PR: https://github.com/OData/odata.net/pull/2862 which has a similar pattern, I'm concerned about letting this in as is and then resulting in a proliferation of `TestContext` pattern in future tests. I think it would be easier to refactor it at this point.

habbes (commented):
Having gone through this other PR: #2862 which has a similar pattern, I'm concerned about letting this in as is and then resulting in a proliferation of redundant `TestContext` classes in future tests. I think it would be easier to refactor it at this point.

joecarl:
> Having gone through this other PR: #2862 which has a similar pattern, I'm concerned about letting this in as is and then resulting in a proliferation of redundant `TestContext` classes in future tests. I think it would be easier to refactor it at this point.

Refactored

habbes (approved):

src/Microsoft.OData.Client/ALinq/DataServiceQueryProvider.cs
```diff
@@ -196,6 +218,52 @@ internal QueryComponents Translate(Expression e)
...
+            return query.GetValue(Context, ParseQuerySetCount<TElement>);
+        }
+
+        /// <summary>
+        /// Transforms the expression type to one of type 'where'.
+        /// Then it wraps this 'where' expression into one of the received type but without a predicate.
+        /// </summary>
+        /// <param name="mce">The original expression with predicate.</param>
+        /// <returns>The wrapped expression.</returns>
+        private static MethodCallExpression NestPredicateExpression(MethodCallExpression mce)
```
gathogojr, line 247: nit: A better name given the method doc would be "TransformToWhereExpression" or "ConstructWhereExpression" or "ReconstructAsWhereExpression"

src/Microsoft.OData.Client/ALinq/DataServiceQueryProvider.cs
```diff
@@ -196,6 +218,52 @@ internal QueryComponents Translate(Expression e)
...
+        {
+            Type resourceType = mce.Arguments[0].Type.GetGenericArguments()[0];
+
+            Expression where = Expression.Call(
```
gathogojr, line 251: nit: `where` is a reserved keyword. Maybe "whereMethod" would be a good substitute

gathogojr (approved):

joecarl:
Should I create the corresponding cherry pick PR for the default branch?

habbes:
@joecarl had not seen your comment, I went ahead and just created one https://github.com/OData/odata.net/pull/3061

PR #2864, opened 2024-02-20
[ODL 8]: Use `$filter` for all `Where` clauses and remove `KeyComparisonGeneratesFilter` flag

### LINKED ISSUE #2208: Remove DataServiceContext.KeyComparisonGeneratesFilterQuery in the next breaking change release.

### Assemblies affected

Remove `DataServiceContext.KeyComparisonGeneratesFilterQuery` in the next breaking change release. This was added so as not to break existing clients.

https://github.com/OData/odata.net/blob/59cc9b6526cc3f43b638c2b15dc9060000c86579/src/Microsoft.OData.Client/DataServiceContext.cs#L684-L690

### Reproduce steps

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

habbes:
### Issues

*This pull request fixes #2208*

### Description

- Removes the `DataServiceContext.KeyComparisonGeneratesFilter` property
- Translates `Where` clause into a `$filter` query option even when the where clause only does a key lookup. Previous, this would generate a key segment instead of a `$filter` unless the flag above was set to `true`.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Client/ALinq/QueryableResourceExpression.cs
+++ b/src/Microsoft.OData.Client/ALinq/QueryableResourceExpression.cs
@@ -156,12 +156,6 @@ namespace Microsoft.OData.Client
         }
 
         /// <summary>
-        /// If this expresssion contains at least one non-key predicate
-        /// This indicates that a filter should be used
-        /// </summary>
-        internal bool UseFilterAsPredicate { get; set; }
-
-        /// <summary>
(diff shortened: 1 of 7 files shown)


habbes:
Closing this because we opted to retain the `KeyComparisonGeneratesFilter` flag in ODL 8 but change the default from `false` to `true` and potential mark it as deprecated.

PR #2865, opened 2024-02-20
[ODL 8]: Deprecate `KeyComparisonGeneratesFilterQuery` and set default to `true`.

habbes:
### Issues

Partially addresses #2208 

### Description

- Marks the `DataServiceContext.KeyComparisonGeneratesFilterQuery` property as deprecated for removal in the next major release. This property was added to avoid breaking changes when we made Where clause generate a `$filter` query even when the predicate was only a key lookup.
- Changes the default value of the property from `false` to `true`.

- [x] *Test cases added*


habbes:
Accidentally created the PR against the `main` branch instead of the `dev-8.x` branch.

PR #2866, opened 2024-02-20
[ODL 8]: Deprecate `KeyComparisonGeneratesFilterQuery` and set default to `true`.

habbes:
### Issues

Partially addresses #2208 

### Description

- Marks the `DataServiceContext.KeyComparisonGeneratesFilterQuery` property as deprecated for removal in the next major release. This property was added to avoid breaking changes when we made Where clause generate a `$filter` query even when the predicate was only a key lookup.
- Changes the default value of the property from `false` to `true`.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -288,7 +288,7 @@ namespace Microsoft.OData.Client
             this.httpStack = HttpStack.Auto;
             this.UsingDataServiceCollection = false;
             this.UsePostTunneling = false;
-            this.keyComparisonGeneratesFilterQuery = false;
+            this.keyComparisonGeneratesFilterQuery = true;
             this.deleteLinkUriOption = DeleteLinkUriOption.IdQueryParam;
         }
 
@@ -696,8 +696,10 @@ namespace Microsoft.OData.Client
         public virtual bool EnableWritingODataAnnotationWithoutPrefix { get; set; }
 
         /// <summary>
-        /// Indicates whether a Where clause that just compares the key property generates a $filter query option.
+        /// When true, a Where clause that just compares the key property generates a $filter query option, otherwise a key segment is generated.
+        /// The default value is true.
         /// </summary>
+        [Obsolete("This property will be removed in a future major release.")]
         public virtual bool KeyComparisonGeneratesFilterQuery
         {
             get { return this.keyComparisonGeneratesFilterQuery; }
--- a/src/Microsoft.OData.Client/DataServiceContext.cs
+++ b/src/Microsoft.OData.Client/DataServiceContext.cs
@@ -696,7 +696,7 @@ namespace Microsoft.OData.Client
         public virtual bool EnableWritingODataAnnotationWithoutPrefix { get; set; }
 
         /// <summary>
-        /// When true, a Where clause that just compares the key property generates a $filter query option, otherwise a key segment is generated.
+        /// When true, a Where clause that has only the key property in the predicate generates a $filter query option, otherwise a key segment is generated.
         /// The default value is true.
         /// </summary>
         [Obsolete("This property will be removed in a future major release.")]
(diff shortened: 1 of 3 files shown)


test/EndToEndTests/Tests/Client/AsynchronousTests/AsynchronousQueryTests.cs
```diff
@@ -1323,6 +1323,8 @@ public void Linq_Where_GreaterThan()
         public void Linq_Where_Generates_Filter_When_KeyComparisonGeneratesFilterQuery_Is_True()
         {
             var context = this.CreateWrappedContext<DefaultContainer>().Context;
+            // By default context.KeyComparisonGeneratesFilterQuery = true;
+            // So we don't need to set it
```
KenitoInc, line 1327: Remove line 1328

KenitoInc (commented):
Minor comment
Otherwise LGTM

test/EndToEndTests/Tests/Client/AsynchronousTests/AsynchronousQueryTests.cs
habbes: Thanks. I thought I had.

KenitoInc (approved):

ElizabethOkerio (approved):

src/Microsoft.OData.Client/DataServiceContext.cs
```diff
@@ -696,8 +696,10 @@ public virtual EntityTracker EntityTracker
         public virtual bool EnableWritingODataAnnotationWithoutPrefix { get; set; }
 
         /// <summary>
-        /// Indicates whether a Where clause that just compares the key property generates a $filter query option.
+        /// When true, a Where clause that just compares the key property generates a $filter query option, otherwise a key segment is generated.
```
gathogojr, line 699:
```suggestion
        /// When true, a Where clause that has only the key property in the predicate generates a $filter query option, otherwise a key segment is generated.
```

test/EndToEndTests/Tests/Client/AsynchronousTests/AsynchronousQueryTests.cs
```diff
@@ -1323,7 +1323,8 @@ public void Linq_Where_GreaterThan()
         public void Linq_Where_Generates_Filter_When_KeyComparisonGeneratesFilterQuery_Is_True()
         {
             var context = this.CreateWrappedContext<DefaultContainer>().Context;
-            context.KeyComparisonGeneratesFilterQuery = true;
+            // By default context.KeyComparisonGeneratesFilterQuery = true;
+            // So we don't need to set it
```
gathogojr, line 1327: Remove the two comments. I don't think they're necessary

test/FunctionalTests/Microsoft.OData.Client.Tests/DataServiceContextTests.cs
```diff
@@ -0,0 +1,14 @@
+﻿using Xunit;
+
+namespace Microsoft.OData.Client.Tests
+{
+    public class DataServiceContextTests
+    {
+        [Fact]
+        public void KeyComparisonGeneratesFilterQuery_Defaults_To_True()
```
gathogojr, line 8: You couldn't find a home for this 1 test where other tests for `DataServiceContext` are domiciled?

gathogojr (approved):
Left a few comments, otherwise LGTM

test/FunctionalTests/Microsoft.OData.Client.Tests/DataServiceContextTests.cs
habbes: The other `DataServiceContext`-related tests are in the E2E tests. I thought it would be good to have some coverage in the unit tests as well but didn't find a relevant class, so I created one.

gathogojr (approved):

KenitoInc (approved):

Issue #2599, opened 2023-01-17
Default leaveOpen to false in ODataBinaryStreamValue constructor.


mikepizzo:
*#2597 adds the a constructor overload to specify that the stream wrapped by an ODataBinaryStreamValue is automatically disposed when the ODataBinaryStreamValue is disposed.  Because this optional parameter was added later, and the previous behavior was not to dispose the underlying stream, we defaulted to not disposing the stream.  However, it would be more consistent with things like StreamWriter to default to taking over ownership and disposing, so in 8.x we it probably make sense to make the default false (don't leave open).*

### Assemblies affected

*OData .Net lib 7.x*

### Reproduce steps

*Create ODataBinaryStreamValue, passing a stream but without specifying a value for the optional parameter leaveOpen.*

### Expected result

*Like StreamWriter, disposing the ODataBinaryStreamValue should call dispose on the underyling stream.*

### Actual result

*The underlying stream is only disposed if the ODataBinaryStreamValue constructor is called with leaveOpen=false.  leaveOpen=false should be the default, in order to be consistent with other APIs like StreamWriter.*

### Additional detail

*We weren't able to make leaveOpen=false the default when we introduced the optional leaveOpen parameter for backwards compatibility. We can change this in 8.x.*

gathogojr:
Fixed by https://github.com/OData/odata.net/pull/2868

PR #2868, opened 2024-02-20
Default leaveOpen to false in ODataBinaryStreamValue constructor

ElizabethOkerio:
### Issues

*This pull request fixes https://github.com/OData/odata.net/issues/2599*

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Value/ODataBinaryStreamValue.cs
+++ b/src/Microsoft.OData.Core/Value/ODataBinaryStreamValue.cs
@@ -20,7 +20,7 @@ namespace Microsoft.OData
         /// </summary>
         /// <param name="stream">Input stream</param>
         public ODataBinaryStreamValue(Stream stream) :
-            this(stream, true) // Leaves the stream open by default, for backward compatibility.
+            this(stream, leaveOpen: false)
         {
         }
 
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightEntryAndFeedSerializerUndecalredTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightEntryAndFeedSerializerUndecalredTests.cs
@@ -235,6 +235,29 @@ namespace Microsoft.Test.OData.TDD.Tests.Writer.JsonLight
             Assert.Equal(result, actual);
         }
 
+        [Fact]
+        public void ODataBinaryStreamValue_Leaves_Stream_Not_Open_By_Default()
+        {
+            // With type name
+            // ODataBinaryStreamValue
+            var stream = new MemoryStream();
+            var writer = new BinaryWriter(stream);
+            writer.Write("1234567890");
+            writer.Flush();
+            stream.Position = 0;
+
+            var property = new ODataProperty
+            {
+                Name = "Data",
+                Value = new ODataBinaryStreamValue(stream)
+            };
+
+            string result = WriteDeclaredUntypedProperty(property);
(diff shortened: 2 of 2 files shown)


src/Microsoft.OData.Core/Value/ODataBinaryStreamValue.cs
```diff
@@ -20,7 +20,7 @@ public sealed class ODataBinaryStreamValue : ODataValue
         /// </summary>
         /// <param name="stream">Input stream</param>
         public ODataBinaryStreamValue(Stream stream) :
-            this(stream, true) // Leaves the stream open by default, for backward compatibility.
+            this(stream, false) // Leaves the stream open by default, for backward compatibility.
```
habbes, line 23:
```suggestion
            this(stream, leaveOpen: false)

```

src/Microsoft.OData.Core/Value/ODataBinaryStreamValue.cs
```diff
@@ -20,7 +20,7 @@ public sealed class ODataBinaryStreamValue : ODataValue
         /// </summary>
         /// <param name="stream">Input stream</param>
         public ODataBinaryStreamValue(Stream stream) :
-            this(stream, true) // Leaves the stream open by default, for backward compatibility.
+            this(stream, leaveOpen: false)
```
mikepizzo, line 23: >leaveOpen: false [](http://example.com/codeflow?start=25&length=16)

are there no tests that need to be updated by changing this default? can we add tests to verify this default?

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightEntryAndFeedSerializerUndecalredTests.cs
```diff
@@ -235,6 +235,29 @@ public void WriteResourceCollectionUntypedProperty_WorksResourceInNestedCollecti
             Assert.Equal(result, actual);
         }
 
+        [Fact]
+        public void ODataBinaryStreamValue_Leaves_Stream_Not_Open_By_Default()
```
habbes, line 239:
```suggestion
        public void ODataBinaryStreamValue_Closes_Stream_By_Default()

```

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightEntryAndFeedSerializerUndecalredTests.cs
```diff
@@ -235,6 +235,29 @@ public void WriteResourceCollectionUntypedProperty_WorksResourceInNestedCollecti
...
+                Name = "Data",
+                Value = new ODataBinaryStreamValue(stream)
+            };
+
+            string result = WriteDeclaredUntypedProperty(property);
+            Assert.Equal("{\"@odata.context\":\"http://www.sampletest.com/$metadata#serverEntitySet/$entity\",\"Data\":\"CjEyMzQ1Njc4OTA=\"}", result);
+
+            Assert.False(stream.CanRead);
+        }
+
```
habbes, line 260: Is there a test that shows if you set `leaveOpen: true` that it won't automatically close the stream? If not, I think you should add that as well.

ElizabethOkerio: It's there: https://github.com/OData/odata.net/blob/main/test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightValueSerializerTests.cs#L320

habbes (approved):

gathogojr (approved):

PR #2869, opened 2024-02-20
[ODL 8]: Merge JsonReader interfaces into one

### LINKED ISSUE #2435: Merge default Json reader synchronous and asynchronous interfaces
Merge default Json reader synchronous and asynchronous interfaces. Approach taken during implementation of the async API was intended to prevent a breaking change.
- Merge `IJsonReader` and `IJsonReaderAsync` interfaces
- Merge `IJsonReaderFactory` and `IJsonReaderFactoryAsync` interfaces
- Merge `IJsonStreamReader` and `IJsonStreamReaderAsync` interfaces

### Assemblies affected

Microsoft.OData.Core 7.x*

### Reproduce steps

N/A

### Expected result

N/A

### Actual result

N/A

habbes:
### Issues

*This pull request fixes #2435 *

### Description

- Merges `IJsonReader`, `IJsonReaderAsync`, `IJsonStreamReader`, `IJsonStreamReaderAsync` into a single `IJsonReader` interface. I updated internal classes and tests to reflect this.
- We had scenarios where we would check whether the `IJsonReader` implements `IStreamReader` and if it does, we create a read stream, otherwise we read the contents into an in-memory stream as a fallback. Since we're not guaranteed that `IJsonReader` has the streaming APIs, there's no need to perform these fallbacks so I removed them.
- Implements `IAsyncDisposable` on `JsonReader` class. I did this mostly for completeness and consistency with the writer class since the `JsonReader.DisposeAsync()` method doesn't perform any I/O or async operation.
- Changed `IJsonReader.Value` property into `IJsonReader.GetValue()` method. Since the property actually performed I/O, it made more sense as a method than a property. It also closely mirrors the async counterpart `GetValueAsync` now.

- [x] *Test cases added*


habbes:
Closed because CI pipeline seems messed up as a result of targeting the wrong branch when I created the PR

PR #2870, opened 2024-02-20
[ODL 8]: Merge JsonReader interfaces into one

### LINKED ISSUE #2435: Merge default Json reader synchronous and asynchronous interfaces

habbes:
### Issues

*This pull request fixes #2435 *

### Description

- Merges `IJsonReader`, `IJsonReaderAsync`, `IJsonStreamReader`, `IJsonStreamReaderAsync` into a single `IJsonReader` interface. I updated internal classes and tests to reflect this.
- We had scenarios where we would check whether the `IJsonReader` implements `IStreamReader` and if it does, we create a read stream, otherwise we read the contents into an in-memory stream as a fallback. Since we're not guaranteed that `IJsonReader` has the streaming APIs, there's no need to perform these fallbacks so I removed them.
- Implements `IAsyncDisposable` on `JsonReader` class. I did this mostly for completeness and consistency with the writer class since the `JsonReader.DisposeAsync()` method doesn't perform any I/O or async operation.
- Changed `IJsonReader.Value` property into `IJsonReader.GetValue()` method. Since the property actually performed I/O, it made more sense as a method than a property. It also closely mirrors the async counterpart `GetValueAsync` now.

- [x] *Test cases added*


test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/JsonReaderWriterInjectionTests.cs
```diff
@@ -610,6 +610,46 @@ public bool Read()
...
+
+            public Task<TextReader> CreateTextReaderAsync()
+            {
+                throw new NotImplementedException();
+            }
+
+            public Task<object> GetValueAsync()
+            {
+                throw new NotImplementedException();
+            }
```
gathogojr, line 652: We could use the terse form, e.g.,
```csharp
public Task<bool> ReadAsync() => throw new NotImplementedException();
```
It keeps the file shorter.

src/Microsoft.OData.Core/Json/IJsonReader.cs
```diff
@@ -4,6 +4,9 @@
 // </copyright>
 //---------------------------------------------------------------------
 
+using System.IO;
```
gathogojr, line 7: nit: Across this project, I see imported namespaces being nested within the namespace block

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -195,14 +182,13 @@ internal bool IsBuffering
...
-            if (!this.isBuffering && streamReader != null)
+            if (!this.isBuffering)
             {
-                return streamReader.CreateReadStream();
+                return this.innerReader.CreateReadStream();
             }
 
-            Stream result = this.Value == null ? Stream.Null :
-                new MemoryStream(Convert.FromBase64String((string)this.Value));
+            Stream result = this.GetValue() == null ? Stream.Null :
```
gathogojr, line 190: I think it'd be better to call `this.GetValue()` and store the result in a variable instead of calling it twice in the same statement in case the two calls result into reading from the stream twice

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -213,13 +199,12 @@ public virtual Stream CreateReadStream()
...
-            IJsonStreamReader streamReader = this.innerReader as IJsonStreamReader;
-            if (!this.isBuffering && streamReader != null)
+            if (!this.isBuffering)
             {
-                return streamReader.CreateTextReader();
+                return this.innerReader.CreateTextReader();
             }
 
-            TextReader result = new StringReader(this.Value == null ? "" : (string)this.Value);
+            TextReader result = new StringReader(this.GetValue() == null ? "" : (string)this.GetValue());
```
gathogojr, line 207: See my comment here https://github.com/OData/odata.net/pull/2870/files#r1496926105

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -230,13 +215,12 @@ public virtual TextReader CreateTextReader()
...
-            IJsonStreamReader streamReader = this.innerReader as IJsonStreamReader;
-            if (!this.isBuffering && streamReader != null)
+            if (!this.isBuffering)
             {
-                return streamReader.CanStream();
+                return this.innerReader.CanStream();
             }
 
-            return (this.Value is string || this.Value == null || this.NodeType == JsonNodeType.StartArray || this.NodeType == JsonNodeType.StartObject);
+            return (this.GetValue() is string || this.GetValue() == null || this.NodeType == JsonNodeType.StartArray || this.NodeType == JsonNodeType.StartObject);
```
gathogojr, line 223: See my comment here https://github.com/OData/odata.net/pull/2870/files#r1496926105

src/Microsoft.OData.Core/JsonLight/ODataJsonLightBatchBodyContentReaderStream.cs
```diff
@@ -387,9 +387,9 @@ private static void WriteCurrentJsonObject(IJsonReader reader, IJsonWriter jsonW
                 {
                     case JsonNodeType.PrimitiveValue:
                         {
-                            if (reader.Value != null)
+                            if (reader.GetValue() != null)
```
gathogojr, line 390: See my comment here https://github.com/OData/odata.net/pull/2870/files#r1496926105

src/Microsoft.OData.Core/JsonLight/ODataJsonLightResourceDeserializer.cs
```diff
@@ -173,9 +173,9 @@ internal ODataDeletedResource ReadDeletedResource()
                             }
                         }
                     }
-                    else if (this.JsonReader.Value != null)
+                    else if (this.JsonReader.GetValue() != null)
```
gathogojr, line 176: See my comment here https://github.com/OData/odata.net/pull/2870/files#r1496926105

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -464,40 +464,14 @@ await this.ReadNextResourceSetItemAsync()
...
-            }
-            else
-            {
-                // JSONReader doesn't support streaming; read as a string and convert
-                // Skip over property or start array
-                this.jsonLightInputContext.JsonReader.Read();
-                string valueAsString = this.jsonLightInputContext.JsonReader.ReadStringValue();
-                reader = new StringReader(valueAsString);
-            }
+            TextReader reader = this.jsonLightInputContext.JsonReader.CreateTextReader();
```
gathogojr, line 474:
```suggestion
            return this.jsonLightInputContext.JsonReader.CreateTextReader();
```

src/Microsoft.OData.Core/JsonLight/ODataJsonLightReader.cs
```diff
@@ -464,40 +464,14 @@ await this.ReadNextResourceSetItemAsync()
...
-            }
-            else
-            {
-                // JSONReader doesn't support streaming; read as a string and convert
-                // Skip over property or start array
-                this.jsonLightInputContext.JsonReader.Read();
-                string valueAsString = this.jsonLightInputContext.JsonReader.ReadStringValue();
-                stream = new MemoryStream(Convert.FromBase64String(valueAsString.Replace('_', '/').Replace('-', '+')));
-            }
+            Stream stream = this.jsonLightInputContext.JsonReader.CreateReadStream();
```
gathogojr, line 467:
```suggestion
            return this.jsonLightInputContext.JsonReader.CreateReadStream();
```

src/Microsoft.OData.Core/Json/JsonReader.cs
```diff
@@ -22,7 +22,11 @@ namespace Microsoft.OData.Json
     /// Reader for the JSON format. http://www.json.org
     /// </summary>
     [DebuggerDisplay("{NodeType}: {Value}")]
```
gathogojr, line 24: Check if `DebuggerDisplay` attribute can accommodate the `GetValue` method and if it cannot remove the `Value` property since it no longer exists

habbes: Using `GetValue()` works as expected. Let me update the PR:

image

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -127,24 +117,21 @@ public JsonNodeType NodeType
         /// Depending on whether buffering is on or off this will return the node type of the last
         /// buffered read or the node type of the last unbuffered read.
         /// </remarks>
-        public object Value
+        public object GetValue()
```
ElizabethOkerio, line 120: Is this a necessary change? Are there any benefits of changing thit to a method?

ElizabethOkerio:
There are no E2E tests that need updating?

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
habbes: The `Value` property is not a simple a property, it actually performs an I/O operation to read a value from the stream. The change from `Value` property to `GetValue()` method serves to better reflect that this method does some heavy lifting. It also serves to create better parity with the async version, which is `GetValueAsync()`. This change was proposed by @gathogojr

habbes:
> There are no E2E tests that need updating?

Not really. E2E are mostly affected by OData client and public config changes.

ElizabethOkerio (approved):

gathogojr:
Fixed by https://github.com/OData/odata.net/pull/2870

Issue #2822, opened 2023-12-01
Consider making ODataUtf8JsonWriter the default JsonWriter


habbes:
Since `ODataUtf8JsonWriter` has better performance and uses less memory, we should consider using its implementation as out default JsonWriter. However, before we do that, it should implement all the features of the default JsonWriter, namely `IJsonStreamWriter`: #2422

gathogojr:
Fixed by https://github.com/OData/odata.net/pull/2980

PR #2874, opened 2024-02-22
IJsonWriterFactory.CreateJsonWriter method to accept Stream instead of TextWriter

### LINKED ISSUE #2821: JsonWriter factories should accept `Stream` instead of `TextWriter`.
We have two interfaces for creating JsonWriter, `IJsonWriterFactory` that accepts an `TextWriter` for the output, and an `IJsonStreamBasedWriter` that accepts a `Stream`. The latter was created because we need to support passing a `Stream` without introduce breaking changes. But passing a stream makes more sense generally, if the implementation uses a TextWriter, it can wrap one on top of the stream. So we only need one interface. We can remove unnecessary interfaces in a future major release.

### Assemblies affected

### Reproduce steps

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

gathogojr:
### Issues

*This pull request fixes #2821.*

### Description

`IJsonWriterFactory.CreateJsonWriter` method to accept `Stream` instead of `TextWriter`

In this PR,
- I propose to drop [`IStreamBasedJsonWriterFactory` interface](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/Json/IStreamBasedJsonWriterFactory.cs#L18). All JSON writers will implement `IJsonWriterFactory` in ODL 8.
- I propose to rename [`DefaultStreamBasedJsonWriterFactory` class](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/Json/DefaultStreamBasedJsonWriterFactory.cs#L21) to `ODataUtf8JsonWriterFactory`. All JSON writers have a `CreateJsonWriter` method that accepts a `Stream`, therefore, naming a specific one `StreamBasedJsonWriterFactory` is no longer intuitive.

- An internal [`ODataJsonLightOutputContext` constructor](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs#L174) that accepts a `TextWriter` as the first parameter will instead accept a `Stream` parameter instead. The constructor is invoked from [`ODataUriConversionUtils`](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs#L579) and a refactor was done to pass the `Stream` property and also to flush the writer before reading the contents of the output stream.

- I propose to drop the odd `internal` [`StringWriter` property](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs#L33) accessed from `ODataJsonLightWriter` class' [`StartTextWriter`](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs#L952), [`EndTextWriter`](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs#L986), [`StartTextWriterAsync`](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs#L2011), and [`EndTextWriterAsync`](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs#L2049) methods. Based on the existing logic, the field would only be accessed if the [`jsonWriter` class member](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs#L53) is null. Based on my investigation, all the 4 code blocks where it's accessed are dead code for the following reasons:

  - The [`jsonWriter` field is set from the constructor](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs#L123) and a debug assertion is thrown if [`ODataJsonLightOutputContext.JsonWriter` property returns null](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs#L205). The fact that the inconsistency has never been flagged also means there's no test coverage for those blocks.

  - In `ODataJsonLightOutputContext`, [an exception is thrown if the factory returns a null as `IJsonWriter`](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs#L1045).

  - In ODL 8, we merged the streaming APIs (that include [`StartTextWriter`](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs#L952), [`EndTextWriter`](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs#L986), [`StartTextWriterAsync`](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs#L2011), and [`EndTextWriterAsync`](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightWriter.cs#L2049) methods) into `IJsonWriter`. This means the methods need to be implemented and that provides an opportunity to supply a `StringWriter` as the `TextWriter` to be used it that's what is desired.

- In the `private` [`ODataJsonLightOutputContext.CreateJsonWriter` method](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs#L999), we check if the injected `IJsonWriterFactory` is `ODataUtf8JsonWriterFactory`. If it is and we are writing asynchronously, we don't wrap the message output stream with a `BufferedStream` since the `Utf8JsonWriter` implements it's own optimized buffering. Checking whether the injected `IJsonWriterFactory` is `ODataUtf8JsonWriterFactory` feels hack-ish!

- In the `private` [`ODataJsonLightOutputContext.CreateJsonWriter` method](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs#L999), we initialize the JSON writer to the concrete built-in `JsonWriter` [if the dependency injection container is null](https://github.com/OData/odata.net/blob/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs#L1006). Should we choose to make `ODataUtf8JsonWriter` to be the default writer, that bit of logic would _most probably_ need to change. Changing the default writer to `ODataUtf8JsonWriter` is tracked as a separate issue #2822.

- [x] *Test cases added*

### Additional work necessary


src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -85,43 +72,33 @@ internal sealed class ODataJsonLightOutputContext : ODataOutputContext
...
-                    if (this.Synchronous)
-                    {
-                        outputStream = this.messageOutputStream;
-                    }
-                    else
-                    {
-                        this.asynchronousOutputStream = new BufferedStream(this.messageOutputStream, this.MessageWriterSettings.BufferSize);
-                        outputStream = this.asynchronousOutputStream;
-                    }
+                    Stream outputStream = InitializeOutputStream();
```
habbes, line 99: Since both factories now accept a stream, I think the code would look cleaner if the `DefaultJsonWriterFactory` took encapsulated the buffered stream and wrapped it around the stream it received before passing it to the `JsonWriter`. That way you don't have don't have to explicitly check which type of `jsonWriterFactory` we have here. Ideally those details would be hidden within each `IJsonWriterFactory` implementation.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -939,29 +886,37 @@ protected override async ValueTask DisposeAsyncCore()
...
+                {
+                    throw new ODataException(Strings.ODataMessageWriter_JsonWriterFactory_ReturnedNull(isIeee754Compatible, encoding.WebName));
+                }
             }
 
-            if (jsonWriter is Json.JsonWriter defaultJsonWriter)
+            if (jsonWriter is JsonWriter defaultJsonWriter)
             {
                 defaultJsonWriter.ArrayPool = writerSettings.ArrayPool;
             }
```
habbes, line 922: nit: This block is a code smell in my opinion. Seems like the json writer factory abstraction isn't that clean if we have to resort to these checks. But don't know if there's a quick way to fix without investigating time in some design changes (which is likely outside the scope). But I do hope we get rid of this array pool one day.

Maybe the `ArrayPool` could be a property of the `DefaultJsonWriterFactory` instead, and it could pass it down to the `JsonWriter`. Feel free to ignore this comment if it distracts from the scope.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
```diff
@@ -970,29 +925,20 @@ protected override async ValueTask DisposeAsyncCore()
...
-            bool isIeee754Compatible,
-            Encoding encoding)
+        /// <param name="stream">The supplied message stream.</param>
+        /// <returns>The output stream.</returns>
+        private Stream InitializeOutputStream()
         {
-            IJsonWriter jsonWriter = factory.CreateJsonWriter(outputStream, isIeee754Compatible, encoding);
-
-            if (jsonWriter == null)
+            if (this.Synchronous)
```
habbes, line 934: I think the fact that we do something different based on whether we're in sync or async mode would further complicate my previous suggestion of encapsulating this buffer stream in the `DefaultJsonWriterFactory`. Feel free to #wontfix the suggestion if it distracts from the scope.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/MockODataUtf8JsonWriterFactory.cs
```diff
@@ -11,11 +11,11 @@
 
 namespace Microsoft.OData.Tests.Json
 {
-    internal sealed class MockStreamBasedJsonWriterFactory : IStreamBasedJsonWriterFactory
+    internal sealed class MockODataUtf8JsonWriterFactory : IJsonWriterFactory
```
habbes, line 14:
```suggestion
    internal sealed class MockJsonWriterFactory : IJsonWriterFactory

```

Since this class takes any `IJsonWriter` implementation, I don't think it makes sense to couple the name to `ODataUtf8JsonWriter`.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/MockODataUtf8JsonWriterFactoryWrapper.cs
```diff
@@ -13,15 +13,15 @@
...
 {
     /// <summary>
-    /// This wraps an <see cref="IStreamBasedJsonWriterFactory"/> to
-    /// intercept calls made to create a writer in order to allow
+    /// This wraps an <see cref="IJsonWriterFactory"/> to
+    /// intercept calls made to create an <see cref="ODataUtf8JsonWriter"/> in order to allow
     /// testing code paths around JSON writer construction.
     /// </summary>
-    internal sealed class MockStreamBasedJsonWriterFactoryWrapper : IStreamBasedJsonWriterFactory
+    internal sealed class MockODataUtf8JsonWriterFactoryWrapper : IJsonWriterFactory
```
habbes, line 20:
```suggestion
    internal sealed class MockJsonWriterFactoryWrapper : IJsonWriterFactory

```

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/MockODataUtf8JsonWriterFactoryWrapper.cs
```diff
@@ -13,15 +13,15 @@
 namespace Microsoft.OData.Tests.Json
 {
     /// <summary>
-    /// This wraps an <see cref="IStreamBasedJsonWriterFactory"/> to
-    /// intercept calls made to create a writer in order to allow
+    /// This wraps an <see cref="IJsonWriterFactory"/> to
+    /// intercept calls made to create an <see cref="ODataUtf8JsonWriter"/> in order to allow
```
habbes, line 17:
```suggestion
    /// intercept calls made to create an <see cref="IJsonWriter"/> in order to allow

```

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/MockODataUtf8JsonWriterFactoryWrapper.cs
```diff
@@ -13,15 +13,15 @@
...
     /// testing code paths around JSON writer construction.
     /// </summary>
-    internal sealed class MockStreamBasedJsonWriterFactoryWrapper : IStreamBasedJsonWriterFactory
+    internal sealed class MockODataUtf8JsonWriterFactoryWrapper : IJsonWriterFactory
     {
-        private readonly IStreamBasedJsonWriterFactory innerFactory;
+        private readonly IJsonWriterFactory innerFactory;
 
-        public MockStreamBasedJsonWriterFactoryWrapper(IStreamBasedJsonWriterFactory wrappedFactory)
+        public MockODataUtf8JsonWriterFactoryWrapper(IJsonWriterFactory wrappedFactory)
```
habbes, line 24:
```suggestion
        public MockJsonWriterFactoryWrapper(IJsonWriterFactory wrappedFactory)

```

test/FunctionalTests/Microsoft.OData.Core.Tests/JsonLight/ODataJsonLightBatchWriterTests.cs
```diff
@@ -505,7 +505,7 @@ public async Task WriteBatchResponseAsync_WithStreamCopy_UsingODataUtf8JsonWrite
         {
             Action<IContainerBuilder> configureServices = (builder) =>
             {
-                builder.AddService<IStreamBasedJsonWriterFactory>(ServiceLifetime.Singleton, _ =>DefaultStreamBasedJsonWriterFactory.Default);
+                builder.AddService<IJsonWriterFactory>(ServiceLifetime.Singleton, _ =>ODataUtf8JsonWriterFactory.Default);
```
habbes, line 508: nit:
```suggestion
                builder.AddService<IJsonWriterFactory>(ServiceLifetime.Singleton, _ => ODataUtf8JsonWriterFactory.Default);

```

xuzhg (dismissed):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
gathogojr: @habbes The challenge is as you [noted here](https://github.com/OData/odata.net/pull/2874#discussion_r1502062328). I didn't feel it'd be good design to introduce a parameter for passing the synchronous/asynchronous value. One thing I'd like to look into deeper into is whether we could use the buffered stream in both synchronous and asynchronous scenarios. But I wouldn't wish to expand the scope for this work item for now. I'll create a work item to track the follow up work.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
gathogojr: @habbes I created an issue to track the redesign for a cleaner approach https://github.com/OData/odata.net/issues/2881

src/Microsoft.OData.Core/JsonLight/ODataJsonLightOutputContext.cs
gathogojr: @habbes https://github.com/OData/odata.net/issues/2882

habbes (approved):

src/Microsoft.OData.Core/Json/DefaultJsonWriterFactory.cs
```diff
@@ -37,11 +38,13 @@ public DefaultJsonWriterFactory(ODataStringEscapeOption stringEscapeOption)
         /// <summary>
         /// Creates a new JSON writer of <see cref="IJsonWriter"/>.
         /// </summary>
-        /// <param name="textWriter">Writer to which text needs to be written.</param>
+        /// <param name="stream">The stream write to.</param>
```
ElizabethOkerio, line 41: nit: 
```suggestion
        /// <param name="stream">The stream to write to.</param>

```

src/Microsoft.OData.Core/Json/IJsonWriterFactory.cs
```diff
@@ -18,9 +19,10 @@ public interface IJsonWriterFactory
         /// <summary>
         /// Creates a new JSON writer of <see cref="IJsonWriter"/>.
         /// </summary>
-        /// <param name="textWriter">Writer to which text needs to be written.</param>
+        /// <param name="stream">The stream write to.</param>
```
ElizabethOkerio, line 22: why `the stream write to` and not `the stream to write to? `

src/Microsoft.OData.Core/Json/DefaultJsonWriterFactory.cs
```diff
@@ -37,11 +38,13 @@ public DefaultJsonWriterFactory(ODataStringEscapeOption stringEscapeOption)
         /// <summary>
         /// Creates a new JSON writer of <see cref="IJsonWriter"/>.
         /// </summary>
-        /// <param name="textWriter">Writer to which text needs to be written.</param>
+        /// <param name="stream">The stream write to.</param>
         /// <param name="isIeee754Compatible">True if it is IEEE754Compatible.</param>
         /// <returns>The JSON writer created.</returns>
-        public IJsonWriter CreateJsonWriter(TextWriter textWriter, bool isIeee754Compatible)
+        public IJsonWriter CreateJsonWriter(Stream stream, bool isIeee754Compatible, Encoding encoding)
```
wachugamaina, line 44: This seems like a breaking change. Why not have two constructors, one for steams and one for text and in the underlying implementation convert the text into a stream ?

src/Microsoft.OData.Core/Json/IJsonWriterFactory.cs
gathogojr: @ElizabethOkerio Fixed. Good catch! It should be "The stream to write to". The error was replicated because I tend to copy the param docstring across different places where the param represents the same thing.

src/Microsoft.OData.Core/Json/IJsonWriterFactory.cs
```diff
@@ -18,9 +19,10 @@ public interface IJsonWriterFactory
         /// <summary>
         /// Creates a new JSON writer of <see cref="IJsonWriter"/>.
         /// </summary>
-        /// <param name="textWriter">Writer to which text needs to be written.</param>
+        /// <param name="stream">The stream to write to.</param>
         /// <param name="isIeee754Compatible">True if it is IEEE754Compatible.</param>
+        /// <param name="encoding">The character encoding to use.</param>
         /// <returns>The JSON writer created.</returns>
-        IJsonWriter CreateJsonWriter(TextWriter textWriter, bool isIeee754Compatible);
+        IJsonWriter CreateJsonWriter(Stream stream, bool isIeee754Compatible, Encoding encoding);
```
wachugamaina, line 26: Why not use StreamWriter ?

src/Microsoft.OData.Core/Json/DefaultJsonWriterFactory.cs
gathogojr: @wachugamaina It is a breaking change. ODL 8.x will be a major release and it affords us the opportunity to roll out breaking changes that we can't safely make in ODL 7

src/Microsoft.OData.Core/Json/IJsonWriterFactory.cs
gathogojr: @wachugamaina Why `StreamWriter`?

gathogojr: @wachugamaina Passing the output stream allows the freedom for the `IJsonWriter` implementation to choose which means to use to write to the stream, e.g., write to the `Stream` directly by calling the `Write`/`WriteAsync` methods, use `StreamWriter` like you suggested, use `Utf8JsonWriter`, etc.

ElizabethOkerio (approved):

wachugamaina (approved):

habbes (approved):

PR #2883, opened 2024-03-04
Allow custom `HttpClient` via `IHttpClientFactory`

### LINKED ISSUE #2037: Provide a convenient way to specify a HttpClient instance

### LINKED ISSUE #2485: Allow injecting HttpClient instead of HttpClientHandler and reduce DataServiceClientRequest API surface.
This PR: https://github.com/OData/odata.net/pull/2431 provided support for injecting a custom `HttpClientHandler` which allowed us to address a number of long-standing issues, especially performance issues related to re-using http connections.

However, injecting `HttpClientHandler` is not the most ideal choice. I believe it would be better to inject `HttpClient` instead, namely because this allows customers to use `IHttpClientFactory` which is the recommended approach for working `HttpClient` in long-running services. If we only have provisions for `HttpClientHandler`, this means that customers will not benefit from `IHttpClientFactory` and will be required to manage the lifetime of `HttpClientHandler` and potential DNS changes on their own. If they still wish to customize a handler directly, they can still pass a custom handler to instantiate a custom `HttpClient`. So, this offers the most flexibility.

The reason we exposed `HttpClientHandler` is because `DataServiceClientRequest`, and consequently `HttpClientRequestMessgae` expose (writable) properties that can only be accessed from the handler. But since we allow injecting a custom client, we don't need to provide customization via properties, customers can configure the clients how they wish. Therefore, I propose we get rid of properties such as `Credentials`, `Timeout`, etc.

### Assemblies affected

Microsoft.OData.Core 7.x

### Reproduce steps

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

habbes:
### Issues

Fixes #2485 
Fixes #2037

### Description

This PR allows the developer to inject custom `HttpClient` instances via the standard `IHttpClientFactory` interfaces. This interface replaces the `IHttpClientHandlerProvider` used to provide custom `HttpClientHandler` instances. With this change, I also removed obsolete properties from `IODataRequestMessage` like `Credentials` and `ReadWriteTimeout`. If the developer wants to customize credentials, they should configure that on the `HttpClientHandler` they use to create a `HttpClient` and pass down an already-configured `HttpClient` via the `IHttpClientProvider` interface.

Now that we allow the user can provide custom `HttpClient` instances, it's possible (and very likely) that the same `HttpClient` instance will be used to send multiple request messages. For this reason, it's important the `HttpClientRequestMessage` properties are isolated to the single request they're wrapping and not impact other requests sent from the same client. In light of this, I made the following changes:
- Refactored `HttpClientRequestMessage.Abort()` to abort only the current requests instead of aborting all pending requests from the client. Before, it was calling `HttpClient.CancelPendingRequests`. Now, it uses a cancellation token to cancel only the single request in scope (`_abortRequestCancellationToken`).
- Refactored `HttpClientRequestMessage.Timeout` to set the timeout only for the single request instead of the entire `HttpClient`. Before, it was a proxy`HttpClient.Timeout` property, but not it's a separate field that schedule cancellation of the aforemention `_abortRequestCancellationToken` when set. When it's not set, the timeout from the `HttpClient` will still take effect.

I also changed the `HttpClientRequestMessage.Url` setter from throw an `OperationNotSupportedException` to actually updating the URL of the underlying `HttpRequestMessage`. I'm not sure why this was not done before.

I was contemplating whether to change the type of the `Timeout` property from `int` to `TimeSpan` for better consistency with `HttpClient.Timeout` property. I did not make this change because I didn't want to introduce extra disruptions. But do let me know what you think about that.

- [x] *Test cases added*

### Additional work necessary


test/FunctionalTests/Tests/DataServices/UnitTests/ClientCSharpUnitTests/TestHttpClientProvider.cs
```diff
@@ -0,0 +1,27 @@
...
+namespace AstoriaTestFramework.Client.CLWrappers
+{
+    internal class TestHttpClientProvider : IHttpClientProvider
+    {
+        readonly HttpClient _client;
+        public TestHttpClientProvider(TestHttpClientProviderOptions options)
+        {
+            HttpClientHandler handler = new HttpClientHandler();
+            handler.Credentials = options.Credentials;
+            _client = new HttpClient(handler);
```
wachugamaina, line 14: Why are we newing this up, why not inject it or use some factory?

test/FunctionalTests/Tests/DataServices/ddbasics/TestHttpClientProvider.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Microsoft.OData.Client;
```
wachugamaina, line 1: Is this not the same class as
https://github.com/OData/odata.net/pull/2883/files#diff-f4748932855371bed900c751c5215fdb57f98380b5cff81c55af54288b58d4c5
Why not re-use it ?

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -40,17 +41,18 @@ public class HttpClientRequestMessage : DataServiceClientRequestMessage, ISendin
...
-
         /// <summary> The effective HTTP method. </summary>
         private readonly string _effectiveHttpMethod;
 
         /// <summary>HttpRequestMessage instance.</summary>
         private readonly HttpRequestMessage _requestMessage;
         private readonly HttpClient _client;
-        private readonly HttpClientHandler _handler;
         private readonly MemoryStream _messageStream;
+        private readonly bool _shouldDisposeClient;
```
wachugamaina, line 51: In what instance would we use this, I think with HttpClient we want to always keep it around for the life of the application.

src/Microsoft.OData.Client/IHttpClientProvider.cs
```diff
@@ -9,21 +9,19 @@
...
-    public interface IHttpClientHandlerProvider
+    public interface IHttpClientProvider
     {
         /// <summary>
-        /// Returns an <see cref="HttpClientHandler"/> instance that OData Client should
+        /// Returns an <see cref="HttpClient"/> instance that OData Client should
         /// used to execute a request. This method is called per request.
         /// </summary>
-        /// <returns>The <see cref="HttpClientHandler"/> instance that will be used to execute the request.</returns>
+        /// <returns>The <see cref="HttpClient"/> instance that will be used to execute the request.</returns>
```
wachugamaina, line 21: Why not use IHttpClientFactory ?

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -77,18 +79,23 @@ public HttpClientRequestMessage(DataServiceClientRequestMessageArgs args)
...
-            IHttpClientHandlerProvider clientHandlerProvider = args.HttpClientHandlerProvider;
-            if (clientHandlerProvider == null)
+            _abortRequestCancellationTokenSource = new CancellationTokenSource();
+
+            IHttpClientProvider clientProvider = args.HttpClientProvider;
+            if (clientProvider == null)
             {
-                _handler = new HttpClientHandler();
-                _client = new HttpClient(_handler, disposeHandler: true);
+                _client = new HttpClient();
```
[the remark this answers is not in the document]
gathogojr: @habbes In addition to @wachugamaina's question, please confirm whether we're supporting injecting of `HttpClient` or `HttpClientFactory` or `IHttpClientProvider` via DI in a separate work item. Actually, do we support DI in OData client?

src/Microsoft.OData.Client/IHttpClientProvider.cs
habbes: Good question. One concern with `IHttpClientFactory` is .NET Framework support which we have not yet finalized a decision on. Another concern is whether requesting `IHttpClientFactory` would be a barrier to people who are using OData client in an environment where they don't have an instance of `IHttpClientFactory`.

src/Microsoft.OData.Client/IHttpClientProvider.cs
```diff
@@ -9,21 +9,19 @@
...
         /// <remarks>
-        /// It's possible that the returned <see cref="HttpClientHandler"/> may be mutated inside a request hook
-        /// e.g. via <see cref="HttpClientRequestMessage.Credentials"/>.
-        /// OData Client does not dispose the <see cref="HttpClientHandler"/> instance.
+        /// OData Client does not dispose the <see cref="HttpClient"/> instance.
         /// </remarks>
-        HttpClientHandler GetHttpClientHandler();
+        HttpClient GetHttpClient();
```
habbes, line 25: Should this method accept arguments? Do you think there's a use case for creating different `HttpClient` instances based on some request context information, like the request URI?

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockHttpClientProvider.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Microsoft.OData.Client.Tests.Serialization
+{
+    /// <summary>
+    /// A mock implementation of <see cref="IHttpClientProvider"/>
+    /// for testing purposes.
+    /// </summary>
+    internal sealed class MockHttpClientProvider
+
+        : IHttpClientProvider
```
gathogojr, line 19:
```suggestion
    internal sealed class MockHttpClientProvider
        : IHttpClientProvider
```

test/FunctionalTests/Tests/DataServices/UnitTests/ClientExtensions/DataServiceContextWithCustomTransportLayer.cs
```diff
@@ -65,7 +65,6 @@ public DataServiceContextWithCustomTransportLayer(Uri serviceUri, ODataProtocolV
         internal DataServiceClientRequestMessage CreateMessage(DataServiceClientRequestMessageArgs requestMessageArgs)
         {
             var requestMessage = this.getRequestMessage();
-            requestMessage.Url = requestMessageArgs.RequestUri;
```
gathogojr, line 68: We don't need to do this anymore?

test/FunctionalTests/Tests/DataServices/ddbasics/TestHttpClientProvider.cs
gathogojr: @habbes Is there a "TestCommon" project that you can move the test classes that you have redefined in different test projects?

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -39,18 +40,23 @@ public class HttpClientRequestMessage : DataServiceClientRequestMessage, ISendin
             "Expires",
             "Last-Modified",
         }, StringComparer.OrdinalIgnoreCase);
-
-        /// <summary>Request Url.</summary>
-        private readonly Uri _requestUrl;
+        /// <summary>
+        /// HttpClient to use when the caller does not provide one.
+        /// </summary>
+        private static readonly HttpClient DefaultClient = new HttpClient();
```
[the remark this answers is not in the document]
habbes: Renamed to `_defaultClient`.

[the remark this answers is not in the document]
habbes: Given that now the default client is static, where else would you suggest initializing it from?

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
habbes: Good call out. I no longer need this now that the default client is static.

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockDelayedHttpClientHandler.cs
```diff
@@ -0,0 +1,82 @@
...
+                if (stopwatch.ElapsedMilliseconds > _delay)
+                {
+                    break;
+                }
+            }
+
+            stopwatch.Stop();
+            return base.Send(request, cancellationToken);
+        }
+#endif
```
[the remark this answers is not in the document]
habbes: The difference is that in `NETCOREAPP` we use `protected override` otherwise we use `public override`. Note that this class inherits from the test `MockHttpClientHandler` which in turns inherits from the framework's `HttpClientHandler`. In .NET, `HttpClientHandler` defines a `protected virtual` `Send` method, but that's not the case in .NET FX. So in NET Fx, the `MockHttpClientHandler` implements a `public virtual Send` overload, which is what is being overridden here in the `#else` block.

I could simplify this code by having the `#if-#else` block only on the method signature.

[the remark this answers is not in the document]
gathogojr: > I could simplify this code by having the #if-#else block only on the method signature.

@habbes This makes more sense. We have done it that way in other places

test/FunctionalTests/Tests/DataServices/ddbasics/TestHttpClientProvider.cs
habbes: The `TestCommon` project doesn't have references to OData core libraries and I'm not too fond of changing the dependencies in this pipeline solution until we port/fix it for ODL 8. Otherwise I'd have moved it there.

Also, I don't think it's a bad thing to have this class duplicated. It's small and doesn't do much. And if it changed in different ways in each of the 3 projects, that would be fine since each project might have unique scenarios it tests for.

test/FunctionalTests/Tests/DataServices/UnitTests/ClientExtensions/DataServiceContextWithCustomTransportLayer.cs
habbes: Hmm. Not sure why I removed this. I've put it back.

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -39,18 +40,23 @@ public class HttpClientRequestMessage : DataServiceClientRequestMessage, ISendin
...
         /// <summary>HttpRequestMessage instance.</summary>
         private readonly HttpRequestMessage _requestMessage;
         private readonly HttpClient _client;
-        private readonly HttpClientHandler _handler;
         private readonly MemoryStream _messageStream;
+        private readonly bool _shouldDisposeClient;
+        private CancellationTokenSource _abortRequestCancellationTokenSource;
+        private TimeSpan _timeout;
+        // Whether the _timeout value has been changed from the default or not
+        bool _isCustomTimeout = false;
```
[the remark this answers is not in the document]
habbes: I feel like `defaultTimeoutChanged` can give the impression that some changed the value of the default timeout rather someone is using a timeout other than the default value.

src/Microsoft.OData.Client/IHttpClientProvider.cs
habbes: I may have been wrong about the .NET Framework support. `IHttpClientFactory` is defined in .NET Platform Extensions and not in the core standard lib. And according to this chart on NuGet, it seems to be supported in .NET FX: https://www.nuget.org/packages/Microsoft.Extensions.Http#supportedframeworks-body-tab

habbes:
@gathogojr @wachugamaina I replaced `IHttpClientProvider` with the standard `IHttpClientFactory` interface after confirming it's support in .NET FX.

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockDelayedHttpClientHandler.cs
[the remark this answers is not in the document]
habbes: I made that change.

src/Microsoft.OData.Client/DataServiceClientRequestMessageArgs.cs
```diff
@@ -37,11 +36,10 @@ public DataServiceClientRequestMessageArgs(string method, Uri requestUri, bool u
         /// </summary>
         /// <param name="method">Method of the request.</param>
         /// <param name="requestUri">The Request Uri.</param>
-        /// <param name="useDefaultCredentials">True if the default credentials need to be sent with the request. Otherwise false.</param>
         /// <param name="usePostTunneling">True if the request message must use POST verb for the request and pass the actual verb in X-HTTP-Method header, otherwise false.</param>
         /// <param name="headers">The set of headers for the request.</param>
-        /// <param name="httpClientHandlerProvider">The <see cref="IHttpClientHandlerProvider"/> that provides the <see cref="HttpClientHandler"/> that should be used to send the request message.</param>
-        public DataServiceClientRequestMessageArgs(string method, Uri requestUri, bool useDefaultCredentials, bool usePostTunneling, IDictionary<string, string> headers, IHttpClientHandlerProvider httpClientHandlerProvider)
+        /// <param name="httpClientFactory">The <see cref="IHttpClientProvider"/> that provides the <see cref="HttpClient"/> that should be used to send the request message.</param>
```
gathogojr, line 41:
```suggestion
        /// <param name="httpClientFactory">The <see cref="IHttpClientFactory"/> that provides the <see cref="HttpClient"/> that should be used to send the request message.</param>
```

src/Microsoft.OData.Client/Microsoft.OData.Client.csproj
```diff
@@ -160,6 +160,7 @@
       <PrivateAssets>all</PrivateAssets>
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
     </PackageReference>
+    <PackageReference Include="Microsoft.Extensions.Http" Version="8.0.0" />
```
gathogojr, line 163: I'm assuming this reference is needed for `IHttpClientFactory`...

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -14,9 +14,10 @@ namespace Microsoft.OData.Client
     using System.Net;
     using System.Net.Http;
     using System.Net.Http.Headers;
+    using System.Threading;
     using System.Threading.Tasks;
     using Microsoft.OData;
-
+
```
gathogojr, line 20:
```suggestion

```

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -39,18 +40,22 @@ public class HttpClientRequestMessage : DataServiceClientRequestMessage, ISendin
             "Expires",
             "Last-Modified",
         }, StringComparer.OrdinalIgnoreCase);
-
-        /// <summary>Request Url.</summary>
-        private readonly Uri _requestUrl;
+        /// <summary>
+        /// HttpClient to use when the caller does not provide one.
+        /// </summary>
+        private static readonly HttpClient _defaultClient = new HttpClient();
```
gathogojr, line 46:
```suggestion

        /// <summary>
        /// HttpClient to use when the caller does not provide one.
        /// </summary>
        private static readonly HttpClient _defaultClient = new HttpClient();
```

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -541,7 +513,7 @@ protected virtual void Dispose(bool disposing)
                     ((IDisposable)response).Dispose();
                 }
 
-                _client.Dispose();
+                _abortRequestCancellationTokenSource.Dispose();
```
gathogojr, line 516: We don't need to dispose the HttpClient is we didn't use the HttpClientFactory to initialize it?

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
```diff
@@ -156,18 +153,81 @@ public async void Abort_CancelsPendingRequest()
...
-        public async void Timeout_CancelsPendingRequestAfterTimeout()
+        public async Task Abort_DoesNotCancelOtherRequestsFromTheSameClient()
         {
             // Arrange
-            using (var handler = new MockUnresponsiveHttpClientHandler())
+            using (var handler = new MockDelayedHttpClientHandler("Success", delayMilliseconds: 2000))
             {
 
-                var httpClientProvider = new MockHttpClientHandlerProvider(handler);
+                var httpClientProvider = new MockHttpClientFactory(handler);
```
gathogojr, line 162:
```suggestion
                var httpClientFactory = new MockHttpClientFactory(handler);
```

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
```diff
@@ -197,35 +257,100 @@ public async void Timeout_CancelsPendingRequestAfterTimeout()
...
 
         [Fact]
-        public void WhenClientHandlerIsProvided_CredentialsProperty_AccessHandlerCredentials()
+        public async Task Timeout_DoesNotCancelOtherRequestsFromTheSameClient()
         {
             // Arrange
-            using (var handler = new MockHttpClientHandler("Foo"))
+            using (var handler = new MockDelayedHttpClientHandler("Success", 5000))
             {
+                var httpClientProvider = new MockHttpClientFactory(handler);
```
gathogojr, line 265:
```suggestion
                var httpClientFactory = new MockHttpClientFactory(handler);
```

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/HttpClientRequestMessageTests.cs
```diff
@@ -197,35 +257,100 @@ public async void Timeout_CancelsPendingRequestAfterTimeout()
...
+            }
+        }
+
+        [Fact]
+        public async Task WhenTimeoutNotSet_HttpClientTimeoutStillApplies()
+        {
+            // Arrange
+            using (var handler = new MockDelayedHttpClientHandler("success", delayMilliseconds: 5000))
+            {
+                var httpClientProvider = new MockHttpClientFactory(handler, new MockHttpClientFactoryOptions
```
gathogojr, line 322:
```suggestion
                var httpClientFactory = new MockHttpClientFactory(handler, new MockHttpClientFactoryOptions
```

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
ElizabethOkerio: Shouldn't the factory be used to create the client here?

ElizabethOkerio: Or why are we creating a HttpClient beforehand instead of doing it when factory is not provided? Why can't OData CLient use the IHttpClientFactory to create httpclients instead of newing them directly?

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -356,7 +321,7 @@ public override Stream GetStream()
         /// </summary>
         public override void Abort()
         {
-            _client.CancelPendingRequests();
+            _abortRequestCancellationTokenSource.Cancel();
```
ElizabethOkerio, line 324: Did you add tests to verify this?

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
```diff
@@ -447,7 +412,14 @@ private Task<HttpResponseMessage> CreateSendTask()
             }
 
             _requestMessage.Method = new HttpMethod(_effectiveHttpMethod);
-            return _client.SendAsync(_requestMessage);
+            // If the timeout value is still the default, don't schedule cancellation.
+            // The timeout from the HttpClient will take effect.
+            if (_isTimeoutProvidedByCaller)
+            {
```
ElizabethOkerio, line 418: Add tests tp verify this.

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockHttpClientHandler.cs
```diff
@@ -17,7 +17,7 @@ namespace Microsoft.OData.Client.Tests.Serialization
     /// A mock implementation of <see cref="HttpClientHandler"/>
     /// for testing purposes.
     /// </summary>
-    internal sealed class MockHttpClientHandler : HttpClientHandler
+    internal class MockHttpClientHandler : HttpClientHandler
```
ElizabethOkerio, line 20: Is this class used anywhere since we did away with the `HttpClientHandler`?

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockDelayedHttpClientHandler.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Net.Http;
+using System.Threading;
+
+namespace Microsoft.OData.Client.Tests.Serialization
+{
+    /// <summary>
+    /// A mock HttpClientHandler that responds to HTTP requests with a mocked
+    /// response after a specified time delay.
+    /// </summary>
+    internal sealed class MockDelayedHttpClientHandler : MockHttpClientHandler
```
ElizabethOkerio, line 18: I'm not sure I understand the use of this class or it's the naming that's confusing.

test/FunctionalTests/Tests/DataServices/UnitTests/Client.TDD.Tests/DataServiceContextUtil.cs
```diff
@@ -29,7 +29,6 @@ internal static HttpClientRequestMessage InjectFakeEdmxRequest()
                     "GET",
                     new Uri("http://temp.org/"),
                     false,
-                    false,
```
ElizabethOkerio, line 32: why this change?

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockDelayedHttpClientHandler.cs
habbes: It's a HttpClientHandler implementation that responds to requests after a configured time delay. E.g. if the time delay is 3s, then it will respond to requests with a mock response after 3s.

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockHttpClientHandler.cs
habbes: Yes, it's used in tests, specifically it's used to created mock `HttpClientHandler` implementations that are passed to the `MockHttpClientFactory` to test things like timeouts, request cancellation, response handling, etc.

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
habbes: This is the "default" client that will be used just in case the user does not provide an `IHttpClientFactory`. If they provide one, we don't use this. If they don't provide one, then we don't have an `IHttpClientFactory` to use, and so we default to this one.

Note that we're not creating the `HttpClient` "beforehand". This is a static field initialization, the runtime will only create the client if the `_defaultClient` field is accessed. This means that if the application always provides an `IHttpCientFactory` then the default `HttpClient` will never be allocated.
 
See how it's used here: https://github.com/OData/odata.net/pull/2883/files#diff-7ffa90011377965c2da0275b44b7914818cf3aacaaaabdd766231a449f3ad656R88

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
habbes: Yes I did, see: https://github.com/OData/odata.net/pull/2883/files#diff-b4128975c89acf84c4f440fc4c0b2af782a08e0fb8955d0195ed344bab8562a8R115

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
habbes: I did. See: https://github.com/OData/odata.net/pull/2883/files#diff-b4128975c89acf84c4f440fc4c0b2af782a08e0fb8955d0195ed344bab8562a8R317

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
habbes: Since we made the default `HttpClient` static, we should not dispose it since it will be re-used by different requests, and its lifetime (if it ever gets created) will be the process's lifetime.

test/FunctionalTests/Microsoft.OData.Client.Tests/Serialization/MockDelayedHttpClientHandler.cs
habbes: See sample usage: https://github.com/OData/odata.net/pull/2883/files#diff-b4128975c89acf84c4f440fc4c0b2af782a08e0fb8955d0195ed344bab8562a8R159

src/Microsoft.OData.Client/Microsoft.OData.Client.csproj
habbes: Yes is.

src/Microsoft.OData.Client/Serialization/HttpClientRequestMessage.cs
gathogojr: From the perspective of isolating requests, would it not be better to new up `HttpClient` instead of using a static instance?
Or at least have PooledConnectionLifetime set - https://learn.microsoft.com/en-us/dotnet/fundamentals/networking/http/httpclient-guidelines#recommended-use

habbes: @gathogojr that was my initial implementation. But you and @wachugamaina  suggested that I should make the default client static instead. I think having a static client is a better default than a new instance per request. Creating a new client each time, means a new connection each time, this results in higher latencies and more resource utilization. Using the `PooledConnectionLifetime` seems like a good compromise, but how do we decide what the appropriate value is for most users? If we pick a default connection lifetime, then I think it's not any better than picking none and letting the user decide what it should be, but that would be an extra configuration that we'd be exposing to the user. If customers want better control of HttpClient lifetimes, they should provide their `IHttpClientFactory` in my opinion.

As for isolating requests, I'm not sure I understand the scenario. I don't see how sharing the client would cause conflicts between requests. Note that we do allow some request-specific settings like timeouts and cancellation which can be isolated despite sharing the client.

gathogojr (approved):

ElizabethOkerio (approved):

Issue #2884, opened 2024-03-04
OData reader fails to read payloads having lowercase EDM types


budt12:
*Short summary (3-5 sentences) describing the issue.*
Running into issues when odata would not parse payloads having lowercase edm types and throws up ODataException

### Assemblies affected

*Which assemblies and versions are known to be affected e.g. OData .Net lib 7.x*
Microsoft.OData.Edm, all versions.

### Reproduce steps
Any payload with lowercase types like 'string', 'int32', 'binary' etc would break payload reader, ODataMessageReader while reading the payload. 

Here's the exception stack for following example payload - 
```
Callsite Method Name: [ReadEntryAsync], Callsite line number: [73]; Exception=Microsoft.OData.ODataException: Incompatible type kinds were found. The type 'string' was found to be of kind 'Complex' instead of the expected kind 'Primitive'.
  at Microsoft.OData.ReaderValidationUtils.ResolvePayloadTypeNameAndComputeTargetType(EdmTypeKind expectedTypeKind, Nullable`1 expectStructuredType, IEdmType defaultPrimitivePayloadType, IEdmTypeReference expectedTypeReference, String payloadTypeName, IEdmModel model, Func`3 clientCustomTypeResolver, Boolean throwIfTypeConflictsWithMetadata, Boolean enablePrimitiveTypeConversion, Func`1 typeKindFromPayloadFunc, EdmTypeKind& targetTypeKind, ODataTypeAnnotation& typeAnnotation)
  at Microsoft.OData.JsonLight.ODataJsonLightPropertyAndValueDeserializer.ReadNonEntityValueImplementationAsync(String payloadTypeName, IEdmTypeReference expectedTypeReference, PropertyAndAnnotationCollector propertyAndAnnotationCollector, CollectionWithoutExpectedTypeValidator collectionValidator, Boolean validateNullValue, Boolean isTopLevelPropertyValue, Boolean insideResourceValue, String propertyName, Nullable`1 isDynamicProperty)
  at Microsoft.OData.JsonLight.ODataJsonLightPropertyAndValueDeserializer.ReadNonEntityValueAsync(String payloadTypeName, IEdmTypeReference expectedValueTypeReference, PropertyAndAnnotationCollector propertyAndAnnotationCollector, CollectionWithoutExpectedTypeValidator collectionValidator, Boolean validateNullValue, Boolean isTopLevelPropertyValue, Boolean insideResourceValue, String propertyName, Nullable`1 isDynamicProperty)
[...]

PR #2885, opened 2024-03-04
Adding ignore case to Coreschema types dictionary in edm core model

budt12:
### Issues

*This pull request fixes issue with [OData reader fails to read payloads having lowercase EDM types](https://github.com/OData/odata.net/issues/2884).*

### Description

Adding ignore case to coreSchemaTypes dictionary declaration in EdmCoreModel.cs so that correct types with lowercase are mapped as well.

- [ X ] *Test cases added*
- [ X ] *Build and test with one-click build and test script passed*

### Additional work necessary

Not needed.


budt12:

--
Tanmay

wachugamaina (approved):

src/Microsoft.OData.Edm/Schema/CoreModel/EdmCoreModel.cs
```diff
@@ -31,7 +31,7 @@ public class EdmCoreModel : EdmElement, IEdmModel, IEdmCoreModelElement
 
         private readonly IList<IEdmSchemaElement> coreSchemaElements = new List<IEdmSchemaElement>();
 
-        private readonly IDictionary<string, IEdmSchemaType> coreSchemaTypes = new Dictionary<string, IEdmSchemaType>();
+        private readonly IDictionary<string, IEdmSchemaType> coreSchemaTypes = new Dictionary<string, IEdmSchemaType>(StringComparer.OrdinalIgnoreCase);
```
corranrogue9, line 34: Can we please parameterize this in a new `EdmCoreModel` public constructor? That way the `Instance` property remains the same and a caller could configure this if desired?

budt12: Parameterizing and adding a public constructor here would require the new constructor to be explicitly called to initialize the dictionary with a string comparer.

I'm not quite sure if bubbling up that EdmCoreModel initialization all the way up to the consumer is necessary here.

To you point about benchmarking this change, this is quite a small dictionary with a very low number of keys. Not quite sure if this will have that big of an impact on performance.

habbes (approved):

gathogojr (approved):

src/Microsoft.OData.Edm/Schema/CoreModel/EdmCoreModel.cs
xuzhg, line 34: I do think it could be better to have a configuration to enable/disable the case sensitive, otherwise it could be a breaking change.

habbes: @xuzhg that's a good point. Where would you suggest adding such configuration given that this is in the `EdmCoreModel`?

habbes: I don't think parameterizing the constructor here is practical since `EdmCoreModel` is used as a singleton, which makes sense because it doesn't change. If we were to paramterize the constructor, we'd need to ensure the configures the `EdmCoreModel` once and configure the library to use that coremodel throughout, in all call sites. This would definitely be huge refactor and may have unintended consequences.

Alternatively, we could create definite a separate `caseInsensitiveSchemaTypes` dictionary that is created only once from a copying the data in the `coreSchemaTypes`. Then we could add an optional `caseSensitive` parameter to the `FindDeclaredType` method, when this flag is set to false, then it would make use of the `caseInsensitiveSchemaTypes` or case insensitive `EdmCoreModel` instead. But this approach means that we would have to bubble this parameter to all the methods that call `FindDeclaredType` all the way up to the public APIs. And if `FindDeclaredType` is called behind an interface instead of the `EdmCoreModel.Instance` directly (not sure if there's such a case), we would have to either add a new overload to the interface (which could be a breaking change unless we use default interface implementations) or we would have to perform casts to check whether the interface is `EdmCoreModel`.

habbes: Instead of bubbling this up everywhere, we could constrain it to just this use case of validating types during reads. We could have flag in the reader validation settings to enable case-insensitive type name validation. But the validator calls the `IEdmModel.FindType` extension method which looks up types in the model and its referenced models. So to pass this config down we may need to add overloads to the `IEdmModel.FindDeclaredType` interface method or explicitly check for `EdmCoreModel`. I don't find this path to be that clean.

habbes: The `ODataUriResolver` can be configured to do case-insensitive look ups. When case-insensitive mode is enabled, it first tries to find a type using `IEdmModel.FindType`, it falls back to using doing a case-insensitive search. If the model is marked as immutable it performs the case-insensitive lookup using a cache that performs a one-time copy of the model's types and schema elements into case-insensitive dictionaries. If the model is not marked as immutable it performs a case-insensitive linear search across the model elements. The linear search would definitely be prohibitively expensive for this use case.

With this in mind, another approach/compromise we could consider is to do the same thing for type validation during reading. If the case-insensitive validation setting is enabled, the we fall back to using the cache (if the model is immutable). This would probably be more costly than the approach implemented in this PR, but it would prevent people who don't need case-insensitivity from being affected. It would also require fewer code changes than propagating the setting down to the `EdmCoreLib`, we could localize the changes just to the reader validator and the helper methods it calls.

habbes (changes requested):
Blocking this PR pending changes to make it an opt-in feature. Blocking to prevent an accidental merge since it had passed all required checks.

src/Microsoft.OData.Edm/Schema/CoreModel/EdmCoreModel.cs
xuzhg: We can define an annotation (for configuration) for customer to config the case-insensitive or case-sensitive. 
If customer wants to support case-insensitive, create an annotation and attach it to the Edm Model.
OData.Edm library should monitor the EdmModel to look up such annotation, if attached and configured, we can pass the configuration to EdmCoreLib's FindType method?

ElizabethOkerio:
Work around provided

PR #2887, opened 2024-03-06
Add EntityType property to IEdmNavigationSource

### LINKED ISSUE #2820: Consider adding an `EntityType` property to `IEdmNavigationSource`
This issue mentions the perf overhead of the `IEdmNavigationSource.EntityType()` extension method. I made slight improvements to that method but I think it would be easier if `IEdmNavigationSource` had an `EntityType` property that could be easily accessed without casting. Each implementation of `IEdmNavigationSource` could store the entity type in the property so that it can be accessed without having to recompute it every time (for example calling `AsElementType()` for entity sets).

Generally, I think we should revisit other EDM interfaces where we are doing recurring computations and casting that could be simplified and made more performant with simpler/unified interfaces with the right properties, e.g. see: https://github.com/OData/odata.net/issues/2528

gathogojr:
### Issues

*This pull request fixes #2820.*

### Description

This pull request adds a `EntityType` property to `IEdmNavigationSource` interface to eliminate the constant type casting that happens when `EntityType(IEdmNavigationSource)` extension method is called.

The extension method is invoked very frequently and on hot paths in some cases.

Changes included implementing the property in classes that implement the `IEdmNavigationSource` interface.
In the crucial `EdmNavigationSource` class, the implementation looks as follows:
```csharp
// ...

private IEdmEntityType entityType;
private bool entityTypeSet = false;

// ...

public IEdmEntityType EntityType
{
    get
    {
        if (!this.entityTypeSet)
        {
            this.entityType = this.Type?.AsElementType() as IEdmEntityType;
            this.entityTypeSet = true;
        }

        return this.entityType;
    }
}
```

Since the `EntityType` extension method is public, I retained it but applied the `Obsolete` attribute.

Refactoring was also done to replace calls to the `EntityType` extension method with `IEdmNavigationSource.EntityType` property access.

**NOTE:**
The build pipeline for 8.x is not running the E2E tests currently.
A previous 8.x pull request got rid of `CreateODataDeltaReader` since it's deprecated and inadvertently forgot to update the tests in _/test/EndToEndTests/Tests/Client/Build.Desktop/DeltaTests/DeltaTests.cs_.
The E2E solution couldn't build locally so I fixed the delta tests in this pull request.
In addition, the tests had very few assert statements with little to no verification of the expected payload structure and data so I refactored them extensively.

- [x] *Test cases added*

### DIFF
--- a/src/Microsoft.OData.Core/UnknownEntitySet.cs
+++ b/src/Microsoft.OData.Core/UnknownEntitySet.cs
@@ -14,6 +14,7 @@ namespace Microsoft.OData
         private readonly IEdmNavigationProperty navigationProperty;
         private readonly IEdmNavigationSource parentNavigationSource;
         private IEdmPathExpression path;
+        private IEdmEntityType entityType;
 
         public UnknownEntitySet(IEdmNavigationSource parentNavigationSource, IEdmNavigationProperty navigationProperty)
         {
@@ -41,6 +42,20 @@ namespace Microsoft.OData
             get
             {
                 return this.navigationProperty.Type.Definition;
+            }
+        }
+
+        /// <inheritdoc/>
+        public IEdmEntityType EntityType
+        {
+            get
+            {
+                if (this.entityType == null)
+                {
+                    this.entityType = this.Type.AsElementType() as IEdmEntityType;
+                }
+
+                return this.entityType;
             }
         }
 
--- a/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsNavigationSource.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsNavigationSource.cs
@@ -45,6 +45,9 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
         private readonly ConcurrentDictionary<IEdmNavigationProperty, IEnumerable<IEdmNavigationPropertyBinding>> navigationPropertyBindingCache = 
             new ConcurrentDictionary<IEdmNavigationProperty, IEnumerable<IEdmNavigationPropertyBinding>>();
 
+        private IEdmEntityType entityType;
+        private bool entityTypeSet = false;
+
         public CsdlSemanticsNavigationSource(CsdlSemanticsEntityContainer container, CsdlAbstractNavigationSource navigationSource)
             : base(navigationSource)
         {
@@ -85,6 +88,21 @@ namespace Microsoft.OData.Edm.Csdl.CsdlSemantics
         public IEnumerable<IEdmNavigationPropertyBinding> NavigationPropertyBindings
         {
             get { return this.navigationTargetsCache.GetValue(this, ComputeNavigationTargetsFunc, null); }
+        }
+
+        /// <inheritdoc/>
+        public IEdmEntityType EntityType
+        {
+            get
+            {
+                if (!this.entityTypeSet)
+                {
+                    this.entityType = this.Type.AsElementType() as IEdmEntityType;
+                    this.entityTypeSet = true;
+                }
+
+                return this.entityType;
+            }
         }
 
         public IEdmNavigationSource FindNavigationTarget(IEdmNavigationProperty property, IEdmPathExpression bindingPath)
--- a/src/Microsoft.OData.Edm/Schema/AmbiguousEntitySetBinding.cs
+++ b/src/Microsoft.OData.Edm/Schema/AmbiguousEntitySetBinding.cs
@@ -12,6 +12,8 @@ namespace Microsoft.OData.Edm
 {
     internal class AmbiguousEntitySetBinding : AmbiguousBinding<IEdmEntitySet>, IEdmEntitySet
     {
+        private IEdmEntityType entityType;
+
         public AmbiguousEntitySetBinding(IEdmEntitySet first, IEdmEntitySet second)
             : base(first, second)
         {
@@ -39,6 +41,20 @@ namespace Microsoft.OData.Edm
         public IEdmType Type
         {
             get { return new EdmCollectionType(new EdmEntityTypeReference(new BadEntityType(String.Empty, this.Errors), false)); }
+        }
+
+        /// <inheritdoc/>
+        public IEdmEntityType EntityType
+        {
+            get
+            {
+                if (this.entityType == null)
+                {
+                    this.entityType = this.Type.AsElementType() as IEdmEntityType;
+                }
+
+                return this.entityType;
+            }
         }
 
         public bool IncludeInServiceDocument
--- a/src/Microsoft.OData.Edm/Schema/AmbiguousSingletonBinding.cs
+++ b/src/Microsoft.OData.Edm/Schema/AmbiguousSingletonBinding.cs
@@ -12,6 +12,8 @@ namespace Microsoft.OData.Edm
 {
     internal class AmbiguousSingletonBinding : AmbiguousBinding<IEdmSingleton>, IEdmSingleton
     {
+        private IEdmEntityType entityType;
+
         public AmbiguousSingletonBinding(IEdmSingleton first, IEdmSingleton second)
             : base(first, second)
         {
@@ -22,6 +24,19 @@ namespace Microsoft.OData.Edm
             get { return new BadEntityType(String.Empty, this.Errors); }
         }
 
+        /// <inheritdoc/>
+        public IEdmEntityType EntityType
+        {
+            get
+            {
+                if (this.entityType == null)
+                {
+                    this.entityType = this.Type.AsElementType() as IEdmEntityType;
+                }
+
+                return this.entityType;
+            }
+        }
 
         public EdmContainerElementKind ContainerElementKind
         {
--- a/src/Microsoft.OData.Edm/Schema/BadEntitySet.cs
+++ b/src/Microsoft.OData.Edm/Schema/BadEntitySet.cs
@@ -18,6 +18,7 @@ namespace Microsoft.OData.Edm
     {
         private readonly string name;
         private readonly IEdmEntityContainer container;
+        private IEdmEntityType entityType;
 
         public BadEntitySet(string name, IEdmEntityContainer container, IEnumerable<EdmError> errors)
             : base(errors)
@@ -54,6 +55,20 @@ namespace Microsoft.OData.Edm
         public IEdmType Type
         {
             get { return new EdmCollectionType(new EdmEntityTypeReference(new BadEntityType(String.Empty, this.Errors), false)); }
+        }
+
+        /// <inheritdoc/>
+        public IEdmEntityType EntityType
+        {
+            get
+            {
+                if (this.entityType == null)
+                {
+                    this.entityType = this.Type.AsElementType() as IEdmEntityType;
+                }
+
+                return this.entityType;
+            }
         }
 
         public bool IncludeInServiceDocument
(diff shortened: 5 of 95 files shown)


src/Microsoft.OData.Core/UnknownEntitySet.cs
```diff
@@ -44,6 +45,19 @@ public IEdmType Type
             }
         }
 
+        public IEdmEntityType EntityType
+        {
+            get
+            {
+                if (this.entityType == null)
+                {
+                    this.entityType = this.Type?.AsElementType() as IEdmEntityType;
```
wachugamaina, line 54: Is there a possibility that this casting may result in null ?

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsNavigationSource.cs
```diff
@@ -87,6 +90,20 @@ public IEnumerable<IEdmNavigationPropertyBinding> NavigationPropertyBindings
             get { return this.navigationTargetsCache.GetValue(this, ComputeNavigationTargetsFunc, null); }
         }
 
+        public IEdmEntityType EntityType
+        {
+            get
+            {
+                if (!this.entityTypeSet)
```
wachugamaina, line 97: Why not check for !this.entityType  == null ?

test/EndToEndTests/Tests/Client/Build.Desktop/DeltaTests/DeltaTests.cs
```diff
@@ -48,20 +48,21 @@ public void RequestDeltaLink()
                 {
                     using (var messageReader = new ODataMessageReader(responseMessage, readerSettings, Model))
                     {
-                        var deltaReader = messageReader.CreateODataDeltaReader(customersSet, customerType);
```
wachugamaina, line 51: This change seems unrelated to this Pr, is it ?

test/EndToEndTests/Tests/Client/Build.Desktop/DeltaTests/DeltaTests.cs
```diff
@@ -161,42 +162,47 @@ public void RequestDeltaLink_Expanded()
...
-                                    case ODataReaderState.Completed:
-                                        Assert.NotNull(deltaReader.Item);
-                                        break;
-                                }
+
+                            }
+                            else if (deltaReader.State == ODataReaderState.NestedResourceInfoEnd)
+                            {
+                                Assert.NotNull(deltaReader.Item as ODataNestedResourceInfo);
+                                //switch (deltaReader.SubState)
```
wachugamaina, line 184: Please delete unused code, you can always retrieve them from Git.

src/Microsoft.OData.Core/UnknownEntitySet.cs
gathogojr: @wachugamaina That's correct. It may result into null and even the previous extension method that was being used in place of this property could return null in some cases. The caller is supposed to anticipate that. I'll mention that it can be null in the property doc.

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsNavigationSource.cs
gathogojr: @wachugamaina Because the value can be null, I didn't want the expression `this.Type?.AsElementType() as IEdmEntityType;` to be executed multiple times since this property is referenced a lot across the library including in hot paths

test/EndToEndTests/Tests/Client/Build.Desktop/DeltaTests/DeltaTests.cs
gathogojr: Yes it is. The build pipeline for 8.x is not running the E2E tests currently so a previous 8.x pull request got rid of `CreateODataDeltaReader` since it's deprecated and inadvertently forgot to update this test. I tried to build the E2E solution locally and it couldn't run so I fixed it in this pull request

test/EndToEndTests/Tests/Client/Build.Desktop/DeltaTests/DeltaTests.cs
gathogojr: @wachugamaina I'll replace the commented lines with better assert statements. I noticed that tests have insufficient assert statements. Checking that the item is not null doesn't feel sufficient enough to me

habbes (approved):

src/Microsoft.OData.Edm/Schema/AmbiguousEntitySetBinding.cs
```diff
@@ -41,6 +43,19 @@ public IEdmType Type
             get { return new EdmCollectionType(new EdmEntityTypeReference(new BadEntityType(String.Empty, this.Errors), false)); }
         }
 
+        public IEdmEntityType EntityType
```
ElizabethOkerio, line 46: Can this method be shared instead of duplicating it in most of these classes?

src/Microsoft.OData.Edm/Schema/BadEntitySet.cs
```diff
@@ -56,6 +57,19 @@ public IEdmType Type
             get { return new EdmCollectionType(new EdmEntityTypeReference(new BadEntityType(String.Empty, this.Errors), false)); }
         }
```
ElizabethOkerio, line 59: Add doc string

src/Microsoft.OData.Edm/Schema/AmbiguousSingletonBinding.cs
```diff
@@ -22,6 +24,19 @@ public IEdmType Type
             get { return new BadEntityType(String.Empty, this.Errors); }
         }
 
+        public IEdmEntityType EntityType
```
ElizabethOkerio, line 27: Add doc string

src/Microsoft.OData.Edm/Schema/AmbiguousEntitySetBinding.cs
ElizabethOkerio: Add doc string

test/EndToEndTests/Tests/Client/Build.Desktop/DeltaTests/DeltaTests.cs
```diff
@@ -22,217 +25,918 @@ public DeltaTests()
...
-            ODataMessageReaderSettings readerSettings = new ODataMessageReaderSettings() { BaseUri = ServiceBaseUri };
+            var readerSettings = new ODataMessageReaderSettings { BaseUri = ServiceBaseUri };
 
-            foreach (var mimeType in deltaMimeTypes)
-            {
-                var requestMessage = new HttpWebRequestMessage(new Uri(ServiceBaseUri.AbsoluteUri + "$delta", UriKind.Absolute));
-                requestMessage.SetHeader("Accept", mimeType);
-                var responseMessage = requestMessage.GetResponse();
-                Assert.Equal(200, responseMessage.StatusCode);
+            var requestMessage = new HttpWebRequestMessage(new Uri(ServiceBaseUri.AbsoluteUri + "$delta", UriKind.Absolute));
```
ElizabethOkerio, line 44: Is this class from OData Client.. I thought we moved away from it?

src/Microsoft.OData.Edm/Schema/AmbiguousEntitySetBinding.cs
gathogojr: @ElizabethOkerio Would you be happy with this:
```csharp
        public IEdmEntityType EntityType
        {
            get
            {
                if (this.entityType == null)
                {
#pragma warning disable CS0618 // Type or member is obsolete
                    this.entityType = this.EntityType();
#pragma warning restore CS0618 // Type or member is obsolete
                }

                return this.entityType;
            }
        }
```
Basically mark the existing `EntityType` extension method as `Obsolete` with the message that it'll not be publicly accessible starting 9.0. But because it's obsolete and I'm calling it from this property (to avoid repeating `this.Type.AsElementType() as IEdmEntityType;`  in a lot of places), apply the `#pragma warning disable` statements to suppress the suggestion/warning by Visual Studio to mark to property as obsolete too...

test/EndToEndTests/Tests/Client/Build.Desktop/DeltaTests/DeltaTests.cs
gathogojr: @ElizabethOkerio This class is not from the OData client library. It is defined in a test project for testing purposes https://github.com/OData/odata.net/blob/fce8fd9bf4942de679918ba3258d4b8dba7660f0/test/EndToEndTests/Tests/Client/Build.Desktop/Common/HttpWebRequestMessage.cs#L18. It implements `IODataRequestMessageAsync` and `IContainerProvider` interfaces.

src/Microsoft.OData.Edm/Schema/AmbiguousEntitySetBinding.cs
ElizabethOkerio: What of the property? The property does not need a doc string describing what it does? Also the method will not be removed per se...Maybe it will be converted to an internal or private method. But having it as obsolete is ok.

wachugamaina (dismissed):

src/Microsoft.OData.Edm/Schema/BadEntitySet.cs
gathogojr: The core implementations of `IEdmNavigationSource` are `EdmNavigationSource` and `CsdlSemanticsNavigationSource`

The following for implementations of `IEdmNavigationSource` are used in rare and special scenarios: `AmbiguousEntitySetBinding`, `AmbiguousSingletonBinding`, `BadEntitySet` and `UnknownEntitySet`. For these category of implementations, I have simply applied `/// <inheritdoc>`. That should suffice

src/Microsoft.OData.Edm/Schema/AmbiguousSingletonBinding.cs
gathogojr: See this comment https://github.com/OData/odata.net/pull/2887#discussion_r1533480789

src/Microsoft.OData.Edm/Schema/AmbiguousEntitySetBinding.cs
gathogojr: REF: https://github.com/OData/odata.net/pull/2887#discussion_r1533480789

See this comment https://github.com/OData/odata.net/pull/2887#discussion_r1533480789

gathogojr: REF: https://github.com/OData/odata.net/pull/2887#discussion_r1524302761

In the attempt to avoid repeating the statement `Type.AsElementType() as IEdmEntityType` that is repeated across `EdmNavigationSource`, `CsdlSemanticsNavigationSource`, `AmbiguousEntitySetBinding`, `AmbiguousSingletonBinding`, `BadEntitySet`, and `UnknownEntitySet`, I considered the following:
- Invoking the existing `EntityType` extension method from the the various implementations of `IEdmNavigationSource.EntityType` property. Considering that we have to add the `Obsolete` attribute to the extension method to signal that it'll cease being publicly available starting 9.0, this introduces warnings suggesting that the properties from where the deprecated method is called from should be also be marked with `Obsolete` attribute too. Suppressing those warnings with `pragma` statements makes the code look worse that it did before.
- Introducing a new internal extension method implemented in a similar manner as the deprecated `EntityType` extension method. This method would be called from the various implementations of `IEdmNavigationSource.EntityType` property. This option seems to introduce a problem to solve another for very little gain.

For the above reasons, I resolved to retain my initial implementation. The implementations are not subject to change with regularity so the angle of maintenance is moot (arguably)

habbes (approved):

ElizabethOkerio (approved):

PR #2892, opened 2024-03-14
Throw an ODataException with helpful message when type name in context URL is not qualified

### LINKED ISSUE #2891: Mitigate ArgumentOutOfRangeException when the typename in context URL is not fully qualified
When the typename in the context URL is not fully qualified, an `ArgumentOutOfRangeException` is thrown when trying to materialize the payload. The exception happens when trying to split the typename into namespace and type name (unqualified).

### Assemblies affected

- Microsoft.OData.Client 7.20.0
- Microsoft.OData.Core 7.20.0

### Reproduce steps

Try to materialize a payload with a context URL containing unqualified typename as follows:
```json
{
    "@odata.context": "http://localhost:5150/$metadata#BookInfo",
    "NoOfPages": 2878,
    "NoOfWords": 81180
}
```
where,
- `BookInfo` is a complex type

### Expected result

An `ODataException` with a descriptive error message is thrown.

### Actual result

An `ArgumentOutOfRangeException` with the message `length ('-1') must be a non-negative value. (Parameter 'length'). Actual value was -1.` is thrown.

### Additional detail

Payload is materialized successfully if the typename is fully qualified as follows:
```json
{
    "@odata.context": "http://localhost:5150/$metadata#Ns.Models.BookInfo",
    "NoOfPages": 2878,
    "NoOfWords": 81180
}
```

Related issue https://github.com/OData/odata.net/issues/2889

gathogojr:
### Issues

*This pull request fixes #2891.*

### Description

Throw an `ODataException` with helpful message when type name in context URL is not qualified - An unhandled `ArgumentOutOfRangeException` was being thrown before.

- [x] *Test cases added*


xuzhg (approved):

src/Microsoft.OData.Core/TypeUtils.cs
```diff
@@ -103,6 +103,11 @@ internal static void ParseQualifiedTypeName(string qualifiedTypeName, out string
             }
 
             int separator = qualifiedTypeName.LastIndexOf(".", StringComparison.Ordinal);
+            if (separator == -1)
```
xuzhg, line 106: I assume it's for reading. 
How about to writing? It's writing the full-qualified type name in the context Url?

xuzhg (dismissed):

test/FunctionalTests/Microsoft.OData.Core.Tests/TypeUtilsTests.cs
```diff
@@ -0,0 +1,37 @@
...
+            Assert.Equal("Ns", namespaceName);
+            Assert.Equal("TypeName", typeName);
+            Assert.Equal(isCollection, outIsCollection);
+        }
+
+        [Fact]
+        public void ParseQualifiedTypeNameThrowsExceptionForTypeNameNotQualified()
+        {
+            // Arrange & Act
+            var exception = Assert.Throws<ODataException>(() => TypeUtils.ParseQualifiedTypeName("TypeName", out _, out _, out _));
```
wachugamaina, line 31: Could we also test an unqualified collection ?

wachugamaina (approved):

gathogojr:
The bulk of the logic for parsing the fragment after the `#` is in [`ParseContextUriFragment`](https://github.com/OData/odata.net/blob/fce8fd9bf4942de679918ba3258d4b8dba7660f0/src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs#L280).

Below is how the fragment looks like for different results returned from an Edm function
- Function returning an entity set/entity collection: `#<EntitySet>` - e.g. `#Books`
- Function returning an entity: `#<EntitySet>/$entity` - e.g. `#Books/$entity`
- Function returning a complex collection: `#Collection(<QualifiedTypeName>)` - e.g. `#Collection(Ns.Models.BookInfo)`
- Function returning a complex type: `#<QualifiedTypeName>` - e.g. `#Ex286.Models.BookInfo`
- Function returning a primitive collection: `#Collection(<QualifiedTypeName>)` - e.g. `#Collection(Edm.Int32)`
- Function returning a primitive: `#<QualifiedTypeName>` - e.g. `#Edm.Int32`

The fragment from the result of calling an Edm function returning an entity is handled by first handled by the code block in the link below:

_NOTE: This execution flow does not involve calling the `ParseQualifiedTypeName` method_

https://github.com/OData/odata.net/blob/fce8fd9bf4942de679918ba3258d4b8dba7660f0/src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs#L286-L291
and subsequently by the `else` block corresponding to:

https://github.com/OData/odata.net/blob/fce8fd9bf4942de679918ba3258d4b8dba7660f0/src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs#L359

<br/>
The fragment from the result of calling an Edm function returning an entity set is handled by the `if` block in the link below:
https://github.com/OData/odata.net/blob/fce8fd9bf4942de679918ba3258d4b8dba7660f0/src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs#L359
We try to find the navigation source (e.g. `Books`) from the Edm model:

_NOTE: This execution flow does not involve calling the `ParseQualifiedTypeName` method_

https://github.com/OData/odata.net/blob/fce8fd9bf4942de679918ba3258d4b8dba7660f0/src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs#L376-L385

<br/>
The else block handles the fragment from the result of calling an Edm function returning a complex collection, complex object, primitive collection and a primitive. It relies on the `ResolveType` method to do that:

https://github.com/OData/odata.net/blob/fce8fd9bf4942de679918ba3258d4b8dba7660f0/src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs#L385-L392

As per the code comment, the else block handles: `Property: {schema.type} or Collection({schema.type}) where schema.type is primitive or complex.`
This is the execution flow that might involve calling the `ParseQualifiedTypeName` method. This call happens in very limited cases where the type is not found in the model:

https://github.com/OData/odata.net/blob/fce8fd9bf4942de679918ba3258d4b8dba7660f0/src/Microsoft.OData.Core/JsonLight/ODataJsonLightContextUriParser.cs#L504-L511

In the case reported by the customer, if the typename was fully qualified (i.e., `microsoft.graph.workbookRange`), it would be found in the model and there would be no need to call `ParseQualifiedTypeName` method.

If for any reason we are unable to resolve the model, we call `ParseQualifiedTypeName` method to split the input into namespace and typename and then create an `EdmUntypedStructureType` object.

<br/>
The fix in this PR is therefore safe to the extent that the logic is executed when all other ways of resolving the expected types have failed...

habbes (approved):

wachugamaina (approved):

PR #2918, opened 2024-04-03
fix ArgumentNullException exception when loading invalid CSDL

quanterion:
When loading invalid CSDL model via
**CsdlReader.Parse** we get correct exception **EdmException**. But if we do it via **CsdlReader.TryParse** with **ignoreUnexpectedAttributesAndElements** option enabled we get instead **System.ArgumentNullException: Value cannot be null. (Parameter 'key'**) because adding null version fails here https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Edm/Csdl/CsdlReader.cs#L339 and as get no critical errors due to https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Edm/Csdl/CsdlReader.cs#L483

### DIFF
--- a/src/Microsoft.OData.Edm/Csdl/CsdlReader.cs
+++ b/src/Microsoft.OData.Edm/Csdl/CsdlReader.cs
@@ -312,9 +312,9 @@ namespace Microsoft.OData.Edm.Csdl
             Version edmxVersion;
             CsdlModel astModel;
 
-            TryParseCsdlFileToCsdlModel(out edmxVersion, out astModel);
+            bool parsed = TryParseCsdlFileToCsdlModel(out edmxVersion, out astModel);
 
-            if (!this.HasIntolerableError())
+            if (!this.HasIntolerableError() && parsed)
             {
                 List<CsdlModel> referencedAstModels = this.LoadAndParseReferencedCsdlFiles(edmxVersion);
 


gathogojr:
Hi @quanterion. Thank you for your pull request contribution. Could you please add a test to confirm the behaviour resulting from your change?

src/Microsoft.OData.Edm/Csdl/CsdlReader.cs
```diff
@@ -312,9 +312,9 @@ private bool TryParse(IEnumerable<IEdmModel> referencedModels, bool includeDefau
             Version edmxVersion;
             CsdlModel astModel;
 
-            TryParseCsdlFileToCsdlModel(out edmxVersion, out astModel);
+            bool parsed = TryParseCsdlFileToCsdlModel(out edmxVersion, out astModel);
 
-            if (!this.HasIntolerableError())
+            if (!this.HasIntolerableError() && parsed)
```
wachugamaina, line 317: Do we have tests covering both parsed is true and false ?

ElizabethOkerio:
@quanterion Add tests

ElizabethOkerio (changes requested):
Add tests

WanjohiSammy:
@quanterion as there has been no activity or response on this PR for a while, I am going to close it for now. If you would like to continue working on it, please feel free to reopen it.

Issue #2228, opened 2021-10-21
UseDefault could be add into IEdmVocabularyAnnotation interface


xuzhg:
In the PR: https://github.com/OData/odata.net/pull/2173, we will use "UseDefault" property to figure out whether it's used the default value from term or not.

In the next main (breaking change) version, we should reconsider it:

1) Add "UseDefault" to the IEdmVocabularyAnnotation interface
2) Or if IEdmVocabularyAnnotation doesn't have the value, we should not write it into the CSDL.

corranrogue9:
From the investigations meeting:

It has been suggested (by myself) that removing the EDM interfaces and replacing them with concrete types where appropriate would be a good idea. 2 existing scenarios were enumerated for why the interfaces are necessary, 1 existing scenario was enumerated for why the interfaces are helpful (but not strictly necessary), 1 known future scenario was enumerated for why the interaces continue to be necessary, and 1 potential future scenario was enumerate for why the interfaces provide existensibility that wouldn't be available with concrete types.

### Existing necessary scenarios
1. The interfaces are used today for mocking in unit tests
2. The interfaces could be used by customers to provide custom implementations for each type

### Existing helpful scenarios
1. It is helpful to have a read-only view into the EDM models

### Known future scenario
1. We want to allow customers to update their models after, for example, the model has been read from a CSDL file. The interfaces being a read-only view give the option for the concrete types to be used for post-read modifications, with the interfaces being used once those modifications are complete. 

### Potential future scenario
1. There is a potential scenario where the full model being loaded into memory is not necessary for a customer. With the current structure of the `IEdmModel` interface, this means that, if all other EDM types were concrete types, then any instance of `IEdmModel` would necessarily need to load portions of the model into memory even if those portions aren't used. A new implementation of non-model EDM interfaces would allow for this even though the concrete types would not.

corranrogue9:
With the above scenarios in mind, we should move the `UseDefault` property to the interface assuming that the EDM types haven't been re-structured in such a way that makes this work no longer relevant.

corranrogue9:
Garrett, follow-up with Sam about the all-up investigation for EDM since a lot of those items are assigned to you.

corranrogue9:
[...]

PR #2935, opened 2024-04-17
Add UseDefault property to IEdmVocabularyAnnotation

ElizabethOkerio:
### Issues

This PR fixes https://github.com/OData/odata.net/issues/2228

### Description

This PR adds the `UsesDefault` property to `IEdmVocabularyAnnotation`.  
We do not write an `EdmVocabularyAnnotation's` value if it has a default value. 
That's why we do the check and set the `UsesDefault` value to either true or false. 

When creating an `EdmVocabularyAnnotation` if the value is not provided, we check whether the EdmTerm has a default value or not. If the term has a default value and no value was provided, we set the `UsesDefault` to true, otherwise false. 

This PR already has this logic and tests. The only change that I made was: 
1. I added a new extension method for creating an `EdmVocabularyAnnotation` when a value is provided as we already have an extension method for creating an `EdmVocabularyAnnotation` when a value is not provided. 
2. I removed the logic that we were using to cast the `IEdmVocabularyAnnotation` to internal implementations of the interface and replaced it with a call to the interface's new added property `UsesDefault`. This now means that we can support any custom implementation of the `IEdmVocabularyAnnotation`. 
3. I added a test for a custom implementation of the `IEdmVocabularyAnnotation`. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
@@ -446,23 +446,6 @@ namespace Microsoft.OData.Edm.Csdl.Serialization
             }
         }
 
-        internal static bool IsUsingDefaultValue(IEdmVocabularyAnnotation annotation)
-        {
-            EdmVocabularyAnnotation edmAnnotation = annotation as EdmVocabularyAnnotation;
-            if (edmAnnotation != null)
-            {
-                return edmAnnotation.UseDefault;
-            }
-
-            CsdlSemanticsVocabularyAnnotation csdlAnnotation = annotation as CsdlSemanticsVocabularyAnnotation;
-            if (csdlAnnotation != null)
-            {
-                return csdlAnnotation.UseDefault;
-            }
-
-            return false;
-        }
-
         internal override void WriteInlineExpression(IEdmExpression expression)
         {
             IEdmPathExpression pathExpression = expression as IEdmPathExpression;
@@ -530,7 +513,7 @@ namespace Microsoft.OData.Edm.Csdl.Serialization
             this.WriteRequiredAttribute(CsdlConstants.Attribute_Term, annotation.Term, this.TermAsXml);
             this.WriteOptionalAttribute(CsdlConstants.Attribute_Qualifier, annotation.Qualifier, EdmValueWriter.StringAsXml);
 
-            if (isInline && !IsUsingDefaultValue(annotation))
+            if (isInline && !annotation.UsesDefault)
             {
                 // in xml format, we can (should) skip writing the expression value if it matches the term default value.
                 this.WriteInlineExpression(annotation.Value);
--- a/src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
+++ b/src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
@@ -44,7 +44,7 @@ namespace Microsoft.OData.Edm.Vocabularies
             this.term = term;
             this.qualifier = qualifier;
             this.value = value;
-            UseDefault = false;
+            UsesDefault = false;
         }
 
         /// <summary>
@@ -81,8 +81,7 @@ namespace Microsoft.OData.Edm.Vocabularies
 
         /// <summary>
         /// Gets whether the annotation uses a default value.
-        /// In vNext, need refactor this.
         /// </summary>
-        internal bool UseDefault { get; set; }
+        public bool UsesDefault { get; set; }
     }
 }
--- a/src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmTermExtensions.cs
+++ b/src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmTermExtensions.cs
@@ -45,8 +45,25 @@ namespace Microsoft.OData.Edm.Vocabularies
 
             return new EdmVocabularyAnnotation(target, term, qualifier, value)
             {
-                UseDefault = true
+                UsesDefault = true
             };
+        }
+
+        /// <summary>
+        /// Creates an <see cref="IEdmVocabularyAnnotation"/> when the annotation value is provided.
+        /// </summary>
+        /// <param name="term">Term bound by the annotation.</param>
+        /// <param name="target">Element the annotation applies to.</param>
+        /// <param name="value">Expression producing the value of the annotation.</param>
+        /// <param name="qualifier">Qualifier used to discriminate between multiple bindings of the same property or type.</param>
+        /// <returns>The <see cref="IEdmVocabularyAnnotation"/> built.</returns>
+        public static IEdmVocabularyAnnotation CreateVocabularyAnnotation(this IEdmTerm term, IEdmVocabularyAnnotatable target, IEdmExpression value, string qualifier = null)
+        {
+            EdmUtil.CheckArgumentNull(target, "target");
+            EdmUtil.CheckArgumentNull(term, "term");
+            EdmUtil.CheckArgumentNull(value, "value");
+
+            return new EdmVocabularyAnnotation(target, term, qualifier, value);
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmVocabularyAnnotation.cs
+++ b/src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmVocabularyAnnotation.cs
@@ -30,5 +30,10 @@ namespace Microsoft.OData.Edm.Vocabularies
         /// Gets the expression producing the value of the annotation.
         /// </summary>
         IEdmExpression Value { get; }
+
+        /// <summary>
+        /// Gets whether the annotation uses a default value.
+        /// </summary>
+        bool UsesDefault { get; }
     }
 }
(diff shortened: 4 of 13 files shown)


src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmVocabularyAnnotation.cs
```diff
@@ -30,5 +30,10 @@ public interface IEdmVocabularyAnnotation : IEdmElement
         /// Gets the expression producing the value of the annotation.
         /// </summary>
         IEdmExpression Value { get; }
+
+        /// <summary>
+        /// Checks if the annotation uses a default value.
+        /// </summary>
+        bool UseDefault { get; }
```
xuzhg, line 37: We don't need a boolean to indicate whether the annotation contains a default value.
We can add a `String DefaultValue {get;}` to the interface.

xuzhg (changes requested):

src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmVocabularyAnnotation.cs
ElizabethOkerio: @xuzhg  can you please provide more info on how this will be used. @mikepizzo.

ElizabethOkerio: I understand that it will be used to determine when to write a vocabulary annotation's default value. That's if the annotation doesn't have a value but has a default value, then the default value is written. This support already exists but maybe was done internally to avoid breaking customers. An annotation term already has a `DefaultValue` property. By having this property in the `IEdmVocabularyAnnotation` how are we envisioning it to be used?

ElizabethOkerio: @corranrogue9 Could you please provide more info here if you have more info.

xuzhg: Context:

1) We can build a EdmTerm with default value.  See https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmTerm.cs#L27

2) When we create a Vocabulary annotation, we need the "Target Term" and the "Annotation Value" which is specified using the 'EdmVocabularyAnnotation" constructor. **When we write this annotation into CSDL, we need to write the 'Annotation Value' into the CSDL.** [Example](https://github.com/OData/odata.net/blob/main/test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs#L2230)

3) Or, When we create a vocabulary annotation, we can use the `Default Value` from the target term. In this case, we don't need to specify the 'Annotation Value' and build it from the `Default Value` of target term.  **When we write this annotation into CSDL, we don't need to write the annotation value into the CSDL.**  [Example](https://github.com/OData/odata.net/blob/main/test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs#L2235)

4)  Since we have two kinds of vocabulary annotation to write. We need a value to distinguish them. That's why we have 'bool UseDefault'.

So, for the new implementation, my suggestion is that:

1) we don't need 'bool UseDefault { get; }' in `EdmVocabularyAnnotation.cs`  .
2) we don't need 'bool UseDefault { get; }' added into `IEdmVocabularyAnnotation` if we can distinguish using the 'Value'.
 or if the 'Value==null', can we say it's using the default value of Term?
     if the 'Value!=null', can we say it's using it's specified value?
     
3) or can we create a new type to represent the 'default value' it's just a 'tag' or a pointer. 
    ```C#
    public class DefaultValueExpression : IEdmExpression 
    { //we don't need to create the value, maybe add a pointer to the term or the default value in the term? }
   ```

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
```diff
@@ -525,7 +508,7 @@ internal override void WriteVocabularyAnnotationElementHeader(IEdmVocabularyAnno
             this.WriteRequiredAttribute(CsdlConstants.Attribute_Term, annotation.Term, this.TermAsXml);
             this.WriteOptionalAttribute(CsdlConstants.Attribute_Qualifier, annotation.Qualifier, EdmValueWriter.StringAsXml);
 
-            if (isInline && !IsUsingDefaultValue(annotation))
+            if (isInline && !annotation.UseDefault)
```
wachugamaina, line 511: Will this retain the same behavior as the deleted function?

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
```diff
@@ -80,9 +80,8 @@ public IEdmExpression Value
         }
 
         /// <summary>
-        /// Gets whether the annotation uses a default value.
-        /// In vNext, need refactor this.
+        /// Checks if the annotation uses a default value. </param>
         /// </summary>
-        internal bool UseDefault { get; set; }
+        public bool UseDefault { get; set; }
```
wachugamaina, line 85: Why are we making these public ?

src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmVocabularyAnnotation.cs
ElizabethOkerio: @xuzhg Is this an optimization thing, not writing the default value into CSDL when the value is not provided? The issue with the current implementation is that we do not support custom implementations? Like when checking when not to write the annotation value to CSDL? Is there a use case for this? Are there customers using custom implementations? 

For the third option, I'm not sure what the right `EdmExpressionKind` will be? We'll also have to implement all the different interfaces for the various `EdmExpressionKinds` which I'm not sure if it's ok or not?

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
ElizabethOkerio: Yes

src/Microsoft.OData.Edm/Vocabularies/Annotations/EdmVocabularyAnnotation.cs
ElizabethOkerio: We added this to a public interface that this class implements.

src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmVocabularyAnnotation.cs
ElizabethOkerio: It's ok to have this value as a boolean in `IEdmVocabularyAnnotation`

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.VocabularyAnnotation.cs
habbes, line 1: I thought part of the goal of this PR was to support round-tripping: i.e. if you read a CSDL without with terms using default (omitted) values and then serialize the CSDL you should be able to get an output CSDL where default value are omitted.

Bearing that, I think you should add tests that demonstrate reading a CSDL with omitted annotation values and show the `UsesDefault` flag is applied correctly.

ElizabethOkerio: There are tests already for this: https://github.com/lisicase/odata.net/blob/878c814662835ba05de7f4495de8537741af91d1/test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs#L416C9-L416C16

habbes: So the main point of this PR was to allow support custom implementations?

habbes (approved):

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.VocabularyAnnotation.cs
ElizabethOkerio: Yea somehow and to get rid of the casts that we were doing.

src/Microsoft.OData.Edm/Vocabularies/Annotations/IEdmTermExtensions.cs
```diff
@@ -43,12 +31,28 @@ public static IEdmVocabularyAnnotation CreateVocabularyAnnotation(this IEdmTerm
                 throw new InvalidOperationException(Strings.EdmVocabularyAnnotations_DidNotFindDefaultValue(term.Type));
             }
 
-            return new EdmVocabularyAnnotation(target, term, qualifier, value)
```
mikepizzo, line 46: why did we get rid of the overload that takes a qualifier?

ElizabethOkerio: It's because we were passing null to it and constructor call also passes null : `        public EdmVocabularyAnnotation(IEdmVocabularyAnnotatable target, IEdmTerm term, IEdmExpression value)
            : this(target, term, null, value)`

ElizabethOkerio: It will be different though if someone calls the extension overload that takes the `qualifier`. I'll add it back.

mikepizzo (dismissed):

habbes (dismissed):

PR #2939, opened 2024-04-19
Add public ShouldIncludeAnnotation property

### LINKED ISSUE #2844: Instance annotations are not written by default
When the caller adds instance annotations to a response resource, those annotations are only included in the response payload if the client has provided the `Prefer: odata.include-annotations=` request header. We should included instance annotations added by the caller by default and honor the `Prefer` header by filtering out those annotations which do not meet the header.

habbes:
### Issues

*This pull request fixes #2844 

### Description

This PR adds a `ShouldIncludeAnnotation` property to the `ODataMessageWriterSettings` that allows developers to force a custom instance annotation to be written even if it was not included in the preference header annotation filters (`odata.include-annotations:`)

It works as follows:
- The property is a `Func<string, bool>` that receives an annotation name and returns true if the annotation should be written
- If the writer was going to skip a certain annotation (e.g. because it was not part of the annotations filter), it will check this delegate before skipping the annotation. If the delegate returns true, then the annotation will not be skipped.

Note that if the writer was going to write an annotation (e.g. because of it's part of the annotation filters), then the annotation will still be written even if this `ShouldIncludeAnnotation` delegate returns false. **That is to say, this property does not prevent an annotation from being written, it can only prevent an annotation from being skipped**.  I did this intentionally because I don't think this property should prevent writing annotations where the standard requires that they be written. However, if you have a contrary opinion, please share your thoughts in the discussion or review comments.

There was already an internal property called `ShouldIncludeAnnotation`, which is also a `Func<string, bool>`. This is generated automatically by `ODataMessageWriter` based on the annotations filter in the header. To avoid confusion, I renamed this property to `ShouldIncludeAnnotationInternal`.

The main logic of this PR is implemented in `ODataMessageWriterSettings.ShouldSkipAnnotation` method.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataMessageWriter.cs
+++ b/src/Microsoft.OData.Core/ODataMessageWriter.cs
@@ -116,8 +116,8 @@ namespace Microsoft.OData
             this.message = new ODataRequestMessage(requestMessage, /*writing*/ true, this.settings.EnableMessageStreamDisposal, /*maxMessageSize*/ -1);
 
             // Always include all annotations when writing request message.
-            Debug.Assert(this.settings.ShouldIncludeAnnotation == null, "this.settings.ShouldIncludeAnnotation == null");
-            this.settings.ShouldIncludeAnnotation = AnnotationFilter.CreateIncludeAllFilter().Matches;
+            Debug.Assert(this.settings.IsAnnotationFiltered == null, "this.settings.ShouldIncludeAnnotation == null");
+            this.settings.IsAnnotationFiltered = AnnotationFilter.CreateIncludeAllFilter().Matches;
         }
 
         /// <summary> Creates a new <see cref="Microsoft.OData.ODataMessageWriter" /> for the given response message. </summary>
@@ -159,7 +159,7 @@ namespace Microsoft.OData
             string annotationFilter = responseMessage.PreferenceAppliedHeader().AnnotationFilter;
             if (!string.IsNullOrEmpty(annotationFilter))
             {
-                this.settings.ShouldIncludeAnnotation = ODataUtils.CreateAnnotationFilter(annotationFilter);
+                this.settings.IsAnnotationFiltered = ODataUtils.CreateAnnotationFilter(annotationFilter);
             }
         }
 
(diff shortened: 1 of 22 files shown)


corranrogue9 (dismissed):

src/Microsoft.OData.Core/ODataMessageWriter.cs
```diff
@@ -116,8 +116,8 @@ public ODataMessageWriter(IODataRequestMessage requestMessage, ODataMessageWrite
             this.message = new ODataRequestMessage(requestMessage, /*writing*/ true, this.settings.EnableMessageStreamDisposal, /*maxMessageSize*/ -1);
 
             // Always include all annotations when writing request message.
-            Debug.Assert(this.settings.ShouldIncludeAnnotation == null, "this.settings.ShouldIncludeAnnotation == null");
-            this.settings.ShouldIncludeAnnotation = AnnotationFilter.CreateIncludeAllFilter().Matches;
+            Debug.Assert(this.settings.IsAnnotationFiltered == null, "this.settings.ShouldIncludeAnnotation == null");
```
gathogojr, line 119:
```suggestion
            Debug.Assert(this.settings.IsAnnotationFiltered == null, "this.settings.IsAnnotationFiltered == null");
```

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
```diff
@@ -367,19 +377,24 @@ internal SelectedPropertiesNode SelectedProperties
...
 
         /// <summary>
-        /// Func to evaluate whether an annotation should be written by the writer. The func should return true if the annotation should
+        /// Func to evaluate whether an annotation should be written by the writer based on the annotations filtered in the odata.include-annotations header and OData standard.
+        /// The func should return true if the annotation should
         /// be written and false if the annotation should be skipped.
         /// </summary>
-        internal Func<string, bool> ShouldIncludeAnnotation
+        /// <remarks>
+        /// This property is internal and automatically set based on annotation filters. The developer can override this using the <see cref="ShouldIncludeAnnotation"/> property
```
ElizabethOkerio, line 385: Is it really overriden?

ElizabethOkerio (approved):

src/Microsoft.OData.Core/ODataMessageWriterSettings.cs
habbes: I've removed the word `override` to avoid any confusion

gathogojr (approved):

ElizabethOkerio (approved):

xuzhg (approved):

PR #2943, opened 2024-04-23
Remove .NET Framework 4.8 target framework

habbes:
We are not planning to support .NET Framework in the first preview release. This PR therefore removes .net48 from target frameworks. This removes support, but it would be easy to add support again if we change our plan before the final release

Note that as a result of this change the **E2E test projects will not compile**.

### DIFF
--- a/src/Microsoft.OData.Client/Microsoft.OData.Client.csproj
+++ b/src/Microsoft.OData.Client/Microsoft.OData.Client.csproj
@@ -1,6 +1,6 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
-    <TargetFrameworks>net8.0;net48</TargetFrameworks>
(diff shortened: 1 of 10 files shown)


ElizabethOkerio:
I think this PR needs more stuff removed. I believe it can be reverted if we need to retain the .net framework.

habbes:
> I think this PR needs more stuff removed. I believe it can be reverted if we need to retain the .net framework.

@ElizabethOkerio which stuff to be specific?

ElizabethOkerio:
We have conditionals like `if netcoreapp`. I believe those will no longer be needed. There are also instances where we have conditionals like `if netstandard`  and things like that.. This is like having code that might never be executed.

gathogojr (approved):

habbes:
Got an approval from @ElizabethOkerio who was away from keyboard.

PR #2952, opened 2024-04-29
Rename IBaseEntityType.Context to DataServiceContext

### LINKED ISSUE #2291: Context Attribute on Connected Service causes conflict in generated code
When using the Odata Connected Service Plugin for VS to import a metadata document and generate project code for our entities, there is a virtual property added with the POCO name "Context" which conflicts with another property. 

### Assemblies affected

### Reproduce steps
Attaching a file for the Odata metadata endpoint of a D365 service. This should generate and compile correctly, instead you will see an error expecting an Object instead of "string"

### Expected result
The generated code would successfully build

### Actual result
Severity	Code	Description	Project	File	Line	Suppression State
Error	CS1061	'string' does not contain a definition for 'EntityTracker' and no accessible extension method 'EntityTracker' accepting a first argument of type 'string' could be found (are you missing a using directive or an assembly reference?)	ConsoleApp1	C:\Users\anschmidt\source\repos\ConsoleApp1\ConsoleApp1\Connected Services\OData Service\Reference.cs	973169	Active

### Additional detail

        /// <summary>
        /// There are no comments for Property Context in the schema.
        /// </summary>
        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.OData.Client.Design.T4", "#VersionNumber#")]

        [global::Microsoft.OData.Client.OriginalNameAttribute("Context")]
        public virtual string Context
        {
            get
            {
                return this._Context;
            }
            set
            {
                this.OnContextChanging(value);
                this._Context = value;
                this.OnContextChanged();
                this.OnPropertyChanged("Context");
            }
        }

### LINKED ISSUE #2593: Fix naming of Context property in IBaseEntityType interface and BaseEntityType classes that causes conflicts
Fix naming of `Context` property in `IBaseEntityType` interface and `BaseEntityType` classes that causes conflicts whenever an entity type also contains a property named `Context`. Probably by renaming the property to something that is less likely to result into a conflict

### Assemblies affected

Microsoft.OData.Client 7.13.0

### Reproduce steps

1. Create an ASP.NET Core OData application that has a class with the property named `Context`
2. Use the class to define an entity type in the EDM
3. Create a OData client application - using the Console application template for instance
4. Generate proxy classes using OData Connected Service
5. Try to compile the project - You should get a compiler error due to the naming conflict since the auto-generated types on OData client inherit from `BaseEntityType`

### Expected result

The OData client application would compile successfully.

### Actual result

The OData client application does not compile

### Additional detail

The developer can mitigate the issue by renaming the property on the auto-generated types and then applying the `OriginalNameAttribute` to the property with the actual name (`Context`) used on the service side

habbes:
### Issues

*This pull request fixes #2593 and fixes #2291 

### Description

This PR renames the `IBaseEntityType.Context` property to `DataServiceContext`. The name `Context` is a relatively common property name in customer schemas and this is therefore a common source of naming conflicts which result in compilation or runtime errors. We expect `DataServiceContext` to be less susceptible to conflicts, it's also more specific, so this doesn't hurt readability of the code.

The issue could be resolved by the customers by changing the generated property name manually to remove the conflict. But this would means they would have to make that change each time the regenerate the code (e.g. after an update to the CSDL). This leads to unfavorable developer experience.

We may also make changes to the code generation logic in the code gen tools such it generates a different property name when the customer CSDL contains property name that may conflict with built-in property names. That would be more work but definitely not unreasonable. In fact we have merged a pull request that changes generated property names if they would be invalid identifiers, so this is not far off. I'm not opposite to doing that if it's preferrable. But since this issue affects `Context` specifically, and since that not a far-fetched property name, I think it's safe to make change to the internal interface in this case.

Another approach that we have not discussed would be to provide a way for customers to inject custom scripts or hook that would apply user-defined transformations during code generations and these scripts could be saved along with the config so they would be re-applied automatically each time the code generator is executed. But this probably overkill and should be its own feature request that ought to be separately designed.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/BaseEntityType.cs
+++ b/src/Microsoft.OData.Client/BaseEntityType.cs
@@ -16,7 +16,7 @@ namespace Microsoft.OData.Client
         /// </summary>
         protected internal DataServiceContext Context { get; set; }
 
-        DataServiceContext IBaseEntityType.Context
+        DataServiceContext IBaseEntityType.DataServiceContext
         {
             get => Context;
             set => Context = value;
--- a/src/Microsoft.OData.Client/BaseEntityType.cs
+++ b/src/Microsoft.OData.Client/BaseEntityType.cs
@@ -16,7 +16,7 @@ namespace Microsoft.OData.Client
         /// </summary>
         protected internal DataServiceContext Context { get; set; }
 
-        DataServiceContext IBaseEntityType.EntityTrackingContext
+        DataServiceContext IBaseEntityType.DataServiceContext
         {
             get => Context;
             set => Context = value;
--- a/src/Microsoft.OData.Client/BaseEntityType.cs
+++ b/src/Microsoft.OData.Client/BaseEntityType.cs
@@ -16,7 +16,7 @@ namespace Microsoft.OData.Client
         /// </summary>
         protected internal DataServiceContext Context { get; set; }
 
-        DataServiceContext IBaseEntityType.DataServiceContext
+        DataServiceContext IBaseEntityType.EntityTrackingContext
         {
             get => Context;
             set => Context = value;
--- a/src/Microsoft.OData.Client/IBaseEntityType.cs
+++ b/src/Microsoft.OData.Client/IBaseEntityType.cs
@@ -15,7 +15,7 @@ namespace Microsoft.OData.Client
         /// <summary>
         /// DataServiceContext for query provider.
         /// </summary>
-        DataServiceContext Context { get; set; }
+        DataServiceContext DataServiceContext { get; set; }
 
         /// <summary>
         /// Entity descriptor containing entity stream links.
--- a/src/Microsoft.OData.Client/IBaseEntityType.cs
+++ b/src/Microsoft.OData.Client/IBaseEntityType.cs
@@ -15,7 +15,7 @@ namespace Microsoft.OData.Client
         /// <summary>
         /// DataServiceContext for query provider.
         /// </summary>
-        DataServiceContext DataServiceContext { get; set; }
+        DataServiceContext EntityTrackingContext { get; set; }
 
         /// <summary>
         /// Entity descriptor containing entity stream links.
--- a/src/Microsoft.OData.Client/IBaseEntityType.cs
+++ b/src/Microsoft.OData.Client/IBaseEntityType.cs
@@ -15,7 +15,7 @@ namespace Microsoft.OData.Client
         /// <summary>
         /// DataServiceContext for query provider.
         /// </summary>
-        DataServiceContext EntityTrackingContext { get; set; }
+        DataServiceContext DataServiceContext { get; set; }
 
         /// <summary>
         /// Entity descriptor containing entity stream links.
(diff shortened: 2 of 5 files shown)


robertmclaws:
If the value that's being set if coming from `EntityTrackingAdapter.Context`, wouldn't a better name be `EntityTrackingContext`? `DataServiceContext` is entirely too close to a major piece of functionality in WCF Data Services that has an entirely different purpose.

gathogojr (approved):

src/Microsoft.OData.Client/BaseEntityType.cs
```diff
@@ -16,7 +16,7 @@ public class BaseEntityType : IBaseEntityType
         /// </summary>
         protected internal DataServiceContext Context { get; set; }
 
-        DataServiceContext IBaseEntityType.DataServiceContext
+        DataServiceContext IBaseEntityType.EntityTrackingContext
```
robertmclaws, line 19: The source class should also probably change to `EntityTrackingContext` as well.

habbes: I think changing the actual class name from `DataServiceContext` to `EntityTrackingContext` would be a much bigger change. The `DataServiceContext` also does more than entity tracking, it is basically the entry point to the client and can be used even when entity tracking is disabled. I think renaming that class to something else should be a larger discussion.

robertmclaws: Interesting. In that case, I think the property name should change to `DataServiceContext` to match the class name. Sorry about that, I should have dug deeper.

It does seem like the DataServiceContext class could use a solid streamlining.

src/Microsoft.OData.Client/IBaseEntityType.cs
```diff
@@ -15,7 +15,7 @@ public interface IBaseEntityType
         /// <summary>
         /// DataServiceContext for query provider.
         /// </summary>
-        DataServiceContext Context { get; set; }
+        DataServiceContext EntityTrackingContext { get; set; }
```
ElizabethOkerio, line 18: I don't think we should name this property `EntityTrackingContext` maybe `ServiceContext` or `DataServiceContext` could do.

ElizabethOkerio: Also. there are no tests that need to be updated by this change?

src/Microsoft.OData.Client/Materialization/EntryValueMaterializationPolicy.cs
```diff
@@ -682,7 +682,7 @@ private void MaterializeResolvedEntry(MaterializerEntry entry, bool includeLinks
 
             if (entry.ResolvedObject is IBaseEntityType entity)
             {
-                entity.Context = this.EntityTrackingAdapter.Context;
+                entity.EntityTrackingContext = this.EntityTrackingAdapter.Context;
```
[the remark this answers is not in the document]
habbes: I don't think this is necessary.

src/Microsoft.OData.Client/IBaseEntityType.cs
habbes: I have renamed the property to `DataServiceContext`.

ElizabethOkerio (approved):

gathogojr (approved):

PR #2955, opened 2024-05-01
Bump version to 7.21.1

habbes:
Updates version to 7.21.1. This release includes a dependency configuration change that fixes a regression #2954

### DIFF
--- a/tools/CustomMSBuild/Versioning.props
+++ b/tools/CustomMSBuild/Versioning.props
@@ -4,7 +4,7 @@
   <PropertyGroup>
     <VersionMajor Condition="'$(VersionMajor)' == ''">7</VersionMajor>
     <VersionMinor Condition="'$(VersionMinor)' == ''">21</VersionMinor>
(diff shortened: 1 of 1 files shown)


xuzhg (approved):

ElizabethOkerio (approved):

gathogojr (approved):

robertmclaws:
I just want to point out that if we modernized the solution configuration and build system here like I did with Restier, we wouldn't need to do these things... because Visual Studio would calculate the dependencies for the .nupkg file automatically.

habbes:
> I just want to point out that if we modernized the solution configuration and build system here like I did with Restier, we wouldn't need to do these things... because Visual Studio would calculate the dependencies for the .nupkg file automatically.

I think we should do these once we get over the ODL 8 release (or after we've completed the major ODL 8 work, unless modernizing the solution would be a quick fix), so the ODL 8 branch could move to modern build config, modern analyzers, etc. I think similar discussions have been raised. Let me see if I can locate the issues, otherwise we should create one to track this.

PR #2957, opened 2024-05-03
Refactor ODataError, ODataErrorDetail and ODataInnerError classes

### LINKED ISSUE #1421: Why ODataError.ErrorCode instead of ODataError.Code
While reading the [OASIS error response specification](http://docs.oasis-open.org/odata/odata-json-format/v4.0/os/odata-json-format-v4.0-os.html#_Toc372793091) I noticed that the ODataError has a ErrorCode property instead of a Code property.

Does a way exist to serialize ODataError with JsonConvert.SerializeObject so that it spits out a "code" field instead of the "errorCode" field?

I would like my services to be compliant with the spec.
How do you folks managed that? Did some of you implement your own ODataError (that's what I'm currently thinking of)?

gathogojr:
### Issues

*This pull request fixes #1421.*

### Description

Refactor `ODataError`, `ODataErrorDetail` and `ODataInnerError` classes

The OData standard provides the following guidance on error response - [19 Error Response](http://docs.oasis-open.org/odata/odata-json-format/v4.0/os/odata-json-format-v4.0-os.html#_Toc372793091):
- _The error response MUST be a single JSON object. This object MUST have a single name/value pair named `error`. The value must be a JSON object._
- _This object MUST contain name/value pairs with the names `code` and `message`, and it MAY contain name/value pairs with the names `target`, `details` and `innererror`._
- _The value for the `code` name/value pair is a language-independent string. Its value is a service-defined error code. This code serves as a sub-status for the HTTP error code specified in the response._
- _The value for the `message` name/value pair MUST be a human-readable, language-dependent representation of the error. The Content-Language header MUST contain the language code from [RFC5646] corresponding to the language in which the value for message is written._
- _The value for the `target` name/value pair is the target of the particular error (for example, the name of the property in error)._
- _The value for the `details` name/value pair MUST be an array of JSON objects that MUST contain name/value pairs for `code` and `message`, and MAY contain a name/value pair for target, as described above._
- _The value for the `innererror` name/value pair MUST be an object. The contents of this object are service-defined. Usually this object contains information that will help debug the service. The `innererror` name/value pair SHOULD only be used in development environments in order to guard against potential security concerns around information disclosure._

The differences between the OData spec guidance and the ODL implementation are summarized in the following comment https://github.com/OData/odata.net/issues/1421#issuecomment-1309673556

The existing implementation of `ODataError` `ToString` made an attempt to construct a JSON string representing the error response that ODL serializes an `ODataError` object into. The logic however falls short in several ways:
1. While `ODataInnerError`'s `ToJson` method skips any of the 3 string properties of `ODataInnerError` (i.e., `message`, `type`, and `stacktrace`) when the values are not provided, the OData serializers default to empty strings.

    For example,
    ```csharp
    new ODataInnerError()
    ```
    OData serializers produce:

    ```json
    {
        "message": "",
        "type": "",
        "stacktrace": ""
    }
    ```
    `odataInnerError.ToJson()`: `{}`

    In this pull request, I have made a change such that `ToJson` produces the same result as the ODL serializers do.
2. One of the `ODataInnerError`'s constructors accepts a `Dictionary<string, ODataValue>` parameter.
    The thing about `ODataValue` is that it has many subclasses, namely `ODataNullValue`, `ODataPrimitiveValue`, `ODataEnumValue`, `ODataBinaryStreamValue`, `ODataStreamReferenceValue`, `ODataResourceValue`, `ODataCollectionValue`, and `ODataUntypedValue`.
    Out of these, the OData serializers only handle `ODataNullValue`, `ODataPrimitiveValue`, `ODataResourceValue`, and `ODataCollectionValue`. If it finds an instance of any of the other subclasses when serializing, an exception is thrown.
    The problem with the `ToJson` method implementation however is that when an `ODataPrimitiveValue` containing primitive values like `byte[]`, `DateTimeOffset`, `Guid`, `TimeSpan`, `Microsoft.OData.Edm.Date`, and `Microsoft.OData.Edm.TimeOfDay` is present in the dictionary, such values not converted into the expected string literals and the converted value is also not enclosed in double quotes as would be expected.

    For example, the following `ODataInnerError` instance would yield the invalid JSON payload further below:

    ```csharp
    new ODataInnerError(
        new Dictionary<string, ODataValue>
        {
            { "p13", new ODataPrimitiveValue(new byte[] { 1, 2, 3, 4, 5, 6, 7, 8, 9, 0 }) },
            { "p14", new ODataPrimitiveValue(new DateTimeOffset(2024, 5, 2, 13, 27, 30, TimeSpan.Zero)) },
            { "p15", new ODataPrimitiveValue(Guid.Parse("6693ceb2-5d47-45c7-b928-900ebdebe898")) },
            { "p16", new ODataPrimitiveValue(new TimeSpan(1, 12, 0)) },
            { "p17", new ODataPrimitiveValue(new Microsoft.OData.Edm.Date(2024, 5, 2)) },
            { "p18", new ODataPrimitiveValue(new Microsoft.OData.Edm.TimeOfDay(12, 42, 30, 100)) }
        })
      ```
      `ToJson` result:

      ```json
      {
        "p13": System.Byte[],
        "p14": 02-May-24 1:27:30 PM +00:00,
        "p15": 6693ceb2-5d47-45c7-b928-900ebdebe89,
        "p16": 01:12:00,
        "p17": 2024-05-02,
        "p18": 12:42:30.1000000
      }
      ```
   The reason for this invalid result is because of the [buggy logic in `ODataValueToString`](https://github.com/OData/odata.net/blob/bbe2ac74487fb01fe14d97aeca131d0ca89d2452/src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs#L124) that just calls `.ToString()` on the value before appending the result to the string builder.
    In this pull request, I have remedied this by making use of `ODataRawValueUtils.TryConvertPrimitiveToString(object, out string)` to convert the value to the expected string literals and also enclosing the converted value in double quotes.
4. When an `ODataInnerError` object has a nested `ODataInnerError`, the ODL serializers serializes the property as `internalexception` while `ToJson` method adopts `innererror`. In addition, `ToJson` always adds an `innererror` property with an empty object when the `InnerError` nested property is not initialized.

   For example, 
    ```csharp
    var odataError = new ODataError
    {
        InnerError = new ODataInnerError
        {
            InnerError = new ODataInnerError()
        }
    }
    ```
    `ODataJsonWriterUtils.WriteError(odataError, /* other parameters */)`:

    ```json
    {
        "error": {
            "code": "",
            "message": "",
            "innererror": {
                "message": "",
                "type": "",
                "stacktrace": "",
                "internalexception": {
                    "message": "",
                    "type": "",
                    "stacktrace": ""
                }
            }
        }
    }
    ```
    `odataError.ToString()`:

    ```json
    {
        "error": {
            "code": "",
            "message": "",
            "target": "",
            "details": {},
            "innererror": {
                "message": "",
                "type": "",
                "stacktrace": "",
                "innererror": {
                    "message": "",
                    "type": "",
                    "stacktrace": "",
                    "innererror": {}
                }
            }
        }
    }
    ```
    In this pull request, I have made a change such that the nested `ODataInnerError` is named `internalexception` when `ToJson` method is called. In addition, when the nested `InnerError` property is null, it doesn't get added as a property. This mirrors what is produced by ODL serializers.

#### Other changes:
- Renamed `ToJson` methods in `ODataInnerError` and `ODataErrorDetail` to `ToJsonString`.
- When serializing an `ODataErrorDetail` instance, the OData serializers would write the optional `target` property ahead of the required `message` property. I switched the order such that the required `message` property is written first. This makes logical sense and change of property order does not qualify as a breaking change.

    **Before:**

    ```csharp
    var odataError = new ODataError
    {
        Details = new Collection<ODataErrorDetail>
        {
            new ODataErrorDetail { Code = "OEDC1", Message = "OEDM1", Target = "OEDT1" },
            new ODataErrorDetail { Code = "OEDC1", Message = "OEDM1" },
            new ODataErrorDetail {}
        }
    };
    ```
    `ODataJsonWriterUtils.WriteError(odataError, /* other parameters */)`:

    ```json
    {
        "error": {
            "code": "",
            "message": "",
            "details": [
                {
                    "code": "OEDC1",
                    "target": "OEDT1",
                    "message": "OEDM1"
                },
                {
                    "code": "OEDC1",
                    "message": "OEDM1"
                },
                {
                    "code": "",
                    "message": ""
                }
            ]
        }
    }
    ```
    `odataError.ToString()`:

    ```json
    {
      "error": {
        "code": "",
        "message": "",
        "target": "",
        "details": [
          {
            "errorcode": "OEDC1",
            "message": "OEDM1",
            "target": "OEDT1"
          },
          {
            "errorcode": "OEDC1",
            "message": "OEDM1",
            "target": ""
          },
          {
            "errorcode": "",
            "message": "",
            "target": ""
          }
        ],
        "innererror": {}
      }
    }
    ```
    **After:**

    Both `ODataJsonWriterUtils.WriteError(odataError, /* other parameters */)` and `odataError.ToString()`:

    ```json
    {
      "error": {
        "code": "",
        "message": "",
        "details": [
          {
            "code": "OEDC1",
            "message": "OEDM1",
            "target": "OEDT1"
          },
          {
            "code": "OEDC1",
            "message": "OEDM1"
          },
          {
            "code": "",
            "message": ""
          }
        ]
      }
    }
    ```
- Throwing an exception from `ToString()` is considered a very bad idea. For that reason, whenever an `ODataValue` that cannot be converted into a JSON representation is encountered in the `ODataInnerError` dictionary, an error message is written in its place:

    For example:
    ```csharp
    var odataInnerError = new ODataInnerError(
        new Dictionary<string, ODataValue>
        {
            { "p1", new ODataEnumValue("Black") },
            { "p2", new ODataUntypedValue { RawValue = "\"roygbiv\"" } },
            { "p3", new ODataBinaryStreamValue(stream, leaveOpen: false) },
            { "p4", new ODataStreamReferenceValue() },
            { "p5", new ODataJsonElementValue(jsonDocument.RootElement) }
        }
    ```
    `odataInnerError.ToJsonString()`:

    ```json
    {
        "p1": "The value of type 'Microsoft.OData.ODataEnumValue' is not supported and cannot be converted to a JSON representation.",
        "p2": "The value of type 'Microsoft.OData.ODataUntypedValue' is not supported and cannot be converted to a JSON representation.",
        "p3": "The value of type 'Microsoft.OData.ODataBinaryStreamValue' is not supported and cannot be converted to a JSON representation.",
        "p4": "The value of type 'Microsoft.OData.ODataStreamReferenceValue' is not supported and cannot be converted to a JSON representation.",
        "p5": "The value of type 'Microsoft.OData.ODataJsonElementValue' is not supported and cannot be converted to a JSON representation.",
        "message": "",
        "type": "",
        "stacktrace": ""
    }
    ```
    The option of validating that an unsupported `ODataValue` is not added to the dictionary and throwing an exception in such event was considered. However, since `ODataResourceValue` and `ODataCollectionValue` can be heavily nested, with `ODataValue` likely to be present at any level, the idea of such validation looks less attractive and potentially costly.
   Tests were added to confirm the output produced when unsupported values are encountered.
- The OData serializers write the properties in the `ODataInnerError` dictionary ahead of the declared properties. This behaviour was retained and this also meant a change to the `ToJsonString` method to produce a similar output.

    **Before:**
    ```csharp
    var odataInnerError = new ODataInnerError(
        new Dictionary<string, ODataValue>
        {
            { "p1", new ODataPrimitiveValue(1) }
        }
    ```
    `odataInnerError.ToJson()`:

    ```json
    {
        "message": "",
        "type": "",
        "stacktrace": "",
        "p1": 1
    }
    ```
    **After:**
    `odataInnerError.ToJsonString()`:

    ```json
    {
        "p1": 1,
        "message": "",
        "type": "",
        "stacktrace": ""
    }
    ```

#### Be noted
- As per the OData standard, the contents of the `innererror` object are service-defined. The sample in the OData spec has the following `innererror` object:
    ```json
    "innererror": {
        "trace": [...],
        "context": {...}
    }
    ```
    The two properties here are just sample properties. The OData spec doesn't designate them as required or optional properties.
    In ODL, the `ODataInnerError` class has 3 declared properties - `Message`, `Type` and `StackTrace`. ~The OData serializers always write those 3 properties. When any of their values are not provided, an empty string is written. This behaviour has been retained.~ The logic to support Astoria compatibility has been removed. If anyone needs `message`, `type` and `stacktrace` to be included in the serialized `ODataError`, they can add the properties to the `ODataInnerError`'s `Properties` dictionary explicitly.

    In addition, a nested inner error is an ODL implementation, not an OData standard requirement. Using `innerexception` for the nested inner error will be toggled using a compatibility flag - `ODataLibraryCompatibility.UseLegacyODataInnerErrorSerialization`. Going forward, `innererror` will be used by default.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ErrorUtils.cs
+++ b/src/Microsoft.OData.Core/ErrorUtils.cs
@@ -33,7 +33,7 @@ namespace Microsoft.OData
         {
             Debug.Assert(error != null, "error != null");
 
-            code = error.ErrorCode ?? string.Empty;
+            code = error.Code ?? string.Empty;
             message = error.Message ?? string.Empty;
         }
 
--- a/src/Microsoft.OData.Core/ErrorUtils.cs
+++ b/src/Microsoft.OData.Core/ErrorUtils.cs
@@ -4,9 +4,9 @@
 // </copyright>
 //---------------------------------------------------------------------
 
-using System;
 using System.Diagnostics;
 using System.Xml;
+using Microsoft.OData.Json;
 using Microsoft.OData.Metadata;
 
 namespace Microsoft.OData
@@ -43,8 +43,12 @@ namespace Microsoft.OData
         /// <param name="writer">The Xml writer to write to.</param>
         /// <param name="error">The error instance to write.</param>
         /// <param name="includeDebugInformation">A flag indicating whether error details should be written (in debug mode only) or not.</param>
-        /// <param name="maxInnerErrorDepth">The maximum number of nested inner errors to allow.</param>
-        internal static void WriteXmlError(XmlWriter writer, ODataError error, bool includeDebugInformation, int maxInnerErrorDepth)
+        /// <param name="messageWriterSettings">Configuration settings of the OData writer.</param>
+        internal static void WriteXmlError(
+            XmlWriter writer,
+            ODataError error,
+            bool includeDebugInformation,
+            ODataMessageWriterSettings messageWriterSettings)
         {
             Debug.Assert(writer != null, "writer != null");
             Debug.Assert(error != null, "error != null");
@@ -53,7 +57,7 @@ namespace Microsoft.OData
             ErrorUtils.GetErrorDetails(error, out code, out message);
 
             ODataInnerError innerError = includeDebugInformation ? error.InnerError : null;
-            WriteXmlError(writer, code, message, innerError, maxInnerErrorDepth);
+            WriteXmlError(writer, code, message, innerError, messageWriterSettings);
         }
 
         /// <summary>
@@ -63,8 +67,13 @@ namespace Microsoft.OData
         /// <param name="code">The code of the error.</param>
         /// <param name="message">The message of the error.</param>
         /// <param name="innerError">Inner error details that will be included in debug mode (if present).</param>
-        /// <param name="maxInnerErrorDepth">The maximum number of nested inner errors to allow.</param>
-        private static void WriteXmlError(XmlWriter writer, string code, string message, ODataInnerError innerError, int maxInnerErrorDepth)
+        /// <param name="messageWriterSettings">Configuration settings of the OData writer.</param>
+        private static void WriteXmlError(
+            XmlWriter writer,
+            string code,
+            string message,
+            ODataInnerError innerError,
+            ODataMessageWriterSettings messageWriterSettings)
         {
             Debug.Assert(writer != null, "writer != null");
             Debug.Assert(code != null, "code != null");
@@ -81,7 +90,7 @@ namespace Microsoft.OData
 
             if (innerError != null)
             {
-                WriteXmlInnerError(writer, innerError, ODataMetadataConstants.ODataInnerErrorElementName, /* recursionDepth */ 0, maxInnerErrorDepth);
+                WriteXmlInnerError(writer, innerError, ODataMetadataConstants.ODataInnerErrorElementName, recursionDepth: 0, messageWriterSettings);
             }
 
             // </m:error>
@@ -95,11 +104,17 @@ namespace Microsoft.OData
         /// <param name="innerError">The inner error to write.</param>
         /// <param name="innerErrorElementName">The local name of the element representing the inner error.</param>
         /// <param name="recursionDepth">The number of times this method has been called recursively.</param>
-        /// <param name="maxInnerErrorDepth">The maximum number of nested inner errors to allow.</param>
-        private static void WriteXmlInnerError(XmlWriter writer, ODataInnerError innerError, string innerErrorElementName, int recursionDepth, int maxInnerErrorDepth)
+        /// <param name="messageWriterSettings">Configuration settings of the OData writer.</param>
+        private static void WriteXmlInnerError(
+            XmlWriter writer,
+            ODataInnerError innerError,
+            string innerErrorElementName,
+            int recursionDepth,
+            ODataMessageWriterSettings messageWriterSettings)
         {
             Debug.Assert(writer != null, "writer != null");
 
+            int maxInnerErrorDepth = messageWriterSettings.MessageQuotas.MaxNestingDepth;
             recursionDepth++;
             if (recursionDepth > maxInnerErrorDepth)
             {
@@ -113,30 +128,48 @@ namespace Microsoft.OData
             // <m:innererror> or <m:internalexception>
             writer.WriteStartElement(ODataMetadataConstants.ODataMetadataNamespacePrefix, innerErrorElementName, ODataMetadataConstants.ODataMetadataNamespace);
 
-            //// NOTE: we add empty elements if no information is provided for the message, error type and stack trace
-            ////       to stay compatible with Astoria.
+            ODataValue propertyValue = null;
 
             // <m:message>...</m:message>
-            string errorMessage = innerError.Message ?? String.Empty;
-            writer.WriteStartElement(ODataMetadataConstants.ODataInnerErrorMessageElementName, ODataMetadataConstants.ODataMetadataNamespace);
-            writer.WriteString(errorMessage);
-            writer.WriteEndElement();
+            if (innerError.Properties.TryGetValue(ODataMetadataConstants.ODataInnerErrorMessageElementName, out propertyValue)
+                && propertyValue is ODataPrimitiveValue innerErrorMessage)
+            {
+                writer.WriteStartElement(ODataMetadataConstants.ODataInnerErrorMessageElementName, ODataMetadataConstants.ODataMetadataNamespace);
+                writer.WriteString((string)innerErrorMessage.Value);
+                writer.WriteEndElement();
+            }
 
             // <m:type>...</m:type>
-            string errorType = innerError.TypeName ?? string.Empty;
-            writer.WriteStartElement(ODataMetadataConstants.ODataInnerErrorTypeElementName, ODataMetadataConstants.ODataMetadataNamespace);
-            writer.WriteString(errorType);
-            writer.WriteEndElement();
+            if (innerError.Properties.TryGetValue(ODataMetadataConstants.ODataInnerErrorTypeElementName, out propertyValue)
+                && propertyValue is ODataPrimitiveValue innerErrorTypeName)
+            {
+                writer.WriteStartElement(ODataMetadataConstants.ODataInnerErrorTypeElementName, ODataMetadataConstants.ODataMetadataNamespace);
+                writer.WriteString((string)innerErrorTypeName.Value);
+                writer.WriteEndElement();
+            }
 
             // <m:stacktrace>...</m:stacktrace>
-            string errorStackTrace = innerError.StackTrace ?? String.Empty;
-            writer.WriteStartElement(ODataMetadataConstants.ODataInnerErrorStackTraceElementName, ODataMetadataConstants.ODataMetadataNamespace);
-            writer.WriteString(errorStackTrace);
-            writer.WriteEndElement();
+            if (innerError.Properties.TryGetValue(ODataMetadataConstants.ODataInnerErrorTypeElementName, out propertyValue)
+                && propertyValue is ODataPrimitiveValue innerErrorStackTrace)
+            {
+                writer.WriteStartElement(ODataMetadataConstants.ODataInnerErrorStackTraceElementName, ODataMetadataConstants.ODataMetadataNamespace);
+                writer.WriteString((string)innerErrorStackTrace.Value);
+                writer.WriteEndElement();
+            }
 
             if (innerError.InnerError != null)
             {
-                WriteXmlInnerError(writer, innerError.InnerError, ODataMetadataConstants.ODataInnerErrorInnerErrorElementName, recursionDepth, maxInnerErrorDepth);
+                string nestedInnerErrorPropertyName;
+                if (messageWriterSettings.LibraryCompatibility.HasFlag(ODataLibraryCompatibility.UseLegacyODataInnerErrorSerialization))
+                {
+                    nestedInnerErrorPropertyName = JsonConstants.ODataErrorInnerErrorInnerErrorName;
+                }
+                else
+                {
+                    nestedInnerErrorPropertyName = JsonConstants.ODataErrorInnerErrorName;
+                }
+
+                WriteXmlInnerError(writer, innerError.InnerError, nestedInnerErrorPropertyName, recursionDepth, messageWriterSettings);
             }
 
             // </m:innererror> or </m:internalexception>
--- a/src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
+++ b/src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
@@ -520,7 +520,7 @@ namespace Microsoft.OData.Json
         /// The value of the TResult parameter contains a tuple comprising of:
         /// 1). A value of true if the current value is an in-stream error value; otherwise false.
         /// 2). An <see cref="ODataError"/> instance that was read from the payload.</returns>
-        internal async Task<Tuple<bool, ODataError>> StartBufferingAndTryToReadInStreamErrorPropertyValueAsync()
+        internal async Task<(bool IsReadSuccessfully, ODataError Error)> StartBufferingAndTryToReadInStreamErrorPropertyValueAsync()
         {
             this.AssertNotBuffering();
             this.AssertAsynchronous();
@@ -531,8 +531,9 @@ namespace Microsoft.OData.Json
 
             try
             {
-                Tuple<bool, ODataError> readInStreamErrorPropertyResult = await this.TryReadInStreamErrorPropertyValueAsync()
+                (bool IsReadSuccessfully, ODataError Error) readInStreamErrorPropertyResult = await this.TryReadInStreamErrorPropertyValueAsync()
                     .ConfigureAwait(false);
+
                 return readInStreamErrorPropertyResult;
             }
             finally
@@ -770,7 +771,7 @@ namespace Microsoft.OData.Json
 
                 // We only consider this to be an in-stream error if the object has a single 'error' property
                 bool errorPropertyFound = false;
-                Tuple<bool, ODataError> readInStreamErrorPropertyResult = null;
+                (bool IsReadSuccessfully, ODataError Error)? readInStreamErrorPropertyResult = null;
 
                 while (this.currentBufferedNode.NodeType == JsonNodeType.Property)
                 {
@@ -791,7 +792,7 @@ namespace Microsoft.OData.Json
 
                     readInStreamErrorPropertyResult = await this.TryReadInStreamErrorPropertyValueAsync()
                         .ConfigureAwait(false);
-                    if (!readInStreamErrorPropertyResult.Item1)
+                    if (readInStreamErrorPropertyResult?.IsReadSuccessfully == false)
                     {
                         // This means we thought we saw an in-stream error, but then
                         // we didn't see an intelligible error object, so we give up on reading the in-stream error
@@ -802,7 +803,7 @@ namespace Microsoft.OData.Json
 
                 if (errorPropertyFound)
                 {
-                    throw new ODataErrorException(readInStreamErrorPropertyResult.Item2);
+                    throw new ODataErrorException(readInStreamErrorPropertyResult?.Error);
                 }
             }
         }
@@ -1185,7 +1186,7 @@ namespace Microsoft.OData.Json
                         string message;
                         if (this.TryReadErrorStringPropertyValue(out message))
                         {
-                            innerError.Message = message;
+                            innerError.Properties.Add(JsonConstants.ODataErrorInnerErrorMessageName, new ODataPrimitiveValue(message));
                         }
                         else
                         {
@@ -1203,7 +1204,7 @@ namespace Microsoft.OData.Json
                         string typeName;
                         if (this.TryReadErrorStringPropertyValue(out typeName))
                         {
-                            innerError.TypeName = typeName;
+                            innerError.Properties.Add(JsonConstants.ODataErrorInnerErrorTypeNameName, new ODataPrimitiveValue(typeName));
                         }
                         else
                         {
@@ -1221,7 +1222,7 @@ namespace Microsoft.OData.Json
                         string stackTrace;
                         if (this.TryReadErrorStringPropertyValue(out stackTrace))
                         {
-                            innerError.StackTrace = stackTrace;
+                            innerError.Properties.Add(JsonConstants.ODataErrorInnerErrorStackTraceName, new ODataPrimitiveValue(stackTrace));
                         }
                         else
                         {
@@ -1231,6 +1232,7 @@ namespace Microsoft.OData.Json
                         break;
 
                     case JsonConstants.ODataErrorInnerErrorInnerErrorName:
+                    case JsonConstants.ODataErrorInnerErrorName:
                         if (!ODataJsonReaderUtils.ErrorPropertyNotFound(ref propertiesFoundBitmask, ODataJsonReaderUtils.ErrorPropertyBitMask.InnerError))
                         {
                             return false;
@@ -1404,7 +1406,7 @@ namespace Microsoft.OData.Json
         /// The value of the TResult parameter contains a tuple comprising of:
         /// 1). A value of true if an <see cref="ODataError"/> instance that was read; otherwise false.
         /// 2). An <see cref="ODataError"/> instance that was read from the reader or null if none could be read.</returns>
-        private async Task<Tuple<bool, ODataError>> TryReadInStreamErrorPropertyValueAsync()
+        private async Task<(bool IsReadSuccessfully, ODataError Error)> TryReadInStreamErrorPropertyValueAsync()
         {
             Debug.Assert(this.parsingInStreamError, "this.parsingInStreamError");
             this.AssertBuffering();
@@ -1413,7 +1415,7 @@ namespace Microsoft.OData.Json
             // We expect a StartObject node here
             if (this.currentBufferedNode.NodeType != JsonNodeType.StartObject)
             {
-                return Tuple.Create(false, (ODataError)null);
+                return (false, (ODataError)null);
             }
 
             // Read the StartObject node
@@ -1435,17 +1437,17 @@ namespace Microsoft.OData.Json
                             ref propertiesFoundBitmask,
                             ODataJsonReaderUtils.ErrorPropertyBitMask.Code))
                         {
-                            return Tuple.Create(false, error);
+                            return (false, error);
                         }
 
-                        Tuple<bool, string> readErrorCodePropertyResult = await this.TryReadErrorStringPropertyValueAsync()
+                        (bool IsReadSuccessfully, string PropertyValue) readErrorCodePropertyResult = await this.TryReadErrorStringPropertyValueAsync()
                             .ConfigureAwait(false);
-                        if (!readErrorCodePropertyResult.Item1)
+                        if (!readErrorCodePropertyResult.IsReadSuccessfully)
                         {
-                            return Tuple.Create(false, error);
+                            return (false, error);
                         }
 
-                        error.Code = readErrorCodePropertyResult.Item2;
+                        error.Code = readErrorCodePropertyResult.PropertyValue;
                         break;
 
                     case JsonConstants.ODataErrorMessageName:
@@ -1453,17 +1455,17 @@ namespace Microsoft.OData.Json
                             ref propertiesFoundBitmask,
                             ODataJsonReaderUtils.ErrorPropertyBitMask.Message))
                         {
-                            return Tuple.Create(false, error);
+                            return (false, error);
                         }
 
-                        Tuple<bool, string> readErrorMessagePropertyResult = await this.TryReadErrorStringPropertyValueAsync()
+                        (bool IsReadSuccessfully, string PropertyValue) readErrorMessagePropertyResult = await this.TryReadErrorStringPropertyValueAsync()
                             .ConfigureAwait(false);
-                        if (!readErrorMessagePropertyResult.Item1)
+                        if (!readErrorMessagePropertyResult.IsReadSuccessfully)
                         {
-                            return Tuple.Create(false, error);
+                            return (false, error);
                         }
 
-                        error.Message = readErrorMessagePropertyResult.Item2;
+                        error.Message = readErrorMessagePropertyResult.PropertyValue;
                         break;
 
                     case JsonConstants.ODataErrorTargetName:
@@ -1471,14 +1473,14 @@ namespace Microsoft.OData.Json
                             ref propertiesFoundBitmask,
                             ODataJsonReaderUtils.ErrorPropertyBitMask.Target))
                         {
-                            return Tuple.Create(false, error);
+                            return (false, error);
                         }
 
-                        Tuple<bool, string> readErrorTargetPropertyResult = await this.TryReadErrorStringPropertyValueAsync()
+                        (bool IsReadSuccessfully, string PropertyValue) readErrorTargetPropertyResult = await this.TryReadErrorStringPropertyValueAsync()
                             .ConfigureAwait(false);
-                        if (!readErrorTargetPropertyResult.Item1)
+                        if (!readErrorTargetPropertyResult.IsReadSuccessfully)
                         {
-                            return Tuple.Create(false, error);
+                            return (false, error);
                         }
 
                         error.Target = readErrorTargetPropertyResult.Item2;
@@ -1489,17 +1491,17 @@ namespace Microsoft.OData.Json
                                 ref propertiesFoundBitmask,
                                 ODataJsonReaderUtils.ErrorPropertyBitMask.Details))
                         {
-                            return Tuple.Create(false, error);
+                            return (false, error);
                         }
 
-                        Tuple<bool, List<ODataErrorDetail>> readErrorDetailsPropertyResult = await this.TryReadErrorDetailsPropertyValueAsync()
+                        (bool IsReadSuccessfully, List<ODataErrorDetail> ErrorDetails) readErrorDetailsPropertyResult = await this.TryReadErrorDetailsPropertyValueAsync()
                             .ConfigureAwait(false);
-                        if (!readErrorDetailsPropertyResult.Item1)
+                        if (!readErrorDetailsPropertyResult.IsReadSuccessfully)
                         {
-                            return Tuple.Create(false, error);
+                            return (false, error);
                         }
 
-                        error.Details = readErrorDetailsPropertyResult.Item2;
+                        error.Details = readErrorDetailsPropertyResult.ErrorDetails;
                         break;
 
                     case JsonConstants.ODataErrorInnerErrorName:
@@ -1507,14 +1509,14 @@ namespace Microsoft.OData.Json
                             ref propertiesFoundBitmask,
                             ODataJsonReaderUtils.ErrorPropertyBitMask.InnerError))
                         {
-                            return Tuple.Create(false, error);
+                            return (false, error);
                         }
 
                         Tuple<bool, ODataInnerError> readInnerErrorPropertyResult = await this.TryReadInnerErrorPropertyValueAsync(0 /*recursionDepth */)
                             .ConfigureAwait(false);
                         if (!readInnerErrorPropertyResult.Item1)
                         {
-                            return Tuple.Create(false, error);
+                            return (false, error);
                         }
 
                         error.InnerError = readInnerErrorPropertyResult.Item2;
@@ -1522,7 +1524,7 @@ namespace Microsoft.OData.Json
 
                     default:
                         // If we find a non-supported property we don't treat this as an error
-                        return Tuple.Create(false, error);
+                        return (false, error);
                 }
 
                 await this.ReadInternalAsync()
@@ -1538,8 +1540,7 @@ namespace Microsoft.OData.Json
                 .ConfigureAwait(false);
 
             // If we don't find any properties it is not a valid error object
-            return Tuple.Create(
-                propertiesFoundBitmask != ODataJsonReaderUtils.ErrorPropertyBitMask.None,
+            return (propertiesFoundBitmask != ODataJsonReaderUtils.ErrorPropertyBitMask.None,
                 error);
         }
 
@@ -1550,7 +1551,7 @@ namespace Microsoft.OData.Json
         /// The value of the TResult parameter contains a tuple comprising of:
         /// 1). A value of true if an <see cref="ODataErrorDetail"/> collection was read; otherwise false.
         /// 2). An <see cref="ODataErrorDetail"/> collection that was read from the reader or null if none could be read.</returns>
-        private async Task<Tuple<bool, List<ODataErrorDetail>>> TryReadErrorDetailsPropertyValueAsync()
+        private async Task<(bool IsReadSuccessfully, List<ODataErrorDetail> ErrorDetails)> TryReadErrorDetailsPropertyValueAsync()
         {
             Debug.Assert(
                 this.currentBufferedNode.NodeType == JsonNodeType.Property,
@@ -1566,7 +1567,7 @@ namespace Microsoft.OData.Json
             // We expect a StartArray node here
             if (this.currentBufferedNode.NodeType != JsonNodeType.StartArray)
             {
-                return Tuple.Create(false, (List<ODataErrorDetail>)null);
+                return (false, (List<ODataErrorDetail>)null);
             }
 
             // [
@@ -1577,22 +1578,22 @@ namespace Microsoft.OData.Json
 
             while (this.currentBufferedNode.NodeType != JsonNodeType.EndArray)
             {
-                Tuple<bool, ODataErrorDetail> readErrorDetailResult = await TryReadErrorDetailAsync()
+                (bool IsReadSuccessfully, ODataErrorDetail ErrorDetail) readErrorDetailResult = await TryReadErrorDetailAsync()
                     .ConfigureAwait(false);
 
-                if (!readErrorDetailResult.Item1)
+                if (!readErrorDetailResult.IsReadSuccessfully)
                 {
-                    return Tuple.Create(false, details);
+                    return (false, details);
                 }
 
-                details.Add(readErrorDetailResult.Item2);
+                details.Add(readErrorDetailResult.ErrorDetail);
 
                 // ] or { (next error detail object)
                 await ReadInternalAsync()
                     .ConfigureAwait(false);
             }
 
-            return Tuple.Create(true, details);
+            return (true, details);
         }
 
         /// <summary>
@@ -1602,7 +1603,7 @@ namespace Microsoft.OData.Json
         /// The value of the TResult parameter contains a tuple comprising of:
         /// 1). A value of true if an <see cref="ODataErrorDetail"/> instance was read; otherwise false.
         /// 2). An <see cref="ODataErrorDetail"/> instance that was read from the reader or null if none could be read.</returns>
-        private async Task<Tuple<bool, ODataErrorDetail>> TryReadErrorDetailAsync()
+        private async Task<(bool IsReadSuccessfully, ODataErrorDetail ErrorDetail)> TryReadErrorDetailAsync()
         {
             Debug.Assert(
                 this.currentBufferedNode.NodeType == JsonNodeType.StartObject,
@@ -1613,7 +1614,7 @@ namespace Microsoft.OData.Json
 
             if (this.currentBufferedNode.NodeType != JsonNodeType.StartObject)
             {
-                return Tuple.Create(false, (ODataErrorDetail)null);
+                return (false, (ODataErrorDetail)null);
             }
 
             // {
@@ -1635,17 +1636,17 @@ namespace Microsoft.OData.Json
                                 ref propertiesFoundBitmask,
                                 ODataJsonReaderUtils.ErrorPropertyBitMask.Code))
                         {
-                            return Tuple.Create(false, detail);
+                            return (false, detail);
                         }
 
-                        Tuple<bool, string> readErrorCodePropertyResult = await this.TryReadErrorStringPropertyValueAsync()
+                        (bool IsReadSuccessfully, string PropertyValue) readErrorCodePropertyResult = await this.TryReadErrorStringPropertyValueAsync()
                             .ConfigureAwait(false);
-                        if (!readErrorCodePropertyResult.Item1)
+                        if (!readErrorCodePropertyResult.IsReadSuccessfully)
                         {
-                            return Tuple.Create(false, detail);
+                            return (false, detail);
                         }
 
-                        detail.Code = readErrorCodePropertyResult.Item2;
+                        detail.Code = readErrorCodePropertyResult.PropertyValue;
                         break;
 
                     case JsonConstants.ODataErrorTargetName:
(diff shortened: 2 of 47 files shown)


habbes:
@gathogojr RE:
> While ODataInnerError's ToJson method skips any of the 3 string properties of ODataInnerError (i.e., message, type, and stacktrace) when the values are not provided, the OData serializers default to empty strings.

You say that the pull requests will default to the same representation as the serializers, which defaults to empty strings. But the comment in the issue you linked says that the standard requires the `code` and `message` to have non-empty values. Is that something we're validating, or we're not targeting strict adherence to the standard?

gathogojr:
> @gathogojr RE:
> 
> > While ODataInnerError's ToJson method skips any of the 3 string properties of ODataInnerError (i.e., message, type, and stacktrace) when the values are not provided, the OData serializers default to empty strings.
> 
> You say that the pull requests will default to the same representation as the serializers, which defaults to empty strings. But the comment in the issue you linked says that the standard requires the `code` and `message` to have non-empty values. Is that something we're validating, or we're not targeting strict adherence to the standard?

We're not targeting strict adherence to the standard here. The OData serializers too don't adhere to the standard with regard to the same. I don't know what informed the decision not to err when those required properties are empty. But I don't find it to be a gross deviation from the standard for the applicable scenarios.

habbes:
@gathogojr 
> In this pull request, I have made a change such that the nested ODataInnerError is named internalexception when ToJson method is called. In addition, when the nested InnerError property is null, it doesn't get added as a property. This mirrors what is produced by ODL serializers.

I'm curious about your choice to rename the nester inner error field in `ToJson` to `internalexception`. I assume using `innererror` for nested inner errors would be more natural since this is a recursive structure. I think it also more closely adheres to the expectation from the spec as mentioned by Garrett:

> The standard allows for complete flexibility within the innererror property, but ODL requires innererrors to have at least the same structure as ODataError

If a new field is used `internalexception`, that essentially makes the object a different type. I think this would make it harder to parse and represent in a strongly typed scenario, because you can't just represent the `ODataInnerError` as a recursive structure.

gathogojr:
> @gathogojr
> 
> > In this pull request, I have made a change such that the nested ODataInnerError is named internalexception when ToJson method is called. In addition, when the nested InnerError property is null, it doesn't get added as a property. This mirrors what is produced by ODL serializers.
> 
> I'm curious about your choice to rename the nester inner error field in `ToJson` to `internalexception`. I assume using `innererror` for nested inner errors would be more natural since this is a recursive structure. I think it also more closely adheres to the expectation from the spec as mentioned by Garrett:
> 
> > The standard allows for complete flexibility within the innererror property, but ODL requires innererrors to have at least the same structure as ODataError
> 
> If a new field is used `internalexception`, that essentially makes the object a different type. I think this would make it harder to parse and represent in a strongly typed scenario, because you can't just represent the `ODataInnerError` as a recursive structure.



> @gathogojr
> 
> > In this pull request, I have made a change such that the nested ODataInnerError is named internalexception when ToJson method is called. In addition, when the nested InnerError property is null, it doesn't get added as a property. This mirrors what is produced by ODL serializers.
> 
> I'm curious about your choice to rename the nester inner error field in `ToJson` to `internalexception`. I assume using `innererror` for nested inner errors would be more natural since this is a recursive structure. I think it also more closely adheres to the expectation from the spec as mentioned by Garrett:
> 
> > The standard allows for complete flexibility within the innererror property, but ODL requires innererrors to have at least the same structure as ODataError
> 
> If a new field is used `internalexception`, that essentially makes the object a different type. I think this would make it harder to parse and represent in a strongly typed scenario, because you can't just represent the `ODataInnerError` as a recursive structure.

@habbes Here's what the standard says about the
innererror:

_The value for the innererror name/value pair MUST be an object. The contents of this object are service-defined. Usually this object contains information that will help debug the service. The innererror name/value pair SHOULD only be used in development environments in order to guard against potential security concerns around information disclosure._

If I was to hazard a guess, I'd say the use of `internalexception` could have been a way of differentiating between `ODataInnerError.InnerError` from `ODataError.InnerError`.

To what depth we get when serializing the nested `InnerError` is controlled by the [`MessageWriterSettings.MessageQuotas.MaxNestingDepth`](https://github.com/OData/odata.net/blob/bbe2ac74487fb01fe14d97aeca131d0ca89d2452/src/Microsoft.OData.Core/JsonLight/ODataJsonLightSerializer.cs#L250) property. The default value for the property is 0, for the reason that "**innererror name/value pair SHOULD only be used in development environments**".

If we should decide to change the serializers to use `innererror` we could place the change behind a feature flag. I'm however not convinced it's a necessary change given that the inner error object is quite flexible. 

The type is so flexible that one could achieve to output nested `innererror` with the following - to whatever level one wants:
```
var odataError = new ODataError
{
    Code = "OEC1",
    Message = "OEM1",
    InnerError = new ODataInnerError(
        new Dictionary<string, ODataValue>
        {
            { "message", new ODataPrimitiveValue("OIET1") },
            { "type", new ODataPrimitiveValue("OIET1") },
            { "stacktrace", new ODataPrimitiveValue("OIES1") },
            { "innererror", new ODataResourceValue
                {
                    Properties = new List<ODataProperty>
                    {
                        new ODataProperty { Name = "message", Value = new ODataPrimitiveValue("OIEM2") },
                        new ODataProperty { Name = "type", Value = new ODataPrimitiveValue("OIET2") },
                        new ODataProperty { Name = "stacktrace", Value = new ODataPrimitiveValue("OIES2") },
                        new ODataProperty { Name = "innererror", Value = new ODataResourceValue
                            {
                                Properties = new List<ODataProperty>
                                {
                                    new ODataProperty { Name = "message", Value = new ODataPrimitiveValue("OIEM3") },
                                    new ODataProperty { Name = "type", Value = new ODataPrimitiveValue("OIET3") },
                                    new ODataProperty { Name = "stacktrace", Value = new ODataPrimitiveValue("OIES3") }
                                }
                            }
                        }
                    }
                }
            }
        })
};
```

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -98,32 +99,52 @@ internal static void ODataValueToString(StringBuilder sb, ODataValue value)
...
-                if (primitiveValue.FromODataValue() is string)
+                object valueAsObject = primitiveValue.FromODataValue();
+                string valueAsString;
+                if (ODataRawValueUtils.TryConvertPrimitiveToString(valueAsObject, out valueAsString))
                 {
-                    sb.Append(string.Concat("\"", JsonValueUtils.GetEscapedJsonString(value.FromODataValue()?.ToString()), "\""));
+                    if (valueAsObject is string)
+                    {
+                        valueAsString = JsonValueUtils.GetEscapedJsonString(valueAsString);
+                    }
```
habbes, line 126: If `TryConvertPrimitiveToString()` returns `true`, that does not guarantee that `valueAsObject` is `string`? Or the point here is that any primitive can be converted to a string, but only values that are originally strings need to be escaped?

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -98,32 +99,52 @@ internal static void ODataValueToString(StringBuilder sb, ODataValue value)
...
+
+                    if (valueAsObject is string || valueAsObject is byte[] || valueAsObject is DateTimeOffset || valueAsObject is Guid || valueAsObject is TimeSpan | valueAsObject is Date || valueAsObject is TimeOfDay)
+                    {
+                        sb.Append('"').Append(valueAsString).Append('"');
+                    }
+                    else
+                    {
+                        sb.Append(valueAsString);
```
habbes, line 134: The rest are numerical values and booleans?

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -98,32 +99,52 @@ internal static void ODataValueToString(StringBuilder sb, ODataValue value)
...
+                    }
                 }
-                else
-                {
-                    sb.Append(JsonValueUtils.GetEscapedJsonString(value.FromODataValue()?.ToString()));
+                else
+                {
+                    // In the unlikely event
```
habbes, line 139: What event would this be?

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -98,32 +99,52 @@ internal static void ODataValueToString(StringBuilder sb, ODataValue value)
...
+                    sb.Append('"').Append(JsonValueUtils.GetEscapedJsonString(ODataErrorStrings.ODataJsonWriter_UnsupportedValueType(value.GetType().FullName))).Append('"');
                 }
+
+                return;
             }
+
+            // Subclasses of ODataValue that are not supported in ODataInnerError.Properties dictionary
+            sb.Append('"').Append(JsonValueUtils.GetEscapedJsonString(ODataErrorStrings.ODataJsonWriter_UnsupportedValueType(value.GetType().FullName))).Append('"');
```
habbes, line 147: What was the concern with throwing an exception in this case?

gathogojr: Throwing an exception from `ToString()` is considered a very bad idea. The logic in this method is only executed when `ODataError.ToString()` and `ODataInnerError.ToJsonString()` are executed.

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
gathogojr: Yes. They don't need to be enclosed with double quotes

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -505,10 +529,6 @@ private static void ODataResourceValueToString(StringBuilder sb, ODataResourceVa
                     await jsonWriter.WriteValueAsync(detail.Target).ConfigureAwait(false);
                 }
 
-                // "message": "$search query option not supported",
-                await jsonWriter.WriteNameAsync(JsonConstants.ODataErrorMessageName).ConfigureAwait(false);
-                await jsonWriter.WriteValueAsync(detail.Message ?? string.Empty).ConfigureAwait(false);
-
                 // }
```
habbes, line 532: What's up with this commented out brace?

gathogojr: I believe that is logic that I only moved. I'll check and remove if it's serving any purpose

gathogojr: @habbes argh! This is not an isolated case. The original author added the comment to document which character is being written. Do you think we should remove them? I think we should...

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
gathogojr: @habbes I'll try and write a test(s) that attempts to get this block triggered. If I can't identify an event where this block of code would be executed, I'll make the call whether to remove it.

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
habbes: Ooh. I see. Well if the comment is still accurate, you could make that call. I didn't know it was documenting the intended output.

ElizabethOkerio:
> > @gathogojr
> > > In this pull request, I have made a change such that the nested ODataInnerError is named internalexception when ToJson method is called. In addition, when the nested InnerError property is null, it doesn't get added as a property. This mirrors what is produced by ODL serializers.
> > 
> > 
> > I'm curious about your choice to rename the nester inner error field in `ToJson` to `internalexception`. I assume using `innererror` for nested inner errors would be more natural since this is a recursive structure. I think it also more closely adheres to the expectation from the spec as mentioned by Garrett:
> > > The standard allows for complete flexibility within the innererror property, but ODL requires innererrors to have at least the same structure as ODataError
> > 
> > 
> > If a new field is used `internalexception`, that essentially makes the object a different type. I think this would make it harder to parse and represent in a strongly typed scenario, because you can't just represent the `ODataInnerError` as a recursive structure.
> 
> > @gathogojr
> > > In this pull request, I have made a change such that the nested ODataInnerError is named internalexception when ToJson method is called. In addition, when the nested InnerError property is null, it doesn't get added as a property. This mirrors what is produced by ODL serializers.
> > 
> > 
> > I'm curious about your choice to rename the nester inner error field in `ToJson` to `internalexception`. I assume using `innererror` for nested inner errors would be more natural since this is a recursive structure. I think it also more closely adheres to the expectation from the spec as mentioned by Garrett:
> > > The standard allows for complete flexibility within the innererror property, but ODL requires innererrors to have at least the same structure as ODataError
> > 
> > 
> > If a new field is used `internalexception`, that essentially makes the object a different type. I think this would make it harder to parse and represent in a strongly typed scenario, because you can't just represent the `ODataInnerError` as a recursive structure.
> 
> @habbes Here's what the standard says about the innererror:
> 
> _The value for the innererror name/value pair MUST be an object. The contents of this object are service-defined. Usually this object contains information that will help debug the service. The innererror name/value pair SHOULD only be used in development environments in order to guard against potential security concerns around information disclosure._
> 
> If I was to hazard a guess, I'd say the use of `internalexception` could have been a way of differentiating between `ODataInnerError.InnerError` from `ODataError.InnerError`.
> 
> To what depth we get when serializing the nested `InnerError` is controlled by the [`MessageWriterSettings.MessageQuotas.MaxNestingDepth`](https://github.com/OData/odata.net/blob/bbe2ac74487fb01fe14d97aeca131d0ca89d2452/src/Microsoft.OData.Core/JsonLight/ODataJsonLightSerializer.cs#L250) property. The default value for the property is 0, for the reason that "**innererror name/value pair SHOULD only be used in development environments**".
> 
> If we should decide to change the serializers to use `innererror` we could place the change behind a feature flag. I'm however not convinced it's a necessary change given that the inner error object is quite flexible.
> 
> The type is so flexible that one could achieve to output nested `innererror` with the following - to whatever level one wants:
> 
> ```cs
> var odataError = new ODataError
> {
>     Code = "OEC1",
>     Message = "OEM1",
>     InnerError = new ODataInnerError(
>         new Dictionary<string, ODataValue>
>         {
>             { "message", new ODataPrimitiveValue("OIET1") },
>             { "type", new ODataPrimitiveValue("OIET1") },
>             { "stacktrace", new ODataPrimitiveValue("OIES1") },
>             { "innererror", new ODataResourceValue
>                 {
>                     Properties = new List<ODataProperty>
>                     {
>                         new ODataProperty { Name = "message", Value = new ODataPrimitiveValue("OIEM2") },
>                         new ODataProperty { Name = "type", Value = new ODataPrimitiveValue("OIET2") },
>                         new ODataProperty { Name = "stacktrace", Value = new ODataPrimitiveValue("OIES2") },
>                         new ODataProperty { Name = "innererror", Value = new ODataResourceValue
>                             {
>                                 Properties = new List<ODataProperty>
>                                 {
>                                     new ODataProperty { Name = "message", Value = new ODataPrimitiveValue("OIEM3") },
>                                     new ODataProperty { Name = "type", Value = new ODataPrimitiveValue("OIET3") },
>                                     new ODataProperty { Name = "stacktrace", Value = new ODataPrimitiveValue("OIES3") }
>                                 }
>                             }
>                         }
>                     }
>                 }
>             }
>         })
> };
> ```
> 
> **UPDATE**: Based on a comment in _JsonContants.cs_ the reason for naming `ODataInnerError.InnerError` as `internalexception` is for compatibility with `Astoria`,
> 
> https://github.com/OData/odata.net/blob/bbe2ac74487fb01fe14d97aeca131d0ca89d2452/src/Microsoft.OData.Core/Json/JsonConstants.cs#L79
> 
> We could make the change and allow someone to toggle the legacy behave by use of a feature flag

Astoria I believe is an old thing that we should not be worrying about. I also agree using `innererror` sounds more natural.

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
ElizabethOkerio, line 126:
```suggestion
                    if (valueAsObject is string)
                    {
                        valueAsString = JsonValueUtils.GetEscapedJsonString(valueAsString);
                        sb.Append('"').Append(valueAsString).Append('"');
                    }
```

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -98,32 +99,52 @@ internal static void ODataValueToString(StringBuilder sb, ODataValue value)
...
-                    sb.Append(string.Concat("\"", JsonValueUtils.GetEscapedJsonString(value.FromODataValue()?.ToString()), "\""));
+                    if (valueAsObject is string)
+                    {
+                        valueAsString = JsonValueUtils.GetEscapedJsonString(valueAsString);
+                    }
+
+                    if (valueAsObject is string || valueAsObject is byte[] || valueAsObject is DateTimeOffset || valueAsObject is Guid || valueAsObject is TimeSpan | valueAsObject is Date || valueAsObject is TimeOfDay)
+                    {
+                        sb.Append('"').Append(valueAsString).Append('"');
+                    }
```
ElizabethOkerio, line 131:
```suggestion
                    if (valueAsObject is byte[] || valueAsObject is DateTimeOffset || valueAsObject is Guid || valueAsObject is TimeSpan | valueAsObject is Date || valueAsObject is TimeOfDay)
                    {
                        sb.Append('"').Append(valueAsString).Append('"');
                    }

```

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
```diff
@@ -348,7 +369,7 @@ internal static void ODataValueToString(StringBuilder sb, ODataValue value)
         private static void ODataCollectionValueToString(StringBuilder sb, ODataCollectionValue value)
         {
             bool isFirst = true;
-            sb.Append("[");
+            sb.Append('[');
             foreach (object item in value.Items)
```
ElizabethOkerio, line 373: why this change?

src/Microsoft.OData.Core/ODataErrorDetail.cs
```diff
@@ -8,8 +8,8 @@
 namespace Microsoft.OData
 {
     #region Namespaces
-    using System.Globalization;
     using Microsoft.OData.Json;
+    using System.Text;
```
ElizabethOkerio, line 12: sort usings

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -7,12 +7,11 @@
 namespace Microsoft.OData
 {
     #region Namespaces
+    using Microsoft.OData.Json;
     using System;
     using System.Collections.Generic;
     using System.Diagnostics;
```
ElizabethOkerio, line 13: sort usings

src/Microsoft.OData.Core/ODataError.cs
```diff
@@ -7,24 +7,23 @@
...
     #region Namespaces
-    using System;
+    using Microsoft.OData.Json;
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
-    using System.Globalization;
     using System.Linq;
-    using Microsoft.OData.Json;
+    using System.Text;
```
ElizabethOkerio, line 15: sort usings

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -21,19 +20,20 @@ namespace Microsoft.OData
...
-
-            //// NOTE: we add empty elements if no information is provided for the message, error type and stack trace
-            ////       to stay compatible with Astoria.
-            Properties.Add(JsonConstants.ODataErrorInnerErrorMessageName, new ODataNullValue());
-            Properties.Add(JsonConstants.ODataErrorInnerErrorTypeNameName, new ODataNullValue());
-            Properties.Add(JsonConstants.ODataErrorInnerErrorStackTraceName, new ODataNullValue());
+            Properties = new Dictionary<string, ODataValue>
+            {
+                //// NOTE: we add empty elements if no information is provided for the message, error type and stack trace
+                ////       to stay compatible with Astoria.
```
ElizabethOkerio, line 29: what's this Astoria. I thought it was some old thing that we moved away from?

src/Microsoft.OData.Core/ODataInnerError.cs
```diff
@@ -124,23 +128,47 @@ internal string ToJson()
                     continue;
                 }
 
-                sb.Append(",");
-                sb.Append("\"").Append(keyValuePair.Key).Append("\"").Append(":");
-                ODataJsonWriterUtils.ODataValueToString(sb, keyValuePair.Value);
+                if (!firstProperty)
```
ElizabethOkerio, line 131: What is the use of this property? How was this working without it before?

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataJsonWriterUtilsTests.cs
```diff
@@ -283,23 +283,20 @@ public void WriteError_InnerErrorWithCollectionAndNulls()
...
-                               "\"CollectionValue\":[null,\"CollectionValue\",1]" +
-                               "}", result);
+            Assert.Equal("{\"ResourceValue\":{" +
+                "\"PropertyName\":\"PropertyValue\"," +
+                "\"NullProperty\":null}," +
+                "\"NullProperty\":null," +
+                "\"CollectionValue\":[null,\"CollectionValue\",1]," +
+                "\"message\":\"\"," +
+                "\"type\":\"\"," +
+                "\"stacktrace\":\"\"}", result);
```
ElizabethOkerio, line 299: Is the order meant to change here?

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
gathogojr: Apparently, if you're adding a character to the string builder, passing it as a character gives you better performance than as a string.

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
gathogojr: @ElizabethOkerio Why did you drop, `valueAsObject is string || ` from the `if` statement? Your suggestion is not logically equivalent to the existing statement. We want to enclose the following with doubel quotes: `string`, `byte[]` as string, `DateTimeOffset` as string, `Guid` as string, `TimeSpan` as string, `Microsoft.Edm.Date` as string, `Microsoft.Edm.TimeOfDay` as string

gathogojr:
> > > @gathogojr
> > > > In this pull request, I have made a change such that the nested ODataInnerError is named internalexception when ToJson method is called. In addition, when the nested InnerError property is null, it doesn't get added as a property. This mirrors what is produced by ODL serializers.
> > > 
> > > 
> > > I'm curious about your choice to rename the nester inner error field in `ToJson` to `internalexception`. I assume using `innererror` for nested inner errors would be more natural since this is a recursive structure. I think it also more closely adheres to the expectation from the spec as mentioned by Garrett:
> > > > The standard allows for complete flexibility within the innererror property, but ODL requires innererrors to have at least the same structure as ODataError
> > > 
> > > 
> > > If a new field is used `internalexception`, that essentially makes the object a different type. I think this would make it harder to parse and represent in a strongly typed scenario, because you can't just represent the `ODataInnerError` as a recursive structure.
> > 
> > 
> > > @gathogojr
> > > > In this pull request, I have made a change such that the nested ODataInnerError is named internalexception when ToJson method is called. In addition, when the nested InnerError property is null, it doesn't get added as a property. This mirrors what is produced by ODL serializers.
> > > 
> > > 
> > > I'm curious about your choice to rename the nester inner error field in `ToJson` to `internalexception`. I assume using `innererror` for nested inner errors would be more natural since this is a recursive structure. I think it also more closely adheres to the expectation from the spec as mentioned by Garrett:
> > > > The standard allows for complete flexibility within the innererror property, but ODL requires innererrors to have at least the same structure as ODataError
> > > 
> > > 
> > > If a new field is used `internalexception`, that essentially makes the object a different type. I think this would make it harder to parse and represent in a strongly typed scenario, because you can't just represent the `ODataInnerError` as a recursive structure.
> > 
> > 
> > @habbes Here's what the standard says about the innererror:
> > _The value for the innererror name/value pair MUST be an object. The contents of this object are service-defined. Usually this object contains information that will help debug the service. The innererror name/value pair SHOULD only be used in development environments in order to guard against potential security concerns around information disclosure._
> > If I was to hazard a guess, I'd say the use of `internalexception` could have been a way of differentiating between `ODataInnerError.InnerError` from `ODataError.InnerError`.
> > To what depth we get when serializing the nested `InnerError` is controlled by the [`MessageWriterSettings.MessageQuotas.MaxNestingDepth`](https://github.com/OData/odata.net/blob/bbe2ac74487fb01fe14d97aeca131d0ca89d2452/src/Microsoft.OData.Core/JsonLight/ODataJsonLightSerializer.cs#L250) property. The default value for the property is 0, for the reason that "**innererror name/value pair SHOULD only be used in development environments**".
> > If we should decide to change the serializers to use `innererror` we could place the change behind a feature flag. I'm however not convinced it's a necessary change given that the inner error object is quite flexible.
> > The type is so flexible that one could achieve to output nested `innererror` with the following - to whatever level one wants:
> > ```cs
> > var odataError = new ODataError
> > {
> >     Code = "OEC1",
> >     Message = "OEM1",
> >     InnerError = new ODataInnerError(
> >         new Dictionary<string, ODataValue>
> >         {
> >             { "message", new ODataPrimitiveValue("OIET1") },
> >             { "type", new ODataPrimitiveValue("OIET1") },
> >             { "stacktrace", new ODataPrimitiveValue("OIES1") },
> >             { "innererror", new ODataResourceValue
> >                 {
> >                     Properties = new List<ODataProperty>
> >                     {
> >                         new ODataProperty { Name = "message", Value = new ODataPrimitiveValue("OIEM2") },
> >                         new ODataProperty { Name = "type", Value = new ODataPrimitiveValue("OIET2") },
> >                         new ODataProperty { Name = "stacktrace", Value = new ODataPrimitiveValue("OIES2") },
> >                         new ODataProperty { Name = "innererror", Value = new ODataResourceValue
> >                             {
> >                                 Properties = new List<ODataProperty>
> >                                 {
> >                                     new ODataProperty { Name = "message", Value = new ODataPrimitiveValue("OIEM3") },
> >                                     new ODataProperty { Name = "type", Value = new ODataPrimitiveValue("OIET3") },
> >                                     new ODataProperty { Name = "stacktrace", Value = new ODataPrimitiveValue("OIES3") }
> >                                 }
> >                             }
> >                         }
> >                     }
> >                 }
> >             }
> >         })
> > };
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > **UPDATE**: Based on a comment in _JsonContants.cs_ the reason for naming `ODataInnerError.InnerError` as `internalexception` is for compatibility with `Astoria`,
> > https://github.com/OData/odata.net/blob/bbe2ac74487fb01fe14d97aeca131d0ca89d2452/src/Microsoft.OData.Core/Json/JsonConstants.cs#L79
> > 
> > We could make the change and allow someone to toggle the legacy behave by use of a feature flag
> 
> Astoria I believe is an old thing that we should not be worrying about. I also agree using `innererror` sounds more natural.

@ElizabethOkerio What your suggestion means is that we change the serializers too such that they write `innererror` instead of `internalexception`. What it also means is that since we're changing what we write on the wire, we place the change behind a feature flag to make it possible to toggle the old behaviour. Are we all in agreement with that? cc @habbes

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataJsonWriterUtilsTests.cs
gathogojr: I have explained this change in the pull request description:
> When serializing an ODataErrorDetail instance, the OData serializers would write the optional target property ahead of the required message property. I switched the order such that the required message property is written first. This makes logical sense and change of property order does not qualify as a breaking change.

That's why I refactored this plus a few other tests

src/Microsoft.OData.Core/Json/ODataJsonWriterUtils.cs
ElizabethOkerio: @gathogojr I moved it to the check for string. Where you escape if `valueObject` is `string`

habbes:
@gathogojr I think using `innererror` instead of `internalexception` is better for consistency, and it would make the JSON easier to parse (if at all there's any production scenario where the JSON would be parsed). However, given the counter-arguments that you raised: it being a breaking change that would require a feature flag, it only being used in dev environments, you could assess how much would be required to make it work and advise on whether or not it's worth it.

robertmclaws:
Just chiming in here, I think there have been issues with OData architecture in the past that have done incorrect things with "Inner" objects, and I'm glad to see this getting resolved.

The serializer should error on the side of nested serialization working as expected, which means identical purposes should not have different names. The only time "Exception" should show up in any part of the error is if a .NET Exception is serialized directly (which by the way should easily be a platform-specific thing that doesn't "break the spec").

Regarding feature flags, I don't believe we should be concerned *at all* about backwards compatibility on breaking change releases. _We're breaking it for a reason._ You will not be able to use this release with older versions of the platform (and will need to make changes to the build & release process to enforce that).

The only feature flag I see being necessary in this case is leveraging any existing .NET "include exceptions" functionality to include the serialized exception in the output, as we do on Restier.

robertmclaws:
Also I want to point out in this example here:

```json
{
    "error": {
        "code": "",
        "message": "",
        "details": [
            {
                "code": "OEDC1",
                "target": "OEDT1",
                "message": "OEDM1"
            },
            {
                "code": "OEDC1",
                "message": "OEDM1"
            },
            {
                "code": "",
                "message": ""
            }
        ]
    }
}
```
The error serializer should omit anything that is empty like that last item OR validation should be added so that empty items are not added in the first place.

gathogojr:
> Just chiming in here, I think there have been issues with OData architecture in the past that have done incorrect things with "Inner" objects, and I'm glad to see this getting resolved.
> 
> The serializer should error on the side of nested serialization working as expected, which means identical purposes should not have different names. The only time "Exception" should show up in any part of the error object is if a .NET Exception is serialized directly (which by the way should easily be a platform-specific thing that doesn't "break the spec").
> 
> Regarding feature flags, I don't believe we should be concerned _at all_ about backwards compatibility on breaking change releases. _We're breaking it for a reason._ You will not be able to use this release with older versions of the platform (and will need to make changes to the build & release process to enforce that).
> 
> The only feature flag I see being necessary in this case is leveraging any existing .NET "include exceptions" functionality to include the serialized exception in the output, as we do on Restier.

ODL 8 will be an STS to help us get into a disciplined release cadence that adheres to a formal support policy.
For that reason and a few more, we purposed to not have any breaking changes in respect of what we write to the wire. To achieve this and still be able to implement requested breaking changes, we resolved that if such changes affect writing (or reading) of payloads, we'd introduce a behaviour flag to make it possible to toggle the old behaviour.

src/Microsoft.OData.Core/ODataInnerError.cs
gathogojr: @ElizabethOkerio Astoria was codename for "WCF Data Services". There could still be services that depend on the behaviour associated with Astoria

gathogojr:
I implemented the change such that we write `innererror` for nested `ODataInnerError`. However, because we decided to not introduce breaking changes in respect to writing and readings payloads in ODL 8, I introduced a feature flag `UseLegacyODataInnerErrorSerialization` that can be toggled to retain legacy behaviour.

In addition to this, I marked "Message", "Type", and "StackTrace" properties on `ODataInnerError` as deprecated: _This will be dropped in the 9.x release. The contents of inner error object should be service-defined. There are no required properties._
This will allow us to drop the properties in ODL 9 and drop the behaviour where they're always serialized irrespective of whether they are null or empty. I felt it made better sense to notify customers first.

Since the string representation of `ODataInnerError` does not fall under the category of payload on the wire, nested `ODataInnerError` will use `innererror` and the option of retaining legacy behaviour does not apply.

src/Microsoft.OData.Core/PublicAPI/net8.0/PublicAPI.Unshipped.txt
```diff
@@ -51,11 +51,16 @@ Microsoft.OData.ODataBatchOperationRequestMessage.ServiceProvider.get -> System.
...
+Microsoft.OData.ODataLibraryCompatibility.UseLegacyODataInnerErrorSerialization = 16 -> Microsoft.OData.ODataLibraryCompatibility
 Microsoft.OData.ODataLibraryCompatibility.UseLegacyVariableCasing = 1 -> Microsoft.OData.ODataLibraryCompatibility
-Microsoft.OData.ODataLibraryCompatibility.Version6 = Microsoft.OData.ODataLibraryCompatibility.UseLegacyVariableCasing | Microsoft.OData.ODataLibraryCompatibility.WriteTopLevelODataNullAnnotation | Microsoft.OData.ODataLibraryCompatibility.WriteODataContextAnnotationForNavProperty | Microsoft.OData.ODataLibraryCompatibility.DoNotThrowExceptionForTopLevelNullProperty -> Microsoft.OData.ODataLibraryCompatibility
-Microsoft.OData.ODataLibraryCompatibility.Version7 = 1 -> Microsoft.OData.ODataLibraryCompatibility
+Microsoft.OData.ODataLibraryCompatibility.Version6 = Microsoft.OData.ODataLibraryCompatibility.WriteTopLevelODataNullAnnotation | Microsoft.OData.ODataLibraryCompatibility.WriteODataContextAnnotationForNavProperty | Microsoft.OData.ODataLibraryCompatibility.DoNotThrowExceptionForTopLevelNullProperty | Microsoft.OData.ODataLibraryCompatibility.Version7 -> Microsoft.OData.ODataLibraryCompatibility
```
ElizabethOkerio, line 56: Do you need to manually update this in net 48 or it's not necessary since we dropping it.

gathogojr: That's a good question. @habbes What guidance can you provide?

gathogojr: @ElizabethOkerio My opinion is that since it's easy to populate the file if the framework is re-added, there's no reason for us to break our backs updating the file manually

src/Microsoft.OData.Core/ODataLibraryCompatibility.cs
```diff
@@ -39,14 +39,20 @@ public enum ODataLibraryCompatibility
         /// </summary>
         DoNotThrowExceptionForTopLevelNullProperty = 1 << 3,
 
+        /// <summary>
+        /// When enabled, OData inner error is serialized with "message", "type", and "stacktrace" properties irrespective of whether
+        /// they are null or empty. Also, nested inner error is serialized with property name as "internalexception".
+        /// </summary>
+        UseLegacyODataInnerErrorSerialization = 1 << 4,
```
ElizabethOkerio, line 46: Do we have a test for this new flag?

src/Microsoft.OData.Core/ODataInnerError.cs
ElizabethOkerio: We are moving away from these WCF services as they are not supported in .net 8.0

src/Microsoft.OData.Core/ODataLibraryCompatibility.cs
habbes: There's a test here: https://github.com/OData/odata.net/pull/2957/files#diff-afe9d5bc87d1806f378774f576fae48802d45003546b8c789fa6344bd554c001R96

gathogojr: @ElizabethOkerio Yes. I went to existing tests and parameterized them such that they are executed both when the flag is toggled and when it's not

src/Microsoft.OData.Core/ODataInnerError.cs
gathogojr: @ElizabethOkerio I have removed the logic to support Astoria compatibility. `message`, `type`, and `stacktrace` will no longer be added to the dictionary by default. I also dropped `Message`, `Type` and `StackTrace` properties since they were added specifically for Astoria compatibility. One can can add `message`, `type`, and `stacktrace` to the `Properties` dictionary is they want to preserve Astoria compatibility.

src/Microsoft.OData.Core/ErrorUtils.cs
```diff
@@ -43,8 +43,12 @@ internal static void GetErrorDetails(ODataError error, out string code, out stri
...
         /// <param name="error">The error instance to write.</param>
         /// <param name="includeDebugInformation">A flag indicating whether error details should be written (in debug mode only) or not.</param>
-        /// <param name="maxInnerErrorDepth">The maximum number of nested inner errors to allow.</param>
-        internal static void WriteXmlError(XmlWriter writer, ODataError error, bool includeDebugInformation, int maxInnerErrorDepth)
+        /// <param name="messageWriterSettings">Configuration settings of the OData writer.</param>
+        internal static void WriteXmlError(
+            XmlWriter writer,
+            ODataError error,
+            bool includeDebugInformation,
+            ODataMessageWriterSettings messageWriterSettings)
```
ElizabethOkerio, line 51: Is there any advantage of passing the whole `ODataMessageWriterSettings` here in place of the int? Do these methods require other properties from `ODatamessageWriterSettings` other than the max nesting depth?

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -531,9 +531,10 @@ await this.StartBufferingAsync()
 
             try
             {
-                Tuple<bool, ODataError> readInStreamErrorPropertyResult = await this.TryReadInStreamErrorPropertyValueAsync()
+                (bool IsSuccess, ODataError Error) readInStreamErrorPropertyResult = await this.TryReadInStreamErrorPropertyValueAsync()
                     .ConfigureAwait(false);
-                return readInStreamErrorPropertyResult;
+
+                return readInStreamErrorPropertyResult.ToTuple();
```
ElizabethOkerio, line 537: Is this call `ToTuple` necessary here?

src/Microsoft.OData.Core/ErrorUtils.cs
gathogojr: @ElizabethOkerio Yes. We're using the same `ODataMessageWriterSettings` object to check whether the backward compatibility flag `UseLegacyODataInnerErrorSerialization` flag has been toggled. I reasoned it's better to pass the object since it contains settings instead of adding an extra parameter. That way we don't need to change the method definition if another writer settings needs to be referenced.

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
gathogojr: I changed the `private` method to use value tuples: `(T, U)`. But the internal methods that's that is referenced by other calls returns a `Tuple<T, U>`. `ToTuple` tranforms a `ValueTuple` into a `Tuple`.

We could change the internal methods to return a `ValueType` as well but I was concerned about scope creep.

gathogojr: @ElizabethOkerio I have changed the internal methods signature to return a value tuple and removed the `.ToTuple()` method call.

ElizabethOkerio (approved):

habbes (approved):

PR #2970, opened 2024-05-18
FIx #2969 targetPath with collection nav props and add test.

### LINKED ISSUE #2969: EdmTargetPath does not get parsed correctly if it contains a collection-valued NavigationProperty
EdmTargetPaths cannot be created by parsing from paths such as the following:
`NS.Container/Singleton/CollectionNavProp1/CollectionNavProp2`

This prevents such paths being used as the target of annotations.

### Assemblies affected

OData .Net lib 7.21.1

### Reproduce steps

``` csharp
 string inputFile = args.Length > 0 ? args[0] : "";

 string csdl = File.ReadAllText(inputFile);
 IEdmModel model = CsdlReader.Parse(XElement.Parse(csdl).CreateReader());

 foreach (var annotation in model.VocabularyAnnotations)
 {
     if (annotation.Target is not IEdmTargetPath path) continue;
     var explicitSegmentCount = path.Path.Split('/').Length;
     if (explicitSegmentCount != path.Segments.Count)
     {
         Console.WriteLine($"Path '{path.Path}' has {explicitSegmentCount} segments, but {path.Segments.Count} segments are specified.");
     }
 }
```

### Expected result

Target path parsed correctly.

### Actual result

Target path's `Segments` collection stops at first collection-valued navigation property.

garethj-msft:
### Issues

*This pull request fixes #2969 *

### Description

Fixes TargetPathHelper to correctly handle collection-valued navigation properties and provides a test.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/FunctionalTests/Microsoft.OData.Edm.Tests/ExtensionMethods/TargetHelperTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Edm.Tests/ExtensionMethods/TargetHelperTests.cs
@@ -101,6 +101,22 @@ namespace Microsoft.OData.Edm.Tests.ExtensionMethods
         }
 
         [Fact]
+        public void GetTargetSegmentsForSingletonNavigationPropertyUnderComplexProperty()
+        {
+            string targetPath = "NS.Default/Me/Address/City";
+            List<IEdmElement> segments = edmModel.GetTargetSegments(targetPath.Split('/'), true).ToList();
+
+            Assert.Equal(4, segments.Count);
+            Assert.IsAssignableFrom<IEdmEntityContainer>(segments[0]);
+            Assert.IsAssignableFrom<IEdmSingleton>(segments[1]);
+            Assert.IsAssignableFrom<IEdmStructuralProperty>(segments[2]);
+            Assert.IsAssignableFrom<IEdmNavigationProperty>(segments[3]);
+            Assert.Equal("Me", (segments[1] as IEdmSingleton).Name);
+            Assert.Equal("Address", (segments[2] as IEdmStructuralProperty).Name);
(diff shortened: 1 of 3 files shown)


garethj-msft:
> @microsoft-github-policy-service agree company="Microsoft"

test/FunctionalTests/Microsoft.OData.Edm.Tests/ExtensionMethods/TargetHelperTests.cs
```diff
@@ -101,19 +101,19 @@ public void GetTargetSegmentsForEntitySetNavigationPropertyUnderDerivedComplexPr
         }
 
         [Fact]
-        public void GetTargetSegmentsForSingletonNavigationPropertyUnderComplexProperty()
```
mikepizzo, line 104: Rather than replace an existing test, can we add this as a new test?

mikepizzo (changes requested):
Unless the existing test for nav prop under a complex property is already covered, or invalid for some reason, I would prefer we add GetTargetSegmentsForSingletonNavigationPropertyUnderNavigationProperty as a new test, rather than replace the existing GetTargetSegmentsForSingletonNavigationPropertyUnderComplexProperty.

Otherwise, PR looks good.

test/FunctionalTests/Microsoft.OData.Edm.Tests/ExtensionMethods/TargetHelperTests.cs
garethj-msft: Sorry, that was just a goof - intended to add it as new.

mikepizzo (approved):
lgtm -- thanks!

KenitoInc (approved):
LGTM

PR #2980, opened 2024-05-27
Set `ODataUtf8JsonWriter` as default `JsonWriter`

### LINKED ISSUE #2822: Consider making ODataUtf8JsonWriter the default JsonWriter

habbes:
### Issues

*This pull request fixes #2822*

### Description

This PR:
- Sets `ODataUtf8JsonWriter` as the default `IJsonWriter` used in `ODataMessageWriter`.
- Sets `JavaScriptEncoder.UnsafeRelaxedJsonEscaping` as the default `JavaScriptEncoder` for `ODataUtf8JsonWriter`.
- Fixes some bugs with `ODataUtf8JsonWriter` streaming (see details at the end)
- Normalizes JSON output for some edge cases for better consistency with `JsonWriter` (see details at the end).

Both of these changes aim to improve _default_ serialization performance. `ODataUtf8JsonWriter` has demonstrated better performance and memory efficiency over the current default `JsonWriter` in benchmarks and in production workloads. One of the major blockers to it becoming the default was lack of support for streaming, but that has been addressed by #2880 

The main concern about this change is that the JSON is different between the different writers as far as character escaping is concerned.

In any case, while the output is different, both are valid JSON and semantically equivalent. Compliant JSON parsers should interpret them the same way.

Differences between `ODataUtf8JsonWriter` and `JsonWriter`

|  Category| Description |  Examples
-------|--------------|-------------
String Escaping | Utf8JsonWriter escapes more characters in a string than the default Jsonwriter. By default, Utf8JsonWriter escapes even HTML-unsafe characters like <. However, you can override that by using the UnsafeRelaxedJsonEscaping as documented here: https://learn.microsoft.com/en-us/odata/odatalib/using-ut8jsonwriter-for-better-performance#choosing-a-javascriptencoder. That said, even with this encoding, there will be some differences in character escaping, but all are valid UTF-8 encoded strings. | Utf8JsonWriter uses uppercase letters for unicode code points, JsonWriter uses lowercase letters: (`JsonWriter`: "Cust1 \ud800\udc05 \u00e4" vs `Utf8JsonWriter`: "Cust1 \uD800\uDC05 \u00E4"). Encoder differences `JsonWriter`: "CityA1 'A1' + 3 <>" vs `Utf8JsonWriter` with relaxed encoder: "CityA1 'A1' + 3 <>" vs `Utf8JsonWriter` with default encoder: "CityA1 \u0027A1\u0027 \u002B 3 \u003C\u003E")
Number formatting | `Utf8JsonWriter` serializes the decimal `1.0M` as `1.0`, but `JsonWriter` serializes it as `1` (the opposite of the previous scenario). | `JsonWriter`: `1` vs `Utf8JsonWriter`: `1.0`

We also used to have a difference in `DateTimeOffset` formatting because `Utf8JsonWriter` uses `+00:00` timezone suffix when the timezone offset is 0 (e.g. `2022-11-09T09:42:30+00:00`) whereas `JsonWriter` uses `Z`  (e.g. `2022-11-09T09:42:30Z`). However, we addressed this in a past PR because we got feedback that this one has higher chances of breaking clients (not following standards). This type of difference is also something customers have raised in the past. Now both `JsonWriter` and `ODataUtf8JsonWriter` use the `Z` suffix.

Justification for changing the default encoder to `JavaScriptEncoder.UnsafeRelaxedJsonEscaping`:

- None of the built-in encoder matches exactly any of the `ODataStringEscapeOption` used by `JsonWriter`. So there's no reason to prefer `JavaScriptEncoder.Default` based on consistency. It escapes more characters than either of the string escape options available for `JsonWriter`.
- Performance: serializing strings that need to be escaped uses more CPU and allocates more memory than strings that don't need escaping. Therefore, an encoder that is less likely to escape strings is more efficient.
- This encoder is used by default in AspNetCore, therefore this is more consistent with JSON-based APIs written in modern ASP.NET Core.
- The security concerns that make `JavaScriptEncoder.Default` (e.g. escaping HTML special characters) do not apply if we don't dump the output directly in a HTML document. Since our use case is REST APIs, it's up to the client to escape the response if they intend to render their contents in a web page.

Main changes:
- Registered `ODataUtf8JsonWriterFactory` as the default implementation of `IJsonWriterFactory` in `AddDefaultODataServices` extension method.
- I've renamed `DefaultJsonWriterFactory` to `ODataJsonWriterFactory` since it's no longer the default.
- Changed the default `JavaScriptEncoder` of `ODataUtf8JsonWriter` to `JavaScriptEncoder.UnsafeRelaxedJsonEscaping`.
- Update strings in tests that were escaped using `JavaScriptEncoder.Default` to match the output returned by the new encoder
- Replaced `DefaultJsonWriterFactory` with `ODataUtf8JsonWriterFactory` in a few tests.
- Fix bug in `ODataUtf8JsonWriter.CreateStreamValueScopeAsync`. Writing to the stream asynchronously would overwrite the content previously written by the JsonWriter and result in jumbled corrupted output. I fixed this by making sure the contents of the `Utf8JsonWriter` are committed to the buffer before starting the stream scope. This was correctly implemented and tested in the sync version, but not the async version.
- Fixed a bug in `ODataUtf8JsonWriter.TextWriter`. The `TextWriter` did not override `Write(char)`, as a result calling `Write(char)` would not actually write the char. I've overridden both `TextWriter.Write(char)` and `TextWriter.WriteAsync(char)`.
- I've added support for `float.PositiveInfinity` ( -> `"INF"`), `float.NegativityInfinity (-> `"-INF"`) and `float.NaN` (-> `"NaN"`) to `ODataUtf8JsonWriter`. Support was already there for `double`, but not `float` (an exception would be thrown when `Utf8JsonWriter` tries to write `float.NaN` for example)
- I've changed the implementation of `ODataUtf8JsonWriter.WriteValue(double)` such that it adds a `.0` when the value does not have a decimal point (e.g. `234.0` gets written as `234.0` instead of `234`). This is not required by the spec, but it's consistent with `JsonWriter` current behaviour. I spoke to @mikepizzo and he suggested it could be a left-over of OData v3. We can drop the `.0`, potentially with a feature flag. I thought it wise to keep the behaviour consistent for now to avoid disruptions since it broke a lot of our internal tests that checked for the presense of `.0` in the output.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/ODataJsonOutputContext.cs
+++ b/src/Microsoft.OData.Core/Json/ODataJsonOutputContext.cs
@@ -74,32 +74,11 @@ namespace Microsoft.OData.Json
 
                 bool isIeee754Compatible = messageInfo.MediaType.HasIeee754CompatibleSetToTrue();
 
-                IJsonWriterFactory jsonWriterFactory = this.Container?.GetService<IJsonWriterFactory>();
-
-#if NETCOREAPP
-                if (jsonWriterFactory is ODataUtf8JsonWriterFactory)
-                {
-                    // When using ODataUtf8JsonWriterFactory, we do not wrap the ouput stream in a buffered stream.
-                    // The default ODataUtf8JsonWriter will be used, and that writer handles
-                    // its own buffering (more efficiently). It leads to too many byte[] array allocations if we have
-                    // both the BufferedStream and ODataUtf8JsonWriter allocating buffers unnecessarily.
-                    this.asynchronousOutputStream = this.messageOutputStream;
-                    this.jsonWriter = CreateJsonWriter(
+                this.jsonWriter = this.CreateJsonWriter(
                         this.Container,
-                        this.messageOutputStream,
                         isIeee754Compatible,
                         messageWriterSettings,
                         messageInfo.Encoding);
-                }
-#endif
-
-                // Fallback if not using ODataUtf8JsonWriterFactory
-                if (this.jsonWriter == null)
-                {
-                    Stream outputStream = InitializeOutputStream();
-
-                    this.jsonWriter = CreateJsonWriter(this.Container, outputStream, isIeee754Compatible, messageWriterSettings, messageInfo.Encoding);
-                }
             }
             catch (Exception e)
             {
@@ -135,7 +114,8 @@ namespace Microsoft.OData.Json
             Debug.Assert(messageWriterSettings != null, "messageWriterSettings != null");
 
             bool ieee754CompatibleSetToTrue = (messageInfo.MediaType != null) ? messageInfo.MediaType.HasIeee754CompatibleSetToTrue() : false;
-            this.jsonWriter = CreateJsonWriter(messageInfo.ServiceProvider, stream, ieee754CompatibleSetToTrue, messageWriterSettings, messageInfo.Encoding);
+            this.messageOutputStream = stream;
+            this.jsonWriter = this.CreateJsonWriter(messageInfo.ServiceProvider, ieee754CompatibleSetToTrue, messageWriterSettings, messageInfo.Encoding);
 
             this.metadataLevel = new JsonMinimalMetadataLevel();
             this.propertyCacheHandler = new PropertyCacheHandler();
@@ -891,24 +871,32 @@ namespace Microsoft.OData.Json
         /// <param name="writerSettings">Configuration settings for the OData writer.</param>
         /// <param name="encoding">The character encoding to use.</param>
         /// <returns>An <see cref="IJsonWriter"/> instance.</returns>
-        private static IJsonWriter CreateJsonWriter(
+        private IJsonWriter CreateJsonWriter(
             IServiceProvider container,
-            Stream stream,
             bool isIeee754Compatible,
             ODataMessageWriterSettings writerSettings,
             Encoding encoding)
         {
             IJsonWriter jsonWriter;
+            Stream writeStream = this.messageOutputStream;
+            this.asynchronousOutputStream = this.messageOutputStream;
             if (container == null)
             {
-                TextWriter textWriter = new StreamWriter(stream, encoding);
-
-                jsonWriter = new JsonWriter(textWriter, isIeee754Compatible);
+                jsonWriter = ODataUtf8JsonWriterFactory.Default.CreateJsonWriter(writeStream, isIeee754Compatible, encoding);
             }
             else
             {
                 IJsonWriterFactory jsonWriterFactory = container.GetRequiredService<IJsonWriterFactory>();
-                jsonWriter = jsonWriterFactory.CreateJsonWriter(stream, isIeee754Compatible, encoding);
+
+                // If we're dealing with the legacy ODataJsonWriterFactory, let's wrap
+                // the stream with a BufferedStream
+                if (jsonWriterFactory is ODataJsonWriterFactory)
+                {
+
+                    writeStream = this.InitializeOutputStream();
+                }
+
+                jsonWriter = jsonWriterFactory.CreateJsonWriter(writeStream, isIeee754Compatible, encoding);
 
                 if (jsonWriter == null)
                 {
--- a/src/Microsoft.OData.Core/Json/ODataJsonOutputContext.cs
+++ b/src/Microsoft.OData.Core/Json/ODataJsonOutputContext.cs
@@ -74,11 +74,31 @@ namespace Microsoft.OData.Json
 
                 bool isIeee754Compatible = messageInfo.MediaType.HasIeee754CompatibleSetToTrue();
 
-                this.jsonWriter = this.CreateJsonWriter(
+                IJsonWriterFactory jsonWriterFactory = this.Container?.GetService<IJsonWriterFactory>();
+
+                if (jsonWriterFactory is ODataJsonWriterFactory)
+                {
+                    // When using the legacy ODataJsonWriterFactory, we wrap the stream in a buffered stream
+                    // when the async API is used.
+                    Stream outputStream = this.InitializeOutputStream();
+                    this.jsonWriter = CreateJsonWriter(this.Container, outputStream, isIeee754Compatible, messageWriterSettings, messageInfo.Encoding);
+                }
+                else
+                {
+                    // When using ODataUtf8JsonWriterFactory, we do not wrap the ouput stream in a buffered stream.
+                    // The default ODataUtf8JsonWriter will be used, and that writer handles
+                    // its own buffering (more efficiently). It leads to too many byte[] array allocations if we have
+                    // both the BufferedStream and ODataUtf8JsonWriter allocating buffers unnecessarily.
+                    this.asynchronousOutputStream = this.messageOutputStream;
+                    this.jsonWriter = CreateJsonWriter(
                         this.Container,
+                        this.messageOutputStream,
                         isIeee754Compatible,
                         messageWriterSettings,
-                        messageInfo.Encoding);
+                        messageInfo.Encoding
+                    );
+                }
+                
             }
             catch (Exception e)
             {
@@ -114,8 +134,8 @@ namespace Microsoft.OData.Json
             Debug.Assert(messageWriterSettings != null, "messageWriterSettings != null");
 
             bool ieee754CompatibleSetToTrue = (messageInfo.MediaType != null) ? messageInfo.MediaType.HasIeee754CompatibleSetToTrue() : false;
-            this.messageOutputStream = stream;
-            this.jsonWriter = this.CreateJsonWriter(messageInfo.ServiceProvider, ieee754CompatibleSetToTrue, messageWriterSettings, messageInfo.Encoding);
+
+            this.jsonWriter = CreateJsonWriter(messageInfo.ServiceProvider, stream, ieee754CompatibleSetToTrue, messageWriterSettings, messageInfo.Encoding);
 
             this.metadataLevel = new JsonMinimalMetadataLevel();
             this.propertyCacheHandler = new PropertyCacheHandler();
@@ -871,32 +891,23 @@ namespace Microsoft.OData.Json
         /// <param name="writerSettings">Configuration settings for the OData writer.</param>
         /// <param name="encoding">The character encoding to use.</param>
         /// <returns>An <see cref="IJsonWriter"/> instance.</returns>
-        private IJsonWriter CreateJsonWriter(
+        private static IJsonWriter CreateJsonWriter(
             IServiceProvider container,
+            Stream stream,
             bool isIeee754Compatible,
             ODataMessageWriterSettings writerSettings,
             Encoding encoding)
         {
             IJsonWriter jsonWriter;
-            Stream writeStream = this.messageOutputStream;
-            this.asynchronousOutputStream = this.messageOutputStream;
+
             if (container == null)
             {
-                jsonWriter = ODataUtf8JsonWriterFactory.Default.CreateJsonWriter(writeStream, isIeee754Compatible, encoding);
+                jsonWriter = new ODataUtf8JsonWriter(stream, isIeee754Compatible, encoding, leaveStreamOpen: true);
             }
             else
             {
                 IJsonWriterFactory jsonWriterFactory = container.GetRequiredService<IJsonWriterFactory>();
-
-                // If we're dealing with the legacy ODataJsonWriterFactory, let's wrap
-                // the stream with a BufferedStream
-                if (jsonWriterFactory is ODataJsonWriterFactory)
-                {
-
-                    writeStream = this.InitializeOutputStream();
-                }
-
-                jsonWriter = jsonWriterFactory.CreateJsonWriter(writeStream, isIeee754Compatible, encoding);
+                jsonWriter = jsonWriterFactory.CreateJsonWriter(stream, isIeee754Compatible, encoding);
 
                 if (jsonWriter == null)
                 {
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
@@ -60,9 +60,11 @@ namespace Microsoft.OData.Json
         /// <returns>A task representing the asynchronous operation. The task result contains a stream for writing the stream value.</returns>
         public async Task<Stream> StartStreamValueScopeAsync()
         {
-            this.WriteSeparatorIfNecessary();
+            this.CommitUtf8JsonWriterContentsToBuffer();
+
+            this.WriteItemWithSeparatorIfNeeded();
             this.bufferWriter.Write(this.DoubleQuote.Slice(0, 1).Span);
-            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+            await this.FlushAsync().ConfigureAwait(false);
 
             this.binaryValueStream = new ODataUtf8JsonWriteStream(this);
 
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
@@ -202,6 +202,24 @@ namespace Microsoft.OData.Json
             }
 
             /// <summary>
+            /// Writes the specifid character to the ODataUtf8JsonWriter.
+            /// </summary>
+            /// <param name="value">The character to write.</param>
+            public override void Write(char value)
+            {
+                if (!this.jsonWriter.isWritingJson)
+                {
+                    ReadOnlySpan<char> charToWrite = stackalloc char[1] { value };
+                    this.WriteCharsInChunks(charToWrite);
+                }
+                else
+                {
+                    ReadOnlySpan<char> charToWrite = stackalloc char[1] { value };
+                    this.WriteCharsInChunksWithoutEscaping(charToWrite);
+                }
+            }
+
+            /// <summary>
             /// Writes a specified number of characters from the given character array to the ODataUtf8JsonWriter.
             /// </summary>
             /// <param name="value">The character array from which to write characters.</param>
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
@@ -140,13 +140,13 @@ namespace Microsoft.OData.Json
         internal sealed class ODataUtf8JsonTextWriter : TextWriter
         {
             private readonly ODataUtf8JsonWriter jsonWriter = null;
-            // Buffer used to store chars that were could not be encoded due to
+            // Buffer used to store chars that could not be encoded due to
             // insufficient data in the input buffer. The chars will be prepended
             // to the next chunk of input.
             private char[] buffer;
-            // Number of chars in the buffer that are left over frevious chunk.
+            // Number of chars in the buffer that are left over from the previous chunk.
             private int numOfCharsNotWrittenFromPreviousChunk = 0;
-            // This buffer is used to by Write(char) to store the char so
+            // This buffer is used by Write(char) to store the char so
             // that we can re-use our Write(char[], ...) method.
             private char[] singleCharBuffer;
 
@@ -217,19 +217,18 @@ namespace Microsoft.OData.Json
             }
 
             /// <summary>
-            /// Writes the specifid character to the ODataUtf8JsonWriter.
+            /// Writes the specified character to the ODataUtf8JsonWriter.
             /// </summary>
             /// <param name="value">The character to write.</param>
             public override void Write(char value)
             {
+                ReadOnlySpan<char> charToWrite = stackalloc char[1] { value };
                 if (!this.jsonWriter.isWritingJson)
                 {
-                    ReadOnlySpan<char> charToWrite = stackalloc char[1] { value };
                     this.WriteCharsInChunks(charToWrite);
                 }
                 else
                 {
-                    ReadOnlySpan<char> charToWrite = stackalloc char[1] { value };
                     this.WriteCharsInChunksWithoutEscaping(charToWrite);
                 }
             }
@@ -246,7 +245,6 @@ namespace Microsoft.OData.Json
 
                 if (!this.jsonWriter.isWritingJson)
                 {
-                
                     await this.WriteCharsInChunksAsync(input).ConfigureAwait(false);
                 }
                 else
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
@@ -9,12 +9,10 @@ namespace Microsoft.OData.Json
 {
     using System;
     using System.Buffers;
-    using System.Diagnostics;
     using System.IO;
     using System.Text;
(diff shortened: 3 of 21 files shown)


src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -198,9 +210,53 @@ public override async ValueTask DisposeAsync()
...
+                }
+                else
+                {
+                    ReadOnlySpan<char> charToWrite = stackalloc char[1] { value };
+                    this.WriteCharsInChunksWithoutEscaping(charToWrite);
+                }
+            }
+
+            /// <summary>
+            /// Writes the specifid character to the ODataUtf8JsonWriter.
```
ElizabethOkerio, line 240:
```suggestion
            /// Writes the specified character to the ODataUtf8JsonWriter.

```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -198,9 +210,53 @@ public override async ValueTask DisposeAsync()
...
+                if (this.singleCharBuffer != null)
+                {
+                    ArrayPool<char>.Shared.Return(this.singleCharBuffer);
+                }
+
                 await this.FlushAsync().ConfigureAwait(false);
             }
 
+            /// <summary>
+            /// Writes the specifid character to the ODataUtf8JsonWriter.
```
ElizabethOkerio, line 222:
```suggestion
            /// Writes the specified character to the ODataUtf8JsonWriter.

```

src/Microsoft.OData.Core/Json/ODataJsonOutputContext.cs
```diff
@@ -76,8 +76,14 @@ internal sealed class ODataJsonOutputContext : ODataOutputContext
 
                 IJsonWriterFactory jsonWriterFactory = this.Container?.GetService<IJsonWriterFactory>();
 
-#if NETCOREAPP
-                if (jsonWriterFactory is ODataUtf8JsonWriterFactory)
+                if (jsonWriterFactory is ODataJsonWriterFactory)
+                {
+                    // When using the legacy ODataJsonWriterFactory, we wrap the stream in a buffered stream
```
ElizabethOkerio, line 81: Why are we calling it legacy? Can a user use it if they wanted to use it or the plan is to get rid of it all together?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
```diff
@@ -60,9 +60,11 @@ public void EndStreamValueScope()
         /// <returns>A task representing the asynchronous operation. The task result contains a stream for writing the stream value.</returns>
         public async Task<Stream> StartStreamValueScopeAsync()
         {
-            this.WriteSeparatorIfNecessary();
+            this.CommitUtf8JsonWriterContentsToBuffer();
+
+            this.WriteItemWithSeparatorIfNeeded();
             this.bufferWriter.Write(this.DoubleQuote.Slice(0, 1).Span);
-            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+            await this.FlushAsync().ConfigureAwait(false);
```
ElizabethOkerio, line 67: Why the change?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -142,8 +142,15 @@ private bool CheckIfWritingJson(string currentContentType)
         internal sealed class ODataUtf8JsonTextWriter : TextWriter
         {
             private readonly ODataUtf8JsonWriter jsonWriter = null;
+            // Buffer used to store chars that were could not be encoded due to
```
ElizabethOkerio, line 145:
```suggestion
            // Buffer used to store chars that could not be encoded due to

```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -142,8 +142,15 @@ private bool CheckIfWritingJson(string currentContentType)
         internal sealed class ODataUtf8JsonTextWriter : TextWriter
         {
             private readonly ODataUtf8JsonWriter jsonWriter = null;
+            // Buffer used to store chars that were could not be encoded due to
+            // insufficient data in the input buffer. The chars will be prepended
+            // to the next chunk of input.
             private char[] buffer;
+            // Number of chars in the buffer that are left over frevious chunk.
```
ElizabethOkerio, line 149:
```suggestion
            // Number of chars in the buffer that are left over from previous chunk.

```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -142,8 +142,15 @@ private bool CheckIfWritingJson(string currentContentType)
         internal sealed class ODataUtf8JsonTextWriter : TextWriter
         {
             private readonly ODataUtf8JsonWriter jsonWriter = null;
+            // Buffer used to store chars that were could not be encoded due to
+            // insufficient data in the input buffer. The chars will be prepended
+            // to the next chunk of input.
             private char[] buffer;
+            // Number of chars in the buffer that are left over frevious chunk.
             private int numOfCharsNotWrittenFromPreviousChunk = 0;
+            // This buffer is used to by Write(char) to store the char so
```
ElizabethOkerio, line 151:
```suggestion
            // This buffer is used to Write(char) to store the char so

```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -123,7 +124,7 @@ public ODataUtf8JsonWriter(Stream outputStream, bool isIeee754Compatible, Encodi
             this.bufferFlushThreshold = 0.9f * this.bufferSize;
             this.leaveStreamOpen = leaveStreamOpen;
             this.outputEncoding = encoding;
-            this.encoder = encoder ?? JavaScriptEncoder.Default;
+            this.encoder = encoder ?? JavaScriptEncoder.UnsafeRelaxedJsonEscaping;
```
ElizabethOkerio, line 127: What are the limitations or risks of using this encoder? With SFI in mind😂.

src/Microsoft.OData.Core/Json/ODataJsonOutputContext.cs
ElizabethOkerio: If the intention is not to get rid of the default `JsonWriter`.. Are there tests to show how a user can set to use it if they wanted to use it instead of the `ODataUtf8JsonWriter` or we'll never be in that situation? If we'll never be in that situation can we get rid of the `JsonWriter` all together?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: I've mentioned this in the PR description, let me quote it here:

> Justification for changing the default encoder to JavaScriptEncoder.UnsafeRelaxedJsonEscaping:

> None of the built-in encoder matches exactly any of the ODataStringEscapeOption used by JsonWriter. So there's no reason to prefer JavaScriptEncoder.Default based on consistency. It escapes more characters than either of the string escape options available for JsonWriter.
Performance: serializing strings that need to be escaped uses more CPU and allocates more memory than strings that don't need escaping. Therefore, an encoder that is less likely to escape strings is more efficient.
This encoder is used by default in AspNetCore, therefore this is more consistent with JSON-based APIs written in modern ASP.NET Core.
The security concerns that make JavaScriptEncoder.Default (e.g. escaping HTML special characters) do not apply if we don't dump the output directly in a HTML document. Since our use case is REST APIs, it's up to the client to escape the response if they intend to render their contents in a web page.

src/Microsoft.OData.Core/Json/ODataJsonOutputContext.cs
habbes: The user can still use it by injecting the `ODataJsonWriterFactory` (formerly known as `DefaultJsonWriterFactory`) to the service container.

habbes: Here's an example of a test that shows how the user could configure message writer to use `ODataJsonWriterFactory`: https://github.com/OData/odata.net/pull/2980/files#diff-43dd0673ed97d32066331d297299cab927a8992b7f5cc934f82fecf2d1eae65cR118

habbes: I used the term `legacy` because it's the original writer, but is no longer default, it's there mostly for backwards compatibility.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
habbes: I made this change in order to be consistent with `StartStreamValueScope` (the sync version). I would say the reasoning behind forcing a flush here is to ensure that the buffer is empty before we proceed with writing to the stream, with the expectation that writing to a stream may often involve writing large chunks of data. If the next value they write is smaller than the chunk size, then we'll attempt to write it all without buffering, which might lead to our bufferWriter being resizied (i.e. extra allocations). So the forced flush is just to reduce the chances of extending the internal buffer size.

ElizabethOkerio (dismissed):

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -142,8 +142,18 @@ private bool CheckIfWritingJson(string currentContentType)
         internal sealed class ODataUtf8JsonTextWriter : TextWriter
         {
             private readonly ODataUtf8JsonWriter jsonWriter = null;
+            // Buffer used to store chars that could not be encoded due to
+
+            // insufficient data in the input buffer. The chars will be prepended
```
gathogojr, line 147: nit: Remove empty line in the middle of the statement
```suggestion
            // Buffer used to store chars that could not be encoded due to
            // insufficient data in the input buffer. The chars will be prepended
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -142,8 +142,18 @@ private bool CheckIfWritingJson(string currentContentType)
         internal sealed class ODataUtf8JsonTextWriter : TextWriter
         {
             private readonly ODataUtf8JsonWriter jsonWriter = null;
+            // Buffer used to store chars that could not be encoded due to
+
+            // insufficient data in the input buffer. The chars will be prepended
+            // to the next chunk of input.
             private char[] buffer;
+            // Number of chars in the buffer that are left over from the previous chunk.
+
```
gathogojr, line 151: nit: I believe the comment is relevant to the field below. The empty line is more likely to throw someone off
```suggestion
            // Number of chars in the buffer that are left over from the previous chunk.
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -142,8 +142,18 @@ private bool CheckIfWritingJson(string currentContentType)
...
+            // insufficient data in the input buffer. The chars will be prepended
+            // to the next chunk of input.
             private char[] buffer;
+            // Number of chars in the buffer that are left over from the previous chunk.
+
             private int numOfCharsNotWrittenFromPreviousChunk = 0;
+            // This buffer is used to Write(char) to store the char so
+
+            // that we can re-use our Write(char[], ...) method.
```
gathogojr, line 155: nit: Remove empty line in the middle of the statement
```suggestion
            // This buffer is used to Write(char) to store the char so
            // that we can re-use our Write(char[], ...) method.
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -198,9 +213,55 @@ public override async ValueTask DisposeAsync()
...
+                    ArrayPool<char>.Shared.Return(this.singleCharBuffer);
+                }
+
                 await this.FlushAsync().ConfigureAwait(false);
             }
 
+            /// <summary>
+            /// Writes the specified character to the ODataUtf8JsonWriter.
+
+            /// </summary>
```
gathogojr, line 227: nit: Remove odd empty line
```suggestion
            /// Writes the specified character to the ODataUtf8JsonWriter.
            /// </summary>
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -198,9 +213,55 @@ public override async ValueTask DisposeAsync()
...
+                {
+                    ReadOnlySpan<char> charToWrite = stackalloc char[1] { value };
+                    this.WriteCharsInChunksWithoutEscaping(charToWrite);
+                }
+            }
+
+            /// <summary>
+            /// Writes the specified character to the ODataUtf8JsonWriter.
+
+            /// </summary>
```
gathogojr, line 246: nit: Remove odd empty line
```suggestion
            /// Writes the specified character to the ODataUtf8JsonWriter.
            /// </summary>
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -198,9 +213,55 @@ public override async ValueTask DisposeAsync()
...
+            /// <param name="value">The character to write.</param>
+            public override void Write(char value)
+            {
+                if (!this.jsonWriter.isWritingJson)
+                {
+                    ReadOnlySpan<char> charToWrite = stackalloc char[1] { value };
```
gathogojr, line 233: Same statement appears 5 lines below. Move it outside the `if...else` blocks

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -198,9 +213,55 @@ public override async ValueTask DisposeAsync()
...
+            public override async Task WriteAsync(char value)
+            {
+                this.singleCharBuffer ??= ArrayPool<char>.Shared.Rent(1);
+                this.singleCharBuffer[0] = value;
+                ReadOnlyMemory<char> input = this.singleCharBuffer.AsMemory().Slice(0, 1);
+
+                if (!this.jsonWriter.isWritingJson)
+                {
+                
+                    await this.WriteCharsInChunksAsync(input).ConfigureAwait(false);
```
gathogojr, line 257: nit: Remove odd empty line
```suggestion
                if (!this.jsonWriter.isWritingJson)
                {                
                    await this.WriteCharsInChunksAsync(input).ConfigureAwait(false);
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -295,7 +312,18 @@ public void WriteValue(double value)
...
+                string valueToWrite = XmlConvert.ToString(value);
+
+                // We write doubles like 100 as 100.0 (with .0 decimal point).
+                // We do this to retain consistency with the legacy JsonWriter and with older
+                // versions. However, it's not a hard requirement. Clients should not rely on the .0
+                // to decide whether a value is an integer or double. We should consider
+                // dropping this in the future (potentially behind a compatibility flag) to avoid
+                // the cost of converting to a string and checking whether a decimal point is needed.
+                bool needsFractionalPart = valueToWrite.IndexOfAny(JsonValueUtils.DoubleIndicatingCharacters) == -1;
+                this.utf8JsonWriter.WriteRawValue(needsFractionalPart ? $"{valueToWrite}.0" : valueToWrite, skipInputValidation: false);
```
gathogojr, line 326: @habbes Why `skipInputValidation: false`?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
```diff
@@ -979,7 +1023,18 @@ public async Task WriteValueAsync(double value)
...
```
gathogojr, line 1037: @habbes Why skipInputValidation: false?

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataJsonServiceDocumentSerializerTests.cs
```diff
@@ -25,7 +25,7 @@ public void FunctionImportShouldWriteOutNameAndUrlCorrectly()
                 {
                     FunctionImports = new ODataFunctionImportInfo[] {new ODataFunctionImportInfo() {Name = "name1", Url = new Uri("http://Service/FunctionImport")}}
                 },
-                @"﻿{""value"":[{""name"":""name1"",""kind"":""FunctionImport"",""url"":""http://service/FunctionImport""}]}");
+                @"{""value"":[{""name"":""name1"",""kind"":""FunctionImport"",""url"":""http://service/FunctionImport""}]}");
```
gathogojr, line 28: I can't spot the changes in this file...

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataUtf8JsonWriterAsyncTests.cs
```diff
@@ -401,8 +435,7 @@ public async Task WriteODataValueAsync_WritesODataCollectionValue()
                 new object[] { Encoding.UTF8 },
                 new object[] { Encoding.Unicode },
                 new object[] { Encoding.UTF32 },
-                new object[] { Encoding.BigEndianUnicode },
-                new object[] { Encoding.ASCII }
```
gathogojr, line 405: Why remove `Encoding.ASCII`?

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
```diff
@@ -225,7 +222,7 @@ public void SupportsODataUtf8JsonWriter()
 
             IJsonWriterFactory factory = request.ServiceProvider.GetService<IJsonWriterFactory>();
             Assert.IsType<ODataUtf8JsonWriterFactory>(factory);
-            Assert.Equal("{\"@odata.context\":\"http://www.example.com/$metadata#Edm.String\",\"value\":\"This is a test \\u0438\\u044F\"}", output);
+            Assert.Equal("{\"@odata.context\":\"http://www.example.com/$metadata#Edm.String\",\"value\":\"This is a test ия\"}", output);
```
gathogojr, line 225: Does this mean that `Default` was escaping `ия` while `RelaxedUnsafe` doesn't? What's the behaviour when using `JsonWriter`?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs
habbes: Hmm good catch. Since we're writing a valid JSON string, we don't need to validate it. I don't know why I put `false`. Maybe I thought I was doing the opposite. Let me change that.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataJsonServiceDocumentSerializerTests.cs
habbes: Seems like the original version has an unprintable character between `"{`. Not sure how it got there or why it didn't throw an error.

habbes: When I restore the original version, the test fails:

image

habbes: The test passes when using `JsonWriter` so this is an interesting catch.

habbes: That character is a [byte order mark](https://en.wikipedia.org/wiki/Byte_order_mark). And I just remembered that I already had a discussion with you about this.

habbes: In this case, the BOM is inserted by the `StreamWriter`. The BOM is based on the "preamble" returned by the `Encoding` instance's `GetPreamble()` value. By default `StreamWriter` uses an instance of `UTF8Encoding` that does not have a BOM. But if you pass a different encoder to the constructor, it might result in the BOM being written. It so happens that the instance returned by `Encoding.UTF8` has a preamble and results in the BOM being inserted at the front of the stream. And that's why the original tests passed for `JsonWriter` which uses a `StreamWriter` and don't pass for `ODataUtf8JsonWriter` unless we remove the BOM.

The reason we didn't have to do this across all our tests is because a lot of our tests either don't specify a BOM or don't use the StreamWriter, e.g. the `JsonWriterTests` use a `StringWriter`.

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataUtf8JsonWriterAsyncTests.cs
habbes: Good catch. Had forgotten to mention that. The reason I removed `Encoding.ASCII` is because when we don't escape these non-ascii characters, their binary representation is outside the ASCII range, so the output text contains characters that ASCII can't represent and end up getting rendered with placeholders like `?`. Technically the writer still supports ascii so long as the input is ascii (since non-ascii characters are not automatically escaped).

I have added a new test that verifies ASCII support when `JavaScriptEncoder.Default` is used (which escapes non-ascii characters).

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
habbes: Yes. Previously the default `ODataUtf8JsonWriter` uses `JavaScriptEncoder.Default` which escapes non-ASCII characters. The new default, `JavaScriptEncoder.UnsafeRelaxed...` does not.

gathogojr: @habbes Let's keep an eye on this especially if the previous default writer was escaping such characters. Because it essentially changes what we write on the wire materially and one of the goals of ODL 8 was to try as much as possible to avoid that change the output payload. Luckily, someone can swap in the native `JsonWriter`.

gathogojr (dismissed):

test/FunctionalTests/Microsoft.OData.Core.Tests/ODataMessageWriterTests.cs
habbes: True. I'll have that explicitly mentioned in the release candidate blog post. Let's see if someone will bring that up.

gathogojr (approved):

PR #2981, opened 2024-05-28
Commit Utf8JsonWriter contents to buffer before writing directly to bufferwriter in StartStreamValueScopeAsync

ElizabethOkerio:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
@@ -60,9 +60,11 @@ namespace Microsoft.OData.Json
         /// <returns>A task representing the asynchronous operation. The task result contains a stream for writing the stream value.</returns>
         public async Task<Stream> StartStreamValueScopeAsync()
         {
-            this.WriteSeparatorIfNecessary();
+            this.CommitUtf8JsonWriterContentsToBuffer();
+            this.WriteItemWithSeparatorIfNeeded();
             this.bufferWriter.Write(this.DoubleQuote.Slice(0, 1).Span);
-            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+
+            await this.FlushAsync().ConfigureAwait(false);
 
             this.binaryValueStream = new ODataUtf8JsonWriteStream(this);
 
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
@@ -142,8 +142,15 @@ namespace Microsoft.OData.Json
         internal sealed class ODataUtf8JsonTextWriter : TextWriter
         {
             private readonly ODataUtf8JsonWriter jsonWriter = null;
+            // Buffer used to store chars that were could not be encoded due to
+            // insufficient data in the input buffer. The chars will be prepended
+            // to the next chunk of input.
             private char[] buffer;
+            // Number of chars in the buffer that are left over frevious chunk.
             private int numOfCharsNotWrittenFromPreviousChunk = 0;
+            // This buffer is used to by Write(char) to store the char so
+            // that we can re-use our Write(char[], ...) method.
+            private char[] singleCharBuffer;
 
             public ODataUtf8JsonTextWriter(ODataUtf8JsonWriter jsonWriter)
             {
@@ -183,6 +190,11 @@ namespace Microsoft.OData.Json
                     ArrayPool<char>.Shared.Return(this.buffer);
                 }
 
+                if (this.singleCharBuffer != null)
+                {
+                    ArrayPool<char>.Shared.Return(this.singleCharBuffer);
+                }
+
                 this.Flush();
                 base.Dispose(disposing);
             }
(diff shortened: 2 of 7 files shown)


habbes:
@ElizabethOkerio found other gaps with `ODataUtf8JsonWriter` that we should close. You could make the judgement where there should be done in the same PR or separate. But it would be good if the patch is shipped with all of them addressed.

We don't have an implementation of `TextWriter.Write(char)`. The default implementation is empty, so nothing is written when this method is called.

Here are some test cases that you may consider to get started:

```c#
// in JsonWriterBaseTests
[Theory]
[InlineData("application/json", 'a', "a")]
[InlineData("text/html", 'a', "\"a\"")]
[InlineData("text/plain", 'a', "\"a\"")]
// JSON special char
[InlineData("application/json", '"', "\"")]
[InlineData("text/html", '"', "\"\\\"\"")]
[InlineData("text/plain", '"', "\"\\\"\"")]
// non-ascii
[InlineData("application/json", '你', "你")]
[InlineData("text/html", '你', "\"你\"")]
[InlineData("text/plain", '你', "\"你\"")]
public void TextWriter_CorrectlyWritesSingleCharacter(string contentType, char value, string expectedOutput)
{
    using MemoryStream stream = new MemoryStream();
    IJsonWriter jsonWriter = CreateJsonWriter(stream, isIeee754Compatible: false, Encoding.UTF8);
    var tw = jsonWriter.StartTextWriterValueScope(contentType);
    tw.Write(value);
    jsonWriter.EndTextWriterValueScope();
    jsonWriter.Flush();

    stream.Seek(0, SeekOrigin.Begin);

    using StreamReader reader = new StreamReader(stream, encoding: Encoding.UTF8);
    string rawOutput = reader.ReadToEnd();
    Assert.Equal(expectedOutput, rawOutput);
}
```

For the following tests on handling of surrogate pairs, I found that `JsonWriter` and `Utf8JsonWriter` handles string escaping differently, but both outputs will be parsed to the same thing.
```c#
// ODataUtf8JsonWriterTests
[Theory]
// both the escaped and non-escaped versions are valid
// and compliant JSON parsers should be able to handle both
[InlineData("application/json", "🐂")]
[InlineData("text/html", "\"\\uD83D\\uDC02\"")]
[InlineData("text/plain", "\"\\uD83D\\uDC02\"")]
public void TextWriter_CorrectlyHandlesSurrogatePairs(string contentType, string expectedOutput)
{
    using MemoryStream stream = new MemoryStream();
    IJsonWriter jsonWriter = CreateJsonWriter(stream, isIeee754Compatible: false, Encoding.UTF8);
    var tw = jsonWriter.StartTextWriterValueScope(contentType);
    tw.Write('\ud83d');
    tw.Write('\udc02');
    jsonWriter.EndTextWriterValueScope();
    jsonWriter.Flush();

    stream.Seek(0, SeekOrigin.Begin);

    using StreamReader reader = new StreamReader(stream, encoding: Encoding.UTF8);
    string rawOutput = reader.ReadToEnd();
    Assert.Equal(expectedOutput, rawOutput);
}
```

```c#
// JsonWriterTests

        [Theory]
        [InlineData("application/json", "🐂")]
        [InlineData("text/html", "\"🐂\"")]
        [InlineData("text/plain", "\"🐂\"")]
        public void TextWriter_CorrectlyHandlesSurrogatePairs(string contentType, string expectedOutput)
        {
            using MemoryStream stream = new MemoryStream();
            IJsonWriter jsonWriter = CreateJsonWriter(stream, isIeee754Compatible: false, Encoding.UTF8);
            var tw = jsonWriter.StartTextWriterValueScope(contentType);
            tw.Write('\ud83d');
            tw.Write('\udc02');
            jsonWriter.EndTextWriterValueScope();
            jsonWriter.Flush();

            stream.Seek(0, SeekOrigin.Begin);

            using StreamReader reader = new StreamReader(stream, encoding: Encoding.UTF8);
            string rawOutput = reader.ReadToEnd();
            Assert.Equal(expectedOutput, rawOutput);
        }
```

Of course we should also have async versions of the tests.

To make implementation easier, we put the char in buffer and reuse the existing Write methods:

```c#
public override void Write(char value)
{
    if (!this.jsonWriter.isWritingJson)
    {
        ReadOnlySpan<char> charToWrite = stackalloc char[1] { value };
        this.WriteCharsInChunks(charToWrite);
    }
    else
    {
        ReadOnlySpan<char> charToWrite = stackalloc char[1] { value };
        this.WriteCharsInChunksWithoutEscaping(charToWrite);
    }
}
```

In the async version we could used a pooled buffer that's cached and released when the writer is disposed.

habbes:
@ElizabethOkerio another gap that we should close: support for `float.PositiveInfinity`, `float.NegativeInfinity` and `float.NaN`. We also ready support this for `double`, so we just need to do the same thing for float. Both sync and async.

See current handling for doubles: https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.cs#L281

And sample test cases: https://github.com/OData/odata.net/blob/main/test/FunctionalTests/Microsoft.OData.Core.Tests/Json/ODataUtf8JsonWriterTests.cs#L116-L131

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
```diff
@@ -60,7 +60,8 @@ public void EndStreamValueScope()
         /// <returns>A task representing the asynchronous operation. The task result contains a stream for writing the stream value.</returns>
         public async Task<Stream> StartStreamValueScopeAsync()
         {
-            this.WriteSeparatorIfNecessary();
+            await this.utf8JsonWriter.FlushAsync();
+            this.WriteItemWithSeparatorIfNeeded();
```
habbes, line 64: I think it's wasteful to flush twice in this method. We should use `this.CommitUtf8JsonWrtierContentsToBuffer` like we do in the sync version. We only need to make sure that the buffer writer's position is advanced passed the Utf8JsonWriter's content before we write to the buffer directly, we don't need a full-blown flush yet.

PR #2985, opened 2024-05-30
Commit Utf8JsonWriter contents to buffer before writing directly to b…

ElizabethOkerio:
### Issues

### Description

This PR fixes a bug in the ODataUtf8JsonWriter's `StartStreamValueScopeAsync` where the contents of the utf8jsonwriter are not committed to the buffer before writing stream content to the buffer directly. This leads to content previously written by the utf8jsonwriter being overwritten which leads to corrupt data. Committing utf8jsonwriter's content to buffer before streaming data to the buffer directly fixes this issue. 

The PR also has logic to  add support for float.PositiveInfinity ( -> "INF"), float.NegativityInfinity (-> "-INF") and float.NaN(->"NaN") to ODataUtf8JsonWriter. Support was already there for double, but not float(an exception would be thrown whenUtf8JsonWritertries to writefloat.NaN` for example)

The ODataUtf8JsonWriter.TextWriter did not override Write(char), This PR has changes to override both TextWriter.Write(char) and TextWriter.WriteAsync(char).

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
@@ -64,7 +64,7 @@ namespace Microsoft.OData.Json
             this.WriteItemWithSeparatorIfNeeded();
             this.bufferWriter.Write(this.DoubleQuote.Slice(0, 1).Span);
 
-            await this.FlushAsync().ConfigureAwait(false);
+            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
 
             this.binaryValueStream = new ODataUtf8JsonWriteStream(this);
 
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
@@ -60,9 +60,11 @@ namespace Microsoft.OData.Json
         /// <returns>A task representing the asynchronous operation. The task result contains a stream for writing the stream value.</returns>
         public async Task<Stream> StartStreamValueScopeAsync()
         {
-            this.WriteSeparatorIfNecessary();
+            this.CommitUtf8JsonWriterContentsToBuffer();
+            this.WriteItemWithSeparatorIfNeeded();
             this.bufferWriter.Write(this.DoubleQuote.Slice(0, 1).Span);
-            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+
+            await this.FlushAsync().ConfigureAwait(false);
 
             this.binaryValueStream = new ODataUtf8JsonWriteStream(this);
 
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
@@ -142,14 +142,20 @@ namespace Microsoft.OData.Json
         internal sealed class ODataUtf8JsonTextWriter : TextWriter
         {
             private readonly ODataUtf8JsonWriter jsonWriter = null;
-            // Buffer used to store chars that were could not be encoded due to
-            // insufficient data in the input buffer. The chars will be prepended
-            // to the next chunk of input.
+            /// <summary>
+            /// Buffer used to store characters that could not be encoded due to
+            /// insufficient data in the input buffer. These characters will be prepended
+            /// to the next chunk of input.
+            /// </summary>
             private char[] buffer;
-            // Number of chars in the buffer that are left over frevious chunk.
+            /// <summary>
+            /// Number of characters in the buffer that are left over from the previous chunk.
+            /// </summary>
             private int numOfCharsNotWrittenFromPreviousChunk = 0;
-            // This buffer is used to by Write(char) to store the char so
-            // that we can re-use our Write(char[], ...) method.
+            /// <summary>
+            /// This buffer is used by the Write(char) method to store a single character,
+            /// that we can reuse the methods for writing chars in chunks.
+            /// </summary>
             private char[] singleCharBuffer;
 
             public ODataUtf8JsonTextWriter(ODataUtf8JsonWriter jsonWriter)
@@ -219,7 +225,7 @@ namespace Microsoft.OData.Json
             }
 
             /// <summary>
-            /// Writes the specifid character to the ODataUtf8JsonWriter.
+            /// Writes the specified character to the ODataUtf8JsonWriter.
             /// </summary>
             /// <param name="value">The character to write.</param>
             public override void Write(char value)
@@ -237,7 +243,7 @@ namespace Microsoft.OData.Json
             }
 
             /// <summary>
-            /// Writes the specifid character to the ODataUtf8JsonWriter.
+            /// Asynchronously writes the specified character to the ODataUtf8JsonWriter.
             /// </summary>
             /// <param name="value">The character to write.</param>
             public override async Task WriteAsync(char value)
--- a/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
+++ b/src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
@@ -142,8 +142,15 @@ namespace Microsoft.OData.Json
         internal sealed class ODataUtf8JsonTextWriter : TextWriter
         {
             private readonly ODataUtf8JsonWriter jsonWriter = null;
+            // Buffer used to store chars that were could not be encoded due to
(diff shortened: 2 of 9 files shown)


src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
```diff
@@ -60,9 +60,11 @@ public void EndStreamValueScope()
         /// <returns>A task representing the asynchronous operation. The task result contains a stream for writing the stream value.</returns>
         public async Task<Stream> StartStreamValueScopeAsync()
         {
-            this.WriteSeparatorIfNecessary();
+            this.CommitUtf8JsonWriterContentsToBuffer();
+            this.WriteItemWithSeparatorIfNeeded();
             this.bufferWriter.Write(this.DoubleQuote.Slice(0, 1).Span);
-            await this.FlushIfBufferThresholdReachedAsync().ConfigureAwait(false);
+
+            await this.FlushAsync().ConfigureAwait(false);
```
habbes, line 67: Not a request for change, but I wonder why opted for `FlushAsync()` instead of `FlushIfBufferThresholdReachedAsync()` here.

habbes (approved):

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
ElizabethOkerio: No reason really but I think it'll be good to flush when buffer threshold is reached instead of flushing every other time

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -198,9 +210,53 @@ public override async ValueTask DisposeAsync()
...
+                if (this.singleCharBuffer != null)
+                {
+                    ArrayPool<char>.Shared.Return(this.singleCharBuffer);
+                }
+
                 await this.FlushAsync().ConfigureAwait(false);
             }
 
+            /// <summary>
+            /// Writes the specifid character to the ODataUtf8JsonWriter.
```
gathogojr, line 222:
```suggestion
            /// Writes the specified character to the ODataUtf8JsonWriter.
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
```diff
@@ -60,8 +60,10 @@ public void EndStreamValueScope()
         /// <returns>A task representing the asynchronous operation. The task result contains a stream for writing the stream value.</returns>
         public async Task<Stream> StartStreamValueScopeAsync()
         {
-            this.WriteSeparatorIfNecessary();
+            this.CommitUtf8JsonWriterContentsToBuffer();
+            this.WriteItemWithSeparatorIfNeeded();
```
gathogojr, line 64: Why the change from `WriteSeparatorIfNecessary` to `WriteItemWithSeparatorIfNeeded`? Which item would get written in this case?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -142,8 +142,15 @@ private bool CheckIfWritingJson(string currentContentType)
         internal sealed class ODataUtf8JsonTextWriter : TextWriter
         {
             private readonly ODataUtf8JsonWriter jsonWriter = null;
+            // Buffer used to store chars that were could not be encoded due to
+            // insufficient data in the input buffer. The chars will be prepended
+            // to the next chunk of input.
             private char[] buffer;
+            // Number of chars in the buffer that are left over frevious chunk.
```
gathogojr, line 149:
```suggestion
            // Number of chars in the buffer that are left over previous chunk.
```
Also are you missing "from" after "over"?

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -198,9 +210,53 @@ public override async ValueTask DisposeAsync()
...
+                }
+                else
+                {
+                    ReadOnlySpan<char> charToWrite = stackalloc char[1] { value };
+                    this.WriteCharsInChunksWithoutEscaping(charToWrite);
+                }
+            }
+
+            /// <summary>
+            /// Writes the specifid character to the ODataUtf8JsonWriter.
```
gathogojr, line 240:
```suggestion
            /// Asynchronously writes the specified character to the ODataUtf8JsonWriter.
```

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -142,8 +142,15 @@ private bool CheckIfWritingJson(string currentContentType)
         internal sealed class ODataUtf8JsonTextWriter : TextWriter
         {
             private readonly ODataUtf8JsonWriter jsonWriter = null;
+            // Buffer used to store chars that were could not be encoded due to
```
gathogojr, line 145: Since you decided to describe what the private fields do, it'd be better to use `/// <summary></summary>`. That way the intellisense feature display the description wherever the field is used

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -142,8 +142,15 @@ private bool CheckIfWritingJson(string currentContentType)
...
         {
             private readonly ODataUtf8JsonWriter jsonWriter = null;
+            // Buffer used to store chars that were could not be encoded due to
+            // insufficient data in the input buffer. The chars will be prepended
+            // to the next chunk of input.
             private char[] buffer;
+            // Number of chars in the buffer that are left over frevious chunk.
             private int numOfCharsNotWrittenFromPreviousChunk = 0;
+            // This buffer is used to by Write(char) to store the char so
+            // that we can re-use our Write(char[], ...) method.
```
gathogojr, line 152:
```suggestion
            // that we can reuse the methods for writing chars in chunks.
```

test/FunctionalTests/Microsoft.OData.Core.Tests/Json/JsonWriterAsyncTests.cs
```diff
@@ -383,6 +383,39 @@ private async Task VerifyWriterPrimitiveValueWithIeee754CompatibleAsync<T>(T par
...
         }
 
+        [Theory]
+        [InlineData(124.45, "124.45")]
+        // We write the .0 for doubles without a fractional part
+        // for consistency with the original JsonWriter implementation
+        // and with previous versions. However, it's not a hard requirement.
+        // Clients should not rely on the .0 to decide whether a value
+        // is an integer or double.
+        // In the future we can consider relaxing the need to add a .0 (possibly behind a feature flag).
```
gathogojr, line 393: Maybe we should create an issue to streamline this in ODL 9. The issue would also serve to document our intentions...

ElizabethOkerio: We didn't make this change in 7.x.We'll make it in 8.x.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.Stream.cs
ElizabethOkerio: It's still the same separator, it's just that `WriteItemWithSeparatorIfNeeded` has logic responsible for writing a separator before an item if the item is within an array and it is not the first item.

src/Microsoft.OData.Core/Json/ODataUtf8JsonWriter.TextWriter.cs
```diff
@@ -198,9 +210,53 @@ public override async ValueTask DisposeAsync()
...
+                }
+            }
+
+            /// <summary>
+            /// Writes the specifid character to the ODataUtf8JsonWriter.
+            /// </summary>
+            /// <param name="value">The character to write.</param>
+            public override async Task WriteAsync(char value)
+            {
+                this.singleCharBuffer ??= ArrayPool<char>.Shared.Rent(1);
```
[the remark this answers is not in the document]
ElizabethOkerio: This is a performance optimization thing. `stackalloc` allocates memory on the stack which is fast and automatically freed when method returns.  The synchronous `write` method completes it's work in a single, uninterrupted flow of execution. This is not the case in the asynchronous method which can yield control back to called and continue with execution later. The buffer in this case will need to exist across asynchronous waits which will require something more durable than a stack hence the use of the `ArrayPool`

gathogojr (approved):

habbes (approved):

PR #2993, opened 2024-06-06
Remove Hardcoded 'Fred' from EdmModel_Validator_Semantic_TypeMismatchRelationshipConstraint Error Message

### LINKED ISSUE #2888: Referential constraint 'Fred'
# Description

I was attempting to convert a very large CSDL, I ran validation on the parsed document and found an odd issue. It took me a bit to realize that the word 'Fred' had nothing to do with my code or the document I was parsing but has likely been a bug in the code base for about 10 years if the git blame is to be believed.

I attempted to make the fix myself but I don't have the time. It was painful getting a successful build, and I am not familiar enough with the OData spec to say what should be used instead of 'Fred'.

Here is an example set of error messages:

```
> The types of all properties in the dependent role of a referential constraint must be the same as the corresponding property types in the principal role. The type of property 'contentfileref' on entity 'Microsoft.Dynamics.CRM.webresource' does not match the type of property 'fileattachmentid' on entity 'fileattachment' in the referential constraint 'Fred'.
> Because all dependent properties of the navigation 'ContentFileRef' are nullable, the multiplicity of the principal end must be '0..1'.
> The types of all properties in the dependent role of a referential constraint must be the same as the corresponding property types in the principal role. The type of property 'contentjsonfileref' on entity 'Microsoft.Dynamics.CRM.webresource' does not match the type of property 'fileattachmentid' on entity 'fileattachment' in the referential constraint 'Fred'.
```

## Code Reference

https://github.com/OData/odata.net/blame/57572612f1ff833d395b44e6b2c498d7d0d31890/src/Microsoft.OData.Edm/Validation/ValidationRules.cs#L1614

```c#
if (!(dependentType is BadType) && !(principalType is BadType) && !dependentType.IsEquivalentTo(principalType))
{
    string errorMessage = Strings.EdmModel_Validator_Semantic_TypeMismatchRelationshipConstraint(
        navigationProperty.DependentProperties().ToList()[i].Name,
        navigationProperty.DeclaringEntityType().FullName(),
        principalProperties.ToList()[i].Name,
        principalEntityType.Name,
        "Fred"
    );

    context.AddError(navigationProperty.Location(), EdmErrorCode.TypeMismatchRelationshipConstraint, errorMessage);
}
```

## Non-Annotated CSDL

I don't know if this is helpful but this is one of the documents I am working with, it should help with test cases but be warned validating this document takes 7 minutes on my PC.
[$metadata.xml](https://gist.github.com/DesuNinshu/649b84044444d5218d5d97d54ddce88f)

WanjohiSammy:
### Issues

*This pull request fixes #2888*

### Description

To address the issue of the hardcoded 'Fred' argument in the `EdmModel_Validator_Semantic_TypeMismatchRelationshipConstraint` error message, this PR modify the error message template by removing the fourth parameter. This will prevent the display of the hardcoded value and ensure that the error message only contains relevant dynamic content. 

**Original error message template with 'Fred':**
_

> The types of all properties in the dependent role of a referential constraint must be the same as the corresponding property types in the principal role. The type of property '{0}' on entity '{1}' does not match the type of property '{2}' on entity '{3}' in the referential constraint '{4}'.

_

**Updated error message template without the hardcoded 'Fred':**

> The types of all properties in the dependent role of a referential constraint must be the same as the corresponding property types in the principal role. The type of property '{0}' on entity '{1}' does not match the type of property '{2}' on entity '{3}' in the referential constraint.

Main Changes:
- Removed 4th parameter of _`EdmModel_Validator_Semantic_TypeMismatchRelationshipConstraint`_ in _`Microsoft.OData.Edm.txt`_. Running **`Parameterized.Microsoft.OData.Edm.tt`** will generate `EdmModel_Validator_Semantic_TypeMismatchRelationshipConstraint` method with 3 parameters.
- Modified _`NavigationPropertyTypeMismatchRelationshipConstraint`_ method in **`ValidationRules.cs`** to remove '**Fred**' passed as 4th argument in **`Strings.EdmModel_Validator_Semantic_TypeMismatchRelationshipConstraint`**.
- Modified _`NavigationPropertyTypeMismatchOnReferentialConstraint`_ test case to remove '**Fred**' passed as 4th argument in **`Strings.EdmModel_Validator_Semantic_TypeMismatchRelationshipConstraint`**.

- [ ] *Test cases added*
- [x] *Test case modified*


src/Microsoft.OData.Edm/Validation/ValidationRules.cs
```diff
@@ -1611,7 +1611,7 @@ public static class ValidationRules
                               var principalType = principalProperties.ElementAtOrDefault(i).Type.Definition;
                               if (!(dependentType is BadType) && !(principalType is BadType) && !dependentType.IsEquivalentTo(principalType))
                               {
-                                  string errorMessage = Strings.EdmModel_Validator_Semantic_TypeMismatchRelationshipConstraint(navigationProperty.DependentProperties().ToList()[i].Name, navigationProperty.DeclaringEntityType().FullName(), principalProperties.ToList()[i].Name, principalEntityType.Name, "Fred");
+                                  string errorMessage = Strings.EdmModel_Validator_Semantic_TypeMismatchRelationshipConstraint(navigationProperty.DependentProperties().ToList()[i].Name, navigationProperty.DeclaringEntityType().FullName(), principalProperties.ToList()[i].Name, principalEntityType.Name);
```
habbes, line 1614: Just realized we're computing `principaleProperties.ToList()`  just to obtain the name. But on line 1611, we call `principalProperties.ElementAtOrDefault(i)`.  If `principalProperties.ElementAtOrDefault(i)` returns the same object as `principalProperties.ToLits()[i]` then we just probably re-use the object instead of computing `ToList()`:


```c#
var principalProperty = principalProperties.ElementAtOrDefault(i);
var principalType = principaleProperty.Type.Definition;

if (...)
{
   string errorMessage = ... principalProperty?.Name
}

File: src/Microsoft.OData.Edm/Validation/ValidationRules.cs [hunk 2]
```diff
@@ -1611,7 +1611,7 @@ public static class ValidationRules
                               var principalType = principalProperties.ElementAtOrDefault(i).Type.Definition;
```
habbes, line 1611: We'er calling `ElementAtOrDefault(i)` inside of a loop. Unless the collection is backed by an `IList`, `ElementAtOrDefault(i)` will traverse the collection up to `i` for each iteration of the loop. So, this loop is essentially `O(n^2)`.

I see that we call `Count()` on these collections earlier in the code as well. These could also be performing linear scans unless the collections are backed by arrays.

This method appears to be really inefficient, performing many redundant scans. 

It may be possible to refactor this entire method to perform a single scan the steps through both collections per iteration. This might be possible if we use enumerator.MoveNext() API directly.

we could also consider performing `ToList()` on both collections at the beginning and using `.Count` (property not method) and indexing against the lists directly. Though this allocates more memory than using enumerator directly.

habbes: I understand that these issues have nothing to do with this PR. We may consider them separate concerns for a different PR unless you're open to fixing the perf issues in this PR.

WanjohiSammy: Thanks @habbes for this. Please see my previous comment. Can we create a new issue to handle perf concerns for these validation rules and handle them in a separate PR.

KenitoInc (approved):

PR #2999, opened 2024-06-10
Remove Hardcoded 'Fred' from EdmModel_Validator_Semantic_TypeMismatchRelationshipConstraint Error Message #2993

### LINKED ISSUE #2888: Referential constraint 'Fred'

WanjohiSammy:
### Issues

*This pull request fixes #2888*

### Description

Cherry-pick https://github.com/OData/odata.net/commit/98155ac466843da9c86d962977439317449f7ee0 into dev-8.x branch

- [ ] *Test cases added*
- [x] *Test case modified*


WanjohiSammy:
Duplicate

PR #3002, opened 2024-06-11
E2E test framework

### LINKED ISSUE #2878: Port E2E tests to ODL 8

### Assemblies affected

### Reproduce steps

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

ElizabethOkerio:
### Issues

*This pull request fixes #2878*

### Description

This PR adds a framework that we will use to transition the E2E tests from using WCF services to create OData services to using to using `Microsoft.AspNetCore.OData`. 
One of the reasons for this is: WCF services are not supported in .net 8.0. 

I've leveraged the `Microsoft.AspNetCore.Mvc.Testing` package to create an in-memory test server to enable tests to communicate with OData endpoints: 

Some of the classes added: 
1. `TestStartupBase` - This class provides a bae class for configuring startup behavior in a .NET core or ASP.NET Core application. The methods in this class are virtual to allow developers to override them to customize the start-up behavior by for example adding some services to the pipeline etc. 
2. `TestWebApplicationFactory<TStartup>` - This class sets up a server for integration testing of ASP.NET Core applications. It derives from `WebApplicationFactory<TStartup>` a class in `Microsoft.AspNetCore.Mvc.Testing` which creates an instance of an application for testing purposes. An HttpClient is also automatically created when you derive from this class which is used to communicate with the test server. Since this server is in-memory and is not accessible from the browser for example, I've added a custom implementation of `IHttpClientFactory` that returns this instance of `HttpClient`. The `IHttpClientFactory` will set when creating a `DataServiceContext`. This will allow the requests to be made using the test server's `HttpClient`. 
3. `EndToEndTestBase<T>` - This class provides a convenient way to set up and manage resources needed for end-to-end testing, including HTTP client management and customization of data service context creation. Test will be deriving from this class. In the test classes, if users want to add custom services to the request pipeline, they can create a class that overrides the `TestStartupBase` class and use that class as the startup class to pass to `WebApplicationFactory` for example. 
4. I've added a samples folder that shows how to set-up the server folder- which will contain the ODataControllers with the OData APIs, the model classes, the data source, the EdmModel and the client folder- This will contain the generated code (This is code that is generated using OData Connected Service or code gen).  This generated code will be used to set-up the `DataServiceContext`. 

This Project has some of the common classes that will be shared by other test classes. We'll need to create another project to contain the server-client-test structure in the samples folder for all the tests to be rewritten. 

I borrowed some of the code for structuring the in-memory test server from `Microsoft.AspNetCore.OData` repo tests. 

### Checklist (Uncheck if it is not completed).
- [ ] *Test cases added*

### Additional work necessary


OData.E2E.sln
```diff
@@ -0,0 +1,1568 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
```
habbes, line 1: Has this file changed? If so, what has changed?

OData.E2E.sln
habbes, line 1: Why is this file in the root folder?

sln/OData.E2E.sln
```diff
@@ -166,6 +166,8 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Spatial.Tests", "
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.OData.PublicApi.Tests", "..\test\PublicApiTests\Microsoft.OData.PublicApi.Tests.csproj", "{D482D182-ABFA-41FF-B4E2-67ED968CEE99}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.OData.Client.E2E.TestCommon", "..\test\EndToEndTests\Common\Microsoft.OData.Client.E2E.TestCommon\Microsoft.OData.Client.E2E.TestCommon.csproj", "{73DB460C-D581-4E2F-B692-A9508CE8C9AD}"
```
habbes, line 169: Can you explain the changes made to this file? Also, I would assume we would be creating a new solution

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/Samples/Tests/SampleTest.cs
```diff
@@ -0,0 +1,45 @@
...
+// <copyright file="SampleTest.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using Microsoft.AspNetCore.OData;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.OData.Client.E2E.TestCommon.Samples.Client;
+using Microsoft.OData.Client.E2E.TestCommon.Samples.Server;
```
habbes, line 11: Do the client and server exist in the same project?

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/Samples/Tests/SampleTest.cs
```diff
@@ -0,0 +1,45 @@
...
+            : base(fixture)
+        {
+        }
+
+        [Fact]
+        public void ExecuteOverloadedOperations()
+        {
+            Uri uri = new Uri(Client.BaseAddress, "actionresult/RetrieveProduct");
+            CreateDataServiceContext = uri => new DefaultContainer(uri);
+            var context = CreateDataServiceContext(uri);
```
habbes, line 38: What's the value of wrapping the constructor in a delegate and createing the context via the delegate instead of calling the constructor directly as follows?:
```suggestion
            var context = new DefaultContainer(uri);

```

sln/OData.E2E.sln
ElizabethOkerio: When setting up the `WebApplicationFactory<T>' we need to set-up the `UseContentRoot`. This setting takes in the path to the `.sln` file. The way the project has been set-up the .sln file is in a different folder from the tests folder. This leads to exceptions being thrown. Having the .sln file in the base folder seems to solve this problem. I'm not sure whether there is a better way of doing this.

OData.E2E.sln
ElizabethOkerio: check here: https://github.com/OData/odata.net/pull/3002#discussion_r1635891401

OData.E2E.sln
ElizabethOkerio: check here: https://github.com/OData/odata.net/pull/3002#discussion_r1635891401

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/Samples/Tests/SampleTest.cs
ElizabethOkerio: The way have done it in this sample project is to have them in the same project but different folders. I don't see why they need to be in different projects.

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/Samples/Tests/SampleTest.cs
ElizabethOkerio: Yea. There is no need for the delegate here. We can do away with it.

sln/OData.E2E.sln
ElizabethOkerio: We'll create a new solution for the tests? I used the current set-up where the tests are included in the other projects solution.

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/Samples/Tests/SampleTest.cs
habbes: I think it's fine so long as we can easily tell the model classes apart, since they'll probably have the same names.

sln/OData.E2E.sln
habbes: What I mean was, a new solution that includes the new tests and the source projects that the tests reference.

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/EndToEndTestBaseOfT.cs
```diff
@@ -0,0 +1,58 @@
...
+
+        /// <summary>
+        /// Gets the HttpClientFactory.
+        /// </summary>
+        public IHttpClientFactory HttpClientFactory => _httpClientFactory;
+
+        /// <summary>
+        /// Gets or sets the delegate for creating a DataServiceContext for the service.
+        /// </summary>
+        public Func<Uri, DataServiceContext> CreateDataServiceContext { get; set; }
```
habbes, line 56: Since this delegate is not used anywhere in this class, I don't think it serves any purpose. I think this property should be removed.

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/EndToEndTestBaseOfT.cs
```diff
@@ -0,0 +1,58 @@
...
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using Microsoft.Extensions.DependencyInjection;
+using Xunit;
+
+namespace Microsoft.OData.Client.E2E.TestCommon
+{
+    public class EndToEndTestBase<T> : IClassFixture<TestWebApplicationFactory<T>> where T : class
```
habbes, line 13: What does `T` represent?

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/EndToEndTestBaseOfT.cs
```diff
@@ -0,0 +1,58 @@
...
+    {
+        private HttpClient _client;
+        private readonly IHttpClientFactory _httpClientFactory;
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="WebODataTestBase{TStartup}"/> class.
```
habbes, line 19: What is `WebODataTestBase{TStartup}`? Is that reference a relic from the past that should be removed?

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/EndToEndTestBaseOfT.cs
ElizabethOkerio: It is the custom start-up class that has custom configurations for the services you want added to the request pipeline.

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/EndToEndTestBaseOfT.cs
ElizabethOkerio: This was a type. Fixed.

sln/OData.E2E.sln
ElizabethOkerio: Added a new solution

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/Samples/Server/SampleController.cs
```diff
@@ -0,0 +1,51 @@
...
+
+namespace Microsoft.OData.Client.E2E.TestCommon.Samples.Server
+{
+    public class SampleController : ODataController
+    {
+        public SampleController()
+        {
+        }
+
+        [HttpGet("actionresult/Products")]
```
gathogojr, line 20: Is it not more intuitive to use `Sample` instead of `actionresult`?

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/Samples/Tests/SampleTest.cs
```diff
@@ -0,0 +1,43 @@
...
+        {
+        }
+
+        [Fact]
+        public void ExampleTest()
+        {
+            Uri uri = new Uri(Client.BaseAddress, "actionresult");
+            var context = new Default.Container(uri);
+            context.HttpClientFactory = HttpClientFactory;
+            var product = context?.Products.Execute().FirstOrDefault(a=>a.ProductId == -10);
```
gathogojr, line 38:
```suggestion
            var product = context.Products.Execute().FirstOrDefault(a=>a.ProductId == -10);
```
nit: `context` cannot be null

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/Samples/Server/SampleModel.cs
```diff
@@ -0,0 +1,36 @@
+﻿//-----------------------------------------------------------------------------
+// <copyright file="SampleModel.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using EfKey = System.ComponentModel.DataAnnotations.KeyAttribute;
```
gathogojr, line 8: Is there any specific reason for aliasing `KeyAttribute`? Is it just a legacy thing?

ElizabethOkerio: It's because the project references `Microsoft.OData.Client` which also has KeyAttribute with the same namespace.

habbes (approved):

gathogojr (approved):

PR #3005, opened 2024-06-18
Port ActionOverloadingTests, AnnotationTests and AsyncRequestTests

ElizabethOkerio:
### Issues

### Description

This PR ports the `ActionOverloadingTests`, `AnnotationTests` and `AsyncRequestTests` from using WCF services to use `Microsoft.AspNetCore.OData`

The `ActionOverloading` tests test action overloading. Action Overloading is where we have several actions with the same name but with different binding parameters. 

The `AnnotationTests` tests the use of custom annotations and how users can add annotations to their model classes.  `Microsoft.AspNetCore.OData` does not have a way for customers to add annotations to their model classes. There is on-going work to add support for this. These tests will be enabled once we have a way of adding these annotations when creating OData APIs. 

The `AsyncRequestsTests` test the use of the Prefer header `Respond-Async`. More info on what this `prefer-header` is here: https://docs.oasis-open.org/odata/odata/v4.01/cs01/part1-protocol/odata-v4.01-cs01-part1-protocol.html#sec_AsynchronousRequests.

The structure of these tests will be: 
We'll have folders for the various scenarios that we are testing just like is in the current set-up. For example for the above tests: We have folders named: 
1. ActionOverloadingTests 
2. AnnotationsTests
3. AsyncRequestsTests

These folders will contain 3 sub-folders: 
1. Server - This server folder will contain the controllers, EdmModels, DataSources and Data Model classes that will be used to create the OData APIs. 
2. Client - This folder will contain proxy classes for the `EdmModel` that's used by all the services in that particular folder. Using `OData Connected Service` to generate the proxy classes. It is this generated class that will be used to create an instance of the `DataServiceContext` that will be used to communicate the various OData endpoints. 
3. Tests - This folder will contain all tests classes that contain tests for the particular scenario that we're testing. 

I've added a `Common` folder in `Microsoft.OData.Client.E2E.Tests.csproj` that will have a common DataModel and Data Source for some test scenarios that might share the various Model classes or Data Sources. 

- [ ] *Test cases added*

### Additional work necessary


test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/Program.cs
```diff
@@ -0,0 +1,2 @@
+﻿// See https://aka.ms/new-console-template for more information
+Console.WriteLine("Hello, World!");
```
habbes, line 2: What is this for?

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/TestPriorityAttribute.cs
```diff
@@ -0,0 +1,18 @@
+﻿//-----------------------------------------------------------------------------
+// <copyright file="TestPriorityAttribute.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+namespace Microsoft.OData.Client.E2E.TestCommon
+{
+    public class TestPriorityAttribute : Attribute
```
habbes, line 10: What does this attribute do? Maybe add a doc comment?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,285 @@
...
+            for (int i = 0; i < 2; i++)
+            {
+                if (i == 1)
+                {
+                    _context.Format.UseJson();
+                }
+
+                // unbound action
+                _context.Execute(new Uri(_baseUri + "Default.UpdatePersonInfo", UriKind.Absolute), "POST");
+            }
```
habbes, line 84: I'm having a hard time understanding what this loop is for or what is actually being tested in this method, with no assertions.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,285 @@
...
+            _baseUri = new Uri(Client.BaseAddress, "odata/");
+            _context = new Container(_baseUri);
+            _context.HttpClientFactory = HttpClientFactory;
+        }
+
+        [Fact]
+        public void ExecuteOverloadedOperations()
+        {
+            var product = _context.Execute<Common.Clients.EndToEnd.Product>(new Uri(_baseUri + "Products(-10)", UriKind.Absolute)).Single();
+            OperationDescriptor productOperationDescriptor = _context.GetEntityDescriptor(product).OperationDescriptors.Single();
```
habbes, line 49: This returns the operations bound to this entity, is that correct?

If someone adds a new operation to the controller, this test would fail, right? Is that intentional, i.e. does this test depend on the fact that the entity type has only one bound operation?

Would it be more robust, and arguably make the intent clearer if we filtered the operations we care about?

```c#
OperationDescriptor productOperationDescriptor = _context.GetEntityDescriptor(product).OperationDescriptors.Single(op => op.Target.AbsoluteUri === "http://localhost..../Default.RetrieveProduct"); // or use a better filter.
```
Doing it this way not only makes it robust, but eliminates the need for at least one Assert.Equal call.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,285 @@
...
+            OperationDescriptor orderLineOperationDescriptor = _context.GetEntityDescriptor(orderLine).OperationDescriptors.Single();
+            Assert.Equal("http://localhost/odata/OrderLines(OrderId=-10,ProductId=-10)/Default.RetrieveProduct", orderLineOperationDescriptor.Target.AbsoluteUri, true);
+            Assert.Equal("http://localhost/odata/$metadata#Default.RetrieveProduct", orderLineOperationDescriptor.Metadata.AbsoluteUri, true);
+
+            // service operation
+            int serviceOperationResult = _context.Execute<int>(new Uri(_baseUri + "RetrieveProduct", UriKind.Absolute), "POST", true).Single();
+            Assert.Equal(-10, serviceOperationResult);
+
+            // actions
+            int productodata = _context.Execute<int>(productOperationDescriptor.Target, "POST", true).Single();
```
habbes, line 65:
```suggestion
            int productOData = _context.Execute<int>(productOperationDescriptor.Target, "POST", true).Single();

```

This variable name is strange. Even the first `product` variable is the result of an OData request. Also, the result is an ID, not a product.

```suggestion
            int resultProductId= _context.Execute<int>(productOperationDescriptor.Target, "POST", true).Single();

```

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,285 @@
...
+            Assert.Equal(-10, productodata);
+
+            int orderLineodata = _context.Execute<int>(orderLineOperationDescriptor.Target, "POST", true).Single();
```
habbes, line 68:
```suggestion
            int orderLineId= _context.Execute<int>(orderLineOperationDescriptor.Target, "POST", true).Single();

```

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,285 @@
...
+        public ActionOverloadingEndToEndTests(TestWebApplicationFactory<TestsStartup> fixture)
+            : base(fixture)
+        {
+            _baseUri = new Uri(Client.BaseAddress, "odata/");
+            _context = new Container(_baseUri);
+            _context.HttpClientFactory = HttpClientFactory;
+        }
+
+        [Fact]
+        public void ExecuteOverloadedOperations()
```
habbes, line 46: What kind of overloads does this test? Calling operations that have the same name but different binding parameters? If this is the entire scope of the test, that maybe the test name should be more specific. Since, for example, this is not testing unbound operations, operations with different parameters (other than the binding parameter), etc.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,285 @@
...
+                _context.Execute(new Uri(_baseUri + "Default.UpdatePersonInfo", UriKind.Absolute), "POST");
+            }
+        }
+
+        [Fact]
+        public void ExecuteActionImport()
+        {
+            var productId = _context.RetrieveProduct().GetValue();
+            Assert.Equal(-10, productId);
+            _context.UpdatePersonInfo();
```
habbes, line 92: Why call `_context.UpdatePersonInfo()`? And why is its effect not tested?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,285 @@
...
+        }
+
+        [Fact, TestPriority(1)]
```
habbes, line 95: What does the `TestPriority` attribute do? And given that it's defined in a custom class in this solution, is this something that Xunit is aware of?

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/TestPriorityAttribute.cs
ElizabethOkerio: It controls the order in which tests are executed.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
ElizabethOkerio: Yes. It is for determining the order in which tests are executed. A test with a high priority will be executed before a test with a lower priority. It's an xUnit construct.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
ElizabethOkerio: It tests the different formats that the `context` can use. It ensures that when `i == 1` then the `_context` has to use the `json` format. I think we only support the JSON format curretly.. Could get rid of the loop.

WanjohiSammy:
@ElizabethOkerio Please resolve the build error.

PR #3006, opened 2024-06-20
Provide async APIs for CsdlWriter and SchemaWriter

### LINKED ISSUE #2684: Provide async API for CsdlWriter
The `CsdlWriter` only exposes synchronous APIs for writing CSDL (`WriteCsdl` and `TryWriteCsdl`). This means that any async API that relies on `CsdlWriter` to serialize CSDLs (e.g. `ODataMessageWriter.WriteMetadataAsync`) will have to perform synchronous I/O, which may lead to undesired effects (sync over async which may lead to thread starvation, exceptions in environments that don't allow synchronous I/O, etc.).

### Assemblies affected

`Micrososft.OData.Edm` 7.x

### Reproduce steps

### Expected result

- We should provide APIs like `CsdlWriter.WriteCsdlAsync` and `CsdlWriter.TryWriteCsdlWriterAsync`.
- Methods like `ODataMessageWriter.WriteMetadataAsync` should be purely asynchronous and not perform any synchronous I/O.

### Actual result

- `CsdlWriter` does not expose async APIs
- `ODataMessageWriter.WriteMetadataAsync` performs synchronous I/O

### Additional detail

You may be wondering: If `ODataMessageWriter.WriteMetadataAsync` performs synchronous I/O and ASP.NET Core disables synchronous I/O by default then how come `/$metadata` request in ASP.NET Core OData do not throw exceptions? Well, ASP.NET Core OData 8 implements a hack to get around this issue. It has a [StreamWrapper](https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Formatter/ODataOutputFormatterHelper.cs#L244) that implements synchronous I/O calls like `Flush()` using blocking async calls calls like `FlushAsync().Wait()` (which is also bad for perf). We should also address this when we make `WriteMetadataAsync` truly async.

For customers who use `ODataMessageWriter` directly and don't want to perform synchronous I/O, a workaround could be to write the CSDL in a memory stream and cache the byte array, then use the byte array as the input for async I/O.

WanjohiSammy:
### Issues

*This pull request fixes #2684*

### Description

The current `CsdlWriter `class provides only synchronous methods for writing **CSDL** `(WriteCsdl and TryWriteCsdl)`. 
Also `SchemaWriter `class exposes only syncronous methods for writing **Schema** `(TryWriteSchema)`

### Main Changes

**1. Provide APIs for writing `CSDL` asynchronously:**

> - WriteCsdlAsync
> - TryWriteCsdlAsync
> - TryWriteSchemaAsync

To add these async APIs methods necessitates updates to the following files to support asynchronous operations:
```
- src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
- src/Microsoft.OData.Edm/Csdl/CsdlWriter.cs
- src/Microsoft.OData.Edm/Csdl/CsdlXmlWriter.cs
- src/Microsoft.OData.Edm/Csdl/SchemaWriter.cs
- src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaJsonWriter.cs
- src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaWriter.cs
- src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
- src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSerializationVisitor.cs
- src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelReferenceElementsJsonVisitor.cs
- src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelReferenceElementsXmlVisitor.cs
- src/Microsoft.OData.Edm/EdmModelVisitor.cs
```

These new asynchronous APIs have been documented in `PublicAPI.Unshipped.txt`.

**2. Writing tests for the new asynchronous `CsdlWriter `APIs:**

Adapting the existing tests to utilize the async/await pattern for the newly introduced asynchronous methods in the:
```
- test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/Roundtrip/ContextUrlWriterReaderTests.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.TargetPath.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.VocabularyAnnotation.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/Serialization/EdmModelCsdlSchemaWriterTests.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/Serialization/EdmModelCsdlSerializationVisitorTests.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/ScenarioTests/OasisActionsFunctionsRelationshipChangesAcceptanceTests.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/ScenarioTests/OasisRelationshipChangesAcceptanceTests.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/Vocabularies/AlternateKeysVocabularyTests.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/Vocabularies/AuthorizationVocabularyTests.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/Vocabularies/CapabilitiesVocabularyTests.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/Vocabularies/CommunityVocabularyTests.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/Vocabularies/CoreVocabularyTests.cs
- test/FunctionalTests/Microsoft.OData.Edm.Tests/Vocabularies/ValidationVocabularyTests.cs
```

- [x] *Test cases added*


src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
```diff
@@ -54,6 +55,23 @@ protected override void WriteCsdl()
             WriteCsdlEnd();
         }
 
+        /// <summary>
+        /// Write the JSON CSDL asynchronously.
+        /// </summary>
+        /// <returns></returns>
+        protected override async Task WriteCsdlAsync()
+        {
+            await WriteCsdlStartAsync();
```
habbes, line 64: We normally call [`ConfigureAwait(false)`](https://learn.microsoft.com/en-us/dotnet/api/system.threading.tasks.task.configureawait?view=net-8.0) to all the awaited async calls in our library. (I thought we had an analyzer to check this). You should do this across the async calls.
```suggestion
            await WriteCsdlStartAsync().ConfigureAwait(false);

```
Setting this to false, means the continuation of the awaited call does not have to wait to run on the same thread or context it was create from, it can continue on the next available thread. This can help improve perf and avoid deadlocks and is recommended practice inside library code.

To learn more, check this blog post: https://devblogs.microsoft.com/dotnet/configureawait-faq/

src/Microsoft.OData.Edm/Csdl/CsdlXmlWriter.cs
```diff
@@ -96,22 +127,50 @@ private void WriteEFCsdl()
...
+
         private void WriteRuntimeElement()
         {
             this.writer.WriteStartElement(CsdlConstants.Prefix_Edmx, CsdlConstants.Element_Runtime, this.edmxNamespace);
         }
 
+        private async Task WriteRuntimeElementAsync()
+        {
+            await this.writer.WriteStartElementAsync(CsdlConstants.Prefix_Edmx, CsdlConstants.Element_Runtime, this.edmxNamespace);
+        }
```
habbes, line 162: In cases where you're only performing one async call in the method, you could just "foward" the returned task instead of using `async/await` this way we can avoid the overhead of the state machine associated with `async/await`.
```suggestion
        private Task WriteRuntimeElementAsync()
        {
            return this.writer.WriteStartElementAsync(CsdlConstants.Prefix_Edmx, CsdlConstants.Element_Runtime, this.edmxNamespace);
        }

```

src/Microsoft.OData.Edm/Csdl/CsdlXmlWriter.cs
```diff
@@ -96,22 +127,50 @@ private void WriteEFCsdl()
...
+
         private void WriteConceptualModelsElement()
         {
             this.writer.WriteStartElement(CsdlConstants.Prefix_Edmx, CsdlConstants.Element_ConceptualModels, this.edmxNamespace);
         }
 
+        private async Task WriteConceptualModelsElementAsync()
+        {
+            await this.writer.WriteStartElementAsync(CsdlConstants.Prefix_Edmx, CsdlConstants.Element_ConceptualModels, this.edmxNamespace);
+        }
```
habbes, line 172:
```suggestion
        private Task WriteConceptualModelsElementAsync()
        {
            return this.writer.WriteStartElementAsync(CsdlConstants.Prefix_Edmx, CsdlConstants.Element_ConceptualModels, this.edmxNamespace);
        }

```

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
WanjohiSammy: It will save on performance, and it is usually best practice. I tried to research online if there is a default way of making ConfigureAwait false by default but that is not currently available. I will definitely add it to all of the awaits.

src/Microsoft.OData.Edm/Csdl/CsdlXmlWriter.cs
WanjohiSammy: The reason why I used `async/await` in this case is to avoid unobserved task exceptions and unexpected behavior if the caller expects the operation to be completed when the method returns.

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
WanjohiSammy: @habbes Thanks for bringing this up. I have added `ConfigureAwait(false)` to all the awaited async calls.

src/Microsoft.OData.Edm/Csdl/CsdlXmlWriter.cs
WanjohiSammy: @habbes I have resolve this. The reason why I used `async/await` in this case was to avoid unobserved task exceptions and unexpected behavior. `Non-awaited Tasks hide exceptions.`

odero: @WanjohiSammy this is a well-known performance improvement technique wrt async - it will not have result in any unexpected behaviour. Search/read about "eliding async/await "

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
```diff
@@ -289,6 +290,77 @@ public void ValidateNavigationPropertyBindingPathEndingInTypeCast(string odataVe
 
         }
 
+        [InlineData("4.0")]
+        [InlineData("4.01")]
+        [Theory]
+        public async Task ValidateNavigationPropertyBindingPathEndingInTypeCast_Async(string odataVersion)
```
ElizabethOkerio, line 296: Can we move these async tests to a new class to avoid making this class so huge. We can have a new test class: `CsdlReaderAsyncTests`  The same for the CsdlWriterTests class. `CsdlWriterAsyncTests`

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
```diff
@@ -54,6 +55,23 @@ protected override void WriteCsdl()
             WriteCsdlEnd();
         }
 
+        /// <summary>
+        /// Write the JSON CSDL asynchronously.
+        /// </summary>
+        /// <returns></returns>
```
ElizabethOkerio, line 61: Returns what? complete this doc string.

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
```diff
@@ -72,6 +90,27 @@ private void WriteCsdlStart()
             }
         }
 
+        /// <summary>
+        /// CSDL JSON Document Object asynchronously
+        /// </summary>
+        /// <returns></returns>
```
ElizabethOkerio, line 96: Same here. Complete the doc string.

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
```diff
@@ -72,6 +90,27 @@ private void WriteCsdlStart()
             }
         }
 
+        /// <summary>
+        /// CSDL JSON Document Object asynchronously
```
ElizabethOkerio, line 94:
```suggestion
        /// Writes CSDL JSON Document Object asynchronously

```

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
```diff
@@ -100,13 +150,43 @@ private void WriteSchemata()
             }
         }
 
+        /// <summary>
+        /// Schema Object asynchronously
```
ElizabethOkerio, line 154:
```suggestion
        /// writes schema objects asynchronously

```

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
```diff
@@ -100,13 +150,43 @@ private void WriteSchemata()
             }
         }
 
+        /// <summary>
+        /// Schema Object asynchronously
+        /// </summary>
+        /// <returns></returns>
```
ElizabethOkerio, line 156: Same here: Complete the doc string.

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
```diff
@@ -100,13 +150,43 @@ private void WriteSchemata()
             }
         }
 
+        /// <summary>
+        /// Schema Object asynchronously
+        /// </summary>
+        /// <returns></returns>
+        private async Task WriteSchemataAsync()
```
ElizabethOkerio, line 157:
```suggestion
        private async Task WriteSchemaAsync()

```

src/Microsoft.OData.Edm/Csdl/CsdlWriter.cs
```diff
@@ -56,6 +60,17 @@ public static bool TryWriteCsdl(IEdmModel model, Utf8JsonWriter writer, out IEnu
             return TryWriteCsdl(model, writer, CsdlJsonWriterSettings.Default, out errors);
         }
 
+        /// <summary>
+        /// Outputs a CSDL JSON artifact to the provided <see cref="Utf8JsonWriter"/> Asynchronously.
+        /// </summary>
+        /// <param name="model"></param>
+        /// <param name="writer"></param>
```
ElizabethOkerio, line 67: Complete the doc strings.

ElizabethOkerio: Do this in all other places.

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
WanjohiSammy: @ElizabethOkerio Thanks for pointing this out. I have fix this.

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
WanjohiSammy: @ElizabethOkerio This is done across the other methods.

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
WanjohiSammy: @ElizabethOkerio This is done for both `WriteSchemata `and `WriteSchemataAsync`

> WriteSchemata to WriteSchema
> WriteSchemataAsync to WriteSchemaAsync

src/Microsoft.OData.Edm/Csdl/CsdlWriter.cs
WanjohiSammy: @ElizabethOkerio I have added params description across other places as well.

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
WanjohiSammy: @ElizabethOkerio I have moved all the async methods to individual `.Async partial classes`

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
```diff
@@ -49,13 +50,30 @@ protected override void WriteCsdl()
             WriteReferenceElements();
 
             // It also MAY contain members for schemas.
-            WriteSchemata();
+            WriteSchema();
```
habbes, line 53: Why this change?

src/Microsoft.OData.Edm/Csdl/CsdlWriter.cs
```diff
@@ -132,6 +208,15 @@ protected virtual void WriteCsdl()
             // nothing here
         }
 
+        /// <summary>
+        /// Asynchronously Writes CSDL output.
+        /// </summary>
+        /// <returns>Task represents an asynchronous operation that may or may not return a result.</returns>
+        protected virtual Task WriteCsdlAsync()
```
habbes, line 215: This is an observation, not a suggestion for change, but I find this virtual `Write` methods that do nothing to be a potential source for bugs. The fact that it has a default implementation means that child classes are not required to implement it, but a `Write` method that does nothing is in my opinion unintuitive and unexpected. So, there's no practical use case of child class that does not override this method. That makes this default implementation practically useless. I think this would be better off being an abstract method than empty virtual method. That way subclasses would be forced to provide an implementation, and if they actually wanted a no-op, they would have to provide a no-op implementation explicitly. This refers to the existing `WriteCsdl` method. This would be a breaking change, but I think the impact would small because it's not common for services to create custom subclasses of this class. And since ODL 8 is major version release, it would be a good opportunity for such change.

This is not just a hypothetical problem. We have actually gotten bitten by something similar. We had a custom `TextWriter` implementation overrode `TextWriter.Writer(char[])` method but not the `TextWriter.Write(char)` overload. It just so happens that the `TextWriter.Writer(char)` default implementation is a no-op (I would expect it to call `TextWrtier.Write(new [] {char})` at least). But since the compiler didn't force us to implement this method, and we didn't have test coverage for this, we did not catch the bug when the PR was merged. Luckily the issue was caught later on and a patch made.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaWriter.cs
```diff
@@ -47,227 +48,372 @@ internal EdmModelCsdlSchemaWriter(IEdmModel model, Version edmVersion)
...
         internal abstract void WriteLabeledExpressionReferenceExpression(IEdmLabeledExpressionReferenceExpression labeledExpressionReference);
+        internal abstract Task WriteLabeledExpressionReferenceExpressionAsync(IEdmLabeledExpressionReferenceExpression labeledExpressionReference);
 
         internal abstract void WriteTimeOfDayConstantExpressionElement(IEdmTimeOfDayConstantExpression expression);
+        internal abstract Task WriteTimeOfDayConstantExpressionElementAsync(IEdmTimeOfDayConstantExpression expression);
 
         internal abstract void WriteIsTypeExpressionElementHeader(IEdmIsTypeExpression expression, bool inlineType);
+        internal abstract Task WriteIsTypeExpressionElementHeaderAsync(IEdmIsTypeExpression expression, bool inlineType);
 
         internal virtual void WriteIsOfExpressionType(IEdmIsTypeExpression expression, bool inlineType)
```
habbes, line 374: I wonder why these were left empty instead of made abstract like the rest.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
```diff
@@ -280,6 +742,27 @@ internal override void WriteBinaryTypeAttributes(IEdmBinaryTypeReference referen
...
+        /// </summary>
+        /// <param name="reference">The Edm binary type reference.</param>
+        /// <returns>Task represents an asynchronous operation.</returns>
+        internal override Task WriteBinaryTypeAttributesAsync(IEdmBinaryTypeReference reference)
+        {
+            if (reference.IsUnbounded)
+            {
+                return this.WriteRequiredAttributeAsync(CsdlConstants.Attribute_MaxLength, CsdlConstants.Value_Max, EdmValueWriter.StringAsXml);
+            }
+            else
```
habbes, line 756: The `else` is unnecessary since it would not be reached if the if conditioned passed.

habbes (approved):

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
WanjohiSammy: @habbes This was suggested in the previous comment to be renamed to `WriteSchema`

src/Microsoft.OData.Edm/Csdl/CsdlWriter.cs
WanjohiSammy: @habbes Thanks for bringing this up. **I see a lot of similar methods which are virtual with empty implementation across the project**. In this case, `base class` expects the` child classes` will `override `these methods and provide their own implementation. To mark these methods as `abstract`, we will need also to mark the `base classes` as `abstract`.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaWriter.cs
WanjohiSammy: @habbes 
Noted this as well. Probably there was a plan to provide default implementation for them.

**However, this can be clean up. What do you think?**

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaXmlWriter.cs
WanjohiSammy: I have removed the unnecessary `else`

ElizabethOkerio:
If you do `$metadata` on your endpoint will the metadata document be written asynchronously or that is not yet supported?

WanjohiSammy:
> If you do `$metadata` on your endpoint will the metadata document be written asynchronously or that is not yet supported? If not, do we plan on supporting that?

@ElizabethOkerio 

Thanks for bringing this up. However, this might be a breaking change since it will involve modify the `WriteMetadataDocumentAsync` implementation in both [ODataMetadataJsonOutputContext](https://github.com/OData/odata.net/blob/dev-8.x/src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs) and [ODataMetadataOutputContext](https://github.com/OData/odata.net/blob/dev-8.x/src/Microsoft.OData.Core/ODataMetadataOutputContext.cs)

- [ODataMetadataJsonOutputContext - Task WriteMetadataDocumentAsync()](https://github.com/OData/odata.net/blob/dev-8.x/src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs#L92)

**From** 
```cs
internal override Task WriteMetadataDocumentAsync()
{
    this.AssertAsynchronous();

    return TaskUtils.GetTaskForSynchronousOperationReturningTask(
        () =>
        {
            this.WriteMetadataDocumentImplementation();
            return this.FlushAsync();
        });
}
```
**to** 

```cs
internal override async Task WriteMetadataDocumentAsync()
{
    this.AssertAsynchronous();

    await this.WriteMetadataDocumentImplementationAsync().ConfigureAwait(false);
}
```

- [ODataMetadataOutputContext - Task WriteMetadataDocumentAsync()](https://github.com/OData/odata.net/blob/dev-8.x/src/Microsoft.OData.Core/ODataMetadataOutputContext.cs#L121)

**From** 

```cs
internal override Task WriteMetadataDocumentAsync()
{
    this.AssertAsynchronous();

    return TaskUtils.GetTaskForSynchronousOperationReturningTask(
        () =>
        {
            this.WriteMetadataDocumentImplementation();
            return this.FlushAsync();
        });
}
```

**to**

```cs
internal override async Task WriteMetadataDocumentAsync()
{
    this.AssertAsynchronous();

    await WriteMetadataDocumentImplementationAsync().ConfigureAwait(false);
}
```

WanjohiSammy:
> > If you do `$metadata` on your endpoint will the metadata document be written asynchronously or that is not yet supported? If not, do we plan on supporting that?
> 
> @ElizabethOkerio
> 
> Thanks for bringing this up. However, this might be a breaking change since it will involve modify the `WriteMetadataDocumentAsync` implementation in both [ODataMetadataJsonOutputContext](https://github.com/OData/odata.net/blob/dev-8.x/src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs) and [ODataMetadataOutputContext](https://github.com/OData/odata.net/blob/dev-8.x/src/Microsoft.OData.Core/ODataMetadataOutputContext.cs)
>

@ElizabethOkerio 

**Should I go forward and change these implementations in this PR or can be done by a different PR.**

gathogojr:
> > If you do `$metadata` on your endpoint will the metadata document be written asynchronously or that is not yet supported? If not, do we plan on supporting that?
> 
> @ElizabethOkerio
> 
> Thanks for bringing this up. However, this might be a breaking change since it will involve modify the `WriteMetadataDocumentAsync` implementation in both [ODataMetadataJsonOutputContext](https://github.com/OData/odata.net/blob/dev-8.x/src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs) and [ODataMetadataOutputContext](https://github.com/OData/odata.net/blob/dev-8.x/src/Microsoft.OData.Core/ODataMetadataOutputContext.cs)
> 
> **In both:**
> 
> * [ODataMetadataJsonOutputContext - Task WriteMetadataDocumentAsync()](https://github.com/OData/odata.net/blob/dev-8.x/src/Microsoft.OData.Core/ODataMetadataJsonOutputContext.cs#L92)
> * [ODataMetadataOutputContext - Task WriteMetadataDocumentAsync()](https://github.com/OData/odata.net/blob/dev-8.x/src/Microsoft.OData.Core/ODataMetadataOutputContext.cs#L121)
> 
> **Change From**
> 
> ```cs
> internal override Task WriteMetadataDocumentAsync()
> {
>     this.AssertAsynchronous();
> 
>     return TaskUtils.GetTaskForSynchronousOperationReturningTask(
>         () =>
>         {
>             this.WriteMetadataDocumentImplementation();
>             return this.FlushAsync();
>         });
> }
> ```
> 
> **to**
> 
> ```cs
> internal override async Task WriteMetadataDocumentAsync()
> {
>     this.AssertAsynchronous();
> 
>     await this.WriteMetadataDocumentImplementationAsync().ConfigureAwait(false);
>     await this.FlushAsync().ConfigureAwait(false);
> }
> ```

It wouldn't be a breaking change if it doesn't change what's written on the wire...

In addition, part of the objective for implementing the async APIs was so we could remove these async wrapper over sync APIs. The change you mentioned above is the right/desired change. Make sure to add tests to verify that the output is the same from both `WriteMetadataDocument` and `WriteMetadataDocumentAsync`

WanjohiSammy:
> It wouldn't be a breaking change if it doesn't change what's written on the wire...
> 
> In addition, part of the objective for implementing the async APIs was so we could remove these async wrapper over sync APIs. The change you mentioned above is the right/desired change. Make sure to add tests to verify that the output is the same from both `WriteMetadataDocument` and `WriteMetadataDocumentAsync`

@gathogojr 

This has been done and pushed as commit: [Added WriteMetadataDocumentAsync methods and tests](https://github.com/OData/odata.net/pull/3006/commits/64977a61dfcd5a20c48884b4cf304c14ef9b9c83)

I have added tests to compare for both `XML` and `JSON` metadata generated by `WriteMetadataDocument` and `WriteMetadataDocumentAsync`

src/Microsoft.OData.Edm/Csdl/CsdlWriter.cs
habbes: My point was that an empty virtual method that's always expected to write something is a bad idea. I would prefer that they bet turned abstract instead, but I don't know if there are unknown repercussions that would follow, and maybe that could derail this PR.

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaWriter.cs
habbes: The purpose of a default implementation is that the child class should still be able to behave reasonably well without overriding the virtual method. If child classes are always expected to override the virtual methods to function correctly, then they should be abstract.

src/Microsoft.OData.Edm/Csdl/CsdlWriter.cs
habbes: On a side note, interesting discussion about `TextWriter.Write(char)` being an empty virtual method instead of an abstract method: https://github.com/dotnet/runtime/discussions/104345

habbes: I also wanted to point out that this is not a blocker for me.

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
habbes: Just wanted to point out that `Schemata` is the plural of `Schema`. Maybe this method was intentionally named so to indicate that it's meant to write more than one _schemas_. With that in mind, if we have a problem with the word "schemata", maybe we could rename the method to `WriteSchemas` instead.

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
habbes: See my comments here: https://github.com/OData/odata.net/pull/3006#discussion_r1675475503

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaWriter.cs
habbes: I don't know what the impact of making them abstract would be. Maybe it might open a can of worms that derails the PR. For that reason, this is not a blocker for me.

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
WanjohiSammy: @ElizabethOkerio Please see the previous comment.
@habbes

src/Microsoft.OData.Edm/Csdl/Serialization/EdmModelCsdlSchemaWriter.cs
WanjohiSammy: Noted

src/Microsoft.OData.Edm/Csdl/CsdlWriter.cs
WanjohiSammy: I will close this thread as resolved

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
WanjohiSammy: @habbes @ElizabethOkerio 

Lets rename these private methods to `WriteSchemas` and `WriteSchemasAsync`

WanjohiSammy: This has been done.

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
```diff
@@ -49,13 +50,30 @@ protected override void WriteCsdl()
             WriteReferenceElements();
 
             // It also MAY contain members for schemas.
-            WriteSchemata();
```
xuzhg, line 52: Schemata is correct plural of Schema, right?

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
```diff
@@ -84,9 +123,20 @@ EdmModelReferenceElementsJsonVisitor visitor
         }
 
         /// <summary>
-        /// Schema Object.
+        /// Asynchronously writes $Reference Object.
         /// </summary>
-        private void WriteSchemata()
+        /// <returns>A task that represents the asynchronous operation</returns>
+        private Task WriteReferenceElementsAsync()
```
xuzhg, line 129: no async key word for this private method?

xuzhg (dismissed):

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
WanjohiSammy: @xuzhg That is correct. However, I have used `WriteSchemas` and `WriteSchemasAsync` to be the same way they are named in [CsdlXmlWriter](https://github.com/OData/odata.net/blob/fix/2684-csdlwriter-async-apis/src/Microsoft.OData.Edm/Csdl/CsdlXmlWriter.cs#L19) where we have [WriteSchemas](https://github.com/OData/odata.net/blob/fix/2684-csdlwriter-async-apis/src/Microsoft.OData.Edm/Csdl/CsdlXmlWriter.cs#L246) and [WriteSchemasAsync](https://github.com/OData/odata.net/blob/fix/2684-csdlwriter-async-apis/src/Microsoft.OData.Edm/Csdl/CsdlXmlWriter.cs#L259)

src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs
WanjohiSammy: @xuzhg This is the method that is in question to either rename it to `WriteSchemas` or maintain it as `WriteSchemata`. The corresponding async method is here this 
[async Task WriteSchemasAsync()](https://github.com/OData/odata.net/blob/fix/2684-csdlwriter-async-apis/src/Microsoft.OData.Edm/Csdl/CsdlJsonWriter.cs#L157)

habbes (approved):

PR #3013, opened 2024-07-04
Support Enums as Keys in OData Client

### LINKED ISSUE #2964: Odata returns generic type error when enum is a key
Our project uses enums as keys. For example we use weekday as a key here.
 
image

This wouldn’t be a problem expect that when getting the data from odata the mapping fails with the following error:

           This operation is only valid on generic types.

### Assemblies affected

Microsoft.OData.Core 7.x

### Reproduce steps

Add an enum as a key

### Expected result

We expect enum's to be useable as keys

### Additional detail

In the file OdatatypeInfo.cs in the method GetKeyProperties() we found that the program assumes that the property should be generic type and afterwards checks if it is an enum. 

 
image

We think the order should be changed in the if statement and the IsEnum() should be inverted.  

Are we over looking something? We tried getting, updating and posting data with this change. Which all worked without problem

image

WanjohiSammy:
### Issues

*This pull request fixes #2964.*

### Description

*This change involves adding functionality to support the use of `enumeration types` as `keys`.*

Currently, attempting to query entities with `enums as keys` through `OData Client`, results in an error stating that the operation is not valid for non-generic types. 

**Exception**:
`This operation is only valid on generic types.`

**Change**:
This modification will address this issue, allowing `enums` as keys without causing exceptions.
For example, the below `Employee` `object` has `EmpType` which is a key of `Type Enum` and `NullableEmpType` which is `nullable generic Type with Enum as generic argument`.

```cs
public class Employee
{
    [System.ComponentModel.DataAnnotations.Key]
    public int EmpNumber { get; set; }

    // Enum - Employee Type Key
    [System.ComponentModel.DataAnnotations.Key]
    public EmployeeType EmpType { get; set; }

    [System.ComponentModel.DataAnnotations.Key]
    public EmployeeType? NullableEmpType { get; set; }

    public decimal Salary { get; set; }
}

public enum EmployeeType
{
    Hybrid = 1,
    FullTime = 2,
    PartTime = 3
}
```

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
+++ b/src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
@@ -321,7 +321,9 @@ namespace Microsoft.OData.Client.Metadata
                     throw c.Error.InvalidOperation(c.Strings.ClientType_KeysOnDifferentDeclaredType(typeName));
                 }
 
-                if (!PrimitiveType.IsKnownType(key.PropertyType) && !(key.PropertyType.GetGenericTypeDefinition() == typeof(System.Nullable<>) && key.PropertyType.GetGenericArguments().First().IsEnum()))
+                // Check if the key property's type is a known primitive, an enum, or a nullable generic.
+                // If it doesn't meet any of these conditions, throw an InvalidOperationException.
+                if (!PrimitiveType.IsKnownType(key.PropertyType) && !key.PropertyType.IsEnum() && !(key.PropertyType.IsGenericType() && key.PropertyType.GetGenericTypeDefinition() == typeof(System.Nullable<>)))
                 {
                     throw c.Error.InvalidOperation(c.Strings.ClientType_KeysMustBeSimpleTypes(key.Name, typeName, key.PropertyType.FullName));
                 }
--- a/src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
+++ b/src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
@@ -323,7 +323,7 @@ namespace Microsoft.OData.Client.Metadata
 
                 // Check if the key property's type is a known primitive, an enum, or a nullable generic.
                 // If it doesn't meet any of these conditions, throw an InvalidOperationException.
-                if (!PrimitiveType.IsKnownType(key.PropertyType) && !key.PropertyType.IsEnum() && !(key.PropertyType.IsGenericType() && key.PropertyType.GetGenericTypeDefinition() == typeof(System.Nullable<>)))
+                if (!PrimitiveType.IsKnownType(key.PropertyType) && !key.PropertyType.IsEnum() && !(key.PropertyType.IsGenericType() && key.PropertyType.GetGenericTypeDefinition() == typeof(System.Nullable<>) && key.PropertyType.GetGenericArguments().First().IsEnum()))
                 {
                     throw c.Error.InvalidOperation(c.Strings.ClientType_KeysMustBeSimpleTypes(key.Name, typeName, key.PropertyType.FullName));
                 }
--- a/test/FunctionalTests/Microsoft.OData.Client.Tests/Metadata/ClientTypeUtilTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Client.Tests/Metadata/ClientTypeUtilTests.cs
@@ -7,6 +7,7 @@
 using Microsoft.OData.Client.Metadata;
 using System;
 using System.Linq;
+using System.Reflection;
 using Xunit;
 
 namespace Microsoft.OData.Client.Tests.Metadata
@@ -70,10 +71,10 @@ namespace Microsoft.OData.Client.Tests.Metadata
             //Arrange
             Type employee = typeof(Employee);
 
-            int expectedNumberOfKeyProperties = 4; // 2 Primitive Known Types, 1 Enum Type, 1 Nullable Generic Type
+            int expectedNumberOfKeyProperties = 4; // 2 Primitive Known Types, 1 Enum Type, 1 Enum Nullable Generic Type
 
             //Act
-            var keyProperties = ClientTypeUtil.GetKeyPropertiesOnType(employee);
+            PropertyInfo[] keyProperties = ClientTypeUtil.GetKeyPropertiesOnType(employee);
 
             //Assert
             Assert.Equal(expectedNumberOfKeyProperties, keyProperties.Length);
@@ -86,8 +87,8 @@ namespace Microsoft.OData.Client.Tests.Metadata
             Type employee = typeof(Employee);
 
             //Act
-            var keyProperties = ClientTypeUtil.GetKeyPropertiesOnType(employee);
-            var key = keyProperties.Single(k => k.Name == "EmpType");
+            PropertyInfo[] keyProperties = ClientTypeUtil.GetKeyPropertiesOnType(employee);
+            PropertyInfo key = keyProperties.Single(k => k.Name == "EmpType");
 
             //Assert
             Assert.True(key.PropertyType.IsEnum());
@@ -101,10 +102,10 @@ namespace Microsoft.OData.Client.Tests.Metadata
             Type employee = typeof(Employee);
 
             //Act
-            var keyProperties = ClientTypeUtil.GetKeyPropertiesOnType(employee);
+            PropertyInfo[] keyProperties = ClientTypeUtil.GetKeyPropertiesOnType(employee);
 
-            var empNumKey = keyProperties.Single(k => k.Name == "EmpNumber");
-            var deptNumKey = keyProperties.Single(k => k.Name == "DeptNumber");
+            PropertyInfo empNumKey = keyProperties.Single(k => k.Name == "EmpNumber");
+            PropertyInfo deptNumKey = keyProperties.Single(k => k.Name == "DeptNumber");
 
             //Assert
             Assert.True(PrimitiveType.IsKnownType(empNumKey.PropertyType) && empNumKey.PropertyType == typeof(int));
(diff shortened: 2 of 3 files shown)


src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs
```diff
@@ -321,7 +321,9 @@ private PropertyInfo[] GetKeyProperties()
                     throw c.Error.InvalidOperation(c.Strings.ClientType_KeysOnDifferentDeclaredType(typeName));
                 }
 
-                if (!PrimitiveType.IsKnownType(key.PropertyType) && !(key.PropertyType.GetGenericTypeDefinition() == typeof(System.Nullable<>) && key.PropertyType.GetGenericArguments().First().IsEnum()))
+                // Check if the key property's type is a known primitive, an enum, or a nullable generic.
+                // If it doesn't meet any of these conditions, throw an InvalidOperationException.
+                if (!PrimitiveType.IsKnownType(key.PropertyType) && !key.PropertyType.IsEnum() && !(key.PropertyType.IsGenericType() && key.PropertyType.GetGenericTypeDefinition() == typeof(System.Nullable<>)))
```
mikepizzo, line 326: Prior to the proposed change, we *only* allowed nullable<enum>, and not other nullable types.  After the proposed change, we allow any nullable<T>.  I'm not sure why we support nullable<enum>, since keys have to be non-nullable, but it would be a breaking change to fix that at this point. So, we should add support for enum, make sure we only call GetGenericTypeDefinition if it is a GenericType, and limit generic types to nullable<enum>.

```suggestion
                if (!PrimitiveType.IsKnownType(key.PropertyType) && !key.PropertyType.IsEnum() && !(key.PropertyType.IsGenericType() && key.PropertyType.GetGenericTypeDefinition() == typeof(System.Nullable<> && key.PropertyType.GetGenericArguments().First().IsEnum())))
```

WanjohiSammy: @mikepizzo 

Thank you for bringing this up. I have limited it to allow nullable generic type keys with enum as generic argument.

This is fixed as suggested.

martinfbluesoftcz: Yes it looks fine now.

habbes (commented):
I think it would be good to add a test that shows that querying data by keys with enum executes corectly without throwing an exception, just to make sure that the big-picture scenario that is describe in the original issue is resolved.

mikepizzo (approved):

WanjohiSammy:
> I think it would be good to add a test that shows that querying data by keys with enum executes corectly without throwing an exception, just to make sure that the big-picture scenario that is describe in the original issue is resolved.

@habbes Yes, this can be done. However, I have added tests on the specific methods that are called when getting `KeyProperties`. [GetKeyProperties()](https://github.com/OData/odata.net/blob/fix/2964-support-enum-as-key/src/Microsoft.OData.Client/Metadata/ODataTypeInfo.cs#L273) method is the one that was throwing exception when enum was used as keys.

[test/FunctionalTests/Microsoft.OData.Client.Tests/Metadata/ClientTypeUtilTests.cs](https://github.com/OData/odata.net/pull/3013/files#diff-bb00d0b81ba8b4237fc55f0e85caee3321de1afd6f918b1913cd129e2d1b73ba)

WanjohiSammy:
> [#3013 (comment)](https://github.com/OData/odata.net/pull/3013#discussion_r1674783033)
> 
> @WanjohiSammy Probably add a test(s) to verify that nullable properties are not permitted (other than nullable enum property). Also create an issue to fix it in next major release such that no nullable property area allowed. We published ODL 8 RC so I don't think we can sneak in the change in ODL 8

@gathogojr I already added a test for this case where I am using a `Key` of `Type struct.` This will throw an exception.
[IFTypeProperty_HasNullableGenericTypeKey_OfTypeStruct_GetKeyPropertiesOnType_ThrowsException()](https://github.com/OData/odata.net/pull/3013/files#diff-bb00d0b81ba8b4237fc55f0e85caee3321de1afd6f918b1913cd129e2d1b73ba:~:text=IFTypeProperty_HasNullableGenericTypeKey_OfTypeStruct_GetKeyPropertiesOnType_ThrowsException)

ElizabethOkerio (approved):

WanjohiSammy:
> I think it would be good to add a test that shows that querying data by keys with enum executes corectly without throwing an exception, just to make sure that the big-picture scenario that is describe in the original issue is resolved.

@habbes Thanks for this. I have added these tests in the file [test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextQueryTests.cs](https://github.com/OData/odata.net/pull/3013/files#diff-925c126a9411dd12bce58e13d4c4c48f79e51bcb1226daa61dae647fcad0263a)

mikepizzo (dismissed):

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextQueryTests.cs
```diff
@@ -0,0 +1,189 @@
...
+            ""EmpType"": ""PartTime"",
+            ""OrgId"": 1,
+            ""Name"": ""Employee Two""
+        }
+    ]
+}";
+            SetupContextWithRequestPipeline(new DataServiceContext[] { _defaultContext }, response, "employees");
+
+            // Act
+            Employee employee = _defaultContext.Employees.Where(e => e.EmpNumber == 8).First();
```
habbes, line 122: Could you have a test where you filter by the enum property (i.e. `e.EmpType`)? And could you have another test where you fetch by key (using `ByKey` with the composite key instead of `Where`)?

test/FunctionalTests/Microsoft.OData.Client.Tests/Metadata/ClientTypeUtilTests.cs
```diff
@@ -50,6 +52,99 @@ public void IFTypeProperty_HasKeyAttributeAndOneProperty_TypeIsEntityAndDoesNotT
...
+            //Arrange
+            Type employee = typeof(Employee);
+
+            int expectedNumberOfKeyProperties = 4; // 2 Primitive Known Types, 1 Enum Type, 1 Enum Nullable Generic Type
+
+            //Act
+            PropertyInfo[] keyProperties = ClientTypeUtil.GetKeyPropertiesOnType(employee);
+
+            //Assert
+            Assert.Equal(expectedNumberOfKeyProperties, keyProperties.Length);
```
habbes, line 80: Maybe we should assert the names of the actual properties are what we expect.

habbes (dismissed):

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextQueryTests.cs
```diff
@@ -0,0 +1,274 @@
...
+            ""EmpType"": ""PartTime"",
+            ""OrgId"": 1,
+            ""Name"": ""Employee 24""
+        }
+    ]
+}";
+            SetupContextWithRequestPipeline(new DataServiceContext[] { _defaultContext }, response, "Employees");
+
+            // Act
+            EmployeeSingle query = _defaultContext.Employees.ByKey(
```
habbes, line 186: nit: since this is a custom extension method, would be good to assert that the generated request URL is as expected.

WanjohiSammy: @habbes I have added the assert for query URL.

[Assert.Equal(expectedUri, query.Query.ToString());](https://github.com/OData/odata.net/pull/3013/files/04d79b127d87e6222d99cec5f14c6ae27cc39b78#diff-925c126a9411dd12bce58e13d4c4c48f79e51bcb1226daa61dae647fcad0263a:~:text=Assert.Equal(expectedUri%2C%20query.Query.ToString())%3B)

test/FunctionalTests/Microsoft.OData.Client.Tests/Metadata/ClientTypeUtilTests.cs
WanjohiSammy: Yes, the tests for actual properties expected are on separate methods below.

mikepizzo (approved):
Thanks for the contribution and follow-up on feedback.

test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextQueryTests.cs
```diff
@@ -0,0 +1,274 @@
...
+
+            Employee employee = query.GetValue();
+
+            // Assert
+            Assert.Equal(expectedUri, query.Query.ToString());
+            Assert.Equal("Employee 24", employee.Name);
+            Assert.Equal(EmployeeType.PartTime, employee.EmpType);
+        }
+
+        private void SetupContextWithRequestPipeline(DataServiceContext[] contexts, string response, string path)
```
gathogojr, line 197: @WanjohiSammy What is the rationale behind make the first parameter an array? I don't see any scenario where you'd need to pass an array of `DataServiceContext`s

WanjohiSammy: @gathogojr 

That's correct John. To ensure code reusability, I reused the same logic and code from these tests:

- [DataServiceContextNoTrackingStreamsTests.cs#L321](https://github.com/OData/odata.net/blob/main/test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/DataServiceContextNoTrackingStreamsTests.cs#L321)
- [ClientEntityTrackerTests.cs#L229](https://github.com/OData/odata.net/blob/main/test/FunctionalTests/Microsoft.OData.Client.Tests/Tracking/ClientEntityTrackerTests.cs#L229)

**For best practices, we can abstract this similar code in a helper method and reuse it.** Let me know if I can clean up this part of code in this PR.

WanjohiSammy: @gathogojr 

I have resolve this and also added a test to filter by enum key only using ByKey() in [commit 6a5667f](https://github.com/OData/odata.net/pull/3013/commits/6a5667f941b8fbe9d10929ef72909495775f2582)

gathogojr (approved):

PR #3014, opened 2024-07-10
Allow $filtering with enum integer value in squote and without squote

### LINKED ISSUE #2542: Filter with enum integer value

Filtering an enum with it's integer value causes an exception

### Assemblies affected

Microsoft.AspNetCore.OData 8.0.11
Also whatever the Tripin service on odata.org is using

### Reproduce steps

the Tripin service used on odata.org is using this library, as the error message is the same.

https://services.odata.org/TripPinRESTierService/(S(bhxqknjf3hgdlqxrtt5uuffr))/People?$count=true&$filter=Gender%20eq%201

### Expected result

Same results as https://services.odata.org/TripPinRESTierService/(S(bhxqknjf3hgdlqxrtt5uuffr))/People?$count=true&$filter=Gender%20eq%20%27Female%27

### Actual result

*What is actually happening.*

```json
{"error":{"code":"","message":"A binary operator with incompatible types was detected. Found operand types 'Trippin.PersonGender' and 'Edm.Int32' for operator kind 'Equal'."}}
```

```json
{"stacktrace": "   at Microsoft.OData.UriParser.BinaryOperatorBinder.PromoteOperandTypes(BinaryOperatorKind binaryOperatorKind, SingleValueNode& left, SingleValueNode& right, TypeFacetsPromotionRules facetsPromotionRules)\r\n   at Microsoft.OData.UriParser.BinaryOperatorBinder.BindBinaryOperator(BinaryOperatorToken binaryOperatorToken)\r\n   at Microsoft.OData.UriParser.MetadataBinder.Bind(QueryToken token)\r\n   at Microsoft.OData.UriParser.BinaryOperatorBinder.GetOperandFromToken(BinaryOperatorKind operatorKind, QueryToken queryToken)\r\n   at Microsoft.OData.UriParser.BinaryOperatorBinder.BindBinaryOperator(BinaryOperatorToken binaryOperatorToken)\r\n   at Microsoft.OData.UriParser.MetadataBinder.Bind(QueryToken token)\r\n   at Microsoft.OData.UriParser.FilterBinder.BindFilter(QueryToken filter)\r\n   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseFilterImplementation(String filter, ODataUriParserConfiguration configuration, ODataPathInfo odataPathInfo)\r\n   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseFilter()\r\n   at Microsoft.AspNetCore.OData.Query.FilterQueryOption.get_FilterClause()\r\n   at Microsoft.AspNetCore.OData.Query.Validator.FilterQueryValidator.Validate(FilterQueryOption filterQueryOption, ODataValidationSettings settings)\r\n   at Microsoft.AspNetCore.OData.Query.FilterQueryOption.Validate(ODataValidationSettings validationSettings)\r\n   at Microsoft.AspNetCore.OData.Query.Validator.ODataQueryValidator.Validate(ODataQueryOptions options, ODataValidationSettings validationSettings)\r\n   at Microsoft.AspNetCore.OData.Query.ODataQueryOptions.Validate(ODataValidationSettings validationSettings)\r\n   at Microsoft.AspNetCore.OData.Query.EnableQueryAttribute.ValidateQuery(HttpRequest request, ODataQueryOptions queryOptions)\r\n   at Microsoft.AspNetCore.OData.Query.EnableQueryAttribute.OnActionExecuting(ActionExecutingContext actionExecutingContext)"}
```

### Additional detail

```
enum  = [ qualifiedEnumTypeName ] SQUOTE enumValue SQUOTE
enumValue  = singleEnumValue
singleEnumValue = enumerationMember / enumMemberValue
enumMemberValue = int64Value
```

which I assume means we should able to use a long when referring to an enum

WanjohiSammy:
### Issues

*This pull request fixes #2542.*

### Description

This change is to allow `$filter` using `enum` `integer` `value`.

For example, to allow:
- [/People?$count=true&$filter=Gender eq 1](https://services.odata.org/TripPinRESTierService/(S(bhxqknjf3hgdlqxrtt5uuffr))/People?$count=true&$filter=Gender%20eq%201)
- [/People?$count=true&$filter=Gender eq '1'](https://services.odata.org/TripPinRESTierService/(S(bhxqknjf3hgdlqxrtt5uuffr))/People?$count=true&$filter=Gender%20eq%20%271%27)
- [/People?$count=true&$filter=Features/Any(f: f eq 1)](https://services.odata.org/TripPinRESTierService/(S(bhxqknjf3hgdlqxrtt5uuffr))/People?$count=true&$filter=Features/Any(f:%20f%20eq%201))
- [/People?$count=true&$filter=Features/Any(f: f eq '1')](https://services.odata.org/TripPinRESTierService/(S(bhxqknjf3hgdlqxrtt5uuffr))/People?$count=true&$filter=Features/Any(f:%20f%20eq%20%271%27))

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
@@ -58,19 +58,31 @@ namespace Microsoft.OData.UriParser
                 }
 
                 ConstantNode constantNode = source as ConstantNode;
-                if (constantNode != null && constantNode.Value != null && source.TypeReference.IsString() && targetTypeReference.IsEnum())
+                // Check if the source node is a constant node, not null, and the source type is either string or integral
+                // and the target type is an enum.
+                if (constantNode != null && constantNode.Value != null && (source.TypeReference.IsString() || source.TypeReference.IsIntegral()) && targetTypeReference.IsEnum())
                 {
                     string memberName = constantNode.Value.ToString();
                     IEdmEnumType enumType = targetTypeReference.Definition as IEdmEnumType;
                     if (enumType.Members.Any(m => string.Compare(m.Name, memberName, StringComparison.Ordinal) == 0))
                     {
                         string literalText = ODataUriUtils.ConvertToUriLiteral(constantNode.Value, default(ODataVersion));
-                        return new ConstantNode(new ODataEnumValue(constantNode.Value.ToString(), targetTypeReference.Definition.ToString()), literalText, targetTypeReference);
+                        return new ConstantNode(new ODataEnumValue(memberName, enumType.ToString()), literalText, targetTypeReference);
                     }
-                    else
+
+                    // If the member name is an integral value, we should try to convert it to the enum member name.
+                    if (int.TryParse(memberName, out int memberIntegralValue))
                     {
-                        throw new ODataException(ODataErrorStrings.Binder_IsNotValidEnumConstant(memberName));
+                        // Find the enum member with the matching integral value
+                        var enumMember = enumType.Members.FirstOrDefault(m => m.Value.Value == memberIntegralValue);
+                        if (enumMember != null)
+                        {
+                            string literalText = ODataUriUtils.ConvertToUriLiteral(enumMember.Name, default(ODataVersion));
+                            return new ConstantNode(new ODataEnumValue(enumMember.Name, enumType.ToString()), literalText, targetTypeReference);
+                        }
                     }
+
+                    throw new ODataException(ODataErrorStrings.Binder_IsNotValidEnumConstant(memberName));
                 }
 
                 if (!TypePromotionUtils.CanConvertTo(source, source.TypeReference, targetTypeReference))
--- a/src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
@@ -64,18 +64,17 @@ namespace Microsoft.OData.UriParser
                 {
                     string memberName = constantNode.Value.ToString();
                     IEdmEnumType enumType = targetTypeReference.Definition as IEdmEnumType;
-                    if (enumType.Members.Any(m => string.Compare(m.Name, memberName, StringComparison.Ordinal) == 0))
+                    if(enumType.EnumMemberExists(memberName, StringComparison.Ordinal))
                     {
                         string literalText = ODataUriUtils.ConvertToUriLiteral(constantNode.Value, default(ODataVersion));
                         return new ConstantNode(new ODataEnumValue(memberName, enumType.ToString()), literalText, targetTypeReference);
                     }
 
                     // If the member name is an integral value, we should try to convert it to the enum member name.
-                    if (int.TryParse(memberName, out int memberIntegralValue))
+                    if (long.TryParse(memberName, out long memberIntegralValue))
                     {
                         // Find the enum member with the matching integral value
-                        var enumMember = enumType.Members.FirstOrDefault(m => m.Value.Value == memberIntegralValue);
-                        if (enumMember != null)
+                        if (enumType.TryParseEnum(memberIntegralValue, out IEdmEnumMember enumMember))
                         {
                             string literalText = ODataUriUtils.ConvertToUriLiteral(enumMember.Name, default(ODataVersion));
                             return new ConstantNode(new ODataEnumValue(enumMember.Name, enumType.ToString()), literalText, targetTypeReference);
--- a/src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
@@ -64,21 +64,17 @@ namespace Microsoft.OData.UriParser
                 {
                     string memberName = constantNode.Value.ToString();
                     IEdmEnumType enumType = targetTypeReference.Definition as IEdmEnumType;
-                    if(enumType.EnumMemberExists(memberName, StringComparison.Ordinal))
+                    if(enumType.ContainsMember(memberName, StringComparison.Ordinal))
                     {
                         string literalText = ODataUriUtils.ConvertToUriLiteral(constantNode.Value, default(ODataVersion));
                         return new ConstantNode(new ODataEnumValue(memberName, enumType.ToString()), literalText, targetTypeReference);
                     }
 
-                    // If the member name is an integral value, we should try to convert it to the enum member name.
-                    if (long.TryParse(memberName, out long memberIntegralValue))
+                    // If the member name is an integral value, we should try to convert it to the enum member name and find the enum member with the matching integral value
+                    if (long.TryParse(memberName, out long memberIntegralValue) && enumType.TryParse(memberIntegralValue, out IEdmEnumMember enumMember))
                     {
-                        // Find the enum member with the matching integral value
-                        if (enumType.TryParseEnum(memberIntegralValue, out IEdmEnumMember enumMember))
-                        {
-                            string literalText = ODataUriUtils.ConvertToUriLiteral(enumMember.Name, default(ODataVersion));
-                            return new ConstantNode(new ODataEnumValue(enumMember.Name, enumType.ToString()), literalText, targetTypeReference);
-                        }
+                        string literalText = ODataUriUtils.ConvertToUriLiteral(enumMember.Name, default(ODataVersion));
+                        return new ConstantNode(new ODataEnumValue(enumMember.Name, enumType.ToString()), literalText, targetTypeReference);
                     }
 
                     throw new ODataException(ODataErrorStrings.Binder_IsNotValidEnumConstant(memberName));
--- a/src/Microsoft.OData.Core/UriParser/TypePromotionUtils.cs
+++ b/src/Microsoft.OData.Core/UriParser/TypePromotionUtils.cs
@@ -267,14 +267,15 @@ namespace Microsoft.OData.UriParser
                     return true;
                 }
 
-                // Comparing an enum with a string is valid
-                if (left != null && right != null && left.IsEnum() && right.IsString())
+                // Comparing an enum with a string or int is valid
+                if (left != null && right != null && left.IsEnum() && (right.IsString() || right.IsIntegral()))
                 {
                     right = left;
                     return true;
                 }
 
-                if (left != null && right != null && right.IsEnum() && left.IsString())
+                // Comparing an enum with a string or int is valid
+                if (left != null && right != null && right.IsEnum() && (left.IsString() || left.IsIntegral()))
                 {
                     left = right;
                     return true;
--- a/src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs
+++ b/src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs
@@ -127,6 +127,48 @@ namespace Microsoft.OData.Edm
         }
 
         /// <summary>
+        /// Parse an enum integral value to enum member.
+        /// </summary>
+        /// <param name="enumType">edm enum type</param>
+        /// <param name="value">input integral value.</param>
+        /// <param name="enumMember">parsed result.</param>
+        /// <returns>true if parse succeeds, false if parse fails.</returns>
+        public static bool TryParseEnum(this IEdmEnumType enumType, long value, out IEdmEnumMember enumMember)
+        {
+            enumMember = null;
+            foreach (var member in enumType.Members)
+            {
+                if (member.Value.Value == value)
+                {
+                    enumMember = member;
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        /// <summary>
+        /// Checks if the given member name exists in the enum type.
+        /// </summary>
+        /// <param name="enumType">The enum type.</param>
+        /// <param name="memberName">The member name to check.</param>
+        /// <param name="comparison">The comparison type to use for string comparison. Default is Ordinal.</param>
+        /// <returns>True if the member name exists in the enum type; otherwise, false.</returns>
+        public static bool EnumMemberExists(this IEdmEnumType enumType, string memberName, StringComparison comparison = StringComparison.Ordinal)
+        {
+            foreach (var member in enumType.Members)
+            {
+                if (string.Equals(member.Name, memberName, comparison))
+                {
+                    return true;
+                }
+            }
+
+            return false;
+        }
+
+        /// <summary>
         /// Convert enum int value to string
         /// </summary>
         /// <param name="type">edm enum type reference</param>
--- a/src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs
+++ b/src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs
@@ -133,7 +133,7 @@ namespace Microsoft.OData.Edm
         /// <param name="value">input integral value.</param>
         /// <param name="enumMember">parsed result.</param>
         /// <returns>true if parse succeeds, false if parse fails.</returns>
-        public static bool TryParseEnum(this IEdmEnumType enumType, long value, out IEdmEnumMember enumMember)
+        public static bool TryParse(this IEdmEnumType enumType, long value, out IEdmEnumMember enumMember)
         {
             enumMember = null;
             foreach (IEdmEnumMember member in enumType.Members)
@@ -155,7 +155,7 @@ namespace Microsoft.OData.Edm
         /// <param name="memberName">The member name to check.</param>
         /// <param name="comparison">The comparison type to use for string comparison. Default is Ordinal.</param>
         /// <returns>True if the member name exists in the enum type; otherwise, false.</returns>
-        public static bool EnumMemberExists(this IEdmEnumType enumType, string memberName, StringComparison comparison = StringComparison.Ordinal)
+        public static bool ContainsMember(this IEdmEnumType enumType, string memberName, StringComparison comparison = StringComparison.Ordinal)
         {
             foreach (IEdmEnumMember member in enumType.Members)
             {
--- a/src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs
+++ b/src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs
@@ -127,48 +127,6 @@ namespace Microsoft.OData.Edm
         }
 
         /// <summary>
-        /// Parse an enum integral value to enum member.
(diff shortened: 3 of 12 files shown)


src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
```diff
@@ -58,19 +58,31 @@ internal static SingleValueNode ConvertToTypeIfNeeded(SingleValueNode source, IE
...
                     if (enumType.Members.Any(m => string.Compare(m.Name, memberName, StringComparison.Ordinal) == 0))
                     {
                         string literalText = ODataUriUtils.ConvertToUriLiteral(constantNode.Value, default(ODataVersion));
-                        return new ConstantNode(new ODataEnumValue(constantNode.Value.ToString(), targetTypeReference.Definition.ToString()), literalText, targetTypeReference);
+                        return new ConstantNode(new ODataEnumValue(memberName, enumType.ToString()), literalText, targetTypeReference);
                     }
-                    else
+
+                    // If the member name is an integral value, we should try to convert it to the enum member name.
+                    if (int.TryParse(memberName, out int memberIntegralValue))
```
gathogojr, line 74: What if someone has defined their enum as follows:
```csharp
public enum MyEnum : long
{
    MyEnumMember = (long)int.MaxValue + (long)int.MaxValue
}
```
Maybe we should use `long.TryParse` since it'd work for both `int` and `long` scenarios...

gathogojr: You should probably add a test case for that to confirm...

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
```diff
@@ -58,19 +58,31 @@ internal static SingleValueNode ConvertToTypeIfNeeded(SingleValueNode source, IE
...
                     {
-                        throw new ODataException(ODataErrorStrings.Binder_IsNotValidEnumConstant(memberName));
+                        // Find the enum member with the matching integral value
+                        var enumMember = enumType.Members.FirstOrDefault(m => m.Value.Value == memberIntegralValue);
```
gathogojr, line 77: Across this library's source code, it's the practice to use actual type name instead of `var`

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
```diff
@@ -58,19 +58,31 @@ internal static SingleValueNode ConvertToTypeIfNeeded(SingleValueNode source, IE
                 }
 
                 ConstantNode constantNode = source as ConstantNode;
-                if (constantNode != null && constantNode.Value != null && source.TypeReference.IsString() && targetTypeReference.IsEnum())
+                // Check if the source node is a constant node, not null, and the source type is either string or integral
+                // and the target type is an enum.
+                if (constantNode != null && constantNode.Value != null && (source.TypeReference.IsString() || source.TypeReference.IsIntegral()) && targetTypeReference.IsEnum())
```
gathogojr, line 63: Curious about `IsIntegral`. Does it return true for floating point numbers? If it does, that could be cause some grief. At least confirm that if someone uses a floating point number (e.g. `$filter=Gender eq 1.0`), the logic handles that gracefully and the expected exemption is thrown

src/Microsoft.OData.Core/UriParser/TypePromotionUtils.cs
```diff
@@ -267,14 +267,15 @@ internal static void GetTypeFacets(IEdmTypeReference type, out int? precision, o
                     return true;
                 }
 
-                // Comparing an enum with a string is valid
-                if (left != null && right != null && left.IsEnum() && right.IsString())
+                // Comparing an enum with a string or int is valid
+                if (left != null && right != null && left.IsEnum() && (right.IsString() || right.IsIntegral()))
```
gathogojr, line 271: This comment https://github.com/OData/odata.net/pull/3014/files#r1673723890 is relevant here as well. We wouldn't want to return `true` for floating point numbers

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
habbes: I believe Integral precludes floats, but I stand to be corrected. A confirmation with a test would great.

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
habbes, line 77: This predicate passed to `FirstOrDefault` will cause allocations because of capturing the `memberIntegralValue` variable.

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/MetadataBindingUtilsTests.cs
```diff
@@ -50,6 +51,86 @@ public void IfTypesCannotPromoteErrorIsThrown()
...
+
+            SingleValueNode source = new ConstantNode(enumValue);
+            IEdmTypeReference targetTypeReference = new EdmEnumTypeReference(WeekDayEmumType, false);
+
+            // Act
+            var result = MetadataBindingUtils.ConvertToTypeIfNeeded(source, targetTypeReference);
+
+            // Assert
+            var member = WeekDayEmumType.Members.First(m => m.Value.Value == enumValue);
+            result.ShouldBeEnumNode(WeekDayEmumType, member.Name);
```
habbes, line 69: I think we should also assert the actual expected member name, just to make sure we're getting the enum value that we expected.

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
```diff
@@ -58,19 +58,31 @@ internal static SingleValueNode ConvertToTypeIfNeeded(SingleValueNode source, IE
...
+
+                    // If the member name is an integral value, we should try to convert it to the enum member name.
+                    if (int.TryParse(memberName, out int memberIntegralValue))
                     {
-                        throw new ODataException(ODataErrorStrings.Binder_IsNotValidEnumConstant(memberName));
+                        // Find the enum member with the matching integral value
+                        var enumMember = enumType.Members.FirstOrDefault(m => m.Value.Value == memberIntegralValue);
+                        if (enumMember != null)
+                        {
+                            string literalText = ODataUriUtils.ConvertToUriLiteral(enumMember.Name, default(ODataVersion));
```
xuzhg, line 80: Is it possible to use the version from configuration?

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
xuzhg: IsIntegral covers: 
               case EdmPrimitiveTypeKind.Int64:
                case EdmPrimitiveTypeKind.Int32:
                case EdmPrimitiveTypeKind.Int16:
                case EdmPrimitiveTypeKind.Byte:
                case EdmPrimitiveTypeKind.SByte:

xuzhg (commented):
Donot we need to change 'EnumAsStringUriResolver'?

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
WanjohiSammy: @gathogojr 

I have used `long.TryParse` as it is able to handle the following integral scenarios:
```cs
case EdmPrimitiveTypeKind.Int64:
case EdmPrimitiveTypeKind.Int32:
case EdmPrimitiveTypeKind.Int16:
case EdmPrimitiveTypeKind.Byte:
case EdmPrimitiveTypeKind.SByte:
```

I have added the following tests for the scenarios you mentioned:
- [IfTypePromotionNeeded_SourceIsLongAsStringMemberValueAndTargetIsEnum_ConstantNodeIsCreated()](https://github.com/OData/odata.net/blob/fix/2542-filter-with-enum-integer/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/MetadataBindingUtilsTests.cs#L113)
- [IfTypePromotionNeeded_SourceIsLongMemberValueAndTargetIsEnum_ConstantNodeIsCreated()](https://github.com/OData/odata.net/blob/fix/2542-filter-with-enum-integer/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/MetadataBindingUtilsTests.cs#L75)
- [IfTypePromotionNeeded_SourceIsHugeLongMemberValueAndTargetIsEnum_ConstantNodeIsCreated()](https://github.com/OData/odata.net/blob/fix/2542-filter-with-enum-integer/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/MetadataBindingUtilsTests.cs#L94)
- [IfTypePromotionNeededForEnum_SourceIsIntegralMemberValueInStringAndTargetIsEnumType_ConstantNodeIsCreated()](https://github.com/OData/odata.net/blob/fix/2542-filter-with-enum-integer/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/MetadataBindingUtilsTests.cs#L132)

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
WanjohiSammy: @gathogojr 

Thanks for this John. I have replaced the `var `types with explicit types

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
WanjohiSammy: @gathogojr, @habbes , @xuzhg

Floating point numbers will throw `CannotConvertToType` exception when used. Here is the test for this:
- [IfTypePromotionNeeded_SourceIsFloatMemberValuesAndTargetIsEnum_CannotConvertToTypeExceptionIsThrown()](https://github.com/OData/odata.net/blob/fix/2542-filter-with-enum-integer/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/MetadataBindingUtilsTests.cs#L181)

src/Microsoft.OData.Core/UriParser/TypePromotionUtils.cs
WanjohiSammy: @gathogojr 

See my previous comment:

> Floating point numbers will throw CannotConvertToType exception when used. Here is the test for this:
> 
> [IfTypePromotionNeeded_SourceIsFloatMemberValuesAndTargetIsEnum_CannotConvertToTypeExceptionIsThrown()](https://github.com/OData/odata.net/blob/fix/2542-filter-with-enum-integer/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/MetadataBindingUtilsTests.cs#L181)

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
WanjohiSammy: @habbes 

Thank you for pointing this perf issue. I have added a public method in [EnumHelper](https://github.com/OData/odata.net/blob/fix/2542-filter-with-enum-integer/src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs) called [bool TryParseEnum(this IEdmEnumType enumType, long value, out IEdmEnumMember enumMember)](https://github.com/OData/odata.net/blob/fix/2542-filter-with-enum-integer/src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs#L136) that finds enum member value using enum integral value. I then used this method here like so:

```cs
if (enumType.TryParseEnum(memberIntegralValue, out IEdmEnumMember enumMember))
{
    string literalText = ODataUriUtils.ConvertToUriLiteral(enumMember.Name, default(ODataVersion));
    ...
}
```

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/MetadataBindingUtilsTests.cs
WanjohiSammy: @habbes 

Yes, this has been done. I have added this `assertion` to all the tests that I have added for this case

[MetadataBindingUtilsTests.cs](https://github.com/OData/odata.net/blob/fix/2542-filter-with-enum-integer/test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/MetadataBindingUtilsTests.cs)

test/FunctionalTests/Microsoft.OData.Core.Tests/UriParser/Binders/MetadataBindingUtilsTests.cs
```diff
@@ -50,6 +51,86 @@ public void IfTypesCannotPromoteErrorIsThrown()
...
+            var enumValue = "5";
+            SingleValueNode source = new ConstantNode(enumValue);
+            IEdmTypeReference targetTypeReference = new EdmEnumTypeReference(WeekDayEmumType, false);
+
+            // Act
+            var result = MetadataBindingUtils.ConvertToTypeIfNeeded(source, targetTypeReference);
+
+            // Assert
+            var member = WeekDayEmumType.Members.First(m => m.Value.Value == int.Parse(enumValue));
+            result.ShouldBeEnumNode(WeekDayEmumType, member.Name);
```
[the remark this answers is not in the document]
WanjohiSammy: @habbes 

Same here. This has been done. See my comment: https://github.com/OData/odata.net/pull/3014#discussion_r1678994226

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
WanjohiSammy: @habbes

I have also added a method in [EnumHelper](https://github.com/OData/odata.net/blob/fix/2542-filter-with-enum-integer/src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs) called [bool EnumMemberExists(this IEdmEnumType enumType, string memberName, StringComparison comparison = StringComparison.Ordinal)](https://github.com/OData/odata.net/blob/fix/2542-filter-with-enum-integer/src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs#L158) that checks if the given member name exists in the enum type.

This is used in the place where it was using `.Any`
```cs
if(enumType.EnumMemberExists(memberName, StringComparison.Ordinal))
{
    string literalText = ODataUriUtils.ConvertToUriLiteral(constantNode.Value, default(ODataVersion));
   ...
}
```

habbes (approved):

src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs
```diff
@@ -126,6 +126,48 @@ public static bool TryParseEnum(this IEdmEnumType enumType, string value, bool i
...
+        }
+
+        /// <summary>
+        /// Checks if the given member name exists in the enum type.
+        /// </summary>
+        /// <param name="enumType">The enum type.</param>
+        /// <param name="memberName">The member name to check.</param>
+        /// <param name="comparison">The comparison type to use for string comparison. Default is Ordinal.</param>
+        /// <returns>True if the member name exists in the enum type; otherwise, false.</returns>
+        public static bool EnumMemberExists(this IEdmEnumType enumType, string memberName, StringComparison comparison = StringComparison.Ordinal)
```
gathogojr, line 158: I feel that `ContainsMember` is probably a better fit since you're checking if the member specified as a string is contained in the enum - `myEnum.ContainsMember("foobar")`

src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs
```diff
@@ -126,6 +126,48 @@ public static bool TryParseEnum(this IEdmEnumType enumType, string value, bool i
...
         }
 
+        /// <summary>
+        /// Parse an enum integral value to enum member.
+        /// </summary>
+        /// <param name="enumType">edm enum type</param>
+        /// <param name="value">input integral value.</param>
+        /// <param name="enumMember">parsed result.</param>
+        /// <returns>true if parse succeeds, false if parse fails.</returns>
+        public static bool TryParseEnum(this IEdmEnumType enumType, long value, out IEdmEnumMember enumMember)
```
gathogojr, line 136: If you used `TryParse`, would it conflict with existing framework `TryParse` methods? If not, `TryParse` is a better fit than `TryParseEnum`

src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs
```diff
@@ -126,6 +126,48 @@ public static bool TryParseEnum(this IEdmEnumType enumType, string value, bool i
             return true;
         }
 
+        /// <summary>
```
gathogojr, line 129: Do the two methods need to be public? Probably not. Make them `internal` if possible and remove the public API changes.

WanjohiSammy: @gathogojr 

The two methods are in `Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs` and I want to use them in a different project `Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs`. If I make them `internal`, they are accessible in `Microsoft.OData.Core` project. If I make them `public`, they must be added to Public API.

src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs
WanjohiSammy: @gathogojr 

I have renamed the method to `TryParse`. Yes, you are right. It won't cause any conflict

src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs
WanjohiSammy: @gathogojr 

Yes, this is correct. I have renamed it to `ContainsMember`

src/Microsoft.OData.Edm/ExtensionMethods/EnumHelper.cs
WanjohiSammy: @gathogojr 

I have moved the 2 methods to [src/Microsoft.OData.Core/EdmExtensionMethods.cs](https://github.com/OData/odata.net/pull/3014/files#diff-9515e6bf05155cabcb114b3a0b071152eae0435663121151fde82b3fccffdb0b)

gathogojr (approved):
LGTM

PR #3017, opened 2024-07-16
Port action overloading tests

ElizabethOkerio:
### Issues

### Description

This PR ports the `ActionOverloadingTests` from using WCF services to use `Microsoft.AspNetCore.OData`

The `ActionOverloading` tests test action overloading. Action Overloading is where we have several actions with the same name but with different binding parameters. 

The PR also has some common classes that will be shared across the various tests.  These common classes are in the `Common` folder in the Microsoft.OData.Client.E2E.Tests.csproj`. Some of the common classes are: Class models, EdmModels, DataServiceContainers, etc.. 

- [ ] *Test cases added*

### Additional work necessary


buildandtest.yml
```diff
@@ -44,13 +41,11 @@ steps:
     projects: |      
      $(Build.SourcesDirectory)\sln\OData.Pipeline.sln
```
habbes, line 42: Are the e2e tests included back to OData.Pipeline.sln?

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/PriorityOrderer.cs
```diff
@@ -0,0 +1,52 @@
...
+
+            foreach (var list in sortedMethods.Keys.Select(priority => sortedMethods[priority]))
+            {
+                list.Sort((x, y) => StringComparer.OrdinalIgnoreCase.Compare(x.TestMethod.Method.Name, y.TestMethod.Method.Name));
+                foreach (TTestCase testCase in list)
+                    yield return testCase;
+            }
+        }
+
+        static TValue GetOrCreate<TKey, TValue>(IDictionary<TKey, TValue> dictionary, TKey key) where TValue : new()
```
habbes, line 40: nitpick: I wonder why this had to be a generic method yet it's only used once, and only sed against known key of and value types `int` and `List<TTestCase>>`. I think that made slightly harder to read. But not a big deal.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/OrderLinesController.cs
```diff
@@ -0,0 +1,49 @@
...
+    public class OrderLinesController : ODataController
+    {
+        public OrderLinesController()
+        {
+        }
+
+        [EnableQuery]
+        public IActionResult Get()
+        {
+            var products = CommonEndToEndDataSource.OrderLines;
```
habbes, line 25: Why is this called products? Maybe you could show the actual type that's being returned instead of using `var` here to avoid confusion since it's not evident from the value.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/PeopleController.cs
```diff
@@ -0,0 +1,124 @@
...
+            var people = CommonEndToEndDataSource.People;
+            return Ok(people);
+        }
+
+        [EnableQuery]
+        [HttpGet("odata/People({key})")]
+        public IActionResult Get(int key)
+        {
+            var person = CommonEndToEndDataSource.People?.FirstOrDefault(a => a.PersonId == key);
+            return Ok(person);
```
habbes, line 33: Why don't we return 404 if the person is not found?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/PeopleController.cs
```diff
@@ -0,0 +1,124 @@
...
+
+        [HttpPost("odata/People({key})/Microsoft.OData.Client.E2E.Tests.Common.Server.EndToEnd.Person/Default.UpdatePersonInfo")]
+        public IActionResult UpdateEmployeeInfo([FromODataUri] int key)
+        {
+            if (!ModelState.IsValid)
+            {
+                return BadRequest();
+            }
+
+            CommonEndToEndDataSource.People.First(a => a.PersonId == key).Name += "[UpdataPersonName]";
```
habbes, line 58: This will throw an unhandled exception if the entity by the key doesn't exist. Why not throw an exception instead?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/PeopleController.cs
```diff
@@ -0,0 +1,124 @@
...
+            {
+                specialEmployee.Salary += 1;
+                return Ok(specialEmployee.Salary);
+            }
+            else
+            {
+                var employee = (Employee)CommonEndToEndDataSource.People.First(a => a.PersonId == key);
+                employee.Salary += (int)parameters["n"];
+
+                return Ok(true);
```
habbes, line 90: I wonder what the return type of this action is declared in the IEdmModel since here it appears to return different things depending on the type.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/ProductsController.cs
```diff
@@ -0,0 +1,56 @@
...
+        public IActionResult Get()
+        {
+            var products = CommonEndToEndDataSource.Products;
+            return Ok(products);
+        }
+
+        [EnableQuery]
+        public IActionResult Get(int key)
+        {
+            var product = CommonEndToEndDataSource.Products?.FirstOrDefault(a => a.ProductId == key);
```
habbes, line 32: What if the product doesn't exist, why not return a 404? Also, why does `Products` have the `?` operator here?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingQueryTests.cs
```diff
@@ -0,0 +1,253 @@
...
+
+        /// <summary>
+        /// Verify metadata of the operations defined in the test service.
+        /// </summary>
+        [Fact]
+        public void QueryMetadataTest()
+        {
+            Dictionary<string, string> expectedUpdatePersonInfoOperations = new Dictionary<string, string>()
+            {
+                { "", null }, // the unbound action
```
habbes, line 63: What do these values mean?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingQueryTests.cs
```diff
@@ -0,0 +1,253 @@
...
+            string retrieveProduct = "RetrieveProduct";
+            string retrieveProductName = "Default.RetrieveProduct";
+            string increaseEmployeeSalary = "IncreaseEmployeeSalary";
+            string increaseEmployeeSalaryName = "Default.IncreaseEmployeeSalary";
+            string sack = "Sack";
+            string sackName = "Default.Sack";
+
+            foreach (string mimeType in mimeTypes)
+            {
+                List<Tuple<string, string>> expectedActionsOnPerson = new List<Tuple<string, string>>()
```
habbes, line 99: Consider`ValueTuple` instead of `Tuple`, also consider the shorthand syntax: `List<(string, string)> expectedActinsOnPerson = ...`

If you want, you could also name the fields of the tuple `List<(string ActionName, string Endpoint)> ...`

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingQueryTests.cs
```diff
@@ -0,0 +1,253 @@
...
+                {
+                    new Tuple<string, string>(MetadataPrefix + updatePersonInfo, "People(-1)/" + updatePersonInfoName),
```
habbes, line 101: Consider the shorthand ValueTuple syntax:
```suggestion
                    (MetadataPrefix + updatePersonInfo, "People(-1)/" + updatePersonInfoName),

```

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingQueryTests.cs
```diff
@@ -0,0 +1,253 @@
...
+            _context.HttpClientFactory = HttpClientFactory;
+            _model = CommonEndToEndEdmModel.GetEdmModel();
+        }
+
+        /// <summary>
+        /// Verify metadata of the operations defined in the test service.
+        /// </summary>
+        [Fact]
+        public void QueryMetadataTest()
+        {
```
habbes, line 60: It's not easy to tell what these tests are doing.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingQueryTests.cs
```diff
@@ -0,0 +1,253 @@
...
+
+            this.VerifyOperationsInMetadata(expectedUpdatePersonInfoOperations, actionImports.Where(fi => fi.Name == "UpdatePersonInfo"));
+            this.VerifyOperationsInMetadata(expectedRetrieveProductOperations, actionImports.Where(fi => fi.Name == "RetrieveProduct"));
+        }
+
+        /// <summary>
+        /// Verify actions in entry payload format atom, json verbose, and json fullmetadata.
+        /// </summary>
+        [Fact]
+        public void QueryEntryTest()
```
habbes, line 80: Having a hard time wrapping my head around what these tests are actual doing.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingQueryTests.cs
```diff
@@ -0,0 +1,253 @@
...
+            string sack = "Sack";
+            string sackName = "Default.Sack";
+
+            foreach (string mimeType in mimeTypes)
+            {
+                List<Tuple<string, string>> expectedActionsOnPerson = new List<Tuple<string, string>>()
+                {
+                    new Tuple<string, string>(MetadataPrefix + updatePersonInfo, "People(-1)/" + updatePersonInfoName),
+                };
+                List<Tuple<string, string>> expectedActionsOnEmployee = new List<Tuple<string, string>>
```
habbes, line 103: I wonder whether using parameterized tests with `[Theory]` would make the test clearer and more readable.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/Program.cs
habbes, line 1: Should this file exist?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/Microsoft.OData.Client.E2E.Tests.csproj
```diff
@@ -0,0 +1,38 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <OutputType>Exe</OutputType>
```
habbes, line 4: Should the output type really be an executable given that this is a test project?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,266 @@
...
+using Microsoft.OData.Client.E2E.Tests.Common.Server.EndToEnd;
+using Microsoft.OData.Edm;
+using Microsoft.OData.Edm.Csdl;
+using Microsoft.OData.Edm.Validation;
+using Xunit;
+
+namespace Microsoft.OData.Client.E2E.Tests.ActionOverloadingTests.Tests
+{
+    [TestCaseOrderer("Microsoft.OData.Client.E2E.TestCommon.PriorityOrderer", "Microsoft.OData.Client.E2E.TestCommon")]
+    public class ActionOverloadingEndToEndTests : EndToEndTestBase<ActionOverloadingEndToEndTests.TestsStartup>
```
habbes, line 23: Why does the test order matter? Do the tests modify shared/static data?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,266 @@
...
+            _baseUri = new Uri(Client.BaseAddress, "odata/");
+            _context = new Container(_baseUri);
+            _context.HttpClientFactory = HttpClientFactory;
+        }
+
+        [Fact]
+        public void ExecuteOverloadedOperations()
+        {
+            var product = _context.Execute<Common.Clients.EndToEnd.Product>(new Uri(_baseUri + "Products(-10)", UriKind.Absolute)).Single();
+            OperationDescriptor productOperationDescriptor = _context.GetEntityDescriptor(product).OperationDescriptors.Single();
```
habbes, line 49: Maybe we should retrieve the descriptor by name instead of `Single()`, that way adding new operations to the model would not cause this test to fail. It would also make the intent clearer in my opinion.

habbes: If it's critical to this test that there only be exactly one overload of that operation, then maybe use Single with a filter against the name instead of an unfiltered `Single`.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,266 @@
...
+
+            Assert.Equal("http://localhost/odata/Products(-10)/Default.RetrieveProduct", productOperationDescriptor.Target.AbsoluteUri, true);
+            Assert.Equal("http://localhost/odata/$metadata#Default.RetrieveProduct", productOperationDescriptor.Metadata.AbsoluteUri, true);
+
+            Common.Clients.EndToEnd.OrderLine orderLine = _context.Execute<Common.Clients.EndToEnd.OrderLine>(new Uri(_baseUri + "OrderLines(OrderId=-10,ProductId=-10)", UriKind.Absolute)).Single();
+
+            OperationDescriptor orderLineOperationDescriptor = _context.GetEntityDescriptor(orderLine).OperationDescriptors.Single();
```
habbes, line 56: Same concerns as https://github.com/OData/odata.net/pull/3017/files#r1678985648

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,266 @@
...
+            Assert.Equal("http://localhost/odata/OrderLines(OrderId=-10,ProductId=-10)/Default.RetrieveProduct", orderLineOperationDescriptor.Target.AbsoluteUri, true);
+            Assert.Equal("http://localhost/odata/$metadata#Default.RetrieveProduct", orderLineOperationDescriptor.Metadata.AbsoluteUri, true);
+
+            // service operation
+            int serviceOperationResult = _context.Execute<int>(new Uri(_baseUri + "RetrieveProduct", UriKind.Absolute), "POST", true).Single();
+            Assert.Equal(-10, serviceOperationResult);
+
+            // actions
+            int productId = _context.Execute<int>(productOperationDescriptor.Target, "POST", true).Single();
```
habbes, line 65: What does the `true` parameter represent?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,266 @@
...
+            Assert.Equal(-10, productId);
+
+            int orderLineId = _context.Execute<int>(orderLineOperationDescriptor.Target, "POST", true).Single();
+            Assert.Equal(-10, orderLineId);
```
habbes, line 69: All these different overloads return the same thing. That makes it harder to confirm that different overloads were actually called on the service. Maybe you could have used different products with different IDs for a clearer demonstration.

buildandtest.yml
ElizabethOkerio: Yes

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
ElizabethOkerio: Yes. Some tests modify static data.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/Microsoft.OData.Client.E2E.Tests.csproj
ElizabethOkerio: Yes. Updated this. It should be a test project.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/Program.cs
ElizabethOkerio: removed.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,266 @@
...
+            _context.UpdatePersonInfo();
+        }
+
+        [Fact, TestPriority(1)]
+        public void ExcuteBoundAction()
+        {
+            _context.MergeOption = MergeOption.OverwriteChanges;
+
+            Common.Clients.EndToEnd.Employee employee = (Common.Clients.EndToEnd.Employee)_context.People.Where(p => p.PersonId == 0).Single();
+            Assert.Equal(85, employee.Salary);
```
habbes, line 89: Is this employee instance used anywhere else in this method? What's the point of fetching this instance?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,266 @@
...
+
+            DataServiceQuerySingle<Common.Clients.EndToEnd.Employee> singleEmployee = new DataServiceQuerySingle<Common.Clients.EndToEnd.Employee>(_context, "People(0)/Microsoft.OData.Client.E2E.Tests.Common.Server.EndToEnd.Employee");
+            singleEmployee.UpdatePersonInfo().Execute();
```
habbes, line 92: What does this statement do? How is the `singleEmployee` updated? Any assertions?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,266 @@
...
+            {
+                { metadataPrefix + actionName, "People(-1)/" + actionTitle },
+            };
+            this.VerifyLinks(expectedLinkValues, peopleDescriptors);
+            this.ExecuteActions(_context, peopleDescriptors);
+
+            // action bound with employee instance
+            var employee = _context.Execute<Common.Clients.EndToEnd.Employee>(new Uri(_baseUri + "People(0)", UriKind.Absolute)).Single();
+            IEnumerable<OperationDescriptor> employeeDescriptors = _context.GetEntityDescriptor(employee).OperationDescriptors.Where(od => od.Title.Equals(actionTitle));
+            List<Tuple<string, string>> expectedLinkValuesList = new List<Tuple<string, string>>
```
habbes, line 125:
```suggestion
            List<(string, string)> expectedLinkValuesList = new List<(string, string)>

```

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/PeopleController.cs
```diff
@@ -0,0 +1,124 @@
...
+        {
+            var people = CommonEndToEndDataSource.People;
+            return Ok(people);
+        }
+
+        [EnableQuery]
+        [HttpGet("odata/People({key})")]
+        public IActionResult Get(int key)
+        {
+            var person = CommonEndToEndDataSource.People?.FirstOrDefault(a => a.PersonId == key);
```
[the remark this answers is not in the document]
ElizabethOkerio: No. It cannot.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,266 @@
...
+                    new BodyOperationParameter("n", 3));
+            }
+        }
+
+        [Fact]
+        public void OverloadedActionsProjection()
+        {
+            for (int i = 1; i < 2; i++)
+            {
+                Common.Clients.EndToEnd.Product product = _context.Execute<Common.Clients.EndToEnd.Product>(new Uri(_baseUri + "Products(-10)?$select=Default.*", UriKind.Absolute)).Single();
```
habbes, line 213: What does select `Default.*` implies?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,266 @@
...
+                OperationDescriptor productOperationDescriptor = _context.GetEntityDescriptor(product).OperationDescriptors.Single();
+                Assert.Equal(_baseUri + "$metadata#Default.RetrieveProduct", productOperationDescriptor.Metadata.AbsoluteUri, true);
+                Assert.Equal(_baseUri + "Products(-10)/Default.RetrieveProduct", productOperationDescriptor.Target.AbsoluteUri, true);
+
+                Common.Clients.EndToEnd.OrderLine orderLine = _context.Execute<Common.Clients.EndToEnd.OrderLine>(new Uri(_baseUri + "OrderLines(OrderId=-10,ProductId=-10)?$select=*")).Single();
```
habbes, line 218: Maybe you could have scenarios where you select something other than `*`, like select specific property names?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,266 @@
...
+                    if (_baseUri + expected.Item2 == od.Target.AbsoluteUri)
+                    {
+                        matched = true;
+                        break;
+                    }
+                }
+
+                Assert.True(matched);
+            }
+        }
```
habbes, line 252: Are there tests where you test the return values of actions?

habbes (commented):
In my opinion, some of the tests are hard to read, in the sense that it's not always immediately clear what the test is trying to verify. I understand that many of the tests require a lot of setup code and boilerplate to get working, but I'm wondering whether there ways to make the tests easier to read, to make it clear in 1-2 passes what's going on.

Here are my overall suggestions:
- Avoid cramming multiple test scenarios into one when possible, maybe break them apart into separate test methods. Or consider parameterizing tests with `[Theory]` where applicable, instead of having multiple lists or loops of scenarios within a single test.
- Use descriptive test names. Or use doc descriptions that help explain the scenario being tested
- Some tests don't have assertions, making it harder to tell what is being tested or verified in that test.
- Use descriptive names helper methods that hide important test details (it was not clear to me what `VerifyLinks` does until I got to the implementation).
- Make tests robust: future changes to the data should not break unrelated tests. I think the extensive use of unfiltered `Single()` against the source data set could lead to maintainability issues. If tests depend on there being only instance of something, we should have ways to indicate that in the source data as well to avoid unintended consequences.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/PeopleController.cs
ElizabethOkerio: This is how it is declared: 

```c#
            builder.EntityType<SpecialEmployee>()
                .Action("IncreaseEmployeeSalary")
                .Returns<int>();

            builder.EntityType<Employee>()
                .Action("IncreaseEmployeeSalary")
                .Returns<bool>()
                .Parameter<int>("n");
```

ElizabethOkerio: For this particular test it is the second one.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
ElizabethOkerio: I think the main reason for maintaining some of these values is not to have to change the tests as much and to maintain them as they were before. I guess though we can improve this to make it clearer and simple to understand the intent of the tests.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
ElizabethOkerio: Yea. It's not used. Can get rid of it.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
ElizabethOkerio: ignoreCase

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
ElizabethOkerio: Most of these tests were written to show that the actions execute but I've added some asserts for the various actions.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingQueryTests.cs
ElizabethOkerio: I've tried to simplify these tests without introducing a lot of complexity.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingQueryTests.cs
ElizabethOkerio: These are placeholder values for the expected operations.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
```diff
@@ -0,0 +1,266 @@
...
+
+            int orderLineId = _context.Execute<int>(orderLineOperationDescriptor.Target, "POST", true).Single();
+            Assert.Equal(-10, orderLineId);
+        }
+
+        [Fact]
+        public void ExecuteOverloadedActionNotBound()
+        {
+            // unbound action
+            _context.Execute(new Uri(_baseUri + "Default.UpdatePersonInfo", UriKind.Absolute), "POST");
```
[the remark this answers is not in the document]
ElizabethOkerio: Added assertions.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingEndToEndTests.cs
ElizabethOkerio: It is a namespace.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingQueryTests.cs
ElizabethOkerio: These tests are meant to verify metadata of the specific operations defined in the service model. In this case it checks the metadata for `UpdatePersonInfo` and `RetrieveProduct`

ElizabethOkerio: This tests specifically checks whether the various operations are present in the model.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/PeopleController.cs
habbes: Doesn't seem like the unhandled exception when the entity has not been found has been addressed. Should return a 404 if the person is not found?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/ProductsController.cs
habbes: Why not return a 404 if the entity is not found?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingQueryTests.cs
```diff
@@ -0,0 +1,205 @@
...
+                "UpdatePersonInfo",
+                "RetrieveProduct"
+            };
+
+            var actionImports = _model.EntityContainer.OperationImports();
+
+            foreach (var operationName in expectedOperations)
+            {
+                var filteredImports = actionImports.Where(fi => fi.Name == operationName);
+                Assert.NotNull(filteredImports);
```
habbes, line 72: Even if `filteredImports` was empty, this assertion would not fail, because the return of `Where()` clause would be a non-null object, even if it represent an empty collection.

Maybe should use `Assert.True(actionImports.Any(fi => fi.Name == operationName)` if you want to use not null:

```c#
var import = actionImports.FirstOrDefault(fi => fi.Name == operationName);
Assert.NotNull(import);
```

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingQueryTests.cs
```diff
@@ -0,0 +1,205 @@
...
+            Assert.Equal(expectedActions.Count, entry.Actions.Count());
+            foreach (var expected in expectedActions)
+            {
+                var actions = entry.Actions.Where(a => a.Metadata.AbsoluteUri == _baseUri + expected.Item1);
+                bool matched = false;
+                foreach (var action in actions)
+                {
+                    if (_baseUri + expected.Item2 == action.Target.AbsoluteUri)
+                    {
+                        matched = true;
```
habbes, line 198: Nit: maybe break out of the inner loop once matched.
```suggestion
                        matched = true;
                        break;

```

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Tests/ActionOverloadingQueryTests.cs
```diff
@@ -0,0 +1,205 @@
...
+                    }
+                }
+                Assert.True(matched);
```
habbes, line 201: Nit: Since this assertion runs in a loop, if it were to fail, it would be hard to debug which action failed to match. Maybe you should add a custom error message to the assertion that specifies which action it is that failed to match, that would make debugging slightly easier.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/ProductsController.cs
ElizabethOkerio: fixed

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/OrderLinesController.cs
```diff
@@ -0,0 +1,63 @@
...
+
+        [EnableQuery]
+        public IActionResult Get()
+        {
+            var orderLines = CommonEndToEndDataSource.OrderLines;
+
+            if (orderLines == null)
+            {
+                return NotFound();
+            }
```
habbes, line 30: We don't need to return not found for `GET /entityset` requests, we should just return an empty collection if there are not items. Under what circumstances would the `CommonEndToEndDataSource.OrderLines` property be null?

habbes (approved):

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/OrderLinesController.cs
ElizabethOkerio: I'll update this in the next PR.

PR #3023, opened 2024-07-25
Fix for Error occurs when you check whether a string or integer literal is in an enum collection

### LINKED ISSUE #2371: Error occurs when you check whether a string literal is in an enum collection
Related to https://github.com/OData/AspNetCoreOData/issues/536

The `in` operator does not allow you to use a string literal as an operand when comparing against a collection of enums.
For example, let's assume `WorkingDays` property is a collection of enums, the following would fail with an error:

```
$filter='Monday' in WorkingDays
```

For it to work, you have to use the fully qualified enum literal:

```
$filter=ODataRoutingSample.Models.WorkingDay'Monday' in WorkingDays
```

However, binary operators like `eq` do allow comparisons between string literals and enums, so the following does work:

```
$filter=WorkingDays/any(t:t eq 'Monday')
```

I think both approaches should work. According to the [spec](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_PrimitiveLiterals):

      Enumeration literals in OData 4.0 required prefixing with the qualified type name of the enumeration.

      In OData 4.01, services MUST support duration and enumeration literals with or without the type prefix.
      OData clients that want to operate across OData 4.0 and OData 4.01 services should always include the prefix for duration and enumeration types.

### Assemblies affected

Microsoft.OData.Core 7.10.0

### Reproduce steps

Create a service that exposes an entity set with a property that contains a collection of enums, e.g.: `WorkingDays` that contains enum values: `WorkingDay.Monday`, `WorkingDay.Tueday`, etc.

Make a query against the service with the filter expression: `$filter='Monday' in WorkingDays`.

See [this related issue](https://github.com/OData/AspNetCoreOData/issues/536) for a more detailed repro example.

### Expected result

The request should return the expected filtered data, same to what you would get with: `$filter=WorkingDays/any(t:t eq 'Monday')`

### Actual result

The following error is thrown:
```
System.ArgumentException: An instance of InNode can only be created where the item types of the right operand 'ODataRoutingSample.Models.WorkingDay' and the left operand 'Edm.String' can be compared.
```

### Additional detail

For binary operators, the [`BinaryOperatorBinder` promotes the operands to compatible types](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/UriParser/Binders/BinaryOperatorBinder.cs#L52), specifically, strings are promoted to enums where applicable: 
```c#
internal QueryNode BindBinaryOperator(BinaryOperatorToken binaryOperatorToken)
{
    ExceptionUtils.CheckArgumentNotNull(binaryOperatorToken, "binaryOperatorToken");

    SingleValueNode left = this.GetOperandFromToken(binaryOperatorToken.OperatorKind, binaryOperatorToken.Left);
    SingleValueNode right = this.GetOperandFromToken(binaryOperatorToken.OperatorKind, binaryOperatorToken.Right);

    IEdmTypeReference typeReference;
    this.resolver.PromoteBinaryOperandTypes(binaryOperatorToken.OperatorKind, ref left, ref right, out typeReference);

    return new BinaryOperatorNode(binaryOperatorToken.OperatorKind, left, right, typeReference);
}
```
The [`InBinder`](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs#L44) does not promote types. The [`InNode` throws an exception](https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/UriParser/SemanticAst/InNode.cs#L58) if the type of one operand is not assignable to the other.

WanjohiSammy:
### Issues

*This pull request fixes #2371.*

### Description

This change is to fix the issue raised where the `in` operator does not allow using a `string` or `integer` literal as left operand when comparing against a collection of enums.

The change involves:
- Checking if the` left operand` is either an `integral` or a `string` type and the `right operand` is a collection of enums. 
- If this is true, it calls `MetadataBindingUtils.ConvertToTypeIfNeeded()` method to try convert the left operand to the same enum type if possible. 
- If this is not possible, exception is thrown.

Assume you have the following entity:
```cs
public class Employee 
{
    public int Id { get; set; }
    public List<WeekDay> WorkingDays { get; set; }
}

public enum WeekDay
{
    Monday = 1,
    Tuesday = 2,
    Wednesday = 3,
    Thursday = 4,
    Friday = 5,
    Saturday = 6,
    Sunday = 7
}
```
This change enables support of the following scenarios for `$filter` with `In` Operator:

```cs
// left operand without fully qualified namespace
/Employees?$filter='Monday' in WorkingDays

// left operand with integral value in single quotes
/Employees?$filter='1' in WorkingDays

// left operand with integral value without single quotes
/Employees?$filter=1 in WorkingDays
```

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -60,7 +60,7 @@ namespace Microsoft.OData.UriParser
 
             // If the left operand is either an integral or a string type and the right operand is a collection of enums,
             // Calls the MetadataBindingUtils.ConvertToTypeIfNeeded() method to convert the left operand to the same enum type as the right operand.
-            if ((right is CollectionPropertyAccessNode && right.ItemType.IsEnum()) && (left.TypeReference.IsString() || left.TypeReference.IsIntegral()))
+            if ((right is CollectionPropertyAccessNode && right.ItemType.IsEnum()) && (left.TypeReference != null && (left.TypeReference.IsString() || left.TypeReference.IsIntegral())))
             {
                 left = MetadataBindingUtils.ConvertToTypeIfNeeded(left, right.ItemType);
             }
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -58,14 +58,25 @@ namespace Microsoft.OData.UriParser
                     inToken.Right, new EdmCollectionTypeReference(new EdmCollectionType(EdmCoreModel.Instance.GetUntyped())), state.Model);
             }
 
-            // If the left operand is either an integral or a string type and the right operand is a collection of enums,
-            // Calls the MetadataBindingUtils.ConvertToTypeIfNeeded() method to convert the left operand to the same enum type as the right operand.
-            if ((right is CollectionPropertyAccessNode && right.ItemType.IsEnum()) && (left.TypeReference.IsString() || left.TypeReference.IsIntegral()))
+            if (ShouldConvertLeftOperand(left, right))
             {
                 left = MetadataBindingUtils.ConvertToTypeIfNeeded(left, right.ItemType);
             }
 
             return new InNode(left, right);
+        }
+
+        /// <summary>
+        /// Checks if the left operand is either an integral or a string type and the right operand is a collection of enums.
+        /// </summary>
+        /// <param name="leftNode">The left operand.</param>
+        /// <param name="rightNode">The right operand.</param>
+        /// <returns>True if the condition is met, otherwise false.</returns>
+        private static bool ShouldConvertLeftOperand(SingleValueNode leftNode, CollectionNode rightNode)
+        {
+            // If the left operand is either an integral or a string type and the right operand is a collection of enums,
+            // Calls the MetadataBindingUtils.ConvertToTypeIfNeeded() method to convert the left operand to the same enum type as the right operand.
+            return (rightNode is CollectionPropertyAccessNode && rightNode.ItemType.IsEnum()) && (leftNode.TypeReference.IsString() || leftNode.TypeReference.IsIntegral());
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -58,6 +58,13 @@ namespace Microsoft.OData.UriParser
                     inToken.Right, new EdmCollectionTypeReference(new EdmCollectionType(EdmCoreModel.Instance.GetUntyped())), state.Model);
             }
 
+            // If the left operand is either an integral or a string type and the right operand is a collection of enums,
+            // Calls the MetadataBindingUtils.ConvertToTypeIfNeeded() method to convert the left operand to the same enum type as the right operand.
+            if ((right is CollectionPropertyAccessNode && right.ItemType.IsEnum()) && (left.TypeReference.IsString() || left.TypeReference.IsIntegral()))
+            {
+                left = MetadataBindingUtils.ConvertToTypeIfNeeded(left, right.ItemType);
+            }
+
             return new InNode(left, right);
         }
 
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -58,25 +58,14 @@ namespace Microsoft.OData.UriParser
                     inToken.Right, new EdmCollectionTypeReference(new EdmCollectionType(EdmCoreModel.Instance.GetUntyped())), state.Model);
             }
 
-            if (ShouldConvertLeftOperand(left, right))
+            // If the left operand is either an integral or a string type and the right operand is a collection of enums,
+            // Calls the MetadataBindingUtils.ConvertToTypeIfNeeded() method to convert the left operand to the same enum type as the right operand.
+            if ((right is CollectionPropertyAccessNode && right.ItemType.IsEnum()) && (left.TypeReference.IsString() || left.TypeReference.IsIntegral()))
             {
                 left = MetadataBindingUtils.ConvertToTypeIfNeeded(left, right.ItemType);
             }
 
             return new InNode(left, right);
-        }
-
-        /// <summary>
-        /// Checks if the left operand is either an integral or a string type and the right operand is a collection of enums.
-        /// </summary>
-        /// <param name="leftNode">The left operand.</param>
-        /// <param name="rightNode">The right operand.</param>
-        /// <returns>True if the condition is met, otherwise false.</returns>
-        private static bool ShouldConvertLeftOperand(SingleValueNode leftNode, CollectionNode rightNode)
-        {
-            // If the left operand is either an integral or a string type and the right operand is a collection of enums,
-            // Calls the MetadataBindingUtils.ConvertToTypeIfNeeded() method to convert the left operand to the same enum type as the right operand.
-            return (rightNode is CollectionPropertyAccessNode && rightNode.ItemType.IsEnum()) && (leftNode.TypeReference.IsString() || leftNode.TypeReference.IsIntegral());
         }
 
         /// <summary>
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
@@ -99,11 +99,11 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
 
             binaryNode
                 .Left
-                .ShouldBeSingleValuePropertyAccessQueryNode(GetColorProp(this.userModel));
+                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("Color"));
 
             binaryNode
                 .Right
-                .ShouldBeEnumNode(this.GetColorType(this.userModel), (int)Color.Green);
+                .ShouldBeEnumNode(this.GetIEdmType<IEdmEnumType>("NS.Color"), (int)Color.Green);
         }
 
         [Fact]
@@ -121,11 +121,11 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
 
             binaryNode
                 .Left
-                .ShouldBeSingleValuePropertyAccessQueryNode(GetColorProp(this.userModel));
+                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("Color"));
 
             binaryNode
                 .Right
-                .ShouldBeEnumNode(this.GetColorType(this.userModel), (int)Color.Green);
+                .ShouldBeEnumNode(this.GetIEdmType<IEdmEnumType>("NS.Color"), (int)Color.Green);
         }
 
         [Fact]
@@ -143,11 +143,11 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
 
             binaryNode
                 .Left
-                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetColorFlagsProp(this.userModel));
+                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("ColorFlags"));
 
             binaryNode
                 .Right
-                .ShouldBeEnumNode(this.GetColorFlagsType(this.userModel), (int)ColorFlags.Green);
+                .ShouldBeEnumNode(this.GetIEdmType<IEdmEnumType>("NS.ColorFlags"), (int)ColorFlags.Green);
         }
 
         [Fact]
@@ -165,11 +165,11 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
 
             binaryNode
                 .Left
-                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetColorFlagsProp(this.userModel));
+                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("ColorFlags"));
 
             binaryNode
                 .Right
-                .ShouldBeEnumNode(this.GetColorFlagsType(this.userModel), (int)ColorFlags.Green);
+                .ShouldBeEnumNode(this.GetIEdmType<IEdmEnumType>("NS.ColorFlags"), (int)ColorFlags.Green);
         }
 
         [Fact]
@@ -187,11 +187,11 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
 
             binaryNode
             .Left
-            .ShouldBeEnumNode(this.GetColorFlagsType(this.userModel), (int)(ColorFlags.Green | ColorFlags.Red));
+            .ShouldBeEnumNode(this.GetIEdmType<IEdmEnumType>("NS.ColorFlags"), (int)(ColorFlags.Green | ColorFlags.Red));
 
             binaryNode
             .Right
-            .ShouldBeSingleValuePropertyAccessQueryNode(this.GetColorFlagsProp(this.userModel));
+            .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("ColorFlags"));
         }
 
         [Fact]
@@ -209,11 +209,11 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
 
             binaryNode
                 .Left
-                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetColorProp(this.userModel));
+                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("Color"));
 
             binaryNode
                 .Right
-                .ShouldBeEnumNode(this.GetColorType(this.userModel), (int)Color.Green);
+                .ShouldBeEnumNode(this.GetIEdmType<IEdmEnumType>("NS.Color"), (int)Color.Green);
         }
 
         [Fact]
@@ -231,12 +231,12 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
 
             binaryNode
                 .Left
-                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetColorFlagsProp(this.userModel));
+                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("ColorFlags"));
 
             binaryNode
                 .Right
                 .ShouldBeEnumNode(
-                this.GetColorFlagsType(this.userModel),
+                this.GetIEdmType<IEdmEnumType>("NS.ColorFlags"),
                 (int)(ColorFlags.Green | ColorFlags.Red));
         }
 
@@ -255,12 +255,12 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
 
             binaryNode
                 .Left
-                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetColorFlagsProp(this.userModel));
+                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("ColorFlags"));
 
             binaryNode
                 .Right
                 .ShouldBeEnumNode(
-                this.GetColorFlagsType(this.userModel),
+                this.GetIEdmType<IEdmEnumType>("NS.ColorFlags"),
                 (int)(ColorFlags.Green | ColorFlags.Red));
         }
 
@@ -279,12 +279,12 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
 
             binaryNode
                 .Left
-                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetColorFlagsProp(this.userModel));
+                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("ColorFlags"));
 
             binaryNode
                  .Right
                  .ShouldBeEnumNode(
-                 this.GetColorFlagsType(this.userModel),
+                 this.GetIEdmType<IEdmEnumType>("NS.ColorFlags"),
                  "Red");
         }
 
@@ -303,12 +303,12 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
 
             binaryNode
                 .Left
-                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetColorFlagsProp(this.userModel));
+                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("ColorFlags"));
 
             binaryNode
                 .Right
                 .ShouldBeEnumNode(
-                this.GetColorFlagsType(this.userModel),
+                this.GetIEdmType<IEdmEnumType>("NS.ColorFlags"),
                 (int)(ColorFlags.Green | ColorFlags.Red));
         }
 
@@ -327,12 +327,12 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
 
             binaryNode
                 .Left
-                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetColorFlagsProp(this.userModel));
+                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("ColorFlags"));
 
             binaryNode
                 .Right
                 .ShouldBeEnumNode(
-                this.GetColorFlagsType(this.userModel),
+                this.GetIEdmType<IEdmEnumType>("NS.ColorFlags"),
                 (int)(ColorFlags.Green | ColorFlags.Red));
         }
 
@@ -351,12 +351,12 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
 
             binaryNode
                 .Left
-                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetColorProp(this.userModel));
+                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("Color"));
 
             binaryNode
                 .Right
                 .ShouldBeEnumNode(
-                this.GetColorType(this.userModel),
+                this.GetIEdmType<IEdmEnumType>("NS.Color"),
                 (int)(Color.White));
 
             var constantNode = Assert.IsType<ConstantNode>(binaryNode.Right);
@@ -379,12 +379,12 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
 
             binaryNode
                 .Left
-                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetColorProp(this.userModel));
+                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("Color"));
 
             binaryNode
                 .Right
                 .ShouldBeEnumNode(
-                this.GetColorType(this.userModel),
+                this.GetIEdmType<IEdmEnumType>("NS.Color"),
                 -132534290);
 
             var constantNode = Assert.IsType<ConstantNode>(binaryNode.Right);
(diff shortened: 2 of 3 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -10,6 +10,7 @@
 using Microsoft.OData.UriParser;
 using Microsoft.OData.Edm;
 using Xunit;
+using System.Linq;
```
KenitoInc, line 13: `System.X` namespaces should always be at the top.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -515,13 +520,205 @@ public void ParseFilterEnumTypesWrongCast3()
...
+
+            // Assert
+            test.Throws<ArgumentException>(Strings.Nodes_InNode_CollectionItemTypeMustBeSameAsSingleItemType("NS.Color", "Edm.Single")); // Float are of Type Edm.Single
+        }
+
+        [Theory]
+        [InlineData(-20)]
+        [InlineData("-20")]
+        [InlineData("'-20'")]
+        public void ParseFilterWithInOperatorWithEnumsMemberInvalidIntegralValue(object integralValue)
```
KenitoInc, line 696: Add a test to validate that an error is thrown if we pass a string that doesn't exist
E.g `$filter='Blue10' in Colors`

WanjohiSammy: Thanks @KenitoInc for this. I have added the tests. Please see using the link below:

- [ParseFilterWithInOperatorWithEnumsInvalidMemberNames_ThrowsIsNotValidEnumConstantException](https://github.com/OData/odata.net/pull/3023/files#diff-9f67b5b315ba1bb4ddbcb823f5c4183658c6a7f7a3b06d0da900b0ea7e21c7d6:~:text=ParseFilterWithInOperatorWithEnumsInvalidMemberNames_ThrowsIsNotValidEnumConstantException)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
WanjohiSammy: @KenitoInc : This has been moved to the top

ElizabethOkerio:
// left operand with fully qualified namespace
/Employees?$filter=Fully.Qualified.Namespace'Monday' in WorkingDays

// left operand without fully qualified namespace
/Employees?$filter='Monday' in WorkingDays

// left operand with integral value in single quotes
/Employees?$filter='1' in WorkingDays

// left operand with integral value without single quotes
/Employees?$filter=1 in WorkingDays

From the issue description, the first and last scenario is currently supported..is that true?

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -58,9 +58,27 @@ internal QueryNode BindInOperator(InToken inToken, BindingState state)
...
             return new InNode(left, right);
         }
 
+        /// <summary>
+        /// Checks if the left operand is either an integral or a string type and the right operand is a collection of enums.
+        /// </summary>
+        /// <param name="leftNode">The left operand.</param>
+        /// <param name="rightNode">The right operand.</param>
+        /// <returns>True if the condition is met, otherwise false.</returns>
+        private static bool ShouldConvertLeftOperand(SingleValueNode leftNode, CollectionNode rightNode)
```
ElizabethOkerio, line 75: We currently have a method that promotes types to compatible types, why can't we use it instead of adding this method to do this check? Can the check be added to that method if need be?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -94,11 +99,11 @@ public void ParseFilterWithEnum()
 
             binaryNode
                 .Left
-                .ShouldBeSingleValuePropertyAccessQueryNode(GetColorProp(this.userModel));
+                .ShouldBeSingleValuePropertyAccessQueryNode(this.GetIEdmProperty("Color"));
```
ElizabethOkerio, line 102: Why are we changing these tests? Should we add new tests to show the new behavior, or the tests have to change?

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
WanjohiSammy: @ElizabethOkerio This is just a private method that checks for the condition before calling `MetadataBindingUtils.ConvertToTypeIfNeeded` which is the existing method that does the promotion. It was for readability purposes. 

```cs
if (ShouldConvertLeftOperand(left, right))
{
    left = MetadataBindingUtils.ConvertToTypeIfNeeded(left, right.ItemType);
}
```

where:
```cs
private static bool ShouldConvertLeftOperand(SingleValueNode leftNode, CollectionNode rightNode)
{
    // If the left operand is either an integral or a string type and the right operand is a collection of enums,
    // Calls the MetadataBindingUtils.ConvertToTypeIfNeeded() method to convert the left operand to the same enum type as the right operand.
    return (rightNode is CollectionPropertyAccessNode && rightNode.ItemType.IsEnum()) && (leftNode.TypeReference.IsString() || leftNode.TypeReference.IsIntegral());
}
```

If not necessary, I can remove the ` bool ShouldConvertLeftOperand(SingleValueNode leftNode, CollectionNode rightNode)' and put the condition as below:
```cs
if ((rightNode is CollectionPropertyAccessNode && rightNode.ItemType.IsEnum()) && (leftNode.TypeReference.IsString() || leftNode.TypeReference.IsIntegral()))
{
    left = MetadataBindingUtils.ConvertToTypeIfNeeded(left, right.ItemType);
}
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
WanjohiSammy: @ElizabethOkerio Thanks for bringing this up.

I only change the private methods for getting IEdmProperty and IEdmType from IEdmModel to make them reused across all the tests. I noted a lot of code repetition. For example, if I want to get Person property from the IEdmModel, I am forced to create `GetPersonProp` and the same way if I want to get Person type, I have to create `GetPersonType`.

I have created one method `private IEdmStructuralProperty GetIEdmProperty(string propertyName)` to replace the following methods so that in future you do not need to add another method to achieve the same thing. The only thing you need to do is to pass propertyName to `GetIEdmProperty()` and you are good to go.
```cs
private IEdmStructuralProperty GetColorProp(IEdmModel model) { }
private IEdmStructuralProperty GetColorFlagsProp(IEdmModel model) { }
```

Also instead of many methods for getting different IEdmTypes, I have created one method `private T GetIEdmType<T>(string typeName) where T : IEdmType` to replace the following many methods so that in future if you want to get IEdmType of another entity, you just need to provide typeName and the type to the `GetIEdmType<T>()` method instead of adding another method:
```cs
private IEdmEnumType GetColorFlagsType(IEdmModel model)
private IEdmEnumType GetColorType(IEdmModel model)
```

WanjohiSammy:
> // left operand with fully qualified namespace /Employees?$filter=Fully.Qualified.Namespace'Monday' in WorkingDays
> 
> // left operand without fully qualified namespace /Employees?$filter='Monday' in WorkingDays
> 
> // left operand with integral value in single quotes /Employees?$filter='1' in WorkingDays
> 
> // left operand with integral value without single quotes /Employees?$filter=1 in WorkingDays
> 
> From the issue description, the first and last scenarios are currently supported..is that true? What this PR is supporting is the second and third scenario?

@ElizabethOkerio Currently, we are supporting only the 1st scenario. This PR is to support `2nd, 3rd and 4th` scenarios. I have edited the description.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -515,28 +520,245 @@ public void ParseFilterEnumTypesWrongCast3()
...
+
+            IEdmEnumType colorType = this.GetIEdmType<IEdmEnumType>("NS.Color");
+            bool success = colorType.TryParse(enumValue, out IEdmEnumMember enumMember);
+
+            // Act
+            FilterClause filterQueryNode = ParseFilter(filterQuery, this.userModel, this.entityType, this.entitySet);
+            SingleValueNode expression = filterQueryNode.Expression;
+
+            // Assert
+            Assert.Equal("3 in Colors", filterQuery);
```
habbes, line 629: See: https://github.com/OData/odata.net/pull/3023/files#r1716585909

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -515,28 +520,245 @@ public void ParseFilterEnumTypesWrongCast3()
...
```
habbes, line 660: See: https://github.com/OData/odata.net/pull/3023/files#r1716585909

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -515,28 +520,245 @@ public void ParseFilterEnumTypesWrongCast3()
...
+            Assert.Equal(enumMember.Name, leftNode.Value.ToString());
+            Assert.Equal(expectedLiteral, leftNode.LiteralText);
+
+            CollectionPropertyAccessNode rightNode = Assert.IsType<CollectionPropertyAccessNode>(inNode.Right);
+            rightNode.ShouldBeCollectionPropertyAccessQueryNode(this.GetIEdmProperty("Colors"));
+            Assert.Equal(colorType, rightNode.ItemType.Definition);
+        }
+
+        [Fact]
+        public void ParseFilterWithInOperatorWithEnumsMemberDoubleIntegralValue()
```
habbes, line 645: What do you mean by `DoubleIntegralValues`? How is this test different from this: https://github.com/OData/odata.net/pull/3023/files#diff-9f67b5b315ba1bb4ddbcb823f5c4183658c6a7f7a3b06d0da900b0ea7e21c7d6R614

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -515,28 +520,245 @@ public void ParseFilterEnumTypesWrongCast3()
...
+            // Arrange
+            string filterQuery = "Red in Colors";
+
+            string erroMessage = "Object reference not set to an instance of an object.";
+
+            // Act
+            Action action = () => ParseFilter(filterQuery, this.userModel, this.entityType, this.entitySet);
+
+            // Assert
+            action.Throws<NullReferenceException>(erroMessage);
```
habbes, line 720: Is the cause of the exception the fact that `Red` in the `filterQuery` is not in single quotes? If so, why does it throw a null reference exception? Is that by design or a bug? Should it throw a friendlier error message that states what the error is? (e.g. `Red` is not a known property or is not of the correct type or whatever the appropriate error should be)?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -515,28 +520,245 @@ public void ParseFilterEnumTypesWrongCast3()
...
+        }
+
+        [Theory]
+        [InlineData("'Yellow'")]
+        [InlineData("'Teal'")]
+        [InlineData("NS.Color'Yellow'")]
+        [InlineData("NS.Color'Teal'")]
+        public void ParseFilterWithInOperatorWithEnumsInvalidMemberNames_ThrowsIsNotValidEnumConstantException(string memberName)
```
habbes, line 728: Surprised if we didn't test such scenarios before.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -515,28 +520,245 @@ public void ParseFilterEnumTypesWrongCast3()
...
+            return (T)model.FindType(typeName);
         }
 
-        private IEdmStructuralProperty GetColorFlagsProp(IEdmModel model)
+        private IEdmStructuralProperty GetIEdmProperty(string propertyName)
         {
-            return (IEdmStructuralProperty)((IEdmStructuredType)model
-                .FindType("NS.MyEntityType"))
-                .FindProperty("ColorFlags");
+            return (IEdmStructuralProperty)this.GetIEdmType<IEdmStructuredType>("NS.MyEntityType")
```
habbes, line 778: Does this have model have a single entity type?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -515,28 +520,245 @@ public void ParseFilterEnumTypesWrongCast3()
...
+                .FindProperty(propertyName);
         }
 
-        private IEdmEnumType GetColorFlagsType(IEdmModel model)
+        private IEdmStructuralProperty GetIEdmProperty(string entityName, string propertyName)
```
habbes, line 782: Could also use an optional param with a default value instead of overloads. I think that would make it easier for the function caller to see what the default is?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
WanjohiSammy: @habbes Thanks for bringing this up. This was happening because the `EdmEntityType` for testing was set to be open. In case where isOpen is false, the exception is thrown: "`Could not find a property named '{1}' on type '{0}'.`"

```cs
internal SingleValueOpenPropertyAccessNode GeneratePropertyAccessQueryForOpenType(
EndPathToken endPathToken, SingleValueNode parentNode)
{
    if (parentNode.TypeReference == null ||
        parentNode.TypeReference.Definition.IsOpen() ||
        IsAggregatedProperty(endPathToken))
    {
        return new SingleValueOpenPropertyAccessNode(parentNode, endPathToken.Identifier);
    }
    else
    {
        throw new ODataException(ODataErrorStrings.MetadataBinder_PropertyNotDeclared(
            parentNode.TypeReference.FullName(),
            endPathToken.Identifier));
    }
}
```

I have modified the `isOpen` to `false`
```cs
this.entityType = new EdmEntityType("NS", "MyEntityType", isAbstract: false, isOpen: false, baseType: null);
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
WanjohiSammy: @habbes: I meant to test for `long` type.
I have removed this test as it is similar to https://github.com/OData/odata.net/pull/3023/files#diff-9f67b5b315ba1bb4ddbcb823f5c4183658c6a7f7a3b06d0da900b0ea7e21c7d6R614

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
WanjohiSammy: @habbes Yes, the model already existed in this test has a single entity type
```cs
this.entityType = new EdmEntityType("NS", "MyEntityType", isAbstract: false, isOpen: false, baseType: null);
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
WanjohiSammy: @habbes The tests that are there are for the `has` operator.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -515,28 +520,207 @@ public void ParseFilterEnumTypesWrongCast3()
...
+            leftNode.ShouldBeEnumNode(colorType, enumMember.Name);
+
+            Assert.True(leftNode.TypeReference.IsEnum());
+            Assert.Equal(enumMember.Name, leftNode.Value.ToString());
+            Assert.Equal(expectedLiteral, leftNode.LiteralText);
+
+            CollectionPropertyAccessNode rightNode = Assert.IsType<CollectionPropertyAccessNode>(inNode.Right);
+            rightNode.ShouldBeCollectionPropertyAccessQueryNode(this.GetIEdmProperty("Colors"));
+            Assert.Equal(colorType, rightNode.ItemType.Definition);
+        }
```
gathogojr, line 637: You could instead use a single parameterized test that verifies that expressing the filter in the different forms results into the same outcome, i.e.:

```csharp
[Theory]
[InlineData("NS.Color'Green'")]
[InlineData("'Green'")]
[InlineData("'2'")]
[InlineData("2")]
public void ParseFilterWithInOperatorWithEnums(string filterOptionValue)
{
    // Arrange
    string filterQuery = $"{filterOptionValue} in Colors";

    // ...
}
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -515,28 +520,207 @@ public void ParseFilterEnumTypesWrongCast3()
...
+
+        [Theory]
+        [InlineData("3.0")]
+        [InlineData("3.1")]
+        [InlineData("4.0")]
+        [InlineData("5.0")]
```
gathogojr, line 643: I'd think one of these is sufficient.
What you could add is a float as a string - `'3.0'`, i.e., the following should suffice

```csharp
[InlineData("3.0")]
[InlineData("'3.0'")]
```

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -515,28 +520,207 @@ public void ParseFilterEnumTypesWrongCast3()
...
+            // Act
+            Action test = () => ParseFilter(filterQuery, this.userModel, this.entityType, this.entitySet);
+
+            // Assert
+            test.Throws<ArgumentException>(Strings.Nodes_InNode_CollectionItemTypeMustBeSameAsSingleItemType("NS.Color", "Edm.Single")); // Float are of Type Edm.Single
+        }
+
+        [Theory]
+        [InlineData(-20)]
+        [InlineData("-20")]
```
gathogojr, line 658: There's no difference between these two when they're applied to `$"{integralValue} in Colors"`. You should get rid of the first and change the parameter type to `int`

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -2041,6 +2041,108 @@ public void FilterWithInOperationWithEnums()
...
+            ODataEnumValue enumValue = Assert.IsType<ODataEnumValue>(left.Value);
+
+            Assert.Equal(expectedLiteralText, left.LiteralText);
+            Assert.Equal(expectedfullyQualifiedName, (enumValue.TypeName + left.LiteralText));
+            Assert.Equal(expectedCollectionPropertyName, Assert.IsType<CollectionPropertyAccessNode>(inNode.Right).Property.Name);
+        }
+
+        [Theory]
+        [InlineData(53)]
+        [InlineData("53")]
```
gathogojr, line 2115: See my comment here https://github.com/OData/odata.net/pull/3023/files#r1718016255

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -2041,6 +2041,108 @@ public void FilterWithInOperationWithEnums()
...
+
+            // Assert
+            InNode inNode = Assert.IsType<InNode>(filter.Expression);
+            ConstantNode left = Assert.IsType<ConstantNode>(inNode.Left);
+            ODataEnumValue enumValue = Assert.IsType<ODataEnumValue>(left.Value);
+
+            Assert.Equal(expectedLiteralText, left.LiteralText);
+            Assert.Equal(expectedfullyQualifiedName, (enumValue.TypeName + left.LiteralText));
+            Assert.Equal(expectedCollectionPropertyName, Assert.IsType<CollectionPropertyAccessNode>(inNode.Right).Property.Name);
+        }
```
gathogojr, line 2088: You could use a single parameterized test here, i.e.:

[Theory]
[InlineData("'SolidYellow'")]
[InlineData("'12'")] // Or the equivalent integral value for SolidYellow
public void FilterWithInOperationWithEnumsMemberIntegralValue(string filterOptionValue)
{
    // ...
}

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
WanjohiSammy: @gathogojr The exception will be different.
```js
[InlineData("3.0")] //Throws<ArgumentException>(Strings.Nodes_InNode_CollectionItemTypeMustBeSameAsSingleItemType("NS.Color", "Edm.Single")); // Float are of Type Edm.Single

[InlineData("'3.0'")] // Throws<ODataException>(Strings.Binder_IsNotValidEnumConstant("3.0"))
```

I have moved `[InlineData("'3.0'")]` to the test [ParseFilterWithInOperatorWithEnumsInvalidIntegralValues_ThrowsIsNotValidEnumConstantException](https://github.com/OData/odata.net/blob/fix/2371-allow-filter-stringliterals-in-enum-collection/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs#L583)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
WanjohiSammy: @gathogojr Thanks for pointing this out. I have resolve this and push the changes. [test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs#L583](https://github.com/OData/odata.net/blob/fix/2371-allow-filter-stringliterals-in-enum-collection/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs#L583)

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -2041,6 +2041,64 @@ public void FilterWithInOperationWithEnums()
...
+
+        [Theory]
+        [InlineData("53")]
+        [InlineData("'53'")]
+        public void FilterWithInOperationWithEnumsInvalidMemberIntegralValue_ThrowsIsNotValidEnumConstantException(object integralValue)
```
gathogojr, line 2073: nit:
```suggestion
        public void FilterWithInOperationWithEnumsInvalidMemberIntegralValue_ThrowsIsNotValidEnumConstantException(string integralValue)
```

gathogojr (approved):

gathogojr (approved):

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
habbes: Okay that makes sense. But it seems like there are two issues. This has not addressed the issue when we have an open type, which I assume would still throw a null-reference exception. I'd still be curious about that, because it's possible that customers are using open types. I guess the challenge is that we may not know the type. But regardless, if it's an error, it should throw a meaningful error, not a null-reference exception.

habbes (approved):

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
WanjohiSammy: @habbes That seems like an existing issue. Even if we remove this fix, we will still get the same error. If you pass `Red` without single quotes, it will be treated like a property and hence the error when trying to check if `left node` can be assignableFrom `right node`. This is happening because the left node does not exist as a property and therefore has a `null` `EdmTypeReference`:

This image shows where the error is thrown:

image

I found that this is where the issue is in the case of `In Operator`: 
```js
if (!this.left.GetEdmTypeReference().IsAssignableFrom(this.right.ItemType) &&
!this.right.ItemType.IsAssignableFrom(this.left.GetEdmTypeReference()))
```
In the above code, `this.left.GetEdmTypeReference()` is always null for this example: `?$filter=Red in Colors`.

```cs
public InNode(SingleValueNode left, CollectionNode right)
{
    ExceptionUtils.CheckArgumentNotNull(left, "left");
    ExceptionUtils.CheckArgumentNotNull(right, "right");
    this.left = left;
    this.right = right;

    if (!right.GetEdmTypeReference().IsUntyped())
    {
        if (!this.left.GetEdmTypeReference().IsAssignableFrom(this.right.ItemType) &&
            !this.right.ItemType.IsAssignableFrom(this.left.GetEdmTypeReference()))
        {
            throw new ArgumentException(ODataErrorStrings.Nodes_InNode_CollectionItemTypeMustBeSameAsSingleItemType(
                this.right.ItemType.FullName(), this.left.GetEdmTypeReference().FullName()));
        }
    }
}
```

WanjohiSammy: @habbes Unless the spec permit, we can try to convert the left operand `Red` to the same enum type if possible. Hence, to support the following. But this will only work if the `EdmEntityType` is `open`
```js
?$filter=Red in Colors
```

habbes: Since this PR was specific about the literals, may no need to add support for type coercions of non-literals. I was just concerned about the error message.

PR #3025, opened 2024-07-25
Replace 'IsType' using 'IsOf'

xuzhg:
OData spec is using 'IsOf' not 'IsType' as below

14.4.8 Is-Of
The is-of expression checks whether the value obtained from its single child expression is compatible with the specified type. It returns true if the child expression returns a type that is compatible with the specified type, and false otherwise.

Expression edm:IsOf

The edm:IsOf element MUST contain the Type attribute and MUST contain exactly one child expression.

It MAY contain edm:Annotation elements.

Attribute Type

The value of Type is the qualified name of a type in scope, or the character sequence Collection( followed by the qualified name of a type in scope, followed by a closing parenthesis ).

If the specified type is a primitive type or a collection of a primitive type, the facet attributes MaxLength, Unicode, Precision, Scale, and SRID MAY be specified if applicable to the specified primitive type. If the facet attributes are not specified, their values are considered unspecified.

Example 81:

```xml
<Annotation Term="self.IsPreferredCustomer">
  <IsOf Type="self.PreferredCustomer">
    <Path>Customer</Path>
  </IsOf>
</Annotation>
```

OData.Edm has the wrong definition. This PR is to fix it.
Be noted, JSON CSDL uses the correct 'IsOf' already


mikepizzo (approved):

habbes:
Would this break existing code that uses or relies on `IsType`? Does it affect only schema generation or also parsing? Does it affect round-tripping? Is it opt-in?

xuzhg:
> Would this break existing code that uses or relies on `IsType`? Does it affect only schema generation or also parsing? Does it affect round-tripping? Is it opt-in?

You mean to post pone to 9.x version? At that time, we could have the same question/concern?
Since it's breaking change version, What's the difference to do like this within 8.x and 9.x?

PR #3028, opened 2024-07-31
Port collection tests

ElizabethOkerio:
### Issues

### Description

This PR ports collection tests from depending on WCF services to use `Microsoft.AspNetCore.OData` APIs. 
Here are the tests being ported: https://github.com/OData/odata.net/blob/main/test/EndToEndTests/Tests/Client/Build.Desktop/CollectionTests/CollectionNullableFacetTest.cs

These tests verify the behavior of a collection property in a structural entity specifically focussing on the nullable facet of the collection.  For instance, the test checks that a collection structural property with the nullable facet set to false cannot have a null element.  It also verifies that the collection can be empty. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/CollectionTests/Server/CollectionNullableFacetTestController.cs
+++ b/test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/CollectionTests/Server/CollectionNullableFacetTestController.cs
@@ -1,0 +1,56 @@
+﻿//-----------------------------------------------------------------------------
+// <copyright file="CollectionNullableFacetTestController.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.OData.Query;
+using Microsoft.AspNetCore.OData.Routing.Controllers;
+using Microsoft.OData.Client.E2E.Tests.Common.Server.Default;
+
+namespace Microsoft.OData.Client.E2E.Tests.CollectionTests.Server
+{
+    public class CollectionNullableFacetTestController : ODataController
+    {
+        [EnableQuery]
+        [HttpGet("odata/Customers")]
+        public IActionResult GetCustomers()
+        {
+            var rowIndex = DefaultDataSource.Customers;
+            return Ok(rowIndex);
+        }
+
+        [EnableQuery]
+        [HttpGet("odata/Customers({key})")]
+        public IActionResult GetCustomer([FromRoute] int key)
+        {
+            var customer = DefaultDataSource.Customers.SingleOrDefault(a => a.PersonID == key);
+
+            if (customer == null)
+            {
+                return NotFound();
+            }
+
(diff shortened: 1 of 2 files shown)


test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/CollectionTests/Server/CollectionNullableFacetTestController.cs
```diff
@@ -0,0 +1,56 @@
...
+
+namespace Microsoft.OData.Client.E2E.Tests.CollectionTests.Server
+{
+    public class CollectionNullableFacetTestController : ODataController
+    {
+        [EnableQuery]
+        [HttpGet("odata/Customers")]
+        public IActionResult GetCustomers()
+        {
+            var rowIndex = DefaultDataSource.Customers;
```
WanjohiSammy, line 21: Nitpick: We should stick with explicit type instead of 'var'

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/CollectionTests/Server/CollectionNullableFacetTestController.cs
```diff
@@ -0,0 +1,56 @@
...
+            return Ok(rowIndex);
+        }
+
+        [EnableQuery]
+        [HttpGet("odata/Customers({key})")]
+        public IActionResult GetCustomer([FromRoute] int key)
```
WanjohiSammy, line 27: Not a must-fix: Is it possible to convert this method and the rest to `async`
```suggestion
        public async Task<IActionResult> GetCustomerAsync([FromRoute] int key)

```

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/CollectionTests/Server/CollectionNullableFacetTestController.cs
```diff
@@ -0,0 +1,56 @@
...
+            }
+
+            return Ok(customer);
+        }
+
+        [EnableQuery]
+        [HttpPut("odata/Customers({key})")]
+        public IActionResult UpdateCustomers([FromRoute] int key, [FromBody] Customer customer)
+        {
+            var updateCustomer = DefaultDataSource.Customers.FirstOrDefault(a => a.PersonID == key);
```
WanjohiSammy, line 43: Same for all '`var`', use explicit type

ElizabethOkerio: I don't think it's necessary here to have explicit types.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/CollectionTests/Server/CollectionNullableFacetTestController.cs
ElizabethOkerio: Is there any advantage of having async methods here. I do not have any IO or compute intensive logic here. The data here is not from the database also it's static data.

WanjohiSammy (approved):

PR #3032, opened 2024-08-09
Update common models and data sources

ElizabethOkerio:
### Issues

This PR contains some of the common data sources, Edm models and model classes used by the various E2E tests. 

The ported sources and models can be found here:  

https://github.com/OData/odata.net/blob/release-7.x/test/EndToEndTests/Services/Astoria/OpenTypes/DefaultContainer.cs
https://github.com/OData/odata.net/blob/release-7.x/test/EndToEndTests/Services/Astoria/Default/DefaultContainer.cs
https://github.com/OData/odata.net/blob/release-7.x/test/EndToEndTests/Services/Astoria/Default/Entities.cs

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/Common/Server/Default/DefaultDataModel.cs
+++ b/test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/Common/Server/Default/DefaultDataModel.cs
@@ -17,7 +17,7 @@ namespace Microsoft.OData.Client.E2E.Tests.Common.Server.Default
         public string? Street { get; set; }
         public string? City { get; set; }
         public string? PostalCode { get; set; }
-        public DateTime UpdatedTime { get; set; }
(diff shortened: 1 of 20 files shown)


test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/Common/Server/Default/DefaultDataModel.cs
```diff
@@ -73,6 +74,7 @@ public class Person
         public Address? HomeAddress { get; set; }
         public GeographyPoint? Home { get; set; }
         public Person? Parent { get; set; }
+        public DateTime UpdatedTime { get; set; }
```
xuzhg, line 77: why do you use 'DateTime'? It's not compatible to OData spec. Maybe to use "DateTimeOffset"?

xuzhg (approved):

PR #3033, opened 2024-08-12
Update version to 8.0.0

habbes:
Updates version to 8.0.0 (official release). Also moves public API to shipped.


ElizabethOkerio:
cxcx
I think we need to get rid of this. Or this is ok? We should only have .net 8.0 as the dependencies.

gathogojr:
What are the next steps after merging this pull request to **dev-8.x**? Will we continue to do 8.x releases from that branch and 7.x releases from the **main** branch?

gathogojr:
For the empty _*/PublicAPI/net48/PublicAPI.Shipped.txt_ files, I think you should have deleted them altogether after deleting the equivalent _*/PublicAPI/net48/PublicAPI.Unshipped.txt_ files

gathogojr:
Maybe double-check that _src/Microsoft.OData.Core/Build.NuGet/Microsoft.OData.Core.Nightly.Release.nuspec_ and _src/Microsoft.OData.Core/Build.NuGet/Microsoft.OData.Core.Nightly.Release.nuspec_ are updated with all referenced dependencies...

habbes:
> For the empty _*/PublicAPI/net48/PublicAPI.Shipped.txt_ files, I think you should have deleted them altogether after deleting the equivalent _*/PublicAPI/net48/PublicAPI.Unshipped.txt_ files

I deleted them already. Not sure why they're showing up in the diff as "empty". Maybe it's because the files were already empty. But I deleted the actual folders.

gathogojr:
> > For the empty _*/PublicAPI/net48/PublicAPI.Shipped.txt_ files, I think you should have deleted them altogether after deleting the equivalent _*/PublicAPI/net48/PublicAPI.Unshipped.txt_ files
> 
> I deleted them already. Not sure why they're showing up in the diff as "empty". Maybe it's because the files were already empty. But I deleted the actual folders.

Okay. Maybe just check to ensure that the logic that generates them isn't causing them to be regenerated as empty files

habbes:
> > > For the empty _*/PublicAPI/net48/PublicAPI.Shipped.txt_ files, I think you should have deleted them altogether after deleting the equivalent _*/PublicAPI/net48/PublicAPI.Unshipped.txt_ files
> > 
> > 
> > I deleted them already. Not sure why they're showing up in the diff as "empty". Maybe it's because the files were already empty. But I deleted the actual folders.
> 
> Okay. Maybe just check to ensure that the logic that generates them isn't causing them to be regenerated as empty files

Since it's no longer a target framework, they don't get regenerated. Those were leftover files from before we removed .net48 as a target.

habbes:
> What are the next steps after merging this pull request to **dev-8.x**? Will we continue to do 8.x releases from that branch and 7.x releases from the **main** branch?

We'll create a new branch for 7.x from and make main point to what is now dev-8.x

ElizabethOkerio (approved):

PR #3042, opened 2024-08-20
Update data sources

ElizabethOkerio:
### Issues

### Description

This PR updates the common data sources. It adds logic for resetting the data source and creating a copy of the data source by each test to avoid data corruption in cases where tests run in parallel, 

It is important to note that `xunit` tests in different classes run in parallel. This therefore means that if they are updating the same data source then we'll run into data corruption issues which necessitated the tests in 7.x to use `TestPriority` in running the tests. 

To fix this, I've added an action for the different data sources that creates a copy of the original data. This means that each test class will implement this action. The implementation will then be included in the test class's constructor. This will mean that each test will work on a new copy of data every other time and we won't need to use the `TestPriority` attribute that we were using before. This will also make it easy to add assertions to assert that what we are testing is correct. Most of the tests that we are porting didn't have the assertions as the data was corrupted as it was not being reset correctly. 

For creating copies, I'm using `System.Text.Json` library which does deep copies as the source data has some deep nesting.  The other options for copying these data were a bit manual, others depended on reflection hence why I decided to use this library. 

- [ ] *Test cases added*

### Additional work necessary


test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/ActionOverloadingQueryTestsController.cs
```diff
@@ -0,0 +1,303 @@
...
+using Microsoft.AspNetCore.OData.Routing.Controllers;
+using Microsoft.OData.Client.E2E.Tests.Common.Server.EndToEnd;
+
+namespace Microsoft.OData.Client.E2E.Tests.ActionOverloadingTests.Server
+{
+    internal class ActionOverloadingQueryTestsController : ODataController
+    {
+        private static CommonEndToEndDataSource _dataSource;
+
+        public ActionOverloadingQueryTestsController() { }
```
gathogojr, line 20: nit: Is this constructor necessary?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/ActionOverloadingTestsController.cs
```diff
@@ -0,0 +1,303 @@
...
+
+using Microsoft.AspNetCore.Mvc;
+using Microsoft.AspNetCore.OData.Formatter;
+using Microsoft.AspNetCore.OData.Query;
+using Microsoft.AspNetCore.OData.Routing.Controllers;
+using Microsoft.OData.Client.E2E.Tests.Common.Server.EndToEnd;
+
+namespace Microsoft.OData.Client.E2E.Tests.ActionOverloadingTests.Server
+{
+    public class ActionOverloadingTestsController : ODataController
```
gathogojr, line 16: Not for the first time, I have noticed `ActionOverloadingQueryTestsController` and `ActionOverloadingTestsController`. What is the difference between the two?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/Common/Server/EndToEnd/CommonEndToEndDataSource.cs
```diff
@@ -6,19 +6,48 @@
...
+        }
+
+        public CommonEndToEndDataSource()
         {
+            ResetData();
             InitializeData();
         }
 
-        public static void InitializeData()
+        public CommonEndToEndDataSource CreateCopy()
```
gathogojr, line 34: Hopefully all types are serializable... Is it difficult to write a test to verify that the object returned by `CreateCopy` contains the expected data?

marabooy (dismissed):
LGTM

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/Common/Server/EndToEnd/CommonEndToEndDataSource.cs
gathogojr: Would it make better sense to remove the static `Instance` property and instead rename this method `CreateInstance` such that it's what is called when someone want to create an instance of the `CommonEndToEndDataSource`?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/ActionOverloadingQueryTestsController.cs
ElizabethOkerio: I can remove it.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ActionOverloadingTests/Server/ActionOverloadingTestsController.cs
ElizabethOkerio: They are not different per say but we have two test classes. `ActionOverloadingEndToEndTests` and `ActionOverloadingQueryTests` that test different things but since they share the same data source and they can run in parallel, they lead to data corruption. That's why have created a different controller for each test class although some of the APIs could be the same.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/Common/Server/EndToEnd/CommonEndToEndDataSource.cs
ElizabethOkerio: The spatial types might not be serializable but in those cases we can create a different data source for them.

ElizabethOkerio: I think the spatial types might not be serializable but in those cases then we can have a different data source for them i.e if there are any tests on these types.

ElizabethOkerio: ` ``c#  public static DefaultDataSource CreateInstance()
        {
            return new DefaultDataSource();
        }`

gathogojr: https://github.com/OData/odata.net/pull/3042#discussion_r1726596861

That is what I had in mind. Otherwise, how will people know when it's okay to use the `Instance` property and when it's necessary to call `CreateCopy`?

Curious, does the class expose any other methods publicly?

ElizabethOkerio: No. I know some of the methods are public but they can be private.

ElizabethOkerio: updated this.

gathogojr (approved):

PR #3058, opened 2024-09-09
Update message writer and reader to ignore Message info from DI 

marabooy:
Removes calls to GetRequiredService<ODataMessageInfo> as these instances are overwritten immediately.
Simplifies their creation to call stacks where they are needed.

### DIFF
--- a/src/Microsoft.OData.Core/ODataMessageReader.cs
+++ b/src/Microsoft.OData.Core/ODataMessageReader.cs
@@ -904,24 +904,18 @@ namespace Microsoft.OData
         {
             if (this.messageInfo == null)
             {
-                if (this.serviceProvider == null)
+                this.messageInfo = new ODataMessageInfo
                 {
-                    this.messageInfo = new ODataMessageInfo();
-                }
-                else
-                {
-                    this.messageInfo = this.serviceProvider.GetRequiredService<ODataMessageInfo>();
-                }
-
-                this.messageInfo.Encoding = this.encoding;
-                this.messageInfo.IsResponse = this.readingResponse;
-                this.messageInfo.IsAsync = isAsync;
-                this.messageInfo.MediaType = this.contentType;
-                this.messageInfo.Model = this.model;
-                this.messageInfo.PayloadUriConverter = this.payloadUriConverter;
-                this.messageInfo.ServiceProvider = this.serviceProvider;
-                this.messageInfo.MessageStream = messageStream;
-                this.messageInfo.PayloadKind = this.readerPayloadKind;
+                    Encoding = this.encoding,
+                    IsResponse = this.readingResponse,
+                    IsAsync = isAsync,
+                    MediaType = this.contentType,
+                    Model = this.model,
+                    PayloadUriConverter = this.payloadUriConverter,
+                    ServiceProvider = this.serviceProvider,
+                    MessageStream = messageStream,
+                    PayloadKind = this.readerPayloadKind
+                };
             }
 
             return this.messageInfo;
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
@@ -28,6 +28,7 @@ namespace Microsoft.OData.Core.Tests
             services.AddDefaultODataServices();
             ServiceProvider serviceProvider = services.BuildServiceProvider();
 
+            await Task.CompletedTask; // Added due to dotnet < 5 as async await cannot be used only in a loop
             var content1 = string.Concat(Enumerable.Repeat('A', 1000_000));
             var content2 = string.Concat(Enumerable.Repeat('B', 1000_000));
             for (int i = 0; i < 1000; i++)
@@ -54,13 +55,16 @@ namespace Microsoft.OData.Core.Tests
 
             var message = new ODataMessage(outputStream, serviceProvider);
             await using ODataMessageWriter writer = new ODataMessageWriter(message);
+
             await Task.Yield();
 
             await writer.WriteValueAsync(content);
 
             outputStream.Position = 0;
             using var reader = new StreamReader(outputStream);
+
             await Task.Yield();
+
             string writen = await reader.ReadToEndAsync();
             await writer.DisposeAsync();
             return writen;
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
@@ -4,13 +4,13 @@
 // </copyright>
 //---------------------------------------------------------------------
 
-using System.Collections.Generic;
 using System.IO;
 using System.Threading.Tasks;
 using System;
 using Xunit;
 using Microsoft.Extensions.DependencyInjection;
 using System.Linq;
+using Microsoft.OData.Tests;
 
 namespace Microsoft.OData.Core.Tests
 {
@@ -52,7 +52,12 @@ namespace Microsoft.OData.Core.Tests
         {
             using Stream outputStream = new MemoryStream();
 
-            var message = new ODataMessage(outputStream, serviceProvider);
+            var message = new InMemoryMessage
+            {
+                Stream = outputStream,
+                ServiceProvider = serviceProvider
+            };
+
             var responseMessage = new ODataResponseMessage(message, writing: true, enableMessageStreamDisposal: true, maxMessageSize: -1);
             await using ODataMessageWriter writer = new ODataMessageWriter(responseMessage);
 
@@ -68,48 +73,6 @@ namespace Microsoft.OData.Core.Tests
             string written = await reader.ReadToEndAsync();
             await writer.DisposeAsync();
             return written;
-        }
-
-
-        class ODataMessage : IODataResponseMessage, IODataResponseMessageAsync, IServiceCollectionProvider
-        {
-            private Dictionary<string, string> _headers = new();
-            private Stream _outputStream;
-            public ODataMessage(Stream outputStream, IServiceProvider serviceProvider)
-            {
-                this.ServiceProvider = serviceProvider;
-                _outputStream = outputStream;
-            }
-            public IEnumerable<KeyValuePair<string, string>> Headers => _headers;
-
-            public int StatusCode { get => throw new NotImplementedException(); set => throw new NotImplementedException(); }
-
-            public IServiceProvider ServiceProvider { get; private set; }
-
-            public string GetHeader(string headerName)
-            {
-                if (_headers.TryGetValue(headerName, out var value))
-                {
-                    return value;
-                }
-
-                return null;
-            }
-
-            public Stream GetStream()
-            {
-                return _outputStream;
-            }
-
-            public Task<Stream> GetStreamAsync()
-            {
-                return Task.FromResult(_outputStream);
-            }
-
-            public void SetHeader(string headerName, string headerValue)
-            {
-                _headers[headerName] = headerValue;
-            }
         }
     }
 }
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
@@ -1,0 +1,112 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="MessageWriterConcurrencyTests.cs" company="Microsoft">
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using System.Collections.Generic;
+using System.IO;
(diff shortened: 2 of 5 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
```diff
@@ -0,0 +1,112 @@
...
+        /// <returns>A task for the asyncronous test</returns>
+
+        [Fact]
+        public async Task VerifyConcurrentResultsAreConsistentAsync()
+        {
+            ServiceCollection services = new();
+            services.AddDefaultODataServices();
+            ServiceProvider serviceProvider = services.BuildServiceProvider();
+
+            await Task.CompletedTask;
```
habbes, line 31: Why are we awaiting this completed task? Is it to ensure the code below is scheduled to run concurrently? Would `Task.Yield()` achieve the same goal, I see you're using it down below and it seems to express the intent more clearly.

habbes (approved):

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
marabooy: removed

mikepizzo (dismissed):

habbes (dismissed):

mikepizzo (dismissed):

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            await Task.Yield();
+
+            string writen = await reader.ReadToEndAsync();
+            await writer.DisposeAsync();
+            return writen;
+        }
+
+
+        class ODataMessage : IODataResponseMessage, IODataResponseMessageAsync, IServiceCollectionProvider
```
xuzhg, line 74: I remember ODL does have the similar classes created ahead. If yes, use that.

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+
+        /// <summary>
+        /// A helper function that writes to a strem using the message writer and returns the content that is present in the stream.
+        /// </summary>
+        /// <param name="content">String content to write.</param>
+        /// <param name="serviceProvider">A service provider with the default configurations.</param>
+        /// <returns>A task that resolves to the string present in the output stream.</returns>
+        private async Task<string> WritePayload(string content, IServiceProvider serviceProvider)
```
xuzhg, line 52: better rename it as "WritePaylaodAsync" and make it as static?

src/Microsoft.OData.Core/ODataMessageReader.cs
```diff
@@ -904,24 +904,18 @@ private ODataMessageInfo GetOrCreateMessageInfo(Stream messageStream, bool isAsy
         {
             if (this.messageInfo == null)
             {
-                if (this.serviceProvider == null)
```
xuzhg, line 907: so, why do we still keep "https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/ODataServiceCollectionExtensions.cs#L52" ?

habbes: Strictly speaking, that could be a breaking change. The likelihood of breaking existing code is low, likely 0, but here's sample code that will break today if we removed `ODataMessageInfo` from the DI:

```c#
var serviceCollection = new ServiceCollection();
serviceCollection.AddDefaultODataServices();
var serviceProvider = serviceCollection.BuildServiceProvider();
var messageInfo = serviceProvider.GetRequiredService<ODataMessageInfo>();
```
In the current version, this would return a messageInfo instance, if we remove it from DI and ship a new version, then this code would suddenly throw an exception when the user updates.

If we want to be strict about non-breaking minor and patch releases, we can remove it from the next major version and consider making the class private.

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
```diff
@@ -0,0 +1,115 @@
...
+{
+    public class MessageWriterConcurrencyTests
+    {
+        /// <summary>
+        /// Verifies that concurrent message writer does not interleave execution and isolates the underlying streams.
+        /// </summary>
+        /// <returns>A task for the asyncronous test</returns>
+
+        [Fact]
+        public async Task VerifyConcurrentResultsAreConsistentAsync()
```
odero, line 25:
```suggestion
        public async Task VerifyConcurrentResultsAreIsolatedAsync()

```

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
```diff
@@ -0,0 +1,115 @@
...
+using System.Linq;
+
+namespace Microsoft.OData.Core.Tests
+{
+    public class MessageWriterConcurrencyTests
+    {
+        /// <summary>
+        /// Verifies that concurrent message writer does not interleave execution and isolates the underlying streams.
+        /// </summary>
+        /// <returns>A task for the asyncronous test</returns>
```
odero, line 22:
```suggestion
        /// <returns>A task for the asynchronous test</returns>

```

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
```diff
@@ -0,0 +1,115 @@
...
+                Assert.Equal(content2.Length, values[1].Length);
+
+                Assert.Equal(content1, values[0]);
+                Assert.Equal(content2, values[1]);
+            }
+        }
+
+
+        /// <summary>
+        /// A helper function that writes to a strem using the message writer and returns the content that is present in the stream.
```
odero, line 47:
```suggestion
        /// A helper function that writes to a stream using the message writer and returns the content that is present in the stream.

```

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            outputStream.Position = 0;
+            using var reader = new StreamReader(outputStream);
+
+            await Task.Yield();
+
+            string writen = await reader.ReadToEndAsync();
+            await writer.DisposeAsync();
+            return writen;
+        }
```
odero, line 71:
```suggestion
            string written = await reader.ReadToEndAsync();
            await writer.DisposeAsync();
            return written;
        }

```

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
odero, line 74: This class seems like a lot of implementation for a test. And it seems like you're testing test code - why not use actual code that would be ran?

habbes: I think there's an `InMemoryMessage` class that's designed and used for testing, it doesn't wrap around implementation, it has setters for the stream, service provider and other things you may want to customize.

src/Microsoft.OData.Core/ODataMessageReader.cs
marabooy: Resolved

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
marabooy: Found a In memory class used in the tests and moved to it. Odata doesn't have a concrete message so I ended up using the message and the wrapper.

src/Microsoft.OData.Core/ODataMessageReader.cs
marabooy: It could be a breaking change but it seems it could have very few users since a search shows that this was only being used in that location. Also the DI version is scoped which should not reproduce this bug/issue.

habbes (approved):

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
```diff
@@ -0,0 +1,78 @@
...
+
+            await writer.WriteValueAsync(content);
+
+            outputStream.Position = 0;
+            using var reader = new StreamReader(outputStream);
+
+            await Task.Yield();
+
+            string written = await reader.ReadToEndAsync();
+            await writer.DisposeAsync();
```
gathogojr, line 74: Why are we calling `DisposeAsync`? I'd expect use of `await using` would guarantee that the method is called

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// </summary>
+        /// <returns>A task for the asynchronous test</returns>
+        [Fact]
+        public async Task VerifyConcurrentResultsAreIsolatedAsync()
+        {
+            ServiceCollection services = new();
+            services.AddDefaultODataServices();
+            ServiceProvider serviceProvider = services.BuildServiceProvider();
+
+            await Task.CompletedTask; // Added due to dotnet < 5 as async await cannot be used only in a loop
```
gathogojr, line 30: Do we need this? The target framework for Microsoft.OData.Core is net8.0

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
```diff
@@ -0,0 +1,78 @@
...
+            var content1 = string.Concat(Enumerable.Repeat('A', 1000_000));
+            var content2 = string.Concat(Enumerable.Repeat('B', 1000_000));
```
gathogojr, line 32: Is it necessary to create such long strings?

marabooy: Not necessarily but it would make the interleaving more apparent if it happens

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
gathogojr: @marabooy Which projects are these? Are they loaded when you open _OData.sln_?

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
gathogojr: How was https://github.com/OData/odata.net/pull/3058/files#r1753505077 resolved?

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
marabooy: I believe these are the tests let me find them and share here.

src/Microsoft.OData.Core/ODataMessageReader.cs
marabooy: https://github.com/OData/odata.net/blob/main/buildandtest.yml#L24 this build step is the one that fails to build

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
marabooy: Seems I forgot to push this change 🤦🏾 thanks for catching this

test/FunctionalTests/Microsoft.OData.Core.Tests/MessageWriterConcurrencyTests.cs
marabooy: https://github.com/OData/odata.net/blob/main/buildandtest.yml#L24 This line when building will compile vs an older version and causes the failure

gathogojr: @marabooy Could you try moving this task https://github.com/OData/odata.net/blob/95233fc8a399cbae6e816216528d02a581b77b48/buildandtest.yml#L24 to here https://github.com/OData/odata.net/blob/95233fc8a399cbae6e816216528d02a581b77b48/nightly.yml#L14 (since that's the task that require .NET 2.1). Let's see if the build pipeline runs okay and whether you can drop this line.

habbes (approved):

habbes (approved):

PR #3069, opened 2024-09-20
Fixes #3068: Throws ResourceTypeAnnotationNotFirst error when payload…

### LINKED ISSUE #3068: WebApi v7 throws `ResourceTypeAnnotationNotFirst` error when payload has both @removed and @odata.type
Webapiv7 throws ```ResourceTypeAnnotationNotFirst``` error when payload has both @removed and @odata.type. 

Our clients seem to be reusing the same class structure for both normal OData payloads for POST and Delta payloads.
This issue was not encountered by our clients in the older OData versions, possibly because the concept of @removed did not exist in older libraries.

The library seems to be reordering the json payload in this order: ```@odata.context > @removed > @odata.type```. 
So, when @removed property is encountered, it skips checking for @odata.type property. However, the logic later in the ```ReadEntryInstanceAnnotation()``` method in ```ODataJsonLightResourceDeserializer.cs``` assumes that @odata.type was already encountered in ReadResourceStart and hence considers the occurrence as duplicate or occurring in an incorrect order.

### Assemblies affected

OData WebApi lib 7x.

### Reproduce steps

Consider the following payload sent to the POST planner/plans endpoint to create a new plan
```
{
  "@odata.type": "#microsoft.taskServices.plan",
  "title": "The Plan",
  "@removed": false,
  "owner": "abcd"
}
```

### Expected result

201 Created
```
{"@odata.context":"http://localhost/V3.0/$metadata#plans/$entity","owner":"abcd","title":"The Plan"}
```

### Actual result
400 BadRequest

```
"error":{"code":"","message":"The request is invalid:\r\nThe 'odata.type' instance annotation in a resource object is preceded by an invalid property. In OData, the 'odata.type' instance annotation must be either the first property in the JSON object or the second if the 'odata.context' instance annotation is present."
```

### Additional detail
We have upgraded odata to v7 and deployed the changes to Prod and are receiving incidents from our clients due to this issue. We are blocked from deploying to further regions.

xuzhg:
… has both @removed and @odata.type

### Issues

*This pull request fixes #3068.*

### Description

1) Make sure the reordering json reader reorder @context, @removed, @type, @id in this ordering.
   @mikepizzo, Why does @removed should be property before @type?

2) We cannot assume the end customer is reading the payload using delta reader or normal entity reader, so, we don't know whether the @type position is the first property or second property if it contains @context.  So, in this PR, try to retrieve the @type directly from the buffer.  But, we should make sure:
      a) Can a normal entity payload contain 'odata.removed'? 
      b) Can we allow use 'resource reader' to read the payload containing 'odata.removed'?
       @mikepizzo 
  
3) We can skip the value/validation if the 'odata.removed' has the wrong value. It's same concern that if we do like this solution? 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
+++ b/src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
@@ -81,6 +81,64 @@ namespace Microsoft.OData.Json
             this.currentBufferedNode = null;
         }
 
+        internal bool TryGetValueFromBuffered(string propertyName, bool removeIfExist, out object value)
+        {
+            bool found = false;
+            BufferedNode currentNode = this.bufferedNodesHead;
+            if (this.bufferedNodesHead != null)
+            {
+                do
+                {
+                    if (currentNode.NodeType == JsonNodeType.StartObject ||
+                        currentNode.NodeType == JsonNodeType.StartArray)
+                    {
+                        break;
+                    }
+
+                    if (currentNode.NodeType == JsonNodeType.Property &&
+                        currentNode.Value.ToString() == propertyName &&
+                        currentNode.Next != currentNode) // must have another node
+                    {
+                        found = true;
+                        break;
+                    }
+
+                    currentNode = currentNode.Next;
+                }
+                while (currentNode != this.bufferedNodesHead);
+            }
+
+            value = null;
+            if (!found)
+            {
+                return false;
+            }
+
+            BufferedNode valueNode = currentNode.Next;
+
+            if (removeIfExist)
+            {
+                currentNode.Previous.Next = valueNode.Next;
+                valueNode.Next.Previous = currentNode.Previous;
+
+                if (this.bufferedNodesHead == currentNode)
+                {
+                    this.bufferedNodesHead = valueNode.Next;
+                }
+
+                if (this.isBuffering)
+                {
+                    if (this.currentBufferedNode == currentNode || this.currentBufferedNode == valueNode)
+                    {
+                        this.currentBufferedNode = this.bufferedNodesHead;
+                    }
+                }
+            }
+
+            value = valueNode.Value;
+            return true;
+        }
+
         /// <summary>
         /// The type of the last node read.
         /// </summary>
--- a/src/Microsoft.OData.Core/Json/ODataJsonResourceDeserializer.cs
+++ b/src/Microsoft.OData.Core/Json/ODataJsonResourceDeserializer.cs
@@ -100,6 +100,14 @@ namespace Microsoft.OData.Json
             Debug.Assert(resourceState != null, "resourceState != null");
             this.AssertJsonCondition(JsonNodeType.Property, JsonNodeType.EndObject);
 
+            object value;
+            if (this.JsonReader.TryGetValueFromBuffered(ODataJsonConstants.PrefixedODataTypePropertyName, true, out value) ||
+                this.JsonReader.TryGetValueFromBuffered(ODataJsonConstants.SimplifiedODataTypePropertyName, true, out value))
+            {
+                resourceState.Resource.TypeName = ReaderUtils.AddEdmPrefixOfTypeName(ReaderUtils.RemovePrefixOfTypeName(value as string));
+                return;
+            }
+
             // If the current node is the odata.type property - read it.
             if (this.JsonReader.NodeType == JsonNodeType.Property)
             {
@@ -114,6 +122,8 @@ namespace Microsoft.OData.Json
 
                     // Read the annotation value.
                     resourceState.Resource.TypeName = this.ReadODataTypeAnnotationValue();
+
+                 //   resourceState.TypeAnnotationFound = true;
                 }
             }
 
@@ -821,9 +831,15 @@ namespace Microsoft.OData.Json
                     return this.ReadAndValidateAnnotationStringValue(annotationName);
 
                 case ODataAnnotationNames.ODataRemoved: // 'odata.removed'
+                    // If the value of 'odata.removed' is an object, let's throw exception since it should be not read here.
+                    if (this.JsonReader.NodeType == JsonNodeType.StartObject)
                     {
                         throw new ODataException(ODataErrorStrings.ODataJsonResourceDeserializer_UnexpectedDeletedEntryInResponsePayload);
                     }
+
+                    // for others, let's skip it
+                    this.JsonReader.SkipValue();
+                    return null;
 
                 default:
                     ODataAnnotationNames.ValidateIsCustomAnnotationName(annotationName);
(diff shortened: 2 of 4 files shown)


src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -81,6 +81,64 @@ internal BufferingJsonReader(IJsonReader innerReader, string inStreamErrorProper
...
+                        currentNode.Value.ToString() == propertyName &&
+                        currentNode.Next != currentNode) // must have another node
+                    {
+                        found = true;
+                        break;
+                    }
+
+                    currentNode = currentNode.Next;
+                }
+                while (currentNode != this.bufferedNodesHead);
```
habbes, line 108: Was wondering how `currentNode` ever be `bufferedNodeHead` then I confirmed that this is a circular linked list. I think we should abstract traversal and access of the nodes through some helper method or inner class to avoid accidentally accessing the nodes the wrong way.

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -81,6 +81,64 @@ internal BufferingJsonReader(IJsonReader innerReader, string inStreamErrorProper
...
+
+                    if (currentNode.NodeType == JsonNodeType.Property &&
+                        currentNode.Value.ToString() == propertyName &&
+                        currentNode.Next != currentNode) // must have another node
+                    {
+                        found = true;
+                        break;
+                    }
+
+                    currentNode = currentNode.Next;
```
habbes, line 106: can `currentNode.Next` ever return `null`? Should account for that?

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -81,6 +81,64 @@ internal BufferingJsonReader(IJsonReader innerReader, string inStreamErrorProper
...
+                {
+                    if (currentNode.NodeType == JsonNodeType.StartObject ||
+                        currentNode.NodeType == JsonNodeType.StartArray)
+                    {
+                        break;
+                    }
+
+                    if (currentNode.NodeType == JsonNodeType.Property &&
+                        currentNode.Value.ToString() == propertyName &&
+                        currentNode.Next != currentNode) // must have another node
```
habbes, line 100: Why do check whether there's another node after this?

habbes: Just to see if I understand, property and value are stored in separate node, so if we find the expected property node, we want to make sure there's a value node after it. Is that correct?

src/Microsoft.OData.Core/Json/ODataJsonResourceDeserializer.cs
```diff
@@ -100,6 +100,14 @@ internal void ReadResourceTypeName(IODataJsonReaderResourceState resourceState)
             Debug.Assert(resourceState != null, "resourceState != null");
             this.AssertJsonCondition(JsonNodeType.Property, JsonNodeType.EndObject);
 
+            object value;
+            if (this.JsonReader.TryGetValueFromBuffered(ODataJsonConstants.PrefixedODataTypePropertyName, true, out value) ||
+                this.JsonReader.TryGetValueFromBuffered(ODataJsonConstants.SimplifiedODataTypePropertyName, true, out value))
```
habbes, line 105: nit: make it clear what the `true` arguments imply

```suggestion
            if (this.JsonReader.TryGetValueFromBuffered(ODataJsonConstants.PrefixedODataTypePropertyName, removeIfExists: true, out value) ||
                this.JsonReader.TryGetValueFromBuffered(ODataJsonConstants.SimplifiedODataTypePropertyName, removeIfExists: true, out value))

```

src/Microsoft.OData.Core/Json/ODataJsonResourceDeserializer.cs
```diff
@@ -114,6 +122,8 @@ internal void ReadResourceTypeName(IODataJsonReaderResourceState resourceState)
 
                     // Read the annotation value.
                     resourceState.Resource.TypeName = this.ReadODataTypeAnnotationValue();
+
+                 //   resourceState.TypeAnnotationFound = true;
```
habbes, line 126: Why was this statement added then commented out? Should we remove it?

habbes (approved):

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
xuzhg: That's correct. Property and value are two node types and they are stored in separate node.

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
xuzhg: If we have the 'bufferedNodesHead', the 'Next' is not null because if it's empty, 'bufferedNodesHead.Next == bufferedNodesHead'.

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
xuzhg: I checked the logics in 'BefferingJsonReader' and that's the way to check the end of link.

src/Microsoft.OData.Core/Json/ODataJsonResourceDeserializer.cs
xuzhg: Yep, miss to clean it.

src/Microsoft.OData.Core/Json/ODataJsonResourceDeserializer.cs
xuzhg: #Resolved.

src/Microsoft.OData.Core/Json/BufferingJsonReader.cs
```diff
@@ -81,6 +81,64 @@ internal BufferingJsonReader(IJsonReader innerReader, string inStreamErrorProper
             this.currentBufferedNode = null;
         }
 
+        internal bool TryGetValueFromBuffered(string propertyName, bool removeIfExist, out object value)
```
mikepizzo, line 84: This could have some pretty severe performance ramifications.

If a payload doesn't have the value that we try to get, we will read the entire payload into memory.  For very large, or streamed, responses, this could overflow memory.

This kinda defeats the whole point of streaming and ordering, which is that we can process the response in order.

The previous implementation assumes that type, if present, is the first thing we read. So we read the first node and, if it's type, we store that, and if it is not type, we assume there is no type and continue reading the payload as normal.

I would have expected we simply expand that logic.  That is, we read the first node. If it is one of our "metadata" nodes (context, removed, type, id), then we get the value and repeat for the next node. As soon as we find a node that is not one of our metadata nodes, we break and start reading the payload as normal.

mikepizzo (changes requested):
The current design defeats the whole purpose of streaming -- for example, if the resource does not have an @odata.type we will read to the end.

mikepizzo:
See related https://github.com/OData/AspNetCoreOData/discussions/1523. This is a similar issue, but instead of using @id the payload uses the key property (named "id").   For [this](https://github.com/OData/odata.net/pull/3069#discussion_r1783394751) to work, we would probably have to pre-read the control information *and key properties* before deciding whether this or not was a deleted resource.

PR #3071, opened 2024-09-23
Isolate scoped settings in default OData services

### LINKED ISSUE #3070: Changes in reader/writer/uri parser settings in one request affect other requests
There's a flaw in the `AddDefaultODataServices()` that leads to the same instances of the settings classes to be shared across threads even when the services are scoped. This means if one request changes the writer settings, this change will affect other requests as well. Settings should be isolated per request, hence why they are registered as scoped services.

### Assemblies affected

`Microsoft.OData.Core` 8.x

### Reproduce steps

Here's a test case that reproduces the scenario:
```c#
var services = new ServiceCollection();
services.AddDefaultODataServices();
var provider = services.BuildServiceProvider();
using var scope1 = provider.CreateScope();
var settings = scope1.ServiceProvider.GetRequiredService<ODataMessageWriterSettings>();
settings.EnableCharactersCheck = true;

var settingsFromSameScope = scope1.ServiceProvider.GetRequiredService<ODataMessageWriterSettings>();

using var scope2 = provider.CreateScope();
var settingsFromOtherScope = scope2.ServiceProvider.GetRequiredService<ODataMessageWriterSettings>();

// Instances from the same scope should be the same
Assert.True(object.ReferenceEquals(settings, settingsFromSameScope));
Assert.True(settingsFromSameScope.EnableCharactersCheck);
// Instances from different scopes should be different
Assert.False(object.ReferenceEquals(settings, settingsFromOtherScope));
Assert.False(settingsFromOtherScope.EnableCharactersCheck);
```

### Expected result

Test should pass.

### Actual result

Test fails.

### Additional detail

Here's the offending snippet from `AddDefaulODataServices()`

```c#
 // Finally, register configurable settings.
 var readerSettings = new ODataMessageReaderSettings(odataVersion);
 configureReaderAction?.Invoke(readerSettings);
 services.AddScoped(sp => readerSettings);

 var writerSettings = new ODataMessageWriterSettings(odataVersion);
 configureWriterAction?.Invoke(writerSettings);
 services.AddScoped(sp => writerSettings);

 var parserSettings = new ODataUriParserSettings();
 configureUriParserAction?.Invoke(parserSettings);
 services.AddScoped(sp => parserSettings);
```

Let's take the `readerSettings` as an example. This implementation was meant to provide a way to customize default options for the settings classes, but each request should be able to independently update its settings when appropriate. However, since the same `readerSettings` instance will be returned from the service provider, if one request changes a setting, the change will affect other requests as well.

We could fix the issue by either constructing the settings in the `AddScope()` delegate, or by cloning the settings instance:

```c#
services.AddScope(_ => {
    var readerSettings = new ODataMessageReaderSettings(...);
     configureReaderAction?.Invoke(readerSettings);
    return readerSettings;
});
```
or
```c#
 var readerSettings = new ODataMessageReaderSettings(odataVersion);
 configureReaderAction?.Invoke(readerSettings);
 services.AddScoped(sp => readerSettings.Clone());
```

I think the latter is a better solution since we can't predict the cost of calling the user-provided delegate each time.

habbes:
### Issues

*This pull request fixes #3070 *

### Description

Clones the reader, writer and URI parser settings when registering them as scoped services in `AddDefaultODataServices()` to ensure that different request scopes get different copies that can be independently modified.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/ODataServiceCollectionExtensions.cs
+++ b/src/Microsoft.OData.Core/ODataServiceCollectionExtensions.cs
@@ -55,15 +55,15 @@ namespace Microsoft.Extensions.DependencyInjection
             // Finally, register configurable settings.
             var readerSettings = new ODataMessageReaderSettings(odataVersion);
             configureReaderAction?.Invoke(readerSettings);
-            services.AddScoped(sp => readerSettings);
+            services.AddScoped(sp => readerSettings.Clone());
 
             var writerSettings = new ODataMessageWriterSettings(odataVersion);
             configureWriterAction?.Invoke(writerSettings);
-            services.AddScoped(sp => writerSettings);
+            services.AddScoped(sp => writerSettings.Clone());
 
             var parserSettings = new ODataUriParserSettings();
             configureUriParserAction?.Invoke(parserSettings);
-            services.AddScoped(sp => parserSettings);
+            services.AddScoped(sp => parserSettings.Clone());
 
             return services;
         }
--- a/src/Microsoft.OData.Core/UriParser/ODataUriParserSettings.cs
+++ b/src/Microsoft.OData.Core/UriParser/ODataUriParserSettings.cs
@@ -91,7 +91,7 @@ namespace Microsoft.OData.UriParser
 
         internal ODataUriParserSettings Clone()
         {
-            ODataUriParserSettings copy = new()
+            return new()
             {
                 FilterLimit = this.FilterLimit,
                 OrderByLimit = this.OrderByLimit,
@@ -101,8 +101,6 @@ namespace Microsoft.OData.UriParser
                 MaximumExpansionCount = this.MaximumExpansionCount,
                 EnableParsingKeyAsSegment = this.EnableParsingKeyAsSegment
             };
-
-            return copy;
         }
 
         /// <summary>
--- a/src/Microsoft.OData.Core/UriParser/ODataUriParserSettings.cs
+++ b/src/Microsoft.OData.Core/UriParser/ODataUriParserSettings.cs
@@ -91,14 +91,17 @@ namespace Microsoft.OData.UriParser
 
         internal ODataUriParserSettings Clone()
         {
-            ODataUriParserSettings copy = new ODataUriParserSettings();
-            copy.FilterLimit = this.FilterLimit;
-            copy.OrderByLimit = this.OrderByLimit;
-            copy.PathLimit = this.PathLimit;
-            copy.SearchLimit = this.SearchLimit;
-            copy.MaximumExpansionDepth = this.MaximumExpansionDepth;
-            copy.MaximumExpansionCount = this.MaximumExpansionCount;
-            copy.EnableParsingKeyAsSegment = this.EnableParsingKeyAsSegment;
+            ODataUriParserSettings copy = new()
+            {
+                FilterLimit = this.FilterLimit,
+                OrderByLimit = this.OrderByLimit,
+                PathLimit = this.PathLimit,
+                SearchLimit = this.SearchLimit,
+                MaximumExpansionDepth = this.MaximumExpansionDepth,
+                MaximumExpansionCount = this.MaximumExpansionCount,
+                EnableParsingKeyAsSegment = this.EnableParsingKeyAsSegment
+            };
+
             return copy;
         }
 
--- a/src/Microsoft.OData.Core/UriParser/ODataUriParserSettings.cs
+++ b/src/Microsoft.OData.Core/UriParser/ODataUriParserSettings.cs
@@ -89,6 +89,19 @@ namespace Microsoft.OData.UriParser
             this.EnableParsingKeyAsSegment = true;
         }
 
+        internal ODataUriParserSettings Clone()
+        {
+            ODataUriParserSettings copy = new ODataUriParserSettings();
+            copy.FilterLimit = this.FilterLimit;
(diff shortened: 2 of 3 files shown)


src/Microsoft.OData.Core/ODataServiceCollectionExtensions.cs
```diff
@@ -55,15 +55,15 @@ public static IServiceCollection AddDefaultODataServices(this IServiceCollection
             // Finally, register configurable settings.
             var readerSettings = new ODataMessageReaderSettings(odataVersion);
             configureReaderAction?.Invoke(readerSettings);
-            services.AddScoped(sp => readerSettings);
+            services.AddScoped(sp => readerSettings.Clone());
```
marabooy, line 58: question: Would there be an issue if we called configure reader action multiple times e.g. if we used a delegate instead of the cloned instance?

src/Microsoft.OData.Core/UriParser/ODataUriParserSettings.cs
```diff
@@ -89,6 +89,19 @@ public ODataUriParserSettings()
...
+        {
+            ODataUriParserSettings copy = new ODataUriParserSettings();
+            copy.FilterLimit = this.FilterLimit;
+            copy.OrderByLimit = this.OrderByLimit;
+            copy.PathLimit = this.PathLimit;
+            copy.SearchLimit = this.SearchLimit;
+            copy.MaximumExpansionDepth = this.MaximumExpansionDepth;
+            copy.MaximumExpansionCount = this.MaximumExpansionCount;
+            copy.EnableParsingKeyAsSegment = this.EnableParsingKeyAsSegment;
+            return copy;
```
marabooy, line 102:
```suggestion
          return new ODataUriParserSettings()
           {
            FilterLimit = this.FilterLimit;
            OrderByLimit = this.OrderByLimit;
            PathLimit = this.PathLimit;
            SearchLimit = this.SearchLimit;
            MaximumExpansionDepth = this.MaximumExpansionDepth;
            MaximumExpansionCount = this.MaximumExpansionCount;
            EnableParsingKeyAsSegment = this.EnableParsingKeyAsSegment;
           };

```
nit: we could use theObject initializer here to make things clearer. Also not sure if we could use => new .... instead

src/Microsoft.OData.Core/ODataServiceCollectionExtensions.cs
habbes: Are you referring to this approach:

```c#
services.AddScope(_ => {
    var readerSettings = new ODataMessageReaderSettings(...);
    configureReaderAction?.Invoke(readerSettings);
    return readerSettings;
});
```
If so, I considered it (and mentioned it in the attached issue). The reason I opted for cloning instead of that approach is because we can't predict the cost of the delegate since its user defined, for all we know it could be reading from a file or database. On the other hand, since we control the implementation of the `Clone()` method, we can make better assumptions about it cost and we can optimize if it becomes a hotspot, so I think it's safer to err on the side of calling `Clone()` multiple times rather than calling the opaque user-defined delegate multiple times.

marabooy (approved):
LGTM

src/Microsoft.OData.Core/ODataServiceCollectionExtensions.cs
```diff
@@ -55,15 +55,15 @@ public static IServiceCollection AddDefaultODataServices(this IServiceCollection
             // Finally, register configurable settings.
             var readerSettings = new ODataMessageReaderSettings(odataVersion);
             configureReaderAction?.Invoke(readerSettings);
-            services.AddScoped(sp => readerSettings);
+            services.AddScoped(sp => readerSettings.Clone());
 
             var writerSettings = new ODataMessageWriterSettings(odataVersion);
             configureWriterAction?.Invoke(writerSettings);
-            services.AddScoped(sp => writerSettings);
+            services.AddScoped(sp => writerSettings.Clone());
```
syprieur, line 62: When an `ODataMessageWriterSettings` is passed as argument to an `ODataMessageWriter`, the scoped `ODataMessageWriterSettings` service is being updated to that value?

habbes: That's a very question, and after checking the code, apparently we do :(. The settings from the `ODataMessageWriterSettings` are copied to the scoped `ODataMessageWriterSettings` service:

see:
- https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/ODataMessageWriter.cs#L110
- https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/ODataMessageWriterSettings.cs#L485

which calls:

```c#
internal static ODataMessageWriterSettings CreateWriterSettings(
    IServiceProvider container,
    ODataMessageWriterSettings other)
{
    ODataMessageWriterSettings writerSettings;
    if (container == null)
    {
        writerSettings = new ODataMessageWriterSettings();
    }
    else
    {
       // we get the scoped settings
        writerSettings = container.GetRequiredService<ODataMessageWriterSettings>();
    }

    if (other != null)
    {
        // then overwrite the scoped service based on the settings passed to the writer
        writerSettings.CopyFrom(other);
    }

    return writerSettings;
}
```

syprieur: That explains the observed bug and confirms the hypothesis: the same settings instance is indeed re-used and updated concurrently.
Your change will fix it.

gathogojr (approved):

PR #3077, opened 2024-09-30
Use ReadOnlySpan<char> to replace the 'new string(...)' to reduce memory allocation

xuzhg:
Use ExpressionLexer to do ReadOnlySpan investigation and seek feebacks to move forward to all other ODL parts.

### Be noted, 

1) the ReadOnlySpan<char> doesn't override the operator == and != for string. Remember to use 'Equals(...)' to replace them.

```C#
ReadOnlySpan<char> s1 = "nav/$ref".AsSpan().Slice(4,4);

Console.WriteLine(s1.ToString());
Console.WriteLine(s1 == "$ref");
Console.WriteLine(s1 != "$ref");
Console.WriteLine(s1.Equals("$ref", StringComparison.Ordinal));
```

It outputs
```cmd
$ref
False
True
True
```

2) Anytime to call 'ToString()' on ReadOnlySpan<char> will do the 'new string(...)', so be noted, not to call it multiple times until you need the string and remember to cache it.

### Benchmark:

image

See the allocated, using 'ReadOnlySpan<char>' is 1/3 memory allocation comparing to the 'new string(...)'.

#### Codes

```C#
   int NumberOfItems = 100000;
   public string expression = "abc(2014-09-19T12:13:14+00:00)/3258.678765765489753678965390/SRID=1234(POINT(10 20))/Function(foo=@x,bar=1,baz=@y)";

    [Benchmark]
    public void ExpressionLexerUsingNewStringSingleRun()
    {
        ExpressionLexer2 lexer = new Microsoft.OData.UriParser.ExpressionLexer2(expression, true, false);
        while (lexer.CurrentToken.Kind != ExpressionTokenKind.End)
        {
            lexer.NextToken();
        }
    }

    [Benchmark]
    public void ExpressionLexerUsingReadOnlySingleRun()
    {

        ExpressionLexer lexer = new Microsoft.OData.UriParser.ExpressionLexer(expression, true, false);
        while (lexer.CurrentToken.Kind != ExpressionTokenKind.End)
        {
            lexer.NextToken();
        }
    }

    [Benchmark]
    public void ExpressionLexerUsingNewString()
    {
        for (int i = 0; i < NumberOfItems; i++)
        {
            ExpressionLexer2 lexer = new Microsoft.OData.UriParser.ExpressionLexer2(expression, true, false);
            while (lexer.CurrentToken.Kind != ExpressionTokenKind.End)
            {
                lexer.NextToken();
            }
        }
    }

    [Benchmark]
    public void ExpressionLexerUsingReadOnly()
    {
        for (int i = 0; i < NumberOfItems; i++)
        {
            ExpressionLexer lexer = new Microsoft.OData.UriParser.ExpressionLexer(expression, true, false);
            while (lexer.CurrentToken.Kind != ExpressionTokenKind.End)
            {
                lexer.NextToken();
            }
        }
    }
```

### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/CustomUriLiteralPrefixes.cs
+++ b/src/Microsoft.OData.Core/UriParser/CustomUriLiteralPrefixes.cs
@@ -92,12 +92,13 @@ namespace Microsoft.OData.UriParser
         /// </summary>
         /// <param name="literalPrefix">The literal prefix of the EdmType</param>
         /// <returns>Null if the custom literal prefix has no registered EdmType.</returns>
-        internal static IEdmTypeReference GetEdmTypeByCustomLiteralPrefix(string literalPrefix)
+        internal static IEdmTypeReference GetEdmTypeByCustomLiteralPrefix(ReadOnlySpan<char> literalPrefix)
         {
             lock (Locker)
             {
                 IEdmTypeReference edmTypeReference;
-                if (CustomLiteralPrefixesOfEdmTypes.TryGetValue(literalPrefix, out edmTypeReference))
+                string literal = literalPrefix.ToString();
+                if (CustomLiteralPrefixesOfEdmTypes.TryGetValue(literal, out edmTypeReference))
                 {
                     return edmTypeReference;
                 }
--- a/src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
@@ -82,8 +82,8 @@ namespace Microsoft.OData.UriParser
                 while (this.lexer.CurrentToken.Kind != ExpressionTokenKind.CloseParen)
                 {
                     string text = this.UriQueryExpressionParser.EnableCaseInsensitiveBuiltinIdentifier
-                        ? this.lexer.CurrentToken.Text.ToLowerInvariant()
-                        : this.lexer.CurrentToken.Text;
+                        ? this.lexer.CurrentToken.Text.ToString().ToLowerInvariant()
+                        : this.lexer.CurrentToken.Text.ToString();
 
                     // Prepend '$' prefix if needed.
                     if (this.UriQueryExpressionParser.EnableNoDollarQueryOptions && !text.StartsWith(UriQueryConstants.DollarSign, StringComparison.Ordinal))
--- a/src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
@@ -78,33 +78,30 @@ namespace Microsoft.OData.UriParser
                     throw new ODataException(ODataErrorStrings.UriParser_EmptyParenthesis);
                 }
 
+                StringComparison comparison = this.UriQueryExpressionParser.EnableCaseInsensitiveBuiltinIdentifier ?
+                        StringComparison.OrdinalIgnoreCase :
+                        StringComparison.Ordinal;
+
                 // Look for all the supported query options
                 while (this.lexer.CurrentToken.Kind != ExpressionTokenKind.CloseParen)
                 {
-                    string text = this.UriQueryExpressionParser.EnableCaseInsensitiveBuiltinIdentifier
-                        ? this.lexer.CurrentToken.Text.ToString().ToLowerInvariant()
-                        : this.lexer.CurrentToken.Text.ToString();
+                    ReadOnlySpan<char> textSpan = this.lexer.CurrentToken.Span;
 
-                    // Prepend '$' prefix if needed.
-                    if (this.UriQueryExpressionParser.EnableNoDollarQueryOptions && !text.StartsWith(UriQueryConstants.DollarSign, StringComparison.Ordinal))
+                    if (textSpan.Equals(ExpressionConstants.QueryOptionFilter, comparison) ||
+                        (this.UriQueryExpressionParser.EnableNoDollarQueryOptions && textSpan.Equals("filter", comparison)))
                     {
-                        text = string.Format(CultureInfo.InvariantCulture, "{0}{1}", UriQueryConstants.DollarSign, text);
+                        // $filter within $count segment
+                        filterToken = ParseInnerFilter();
                     }
-
-                    switch (text)
+                    else if (textSpan.Equals(ExpressionConstants.QueryOptionSearch, comparison) ||
+                        (this.UriQueryExpressionParser.EnableNoDollarQueryOptions && textSpan.Equals("search", comparison)))
                     {
-                        case ExpressionConstants.QueryOptionFilter: // $filter within $count segment
-                            filterToken = ParseInnerFilter();
-                            break;
-
-                        case ExpressionConstants.QueryOptionSearch: // $search within $count segment
-                            searchToken = ParseInnerSearch();
-                            break;
-
-                        default:
-                            {
-                                throw new ODataException(ODataErrorStrings.UriQueryExpressionParser_IllegalQueryOptioninDollarCount);
-                            }
+                        // $search within $count segment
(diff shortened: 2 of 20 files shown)


src/Microsoft.OData.Core/UriParser/CustomUriLiteralPrefixes.cs
```diff
@@ -92,12 +92,13 @@ public static bool RemoveCustomLiteralPrefix(string literalPrefix)
...
         /// <returns>Null if the custom literal prefix has no registered EdmType.</returns>
-        internal static IEdmTypeReference GetEdmTypeByCustomLiteralPrefix(string literalPrefix)
+        internal static IEdmTypeReference GetEdmTypeByCustomLiteralPrefix(ReadOnlySpan<char> literalPrefix)
         {
             lock (Locker)
             {
                 IEdmTypeReference edmTypeReference;
-                if (CustomLiteralPrefixesOfEdmTypes.TryGetValue(literalPrefix, out edmTypeReference))
+                string literal = literalPrefix.ToString();
+                if (CustomLiteralPrefixesOfEdmTypes.TryGetValue(literal, out edmTypeReference))
```
habbes, line 101: It sucks that we have to resort to allocations like this when we want to do cache lookups. In .NET 9, we can avoid this with the help of [`IAlternateEqualityComparer`](https://learn.microsoft.com/en-us/dotnet/api/system.collections.generic.ialternateequalitycomparer-2?view=net-9.0) which could allow us to look up a dictionary that uses string keys using a `ReadOnlySpan<char>` (see https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-9/#core-collections).

In .NET 8 we can decide whether to settle for the allocations or to implement a workaround if allocation cost is significant.

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
```diff
@@ -82,8 +82,8 @@ public CountSegmentToken CreateCountSegmentToken(QueryToken countedInstance)
                 while (this.lexer.CurrentToken.Kind != ExpressionTokenKind.CloseParen)
                 {
                     string text = this.UriQueryExpressionParser.EnableCaseInsensitiveBuiltinIdentifier
-                        ? this.lexer.CurrentToken.Text.ToLowerInvariant()
-                        : this.lexer.CurrentToken.Text;
+                        ? this.lexer.CurrentToken.Text.ToString().ToLowerInvariant()
+                        : this.lexer.CurrentToken.Text.ToString();
```
habbes, line 86: Is it necessary to call `ToString()` here? Since we are only using the strings for switch matching and not returning it, maybe we can avoid the extra allocations. Maybe we can use `Text.Equals(...)` instead of the `switch` statement. And for the `$` prefix, maybe we can also perform a comparison without having to do the string concatenation.

src/PlatformHelper.cs
```diff
@@ -241,15 +243,23 @@ internal static Date ConvertStringToDate(string text)
...
             }
 
-            return TimeOfDay.TryParse(text, CultureInfo.InvariantCulture, out timeOfDay);
+            TimeSpan time;
+            bool b = TimeSpan.TryParse(text, CultureInfo.InvariantCulture, out time);
+            if (b && time.Ticks >= TimeOfDay.MinTickValue && time.Ticks <= TimeOfDay.MaxTickValue)
+            {
+                timeOfDay = new TimeOfDay(time.Ticks);
+                return true;
+            }
```
habbes, line 260: Why this change? Does the change from string to span change the `TryParse` behaviour?

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
```diff
@@ -97,14 +97,14 @@ public bool TryParseIdentifierAsFunction(QueryToken parent, out QueryToken resul
...
-                functionName = this.Lexer.CurrentToken.Text;
+                functionName = this.Lexer.CurrentToken.Span;
                 this.Lexer.NextToken();
             }
 
             FunctionParameterToken[] arguments = this.ParseArgumentListOrEntityKeyList(() => lexer.RestorePosition(position));
             if (arguments != null)
             {
-                result = new FunctionCallToken(functionName, arguments, parent);
+                result = new FunctionCallToken(functionName.ToString(), arguments, parent);
```
habbes, line 107: Should we consider making the `FunctionCallToken` and other query tokens accept `ReadOnlySpan<char>` or `ReadOnlyMemory<T>` instead of `string`? I think that would reduce allocations further. I'm going off the assumption that these tokens are the generated from a lexer, and therefore always tied to some input query string expression, and so the tokens can also point into segments of the original string instead of being requiring allocations. The downside is the source string would remain in memory for the lifetime of the tokens.

I don't know if we expect users to instantiate these tokens without going through a lexer, but since the constructors are public, I guess that possible. But we can still store span/memory without a breaking change I think, and the string properties could be computed from the span/memory on demand then cached. If we have a clean separation between the syntax and semantic parse phases, then the lexer and lexical tokens could be short-lived most of the time, maybe we could be more aggressive about using spans and even ref structs.

xuzhg:
> @xuzhg could you share the benchmarks that you used for the perf tests, did you also try to run the existing UriParser benchmarks before and after these changes?

I create a branch to hold the perf test at: https://github.com/OData/odata.net/tree/ReadOnlySpanBenchMark

This PR is just a starting point. We will move to UriParser and others in the comming step.

src/Microsoft.OData.Core/UriParser/CustomUriLiteralPrefixes.cs
xuzhg: Yep, I should to change to avoid the allocation using .NET 9. In 8 version, current change is a solution because there's no extra allocation happening comparing to the existing code. 
an alternate solution is to create a helper method to compare the `ReadOnlySpan<Char>` and `Dictionary<string, T>`. The method is to traver the dictionary key and get a span of the key, and use the span to compare `ReadOnlySpan<Char>`, It could be a little bit bad perf, I need more investigation.

src/Microsoft.OData.Core/UriParser/Parsers/CountSegmentParser.cs
xuzhg: Thanks. Improved it.

src/Microsoft.OData.Core/UriParser/Parsers/SelectExpandOptionParser.cs
```diff
@@ -137,8 +137,8 @@ internal SelectTermToken BuildSelectTermToken(PathSegmentToken pathToken, string
                 while (this.lexer.CurrentToken.Kind != ExpressionTokenKind.CloseParen)
                 {
                     string text = this.enableCaseInsensitiveBuiltinIdentifier
-                        ? this.lexer.CurrentToken.Text.ToLowerInvariant()
-                        : this.lexer.CurrentToken.Text;
+                        ? this.lexer.CurrentToken.Text.ToString().ToLowerInvariant()
+                        : this.lexer.CurrentToken.Text.ToString();
```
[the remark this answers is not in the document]
xuzhg: Thanks, Improved it.

src/PlatformHelper.cs
xuzhg: Yes. TimeOfDay.TryParse calls 'TimeSpan.TryParse' and do the if clause.
Since it is using the 'System.String' version, and now we'd call 'ReadOnlySpan<char>' version. I don't want to change the definition of TimeOfDay. So, use the codes directly here. #won't fix

xuzhg:
@habbes Resolved the comment. Please take a look again. I will keep moving to next step. It could be 'UriParser'.

xuzhg:
Welcome any new comments/reviews. I will triage them together, but I have to merge it now and move my work forward.

Issue #1320, opened 2018-11-09
OData Client fails to parse $batch response for Linux


VirusQuartirus:
*When $batch response created by ExecuteBatchAsync is returned from Linux machine (so it uses 0x0A as a new line instead of 0x0D0A) OData client fails to deserialize last batch item*

### Assemblies affected

*OData .Net Client 5.x-7x.*

### Reproduce steps

*Use a Linux as a service host machine or use Fiddler autoresponder with modified response (replace \r\n line breaks with \n).*

### Expected result

*All responses have to be parsed by client.*

### Actual result

*Exception occurs.
```
System.InvalidOperationException
  HResult=0x80131509
  Message=Invalid JSON. More than one value was found at the root of the JSON content. JSON content can only have one value at the root level, which is an array, an object or a primitive value.
  Source=Microsoft.OData.Client
  StackTrace:
   at Microsoft.OData.Client.Materialization.FeedAndEntryMaterializerAdapter.TryRead()

Inner Exception 1:
ODataException: Invalid JSON. More than one value was found at the root of the JSON content. JSON content can only have one value at the root level, which is an array, an object or a primitive value.

Microsoft.OData.ODataException
  HResult=0x80131509
  Message=Invalid JSON. More than one value was found at the root of the JSON content. JSON content can only have one value at the root level, which is an array, an object or a primitive value.
  Source=Microsoft.OData.Core
  StackTrace:
   at Microsoft.OData.Json.JsonReader.Read()
   at Microsoft.OData.Json.BufferingJsonReader.ReadInternal()
   at Microsoft.OData.Json.BufferingJsonReader.ReadNextAndCheckForInStreamError()
   at Microsoft.OData.Json.BufferingJsonReader.ReadInternal()
   at Microsoft.OData.Json.BufferingJsonReader.Read()
   at Microsoft.OData.JsonLight.ODataJsonLightReader.ReadAtResourceEndImplementationSynchronously()
   at Microsoft.OData.JsonLight.ODataJsonLightReader.ReadAtResourceEndImplementation()
   at Microsoft.OData.ODataReaderCore.ReadImplementation()
   at Microsoft.OData.ODataReaderCore.ReadSynchronously()
   at Microsoft.OData.ODataReaderCore.InterceptException[T](Func`1 action)
   at Microsoft.OData.ODataReaderCore.Read()
   at Microsoft.OData.Client.Materialization.ODataReaderWrapper.Read()
[...]

PR #3084, opened 2024-10-14
Port to 7.x: Fix issue #1320 - write metadata async issue

WanjohiSammy:
### Issues

### Description

- Cherry-pick 17840516e3eba4c670f46d715dce9f811415898c from 8.x to 7.x
- Ensure compatibility with older .NET versions.

- [X] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Edm/EdmModelVisitor.cs
+++ b/src/Microsoft.OData.Edm/EdmModelVisitor.cs
@@ -91,7 +91,7 @@ namespace Microsoft.OData.Edm
                     return this.ProcessSchemaElementAsync(element);
                 default:
                     Debug.Assert(false, Edm.Strings.UnknownEnumVal_SchemaElementKind(element.SchemaElementKind));
-                    return Task.FromResult<InvalidOperationException>(new InvalidOperationException(Edm.Strings.UnknownEnumVal_SchemaElementKind(element.SchemaElementKind)));
+                    return TaskUtils.GetFaultedTask<InvalidOperationException>(new InvalidOperationException(Edm.Strings.UnknownEnumVal_SchemaElementKind(element.SchemaElementKind)));
             }
         }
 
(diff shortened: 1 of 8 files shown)


test/FunctionalTests/Microsoft.OData.Core.Tests/AsyncYieldStream.cs
```diff
@@ -0,0 +1,138 @@
...
+
+namespace Microsoft.OData.Tests
+{
+    /// <summary>
+    /// Stream wrapper that yields in every async method
+    /// to ensure the read/write operations complete asynchronously
+    /// even when the operate completely in-memory. This helps simulate
+    /// async I/O in unit tests without making requests out of the process.
+    /// </summary>
+    internal class AsyncYieldStream : Stream
```
xuzhg, line 14: ODL test project doesn't contain similar class?

WanjohiSammy: @xuzhg With the `AsyncStream`, I was not able to repro the issue faced by the customer https://github.com/OData/AspNetCoreOData/issues/1320. I added class `AsyncYieldStream` with `await Task.Yield()` in the async methods to help repro the issue and fix it. 

`await Task.Yield()` ensures that method yields control back to the caller immediately, allowing other tasks to run without blocking the thread.

src/Microsoft.OData.Edm/EdmModelVisitor.cs
```diff
@@ -194,11 +264,77 @@ public void VisitExpression(IEdmExpression expression)
...
+                    return this.ProcessRecordExpressionAsync((IEdmRecordExpression)expression);
+                case EdmExpressionKind.StringConstant:
+                    return this.ProcessStringConstantExpressionAsync((IEdmStringConstantExpression)expression);
+                case EdmExpressionKind.TimeOfDayConstant:
+                    return this.ProcessTimeOfDayConstantExpressionAsync((IEdmTimeOfDayConstantExpression)expression);
+                case EdmExpressionKind.DurationConstant:
+                    return this.ProcessDurationConstantExpressionAsync((IEdmDurationConstantExpression)expression);
+                case EdmExpressionKind.None:
+                    return this.ProcessExpressionAsync(expression);
+                default:
```
marabooy, line 323: Shouldn't we throw this instead?

habbes (approved):

gathogojr (approved):

PR #3087, opened 2024-10-15
Update Core Vocabulary to include IsDelta

### LINKED ISSUE #3086: Update Core vocabulary to include IsDelta
The OASIS Core vocabulary has some new definitions since we last integrated with EdmLib; most notably, the IsDelta annotation term.

### Assemblies affected

All Versions

### Reproduce steps

Load the Core Model and observe that there is no IsDelta term defined.

### Expected result

IsDelta term to be defined.

### Actual result

IsDelta term is not defined.

### Additional detail

We have a manual process for copying over the latest OASIS vocabularies to create core models when building our EdmLib.  The IsDelta term was added since we last did this manual updating.

mikepizzo:
### Issues

*This pull request fixes #3086.*

### Description

Updated the OASIS Core vocabulary to include the IsDelta term.
Updated the tests to validate this term.

- [x] *Test cases added*

### Additional work necessary

None


test/FunctionalTests/Microsoft.OData.Edm.Tests/Vocabularies/CoreVocabularyTests.cs
```diff
@@ -566,6 +572,7 @@ public void TestCoreVocabularyEnumType(string typeName, string members, bool isF
         [InlineData("SchemaVersion", "Edm.String", "Schema Reference")]
         [InlineData("ODataVersions", "Edm.String", "EntityContainer")]
         [InlineData("Example", "Org.OData.Core.V1.ExampleValue", "EntityType ComplexType TypeDefinition Term Property NavigationProperty Parameter ReturnType")]
+        [InlineData("IsDelta","Org.OData.Core.V1.Tag","ReturnType Parameter")]
```
xuzhg, line 575: This new line looks unnecessary.

xuzhg (approved):

test/FunctionalTests/Microsoft.OData.Edm.Tests/Vocabularies/CoreVocabularyTests.cs
mikepizzo: This is testing that the IsDelta term is defined in the vocabulary. Of course, that should also be apparent from comparing the expected text, but that's how the test is designed.

PR #3091, opened 2024-10-17
Improve handling of dynamic properties in expressions

### LINKED ISSUE #3090: Should support dynamic properties as right-hand operator in In clause using type of left-hand operand
According to [OData ABNF](https://[docs.oasis-open.org/odata/odata/v4.02/csd01/abnf/odata-abnf-construction-rules.txt](https://docs.oasis-open.org/odata/odata/v4.02/csd01/abnf/odata-abnf-construction-rules.txt)), the right operand of an in expression can be a common expression.  Specifically:

`inExpr = RWS "in" RWS ( listExpr / commonExpr )`

allows the right-hand operand of an in expression to be a collection-valued property.

This works in the library today if the right-hand expression is a schematized property, for example:

`https://services.odata.org/TripPinRESTierService/Me/Trips?$filter='Travel' in Tags`

However, if the right-hand expression is a dynamic property, the query fails because we are unable to determine the type of the right-hand operand.

We should support dynamic properties as the right-hand operand, using the type of the left-hand operator.

### Assemblies affected

All

### Reproduce steps

Try to parse the following filter expression, where "trip" is an open type and "undefined" is an undefined property:

trips?$filter='Europe' in undefined

### Expected result

A valid In expression with a right-hand operand that is a CollectionOpenPropertyAccessNode with a CollectionType of Collection(Edm.String).

### Actual result

An error is thrown stating that the right hand operand isn't a collection.

### Additional detail

Note that, for the in operator, user can work around this limitation by use of the Any operator. However, the Any operator is less intuitive and potentially more expensive to compute than the in operator, and doesn't solve the general case.

mikepizzo:
### Issues

*This pull request fixes #3090.*

### Description

Allows open property access nodes to be created with a type when the type can be inferred from the context.
Supports creating an In expression with a CollectionOpenPropertyAccessNode as the right-hand operand, using the type of the left-hand operand

- [x] *Test cases added*

### Additional work necessary

None

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -60,7 +60,7 @@ namespace Microsoft.OData.UriParser
 
             // If the left operand is either an integral or a string type and the right operand is a collection of enums,
             // Calls the MetadataBindingUtils.ConvertToTypeIfNeeded() method to convert the left operand to the same enum type as the right operand.
-            if ((right is CollectionPropertyAccessNode && right.ItemType.IsEnum()) && (left.TypeReference != null && (left.TypeReference.IsString() || left.TypeReference.IsIntegral())))
+            if ((!(right is CollectionConstantNode) && right.ItemType.IsEnum()) && (left.TypeReference != null && (left.TypeReference.IsString() || left.TypeReference.IsIntegral())))
             {
                 left = MetadataBindingUtils.ConvertToTypeIfNeeded(left, right.ItemType);
             }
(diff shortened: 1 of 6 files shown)


src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -151,7 +151,15 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
             }
             else
             {
-                operand = this.bindMethod(queryToken) as CollectionNode;
+                var node = this.bindMethod(queryToken);
+                if (node is SingleValueOpenPropertyAccessNode openNode)
```
xuzhg, line 155: why doesn't the 'bindMethod' return 'CollectionOpenPropertyAccessNode' directly?

xuzhg (approved):

PR #3097, opened 2024-10-24
Fixes #3094: continue to replace substring in HttpHeaderValueLexer

### LINKED ISSUE #3094: Use ReadOnlySpan<char> to replace 'new string(...)'

### Assemblies affected

ODL has lot of string parsing scenarios, for example:

	1) OData Uri parsing
	2) OData query option parsing
	3) OData json payload parsing

The typical parsing process is:

1) Read character one by one
2) 'Create a new substring' every time if it could be a 'token'
3) Create the corresponding token instance using the substring
4) Do the further process

For example:

"abc(2014-09-19T12:13:14+00:00)/3258.678765765489753678965390/SRID=1234(POINT(1020))/Function(foo=@x,bar=1,baz=@y)"

If do a lexer, we can get the following tokens:
```table
Kind: Identifier: Text: abc
Kind: OpenParen: Text: (
Kind: DateTimeOffsetLiteral: Text: 2014-09-19T12:13:14+00:00
Kind: CloseParen: Text: )
Kind: Slash: Text: /
Kind: DecimalLiteral: Text: 3258.678765765489753678965390
Kind: Slash: Text: /
Kind: Identifier: Text: SRID
Kind: Equal: Text: =
Kind: IntegerLiteral: Text: 1234
Kind: OpenParen: Text: (
Kind: Identifier: Text: POINT
Kind: OpenParen: Text: (
Kind: IntegerLiteral: Text: 10
Kind: IntegerLiteral: Text: 20
Kind: CloseParen: Text: )
Kind: CloseParen: Text: )
Kind: Slash: Text: /
Kind: Identifier: Text: Function
Kind: OpenParen: Text: (
Kind: Identifier: Text: foo
Kind: Equal: Text: =
Kind: Identifier: Text: @x
Kind: Comma: Text: ,
Kind: Identifier: Text: bar
Kind: Equal: Text: =
Kind: IntegerLiteral: Text: 1
Kind: Comma: Text: ,
Kind: Identifier: Text: baz
Kind: Equal: Text: =
Kind: Identifier: Text: @y
Kind: CloseParen: Text: )
```
As an example: 
	1) We have to create a string for literal l'2014-09-19T12:13:14+00:00'
	2) Test it using the created substring
	3) Create the DateTimeOffset instance using that string
	4) The created substring goes into the GC

Moreover, most of time, we don't need the 'substring', for example, "(", ")", etc, They are special but no other meaning, just a label. We don't need to create a substring for all of them.

In a summary, it seems we don't need to create a 'new string'. Because the literal is there (in the original raw string), we just need to retrieve it when we do need it. And most importantly, if we just test it (to see whether it's a valid DateTimeOffset literal), we can just test the PART of original raw string, don't need to create a new substring.

### Expected result

### Actual result

*What is actually happening.*

### Additional detail

xuzhg:
Add the test cases for HttpUtils

### Issues

*This pull request fixes #3094.*

### Description

* This PR focuses on HttpHeaderValueLexer, try to replace 'substring' using ReadOnlyMemory<char> to save the memory allocatioin.
* Add test cases for httpUtils

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/HttpHeaderValueLexer.cs
+++ b/src/Microsoft.OData.Core/HttpHeaderValueLexer.cs
@@ -6,6 +6,7 @@
 
 namespace Microsoft.OData
 {
+    using System;
     using System.Collections.Generic;
     using System.Diagnostics;
     using System.Linq;
@@ -49,13 +50,13 @@ namespace Microsoft.OData
         /// The value of the current parsed item. If the item type is quoted-string, this returns the unescaped and unquoted string value. For other item types,
         /// the value is the same as the original text from the header.
         /// </summary>
-        private readonly string value;
+        private readonly ReadOnlyMemory<char> value;
 
         /// <summary>
         /// The original text of the current parsed item. If the item type is quoted-string, this returns the escaped and quoted string value straight from the header.
         /// For other item types, the original text is the same as the item value.
         /// </summary>
-        private readonly string originalText;
+        private readonly ReadOnlyMemory<char> originalText;
 
         /// <summary>
         /// Constructs a new instance of <see cref="HttpHeaderValueLexer"/>.
@@ -67,7 +68,7 @@ namespace Microsoft.OData
         /// <param name="originalText">The original text of the current parsed item. If the item type is quoted-string, this returns the escaped and quoted string value straight from the header.
         /// For other item types, the original text is the same as the item value.</param>
         /// <param name="startIndexOfNextItem">The start index of the next item to be parsed.</param>
-        private HttpHeaderValueLexer(string httpHeaderName, string httpHeaderValue, string value, string originalText, int startIndexOfNextItem)
+        private HttpHeaderValueLexer(string httpHeaderName, string httpHeaderValue, ReadOnlyMemory<char> value, ReadOnlyMemory<char> originalText, int startIndexOfNextItem)
         {
             this.httpHeaderName = httpHeaderName;
             this.httpHeaderValue = httpHeaderValue;
@@ -114,25 +115,13 @@ namespace Microsoft.OData
         /// The value of the current parsed item. If the item type is quoted-string, this returns the unescaped and unquoted string value. For other item types,
         /// the value is the same as the original text from the header.
         /// </summary>
-        internal string Value
-        {
-            get
-            {
-                return this.value;
-            }
-        }
+        internal ReadOnlyMemory<char> Value => this.value;
 
         /// <summary>
         /// The original text of the current parsed item. If the item type is quoted-string, this returns the escaped and quoted string value straight from the header.
         /// For other item types, the original text is the same as the item value.
         /// </summary>
-        internal string OriginalText
-        {
-            get
-            {
-                return this.originalText;
-            }
-        }
+        internal ReadOnlyMemory<char> OriginalText => this.originalText;
 
         /// <summary>
         /// The type of the current parsed item.
@@ -246,7 +235,7 @@ namespace Microsoft.OData
         {
             Debug.Assert(lexer.Type == HttpHeaderValueItemType.Token, "lexer.Type == HttpHeaderValueItemType.Token");
 
-            string name = lexer.OriginalText;
+            ReadOnlyMemory<char> name = lexer.OriginalText;
             string value = null;
             lexer = lexer.ReadNext();
             if (lexer.Type == HttpHeaderValueItemType.ValueSeparator)
@@ -256,7 +245,7 @@ namespace Microsoft.OData
                     lexer.Type == HttpHeaderValueItemType.Token || lexer.Type == HttpHeaderValueItemType.QuotedString,
                     "lexer.Type == HttpHeaderValueItemType.Token || lexer.Type == HttpHeaderValueItemType.QuotedString");
 
-                value = lexer.OriginalText;
+                value = lexer.OriginalText.ToString();
                 lexer = lexer.ReadNext();
             }
 
@@ -264,7 +253,7 @@ namespace Microsoft.OData
                 lexer.Type == HttpHeaderValueItemType.End || lexer.Type == HttpHeaderValueItemType.ElementSeparator || lexer.Type == HttpHeaderValueItemType.ParameterSeparator,
                 "lexer.Type == HttpHeaderValueItemType.End || lexer.Type == HttpHeaderValueItemType.ElementSeparator || lexer.Type == HttpHeaderValueItemType.ParameterSeparator");
 
-            return new KeyValuePair<string, string>(name, value);
+            return new KeyValuePair<string, string>(name.ToString(), value);
         }
 
         /// <summary>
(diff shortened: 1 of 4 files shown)


src/Microsoft.OData.Core/HttpHeaderValueLexer.cs
```diff
@@ -256,15 +245,15 @@ private static KeyValuePair<string, string> ReadKeyValuePair(ref HttpHeaderValue
                     lexer.Type == HttpHeaderValueItemType.Token || lexer.Type == HttpHeaderValueItemType.QuotedString,
                     "lexer.Type == HttpHeaderValueItemType.Token || lexer.Type == HttpHeaderValueItemType.QuotedString");
 
-                value = lexer.OriginalText;
+                value = lexer.OriginalText.ToString();
```
gathogojr, line 248: Can `value` be of type `ReadOnlyMemory<char>` instead of `string` to avoid calling `ToString`?

gathogojr: Probably this line https://github.com/OData/odata.net/pull/3097/files#diff-d4037d6fdf2743a1c6adbc34ea8ea80637de11e94ed36d4ab9e3250b396a3981R256 means there might be no value in doing that...

src/Microsoft.OData.Core/HttpHeaderValueLexer.cs
```diff
@@ -343,7 +333,7 @@ private sealed class HttpHeaderStart : HttpHeaderValueLexer
             /// <param name="httpHeaderName">The name of the HTTP header being parsed.</param>
             /// <param name="httpHeaderValue">The value of the HTTP header being parsed.</param>
             internal HttpHeaderStart(string httpHeaderName, string httpHeaderValue)
-                : base(httpHeaderName, httpHeaderValue, /*value*/ null, /*originalText*/ null, 0)
+                : base(httpHeaderName, httpHeaderValue, /*value*/ default, /*originalText*/ default, 0)
```
gathogojr, line 336: nit:
```suggestion
                : base(httpHeaderName, httpHeaderValue, value: default, originalText: default, <argumentName>: 0)
```

src/Microsoft.OData.Core/HttpHeaderValueLexer.cs
```diff
@@ -581,20 +562,14 @@ private sealed class HttpHeaderEnd : HttpHeaderValueLexer
             /// Constructs a new instance of <see cref="HttpHeaderEnd"/>.
             /// </summary>
             private HttpHeaderEnd()
-                : base(/*httpHeaderName*/ null, /*httpHeaderValue*/ null, /*value*/ null, /*originalText*/ null, /*startIndexOfNextItem*/ -1)
+                : base(/*httpHeaderName*/ null, /*httpHeaderValue*/ null, /*value*/ default, /*originalText*/ default, /*startIndexOfNextItem*/ -1)
```
gathogojr, line 565: nit:
```suggestion
                : base(httpHeaderName: null, httpHeaderValue: null, value: default, originalText: default, startIndexOfNextItem: -1)
```

src/Microsoft.OData.Core/HttpUtils.cs
```diff
@@ -453,37 +455,82 @@ internal static string ReadTokenOrQuotedStringValue(string headerName, string he
...
+            }
+        }
+
+        /// <summary>
+        /// Skips whitespace in the specified text by advancing an index to
+        /// the next non-whitespace character.
+        /// </summary>
+        /// <param name="text">Text to scan.</param>
+        /// <param name="textIndex">Index to begin scanning from.</param>
+        /// <returns>true if the end of the string was reached, false otherwise.</returns>
```
gathogojr, line 513: This `<returns>` description differs from the actual return value from the method.

src/Microsoft.OData.Core/HttpUtils.cs
```diff
@@ -433,18 +431,22 @@ internal static string ReadTokenOrQuotedStringValue(string headerName, string he
                 }
             }
 
-            char currentChar = default(char);
+            // if not a quoted string, reading it as Token
+            if (!isQuotedString)
+            {
+                return ReadTokenValue(headerName, headerText, ref textIndex, createException);
```
habbes, line 437: The previous behaviour looks different from the previous version where we throw an exception, is this adding a new feature or support for new behaviour?

src/Microsoft.OData.Core/HttpUtils.cs
```diff
@@ -453,37 +455,82 @@ internal static string ReadTokenOrQuotedStringValue(string headerName, string he
                         break;
                     }
 
+                    if (parameterValue == null)
+                    {
+                        parameterValue = new StringBuilder();
```
habbes, line 460: I'm curious how using the string builder compares to allocating/renting a block large enough to store the encoded escaped string then using the string constructor with a slice at the end.

```c#
int maxLength = estimateUpperboundSizeOfParameterValue;

char[] array = null;
Span<char> parameterValue = maxLength < 128 ? stackalloc char[128] : array = ArrayPool<char>.Shared.Rent(maxLength);

while (...) {
  // copy text into parameterValue buffer instead of string builder.Append which can resize itself
}


// get string from span buffer:
return parameterValue.Slice(0, textLength).ToString();
if (array != null)
{
    ArrayPool<char>.Shared.Return(array);
}

habbes: The difference here is that `StringBuilder.Append()` can allocate multiple arrays depending on the size of the string we append. The approach above proposes allocating a single buffer that is guaranteed to be large enough for the parameter value. We can estimate a minimal upper bound, it doesn't have to be the exact size. At the end we slice only the portion we've written to and create a string from that.

We'd need to compare both approaches to determine conclusively which is one more performant for the common sizes of parameterValue we expect to encounter.

The `stackalloc` and `ArrayPool` also help minimize allocations compared to `StringBuilder`.

src/Microsoft.OData.Core/HttpHeaderValueLexer.cs
xuzhg: it definitely needs to be a string to the KeyValuePair. So, ToString() is necessary and You may notice it only happens once

src/Microsoft.OData.Core/HttpHeaderValueLexer.cs
xuzhg: #Resolved. I removed the argument name since it's straightforward.

src/Microsoft.OData.Core/HttpUtils.cs
xuzhg: #Resolved.

src/Microsoft.OData.Core/HttpUtils.cs
xuzhg: Header typically doesn't contain a long original string. So, I think The StringBuilder used here is enough. If it's critical issue for performance, we can file new issue to improve this part.

xuzhg: #Resolved.

xuzhg:
Let me know your more concerns.

Issue #1744, opened 2020-04-15
Isof type parameter quotes


VirusQuartirus:
According [OData V4 spec](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_isof) `isof` function type parameter should not be enclosed in quotes. However both .NET client lib sends quoted type parameter and service impl requires type to be quoted.

### Assemblies affected

OData .Net lib 6.x, 7.x

### Reproduce steps

Execute query like
https://services.odata.org/TripPinRESTierService/(S(4dte2qgkqy1lkpudzkpnk4om))/People?$filter=isof(Microsoft.OData.Service.Sample.TrippinInMemory.Models.Employee)

### Expected result

Request should return list of Employees like https://services.odata.org/TripPinRESTierService/(S(4dte2qgkqy1lkpudzkpnk4om))/People?$filter=isof('Microsoft.OData.Service.Sample.TrippinInMemory.Models.Employee') request does.

### Actual result

Request fails with "Cast or IsOf Function must have a type in its arguments." message.

mikepizzo:
Here is the stack from RESTier, which suggests that the issue is in the UriParser within OData Library:

[...]

PR #3117, opened 2024-11-12
Fix type casting exceptions in 'isof' and 'cast' calls.

WanjohiSammy:
### Issues

*This pull request fixes [#1744](https://github.com/OData/odata.net/issues/1744).*

### Description

When using unquoted type parameters in `isof` and `cast`, the operations perform as expected provided that the source type can be cast to the target type. Issues arise when casting is not possible, resulting in the following exception:
```
Encountered invalid type cast. '{source type}' is not assignable from '{target type}'.
```
Lets say you have the following data model:
```cs
namespace NS;

public class Product
{
    public int ProductID { get; set; }
    // Others properties
    public Address SupplierAddress { get; set; }
}

public class DerivedProduct : Product
{
   public string AnotherProperty { get; set; }
    // Others properties
}

public class Address
{
    // Others properties
    public string City { get; set; }
}
```
Executing the query `isof(NS.DerivedCategory)` will throw the exception:
```
"Encountered invalid type cast. 'NS.Category' is not assignable from 'NS.Product'."
```

This occurs because ODL checks if `NS.Product` is related to `NS.Category` using the [CheckRelatedTo ](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/Binders/DottedIdentifierBinder.cs#L104) function and since the types are not related or one is not a derivative of the other, the exception is thrown.

### Main changes
According to the [spec_cast](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_cast) and [spec_isof](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_isof) specifications, there is no indication that an exception should be thrown if casting fails. This pull request introduces the `TryBindDottedIdentifierForIsOfOrCastFunctionCall` method, which handles the binding of `DottedIdentifierToken` specifically for the `isof` and `cast` function calls without performing the `CheckRelatedTo` validation, thereby preventing exceptions when type casting is not possible.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
@@ -240,9 +240,9 @@ namespace Microsoft.OData.UriParser
                 QueryNode argumentNode;
 
                 // If the function is IsOf and the argument is a dotted identifier, we should bind it as a single resource node
-                if ((functionCallToken.Name == ExpressionConstants.UnboundFunctionIsOf || functionCallToken.Name == ExpressionConstants.UnboundFunctionCast) && argument.ValueToken is DottedIdentifierToken dottedIdentifier)
+                if (UnboundFunctionNames.Contains(functionCallToken.Name) && argument.ValueToken is DottedIdentifierToken dottedIdentifier)
                 {
-                    argumentNode = this.TryBindDottedIdentifierAsIsOfFunctionCall(dottedIdentifier);
+                    argumentNode = this.TryBindDottedIdentifierForIsOfOrCastFunctionCall(dottedIdentifier);
                 }
                 else
                 {
@@ -449,7 +449,7 @@ namespace Microsoft.OData.UriParser
         /// <param name="dottedIdentifierToken">The dotted identifier token to bind.</param>
         /// <returns>A <see cref="QueryNode"/> representing the bound single resource node.</returns>
         /// <exception cref="ODataException">Thrown when the token cannot be bound as a single resource node.</exception>
-        private QueryNode TryBindDottedIdentifierAsIsOfFunctionCall(DottedIdentifierToken dottedIdentifierToken)
+        private QueryNode TryBindDottedIdentifierForIsOfOrCastFunctionCall(DottedIdentifierToken dottedIdentifierToken)
         {
             QueryNode parent = null;
             IEdmType parentType = null;
--- a/src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
@@ -234,7 +234,24 @@ namespace Microsoft.OData.UriParser
 
             // If there isn't, bind as Uri function
             // Bind all arguments
-            List<QueryNode> argumentNodes = new List<QueryNode>(functionCallToken.Arguments.Select(ar => this.bindMethod(ar)));
+            List<QueryNode> argumentNodes = new List<QueryNode>();
+            foreach (FunctionParameterToken argument in functionCallToken.Arguments)
+            {
+                QueryNode argumentNode;
+
+                // If the function is IsOf and the argument is a dotted identifier, we should bind it as a single resource node
+                if ((functionCallToken.Name == ExpressionConstants.UnboundFunctionIsOf || functionCallToken.Name == ExpressionConstants.UnboundFunctionCast) && argument.ValueToken is DottedIdentifierToken dottedIdentifier)
+                {
+                    argumentNode = this.TryBindDottedIdentifierAsIsOfFunctionCall(dottedIdentifier);
+                }
+                else
+                {
+                    argumentNode = this.bindMethod(argument);
+                }
+
+                argumentNodes.Add(argumentNode);
+            }
+
             return BindAsUriFunction(functionCallToken, argumentNodes);
         }
 
@@ -424,6 +441,54 @@ namespace Microsoft.OData.UriParser
             }
 
             return true;
+        }
+
+        /// <summary>
+        /// Binds a <see cref="DottedIdentifierToken"/> for the 'isof' and 'cast' function calls.
+        /// </summary>
+        /// <param name="dottedIdentifierToken">The dotted identifier token to bind.</param>
+        /// <returns>A <see cref="QueryNode"/> representing the bound single resource node.</returns>
+        /// <exception cref="ODataException">Thrown when the token cannot be bound as a single resource node.</exception>
+        private QueryNode TryBindDottedIdentifierAsIsOfFunctionCall(DottedIdentifierToken dottedIdentifierToken)
+        {
+            QueryNode parent = null;
+            IEdmType parentType = null;
+
+            if (state.ImplicitRangeVariable != null)
+            {
+                if (dottedIdentifierToken.NextToken == null)
+                {
+                    parent = NodeFactory.CreateRangeVariableReferenceNode(state.ImplicitRangeVariable);
+                    parentType = state.ImplicitRangeVariable.TypeReference.Definition;
+                }
+                else
+                {
+                    parent = this.bindMethod(dottedIdentifierToken.NextToken);
+                    parentType = parent.GetEdmType();
+                }
+            }
+
+            SingleResourceNode parentAsSingleResource = parent as SingleResourceNode;
+            IEdmSchemaType childType = UriEdmHelpers.FindTypeFromModel(state.Model, dottedIdentifierToken.Identifier, this.Resolver);
+            IEdmStructuredType childStructuredType = childType as IEdmStructuredType;
+
+            if (childStructuredType == null)
+            {
+                return this.bindMethod(dottedIdentifierToken);
+            }
+
+            if (parentType != null)
+            {
+                this.state.ParsedSegments.Add(new TypeSegment(childType, parentType, null));
+            }
+
+            CollectionResourceNode parentAsCollection = parent as CollectionResourceNode;
+            if (parentAsCollection != null)
+            {
+                return new CollectionResourceCastNode(parentAsCollection, childStructuredType);
+            }
+
+            return new SingleResourceCastNode(parentAsSingleResource, childStructuredType);
         }
 
         /// <summary>
--- a/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
+++ b/test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
@@ -844,27 +844,31 @@ namespace Microsoft.OData.Tests.ScenarioTests.UriParser
         }
 
         [Theory]
-        [InlineData("isof(Fully.Qualified.Namespace.Pet1)", "Fully.Qualified.Namespace.Pet1")]
-        [InlineData("cast(Fully.Qualified.Namespace.HomeAddress)/City eq 'City1'", "Fully.Qualified.Namespace.HomeAddress")]
-        public void IsOfAndCastFunctionsWithSingleParameterWithoutSingleQuotes_WithIncorrectType_ThrowException(string filterQuery, string fullyQualifiedTypeName)
+        [InlineData("isof(Fully.Qualified.Namespace.Pet1)")]
+        [InlineData("isof(MyAddress,Fully.Qualified.Namespace.Pet1)")]
+        [InlineData("isof(null,Fully.Qualified.Namespace.Person)")]
+        [InlineData("isof('',Fully.Qualified.Namespace.Person)")]
+        public void IsOfFunctionsWithUnquotedTypeParameter_WithIncorrectType_DoesNotThrowException(string filterQuery)
         {
             // Arrange & Act
-            Action test = () => ParseFilter(filterQuery, HardCodedTestModel.TestModel, HardCodedTestModel.GetPersonType(), HardCodedTestModel.GetPeopleSet());
+            var exception = Record.Exception(() => ParseFilter(filterQuery, HardCodedTestModel.TestModel, HardCodedTestModel.GetPersonType(), HardCodedTestModel.GetPeopleSet()));
 
             // Assert
-            test.Throws<ODataException>(Strings.MetadataBinder_HierarchyNotFollowed(fullyQualifiedTypeName, "Fully.Qualified.Namespace.Person"));
+            Assert.Null(exception);
         }
 
         [Theory]
-        [InlineData("isof(MyAddress,Fully.Qualified.Namespace.Pet1)", "Fully.Qualified.Namespace.Pet1")]
-        [InlineData("cast(MyAddress,Fully.Qualified.Namespace.Employee)/WorkID eq 345", "Fully.Qualified.Namespace.Employee")]
-        public void IsOfAndCastFunctionsWithTwoParameterWhereTypeParameterIsWithoutSingleQuotes_WithIncorrectType_ThrowException(string filterQuery, string fullyQualifiedTypeName)
+        [InlineData("cast(Fully.Qualified.Namespace.HomeAddress)/City eq 'City1'")]
+        [InlineData("cast(MyAddress,Fully.Qualified.Namespace.Employee)/WorkID eq 345")]
+        [InlineData("cast(null,Fully.Qualified.Namespace.Employee)/WorkID eq 345")]
+        [InlineData("cast('',Fully.Qualified.Namespace.Employee)/WorkID eq 345")]
+        public void CastFunctionWithUnquotedTypeParameter_WithIncorrectType_DoesNotThrowException(string filterQuery)
         {
             // Arrange & Act
-            Action test = () => ParseFilter(filterQuery, HardCodedTestModel.TestModel, HardCodedTestModel.GetPersonType(), HardCodedTestModel.GetPeopleSet());
+            var exception = Record.Exception(() => ParseFilter(filterQuery, HardCodedTestModel.TestModel, HardCodedTestModel.GetPersonType(), HardCodedTestModel.GetPeopleSet()));
 
             // Assert
-            test.Throws<ODataException>(Strings.MetadataBinder_HierarchyNotFollowed(fullyQualifiedTypeName, "Fully.Qualified.Namespace.Address"));
+            Assert.Null(exception);
         }
 
         [Fact]


xuzhg (approved):

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
```diff
@@ -234,7 +234,22 @@ internal QueryNode BindFunctionCall(FunctionCallToken functionCallToken)
 
             // If there isn't, bind as Uri function
             // Bind all arguments
-            List<QueryNode> argumentNodes = new List<QueryNode>(functionCallToken.Arguments.Select(ar => this.bindMethod(ar)));
+            List<QueryNode> argumentNodes = new List<QueryNode>(functionCallToken.Arguments.Select(argument =>
```
habbes, line 237: It may be better to call `functionCallToken.Arguments.Select(...).ToList()` than `new List(...)`. If `Arguments` has a known collection size, then using `ToList()` could use an optimized path that avoids resizing. See: 
https://github.com/dotnet/runtime/issues/107560

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -844,27 +844,31 @@ public void IsOfFunctionWithTwoParameters_WithoutSingleQuotesOnTypeParameter_Sho
...
 
         [Theory]
-        [InlineData("isof(MyAddress,Fully.Qualified.Namespace.Pet1)", "Fully.Qualified.Namespace.Pet1")]
-        [InlineData("cast(MyAddress,Fully.Qualified.Namespace.Employee)/WorkID eq 345", "Fully.Qualified.Namespace.Employee")]
-        public void IsOfAndCastFunctionsWithTwoParameterWhereTypeParameterIsWithoutSingleQuotes_WithIncorrectType_ThrowException(string filterQuery, string fullyQualifiedTypeName)
+        [InlineData("cast(Fully.Qualified.Namespace.HomeAddress)/City eq 'City1'")]
+        [InlineData("cast(MyAddress,Fully.Qualified.Namespace.Employee)/WorkID eq 345")]
+        [InlineData("cast(null,Fully.Qualified.Namespace.Employee)/WorkID eq 345")]
+        [InlineData("cast('',Fully.Qualified.Namespace.Employee)/WorkID eq 345")]
+        public void CastFunctionWithUnquotedTypeParameter_WithIncorrectType_DoesNotThrowException(string filterQuery)
```
habbes, line 865: What is the expected behaviour of `cast()` when the types are incorrect? If it doesn't throw an exception, what will happen? How is this scenario expected to be handled?

habbes, line 865: Is this behaviour consistent with how cast is handled when the type param is quoted?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -844,27 +844,31 @@ public void IsOfFunctionWithTwoParameters_WithoutSingleQuotesOnTypeParameter_Sho
...
 
         [Theory]
-        [InlineData("isof(Fully.Qualified.Namespace.Pet1)", "Fully.Qualified.Namespace.Pet1")]
-        [InlineData("cast(Fully.Qualified.Namespace.HomeAddress)/City eq 'City1'", "Fully.Qualified.Namespace.HomeAddress")]
-        public void IsOfAndCastFunctionsWithSingleParameterWithoutSingleQuotes_WithIncorrectType_ThrowException(string filterQuery, string fullyQualifiedTypeName)
+        [InlineData("isof(Fully.Qualified.Namespace.Pet1)")]
+        [InlineData("isof(MyAddress,Fully.Qualified.Namespace.Pet1)")]
+        [InlineData("isof(null,Fully.Qualified.Namespace.Person)")]
+        [InlineData("isof('',Fully.Qualified.Namespace.Person)")]
+        public void IsOfFunctionsWithUnquotedTypeParameter_WithIncorrectType_DoesNotThrowException(string filterQuery)
```
habbes, line 851: Is this behaviour consistent with how `isof` is handled when the type param is quoted?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
WanjohiSammy: This is what is happening currently with `cast` with `quoted type parameter`
- ODL will bind `literal type parameter token` to `ConstantNode` using [BindLiteral(LiteralToken literalToken)](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/Binders/LiteralBinder.cs#L21). The AspNetCore library will get the `cast` node and tries to create LINQ Expression using [Expression BindCastSingleValue(SingleValueFunctionCallNode node, QueryBinderContext context)](https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs#L562). It knows how to handle `ConstantNode`

For `cast` with `unquoted type parameter`
- ODL will bind `DottedIDentifierToken` as `SingleResourceCastNode` using a method I have added with this PR. To support this in AspNetCore, I will update function [Expression BindCastSingleValue(SingleValueFunctionCallNode node, QueryBinderContext context)](https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs#L562) to add a condition to cast QueryNode to SingleResourceCastNode like what was done with the PR: https://github.com/OData/AspNetCoreOData/pull/1313

WanjohiSammy: - For `quoted type parameter`, ODL bind the token to `ConstantNode` and AspNetCore is responsible to create LINQ expression using [Expression BindCastSingleValue(SingleValueFunctionCallNode node, QueryBinderContext context)](https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs#L562). It knows how to handle `ConstantNode`.
- For `unquoted type parameter`, ODL bind the token to `SingleResourceCastNode`. When bumping the ODL version in AspNetCore, I will have to update the method [Expression BindCastSingleValue(SingleValueFunctionCallNode node, QueryBinderContext context)](https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs#L562) to allow cast to `SingleResourceCastNode` as well.

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
WanjohiSammy: Same applied to `isof` as discussed in https://github.com/OData/odata.net/pull/3117#discussion_r1840468666, however, the method in `AspNetCore` that is responsible to create LINQ is different. Here is the method [Expression BindIsOf(SingleValueFunctionCallNode node, QueryBinderContext context)](https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.SingleValueFunctionCall.cs#L637). I will also update this method to allow cast to `SingleResourceCastNode` as it currently only support cast to `ConstantNode`

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
habbes: Thanks for this added context. Can you also shed light on the behaviour with respect to handling of types where a cast is not possible? Ideally, from the user's endpoint the behaviour should be the same regardless of whether they user quoted or non-quoted syntax, and regardless of our internal implementation details which you have highlighted.

So could you clarify how we handle both supported and invalid cast scenarios in both quoted and unquoted variants? In which cases are we throwing an exception? Is the exception thrown consistent? Is that the expected behaviour?

My expectation is that `isof` should not throw an exception when a cast is invalid, but rather evaluate to a boolean that will perform the correct filter. I think `cast` throwing an exception when a cast is invalid is reasonable. What is not clear to me is whether such an exception should be thrown in ODL or AspNetCore. For the case of `cast` I would be cautious about changing the existing behaviour to avoid introducing a breaking change.

WanjohiSammy: With the current behavior, ODL does not throw exception or to be more specific, it doesn't check if target type is related to source type. It only bind the `quoted type parameters` to `constant node` and return the entire token as FunctionCall token. The `AspNetCore`, on the other end, will get the token and try to create LINQ expression. If expression is not possible, AspNetCore returns null [BindSingleResourceCastFunctionCall](https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs#L655) will return `null` expression. 

Let's look at this query:
```
/products?$filter=cast('Microsoft.AspNetCore.OData.E2E.Tests.Cast.Category')/Name eq 'Cat
```

AspNetCore will try to relate `Entity Product` to `Entity Category` in [BindSingleResourceCastFunctionCall](https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs#L655) and returns null expression if one is not assignable from the other.

And then [CreatePropertyAccessExpression](https://github.com/OData/AspNetCoreOData/blob/main/src/Microsoft.AspNetCore.OData/Query/Expressions/QueryBinder.cs#L1114) will throw an expression when trying to get the Property Name from a null expression.

image

An exception like below will be thrown:
```
 Message: 
System.ArgumentException : Instance property 'Name' is not defined for type 'System.Object' (Parameter 'propertyName')

  Stack Trace: 
Expression.Property(Expression expression, String propertyName)
QueryBinder.CreatePropertyAccessExpression(Expression source, QueryBinderContext context, IEdmProperty property, String propertyPath) line 1124
QueryBinder.BindPropertyAccessQueryNode(SingleValuePropertyAccessNode propertyAccessNode, QueryBinderContext context) line 360
QueryBinder.BindSingleValueNode(SingleValueNode node, QueryBinderContext context) line 147
QueryBinder.Bind(QueryNode node, QueryBinderContext context) line 73
QueryBinder.BindBinaryOperatorNode(BinaryOperatorNode binaryOperatorNode, QueryBinderContext context) line 453
QueryBinder.BindSingleValueNode(SingleValueNode node, QueryBinderContext context) line 132
QueryBinder.Bind(QueryNode node, QueryBinderContext context) line 73
FilterBinder.BindFilter(FilterClause filterClause, QueryBinderContext context) line 39
```

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
```diff
@@ -426,6 +436,52 @@ private bool TryBindIdentifier(string identifier, IEnumerable<FunctionParameterT
...
+                }
+            }
+
+            IEdmSchemaType childType = UriEdmHelpers.FindTypeFromModel(state.Model, dottedIdentifierToken.Identifier, this.Resolver);
+            IEdmStructuredType childStructuredType = childType as IEdmStructuredType;
+
+            if (childStructuredType == null)
+            {
+                return this.bindMethod(dottedIdentifierToken);
+            }
```
marabooy, line 470:
```suggestion

            if ( childType is IEdmStructuredType childStructuredType)
            {
                return this.bindMethod(dottedIdentifierToken);
            }

```
nit: Use is operator instead

WanjohiSammy: Thanks for this. Implemented

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
```diff
@@ -409,20 +409,24 @@ public void ParseFilterWithNullEnumValue()
             convertNode.Source.ShouldBeConstantQueryNode((object)null);
         }
 
-        [Fact]
-        public void ParseFilterCastMethod1()
+        [Theory]
+        [InlineData("cast(NS.Color'Green', 'Edm.String') eq 'blue'")]
+        [InlineData("cast(NS.Color'Green', Edm.String) eq 'blue'")]
+        public void ParseFilterCastMethod1(string filterQuery)
```
xuzhg, line 415: The test name contains XXX1 AND XXX2 doesn't make sense. Can you rename them?

And can combine 'ParseFilterCastMethod1' and 'ParseFilterCastMethod2'?

xuzhg (approved):

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
```diff
@@ -234,7 +234,27 @@ internal QueryNode BindFunctionCall(FunctionCallToken functionCallToken)
 
             // If there isn't, bind as Uri function
             // Bind all arguments
-            List<QueryNode> argumentNodes = new List<QueryNode>(functionCallToken.Arguments.Select(ar => this.bindMethod(ar)));
+            List<QueryNode> argumentNodes = functionCallToken.Arguments.Select(argument =>
+            {
+                // If the function is IsOf or Cast and the argument is a dotted identifier, we need to bind it differently
+                if (UnboundFunctionNames.Contains(functionCallToken.Name) && argument.ValueToken is DottedIdentifierToken dottedIdentifier)
```
habbes, line 240: Does `UnboundFunctionNames` check for `isof` and `cast`?

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
```diff
@@ -234,7 +234,27 @@ internal QueryNode BindFunctionCall(FunctionCallToken functionCallToken)
...
+            List<QueryNode> argumentNodes = functionCallToken.Arguments.Select(argument =>
+            {
+                // If the function is IsOf or Cast and the argument is a dotted identifier, we need to bind it differently
+                if (UnboundFunctionNames.Contains(functionCallToken.Name) && argument.ValueToken is DottedIdentifierToken dottedIdentifier)
+                {
+                    // Find the type of the dotted identifier by resolving it against the model. This also ensure case-insensitive resolution.
+                    IEdmSchemaType dottedIdentifierType = UriEdmHelpers.FindTypeFromModel(state.Model, dottedIdentifier.Identifier, this.Resolver);
+
+                    // If the dotted identifier is a primitive type and the next token is not null, we need to ensure next token is null
+                    if (dottedIdentifierType is IEdmPrimitiveType && dottedIdentifier.NextToken != null)
```
habbes, line 246: Why do we need to ensure that the next token is null?

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
```diff
@@ -234,7 +234,27 @@ internal QueryNode BindFunctionCall(FunctionCallToken functionCallToken)
...
+                    {
+                        // Push the current argument onto the stack to keep track of the previous argument
+                        dottedIdentifier.NextToken = null;
```
habbes, line 249: Is it safe to modify tokens like this? Is this something we do elsewhere in the code?

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
```diff
@@ -426,6 +446,50 @@ private bool TryBindIdentifier(string identifier, IEnumerable<FunctionParameterT
...
+        /// <summary>
+        /// Binds a <see cref="DottedIdentifierToken"/> for the 'isof' and 'cast' function calls.
+        /// </summary>
+        /// <param name="dottedIdentifierToken">The dotted identifier token to bind.</param>
+        /// <param name="childType">The child type to bind to.</param>
+        /// <returns>A <see cref="QueryNode"/> representing the bound single resource node.</returns>
+        /// <exception cref="ODataException">Thrown when the token cannot be bound as a single resource node.</exception>
+        private QueryNode TryBindDottedIdentifierForIsOfOrCastFunctionCall(DottedIdentifierToken dottedIdentifierToken, IEdmSchemaType childType)
+        {
+            QueryNode parent = null;
```
habbes, line 458: Could you explain what is going on in this method?

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
WanjohiSammy: Consider the following:
- `cast` or `isof` with 2 parameters. For example, cast(ID, Edm.String):
   - The parser will create a `FunctionCallToken` with 2 arguments where the first argument (`ID`) is an `EndPathToken` and the second argument (`Edm.String`) is a `DottedIdentifierToken` because it is unquoted. When the type parameter is quoted, the second argument becomes a `LiteralToken`.
   - The second argument's Next Token will point to the first argument. This is important for determining if the type of the second argument can be assigned from the type of the first argument.
   - For primitive DottedIdentifier types like Edm.String, Edm.Int32, Edm.Byte (including their lowercase variants), the presence of NextToken will cause the DottedIdentifier to bind as a `SingleValueCastNode`, leading to potential breaking changes as reported in issue https://github.com/OData/odata.net/issues/3123
   
- `cast` or `isof` with 1 parameter. For example, cast(Edm.String):
  - The parser will create a `FunctionCallToken` with 1 argument where the argument (`Edm.String`) is a `DottedIdentifierToken`. This argument does not have a NextToken.

Here we are checking DottedIdentifier is primitive type and next token is null to ensure we handle the primitive types properly and cases where argument is 1.

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
WanjohiSammy: Instead of `dottedIdentifier.NextToken = null` I have opted to do the following:
```cs
// If the next token is not null, we need to ensure it is null
dottedIdentifier = new DottedIdentifierToken(dottedIdentifier.Identifier, null);
```
This is done in [ParseIdentifier(QueryToken parent)](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/Parsers/IdentifierTokenizer.cs#L71) where `parent` is the `NextToken`  and for DottedIdentifier, parent is always null

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
```diff
@@ -234,7 +234,28 @@ internal QueryNode BindFunctionCall(FunctionCallToken functionCallToken)
 
             // If there isn't, bind as Uri function
             // Bind all arguments
-            List<QueryNode> argumentNodes = new List<QueryNode>(functionCallToken.Arguments.Select(ar => this.bindMethod(ar)));
+            List<QueryNode> argumentNodes = functionCallToken.Arguments.Select(argument =>
+            {
+                // If the function is IsOf or Cast and the argument is a dotted identifier, we need to bind it differently
+                string functionCallTokenName = IsUnboundFunction(functionCallToken.Name);
```
xuzhg, line 240: IsUnboundFunction returns a string? It looks non-sense.

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
```diff
@@ -234,7 +234,28 @@ internal QueryNode BindFunctionCall(FunctionCallToken functionCallToken)
...
+                if ((ExpressionConstants.UnboundFunctionIsOf == functionCallTokenName || ExpressionConstants.UnboundFunctionCast == functionCallTokenName) && argument.ValueToken is DottedIdentifierToken dottedIdentifier)
+                {
+                    // Find the type of the dotted identifier by resolving it against the model. This also ensure case-insensitive resolution.
+                    IEdmSchemaType dottedIdentifierType = UriEdmHelpers.FindTypeFromModel(state.Model, dottedIdentifier.Identifier, this.Resolver);
+
+                    // If the dotted identifier is a primitive type and the next token is not null, we need to ensure next token is null
+                    if (dottedIdentifierType is IEdmPrimitiveType && dottedIdentifier.NextToken != null)
+                    {
+                        // If the next token is not null, we need to ensure it is null
```
xuzhg, line 249: This comment line doesn't make sense?

xuzhg (approved):

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
WanjohiSammy: `IsUnboundFunction` gets `functionCallToken.Name` in any case (lower-case, upper case or camel case) and compare with the `isof` and `cast` using `this.state.Configuration.EnableCaseInsensitiveUriFunctionIdentifier` config

[private string IsUnboundFunction(string functionName)](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs#L635)

[private static readonly string[] UnboundFunctionNames](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs#L29)

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
```diff
@@ -758,7 +765,7 @@ private static IEdmTypeReference ValidateIsOfOrCast(BindingState state, bool isC
             {
                 // throw if cast enum to not-string :
                 if ((args[0].GetEdmTypeReference() is IEdmEnumTypeReference)
-                    && !string.Equals(typeArgument.Value as string, Microsoft.OData.Metadata.EdmConstants.EdmStringTypeName, StringComparison.Ordinal))
+                    && !string.Equals(typeArgumentFullName, Microsoft.OData.Metadata.EdmConstants.EdmStringTypeName, state.Configuration.Resolver.EnableCaseInsensitive ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal))
```
gathogojr, line 768: Can `state.Configuration.Resolver.` be null?

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
```diff
@@ -33,6 +34,11 @@ internal sealed class FunctionCallParser : IFunctionCallParser
         /// </summary>
         private readonly bool restoreStateIfFail;
 
+        /// <summary>
+        /// If the function call is cast or isof, set to true.
+        /// </summary>
+        private bool isFunctionCallNameCastOrIsOf = false;
```
xuzhg, line 40: Why do you use the private to cache a transient state?
If it's a state/signal within the parser, please find a different way to save/cache it,?

mikepizzo:
@WanjohiSammy -- I ran into this last week, and it may shortly be blocking some agentic API work.  What's the status of getting this fix in?

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
WanjohiSammy: It can be null. In case the [DI container is nul](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs#L567)l, the Resolver will always be an instance of `ODataUriResolver()`:

https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs#L25

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
WanjohiSammy: Instead of having to cache, I have passed it as argument.

WanjohiSammy:
> I ran into this last week, and it may shortly be blocking some agentic API work. What's the status of getting this fix in?

@mikepizzo This is ready. I will remind the team to review.

gathogojr:
@WanjohiSammy I had a question regarding this part of your pull request description:

> According to the https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_cast and https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_isof specifications, there is no indication that an exception should be thrown if casting fails. This pull request introduces the `TryBindDottedIdentifierForIsOfOrCastFunctionCall` method, which handles the binding of `DottedIdentifierToken` specifically for the `isof` and `cast` function calls without performing the `CheckRelatedTo` validation, thereby preventing exceptions when type casting is not possible.

I understand the intent to align with the spec and avoid throwing exceptions on failed casts. However, I’m wondering about the practical implications of this behavior.

For example, consider the following scenario:

```csharp
public class ProductsController : ODataController
{
    [EnableQuery]
    public ActionResult<IEnumerable<Product>> Get()
    {
        return new List<Product>
        {
            new Product { Id = 1, Name = "Banana" },
            new Product { Id = 2, Name = "Orange" },
            new Product { Id = 3, Name = "Apple" }
        };
    }
}
```

With the following model:

```csharp
namespace NS
{
    public class Product
    {
        public int Id { get; set; }
    }

    public class DerivedProduct : Product
    {
        public DateTimeOffset ExpiryDate { get; set; }
    }

    public class Category
    {
        public int Id { get; set; }
    }
}
```

Now, if someone runs:

```
/Products?$filter=isof(NS.DerivedProduct)
/Products?$filter=isof(NS.Category)
```

Both filters would yield the same result—an empty set—because there are no `DerivedProduct` instances in the collection, and `Category` is unrelated to `Product`.

While this behavior is technically correct per the spec, it raises a usability concern: if a user mistakenly writes `isof(NS.Category)` thinking it's a valid filter, the system silently returns no results. This could lead to confusion or debugging difficulties, especially when the type is completely unrelated.

I understand the idea of staying spec-compliant but how can we help users catch likely mistakes when the cast target is unrelated to the base type?

Curious to hear your thoughts on this trade-off.
cc @mikepizzo

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
```diff
@@ -720,12 +720,19 @@ private static IEdmTypeReference ValidateIsOfOrCast(BindingState state, bool isC
                     Error.Format(SRResources.MetadataBinder_CastOrIsOfExpressionWithWrongNumberOfOperands, args.Count));
             }
 
-            ConstantNode typeArgument = args.Last() as ConstantNode;
+            QueryNode queryNode = args.Last();
```
gathogojr: Given the `args.Count` above this line, that suggests to me that we're dealing with a `List<T>`. I think we can avoid using LINQ here

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
gathogojr: The reason I was asking if it can be null is because `state.Configuration.Resolver.EnableCaseInsensitive` would throw an exception if it was `state.Configuration.Resolver` is null. Maybe we could change the expression to
```csharp
state.Configuration.Resolver?.EnableCaseInsensitive == true ? StringComparison.OrdinalIgnoreCase : StringComparison.Ordinal
```

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
```diff
@@ -161,33 +163,54 @@ public FunctionParameterToken[] ParseArgumentListOrEntityKeyList(Action restoreA
...
+        /// <param name="functionName">The name of the function being called. Default is an empty span.</param>
         /// <returns>A list of FunctionParameterTokens representing each argument</returns>
-        private List<FunctionParameterToken> ReadArgumentsAsPositionalValues()
+        private List<FunctionParameterToken> ReadArgumentsAsPositionalValues(ReadOnlySpan<char> functionName = default)
         {
+            // Store the parent expression of the current argument.
+            Stack<QueryToken> expressionParents = new Stack<QueryToken>();
+            bool isFunctionCallNameCastOrIsOf = functionName.Length > 0 &&
+                functionName.SequenceEqual(ExpressionConstants.UnboundFunctionCast.AsSpan()) || functionName.SequenceEqual(ExpressionConstants.UnboundFunctionIsOf.AsSpan());
+
```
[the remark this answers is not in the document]
gathogojr: @WanjohiSammy Are you going to consider this suggestion by Copilot?

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
```diff
@@ -213,5 +236,37 @@ private bool TryReadArgumentsAsNamedValues(out ICollection<FunctionParameterToke
...
+        /// <param name="parameterToken">The current parameter token.</param>
+        /// <returns>The updated parameter token.</returns>
+        private static QueryToken SetParentForCurrentParameterToken(QueryToken parentExpression, QueryToken parameterToken)
+        {
+            // If the parameter is a dotted identifier, we need to set the parent of the next argument to the current argument.
+            // But if it is primitive literal, no need to set the parent.
+            if (parameterToken is DottedIdentifierToken dottedIdentifierToken && dottedIdentifierToken.NextToken == null)
+            {
+                // Check if the dottedIdentifier is a primitive type
+                IEdmSchemaElement schemaElement = EdmCoreModel.Instance.SchemaElements.Where(e => string.Equals(dottedIdentifierToken.Identifier, e.FullName(), StringComparison.OrdinalIgnoreCase)).FirstOrDefault();
```
gathogojr: Do we really need to loop through all the schema elements in the EdmCoreModel to confirm if `dottedIdentifierToken.Identifier` is primitive? I think in ODL there are checks for primitive that don't involve scanning through the EdmCoreModel schema elements using LINQ.

gathogojr: Just realized that identifier will look something like this:
image

All the same, I think we should find a more efficient way of evaluating that. Maybe we can maintain an unordered collection (e.g. `new Dictionary<string, IEdmSchemaElement>(StringComparer.OrdinalIgnoreCase);`) of primitive identifiers (in a class variable) that we initialize at first use and subsequently reuse it for similar checks. Use of `EdmCoreModel.Instance.SchemaElements` feel a bit too suboptimal for me.

If we don't go the class variable way, we could at least use `EdmCoreModel.Instance.FindDeclaredType(dottedIdentifierToken.Identifier)` if it can handle case-insensitive search

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
```diff
@@ -213,5 +236,37 @@ private bool TryReadArgumentsAsNamedValues(out ICollection<FunctionParameterToke
...
+                //  isof(MyAddress, NS.Address) -> NS.Address is a dottedIdentifierToken but not a primitive type
+                if (schemaElement is null or not IEdmPrimitiveType)
+                {
+                    dottedIdentifierToken.NextToken = parentExpression;
+                    parameterToken = dottedIdentifierToken;
+                }
+            }
+
+            return parameterToken;
+        }
```
gathogojr:
```csharp
private static QueryToken SetParentForCurrentParameterToken(QueryToken parentExpression, QueryToken parameterToken)
{
    // If the parameter is not a DottedIdentifierToken or already has a NextToken
    if (parameterToken is not DottedIdentifierToken dottedIdentifierToken || dottedIdentifierToken.NextToken is not null)
    {
        return parameterToken;
    }

    // Check if the identifier is a primitive type
    IEdmSchemaElement schemaElement = EdmCoreModel.Instance.SchemaElements
        .FirstOrDefault(e => string.Equals(dottedIdentifierToken.Identifier, e.FullName(), StringComparison.OrdinalIgnoreCase));

    // If the identifier is a primitive type
    if (schemaElement is IEdmPrimitiveType)
    {
        return parameterToken;
    }

    // Set the parent of the next argument to the current argument
    dottedIdentifierToken.NextToken = parentExpression;
    return dottedIdentifierToken;
}
```

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
WanjohiSammy: Updated to use the suggestion

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
WanjohiSammy: Use the below instead of LINQ:
```c#
if (args.Count == 1)
{
    queryNode = args[0];
}
else
{
    queryNode = args[1];
}
```

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
WanjohiSammy: Updated

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
```diff
@@ -720,12 +720,27 @@ private static IEdmTypeReference ValidateIsOfOrCast(BindingState state, bool isC
...
-            ConstantNode typeArgument = args.Last() as ConstantNode;
+            QueryNode queryNode;
+            if (args.Count == 1)
+            {
+                queryNode = args[0];
+            }
+            else
+            {
+                queryNode = args[1];
+            }
```
gathogojr: @WanjohiSammy `args.Last()` would be equivalent to `args[args.Count - 1]`, no? Or is there a chance or having more than 2 elements in the args collection such that we want to particularly target the element at index 1 in all other cases where there's more than 1 elements?

test/FunctionalTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/EnumFilterFunctionalTests.cs
WanjohiSammy: Updated

src/Microsoft.OData.Core/UriParser/Binders/FunctionCallBinder.cs
WanjohiSammy: Since `isof` and `cast` can have 1 or 2 parameters where the last parameter is the type, we can use your suggestion `args[args.Count - 1]` or `args[^1]` using the ["hat" operator (^)](https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/proposals/csharp-8.0/ranges#overview)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
WanjohiSammy: @gathogojr By doing a simple benchmark, it seems like the loop is faster and memory efficient:
image

### Benchmark code:
```cs
[MemoryDiagnoser]
[CPUUsageDiagnoser]
public class MyBenchmarks
{
    [Params("", "edm.int64", "edm.string", "Edm.String", "EDM.decimal", "EDM.GUID", "fully.qualified.ns")]
    public string TypeName { get; set; }

    [Benchmark(Baseline = true)]
    public object? Using_EdmCoreModel_CoreSchemaElements()
    {
        return EdmCoreModel.Instance.SchemaElements.FirstOrDefault(se => string.Equals(se.Namespace, TypeName, StringComparison.OrdinalIgnoreCase));
    }

    [Benchmark]
    public object? Using_NormalizedModelElementsCache_FindSchemaTypes()
    {
        var cache = new NormalizedModelElementsCache(EdmCoreModel.Instance);
        var schemaTypes = cache.FindSchemaTypes(TypeName);
        return schemaTypes?.FirstOrDefault();
    }

    [Benchmark]
    public object? Using_ODataUriResolver()
    {
        var resolver = new ODataUriResolver { EnableCaseInsensitive = true };
        var edmSchemaType = resolver.ResolveType(EdmCoreModel.Instance, TypeName);
        return edmSchemaType;
    }
}
```

WanjohiSammy:
> Both filters would yield the same result—an empty set—because there are no `DerivedProduct` instances in the collection, and `Category` is unrelated to `Product`.
> 
> While this behavior is technically correct per the spec, it raises a usability concern: if a user mistakenly writes `isof(NS.Category)` thinking it's a valid filter, the system silently returns no results. This could lead to confusion or debugging difficulties, especially when the type is completely unrelated.
> 
> I understand the idea of staying spec-compliant but how can we help users catch likely mistakes when the cast target is unrelated to the base type?
> 
> Curious to hear your thoughts on this trade-off. cc @mikepizzo

@gathogojr I have updated the PR description of what the change is about. I have also added tests to ensure that exceptions are thrown when the types do not match.

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
habbes: @WanjohiSammy  this benchmark is flawed because in the `Using_NormalizedModelElementsCache_FindSchemaTypes` you're creating a new instance of the `NormalizedModelElementsCache`, which will get created a new for each run, whereas in the first method you seem to be using a singleton, in the third method it's probably also using a singleton cache internally. If you want a fairer comparison, the `NormalizedModelElementsCache` should be a singleton initialized outside of the benchmark method. You should probably also create the `ODataUriResolver` instance outside. I'd be curious to see the CPU profile session and verify whether the majority of the runtime is not spent on building the cache rather than performing the search.

habbes:
@WanjohiSammy the issues listed in the description that this PR aims to fix happen to be closed. What is the exact behaviour that this PR aims to change or rectify? Is it to support case-insensitivity in the `isof` and `cast` operator or supporting `isof` on unrelated types without throwing an error? It's not clear to me from reading the description what did not work before vs what now works after this PR.

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
```diff
@@ -161,33 +163,53 @@ public FunctionParameterToken[] ParseArgumentListOrEntityKeyList(Action restoreA
...
         /// <summary>
         /// Read the list of arguments as a set of positional values
         /// </summary>
+        /// <param name="functionName">The name of the function being called. Default is an empty span.</param>
         /// <returns>A list of FunctionParameterTokens representing each argument</returns>
-        private List<FunctionParameterToken> ReadArgumentsAsPositionalValues()
+        private List<FunctionParameterToken> ReadArgumentsAsPositionalValues(ReadOnlySpan<char> functionName = default)
         {
+            // Store the parent expression of the current argument.
+            Stack<QueryToken> expressionParents = new Stack<QueryToken>();
```
habbes, line 187: nit: Since there are likely to be few arguments in most cases, and this stack is not used outside of this method, might be an option to consider a stack allocated array or a struct-based stack that only allocates to the heap once it reaches a certain size threshold (and usually we don't get functions with > 3/4 args). Might be overkill for this PR, but I think there a couple of places where we build this short-lived single-scoped stacks that may end up causing avoidable allocations in the hot path. We can talk more about this if interested.

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
WanjohiSammy: @habbes The real scenario is that we will need to initialize `NormalizedModelElementsCache` or `ODataUriResolver`, that is why the initialization is not outside the benchmark.

WanjohiSammy: Added a `singleton` for `EdmCoreModel` in `NormalizedModelElementsCache` has significantly reduce the 

image

### Note
- Using of `ODataResolver` still performs poorly even if the initialization is done outside the benchmark.
- `EdmCoreModelInstance` is a singleton specifically for `EDM core model`
   ```cs
   public static NormalizedModelElementsCache EdmCoreModelCacheInstance = new NormalizedModelElementsCache(EdmCoreModel.Instance);
  ```

### Benchmarks code
```cs
[MemoryDiagnoser]
[CPUUsageDiagnoser]
public class MyBenchmarks
{
    [Params("", "edm.int64", "edm.string", "Edm.String", "EDM.decimal", "EDM.GUID", "fully.qualified.ns")]
    public string TypeName { get; set; }

    private ODataUriResolver Resolver { get; set; }

    [GlobalSetup]
    public void Setup()
    {
        Resolver = new ODataUriResolver { EnableCaseInsensitive = true };
    }

    [Benchmark(Baseline = true)]
    public object? Using_EdmCoreModel_CoreSchemaElements()
    {
        return EdmCoreModel.Instance.SchemaElements.FirstOrDefault(se => string.Equals(se.Namespace, TypeName, StringComparison.OrdinalIgnoreCase));
    }

    [Benchmark]
    public object? Using_NormalizedModelElementsCache_FindSchemaTypes()
    {
        var schemaTypes = NormalizedModelElementsCache.EdmCoreModelInstance.FindSchemaTypes(TypeName);
        return schemaTypes?.FirstOrDefault();
    }

    [Benchmark]
    public object? Using_ODataUriResolver()
    {
        var edmSchemaType = Resolver.ResolveType(EdmCoreModel.Instance, TypeName);
        return edmSchemaType;
    }
}
```

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
habbes: Here's a sample implementation (I asked Copilot to generate the code, might contain errors, but you hopefully get the idea)

```c#

using System;

using System.Runtime.CompilerServices;

[InlineArray(4)]
public struct InlineBuffer<T>
{
    private T _element0; // Compiler generates _element1, _element2, _element3
}

public struct StructStack<T>
{
    private InlineBuffer<T> _inline;
    private List<T>? _overflow;
    private int _count;
    private int _overflowCount;

    private const int InlineCapacity = 4;

    public int Count => _count;

    public void Push(T item)
    {
        if (_count < InlineCapacity)
        {
            _inline[_count] = item;
        }
        else
        {
            _overflow ??= new List<T>(4);
            if (_overflowCount < _overflow.Count)
            {
                _overflow[_overflowCount] = item;
            }
            else
            {
                _overflow.Add(item);
            }
            _overflowCount++;
        }

        _count++;
    }

    public T Pop()
    {
        if (_count == 0)
            throw new InvalidOperationException("Stack is empty.");

        _count--;

        if (_count < InlineCapacity)
        {
            return _inline[_count];
        }
        else
        {
            _overflowCount--;
            return _overflow![_overflowCount];
        }
    }

    public T Peek()
    {
        if (_count == 0)
            throw new InvalidOperationException("Stack is empty.");

        if (_count <= InlineCapacity)
        {
            return _inline[_count - 1];
        }
        else
        {
            return _overflow![_overflowCount - 1];
        }
    }
}
```

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
```diff
@@ -213,5 +235,36 @@ private bool TryReadArgumentsAsNamedValues(out ICollection<FunctionParameterToke
...
+        /// <returns>The updated parameter token.</returns>
+        private static QueryToken SetParentForCurrentParameterToken(QueryToken parentExpression, QueryToken parameterToken)
+        {
+            if (parameterToken is not DottedIdentifierToken dottedIdentifierToken || dottedIdentifierToken?.NextToken is not null)
+            {
+                return parameterToken;
+            }
+
+            // Check if the identifier is a primitive type
+            IEdmSchemaElement schemaElement = EdmCoreModel.Instance.SchemaElements
```
habbes: Is there where it's been suggested to do a cache look-up? How many elements do we have in the core model schema elements?

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
WanjohiSammy: I have something similar. I will push and let you review.

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
WanjohiSammy: @habbes Around 40 items. Though the code you are referring to is outdated . I have opted to use a singleton in `NormalizedModelElementsCache` as indicated in this comment https://github.com/OData/odata.net/pull/3117#discussion_r2260292632

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
WanjohiSammy: @habbes I have added the [StackStruct](https://github.com/OData/odata.net/pull/3117/files#diff-5dfd5cfa37315d4929c9b561905189f98615e9da6692511ef62515daaa2b6cea)

src/Microsoft.OData.Core/UriParser/Parsers/FunctionCallParser.cs
habbes: Interesting, the speedup is even more significant than I expected. I also find it curious that the ODataUriResolver is faster only when the type name is Edm.String.

src/Microsoft.OData.Core/StackStructOfT.cs
```diff
@@ -0,0 +1,175 @@
...
+        return Equals(other);
+    }
+
+    /// <summary>
+    /// Determines whether the current stack is equal to another <see cref="StackStruct{T}"/>.
+    /// Two stacks are equal if they have the same count and all elements are equal in order.
+    /// </summary>
+    /// <param name="other">The stack to compare with the current stack.</param>
+    /// <returns>true if the stacks are equal; otherwise, false.</returns>
+    public readonly bool Equals(StackStruct<T> other)
```
habbes: Is this a feature we actually need or use?

src/Microsoft.OData.Core/StackStructOfT.cs
```diff
@@ -0,0 +1,175 @@
...
+using Microsoft.OData.Core;
+
+namespace Microsoft.OData;
+
+/// <summary>
+/// A lightweight, struct-based generic stack implementation for value and reference types.
+/// Provides efficient push, pop, and peek operations with dynamic resizing.
+/// </summary>
+/// <typeparam name="T">The type of elements in the stack.</typeparam>
+internal struct StackStruct<T>
```
habbes: This is only light-weight in the wrapper `StackStruct` is a struct and not an object. So it may save one allocation, but you still use a heap allocated array even for the base case. So It's not really much different from a regular `Stack` and I don't think it's worth creating a custom type.

This is different from the version I proposed which would use an inline array buffer (e.g. using `InlineArray` attribute, could use `fixed` array buffer, but the latter is `unsafe` and limited to simple primitive types).

But I wouldn't want to block this PR because of this. My suggestion would be to remove this altogether (since it doesn't provide much optimization over a standard Stack, then we can think of a lightweight stack helper in a separate PR that can be used across the codebase)

src/Microsoft.OData.Core/UriParser/Resolver/NormalizedModelElementsCache.cs
```diff
@@ -20,6 +20,11 @@ namespace Microsoft.OData.Edm
     /// </summary>
     internal sealed class NormalizedModelElementsCache
     {
+        /// <summary>
+        /// Provides a shared, normalized case-insensitive cache of model elements for the EDM core model.
+        /// </summary>
+        public static NormalizedModelElementsCache EdmCoreModelCacheInstance = new NormalizedModelElementsCache(EdmCoreModel.Instance);
```
habbes: Make it readonly.

```suggestion
        public static readonly NormalizedModelElementsCache EdmCoreModelCacheInstance = new NormalizedModelElementsCache(EdmCoreModel.Instance);

```

nit: since this is already in a class called cache, we could simplify the model. But feel free to ignore this part

```suggestion
        public static readonly NormalizedModelElementsCache EdmCoreModelInstance = new NormalizedModelElementsCache(EdmCoreModel.Instance);
```

or 

```suggestion
        public static readonly NormalizedModelElementsCache EdmCoreModel = new NormalizedModelElementsCache(EdmCoreModel.Instance);

```

src/Microsoft.OData.Core/StackStructOfT.cs
WanjohiSammy: I have reverted to use the regular `stack`

src/Microsoft.OData.Core/StackStructOfT.cs
WanjohiSammy: @habbes I have reverted to use the regular `stack`

habbes (approved):

Issue #3092, opened 2024-10-17
ArgumentNullException for: $filter=property in ['']


sriram-d:
Issue is similar to #2539, except for the use of square brackets rather than parenthesis. Also, issue happens for any regular property.

### Assemblies affected
ODL version 8.0.1

### Reproduce steps
```C#
// Setup basic model
var model = new EdmModel();
var user = model.AddEntityType("graph", "user", baseType: null, isAbstract: false, isOpen: true, hasStream: false);
var container = model.AddEntityContainer("graph", "service");
container.AddEntitySet("users", user);

// Setup URI parser
var parser = new ODataUriParser(model, new System.Uri("/users?$filter=openPropertyFoo in ('')", System.UriKind.Relative));
parser.ParsePath();

// This throws an ArgumentNullException:
var filter = parser.ParseFilter();
```

### Expected result
Since syntax is not as expected, throw ODataException.

### Actual result
Below exception is thrown.
```
System.ArgumentNullException: Value cannot be null or empty. (Parameter 'literalText')
   at Microsoft.OData.ExceptionUtils.CheckArgumentStringNotNullOrEmpty(String value, String parameterName)
   at Microsoft.OData.UriParser.ConstantNode..ctor(Object constantValue, String literalText, IEdmTypeReference typeReference)
   at Microsoft.OData.UriParser.CollectionConstantNode..ctor(IEnumerable`1 objectCollection, String literalText, IEdmCollectionTypeReference collectionType)
   at Microsoft.OData.UriParser.LiteralBinder.BindInLiteral(LiteralToken literalToken)
   at Microsoft.OData.UriParser.MetadataBinder.<BindIn>b__22_0(QueryToken queryToken)
   at Microsoft.OData.UriParser.InBinder.GetCollectionOperandFromToken(QueryToken queryToken, IEdmTypeReference expectedType, IEdmModel model)
   at Microsoft.OData.UriParser.InBinder.BindInOperator(InToken inToken, BindingState state)
   at Microsoft.OData.UriParser.MetadataBinder.BindIn(InToken inToken)
   at Microsoft.OData.UriParser.MetadataBinder.Bind(QueryToken token)
   at Microsoft.OData.UriParser.FilterBinder.BindFilter(QueryToken filter)
   at Microsoft.OData.UriParser.ODataQueryOptionParser.ParseFilterImplementation(String filter, ODataUriParserConfiguration configuration, ODataPathInfo odataPathInfo)
[...]

PR #3122, opened 2024-11-15
Fix ArgumentNullException for: $filter=property in ['']

WanjohiSammy:
### Issues

*This pull request fixes [#3092 ](https://github.com/OData/odata.net/issues/3092)*

### Description

This PR addresses an issue where the OData URI parser throws an `ArgumentNullException` when encountering a list with empty string in `square brackets`.

For example, the following query will throw an ArgumentNullException:
- `/People?$filter=Name in ['']`
- `/People?$filter=Name in [ ' ' ]`
- `/People?$filter=Name in [ \"\", ' ' ]`
- `/People?$filter=Name in [ \"\", '' ]`
- `/People?$filter=Name in [\"\"]`
- `/People?$filter=Name in [ \"\", \" \" ]`
- `/People?$filter=Name in [ \"\" ]`
- `/People?$filter=Name in [  '' , ' ' ]`

```
System.InvalidOperationException : An error occurred while processing this request.
---- Microsoft.OData.Client.DataServiceTransportException : System.ArgumentNullException: Value cannot be null or empty. (Parameter 'literalText')
   at Microsoft.OData.ExceptionUtils.CheckArgumentStringNotNullOrEmpty(String value, String parameterName) in C:\Work\odata.net\src\Microsoft.OData.Core\ExceptionUtils.cs:line 101
   .....
```
### Change

This change ensures the Empty String in `square brackets` are handle the same way as `parentheses` when constructing `CollectionNode` in `GetCollectionOperandFromToken` method. This method calls the following methods to handle Single and Double quotes:
 - `ProcessSingleQuotedStringItem` is called to handle single-quoted string items within a collection. This method converts `single-quoted` strings to `double-quoted` strings, ensuring compatibility with JSON format. It manages the escaping of single quotes by unescaping double single quotes and escaping double quotes. For example, it converts ''' to \"\" and 'ghi''' to "ghi'"
 - `ProcessDoubleQuotedStringItem` is then called to process double-quoted string items within a collection. This method ensures that the string is properly escaped for JSON format, handling the escaping of double quotes and backslashes. For example, it converts "" to \"\" to avoid passing an empty string to the `ConstantNode`.

- [x] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -104,13 +104,10 @@ namespace Microsoft.OData.UriParser
 
                 if (bracketLiteralText[0] == '(' || bracketLiteralText[0] == '[')
                 {
-                    bool isParenthesesBased = bracketLiteralText[0] == '(' && bracketLiteralText[^1] == ')';
-                    bool isBracketBased = bracketLiteralText[0] == '[' && bracketLiteralText[^1] == ']';
+                    Debug.Assert((bracketLiteralText[0] == '(' && bracketLiteralText[^1] == ')') || (bracketLiteralText[0] == '[' && bracketLiteralText[^1] == ']'),
+                        $"Collection with opening '{bracketLiteralText[0]}' should have corresponding '{(bracketLiteralText[0] == '(' ? ')' : ']')}'");
 
-                    Debug.Assert(isParenthesesBased || isBracketBased,
-                        $"Collection with opening '{bracketLiteralText[0]}' should have corresponding '{bracketLiteralText[^1]}'");
-
-                    if(isParenthesesBased)
+                    if (bracketLiteralText[0] == '(' && bracketLiteralText[^1] == ')')
                     {
                         bracketLiteralText = string.Create(bracketLiteralText.Length, bracketLiteralText, (span, state) =>
                         {
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -97,21 +97,23 @@ namespace Microsoft.OData.UriParser
             LiteralToken literalToken = queryToken as LiteralToken;
             if (literalToken != null)
             {
-                string originalLiteralText = literalToken.OriginalText;
-
                 // Parentheses-based collections are not standard JSON but bracket-based ones are.
                 // Temporarily switch our collection to bracket-based so that the JSON reader will
                 // correctly parse the collection. Then pass the original literal text to the token.
-                string bracketLiteralText = originalLiteralText;
-                if (bracketLiteralText[0] == '(')
-                {
-                    Debug.Assert(bracketLiteralText[bracketLiteralText.Length - 1] == ')',
-                        "Collection with opening '(' should have corresponding ')'");
+                ReadOnlySpan<char> bracketLiteralText = literalToken.OriginalText.AsSpan();
 
-                    StringBuilder replacedText = new StringBuilder(bracketLiteralText);
-                    replacedText[0] = '[';
-                    replacedText[replacedText.Length - 1] = ']';
-                    bracketLiteralText = replacedText.ToString();
+                if (bracketLiteralText[0] == '(' || bracketLiteralText[0] == '[')
+                {
+                    Debug.Assert((bracketLiteralText[0] == '(' && bracketLiteralText[^1] == ')') || (bracketLiteralText[0] == '[' && bracketLiteralText[^1] == ']'),
+                        $"Collection with opening '{bracketLiteralText[0]}' should have corresponding '{bracketLiteralText[^1]}'");
+
+                    if(bracketLiteralText[0] == '(' && bracketLiteralText[^1] == ')')
+                    {
+                        Span<char> replacedText = new Span<char>(bracketLiteralText.ToArray());
+                        replacedText[0] = '[';
+                        replacedText[replacedText.Length - 1] = ']';
+                        bracketLiteralText = replacedText;
+                    }
 
                     Debug.Assert(expectedType.IsCollection());
                     string expectedTypeFullName = expectedType.Definition.AsElementType().FullTypeName();
@@ -122,7 +124,7 @@ namespace Microsoft.OData.UriParser
                         // and also, per ABNF, a single quote within a string literal is "encoded" as two consecutive single quotes in either
                         // literal or percent - encoded representation.
                         // Sample: ['a''bc','''def','xyz'''] ==> ["a'bc","'def","xyz'"], which is legitimate Json format.
-                        bracketLiteralText = NormalizeStringCollectionItems(bracketLiteralText);
+                        bracketLiteralText = NormalizeStringCollectionItems(bracketLiteralText.ToString());
                     }
                     else if (expectedTypeFullName.Equals("Edm.Guid", StringComparison.Ordinal))
                     {
@@ -130,7 +132,7 @@ namespace Microsoft.OData.UriParser
                         // with the Json reader used for deserialization.
                         // Sample: [D01663CF-EB21-4A0E-88E0-361C10ACE7FD, 492CF54A-84C9-490C-A7A4-B5010FAD8104]
                         //    ==>  ['D01663CF-EB21-4A0E-88E0-361C10ACE7FD', '492CF54A-84C9-490C-A7A4-B5010FAD8104']
-                        bracketLiteralText = NormalizeGuidCollectionItems(bracketLiteralText);
+                        bracketLiteralText = NormalizeGuidCollectionItems(bracketLiteralText.ToString());
                     }
                     else if (expectedTypeFullName.Equals("Edm.DateTimeOffset", StringComparison.Ordinal) ||
                              expectedTypeFullName.Equals("Edm.Date", StringComparison.Ordinal) ||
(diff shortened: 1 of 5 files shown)


xuzhg (approved):

gathogojr:
@WanjohiSammy 
_This change ensures the square brackets are handle the same way as parentheses when constructing CollectionNode in GetCollectionOperandFromToken method:_
- Is use of square brackets in the same manner as parentheses supported by the OData standard?
- Can you confirm that we support non-empty square brackets? For example, `$filter=Name in ['Sue','Joe']`. Maybe even add a test

WanjohiSammy:
@gathogojr 
> * Is use of square brackets in the same manner as parentheses supported by the OData standard?
> * Can you confirm that we support non-empty square brackets? For example, `$filter=Name in ['Sue','Joe']`. Maybe even add a test

I have updated the PR description to reflect correctly that this was failing because the `empty string` in `square brackets` was not handle the same as `empty string` in `parenthesis`

gathogojr:
@WanjohiSammy Can you confirm what I asked here https://github.com/OData/odata.net/pull/3122#issuecomment-2485561072

gathogojr:
@WanjohiSammy I have confirmed that we support brackets, i.e., `Books?$filter=Title in ('Harry Porter')` and `Books?$filter=Title in ['Harry Porter']` are both handled fine by the library

WanjohiSammy:
@gathogojr
> Can you confirm what I asked here [#3122 (comment)](https://github.com/OData/odata.net/pull/3122#issuecomment-2485561072)

I thought I confirmed here https://github.com/OData/odata.net/pull/3122#issuecomment-2485847253
The issue solved by this PR is when there is an empty string in square brackets. 

-We do support square brackets but in case of empty string in these brackets an exception is thrown. 
-Since there is no exception with Empty string in parenthesis, this change ensures empty string in square brackets are treated the same as empty string in parenthesis.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -122,15 +124,15 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
...
+                        bracketLiteralText = NormalizeStringCollectionItems(bracketLiteralText.ToString());
                     }
                     else if (expectedTypeFullName.Equals("Edm.Guid", StringComparison.Ordinal))
                     {
                         // For collection of Guids, need to convert the Guid literals to single-quoted form, so that it is compatible
                         // with the Json reader used for deserialization.
                         // Sample: [D01663CF-EB21-4A0E-88E0-361C10ACE7FD, 492CF54A-84C9-490C-A7A4-B5010FAD8104]
                         //    ==>  ['D01663CF-EB21-4A0E-88E0-361C10ACE7FD', '492CF54A-84C9-490C-A7A4-B5010FAD8104']
-                        bracketLiteralText = NormalizeGuidCollectionItems(bracketLiteralText);
+                        bracketLiteralText = NormalizeGuidCollectionItems(bracketLiteralText.ToString());
```
habbes, line 135: Similar to: https://github.com/OData/odata.net/pull/3122/files#r1850669734

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -97,21 +97,23 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
...
+                    Debug.Assert((bracketLiteralText[0] == '(' && bracketLiteralText[^1] == ')') || (bracketLiteralText[0] == '[' && bracketLiteralText[^1] == ']'),
+                        $"Collection with opening '{bracketLiteralText[0]}' should have corresponding '{bracketLiteralText[^1]}'");
 
-                    StringBuilder replacedText = new StringBuilder(bracketLiteralText);
-                    replacedText[0] = '[';
-                    replacedText[replacedText.Length - 1] = ']';
-                    bracketLiteralText = replacedText.ToString();
+                    if(bracketLiteralText[0] == '(' && bracketLiteralText[^1] == ')')
+                    {
+                        Span<char> replacedText = new Span<char>(bracketLiteralText.ToArray());
```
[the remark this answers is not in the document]
habbes: Note that the `string.Create` method could be used to create string from a delegate without having to create a temporary buffer.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -122,15 +124,15 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
                         // and also, per ABNF, a single quote within a string literal is "encoded" as two consecutive single quotes in either
                         // literal or percent - encoded representation.
                         // Sample: ['a''bc','''def','xyz'''] ==> ["a'bc","'def","xyz'"], which is legitimate Json format.
-                        bracketLiteralText = NormalizeStringCollectionItems(bracketLiteralText);
+                        bracketLiteralText = NormalizeStringCollectionItems(bracketLiteralText.ToString());
```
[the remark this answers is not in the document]
WanjohiSammy: Yes, this can be refactored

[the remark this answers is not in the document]
WanjohiSammy: Used `string.Create` instead

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
WanjohiSammy: Used `string.Create` instead

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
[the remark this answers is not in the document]
WanjohiSammy: Used `string.Create` instead

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -97,21 +97,28 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
...
+                    bool isBracketBased = bracketLiteralText[0] == '[' && bracketLiteralText[^1] == ']';
+
+                    Debug.Assert(isParenthesesBased || isBracketBased,
+                        $"Collection with opening '{bracketLiteralText[0]}' should have corresponding '{bracketLiteralText[^1]}'");
 
-                    StringBuilder replacedText = new StringBuilder(bracketLiteralText);
-                    replacedText[0] = '[';
-                    replacedText[replacedText.Length - 1] = ']';
-                    bracketLiteralText = replacedText.ToString();
+                    if(isParenthesesBased)
```
gathogojr, line 113: Can be simplified... No need for the variables, we only need to verify that the last character matches the first character:
```csharp
Debug.Assert(bracketLiteralText[0] == 'bracketLiteralText[^1],
    $"Collection with opening '{bracketLiteralText[0]}' should have corresponding '{(bracketLiteralText[0] == '(' ? ')' : ']')}'");

if (bracketLiteralText[0] == '(')
// ...
```

Note that with your current code, if `bracketLiteralText` had the value `(foobar]`, the logic would result into a debug assertion with the following message:
```
Collection with opening '(' should have corresponding ']'
```
Likewise, if `bracketLiteralText` had the value `(foobar!`, the logic would result into a debug assertion with the following message:
```
Collection with opening '(' should have corresponding '!'
```

WanjohiSammy: Thanks @gathogojr for pointing this out. 
Resolved.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -97,21 +97,25 @@ private CollectionNode GetCollectionOperandFromToken(QueryToken queryToken, IEdm
...
                 // correctly parse the collection. Then pass the original literal text to the token.
-                string bracketLiteralText = originalLiteralText;
-                if (bracketLiteralText[0] == '(')
+                string bracketLiteralText = literalToken.OriginalText;
+
+                if (bracketLiteralText[0] == '(' || bracketLiteralText[0] == '[')
                 {
-                    Debug.Assert(bracketLiteralText[bracketLiteralText.Length - 1] == ')',
-                        "Collection with opening '(' should have corresponding ')'");
+                    Debug.Assert((bracketLiteralText[0] == '(' && bracketLiteralText[^1] == ')') || (bracketLiteralText[0] == '[' && bracketLiteralText[^1] == ']'),
```
gathogojr, line 107: Can we get here in reality? Or is this designed just to assert our expectations during development? If we can get here in reality, we should throw an exception instead

WanjohiSammy: We can't get here with `unbalanced brackets` in reality. The unbalanced brackets will be captured in the `ExpressionLexer` and throw an expression. The method that does that is [AdvanceThroughBalancedExpression](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs#L1209)

```cs
private void AdvanceThroughBalancedExpression(char startingCharacter, char endingCharacter)
{
    int currentBracketDepth = 1;

    while (currentBracketDepth > 0)
    {
        ...
        // the rest of code
        if (this.ch == null)
        {
            throw new ODataException(ODataErrorStrings.ExpressionLexer_UnbalancedBracketExpression); <=Exception thrown here
        }

        this.NextChar();
    }
}
```

The method severally:
- [AdvanceThroughBalancedExpression('(', ')')](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs#L433)
- [AdvanceThroughBalancedExpression('(', ')')](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs#L548)
- [AdvanceThroughBalancedExpression('{', '}')](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs#L647)
- [AdvanceThroughBalancedExpression('[', ']')](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/ExpressionLexer.cs#L652)

gathogojr (approved):

PR #3125, opened 2024-11-18
Fixes #3120: Throws exception if non defined enum member used in `in` operator

### LINKED ISSUE #3120: Enum value not parsed for `in` query operator
| Request URL | Forwarded URL |
|--------|--------|
| `/identityProtection/RiskyUsers?$filter=riskLevel eq 'low'`  | `identityProtection/riskyUsers?$filter=riskLevel+eq+Microsoft.IdentityProtectionServices.riskLevel'low'` |
| `/identityProtection/RiskyUsers?$filter=riskLevel in ('low')` | `/identityProtection/riskyUsers?$filter=riskLevel+in+('low')` |

ODL doesn't parse the value as Enum for the `in` case.

xuzhg:
… 

### Issues

*This pull request fixes #3120.*

### Description

* Should throw exception if the non-defined enum member used in 'in' operator.
* Make sure it's the same behavior comparing to the binary operator.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -27,12 +27,19 @@ namespace Microsoft.OData.UriParser
         private readonly Func<QueryToken, QueryNode> bindMethod;
 
         /// <summary>
+        /// Resolver for parsing
+        /// </summary>
+        private readonly ODataUriResolver resolver;
+
+        /// <summary>
         /// Constructs a InBinder with the given method to be used binding the parent token if needed.
         /// </summary>
         /// <param name="bindMethod">Method to use for binding the parent token, if needed.</param>
-        internal InBinder(Func<QueryToken, QueryNode> bindMethod)
+        /// <param name="resolver">Resolver for parsing.</param>
+        internal InBinder(Func<QueryToken, QueryNode> bindMethod, ODataUriResolver resolver)
         {
             this.bindMethod = bindMethod;
+            this.resolver = resolver;
         }
 
         /// <summary>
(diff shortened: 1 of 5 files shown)


src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
```diff
@@ -64,7 +64,7 @@ internal static SingleValueNode ConvertToTypeIfNeeded(SingleValueNode source, IE
                 {
                     string memberName = constantNode.Value.ToString();
                     IEdmEnumType enumType = targetTypeReference.Definition as IEdmEnumType;
-                    if(enumType.ContainsMember(memberName, StringComparison.Ordinal))
+                    if (enumType.ContainsMember(memberName, StringComparison.Ordinal))
```
syprieur, line 67: Should that honor the resolver's preference for case-sensitivity?

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -65,6 +72,8 @@ internal QueryNode BindInOperator(InToken inToken, BindingState state)
                 left = MetadataBindingUtils.ConvertToTypeIfNeeded(left, right.ItemType);
             }
 
+            MetadataBindingUtils.VerifyCollectionNode(right, this.resolver.EnableCaseInsensitive);
```
syprieur, line 75: I'm not familiar with this codebase and it does not use nullable context (`#nullable enable`) so I can't say: could `resolver` ever be `null` here?

syprieur (approved):

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
xuzhg: From other similar codes, it seems it should not be null exception the customer explicitly sets as null. .

src/Microsoft.OData.Core/UriParser/Binders/MetadataBindingUtils.cs
xuzhg: This is not related to this PR. I'd prefer to leave it as it now and fix it if we do meet the problem.

Issue #3123, opened 2024-11-16
Breaking change in 8.2.0


syprieur:
The 8.2.0 OData upgrade contained a breaking change via [Fix/1744 isof unquoted type params issue](https://github.com/OData/odata.net/pull/3108/files#diff-4353a0066027cf1ccd9e21c7ddfda66c1b86615190edba076336ccc1e28cdd88)

A side-effect of this PR seems to be new case-sensitivity.
Example: `?$filter=cast(stringProperty, edm.int64) lt 1`

This used to work in 8.1.0, but now throws an `ODataException` in 8.2.0:
_"Cast or IsOf Function must have a type in its argument"_

Note that with the proper case, the filter still works in 8.2.0:
`?$filter=cast(stringProperty, Edm.Int64) lt 1`

WanjohiSammy:
@syprieur Thanks for pointing this out. I have tested the scenario with versions `8.1.0` and `8.0.0`, but I am still getting the same error. Could you please confirm if it used to work previously as we continue to investigate?

You can find a repo below to repro the issue with different ODL versions. The ODL version in this repo is 8.0.0, but you can update it to **8.1.0** or even **8.2.0**:

https://github.com/WanjohiSammy/ODLTests.git

syprieur:
With your test setup, simply using an `ODataUriParser` with `UnqualifiedODataUriResolver` instead of an `ODataQueryOptionParser` repros with `string filterQuery = "cast(ID, edm.int64) lt 1";`

``` cs
            var parser = new ODataUriParser(HardCodedTestModel.TestModel, new Uri($"/People?$filter={filterQuery}", UriKind.Relative))
            {
                Resolver = new UnqualifiedODataUriResolver()
                {
                    EnableCaseInsensitive = true,
                },
                UrlKeyDelimiter = ODataUrlKeyDelimiter.Slash,
            };
            parser.Settings.MaximumExpansionDepth = 2;
            parser.ParsePath();
            var filter = parser.ParseFilter();
```

Results matrix:

| filterQuery/OData Version | 8.1.0 | 8.2.0 |
| - | - | - |
| "cast(ID, edm.int64) lt 1" | Ok | Throws |
| "cast(ID, edm.Int64) lt 1" | Ok | Ok |

WanjohiSammy:
Thanks @syprieur for adding further repro steps.

WanjohiSammy:
Resolved. 

https://github.com/OData/odata.net/releases/tag/8.2.2

PR #3108, opened 2024-11-07
Fix/1744 isof unquoted type params issue

WanjohiSammy:
### Issues

*This pull request fixes [#1744](https://github.com/OData/odata.net/issues/1744).*

### Description

According to [OData V4 spec](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_isof) `isof` and `cast` function type parameter can be used without enclosed in quotes. 

The isof function has the following signatures
```
Edm.Boolean isof(type)
Edm.Boolean isof(expression,type)
```

This update supports unquoted type parameters in `isof` and `cast` functions through the following changes:
- Add support to handle `SingleResourceCastNode` in `ValidateIsOfOrCast` method since the unquoted type parameter is bind as **SingleResourceCastNode**
- Update the FunctionCallParser to handle `isof` and `cast` function calls with two parameters, where the second parameter is an unquoted type parameter. For two parameters, the second unquoted type parameter's next token should point to the first parameter. Previously, this was only handled for quoted type parameters. The change has now been made to include unquoted type parameters as well.
  For example,
  - Cast(Location, NS.HomeAddress) -> NS.HomeAddress will be a DottedIdentifierToken, and its NextToken should point to the Location parameter token.
  - Isof(Location, NS.HomeAddress) -> This also applies to the `isof()` function.

With this change, the following queries with unquoted type param are now supported:
```cs
cast(NS.Employee) // cast Person to Employee
cast(Location, NS.WorkAddress) // cast Location property to NS.WorkAddress
isof(NS.Employee) // cast Person to Employee
isof(Location, NS.WorkAddress) // cast Location property to NS.WorkAddress
```

### Checklist (Uncheck if it is not completed)

- [x] *Test cases added*
- [x] *Build and test with one-click build and test script passed*

### Additional work necessary

*If documentation update is needed, please add "Docs Needed" label to the issue and provide details about the required document change in the issue.*

PR #3126, opened 2024-11-19
Revert "Fix/1744 isof unquoted type params issue (#3108)"

WanjohiSammy:
This reverts commit 1aed366fe1df4d3ae2764bc20cfd8c50fa7dcb04.

### Issues

### Description

This PR removes the feature to `support unquoted type parameters` in `cast` and `isof` function calls. The feature caused a breaking change https://github.com/OData/odata.net/issues/3123. We are taking this step back to properly prepare a fix that reintroduces the feature without causing any breaking changes. It's important to note that this feature was not released to AspNetCore.OData, and we do not anticipate significant usage by existing ODL customers.

- [ ] *Test cases added*

### Additional work necessary


habbes:
This PR reverts: https://github.com/OData/odata.net/pull/3108/

habbes (approved):

PR #3127, opened 2024-11-19
architecture doc

corranrogue9:


docs/architecture.md
```diff
@@ -0,0 +1,545 @@
...
+    participant odata request AST
+
+    user defined types->>odata request AST: serialized to
+```
+
+#### odata request to http request
+
+```mermaid
+sequenceDiagram
+    participant odata request AST
```
corranrogue9, line 296: Did I get these backwards? Which comes first, the AST, or the CST, when the input is the http request?

corranrogue9: Currently, we are calling the CST "syntactic ast" and the AST "semantic ast"

docs/architecture.md
```diff
@@ -0,0 +1,545 @@
...
+    participant odata request CST
```
corranrogue9, line 297: From @habbes : should we remove the CST and treat it as a "detail" or parsing the payloads?

corranrogue9: My gut reaction here is: it's almost "hard" to write the code **wrong** if there's both an AST and a CST. Because of this, I think it's really useful to have both concepts defined and implemented; it would be really useful to have those implementations to use as comparisons in test cases and such to make sure our transcribers, readers, and writers are implemented correctly. Those tests would be "complete" because they would be comparing the entire payload instead of just whatever assertions the test writer thought were meaningful at the time. This also could give us significantly better coverage on breaking changes. 

Let me think about this more.

docs/architecture.md
corranrogue9: @mikepizzo I am not finding any sources that refer to these concepts by the names "syntactic AST" and "semantic AST", but I am finding lots of literature that uses "AST" and either "CST" or "parse tree". I'm happy to rename if we really need to, but I'd like to normalize on established names.

corranrogue9: After some more discussion, my previous comments were incorrect. What we have today is a "syntactic AST" that follows a different grammar than the OData ABNF. This grammar is context-free and allows preliminary parsing to happen without the need of the EDM model. The "semantic AST" follows the OData ABNF with the context of the EDM model. 

I will include a note in the follow-ups that there **must** be a concept of this "syntactic AST" to maintain parity.

docs/architecture.md
corranrogue9: I'm going to keep the CST references in the doc because I think they will clarify later design decisions, but I've added a note that the CSTs won't necessarily be exposed to customers.

PR #3129, opened 2024-11-22
Update the package version to fix the vulnerable issue

xuzhg:
Update the package version to fix the vulnerable issue

Part of #https://github.com/OData/odata.net/issues/3132


xuzhg:
It's straightforward. To save the review time, I merged it and let me know any concerns.

PR #3137, opened 2024-11-29
Don't require entityset when writing responses

mikepizzo:
### Issues

Enables writing responses without specifying an entityset.

### Description

UnknownEntitySet is treated as containment -- path becomes path through parent
Id of entities in unknown entity set is path if known, otherwise null
ContextUrl is path, if known, otherwise type

- [x ] *Test cases added*

### Additional Work
-Might be able to get rid of ThrowIfMissingTypeInfo (or validate that it relates only to type info, not entity set info)

### DIFF
--- a/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
+++ b/src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
@@ -393,6 +393,12 @@ namespace Microsoft.OData.Evaluation
         /// </returns>
         internal override bool TryGetIdForSerialization(out Uri id)
         {
+            if(this.ResourceMetadataContext.TypeContext.NavigationSourceName == null)
+            {
+                id = null;
+                return false;
+            }
+
             id = this.ResourceMetadataContext.Resource.IsTransient ? null : this.GetId();
             return true;
         }
@@ -405,10 +411,13 @@ namespace Microsoft.OData.Evaluation
         {
             Uri uri = this.ResourceMetadataContext.Resource.HasNonComputedId ? this.ResourceMetadataContext.Resource.NonComputedId : this.ComputedId;
 
-            Debug.Assert(this.ResourceMetadataContext != null && this.ResourceMetadataContext.TypeContext != null, "this.resourceMetadataContext != null && this.resourceMetadataContext.TypeContext != null");
-            if (this.ResourceMetadataContext.ActualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName)
+            if (uri != null)
             {
-                uri = this.UriBuilder.AppendTypeSegment(uri, this.ResourceMetadataContext.ActualResourceTypeName);
+                Debug.Assert(this.ResourceMetadataContext != null && this.ResourceMetadataContext.TypeContext != null, "this.resourceMetadataContext != null && this.resourceMetadataContext.TypeContext != null");
+                if (this.ResourceMetadataContext.ActualResourceTypeName != this.ResourceMetadataContext.TypeContext.NavigationSourceEntityTypeName)
+                {
+                    uri = this.UriBuilder.AppendTypeSegment(uri, this.ResourceMetadataContext.ActualResourceTypeName);
+                }
             }
 
             return uri;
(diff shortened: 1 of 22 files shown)


src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
```diff
@@ -393,6 +393,12 @@ internal override string GetOperationTitle(string operationName)
         /// </returns>
         internal override bool TryGetIdForSerialization(out Uri id)
         {
+            if(this.ResourceMetadataContext.TypeContext.NavigationSourceName == null)
+            {
+                id = null;
+                return false;
+            }
+
             id = this.ResourceMetadataContext.Resource.IsTransient ? null : this.GetId();
```
xuzhg, line 402: Id could be 'null' at Line 402, but it always returns 'true'? Now, the new return could be 'false'? what's the difference?

src/Microsoft.OData.Core/Evaluation/ODataConventionalIdMetadataBuilder.cs
```diff
@@ -131,11 +131,11 @@ private void ComputeAndCacheId()
                 case EdmNavigationSourceKind.Singleton:
                     uri = this.ComputeIdForSingleton();
                     break;
+                // Treat UnknownEntitySet as a containment
+                case EdmNavigationSourceKind.UnknownEntitySet:
                 case EdmNavigationSourceKind.ContainedEntitySet:
                     uri = this.ComputeIdForContainment();
                     break;
-                case EdmNavigationSourceKind.UnknownEntitySet:
```
xuzhg, line 137: Is it a behavior breaking change?

src/Microsoft.OData.Core/Evaluation/ODataConventionalIdMetadataBuilder.cs
```diff
@@ -152,7 +152,7 @@ private void ComputeAndCacheId()
         /// </returns>
         private Uri ComputeId()
         {
-            if (this.ResourceMetadataContext.KeyProperties.Any())
+            if (this.ResourceMetadataContext.KeyProperties.Any() && this.ResourceMetadataContext.TypeContext.NavigationSourceName != null)
```
xuzhg, line 155: move the  'null' testing ahead of 'Any()' could be better perf?

src/Microsoft.OData.Core/ODataResourceTypeContext.cs
```diff
@@ -55,23 +55,23 @@ private ODataResourceTypeContext(IEdmStructuredType expectedResourceType, bool t
         /// </summary>
         public virtual string NavigationSourceName
         {
-            get { return this.ValidateAndReturn(default(string)); }
```
xuzhg, line 58: can we just use default property get?

public virtual string NavigationSourceName { get; }

xuzhg (commented):
I left some comments and looking forward the PR

src/Microsoft.OData.Core/Evaluation/ODataConventionalEntityMetadataBuilder.cs
mikepizzo: For a transient resource (like in aggregation) null *is* the right value for the id. In this case, the id isn't null, it just can't be computed. 

We have existing code that checks the boolean result of the TryGetIdForSerialization and handles it differently than true with a null value.

src/Microsoft.OData.Core/Evaluation/ODataConventionalIdMetadataBuilder.cs
mikepizzo: It is turning an error case into a success case.

src/Microsoft.OData.Core/Evaluation/ODataConventionalIdMetadataBuilder.cs
mikepizzo: :thumbsup: 

I like it!

src/Microsoft.OData.Core/ODataResourceTypeContext.cs
mikepizzo: :thumbsup: 

yes; that would be simpler. good catch.

xuzhg (approved):

PR #3145, opened 2024-12-06
Fixes #1817: support key alias

### LINKED ISSUE #1817: Support OData key alias

xuzhg:
Fixes #1817: support key alias
Introduce the IEdmPropertyRef interface
Use default interface method for IEdmEntityType avoid breaking changes
This change focus on Edm lib, if it's ready and will move to OData.Core.


xuzhg:
I'd like to move the key alias feature forward. 
It's not a duplicated of PR: https://github.com/OData/odata.net/pull/2411.
It's a different implementation.

src/Microsoft.OData.Edm/Csdl/Semantics/BadElements/UnresolvedPropertyRef.cs
```diff
@@ -0,0 +1,30 @@
+﻿//---------------------------------------------------------------------
+// <copyright file="UnresolvedProperty.cs" company="Microsoft">
```
WanjohiSammy, line 2: This should be `UnresolvedPropertyRef.cs` instead of `UnresolvedProperty.cs`
```suggestion
// <copyright file="UnresolvedPropertyRef.cs" company="Microsoft">

```

src/Microsoft.OData.Edm/Csdl/Parsing/SchemaJsonParser.cs
```diff
@@ -650,21 +650,37 @@ internal static CsdlKey ParseCsdlKey(string name, JsonElement keyArray, JsonPars
             Debug.Assert(keyArray.ValueKind == JsonValueKind.Array);
             Debug.Assert(name == "$Key", "The name should be $Key.");
```
WanjohiSammy, line 651: What if the `$Key` is in lowercase: `$key`

src/Microsoft.OData.Edm/Csdl/Semantics/CsdlSemanticsEntityTypeDefinition.cs
```diff
@@ -105,17 +114,24 @@ private IEdmEntityType ComputeBaseType()
...
+        private IEnumerable<IEdmPropertyRef> ComputeDeclaredKey()
         {
             if (this.entity.Key != null)
             {
-                List<IEdmStructuralProperty> key = new List<IEdmStructuralProperty>();
+                List<IEdmPropertyRef> key = new List<IEdmPropertyRef>();
                 foreach (CsdlPropertyReference keyProperty in this.entity.Key.Properties)
                 {
-                    IEdmStructuralProperty structuralProperty = this.FindProperty(keyProperty.PropertyName) as IEdmStructuralProperty;
+                    IEdmStructuralProperty structuralProperty = FindKeyProperty(keyProperty.PropertyName);
```
WanjohiSammy, line 124:
```suggestion
                    IEdmStructuralProperty structuralProperty = this.FindKeyProperty(keyProperty.PropertyName);

```

Or define `FindKeyProperty` as `static` method

test/FunctionalTests/Microsoft.OData.Edm.Tests/Csdl/CsdlWriterTests.cs
```diff
@@ -2155,6 +2157,200 @@ public void CanWriteUrlEscapeFunction()
 }");
         }
 
+        [Fact]
+        public void CanWriteEntityType_WithSimpleKey_WithOrWithoutAlias()
```
WanjohiSammy, line 2161: You can add equivalent `Async` tests for these `Sync` tests

PR #3160, opened 2024-12-30
Port Primitive Types Tests

WanjohiSammy:
### Issues

### Description

Port QueryOption Tests from [`WCF`](https://learn.microsoft.com/en-us/dotnet/framework/wcf/whats-wcf) to use `AspNetCoreOData`

- [test/EndToEndTests/Tests/Client/Build.Desktop/PrimitiveTypesTests](https://github.com/OData/odata.net/tree/main/test/EndToEndTests/Tests/Client/Build.Desktop/PrimitiveTypesTests)
- [test/EndToEndTests/Tests/Client/Build.Desktop/KeyAsSegmentTests/PrimitiveKeyValuesTests](https://github.com/OData/odata.net/blob/main/test/EndToEndTests/Tests/Client/Build.Desktop/KeyAsSegmentTests/PrimitiveKeyValuesTests.cs)

- [x] *Test cases added*

### Additional work necessary


test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/PrimitiveKeysValuesTests.cs
```diff
@@ -0,0 +1,246 @@
...
+        {
+            return float.IsNaN(key);
+        }
+
+        private static bool IsNotSupportedKey(double key)
+        {
+            return double.IsNaN(key);
+        }
+
+        private void ResetDefaultDataSource()
```
xuzhg, line 238: any HttpRequest test to call this reset method?

xuzhg (approved):

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/PrimitiveKeysValuesTests.cs
WanjohiSammy: This method is called from the constructor of `PrimitiveKeysValuesTests`:

```cs
 public PrimitiveKeysValuesTests(TestWebApplicationFactory<PrimitiveKeysValuesTests.TestsStartup> fixture)
     : base(fixture)
 {
     // The rest of code
     ResetDefaultDataSource(); // here
 }
```

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/PrimitiveKeysValuesTests.cs
```diff
@@ -0,0 +1,270 @@
...
+            };
+
+            _model = PrimitiveKeyValuesEdmModel.GetEdmModel();
+            ResetDefaultDataSource();
+        }
+
+        [Fact]
+        public void BinaryTest()
+        {
+            foreach (var entry in _context.EdmBinarySet)
```
habbes, line 69: Instead of having a loop inside the test, could make this a `[Theory]` with `[TheoryDataSet]` instead of a `[Fact]` and use the data set? That way the test could be executed separately for each input and it would be easier to trace failing tests.

WanjohiSammy: Thanks for noting this. I have used `[Theory]` with `TheoryData`

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/Helpers/ODataMessageReaderTestsHelper.cs
```diff
@@ -0,0 +1,235 @@
...
+                }
+                Assert.Equal(ODataReaderState.Completed, reader.State);
+            }
+
+        }
+
+        return entries;
+    }
+
+    public async Task<(List<ODataResource>, List<ODataResourceSet>)> QueryResourceAndResourceSetsAsync(string queryText, string mimeType)
```
habbes, line 110: Would be good to have doc comments for this method as well. Also it would help understand this method (and other methods in this class) if we could see examples of queries or URLs it expects.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/DurationTests.cs
```diff
@@ -0,0 +1,265 @@
...
+
+        _model = DefaultEdmModel.GetEdmModel();
+        ResetDefaultDataSource();
+    }
+
+    public static IEnumerable<object[]> MimeTypesData
+    {
+        get
+        {
+            yield return new object[] { MimeTypes.ApplicationJson + MimeTypes.ODataParameterFullMetadata };
```
habbes, line 63: Are these mime types `const` strings? If they are constants, then I would suggest using `InlineData` instead of `MemberData`.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/DurationTests.cs
```diff
@@ -0,0 +1,265 @@
...
+            yield return new object[] { MimeTypes.ApplicationJson + MimeTypes.ODataParameterMinimalMetadata };
+            yield return new object[] { MimeTypes.ApplicationJson + MimeTypes.ODataParameterNoMetadata };
+        }
+    }
+
+    #region Entity Set
+
+    [Theory]
+    [MemberData(nameof(MimeTypesData))]
```
habbes, line 72: If these mime types are constants, then I'd suggest using inline data as so:

```suggestion
    [InlineData(MimeTypes.ApplicationJson + ODataParameterFullMetadata)]
    [InlineData(MimeTypes.ApplicationJson + MimeTypes.ODataParameterMinimalMetadata)]
    [InlineData(MimeTypes.ApplicationJson + MimeTypes.ODataParameterNoMetadata)]
```

Why? This makes the tests easier to read. It's easier to know what is being tested without having to search for the definition of the member data collection. We should strive to keep the tests self-contained when feasible, so that we don't have to look at other source code to understand what the test is trying to achieve.

If the fact that these mime types are constant is not important to the feature, we could even go further and write them out as string literals.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/DurationTests.cs
```diff
@@ -0,0 +1,265 @@
...
+    public async Task QueryEntitySetTest(string mimeType)
+    {
+        // Arrange & Act
+        var queryText = "Customers";
+        var entries = await this.TestsHelper.QueryResourceSetsAsync(queryText, mimeType);
+
+        // Assert
+        if (!mimeType.Contains(MimeTypes.ODataParameterNoMetadata))
```
habbes, line 80: Why do we have an if statement here? If this test does not make sense for ODataParameterNoMetadata, then we should not include it in the test. Again, using InlineData would make it easier to include only the relevant mime types.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/DurationTests.cs
```diff
@@ -0,0 +1,265 @@
...
+        {
+            foreach (var entry in entries)
+            {
+                if (entry.TypeName.EndsWith("Customer"))
+                {
+                    var timeBetweenLastTwoOrdersProperty = entry.Properties.Single(p => p.Name == "TimeBetweenLastTwoOrders") as ODataProperty;
+                    Assert.NotNull(timeBetweenLastTwoOrdersProperty);
+                    Assert.NotNull(timeBetweenLastTwoOrdersProperty.Value);
+                }
+            }
```
habbes, line 90: This test will succeed if the return collection is empty or has no entries with a type name that ends with "Customer". This means that we cannot guarantee that the relevant code path is tested when this test runs. I think we should avoid such conditional tests. We should have the confidence that if this test runs and passes, then the associated behavior is correctly implemented.  If not, how can we make this test more specific and deterministic. If we expect that query to return a non-empty array, then we should test for that. If we expect there to be customer entries, we should retrieve them and ensure that set is non empty, or has the expected number of elements. You can look for a way to refactor this such that there's guarantee that when this test runs, the behaviour is actually tested. When possible, let's avoid complex logic such conditional and loops in tests.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/DurationTests.cs
```diff
@@ -0,0 +1,265 @@
...
+        // Arrange & Act
+        var queryText = "Customers(1)";
+        var entries = await this.TestsHelper.QueryResourceEntriesAsync(queryText, mimeType);
+
+        // Assert
+        if (!mimeType.Contains(MimeTypes.ODataParameterNoMetadata))
+        {
+            foreach (var entry in entries)
+            {
+                if (entry.TypeName.EndsWith("Customer"))
```
habbes, line 111: See https://github.com/OData/odata.net/pull/3160/files#r1914339945

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/DurationTests.cs
```diff
@@ -0,0 +1,265 @@
...
+    [Theory]
+    [MemberData(nameof(MimeTypesData))]
+    public async Task QueryEntityPropertyTest(string mimeType)
+    {
+        // Arrange & Act
+        var queryText = "Customers(1)/TimeBetweenLastTwoOrders";
+        var property = await this.TestsHelper.QueryPropertyAsync(queryText, mimeType);
+
+        // Assert
+        if (!mimeType.Contains(MimeTypes.ODataParameterNoMetadata))
```
habbes, line 134: See https://github.com/OData/odata.net/pull/3160/files#r1914339945

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/DurationTests.cs
```diff
@@ -0,0 +1,265 @@
...
+        var queryText = "Orders(7)/CustomerForOrder";
+        var entries = await this.TestsHelper.QueryResourceEntriesAsync(queryText, mimeType);
+
+        // Assert
+        if (!mimeType.Contains(MimeTypes.ODataParameterNoMetadata))
+        {
+            foreach (var entry in entries)
+            {
+                if (entry.TypeName.EndsWith("Customer"))
+                {
```
habbes, line 159: See https://github.com/OData/odata.net/pull/3160/files#r1914339945

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/PrimitiveKeysValuesTests.cs
```diff
@@ -0,0 +1,270 @@
...
+        [Theory]
+        [MemberData(nameof(GetEdmDoubleSet))]
+        public void DoubleTest(EdmDouble entry)
+        {
+            // Arrange & Act
+            var query = _context.CreateQuery<EdmDouble>("EdmDoubleSet").Where(e => e.Id.Equals(entry.Id));
+            var queryResult = query.ToArray();
+
+            // Assert
+            //Expected a single result for key value {0}, entry.Id.ToString(CultureInfo.InvariantCulture)
```
habbes, line 123: Maybe this comment could be used as the custom message of the assertion?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/PrimitiveKeysValuesTests.cs
```diff
@@ -0,0 +1,270 @@
...
+            {
+                HttpClientFactory = HttpClientFactory
+            };
+
+            _model = PrimitiveKeyValuesEdmModel.GetEdmModel();
+            ResetDefaultDataSource();
+        }
+
+        [Theory]
+        [MemberData(nameof(GetEdmBinarySet))]
```
habbes, line 67: In this case, I also think there would be value in spelling out all the keys we want to test with using `InlineData`. We could pass a primtive value to the attribute, then in the test, construct the `EdmBinary` instance. I see that advantage with your approach is that the inputs here are the same keys used in the context, and so if the changing it in on place will automatically update it here. But I'm not sure if that beats the readability of clearly seeing what's being tested, which would also make it easy to spot if we're missing any edge cases. A bit data or code duplication is not too bad in tests.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/PrimitiveKeysValuesTests.cs
```diff
@@ -0,0 +1,270 @@
...
+            var queryResult = query.ToArray();
+
+            // Assert
+            //Expected a single result for key value {0}, entry.Id.ToString()
+            Assert.Single(queryResult);
+        }
+
+        #region TheoryData
+
+        public static TheoryData<T> GetTheoryData<T>(Func<Container, IEnumerable<T>> dataSelector)
```
habbes, line 207:
```suggestion
        private static TheoryData<T> GetTheoryData<T>(Func<Container, IEnumerable<T>> dataSelector)

```

habbes: But do we really need this method. The logic seems complex. Why do we have `IsNotSupportedKey` getting called here? I think we should write the literal keys in `InlineData` to avoid such "magic" in tests.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/PrimitiveValueFormatTest.cs
```diff
@@ -0,0 +1,476 @@
...
+        {
+            yield return new object[] { MimeTypes.ApplicationJson + MimeTypes.ODataParameterFullMetadata };
+            yield return new object[] { MimeTypes.ApplicationJson + MimeTypes.ODataParameterFullMetadata + MimeTypes.ODataParameterIEEE754Compatible };
+        }
+    }
+
+    public string DataPattern = "(" + @"[\+-]?\d\.?\d*E?[\+-]?\d*" + "|" + @"INF|-INF|NaN" + ")";
+
+    [Theory]
+    [MemberData(nameof(MimeTypesData))]
```
habbes, line 74: See https://github.com/OData/odata.net/pull/3160/files#r1914339945

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/PrimitiveValueFormatTest.cs
```diff
@@ -0,0 +1,476 @@
...
+        var (resourceEntries, resourceSetEntries) = await TestsHelper.QueryResourceAndResourceSetsAsync(skipTokenQuery, mimeType);
+
+        var pattern = string.Concat(skipTokenQuery.AsSpan(0, skipTokenQuery.IndexOf('?')), @"\(", DataPattern, @"\)$");
+        var rgx = new Regex(pattern, RegexOptions.IgnoreCase);
+
+        foreach (var entry in resourceEntries)
+        {
+            Assert.True(rgx.Match(entry.Id.ToString()).Success);
+
+            if (mimeType.Equals(MimeTypes.ApplicationJson + MimeTypes.ODataParameterFullMetadata))
```
habbes, line 445: See https://github.com/OData/odata.net/pull/3160/files#r1914339945

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/PrimitiveValueFormatTest.cs
```diff
@@ -0,0 +1,476 @@
...
+
+    [Theory]
+    [MemberData(nameof(MimeTypesData))]
+    public async Task FloatWithoutSuffixAsKeyInURLTest(string mimeType)
+    {
+        // Arrange
+        var expectedKeySegment = "EdmSingleSet(1)";
+
+        // Act & Assert
+        await PrimitiveValueAsKeyInURLAsync("EdmSingleSet(1.0)", mimeType, expectedKeySegment);
```
habbes, line 104: It's not clear what this method is doing. Is it possible to write the relevant "act" and "assert" portions directly?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/DurationTests.cs
WanjohiSammy: I have created them as constants and passed them using InlineData

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/DurationTests.cs
WanjohiSammy: Resolved this

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/Helpers/ODataMessageReaderTestsHelper.cs
WanjohiSammy: Added the doc string

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/DurationTests.cs
WanjohiSammy: I have created them as constants and passed them using InlineData as recommended

test/EndToEndTests/Common/Microsoft.OData.Client.E2E.TestCommon/Helpers/ODataMessageReaderTestsHelper.cs
WanjohiSammy: I have added the doc string

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/DurationTests.cs
WanjohiSammy: This was a list of items, that's why I had to loop and ensure the value is not null. I have rectified this to below:
```cs
// Arrange & Act
var queryText = "Customers";
var entries = await this.TestsHelper.QueryResourceSetsAsync(queryText, mimeType);

entries = entries.Where(e => e.TypeName.EndsWith("Customer")).ToList();

// Assert
var timeBetweenLastTwoOrdersProperty = entries[0].Properties.Single(p => p.Name == "TimeBetweenLastTwoOrders") as ODataProperty;
Assert.NotNull(timeBetweenLastTwoOrdersProperty);
Assert.Equal(new TimeSpan(1), timeBetweenLastTwoOrdersProperty.Value);

timeBetweenLastTwoOrdersProperty = entries[1].Properties.Single(p => p.Name == "TimeBetweenLastTwoOrders") as ODataProperty;
Assert.NotNull(timeBetweenLastTwoOrdersProperty);
Assert.Equal(new TimeSpan(2), timeBetweenLastTwoOrdersProperty.Value);
```

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/DurationTests.cs
WanjohiSammy: Resolved this.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PrimitiveTypesTests/Tests/PrimitiveKeysValuesTests.cs
WanjohiSammy: I have explicitly passed the IDs using `InlineData` instead of using `MemberData` like below:

```cs
[Theory]
[InlineData(new Byte[] { }, "binary''")]
[InlineData(new Byte[] { 1 }, "binary'AQ%3D%3D'")]
[InlineData(new Byte[] { 2, 3, 4 }, "binary'AgME'")]
public void BinaryTest(byte[] entryId, string expectedIdInUrl)
...
```

```cs
[Theory]
[InlineData(true, "true")]
[InlineData(false, "false")]
public void BooleanTest(bool entryId, string expectedIdInUrl)
...
```

```cs
[Theory]
[InlineData(0, "0.0")]
[InlineData(-1, "-1")]
[InlineData(1, "1.0")]
[InlineData(Double.MaxValue, "1.7976931348623157E%2B308")]
[InlineData(Double.MinValue, "-1.7976931348623157E%2B308")]
[InlineData(Double.Epsilon, "5E-324")]
[InlineData(Double.NegativeInfinity, "-INF")]
[InlineData(Double.PositiveInfinity, "INF")]
public void DoubleTest(double entryId, string expectedIdInUrl)
...
```

habbes (approved):

PR #3164, opened 2025-01-13
Port Batch Request Tests

WanjohiSammy:
### Issues

### Description

Port `Batch Request Tests` from [`WCF`](https://learn.microsoft.com/en-us/dotnet/framework/wcf/whats-wcf) to use `AspNetCoreOData`

- [test/EndToEndTests/Tests/Client/Build.Desktop/BatchRequestTests](https://github.com/OData/odata.net/tree/main/test/EndToEndTests/Tests/Client/Build.Desktop/BatchRequestTests)

- [x] *Test cases added*

### Additional work necessary


test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/BatchRequestTests/Tests/BatchRequestClientTests.cs
```diff
@@ -0,0 +1,203 @@
...
+        var response = await _context.SaveChangesAsync(SaveChangesOptions.BatchWithSingleChangeset | SaveChangesOptions.UseRelativeUri);
+
+        // Assert
+        Assert.Equal(4, response.Count());
+
+        var changeResponses = response.Cast<ChangeOperationResponse>();
+
+        Assert.All(changeResponses, changeResponse => Assert.True(changeResponse.StatusCode == 200 || changeResponse.StatusCode == 201));
+
+        var accountCreatedResponse = changeResponses.FirstOrDefault(r => r.StatusCode == 201 && r.Descriptor is EntityDescriptor descriptor && descriptor.Entity is Account);
```
habbes, line 110: Retrieve the request by status code seems brittle and counter-intuitive. Does the `SaveChangesAsync` method make any guarantees in terms of the order of responses? If so we can use that and also implicitly test that behaviour. If not, can we assert based on the type of request?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/BatchRequestTests/Tests/BatchRequestClientTests.cs
```diff
@@ -0,0 +1,203 @@
...
+            new DataServiceRequest<Account>(new Uri(_baseUri + "Accounts(102)")),
+            new DataServiceRequest<PaymentInstrument>(new Uri(_baseUri + "Accounts(101)/MyPaymentInstruments")),
+            new DataServiceRequest<Statement>(new Uri(_baseUri + "Accounts(103)/MyPaymentInstruments(103901)/BillingStatements(103901001)"))
+        };
+
+        // Act
+        var response = await _context.ExecuteBatchAsync(SaveChangesOptions.BatchWithSingleChangeset | SaveChangesOptions.UseRelativeUri, batchRequest);
+
+        // Assert
+        foreach (var operationResponse in response)
```
habbes, line 164: If we expect a fixed number of responses, then I'd suggest not using a loop, and just assert on each response. Does the client guarantee a certain order of response? If so, then let's assert the responses in order as well.

habbes: This was not addressed. Why do we need a loop here? It seems each type of response is handled differently.

WanjohiSammy: This is marked as `Outdated` -> means you need to check for the latest commit.

WanjohiSammy: Already updated to use indexes. You can confirm by going to the files and review.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/BatchRequestTests/BatchRequestClientTests.cs
```diff
@@ -0,0 +1,192 @@
...
+    public class TestsStartup : TestStartupBase
+    {
+        public override void ConfigureServices(IServiceCollection services)
+        {
+            services.ConfigureControllers(typeof(BatchRequestTestsController), typeof(MetadataController));
+
+            services.AddControllers().AddOData(opt =>
+            {
+                opt.EnableQueryFeatures().AddRouteComponents("odata", DefaultEdmModel.GetEdmModel(), new DefaultODataBatchHandler());
+                opt.RouteOptions.EnableNonParenthesisForEmptyParameterFunction = true;
```
ElizabethOkerio, line 37: why do we need this?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/BatchRequestTests/BatchRequestClientTests.cs
```diff
@@ -0,0 +1,192 @@
...
+        };
+        _context.UpdateObject(accountToUpdate);
+
+        // Act
+        var response = await _context.SaveChangesAsync(SaveChangesOptions.BatchWithSingleChangeset | SaveChangesOptions.UseRelativeUri);
+
+        // Assert
+        Assert.Equal(4, response.Count());
+
+        var changeResponses = response.Cast<ChangeOperationResponse>().ToList();
```
ElizabethOkerio, line 106: Are all items returned of type `ChangeOperationResponse` even failures? Or in this case there are no failures.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/BatchRequestTests/BatchRequestWithRelativeUriTests.cs
```diff
@@ -0,0 +1,264 @@
...
+            await batchWriter.WriteEndBatchAsync();
+        }
+
+        // Act
+        var responseMessage = await requestMessage.GetResponseAsync();
+
+        // Assert
+        Assert.Equal(200, responseMessage.StatusCode);
+
+        await ProcessBatchResponseAsync(responseMessage);
```
ElizabethOkerio, line 145: Is it possible to move the logic in the helper method to this test to avoid asserting in the helper class and to make it easy to see what the test is doing.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/BatchRequestTests/BatchRequestClientTests.cs
WanjohiSammy: The expectation at this point is no failure is returned

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/BatchRequestTests/BatchRequestClientTests.cs
WanjohiSammy: It is not needed. Removed

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/BatchRequestTests/BatchRequestWithRelativeUriTests.cs
WanjohiSammy: I have moved the logic from the helper method to the test method. However, this has resulted in the test becoming quite large.

ElizabethOkerio (approved):

PR #3168, opened 2025-01-22
Fixes #3165: UriParser - Aggregation method Average doesn't support Int16

### LINKED ISSUE #3165: UriParser - Aggregation method Average doesn't support Int16
Hello,

when working with `ODataUriParser` instance I've ran into an exception when an aggregation function average does not support attribute of the type Int16 (or Byte for that matter). The exception comes from the file  [ApplyBinder.cs](https://github.com/OData/odata.net/blob/c1c9b84f12b69584e14b402c0acff5cc99b60f2a/src/Microsoft.OData.Core/UriParser/Aggregation/ApplyBinder.cs#L190). In my eyes this is incorrect behavior. If you decide that this is indeed a bug and not desired behaviour then I can provide my capacity to fix this issue.

Example query:  `$apply=aggregate(Short with average as AverageLength)`

Short attribute must be of type Int16 in edm model.

Thank you for response!

davidKolar175:
### Issues

This pull request fixes [issue](https://github.com/OData/odata.net/issues/3165) where the problem was that during the parsing of average aggregation function the parser could throw an exception saying that it does not support int16. (Or Byte and SByte for that matter)

### Description

Easy fix that required just adding few cases into a switch.

- [ ] *Test cases added*
- [ ] *Build and test with one-click build and test script passed* (No since I don't have VS 2019 :/ but in VS 2020 no additional tests failed)

### Additional work necessary

None

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Aggregation/ApplyBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Aggregation/ApplyBinder.cs
@@ -176,6 +176,9 @@ namespace Microsoft.OData.UriParser.Aggregation
                     EdmPrimitiveTypeKind expressionPrimitiveKind = expressionType.PrimitiveKind();
                     switch (expressionPrimitiveKind)
                     {
+                        case EdmPrimitiveTypeKind.SByte:
+                        case EdmPrimitiveTypeKind.Byte:
+                        case EdmPrimitiveTypeKind.Int16:
                         case EdmPrimitiveTypeKind.Int32:
                         case EdmPrimitiveTypeKind.Int64:
                         case EdmPrimitiveTypeKind.Double:
--- a/src/Microsoft.OData.Core/UriParser/Aggregation/ApplyBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Aggregation/ApplyBinder.cs
@@ -176,8 +176,6 @@ namespace Microsoft.OData.UriParser.Aggregation
                     EdmPrimitiveTypeKind expressionPrimitiveKind = expressionType.PrimitiveKind();
                     switch (expressionPrimitiveKind)
                     {
-                        case EdmPrimitiveTypeKind.SByte:
-                        case EdmPrimitiveTypeKind.Byte:
                         case EdmPrimitiveTypeKind.Int16:
                         case EdmPrimitiveTypeKind.Int32:
                         case EdmPrimitiveTypeKind.Int64:
--- a/test/UnitTests/Microsoft.OData.Core.Tests/UriParser/Extensions/Binders/ApplyBinderTests.cs
+++ b/test/UnitTests/Microsoft.OData.Core.Tests/UriParser/Extensions/Binders/ApplyBinderTests.cs
@@ -838,6 +838,27 @@ namespace Microsoft.OData.Tests.UriParser.Extensions.Binders
             Assert.Empty(homeNode.ChildTransformations);
         }
 
+        [Fact]
+        public void BindApplyWithAverageWithInt16PropertyShouldReturnApplyClause()
+        {
+            IEnumerable<QueryToken> tokens = _parser.ParseApply("aggregate(FavoriteNumber with average as AverageFavoriteNumber)");
+
+            ApplyBinder binder = new ApplyBinder(FakeBindMethods.BindMethodReturningASingleFloatPrimitive, _bindingState);
+            ApplyClause actual = binder.BindApply(tokens);
+
+            Assert.NotNull(actual);
+            AggregateTransformationNode aggregate = Assert.IsType<AggregateTransformationNode>(Assert.Single(actual.Transformations));
+
+            Assert.Equal(TransformationNodeKind.Aggregate, aggregate.Kind);
+            Assert.NotNull(aggregate.AggregateExpressions);
+
+            AggregateExpression statement = Assert.IsType<AggregateExpression>(Assert.Single(aggregate.AggregateExpressions));
+            Assert.NotNull(statement.Expression);
+            Assert.Same(FakeBindMethods.FakeSingleFloatPrimitive, statement.Expression);
+            Assert.Equal(AggregationMethod.Average, statement.Method);
+            Assert.Equal("AverageFavoriteNumber", statement.Alias);
+        }
+
         private static ConstantNode _booleanPrimitiveNode = new ConstantNode(true);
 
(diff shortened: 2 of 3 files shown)


davidKolar175:
@microsoft-github-policy-service agree [company="TescoSW"]

davidKolar175:
@microsoft-github-policy-service agree company="TescoSW"

xuzhg:
The change looks straightforward, do you mind to add test to cover the changes?

src/Microsoft.OData.Core/UriParser/Aggregation/ApplyBinder.cs
```diff
@@ -176,6 +176,9 @@ private IEdmTypeReference CreateAggregateExpressionTypeReference(SingleValueNode
                     EdmPrimitiveTypeKind expressionPrimitiveKind = expressionType.PrimitiveKind();
                     switch (expressionPrimitiveKind)
                     {
+                        case EdmPrimitiveTypeKind.SByte:
+                        case EdmPrimitiveTypeKind.Byte:
+                        case EdmPrimitiveTypeKind.Int16:
```
gathogojr, line 181: I just realized that the types we support here correspond with to the types supported by the generic `Average` method supported by the framework https://learn.microsoft.com/en-us/dotnet/api/system.linq.enumerable.average?view=net-9.0. There's no `Average` method that accepts a collection of Int16.

Furthermore, we'd need to ensure parity with the ASP.NET Core OData library otherwise we'd end up with a bug in the dependent library due to the mappings here:
https://github.com/OData/AspNetCoreOData/blob/0ceeb086e4fd7a6396b4c6e0fd4f123283d55736/src/Microsoft.AspNetCore.OData/Query/ExpressionHelperMethods.cs#L86

We could cast `Int16` to `Int32` in order to support your scenario, but we need to discuss this more broadly before moving forward with this change.

Finally, I don't think we should add `SByte` and `Byte` for now. I think such scenarios are much less likely. Either way, this pull request doesn't add tests for those two additions.

test/UnitTests/Microsoft.OData.Core.Tests/UriParser/Extensions/Binders/ApplyBinderTests.cs
```diff
@@ -838,6 +838,27 @@ public void BindApplyWithGroupByComplexAndTypeCastShouldReturnApplyClause()
             Assert.Empty(homeNode.ChildTransformations);
         }
 
+        [Fact]
+        public void BindApplyWithAverageWithInt16PropertyShouldReturnApplyClause()
+        {
+            IEnumerable<QueryToken> tokens = _parser.ParseApply("aggregate(FavoriteNumber with average as AverageFavoriteNumber)");
```
gathogojr, line 844: `FavoriteNumber` is defined as `UInt16` in the referenced Edm model. I'd advise that we also add a test for `Int16` to cover all our bases

src/Microsoft.OData.Core/UriParser/Aggregation/ApplyBinder.cs
davidKolar175: Alright, I'll wait with the changes to the PR till your response with the result of your discussion. Is that ok?

davidKolar175: And just to add my two cents to this conversation. I do know that .NET's `Average` method does not contain sufficient override for Int16, but in my opinion thatis just another bug and `UriParser` should be agnostic to this. Why should a parser determine that having average over Int16 is forbidden? Especially when `Sum` is supported for Int16 by the parser.

Also, even if the types of `sbyte` and `byte` are less common, it still makes sense to average over them.

xuzhg (dismissed):

WanjohiSammy:
@davidKolar175 is it possible to an End to End tests. You can add them here:
https://github.com/OData/odata.net/tree/main/test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests

src/Microsoft.OData.Core/UriParser/Aggregation/ApplyBinder.cs
davidKolar175: @gathogojr How did the internal discussion go?

gathogojr: @davidKolar175 We'll incorporate support for Int16 with average. We'll not incorporate support for `byte` and `sbyte` just because we can - YAGNI principle. When an actual scenario arises (with reasonable justification) then we'll consider it. Let's restrict this pull request to the issue originally raised. @corranrogue9 planned to ensure that the test coverage was sufficient and plug any gaps if necessary.

davidKolar175: @gathogojr Thanks for the response and I understand. What is now required from my part? Should I change this PR to only count with Int16 properties and update the UnitTest I created?

gathogojr: @davidKolar175 Drop the support for `Byte` and `SByte` and update the test if necessary

test/UnitTests/Microsoft.OData.Core.Tests/UriParser/Extensions/Binders/ApplyBinderTests.cs
corranrogue9: @davidKolar175 I created a branch [here](https://github.com/OData/odata.net/tree/corranrogue9/davidKolar175fork/main) to add the extra test if you want to pull it into your working branch. I've assumed that you take @gathogojr's other feedback to remove `byte` and `sbyte`, though I'm happy to add those tests if you currently have a use case for those types.

davidKolar175: @corranrogue9 Thanks, I have used your test methods. :)

src/Microsoft.OData.Core/UriParser/Aggregation/ApplyBinder.cs
davidKolar175: @gathogojr Requested changes have been made. ;)

corranrogue9 (approved):

WanjohiSammy (approved):

habbes (approved):

gathogojr (approved):

PR #3169, opened 2025-01-22
Port EnumerationType Tests

WanjohiSammy:
### Issues

### Description

Port EnumerationType Tests from [`WCF`](https://learn.microsoft.com/en-us/dotnet/framework/wcf/whats-wcf) to use `AspNetCoreOData`

- [test/EndToEndTests/Tests/Client/Build.Desktop/EnumerationTypeTests](https://github.com/OData/odata.net/tree/main/test/EndToEndTests/Tests/Client/Build.Desktop/EnumerationTypeTests)

- [x] *Test cases added*

### Additional work necessary


test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/EnumerationTypeTests/Tests/EnumerationTypeQueryTests.cs
```diff
@@ -0,0 +1,649 @@
...
+using Microsoft.OData.Client.E2E.Tests.EnumerationTypeTests.Server;
+using Microsoft.OData.Edm;
+using Xunit;
+using AccessLevel = Microsoft.OData.Client.E2E.Tests.Common.Client.Default.AccessLevel;
+using Color = Microsoft.OData.Client.E2E.Tests.Common.Client.Default.Color;
+using Product = Microsoft.OData.Client.E2E.Tests.Common.Client.Default.Product;
+
+namespace Microsoft.OData.Client.E2E.Tests.EnumerationTypeTests.Tests;
+
+public class EnumerationTypeQueryTests : EndToEndTestBase<EnumerationTypeQueryTests.TestsStartup>
```
habbes, line 24: I suggest we have description comments to document the intent of these tests. I'm having a hard time finding the relation between the name of the test classes and the actual test methods. It's also not clear from the names of the test methods what is the behaviour that is actually being tested. A lot of the test methods make requests and assert on the response values, so it's hard to understand what's the difference between one test class and another, what is the focus of these tests?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/EnumerationTypeTests/Tests/EnumerationTypeQueryTests.cs
```diff
@@ -0,0 +1,649 @@
...
+        // Arrange
+        var queryText = "Products?$select=SkinColor,UserAccess";
+
+        // Act
+        List<ODataResource> entries = await TestsHelper.QueryResourceSetsAsync(queryText, mimeType);
+
+        // Assert
+        Assert.Equal(5, entries.Count);
+
+        Assert.DoesNotContain(entries[0].Properties, p => p.Name != "SkinColor" && p.Name != "UserAccess");
```
habbes, line 336: Should we also assert that it contains SkinColor and UserAccess? (Demonstrating that it doesn't contain properties that are not SkinColor nor UserAccess is not sufficient that prove that it contains SkinColor and UserAccess), it could contain no properties).

habbes (approved):

PR #3170, opened 2025-01-23
Added CollectionGenericType to ClientPropertyAnnotation

### LINKED ISSUE #3124: OData.Client 8.2.1 does not materialize data service responses
**Assemblies affected**

- Microsoft.OData.Client 8.2.1
- Microsoft.OData.Core 8.2.1
- Microsoft.OData.Edm 8.2.1
- Microsoft.Spatial 8.2.1

**Other libraries used in the project:**

- Microsoft.EntityFrameworkCore 9.0.0
- Microsoft.EntityFrameworkCore.Design 9.0.0
- Microsoft.EntityFrameworkCore.SqlServer 9.0.0
- Microsoft.EntityFrameworkCore.Tools 9.0.0
- Microsoft.AspNetCore.OData 9.1.0

**Describe the bug**

In all requests made by using a unit of work that encapsulates a DataServiceContext, it is not possible to materialize the results, obtaining the same error in all methods "Value cannot be null (Parameter 'key')".

Multiple client applications were programmed on Blazor Server, Wpf Core, among others. These must consume multiple services made with AspNetCore.OData, prior to updating to the assemblies mentioned above, everything worked fine, at this time the calls to the OData services do not materialize. 
AspNetCore.OData services work well and return results correctly.

**Reproduce steps**

- Create an API with AspNetCore.OData, which consumes data from a database in Sql Server with or without primary keys.
- Use the OData Connected Service extension to read the metadata and create the class for consumption in client applications.
- Make a request to the data service, and it should throw the error mentioned above.

**Data Model**

image

**EDM (CSDL) Model**

image

**Request/Response**

Request:
image

Response: 
The following image shows that the data service adequately responds to all methods:
image
The response it´s empty on blazor server app or Wpf app.

**Expected behavior**

Whether the data service returns IQueryable<T>, Enumerable<T>, List<T>, etc. It is expected to be able to materialize these results for consumption in client applications.

**Screenshots**
image
image
image
image

**Additional context**

If you use Microsoft.OData.Client 8.1.0, everything works as expected. Leaving client applications with Net 9 and data services with AspNetCore.Odata 9.1.0 and EntityFrameworkCore 9.0.0.

lawynn:
Modified EntryValueMaterializationPolicy to ensure the correct collection element type is used if a POCO does not have a collection initializer for collection properties.

### Issues

*This pull request fixes #3124.*

### Description

In cases where POCO classes are used, and contain collection properties, the materializer does not correctly identify the element type.  The property.EntityCollectionItemType erroneously returns a null for complex types, which then fails on the MakeGenericType call.

This fix surfaces the CollectionGenericType, and uses that in the 2 logic branches that will fail if the value is null. 

- [ ] *Test cases added*

### Additional work necessary


WanjohiSammy:
@lawynn Could you please add tests

WanjohiSammy:
@lawynn Thanks for taking some time to contribute on this. Please add some tests.

lawynn:
> @lawynn Thanks for taking some time to contribute on this. Please add some tests.

thank you, unfortunately I don't have a lot of time left to work on this at the moment.  It looks as though there may be some test coverage already.  I'll see if I can carve out some time in the next few weeks to get back to this issue.

leoerlandsson:
Any progress on this? Can we help in any way?

We are seeing the exact same problem when updating to later OData Client versions.

lawynn:
> Any progress on this? Can we help in any way?
> 
> We are seeing the exact same problem when updating to later OData Client versions.

Apologies, I got swamped and have not had time to get back to this issue. 

This PR corrected the issue for my particular case, but I see similar patterns of code elsewhere.  Depending on which logic path is taken, it may incorrectly use the EntityCollectionItemType (EntityType), rather than the CollectionGenericType (Collection<EntityType>).   

I only addressed the issue that I was seeing, @leoerlandsson, were you able to build this branch, and did it resolve your issue?  If not, then your EDM setup may be different than mine, and you may be running into some of the other logic branches.

PR #3177, opened 2025-02-14
Port WriteJsonPayloadTests

ElizabethOkerio:
### Description

This Pr ports these tests: [odata.net/test/EndToEndTests/Tests/Client/Build.Desktop/WriteJsonPayloadTests at main · OData/odata.net (github.com)](https://github.com/OData/odata.net/tree/main/test/EndToEndTests/Tests/Client/Build.Desktop/WriteJsonPayloadTests) from using WCF services to using Microsoft.AspNetCore.OData services. 

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
+++ b/test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
@@ -1,0 +1,747 @@
+﻿//-----------------------------------------------------------------------------
+// <copyright file="SerializationInfoEdgeTests.cs" company=".NET Foundation">
+//      Copyright (c) .NET Foundation and Contributors. All rights reserved.
+//      See License.txt in the project root for license information.
+// </copyright>
+//------------------------------------------------------------------------------
+
+using Microsoft.AspNetCore.OData;
+using Microsoft.AspNetCore.OData.Routing.Controllers;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.OData.E2E.TestCommon;
+using Microsoft.OData.E2E.TestCommon.Common;
+using Microsoft.OData.E2E.TestCommon.Common.Server.EndToEnd;
+using Microsoft.OData.Edm;
+
+namespace Microsoft.OData.Core.E2E.Tests.WriteJsonPayloadTests
+{
+    /// <summary>
+    /// This class tests OData serialization behavior, particularly
+    /// how OData payloads are written when serialization information is either included or omitted
+    /// </summary>
+    public class SerializationInfoEdgeTests : EndToEndTestBase<SerializationInfoEdgeTests.TestsStartup>
+    {
+        private readonly Uri _baseUri;
+        private readonly IEdmModel _model;
+        private static string NameSpacePrefix = "Microsoft.OData.E2E.TestCommon.Common.Server.EndToEnd.";
+
+        public class TestsStartup : TestStartupBase
+        {
+            public override void ConfigureServices(IServiceCollection services)
+            {
+                services.ConfigureControllers(typeof(MetadataController));
+
+                services.AddControllers().AddOData(opt => opt.Count().Filter().Expand().Select().OrderBy().SetMaxTop(null)
+                    .AddRouteComponents("odata", CommonEndToEndEdmModel.GetEdmModel()));
+            }
+        }
+
+        public SerializationInfoEdgeTests(TestWebApplicationFactory<TestsStartup> fixture)
+            : base(fixture)
+        {
+            _baseUri = new Uri(Client.BaseAddress, "odata/");
+            _model = CommonEndToEndEdmModel.GetEdmModel();
+        }
+
+        [Theory]
+        [InlineData(MimeTypeODataParameterFullMetadata)]
+        [InlineData(MimeTypeODataParameterMinimalMetadata)]
+        public async Task WritingAnEntryWithoutSerializationInfo_WithMetadata_ShouldThrowAnException(string mimeType)
+        {
+            var settings = CreateODataMessageWriterSettings();
+            var responseMessage = CreateTestResponseMessage(mimeType);
+
+            using var messageWriter = new ODataMessageWriter(responseMessage, settings);
+            var odataWriter = await messageWriter.CreateODataResourceWriterAsync();
+            var entry = CreatePersonEntryWithoutSerializationInfo();
+
+            var expectedError = Error.Format(SRResources.ODataContextUriBuilder_NavigationSourceOrTypeNameMissingForResourceOrResourceSet);
+            var exception = await Assert.ThrowsAsync<ODataException>(async () => await odataWriter.WriteStartAsync(entry));
+
+            Assert.Equal(expectedError, exception.Message);
+        }
+
+        [Fact]
+        public async Task WritingAnEntryWithoutSerializationInfo_WithoutMetadata_ShouldWriteAnEntryWithoutODataMetadata()
+        {
+            var settings = CreateODataMessageWriterSettings();
+            var responseMessage = CreateTestResponseMessage(MimeTypeODataParameterNoMetadata);
+
+            using var messageWriter = new ODataMessageWriter(responseMessage, settings);
+            var odataWriter = await messageWriter.CreateODataResourceWriterAsync();
+            var entry = CreatePersonEntryWithoutSerializationInfo();
+            await odataWriter.WriteStartAsync(entry);
+            await odataWriter.WriteEndAsync();
+
+            Stream stream = await responseMessage.GetStreamAsync();
+            string result = await TestsHelper.ReadStreamContentAsync(stream);
+
+            Assert.Equal("{\"PersonId\":-5,\"Name\":\"xhsdckkeqzvlnprheujeycqrglfehtdocildrequohlffazfgtvmddyqsaxrojqxrsckohrakdxlrghgmzqnyruzu\"}", result);
+        }
+
+        [Theory]
+        [InlineData(MimeTypeODataParameterFullMetadata)]
+        [InlineData(MimeTypeODataParameterMinimalMetadata)]
+        public async Task WritingAFeedWithoutSerializationInfo_WithMetadata_ShouldThrowAnException(string mimeType)
+        {
+            var settings = CreateODataMessageWriterSettings();
+            var responseMessage = CreateTestResponseMessage(mimeType);
+
+            using var messageWriter = new ODataMessageWriter(responseMessage, settings);
+            var odataWriter = await messageWriter.CreateODataResourceSetWriterAsync();
+            var feed = CreatePersonFeed();
+            var entry = CreatePersonEntryWithoutSerializationInfo();
+
+            entry.SetSerializationInfo(new ODataResourceSerializationInfo()
+            {
+                NavigationSourceName = "People",
+                NavigationSourceEntityTypeName = NameSpacePrefix + "Person"
+            });
+
+            var expectedError = Error.Format(SRResources.ODataContextUriBuilder_NavigationSourceOrTypeNameMissingForResourceOrResourceSet);
+            var exception = await Assert.ThrowsAsync<ODataException>(async () => await odataWriter.WriteStartAsync(feed));
+
+            Assert.Equal(expectedError, exception.Message);
+        }
+
+        [Fact]
+        public async Task WritingAFeedWithoutSerializationInfo_WithoutMetadata_ShouldWriteAnEmptyArray()
+        {
+            var settings = CreateODataMessageWriterSettings();
+            var responseMessage = CreateTestResponseMessage(MimeTypeODataParameterNoMetadata);
+
+            using var messageWriter = new ODataMessageWriter(responseMessage, settings);
+            var odataWriter = await messageWriter.CreateODataResourceSetWriterAsync();
+            var feed = CreatePersonFeed();
+            var entry = CreatePersonEntryWithoutSerializationInfo();
+
+            entry.SetSerializationInfo(new ODataResourceSerializationInfo()
+            {
+                NavigationSourceName = "People",
+                NavigationSourceEntityTypeName = NameSpacePrefix + "Person"
+            });
+
+            await odataWriter.WriteStartAsync(feed);
+            await odataWriter.WriteEndAsync();
+
+            Stream stream = await responseMessage.GetStreamAsync();
+            string result = await TestsHelper.ReadStreamContentAsync(stream);
+
+            Assert.Equal("{\"value\":[]}", result);
+        }
+
+        [Theory]
+        [InlineData(MimeTypeODataParameterFullMetadata)]
+        [InlineData(MimeTypeODataParameterMinimalMetadata)]
+        public async Task WritingACollectionWithoutSerializationInfo_WithMetadata_ShouldThrowAnException(string mimeType)
+        {
+            var settings = CreateODataMessageWriterSettings();
+            var responseMessage = CreateTestResponseMessage(mimeType);
+
+            using var messageWriter = new ODataMessageWriter(responseMessage, settings);
+            var odataWriter = await messageWriter.CreateODataCollectionWriterAsync();
+            var collectionStart = new ODataCollectionStart() { Name = "BackupContactInfo" };
+
+            var expectedError = Error.Format(SRResources.ODataContextUriBuilder_TypeNameMissingForTopLevelCollection);
+            var exception = await Assert.ThrowsAsync<ODataException>(async () => await odataWriter.WriteStartAsync(collectionStart));
+
+            Assert.Equal(expectedError, exception.Message);
+        }
+
+        [Fact]
+        public async Task WritingACollectionWithoutSerializationInfo_WithoutMetadata_ShouldWriteAnEmptyArray()
+        {
+            var settings = CreateODataMessageWriterSettings();
+            var responseMessage = CreateTestResponseMessage(MimeTypeODataParameterNoMetadata);
+
+            using var messageWriter = new ODataMessageWriter(responseMessage, settings);
+            var odataWriter = await messageWriter.CreateODataCollectionWriterAsync();
+            var collectionStart = new ODataCollectionStart() { Name = "BackupContactInfo" };
+
+            await odataWriter.WriteStartAsync(collectionStart);
+            await odataWriter.WriteEndAsync();
+
+            Stream stream = await responseMessage.GetStreamAsync();
+            string result = await TestsHelper.ReadStreamContentAsync(stream);
+
+            Assert.Equal("{\"value\":[]}", result);
+        }
+
+        [Theory]
+        [InlineData(MimeTypeODataParameterFullMetadata)]
+        [InlineData(MimeTypeODataParameterMinimalMetadata)]
+        public async Task WritingAReferenceLinkWithoutSerializationInfo_WithMetadata_ShouldContainMetadataReference(string mimeType)
+        {
+            var settings = CreateODataMessageWriterSettings();
+            var responseMessage = CreateTestResponseMessage(mimeType);
+
+            using var messageWriter = new ODataMessageWriter(responseMessage, settings);
+            var link = new ODataEntityReferenceLink() { Url = new Uri(_baseUri + "Orders(-10)") };
+            await messageWriter.WriteEntityReferenceLinkAsync(link);
+
+            Stream stream = await responseMessage.GetStreamAsync();
+            string result = await TestsHelper.ReadStreamContentAsync(stream);
+
+            Assert.Contains("$metadata#$ref", result);
+        }
+
+        [Fact]
+        public async Task WritingAReferenceLinkWithoutSerializationInfo_WithMetadata_ShouldNotContainMetadataReference()
+        {
+            var settings = CreateODataMessageWriterSettings();
+            var responseMessage = CreateTestResponseMessage(MimeTypeODataParameterNoMetadata);
+
+            using var messageWriter = new ODataMessageWriter(responseMessage, settings);
+            var link = new ODataEntityReferenceLink() { Url = new Uri(_baseUri + "Orders(-10)") };
+            await messageWriter.WriteEntityReferenceLinkAsync(link);
+
+            Stream stream = await responseMessage.GetStreamAsync();
+            string result = await TestsHelper.ReadStreamContentAsync(stream);
+
+            Assert.DoesNotContain("$metadata#$ref", result);
+        }
+
+        [Theory]
+        [InlineData(MimeTypeODataParameterFullMetadata)]
+        [InlineData(MimeTypeODataParameterMinimalMetadata)]
+        [InlineData(MimeTypeODataParameterNoMetadata)]
+        public async Task WritingARequestMessageWithEntry_ShouldContainExpectedContent(string mimeType)
+        {
+            var settings = CreateODataMessageWriterSettings();
+            var requestMessage = new TestStreamRequestMessage(new MemoryStream(), new Uri(_baseUri + "People"), "POST");
+            requestMessage.SetHeader("Content-Type", mimeType);
+
+            using var messageWriter = new ODataMessageWriter(requestMessage, settings);
+            var odataWriter = await messageWriter.CreateODataResourceWriterAsync();
+            var entry = CreatePersonEntryWithoutSerializationInfo();
+
+            await odataWriter.WriteStartAsync(entry);
+            await odataWriter.WriteEndAsync();
+
+            Stream stream = await requestMessage.GetStreamAsync();
+            string result = await TestsHelper.ReadStreamContentAsync(stream);
+
+            Assert.Contains("People(-5)\",\"PersonId\":-5,\"Name\":\"xhsdckkeqzvlnprheujeycqrglfehtdocildrequohlffazfgtvmddyqsaxrojqxrsckohrakdxlrghgmzqnyruzu\"", result);
+        }
+
+        [Theory]
+        [InlineData(MimeTypeODataParameterFullMetadata)]
+        [InlineData(MimeTypeODataParameterMinimalMetadata)]
+        public async Task SpecifyingSerializationInfoForBothFeedAndEntry_WithMetadata_ShouldContainExpectedEntryType(string mimeType)
+        {
+            var settings = CreateODataMessageWriterSettings();
+            var responseMessage = CreateTestResponseMessage(mimeType);
+
+            using (var messageWriter = new ODataMessageWriter(responseMessage, settings))
+            {
+                var odataWriter = await messageWriter.CreateODataResourceSetWriterAsync();
+                var feed = CreatePersonFeed();
+                feed.SetSerializationInfo(
+                    new ODataResourceSerializationInfo()
+                    {
+                        NavigationSourceName = "People",
+                        NavigationSourceEntityTypeName = NameSpacePrefix + "Person"
+                    });
+
+                var entry = new ODataResource()
+                {
+                    Id = new Uri(_baseUri + "People(-5)"),
+                    TypeName = NameSpacePrefix + "Employee"
+                };
+
+                var personEntryP1 = new ODataProperty { Name = "PersonId", Value = -5 };
+                var personEntryP2 = new ODataProperty
+                {
+                    Name = "Name",
+                    Value = "xhsdckkeqzvlnprheujeycqrglfehtdocildrequohlffazfgtvmddyqsaxrojqxrsckohrakdxlrghgmzqnyruzu"
+                };
+
+                var personEntryP3 = new ODataProperty
+                {
+                    Name = "ManagersPersonId",
+                    Value = -465010984
+                };
+
+                entry.Properties = new[] { personEntryP1, personEntryP2, personEntryP3 };
+                entry.SetSerializationInfo(
+                    new ODataResourceSerializationInfo()
+                    {
+                        NavigationSourceName = "Person",
+                        NavigationSourceEntityTypeName = NameSpacePrefix + "Person"
+                    });
+
+
+                await odataWriter.WriteStartAsync(feed);
+                await odataWriter.WriteStartAsync(entry);
+                await odataWriter.WriteEndAsync();
+                await odataWriter.WriteEndAsync();
+            }
+
+            Stream stream = await responseMessage.GetStreamAsync();
+            string result = await TestsHelper.ReadStreamContentAsync(stream);
+
+            Assert.Contains(NameSpacePrefix + "Employee", result);
+            Assert.Contains("$metadata#People", result);
+        }
+
+        [Fact]
+        public async Task SpecifyingSerializationInfoForBothFeedAndEntry_WithoutMetadata_ShouldNotContainExpectedEntryType()
+        {
+            var settings = CreateODataMessageWriterSettings();
+            var responseMessage = CreateTestResponseMessage(MimeTypeODataParameterNoMetadata);
+
+            using (var messageWriter = new ODataMessageWriter(responseMessage, settings))
+            {
+                var odataWriter = await messageWriter.CreateODataResourceSetWriterAsync();
+                var feed = CreatePersonFeed();
+                feed.SetSerializationInfo(
+                    new ODataResourceSerializationInfo()
+                    {
+                        NavigationSourceName = "People",
+                        NavigationSourceEntityTypeName = NameSpacePrefix + "Person"
+                    });
+
+                var entry = new ODataResource()
+                {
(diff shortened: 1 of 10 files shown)


test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,488 @@
...
+using Microsoft.AspNetCore.OData.Routing.Controllers;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.OData.E2E.TestCommon;
+using Microsoft.OData.E2E.TestCommon.Common;
+using Microsoft.OData.E2E.TestCommon.Common.Server.EndToEnd;
+using Microsoft.OData.Edm;
+
+namespace Microsoft.OData.Core.E2E.Tests.WriteJsonPayloadTests
+{
+    public class SerializationInfoEdgeTests : EndToEndTestBase<SerializationInfoEdgeTests.TestsStartup>
```
habbes, line 18: Consider adding a doc comment explaining what these tests are about. It's not clear just from the name what the focus area for these tests are.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,488 @@
...
+
+        /// <summary>
+        /// Write payloads without model or serialization info.
+        /// Note that ODL will succeed if writing requests or writing nometadata reponses.
+        /// </summary>
+        [Theory]
+        [InlineData(MimeTypeODataParameterFullMetadata)]
+        [InlineData(MimeTypeODataParameterMinimalMetadata)]
+        [InlineData(MimeTypeODataParameterNoMetadata)]
+        public async Task WriteWithoutSerializationInfo(string mimeType)
```
habbes, line 50: It seems that this test tests different scenarios with different assertions and expectations. We create a new message writer and response for each test. It's hard to tell what scenarios are being tested from a quick glance, and it's harder to tell what the expected behaviour is because we're mixing different scenarios in the same test case. I suggest splitting the different scenarios in separate test methods, and to give each one a name that helps us understand the expected behaviour.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,488 @@
...
+                await odataWriter.WriteStartAsync(entry);
+                await odataWriter.WriteEndAsync();
+                await odataWriter.WriteEndAsync();
+                Stream stream = await responseMessageWithoutModel.GetStreamAsync();
+                string result = await TestsHelper.ReadStreamContentAsync(stream);
+                if (!mimeType.Contains(MimeTypes.ODataParameterNoMetadata))
+                {
+                    Assert.Contains(NameSpacePrefix + "Employee", result);
+                    Assert.Contains("$metadata#People", result);
+                }
```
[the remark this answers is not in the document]
habbes: Alternatively, you could add a test parameter that represents whether or not the context is included in the result

```c#
SpecifySerializationInfoForFeedAndEntry(string mimeType, bool shouldIncludeTypeInContext)
```

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,488 @@
...
+                var entry = this.CreatePersonEntryWithoutSerializationInfo();
+                await odataWriter.WriteStartAsync(entry);
+                await odataWriter.WriteEndAsync();
+                Stream stream = await requestMessageWithoutModel.GetStreamAsync();
+                Assert.Contains("People(-5)\",\"PersonId\":-5,\"Name\":\"xhsdckkeqzvlnprheujeycqrglfehtdocildrequohlffazfgtvmddyqsaxrojqxrsckohrakdxlrghgmzqnyruzu\"", await TestsHelper.ReadStreamContentAsync(stream));
+            }
+        }
+
+        /// <summary>
+        /// Specify serialization info for both feed and entry should fail.
```
habbes, line 153: What exactly is this test testing? It says that specify serialization for both feed and entry should fail, however, there are no failure assertion in the test. I'm not sure I understand the exact behaviour that we're evaluating here (or in this class in general).

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,488 @@
...
+                if (!mimeType.Contains(MimeTypes.ODataParameterNoMetadata))
+                {
+                    // no metadata does not write odata.metadata
+                    Assert.Contains("$metadata#Parsen/$entity", result);
+                    Assert.Contains("People(-5)\",\"PersonId\":-5", result);
+                }
+                else
+                {
+                    Assert.DoesNotContain("People(-5)\",\"PersonId\":-5", result);
+                }
```
habbes, line 242: Consider avoiding conditions that make it harder to tell what is being tested when. We could split the test or parameterize the test to avoid conditions.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,488 @@
...
+                else
+                {
+                    Assert.DoesNotContain(NameSpacePrefix + "Employee", result);
+                }
+            }
+        }
+
+        /// <summary>
+        /// Write response payload with serialization info containing wrong values
```
habbes, line 209: What is the behaviour being tested? What is the expected outcome?

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,488 @@
...
+        /// </summary>
+        [Theory]
+        [InlineData(MimeTypeODataParameterFullMetadata)]
+        [InlineData(MimeTypeODataParameterMinimalMetadata)]
+        [InlineData(MimeTypeODataParameterNoMetadata)]
+        public async Task WriteEntryWithWrongSerializationInfo(string mimeType)
```
habbes, line 215: This test also appears to handle multiple scenarios, each with its own response message and message writer. Consider splitting those into separate tests.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,488 @@
...
+        /// </summary>
+        [Theory]
+        [InlineData(MimeTypeODataParameterFullMetadata)]
+        [InlineData(MimeTypeODataParameterMinimalMetadata)]
+        [InlineData(MimeTypeODataParameterNoMetadata)]
+        public async Task WriteEntryWithWrongSerializationInfoWithModel(string mimeType)
+        {
+            bool[] autoComputeMetadataBools = new bool[] { true, false, };
+
+            foreach (var autoComputeMetadata in autoComputeMetadataBools)
```
habbes, line 325: This looks like a good candidate for parameterizing the test further instead of having a loop.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,488 @@
...
+        }
+
+        /// <summary>
+        /// Change serialization info value after WriteStart
+        /// </summary>
+        [Theory]
+        [InlineData(MimeTypeODataParameterFullMetadata)]
+        [InlineData(MimeTypeODataParameterMinimalMetadata)]
+        [InlineData(MimeTypeODataParameterNoMetadata)]
+        public async Task ChangeSerializationInfoAfterWriteStart(string mimeType)
```
habbes, line 362: What is the expected outcome of this scenario?

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,488 @@
...
+                entry.SetSerializationInfo(new ODataResourceSerializationInfo() { NavigationSourceName = "Parsen", NavigationSourceEntityTypeName = NameSpacePrefix + "Person" });
+                await odataWriter.WriteEndAsync();
+
+                Stream stream = await responseMessageWithoutModel.GetStreamAsync();
+
+                // nometadata option does not write odata.metadata in the payload
+                if (!mimeType.Contains(MimeTypes.ODataParameterNoMetadata))
+                {
+                    Assert.Contains("People/$entity", (await TestsHelper.ReadStreamContentAsync(stream)));
+                }
```
habbes, line 385: Consider making this a parameter. Note that there's a different between:
- not asserting that some condition is true
- asserting that some condition is not true

In most of your tests you're doing the former, which does not give us any insight about the latter. Because we have behaviour that's not defined or acknowledged by this test, and we don't know if that's intentional or not. If nometadata leads to odata.metadata not being written and that's expected, then why not assert that?

Adding a boolean parameter to this test would allow having a single assert statement for all scenarios and we could look at the parameters to tell when it's expected to be true or false. Or we could have separate tests if those scenarios are different enough.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,488 @@
...
+        {
+            var orderFeed = new ODataResourceSet()
+            {
+                Id = new Uri(_baseUri + "People"),
+            };
+
+            return orderFeed;
+        }
+
+        private static async Task AssertThrowsAsync<T>(Func<Task> action, string expectedError) where T : Exception
```
habbes, line 464: Doesn't xUnit already provide helpers for exceptions?

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/WritePayloadTests.cs
```diff
@@ -0,0 +1,1132 @@
...
+            {
+                var rex = new Regex("\"\\w*@odata.type\":\"#[\\w\\(\\)\\.]*\",");
+                var outputWithoutModel2 = rex.Replace(outputWithoutModel, "");
+                TestsHelper.VerifyPayloadString(outputWithModel, outputWithoutModel2, mimeType);
+            }
+            else
+            {
+                //NoMetadata with/out model should result in same output
+                Assert.Equal(outputWithModel, outputWithoutModel);
+            }
```
habbes, line 104: If we can't create suitable parameters to allow us to have a more "linear" and predictable test, maybe we should split the scenario into separate tests.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/WritePayloadTests.cs
```diff
@@ -0,0 +1,1132 @@
...
+            var responseMessageWithoutModel = new TestStreamResponseMessage(new MemoryStream());
+            responseMessageWithoutModel.SetHeader("Content-Type", mimeType);
+            responseMessageWithoutModel.PreferenceAppliedHeader().AnnotationFilter = "*";
+            using (var messageWriter = new ODataMessageWriter(responseMessageWithoutModel, settings))
+            {
+                var odataWriter = await messageWriter.CreateODataResourceWriterAsync();
+                outputWithoutModel = await WriteAndVerifyCarEntryAsync(responseMessageWithoutModel, odataWriter, false,mimeType);
+            }
+
+            TestsHelper.VerifyPayloadString(outputWithModel, outputWithoutModel, mimeType);
```
habbes, line 203: What does this method do?

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/WritePayloadTests.cs
```diff
@@ -0,0 +1,1132 @@
...
+        }
+
+        /// <summary>
+        /// Write an entry containing stream, named stream
+        /// </summary>
+        [Theory]
+        [InlineData(MimeTypeODataParameterFullMetadata)]
+        [InlineData(MimeTypeODataParameterMinimalMetadata)]
+        [InlineData(MimeTypeODataParameterNoMetadata)]
+        public async Task CarEntryTest(string mimeType)
```
habbes, line 171: What does this method test?

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/WritePayloadTests.cs
```diff
@@ -0,0 +1,1132 @@
...
+        }
+
+        /// <summary>
+        /// Write complex collection response
+        /// </summary>
+        [Theory]
+        [InlineData(MimeTypeODataParameterFullMetadata)]
+        [InlineData(MimeTypeODataParameterMinimalMetadata)]
+        [InlineData(MimeTypeODataParameterNoMetadata)]
+        public async Task ComplexCollectionTest(string mimeType)
```
habbes, line 341: What's the expected outcome?

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
ElizabethOkerio: I don't think this is intuitive. We are testing how to write a resource without serialization information using different mimeTypes. When using NoMimeType OData should succeed but if we are using either the MinimalMimeType or the FullMimeType then an exception should be thrown. Are you suggesting we have a test for each mime type when the expected result is the same?

ElizabethOkerio: Maybe the scenarios can be put in different test cases.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
ElizabethOkerio: The test shows specifying serialization info for both a feed and entry should fail. I've updated the doc and test name.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
ElizabethOkerio: I've split these tests

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
ElizabethOkerio: I've update the test

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
ElizabethOkerio: splitted

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
ElizabethOkerio: updated

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/WritePayloadTests.cs
ElizabethOkerio: It asserts if the `outputWithModel` and `outputWithoutModel` are equal. I've gotten rid of it.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/WritePayloadTests.cs
ElizabethOkerio: updated

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/WritePayloadTests.cs
ElizabethOkerio: updated

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/WritePayloadTests.cs
```diff
@@ -0,0 +1,1132 @@
...
+
+                while (await reader.ReadAsync())
+                {
+                    if (reader.State == ODataReaderState.ResourceEnd)
+                    {
+                        entry = reader.Item as ODataResource;
+                    }
+                }
+
+                Assert.True(entry.Id.ToString().Contains("Orders(-10)"), "entry.Id");
```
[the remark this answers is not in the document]
ElizabethOkerio: fixed

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,552 @@
...
+
+            if (containsExpectedTypeInfo)
+            {
+                Assert.Contains(NameSpacePrefix + "Employee", result);
+                Assert.Contains("$metadata#People", result);
+            }
+            else
+            {
+                Assert.DoesNotContain(NameSpacePrefix + "Employee", result);
+            }
```
habbes, line 246: This also looks like another case where separate test cases should be considered.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,552 @@
...
+
+            if (containsWrongMetadata)
+            {
+                Assert.Contains("$metadata#Parsen/$entity", result);
+                Assert.Contains("People(-5)\",\"PersonId\":-5", result);
+            }
+            else
+            {
+                Assert.DoesNotContain("People(-5)\",\"PersonId\":-5", result);
+            }
```
habbes, line 281: See: https://github.com/OData/odata.net/pull/3177/files#r1969052792

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,552 @@
...
+
+            if (containsWrongMetadata)
+            {
+                Assert.Contains("$metadata#Parsen\"", result);
+                Assert.Contains("People(-5)\",\"PersonId\":-5", result);
+            }
+            else
+            {
+                Assert.DoesNotContain("People(-5)\",\"PersonId\":-5", result);
+            }
```
habbes, line 315: See: https://github.com/OData/odata.net/pull/3177/files#r1969052792

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,552 @@
...
+
+            if (containsMetadataWithIncorrectTypeName)
+            {
+                Assert.Contains("\"value\":[{\"@odata.type\":\"#Microsoft.OData.E2E.TestCommon.Common.Server.EndToEnd.ContactDetails\",\"EmailBag", result);
+                Assert.Contains("$metadata#Collection(" + NameSpacePrefix + "ContactDETAIL)", result);
+            }
+            else
+            {
+                Assert.DoesNotContain("\"@odata.type\":\"#Microsoft.OData.E2E.TestCommon.Common.Server.EndToEnd.ContactDetails\"", result);
+            }
```
habbes, line 355: See: https://github.com/OData/odata.net/pull/3177/files#r1969052792

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,552 @@
...
+            Assert.Contains("\"PersonId\":-5", result);
+
+            if (containsIncorrectMetadata)
+            {
+                Assert.Contains("$metadata#Parsen/$entity", result);
+            }
+            else
+            {
+                Assert.DoesNotContain("$metadata", result);
+            }
```
habbes, line 391: See: https://github.com/OData/odata.net/pull/3177/files#r1969052792

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,552 @@
...
+            Assert.Contains("Name", responseContent);
+
+            if (hasMetadata)
+            {
+                Assert.Contains("People/$entity", responseContent);
+            }
+            else
+            {
+                Assert.DoesNotContain("People/$entity", responseContent);
+            }
```
habbes, line 495: See: https://github.com/OData/odata.net/pull/3177/files#r1969052792

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,552 @@
...
+            {
+                var expectedError = Error.Format(SRResources.ODataContextUriBuilder_TypeNameMissingForTopLevelCollection);
+                var exception = await Assert.ThrowsAsync<ODataException>(async () => await odataWriter.WriteStartAsync(collectionStart));
+                Assert.Equal(expectedError, exception.Message);
+            }
+            else
+            {
+                await odataWriter.WriteStartAsync(collectionStart);
+                await odataWriter.WriteEndAsync();
+            }
```
habbes, line 129: See: https://github.com/OData/odata.net/pull/3177/files#r1969049110

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,552 @@
...
+            string result = await TestsHelper.ReadStreamContentAsync(stream);
+
+            if (containsMetadataReference)
+            {
+                Assert.Contains("$metadata#$ref", result);
+            }
+            else
+            {
+                Assert.DoesNotContain("$metadata#$ref", result);
+            }
```
habbes, line 155: https://github.com/OData/odata.net/pull/3177/files#r1969052792

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/SerializationInfoEdgeTests.cs
```diff
@@ -0,0 +1,552 @@
...
+                // Change serialization info after WriteStartAsync
+                entry.SetSerializationInfo(new ODataResourceSerializationInfo() { NavigationSourceName = "Parsen", NavigationSourceEntityTypeName = NameSpacePrefix + "Person" });
+
+                await odataWriter.WriteEndAsync();
+
+                Stream stream = await responseMessageWithoutModel.GetStreamAsync();
+                var result = await TestsHelper.ReadStreamContentAsync(stream);
+
+                // Assert that the metadata still refers to "People", not "Parsen"
+                if (usesOriginalMetadata)
```
habbes, line 425: See: https://github.com/OData/odata.net/pull/3177/files#r1969052792

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/WritePayloadTests.cs
```diff
@@ -0,0 +1,1056 @@
...
+            if (hasODataType)
+            {
+                Assert.True(outputWithoutModel.Contains("{\"@odata.type\":\"" + "#" + NameSpacePrefix + "Employee\","), "odata.type Employee");
+                Assert.True(outputWithoutModel.Contains("{\"@odata.type\":\"" + "#" + NameSpacePrefix + "SpecialEmployee\","), "odata.type SpecialEmployee");
+            }
+
+            if (hasNoODataType)
+            {
+                Assert.False(outputWithoutModel.Contains("{\"@odata.type\":\"" + "#" + NameSpacePrefix + "Person\","), "odata.type Person");
+            }
```
habbes, line 201: If you're going to parameterize the test, is it possible to put parameters that contain actual values you're going to assert against, instead for conditions for whether or not you should execute the assertions.

For example, maybe you could have a parameter that contains the actual metadata information that should appear in the payload, then assert `Assert.True(outputWithoutModel.StartsWith(payloadStart)` instead of using if statements.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/WritePayloadTests.cs
```diff
@@ -0,0 +1,1056 @@
...
+            var orderSet = _model.EntityContainer.FindEntitySet("Orders");
+
+            Stream stream = await responseMessage.GetStreamAsync();
+
+            if (!mimeType.Contains(MimeTypes.ODataParameterNoMetadata))
+            {
+                stream.Seek(0, SeekOrigin.Begin);
+                await TestsHelper.ReadAndVerifyFeedEntryMessageAsync(true, responseMessage, orderSet, orderType, verifyFeed, verifyEntry, verifyNavigation);
+                Assert.True(verifyFeedCalled && verifyEntryCalled && verifyNavigationCalled, "Verification action not called.");
+            }
```
habbes, line 583: We should avoid burying assertion logic inside test helpers. This makes it hard to look at the understand and understand what it's doing or whether it's testing the right thing. This is probably something we do in other places across the library. We should look for better patterns that promote test readability and verifiability. It's easy to obscure incorrect logic and bugs in these helps that cause the tests to pass when they should fail without anyone finding out. In this case the reviewer of the test will not easily tell that this assertion is called and that it's only called when the mime type does not contain the specified constant, which is also obscured. This helper has a lot of conditions making it harder to easily reason about it. As a result, it becomes harder to verify the correctness of any test that relies on it with the naked eye.

ElizabethOkerio: This is not a helper method.

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/WritePayloadHelper.cs
```diff
@@ -0,0 +1,1322 @@
...
+                orderEntry1P1.SetSerializationInfo(new ODataPropertySerializationInfo()
+                {
+                    PropertyKind = ODataPropertyKind.Key
+                });
+            }
+
+            return orderEntry1;
+        }
+
+        public static ODataResourceWrapper CreateOrderEntry2NoMetadata(bool hasModel)
```
WanjohiSammy, line 756: Are there better ways of naming this method so that it can the OrderEntry that is being created. Probably, the different between `CreateOrderEntry1NoMetadata`, `CreateOrderEntry2NoMetadata` ... This will make it easier for anyone to know what this method is doing without having to go through the code.

WanjohiSammy (approved):

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/WritePayloadTests.cs
```diff
@@ -0,0 +1,837 @@
...
+        private async Task<string> WriteAnExpandedCustomerEntryAsync(
+            TestStreamResponseMessage responseMessage,
+            ODataWriter odataWriter,
+            bool hasModel,
+            string mimeType)
+        {
+            ODataResourceWrapper customerEntry = TestsHelper.CreateCustomerEntry(hasModel);
+
+            var loginFeed = new ODataResourceSet() { Id = new Uri(_baseUri + "Customers(-9)/Logins") };
+            if (!hasModel)
```
habbes, line 562: The reliance of compex helper code in our tests is still a concern. I will not block PR because of this but I'd want us to brainstorm and think of ways to avoid this pattern in our code. The risk with reliance on such helpers is that when reading the tests these helpers are blackboxes that we may not fully understand, and we may not always inspect exhaustively. Our tests relies on these helpers being correct, but we don't have guarantees that they are correct since we don't have test cases for most helpers. So if we have a bug in the helpers, it could cause a faulty test to pass and hide a bug.

habbes (approved):

test/EndToEndTests/Tests/Core/Microsoft.OData.Core.E2E.Tests/WriteJsonPayloadTests/WritePayloadTests.cs
ElizabethOkerio: I'll create a work item to track this.

habbes:
Run pipeline

ElizabethOkerio:
Run pipeline

habbes (approved):

PR #3180, opened 2025-02-17
Port PayloadValueConverter tests

ElizabethOkerio:
### Description

This Pr ports these tests: [odata.net/test/EndToEndTests/Tests/Client/Build.Desktop/PayloadValueConverterTests at release-7.x · OData/odata.net](https://github.com/OData/odata.net/tree/release-7.x/test/EndToEndTests/Tests/Client/Build.Desktop/PayloadValueConverterTests) from using WCF services to using Microsoft.AspNetCore.OData services.

- [ ] *Test cases added*

### Additional work necessary


xuzhg (dismissed):

test/EndToEndTests/Common/Microsoft.OData.E2E.TestCommon/Converters/BinaryPayloadConverter.cs
```diff
@@ -0,0 +1,34 @@
...
+//      Copyright (C) Microsoft Corporation. All rights reserved. See License.txt in the project root for license information.
+// </copyright>
+//---------------------------------------------------------------------
+
+using Microsoft.OData.Edm;
+using System.Globalization;
+
+namespace Microsoft.OData.E2E.TestCommon.Converters
+{
+    public class BinaryPayloadConverter : ODataPayloadValueConverter
```
habbes, line 12: Is this class used in other tests? If it's only used in the `PayloadValueConverterTests` then maybe it should be defined there?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PayloadValueConverterTests/PayloadValueConverterTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            {
+                HttpClientFactory = HttpClientFactory
+            };
+
+            _model = DefaultEdmModel.GetEdmModel();
+            ResetDataSource();
+        }
+
+        [Fact]
+        public async Task PostingAndQueryingABinaryProperty_UsingCustomBinaryToHyphenStringConverter_ShouldReturnHyphenSeparatedString()
```
habbes, line 58: If this is only testing that the payload converter generates the correct output, I'm wondering why this is an E2E test as opposed to just a test in AspNetCoreOData. In this case, it seems like client does nothing to influence the output of the server returns. Is there anything special that the client has to do to process this message? Is there any relevant client-side behaviour being verified in this test?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PayloadValueConverterTests/PayloadValueConverterTests.cs
```diff
@@ -0,0 +1,94 @@
...
+            await _context.SaveChangesAsync();
+
+            var requestUrl = new Uri(_baseUri.AbsoluteUri + "People(" + person.Id + ")/Picture", UriKind.Absolute);
+
+            var requestMessage = new TestHttpClientRequestMessage(requestUrl, base.Client)
+            {
+                Method = "GET"
+            };
+
+            var responseMessage = await requestMessage.GetResponseAsync();
```
habbes, line 79: It seems that we're not using the OData Client API to read this response, we're using `TestHttpClientRequestMessage`, which means we might as well have used the plain `HttpClient`, again this makes me wonder why this has to be an E2E test since it doesn't seem like any relevant OData client behaviour is being verified here. It seems like what we're really verifying is whether AspNetCoreOData supports a custom converter and uses it to process the result. Is that the case?

Is it that it would be difficult to test this in AspNetCoreOData and it's easier to test in an E2E test?

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PayloadValueConverterTests/PayloadValueConverterTests.cs
habbes: So what we're testing is the request, not the response? The PayloadConverter is converting the request from the client?

ElizabethOkerio: It's the response.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PayloadValueConverterTests/PayloadValueConverterTests.cs
ElizabethOkerio: The `ODataPayloadValueConverter` that is being overriden is a class in `Microsoft.OData.Core` but I get what you're saying.

test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/PayloadValueConverterTests/PayloadValueConverterTests.cs
habbes: So if I understand this correctly, we're using the client to send some payload that includes a byte array to the server.
On the server we use a binary converter to convert the response back to the client.

In this test case we're testing the following behaviour:
1. That the client can send binary data to the server (Is this the only place where we have a test to verify that the client can send binary data to the server(
2. The server can correctly deserialize a payload that contains a byte array property (is this an explicit goal of these tests)
3. That the server invokes the custom payload converter to transform the binary payload during response serialization

Is this correct?

If so, here was the source of my misunderstanding:
- Scenario 3. can be tested exclusively server side, in fact we are not even using OData client to read the response in the test. If we already have test PayloadConverter unit tests and tests in AspNetCoreOData verifying that the registered payload converters are called, then this test could be redundant.
- Scenario 2 could be tested in AspNetCoreOData
- These are ODL E2E tests, so I think the focus should be testing E2E behaviour involving OData core and Client. We should not have AspNetCoreOData behaviour that are tested here but not tested in AspNetCoreOData (since AspNetCoreOData is built and shipped independently)

So it seems that Scenario 1 is the only part that is actually testing OData client behaviour. So I have is, would that be easily testable with less code? I guess the concern is that the easiest approach to test scenario 1 would be to send to an actual test server, and to verify that it was sent, we would need to read the response. Do we need the payload converter in order to send the response? if not then maybe we could get rid of that part (assuming the PayloadConverter is tested elsewhere), if we do need it to send the response, then we can keep the test as is.

ElizabethOkerio: resolved offline

PR #2711, opened 2023-07-26
Support $filter=dynamicProperty in ('')

KenitoInc:
### Issues

*This pull request fixes #2539 .*

### Description

*Briefly describe the changes of this pull request.*

### Checklist (Uncheck if it is not completed)

- [ ] *Test cases added*
- [ ] *Build and test with one-click build and test script passed*

### Additional work necessary

*If documentation update is needed, please add "Docs Needed" label to the issue and provide details about the required document change in the issue.*

PR #3190, opened 2025-02-28
Fix $filter in (collectionofdynamicprimitivevalues)

### LINKED ISSUE #2875: No longer able to parse IN filter expressions against nested properties of dynamic objects
We use OData filters in our web apis to power queries against a CosmosDB database.
We have built a custom parser that translates the OData syntax (AST) to CosmosDB SQL syntax.
This is how we do it:
image

OData .NET v7.18 broke my unit tests and I believe this [PR](https://github.com/OData/odata.net/pull/2711) might have something to do with it.
Our main entity defines a dynamic complex type property called metadata (open type, essentially just a dictionary of random KeyValue pairs) and IN expressions against integer nested properties e.g. $filter="**metadata/intProp in (1,2)**" used to work up until OData .NET v7.17.

### Assemblies affected

OData.Core .NET >= v7.18

### Reproduce steps

https://gist.github.com/atomaras/335bda36b3c03faeddc4b66f2483386f

### Expected result

Parsing should respond with a valid Node just like it did with previous versions.

### Actual result

It started failing with the following error:
image

### Additional detail

### LINKED ISSUE #3098: ODL requires collection items on the RHS of "in" operator to be strings
ODL requires collection items on the RHS of in operator to be strings.
The following filter expression where `DynamicPrimitiveCollectionProperty` is a dynamic property is not parsed successfully.
```
$filter=DynamicPrimitiveCollectionProperty in (19,17)
```

### Assemblies affected

- Microsoft.OData.Core 8.x
- Microsoft.OData.Core 7.x

### Reproduce steps

```csharp
var model = new EdmModel();

var customerEntityType = model.AddEntityType("NS", "Customer", baseType: null, isAbstract: false, isOpen: true);
customerEntityType.AddKeys(customerEntityType.AddStructuralProperty("Id", EdmPrimitiveTypeKind.Int32));

var entityContainer = model.AddEntityContainer("Default", "Container");

var customersEntitySet = entityContainer.AddEntitySet("Customers",
customerEntityType);

var filterClause = new ODataQueryOptionParser(
    model,
    customerEntityType,
    customersEntitySet, new Dictionary<string, string> { { "$filter", "DynamicPrimitiveCollectionProperty in (19,17)" } }).ParseFilter();
```

### Expected result

The filter expression to be parsed successfully.

### Actual result

image

### Additional detail

Works if the collection items are strings, e.g., `DynamicPrimitiveCollectionProperty in ('19','17')`

ElizabethOkerio:
### Issues

*This pull request fixes #3098, fixes #2875 *

### Description

This PR (https://github.com/OData/odata.net/pull/2711) introduced a fix to support `$filter=dynamicProperty in ('')`. However, the changes in this PR inadvertently broke the handling of `$filter=dynamicProperty in (primitive collections)`, such as `$filter=dynamicProperty in (1,2,3)`.

To address this, I've added a private method, `IsCollectionContentEmptyOrSpaces`, which checks whether a collection of `Edm.Untyped` is either empty or contains only spaces before further processing. If the collection meets this condition, the logic for ensuring a string collection gets properly formatted gets called. 

Previously, this string formatting logic was executed without verifying whether the `Edm.Untyped` collection contained empty spaces or not. As a result, it was being applied even when the collection contained primitive types (e.g., numbers) that do not require string formatting, leading to exceptions being thrown.

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -13,6 +13,7 @@ namespace Microsoft.OData.UriParser
     using System.Text;
     using Microsoft.OData.Edm;
     using Microsoft.OData.Core;
+    using System.Buffers;
 
     /// <summary>
     /// Class that knows how to bind the In operator.
@@ -439,57 +440,64 @@ namespace Microsoft.OData.UriParser
             }
 
             bool isEmptyOrHasOnlySpaces = true;
-            bool isCharinsideQuotes = false;
+            bool isCharInsideQuotes = false;
             char quoteChar = '\0';
-            Span<char> buffer = stackalloc char[content.Length];
+            char[] buffer = ArrayPool<char>.Shared.Rent(content.Length);
             int bufferIndex = 0;
 
-            for (int i = 0; i < content.Length; i++)
+            try
             {
-                char c = content[i];
-
-                if (isCharinsideQuotes)
+                for (int i = 0; i < content.Length; i++)
                 {
-                    if (c == quoteChar)
-                    {
-                        isCharinsideQuotes = false;
-                    }
-                    buffer[bufferIndex++] = c;
-                }
-                else
-                {
-                    if (c == '"' || c == '\'')
-                    {
-                        isCharinsideQuotes = true;
-                        quoteChar = c;
-                        buffer[bufferIndex++] = c;
-                    }
-                    else if (c == ',')
-                    {
-                        string item = new string(buffer[..bufferIndex]).Trim().Trim('\'', '"');
+                    char c = content[i];
 
-                        if (!string.IsNullOrWhiteSpace(item))
+                    if (isCharInsideQuotes)
+                    {
+                        if (c == quoteChar)
                         {
-                            isEmptyOrHasOnlySpaces = false;
-                            break;
+                            isCharInsideQuotes = false;
                         }
-                        bufferIndex = 0;
+                        buffer[bufferIndex++] = c;
                     }
                     else
                     {
-                        buffer[bufferIndex++] = c;
+                        if (c == '"' || c == '\'')
+                        {
+                            isCharInsideQuotes = true;
+                            quoteChar = c;
+                            buffer[bufferIndex++] = c;
+                        }
+                        else if (c == ',')
+                        {
+                            string item = new string(buffer, 0, bufferIndex).Trim().Trim('\'', '"');
+
+                            if (!string.IsNullOrWhiteSpace(item))
+                            {
+                                isEmptyOrHasOnlySpaces = false;
+                                break;
+                            }
+                            bufferIndex = 0;
+                        }
+                        else
+                        {
+                            buffer[bufferIndex++] = c;
+                        }
+                    }
+                }
+
+                if (bufferIndex > 0)
+                {
+                    string lastItem = new string(buffer, 0, bufferIndex).Trim().Trim('\'', '"');
+
+                    if (!string.IsNullOrWhiteSpace(lastItem))
+                    {
+                        isEmptyOrHasOnlySpaces = false;
                     }
                 }
             }
-
-            if (bufferIndex > 0)
+            finally
             {
-                string lastItem = new string(buffer[..bufferIndex]).Trim().Trim('\'', '"');
-
-                if (!string.IsNullOrWhiteSpace(lastItem))
-                {
-                    isEmptyOrHasOnlySpaces = false;
-                }
+                ArrayPool<char>.Shared.Return(buffer);
             }
 
             return isEmptyOrHasOnlySpaces;
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -13,6 +13,7 @@ namespace Microsoft.OData.UriParser
     using System.Text;
     using Microsoft.OData.Edm;
     using Microsoft.OData.Core;
+    using System.Collections.Generic;
 
     /// <summary>
     /// Class that knows how to bind the In operator.
@@ -129,7 +130,7 @@ namespace Microsoft.OData.UriParser
                     Debug.Assert(expectedType.IsCollection());
                     string expectedTypeFullName = expectedType.Definition.AsElementType().FullTypeName();
 
-                    if (expectedTypeFullName.Equals("Edm.String", StringComparison.Ordinal) || expectedTypeFullName.Equals("Edm.Untyped", StringComparison.Ordinal))
+                    if (expectedTypeFullName.Equals("Edm.String", StringComparison.Ordinal) || (expectedTypeFullName.Equals("Edm.Untyped", StringComparison.Ordinal) && IsCollectionContentEmptyOrSpaces(bracketLiteralText)))
                     {
                         // For collection of strings, need to convert single-quoted string to double-quoted string,
                         // and also, per ABNF, a single quote within a string literal is "encoded" as two consecutive single quotes in either
@@ -422,6 +423,77 @@ namespace Microsoft.OData.UriParser
             }
 
             return "[" + String.Join(",", items) + "]";
+        }
+
+        private static bool IsCollectionContentEmptyOrSpaces(string bracketLiteralText)
+        {
+            if (string.IsNullOrWhiteSpace(bracketLiteralText) || bracketLiteralText.Length < 2)
+            {
+                return true;
+            }
+
+            string content = bracketLiteralText[1..^1].Trim();
+
+            if (string.IsNullOrWhiteSpace(content))
+            {
+                return true;
+            }
+
+            bool isEmptyOrHasOnlySpaces = true;
+            bool isCharinsideQuotes = false;
+            char quoteChar = '\0';
+            Span<char> buffer = stackalloc char[content.Length];
+            int bufferIndex = 0;
+
+            for (int i = 0; i < content.Length; i++)
+            {
+                char c = content[i];
+
+                if (isCharinsideQuotes)
+                {
+                    if (c == quoteChar)
+                    {
+                        isCharinsideQuotes = false;
+                    }
+                    buffer[bufferIndex++] = c;
+                }
+                else
+                {
+                    if (c == '"' || c == '\'')
+                    {
+                        isCharinsideQuotes = true;
+                        quoteChar = c;
+                        buffer[bufferIndex++] = c;
+                    }
+                    else if (c == ',')
+                    {
+                        string item = new string(buffer[..bufferIndex]).Trim().Trim('\'', '"');
+
+                        if (!string.IsNullOrWhiteSpace(item))
+                        {
+                            isEmptyOrHasOnlySpaces = false;
+                            break;
+                        }
+                        bufferIndex = 0;
+                    }
+                    else
+                    {
+                        buffer[bufferIndex++] = c;
+                    }
+                }
+            }
+
+            if (bufferIndex > 0)
+            {
+                string lastItem = new string(buffer[..bufferIndex]).Trim().Trim('\'', '"');
+
+                if (!string.IsNullOrWhiteSpace(lastItem))
+                {
+                    isEmptyOrHasOnlySpaces = false;
+                }
+            }
+
+            return isEmptyOrHasOnlySpaces;
         }
     }
 }
--- a/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
+++ b/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
@@ -7,13 +7,13 @@
 namespace Microsoft.OData.UriParser
 {
     using System;
+    using System.Buffers;
     using System.Diagnostics;
     using System.Globalization;
     using System.Linq;
     using System.Text;
-    using Microsoft.OData.Edm;
     using Microsoft.OData.Core;
-    using System.Buffers;
+    using Microsoft.OData.Edm;
 
     /// <summary>
     /// Class that knows how to bind the In operator.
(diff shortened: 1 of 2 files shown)


ElizabethOkerio:
Run pipeline

test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -2130,6 +2130,22 @@ public void FilterWithInOperationWithParensCollection()
             Assert.Equal("(1,2,3)", Assert.IsType<CollectionConstantNode>(inNode.Right).LiteralText);
         }
 
+        [Theory]
+        [InlineData("TestInt", "(1,2,3)")]
+        [InlineData("TestDouble", "(1.1,2.2,3.3)")]
+        [InlineData("TestBool", "(true,false)")]
+        [InlineData("TestString", "('a','b','c')")]
+        [InlineData("TestDecimal", "(1.1,2.2,3.3)")]
+        public void FilterWithInOperationWithParensCollectionConsistingOfDynamicPrimitiveProperties(string propertyName, string values)
```
WanjohiSammy, line 2139: What if you use square brackets '[]' instead of parentheses

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -423,5 +423,76 @@ private static string NormalizeDateTimeCollectionItems(string bracketLiteralText
 
             return "[" + String.Join(",", items) + "]";
         }
+
+        private static bool IsCollectionContentEmptyOrSpaces(string bracketLiteralText)
```
WanjohiSammy, line 427: Will this be able to handle scenarios where some items are quoted, and others are not in the same collection:
- ('1',2,3)
- ('true', true, false)
- ('1.23', 0.4, 1.0)

test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
marabooy: Not entirely sure if square brackets are used but I could be wrong the 4.01 [specification ](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part1-protocol.html#_Toc31358949:~:text=Address/City%20in%20(%27Redmond%27%2C%20%27London%27)) but I also think we may not have support for [it](https://learn.microsoft.com/en-us/odata/odatalib/in-operator)

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -423,5 +423,76 @@ private static string NormalizeDateTimeCollectionItems(string bracketLiteralText
...
+
+            if (string.IsNullOrWhiteSpace(content))
+            {
+                return true;
+            }
+
+            bool isEmptyOrHasOnlySpaces = true;
+            bool isCharinsideQuotes = false;
+            char quoteChar = '\0';
+            Span<char> buffer = stackalloc char[content.Length];
```
marabooy, line 444: Do we know the bounds for content? If it is user allocated how will this work to ensure that we will not incur a Stack overflow exception.

ElizabethOkerio: The bounds for content is the length of the content.. The user cannot set if. We get it from whatever the user passes in brackets..

ElizabethOkerio: Or we throw an exception if an upper bound is exceeded.

test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
WanjohiSammy: We do support square brackets: https://github.com/OData/odata.net/pull/3122

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
ElizabethOkerio: This method only checks to see if a collection has empty spaces or no spaces at all. If it gets a value it will return false. and the code execution will continue as usual. The first scenario is supported. The second scenario we do not support even for typed properties.

ElizabethOkerio: Maybe we can add that to our backlog

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
ElizabethOkerio: Updated this to use an ArrayPool instead.

test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
ElizabethOkerio: Added tests to show this works.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -423,5 +424,83 @@ private static string NormalizeDateTimeCollectionItems(string bracketLiteralText
 
             return "[" + String.Join(",", items) + "]";
         }
+
+        private static bool IsCollectionContentEmptyOrSpaces(string bracketLiteralText)
```
gathogojr, line 428:
```suggestion
        private static bool IsCollectionEmptyOrWhiteSpace(string bracketLiteralText)
```

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -423,5 +424,83 @@ private static string NormalizeDateTimeCollectionItems(string bracketLiteralText
 
             return "[" + String.Join(",", items) + "]";
         }
+
+        private static bool IsCollectionContentEmptyOrSpaces(string bracketLiteralText)
+        {
+            if (string.IsNullOrWhiteSpace(bracketLiteralText) || bracketLiteralText.Length < 2)
```
gathogojr, line 430: Will validation that the first non-whitespace character is `(` or `[` and the last non-whitespace character is `)` or `]` have been done prior to calling this method?

Asking because this method would return true for the input `(1`

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -423,5 +424,83 @@ private static string NormalizeDateTimeCollectionItems(string bracketLiteralText
...
+        private static bool IsCollectionContentEmptyOrSpaces(string bracketLiteralText)
+        {
+            if (string.IsNullOrWhiteSpace(bracketLiteralText) || bracketLiteralText.Length < 2)
+            {
+                return true;
+            }
+
+            string content = bracketLiteralText[1..^1].Trim();
+
+            if (string.IsNullOrWhiteSpace(content))
```
gathogojr, line 437: Depending on the answer to this https://github.com/OData/odata.net/pull/3190/files#r1978721142, is this scenario possible?

I think that the first thing we should check is `string.IsNullOrWhiteSpace` at the beginning of this method

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
```diff
@@ -423,5 +424,83 @@ private static string NormalizeDateTimeCollectionItems(string bracketLiteralText
...
+            {
+                return true;
+            }
+
+            bool isEmptyOrHasOnlySpaces = true;
```
gathogojr, line 442:
```suggestion
            bool isEmptyOrWhiteSpace = true;
```

test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
```diff
@@ -2130,6 +2130,27 @@ public void FilterWithInOperationWithParensCollection()
...
+        {
+            string filterExpression = $"{propertyName} in {values}";
+            FilterClause filter = ParseFilter(filterExpression, HardCodedTestModel.TestModel, HardCodedTestModel.GetOpenEmployeeType());
+
+            var inNode = Assert.IsType<InNode>(filter.Expression);
+            Assert.IsType<SingleValueOpenPropertyAccessNode>(inNode.Left);
+            Assert.Equal(values, Assert.IsType<CollectionConstantNode>(inNode.Right).LiteralText);
+        }
+
```
gathogojr, line 2153: Please add tests for empty parentheses/brackets scenarios - `()`, `[]`. Also add tests for the same with leading and trailing whitespaces

gathogojr: Ensure that every statement in `IsCollectionContentEmptyOrSpaces` is covered by some test

WanjohiSammy: Good point @gathogojr . We already have similar tests. I'm not sure if it's necessary to repeat them:
- [FilterWithInOperationWithEmptyCollection](https://github.com/OData/odata.net/blob/main/test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs#L2532)
- [DollarFilter_WithCollectionWithEmptyString_ExecutesSuccessfully](https://github.com/OData/odata.net/blob/main/test/EndToEndTests/Tests/Client/Microsoft.OData.Client.E2E.Tests/ClientTests/Tests/ClientQueryTests.cs#L89)
- [FilterWithInExpressionContainingEscapedSingleQuotes](https://github.com/OData/odata.net/blob/main/test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs#L2295)
- [FilterWithInOperationWithEmptyString](https://github.com/OData/odata.net/blob/main/test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs#L2547)
- [FilterWithInOperationWithEmptyStringInSquareBrackets](https://github.com/OData/odata.net/blob/main/test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs#L2565)
- [FilterWithInOperationWithWhitespace](https://github.com/OData/odata.net/blob/main/test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs#L2583)
- [FilterWithInOperationWithWhitespaceInSquareBrackets](https://github.com/OData/odata.net/blob/main/test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs#L2603)
- etc

@ElizabethOkerio Maybe check which scenarios are not added and add them.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
ElizabethOkerio: This is handled somewhere else. Where we have unbalanced brackets or an incorrect format.

ElizabethOkerio: Maybe the check here is not necessary.

test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
ElizabethOkerio: Most of these scenarios are already covered by existing tests.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
ElizabethOkerio: There is a check for that at the beginning of the method. This is after removing the first and last brackets and trimmed spaces before and after content.

src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs
ElizabethOkerio: remove the check.

test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs
ElizabethOkerio: https://github.com/OData/odata.net/blob/e50b8ec5cd74438d45bd0baced6f2d1aa4331c5a/test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs#L3096C9-L3096C16 

https://github.com/OData/odata.net/blob/e50b8ec5cd74438d45bd0baced6f2d1aa4331c5a/test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriParser/FilterAndOrderByFunctionalTests.cs#L3077

gathogojr (approved):

WanjohiSammy (approved):

habbes:
I'm not opposed to this PR, but the way handle collection literals seems like a hack and all these PRs we make to patch edge cases are band-aids that are not sustainable in the long term. We should not be hacking the collection string to massage it into a JSON array so that we can parse it and then do special checks for edge cases. I believe the ideal, sustainable solution would be to handle the collection property from the lexer stage. By the time we get to the parser, we have an array start token, and array end token, and we know all the elements in between, the elements would have been properly tokenized based on the tokenizer rules for these literals.

xuzhg (approved):

PR #3192, opened 2025-03-03
Fix endless loop issue when Schema tag has no correct xmlns value.

### LINKED ISSUE #3188: Memory Leak in CsdlReader.TryParse Due to missing xmlns=”http://docs.oasis-open.org/odata/ns/edm”
The `CsdlReader.TryParse` does not throw an exception when parsing CSDL document with missing `xmlns=”http://docs.oasis-open.org/odata/ns/edm”`. Instead, the memory continuously increases until it reaches the maximum limit, resulting in significant memory usage. This condition persists until the operation is either forcibly stopped, canceled, or time out. 

### Assemblies affected

Microsoft.OData.Edm 7.x, 8.x

### Reproduce steps

- Clone the repo: https://github.com/WanjohiSammy/MemoryLeakInCSDLReaderTryParse.git
- Run the application:
   - Remove `xmlns="http://docs.oasis-open.org/odata/ns/edm"` from the `Schema` tag and run the application
   ```xml
   <edmx:Edmx xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx" Version="4.0">
    <edmx:DataServices>
        <Schema Namespace="NS"> // without xmlns="http://docs.oasis-open.org/odata/ns/edm"
            <EntityType Name="Customer">
                <Key>
                    <PropertyRef Name="Id" />
                </Key>
                <Property Name="Id" Type="Edm.Int32" Nullable="false" />
                <Property Name="Name" Type="Edm.String" />
            </EntityType>
            <EntityContainer Name="Container">
                <EntitySet Name="Customers" EntityType="NS.Customer" />
            </EntityContainer>
        </Schema>
    </edmx:DataServices>
  </edmx:Edmx>
   ```
   - Note how the process memory is growing using `Visual Studio Diagnostics Tools`
   
   Image

   - with `<Schema xmlns=""http://docs.oasis-open.org/odata/ns/edm"" Namespace=""NS"">`, the CSDL document is parsed successfully
   ```xml
   <edmx:Edmx xmlns:edmx="http://docs.oasis-open.org/odata/ns/edmx" Version="4.0">
    <edmx:DataServices>
        <Schema xmlns="http://docs.oasis-open.org/odata/ns/edm" Namespace="NS"> // with xmlns
            <EntityType Name="Customer">
                <Key>
                    <PropertyRef Name="Id" />
                </Key>
                <Property Name="Id" Type="Edm.Int32" Nullable="false" />
                <Property Name="Name" Type="Edm.String" />
            </EntityType>
            <EntityContainer Name="Container">
                <EntitySet Name="Customers" EntityType="NS.Customer" />
            </EntityContainer>
        </Schema>
    </edmx:DataServices>
  </edmx:Edmx>
   ```
### Expected result

Schema without `xmlns="http://docs.oasis-open.org/odata/ns/edm"` should throw an exception or should be captured as part of `EdmError` instead of running to forever and increases memory usage.

### Actual result

The memory continuously increases until it reaches the maximum limit, resulting in significant memory usage. This condition persists until the operation is either forcibly stopped, canceled, or time out. 

### Additional detail

WanjohiSammy:
### Issues

*This pull request fixes #3188.*

### Description

This PR addresses an issue where an endless loop occurs when the `Schema` tag does not have `xmlns` attribute or correct `xmlns` value. The fix involves adding `reader.Skip()` before the return statement to ensure the reader skips to the end tag of the current element. 

Adding `reader.Read()` will move the reader to the next element and results to elements with opening and closing tags that do not match.

- [x] *Test cases added*

### Additional work necessary


test/UnitTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
```diff
@@ -2262,6 +2262,90 @@ public void ReadingScaleAndSridValue_VaribaleValue_NotInitializedOnRespectiveEdm
...
         }
 
+        [Theory]
+        [InlineData("", "The root element has no namespace.")]
+        [InlineData(" ", "The root element has no namespace.")]
+        [InlineData("xmlns=\"\"", "The root element has no namespace.")]
+        [InlineData("xmlns=\"http://docs.oasis-open.org/odata/ns/edm-invalid\"", "The namespace 'http://docs.oasis-open.org/odata/ns/edm-invalid' is invalid.")]
+        [InlineData("xmlns=\"''\"", "The namespace '''' is invalid.")]
+        [InlineData("xmlns=\"null\"", "The namespace 'null' is invalid.")]
+        [InlineData(null, "The root element has no namespace.")]
```
gathogojr, line 2272: Would `<Schema Namespace="NS" Alias="self" null>` be valid XML?

gathogojr (approved):
LGTM

test/UnitTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
WanjohiSammy: No, this will return an error:

```log
The root element has no namespace. The root element is expected to belong to one of the following namespaces ...
```

xuzhg (approved):

src/Microsoft.OData.Edm/Csdl/Parsing/Common/XmlDocumentParser.cs
```diff
@@ -148,6 +148,7 @@ internal void ParseDocumentElement()
             else
             {
                 this.ReportUnexpectedRootNamespace(this.reader.LocalName, this.DocumentNamespace, expectedNamespaces);
+                this.reader.Skip();
```
xuzhg, line 151: Can we just tolerate to allow the missing and append it by default by use and only report the wrong namespace used?

xuzhg (approved):

src/Microsoft.OData.Edm/Csdl/Parsing/Common/XmlDocumentParser.cs
WanjohiSammy: I believe this is a good point. But we will end up with some other issues. For example, [void Parse()](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Edm/Csdl/Parsing/Common/XmlDocumentParser.cs#L199) expects that there are errors encountered before start processing the nodes.

PR #3193, opened 2025-03-04
URI Parser rewrite for performance and stability

habbes:
### Issues

### Description

- [ ] *Test cases added*

### Additional work necessary

### DIFF
--- a/docs/uri-parser-rewriter-proposal.md
+++ b/docs/uri-parser-rewriter-proposal.md
@@ -1,0 +1,97 @@
+# OData URI Parser Rewrite for high performance and reliability
+
+This is a proposal to rewrite the OData URI parser stack from the ground up based on a different architecture. The primary motivation for this proposal is performance,
+but it also aims to make improvements to usability and developer experience.
+
+## Motivations
+
+### Performance
+
+This proposal is part of larger effort to turn OData into a high-performance library. The URI parser is a key component of the OData experience as parsing the request URL is the first
+step to handling and OData request. We have made efforts aimed at improving performance, the most recent being the [work by `xuhg` to use `ReadOnlySpan<char>` to cut down
+string allocations](https://github.com/OData/odata.net/pull/3077). While we have seen improvements in such efforts, we're limited by the current architecture and public-facing APIs.
+If we entertain a breaking change, we can change the architecture and redesign the parser for better performance and low resource use.
+
+The aim of this effort to reach at least 5x improvement in performance and reduction in allocations on the core URI parser on benchmarks that demonstrate end-to-end semantic parsing of common URL patterns.
+
+### Usability
+
+The OData URI has two main parts, the path segments and query options. There different parsers for the path and the query options.
+In general, each component has two parsing phases: syntactic parse and semantic parse. The syntactic parse is supposed to return a tree or sequence of nodes that represent the structure without binding model information. The semantic pass adds type information and validation from the model. However, the separation between syntax and semantic parse isn't cleanly implemented in all cases:
+
+- The `SelectExpandSyntactiParser` which performs syntactic parsing of the select and expand query options is internal, meaning that end users cannot access it. It also has a dependency on `IEdmStructuredType`, meaning the model is required to use it. Customers have expressed interest in parsing URLs without the model.
+- The path parser with parses the path segments simply splits the string on `/` slash symbols. As a result it doesn't properly handle cases where the keys in the path may contain slashes. [A comment in the code](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/Parsers/UriPathParser.cs#L55) suggests that this is a known bug.
+- The `UriQueryOptionsParser` with does a lexical parse of some query options like `$filter` reads sequences delimited by `(` and `)` as single string. This means that items of literal collections cannot be individually consumed in the syntactic phase. These are properly parsed later in the semantic binder (for example, `FilterBinder`) where a model is required. This binding is also not ideal. For example, when parsing an `in` or `has` expression, the collection [literal is parsed by replacing the delimiting `(` and `)` with `[` and `]` and then calling the JsonReader to parse the array](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/UriParser/Binders/InBinder.cs#L119). This "hack" is the cause of several bugs and incompalities related to the `in` operator, such as:
+  - [](https://github.com/OData/odata.net/pull/3190)
+  - https://github.com/OData/odata.net/issues/2875
+  - https://github.com/OData/odata.net/issues/3098
+  - https://github.com/OData/odata.net/pull/2711
+  - [Source comment acknowleging flaws of this approach](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs#L677).
+
+The rewrite provides an opportunity to parse each expression properly from the lexical tokens up to the semantic nodes.
+
+## High-level architectural changes
+
+The key changes proposed:
+
+- Avoid premature string allocations. Reference segments in the source string using `ReadOnlySpan<char>` until the user explicitly asks for a `string`.
+- Store parsed nodes in array of structs instead of graph of objects.
+- Frugal use of space, for example pack data in bit flags where applicable instead of separate bool fields. This helps keep the structs small and less expensive to move around.
+
+Using an array of node structs is a common approach for memory-efficient parsers like `System.Text.Json`.
+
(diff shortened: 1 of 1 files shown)


docs/uri-parser-rewriter-proposal.md
```diff
@@ -0,0 +1,272 @@
...
+```csharp
+public struct QueryNode
+{
+    private int _index { get; set; }
+    private ExpressionParser _parser { get; set; }
+    
+    public ExpressionNodeKind Kind => _parser[_index].Kind;
+    public int GetIntValue()
+    {
+        return int.TryParse(_parser[_index].Range.GetValue(_parser.Source));
```
corranrogue9, line 164: I'm assuming this is meant to be `_parser._nodes[_index]`?

docs/uri-parser-rewriter-proposal.md
```diff
@@ -0,0 +1,272 @@
...
+![Array layout of the queyr](./uri-parser-nested-in-query-array.png)
+
+The downside of this approach is that there could be multiple nesting levels between two siblings of the same parent. That would require multiple hops of to skip over from one child to its next sibling. The number of hops is proportional to the level of nesting. If we have highly nested collections, this could noticeably slow down the iteration or indexing beyond O(n). If we anticipate highly nested collections, we could reduce the number of hops to one per sibling by storing more metadata about each child and level (TODO: demonstrate this).
+
+How does the end user process this tree? Since we're using structs, we can't provide a traditional visitor that's based on inheritance. The basic approach is for the user to simply traverse the tree using the APIs on the `QueryNode` struct, like they would any other DOM-like structure. But this may be tedious for some cases. We can propose a handler that exposes a different method based on the type of node for convenience, e.g.:
+
+```csharp
+interface IQueryNodeHandler<T>
+{   
+    public T HandleIntNode(QueryNode intLiteralNode);
```
corranrogue9, line 221: Is there a huge perf overhead for these methods to take in more specific types? Like, it would be great if we could make this `IntQueryNode` instead

docs/uri-parser-rewriter-proposal.md
```diff
@@ -0,0 +1,272 @@
...
+        return QueryNode(this, root);
+    }
+}
+```
+
+The `ExpressioNode` is an internal representation of any kind of node in the "tree". This struct is not exposed publicly.
+Instead, we return a `QueryNode` struct which has the following structure:
+
+```csharp
+public struct QueryNode
```
corranrogue9, line 156: I don't see what prevents this from being a `ref struct` instead

corranrogue9: And I think we don't actually need the setters and we can make this `readonly` right?

docs/uri-parser-rewriter-proposal.md
```diff
@@ -0,0 +1,272 @@
...
+  - [Source comment acknowleging flaws of this approach](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/Uri/ODataUriConversionUtils.cs#L677).
+
+The rewrite provides an opportunity to parse each expression properly from the lexical tokens up to the semantic nodes.
+
+## High-level architectural changes
+
+The key changes proposed:
+
+- Avoid premature string allocations. Reference segments in the source string using `ReadOnlySpan<char>` until the user explicitly asks for a `string`.
+- Store parsed nodes in array of structs instead of graph of objects.
```
corranrogue9, line 38: I want to be clear about this aspect and the terminology that we are using. By not having a tree, this is more akin to the readers that we have. This is not really "parsing" anymore. There's nothing wrong with this, I'm more having an issue with the terminology, and the terminology is important to me because (as you note later in the document), we don't differentiate between the different kinds of notes (like the reader doesn't): the caller is expected to understand what kinds of nodes can come next and needs to be aware of this based on the context of the previous nodes.

docs/uri-parser-rewriter-proposal.md
corranrogue9: Also, we have to be very careful about this because we will need to constantly add support for `async`, `unsafe`, methods that require parameterized context, `allows ref struct`, etc.

docs/uri-parser-rewriter-proposal.md
```diff
@@ -0,0 +1,272 @@
...
+- The `LiteralToken` is used to represent, among others, parenthesized expressions (e.g. collection expressions). The expression is still in unparsed string form.
+
+This is where we propose a drastic change. Instead of nodes with direct reference to each other, we store nodes in a flat array.
+
+A node would have roughly the following structure:
+
+```csharp
+internal struct ExpressionNode
+{
+    public ExpressionNodeKind Kind { get; internal set; }
```
corranrogue9, line 127: If we use `class`es to model the enumeration instead of an `enum`, we could have a visitor on those classes which would help implement the code that generates the `ref struct` nodes that could be used in the handler that I suggested below

docs/uri-parser-rewriter-proposal.md
```diff
@@ -0,0 +1,272 @@
+# OData URI Parser design change for higher performance and reliability
```
corranrogue9, line 1: It was brought up in the meeting that we have the whole URI string in memory, so we should leverage that for performance gains instead of trying to overly generalize. Have you considered what it looks like for someone to compose an OData URI? For example, in the OData Client, we compose the URI as the caller tells us what data they are requesting. In that situation, we actually *don't* have the full URI *anywhere* in memory. Are we thinking we will have an entirely separate AST model for the client code to leverage?

habbes: The client can either consume or produce OData URI. When it consume OData URI, the URI is available in full, in memory. When it produces the URI, for example when translating LINQ expressions to an OData request, it generates the URI on the fly using a StringBuilder, the output of the translation will be string that's in memory, but even if it wasn't, this URI generation is not the role of the parser. The parser as described in this doc (and as implemented by the existing `ODataUriParser`) takes the URI as input, it does not generate it.

docs/uri-parser-rewriter-proposal.md
habbes: Based on your suggestions, I will investigate the cost of adding more specific types on top `QueryNode` for the different methods, e.g. `Handle(IntQueryNode)`

Where `IntQueryNode` could be defined as 

```c#
struct IntQueryNode
{
     private QueryNode _node;
     public int GetIntValue() => _node.GetIntValue();
}
```

Maybe it could be optimized to a ref struct itself or it could reference the expression parser directly if such optimizations prove meaningful and practical.

docs/uri-parser-rewriter-proposal.md
corranrogue9: > We could compare both and see what the perf overhead is for each. But I assume the enum would have less overhead, it's smaller value, which could make it easier to pack the struct more densely and we don't have to cast.

The use of the visitor is actually what stops the need to cast, but your overall point certainly stands: this *will* be slower than an enum. I suppose I'm suggesting that between this change and the change for `ihandler` to take in more specific types, we will lose out *slightly* on performance but get 90% of the type safety benefits of a completely strongly-typed AST (for example, `kind` will now be expressed in a way that ensure we notice backwards compatibility issues, and `ihandler` would be expressed in a way that ensures callers don't have to do their own assertions). I understand if you don't take either suggestion, but they seems a very reasonable middle ground to me.

> Another I opted for an enum is that we could pack additional data in the enum....

I don't know enough about it and can't get a good test to repro the perf benefits. Is there a benefit to doing this instead of using 2 `short` enums?

docs/uri-parser-rewriter-proposal.md
corranrogue9: I guess my point is about the contract that the customer interfaces with. For OData Client, the customer gets a `string` output, and from the parser, the customer has a `string` input, so the "contract" that we are publishing to customers is that an OData URI is a `string`. I'll try to illustrate with something a bit contrived but that I think is reasonable:

A customer is using OData Client to generate a request. They then realize that they want to add an additional query option that the client doesn't support yet. So, they get the URI string from the client. At this point, they can't interact with our strongly-type URI code. If they parse the string, they will end up with types that can't be used to produce a new URI. So, the customer now has to either implement their own minimalized parser, or they will do simple string manipulation that is very error prone.

If instead the parser produced a type that can be added to (even something similar to how immutable collections work), then the customer could take the string from the client and parse it into something strongly typed, make their adjustments, and get a string back by walking the "AST". But if all of this is possible, then there's really no reason that the *client* can't just give the customer the same type that the parser produces. If we do *this*, the "contract" that our customers see is "ah, this AST-looking this is the type of OData URIs".

PR #3197, opened 2025-03-10
adding the stack-based linked list

corranrogue9:


odata.tests/StackBasedLinkedList.cs
```diff
@@ -0,0 +1,861 @@
...
+
+            public static implicit operator T*(BetterSpan<T> betterSpan)
+            {
+                return betterSpan.data;
+            }
+        }
+
+        public static class LinkedListNodeExtensions
+        {
+            public static unsafe LinkedListNode<T> Append<T>(this LinkedListNode<T> source, T value, Span<byte> previousMemory)
```
corranrogue9, line 763: I don't think this is unsafe anymore

PR #3225, opened 2025-04-07
adding uri parser tests for the newly supported average overloads

### LINKED ISSUE #3165: UriParser - Aggregation method Average doesn't support Int16

corranrogue9:
### Issues

*This pull request fixes #3165 .*

### Description

Adding additional unit tests for #3168. These are `UriParser` tests to get a more scenario-based coverage of the functionality.

- [x] *Test cases added*


gathogojr:
@corranrogue9 I believe the contributor for #3168 added a test for `Int16`
- [FullyQualifiedNamespacePerson.AddStructuralProperty("SecondFavoriteNumber", EdmCoreModel.Instance.GetInt16(false));](https://github.com/OData/odata.net/pull/3168/files#diff-9be973737b29aeb97734019e655d16bb24a0770b3173182f9c6ffe434c155bc8R164)
- [BindApplyWithAverageWithInt16PropertyShouldReturnApplyClause](https://github.com/OData/odata.net/pull/3168/files#diff-26fd34e25fa343cc32841bc2dc9d77e21e742e13a630b420f7df74eb8c7e033bR863)

Can you confirm if the scenario that your test in this pull request verifies is different?

corranrogue9:
> @corranrogue9 I believe the contributor for #3168 added a test for `Int16`
> 
> * [FullyQualifiedNamespacePerson.AddStructuralProperty("SecondFavoriteNumber", EdmCoreModel.Instance.GetInt16(false));](https://github.com/OData/odata.net/pull/3168/files#diff-9be973737b29aeb97734019e655d16bb24a0770b3173182f9c6ffe434c155bc8R164)
> * [BindApplyWithAverageWithInt16PropertyShouldReturnApplyClause](https://github.com/OData/odata.net/pull/3168/files#diff-26fd34e25fa343cc32841bc2dc9d77e21e742e13a630b420f7df74eb8c7e033bR863)
> 
> Can you confirm if the scenario that your test in this pull request verifies is different?

@gathogojr those tests don't cover the actual parsing that needs to be supported (which was their original scenario, from what I understand). These tests add the coverage that parsing works end-to-end.

gathogojr (approved):

WanjohiSammy (approved):

PR #3235, opened 2025-04-23
#2977 Support more kinds of relative context URIs

### LINKED ISSUE #2977: ODataLib Client fails to deserialize payload with @odata.context relative URL starting with ../
I have a project where I want to integrate a SAP CAPs with a .NET OData Client.
Querying and write operations called from the data service context like `DataServiceContext.AddToSomeEntity(newEntityInstance)` are not a problem. But if I want to insert related entities I got an error while the response is parsed, because the context url looks like `@odata.context": "../$metadata#SomeEntity`.

And the code changes for the issue [ODataLib Client fails to deserialize payload with @odata.context relative URL](https://github.com/OData/odata.net/issues/1208) seems only to fix the case if url starts with `$metadata`.

Used `Microsoft.OData.Client` version `7.21.1`.

cmackethan-pg:
### Issues

This pull request fixes #2977. It is similar to PRs #1208 and #2132

### Description

SAP's OData V4 APIs will return context URI annotations like `"@odata.context": "../$metadata#Customers(1)/Name"` and `"@odata.context": "../../$metadata#Customers(1)/Name"` for certain requests. These are currently unsupported and result in an exception.

The code changes in this PR cover these scenarios.

- [x] *Test cases added*

`build.cmd` did not work for me... I think I'm missing some build tools, or maybe it's because I don't have VS 2019 Enterprise. However, all tests that I ran in the test explorer passed (21,066 passed, 8 were skipped).

### DIFF
--- a/src/Microsoft.OData.Core/Json/ODataJsonDeserializer.cs
+++ b/src/Microsoft.OData.Core/Json/ODataJsonDeserializer.cs
@@ -12,6 +12,7 @@ namespace Microsoft.OData.Json
     using System.Diagnostics.CodeAnalysis;
     using System.Globalization;
     using System.Linq;
+    using System.Text.RegularExpressions;
     using System.Threading.Tasks;
     using Microsoft.OData;
     using Microsoft.OData.Core;
@@ -24,6 +25,9 @@ namespace Microsoft.OData.Json
     /// </summary>
     internal abstract class ODataJsonDeserializer : ODataDeserializer
     {
+        /// <summary>Regex used to strip unwanted information from the start of a relative contextUriAnnotation value</summary>
+        internal static readonly Regex RelativeContextUriAnnotationRegex = PlatformHelper.CreateCompiled(@"^(\.\.\/)*\$metadata#", RegexOptions.IgnoreCase);
+
         /// <summary>The Json input context to use for reading.</summary>
         private readonly ODataJsonInputContext jsonInputContext;
 
@@ -278,12 +282,12 @@ namespace Microsoft.OData.Json
             {
                 // If the Base uri string is http://odata.org/test
                 // The MetadataDocumentUri will be http://odata.org/test/$metadata
-                // If the contextUriAnnotation value is Customers(1)/Name or $metadata#Customers(1)/Name
+                // If the contextUriAnnotation value is Customers(1)/Name, $metadata#Customers(1)/Name, or ../$metadata#Customers(1)/Name
                 // The generated context uri will be http://odata.org/test/$metadata#Customers(1)/Name
                 ODataUri oDataUri = new ODataUri() { ServiceRoot = this.BaseUri };
-                contextUriAnnotationValue = contextUriAnnotationValue.StartsWith("$metadata#", StringComparison.OrdinalIgnoreCase)
-                    ? this.BaseUri + contextUriAnnotationValue
-                    : oDataUri.MetadataDocumentUri.ToString() + ODataConstants.ContextUriFragmentIndicator + contextUriAnnotationValue;
+                contextUriAnnotationValue = RelativeContextUriAnnotationRegex.Replace(contextUriAnnotationValue, string.Empty);
+                contextUriAnnotationValue = oDataUri.MetadataDocumentUri.ToString()
+                    + ODataConstants.ContextUriFragmentIndicator + contextUriAnnotationValue;
             }
 
             ODataJsonContextUriParseResult parseResult = null;
--- a/src/Microsoft.OData.Core/Json/ODataJsonDeserializer.cs
+++ b/src/Microsoft.OData.Core/Json/ODataJsonDeserializer.cs
@@ -7,17 +7,16 @@
 namespace Microsoft.OData.Json
 {
     #region Namespaces
+    using Microsoft.OData;
+    using Microsoft.OData.Core;
+    using Microsoft.OData.Edm;
+    using Microsoft.OData.Evaluation;
     using System;
     using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
     using System.Globalization;
     using System.Linq;
-    using System.Text.RegularExpressions;
     using System.Threading.Tasks;
-    using Microsoft.OData;
-    using Microsoft.OData.Core;
-    using Microsoft.OData.Edm;
-    using Microsoft.OData.Evaluation;
     #endregion Namespaces
 
     /// <summary>
@@ -25,9 +24,6 @@ namespace Microsoft.OData.Json
     /// </summary>
     internal abstract class ODataJsonDeserializer : ODataDeserializer
     {
-        /// <summary>Regex used to strip unwanted information from the start of a relative contextUriAnnotation value</summary>
-        internal static readonly Regex RelativeContextUriAnnotationRegex = PlatformHelper.CreateCompiled(@"^(\.\.\/)*\$metadata#", RegexOptions.IgnoreCase);
-
         /// <summary>The Json input context to use for reading.</summary>
         private readonly ODataJsonInputContext jsonInputContext;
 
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Core/Json/ODataJsonDeserializer.cs
```diff
@@ -278,12 +282,12 @@ internal void ReadPayloadStart(
...
                 // If the Base uri string is http://odata.org/test
                 // The MetadataDocumentUri will be http://odata.org/test/$metadata
-                // If the contextUriAnnotation value is Customers(1)/Name or $metadata#Customers(1)/Name
+                // If the contextUriAnnotation value is Customers(1)/Name, $metadata#Customers(1)/Name, or ../$metadata#Customers(1)/Name
                 // The generated context uri will be http://odata.org/test/$metadata#Customers(1)/Name
                 ODataUri oDataUri = new ODataUri() { ServiceRoot = this.BaseUri };
-                contextUriAnnotationValue = contextUriAnnotationValue.StartsWith("$metadata#", StringComparison.OrdinalIgnoreCase)
-                    ? this.BaseUri + contextUriAnnotationValue
-                    : oDataUri.MetadataDocumentUri.ToString() + ODataConstants.ContextUriFragmentIndicator + contextUriAnnotationValue;
+                contextUriAnnotationValue = RelativeContextUriAnnotationRegex.Replace(contextUriAnnotationValue, string.Empty);
```
xuzhg, line 288: It looks good to me. However, I prefer to add a condition directly and not use Regex.

xuzhg (dismissed):

src/Microsoft.OData.Core/Json/ODataJsonDeserializer.cs
cmackethan-pg: Hey Sam,

Thanks for the quick review.

I would have liked to avoid the Regex, the issue is that there is no limit on the number of `../` repetitions at the start of the context URI.

Feel free to change it if there's a better way to handle it.

cmackethan-pg:
@microsoft-github-policy-service agree company="Prometheus Group"

src/Microsoft.OData.Core/Json/ODataJsonDeserializer.cs
WanjohiSammy: Are there possibilities of the following context Uris and how do we handle them:
- /$metadata#Customers(1)/Name
- /Customers(1)/Name

WanjohiSammy: Maybe remove the trailing "/" at the start of the Uri:
```suggestion
                contextUriAnnotationValue = RelativeContextUriAnnotationRegex.Replace(contextUriAnnotationValue.TrimStart('/'), string.Empty);

```

WanjohiSammy: Without Regex, you can do something like:
```cs
contextUriAnnotationValue = contextUriAnnotationValue.StartsWith("$metadata#", StringComparison.OrdinalIgnoreCase) || contextUriAnnotationValue.Contains("$metadata#", StringComparison.OrdinalIgnoreCase) 
    ? contextUriAnnotationValue.Substring(contextUriAnnotationValue.IndexOf("$metadata#") + 10) // 10 is the length of "$metadata#"
    : contextUriAnnotationValue.TrimStart('/', '.');

contextUriAnnotationValue = oDataUri.MetadataDocumentUri.ToString() + ODataConstants.ContextUriFragmentIndicator + contextUriAnnotationValue;
```

WanjohiSammy: https://github.com/OData/odata.net/pull/3235#discussion_r2154664238
This should be better in terms of performance compared to Regex

cmackethan-pg: @WanjohiSammy I removed the Regex - let me know what you think.

src/Microsoft.OData.Core/Json/ODataJsonDeserializer.cs
```diff
@@ -7,16 +7,16 @@
 namespace Microsoft.OData.Json
 {
     #region Namespaces
+    using Microsoft.OData;
+    using Microsoft.OData.Core;
+    using Microsoft.OData.Edm;
+    using Microsoft.OData.Evaluation;
```
WanjohiSammy: **nit**: `using System` should come before the `using Microsoft`.
Place `System` directives first.

cmackethan-pg: Updated it

test/UnitTests/Microsoft.OData.Core.Tests/Json/ODataJsonDeserializerTests.cs
```diff
@@ -878,6 +878,18 @@ public void TopLevelPropertyShouldReadContextUriAsRelativeUri()
 
             Assert.NotNull(property);
             Assert.Equal("http://odata.org/test/$metadata#Customers(1)/Name", deserializer.ContextUriParseResult.ContextUri.ToString());
+
```
WanjohiSammy: The test can be simplified using `Theory` with `[InlineData()]`:

```cs
[Theory]
[InlineData("{\"@odata.context\":\"Customers(1)/Name\",\"value\":\"Joe\"}")]
[InlineData("{\"@odata.context\":\"/Customers(1)/Name\",\"value\":\"Joe\"}")]
[InlineData("{\"@odata.context\":\"./Customers(1)/Name\",\"value\":\"Joe\"}")]
[InlineData("{\"@odata.context\":\"$metadata#Customers(1)/Name\",\"value\":\"Joe\"}")]
[InlineData("{\"@odata.context\":\"/$metadata#Customers(1)/Name\",\"value\":\"Joe\"}")]
[InlineData("{\"@odata.context\":\"./$metadata#Customers(1)/Name\",\"value\":\"Joe\"}")]
[InlineData("{\"@odata.context\":\"../$metadata#Customers(1)/Name\",\"value\":\"Joe\"}")]
[InlineData("{\"@odata.context\":\"../../$metadata#Customers(1)/Name\",\"value\":\"Joe\"}")]
[InlineData("{\"@odata.context\":\"../../../$metadata#Customers(1)/Name\",\"value\":\"Joe\"}")]
[InlineData("{\"@odata.context\":\"../../../../$metadata#Customers(1)/Name\",\"value\":\"Joe\"}")]
[InlineData("{\"@odata.context\":\"../../../../../$metadata#Customers(1)/Name\",\"value\":\"Joe\"}")]
[InlineData("{\"@odata.context\":\"../../../../../../$metadata#Customers(1)/Name\",\"value\":\"Joe\"}")]
[InlineData("{\"@odata.context\":\"../../../../../../../$metadata#Customers(1)/Name\",\"value\":\"Joe\"}")]
public void TopLevelPropertyShouldReadContextUriAsRelativeUri(string payload)
{
    // Arrange
    var model = this.CreateEdmModelWithEntity();
    var primitiveTypeRef = ((IEdmEntityType)model.SchemaElements.First()).Properties().First().Type;
    this.messageReaderSettings = new ODataMessageReaderSettings() { BaseUri = new Uri("http://odata.org/test/") };

    // Act
    ODataJsonPropertyAndValueDeserializer deserializer = new ODataJsonPropertyAndValueDeserializer(this.CreateJsonInputContext(payload, model));
    ODataProperty property = deserializer.ReadTopLevelProperty(primitiveTypeRef);

    // Assert
    Assert.NotNull(property);
    Assert.Equal("http://odata.org/test/$metadata#Customers(1)/Name", deserializer.ContextUriParseResult.ContextUri.ToString());
}
```

cmackethan-pg: Updated 🫡

WanjohiSammy: Could you add these 2 as well:
```
[InlineData("{\"@odata.context\":\"/$metadata#Customers(1)/Name\",\"value\":\"Joe\"}")]
[InlineData("{\"@odata.context\":\"./$metadata#Customers(1)/Name\",\"value\":\"Joe\"}")]
```

WanjohiSammy (dismissed):

test/UnitTests/Microsoft.OData.Core.Tests/Json/ODataJsonDeserializerTests.cs
cmackethan-pg: Yep, I just added them.

PR #3243, opened 2025-05-09
updating build pipeline and PR template to reflect CI build instructions

corranrogue9:
For security reasons, the CI build does not run automatically for PRs within this repository. The build can be triggered by a team member commenting `/AzurePipelines run` on the PR. I have added these instructions to the PR template so that we can remind ourselves of this mechanism. I have also updated the YAML to not "auto cancel", which, described [here](https://learn.microsoft.com/en-us/azure/devops/pipelines/repos/github?view=azure-devops&tabs=yaml#multiple-pr-updates) current cancels running builds whenever a new commit is pushed to the PR. 

The intent is to make CI builds easier to run and not interfere with the general workflow of the individual developer.

### DIFF
--- a/azure-pipelines-rolling.yml
+++ b/azure-pipelines-rolling.yml
@@ -11,8 +11,11 @@ trigger:
    
 # Pull request (PR) triggers
 pr:
-- main
-- dev-8.x
+  autoCancel: false
+  brannches:
+    include:
+      - main
+      - dev-8.x
 
 resources:
   repositories:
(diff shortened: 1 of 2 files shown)


azure-pipelines-rolling.yml
```diff
@@ -11,8 +11,11 @@ trigger:
    
 # Pull request (PR) triggers
 pr:
-- main
-- dev-8.x
+  autoCancel: false
+  brannches:
```
marabooy, line 15:
```suggestion
  branches:

```
typo

WanjohiSammy:
Is there a way to prevent the `azure-pipelines bot` from sending the following response when the pipeline is triggered behind the scenes?

`No pipelines are associated with this pull request.`

Issue #2778, opened 2023-10-25
System.NotSupportedException: Could not convert constant MyEnumValue expression to string.


rpallares:
Hello,
I use Microsoft.OData.Client@7.18.0 library to get data from service, generated from Microsoft.OData.Cli@0.2.1.

I have an exception when using In operator on some enumeration values following:
https://learn.microsoft.com/en-us/odata/client/query-options#filter

### Assemblies affected
- Microsoft.OData.Client@7.18.0

### Reproduce steps
```c#
MyEnum[] enumValues = { MyEnum.Value1, MyEnum.Value2 };

var results = client.MyEntitySet
    .Where(entity => enumValues.Contains(entity.EntityEnumValue)
    .ToList();
```

### Expected result
The query is generated and executed correctly using the In operator.

### Actual result
```
System.NotSupportedException: Could not convert constant Value1 expression to string.
   at Microsoft.OData.Client.ExpressionWriter.VisitConstant(ConstantExpression c)
   at Microsoft.OData.Client.ALinqExpressionVisitor.Visit(Expression exp)
   at Microsoft.OData.Client.ExpressionWriter.Visit(Expression exp)
   at Microsoft.OData.Client.ExpressionWriter.VisitMethodCall(MethodCallExpression m)
   at Microsoft.OData.Client.ALinqExpressionVisitor.Visit(Expression exp)
   at Microsoft.OData.Client.ExpressionWriter.Visit(Expression exp)
   at Microsoft.OData.Client.ExpressionWriter.VisitOperand(Expression e, Nullable`1 parentType, Nullable`1 childDirection)
   at Microsoft.OData.Client.ExpressionWriter.VisitBinary(BinaryExpression b)
   at Microsoft.OData.Client.ALinqExpressionVisitor.Visit(Expression exp)
   at Microsoft.OData.Client.ExpressionWriter.Visit(Expression exp)
   at Microsoft.OData.Client.ExpressionWriter.VisitOperand(Expression e, Nullable`1 parentType, Nullable`1 childDirection)
   at Microsoft.OData.Client.ExpressionWriter.VisitBinary(BinaryExpression b)
   at Microsoft.OData.Client.ALinqExpressionVisitor.Visit(Expression exp)
   at Microsoft.OData.Client.ExpressionWriter.Visit(Expression exp)
   at Microsoft.OData.Client.ExpressionWriter.VisitOperand(Expression e, Nullable`1 parentType, Nullable`1 childDirection)
[...]

PR #3253, opened 2025-05-21
Fixes #1375 Allows to filter models using Contains with Enum collections

Glucocorticoid:
### Issues

*This pull request fixes #1375.*

### Description

OData endpoint can filter models by enum property usng collection Contains method: `$filter=Color in ('Blue','Green')`. But odata client throws exception and dont covert code `requestedColors.Contains(model.Color)` to valid query. 

- [ ] *Test cases added*

### Additional work necessary

*None.*

### DIFF
--- a/src/Microsoft.OData.Core/Evaluation/LiteralFormatter.cs
+++ b/src/Microsoft.OData.Core/Evaluation/LiteralFormatter.cs
@@ -267,6 +267,11 @@ namespace Microsoft.OData.Evaluation
                 return enumValue.Value;
             }
 
+            var commonEnum = value as Enum;
+            {
+                return commonEnum.ToString();
+            }
+
             throw SharedUtils.CreateExceptionForUnconvertableType(value);
         }
 
@@ -488,6 +493,11 @@ namespace Microsoft.OData.Evaluation
                 }
 
                 string result = this.FormatAndEscapeLiteral(value);
+
+                if (value is Enum)
+                {
+                    return string.Concat("'", result, "'");
+                }
 
                 if (value is byte[])
                 {
--- a/test/UnitTests/Microsoft.OData.Client.Tests/DataServiceQueryProviderTests.cs
+++ b/test/UnitTests/Microsoft.OData.Client.Tests/DataServiceQueryProviderTests.cs
@@ -62,6 +62,22 @@ namespace Microsoft.OData.Client.Tests
         }
 
         [Fact]
+        public void TranslatesEnumerableContainsWithEnumToInOperator()
+        {
+            // Arrange
+            var sut = new DataServiceQueryProvider(dsc);
+            var productColors = new[] { Color.None, Color.Blue, Color.Green };
+            var products = dsc.CreateQuery<Product>("Products")
+                .Where(product => productColors.Contains(product.Color));
+
+            // Act
+            var queryComponents = sut.Translate(products.Expression);
+
+            // Assert
+            Assert.Equal(@"http://root/Products?$filter=Color in ('None','Blue','Green')", queryComponents.Uri.ToString());
+        }
+
+        [Fact]
         public void TranslatesEnumerableContainsWithSpecialCharactersToInOperator()
         {
             // Arrange
@@ -93,6 +109,21 @@ namespace Microsoft.OData.Client.Tests
         }
 
         [Fact]
+        public void ThrowsForEnumerableContainsWithEmptyEnumCollection()
+        {
+            // Arrange
+            var sut = new DataServiceQueryProvider(dsc);
+            var products = dsc.CreateQuery<Product>("Products")
+                .Where(product => Enumerable.Empty<Color>().Contains(product.Color));
+
+            // Act
+            var exception = Assert.ThrowsAny<InvalidOperationException>(() => sut.Translate(products.Expression));
+
+            // Assert
+            Assert.Equal(SRResources.ALinq_ContainsNotValidOnEmptyCollection, exception.Message);
+        }
+
+        [Fact]
         public void TranslatesEnumerableContainsWithSingleItem()
         {
             // Arrange
@@ -105,6 +136,21 @@ namespace Microsoft.OData.Client.Tests
 
             // Assert
             Assert.Equal(@"http://root/Products?$filter=Name in ('Pancake mix')", queryComponents.Uri.ToString());
+        }
+
+        [Fact]
+        public void TranslatesEnumerableContainsWithSingleEnumItem()
+        {
+            // Arrange
+            var sut = new DataServiceQueryProvider(dsc);
+            var products = dsc.CreateQuery<Product>("Products")
+                .Where(product => new[] { Color.None }.Contains(product.Color));
+
+            // Act
+            var queryComponents = sut.Translate(products.Expression);
+
+            // Assert
+            Assert.Equal(@"http://root/Products?$filter=Color in ('None')", queryComponents.Uri.ToString());
         }
 
         [Fact]
@@ -369,6 +415,8 @@ namespace Microsoft.OData.Client.Tests
 
             public decimal Price { get; set; }
 
+            public Color Color { get; set; }
+
             public Edm.Date LaunchDate { get; set; }
 
             public IEnumerable<string> Comments { get; set; }
@@ -385,6 +433,14 @@ namespace Microsoft.OData.Client.Tests
             {
                 return string.IsNullOrEmpty(data);
             }
+        }
+
+        private enum Color
+        {
+            None = 0,
+            Red = 1,
+            Green = 2,
+            Blue = 3
         }
 
         private static class UriFunctions


test/UnitTests/Microsoft.OData.Client.Tests/DataServiceQueryProviderTests.cs
```diff
@@ -61,6 +61,22 @@ public void TranslatesEnumerableContainsToInOperator()
             Assert.Equal(@"http://root/Products?$filter=Name in ('Milk','Cheese','Donut')", queryComponents.Uri.ToString());
         }
 
+        [Fact]
+        public void TranslatesEnumerableContainsWithEnumToInOperator()
+        {
+            // Arrange
+            var sut = new DataServiceQueryProvider(dsc);
+            var productColors = new[] { Color.None, Color.Blue, Color.Green };
```
WanjohiSammy, line 69: Could you try with these:
```cs
var productColors = new List<Color> { Color.None, Color.Blue, Color.Green };

IEnumerable<CustomerType> productColors = new List<Color> { Color.None, Color.Blue, Color.Green };
```

src/Microsoft.OData.Core/Evaluation/LiteralFormatter.cs
```diff
@@ -267,6 +267,11 @@ private static string FormatRawLiteral(object value)
                 return enumValue.Value;
             }
 
+            var commonEnum = value as Enum;
+            {
+                return commonEnum.ToString();
+            }
+
             throw SharedUtils.CreateExceptionForUnconvertableType(value);
```
WanjohiSammy, line 275: This line will never be reached.

test/UnitTests/Microsoft.OData.Client.Tests/DataServiceQueryProviderTests.cs
Glucocorticoid: > IEnumerable<CustomerType> productColors = new List<Color>
Sorry, I didn`t get it. Can you please add more details about this case? I should use EdmCollectionType for CustumerType or just swap `var` to `IEnumerable`?  I think I missed something about CustumerType to be able to assign collection of Color to collection of CustomerType.

src/Microsoft.OData.Core/Evaluation/LiteralFormatter.cs
WanjohiSammy: I meant `IEnumerable<Color>` or `List<Color>`

Glucocorticoid: I've tried both variants: List<Color> and IEnumerable<Color>. Got same result - without casting `value` to `Enum` it doesn't work in test cases and throws exception for unconvertable type.

test/UnitTests/Microsoft.OData.Client.Tests/DataServiceQueryProviderTests.cs
Glucocorticoid: I made requested changes.

src/Microsoft.OData.Core/Evaluation/LiteralFormatter.cs
WanjohiSammy: With this solution, `throw SharedUtils.CreateExceptionForUnconvertableType(value);` will never be triggered. I think the solution to this issue is in [ uriLiteral = LiteralFormatter.ForConstants.Format(item);](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs#L564)

WanjohiSammy: Changing [uriLiteral = LiteralFormatter.ForConstants.Format(item);](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs#L564) to the below should fix the problem:
```cs
object currentValue = item;
if (item.GetType().IsEnum)
{
    currentValue = item.ToString();
}

uriLiteral = LiteralFormatter.ForConstants.Format(currentValue);
```

However, this will not work with `instance Contains method` unless you cast it to [`LINQ generic Contains`](https://referencesource.microsoft.com/#System.Core/System/Linq/Enumerable.cs,f60bab4c5e27a849,references) method
```cs
....
List<T>.Contains()
HasShet<T>.Contains()
Collection<T>.Contains()
....

```

To use the `generic Contains method` from LINQ with a `List<T>`, `HasShet<T>`, `Collection<T>` etc, you can cast the collection to `IEnumerable<T>` or call the extension method explicitly. For example:
1. Cast to IEnumerable<T>
```cs
var query = context.Customers
    .Where(p => ((IEnumerable<DaysOfWeek>)days).Contains((DaysOfWeek)p.PreferredDay));
```

2. Use Enumerable.Contains
```
List<DaysOfWeek> days = new() { DaysOfWeek.Monday, DaysOfWeek.Tuesday };
var query = context.Customers
    .Where(p => Enumerable.Contains(days, (DaysOfWeek)p.PreferredDay));

```
This is because, if you take a look at [Dictionary<String, SequenceMethod> map = new Dictionary<string, SequenceMethod>(EqualityComparer<string>.Default);](https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Client/ALinq/ReflectionUtil.cs#L39), there are no entries for `List, Collection, HashSet` methods (i.e, Contains) in the **map dictionary**

Glucocorticoid: Thanks for explanation. 

I didn't add any new code to `ReflectionUtil`. That what I found in debugger:

Test case:
```c#
IEnumerable<Color> productColors = new List<Color> { Color.None, Color.Blue, Color.Green };
var products = dsc.CreateQuery<Product>("Products")
    .Where(product => productColors.Contains(product.Color));
```

ExpressionWriter: 

https://github.com/OData/odata.net/blob/67ec5ef4482019e0b9e3f0e283cf53c71dd113c7/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs#L535
`b is null`

https://github.com/OData/odata.net/blob/67ec5ef4482019e0b9e3f0e283cf53c71dd113c7/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs#L536

`m ={value(System.Collections.Generic.List'1[Microsoft.OData.Client.Tests.DataServiceQueryProviderTests+Color]).Contains([10008].Color)}`

`m.Method =
 {Boolean Contains[Color](System.Collections.Generic.IEnumerable'1[Microsoft.OData.Client.Tests.DataServiceQueryProviderTests+Color], Color)}`

So, ReflectionUtis can handle it as IEnumerable.Contains via existing code:
https://github.com/OData/odata.net/blob/67ec5ef4482019e0b9e3f0e283cf53c71dd113c7/src/Microsoft.OData.Client/ALinq/ReflectionUtil.cs#L185

And `ReflectionUtil.IsSequenceMethod(m.Method, SequenceMethod.Contains)` returns true.
https://github.com/OData/odata.net/blob/67ec5ef4482019e0b9e3f0e283cf53c71dd113c7/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs#L549

I guess we don't need to cast to IEnumerable.Contains - it already works as is. 
Next stop on ExpressionWriter.cs#L564 is reached.

https://github.com/OData/odata.net/blob/67ec5ef4482019e0b9e3f0e283cf53c71dd113c7/src/Microsoft.OData.Client/ALinq/ExpressionWriter.cs#L564

It calls LiteralFormatter.cs/Format -> DefaultLiteralFormatter.Format(object value) -> this.FormatLiteralWithTypePrefix(value)  and then proposed patch:
```suggestion
                if (value is Enum)
                {
                    return string.Concat("'", result, "'");
                }
```

I think it would be better to place
```suggestion
            if (value.GetType().IsEnum)
            {
                    return string.Concat("'", result, "'");
            }
```
in `DefaultLiteralFormatter` with other type checks (as proposed) instead of `ExpressionWriter`.

Should I add tests with List.Contains, HashSet.Contains, etc?

Glucocorticoid: There is already test for collection of strings in main:
https://github.com/OData/odata.net/blob/67ec5ef4482019e0b9e3f0e283cf53c71dd113c7/test/UnitTests/Microsoft.OData.Client.Tests/DataServiceQueryProviderTests.cs#L49

It is also works for collection of numbers as well. For example:
```c#
        [Fact]
        public void TranslatesEnumerableContainsToInOperatorWithNumbers()
        {
            // Arrange
            var sut = new DataServiceQueryProvider(dsc);
            var productIds = new[] { 2, 3, 4, 5, 6 };
            var products = dsc.CreateQuery<Product>("Products")
                .Where(product => productIds.Contains(product.Id));

            // Act
            var queryComponents = sut.Translate(products.Expression);

            // Assert
            Assert.Equal(@"http://root/Products?$filter=Id in (2,3,4,5,6)", queryComponents.Uri.ToString());
        }
```

But it didn't work for collection of enums. This problem mentioned in https://github.com/OData/odata.net/issues/2778 and I have this problem too.

My model:
```c#
public enum HistoryStatus
{
        Created = 0,
        Started = 1,
        Failed = 2,
        //...
}

public class HistoryEntry
{
        public long Id { get; init; }
        public HistoryStatus Status { get; init; }
        public DateTime CreatedAt { get; init; }
}
```

When I filter by Id using `e => requestedIds.Contains(e.Id)` it works fine. But when I filter by enum using `e => validHistoryStatuses.Contains(e.Status)` it throws exeption no matter what kind of collection I use (List, HashSet, Array).

OData endpoint can filter by enum with this query string `/odata/history?Status in ('Created', 'Failed')`. But OData client doesn't support it now because can't handle enums inside Contains method.

I had to change type of `Status` property from enum to int to be able to use "Contains" filtration.

I am trying to add enum support for "Contains" because it is already work for other primitive types but not for enums.

WanjohiSammy: I meant collection of strings that uses `List<string>`. For example:
```cs
var names = new List<string> { "Foo", "bar" };
var numbers = new HashSet<int> { 2, 3, 4, 6 };
```

habbes: @WanjohiSammy not sure I understand the thread correctly. Is the issue here enum support or support for arbitrary collection types like `List<T>`?

Glucocorticoid: An issue I am trying to solve in this PR is about enum support. Contains method already works for strings or numbers, but not for enums. Server side OData controller can filter by enum collection like `/odata/history?Status in ('Created', 'Failed')` , but OData client throws an exception for collection of enums as unsupported collection type. 

However, collection definition of any type is not clear to use for filtering with `Contains`. For example:
```c#
var stringFilter = new List<string>("foo", "bar");
\\...
var products = dsc.CreateQuery<Product>("Products")
                .Where(product => stringFilter.Contains(product.Name));
```

will throw an exeption too, because in runtime ReflectionUtil.GetAllLinqOperators collects only public static methods for `Queryable`, `Enumerable` and `DataServiceExtensions` types. List<string> in runtime will not pass collection check in `ExpressionWriter` (`ReflectionUtil.IsSequenceMethod`).

But
 ```c#
IEnumerable<string> stringFilter = new List<string>("foo", "bar");
\\...
var products = dsc.CreateQuery<Product>("Products")
                .Where(product => stringFilter.Contains(product.Name));
``` 
will work fine. Reflection will get an Enumerable static extention method `public static bool Contains<TSource>(this IEnumerable<TSource> source, TSource value);` and pass collection check in `ExpressionWriter`.

There is no compile time warnings or IDE suggestions about using "bad" collection type.

I think it is different kind of problem and support for arbitrary collection types can be added/clarified in another PR.

File: test/UnitTests/Microsoft.OData.Client.Tests/DataServiceQueryProviderTests.cs [hunk 3]
```diff
@@ -61,6 +61,22 @@ public void TranslatesEnumerableContainsToInOperator()
             Assert.Equal(@"http://root/Products?$filter=Name in ('Milk','Cheese','Donut')", queryComponents.Uri.ToString());
         }
 
+        [Fact]
+        public void TranslatesEnumerableContainsWithEnumToInOperator()
+        {
+            // Arrange
+            var sut = new DataServiceQueryProvider(dsc);
+            IEnumerable<Color> productColors = new List<Color> { Color.None, Color.Blue, Color.Green };
```
WanjohiSammy, line 69: Could you add a test for:
```cs
var productColors = new List<Color> { Color.None, Color.Blue, Color.Green };

// OR

List<Color> productColors = new List<Color> { Color.None, Color.Blue, Color.Green };
```

src/Microsoft.OData.Core/Evaluation/LiteralFormatter.cs
```diff
@@ -267,6 +267,12 @@ private static string FormatRawLiteral(object value)
                 return enumValue.Value;
             }
 
+            var commonEnum = value as Enum;
+            if(commonEnum  != null)
```
WanjohiSammy:
```suggestion
            if(value is Enum commonEnum)

```

test/UnitTests/Microsoft.OData.Client.Tests/DataServiceQueryProviderTests.cs
Glucocorticoid: Added test for not supported `Contains` method from generic collection types.

WanjohiSammy (approved):

PR #3260, opened 2025-06-03
OData Writer API and abstraction brainstorming

habbes:
## Issues

Related to issue https://github.com/OData/odata.net/pull/3245

## Description

This proposal/brainstorming session to figure out the right public API and abstractions to expose for the new writer.

### DIFF
--- a/docs/designs/odata-writer-rewrite-proposal.md
+++ b/docs/designs/odata-writer-rewrite-proposal.md
@@ -1,0 +1,318 @@
+# OData Writer re-design for better performance
+
+The issue https://github.com/OData/odata.net/issues/3236 summarizes the performance issues of the current OData Writer design.
+We aim to create a foundation for a new writer this will address these performance issues and scale better with larger inputs.
+
+This document presents a very high-level proposal with aim of figuring the correct layering of abstraction levels and extensibility
+mechanisms to expose to callers.
+
+More concretely, here are the questions I hope to get answers to:
+
+- What layers of abstractions should the new writer have to ensure we offer sufficient flexibility for use cases the OData writer must support, while providing efficient default implementations for common scenarios?
+- How tightly coupled should we be with `JsonSerializer` and `Utf8JsonWriter`?
+- Should we account for other formats other than JSON?
+- Should we require an `IEdmModel` to be provided for serialization?
+
+Here are the design principles that I want to adhere to:
+
+- Avoid forcing the caller to create intermediate data unnecessarily.
+  - A lot of the cost of the existing writer comes from intermediate data and repeated/redundant computations.
+  - The existing writer requires the caller to create intermediate data via `ODataResource`, `ODataValue`, etc.
+- If the caller passes structs as input, they should not be forced to box them.
+  - Or existing abstractions often accept or return `object`, which force value types to be boxed.
+- Avoid paying the cost for things we don't use.
+  - Existing writer allocates things that are not used (e.g. `ODataResource.Actions`).
+  - Existing writer performs validation eagerly (e.g. `ODataResourceBase.VerifyProperties()`) even when not necessary.
+- New writer should expose similar levels of abstractions and extensibility to existing writer
+- New writer should provide support for all the scenarios that the existing writer supports
+  - Different input types (CLR objects, EdmObject, Dictionaries, JObject)
+  - Different types of payloads: resource set, values, aggregation results, service document, raw values, metadata document, etc.
+  - Dynamic properties, delta payloads, stream properties, polymorphic properties, $select/$expand
+  - Standard OData annotations, custom annotations, etc.
+  - Streaming support (start serialization before all data is available)
+
+The sample code in this doc or for demonstration purposes and are subject to change as the design is fleshed out more.
+
+## Top-level writer
+
+This is the equivalent to the existing `ODataMessageWriter` layer, which also exposes the abstract `ODataWriter` and `ODataOutputContext` classes.
+The `ODataMessageWriter` exposes methods for writing top-level payloads like metadata, service document, property values, etc. or methods
+for creating writers for writing more complex, nested payloads like resources and resouce sets.
+
+However, `ODataMessageWriter` doesn't implement any logic, all the logic is delegated to the abstract `ODataOutputContext` and `ODataWriter` classes.
+The `ODataOutputContext` does most of the heavy lifting and is the one that determines the output format.
+
+This means that if you wanted to create your own customer `ODataMessageWriter` from scratch (e.g. to output a different format other than JSON), you would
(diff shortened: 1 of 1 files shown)


docs/designs/odata-writer-rewrite-proposal.md
```diff
@@ -0,0 +1,315 @@
...
+### Concerns:
+
+* Some OData payloads—particularly `$value` responses—aren’t always valid JSON (e.g., unquoted strings), so they cannot be written using `Utf8JsonWriter`. Since `JsonSerializer` doesn’t expose the output stream, we’d need to:
+
+  * Detect the payload type and use a different writer that supports `$value` (e.g., a custom `ODataValueWriter`), or
+  * Design the new OData writer to write directly to the stream.
+
+  Note: `$metadata` currently still defaults to XML.
+
+* If customers implement their own `JsonConverter` to customize serialization for specific types, they would also need to manually handle support for `$select`, `$expand`, annotations, dynamic properties, etc. This includes knowing which annotations to write and where. If they already have generic `JsonConverters`, they likely won't work out-of-the-box. They would need to write OData-specific `JsonConverters`.
```
corranrogue9, line 253: I agree with your point here, it's super important that we try to abstract the odata details from callers. I think we also need to recognize that customers *will* need to know these things if they wish to customize certain behaviors. Again, though, I agree that the converter is *not* the layer where they should be expected to do those things correctly just to get some custom behavior.

docs/designs/odata-writer-rewrite-proposal.md
```diff
@@ -0,0 +1,315 @@
...
+- You can replace the low-level `IJsonWriter` by implementing the interface
+- You can control what gets written through the properties and annotatons that you add to `ODataResource` and similar types
+
+Here's a rough proposal for the new writer:
+
+```csharp
+interface IODataMessageWriter<TContext, TState>
+{
+    IODataWriter<TContext, TState, TValue> GetWriter<TValue>(ODataPayloadKind payloadKind);
+    // Alternatively we could have different `GetWriter` methods for different
```
corranrogue9, line 65: please do this; i think avoiding enums is really helpful for us to have visibility into breaking changes

docs/designs/odata-writer-rewrite-proposal.md
```diff
@@ -0,0 +1,315 @@
...
+interface IODataWriter<TContext, TState, TValue>
+{
+    ValueTask WriteAsync(TValue value, TState state, TContext context);
+}
+
+var odataWriter = ODataWriterBuilder.FromPayloadKind(payloadKind)
+    .WithJsonOptions(jsonSerializationOptions)
+    .WithODataUri(odataUri)
+    .ConfigureResourceWriters(provider => provider.Add<Customer, CustomerResourceWriter>())
+    .ForCollectionOf<Customer>()
```
corranrogue9, line 102: Let's remember that AGS wants to use our readers and writers, and doesn't have CLR types for the types represented in the model

docs/designs/odata-writer-rewrite-proposal.md
```diff
@@ -0,0 +1,315 @@
...
+interface IODataWriter<TContext, TState, TValue>
+{
+    ValueTask WriterAsync(TValue value, TState state, TContext context);
+}
+```
+
+where:
+
+- `TContext`: represents an implementation-specific context type that encapsulates session-specific settings and format information (e.g. JSON options, ODataUri, Output stream, custom services)
+- `TState`: implementation-specific state type that represents the current state of the serialization. Unlike the context, this is expected to change per object or level (e.g. current type being serialized, nesting depth, etc.). I'm not yet sure if it makes sense to have this
```
corranrogue9, line 79: > I'm not yet sure if it makes sense to have this

I was writing an argument against having `TContext` **and** `TState` and I convinced myself otherwise. The use-case: `TState` is "has a property already been written" so that we can avoid trailing ","; `TContext` is a logger that the writer posts to when significant writes have occurred (like the end of an object); `TValue` is the value currently being written

corranrogue9: `TContext` to me is stuff that is not changing throughout the write, which `TState` **can** change throughout the write, and `TValue` is necessary be definition

docs/designs/odata-writer-rewrite-proposal.md
```diff
@@ -0,0 +1,315 @@
...
+## Custom extensible implementation
+
+This layer is equivalent to the `ODataResource` and similar types. Since this design moves away from intermediate data, we'll replace these with interfaces that implement the logic of writing a specific type.
+
+```c#
+interface ODataJsonWriterContext
+{
+    public ODataUri ODataUri { get; }
+    public IEdmModel Model { get; }
+    public Ut8JsonWriter JsonWriter { get; }
```
corranrogue9, line 117: based on the separation i see between `tstate` and `tcontext`, i think the writer goes into `tstate`; please clarify what you're thinking the difference is

docs/designs/odata-writer-rewrite-proposal.md
habbes: In the case of AGS they would inject custom writers for their input types `JArray`, `JObject` or the input stream. The build pattern here is to provide a discoverable way to configure the service, so there could be helper methods for different common use cases (including when it's not an `IEnumerable<T>`), while also providing where to define the type directly for the less common use cases.

docs/designs/odata-writer-rewrite-proposal.md
habbes: The thinking here is that the same writer would be passed along throughout the entire serialization of the payload, the writer here is like a stand-in for the output stream, it's shared from the start to the end of the serialization session.

docs/designs/odata-writer-rewrite-proposal.md
```diff
@@ -0,0 +1,315 @@
...
+
+This is the equivalent to the existing `ODataMessageWriter` layer, which also exposes the abstract `ODataWriter` and `ODataOutputContext` classes.
+The `ODataMessageWriter` exposes methods for writing top-level payloads like metadata, service document, property values, etc. or methods
+for creating writers for writing more complex, nested payloads like resources and resouce sets.
+
+However, `ODataMessageWriter` doesn't implement any logic, all the logic is delegated to the abstract `ODataOutputContext` and `ODataWriter` classes.
+The `ODataOutputContext` does most of the heavy lifting and is the one that determines the output format.
+
+This means that if you wanted to create your own customer `ODataMessageWriter` from scratch (e.g. to output a different format other than JSON), you would
+need to provide custom `ODataOutputContext` and `ODataWriter` implementations. This is a non-trivial undertaking and I doubt there are any implementations
```
xuzhg, line 46: Check this: https://devblogs.microsoft.com/odata/customize-odata-payload-serialization-format-within-asp-net-core-odata/

PR #3265, opened 2025-06-12
Default EnableCaseInsensitive to true when initializing a new instance of ODataUriResolver

### LINKED ISSUE #3066: Default EnableCaseInsensitive to true when initializing a new instance of ODataUriResolver
OData V4.01 states that [services MUST support case-insensitive system query option names specified with or without the $ prefix](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_SystemQueryOptions).

For this reason, we inject a case insensitive `ODataUriResolver` in ASP.NET Core OData:
https://github.com/OData/AspNetCoreOData/blob/1368836752963d88e920d9508d94807eb995894b/src/Microsoft.AspNetCore.OData/ODataOptions.cs#L326-L329

For the same reason, we should consider defaulting `EnableCaseInsensitive` to true when initializing a new instance of `ODataUriResolver` in ODL:
https://github.com/OData/odata.net/blob/95233fc8a399cbae6e816216528d02a581b77b48/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs#L24-L24

### Assemblies affected

- Microsoft.OData.Core 8.x
- Microsoft.OData.Core 7.x

### Expected result

The injected `ODataUriResolver` should have `EnableCaseInsensitive` default to `true`

### Actual result

The injected `ODataUriResolver` has `EnableCaseInsensitive` default to `false`

### Additional detail

Related to https://github.com/OData/AspNetCoreOData/issues/1312, where the call to `AddRouteComponents` configures an `ODataUriResolver` with `EnableCaseInsensitive` set to `true` but when `AddDefaultODataServices` is called, it configures an `ODataUriResolver` with `EnableCaseInsensitive` set to false. This is not a bug per se, but `true` is arguably the more appropriate default given the OData standard and in line with the principle of least astonishment.

gathogojr:
### Issues

*This pull request fixes #3066.*

### Description

This pull request addresses #3066 by aligning the default behavior of `ODataUriResolver` with the OData V4.01 specification.

According to the OData V4.01 specification, services **MUST** support case-insensitive system query option names, whether or not they are prefixed with `$`.

ASP.NET Core OData already complies with this requirement by injecting a case-insensitive `UnqualifiedODataUriResolver` (derived from `ODataUriResolver`) by default:
https://github.com/OData/AspNetCoreOData/blob/1368836752963d88e920d9508d94807eb995894b/src/Microsoft.AspNetCore.OData/ODataOptions.cs#L326-L329

To ensure consistent behavior across the OData ecosystem, this PR proposes setting `EnableCaseInsensitive = true` by default when initializing a new instance of `ODataUriResolver` in ODL:
https://github.com/OData/odata.net/blob/95233fc8a399cbae6e816216528d02a581b77b48/src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs#L24

With this change, both of the following requests will be parsed successfully and equivalently:

1.
```http
GET https://localhost:54001/odata/Books?$expand=Authors
```

2.
```http
GET https://localhost:54001/odata/Books?$expand=authors
```

This change improves spec compliance and developer experience by reducing case-sensitivity issues in query options.

- [x] *Test cases added*

### Additional work necessary

### Repository notes


src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -35,7 +35,7 @@ public class ODataUriResolver
         /// <remarks>
         /// All extensions should look at this property and keep case sensitive behavior consistent.
         /// </remarks>
-        public virtual bool EnableCaseInsensitive { get; set; }
```
xuzhg, line 38: From your notes in the PR, it says that the case insensitive for the query options key name no matter whether it prefixes with '$', Is it right?

Here, if we default to 'true', it means all the Uri parts are case insensitive by default?

gathogojr: @xuzhg That statement is directly from the standard: OData V4.01 states that [services MUST support case-insensitive system query option names specified with or without the $ prefix](https://docs.oasis-open.org/odata/odata/v4.01/odata-v4.01-part2-url-conventions.html#sec_SystemQueryOptions).

With this change, the URI parts are case-insensitive by default - which is what the `UnqualifiedODataUriResolver` does in ASP.NET Core OData here:
https://github.com/OData/AspNetCoreOData/blob/1368836752963d88e920d9508d94807eb995894b/src/Microsoft.AspNetCore.OData/ODataOptions.cs#L326

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
```diff
@@ -35,7 +35,7 @@ public class ODataUriResolver
         /// <remarks>
         /// All extensions should look at this property and keep case sensitive behavior consistent.
         /// </remarks>
-        public virtual bool EnableCaseInsensitive { get; set; }
+        public virtual bool EnableCaseInsensitive { get; set; } = true;
```
WanjohiSammy, line 38: I thought this was already `true` by default.

Could you add a test that makes this to false.

gathogojr: @WanjohiSammy The default was `false`, as you can see from the following tests:
image

For the `UnqualifiedODataUriResolver` in ASP.NET Core OData here
https://github.com/OData/AspNetCoreOData/blob/1368836752963d88e920d9508d94807eb995894b/src/Microsoft.AspNetCore.OData/ODataOptions.cs#L326, we set it to `true` to support case insensitive behaviour when resolving OData URIs.

Regarding your ask, do you need me to add a test like the following:
```csharp
        [Fact]
        public void TestEnableCaseInsensitiveBuiltinIdentifierSetToFalse()
        {
            Assert.False((new ODataUriResolver { EnableCaseInsensitive = false }).EnableCaseInsensitive);
        }
```

WanjohiSammy (dismissed):

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
WanjohiSammy: Something like this will suffice

WanjohiSammy (approved):

habbes (approved):

src/Microsoft.OData.Core/UriParser/Resolver/ODataUriResolver.cs
CAPCHIK: Hi @gathogojr,

Could you clarify why this MR changes the handling of property names? The standard specifies that query parameter names (like $expand) should be case-insensitive, but not their values. For example, $exand and $EXPand should be treated as equivalent, but $expand=shoe and $expand=Shoe should not.

This has been a pain point for me regarding case insensitivity—especially since C# class properties are also resolved in a case-insensitive manner, though I haven’t found a clear justification for it. ([Related PR where I struggled to write tests](https://github.com/OData/AspNetCoreOData/pull/1482))

Could you explain why property names should be handled case-insensitively?

Issue #3264, opened 2025-06-12
Redundant property verifications in OData reader


habbes:
## Is your feature request related to a problem? Please describe.

Related to #3257 
Also related to #3263 

This affects both `Microsoft.OData.Core` 7.x and 8.x

The OData reader performs `ODataResourceBase.VerifyProperties()` excessively causing a lot of CPU usage, especially when resources have a lot properties. The following data was collected from [profiling a benchmark](https://github.com/habbes/ODataReaderPerfInvestigations) that deserializes a resource with 999 properties. It shows at least 68% CPU spent on `VerifyProperties()`

![Image](https://github.com/user-attachments/assets/c93a9ffc-4ebd-4601-9ecb-eda340235ba0)

![Image](https://github.com/user-attachments/assets/a0149411-0522-467f-9408-214820e2f6bb)

During deserialization, the `ODataJsonPropertyAndValueDeserializer.AddResourceProperty` method is called each time we add a deserialized property into the `resource.Properties` collection.

If you look at the code, you'll notice that we update the `resource.Properties` property

```c#
resource.Properties = resource.Properties.ConcatToReadOnlyEnumerable(property)
```

This will invoke `resource.Properties`'s setter which runs the verification. Since the verification is called for each property we add, then the cost scales in quadratic time.

It's important to note that while `ConcatToReadOnlyEnumerable`'s name that we may be creating a new collection and adding the item there, we're actually appending to an existing mutable list. The custom `ReadOnlyEnumerable<T>` collection we use here is actually a wrapper on top of a `List<T>` and exposes a method `AddToSourceList` which appends the internal list using `List<T>.Add()`. So it's actually not read-only, a name such as `AppendOnlyEnumerable` would be less misleading.

There are many ways to fix this issue

- set ODataResource.SkipPropertyVerification to true when creating the resource. The concern with this is that it could potentially be a breaking change in some (narrow theoretical) use cases. I could address this concern by setting this property to false before setting the ResourceStart or  ResourceEnd state, but not sure if that's fool-proof.
- add a flag in ODataMessageReaderSettings that if enabled, will set SkipPropertyVerification to true, so no breaking change. But this will require explicit opt-in, so most users will not benefit from this since they don't know about it.
[...]

Issue #3263, opened 2025-06-12
Move resource property verification out of the Properties setter


habbes:
## Is your feature request related to a problem? Please describe.
*A clear and concise description of the problem or challenge you’re facing.*
**Example:** *Currently, the library does not support grouping with `rollup`.*

## Describe the solution you'd like
*A clear and concise description of what you want to happen.*
**Example:** *Add support for grouping with `rollup` to allow applying set transformations to instances of an input set organized in a leveled hierarchy.*

## Describe alternatives you've considered
*A description of alternative solutions or workarounds you've considered.*
**Example:** *Could not find an alternative solution or workaround.*

## Additional context
- *Screenshots, mockups, or examples (if applicable)*
- *References to similar implementations in other projects*
- *Any other relevant information*

## Would you like to work on this feature?
- [ ] Yes
- [ ] No
- [ ] Maybe (need guidance)

---
🔹 *Thanks for taking the time to suggest this feature! We appreciate your contributions. 🙌*

PR #3267, opened 2025-06-12
Avoid excessive property verifications when adding resource properties in OData reader

habbes:
### Issues

*This pull request addresses #3264 in ODL 7.x*

If this PR is approved, I'll port the fix to ODL 8.x

### Description

To fix the performance issue caused by excessively running property verification each time the reader adds a property to the resource being deserialized, I update the `resource.Properties` in-place in `ODataJsonLightPropertyAndValueDeserializer.AddResourceProperty()` without invoking the setter each time since the verification logic runs in the setter.

In lieu of running the verification logic, I use `Debug.Assert` statements to perform the equivalent verification during debug builds only, to avoid any runtime impact from the verification. I could have implemented the verification using logic that's executed also in prod builds that would also thrown an exception, this would still improve performance since the properties would be checked in O(n) times instead of O(n^2) times, but I found this unnecessary.

Here's the reasoning for doing debug-time verification only: If the reader stores the `ODataResourceValue` instances inside a resource's properties, then we would be getting exceptions already due to the existing property verification that runs by default. The fact that we don't get exceptions thrown in the reader related to this issue, means that the reader does not add such values. In any case, it would be a bug for the reader to add values in the resource properties which are considered invalid. Which is why I believe a `Debug.Assert` would suffice so we could catch such bugs when testing. Since these resources and properties are created inside the reader, we have full control of the internal behaviour.

The reason I chose this approach (over other alternatives listed in the attached issue) is because it's simple and localized to a single internal method. Every user will benefit from this without having to change their code or opt-in through some new flag. This is a non-breaking change. And we can easily change this implementation in the future without breaking.

Again, I do suggest we revisit the property verification and decide how to address it in a more holistic way since it also affects the writer performance and may affect other scenarios as well (see: https://github.com/OData/odata.net/issues/3263)

### Benchmarks

Benchmark code: 

#### Before

```sh
// * Summary *

BenchmarkDotNet v0.15.1, Windows 11 (10.0.26100.3775/24H2/2024Update/HudsonValley)
.NET SDK 10.0.100-preview.1.25120.13
  [Host]     : .NET 8.0.17 (8.0.1725.26602), X64 RyuJIT AVX-512F+CD+BW+DQ+VL
  DefaultJob : .NET 8.0.17 (8.0.1725.26602), X64 RyuJIT AVX-512F+CD+BW+DQ+VL
```

| Method       | Mean     | Error    | StdDev   | Gen0      | Gen1     | Gen2    | Allocated |
|------------- |---------:|---------:|---------:|----------:|---------:|--------:|----------:|
| ReadResource | 23.16 ms | 0.459 ms | 0.969 ms | 2562.5000 | 562.5000 | 62.5000 |  12.65 MB |

image

#### After

```sh
// * Summary *

BenchmarkDotNet v0.15.1, Windows 11 (10.0.26100.3775/24H2/2024Update/HudsonValley)
.NET SDK 10.0.100-preview.1.25120.13
  [Host]     : .NET 8.0.17 (8.0.1725.26602), X64 RyuJIT AVX-512F+CD+BW+DQ+VL
  DefaultJob : .NET 8.0.17 (8.0.1725.26602), X64 RyuJIT AVX-512F+CD+BW+DQ+VL
```

| Method       | Mean     | Error     | StdDev    | Gen0     | Gen1     | Allocated |
|------------- |---------:|----------:|----------:|---------:|---------:|----------:|
| ReadResource | 4.377 ms | 0.0867 ms | 0.1243 ms | 609.3750 | 453.1250 |   3.26 MB |

image

- [ ] *Test cases added*

### Additional work necessary

### Repository notes

### DIFF
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
@@ -12,6 +12,7 @@ namespace Microsoft.OData.JsonLight
     using System.Collections.ObjectModel;
     using System.Diagnostics;
     using System.Diagnostics.CodeAnalysis;
+    using System.Linq;
     using System.Threading.Tasks;
     using Microsoft.OData.Edm;
     using Microsoft.OData.Json;
@@ -1022,7 +1023,31 @@ namespace Microsoft.OData.JsonLight
             resourceState.PropertyAndAnnotationCollector.CheckForDuplicatePropertyNames(property);
             ODataResourceBase resource = resourceState.Resource;
             Debug.Assert(resource != null, "resource != null");
+
+            // The resource.Properties setter would run property verification
+            // each time a property is added to the resource. That is
+            // expensive and scales quadratically with the number of properties + number of items in collection properties.
+            // To avoid this cost, we skip property verification.
+            // Since the resource and properties are created internally by the reader,
+            // we don't need to verify the properties if we can guarantee that we are only
+            // adding properties with acceptable values (i.e. not ODataResourceValue).
+            // We use debug-time assertions instead to catch such bugs in tests.
+
+            // This is a stop-gap measure to address a performance issue. But
+            // we should rethink the current approach to property verification.
+            // See: https://github.com/OData/odata.net/issues/3263
+
+            Debug.Assert(!(property.Value is ODataResourceValue), Strings.ODataResource_PropertyValueCannotBeODataResourceValue(property.Name));
+            Debug.Assert(
+                !(property.Value is ODataCollectionValue collectionValue && collectionValue.Items.Any(item => item is ODataResourceValue)),
+                Strings.ODataResource_PropertyValueCannotBeODataResourceValue(property.Name));
+
+            bool originalSkipPropertyVerificationValue = resource.SkipPropertyVerification;
+            resource.SkipPropertyVerification = true;
+            
             resource.Properties = resource.Properties.ConcatToReadOnlyEnumerable("Properties", property);
+
+            resource.SkipPropertyVerification = originalSkipPropertyVerificationValue;
             return property;
         }
 
--- a/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
+++ b/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
@@ -1024,17 +1024,15 @@ namespace Microsoft.OData.JsonLight
             ODataResourceBase resource = resourceState.Resource;
             Debug.Assert(resource != null, "resource != null");
 
-            // The resource.Properties setter would run property verification
-            // each time a property is added to the resource. That is
-            // expensive and scales quadratically with the number of properties + number of items in collection properties.
-            // To avoid this cost, we skip property verification.
+            // To avoid repeated computations for the property verification logic inside the
+            // resource.Properties setter, we update the resource.Properties in-place
+            // without invoking the setter each time.
             // Since the resource and properties are created internally by the reader,
             // we don't need to verify the properties if we can guarantee that we are only
             // adding properties with acceptable values (i.e. not ODataResourceValue).
             // We use debug-time assertions instead to catch such bugs in tests.
 
-            // This is a stop-gap measure to address a performance issue. But
-            // we should rethink the current approach to property verification.
+            // We should rethink the current approach to property verification.
             // See: https://github.com/OData/odata.net/issues/3263
 
             Debug.Assert(!(property.Value is ODataResourceValue), Strings.ODataResource_PropertyValueCannotBeODataResourceValue(property.Name));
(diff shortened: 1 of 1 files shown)


WanjohiSammy (approved):

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1022,7 +1023,31 @@ in resourceState.PropertyAndAnnotationCollector.GetCustomPropertyAnnotations(pro
...
+
+            Debug.Assert(!(property.Value is ODataResourceValue), Strings.ODataResource_PropertyValueCannotBeODataResourceValue(property.Name));
+            Debug.Assert(
+                !(property.Value is ODataCollectionValue collectionValue && collectionValue.Items.Any(item => item is ODataResourceValue)),
+                Strings.ODataResource_PropertyValueCannotBeODataResourceValue(property.Name));
+
+            bool originalSkipPropertyVerificationValue = resource.SkipPropertyVerification;
+            resource.SkipPropertyVerification = true;
+            
             resource.Properties = resource.Properties.ConcatToReadOnlyEnumerable("Properties", property);
```
WanjohiSammy: Why not concatenate in-place instead of re-assigning everytime

habbes: I think the reason it was written this way is because `ConcatToReadOnlyEnumerable` may create a new `ReadOnlyEnumerable` instance (if the current one is the empty instance). So setting the property like this ensure you always get the right instance. But you can address that edge case by manually checking for the empty instance outside of the method like I've done in the sample code. Of course this now means it will check for the empty instance twice, but that doesn't seem to hurt perf that much. I had also created a separate `AppendToReadOnlyEnumerable()` method that does not perform this check, but I felt like that was becoming too much code.

habbes:
@WanjohiSammy I have

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1022,7 +1023,31 @@ in resourceState.PropertyAndAnnotationCollector.GetCustomPropertyAnnotations(pro
...
+
+            resource.SkipPropertyVerification = originalSkipPropertyVerificationValue;
             return property;
```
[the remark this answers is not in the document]
habbes: Changed implementation. this is no longer a concern.

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
```diff
@@ -1022,7 +1023,31 @@ in resourceState.PropertyAndAnnotationCollector.GetCustomPropertyAnnotations(pro
...
+            // without invoking the setter each time.
+            // Since the resource and properties are created internally by the reader,
+            // we don't need to verify the properties if we can guarantee that we are only
+            // adding properties with acceptable values (i.e. not ODataResourceValue).
+            // We use debug-time assertions instead to catch such bugs in tests.
+
+            // We should rethink the current approach to property verification.
+            // See: https://github.com/OData/odata.net/issues/3263
+
+            Debug.Assert(!(property.Value is ODataResourceValue), Strings.ODataResource_PropertyValueCannotBeODataResourceValue(property.Name));
```
gathogojr, line 1038: Would `Contract.Assert` be a better choice?
That aside, I did investigate when `ODataJsonPropertyAndValueDeserializer.ReadResourceValue[Async]` is executed and found it's during reading of `ODataError` and annotations where resource values are allowed. Specifically I identified that the following tests do trigger the `ReadResourceValue[Async]` method:
- MultipleTypeCustomInstanceAnnotationsOnErrorShouldRoundtrip
- ResourceCustomInstanceAnnotationOnErrorShouldRoundtrip
- ReadTopLevelErrorWithResourceValueInstanceAnnotation_Works
- ReadTopLevelErrorWithCollectionResourceValueInstanceAnnotation_Works
- ReadStreamPrimitivePropertyWithoutValueButWithInstanceAnnotations

gathogojr (approved):
Left a comment, otherwise LGTM

src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs
habbes: Thanks @gathogojr  for the additional extensive checks. I went with `Debug.Assert` with the intention that these checks would be stripped out of release builds.

habbes: There is also existing use of `Debug.Assert` in that method. So, I think I'd need a special reason to use `Contract.Assert`. In this case, the only benefit I can think of for `Contract.Assert` is if we also want the check to run in release build, at which point maybe it would be better to use if statements and throwing exceptions. What do you think?

It would be good for us to have a clear consistent go-to or rules to guide when to use one or the other if we want to use both in our codebase.

gathogojr: @habbes I think we can go with `Debug.Assert` for now. I’m having trouble visualizing what a payload would look like if it contained a property that could be interpreted as an `ODataResourceValue`. Given how `ODataResourceValue` is typically structured — as a kind of nested resource -  it seems likely that, during deserialization, it would be treated as such. That’s why I’m quite skeptical that we’d ever encounter an `ODataResourceValue` directly as the value of an `ODataProperty`.

PR #3268, opened 2025-06-12
Fix CSDL typename validation

### LINKED ISSUE #2894: EdmLib successfully parses type names with unbalanced parens
TypeNames with unbalance parens, for example:

Type="Collection(ns.myType"
or
Type="ns.myType)"

parse just fine.

### Assemblies affected

Microsoft.OData.Edm 7.x

### Reproduce steps
Save the following xml as "badschema.csdl":
```xml
<?xml version="1.0" encoding="utf-8" ?>
    <edmx:DataServices>
        <Schema Namespace="namespace" Alias="ns" xmlns="http://docs.oasis-open.org/odata/ns/edm">
			<ComplexType Name="myPropType"/>
			<ComplexType Name="myType">
				<Property Name="prop1" Type="ns.myType)" />
				<Property Name="prop2" Type="Collection(ns.myType" />
			</ComplexType>
		</Schema>
    </edmx:DataServices>
</edmx:Edmx>
```

run the following:
```c#
static void Main(string[] args)
{
    IEdmModel model;
    IEnumerable<EdmError> errors;
    using (var stream = File.Open("badschema.csdl", FileMode.Open))
    {
        using (XmlReader reader = XmlReader.Create(stream))
        {
            var success = CsdlReader.TryParse(reader, out model, out errors);
            if (success & model.Validate(out errors))
            {
                var myType = model.SchemaElements.FirstOrDefault(e => e.Name == "myType") as IEdmStructuredType;
                if (myType != null)
                {
                    Console.WriteLine("Found type 'myType'");
                    foreach (var prop in myType.StructuralProperties())
                    {
                        Console.WriteLine("Property Name={0}, Type={1}", prop.Name, prop.Type.FullName());
                    }
                }
            }
            else
            {
                WriteErrors(errors);
            }
        }
    }
}
```
##

### Expected result

Error reading the property type names.

### Actual result

Model is valid.
Found type 'myType'
Property Name=prop1, Type=namespace.myType
Property Name=prop2, Type=Collection(namespace.myType)

### Additional detail

"bug" is in EdmXmlDocumentParser.ValidateTypeName. The algorithm is:
1) split on '(',')'
2) If first segment is "Collection", then return second segment as typename, otherwise return the first segment as typename. 
 
So, in either case, the extra paren is parsed away, so validate succeeds.
 
Next, we try and parse the type reference, using the same logic -- first split on '{','}', then return first or second segment depending on whether first segment is "collection". So, again, extra parens are lost and the parse succeeds with the desired (if not expected) value.

!!probably can't fix this in a minor release as it could break existing schemas that had this particular typo and are working correctly today!!

gathogojr:
### Issues

*This pull request fixes #2894.*

### Description

This pull request improves the validation of type names in CSDL parsing to ensure that only valid type names are accepted. The changes add comprehensive checks for invalid type name formats and ensure that clear, consistent error messages are produced for each invalid case.

**Details:**
- Enhanced type name validation logic to reject malformed or unsupported type names, such as:
  - Unmatched or misplaced parentheses (e.g., `ns.myType)`, `Collection(ns.myType`, `(ns.myType)`, `(ns.myType`)
  - Unsupported type prefixes (e.g., `Enumerable(ns.myType)`)
  - Invalid suffixes or extra characters (e.g., `Collection(ns.myType)s`, `Collection)ns.myType(`)
  - Empty or whitespace-only type names
  - Invalid use of Collection or Ref (e.g., `Collection()`, `Ref`, `Ref()`)
- Updated test coverage to assert that each invalid type name produces the expected error message: `"The type name '{typeName}' is invalid. The type name must be that of a primitive type, a fully qualified name or an inline 'Collection' or 'Ref' type."`

**Benefits:**
- Prevents invalid CSDL documents from being parsed without error.
- Improves error reporting for developers authoring or consuming CSDL.

- [x] *Test cases added*

### Additional work necessary

### Repository notes

### DIFF
--- a/src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
+++ b/src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
@@ -336,7 +336,7 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Common
             XmlAttributeInfo attr = GetOptionalAttribute(this.currentElement, attributeName);
             if (!attr.IsMissing)
             {
-                return this.ValidateTypeName(attr.Value);
+                return this.CheckAndReportTypeNameValidity(attr.Value);
             }
 
             return null;
@@ -347,7 +347,7 @@ namespace Microsoft.OData.Edm.Csdl.Parsing.Common
             XmlAttributeInfo attr = this.GetRequiredAttribute(this.currentElement, attributeName);
             if (!attr.IsMissing)
             {
-                return this.ValidateTypeName(attr.Value);
+                return this.CheckAndReportTypeNameValidity(attr.Value);
             }
 
             return null;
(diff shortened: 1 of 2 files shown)


src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
```diff
@@ -407,40 +407,64 @@ protected string ValidateEnumMembersPath(string path)
 
         private string ValidateTypeName(string name)
```
xuzhg: The syntax of method and the method name looks mismatch. If it's only to validate the type name, the return type 'string' doesn't make sense?

src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
```diff
@@ -407,40 +407,64 @@ protected string ValidateEnumMembersPath(string path)
 
         private string ValidateTypeName(string name)
         {
-            string[] typeInformation = name.Split(new char[] { '(', ')' });
-            string typeName = typeInformation[0];
+            Debug.Assert(name != null, "name != null");
```
xuzhg, line 410: If it's validation logic, so 'null' is invalid and should throw exception, right?

src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
```diff
@@ -407,40 +407,64 @@ protected string ValidateEnumMembersPath(string path)
...
         private string ValidateTypeName(string name)
         {
-            string[] typeInformation = name.Split(new char[] { '(', ')' });
-            string typeName = typeInformation[0];
+            Debug.Assert(name != null, "name != null");
+            ReadOnlySpan<char> nameSpan = name.AsSpan();
 
-            // For inline types, we need to check that the name contained inside is a valid type name
-            switch (typeName)
+            int indexOfOpenParen = nameSpan.IndexOf('(');
```
xuzhg, line 413: how about the nested parenthesis?

src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
```diff
@@ -407,40 +407,64 @@ protected string ValidateEnumMembersPath(string path)
...
 
-                    break;
+            if (typeNameSpan.IsEmpty)
+            {
+                typeNameSpan = nameSpan; // If no parentheses, the whole name is the type name
             }
 
-            if (EdmUtil.IsQualifiedName(typeName) || Microsoft.OData.Edm.EdmCoreModel.Instance.GetPrimitiveTypeKind(typeName) != EdmPrimitiveTypeKind.None)
+            // 'Collection' on its own is a valid type string
+            if (namePrefixSpan.IsEmpty && typeNameSpan.SequenceEqual(collectionAsSpan))
```
xuzhg, line 460: ReadOnlySpan<char> has its own 'Equals' overload. Should use that to compare?

src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
gathogojr: @xuzhg Can you share an example? I can add a test for that

src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
gathogojr: @xuzhg This is a **private** helper method that's invoked during the CSDL parsing process. I had difficulty envisioning a scenario where `name` could be `null` at this point, which is why I opted for a `Debug.Assert` rather than a runtime check. My thinking was that if `name` is ever `null`, it likely indicates a deeper issue that should be addressed earlier in the call chain, rather than handled defensively here. That said, I’m open to suggestions if you think a more robust check is warranted at this level.

src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
gathogojr: @xuzhg I'm not sure I understand... This private method existed with that signature. Are you suggesting that I should rename it?

src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
xuzhg: Yes, 1) if no side effects, it's better to rename it to refresh the better functionalities of this method.
2) Since it's private, if we make sense the caller checks the nullability of the name, it's ok to do a defensive using Debug.Assert

src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
xuzhg: "collection(ns.Type(anything)))
"((ns.Type)"
"ns.()()"

src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
gathogojr: @xuzhg I could rename it `CheckAndReportTypeNameValidity`. I think that's intuitive enough.
However, I'd want you to note that all other methods that do something similar in the class are named `Validate*`. I'll leave them as is...
image

gathogojr: @xuzhg Why `Debug.Assert` should suffice for now is because from the two places the method is called, the logic looks as follows:
```csharp
XmlAttributeInfo attr = this.Get[Required|Optional]Attribute(this.currentElement, attributeName);
if (!attr.IsMissing)
{
    return this.CheckAndReportTypeNameValidity(attr.Value);
}
```
We'll not call this method when type name is missing from the element.

src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
gathogojr: @xuzhg You use `Equals` on a `ReadOnlySpan<char>` when you want to compare the content of the span to another span or string using a specific string comparison rule (e.g., ordinal, case-insensitive, culture-aware, etc.)
```csharp
public static bool Equals(this ReadOnlySpan<char> span, ReadOnlySpan<char> other, StringComparison comparisonType);
```
Otherwise the advice is to prefer `SequenceEqual` for fast, exact (ordinal, case-sensitive) binary comparison - https://learn.microsoft.com/en-us/dotnet/api/system.memoryextensions.sequenceequal?view=net-9.0
```csharp
public static bool SequenceEqual<T>(this ReadOnlySpan<T> span, ReadOnlySpan<T> other) where T : IEquatable<T>;
```

test/UnitTests/Microsoft.OData.Edm.Tests/Csdl/CsdlReaderTests.cs
```diff
@@ -2535,6 +2536,80 @@ public void ShouldReportCorrectLineNumbersWithMultiLineElementsAndMultipleSchema
...
+                "<Property Name=\"prop4\" Type=\"(ns.myType\" />" +
+                "<Property Name=\"prop5\" Type=\"Enumerable(ns.myType)\" />" +
+                "<Property Name=\"prop6\" Type=\"Collection(ns.myType)s\" />" +
+                "<Property Name=\"prop7\" Type=\"Collection)ns.myType(\" />" +
+                "<Property Name=\"prop8\" Type=\"Collection()\" />" +
+                "<Property Name=\"prop9\" Type=\"\" />" +
+                "<Property Name=\"prop10\" Type=\" \" />" +
+                "<Property Name=\"prop11\" Type=\"Ref\" />" + // Ref is not valid by itself
+                "<Property Name=\"prop12\" Type=\"Ref()\" />" +
+                "<Property Name=\"prop13\" Type=\"Collection\" />" + // TODO: Verify - Collection is valid by itself
```
WanjohiSammy, line 2560: Will this be valid even without `()`

src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
```diff
@@ -405,42 +405,82 @@ protected string ValidateEnumMembersPath(string path)
...
+                {
+                    this.ReportError(this.currentElement.Location, EdmErrorCode.InvalidTypeName, Error.Format(SRResources.CsdlParser_InvalidTypeName, name));
+                    return name;
+                }
+            }
+
+            string typeName = typeNameSpan.ToString();
 
-            if (EdmUtil.IsQualifiedName(typeName) || Microsoft.OData.Edm.EdmCoreModel.Instance.GetPrimitiveTypeKind(typeName) != EdmPrimitiveTypeKind.None)
+            if (EdmUtil.IsQualifiedName(typeName) || EdmCoreModel.Instance.GetPrimitiveTypeKind(typeName) != EdmPrimitiveTypeKind.None)
```
WanjohiSammy, line 483: `EdmUtil.IsQualifiedName` and `EdmCoreModel.Instance.GetPrimitiveTypeKind` does not currectly accepts `ReadOnlySpan<char>

habbes (approved):

src/Microsoft.OData.Edm/Csdl/Parsing/Common/EdmXmlDocumentParser.cs
gathogojr: @WanjohiSammy No they don't. They have more callers that just this method so I chose to leave as is for now

PR #3274, opened 2025-06-23
Deprecate DataServiceContext KeyComparisonGeneratesFilterQuery

### LINKED ISSUE #2208: Remove DataServiceContext.KeyComparisonGeneratesFilterQuery in the next breaking change release.

WanjohiSammy:
### Issues

*This pull request fixes #2208.*

### Description

Remove `DataServiceContext.KeyComparisonGeneratesFilterQuery` flag. This flag was added so as not to break existing clients. 

In `ODL 8`, `DataServiceContext.KeyComparisonGeneratesFilterQuery` flag is default to true, meaning that we generate a query with `$filter` expression. 

The `KeyComparisonGeneratesFilterQuery` property is marked as `obsolete`. This PR is to remove this flag entirely, requiring users to use `ByKey` for key segment queries.

- [X] *Test cases added*

### Additional work necessary

### Repository notes

### DIFF
--- a/src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
+++ b/src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
@@ -249,51 +249,12 @@ namespace Microsoft.OData.Client
 
                 // Earlier path components must be navigation sources, and navigation sources cannot have query options.
                 Debug.Assert(!target.HasQueryOptions, "Navigation source had query options?");
-
-                target.SetKeyPredicate(keyPredicates);
             }
 
             if (inputPredicates != null)
             {
-                List<Expression> keyPredicates = null;
-                List<Expression> nonKeyPredicates = null;
-
-                // Get the current key predicates
-                List<Expression> currentPredicates = input.KeyPredicateConjuncts.ToList();
-
-                if (input.Filter != null && input.Filter.PredicateConjuncts.Count > 0)
-                {
-                    // Get the filter predicates
-                    currentPredicates = currentPredicates.Union(input.Filter.PredicateConjuncts.Union(inputPredicates)).ToList();
-                }
-                else
-                {
-                    currentPredicates = currentPredicates.Union(inputPredicates).ToList();
-                }
-
-                if (!input.UseFilterAsPredicate && !context.KeyComparisonGeneratesFilterQuery)
-                {
-                    keyPredicates = ExtractKeyPredicate(input, currentPredicates, model, out nonKeyPredicates);
-                }
-
-                if (keyPredicates != null)
-                {
-                    input.SetKeyPredicate(keyPredicates);
-                    input.RemoveFilterExpression();
-                }
-
-                // A key predicate cannot be applied if query options other than 'Expand' are present,
-                // so merge the key predicate into the filter query option instead.
-                if (keyPredicates != null && input.HasSequenceQueryOptions)
-                {
-                    input.ConvertKeyToFilterExpression();
-                }
-
-                if (keyPredicates == null)
-                {
-                    input.ConvertKeyToFilterExpression();
-                    input.AddFilter(inputPredicates);
-                }
+                input.ConvertKeyToFilterExpression();
+                input.AddFilter(inputPredicates);
             }
 
             return input; // No need to adjust this.currentResource - filters are merged in all cases
--- a/test/UnitTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
+++ b/test/UnitTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
@@ -19,9 +19,9 @@ namespace Microsoft.OData.Client.Tests.ALinq
     /// </summary>
     public class SequenceMethodsTests
     {
-        private const string MockCustomer1 = "{\"CustomerID\":\"ALFKI\",\"CompanyName\":\"Alfreds Futterkiste\",\"ContactName\":\"Maria Anders\",\"Address\":\"Obere Str. 57\",\"City\":\"Berlin\"}";
+        private const string MockCustomer1 = "{\"CustomerID\":\"ALFKI\",\"CustomerTypeID\":\"Regular\",\"CompanyName\":\"Alfreds Futterkiste\",\"ContactName\":\"Maria Anders\",\"Address\":\"Obere Str. 57\",\"City\":\"Berlin\"}";
 
-        private const string MockCustomer2 = "{\"CustomerID\":\"CHOPS\",\"CompanyName\":\"Chop-suey Chinese\",\"ContactName\":\"Yang Wang\",\"Address\":\"Hauptstr. 29\",\"City\":\"Bern\"}";
+        private const string MockCustomer2 = "{\"CustomerID\":\"CHOPS\",\"CustomerTypeID\":\"VIP\",\"ContactName\":\"Yang Wang\",\"Address\":\"Hauptstr. 29\",\"City\":\"Bern\"}";
 
         private readonly DataServiceQuery<Customer> _customers;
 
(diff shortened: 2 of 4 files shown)


test/UnitTests/Microsoft.OData.Client.Tests/DataServiceContextTests.cs
```diff
@@ -1,14 +1,164 @@
...
+            // Arrange & Act
+            var context = new DataServiceContext(new Uri("http://service/"));
+            context.UrlKeyDelimiter = DataServiceUrlKeyDelimiter.Slash;
+
+            // Assert
+            Assert.Equal(DataServiceUrlKeyDelimiter.Slash, context.UrlKeyDelimiter);
+        }
+
+        [Fact]
+        public void Constructor_InitializesProperties()
```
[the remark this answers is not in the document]
WanjohiSammy: Resolved

test/UnitTests/Microsoft.OData.Client.Tests/ALinq/SequenceMethodsTests.cs
```diff
@@ -291,10 +369,20 @@ private static EdmModel BuildEdmModel()
         {
             var model = new EdmModel();
 
+            // Define customer type enum
+            var enumType = new EdmEnumType("NS", "CustomerType");
+            enumType.AddMember("None", new EdmEnumMemberValue(0));
+            enumType.AddMember("Regular", new EdmEnumMemberValue(0));
+            enumType.AddMember("Premium", new EdmEnumMemberValue(1));
```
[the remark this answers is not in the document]
WanjohiSammy: Resolved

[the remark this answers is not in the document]
WanjohiSammy: Resolved

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -224,8 +221,7 @@ private static Expression AnalyzePredicate(MethodCallExpression mce, ClientEdmMo
             }
 
             conjuncts = null;
-            List<Expression> inputPredicates;
-            if (predicatesByTarget.TryGetValue(input, out inputPredicates))
+            if (predicatesByTarget.TryGetValue(input, out List<Expression> inputPredicates))
```
gathogojr, line 224: @WanjohiSammy I can understand the `else` before where we assign a `null` to `inputPredicates` to prevent code further below from complaining about the use of un-initialized variable, but in your case, wouldn't `inputPredicates` be null if `TryGetValue` returns false?

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -238,8 +234,7 @@ private static Expression AnalyzePredicate(MethodCallExpression mce, ClientEdmMo
             {
                 QueryableResourceExpression target = predicates.Key;
                 List<Expression> clauses = predicates.Value;
-                List<Expression> nonKeyPredicates;
-                List<Expression> keyPredicates = ExtractKeyPredicate(target, clauses, model, out nonKeyPredicates);
+                List<Expression> keyPredicates = ExtractKeyPredicate(target, clauses, model, out List<Expression> nonKeyPredicates);
```
gathogojr, line 237: We removed the `input` expression from the `predicatesByTarget` collection several lines above here. For the `input` expression, we got rid of the `input.SetKeyPredicate` statement - because we don't want a `/EntitySet({key})` expression generated for a `Where` expression.
Why are we retaining it here for targets that are not `input`? This `foreach` loop remains the only place where `ExtractKeyPredicate` is called, and I don't believe we need to extract the key predicate.

I'm however still unsure in what scenarios we would end up here so I'm still trying to understand what changes need to go here. The typical scenario that I'm familiar with is one that looks as follows:
```csharp
var results = context.Customers.Where(d => d.CreditAmount > 1300 || d.Balance == 0)
```
Here we only have a single target which is the `input` - `Customers`, with two predicates.
I think you first need to think about which scenarios would cause the `foreach` block targeting targets that are not `input` to be executed, and then think about how to refactor the `foreach`. It could be that it needs to disappear altogether...

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
```diff
@@ -255,45 +250,8 @@ private static Expression AnalyzePredicate(MethodCallExpression mce, ClientEdmMo
...
-                {
-                    input.ConvertKeyToFilterExpression();
-                }
-
-                if (keyPredicates == null)
-                {
-                    input.ConvertKeyToFilterExpression();
-                    input.AddFilter(inputPredicates);
-                }
+                input.ConvertKeyToFilterExpression();
```
gathogojr, line 253: @WanjohiSammy I think you need to investigate whether there's any scenario where we'd need to convert key to filter expression. A `ByKey` cannot be used together with a `Where` as far as I know

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
WanjohiSammy: If `TryGetValue` returns false, `inputPredicates` will be null. 
```cs
 List<Expression> targetPredicates = null;
 if (!predicatesByTarget.TryGetValue(boundTarget, out targetPredicates)) { }
```

can be simplified using [Inline variable declaration](https://learn.microsoft.com/en-us/dotnet/fundamentals/code-analysis/style-rules/ide0018) to the following:
```cs
if (!predicatesByTarget.TryGetValue(boundTarget, out List<Expression> targetPredicates)) { }
```

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
WanjohiSammy: Given the following LINQ query:
```cs
var query = (
  from product in contextWrapper.Products.OfType<DiscontinuedProduct>()
  from related in product.RelatedProducts.OfType<DiscontinuedProduct>()
  from photo in related.Photos
  where
        product.ProductId == -9 &&
        related.ProductId == -9 &&
        photo.PhotoId == -4 &&
        photo.ProductId == -4

   select photo) as IQueryable<ProductPhoto>;
```

Once you remove the `input` expression from the `predicatesByTarget` collection, you will remain with 2 `QueryableResourceExpression` i.e., `Products.OfType<DiscontinuedProduct>()` and  `product.RelatedProducts.OfType<DiscontinuedProduct>()`

image

If we do not set keyPredicates for these inputs, the below exception is thrown:
```
System.NotSupportedException : Individual properties can only be selected from a single resource or 
as part of a type. Specify a key predicate to restrict the entity set to a single instance or 
project the property into a named or anonymous type.
```

src/Microsoft.OData.Client/ALinq/ResourceBinder.cs
WanjohiSammy: Yes, you are right. I couldn't not find a scenario where key is converted to a Filter expression. 

If the `input` does not have key predicates, `keyPredicateConjuncts` is empty, so `ConvertKeyToFilterExpression()` does nothing. It is safe to remove it as well.

habbes:
I remember changing the default behaviour in ODL 8 caused a few surprises, which we documented in the release notes: https://learn.microsoft.com/en-us/odata/odatalib/release-notes/odatalib-8#changed-key-lookup-in-where-clause-to-use-filter-by-default

Just wanted to point out that we should remember to document the change in behaviour in release notes, especial handling of navigation property paths.

PR #3278, opened 2025-06-24
Support parameter aliases in key segments

### LINKED ISSUE #3277: Parameter Aliases not supported in Keys
## Describe the bug

The following syntax should be supported in the ODataUriParser but is not:

`Person(id=@id)?@id=7`

## Assemblies affected

Microsoft.OData.Core 8.x

## Steps to Reproduce

Using the HardCodedTestModel.TestModel in the OData.Core unit tests, attempt to execute the following:

```c#
            ODataPath pathSegment = new ODataUriParser(HardCodedTestModel.TestModel, new Uri("http://host"), new Uri("http://host/People(ID = @id)")).ParsePath();
```

## Expected behaviour

The path should be parsed, with a key segment containing a keyvalue pair with the key being the string "ID" and the value being a FunctionParameterAlias with the alias value "@id".

## Actual behaviour

An exception is thrown

mikepizzo:
### Issues

*This pull request fixes #3277.*

### Description

Adds support for aliases in key segments, i.e.,:

`People(id=@id)`

Also supports aliases in alternate keys.

- [x] *Test cases added*

### Additional work necessary

None

### DIFF
--- a/src/Microsoft.OData.Core/UriParser/Parsers/SegmentArgumentParser.cs
+++ b/src/Microsoft.OData.Core/UriParser/Parsers/SegmentArgumentParser.cs
@@ -249,6 +249,14 @@ namespace Microsoft.OData.UriParser
                 return false;
             }
 
+            if (valueText[0] == '@')
+            {
+                FunctionParameterAliasToken alias = new FunctionParameterAliasToken(valueText);
+                alias.ExpectedParameterType = typeReference;
+                convertedValue = alias;
+                return true;
+            }
+
             IEdmPrimitiveTypeReference primitiveType = typeReference.AsPrimitive();
             Type primitiveClrType = EdmLibraryExtensions.GetPrimitiveClrType((IEdmPrimitiveType)primitiveType.Definition, primitiveType.IsNullable);
             LiteralParser literalParser = LiteralParser.ForKeys(this.keysAsSegment);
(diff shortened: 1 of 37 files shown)


src/Microsoft.OData.Core/UriParser/Parsers/SegmentArgumentParser.cs
```diff
@@ -249,6 +249,14 @@ private bool TryConvertValue(IEdmTypeReference typeReference, string valueText,
                 return false;
             }
 
+            if (valueText[0] == '@')
```
gathogojr: @mikepizzo Could `valueText` ever be an empty string? If so, this might result in an exception being thrown. Would it be worth adding a safeguard or clarification?

src/Microsoft.OData.Core/UriParser/Parsers/SegmentArgumentParser.cs
```diff
@@ -275,7 +283,7 @@ private static bool TryParseFromUri(string text, out SegmentArgumentParser insta
             List<string> positionalValues = null;
 
             // parse keys just like function parameters
-            ExpressionLexer lexer = new ExpressionLexer(string.Concat("(", text, ")"), true, false);
+            ExpressionLexer lexer = new ExpressionLexer(string.Concat("(", text, ")"), true, false, true);
```
gathogojr: @mikepizzo We should use named parameters here - especially for the boolean arguments - to improve readability and make the intent of each parameter clearer:
```
ExpressionLexer lexer = new ExpressionLexer(
    string.Concat("(", text, ")"),
    moveToFirstToken: true,
    useSemicolonDelimiter: false,
    parsingFunctionParameters: true);
```

src/Microsoft.OData.Core/UriParser/Parsers/SegmentArgumentParser.cs
mikepizzo: Good call. It will never be null in our implementation, but the method that calls it is virtual so in theory someone could override the method and pass in a null value. I've added the suggested check.

src/Microsoft.OData.Core/UriParser/Parsers/SegmentArgumentParser.cs
mikepizzo: Yes, I thought about at least commenting the arguments with the names (I'm not a fan of passing just boolean values as parameters) but thought that was probably a bigger issue to fix than just the one parameter I was adding. But, even if we don't fix it everywhere, everywhere we do fix is an improvement, so I took the suggested fix.

PR #3283, opened 2025-07-02
Resolve issues with Deserialization of Untyped Types

### LINKED ISSUE #2657: It doesn't make sense to resolve untyped integer value to Edm.Decimal

### Assemblies affected

### Reproduce steps

Supposed I have the dynamic property or (untyped property) value as below:

```json
{
    ....
    "Data": 42
}
```

### Expected result

*The property value of 'Data' should be a "System.Int32".

### Actual result

*The property value of 'Data' is a "System.Decimal"

### Additional detail

* It's weird, because '42' looks like an integer, not a decimal. 

Here's the codes:
https://github.com/OData/odata.net/blob/master/src/Microsoft.OData.Core/JsonLight/ODataJsonLightPropertyAndValueDeserializer.cs#L265-L276

image

In my opinion, we should add extra 'else if' to identify other 'common' types.

WanjohiSammy:
### Issues

*This pull request fixes*
- #2657

### Description

The change ensures that untyped values of various primitive types are correctly deserialized and materialized, supporting a broader range of types.

The previous implementation only handled `bool`, `string`, and defaulted all other types to `decimal`. This was insufficient for untyped properties that could be of unresolved types such as `int, long, double, and collections`.

This change introduces a new `boolean` flag `ReadUntypedNumericAsDecimal` to enhance control over how untyped numeric values are handled during JSON parsing.
- **Default behavior**: `False` by default. Numeric values are inferred and converted to specific types (int, long, decimal, double) based on their runtime type.
- **Alternative behavior**: When set, untyped numeric values are preserved as decimal to maintain precision. Can be set as below:
   ```cs
    var readerSettings = new ODataMessageReaderSettings()
    readerSettings.LibraryCompatibility |= ODataLibraryCompatibility.ReadUntypedNumericAsDecimal; // set the flag to read untyped numerical values as Decimal
   ```

- [ ] *Test cases added*

### Additional work necessary

### Repository notes


src/Microsoft.OData.Core/Json/ODataJsonResourceDeserializer.cs
```diff
@@ -3824,7 +3825,8 @@ await this.JsonReader.GetValueAsync().ConfigureAwait(false),
                 payloadTypeReference,
                 this.MessageReaderSettings.PrimitiveTypeResolver,
                 this.MessageReaderSettings.ReadUntypedAsString,
-                !this.MessageReaderSettings.ThrowIfTypeConflictsWithMetadata);
```
xuzhg, line 3827: If it's for 9.x, do we need to keep back compatible?

test/EndToEndTests/Common/Microsoft.OData.E2E.TestCommon/Common/Server/UntypedTypes/UntypedTypesEdmModel.cs
```diff
@@ -0,0 +1,88 @@
...
+
+namespace Microsoft.OData.E2E.TestCommon.Common.Server.UntypedTypes;
+
+public static class UntypedTypesEdmModel
+{
+    private static readonly Uri _baseUri = new("http://localhost/odata/");
+
+    public static IEdmModel GetEdmModel()
+    {
+        Func<Uri, XmlReader?> getReferencedSchemaFunc = uri =>
```
xuzhg: Is it necessary since your CSDL file is simpler?

xuzhg (dismissed):

gathogojr:
@WanjohiSammy I’d like to highlight two specific comments from the thread on the issue you have referenced that are particularly relevant to this PR:

- https://github.com/OData/odata.net/issues/2657#issuecomment-1548547407  
- https://github.com/OData/odata.net/issues/2657#issuecomment-1949023315 — this one reflects the discussion we had during the initial scoping of the work.

Regarding backward compatibility, we typically manage such changes using the `LibraryCompatibility` enum. You can see how this is structured here:  
https://github.com/OData/odata.net/blob/main/src/Microsoft.OData.Core/ODataLibraryCompatibility.cs

To align with this approach, I suggest introducing a new enum member, e.g., `ReadUntypedNumericAsDecimal`, with a value like `1 << 5`. This flag can then be included in the `Version6` and `Version7` enum members, and you can define `Version8 = ReadUntypedNumericAsDecimal`. Toggling the flag will preserve the old behaviour in 6.x, 7.x, and 8.x.

Since `ODataMessageReaderSettings` already exposes a `LibraryCompatibility` property, this approach wouldn’t require introducing a new setting.

Finally, please refer to how compatibility flags are evaluated and toggled across the library to ensure consistency with existing patterns.

src/Microsoft.OData.Core/Json/ODataJsonPropertyAndValueDeserializer.cs
```diff
@@ -217,7 +218,8 @@ internal static IEdmTypeReference ResolveUntypedType(
                 IEdmTypeReference payloadTypeReference,
                 Func<object, string, IEdmTypeReference> primitiveTypeResolver,
                 bool readUntypedAsString,
-                bool generateTypeIfMissing)
+                bool generateTypeIfMissing,
+                bool readUntypedNumericAsDecimal = true)
```
gathogojr: Given that this method is internal, and we have updated the call points to pass the relevant value, do we need to set a default parameter value?

test/EndToEndTests/Common/Microsoft.OData.E2E.TestCommon/Common/Server/UntypedTypes/UntypedTypesEdmModel.cs
WanjohiSammy: Updated

src/Microsoft.OData.Core/Json/ODataJsonPropertyAndValueDeserializer.cs
WanjohiSammy: Resolved

src/Microsoft.OData.Core/Json/ODataJsonPropertyAndValueDeserializer.cs
```diff
@@ -257,23 +259,34 @@ internal static IEdmTypeReference ResolveUntypedType(
...
+                    }
+                    // This is for backward compatibility with untyped numeric values.
+                    else if (readUntypedNumericAsDecimal)
+                    {
+                        typeReference = EdmCoreModel.Instance.GetDecimal(isNullable: true);
                     }
                     else
                     {
-                        typeReference = EdmCoreModel.Instance.GetDecimal(/*isNullable*/ true);
+                        typeReference = jsonReaderValue switch
```
xuzhg, line 283: These are the other (all) potential types?

xuzhg (approved):
Looks good to me

PR #3318, opened 2025-07-10
Add CustomQueryOptionNode

### LINKED ISSUE #3317: ODataUriParser does not include custom query options
## Describe the bug

When parsing a Uri that includes custom query options, those custom query options are missing from the parsed ODataUri.

## Assemblies affected

Microsoft.OData.Core 8.x

## Steps to Reproduce

~~~c#
        public static void ValidatesQueryOptions(IEdmModel model)
        {
            string request = String.Format("company?customQueryOption=value);
            Uri uri = new Uri(request, UriKind.Relative);
            ODataUriParser parser = new ODataUriParser(model, uri);
            ODataUri odataUri = parser.ParseUri();

            Assert.NotEmpty(odataUri.CustomQueryOptions);
~~~

## Expected behaviour

Code should succeed. odataUri.CustomQueryOptions should have a single entry "customQueryOption" with the value "value".

## Actual behaviour

Assert fails -- odataUri.CustomQueryOptions is empty.

## Additional details

There are multiple issues in the code.

First, ODataUri.CustomQueryOptions is typed as a List<QueryNode>, but QueryNode is abstract and there is no derived type for a custom query option.

Second, ODataUriParser.ParseUri hardcodes the custom query options it passes to the ODataUri constructor as a new empty List<QueryNode>.

Third, ODataUriParser.CustomQueryOptions is typed as List<KeyValuePair<string,string>>, which disagrees with the type of the ODataUri.CustomQueryOptions.

mikepizzo:
### Issues

*This pull request fixes #3317.*

### Description

Adds CustomQueryOptionNode
Populates ODataUri.CustomQueryOptions
Add tests for parsing/validating/visiting

- [x] *Test cases added*

### Repository notes

### DIFF
--- a/azure-pipelines-nightly.yml
+++ b/azure-pipelines-nightly.yml
@@ -37,7 +37,7 @@ extends:
   parameters:
     pool:
       name: MSSecurity-1ES-Build-Agents-Pool
-      image: MSSecurity-1ES-Windows-2019
+      image: MSSecurity-1ES-Windows-2022
       os: windows
     customBuildTags:
     - ES365AIMigrationTooling
--- a/azure-pipelines-rolling.yml
+++ b/azure-pipelines-rolling.yml
@@ -39,7 +39,7 @@ extends:
   parameters:
     pool:
       name: MSSecurity-1ES-Build-Agents-Pool
-      image: MSSecurity-1ES-Windows-2019
+      image: MSSecurity-1ES-Windows-2022
       os: windows
     customBuildTags:
     - ES365AIMigrationTooling


src/Microsoft.OData.Core/UriParser/SemanticAst/CustomQueryOptionNode.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            ExceptionUtils.CheckArgumentStringNotNullOrEmpty(name, "name");
+            ExceptionUtils.CheckArgumentStringNotNullOrEmpty(value, "value");
+            this.optionName = name;
+            this.optionValue = value;
+        }
+
+        /// Gets the Query Option Name.
+        /// </summary>
+        public string Name
```
xuzhg: Since 'public override QueryNodeKind Kind => (QueryNodeKind)this.InternalKind;' 
maybe we can set all other properties using '=>'.

public string Name => this.optionName;

src/Microsoft.OData.Core/UriParser/SemanticAst/CustomQueryOptionNode.cs
```diff
@@ -0,0 +1,96 @@
...
+{
+    #region Namespaces
+
+    using Microsoft.OData.Edm;
+    #endregion Namespaces
+
+    /// <summary>
+    /// Base class for all custom query options.
+    /// </summary>
+    public class CustomQueryOptionNode : QueryNode
```
xuzhg, line 17: It seems it's allowed to have query options like:  /api/values?ids=1&ids=2&ids=3
Then, 'ids' is a collection of [1,2,3]. Does CustomerQueryOptionNode handle this?

test/UnitTests/Microsoft.OData.Core.Tests/UrlValidation/ODataUrlValidatorTests.cs
```diff
@@ -153,6 +153,55 @@ public static void CallingOnUriSucceeds()
             });
         }
 
+        [Fact]
+        public static void ValidatesQueryOptions()
```
xuzhg: This test method name could be more descriptive.

test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/QueryOptionBuilderTests.cs
```diff
@@ -0,0 +1,30 @@
...
+        [Fact]
+        public void CustomQueryOptionWorks()
+        {
+            Uri queryUri = new Uri("People?customQueryOption=customValue", UriKind.Relative);
+            Uri actualUri = UriBuilder(queryUri, ODataUrlKeyDelimiter.Parentheses, settings);
+            Assert.Equal(new Uri("http://gobbledygook/People?customQueryOption=customValue"), actualUri);
+        }
+
+        [Fact]
+        public void MultipleCustomQueryOptionsWork()
```
xuzhg: ...Work ==> ...Works?

test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/QueryOptionBuilderTests.cs
```diff
@@ -0,0 +1,30 @@
...
+
+using System;
+using Xunit;
+
+namespace Microsoft.OData.Tests.ScenarioTests.UriBuilder
+{
+    public class QueryOptionBuilderTests : UriBuilderTestBase
+    {
+        [Fact]
+        public void CustomQueryOptionWorks()
```
xuzhg, line 15: shall we have failing test case?
for example, only contain partial, `customerQueryOption=`

src/Microsoft.OData.Core/UriParser/ODataUriParser.cs
```diff
@@ -607,6 +603,26 @@ private void InitQueryOptionDic()
             }
         }
 
+        /// <summary>
```
xuzhg, line 602: Do you think we should add a public API in the ODataUriParser to get the custom query option nodes?
Same as 'ParseFilter()'?

xuzhg (approved):

src/Microsoft.OData.Core/UriParser/ODataUriParser.cs
```diff
@@ -607,6 +603,26 @@ private void InitQueryOptionDic()
             }
         }
 
+        /// <summary>
+        /// Generate a list of <see cref="CustomQueryOptionNode"/> from the custom query options.
+        /// </summary>
+        /// <param name="customQueryOptions"></param>
+        /// <returns></returns>
```
WanjohiSammy: Missing doc string values for `<param />` and `<returns />`

src/Microsoft.OData.Core/UriParser/ODataUriParser.cs
```diff
@@ -607,6 +603,26 @@ private void InitQueryOptionDic()
...
         }
 
+        /// <summary>
+        /// Generate a list of <see cref="CustomQueryOptionNode"/> from the custom query options.
+        /// </summary>
+        /// <param name="customQueryOptions"></param>
+        /// <returns></returns>
+        private static List<CustomQueryOptionNode> CreateCustomQueryOptionNodes(IList<KeyValuePair<string, string>> customQueryOptions)
+        {
+            var customQueryOptionNodes = new List<CustomQueryOptionNode>();
```
WanjohiSammy: In ODL, we use explicit type instead of `var` unless it is a test

```suggestion
            List<CustomQueryOptionNode> customQueryOptionNodes = new List<CustomQueryOptionNode>();

```

src/Microsoft.OData.Core/UriParser/SemanticAst/CustomQueryOptionNode.cs
mikepizzo: Fixed for this class. I did not change the other node classes, on which the original code was based.

src/Microsoft.OData.Core/UriParser/SemanticAst/CustomQueryOptionNode.cs
mikepizzo: Actually, the HTTP spec doesn't specify what happens when you have duplicates -- it could be turned into an array, retained as duplicates, the first, last, or some random value used, or the server could return an error.

The current code simply preserves the duplicates, which seems the safest way to handle.  Added a test to verify this behavior.

test/UnitTests/Microsoft.OData.Core.Tests/UrlValidation/ODataUrlValidatorTests.cs
mikepizzo: updated to `ValidatorValidatesCustomQueryOptions`, which is really what it tests (that the rule is called if a custom query option exists -- this is what did not happen previously.)

test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/QueryOptionBuilderTests.cs
mikepizzo: Fixed.

test/UnitTests/Microsoft.OData.Core.Tests/ScenarioTests/UriBuilder/QueryOptionBuilderTests.cs
mikepizzo: The HTTP spec is pretty flexible in what can be in a query string -- it doesn't have to be a key/value pair, and can have a key but no value, or just a value with no key.  Also, for backward compatibility, I didn't want to introduce any errors when parsing a URL that previously ignored custom query options. I added tests to verify that any of the above are valid and round-trip (the same URL is created as provided). The only change to previous behavior was that query options without a name (i.e., just "foo") were previously ignored, and now are retained.

src/Microsoft.OData.Core/UriParser/ODataUriParser.cs
mikepizzo: I thought of that, but we already have a CustomQueryOptions public property which returns the key/value pairs (which I would expect is the same as what ParseCustomQueryOption() would do.)

src/Microsoft.OData.Core/UriParser/ODataUriParser.cs
mikepizzo: Fixed.

src/Microsoft.OData.Core/UriParser/ODataUriParser.cs
mikepizzo: You are exactly right -- thanks for catching. I've been jumping back and forth between GS (which uses var) and ODL and applied the wrong convention.

